PR #3, opened 2015-06-26
Fixing Hunt Naming Issues and Updating Hunt Compass / Height and Decay / Zoning

sqt:
There's still some weirdness on zoning (quick change of some kind, perhaps related to player position) but over all seems to work OK.


sqt:
OK, the aliasing thing seems to work too.  I think the pull will grab it too now?

PR #4, opened 2015-07-08
Relatively stable focus-based & hotkey-toggled visibility.

sqt:
Slight modification to how the hotkey toggle worked in order to combine approaches and use a single handler (now both called via Dispatcher).  Excessive calls to browserWindow.Show() were bad news for the focus-checking side of things, so there are status flags and explicit checks for when the state needs to change before calling Show().  Hide() doesn't seem to have the same issue?


quisquous:
This seems to work in most cases, but if I do the following:
(1) focus the game
(2) hit ctrl+esc to bring up the start menu
(3) click on something open other than the game/ACT on the task bar (?)
(4) click on the game window

This seems to make it disappear when it shouldn't.  If I then alt+tab to ACT, I can get it back? It's a little odd, though.

ACTPlugin.cs
```diff
@@ -48,14 +54,52 @@ public void InitPlugin(TabPage pluginScreenSpace, Label pluginStatusText)
...
-                    browserWindow.Show();
-            };
+            {
+                overlayEnabled = !overlayEnabled;
+                currentDispatcher.Invoke(visCheck);
+            };
+            System.Windows.Automation.AutomationFocusChangedEventHandler afceh = (sender, o) =>
+            {
+                System.Windows.Automation.AutomationElement focusedElement = sender as System.Windows.Automation.AutomationElement;
+                if (focusedElement != null)
```
quisquous: Code review nitpicking me suggests early return instead of giant indented block?

ACTPlugin.cs
```diff
@@ -48,14 +54,52 @@ public void InitPlugin(TabPage pluginScreenSpace, Label pluginStatusText)
...
+                {
+                    int processId = focusedElement.Current.ProcessId;
+                    using (System.Diagnostics.Process process = System.Diagnostics.Process.GetProcessById(processId))
```
quisquous: Also, do you really need this using an indented block if you only use |process| once?

ACTPlugin.cs
sqt: Single return point habits kicking in.  I have no strong preference here.

ACTPlugin.cs
sqt: Not knowing what "using" was for, it looked like other pieces of code were using it to save space on null checks, but after looking it up it looks like it's secretly wrapping this with a try/catch block as well, which I don't really care about.  I'll pull this out into another check / early return.

PR #6, opened 2017-06-18
Make hint paths more generic

danakj:
This is where people should be able to find things in the OverlayPlugin if building CactbotOverlay, plus making the FFXIV plugin default to being inside ThirdParty instead of going outside the source dir.


CactbotOverlay/resources/cactbot_test.html
```diff
@@ -1,73 +1,48 @@
...
-            if (typeof data.logs != "undefined" && data.logs.length > 0) {
-                document.getElementById("oneLog").innerText = "log: " + data.logs[0];
-            }
-        }
-
-    </script>
+  <meta charset="utf-8" />
+
+  <!-- OverlayPlugin widget resize handle -->
+	<link rel="stylesheet" href="resize_handle.css">
```
quisquous: Mixed spaces and tabs ;;

PR #7, opened 2017-08-21
bar refactoring / warrior nonsense

quisquous:


CactbotOverlay/resources/dana/bars/bars.js
```diff
@@ -78,22 +74,131 @@ var g_init = false;
...
+    // Due to this bug: https://github.com/ravahn/FFXIV_ACT_Plugin/issues/100
+    // We can not look for log messages from FFXIV "You use X" here. Instead we
+    // look for the actual ability usage provided by the XIV plugin.
+    // Also, the networked parse info is given much quicker than the lines from the game.
+    if (this.startList.indexOf(skillList[0]) == -1) {
+      this.startList.push(skillList[0]);
+    }
+    for (var i = 0; i < skillList.length; ++i) {
+      var node = this.comboNodes[skillList[i]];
+      if (node == undefined) {
```
danakj: FYI These 2 lines I think are usually written like

if (!(skillList[i] in this.comboNodes)) {

but since ur adding it anyway I guess it doesn't matter.

CactbotOverlay/resources/dana/bars/bars.js
```diff
@@ -78,22 +74,131 @@ var g_init = false;
...
+  }
+
+  StateTransition(nextState) {
+    if (this.current == nextState)
+      return;
+
+    window.clearTimeout(this.comboTimer);
+    this.comboTimer = null;
+    this.current = nextState;
+    this.comboTimer = null;
```
danakj: Just did this on L170?

CactbotOverlay/resources/dana/bars/bars.js
```diff
@@ -78,22 +74,131 @@ var g_init = false;
...
+    }
+    if (log.search(this.kReEndCombo) >= 0) {
+      this.AbortCombo();
+      return true;
+    }
+    return false;
+  }
+
+  StateTransition(nextState) {
+    if (this.current == nextState)
```
danakj: I think this early out is wrong. If you start your combo, then start your combo, the timer should be reset?

CactbotOverlay/resources/dana/bars/bars.js
```diff
@@ -78,22 +74,131 @@ var g_init = false;
...
+    this.current = nextState;
+    this.comboTimer = null;
+
+    if (nextState == null) {
+      this.considerNext = [];
+    } else {
+      this.considerNext = this.comboNodes[nextState].next;
+      if (!this.comboNodes[nextState].last) {
+        var kComboDelayMs = 12000;
+        this.comboTimer = window.setTimeout(this.AbortCombo.bind(this), kComboDelayMs);
```
danakj: Woa. TIL about bind(). I've been making a local var that = this that gets bound into a local closure. This is much simpler :)

CactbotOverlay/resources/dana/bars/bars.js
```diff
@@ -78,22 +74,131 @@ var g_init = false;
...
+  }
 
+  AbortCombo() {
+    this.StateTransition(null);
+  }
+}
+
+function setupComboTracker(me, callback) {
+  var comboTracker = new ComboTracker(me, kComboBreakers, callback);
+  comboTracker.AddCombo(['(Enchanted )?Riposte', '(Enchanted )?Zwerchhau', '(Enchanted )?Redoublement']);
```
danakj: The RDM combo has to be Enchanted, and includes Verflare/Verholy at the end.

I didn't try to track the last step in the combo before, I treated it as breaking the combo as the outcome is the same. But if you're going to track the last bit then it should be there.

Though TBH I think in the end you're not using the .last field. So you could make these combos all skip the last one if you wanted, for warrior too. (But still require Enchanted.)

Then this made me wonder about tracking the last one with a timeout.. are your war combos 4 skills long and you omitted the last one here? If not then making a timeout on the actual last thing (like Verflare/holy for rdm) doesn't make sense to me, as the combo is already over. So maybe you should condition that on .last.

I guess there's 2 options:
1. The AddCombo omits the last thing. Make sure it's in the combo breakers set. Use timer on the last thing.
2. The AddCombo includes the last thing. Don't use a timer on it. Reset the combo to empty when it's run.

Right now I think this is a mix of the two?

CactbotOverlay/resources/dana/bars/bars.js
```diff
@@ -172,9 +165,12 @@ class ComboTracker {
     this.comboTimer = null;
 
     if (nextState == null) {
-      this.considerNext = [];
+      this.considerNext = this.startList;
     } else {
-      this.considerNext = this.comboNodes[nextState].next;
+      this.considerNext = [];
+      Array.prototype.push.apply(this.considerNext, this.comboNodes[nextState].next);
+      Array.prototype.push.apply(this.considerNext, this.startList);
```
danakj: It might make sense to clear the combos on job change and have OnJobChange() do AddCombo() for everything for the given job.

CactbotOverlay/resources/dana/bars/bars.js
quisquous: Yeah, I think I know it's not ever a real entry that is undefined, so I was just lazy to avoid looking up twice.

CactbotOverlay/resources/dana/bars/bars.js
quisquous: OOPS

CactbotOverlay/resources/dana/bars/bars.js
quisquous: Oh, nice catch thx thx.

PR #18, opened 2017-12-03
Update royal_city_of_rabanastre.js

Vuzrak:
Updated triggers for mask.


quisquous:
Ohmygosh, thanks! I had thought originally that the mask ability only occurred after the divine commandment debuff wore off, but double checking the logs that's not true at all.

As a side question, have you had any lag problems with triggers in Rabanastre with cactbot, especially on Hashmal or Argath?

Vuzrak:
No issues here with lag during Rabanastre, I do have TTS off so not sure if that would make a difference.

PR #23, opened 2017-12-26
New locale: 'ko'

hibiyasleep:
also:
* move language files into `resources/lang`.
* added *status/action ID* on comments.


ui/jobs/jobs.html
```diff
@@ -12,7 +12,7 @@
   <script src="../../resources/timericon.js"></script>
   <script src="../../resources/widgetlist.js"></script>
   <script src="../../resources/lang.js"></script>
-  <script src="../../resources/lang-en.js"></script>
+  <script src="../../resources/lang/en.js"></script>
```
quisquous: Can you include lang/ko.js here as well?

resources/lang/ko.js
```diff
@@ -0,0 +1,165 @@
...
+    this.losesEffectRegex = function(effect, target, attacker) {
+      this.ValidateEffect(effect);
+      if (!target)
+        target = '[^:]*';
+      if (!attacker)
+        attacker = '[^:]*';
+      return Regexes.Parse(' 1E:' + target + ' loses the effect of ' + effect + ' from ' + attacker + '.*\.');
+    };
+
+    this.countdownStartRegex = function() {
```
quisquous: Are these the same in Korean? Could you add a "TODO: not translated yet" if not?

resources/lang/ko.js
```diff
@@ -0,0 +1,165 @@
...
+    this.kUIStrings = Object.freeze({
+      // jobs: text on the pull countdown.
+      Pull: 'Pull',
+    });
+
+    // Due to this bug: https://github.com/ravahn/FFXIV_ACT_Plugin/issues/100
+    // We can not look for log messages from FFXIV "You use X" here. Instead we
+    // look for the actual ability usage provided by the XIV plugin.
+    // Also, the networked parse info is given much quicker than the lines from the game.
+    this.youUseAbilityRegex = function() {
```
quisquous: I didn't look line by line, but at first glance most of these regex functions look unchanged from the en version.  I didn't know at the time whether or not other languages would have the same log messages.  Could you move them into the base CactbotLang class so both en and ko can share and not duplicate them?

resources/lang/en.js
```diff
@@ -54,27 +54,27 @@ class CactbotLanguageEn extends CactbotLanguage {
...
-      BattleLitany: 'Battle Litany',
-      Embolden: 'Embolden',
-      Balance: 'The Balance',
-      Hypercharge: 'Hypercharge',
-      LeftEye: 'Left Eye',
-      RightEye: 'Right Eye',
-      Brotherhood: 'Brotherhood',
-      Devotion: 'Devotion',
-      FoeRequiem: 'Foe Requiem',
+      BluntResistDown: 'Blunt Resistance Down', // 0x23d, 0x335, 0x3a3
```
quisquous: I wasn't sure at first what these hex codes meant, but I looked through logs and I think I understand more of the log format.  I am very curious about this!

It seems like pairs of `:.*0F:(effectId):` designate the effect id.  Where did all of these values come from? Did you investigate these yourself or is there a list somewhere?

Some of these values, I see in logs, but others I do not.  For example, Foe Requiem up I see `[15:57:43.750] 15:1067BEA6:Azel Nox:73:Foe Requiem:1067BEA6:Azel Nox:90030F:8B:` but I do not see any :8C: for Foe Requiem lost.  Similarly, Left Eye and Right Eye are not abilities but only effects, and so I don't know where the values 0x4a0 and 0x49f come from.

Can you help me understand more? I'd love to know.  :)

resources/lang/en.js
```diff
@@ -7,42 +7,42 @@ class CactbotLanguageEn extends CactbotLanguage {
...
-      ButchersBlock: "Butcher's Block",
-      Maim: 'Maim',
-      StormsEye: "Storm's Eye",
-      StormsPath: "Storm's Path",
-      TrickAttack: 'Trick Attack',
-      Embolden: 'Embolden',
-      Aetherflow: 'Aetherflow',
-      ChainStrategem: 'Chain Strategem',
-      Hypercharge: 'Hypercharge',
+      DragonKick: 'Dragon Kick', // 0x4a
```
quisquous: Were these id comments useful for you when translating? I feel like these ids are all already inside of lang.js and so duplicating them here is a bit noisy.  I think I'd prefer to not have them if it was just me, but if they were useful in translating, then I'm happy to leave them.

resources/lang/en.js
hibiyasleep: This is internal effect ID and used very widely. also skills, zones has unique ID like this.

`0x8C` is [for enemies](http://xivdb.com/status/140). (ps: XIVDB uses the decimal ID to build its URL)

resources/lang/en.js
hibiyasleep: Translators can reference this ID to query XIVDB (`0x8c` -> `/status/140`), or extracted FFXIV data itself.

resources/lang/ko.js
hibiyasleep: Seems only 'gains/loses' and few things are in English, so this doesn't need to translate.

resources/lang/ko.js
```diff
@@ -0,0 +1,165 @@
...
+      this.ValidateEffect(effect);
+      if (!target)
+        target = '[^:]*';
+      if (!attacker)
+        attacker = '[^:]*';
+      return Regexes.Parse(' 1E:' + target + ' loses the effect of ' + effect + ' from ' + attacker + '.*\.');
+    };
+
+    this.countdownStartRegex = function() {
+      return Regexes.Parse(/Battle commencing in (\y{Float}) seconds!/);
```
hibiyasleep: this is TODO, I didn't saw countdown recently.

resources/lang/ko.js
hibiyasleep: can `Regex`es be a method? Since we (yes, Korean) using English version of the plugin, so I can't predict how this should behave on ja/de/fr versions, to always override or not.

ui/jobs/jobs.html
hibiyasleep: Yep.

resources/lang/en.js
hibiyasleep: Oh, it was duplicate, I didn't saw them. I'll remove these comments.

resources/lang/en.js
hibiyasleep: Don't miss out that we can access *unparsed log* like saved to a .log file, like [OverlayPlugin and SpecialSpellTimer do](https://github.com/hibiyasleep/OverlayPlugin/blob/master/OverlayPlugin.Core/Overlays/LogParseOverlay.cs#L29).

quisquous: Oh, I see! Thanks for the explanation.  By not seeing `0x8C`, I meant that I didn't see it anywhere in the log, applied to either players or enemies.

I'll have to investigate the unparsed log, thanks for the pointer!

hibiyasleep:
I (accidentally) rebased onto `upstream/master`, there shouldn't be a conflict.

resources/lang.js
```diff
@@ -109,69 +109,4 @@ class CactbotLanguage {
     Object.freeze(this.kAbilNameToId);
     Object.freeze(this.kAbilIdToName);
   }
-
-  // Due to this bug: https://github.com/ravahn/FFXIV_ACT_Plugin/issues/100
-  // We can not look for log messages from FFXIV "You use X" here. Instead we
-  // look for the actual ability usage provided by the XIV plugin.
-  // Also, the networked parse info is given much quicker than the lines from the game.
-  youUseAbilityRegex() {
```
quisquous: Can you leave these in CactbotLanguage but remove them from CactbotLanguageEn and CactbotLanguageKo so that they are not duplicated? I think youStartUsingRegex, youGainEffectRegex, youLoseEffectRegex, abilityRegex, gainsEffectRegex, and losesEffectRegex should all be in CactbotLanguage if they are shared.  Sorry if I haven't communicated this clearly enough--that was why I moved these functions from CactbotLanguageEn to CactbotLanguage in master.

The countdown regex functions (since they'll be different) can be in CactbotLanguageEn and CactbotLanguageKo.

quisquous:
Thanks for your patience with all the back and forth! I really appreciate all the work.  :)

resources/lang.js
hibiyasleep: This is my merging fault, sorry ;( - cherry-picked and amended on [04501a1](https://github.com/hibiyasleep/cactbot/commit/04501a1216b510e02dc45dbc730648a27f212ea5).

ui/jobs/jobs.js
```diff
@@ -360,8 +360,9 @@ function setupBuffTracker() {
       sortKey: 9,
     },
     requiem: {
-      gainRegex: gLang.gainsEffectRegex(gLang.kEffect.FoeRequiem, '(\\y{Name})', '\\1'),
-      loseRegex: gLang.losesEffectRegex(gLang.kEffect.FoeRequiem, '(\\y{Name})', '\\1'),
+      // TODO: Can't use \y{Name} because https://github.com/quisquous/cactbot/issues/22.
```
quisquous: I think this is accidental too.  Can you revert these three lines?

quisquous:
Thanks! Let me know what I can do to help debug the message issue you're seeing.  That really seems like the next issue to getting this working for you.

PR #24, opened 2018-01-09
Jobs support for DRK, PLD

CptDraco:
Admittedly I am new at this, so I could use some help cleaning these up. Thought you might like to include them in the master.  Love this plugin. Thanks.
  
  
drkbars
pldbars


ui/jobs/jobs.js
```diff
@@ -28,7 +29,7 @@ var Options = {
   BigBuffBorderSize: 1,
 
   FarThresholdOffence: 24,
-
+  LowMPThreshold: 4800,
```
danakj: This option seems very job specific, so probably should have Drk in its name?

ui/jobs/jobs.js
```diff
@@ -1199,17 +1338,22 @@ class Bars {
...
     var far = -1;
+	
     if (this.job == 'RDM' || this.job == 'BLM' || this.job == 'SMN' || this.job == 'ACN')
       far = this.options.FarThresholdOffence;
-
-    if (far >= 0 && this.distance > far)
+	else if (this.job == 'DRK')
+	  lowmp = this.options.LowMPThreshold;
+  
+    if (far >= 0 && this.distance > this.options.FarThresholdOffence)
```
danakj: I think comparing to |far| is still right here

ui/jobs/jobs.js
```diff
@@ -1199,17 +1338,22 @@ class Bars {
...
       this.o.manaBar.fg = computeBackgroundColorFrom(this.o.manaBar, 'mp-color.far');
+	else if (lowmp >= 0 && this.mp <= this.options.LowMPThreshold)
```
danakj: and this could compare to |lowMP|

ui/jobs/jobs.js
```diff
@@ -1199,17 +1338,22 @@ class Bars {
     if (!this.o.manaBar) return;
     this.o.manaBar.value = this.mp;
     this.o.manaBar.maxvalue = this.maxMP;
-
+	var lowmp = -1;
```
danakj: "lowMP" ?

user/jobs-example.js
```diff
@@ -36,7 +38,8 @@ Options.BigBuffBorderSize = 1
 
 // The distance that offensive spells such as Verareo, Bio, etc are castable.
 Options.FarThresholdOffence = 24
-
+// When MP falls below this, the MP bar is highlighted with the .low CSS class.
```
danakj: This shouild mention its for DRK I think

ui/jobs/jobs.js
CptDraco: I had thought to use it elsewhere, BLM for example, but that would have different values anyway. So I agree, just don't know what to change it to.

danakj: I think just sticking a "Drk" at the front would work.

ui/jobs/jobs.js
```diff
@@ -19,6 +19,7 @@ var Options = {
 
   RdmCastTime: 1.94 + 0.5,
   WarGcd: 2.38,
+  PldGcd: 2.43,
```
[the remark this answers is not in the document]
CptDraco: Hmm. Can you use the in game command that displays recast to show GCD? 
/e <recast.HeavySwing>

Have a trigger read it out of the log. Just spitballing here. Or can you get the secondary stat totals from the character panel somehow? We have the SKS statistics intervals available to set the ranges for each GCD. (i.e. 782-848sks = 2.43sec)
If I think of anything else I'll let you know.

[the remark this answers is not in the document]
quisquous: I did find the memory locations for secondary stat totals, but the primary problem there is that it doesn't include things like arrow or selene or huton.  The secondary problem is that the gcd formulas are only well known at level 70, and I think there's different offsets at different levels.

I think I just need to go back to the cheat engine mines and go digging for the actual cast times somewhere.

Hardcoded is good enough for now.

ui/jobs/jobs.js
```diff
@@ -1178,15 +1286,46 @@ class Bars {
...
+      } else if (skill == gLang.kAbility.SavageBlade) {
+        minSkillsUntilgore = 4;
+      } else if (skill == gLang.kAbility.RiotBlade) {
+        minSkillsUntilgore = 1;
+      } else {
+        // End of combo, or broken combo.
+        minSkillsUntilgore = 3;
+      }
+
+      // The new threshold is "can I finish the current combo and still
```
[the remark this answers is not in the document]
CptDraco: That actually depends on if it falls in your Fight or Flight window really. The goal is to refresh Goring Blade right before FoF expires. Even then 100% uptime is still just as important. Believe it or not Goring Blade and Sword Oath Autoattacks make up about 25-30% of a Paladin's DPS. 

*If FoF is up, which should be used immediately before Goring Blade, then it is definitely better to refresh early.* Note:  some people do it as much as 1gcd before when using a faster build.

resources/lang.js
```diff
@@ -48,6 +48,19 @@ class CactbotLanguage {
       RabbitMedium: '8E0',
       OneIlmPunch: '48',
       Bootshine: '35',
+	    FastBlade: '09',
```
quisquous: Can you fix all these tabs, here and in other files?

ui/jobs/jobs.js
```diff
@@ -214,6 +243,17 @@ function setupRegexes() {
...
     gLang.kAbility.StormsEye,
     gLang.kAbility.StormsPath,
+	// pld
+    gLang.kAbility.ShieldLob,
+    gLang.kAbility.TotalEclipse,
+    gLang.kAbility.SavageBlade,
+    gLang.kAbility.RageofHalone,
+    gLang.kAbility.RiotBlade,
+    gLang.kAbility.RoyalAuthority,
+	gLang.kAbility.GoringBlade,
```
quisquous: This file too.  ;)

quisquous:
Thanks for all the work!  I might followup with a few things and make some opinionated css changes.  Let me know if I've mucked with anything that you feel strongly about and we can discuss.  :)

CptDraco:
I was working on the same thing at the same time, but yours is much better. Thanks for this. I was making a default positioning that would be more suitable to use with the default FFXIV HUD layout, but seperating the jobs so that your mnk, war etc defaults were back how you had them. The default has the target bar in a very different place, way up at the top of the screen, but nobody runs it that way so setting these up your way is more practical. Personally I set mine to frame center of the screen more. Keeps important information   in easy view while I am watching my positioning. Everyone is going to have different preferences for UI layout. 
What are your plans for more user friendly configurations later on? Since Cactbot runs through OverlayPlugin, we don't have the option of adding settings directly into ACT for each module. What about an HTML configuration screen for the jobs module much like Hibyasleep made for the kagerou overlay? As new jobs get added that .css file is going to get pretty crowded. Would be nice to have a simple GUI to set those up, and even save multiple user configurations. Reminds me of doing World of Warcraft addons years ago.

quisquous:
Re: configuration.  Personally, I'm ok with javascript and CSS configuration.  I don't think it'll be thaaat big.  The larger problem is going to be folks who don't know how CSS works.  I kind of figured that it'd be like the dps overlay space, where one person figures out a cool way to make something look and then everybody else copies and pastes a configuration file without having to think about it.

I'm not opposed to some GUI tool to configure all of this, but it feels kind of like overkill until there are more users who aren't as css savvy.  I'm not even sure what I would put in such a configuration tool for everything.  Positions, sizes, color pickers, maybe? If it generated the GUI config automatically and spat out the CSS, I think that'd be plausible, but it'd be a lot of work.  It's not something I'm personally planning on working on in the short term.

PR #26, opened 2018-01-11
Fix new Opacity option

CptDraco:
Fix new Opacity option

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -1134,7 +1134,7 @@ class Bars {
     if (blood < 50) {
       this.o.bloodTextBox.classList.add('low');
       this.o.bloodTextBox.classList.remove('mid');
-    } else if (blood < 100) {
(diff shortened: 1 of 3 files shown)


ui/jobs/jobs.js
```diff
@@ -1152,7 +1152,7 @@ class Bars {
     if (oath < 50) {
       this.o.oathTextBox.classList.add('low');
       this.o.oathTextBox.classList.remove('mid');
-    } else if (oath < 100) {
+    } else if (oath < 90) {
```
quisquous: Did you mean to do this for pld? Or it seems like maybe you'd want to do this for both pld and drk?

CptDraco: My fault, was meant for DRK. It was too early and I am only a student!

quisquous:
Thanks a bunch! Sorry for the options typo! :)

PR #31, opened 2018-02-02
O6S Timeline

xephero:
Haven't actually tested it in combat yet (going in tomorrow), but seems to work outside that.

As a style note, even though most public timelines don't sync every line, I prefer to because xiv has some amount of [drift in its combat timings](https://www.reddit.com/r/ffxiv/comments/6qk2mv/timestamps_for_all_deltascape_savage_encounters/dkxzuqt/?context=1). So the time in between adjacent abilities ends up being more important than the time from the beginning of a fight/phase to something several minutes into it, since it could be way off by that point. By syncing every ability, it can make minor 100ms corrections throughout the fight.

I plan on making timelines like this for the other fights too (it's fun!) if none exist by the time I get around to it.

### DIFF
--- a/ui/raidboss/data/timelines/o6s.txt
+++ b/ui/raidboss/data/timelines/o6s.txt
@@ -1,5 +1,102 @@
 # Omega - Sigmascape V2.0 (Savage) - O6S
 
-0 "--Reset--" sync /Removing combatant TBD/ window 10000 jump 0
+hideall "--Reset--"
+hideall "--sync--"
 
-0 "Start" sync /:Engage!/ window 0,1
+0 "--Reset--" sync /(Removing combatant Demon Chadarnook|has initiated a ready check)/ window 10000 jump 0
+
+0 "Start" sync /00:0044:I have claimed the girl in the picture! She's mine! You can't have her!/ window 0,1
+18 "--targetable--"
+21 "Demonic Howl" sync /:Demon Chadarnook:282B:/ window 21,0
+31 "Demonic Shear" sync /:Demon Chadarnook:2829:/
+
+57 "Possession" sync /:Demon Chadarnook:2803:/ window 20,20
+62 "Flash Fire" sync /:Portrayal of Fire:280A:/
+71 "Flash Gale" sync /:Portrayal of Wind:2ABA:/
+73 "Demonic Howl" sync /:Demon Chadarnook:282B:/
+87 "Poltergeist" sync /:Goddess Chadarnook:2824:/
+92 "Demonic Shear" sync /:Demon Chadarnook:2829:/
+98 "Divine Lure" sync /:Goddess Chadarnook:2822:/
+100 "Demonic Howl" sync /:Demon Chadarnook:282B:/
+109 "Demonic Pain" sync /:Demon Chadarnook:2AEC:/
+121 "Demonic Howl" sync /:Demon Chadarnook:282B:/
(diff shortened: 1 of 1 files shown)


quisquous:
Timelines are fun.  :D

One thing you should do is add "Options.Debug = true" to cactbot/user/raidboss.js.  If you do that, then you get an onscreen timeline time to one decimal place.  I use that to sync timelines from video to better precision.  Also, if that option is true then any syncs (from time => sync time) are printed in the plugin log, so you can see what the sync time should be.

A few things that I've done to help with sync times is to always use network syncs instead of battle log line syncs.  Those are a *lot* more reliable.  I find that bosses tend to only slip around particular mechanics.  Like, ultimate coil loses +/- second around each trio, so I just sync the start of each trio.

However, I find that most fights don't need that many syncs, to be honest.  Susano is clockwork.  Other fights like o2s need more.  Mostly I prefer having as few syncs as possible.  One reason is that having few syncs mean that you're more likely to be able to plan cooldowns because the times are accurate.  It also makes retiming the fight from video easier because there's fewer sync points.  There is drift, but it's only particular mechanics that really drift, in my opinion.

quisquous:
Let me know how it goes when you test it and I'll go ahead and merge it after.  Thanks for all the help! :)

xephero:
Yeah, all of the syncs are from the network lines. Not only are the chat log lines not very reliable, but some people may not even have them enabled, which is why I had a chat message line to start it at the actual pull, and then the first network line ability with a window capable of starting it just in case they have those messages turned off. 

As for drifting, it only gets really out of control on very long fights without any pushes or phasing. Neo Exdeath is the fight used in the example I linked, where two different pulls had enrages 15+ seconds apart by the end of it, whereas with a bunch of phases you end up starting fresh every push so it doesn't have time to get too out of sync. That matters even using it for cooldown planning, because cooldown planning isn't going to work on that wide of a difference, either.

O6S is that kind of fight, one long timeline. From the looks of it, any O7S timeline is definitely going to be inviable for cooldown planning use, since there's a slightly random phase order that will necessitate a bunch of jumping, like [this timeline](https://drive.google.com/file/d/0B2MxPBM_A6ybUnI5Qnl6TFJ5dGs/view) from A7S back in the day

In any case, I'm not really a stickler about it; this and any future PRs from me are just contributions and you can change them if you'd like to suit your preference. This project is super cool, thanks for making it! I'll let you know how O6S testing goes.

quisquous:
Cool! Sorry, I wasn't trying to be grumpy at you.  I apparently just have a lot of opinions about timelines!  *<_<*

I wasn't sure what to do about o7s.  There's really only two branches, so I was almost thinking of just having the timeline totally duplicate after the first mechanic, so you could always know 100% what was coming next, almost like there were two totally different fights.

xephero:
Went well! https://www.twitch.tv/videos/225169356

Left side is cut off at the start cause I have to monitor capture that section but it's fixed like 90 seconds in, but you can still see the timers going. No debug timer, though. I make these by just walking through the log looking for enemy casts, rounding the time between them to the nearest integer, and then putting it that many seconds after the prior ability. i.e. if one thing is at 30.583 seconds, the next one is at 37.468 seconds, it'll just count that as +7 seconds. It's been pretty accurate in the past, and seems to have worked here too. O7S is going to be a beast though.

Again, this project rules, thanks so much!

quisquous:
Nice! This looks really good.  <3

PR #37, opened 2018-02-04
Timeline updates

xephero:
Went through and synced things. Luckily, it appears there's actual ability uses when the train becomes targetable and untargetable (28A8 and 28A7 respectively). Saintly Beam timing is still a little fuzzy, might just want to sync the actual hit instead of trying duration; I figured duration made sense for beam/light since your positioning matters for those throughout the window, but maybe not. Up to you, really.

I also added a script that takes a Network_XXXX.log file, start and end timestamps, and makes a rough timeline from them. It's not exactly robust, but it does save a lot of the busywork of transcribing from logs. They're still going to be handcrafted things, ultimately, but hopefully it finds some use as a starting point.

### DIFF
--- a/resources/fights.js
+++ b/resources/fights.js
@@ -59,7 +59,7 @@ var gBossFightTriggers = [
   {
     id: 'o5s',
     zoneRegex: /Sigmascape V1\.0 \(Savage\)/,
-    startRegex: /:Alte Roite uses Wyrm Tail/,
+    startRegex: /:Wroth Ghost:28B6:/,
     endRegex: /:Phantom Train was defeated by/,
     countdownStarts: true,
   },


xephero:
Added Byakko EX timeline, generated by the script and cleaned up a little. Tested all the way to enrage here: https://www.twitch.tv/videos/225851185

util/make_timeline.py
```diff
@@ -0,0 +1,113 @@
...
+        # At this point, we have a combat line for the timeline. Set up the fields we'll use
+        line_fields = line.split('|')
+        entry = {
+            'time': parse_line_time(line),
+            'combatant': line_fields[3],
+            'ability_id': line_fields[4],
+            'ability_name': line_fields[5],
+        }
+
+        # Ignore autos, probably need a better rule than this
```
quisquous: There's not a different rule, sadly.  Autos are just some random ability just like anything else.

xephero: There might be something like filtering the most used attack, or filtering the lowest ability id, something clever like that. I'll probably just add command line arguments to filter lines by combatant or name or id, so you can run it once, see what to ignore, and then run it with the filters to get a better result.

The Byakko timeline took about an hour with this, mostly spent simply removing extraneous lines like the state of shock sequence which has like eight actual ability uses instead of four. I didn't have to touch any of the timings though, even though I double checked them, making the process a lot easier and more feasible to do for normal modes or even dungeon bosses

xephero:
I see you made your own Byakko timeline which probably makes this all inconvenient in one big PR, let me know if you want the O5S/script commits in separate PRs from that.

quisquous:
Sorry for the conflict! I should have let you know that I was looking at adding a Byakko timeline myself so that we didn't collide with each other.

But yeah, I would love the o5s and script.  That script sounds super useful.  Thanks for all the contributions!

PR #39, opened 2018-02-05
O5S Timeline syncs

xephero:
More complete version of O5S, still untested until tomorrow's reset. From #37

### DIFF
--- a/ui/raidboss/data/timelines/o5s.txt
+++ b/ui/raidboss/data/timelines/o5s.txt
@@ -3,88 +3,96 @@
 hideall "--Reset--"
 hideall "--sync--"
 
-# Saintly Beam: 0 light appears, 5 target, 9 hit
+# Saintly Beam: 0 light appears, 5 target, 10 hit
 # Diabolic Light: 0 target appears, 8 target, 13 hit
 
 ##############################################################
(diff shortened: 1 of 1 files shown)


quisquous:
I'll give this a run tomorrow night too and then merge after that if it looks good.

Thanks a bunch! Appreciate the separate pull requests too.  It makes it a lot easier to work with.

xephero:
Just tested it, and it reset. The console has OnSync messages up to 203 (the Diabolic Wind in add phase), and after that no syncs happened. I'm not sure if it should print a message if the jump 0 errantly came up at any point. My log doesn't have any "Removing combatant Phantom Train" until the end though.

xDemonessx:
Same just happened to me, the timeline halted during midphase when ghosts were out.

xephero:
Added gigantic sync windows for the abilities that the boss only uses once in the whole fight. If it does end up stopping for some reason, it should resync once it sees those.

quisquous:
I think there's still some bugs to fix here, but this is an improvement over what was there before.

xephero:
I've been making these for myself and my statics since Gordias, I'm just happy to have a channel to get them out there more effectively. Thanks for your work!

PR #40, opened 2018-02-05
Timeline creation script

xephero:
Script to create timelines from a Network_###.log file, from #37


quisquous:
Also also, used this last night to quickly get ability ids out of a bunch of o8s pulls and I thiiiink I have enough ability ids to add callouts for fake vs true in vs out thunder/ice/fire abilities.  Super useful!

PR #42, opened 2018-02-05
Fix SSL/TLS issue when attempting to read manifest file from github pages

ezsoftware:
Fix allows the web.DownloadString to download the manifest.txt file when remotely hosting the overlay from github pages

### DIFF
--- a/CactbotOverlay/CactbotOverlay.cs
+++ b/CactbotOverlay/CactbotOverlay.cs
@@ -286,6 +286,7 @@ namespace Cactbot {
         notify_state_.sent_data_dir = true;
 
         var web = new System.Net.WebClient();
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for this! I'll give this a try during maintenance and then merge it.  Being able to use github pages for overlays was definitely the intention.

ezsoftware:
I updated this pull request to allow for a local pointer-style file similar to what I did to point to kagerou's overlay.  It adds two static variables to the object, a "realUrl" and "lastConfigUrl" and uses those to prevent expensive file reads and regex operations when unnecessary.  But this worked in testing with the normal file, the URL directly and an html file containing a meta refresh to my test github pages.

quisquous:
Thanks! Looks good to me.  oopsyraidsy doesn't appear to work remotely, possibly due to some race condition.  I probably need to wait for document load before adding ui more properly in the javascript.  I'll see if I can fix this on my end.

```
2/5/2018 10:11:35 PM: Info: testremote: BrowserConsole: Uncaught TypeError: Cannot read property 'children' of null (Source: https://ezsoftware.github.io/cactbot/ui/oopsyraidsy/oopsyraidsy.js, Line: 159)
```

Issue #33, opened 2018-02-03
Timeline stopping on O5S


sactage:
~90% of the time, hen I get sucked into a train car in O5S using raidboss, the timeline stops for the fight and doesn't come back up. Doesn't matter if it's the ghosts after add phase or the normal diabolic whistle ones.

quisquous:
Thanks for the report.  I'll take a look at this when I get back into o5s.

quisquous:
@xephero

ezsoftware:
Had the same issue last night as well, but only once.  Another issue I noticed was getting out of the passenger car too slowly could cause the 2nd half of the fight to be out of sync.  Our WAR 2-shotted his mob and was out so fast that the whole 2nd half of the fight was off by like 12 seconds for me.

quisquous:
Ahahaha.  Yeah, the o5s timeline doesn't have any syncs but the beginning.  It clearly needs something for when it comes back up.

The o5s timeline also restarts if it ever sees "Removing combatant Phantom Train".  I'm wondering if by going into the car, it removes some fake phantom train combatant for you which is triggering the timeline reset.  Can either of you look at your o5s logs in ACT (or re-import the log file from yesterday so you can see the encounter again), and see if that line appears anywhere other than on a wipe? Alternatively, can you look at the logs and let me know the hit points of the "real" phantom train?

e.g. for Byakko, there's a bunch of fake Byakkos and then one real one:
[22:54:50.020] 04:Removing combatant Byakko.  Max HP: 57250.
[22:54:50.020] 04:Removing combatant Byakko.  Max HP: 57250.
[22:54:50.020] 04:Removing combatant Byakko.  Max HP: 10997725.

Ikuni17:
I added this to line 42 of the O5S timeline, which helped sync the second phase most of the time:

330 "Diabolic Headlamp"	sync /uses Diabolic Headlamp/ window 30,20

For some reason there was one time it didn't sync correctly and I don't exactly remember why. I have also had the issue of the timeline stopping.

I have a log with a 1% enrage from last night, never cleared though, and "Removing combatant Phantom Train" only appeared at the wipe. There were 22 results, with one real:

[21:59:27.401] 04:Removing combatant Phantom Train.  Max HP: 12860067.

quisquous:
[...]

PR #43, opened 2018-02-07
O5N Timeline

xephero:
Added an O5N timeline since it's nice to have in general, but also it'll be *way* easier to test #33 without getting a static together.

### DIFF
--- a/ui/raidboss/data/manifest.txt
+++ b/ui/raidboss/data/manifest.txt
@@ -22,6 +22,7 @@ timelines/o1s.txt
 timelines/o2s.txt
 timelines/o3s.txt
 timelines/o4s.txt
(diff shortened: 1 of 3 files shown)


quisquous:
Seems reasonable! Can you add this to o5n.js in the same way that o5s.js references o6s.txt as a timeline?

If you've tested this yourself and it looks reasonable, I'm happy to merge it after that.

xephero:
Ah, I knew I forgot something. Yeah, I tested it and recorded it a few times to make sure it was fine. I never made it to the loop while I had the timeline up, but prior logs I have along with testing long-encounter o5n fflogs seems to prove that loop out to like 11 minutes.

PR #53, opened 2018-02-16
Adding timelines for Sigmascape 6-8N

xephero:
O6N:
Pretty standard. Just like savage, there's no real RNG, and a pretty early loop. One tricky part is that Demonic Wave is two casts; the boss has the long castbar, and then Portrayal of Water casts the real one 1 second after the cast bar finishes. Was always kind of confusing trying to pop out and in of melee range for it, and that explains why. No more, thanks to the timeline!

A nearly final version is here, the only thing I took out is the first (fake, boss castbar) Demonic Wave inside the loop: https://www.twitch.tv/videos/229684132

O7N:
There's no set order that I could detect. The only consistency is that a program will not be one of the last two loaded; that is, if the order goes Ultros -> Air Force, the next one will be either Dadaluma or Bibliotaph. If the third is Dadaluma, then the one after will either be Bibliotaph or Ultros. After the first two it always seems like a 50/50 shot, so in lieu of modelling out every sequential pair possible, I just have it empty after Load until the buff appears.

I encountered a few rare phases, usually in the later half of the fight, for Air Force and Dadaluma. I'm unsure if any alternate phases for Ultros or Bibliotaph exist, though. Either way, I opted to use universal sync windows instead of jumps just in case something unexpected happened, so it would pick up at the next Load. Having the fight be modular phases like that really helps for this purpose. You can see the rare phase happen in the video at 8:00, the Magitek Ray casting is the tell that Dadaluma isn't actually going to spawn.

A video of the final version of this timeline is here: https://www.twitch.tv/videos/229689480

O8N:
I'm not totally sure about the randomness of the first iterations of gazes and aero assault/revolting ruin before the loop, whether or not those are random. Same with the half arena hits. Still, there's triggers that call those out and it's not a huge deal to have some ambiguity there, especially in normal modes. I also opted to exclude the Flagrant Fire cast that simply causes the markers to show up, and instead only have the Flagrant Fire that is the damage actually going off. The stack/spread of those, too, might not be random.

A video of the final version of this timeline is here: https://www.twitch.tv/videos/229679385

### DIFF
--- a/ui/raidboss/data/timelines/o6n.txt
+++ b/ui/raidboss/data/timelines/o6n.txt
@@ -1,0 +1,68 @@
+# Omega - Sigmascape V2.0 - O6N
+
+hideall "--Reset--"
+hideall "--sync--"
+
+0 "--Reset--" sync /(Removing combatant Demon Chadarnook|has initiated a ready check)/ window 10000 jump 0
+
+0 "Start" sync /00:0044:I have claimed the girl in the picture! She's mine! You can't have her!/ window 0,1
+18 "--targetable--"
+21 "Demonic Howl" sync /:Demon Chadarnook:282C:/ window 21,2.5
+30 "Demonic Shear" sync /:Demon Chadarnook:282A:/
+
+54 "Possession" sync /:Demon Chadarnook:2803:/
+59 "Flash Fire" sync /:Portrayal of Fire:280B:/
+66 "Demonic Howl" sync /:Demon Chadarnook:282C:/
+76 "Demonic Shear" sync /:Demon Chadarnook:282A:/
+86 "Release" sync /:Demon Chadarnook:2804:/
+
+99 "Possession" sync /:Demon Chadarnook:2803:/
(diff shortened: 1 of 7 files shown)


ui/raidboss/data/timelines/o6n.txt
```diff
@@ -0,0 +1,68 @@
+# Omega - Sigmascape V2.0 - O6N
+
+hideall "--Reset--"
+hideall "--sync--"
+
+0 "--Reset--" sync /(Removing combatant Demon Chadarnook|has initiated a ready check)/ window 10000 jump 0
```
quisquous: What's with the "has initiated a ready check" log here? Do you find that the timelines aren't stopping when out of combat?

xephero: Nah, just habit from when I made them for the old ACT_Timeline without the combat detection stuff. The start block I mostly just copy/paste from old ones.

quisquous:
Thanks again for all the timelines! :)

PR #55, opened 2018-02-17
O8S (P1) naming cleanup

xephero:
Cleaned up some of the intro kefka namings. They're (intentionally, probably) confusing, but for reference, this is a quick glossary I compiled while doing this and normal mode:

* Gravitational Wave is the west half arena hit, and Intemperate Will is the east half.
* Ave Maria is the question mark gaze looking at the statue, Indolent Will is the gaze looking away.
* Inexorable Will are the ordinary targeted ground aoes.
* Indomitable Will is the four stack markers during first graven.
* These aren't in the timeline, but Indulgent Will is the confuse application, Idyllic Will is the sleep.

I also split it out into Graven sections, combined Blizzard+Thunder to make the actual event happening more clear from the timeline, and added the untargetable line for the enrage.


quisquous:
Looks great! Thanks for the cleanup!

PR #58, opened 2018-02-20
Create fr.js

Amiral-Benson:
New locale 'fr'

### DIFF
--- a/resources/lang/fr.js
+++ b/resources/lang/fr.js
@@ -1,0 +1,78 @@
+"use strict";
+
+class CactbotLanguageFr extends CactbotLanguage {
+  constructor(playerName) {
+    super('fr', playerName);
+  }
+
+  InitStrings(playerName) {
+    this.kZone = Object.freeze({
+      O1S: /Deltascape V1\.0 \(Savage\)/,
+      O2S: /Deltascape V2\.0 \(Savage\)/,
+      O3S: /Deltascape V3\.0 \(Savage\)/,
+      O4S: /Deltascape V4\.0 \(Savage\)/,
+      UCU: /The Unending Coil Of Bahamut \(Ultimate\)/,
+    });
+
+    this.kEffect = Object.freeze({
+      BluntResistDown: 'Résistance au contondant réduite', // 0x23d, 0x335, 0x3a3
+      VerstoneReady: 'VerTerre préparée', // 0x4d3
+      VerfireReady: 'VerFeu préparé', // 0x4d2
+      Impactful: 'Impact préparé', // 0x557
+      FurtherRuin: 'Extra Ruine renforcée', // 0x4bc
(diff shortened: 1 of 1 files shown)


Amiral-Benson:
I encountered an issue with French apostrophe character, it is used in the middle of sentences, buff, etc.

``      OpoOpoForm: "Posture de l'opo-opo", // 0x6b``
``      Aetherflow: "Flux d'éther", // 0x130``

It was breaking the code, I used English quotation marks instead, not sure if that was the right thing to do here.

I didn't replace yet the Seal and Unseal Regex, as in French it is said in the opposite way, and I'm not totally sure how to use the wild card .* there.

In English it is: XXX will be sealed off in
In French it is: Sealing of the XXX in

In English it is: XXX is no longer sealed
In French it is: Opening of XXX

quisquous:
Re: seal regexes, the (.*) is just a "match everything".  So you'd say something like "Sealing of the (.*) in" or "Opening of (.*)".

Re: zones.  Are the zone names the same for you in ACT?

Also, re: French version. Have you seen anything like this? https://www.reddit.com/r/ffxiv/comments/6n6o7h/susano_ex_triggers_for_act/dkjj8ja/.  Just curious what your log lines look like.

quisquous:
Also also, thanks for this! I'll try to take a look at this more closely when I have a chance later this week.

Amiral-Benson:
Thank you for your explanation, I updated the fr.js file with Seal and Unseal.

When ACT FFXIV plug-in is set up to use French language, all the zone names are still in English. Enemies, abilities, etc. are in French, but zone names are in English.


In the Custom Triggers tab, when we click on the Zone button, to insert in the input field of Category the name of the current zone, the zone name will be in English.

I checked your reddit link about the person having a bar separating their network lines, I don't have a bar but a colon:

``[23:43:23.304] 1A:Gardien gains the effect of Morsure De La Tempête from Amiral Benson for 30,00 Seconds.
[23:43:23.406] 15:10507038:Sao Somnus:07:Attaque:40001F4E:Gardien:710003:6530000:0:0:0:0:0:0:0:0:0:0:0:0:0:0:18710226:18731055:12000:12000:1000:1000:-0,1704383:-8,395513:0:43921:43921:5760:5760:1000:1000:1,64906:10,5628:-0,001173166:
[23:43:23.000] 00:12a9:Sao Somnus inflige au gardien 1619 points de dégâts.
``
Some things are in French, some in English.

quisquous:
Thanks thanks! Also, to answer your previous question about " vs ', definitely just use a " when there's an apostrophe in the string.

I'm going to go ahead and land this.  I think there's a tiny fix or two to make, but then after the the cactbot jobs overlay should work for you in French if you set `Options.Language ='fr'` in cactbot/user/jobs.js

PR #60, opened 2018-02-24
allow setting of local user config

ezsoftware:
local user config injection,

had to rush, daughter got a surprise promotion in kung fu today and I have to leave right now for that, but it seems to be working.

### DIFF
--- a/CactbotOverlay/CactbotOverlay.cs
+++ b/CactbotOverlay/CactbotOverlay.cs
@@ -296,7 +296,7 @@ namespace Cactbot {
           }
         }
 
-        var web = new System.Net.WebClient();
+				var web = new System.Net.WebClient();
         System.Net.ServicePointManager.SecurityProtocol = System.Net.SecurityProtocolType.Ssl3 | System.Net.SecurityProtocolType.Tls | System.Net.SecurityProtocolType.Tls11 | System.Net.SecurityProtocolType.Tls12;
 
         var data_file_paths = new List<string>();
@@ -676,7 +676,9 @@ namespace Cactbot {
         DispatchToJS(new JSEvents.SendSaveData(Config.OverlayData));
       } else if (obj.ContainsKey("setSaveData")) {
         Config.OverlayData = obj["setSaveData"];
-      }
+      } else if (obj.ContainsKey("getUserLocation")) {
+				DispatchToJS(new JSEvents.SendUserConfigLocation(Config.UserConfigFile));
+			}
     }
 
     // State that is tracked and sent to JS when it changes.
--- a/CactbotOverlay/CactbotOverlay.csproj
+++ b/CactbotOverlay/CactbotOverlay.csproj
@@ -51,26 +51,25 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Advanced Combat Tracker">
-      <HintPath>ThirdParty\ACT\Advanced Combat Tracker.exe</HintPath>
+      <HintPath>..\..\OverlayPlugin-Compiler\OverlayPlugin\OverlayPlugin\bin\x64\Release\Advanced Combat Tracker.exe</HintPath>
     </Reference>
     <Reference Include="FFXIV_ACT_Plugin">
-      <HintPath>ThirdParty\ACT\FFXIV_ACT_Plugin.dll</HintPath>
+      <HintPath>C:\Users\john_000\AppData\Roaming\Advanced Combat Tracker\Plugins\FFXIV_ACT_Plugin.dll</HintPath>
     </Reference>
-    <Reference Include="HtmlRenderer, Version=1.0.1.4, Culture=neutral, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>ThirdParty\OverlayPlugin\HtmlRenderer.dll</HintPath>
+    <Reference Include="HtmlRenderer">
+      <HintPath>..\..\OverlayPlugin-Compiler\OverlayPlugin\BuildX64\HtmlRenderer.dll</HintPath>
     </Reference>
     <Reference Include="Newtonsoft.Json">
       <HintPath>ThirdParty\OverlayPlugin\Newtonsoft.Json.dll</HintPath>
     </Reference>
     <Reference Include="OverlayPlugin">
-      <HintPath>ThirdParty\OverlayPlugin\OverlayPlugin.dll</HintPath>
+      <HintPath>..\..\OverlayPlugin-Compiler\OverlayPlugin\BuildX64\OverlayPlugin.dll</HintPath>
     </Reference>
     <Reference Include="OverlayPlugin.Common">
-      <HintPath>ThirdParty\OverlayPlugin\OverlayPlugin.Common.dll</HintPath>
+      <HintPath>..\..\OverlayPlugin-Compiler\OverlayPlugin\BuildX64\OverlayPlugin.Common.dll</HintPath>
     </Reference>
     <Reference Include="OverlayPlugin.Core">
-      <HintPath>ThirdParty\OverlayPlugin\OverlayPlugin.Core.dll</HintPath>
+      <HintPath>..\..\OverlayPlugin-Compiler\OverlayPlugin\BuildX64\OverlayPlugin.Core.dll</HintPath>
     </Reference>
     <Reference Include="System" />
     <Reference Include="System.Core" />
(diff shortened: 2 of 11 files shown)


CactbotOverlay/CactbotOverlay.cs
```diff
@@ -301,7 +301,7 @@ public CactbotOverlay(CactbotOverlayConfig config)
           }
         }
 
-        var web = new System.Net.WebClient();
+				var web = new System.Net.WebClient();
```
quisquous: Sorry, but can you convert all these tabs to spaces, here and elsewhere?

I really need to have a style checker commit hook, but I haven't gotten there yet.

CactbotOverlay/CactbotOverlay.csproj
```diff
@@ -51,26 +51,25 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Advanced Combat Tracker">
-      <HintPath>ThirdParty\ACT\Advanced Combat Tracker.exe</HintPath>
+      <HintPath>..\..\OverlayPlugin-Compiler\OverlayPlugin\OverlayPlugin\bin\x64\Release\Advanced Combat Tracker.exe</HintPath>
```
quisquous: These whole file seems like accidental changes.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -681,7 +681,9 @@ public CactbotOverlay(CactbotOverlayConfig config)
         DispatchToJS(new JSEvents.SendSaveData(Config.OverlayData));
       } else if (obj.ContainsKey("setSaveData")) {
         Config.OverlayData = obj["setSaveData"];
-      }
+      } else if (obj.ContainsKey("getUserLocation")) {
```
quisquous: I think this should be a message that is pushed every time.  Instead of making the JS ask for the user location, just send it every time via onSendUserConfigLocation, just like onDataFilesRead is automatically pushed from C# to JS.

The save data is a little bit different because there's a "set" part of it where the pull counter sends back the current pull counts.  So, in that case, it's required that the setSaveData be an overlay message type that the JS can call back into C# with.

If I had to add some brief tweaks here, I'd also ask for:
* remove the automatic loading of user/whatever.js and use this path for that as well, so that there's only one way that user config files get loaded and remote files aren't a special case
* user location if unspecified should be ../../user relative to html file, so that this all works as it did prior to this pull request

I will mention offhand here that I think JS loading order is a bit muddled at the moment.  Currently, C# messages can be sent before the plugin is loaded.  (It'd be nice to delay sending anything but it wasn't clear how.  I'm not keen on JS having to inform C# when it's loaded.  Surely this can be automatic, somehow?)  So, many objects are set up, and then some time later, they're attached to the document to update elements.  This change adds an additional layer of loading weirdness where now user config is going to be loaded (likely) after the JS objects are created, so there may need to be some reinitialization in some cases.  I don't think this is something that needs to get fixed in this pull request, but it's something to keep a look out for.

ui/raidboss/raidboss.js
```diff
@@ -40,4 +40,20 @@ gTimelineController = new TimelineController(Options, new TimelineUI(Options));
...
+OverlayPluginApi.overlayMessage(OverlayPluginApi.overlayName, JSON.stringify({'getUserLocation': true}));
+document.addEventListener("onSendUserConfigLocation", function(e) {
+  var jsLoc = (e.detail.location);
+  var cssLoc = jsLoc.replace('.js', '.css');
+  /*load CSS*/
+  var userCSS = document.createElement('link');
+  userCSS.setAttribute('rel', 'stylesheet');
+  userCSS.setAttribute('type', 'text/css');
+  userCSS.setAttribute('href', cssLoc);
+  document.getElementsByTagName('head')[0].appendChild(userCSS);
```
quisquous: I'm a really bad web developer.  Does this work prior to the document load event? I know other elements don't exist, but I am not sure about the head element.  Just curious if there is a race here.

ui/raidboss/raidboss.js
```diff
@@ -40,4 +40,20 @@ gTimelineController = new TimelineController(Options, new TimelineUI(Options));
 gPopupText = new PopupText(Options);
 // Connect the timelines to the popup text.
 gTimelineController.SetPopupTextInterface(new PopupTextGenerator(gPopupText));
-gPopupText.SetTimelineLoader(new TimelineLoader(gTimelineController));
\ No newline at end of file
+gPopupText.SetTimelineLoader(new TimelineLoader(gTimelineController));
+OverlayPluginApi.overlayMessage(OverlayPluginApi.overlayName, JSON.stringify({'getUserLocation': true}));
+document.addEventListener("onSendUserConfigLocation", function(e) {
```
quisquous: Can this code be encapsulated somehow? Maybe some resources/userconfig.js file can be included and then each module could just have a one liner of UserConfig.addConfigListener('raidboss') or something.

Also, it looks like the location config is how you're getting the "raidboss" part of `user\raidboss.js`? I think I would prefer if the location was a directory and not something like `C:\etc etc\cactbot\user\raidboss` where it ends with a partial file name.

quisquous:
(Also, congrats to your daughter! :tada:)

ui/raidboss/raidboss.js
ezsoftware: no race, the head element is loaded at the moment it parses the head opening tag.  As the parser moves down through the objects within the head element its appending them to the Head element already within the Document Object Model.

ui/raidboss/raidboss.js
ezsoftware: the location config is the returned string from the config file.  its the full path to the file

for CSS I'm then stripping the .js off and appending .css to load the custom user css.

So selecting say "raidboss-custom.js" in the raidboss's window config would load "raidboss-custom.js" and "raidboss-custom.css" but it can be named anything.  I could call it "purple-duck.js" and it would load that and "purple-duck.css"

ezsoftware: As for encapsulation, yes, that should be do-able, I'll have to double check what version of chromium is running to see if it has ES5 or ES6 support, if it has ES6 support I can do it with ES6 imports, if not ES5 imports should work, they just aren't as pretty as ES6 ones.

CactbotOverlay/CactbotOverlay.cs
ezsoftware: sorry, yes, work uses tabs so I have VS set to use tabs and I forget about it.  I'll fix this and send it.

CactbotOverlay/CactbotOverlay.csproj
ezsoftware: yes, this is accidental, was rushing and forgot to exclude this from the commit.

CactbotOverlay/CactbotOverlay.cs
ezsoftware: C# IIRC can get a callback from chromium when the page is loaded, but that doesn't always coincide with when javascript is ready.  It could happen too early or later than you want.

I was concerned about the options being changed after the settings as well, its one of the things I didn't get into testing.

I'll look more into it and make the tweaks you request.

ezsoftware:
and Thanks!

ui/raidboss/raidboss.js
quisquous: Re: location config.  I think I'd still prefer to have it be a directory.  That lets files like user/raidboss-example.js and documentation tell users what files to create directly, and I can answer support questions by saying things like "put this in your cactbot/user/raidboss.js file".

CactbotOverlay/CactbotOverlay.cs
ezsoftware: I've changed my VS settings for this and did an auto-format to fix anywhere that I still had tabs, so my pull in should be fine now.

CactbotOverlay/CactbotOverlay.csproj
ezsoftware: I've fixed this, my next pull *should* be missing this file, or at least have no changes.

CactbotOverlay/CactbotOverlay.cs
ezsoftware: After looking at the system, the only way to ensure that there are no race conditions is to wait for the JS to send a request back.  I have pulled the code out into its own module and revised the code for raidboss, pullcounter, oopsyraidsy and jobs to use a callback pattern and registered events on DOMContentLoaded to make this work correctly and request the data at the earliest possible moment that there will be no issue of race condition.

I have also set everything up in such a way that Options are loaded, then user options will load, then the plugin will load, even though the user options are dependent on a callback, it will ensure those are loaded if needed, before the overlay instantiates the classes that use them.

I have some validation added to the userconfig file in case it doesn't exist, and otherwise if its blank it will load relative to the AppDomain's BaseDirectory, I'm still testing this to ensure it targets %ACT%\cactbot\user.

ui/raidboss/raidboss.js
ezsoftware: I'm used to working with Electron apps, which are node based, and C# chromium doesn't support modules like electron chromium does.  So, I wasn't able to do what I originally planned, but my fall-back to just putting it in the HTML works.

PR #62, opened 2018-02-25
O8S timeline tweaks

xephero:
Adding some lines for the untargetable/targetable instances that make sense, since they aren't right when Graven Image goes off.

Video with these changes: https://www.twitch.tv/videos/232532546?t=1h24m40s


quisquous:
Thanks for this! Looks really good.  :)

PR #66, opened 2018-02-28
Adding O8S timeline syncs

xephero:
The diff may not be super useful, so I'll summarize the changes (beyond simply adding syncs). My group got to 1389 on this, so up to that is script-generated, past that is hand-made from fflogs. Preserved your existing naming for the most part.

- Added the Thunder/Fire hits of Celestriad
- Fixed some slight +/- 1s timing issues
- Added all non-starstrafe Soaks (The Path of Light, renamed for clarity), since they were included, albeit inconsistently, later on in the fight. Having specific timing for them is useful for dodging into it.
- Added All Things Ending to Past/Future End during Starstrafe. The ATE cast on the normal ones is fast enough (3s) that it's not worth timing it, but there's a bigger gap in that one (8s).
- Also added Meteor hit (the stack) during Starstrafe
- Added some missing Ultimas (1328, 1358, 1543)
- Added a missing Ultimate Embrace (1364)
- Changed tether times to activation rather than appearance

This is untested so far. I go in again on Friday, I just wanted to put it up now in case there's anything glaring

### DIFF
--- a/ui/raidboss/data/timelines/o8s.txt
+++ b/ui/raidboss/data/timelines/o8s.txt
@@ -193,13 +193,13 @@ hideall "--sync--"
 1584 "2x Wings of Destruction" sync /:Kefka:2900:/
 1586 "Soak"
 1590 "Ultima" sync /:Kefka:2911:/
-1597 "Statue Half Cleave" sync /:Graven Image:28D(E|F):/
-1598 "1x Wings of Destruction" sync /:Kefka:28FF:/
+1595 "Statue Half Cleave"
+1596 "1x Wings of Destruction" sync /:Kefka:28FF:/
 
 # Enrage
-1604 "Ultima" sync /:Kefka:2911:/
-1612 "Ultima" sync /:Kefka:2911:/
-1620 "Ultima" sync /:Kefka:2911:/
-1628 "Ultima" sync /:Kefka:2911:/
-1632 "--sync--" sync /:2A52:Kefka/
-1642 "Enrage" sync /:Kefka:2A52:/ jump 0
+1602 "Ultima" sync /:Kefka:283D:/
+1610 "Ultima" sync /:Kefka:283D:/
+1618 "Ultima" sync /:Kefka:283D:/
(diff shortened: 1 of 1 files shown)


xephero:
Tested it out tonight, worked well as far as we saw (end of third forsaken).

Longest pull is here: https://www.twitch.tv/videos/234577658?t=2h20m42s

quisquous:
Dang it, I forgot to try this myself tonight.  Let me know if you're able to give this timeline a test all the way through and then I'll be happy to merge it.  I feel like the existing one does pretty well, so mostly just want to give a new one a smoke test before changing it.

xephero:
Tested it all the way through: https://www.twitch.tv/videos/237540234
Log: https://www.fflogs.com/reports/VQ1G82gczRJyCL9Y/#fight=14

There was some weirdness right at the end:
```
OnSync: 1590 (prev: 1590.107) (this ultima synced correctly)
OnSync: 1597 (prev: 1595.377)
OnSync: 1598 (prev: 1597.733)
OnSync: 1632 (prev: 1632.843)
```

The corresponding events don't really line up very well:
```
00:09:55.196 | Kefka 1 Ultima
00:10:00.714 | Graven Image 4 casts Intemperate Will |  
00:10:01.560 | Kefka 1 casts Wings Of Destruction
00:10:36.219 | Kefka 1 begins casting Light Of Judgment
```

I decided to take out the Graven sync, since it's a separate unit and runs on a different timer, even though the earlier tether syncs were more or less right on the mark. The fact that one synced from .377 and the next synced from .733 indicates that the real boss is still operating on whole-second units. I bumped up the timeline since 09:55.196 to 00:10:01.560 is rightly a 6 second gap with rounding.

I also discovered that the ending Ultima crescendo is a different ability ID, so I fixed that as well. The last line was likely just the drift from not having synced in a while.

quisquous:
Congrats on the clear! Thanks again for the testing and all the timeline additions.  I'll go ahead and merge this.

Also, I wonder if the last ultimas are different because it gives a buff and isn't just damage on its own.

PR #67, opened 2018-03-01
make_timeline.py improvements

xephero:
Added a lot of functionality to this script.

- Added FFlogs support. This requires the requests library; if it's not present, the error on line 27 instructs the user how to install it. This also requires an API key, and also provides direction if it's missing. With those things active, it will pull all the cast events and make a timeline the same way it does for files.

- Added event filtering. Using the -ii argument, you can provide a list (like -ii 2ABC 2ABD 2ACE) to ignore ability IDs. This will also ignore them for the purposes of calculating time deltas between two lines, so, say, four players hitting their kefka heads won't throw off the rounding. You can also use -ic and -ia for combatant or ability names respectively.

- Added phase support. You can pass something like "-p 28EC:1000" to make the timeline jump to 1000 the first time it sees the ability id 28EC. This is useful for starting off with God Kefka's offset, but also for stuff like padding out the timing on Phantom Train's return.

- The structure should make it importable by other python scripts now, too, if anyone wants to use any pieces from it.

I figure as patches come, we can use network logs for the window where ACT is working but FFLogs isn't, and then when FFlogs adds support for new fights, make timelines based on the early clears. Should make the turnaround really fast once everything gets updated.

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -5,7 +5,7 @@ import re
 def timestamp_type(arg):
     """Defines the timestamp input format"""
     if re.match('\d{2}:\d{2}:\d{2}\.\d{3}', arg) is None:
-        raise argparse.ArgumentTypeError
+        raise argparse.ArgumentTypeError("Invalid timestamp format. Use the format 12:34:56.789")
     return arg
 
 def parse_time(timestamp):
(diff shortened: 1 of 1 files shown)


quisquous:
This is super cool! Thanks!!

PR #72, opened 2018-03-08
raidboss: O7S timeline fixes

xephero:
Minor oversights. They were close enough to not lose syncing but still enough to be misleading.


quisquous:
Thanks!

PR #85, opened 2018-03-16
Update o5s.js

DrLippe:
Added de Localization / todo: fr/ko

### DIFF
--- a/ui/raidboss/data/triggers/o5s.js
+++ b/ui/raidboss/data/triggers/o5s.js
@@ -57,14 +57,14 @@
   ],
   triggers: [
     {
-      regex: /04:Removing combatant (魔列車|Phantom Train|Phantomzug|Train Fantôme)/,
+      regex: /04:Removing combatant (?:魔列車|Phantom Train|Phantomzug|Train Fantôme)/,
       run: function(data) {
         data.StopCombat();
       },
     },
     {
       id: 'O5S Doom Strike',
-      regex: /14:28B1:(魔列車|Phantom Train|Phantomzug|Train Fantôme) starts using (魔霊撃|Doom Strike|Vernichtungsschlag|Frappe Létale) on (\y{Name})/,
+      regex: /14:28B1:(?:魔列車|Phantom Train|Phantomzug|Train Fantôme) starts using (?:魔霊撃|Doom Strike|Vernichtungsschlag|Frappe Létale) on (\y{Name})/,
       alertText: function(data, matches) {
         if (matches[1] == data.me)
           return {
@@ -87,7 +87,7 @@
     },
     {
       id: 'O5S Head On',
-      regex: /14:28A4:(魔列車|Phantom Train|Phantomzug|Train Fantôme) starts using (追突|Head On|Frontalangriff|Plein Fouet)/,
+      regex: /14:28A4:(?:魔列車|Phantom Train|Phantomzug|Train Fantôme) starts using (?:追突|Head On|Frontalangriff|Plein Fouet)/,
       alertText: function (data) {
           return {
               en: 'Go To Back',
@@ -103,7 +103,7 @@
     },
     {
       id: 'O5S Diabolic Headlamp',
-      regex: /14:28B2:(魔列車|Phantom Train|Phantomzug|Train Fantôme) starts using (魔界の前照灯|Diabolic Headlamp|Diabolische Leuchte|Phare Diabolique)/,
+      regex: /14:28B2:(?:魔列車|Phantom Train|Phantomzug|Train Fantôme) starts using (?:魔界の前照灯|Diabolic Headlamp|Diabolische Leuchte|Phare Diabolique)/,
       alertText: function (data) {
           return {
               en: 'Stack Middle',
@@ -153,7 +153,7 @@
     },
     {
       id: 'O5S Remorse',
-      regex: /Added new combatant (未練のゴースト|Remorse|Melancholischer Geist|Fantôme Mélancolique)/,
+      regex: /Added new combatant (?:未練のゴースト|Remorse|Melancholischer Geist|Fantôme Mélancolique)/,
       infoText: function (data) {
           return {
               en: 'Knockback Ghost',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/triggers/o5s.js
```diff
@@ -14,48 +54,158 @@
...
+			  fr: 'stack middle',
+			  ko: 'stack middle',
+		  };
+	  },
+
     },
     {
       id: 'O5S Diabolic Light',
       regex: /1B:........:(\y{Name}):....:....:0001:0000:0000:0000:/,
+	  regexDe: /1B:........:(\y{Name}):....:....:0001:0000:0000:0000:/,
```
quisquous: If the regexes are the same in other languages, you can just specify regex and not need to specify regexDe/regexFr/regexKo.  I think this would reduce duplication here.

ui/raidboss/data/triggers/o5s.js
```diff
@@ -14,48 +54,158 @@
...
+	  regexKo: /1B:........:(\y{Name}):....:....:0001:0000:0000:0000:/,
+	  
       condition: function(data, matches) { return matches[1] == data.me; },
-      infoText: 'Light',
-      tts: 'light',
+      infoText: function (data) {
+		  return {
+			  en: 'Light',
+			  de: 'Licht',
+			  fr: 'Light',
```
quisquous: Similarly, everything uses 'en' if the specific language isn't specified.  So here you could just remove the fr and the ko lines, as they're not adding anything.

ui/raidboss/data/triggers/o5s.js
```diff
@@ -14,48 +54,158 @@
...
       regex: /14:28B1:Phantom Train starts using Doom Strike on (\y{Name})/,
+	  regexDe: /14:28B1:Phantomzug starts using Vernichtungsschlag on (\y{Name})/,
+	  //regexFr: /14:28B1:/,
+	  //regexKo: /14:28B1:/,
+	  
       alertText: function(data, matches) {
         if (matches[1] == data.me)
-          return 'Tank Buster on YOU';
+          return {
+			  en: 'Tank Buster on YOU',
```
quisquous: Can you use two spaces everywhere instead of tabs?

ui/raidboss/data/triggers/o5s.js
```diff
@@ -14,48 +54,158 @@
...
+		  };
+	  },
     },
     {
       id: 'O5S Remorse',
       regex: /Added new combatant Remorse/,
-      infoText: 'Knockback Ghost',
-      tts: 'knockback ghost',
+	  regexDe: /Added new combatant Melancholischer Geist/,
+	  //regexFr: /Added new combatant /,
```
quisquous: I think no need for the extra comments here and on other triggers for untranslated languages.  I think it's pretty clear how to add new ones.

quisquous:
Also also, feel free to update multiple files in the same commit! They all don't necessarily need separate pull requests.

ui/raidboss/data/triggers/o5s.js
DrLippe: i added these to make future modification easier. french translation here is just a dummy

ui/raidboss/data/triggers/o5s.js
DrLippe: i'm editing the files with notepad++ it does the tabs automaticly

ui/raidboss/data/triggers/o5s.js
DrLippe: realized this in o6s and following

ui/raidboss/data/triggers/o5s.js
DrLippe: actually would it be possible to combine all languages in one regex? like

regex: /Added new combatant (Remorse|Melancholischer Geist|Fantôme Mélancolique|未練のゴースト)/

ui/raidboss/data/triggers/o5s.js
danakj: Settings->Preferences->Language. Tab Settings section.
1. Replace by space: Check the box. 
2. Tab size: 2

ui/raidboss/data/triggers/o5s.js
danakj: It attempts to give a place to fill in the french, but it also gives a false impression that it is translated when it is not. Hopefully it's already clear how to do the french translation for future authors and they can add the 'fr' key.

DrLippe:
i ran in some issues here. seems to be something with adding japan regexes

quisquous:
Thanks! I'll give this a try when I get some time tomorrow and then merge it.  Did all these translations come from fflogs, or somewhere else?

DrLippe:
i took the ability translations from xivdb. the rest i took from my own logs after running the duty in different languages.

the way i used the regex works with all languages regardless of the language setting in FFXIV Plugin.
i prefer this method as it works even with different language settings on plugin and game. 
for example: french client, german plugin. -> regexDe and regexFr wont work but regex (?:De|Fr) will

tested this myself with byakko ex in different languages.
the issue with japan regexes has been a "," i missed after adding them. works fine now

danakj:
Maybe cactbot should just append all the regexes together to do its matching instead of picking one.

Tho you can tell cactbot what language to use right?

DrLippe:
the problem is you get logs with languages mixed up like
frenchmobname starts using frenchabilityname -> right configuration of act -> regexFr works
germanmobname starts using englishabilityname -> wrong configuration of act -> regex/regexDe wont work
with 4 languages you get like 16 possible log outputs
(?:japanmobname|englishmobname|germanmobname|frenchmobname) starts using (?:... solves this problem

quisquous:
Sorry for the delay on getting back to this.  I've been out of town and then had a busy week.  I'll try to take a look and land this today.

I'm not sure this mixing and matching of ACT config and game language is something I want to support.  It seems to me that most folks running the game in other languages are almost certainly going to set ACT to run in that language too.  I don't want to get to a place where users have to set two variables based on what language they want for code elsewhere.

PR #97, opened 2018-04-04
Update fractal_continuum_hard.js

jerrybotticelli:
Death Spin is 27AD and 27AE


quisquous:
It is both, but only the 27AE version has knockback attached to it.  The 27AD one is the unpowered up original one that just hits you in place, if I'm not mistaken.

quisquous:
It used to trigger on both but 532e090 adjusted it.

jerrybotticelli:
You're right, I just noticed it. Closed this ^^

PR #103, opened 2018-04-16
eureka: add ability to import times from web tracker

christianhood:
Hi, I've made it so that you can take the clipboard copy function from ffxiv-eureka.com
image and paste them into the game 
e.g. `/e ★ NMs on cooldown: Sabo (70m) > Emperor (82m) > Pazuzu (85m) > Caym (98m) > Amemet (110m)`
This will grab the times and update them on the cactbot tracker.

### DIFF
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -1,4 +1,4 @@
-"use strict";
+﻿"use strict";
 
 var Options = {
   Language: 'en',
@@ -155,6 +155,7 @@ var Options = {
 
 var gFlagRegex = Regexes.Parse(/00:00..:(.*)Eureka (?:Anemos|Pagos) \( (\y{Float})  , (\y{Float}) \)(.*$)/);
 var gTrackerRegex = Regexes.Parse(/(?:https:\/\/)?ffxiv-eureka\.com\/(\S*)\/?/);
+var gImportRegex = Regexes.Parse(/00:00..:(.*)★ NMs on cooldown: (\S.*\))/);
 var gGalesIcon = "&#x1F300;"
 var gWeatherIcons = {
   Gales: gGalesIcon,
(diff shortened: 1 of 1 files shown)


christianhood:
My code might not be the best so if you just want to take the concept and do it your own way that's fine too

ui/eureka/eureka.js
```diff
@@ -404,6 +405,22 @@ class EurekaTracker {
       if (match) {
         this.currentTracker = match[1];
       }
+      match = log.match(gImportRegex);
+      if (match) {
+        var webKeys = ['Sabo', 'Lord', 'Teles', 'Emperor', 'Callisto', 'Number', 'Jaha', 'Amemet', 'Caym', 'Bomba', 'Serket', 'Julika', 'Rider', 'Poly', 'Strider', 'Hazmat', 'Fafnir', 'Amarok', 'Lamashtu', 'Pazuzu'];
```
quisquous: Can you move these up to zone info dict up top? Maybe just add a trackerKey field to each of them with this string, e.g.
```
lama: {
  name: 'Lama',
  mobName: 'Lamashtu',
  trackerName: 'Lamashtu',
  x: 7.7,
  y: 25.3,
  time: 'Night',
},
```

There's going to be some other Eureka zone soon, so having it in the zone info will make it easier to add the next zone.

christianhood:
I thought about doing it that way, but I wasn't sure how to get the names from there without doing a for loop and looking at each one individually. 
I can do it that way if you want or another option would be to change your keys to match theirs if you aren't attached to them. Then we can just do a toLower on the string to match them up.

ui/eureka/eureka.js
christianhood: @quisquous I thought about doing it that way, but I wasn't sure how to get the names from there without doing a for loop and looking at each one individually. 
I can do it that way if you want or another option would be to change your keys to match theirs if you aren't attached to them. Then we can just do a toLower on the string to match them up.

christianhood:
That commit is how I was suggesting doing it. Didn't realize it would automatically add it to the PR haha.

christianhood:
Damn, left a line in there on accident, knew I should have tested it first.

quisquous:
Thanks for this! I think I'm going to land this and then just follow up with a few small things.  Was hoping to do another release this weekend.  :)

christianhood:
Oh, sorry about that, got busy with some other stuff and completely forgot about it

quisquous:
No worries at all! I figured it was something like that.  I just did some Eureka and having this is *so* convenient.

PR #111, opened 2018-04-29
Last 2 remaining French translations for O8S (Fire spread and stack)

Amiral-Benson:
I added the last 2 remaining French translations for Fire spread and stack on **O8S.js**.


quisquous:
Thanks!!

PR #112, opened 2018-04-29
UCoB partial French translation

Amiral-Benson:
Some small correction to the **UCoB Timeline** to facilitate the automatic replace text Translations, some of the original English abilities there weren't wrote the exact same as in-game ones. (**Liquid Fire** instead of **Liquid Hell**, **Earthshaker** instead of **Earth Shaker**, etc.).

And partial French translation of **UCoB triggers**, I found all the missing quotes of Nael in TeraFlare. 
Now I believe that the file requires some modifications first before what remains can be translated as well.


ui/raidboss/data/triggers/unending_coil_ultimate.js
```diff
@@ -187,96 +199,142 @@
...
-      tts: 'spread then in',
+      tts: {
+        en: 'spread then in',
+        fr: 'Se séparer, puis se rassembler',
+      },
     },
     { id: 'UCU Nael Quote 2',
       regex: /From on high I descend, mine enemies to smite/,
       regexFr: /Du haut des cieux, je vais descendre pour conquérir/,
       infoText: function(data) { return 'Spread => Out'; },
```
quisquous: Did you want to translate these infoTexts too, or should I just duplicate what you put for the tts?

ui/raidboss/data/triggers/unending_coil_ultimate.js
```diff
@@ -287,7 +345,10 @@
       regexJa: /:サンダーウィング:26C7:.*?:........:(\y{Name}):/,
       condition: function(data, matches) { return data.me == matches[1]; },
       alarmText: function(data) { return 'Thunder on YOU'; },
```
quisquous: This one too?

ui/raidboss/data/triggers/unending_coil_ultimate.js
quisquous: Oh I see, you probably skipped these because they were functions.  I think any case where there's a function that just returns a string, we could just drop the function too.

Amiral-Benson:
Thank you for your explanations, this allowed me to translate almost everything of what was remaining.

I also remade all **Spread/In/Out/Stack** French translations, for better fidelity to English ones, and for more consistency.

There's some places for which I didn't knew what to do:

- Line 567
- Line 676 (I don't understand this one enough)
- Line 812 (I'm not sure I understand, is that a smiley? I know it's trying to find Twin target in case of death, but I'm not sure how it's trying to say it).
- Line 957 & 980: There's an unusual way to return the text here, I don't know how to handle it.
- Line 1203: The cardinal and inter-cardinal directions, they are all the same in French, except for **W (it's O)**, **SW (it's SO)**, **NW (it's NO)**. I don't know if it's relevant to change them, and how. I believe know where they are called and I put the correct French word there "au".

I skipped one of the **Morn Afah** one, because it's the same in French, and I skipped all **Akh Morn** ones for the same reason (French TTS voice is able to read them with no ambiguity).

Maybe there's also still some places I missed.

quisquous:
Re: 676.  https://github.com/quisquous/cactbot/blob/3e5386df3ae84b881ac5bdfaa7ff0e0c601e2297/ui/raidboss/data/triggers/unending_coil_ultimate.js#L1190 tries to explain wide dives.   For  N, NE, E, SE, SW dragons, I prefer to put C west instead of C south.  I think it's possible to put B as far counterclockwise as possible and C as far clockwise, but I don't like how tight that is.  So, I put C west in that case and call out "wide" to have everybody move more.

quisquous:
Re: Line 567.  Fixed.  Makes sense why that's confusing.
Re: Line 812.  No smiley, just three question marks.
Re: Line 957 & 980.  Haha, I have no idea why I wrote that like that.  I'll clean that up.  Not sure how to translate that though.
Re: Line 1203.  Fixed.

PR #119, opened 2018-05-16
Added lots of German translations and fixed some small issues here and there

ArkfrostLumas:
I've added German TTS and Info-Texts to many of the raids, changed the TTS of some German callouts to fit the default Microsoft Hedda Desktop voice for better pronounciation and I've also added the French localization for o8n while I was at it.
There might be some things you have to look over, like o8n at the end, where I've just copy&pasted the timeline translations from o8s, but there're still lots of leftover God Kefka lines which I didn't look up.


ui/raidboss/data/triggers/o8n.js
```diff
@@ -6,110 +6,401 @@
...
+      tts: {
+        en: 'knockback on boss',
+        fr: 'Projection depuis le boss',
+        de: 'Rückstoß vom Boss',
+      },
     },
     {
       id: 'O8N Flagrant Fire Single',
       regex: /1B:........:(\y{Name}):....:....:0017:0000:0000:0000:/,
+      regexDe: /1B:........:(\y{Name}):....:....:0017:0000:0000:0000:/,
```
quisquous: Just FYI, no need for these if they're duplicates.  If you specify a language like 'de', then it will use regexDe if it's there, otherwise it will fall back to regex on its own.  Would you mind removing all the 1B head marker duplicate lines you added?

ui/raidboss/data/triggers/o8s.js
```diff
@@ -107,26 +119,30 @@
...
         en: 'stack and through',
         fr: 'Stack et le traverser',
+        de: 'Stek und durchlaufen',
       },
     },
     {
       id: "O8S Past's End",
       regex: /Kefka starts using Past's End/,
-      regexDe: /Kefka starts using Past's End/,
+      regexDe: /Kefka starts using Ende der Welt/,
```
quisquous: Whoops.  Thanks for fixing this.  Not sure why none of these translated.

ui/raidboss/data/triggers/o8n.js
```diff
@@ -6,110 +6,401 @@
...
+        de: 'Wahres Eis: Rausgehen',
+      },
+      tts: {
+        en: 'Get out',
+        fr: 'Sortir',
+        de: 'Rausgehen',
+      },
+    },
+  ],
+  timelineReplace: [
```
quisquous: I think it's fine to just copy and paste this here.  It seems more work than it's worth to separate out what's included and what's not without an automatic tool to do it.

ui/raidboss/data/triggers/o8s.js
```diff
@@ -519,8 +567,8 @@
       replaceSync: {
         'Graven Image': 'Heilige Statue',
         'Kefka': 'Kefka',
-        'Light Of Consecration': 'Licht[p] Der Weihe',
-        'The Mad Head': 'Verrückt[a] Kopf',
+        'Light Of Consecration': 'Licht Der Weihe',
+        'The Mad Head': 'Verrückter Kopf',
```
quisquous: Thanks for fixing these up.  I wasn't sure what to do with this, because this is what fflogs gave me back.

ui/raidboss/data/triggers/o8n.js
```diff
@@ -6,110 +6,401 @@
...
+      regexDe: /14:2927:Heilige Statue starts using Schockwelle/,
+      regexFr: /14:2927:Statue Divine starts using Onde De Choc/,
       delaySeconds: 5,
-      alertText: 'Look for Knockback',
-      tts: 'knockback',
+      alertText: {
+        en: 'Look for Knockback',
+        fr: 'Se préparer à la projection',
+        de: 'Auf Rückstoß achten',
+      },,
```
quisquous: Can you remove one of these commas?

The thing I always do when submitting triggers is to reload raidboss and make sure that it doesn't have any errors.

ui/raidboss/data/triggers/o8n.js
```diff
@@ -6,110 +6,401 @@
...
+
+        'Soak': 'Absorber',
+        'Past/Future': 'Passé/Futur',
+        'Past/Future End': 'Passé/Fin du futur',
+        'Knockback Tethers': 'Liens de projection',
+        'Sleep/Confuse Tethers': 'Liens de Sommeil/Confusion',
+        'Statue Half Cleave': 'Demi clivage de la statue',
+      },
     },
-  ]
```
quisquous: You need to add back in this trailing `]` here.  This is what closes the `timelineReplace: []` block.

ui/raidboss/data/triggers/o8n.js
ArkfrostLumas: Ah good to know. That was my only worry there, that if I don't specify the regexDe it won't display anything when it's set to de. But if it falls back then I'll cut that back out again :)

ui/raidboss/data/triggers/o8s.js
ArkfrostLumas: No problem ^^ I can also translate the rest (fr and jp) while I'm at it so you don't have to do it :)

ui/raidboss/data/triggers/o8n.js
ArkfrostLumas: Yup, I think so, too. Though when it's done you might have to take a look at the function closing brackets if I didn't use one too much or one is missing, since I don't see, which bracket closes what that good. I'll copy it into notepad++ though. Maybe it shows which bracket belongs to which and if there's some overflow of brackets x)

ui/raidboss/data/triggers/o8s.js
ArkfrostLumas: Yup, that's a common mistake ff14angler, xivdb and fflogs do, but in the game and in the german logs it's usually the correct expression. I've checked it from my last o8s log :)

ui/raidboss/data/triggers/o8n.js
ArkfrostLumas: Oh, I didn't see that. Gonna fix it :)

ui/raidboss/data/triggers/o8n.js
ArkfrostLumas: Ah, that was a little accident. Gonna put it back in x)

ArkfrostLumas:
Okay I've just edited the missing mistakes.

Also, I've tried to test the changes, but somehow the timeline isn't registering that my game is set to German and still displays the English timeline. Just to reproduce I'll list the steps I took:

i. Set the game back to German
ii. Set the parsing language in the FFXIV settings from ACT back to German
iii. Disabled test.js (because it was starting the test.js instead of the o8s.js while testing)
iv. Went to middle LaNoscea and started the timeline via /echo :Kefka:28C2:

I've also added the German line for the Timeline-Reset:
'The limit gauge resets!': 'Der Limitrausch-Balken wurde geleert.',
And afterwards I tried the whole thing again:

i. Went to middle LaNoscea
ii. Started the timeline via /echo :Kefka:28C2:
iii. Tried resetting the timeline via /echo Der Limitrausch-Balken wurde geleert. => Didn't work
iv. Tried resetting the timeline via /echo The limit gauge resets! => Worked

And as it turns out the timeline starts and works perfectly - though the timeline-replace doesn't seem to work :/

quisquous:
Thanks for the triggers.  I'll just go ahead and commit this patch now.  More followup can be done in other patches.

I haven't figured out how to have cactbot detect your language.  I know there's some login message like "Welcome to WORLD!" and "Welcome to FINAL FANTASY XIV" in English.  Maybe I could use that to key off of.

At any rate, for now you need to add an option for language in cactbot/user/raidboss.js and set `Options.Language = 'de'` for it to work.  Take a look at https://github.com/quisquous/cactbot/blob/master/user/raidboss-example.js for more options help too.

One other thing I need German translations for is all of the healer lb3s and the status effects for Weakness and Brink of Death, e.g. https://github.com/quisquous/cactbot/blob/master/CactbotOverlay/WipeDetector.cs#L46 and https://github.com/quisquous/cactbot/blob/master/CactbotOverlay/WipeDetector.cs#L23.

PR #121, opened 2018-05-22
More German stuff :)

ArkfrostLumas:


ui/raidboss/data/triggers/o5n.js
```diff
@@ -51,5 +81,127 @@
       infoText: 'Wind',
       tts: 'wind',
     },
-  ]
+  ],
+  timelineReplace: [
```
quisquous: Ok ok, this is why your patch has so many merge conflicts listed.  I went ahead and did that in upstream cactbot.  If you look at your fork, it says "This branch is 19 commits ahead, 7 commits behind quisquous:master."

One is that I didn't take your commits directly, I squashed them into a single commit.  I also added a few commits of my own past yours.  So, now our two repositories have diverged.

It looks like github has a resolve conflicts button down below for me.  Is that something you can see and sort through?  If you haven't handled merge conflicts before, all the stuff between <<< and === is from your patch, and everything between === and >>> is stuff that's been checked in master.  I went through the timelines and removed things that weren't necessary and also added every missing translation that was needed to fully translate the timeline.  If you just copied and pasted from the savage versions, I suspect you could just always accept my version and be ok.

In the future, I'd recommend uploading patches on separate branches other than master.  (https://learngitbranching.js.org/ is pretty good if you want a tutorial).  That way I can merge things in to upstream and you can re pull back to master and then branch from there to make the next patch.  Usually what most folks do is that they keep master up to date by pulling the latest version from upstream and then do feature work in separate branches that they then delete after it's been merged in.

ui/raidboss/data/triggers/unending_coil_ultimate.js
```diff
@@ -212,185 +221,241 @@
...
-        fr: "Se rassembler, puis dehors",
+        fr: 'Se rassembler, puis dehors',
+        de: 'stek dann raus',
       },
     },
     { id: 'UCU Nael Quote 5',
       regex: /O red moon, scorch mine enemies/,
       regexFr: /Que l'incandescence de la lune brûle mes ennemis/,
+      regexDe: /O roter Mond! Umlodere meinen Pfad/,
+      regexJp: /月よ！<br>赤熱し、神敵を焼け/,
```
quisquous: Hmm, surely it doesn't say <br> in Japanese.  I wonder what that ends up looking like in game.  Maybe just replace those with `.*`?

quisquous:
It still looks like this has conflicts, so I can't merge it directly.  (See my previous comments).  Let me know if I can do or explain anything to help you get those resolved.

ui/raidboss/data/triggers/unending_coil_ultimate.js
ArkfrostLumas: I guess those are in there since as it seems those lines are written in html and they wanted to start a new paragraph for each line in the Japanese version. 

I've tried to look up japanese ACT triggers for UCoB though and as it seems in the JP version they just insert the last paragraph of the text, as it's different for every mechanic and that br seems to cause ACT issues when reading the logs. Gonna correct that :)

ui/raidboss/data/triggers/o5n.js
ArkfrostLumas: Oh, just noticed this now. I have looked over the conflicts and corrected the issues that I found :)

quisquous:
I'm going to close this old german pull request, then.

PR #127, opened 2018-05-25
Ridorana Lighthouse

xephero:
Hello again! Happy patch time!

I've added Ridorana Lighthouse timelines for all bosses. I ran it about 4x and refined it each time, and this is what I arrived at.

As a side note, getting pulled into Construct 7's add zone drops combat, not unlike the O5 ghost car issue from before. It fixes once Annihilation Mode casts though, by syncing from the stopped state.

I haven't added a timeline since all the localization stuff, if there's something I could do to them to make them more easily localizable let me know and I'm happy to accomodate.

### DIFF
--- a/ui/raidboss/data/timelines/ridorana_lighthouse.txt
+++ b/ui/raidboss/data/timelines/ridorana_lighthouse.txt
@@ -1,0 +1,264 @@
+# The Ridorana Lighthouse
+
+hideall "--reset--"
+hideall "--sync--"
+
+0 "--reset--" sync /:Removing combatant Famfrit, The Darkening Cloud\.  Max HP: \d{8}\./ window 10000 jump 0
+0 "--reset--" sync /:Removing combatant Belias, The Gigas\.  Max HP: \d{8}\./ window 10000 jump 0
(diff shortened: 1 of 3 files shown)


quisquous:
Thanks for this (and all your timelines)!

Hmm, that's unfortunate that it drops combat.  I need to do some further testing to see if it's really dropping combat (e.g. you can change role actions, gear, or ast stances) or if I need to dig up a better memory signature.  Do you think it'd be worth doing something like o5s here and manually stopping combat (https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/triggers/o5s.js#L9-L19)?  Maybe that's overkill for the whole zone and I need to make it possible to start/stop manually dropping combat just for that one section.

ui/raidboss/data/timelines/ridorana_lighthouse.txt
```diff
@@ -0,0 +1,264 @@
+# The Ridorana Lighthouse
+
+hideall "--reset--"
+hideall "--sync--"
+
+0 "--reset--" sync /:Removing combatant Famfrit, The Darkening Cloud\.  Max HP: \d{8}\./ window 10000 jump 0
```
quisquous: \d{8} is a much nicer way to put this in case they ever change HPs.  Good thought!

PR #128, opened 2018-05-25
Tsukuyomi EX Timeline

xephero:
This timeline is as-of-yet untested. It was made with make_timeline.py and curated a bit from videos, logs, and guides as reference. I've cleared the fight, so I know what each mechanic does, but I did so before ACT was fixed and haven't been back in. This is mostly meant as a starting point.

In particular, the add phase is fairly empty right now and could use spawn times and syncs to Added combatant... lines.


quisquous:
Given that this content is so new, I'm going to land this and we can test and iterate from there.  Thanks again!

PR #129, opened 2018-05-26
Cleaning up timelines

xephero:
Fixed some timing issues in Ridorana and added the add phase in Tsukuyomi. Weirdly, line 17 doesn't work even though I'm sure that line is present in the logs. Not sure what's up with that, probably a FFXIV plugin hiccup, but it only causes Concentrativity to show the wrong timing and it's pretty clearly visible. Might want to just hide that for now though.

Can see the Tsukuyomi timeline in action here: https://www.twitch.tv/videos/265667584?t=1h19m15s

I did learn from this low-dps learning party that the middle phase isn't actually a real loop, it will only run until the second Antitwilight/Perilune, so I might change that from a loop to the real sequence, but this is good enough to be useful as-is.

### DIFF
--- a/ui/raidboss/data/timelines/ridorana_lighthouse.txt
+++ b/ui/raidboss/data/timelines/ridorana_lighthouse.txt
@@ -16,9 +16,9 @@ hideall "--sync--"
 26 "Water IV" sync /:Famfrit, the Darkening Cloud:2C3D:/
 35 "Briny Cannonade" sync /:Famfrit, the Darkening Cloud:2C45:/
 47 "Tsunami sync /:Famfrit, the Darkening Cloud:2C50:/
-53 "Tsunami" duration 3
-65 "Tsunami" duration 3
-77 "Tsunami" duration 3
+54 "Tsunami" duration 3
+66 "Tsunami" duration 3
+78 "Tsunami" duration 3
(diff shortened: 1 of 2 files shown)


jerrybotticelli:
I haven't done Ex yet so I don't have any logs to look at, but could your issue with line 17 be the capital letter o? Line 29 has a similar npc name without a capital o.

ui/raidboss/data/timelines/ridorana_lighthouse.txt
```diff
@@ -34,10 +34,10 @@ hideall "--sync--"
 176 "Dark Cannonade" sync /:Famfrit, the Darkening Cloud:2C42:/
 200 "Water IV" sync /:Famfrit, the Darkening Cloud:2C3D:/
 208 "Tsunami" sync /:Famfrit, the Darkening Cloud:2C50:/
-214 "Tsunami" duration 3
-228 "Tsunami" duraiton 3
```
quisquous: I've been thinking about this a lot, and I really need to get some bots to verify timelines and trigger syntax on pull requests (and master).  I need to automate this for sure with so many contributions from you and other folks.

quisquous:
Thanks, as always!

PR #133, opened 2018-05-27
Fixes for debuff triggers

jerrybotticelli:
@quisquous did you add the additional condition statements? 

I setup a test scenario using echo and it would not work with 2 condition statements in each trigger, so I separated them and it works fine this way.


quisquous:
OOPS.  I am so sorry for the hassle.  Thanks for fixing my obvious mistakes. :sob:

PR #135, opened 2018-05-28
raidboss: Tsukuyomi timeline cleanup

xephero:
This should be the final revision. Just cleared with this version and can't see any further changes unless people discover more patterns in the fight than I'm currently aware of to remove the RNG. Sorry for the spamming of revision commits on this one!


quisquous:
No apologies needed! Thanks for all the hard work.  :heart:

PR #136, opened 2018-05-28
byakko translation

ArkfrostLumas:
Finally got it working :)

### DIFF
--- a/ui/raidboss/data/triggers/byakko-ex.js
+++ b/ui/raidboss/data/triggers/byakko-ex.js
@@ -6,36 +6,55 @@
     {
       id: 'ByaEx Heavenly Strike',
       regex: / 14:27DA:Byakko starts using Heavenly Strike on (\y{Name})/,
+      regexDe: / 14:27DA:Byakko starts using Himmelszorn on (\y{Name})/,
       alertText: function(data, matches) {
         if (matches[1] == data.me)
-          return 'Tank Buster on YOU';
+        return {
+          en: 'Tank Buster on YOU',
+          de: 'Tankbuster auf DIR',
+        };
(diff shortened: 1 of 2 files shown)


ArkfrostLumas:
Okay, Eureka is done now :) 
The second edit on Eureka (1b57a9b) is because I accidentally changed one Regex to RegexDe instead of adding a RegexDe Line (and it wasn't even necessary there). Reverted that change and all should be good now :)

ui/raidboss/data/triggers/byakko-ex.js
```diff
@@ -6,36 +6,55 @@
     {
       id: 'ByaEx Heavenly Strike',
       regex: / 14:27DA:Byakko starts using Heavenly Strike on (\y{Name})/,
+      regexDe: / 14:27DA:Byakko starts using Himmelszorn on (\y{Name})/,
       alertText: function(data, matches) {
         if (matches[1] == data.me)
-          return 'Tank Buster on YOU';
+        return {
```
quisquous: Would you mind indenting these returns and adding curly braces around the conditionals with multi-line bodies, like this:

```
        if (matches[1] == data.me) {
          return {
            en: 'Tank Buster on YOU',
            de: 'Tankbuster auf DIR',
          };
        }
```

...both on this line, but elsewhere in these diffs too.

I'm looking into something like https://github.com/beautify-web/js-beautify to do this automatically but I haven't set that up yet (or committed to a particular pretty printer).

ArkfrostLumas: Sure, I can do that ^^

quisquous:
Closing this as it should be all included in #137.

PR #149, opened 2018-06-05
noticed some mistakes and some regexes didn't work in german

ArkfrostLumas:
I did a ridorana run and happened to have cactbot running, but I noticed some mistakes and some misleading TTS regarding some mechanics. Also the Regexes for the division mechanics of Construct 7 don't seem to work in German. I've searched my logs and there were no 14:2CCA:, 14:2CCC: and 14:2CCD: in the logs. Only the indivisible had 14:2CCE:, but the German ones for numbers weren't logged that way, so I adjusted the Regexes from what I found in the Logs.


quisquous:
Hmm.  That's quite odd that the abilities aren't there.  Can you send me a network log of a ridorana run at some point?

PR #154, opened 2018-06-08
FR Translation & Correction for Tsukuyomi-ex

mooondark:
Hello,
I've corrected and added the missing translation.

I tested it, all is functionnal when game is set on French & Options.Language = 'fr' is present in user/raidboss.js.

This is the first time ever I'm using GitHub so if I made a mistake, just tell me.

PR #158, opened 2018-06-10
Fix Fr Traduction for Tsukuyomi

G0li4th:


quisquous:
Thanks for this pull request, but this just got finished by @mooondark in #154, which is why you're conflicting.  If you're looking to do more French translations, you might want to coordinate with them so you don't step on each other's toes.

PR #159, opened 2018-06-11
FR Localization for o7s / O5-8n / Ridorana / Rabanastre

mooondark:


ui/raidboss/data/triggers/o7s.js
```diff
@@ -1,4 +1,4 @@
-'use strict';
+﻿'use strict';
```
quisquous: Can you remove the unicode BOM that you added to this file? That's what travis ci is complaining about.  If you're using Notepad++ you can save this file as UTF-8 without the byte order mark (BOM).

mooondark:
Yeah check passed :)
I learn something today (BOM) :D

PR #161, opened 2018-06-12
Fix tsukuyomi FR Traduction

G0li4th:


mooondark:
I don't think we need a translation for "start using" in regexFr: / 14:2BBC:Tsukuyomi **starts using** Jeune Nuit/ as it's working fine now (tested by myself).
Some of your translations are better but you deleted some triggers (like _regexFr: /00:0044:Non\, je ne peux pas\.\.\. échouer\.\.\./,_) and I don't know why.

G0li4th:
maybe i failed during merge :( Well lets close that PR

PR #167, opened 2018-06-14
Minor fix for Byakko-ex

mooondark:
My bad, duplicate a 'de' string and forgot to put 'fr' instead.


mooondark:
It was my mistake, you don't need to apologize 😄

PR #174, opened 2018-06-21
Tsukuyomi-ex minor fix

nuks-fr:
there are 2 codes for the same ability : 
[23:20:49.336] 14:2BBB:Tsukuyomi starts using Brimade Meurtrière on Name Hidden.
[23:23:39.119] 14:2EB2:Tsukuyomi starts using Brimade Meurtrière on Name Hidden.


ui/raidboss/data/triggers/tsukuyomi-ex.js
```diff
@@ -29,9 +29,9 @@
     },
     {
       id: 'Tsukuyomi Torment',
-      regex: / 14:2BBB:Tsukuyomi starts using Torment Unto Death on (\y{Name})/,
-      regexDe: / 14:2BBB:Tsukuyomi starts using Todesqualen on (\y{Name})/,
-      regexFr: / 14:2BBB:Tsukuyomi starts using Brimade Meurtrière on (\y{Name})/,
+      regex: / 14:(2BBB|2EB2):Tsukuyomi starts using Torment Unto Death on (\y{Name})/,
```
quisquous: Oh, that makes sense, thanks!

You need to make these `(?:2BBB|2EB2)`, so that it doesn't create a matching group.  The alarmText uses matches[1], which it's expecting to be the name match.

nuks-fr:
it should be ok now, thank you for noticing my error

quisquous:
Thanks again for fixing this trigger bug!

PR #182, opened 2018-07-04
first attempt

ArkfrostLumas:
Translated what was to translate there. Anything I missed?


ui/raidboss/data/triggers/ultima_weapon_ultimate.js
```diff
@@ -8,6 +8,7 @@
     {
       id: 'UWU Feather Rain',
       regex: /Feather Rain/,
+      regexDe: /Federregen/,
```
quisquous: Just FYI, timeline triggers don't need to be translated.  I'll edit this myself and then merge.

PR #190, opened 2018-07-30
util: Adding starting work on automated timeline testing

xephero:
Hello! I haven't done much in a while, but I had a spare weekend and decided to put it towards cactbot.

The goal for this is to have a system where we can point to a timeline, point to a log, and run that timeline against that log to see how it lines up. This will complement (possibly replace, but probably not completely) the old process of making a timeline, recording a vod of it, going over it later to see where it desynced, and so on. It's pretty basic right now, but it should get there eventually. Hopefully before 4.4.

Syntax is similar to make_timeline, with the added argument of a timeline filename. I ran the uwu timeline against the furthest pull i have (suppression gaol) like so:

```
test_timeline.py -f "%APPDATA%\Advanced Combat Tracker\FFXIVLogs\Network_20180727.log" -s 19:11:46.000 -e 19:25:29.442 -t ultima_weapon_ultimate > out.txt
```

and the result is here: [out.txt](https://github.com/quisquous/cactbot/files/2239934/out.txt)

Some things to note about the output, I'll probably clean some of these up eventually.

- Downburst missing at 27.0 actually happened at 29.6 or so. I might make an expanded pseudo-window in order to catch cases like that, to give better context instead of just shrugging and saying they're missing
- All the missing stuff after 100.470 (and several other places later): It synced Eye of the Storm at both 19:13:23.834 and 19:13:26.850. I keep track of the number of times each entry has been synced in order to print when something gets missed. Since it saw the same Eye of the Storm entry twice, it assumed that there'd been a loop and it simply missed all of the syncs since the target of last jump. In this case, there was no jump, so it just replayed from the beginning, as though there was a jump 0 and then fast-forwarded to the repeat Eye of the Storm. I'm sure there's a way to address this case, I just have to spend some more time thinking about it.

Let me know if there's any particular timeline-making pitfalls you run into that could be built into this thing, too! I want to get it up and running with fflogs next, even though it will have the caveats of not syncing encounter start/end (because fflogs doesn't record Heehee HAHA hahaha HEEHEE haha HEEEEEE, for example), and I'll have to re-parse out the specific format of :Actor:####: into its constituent fflogs fields.

### DIFF
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -1,0 +1,221 @@
+import argparse
+from datetime import datetime
+import fflogs
+import os
+from pathlib import Path
+import re
+
+def load_timeline(timeline):
+    """Loads a timeline file into a list of entry dicts"""
+    timelist = []
+    with timeline as file:
+        for line in file:
+            # Ignore comments, alertall, hideall, etc by
+            # only reading lines starting with a number
+            if not line[0].isdigit():
+                continue
+
+            entry = {}
+            # Remove trailing comment, if any
+            clean_line = line.split('#')[0]
+
+            # Split the line into sections
+            match = re.search(r'^(?P<time>\d+)\s+"(?P<label>[^"]+)"\s+(?P<options>.+)', clean_line)
+            if not match:
+                continue
+
+            entry['time'] = float(match.group('time'))
+            entry['label'] = match.group('label')
+
+            # Get the sync format into the file format
+            sync_match = re.search(r'sync /([^/]+)/', match.group('options'))
+            if not sync_match:
+                continue
+
+            entry['regex'] =  sync_match.group(1).replace(':', '\|')
+            entry['seen'] = 0
+
+            # Get the start and end of the sync window
+            window_match = re.search(r'window ([\d\.]+),?(\d\.]+)?', match.group('options'))
+            
+            if window_match:
+                pre_window = float(window_match.group(1))
+                if window_match.group(2):
+                    post_window = float(window_match.group(2))
+                else:
+                    post_window = pre_window
+            else:
+                pre_window = 2.5
+                post_window = 2.5
+
+            entry['start'] = max(0, entry['time'] - pre_window)
+            entry['end'] = entry['time'] + post_window
+
+            # Get the jump time, if any
+            jump_match = re.search(r'jump ([\d\.]+)', match.group('options'))
+
+            if jump_match:
+                entry['jump'] = float(jump_match.group(1))
+
+            # Add completed entry to the timelist
+            timelist.append(entry)
+
+    return timelist
+
+def parse_time(timestamp):
+    """Parses a timestamp into a datetime object"""
+    return datetime.strptime(timestamp, '%Y-%m-%dT%H:%M:%S')
+
+def parse_line_time(line):
+    """Parses the line's timestamp into a datetime object"""
+    time = parse_time(line[3:22])
+    time = time.replace(microsecond=int(line[23:29]))
+    return time
+
+def run_file(args, timelist):
+    """Runs a timeline against a specified file"""
+    last_entry = False
+    last_sync_position = 0
+    last_jump = 0
+    file_started = False
+    timeline_stopped = True
+
+    with args.file as file:
+        for line in file:
+            # Scan the file until the start timestamp
+            if not file_started and line[14:26] != args.start:
+                continue
+
+            if line[14:26] == args.end:
+                break
+
+            # We're at the start of the encounter now.
+            if not file_started:
+                file_started = True
+                last_sync_timestamp = parse_line_time(line)
+
+            # Get amount of time that's passed since last sync point
+            if timeline_stopped:
+                time_progress_seconds = 0
+            else:
+                time_progress_delta = parse_line_time(line) - last_sync_timestamp
+                time_progress_seconds = time_progress_delta.seconds + time_progress_delta.microseconds / 1000000
+
+            # Get where the timeline would be at this time
+            timeline_position = last_sync_position + time_progress_seconds
+
+            # Search timelist for matches
+            for entry in timelist:
+                if (
+                        'regex' in entry and
+                        re.search(entry['regex'], line) and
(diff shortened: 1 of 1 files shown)


xephero:
Cleaned up those issues now. The backwards window doesn't act like a loop, and now a missed sync will output when the event last occurred. This doesn't help with the initial Downburst, since the Downburst actually happens after the Great Whirlwind sync that it's missing (GW is at 29.323 and Downburst is at 29.741, in my log), but it does catch a lot of others. Here's what the output is looking like now, same input arguments as above: [out.txt](https://github.com/quisquous/cactbot/files/2243382/out.txt)

xephero:
Short parts first:
- Yeah I can change that to be relative to the file rather than cwd.
- The --report argument doesn't work yet since there isn't fflogs support yet, I just copied the arguments over and left it there since I plan on implementing it

Long part:
I do agree that the ideal is something that will look at multiple occurrences of a fight and get an average of those; something that will not just look for timeline vs one log, but something that will look at the timeline, many logs, and modify the timeline in-place to massage it to fit how the fight plays out on average. That's a longer-term thing than this though. This is basically inspired by me making the lighthouse timelines, running it, changing a half dozen things, running it again, repeating like 6x and then never wanting to run it again. It's also easier to tell from a log when an actual sync would happen than it is from a video.

All that said, I think this can go a lot further with automating these processes, but it will never be completely removed from manual effort. At the very least, a human is going to have to step in to say that Downburst should be included and Super Cyclone shouldn't, for example. We could get the most perfectly average Super Cyclone timing across thousands of logs and it would still not be useful or accurate since only we humans know that it's player-triggered and not in the encounter script.

So, long term, it should absolutely be something that takes an existing timeline, averages the timings across many runs, and updates the timeline in-place with the new data. This script is just an effort to take the lowest-hanging fruit of that task, and using that to replace the highest-effort part of timeline construction. I definitely want to head in the direction you want to, though.

The answer to the question "what's to say that this output would not look wrong when played in test_timeline against some other log" is "nothing". You could go fix the timings to be perfectly accurate to another log, and run it against another, and then things are off again. It shouldn't be off by that much, however; the whole enterprise of the ability to make timelines rests on the reliability of everything in encounters happening in the same order at the same time (excepting big stuff like branches of course), so if there's a part of the fight that just straight-up varies by 2.5+ seconds, it should probably be an active duration or something like that, but I can't think of anything all that fuzzy timing-wise when looking at a regularly synced timeline, only when dealing with things that are several minutes apart.

For now, though, the averaging process has to be done by hand. It had to be done by hand before, too, though, it just took a lot more actual content running to do multiple revisions, and with this it shouldn't.

I'll keep working on this though as I get some free time. Thank you for your feedback!

quisquous:
Sorry if my previous comment came off as negative.  This script is awesome and it's definitely a tool that's missed.  Having to run something to test timelines takes such a long time, especially when it's Lighthouse.  (I did the same thing for the math triggers /o\ it was a mess.)

I think the sad part for me is looking timelines I thought were "good enough" and then going "oh no, now I need to fix all of these" and not really having a great way to do that.  But, being able to test without running something directly is a huge step.

Thanks!  I'll wait until you add the fflogs support and then merge this, if that seems reasonable?

xephero:
No worries! Feedback is welcome, not negative. You can merge it after FFLogs support, sure.

As for the timelines, they probably are good enough; changes would just be polish. I haven't noticed any severe desyncs or anything.

xephero:
While working on FFLogs support I discovered something interesting that is a big cause of the many missed syncs.

Here are some lines from my actual .log file:
```
20|2018-07-27T19:11:51.8100000-07:00|40003C26|Garuda|2B53|Slipstream|40003C26|Garuda||46813d3246b178d23785fea285003b75
22|2018-07-27T19:11:54.3920000-07:00|40003C26|Garuda|2B53|Slipstream|E0000000||0|0|0|0|0|0|0|0|0|0|0|0|0|0|0|0||||||||||5076564|5293335|-11072|-11072|1000|1000|99.90076|99.50403|0||48d4a8704e67f76f82835a692e46f164
```
Note how both of these contain `|Garuda|2B53|`. Next, here's these corresponding lines in ACT:
```
[19:11:51.810] 14:2B53:Garuda starts using Slipstream on Garuda.
[19:11:54.392] 16:40003C26:Garuda:2B53:Slipstream:E0000000::0:0:0:0:0:0:0:0:0:0:0:0:0:0:0:0::::::::::5076564:5293335:-11072:-11072:1000:1000:99.90076:99.50403:0:
```
So, unfortunately for us, the begincast line matches in the file, but does not match in the parsed ACT version! This is pretty disappointing since it means it's not always going to be 1:1 between the line used for triggers and the line written to the file, but most of the lines we sync against are the same, so it should just be an annoyance at this point. I'm going to try to special case the begin-cast events in both reports and logs tomorrow. I was wanting to do that anyway, since begincast events can be critical to the timeline flow (remember O7S for the week or two when they were broken in ACT?). I've committed the FFLogs support, which your ultima timeline more or less passes with flying colors, and I'll work on the begincast issue soon. Totally fine if you want to hold off on merging until it's actually fixed and proper.

For reference, this is the output against the FFLogs of the same exact pull as the one I used above: [out2.txt](https://github.com/quisquous/cactbot/files/2280796/out2.txt)

xephero:
Alright, everything should be sorted. I was misremembering since O7S uses applydebuff lines rather than begincast lines to sync the branches, but I've added and tested support for both. I've tried to make it organized, should we need to use different lines in the future, rather than just baking in the idea that this is all we'll ever need, since you never know when a particular line type will have its time to shine in some phase transition down the road.

As a bonus, here's an O7S timeline showing all the syncs, including the jump on begincast: 
[out.o7s.txt](https://github.com/quisquous/cactbot/files/2281518/out.o7s.txt)

This is ready to merge now, but in the future I do want to add the ability to run against multiple fights at once and finding an average between them.

quisquous:
Aha! Excellent sleuthing work on the begincast stuff.  I feel much better that both of those timelines aren't as off as it seemed like they were.

Thanks for the great work, as always! <3

PR #207, opened 2018-09-20
util: Changing make_timeline to round to tenths of seconds

xephero:
We'd mentioned this in prior PRs, but the timing on some of the O9S stuff was frustratingly half-second in enough places that I went ahead and made the change. As expected, it makes timelines that gradually increase that decimal place, but it should be a better reflection of what's actually happening, instead of the platonic ideal of how the game is coded. It's also going to mean a lot of variance between pulls, so some care should be taken when combining data from multiple pulls to line up a common starting point; a 100.5 in one pull might well be a 99.9 in another, and both will end up with the timer looking right in-game, but it would mess with it a bit if you then take an ability at 102.5 in the first pull and put it at 102.5 in the second pull, for example.

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -129,7 +129,7 @@ def main(args):
     phases = {}
     for phase in args.phase:
         ability, time = phase.split(':')
-        phases[ability] = int(time)
+        phases[ability] = float(time)
 
     # Get the entry list
     if args.report:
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for this! I definitely feel like half seconds are a thing on some fights, so hopefully this will help have more accurate times.

PR #210, opened 2018-09-21
Adding O10S timeline

xephero:
Technically, I believe these are the flip mechanics:

Coil - first spin, either orientation. 31AC is spin, 31AD is flip
Azure Wings - second spin, either orientation. 31AE is spin, 31B0 is flip
Collectively: Flip/Spin

Azure Wings - hits under the boss into max melee range, changed to In (31B2)
Stygian Maw - hits outside the boss's targetting circle changed to Out (31B4)
Collectively: In/Out

Crimson Wings - hits corners in an X, changed to Cardinals (31B3)
Bloodied Maw - hits a cross to the boss's front, rear, and sides, changed to Corners (31B5)
Collectively: Cardinals/Corners

All final actions together collectively (only shown in the cosmetic fake future parts): Positions

I'm normally very much in favor of leaving original mechanic names in whenever possible, since I believe a timeline should show what is about to happen rather than what to do about it. In this case though, the doubling up of the name Azure Wings and the lack of cast bars makes it difficult to justify. The only time the damage mechanic names appear is when someone gets hit by it, and I don't believe the real flip/spin names are seen anywhere at all. Feel free to change flip/spin to names you prefer if you'd like, I'm not sure how clear they are.

Next, I chose Cardinals/Corners to say where the safe spot is, to continue the style of In/Out which also indicate the safe spot rather than the dangerous spot, but I'm not particularly married to the terms. It feels unintuitive for some reason, so I'm open to swapping those or even changing them for new names I haven't thought of that would better indicate, at a glance, the meaning.

The add phase doesn't really sync properly, and probably never will without an overhaul to the timeline foundations. It does seem to consistently start with Frost Breath at 209, and will do its first Rime Wreath at 215. Frost Breath has a 10 second cooldown, Rime Wreath has a 15 second cooldown. As you might have guessed, both come up around 230, then, and it seems random which one it opts to use first. I've made the consistent ones synced and the rest cosmetic for that reason. It might be a better job for triggers to handle tracking the cooldowns of each and showing Frost Breath to tanks, and both to healers when the cooldowns are nearly up.

Lastly, there's the time-based enrage. The final phase has a loop, which, in the timeline, jumps back to an earlier part. It never completely finishes the second loop, though, as it jumps to enrage at 10:30, and where in the loop you hit 10:30 is determined by your nail phase speed. It is sometimes right after the second round of Exaflare and Akh Morn, and it is sometimes after another set of spin/flip roulette. I opted to not enter a whole third branch for this, and instead reuse the first ones. This means it won't show an enrage timer when nearing enrage, until the actual cast starts, but the enrage timer wouldn't have been accurate anyway.

I've tested up through about 450.0 ingame, and I've tested the whole thing with test_timeline against both a clear log and an enrage log provided by friends. I might be able to test it tomorrow but my raid schedule is in the air right now.

### DIFF
--- a/ui/raidboss/data/timelines/o10s.txt
+++ b/ui/raidboss/data/timelines/o10s.txt
@@ -44,17 +44,8 @@ hideall "--sync--"
 186.3 "Horrid Roar" sync /:Midgardsormr:31B9:/
 
 # Add phase
-209.3 "Frost Breath" sync /:Ancient Dragon:33F1:/ # drift 0.048
-215.3 "Rime Wreath" sync /:Ancient Dragon:33F2:/ window 10,10
-219.3 "Frost Breath" sync /:Ancient Dragon:33F1:/
-229.4 "Frost Breath"
-233.4 "Rime Wreath"
-239.4 "Frost Breath"
-249.5 "Rime Wreath"
-251.5 "Frost Breath"
-261.5 "Frost Breath"
-265.5 "Rime Wreath"
-271.5 "Frost Breath"
+209.3 "Frost Breath" sync /:Ancient Dragon:33F1:/ window 2.5,30
+219.3 "Frost Breath ready"
 
 277.6 "Protostar" sync /:Midgardsormr:31C3:/ window 500,500
 
--- a/ui/raidboss/data/timelines/o10s.txt
+++ b/ui/raidboss/data/timelines/o10s.txt
@@ -1,1 +1,204 @@
 # Omega - Alphascape V2.0 (Savage) - O10S
+# -ii 31FA 3249 3630 3623 35BB 35BC 31D6 31B7 31C4 362C 321E 320D
+
+hideall "--Reset--"
+hideall "--sync--"
+
+0.0 "--Reset--" sync /Removing combatant Midgardsormr.  Max HP: \d{8}/ window 10000 jump 0
+
+0.0 "Start" sync /0044:I am Midgardsormr/ window 0,1
+2.2 "--sync--" sync /:Midgardsormr:31F9:/ window 2.5,0 # first auto
+12.1 "Flip" sync /:Midgardsormr:31AD:/
+23.1 "Spin" sync /:Midgardsormr:31AE:/
+27.1 "Cardinals" sync /:31B3:/ # X hit, not sure of sync yet
+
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/timelines/o10s.txt
```diff
@@ -1 +1,204 @@
...
+
+35.1 "Spin" sync /:Midgardsormr:31AC:/ # drift 0.029
+46.2 "Akh Morn" sync /:Midgardsormr:31AB:/
+52.3 "Spin" sync /:Midgardsormr:31AE:/
+55.4 "Out" sync /:Midgardsormr:31B2:/
+67.5 "Tail End" sync /:Midgardsormr:31AA:/
+
+75.5 "Spin" sync /:Midgardsormr:31AC:/ # drift 0.039
+86.5 "Flip/Spin" sync /:Midgardsormr:31(AE|B0)/
+89.6 "In/Out" sync /:(Midgardsormr)?:31B(2|4):/ # Sometimes missing actor name
```
quisquous: Sometimes missing actor name?!  What in the world.

quisquous:
Ohmygosh, thanks again for all your hard work.

I agree with your tendency here to want to use the right ability names, but I think the right names are just confusing here.  In/Out seems great to me.  That's certainly what our group called it.  Cardinals/Corners also seems good.

Thanks for all the ability ids.  I looked at logs a little but didn't have time to find all the ids to write triggers for in/out/cardinal/corner calls, but those will help a ton.

Do you think it's worth putting in Flip, Spin, Flip/Spin at all? Part of me wonders if it's redundant to have Flip/Spin right next to In/Out.  I also wonder for the tell whether it's a cardinals/corners vs in/out if it'd be better to just drop those in general (except for the one where it branches)?  I can go either way here, but mostly just curious if you feel like it's noisy vs helpful.

Re: nail dragon.  The thing that's rough when tanking/healing seems to be the Frost Breath.  I wonder if it'd be worth just syncing the Frost Breath always (and maybe at the 230 having a "Frost Breath?".  I'd be fine with dropping all the Rime Wreaths if it made the Frost Breaths sync ok.

Enrage not being precise seems fine.  This was also true for things like o2s, and there's not really that much you can do about it without a timeline overhaul.

xephero:
Added in and tested the Frost Breath syncs via script.

PR #213, opened 2018-09-21
changes to o9s adding triggers initial

unfaiyted:
Hey, not sure I did this the way you would do it or if the call out or code follows any specific format that you would like it to be in but this may/may not help.

Thank you

### DIFF
--- a/ui/raidboss/data/triggers/o9s.js
+++ b/ui/raidboss/data/triggers/o9s.js
@@ -1,9 +1,140 @@
 'use strict';
 
-// O9S - Alphascape 1.0 Savage
+/* O9S - Alphascape 1.0 Savage*/
 [{
-  zoneRegex: /^Alphascape V1.0 \(Savage\)$/,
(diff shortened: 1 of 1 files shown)


unfaiyted:
Oh, I guess I didnt know about these checks you have or your coding formats....I will try and fix this stuff so it passes.

JoeKer1:
I had to take the Accretion callout out of this because it called it 4x in a row :)

unfaiyted:
Ahh, I’ll take a look at that next time my group runs it or take a look at the logs

quisquous:
Thanks for all these! In the future, can you use separate pull requests for separate fights.  I was going to go land the o9s stuff, but you added o10s stuff that fails the lint check.

unfaiyted:
Hey, yeah I actually didn't mean to do that!

unfaiyted:
No problem. I’m going to be running o10s tomorrow and will try to test those and push them if they are good too.

PR #215, opened 2018-09-22
raidboss: Adding Suzaku EX timeline

xephero:
Tested this on four runs and it worked well, added in the targetable/untargetable lines afterwards from logs and I don't imagine that will break anything.

32DD and 32DE are both called Phantom Flurry; the first is the rapid tank hits and the second is the half-arena hit. I've renamed the second one to Phantom Half for clarity. 322E is the boss jump to center, but I've opted to keep it as invisible --sync-- lines since she doesn't actually drop target in that time (bless). 

I've opted to show the 10 Scarlet Hymns (DDR) individually, since it's really the only thing going on, but there's a case to make them two separate durations, for the first 3 and last 7, which come in 1 second intervals with 5 seconds between the two groups. It's debatable. The Hotspots (simon says) are durations since there actually is other stuff you care about happening in the midst of them.

### DIFF
--- a/ui/raidboss/data/manifest.txt
+++ b/ui/raidboss/data/manifest.txt
@@ -19,6 +19,7 @@ timelines/sephirot-ex.txt
 timelines/shinryu-ex.txt
 timelines/sophia-ex.txt
 timelines/susano-ex.txt
+timelines/suzaku-ex.txt
 timelines/tsukuyomi-ex.txt
 timelines/o1s.txt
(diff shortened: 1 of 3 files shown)


quisquous:
This all looks reasonable to me.  Thanks for all your thoughtfulness in how to craft good timeslines.  :+1:

Oops, didn't mean to leave the above as a commit comment.  Resposting here.

PR #218, opened 2018-09-24
Timeline fixes

xephero:
I had an oversight when I changed make_timeline to use tenths of seconds and noticed inaccuracies when I would run test_timeline against the same exact timeline used to generate them. I sort of shrugged it off until now but I dug in and found the issue. I corrected the Suzaku one with manual adjustments, but 9 and 10 were still affected; these commits fix both the tools and the timelines. Besides the timings, I think the only change is adding un/targetable lines in 10, and some extra info around Protostar. Hopefully this will make the rest of the timelines a little smoother to produce; I'm going to work on 11 and 12 (which I haven't actually entered yet myself) tonight.


quisquous:
Thanks for all of this!

PR #224, opened 2018-09-24
raidboss: Adding Omega-M/F timeline

xephero:
Not only untested in-game, I don't even really know the fight yet! This tests correctly against several fflogs reports along both branches, but in-game viability is a mystery. Will probably need some double-checking later for things that are important in the fight; I think the big eye moves around the outside and we'd probably want a timeline entry for that, for example, but I have no idea what cast it might be. There's quite a few empty-name casts in the fight (formerly Unknown_#### before the xivdb->xivapi changeover), so any one of those could end up being important. We'll see!

I also added just the first minute or so of the other O12S fight as a placeholder.


quisquous:
Sounds good.  Let's give this a go, and I'll put a warning that this is untested in the release notes, and maybe we can get some feedback tomorrow.

PR #226, opened 2018-09-25
Starting FR translation for o9s/o10n/suzaku-ex. Needs to be tested.

mooondark:


mooondark:
Well if you can help me to fix this, I'm not a developper :)
I tested it yesterday night, it worked fine so why the build fails ?

PR #239, opened 2018-09-27
O9s triggers

unfaiyted:
Added some logic for phase/path tracking in here.


ui/raidboss/data/triggers/o9s.js
```diff
@@ -111,6 +111,10 @@
     {
       id: 'O9S Longitudinal Implosion',
       regex: /14:3172:Chaos starts using Longitudinal Implosion/,
+      run: function(data) {
```
quisquous: Just because some people disable triggers, can you put side effect logic in a separate trigger, rather than the one that calls out sides -> front/back?

ui/raidboss/data/triggers/o9s.js
```diff
@@ -170,8 +178,70 @@
         };
       },
     },
+    {
+      id: 'O9S Orbs Fiendish Phase Tracker',
+      regex: /14:317(C|D):Chaos starts using Fiendish Orbs/,
+      // condition: function(data) {
```
quisquous: Similar to my comments on the long/lat trigger, can you make a separate Phase Tracker trigger so that the alarmText for tanks picking up orbs can be its own trigger?

unfaiyted: Hey thanks for the thoughts yeah I can do that going forward as well and make these adjustment

quisquous:
Sorry, I'm still confused about the initial phase trigger.

It looks like your patch is:
long => fire
lat => water

And I think it's supposed to be?
long => water
lat => water
damning edict => fire

Am I mistaken?

quisquous:
My understanding of what the timeline file is and what it should be is what I listed in my previous comment above: https://github.com/quisquous/cactbot/pull/239#issuecomment-426904086

unfaiyted:
Ok no worries. I might be confused on that I  suppose

PR #240, opened 2018-09-27
Group triggers

unfaiyted:
Hey!

I was able to figure out how things worked and adjust within your current priority system. I tested this and was able to get it working within your current priority system. 

I believe additional documentation would also be updated in this case to reflect the new option within the current system. 

Thanks!


quisquous:
Thanks for this pull request! This doesn't look like too much.  I thought on this some more, and I think this all looks good, but it's hard for me to quite know what the extent of the remaining changes is.

Is your use case that you turn off all audio (sounds and tts) and just want group tts?

I'm trying to think about how to classify triggers, and there's:
(1) things that apply to everybody (larboard/starboard).  tts = groupTTS here.
(2) things that apply to a role (tank buster).  Maybe groupTTS = tts? Some tts is individualized, but not all.
(3) things that apply to an individual (head markers).  Here, tts never is groupTTS? You'd want to be able to disable tts?

Am I missing a case here? How many triggers do you expect to need to touch after adding groupTTS support?

Also, I don't want to have to duplicate the whole trigger block for the (1) case above.  Do you think it would make sense to have like a `triggerAppliesToGroup: true` field which would imply that groupTTS was identical to tts in that case?

unfaiyted:
Well I kind of see a degradation pattern.

So it looks something like:
 IF groupTTS is set AND we have groupTTS turned on then the group specific TTS will override everything else.

If groupTTS is turned off it will ignore it and play it as a single user settings concentric. 

It degrades down like it currently does depending on the users settings, so if group is turned on but not enabled on that trigger it will default to standard tts. 

So no existing settings would really be impacted going backward, but going forward you could design them to better adjust for groups and individuals.

ui/raidboss/popup-text.js
```diff
@@ -396,6 +404,7 @@ class PopupText {
       // * user disabled (play nothing)
       // * if tts options are enabled globally or for this trigger:
       //   * user trigger tts override
+      //   * groupTTS entries in the trigger
```
quisquous: I think this ordering is not correct based on what you have below.  I think groupTTS > user trigger tts override (triggerOptions.TTS).

Can you also list user trigger group tts override here in this list, since that's also a thing you added?

quisquous:
Ah, ok.  That makes sense.  Do you not end up needing to disable tts on triggers individually because you just disable tts overall so that only the group tts plays?

ui/raidboss/popup-text.js
```diff
@@ -414,13 +423,24 @@ class PopupText {
...
-      if ('TTSText' in triggerOptions)
-        ttsText = ValueOrFunction(triggerOptions.TTSText);
-      else if ('tts' in trigger)
-        ttsText = ValueOrFunction(trigger.tts);
-      else
-        ttsText = defaultTTSText;
 
+
+      if (playGroupSpeech) {
+        console.log(trigger.groupTTS);
```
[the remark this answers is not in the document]
unfaiyted: Fixed! thanks for the review!

ui/raidboss/popup-text.js
quisquous: Thanks for the clarification.  I think this all seems reasonable to me.

PR #243, opened 2018-09-28
O10s extended

unfaiyted:
I made some changes to this o10s triggers set, should adress some of the TODO's listed. as well


unfaiyted:
For some reason it keeps saying I deleted some files from here that I didn't so going to resolve that and resubmit.

PR #248, opened 2018-10-05
O11s

unfaiyted:
Working on creating various o11s triggers. I tested most of these yesterday and adjusted the callouts to make more sense over TTS. My raid group just starting progging this fight. Let me know, thoughts.


quisquous:
Hmm.  It seems a little bit like every pull request you do includes every change you've ever made.

Forgive me if you already know all of this, but one thing you might consider is that instead of merging in upstream quisquous/master to your local branch, to just `git reset --hard quisquous/master`.  When I'm working and writing forks, I usually have master be exactly the upstream master.  Then I create individual feature branches that track that (which it looks like you're doing) and submit pull requests for those.  And then after those land, I update master again to be exactly the upstream master.  I don't merge into my local master, I just fast forward.

unfaiyted:
Yeah It got really dumb on my end and sorry for that. Trying to avoid that if possible. I am working out of branches, but I havent really done a lot of stuff with pull request or working with other peoples projects like this so I am trying to branch out some and could use any tips or feedback you have! Sorry if its troublesome on your end too, we dont really use git at work and I use it for personal stuff but mostly just random projects

unfaiyted:
Hey thanks for the commands and such! I'm going to definitely use that and see how that works out for me hopefully less issues.

PR #249, opened 2018-10-05
O11s only

unfaiyted:
Hello again, 

I made updates to these triggers here, hopefully someday I get better with this merge stuff.
I dont know if you have already done some of this but at least wanted to make sure you knew it was out there so you don't waste your time. 

We have tested the callouts and I made some modifications to include the groupTTS and to make the TTS callouts make more sense.

Thank you again!


ui/raidboss/data/timelines/o11s.txt
```diff
@@ -16,7 +16,7 @@ hideall "--sync--"
 31.1 "Afterburner" sync /:Omega:325E:/
 38.6 "Starboard/Larboard Cannon" sync /:Omega:326[24]:/
 44.1 "Starboard/Larboard Cannon" sync /:Omega:326[35]:/
-55.6 "Peripheral Synthesis" sync /:Omega:324A:/
+55.6 "Peripheral Synthesis A" sync /:Omega:324A:/
```
quisquous: I think it'd be clearer if this was small, big, red instead of A, B, C, maybe?

ui/raidboss/data/timelines/o11s.txt
```diff
@@ -67,15 +67,15 @@ hideall "--sync--"
 480.0 "Starboard/Larboard Surge" sync /:Omega:326[79]:/
 
 496.0 "Pantokrator 1" sync /:Omega:3702:/
-500.0 "Ballistic Impact" duration 8
+500.0 "Engage Ballistic Systems" duration 8
```
quisquous: Why did you want to change this?

ui/raidboss/data/triggers/o11s.js
```diff
@@ -1,13 +1,63 @@
...
 [{
   zoneRegex: /^Alphascape V3.0 \(Savage\)$/,
   timelineFile: 'o11s.txt',
+  timelineTriggers: [
+    {
+      id: 'O11S Reset',
+      regex: /Reset/,
+      beforeSeconds: 6,
+      alertText: {
+        en: 'get out, center aoe',
```
quisquous: Personally I don't think there's any reason to callout obvious aoes on the ground that show up.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -1,13 +1,63 @@
...
+      },
+    },
+    {
+      id: 'O11S Reset',
+      regex: /Reformat/,
+      beforeSeconds: 6,
+      alertText: {
+        en: 'donut aoe',
```
quisquous: Same thing here.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -1,13 +1,63 @@
...
+      groupTTS: {
+        en: 'Tank LB',
+      },
+    },
+    {
+      id: 'O11S Ballistic Puddles',
+      regex: /Ballistic Impact/,
+      beforeSeconds: 7,
+      infoText: {
+        en: 'position for puddles',
```
quisquous: I think this is not worth a callout either.  Especially for melee, there's no pre-positioning.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -1,13 +1,63 @@
...
+      infoText: {
+        en: 'party stack',
+      },
+    },
+  ],
   triggers: [
     {
       id: 'O11S Mustard Bomb',
-      regex: / 14:326D:Omega starts using Mustard Bomb on (\y{Name})/,
+      regex: /14:326D:Omega starts using Mustard Bomb on (\y{Name})./,
```
quisquous: I think this works with the space.  Also, if you're going to add a `.` it should be a `\.` because it's regex.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -33,6 +83,29 @@
...
+        };
+      },
+    },
+    {
+      id: 'O11S Atomic Ray',
+      regex: / 14:326C:Omega readies Atomic Ray,/,
+      tts: function(data) {
+        if (data.role == 'healer') {
+          return {
+            en: 'AOE',
```
quisquous: In general, aoes are not things that I think need triggers.  For small aoes like atomic ray, folks will live through regardless.  Also, cast bars generally only last one gcd, so it's not like you can do much about it if you aren't prepped or don't want to burn a swift.  This is as compared to a tank buster where yeah optimal play involves hitting your cooldowns as far ahead as possible (usually), but if you forget, you can still slam two or three cooldowns in before the tank buster hits you.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -85,11 +158,11 @@
       alertText: function(data) {
         if (data.lastWasStarboard) {
           return {
-            en: 'Move (Left)',
+            en: 'Move',
```
quisquous: Personally, I like the "left" in here.  I feel like folks have been very split on what they want here, whether they want "move" or "left" called out, and so I really like having both.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -139,6 +225,116 @@
...
         en: 'Opposite (Right)',
       },
+      groupTTS: {
+        en: 'opposite right',
+      },
+    },
+    // End Starboard / Laboard Stuff
+    {
+      id: 'O11S Flamethrower',
+      regex: /00:282b:Omega readies Flamethrower./,
```
quisquous: I don't think this needs a callout.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -139,6 +225,116 @@
...
+        en: 'Stack',
+      },
+    },
+    {
+      id: 'O11S Blaster',
+      regex: /:3261:/,
+      tts: function(data, matches) {
+        if (data.role == 'healer') {
+          return {
+            en: 'Teather on ' + data.ShortName(matches[1]),
```
quisquous: There's no match in the regex here? Did you mean to include this?

ui/raidboss/data/triggers/o11s.js
```diff
@@ -139,6 +225,116 @@
...
+      groupTTS: {
+        en: 'different out, same in',
+      },
+    },
+    // [20:26:01.568] 1B:105A629D:Person Name:0000:0000:00A3:0000:0000:0000: -
+    // :00A2: Positive // useful at all?
+    // Is it possible to see who is matching who?
+    {
+      id: 'O11S Pantokrator',
+      regex: /:3702:Omega readies Pantokrator/,
```
quisquous: I'm not sure this is that useful.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -139,6 +225,116 @@
...
+    {
+      id: 'O11S Peripheral Synthesis',
+      regex: /14:324A:Omega starts using Peripheral Synthesis/,
+      preRun: function(data) {
+        data.synthCount = (data.synthCount || 0) + 1;
+      },
+      infoText: function(data) {
+        if (data.role == 'tank') {
+          return {
+            en: 'Tank LB',
```
quisquous: I don't think tank LB is really the optimal strat here, so I don't want to bake that into the triggers.

My general philosophy on triggers is to tell the player what to do, if there's only one way to do it (e.g. go left for starboard).  However, if there's not an obvious thing to do, then to tell you what you have so you can make that decision (e.g. you get the twin dive during grand octet, but it's not going to tell you where to go).

I think cactbot is probably already chattier than most folks want, so I don't want to take it any further in that direction.  I think if you want more chatty triggers for groupTTS for aoes and match yellow blue and to essentially replace somebody doing a raid calling monologue, then maybe those are things to add to your personal set of triggers.

ui/raidboss/data/triggers/o11s.js
```diff
@@ -139,6 +225,116 @@
...
+          return { en: 'TANK LB' };
+      },
+    },
+    {
+      id: 'O11S Charybdis',
+      regex: /:326E:/,
+      alertText: function(data) {
+        if (data.role == 'healer') {
+          return {
+            en: '1 HP ',
```
quisquous: This is another example of something that I think doesn't need a callout.

ui/raidboss/data/triggers/o10s.js
```diff
@@ -144,5 +170,117 @@
...
+      id: 'O10S Crumbling Bulwark',
+      regex: ':XXXX:Crumbling Bulwark',
+      tts: function(data) {
+        return {
+          en: 'Bulwark Crumbling',
+        };
+      },
+    },
+    {
+      id: 'O10S Pass Bulwark',
```
quisquous: Did you mean to include all this o10s stuff here?

ui/raidboss/data/triggers/o11s.js
```diff
@@ -1,13 +1,63 @@
+
```
quisquous: Can you not add this line here?

unfaiyted:
Thanks for all the comments and review of everything.! If you find any of it useful feel free to implement it anyway you want.  I’m going to go ahead and close this. I was hoping this would be helpful but it maybe just too far separated from what works best for you. Hope I provided some value.

PR #251, opened 2018-10-06
FR Translation (O11n)

mooondark:
Translation is done. Timeline is working fine but none of the triggers works.
I've double checked the logs, everything seems to be good ... -> forgot to change boss name


ui/raidboss/data/triggers/o11n.js
```diff
@@ -8,19 +8,22 @@
     {
       id: 'O11N Blaster',
       regex: /Blaster/,
+      regexFr: /Électrochoc/,
```
quisquous: You don't need to translate the regex for timeline triggers.  They're all based on the English timeline, even if the timeline is translated.

PR #255, opened 2018-10-09
What do we have here?

ArkfrostLumas:
There is only 1 line where I didn't know what to do with it

### DIFF
--- a/ui/raidboss/data/triggers/o9n.js
+++ b/ui/raidboss/data/triggers/o9n.js
@@ -15,14 +15,14 @@
         if (matches[1] == data.me) {
           return {
             en: 'Tank Buster on YOU',
-            de: 'Tenkbuster auf DIR',
+            de: 'Tankbuster auf DIR',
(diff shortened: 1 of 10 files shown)


ui/raidboss/data/triggers/o9n.js
```diff
@@ -92,9 +91,9 @@
         'Stray Spray': 'Chaosspritzer',
         'Tsunami': 'Tsunami',
         'Umbra Smash': 'Schattenschlag',
+        'Long/Lat Implosion': 'Horizontale/Vertikale Implosion',
 
         // FIXME
-        'Long/Lat Implosion': 'Hz/Vert Implosion',
         '\\(ALL\\)': '(ALL)',
```
ArkfrostLumas: This line. I've got no clue what to do with that or how to fix that.

quisquous: The "ALL" here is just meant to be "everybody" as compared to "T/H" or "DPS" for explosions.

PR #266, opened 2018-10-24
Update o10s.js

jerrybotticelli:
Changing the trigger for the fire marker. Wouldn't the original way never show the infoText because of the condition method being used?


quisquous:
Whoops!! Thanks for this fix.

PR #276, opened 2018-11-13
German Update

ArkfrostLumas:
Updated Pagos Mob Names (didn't see they were there in the first place), did some wording-tweaks in a few triggers and updated what was there to update :>

### DIFF
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -491,13 +491,13 @@ let Options = {
         snowqueen: {
           label: {
             en: 'Queen',
-            de: 'Snow Queen',
+            de: 'Schneekönigin',
             fr: 'Snow Queen',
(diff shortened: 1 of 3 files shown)


ui/eureka/eureka.js
```diff
@@ -714,13 +714,13 @@ let Options = {
...
             en: 'Minotaurs',
-            de: 'Minotaurs',
+            de: 'Minotauren',
             fr: 'Minotaurs',
             ja: 'Minotaurs',
           },
           mobName: {
             en: 'Mindertaur',
-            de: 'Mindertaur',
+            de: 'niederer Minotaurus',
```
quisquous: This is really lower case in game?

quisquous:
Thanks as always!

PR #279, opened 2018-11-15
Raidboss emulator

DrLippe:
This neat little thing can trick the raidboss overlay to run a combat log without having to be in the actual fight.
Just replay a once happened fight and test your trigger/timeline files 😄 

For now the module has to be run through ACT with the game running in the background since browsers wont allow the emulator to read local files.
⚠️ This overlay also works as an usual raidboss overlay
⚠️ The overlay is hidden/shown through the eye handle in the upper right corner
The format of your log file should have the zone name to be emulated in the 1st line and the log copied from ACT below

please enjoy it as much as i do


DrLippe:
demo

quisquous:
This is SUPER cool.  Thank you for this.

One feature I think would make sense would be to add an input text box for the zone (maybe we could have a dropdown), and possibly even for the current time to start, and then you'd be able to just point it at a Network_ log file.  If I were dreaming, I'd also want to be able to change the name and job to play triggers for, so I could test job-specific things.

How are you loading the file even within act? I'm confused why the browser doesn't give you a hard time there for loading local text files.

quisquous:
An alternative to trying to make a standalone solution would be to lean on ACT more.  I think when you import a Network log file, it does most of the things that need to get done (changing the zone, all the log messages).  All of those messages are currently ignored (the isImport check in CactbotOverlay.OnLogLineRead), but there's no reason we couldn't have an overlay type that behaved differently and *only* took the isImport lines.

I could imagine the user path for something like that being: open ACT, import old network file into ACT, overlay figures out where encounters stop and start, and can present the user with a drop down of zones, start times, and pull durations much like fflogs does when you upload logs.  Then there's no need to edit any files directly.

DrLippe:
I'm planning to have both an ACT version and a standalone one. My plan is to import a network log file into the emulator and then be able to create and test triggers inside the emulator. I just need to figure out how to translate the network log file to a FFXIV plugin log file. The import method should bypass this problem for the ACT one.
skizze

quisquous:
This is looking quite fancy!

Showing all the triggers is cool.  I could imagine clicking on one and then letting you skip ahead to see where those triggers hit in the timeline, and have raidboss show you what it looks like when they're hit.

Definitely agree that being able to switch who you are is important.

I looked at the importing step a good bit, and my conclusion is that loading a network log file directly is tricky, as there's too many things that need to be filtered, processed, and combined.  This is a lot of the work that ACT does itself.  So, I think either having a combat log or importing through ACT are the only reasonable ways to go.  I personally would be fine with just having an ACT version.  If anybody else wanted me to debug triggers with this, I would probably have them send me the network log.

I think it's going to be tricky to analyze triggers by role, since that's not something that you can find from data, only from actually running Javascript.  It wouldn't be impossible, but it's definitely not straightforward.

I am also a little skeptical about writing triggers from this UI.  I don't think this would be a workflow that I would use.  I could imagine something where I'm writing a trigger in some other program, and then I come back to the emulator, I hit some Reload Triggers button, and then test it.  But actually writing or creating the trigger there is not something I would plan on doing.  That seems like you would have to write an entire editor there to make that reasonable, unless I'm misunderstanding what you're getting at?

Finally, I looked into this a bit and got a working version with import logs, along with an updated C# plugin.  This ended up rewriting most of the JS code though, so I'm not quite sure how best to integrate this with what you're doing.  I'll upload a branch and copy a build here.  You can try it out, and then we can discuss and go from there.

quisquous:
Commit: https://github.com/quisquous/cactbot/commit/d72fb85a30aaa7f94eb9465ce1e3e862c3389cd1

[emulatorv2.zip](https://github.com/quisquous/cactbot/files/2682966/emulatorv2.zip)

DrLippe:
I got the analyze thing to work by testing for any output of a trigger to a given log line, works great on small scale and soon to be tested on a big log file.
Got a lot of working to do so progression on this might take a while.

quisquous:
No worries!  I'll be excited to see more progress on this whenever it gets done.  :)

DrLippe:
Here is an example of the trigger analysis with O9S
anmerkung 2018-12-23 181523
I have the analyzing `function` suppress all Triggers for 1s if nothing else is declared by the trigger itself for clarity reasons.
I am using an `await` inside of a `for` loop to make sure the triggers get the right `data` for things like phase tracking. Might not be the best idea but it made it work like intended.
[analyzePrototype.zip](https://github.com/quisquous/cactbot/files/2706190/analyzePrototype.zip)

DrLippe:
So I finally checked out your changes to the initial emulator and after days of wondering what the heck is going on, I was able to merge my analyzing tool into your version of the emulator.

⚠️ This will need the updated cactbot.dll

This version will analyze logs per role, if all roles trigger on a log line, the trigger will be listed separately (yellow triggers). Trigger delays and suppressions will be taken into account. When hovering over a trigger, the overlay will display its id string. The job icons serve as player changer. 
Imported fights and the fight selection are stored in localStorage and restored after reloading the overlay so you can just hit that reload button as often as you like without regretting. Restoring comes with a small delay though, to wait for the raidboss overlay to properly load.

[emulatorv3.zip](https://github.com/quisquous/cactbot/files/2716874/emulatorv3.zip) [contains only raidemulator html, js, css and resource image files]

I'm neglecting raiding for this... so I'm still on my aged O9S log 😅

newemulator

Give me your thoughts about this version.

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+                id: playerId,
+              },
+            });
+          }
+        }
+      }
+
+      // This can speed up this function a lot!
+      // Add logic to differ between 4 and 8 player dungeons
+      // TODO: Find a better Way of doing this
```
quisquous: This is an oddball thought, but when walking through all the logs to figure out where the fights are, you could figure out all the people that were hit by protect and use that to figure out party members in the zone before you even got to this point.

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+    document.dispatchEvent(evt);
+  }
+
+  SendZoneEvent(zoneName) {
+    let evt = new CustomEvent('onZoneChangedEvent', { detail: { zoneName: zoneName } });
+    document.dispatchEvent(evt);
+  }
+
+  SendPlayerEvent(name, job, id) {
+    let jobDetail = {
```
quisquous: Do you need to send all of this?

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+      maxCP: 1000,
+      currentGP: 1000,
+      maxGP: 1000,
+      level: 100,
+      debugJob: 'debug',
+      jobDetail: jobDetail,
+      pos: { x: 0, y: 0, z: 0 },
+    } });
+    document.dispatchEvent(evt);
+    document.getElementById('player').textContent = name;
```
quisquous: From an abstraction point of view, it'd be nice if the LogPlayer didn't mess with the dom at all and the EmulatorView did all of that.  That's part of what I was trying to get at with reorganizing here, i.e. a LogCollector that only just handled importing logs and breaking them up, a LogPlayer that handles interfacing with the underlying raidboss, sending events, handling time, and finally an EmulatorView that messes with the dom and shows things.

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+      this.logPlayer.SendZoneEvent(fight.zoneName);
+      this.ShowFightInfo(this.selectedFight);
+
+      // Simply push first Player
+      let name = this.sortedPlayers[0].name;
+      let job = this.sortedPlayers[0].job;
+      let id = this.sortedPlayers[0].id;
+      this.logPlayer.SendPlayerEvent(name, job, id);
+
+      this.Analyze();
```
quisquous: I think this is extremely slow to run this inside of SelectFight.  I think if you want to do the analysis synchronously, it would be better to do that when you hit play, and not when you just click on a fight in the list.  I think you could maybe make this asynchronous if you divided up the work to do and did it in a series of callbacks, but that's definitely more complicated.

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+  }
+
+  StoreFights(fights) {
+    localStorage.clear();
+    localStorage.setItem('fights', JSON.stringify(fights));
+  }
+
+  RestoreOldFights() {
+    if (localStorage.fights)
+      this.fights = JSON.parse(localStorage.getItem('fights'));
```
quisquous: If I try to even click on a ucob fight from my log, I get this error: `12/30/2018 10:30:25 PM: Info: emulator: BrowserConsole: Uncaught QuotaExceededError: Failed to execute 'setItem' on 'Storage': Setting the value of 'fights' exceeded the quota. (Source: file:///C:/Users/enne/cactbot/ui/raidboss/raidemulator.js, Line: 41)`

I agree that it's a huge pain to have to keep importing, but I'm also skeptical about this storage method given these errors.

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+        Object.assign(players, {
+          [p]: {
+            name: this.players[p].name,
+            job: this.players[p].job,
+            role: this.players[p].role,
+          },
+        });
+      }
+    }
+    if (role == 'all')
```
quisquous: This is just a bug report from me, but looking at an o9s log from this last week, I found that it was ignoring all the "stack donut" and "get behind" all triggers in the analyzer.  I can send you the log file if that would help.

quisquous:
[Network_20181226.zip](https://github.com/quisquous/cactbot/files/2717862/Network_20181226.zip)

Here is a network log for you for testing.  On the O9S clear, I was surprised that the stack donut callouts weren't in the "all" track.  Also, the ucob pulls all give local storage errors when you try to analyze them.  Hopefully this helps!

ui/raidboss/raidemulator.css
```diff
@@ -0,0 +1,199 @@
...
+  overflow: hidden;
+  white-space: nowrap;
+}
+/* Item Positioning */
+.emulator-background {
+  position: absolute;
+  bottom: 0px; left: 2.5%;
+  height: 320px; width: 95%;
+  z-index: -1;
+  background: url(res/img/dark-bg-220.png) repeat;
```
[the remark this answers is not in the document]
DrLippe: paint.net 😄
```
  background color: rbg(255,255,255) 
  effect: adding noise (100, 400, 400) x2
  adjustment: black and white
  adjustment: level (input: 255, 0; output: 33, 1.00, 25)
```
going to replace it with a smaller size one

ui/raidboss/raidemulator.js
DrLippe: In an earlier version of my emulator I had the idea of having a party menu like the game itself showing HP etc. I will delete this on my next commit along with some other stuff.

ui/raidboss/raidemulator.js
DrLippe: First thing I thought of were cookies but since they are disabled I went for the localStorage which is pretty limited. Maybe it would help to store just the selected fight?

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,832 @@
...
+      info += '<div class="nowrap">' + this.DateToTimeStr(fight.startDate) + ' - ' + this.DateToTimeStr(fight.endDate) + (isClear ? ' (Clear)' : ' (Wipe?)') + '</div><div class="subtitle">Time</div>';
+      info += '<div class="actors">' + actors.join(', ') + '</div><div class="subtitle">Actors</div>';
+
+      fight.info = info;
+    }
+    this.infoElement.innerHTML = fight.info;
+  }
+
+  // Lippe functions
+  SearchPlayers(logs) {
```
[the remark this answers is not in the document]
DrLippe: Healers might never use the skill that is set as signature move. So you have to iterate over moves anyway.
This approach will break on the first occasion to detect a player and won't check a player twice for its role and will terminate as soon as the party limit is reached.
This function does take 2-4ms with party limit and 20-30ms without party limit to finish.

ui/raidboss/raidemulator.js
DrLippe: Maybe the SearchPlayers function could be integrated into the LogCollector to prevent iterating over the same data again and again. I will try that!

ui/raidboss/raidemulator.js
DrLippe: Argh... no idea why it keeps doing that sometimes... does the fight start with water phase?
It may be some issue with the analyzer being faster than the raidboss module does load the triggers.
But honestly? No idea. This bug is driving me nuts.

DrLippe:
That helped a lot!
I tried to use RestoreFights as callback function but that didnt work. I had to add an additional 1s timeout to RestoreFights because ACT will send a zone change while RestoreFights is working resulting in wrong triggers being loaded.
Also i wonder why the import only works if the game is running in the background?

New: You can now store/remove selected fights until your localStorage is full.
Next: Working on the progress bar(s) for triggerlines.

ui/raidboss/raidemulator.js
```diff
@@ -0,0 +1,822 @@
...
+      fightListElement, timerElements, elapsedElement, infoElement,
+      partyElement, currentPlayerElement, triggerInfoElement
+  );
+  gLogCollector = new LogCollector(gEmulatorView.AddFight.bind(gEmulatorView));
+
+  document.addEventListener('onPlayerChangedEvent', function(e) {
+    document.getElementById('player').textContent = e.detail.name;
+  });
+
+  // Timeout is required because the overlay is slower and will break the Zone
```
quisquous: What does RestoreFights do that depends on the zone? If you remove the "SelectFight" part of RestoreFights, would it do the right thing?

quisquous:
Here's a quick patch to include a file I added, as well as to handle Orbonne via zone sealing messages.  I think it could probably be a little bit better (maybe at least show what it's sealing in the ui somewhere?), but at least this separates out by fights.

This emulator was *amazing* in getting all the Seiryu and Orbonne triggers done.  I wrote everything, tested a bunch, and then only had to go run each of them once to feel like I got everything ironed out.  I love this emulator so much!

```
diff --git a/ui/raidboss/raidemulator.html b/ui/raidboss/raidemulator.html
index d2d816a..40b378f 100644
--- a/ui/raidboss/raidemulator.html
+++ b/ui/raidboss/raidemulator.html
@@ -5,6 +5,7 @@
   <!-- General Overlay Scripts-->
   <link rel="stylesheet" href="../../resources/defaults.css">
   <link rel="stylesheet" href="../../resources/resize_handle.css">
+  <script src="../../resources/common.js"></script>
   <script src="../../resources/regexes.js"></script>
   <script src="../../resources/resize_handle.js"></script>
   <script src="../../resources/timerbar.js"></script>
diff --git a/ui/raidboss/raidemulator.js b/ui/raidboss/raidemulator.js
index deb5328..71ab81e 100644
--- a/ui/raidboss/raidemulator.js
+++ b/ui/raidboss/raidemulator.js
@@ -18,6 +18,8 @@ class LogCollector {
   AppendImportLogs(logs) {
     // Define now to save time later
     let playerRegex = Regexes.Parse(/ 15:(\y{ObjectId}):(\y{Name}):(\y{AbilityCode}):/);
+    let sealRegex = Regexes.Parse(/ 00:0839:.* will be sealed off in /);
+    let unsealRegex = Regexes.Parse(/ 00:0839:.* is no longer sealed/);
 
     for (let i = 0; i < logs.length; ++i) {
       let log = logs[i];
@@ -36,9 +38,9 @@ class LogCollector {
         if (matches)
           this.SearchPlayers(matches);
 
-        if (this.IsWipe(log))
+        if (this.IsWipe(log) || log.match(unsealRegex))
           this.EndFight(log);
-      } else if (log.match(gLang.countdownEngageRegex())) {
+      } else if (log.match(gLang.countdownEngageRegex()) || log.match(sealRegex)) {
         // For consistency, only start fights on a countdown.  This
         // makes it easy to know where to start all fights (vs
         // reading timeline files or some such).
```

ui/raidboss/raidemulator.js
DrLippe: Actually only SelectFight does depend on the zone.
The SelectFight part happens to be in RestoreFights to keep this tool as comfortable as possible while editing trigger files. In my opinion the SelectFight part is a great part of RestoreFights but isn't vital.

quisquous:
Ah, ok.  I just feel like that triggerline color was too subtle for me to notice it was happening at all.  Thanks for pointing it out!

I think in terms of other minor feature requests, I might suggest:
* I think you're missing timeline triggers on the triggerline
* Triggers that don't have an id say "undefined"
* Triggers that run but don't say anything still show up (maybe this is ok?)
* It might be nice to have a personal triggerline for whoever you are selected, so you could be like "ok, did all these triggers fire for X person like I hoped" without playing through the whole thing

Don't feel like you need to do any of those, but just listing things I thought might be useful changes.  If you want to get this to a place where we can merge it in and skip any extra features, that also sounds great.  :+1:

DrLippe:
- Yes, timelines are not monitored by the analyzer yet. I'm not sure how to do this the best way. This might take a lot more time as timelines are currently completely ignored by the emulator.
- I thought showing the id of the trigger might be the easiest way to identify what trigger is actually triggered. I could either add an output text or show output if no id is defined. Minor additions.
- Showing pure run triggers on the triggerline is intended because they caused a lot trouble while I was editing O9S (phase tracking for example). They are pretty hard to monitor. I could add a button to show/hide those with minor changes or style them differently. Alternatively I could add a sort of tooltip overlay to show more detailed info of the trigger (id, regex, output for example).
- Adding another triggerline for the selected player should be possible with minor additions.

I will take a look at this next weekend.

DrLippe:
image

TODO: 
- Add timeline monitoring (a lot of work)
- Rework Progressbar

quisquous:
Tried the newest version.  It doesn't seem like any of the Orbonne monastery logs I have play any more.  They get detected properly and have nice new names, but they don't start the timeline or have any triggers.  I debugged a little but was unable to make any progress here.

For what it's worth, I still don't see any progress bar.  I squinted a bunch but just don't see any different color as time progresses.  I am not using your raidboss skin, in case that changes anything.  I think it would be worth having the progress bar be in a separate track, for what it's worth.

quisquous:
I should add that my Seiryu logs were still working properly.  It was just various Orbonne ones from the same day that were not.

DrLippe:
The progress bar jumps to 100% ignoring the transition time, going to rework this. Skin has nothing to do with this, i'm using the default skin for testing.

The bug you are facing is caused by the log trimming. `00:0839:` wont make it to the player. Already fixed this locally.

quisquous:
Just wanted to check in where this was.  Are you ready for this to get merged, or are there still things you wanted to fix up first?

I still don't see any progress bar, but that can get fixed later too.

DrLippe:
Can be merged.
- For the timeline triggers to be analyzed i need to find a way to analyze the timeline itself and tracking its timing. This might take a lot of time.
- The progress bar (actually 4-5 of them) is working. It is integrated to the triggerlines but maybe hard to see. You can change its visibility by simply changing the alpha / background-color in the css file.
- I thougt of using some sort of compression (like LZString) on the storage to increase its capacity but that slows down the storing function a lot.

quisquous:
Thanks for all the hard work on this! :heart:

PR #280, opened 2018-11-15
Raidboss skin

DrLippe:
A raidboss makeover to fit the FFXIV Theme a bit better
This is by far not the most beautiful way to implement this.
demo2


ArkfrostLumas:
This looks great! I hope, this will get implemented, as the classic bahamut-timeline skin looks extremely outdated xD

DrLippe:
Don't worry they are both still work in progress.

I managed to trim the skin to just a js file using the deep combinator. It will mention that it's deprecated. Also i need to alter the `TimerBar.draw()` function. 
Setting the skin via `Options.skin` might be pretty complicated since the browser overlay wont allow local file loading it has do be done via the plugin and you may need a similar structure as for the trigger and timeline files.

quisquous:
I'm a little surprised about the change to the draw function().  Correct me if I'm wrong, but in local testing, both the old and the new draw functions both have a little gap when timers complete:
image

But the new function really doesn't behave properly with "active" timers with a duration that are supposed to fill in reverse: 
image

It kind of crosses over? That's probably not intentional, right? Am I guessing that the new draw function is to try to address that gap?

quisquous:
Yeah, I think removing the `-2px` from that line you changed instead of doing scale is the correct answer.  I'll see if I can fix up the original css to not need it.

quisquous:
I didn't commit this pull request directly, as I wanted it to be CSS rather than JS, but I set you as the author in my commit.  Let me know with another pull request if 2d60b9c3bba1c is confusing or doesn't work for you, or if I broke anything you intended.

Thanks again for all this work.  Sorry about taking so long to get around to integrating this, but I really appreciate it and I think a whole lot of other folks will too.

PR #294, opened 2018-11-21
style: README Markdown Changes

panicstevenson:
Changes to the main README file to match recommendations from [markdownlint](https://github.com/markdownlint/markdownlint) with some fixes to indentation, etc. to preserve list counting.

Also adding gitter as a means for contributors to communicate outside of GitHub issues. If something else is preferred or gitter is not wanted, we can put something else or remove the badge entirely.

Preview available [here](https://github.com/panicstevenson/cactbot/blob/markdown-changes/README.md).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,7 @@
-[![build status](https://travis-ci.org/quisquous/cactbot.svg?branch=master)](https://travis-ci.org/quisquous/cactbot)
-
 # cactbot (ffxiv raiding overlay)
+
+[![build status](https://travis-ci.org/quisquous/cactbot.svg?branch=master)](https://travis-ci.org/quisquous/cactbot)
+[![gitter](https://badges.gitter.im/gitterHQ/gitter.png)](https://gitter.im/cactbot/Lobby)
 
 1. [About](#about)
 1. [Installing](#installing)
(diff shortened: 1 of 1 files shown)


quisquous:
Sorry to say no here, but I'm going to decline on the gitter chat.  I appreciate the thought, but I don't have the personal bandwidth to monitor, moderate, and respond in yet another space.

In a similar vein, it is a deliberate choice that there is not a cactbot discord server for exactly this reason, which is probably what I would pick if I wanted such a space.  (The https://www.fflogs.com/home/discord #actstuff channel is kind of an unofficial space that a lot of people ask questions in, but I don't feel responsible for that in the same way that I would for a specific cactbot server.)

I'm not trying to be deliberately hard to reach, but I just only have so much time.  Also, a lot of people are angry about cactbot (just search for cactbot in the balance discord!) and so I like having a slight level of insulation from that.  Personally, I'd vastly prefer if random users created github issues for whatever their questions are.  That's the easiest for me to handle asynchronously as needed, and is a pretty obvious and single location for that sort of thing.

If you personally just wanted to chat and sort through things you want to change in cactbot, I'd be happy to chat on discord.

(Also, what's wrong with two spaces after a period??  :wink:)

panicstevenson:
I can understand about the Gitter link; I wasn't entirely sure how best to manage people wanting to contribute back to Cactbot.

As for the two space thing, with Markdown there's a line-break inserted for any line ending in two spaces; my understanding is that people try to avoid putting the two spaces after each period to reduce introducing any possible inconsistency with what you're typing versus what the end result looks like. I've reverted them for now, just focusing on the list separations and ordered list numbering style.

quisquous:
Thanks for the understanding about gitter.  Please please do not take that as me not wanting to talk in general (especially to folks who are contributing!)

Ah, the two spaces just look way better to me in monospace font, but mostly been burned into my brain.  The winky face is mostly for the number of folks who have given me a hard time for it.  <_<

I never knew that Markdown adds a linebreak there, although github's version of that doesn't seem to do that, I guess.

PR #324, opened 2019-01-13
Add infoText to show Short Stack on DPS during Hello World

HailZaros:
This only shows the infoText if you do not have the Short Stack, or if it is only on the first DPS player (any future "short stacks" won't show up)

I made it as infoText, because it looked odd having 2 alertText running at once about the same mechanic. having the difference between info and alert made it look a lot cleaner in my opinion.

Not sure if there's a better of doing this, but if there is I'd be happy to change it.

Also, I don't know German, so the translation there is likely wrong - but I did my best.


HailZaros:
I tried it with 2 seconds and it felt too slow, because of the delay of healing casts and reaction time. I left it at 1 second, though.

quisquous:
Looks good to me, thanks! I super appreciate trigger contributions.  Nobody really writes them but me, so I definitely appreciate useful things you think are missing.  :+1:

PR #325, opened 2019-01-14
General timeline translation + FR Trigger translation

mooondark:


quisquous:
Holy moly! Thanks for this translation so soon!

I was actually wondering how to do the timeline translation for this, because there aren't fflogs for it.  Did you make that section by hand?

quisquous:
I left a number of FIXMEs for timeline translations here: https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/triggers/orbonne_monastery.js#L753

Those are the things that are missing to have the timeline work properly.  (Thanks again for all your help!)

mooondark:
Hi,
I used the translation script with my own logs. By hand, I couldn't do JP & DE locales :)

Ok for 1B triggers, I won't do it next time.
I'll test it today and i'll try to fix the FIXMEs.

PR #330, opened 2019-01-16
Enrage sync and Handprint Move Callouts

Legends0:
Removed the => and <= as they were being called out with TTS when I tested (maybe was my settings?).

Idea is that cardinal is called where it starts, then it is assumed the player went to the cardinal and stood there and needs to move for the next hand.

Out of middle call is added, but may be removed. It's primary use is to clear up the tracking of the first hand, and second give increased emphasis on the middle circle where the deadly hand will land.

The last Fifth Element is an 18 second cast according to my logs.

### DIFF
--- a/ui/raidboss/data/triggers/seiryu-ex.js
+++ b/ui/raidboss/data/triggers/seiryu-ex.js
@@ -197,13 +197,11 @@
             fr: 'Allez vers les serpents',
 		      };
 		    }
-		    else {
-		      return {
+		    return {
 	          en: 'Out of Middle, Toward Snakes',
 		        de: 'Raus aus der Mitte, Zu den Schlangen',
             fr: 'Allez vers les serpents',
-		      };
-		    }
+		    };
       },
 	    run: function(data) {
 		    data.withForce = true;
--- a/ui/raidboss/data/triggers/seiryu-ex.js
+++ b/ui/raidboss/data/triggers/seiryu-ex.js
@@ -135,7 +135,7 @@
     },
     {
       id: 'SeiryuEx Handprint East Move',
-      regex: /:Yama-no-shiki:37E5:/,
+      regex: / 15:\y{ObjectId}:Yama-no-shiki:37E5:Handprint:/,
       regexDe: / 15:\y{ObjectId}:Yama no Shiki:37E5:Handabdruck:/,
       condition: function(data) {
         return data.firstHandprint !== undefined;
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/triggers/seiryu-ex.js
```diff
@@ -133,20 +133,66 @@
...
     },
+    {
+      id: 'SeiryuEx Handprint East Move',
+      regex: / 15:\y{ObjectId}:Yama-no-shiki:37E5:Handprint:/,
+      regexDe: / 15:\y{ObjectId}:Yama no Shiki:37E5:Handabdruck:/,
+      condition: function(data) {
+        return data.firstHandprint !== undefined;
+      },
+      infoText: {
+        en: 'Move, East',
```
quisquous: Sorry to be subjectively picky here, but I don't know that the "move" part here really adds much for me.  I think just saying "east" or "west" should be enough.

Also, if this is messing up tts, then I'd suggest adding a specific `tts: { en: 'East', etc }` section here that skips the `<=`.

ui/raidboss/data/triggers/seiryu-ex.js
```diff
@@ -133,20 +133,66 @@
...
+      run: function(data) {
+        data.firstHandprint = true;
+      },
+    },
+    {
+      id: 'SeiryuEx Force of Nature',
+      regex: /Yama-no-shiki:37E9:/,
+      delaySeconds: 1,
+      infoText: {
+        en: 'Out of middle',
```
quisquous: This seems good! An alternative might be to take "go to snakes" callout and make the ones after the first time say something like "out of middle toward snakes" and combine these two?

Or do you think it's better to have both?

ui/raidboss/data/triggers/seiryu-ex.js
Legends0: I was originally planning to add a tts section to the original, but could not figure out if I was doing that correctly or not as I had not used tts yet.

I am ok with not adding the move text, it was something I just wanted to test out with and this fight's east/west is clearer and less deadly than o10s' starboard/larboard. In my original version (prior to the master update) of this I was also calling out left/right instead of cardinals as well, but did agree that cardinals should be used as the Yama-no-shiki is always from the north.

quisquous:
Thanks for the changes, this looks great!

All this does make me think that maybe I should consider doing some auto processing for the tts.  Other folks have complained that `+` and `&` get dropped, and you pointed out that `<=` doesn't do what you want either.

Issue #334, opened 2019-01-17
Alerts in O9S post-DC Blaze phase trigger early and are wrong for the entire phase


spteege:
I've noticed ever since a few updates ago that when Blaze happens after Dark Crystal, "STACK AND STAY OUT" triggers extremely early in the fight and blurbs like "SPREAD" don't fire, and in general the alerts for the entire phase seem to be wrong. The timeline is correct. This isn't an issue if the Blaze phase occurs before Dark Crystal. This has been an issue for me since at least 0.8.7 and on every cactbot update I update by dragging the update into my ACT folder and overwriting all files there. I haven't seen anyone else report the issue so it could very well just be me, but I'm not sure what would cause an issue like that client-side.

PR #336, opened 2019-01-19
O9S: fix stack marker regex and entropy bug

DrLippe:
this should resolve quisquous/cactbot#334

### DIFF
--- a/ui/raidboss/data/triggers/o9s.js
+++ b/ui/raidboss/data/triggers/o9s.js
@@ -337,7 +337,7 @@
     },
     {
       id: 'O9S Knock Down Marker',
-      regex: / 1B:........:(\y{Name}):0000:E000:0057:0000:0000:0000:/,
+      regex: / 1B:........:(\y{Name}):....:....:0057:/,
       condition: function(data, matches) {
         return matches[1] == data.me;
       },
@@ -496,7 +496,7 @@
     },
     {
       id: 'O9S Earth Stack Marker',
-      regex: / 1B:........:(\y{Name}):0000:E000:003E:0000:0000:0000:/,
+      regex: / 1B:........:(\y{Name}):....:....:003E:/,
       suppressSeconds: 10,
       infoText: function(data) {
         return {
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/triggers/o9s.js
```diff
@@ -333,7 +337,7 @@
     },
     {
       id: 'O9S Knock Down Marker',
-      regex: / 1B:........:(\y{Name}):0000:0000:0057:0000:0000:0000:/,
+      regex: / 1B:........:(\y{Name}):0000:E000:0057:0000:0000:0000:/,
```
quisquous: I think these should really be `/ 1B:........:(\y{Name}):....:....:0057:/,`, and the equivalent for the `003E` marker.

The first two things after the name aren't consistent (and so I think it was a mistake to specify them here), and although the last ones are always zeroes, maybe better to underspecify.

ui/raidboss/data/triggers/o9s.js
```diff
@@ -172,7 +172,7 @@
     // Fire Path
     {
       id: 'O9S Fire Phase Tracking',
-      regex: / 14:3186:Chaos starts using Blaze /,
+      regex: / 14:3186:Chaos starts using Blaze/,
```
quisquous: This isn't a fix, just a cleanup, right? I think the space is still correct even if messy?

ui/raidboss/data/triggers/o9s.js
```diff
@@ -221,6 +221,10 @@
           fr: 'Packez-vous et restez',
         };
       },
+      run: function(data) {
+        if (data.phaseType == 'orb' || data.entropyCount == 2)
+          delete data.entropyCount;
```
quisquous: Aha, whoops! Thanks for this fix.

ui/raidboss/data/triggers/o9s.js
DrLippe: Honestly, no idea what these stand for but they changed on recent FFXIV plugin update.

ui/raidboss/data/triggers/o9s.js
DrLippe: The log line says ` 14:3186:Chaos starts using Blaze on Chaos.` so the white space is fine for now.
Maybe the ` on Chaos.` part will disappear some day?

quisquous:
Thanks!

PR #343, opened 2019-01-26
Orbonne Monastery: Fixing french regexes

### LINKED ISSUE #341: Some Orbonne triggers don't show up
Hi there, the following triggers are not displayed/working in FR : 

```
regex: / 14:375A:The Thunder God starts using Crush Accessory/,
regex: / 1B:........:(\y{Name}):....:....:005C:/,
regex: / 1B:........:\y{Name}:....:....:003E:/,
regex: / 14:3750:The Thunder God starts using/,
regex: / 14:374F:The Thunder God starts using/,
regex: / 14:3749:The Thunder God starts using/,
regex: / 14:374A:The Thunder God starts using/,
regex: / 14:374C:The Thunder God starts using/,
regex: / 14:374D:The Thunder God starts using/,
regex: / 14:38AA:Ultima, The High Seraph starts using Redemption on (\y{Name})/,
regex: / 14:38DA:Demi-Hashmal starts using Extreme Edge/,
regex: / 15:\y{ObjectId}:Ultima, the High Seraph:3895:Ultimate Illusion:/,
```

Here is a log file, from today (in tank but i'm sure the heal stuff do not work either) : https://transfernow.net/b156b1c6k21r

DrLippe:
This will resolve #341

### DIFF
--- a/ui/raidboss/data/triggers/orbonne_monastery.js
+++ b/ui/raidboss/data/triggers/orbonne_monastery.js
@@ -518,7 +518,7 @@
       id: 'Orbonne Ultima Redemption',
       regex: / 14:38AA:Ultima, The High Seraph starts using Redemption on (\y{Name})/,
       regexDe: / 14:38AA:Cherub Ultima starts using Zerstörung on (\y{Name})/,
-      regexFr: / 14:38AA:Ultima la Grande Séraphine starts using Destruction on (\y{Name})/,
+      regexFr: / 14:38AA:Ultima La Grande Séraphine starts using Destruction on (\y{Name})/,
       alertText: function(data, matches) {
         if (matches[1] == data.me) {
           return {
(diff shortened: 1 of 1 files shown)


mooondark:
Meh, shame on me to not see that :)
I don't understant why FR is the only language that mixes capitals and lower case .... The translation team is doing this on purpose, so I have a lot of work after ....
Is there a way to let cacbot choose between capital/lower case ? Like a regex operator ?

DrLippe:
@quisquous would adding the `i` modifier to all regexes be a bad idea?

quisquous:
I'll think about adding the `i` modifier, but it does slow regex matching down and there's a lot of logs and regexes.

English has starting mixing case as well.  `Belias, The Gigas` is now `Belias, the Gigas` but only sometimes.  From looking at logs, it looks like `Belias, The Gigas` only happens on gains effects/loses effects network lines and everything else is `Belias, the Gigas`.  Is this true for French as well?

PR #355, opened 2019-02-13
korean localization

Bing-su:
hello... I'm totally noob at Github.


Bing-su:
???

quisquous:
Thanks for this! Does cactbot work with the Korean version? In the past I remember that the memory signatures were different.

It looks like Travis CI is complaining too (the red X).  If you click on it and follow the links to the details, it looks like there's some trailing spaces to remove on a few lines: https://travis-ci.com/quisquous/cactbot/jobs/177442840

quisquous:
If you fix those travis ci errors, I'm happy to commit this! Thanks for all the translating work.  :)

Bing-su:
I didn't know Cactbot doesn't work in Korean version... I tested and still Cactbot dosen't work. (Only dps works.) so sad. but I fixed all trailing spaces. (until now I never know what is trailing space)

quisquous:
Thanks! I think the memory signatures and offsets need to be updated to get cactbot to work.  I'm planning to write a quick guide to use cheat engine to find the signatures and offsets in the next week or two.  I'll cc this pull request when I do if you want to use that and figure out what we need to do to get the Korean version to work.

PR #358, opened 2019-02-17
fisher: adding place id check to prevent unmatched place from stickin…

### LINKED ISSUE #346: Fishing overlay doesn't handle undiscovered fishing hole
Once it figures out which fishing hole you're on, even if it's `undiscovered fishing hole`, then it gets stuck on that.  Maybe it should ignore fishing holes it doesn't know about? Or handle that one specifically?

Log lines in English:
`You cast your line at the undiscovered fishing hole.`
`Data on the unfrozen pond is added to your fishing log.`

xephero:
This should resolve #346. Thankfully, I had a few undiscovered ones left for testing. It only updates the place if the place isn't set, and it's unset after quitting, but the undiscovered hole would count as a set place while still being (obviously) not an actual place. Would be nice to polish it later to grab the message where the new hole is discovered to record the very first catch's data, but this should resolve the immediate issue.

### DIFF
--- a/ui/fisher/fisher.js
+++ b/ui/fisher/fisher.js
@@ -174,7 +174,7 @@ class Fisher {
     this.fishing = true;
 
     // Set place, if it's unset
-    if (!this.place) {
+    if (!this.place || !this.place.id) {
       this.place = this.seaBase.getPlace(place);
       this.ui.setPlace(place);
     }


quisquous:
Thanks, this looks perfect.  No need to overcomplicate things just for that edge case.

PR #365, opened 2019-02-25
fix errors in Hydatos on non-English client

cfxxldev:
fixes the "Uncaught TypeError: Cannot read property 'replace' of undefined" and "Uncaught TypeError: Cannot read property 'style' of undefined" on non-English clients by making sure nm.spawnTrigger[this.options.Language] is only accessed if it's defined


quisquous:
Thank you for fixing this!!

PR #366, opened 2019-02-26
Updated Hydatos NM-names for German client

cfxxldev:
Updated Hydatos NM-names for German client.
I got the names from https://ffxiv-eureka.com/.
I can't confirm any of these names unless I get the nms to spawn, but at least from what I remember these names seem to be correct.


quisquous:
Thanks for this, but it looks like you and Akurosia in #368 submitted similar fixes at the same time.

I'm going to merge #368 because it has what your patch has, but also fixes Japanese and a few other things too.

PR #368, opened 2019-02-26
Added Hydatos NM translations + removed duplicat entry for Pagos NMs

Akurosia:


quisquous:
Thanks for the translation and the cleanup!

PR #404, opened 2019-04-29
dps: Fix rdmty healing meter hps value

amki:
Healing hps value currently always shows NaN because the healing variable contains a string in parseHealing.

I've added a parseFloat just like formatNumber right below does it.


quisquous:
Thanks!!

Issue #400, opened 2019-04-08
Fisher Overlay not working.


Karmana26:
Have tried looking at others with similar issues but am unable to get this working again.  I used it a week or so ago and it was fine.  Now it registers my bait but does not show fish or location etc.

I have also gone to summerford and am unable to get the /bow etc to work with dummy.  I can do /countdown 5 and it works fine but reading from a log seems to be an issue.  I do have emotes showing in my log window.

Here is what I get at the moment.  Any advice would be appreciated. 

![Fish](https://user-images.githubusercontent.com/49406488/55746737-66bc6c80-5a32-11e9-9744-4187c6dcd988.jpg)

I have done a FRESH install of all newest files twice so am pretty sure it is nothing to do with structure...

Karmana26:
Just worked out why the emotes do not work.  It is searching for "You" and the emotes writes your name instead.  If I write /echo You bow courteously to the striking dummy it works fine.  Should it automatically change You to my name?  If not the test will not work.

Karmana26:
Well I finally fixed it after narrowing it down to a LOG issue.  

If this helps anyone, NEVER change Log Window Settings to Surname Abbreviated as this begins all logs referring to you in the 3rd person.  Instead of "YOU cast your line...." it changes to "Karmana M casts his line" which breaks all addons that are searching for the generic YOU and not a personalised name.   

Phew, Back to Fishing.

chanzie:
im getting the same display issue, firewall disabled, log checked out fine, just not showing any timers at the Arms of Meed for some reason, havent checked any other locations, also was not getting any coloring with the timers or any of the nifty extras.

Akurosia:
Nearly identical issue here, except that for me (germna client) chat lines seams to be correct but still no display output. Is there any way i can see at least at which step he has an error? there is also no log inside the overlay.dll tab inside act.

geekgirl101:
I'm having this issue myself.  I've made sure I'm using the full name and no abbreviations but the cast time still shows 0.0 whilst fishing and no fish are appearing in the list, this is at The Isle of Bekko in The Ruby Sea.

quisquous:
Sorry for delay on fixing this.  Been distracted with a few other things.  Will try to take a look soon.

quisquous:
Akurosia: can you give me some example lines from fishing that you feel like should work, and I can check them against the regex? Are you seeing the fisher window at all (e.g. the window shown in the first comment in this bug)?

[...]

PR #407, opened 2019-05-07
fisher: hotfix when Name Display Type is abbreviated, English

Omegakenshin:
A Hotfix for Issue #400 ,

The Fisher Overlay doesn't work if Name Display Type is not Full Name.

I tried my best, however maybe you can do it better. 
PS: Love your works, thanks.


quisquous:
Thanks for this hotfix, sorry for the delay in getting back to this.  There's a few corner cases here that I need to go test in game, and then I'll merge this.

I never knew about this log name option before!

KattTails:
With these changes, the catch regex would match every time anyone near you lands a fish?

Omegakenshin:
> 
> 
> With these changes, the catch regex would match every time anyone near you lands a fish?

I will try to test this, however I think you might be right, but getting to a way for this to work on every name display type could need a different approach then. I'll try to check it out soon.

KattTails:
It should work as-is with your changes so long as the FFXIV plugin parse filter is set to alliance or stricter, AND you aren't fishing with someone else in a party within /say range.  This could be an acceptable limitation, honestly.

If you wanted to solve this 'properly', you can get the player name from cactbot with e.detail.name in the eventlisteners (OnPlayerChange?), but you would need to rewrite the entire regex handling because you can't use variables in the current implementation.

KattTails:
I did some more testing and my assumption is actually **wrong**. The parse plugin has different logs for different sources, It seems like other peoples catches are logged with a 00|2043|| prefix, whereas your own have 00|0843||. This makes everything I assumed above not matter at all, and your changes should work fine.

Sorry for the confusion..

quisquous:
Thanks for the extra info about other fishers, that was my biggest worry that I wanted to test.

My other concern is that I think names with apostrophes won't be matched here.  (Although this is certainly better than what was there before.)

quisquous:
Thanks for this! I'll follow up and fix apostrophes.

PR #408, opened 2019-05-12
ja localize

r4ha:
Hi.

I localized some part of raidboss triggers.
Can I contribute to the project?

Thanks.

### DIFF
--- a/ui/raidboss/data/triggers/general.js
+++ b/ui/raidboss/data/triggers/general.js
@@ -23,6 +23,7 @@
     {
       id: 'General Ultimatum',
       regex: /:(\y{Name}):1D73:Ultimatum:/,
+      regexJa: /:(\y{Name}):1D73:アルティメイタム:/,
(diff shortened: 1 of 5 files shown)


quisquous:
This is great, thank you for this! If you have other translations, I will take any pull requests you want to take the time to do.

Have you been using cactbot in Japanese successfully with these changes, or do other things need to be updated as well?

r4ha:
Thank you!

Yes I was personally using this. but when i updated cactbot, I accidentally overwrite my translate.
so I didnt try all bosses. only o9s and o10s works fine.

ArkfrostLumas:
Pro-tip from me, who is making the German translations:
Save a separate repository of Cactbot on your PC and make your translations there. Every time you translate some stuff, commit it, but always keep it separate from your ACT folder, so you don't overwrite anything by accident and always can copy your translations back in, if something happened :)

Issue #109, opened 2018-04-25
Feature request Eureka Fairy location on Map.


Pirate144:
Not sure if possible, but I imagine that ACT does track the location of NPCs when they spawn, if so would it be possible to add like a green dot on the eureka map when it detects one nearby? I know the trigger for them like hunt mobs is "03:Added new combatant Anemos Elemental.".

In fact is that is possible it would probably be possible to overhaul the Eureka module into a hunt module as well.

quisquous:
I can't think of a way to do this very well.  The "added new combatant" doesn't have a position, so it's pretty murky where one is.  You could add a raidboss trigger that alerts you when one is near.

Pirate144:
I already have a trigger for fairys being nearby like hunt mobs. 

I see that the added new combatant log makes no mention of their location, if so then how does FFlogs replay feature show where the boss/adds are at?

quisquous:
It's only ability usage that has positions.  I guess we could detect the fairy using its aoe maybe, and that could have a position attached to it.

haasn:
@Pirate144 I'm lazy, mind copy/pasting me your fairy trigger?

haasn:
Made my own

```javascript
Options.Triggers = [
  {
    zoneRegex: /Eureka/,
    triggers: [
      {
        regex: /Added new combatant (Anemos|Pagos|Pyros|Hydatos) Elemental/,
        infoText: 'Fairy nearby',
      },
    ],
   },
]
```

haasn:
I gave implementing this a try after seeing that ACT now added x/y/z coordinates to the 'added new combatant' log lines, but those seem to live in a completely different reference system compared to the map links, so I don't know how of any easy way to 'project' this onto the minimap.

I guess I could measure the x/y/z position at each corner of the map and then translate that to the map coordinates for each map, but I was wondering if there's an easier way I'm missing?

quisquous:
https://github.com/quisquous/cactbot/blob/master/ui/eureka/eureka.js#L1736 This function will do that for you.  The AddFlag function takes things in map coordinates, but you could make an AddFairy function that converts the coordinates like SetStyleFromEntity does and then call AddFlag?

haasn:
Haha, I just found that function myself, *after* having actually gone through and measured these constants out myself.

(I did add an AddFairy function but since I was too lazy to do any actual UI stuff I just made it call AddFlag, hence why I was needing to translate between the two)

quisquous:
[...]

PR #413, opened 2019-05-28
eureka: mark fairies (really lazily)

haasn:
Betterment ideas / TODO:

- Add support for other languages
- Play some sound effect when one gets found
- Mark these in their own fancy way on the map, also reuse
  SetStyleFromEntity instead of hackily duplicating it

WIP for #109

### DIFF
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -24,6 +24,9 @@ let Options = {
       entityToMapXConstant: 21.45210725,
       entityToMapYScalar: .02000892816,
       entityToMapYConstant: 21.4665545,
+      fairy: {
+        en: 'Anemos Elemental',
+      },
       nms: {
         sabo: {
           label: {
@@ -549,6 +552,9 @@ let Options = {
       entityToMapXConstant: 21.48,
       entityToMapYScalar: 0.02,
       entityToMapYConstant: 21.48,
+      fairy: {
+        en: 'Pagos Elemental',
+      },
       nms: {
         snowqueen: {
           label: {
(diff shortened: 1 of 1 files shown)


haasn:
Ah sorry about the whitespace mixup, don't have a development setup atm. I'll fix it when I get back

ui/eureka/eureka.js
```diff
@@ -2150,6 +2171,13 @@ class EurekaTracker {
         container.removeChild(label);
     }, this.options.FlagTimeoutMs);
   }
+
+  AddFairy(name, ex, ey) {
+    let zi = this.zoneInfo;
+    let mx = zi.entityToMapXScalar * ex + zi.entityToMapXConstant;
+    let my = zi.entityToMapYScalar * ey + zi.entityToMapYConstant;
+    this.AddFlag(mx, my, name, "");
```
quisquous: For consistency, it might be worth just making the name "fairy".  That's what other code does if folks link to a fairy in chat.

quisquous:
Thanks for this! I'll get this landed when you fix up the whitespace.  No need to apologize!  That's why there's automatic checks.  :)

haasn:
Fixed the whitespace

ui/eureka/eureka.js
haasn: It actually already goes through `SimplifyText` inside `AddFlag` so it already gets translated to "fairy" because of the "elemental" in its name. But yeah, might as well change it here, especially w.r.t other languages and stuff.

I think ultimately we would just mark it with its own icon and no text anyways.

PR #416, opened 2019-05-30
fixing ttsText passing condition bug

### LINKED ISSUE #415: Tripped this Error Today
https://i.gyazo.com/cb87dd567ef27e7db19bfdc8d2266ea5.png

This popped up in the middle of raid tonight, any ideas what it is / means?

DrLippe:
This should resolve #415


quisquous:
I agree that this will fix the issue, but I would like to understand what is causing this problem before blindly applying a workaround, sorry.

PR #419, opened 2019-06-02
added timelineReplace + smaller german fixes

Akurosia:
added timelineReplace  for o1s - o4s
change the "stek" part for german translation into "stack" (stek is not a translation of stack).
But correct translations for "stack" eg "sammeln" or "zusammen stehen" would be way to confusing in the examples so to keep it short it will just say "stack auf dir"


quisquous:
Thank you!

PR #424, opened 2019-06-18
update util to support cn

Bluefissure:
- fix some mob names containing "-" or " ". IE, "ao-no-shiki" to "Aka-No-Shiki" and "Aka-no-shiki", thus the name in timeline and triggers can be found easily
- add `cn` config to `util/check_translation.py`, `util/fflogs.py` and `util/translate_fight.py`
- remove unnecessary reformatting since `npm run lint` will be used


Bluefissure:
if we have replace text like:
```
        "Larboard Wave Cannon": "左舷齐射·波动炮",
        "Larboard Wave Cannon Surge": "左舷齐射·零式波动炮",
```
the `Larboard Wave Cannon` will become `左舷齐射·波动炮 Surge` if not sorted

util/translate_fight.py
```diff
@@ -192,22 +202,23 @@ def main(args):
...
-
-        # replace all quotes on headers
-        for header in headers:
-            if line.find('"' + header + '":') != -1:
-                line = line.replace('"', '', 2)
-        # replace double with single quotes on any line without apostrophes.
-        if line.find("'") == -1:
-            line = line.replace('"', "'")
-        lines.append(line)
+    # lines = []
```
quisquous: Could you remove all of this instead of commenting it out?  I verified that check_translation.py can import this without formatting just fine.

util/translate_fight.py
```diff
@@ -181,6 +183,14 @@ def main(args):
     for lang in languages:
         if lang == default_language:
             continue
+        # Fix lower mob names like "ao-no-shiki" to "Aka-No-Shiki" and "Aka-no-shiki"
```
quisquous: I think this does not happen frequently enough to do for all abilities with space or hyphen in the name, and is better to just do manually.  Are you thinking of other examples other than just things with "no" in the name?

I do feel like there are continual french issues (e.g. ecbb723f097328c7bd0476352e5135bd5f776248) and so at some point it might be worth making everything case insensitive.

util/check_translation.py
```diff
@@ -61,10 +62,10 @@ def translate_regex(regex, trans):
         for old, new in trans['~effectNames'].items():
             did_work = did_work or re.search(old, line)
             line = re.sub(old, new, line)
-    for old, new in trans['replaceText'].items():
+    for old, new in sorted(trans['replaceText'].items(), key=lambda x: -len(x[0])):
```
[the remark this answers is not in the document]
Bluefissure: oh we need longer ones first, otherwise, the shorter one will replace the leading parts of longer ones if it's a prefix of the longer.
IE, with this 
```
"Larboard Wave Cannon": "左舷齐射·波动炮",
"Larboard Wave Cannon Surge": "左舷齐射·零式波动炮",
```
the `Larboard Wave Cannon Surge` will be replaced by the shorter first, which will become `左舷齐射·波动炮 Surge` and the longer ones cannot match it.

util/translate_fight.py
Bluefissure: I think most words with space or hyphen in Chinese logs have this problem, like `dark crystal` of o9s, `ancient dragon` and `immortal key` in o10s and so on, so maybe case insensitive is a good solution.

util/translate_fight.py
Bluefissure: yes of course https://github.com/quisquous/cactbot/pull/424/commits/1894ed8b9a50ee639efb8fc6ae1370c220fb389f

util/check_translation.py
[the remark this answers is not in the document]
Bluefissure: this prefix problem will also affect `timeline.js`, see https://github.com/quisquous/cactbot/pull/424/commits/523d8f672b1d90d07fa865153a959c330e5b64d3 for more information

ui/raidboss/timeline.js
```diff
@@ -51,6 +51,7 @@ class Timeline {
       if (!r[replaceKey])
         continue;
       let keys = Object.keys(r[replaceKey]);
+      keys.sort().reverse();
```
quisquous: I thought about this a bit over the week, and I think this is not ideal and it would be better to do this ordering inside the translation block.

Javascript key ordering is predictable, so I think it's better to have the translation section be an ordered set of replacements.  Sorting here removes people's control.

If you have replacements that should be reordered, then I think we should just reorder them in the timelineReplace section, rather than trying to figure out the best ordering later.  So, I think this sorting, and both bits of sorting in check_translation.py should be removed.

util/translate_fight.py
quisquous: Lots of mobs have spaces in their name, so this seems like it would generate a lot of entries.

I think I would prefer to remove this section and to leave just the original name (which hopefully matches the one in the timeline).  The timelineReplace section has two purposes: (1) it's used in **util/check_translation.py** to automatically add triggers in other languages and (2) it's used at runtime to modify the timeline names and syncs.

If you have a lot of names that don't match, then what about this proposal:
* leave the original mob name in the timelineReplace section
* modify check_translation.py to be case insensitive when doing replacements
* maybe make regexCn / regexFr in triggers be case insensitive by making them be `/regex/i` instead of just `/regex/`

ui/raidboss/timeline.js
Bluefissure: If JavaScript dictionary has predictable keys order, just order them in the Jason section is a good choice I think.
I thought it acts like python, which is unpredictable lol

util/translate_fight.py
Bluefissure: Okay I’ll try to make it to /regex/i when I get back to work

ui/raidboss/timeline.js
quisquous: If the keys are all strings that cannot be represented as integers (e.g. not "1") then Object.keys will returns you keys in insertion order: https://www.stefanjudis.com/today-i-learned/property-order-is-predictable-in-javascript-objects-since-es2015/.  I am pretty sure the version of Chrome that we are using also behaves this way for Object.keys, so you can depend on that.

Maybe it means that the Python code for check_translation.py needs to be a little bit different to make sure these stay in order, but for the timeline stuff at runtime, that order can be depended on.

PR #427, opened 2019-06-24
util: add support of cn

Bluefissure:
- add cn language to fflogs & check_translations & translate_fight
- sort replace text from long to short to solve prefix problem
- remove translate_fight reformatting

What's more:
- as for the replace text with space or hyphen in it, I don't think it will be compatible without changing the runtime `timeline.js` to become case-insensitive. I've tried to add case insensitive to `check_translations.py` and add them to triggers but timeline is translated in the runtime script.
  What about adding something like ` text = text.replace(Regexes.Parse('/'+keys[j]+'/i'), r[replaceKey][keys[j]]);` to [timeline.js](https://github.com/quisquous/cactbot/blob/master/ui/raidboss/timeline.js#L55)?
- as for the order of the keys, I managed them in the python script instead of runtime script, python3.6+ will preserve the order of dictionary keys according to [this doc](https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-compactdict)

### DIFF
--- a/util/translate_fight.py
+++ b/util/translate_fight.py
@@ -10,12 +10,13 @@ import fflogs
 
 # 'en' here is 'www' which we consider the "base" and do automatically.
 # 'cn' exists on fflogs but does not have proper translations, sorry.
-languages = ['en', 'de', 'fr', 'ja']
+languages = ['en', 'de', 'fr', 'ja', 'cn']
 prefixes = {
     'en': 'www',
     'de': 'de',
     'fr': 'fr',
     'ja': 'ja',
+    'cn': 'cn',
 }
 default_language = 'en'
 ignore_abilities = ['Attack']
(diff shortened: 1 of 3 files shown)


util/translate_fight.py
```diff
@@ -183,36 +185,18 @@ def main(args):
             continue
         timeline_replace.append({
             'locale': lang,
-            'replaceText': ability_replace[lang],
-            'replaceSync': mob_replace[lang],
+            'replaceSync': dict(sorted(mob_replace[lang].items(), reverse=True)),
```
quisquous: This seems reasonable to me.  We can put everything in the list reversed to start as a way to help folks writing translations.  Things can always be reordered as desired after running this script.

quisquous:
>   What about adding something like ` text = text.replace(Regexes.Parse('/'+keys[j]+'/i'), r[replaceKey][keys[j]]);` to [timeline.js](https://github.com/quisquous/cactbot/blob/master/ui/raidboss/timeline.js#L55)?

I don't think I understand why you need case insensitivity here.  The timelineReplace section is applied directly to the timeline itself, and so if you have `Ao-No-Shiki` in the timeline syncs, then you just make sure you have an `Ao-No-Shiki` key in the replaceSync section.  If the replacement sync itself needs to be case insensitive, then I think you'd want to change the output of GetReplacedSync (https://github.com/quisquous/cactbot/blob/master/ui/raidboss/timeline.js#L64) to be a case insensitive regex.

Am I thinking about this wrong?

Bluefissure:
Yeah, just manually fix the capitalized words can solve this, so we can leave it as it is rn.

PR #437, opened 2019-07-02
Triggers for Titania and Innocence Extreme

amki:
I've started to work on triggers for titania-ex but since I'm new to the ffxiv act combat log I'm trying to stay close to other trigger files.


amki:
They look ok now for me and work pretty well.

quisquous:
Thanks for this! I'll probably adjust these a little bit and add timelines, but this is a great start.

PR #442, opened 2019-07-05
Drop all references to Xilium.CefGlue

ngld:
My updated OverlayPlugin doesn't use CefGlue so trying to load Cactbot with that fails right now.
Cactbot either needs to use CefSharp (which I'm using) or drop all references to Xilium.CefGlue.
Switching to CefSharp would break compatibility with the old OverlayPlugin so I'd rather avoid that.

This change should be compatible with both versions. I've tested this change with my updated version but not with the old one. If it breaks, the most likely cause is that trying to execute JavaScript before the page is loaded is silently ignored. In that case I'll have to think of a better solution to the issue.

### DIFF
--- a/CactbotOverlay/CactbotOverlay.cs
+++ b/CactbotOverlay/CactbotOverlay.cs
@@ -299,7 +299,7 @@ namespace Cactbot {
     private int SendFastRateEvents() {
       // Handle startup and shutdown. And do not fire any events until the page has loaded and had a chance to
       // register its event handlers.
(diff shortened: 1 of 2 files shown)


quisquous:
Thanks for this.  I was hoping to get to this today, but I ran out of time.

I'm out of town for a week, but will give this a spin when I get back.  I'm a little hesitant of doing a cactbot release without testing this locally a bit more thoroughly and so don't want to rush this.  I've definitely had a number of race conditions (and still do) with loading, and so I don't want to introduce more.

PR #447, opened 2019-07-08
added translations to titania and innocence + small change to translate_file.py

Akurosia:
added translations to both mentioned files.

for innocence the following there is 1 translation missing which i cannot find inside the games data ("Schwert des Urteils") which seams to be the golden sword/mirrors on the outer edge.

The change for translate_fight.py was made so output from the file can be directly used in trigger files.


Akurosia:
added fix for issue #453

util/translate_fight.py
```diff
@@ -98,10 +98,18 @@ def build_mapping(translations, ignore_list=[]):
                     # raise Exception('Conflict on %s: "%s" and "%s"' % (default_name, existing_name, name))
                     pass
             else:
-                replace[lang][default_name] = name
+                if default_name and name :
+                    replace[lang][default_name] = name
     return replace
 
 
+def format_output_str(output_str):
```
quisquous: I hadn't added this before, because running `npm run lintfix` will auto-format this properly.

If you're going to do this, can you also add trailing commas on `^\s*]$` and `^\s*}$` too?

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -118,8 +118,8 @@ public enum EntityJob {
       SAM = 34,
       RDM = 35,
       BLU = 36,
-      DNC = 37,
-      GNB = 38,
+      GNB = 37,
```
quisquous: Can you remove this from this pull request? I'm pretty sure this is correct, but I want to double check this myself before committing more plugin changes.

Akurosia: should be reverted now

util/translate_fight.py
Akurosia: fixed

quisquous:
Thanks!!

PR #459, opened 2019-07-13
Fixed/Update MCH Support

hitokirixeno:
Fixed the offsets for the heat gauge, overcharge timer, battery gauge, and battery timer for CactbotOverlay.dll. Also added very basic and very ugly MCH support for the jobs module to test.

### DIFF
--- a/CactbotOverlay/FFXIVProcess.cs
+++ b/CactbotOverlay/FFXIVProcess.cs
@@ -955,26 +955,26 @@ namespace Cactbot {
     }
 
     public class MachinistJobData {
-      public uint overheat_ms = 0;
       public int heat = 0;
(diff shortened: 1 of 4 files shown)


quisquous:
Yeah, sorry for the delay in looking at this.  I'm not sure what's going on with the whitespace here, but it looks like every file has changed.  Can you sort that out so that the diffs look more reasonable and I can see what's changed?

Travis CI also has some errors: https://travis-ci.com/quisquous/cactbot/jobs/215815651

hitokirixeno:
That's strange - seems like my local repo was all caddywhompus. I couldn't even reset the repo back without the whitespace issue. I'll close this and open a new pull request since I had to destroy the repo.

PR #465, opened 2019-07-19
Fixed instance regex text

JLGarber:


quisquous:
Thank you for catching that!

PR #476, opened 2019-07-23
Titania Ex and Eden (1-4) normal

mooondark:


quisquous:
Thanks for the note about that English regex.  Whoops, copy and paste errors.

PR #481, opened 2019-07-24
Innocence Ex Fr update

mooondark:
This is a first try as I did not run this instance yet, so triggers may be incorrect.


quisquous:
Thanks for all the continued translation work.  Definitely let me know if there's anything I could do to make life easier for you.

PR #497, opened 2019-08-01
raidboss: initial e4s triggers

panicstevenson:
Updates to the Eden's Gate: Sepulture Savage triggers. These triggers are currently untested.

Should have every head indicator, mechanic trigger, fist pattern, and fracture pattern. These are not gospel -- use best judgement; I used the fflogs replay feature to ground assumptions.

Wish List:
Tank swapback
Geocrush position
Bomb explosion order
2x Evil Earth positions for safe corners
Indicate Force needs two stacks for the 8x Force mechanics


quisquous:
heck yeah, thanks!

PR #503, opened 2019-08-02
Translation for Eden savage + Small QoL update

Akurosia:
Added defaulting to help screen when using translation script + a small notification about for unexpirienced users that they have to use the public key (not private one)

Also added timeline replacements (created via script) for eden savage


quisquous:
Thanks for doing this! The script changes are really good too.

PR #504, opened 2019-08-02
util: adding timeline aggregator

panicstevenson:
Adding functionality to the make_timeline.py util to allow for timeline aggregation/event timestamp averaging.

Updates to make_timeline.py:
* Adding import statement for timeline_aggregator
* Updating print statements to return an array
* Main function call will return expected results to stdout
* Reports can handle a comma-separated list of values to parse multiple fflogs reports at the same time, `-r report1,report2` or `-r "report1,report2"` if you're easily scared without quotes
* Adding argparse option `-at` for the averaging threshold, not required and only used when supplying multiple reports

timeline_aggregator.py:
* Could have a better/more accurate name
* Takes N fflogs reports, uses make_timeline to create the timeline, and returns a new timeline with the event timestamps hopefully smoothed out
* Code is both objectively bad and objectively slow from a big O point of view
* Uses janky methods for doing the grouping/aggregating into averaging; there's probably a _much_ better way to do this via any of the number of libraries that do data-anything in Python
  * If you know of a better way, please assist
* Works, but could work order of magnitudes better

Issues that came up:
* Drift
  * Minor things, such as the boss repositioning or flavor text, can impact long-term drift
* Fast vs slow instances
  * Fast instances seemingly hit enrage ~0.5-1% faster than slow instances, which can add up over a 10+ minute encounter.

With both of these issues, the question becomes what should be the source of truth (fast vs slow, absolute timestamps vs relative offsets), but the bigger question is does any of it really matter if you're syncing with every ability used?

Last second edits:
* Actually works pretty well with `--phase`
* Passing multiple reports should be updated to follow how you pass multiple phases

### DIFF
--- a/util/timeline_aggregator.py
+++ b/util/timeline_aggregator.py
@@ -60,11 +60,11 @@ def output_to_timeline(timeline_event_map):
     return sorted(list(timeline_events), key=lambda s: float(s.split()[0]))
 
 
-def create_averaging_function(treshold):
+def create_averaging_function(threshold):
     """Creates and returns a function to group by.
 
-    Creates the averaging function to be used with the itertools.groupby
-    function. Takes a treshold value in seconds as input to average groups
+    Creates the averaging function to be used with the itertools.groupby()
+    function. Takes a threshold value in seconds as input to average groups
     of similar event times.
     """
     key = None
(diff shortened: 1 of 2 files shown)


util/timeline_aggregator.py
```diff
@@ -0,0 +1,127 @@
...
+
+def average_similar_events(values, treshold):
+    """Return a list where similar numbers have been averaged.
+
+    Items are grouped using the supplied width and criteria and the
+    result is rounded to precision if it is supplied.  Otherwise
+    averages are not rounded.
+
+    """
+    grouped_values = itertools.groupby(values, create_averaging_function(treshold))
```
quisquous: This is very elegant!

Issue #505, opened 2019-08-02
Unfriendly TTS symbols


quetzalsly:
E3S for example has very annoying "slash" sounds, if you change these to commas it speaks them perfectly. It would be really nice to change that and possible also look at other problematic files with a grep? like >= / all of those cause problem TTS. Thanks!

panicstevenson:
@quisquous and I had talked about this a little, and I think we agree that the callout in question `front left / back right` can use some work in general. If you have any proposals for making the call out cleaner/more succinct without the use of any symbols I think we'd be happy to go with that.

As for the overarching intent of this issue, I agree. It has been mentioned for other turns as well.

quetzalsly:
i tried it with a comma and it works well enough, personally i would just keep it like that. Cactbot does not know where you are so it can't narrow it down to one callout, but you simply check "Am i front or back?" and listen for the one that applies to you.

panicstevenson:
Looking from the list of javascript files that hit `^\W+en: '.*[/<>=].*',$`, there's a lot of things that have had slashes in them. Testing out different text to speech options, they all handle slashes a bit differently; some read every slash, some never read the slash, and some won't read it if there's no spaces in the slash.

For things like `In/Out, then Water Positions` in Titania, has this procced the same effect?

quetzalsly:
I dont remember having problems with titania but im not sure, maybe the tts was short enough not to overload my brain haha. Also I don't have windows media player installed so I can't use that API, I use the other API which can only output 1 voice at a time. I imagine a lot of people don't use windows media player anymore so the best solution would just be to fix it for everyone universally.

quisquous:
Yeah, I think the solution here will just be to do some ad hoc replacement before playing the tts.  Maybe `/` turns into nothing, and `=>` turns into `then`.  I think there are some things where I say `go east =>` or `<= go west` and maybe I can clean those up and be consistent.

Are there other bad symbols?

quetzalsly:
I can't think of any other bad symbols right now. Also I don't want to open a new issue but in e3s there was a change which makes the callouts awful. These 4 occurances: 'left front / back right'. First of all it has the slash and second of all it is very very hard for the brain to comprehend this because it is inconsistent! Simply changing it to 'front left, back right' works perfectly. So your brain first understands if its front or back and then which side.

JLGarber:
[...]

PR #508, opened 2019-08-03
raidboss: adding e3s tts triggers without slashes

panicstevenson:
Quick fix for the immediate issue presented in https://github.com/quisquous/cactbot/issues/505.

A longer term fix should be implemented for everything that has unfriendly characters in the text-to-speech, either by including a `tts:` block without the special characters as I've done here, or by updating the alerts themselves. I feel like the latter is the better option in general, as it prevents drift, but this should be an immediate fix for anyone progging E3S with TTS alerts.

Signed-off-by: Panic Stevenson <panic.stevenson@gmail.com>

### DIFF
--- a/ui/raidboss/data/triggers/e3s.js
+++ b/ui/raidboss/data/triggers/e3s.js
@@ -377,6 +377,9 @@
         en: 'left front / back right',
         ja: '左前 / 右後ろ',
       },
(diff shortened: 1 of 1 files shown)


quisquous:
I added e79ecc78b, which does this in a more automatic way.

PR #513, opened 2019-08-04
raidboss: trigger resolving

DrLippe:
#9 
There might be a better way to do this but that's beyond my knowledge.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -386,12 +386,12 @@ class PopupText {
           if ('resolve' in trigger && 'id' in trigger) {
             Object.assign(this.intervals, {
               [trigger.id]: window.setInterval(
-                resolveText.bind(
-                  that,
-                  trigger.resolve,
-                  removeText.bind(that, holder, div),
-                  trigger.id
-                )
+                  resolveText.bind(
+                    that,
+                    trigger.resolve,
+                    removeText.bind(that, holder, div),
+                    trigger.id
+                  )
               ),
             });
           } else {
@@ -417,12 +417,12 @@ class PopupText {
           if ('resolve' in trigger && 'id' in trigger) {
             Object.assign(this.intervals, {
               [trigger.id]: window.setInterval(
-                resolveText.bind(
-                  that,
-                  trigger.resolve,
-                  removeText.bind(that, holder, div),
-                  trigger.id
-                )
+                  resolveText.bind(
+                    that,
+                    trigger.resolve,
+                    removeText.bind(that, holder, div),
+                    trigger.id
+                  )
               ),
             });
           } else {
@@ -448,12 +448,12 @@ class PopupText {
           if ('resolve' in trigger && 'id' in trigger) {
             Object.assign(this.intervals, {
               [trigger.id]: window.setInterval(
-                  resolveText.bind(
-                    that,
-                    trigger.resolve,
-                    removeText.bind(that, holder, div),
-                    trigger.id
-                  )
+                    resolveText.bind(
+                      that,
+                      trigger.resolve,
+                      removeText.bind(that, holder, div),
+                      trigger.id
+                    )
               ),
             });
           } else {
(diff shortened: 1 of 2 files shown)


DrLippe:
Triggers can now stay visible until a condition is met.
Also raidboss should be able to translate english regexes if no localized regexes are present but timelineReplace is.

ui/raidboss/popup-text.js
```diff
@@ -121,6 +122,48 @@ class PopupText {
             if (!trigger.regex)
               console.error('Trigger ' + trigger.id + ': has no regex property specified');
 
+            // If trigger has no localized regex but does have a replacement
+            // create regex
+            // FIXME: move this to loading data files to save time
+            if (!trigger[regexLocale] && set.timelineReplace) {
```
quisquous: I understand what you're getting at here, but I would prefer to do this translation not at runtime and commit the values.  That lets other people look at what's going on and commit changes if needed.  Committing these translations also makes it clear that somebody has given it a first pass who is actually using those languages.

This is what the `check_timeline.py` script does, in that it appends translated regexes based on timelineReplace, and you can evaluate if it did the right thing and then commit them.

ui/raidboss/popup-text.js
```diff
@@ -455,7 +553,14 @@ class PopupText {
       // on (speech=true, text=true, sound=true) but this will
       // not cause tts to play over top of sounds or noises.
       if (ttsText && playSpeech) {
-        ttsText = ttsText.replace(/[#!]/, '');
+        // Heuristics for auto tts.
+        // * Remove a bunch of chars.
+        ttsText = ttsText.replace(/[#!\/]/, '');
```
quisquous: It's a little weird to have these changes in here too.  Can you rebase on top of the latest master?

ui/raidboss/popup-text.js
```diff
@@ -324,6 +374,12 @@ class PopupText {
           }
         }
       };
+      let resolveText = (condition, callback, triggerId) => {
```
quisquous: Can you add some documentation in https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/triggers/README.txt for how this new feature works?

quisquous:
Thanks for the documentation, that helps clarify a lot.  I think I would prefer to not have constant polling here, and instead just have some other trigger disable it.  Since the function won't become untrue until some other trigger runs, it might make sense to just have the other trigger turn it off, if that makes sense?

I'm thinking of an approach where `data` could expose some `HideText(id)` function that took a trigger id, and then another trigger could just hide it directly? Is there a use case this wouldn't cover?  What do you think of this approach?

Is there a particular trigger you have in mind for this, or are you just filling in the feature request I filed a long time ago to do this very thing? (Thanks!)

DrLippe:
The resolve function will update constantly. The idea that the trigger will resolve with updated `data` and a separate trigger to resolve this is not needed. My trigger for testing this was 
```js
{
  id: 'Trigger Resolve Test',
  regex: / hello world resolve/,
  infoText: 'change to tank to resolve',
      return data.role != 'tank';
  },
  resolve: function(data) {
    return data.role == 'tank';
  },
}
```
Actually yes this is the answer on that feature request 😅 
But this may open the door to resolve triggers with forwarded data like positioning, jobdetail etc.

quisquous:
Yeah, I was just trying to think about use cases for when this kind of thing would be used.  Do you have example of things where you think this would really be helpful to do things that aren't possible today?

A loooong time ago, I suggested maybe for stuff like 6 fulms under in o2s, where you could hide the trigger when somebody used the raid action.  But in all honesty, just using a timer to show it for longer (which also didn't really exist at the time) is just as good.

DrLippe:
I admit use cases are pretty hard to find.

quisquous:
Yeah, ok.  I'm sorry to close something you've spent work on, but I think I want to be careful about adding new things that add some complexity but don't have a lot of good use cases.  I just only have so much time to support (or explain) all the features enough as it is, you know?  If this comes up as being needed, I'll resurrect this pull request and merge it.

Thanks!

PR #515, opened 2019-08-05
raidboss: e1s timeline updates

panicstevenson:
The update no one wanted:

* updating the initial Eden's Gravity sync window
* smoothing timeline data aggregated across multiple runs
* adding an additional Mana Slice if warranted during add phase
* updating the final phase and enrage

Timelines aggregated with multiple 10 minute+ kills and wipes, and tested against several of the current fastest clearing reports available.

### DIFF
--- a/ui/raidboss/data/timelines/e1s.txt
+++ b/ui/raidboss/data/timelines/e1s.txt
@@ -11,88 +11,73 @@ hideall "--sync--"
 0 "Start"
 0.0 "--sync--" sync /:Engage!/ window 0,1
 1.5 "--sync--" sync /:Eden Prime:367:/ window 2,0
-12.0 "Eden's Gravity" sync /:Eden Prime:3D70:/ window 12,0
+12.0 "Eden's Gravity" sync /:Eden Prime:3D70:/ window 12,1
 25.6 "Vice And Virtue (D)" sync /:Eden Prime:44EF:/
 32.0 "Eden's Flare" sync /:Eden Prime:3D73:/
-50.7 "Vice And Virtue (T)" sync /:Eden Prime:44EE:/
-58.2 "Spear Of Paradise" sync /:Eden Prime:3D88:/
-59.8 "Heavensunder" sync /:Eden Prime:3D89:/
-69.6 "Vice And Virtue (H)" sync /:Eden Prime:44F0:/
-72.1 "--corner--" sync /:Eden Prime:4683:/
-80.9 "Pure Light" sync /:Eden Prime:3D8A:/
-83.7 "--center--" sync /:Eden Prime:4683:/
-93.8 "Delta Attack (Cross)" sync /:Eden Prime:44F4:/
-99.4 "--sync--" sync /:Eden Prime:4683:/
-106.0 "Dimensional Shift" sync /:Eden Prime:3D7F:/
-112.5 "Paradise Lost" sync /:Eden Prime:3D86:/ duration 9
-122.9 "Pure Beam" sync /:Eden Prime:3D80:/
-125.4 "--corner--" sync /:Eden Prime:4683:/
-134.2 "Pure Light" sync /:Eden Prime:3D8A:/
-137.0 "--center--" sync /:Eden Prime:4683:/
-143.7 "Dimensional Shift" sync /:Eden Prime:3D7F:/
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/timelines/e1s.txt
```diff
@@ -11,88 +11,73 @@ hideall "--sync--"
...
+748.3 "--corner--" sync /:Eden Prime:4683:/
+757.1 "Pure Light" sync /:Eden Prime:3D8A:/
+760.0 "--center--" sync /:Eden Prime:4683:/
+766.5 "Dimensional Shift" sync /:Eden Prime:3D7F:/
+779.2 "Eden's Gravity" sync /:Eden Prime:3D70:/
+790.7 "Eden's Flare" sync /:Eden Prime:3D73:/
+800.2 "Paradise Regained" sync /:Eden Prime:44FC:/
+810.7 "Vice and Virtue (T)" sync /:Eden Prime:3D78:/
+824.2 "Vice and Virtue (DPS)" sync /:Eden Prime:3D7A:/
+836.7 "Enrage" sync /:Eden Prime:45E4:/
```
quisquous: I think technically there can be a delta attack here.  I've seen some folks chat on reddit that depending on if you kill the adds during their enrage or not, you may get all the way to delta attack here.  It's probably worth including the next few mechanics as well, as add phase dps gets faster, seeing more things here is more possible if the final phase is a mess.  I don't know for sure that Spear of Paradise / Heavensunder is next, but it seems like a good guess.

I'd be fine including an "--Enrage Soon--" if you wanted in the 800 range before the jump (or maybe just unrolling the loop)?

Do you know how long the Enrage cast is? If so, could you add a jump to `:45E4:Eden Prime` for the cast start?

ui/raidboss/data/timelines/e1s.txt
```diff
@@ -11,88 +11,73 @@ hideall "--sync--"
 0 "Start"
 0.0 "--sync--" sync /:Engage!/ window 0,1
 1.5 "--sync--" sync /:Eden Prime:367:/ window 2,0
-12.0 "Eden's Gravity" sync /:Eden Prime:3D70:/ window 12,0
+12.0 "Eden's Gravity" sync /:Eden Prime:3D70:/ window 12,1
```
quisquous: This could also be 12,5, as `window 5,5` is the default.

quisquous:
Also, thanks for this! I've always felt a little bad that timelines got generated mostly from one piece of data, and so having a more aggregated set is a really nice improvement.  Maybe most people won't notice, but it'll make everything more dependable.

ui/raidboss/data/timelines/e1s.txt
panicstevenson: Interesting; I looked at the longest kills and wipes I could find and I never once saw a Delta Attack, but it _was_ mentioned within the Balance Discord's guide for Eden Prime. I just thought they took an educated guess as to what happened next since I didn't see it in any of the logs I saw.

Enrage is about 8-9 seconds if I remember correctly. I can do a bit more digging into this tomorrow.

panicstevenson:
I changed the timeline a bit to react to the comments, but I don't really know what's more useful -- how I updated it will only effectively show you how long you have for each mechanic but won't show you the mechanics coming up.

For instance, you do the last two Vice and Virtues then it says --Enrage Soon-- then it says nothing and you're either seeing enrage or you're seeing a Delta Attack (Donut) that says you have 2 more seconds after it resolves before enrage begins casting.

For my own reference so I don't have to go back later depending on what we do: the Delta Attack seems to begin casting 6.75 seconds after the second Vice and Virtue resolves, and has a duration of 5.5 seconds. The auto-attack sync after the Vice and Virtue may or may not happen, and it being there or not being there doesn't give any additional information. Enrage begins casting 2 seconds after Delta Attack resolves, or 8 seconds after Delta Attack begins casting. Enrage has a duration of 10 seconds.

Or, like you said, we can have ~800 sync to /:44FC:Eden Prime:/ and jump up to `518.4 "Paradise Regained" sync /:Eden Prime:44FC:/`. Fragor Maximus can have a window of 500 since it's only cast twice throughout the fight.

Actually, I'm just going to do that and still post this comment anyway.

PR #518, opened 2019-08-06
More FR Translations for E1S & E2S

Snsei987:


ui/raidboss/data/triggers/e2s.js
```diff
@@ -27,21 +28,24 @@
...
+      regex: / 1A:\y{ObjectId}:Voidwalker gains the effect of (?:Spell-In-Waiting|)/,
+      regexFr: / 1A:\y{ObjectId}:Marcheuse Du Néant gains the effect of (?:Déphasage incantatoire|)/,
       run: function(data) {
         data.waiting = true;
       },
     },
     {
       id: 'E2S Spell In Waiting Lose',
-      regex: / 1E:\y{ObjectId}:Voidwalker loses the effect of Spell-In-Waiting/,
+      regex: / 1E:\y{ObjectId}:Voidwalker loses the effect of (?:Spell-In-Waiting|)/,
```
quisquous: Can you undo this? Unlike abilities which have ids, this has no effect id listed, and so needs the name and not a `(?:X|)` optional block.  That construct was only there when I wasn't sure what some names were or I was being future proof when things changed from Unknown to real names.

Snsei987:
Should be good now

quisquous:
Thanks!

PR #519, opened 2019-08-06
adding user config for rdmty

juan-medina:
Hi, I've done this simple change to allow to have user config on the rdmty dps module


quisquous:
Thanks!

PR #533, opened 2019-08-10
Fix & add FR translations for E3S

Snsei987:


ui/raidboss/data/triggers/e3s.js
```diff
@@ -115,6 +117,7 @@
         if (data.role == 'healer') {
           return {
             en: 'Tank Busters',
+            fr: 'Tankbursters',
```
quisquous: Typo?

Snsei987: I don't know, I've checked other fights and sometimes it's 'Tankburster', sometimes 'Tank burster', not very consistant.

Do you still want me do update it ?

quisquous: Oh, I meant `Tankbuster` vs `Tankburster`.  I don't see the latter with the extra `r` anywhere else, so was curious if that was intended.

Snsei987: Oh yeah, mb, gonna fix it

Snsei987: Should be good now

PR #535, opened 2019-08-11
Add E4S FR translations

Snsei987:


quisquous:
Thanks for fixing all these up.  I just wasn't sure whether or not to capitalize everything.  Maybe I should make the auto translator do that by default?

PR #536, opened 2019-08-11
Small german e2s additions + adding missing element for oopsy

Akurosia:
Small german additions for e2s + adding missing triggers element for innocence-ex.js (to avoid error)

### DIFF
--- a/ui/oopsyraidsy/data/innocence-ex.js
+++ b/ui/oopsyraidsy/data/innocence-ex.js
@@ -31,5 +31,4 @@
     'InnoEx God Ray 3': '3EE8',
     'InnoEx Explosion': '3EF0',
   },
-  triggers: [],
 }];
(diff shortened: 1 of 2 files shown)


ui/oopsyraidsy/data/innocence-ex.js
```diff
@@ -31,4 +31,5 @@
     'InnoEx God Ray 3': '3EE8',
     'InnoEx Explosion': '3EF0',
   },
+  triggers: [],
```
quisquous: This shouldn't be needed.  Is this still giving errors on the latest of master branch?

Akurosia: You are right, its not longer needed. Reverted commit.

quisquous:
Thanks for double checking that it's not needed.  I was pretty sure, but the verification is appreciated.

PR #539, opened 2019-08-12
Fix Red Chalice in Fisher overlay

imynwulf:
Red Chalice contained the word the which broke the lookup.


quisquous:
Thanks for letting me know this was broken.  Because the fishing data there is generated from the game itself, I don't want to have to modify it every time I regenerate it, so I did this a different way in 8a2434a.

Issue #527, opened 2019-08-09
Option to select the TTS language separately ?


Snsei987:
Would it be possible to add an option to have the timelines in one language and the TTS in another ?

quisquous:
What's the use case here? I guess I figured that most folks who are using one language what everything the same.

Snsei987:
Well, as I try to read guides and run the raid sometimes with pugs, I would like to have the timeline in english to know exactly what ability they are making the call for etc...

And I would also like to the the tts in my native language as I will react faster to it.

Maybe make an option `Options.TTSLanguage` which is equal to `Options.Language` by default ?

PR #551, opened 2019-08-18
Separate Alerts/TTS Language from timeline (#527)

Snsei987:


quisquous:
Oh, this is a lot easier and simpler than I thought it would be! Thanks for the patch :+1:

PR #555, opened 2019-08-20
Added dancer standard finish buff/aura

tsposato:
first time using git etc. so probably needs to be reviewed...


resources/lang/de.js
```diff
@@ -30,6 +30,7 @@ class CactbotLanguageDe extends CactbotLanguage {
       FoeRequiem: 'Requiem Der Feinde', // up 0x8b, down 0x8c
       LeadenFist: 'Verbesserter Schlag auf Schlag',
       TechnicalFinish: 'Komplexes Finale',
+      StandardFinish: 'FIXME',
```
quisquous: Can you add these translations? https://xivapi.com/Action/16003?pretty=true&columns=Name,Name_de,Name_en,Name_fr,Name_ja

quisquous:
Oh, also if you're up for adding devilment (probably needs to be pretty high on the buff list, since it's 20% crit), feel free to add that too.

tsposato:
> Oh, also if you're up for adding devilment (probably needs to be pretty high on the buff list, since it's 20% crit), feel free to add that too.

Ok done, how is that?

quisquous:
Looks great, thanks a bunch!!

haasn:
Should this perhaps be hidden by default for non-dancers? Not much use to track a buff that has 100% uptime.

quisquous:
It only applies if you get the effect yourself.

PR #559, opened 2019-08-21
Triggers readme work

JLGarber:
I've had this one hanging around half-finished for a week or two, so I sat down to it and finished it up today. I did make some edits from the original, mostly in terms of clarifications on how things work. I also added a brief description of what each of the popup text levels should be, based on what you told me on some of my other pull requests. We might want to abstract that out into an in-depth trigger guide section at some point.

As for the formatting, I'm open to suggestions on that. I'm terrible with formatting overall, so what's there is minimally functional and pretty much exactly matches what was in the .js-formatted readme.


ui/raidboss/data/README.md
```diff
@@ -0,0 +1,151 @@
...
+```
+[{
+  zoneRegex: /match for the zone/,
+  timelineFile: 'filename.txt',
+  timeline: `hideall`,
+  timelineReplace: [
+    {
+	  locale: 'en',
+	  replaceText: {
+      "regexSearch": "strReplace",
```
quisquous: can you indent this correctly? <_<

ui/raidboss/data/README.md
```diff
@@ -0,0 +1,151 @@
...
+    { /* ..trigger 1.. */ },
+    { /* ..trigger 2.. */ },
+    { /* ..trigger 3.. */ },
+  ]
+}]
+```
+# Elements
+
+**zoneRegex**
+Largely self-explanatory. For most situations you will want to match the encounter zone and nothing else.
```
quisquous: Maybe worth saying "zone name as given by ACT", in the sense that it doesn't need to be translated and isn't what's in game.

ui/raidboss/data/README.md
```diff
@@ -0,0 +1,151 @@
...
+  disabled: false,
+},
+```
+# Trigger Elements
+
+**id string**
+ An id string for the trigger, used to disable triggers. Every built-in trigger that has a text/sound output should have an id so it can be disabled. User-defined triggers need not have one.
+
+ **regex**
+The regular expression Cactbot will run against each log line to determine whether the trigger will activate. Positions 1 and 2 are capture groups that can be referenced in trigger functions.
```
quisquous: Could say that the capture groups are returned in the `matches` variable

ui/raidboss/data/README.txt
```diff
@@ -1,150 +0,0 @@
...
-4. durationSeconds
-5. suppressSeconds
-6. infoText
-7. alertText
-8. alarmText
-9. tts
-10. run
-
-
-Timeline Readme
```
quisquous: Would you mind including this bit too?

ui/raidboss/data/README.md
quisquous: And use single quotes <_<

ui/raidboss/data/README.txt
JLGarber: Oof. I was working off a local copy of readme.txt from before you redid the file structure. Sorry I missed including that.

JLGarber:
Thanks much for the changes. I'll should have the rest of it in place tomorrow. I apologize for the piled-up history, I'm still working out my way around Git and GitHub, and the documentation around Rebase and Reset is a little frightening. I will take some time this weekend to work that out properly.

quisquous:
No need for apologies at all! Thank you *so* much for the well-formatted and way better readme doc.  It's much appreciated.

I'm happy to help explain git things if I can help! 100% agreed that rebase and reset are a lot.  As long as you've committed it or added it, you can always get it back though, so don't worry too much about messing things up beyond repair.

JLGarber:
Alright, this should be close to what we want. I think I'll revisit the timeline section this weekend, as I'm not completely happy with my presentation in the Markdown version, but it is at least clearly understandable.

At some point, I might try to go through the raidboss-example.js comments and pull those into the readme as well, if you think that's acceptable. Not that the information shouldn't be there, it's very helpful to have it in-line, but I think it would be helpful to have that also co-located with the rest of the trigger information. (That, and at least for me it's much easier to read Markdown than comments.)

quisquous:
Thanks! I think I'll land this for now, but feel free to make any changes in the future.

I think the raidboss-example.js comments maybe aren't the best to include wholesale.  They're mostly just there to explain what can be done, but are kind of just silly examples that I thought of.

I think if there's missing documentation, it would be to make the "advanced cactbot" doc have some sections about common things that people want to do.  Things like disabling the Game On noise, or setting up tts, or css changes like rearranging elements or making everything bigger.  There's just a whole bunch of common questions that get asked a lot, and maybe it'd be good to have some easy sections to refer people to.

PR #561, opened 2019-08-21
Add Manifest Integrity Checker

panicstevenson:
Add tests for validating the integrity of a manifest.txt file:

  Errors:
   - An entry in the manifest must exist within the data directory
   - A file within the data directory must have a corresponding
     manifest entry

Add test to automation via Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,3 +17,9 @@ matrix:
       - pip install pylint requests
     script:
       - pylint **/*.py --errors-only
+
+  - language: python
+    python:
+      - 3.7
+    script:
+      - python test/manifest_integrity.py
--- a/test/manifest_integrity.py
+++ b/test/manifest_integrity.py
@@ -1,0 +1,94 @@
+"""Validates the integrity of the manifest files and data directories for Cactbot modules."""
+
+from pathlib import Path
+import sys
+
+DATA_DIRECTORY = 'data'
+MANIFEST_FILENAME = 'manifest.txt'
+
+
+def main():
+    """Validates the integrity of the manifest file against the files within the data directory.
+
+    Validation includes checking if all of the files listed in the manifest are present in the
+    module's expected data directory, and if all of the files in the expected data directory have
+    a corresponding entry within the module's manifest.
+
+    Returns:
+        An exit status code of 0 or 1 if the tests passed successfully or failed, respectively.
+    """
+    exit_status = 0
+
+    # TODO: Only goes two directories up -- makes assumptions about project structure
+    project_root_dir = Path(__file__).parent.parent
+
+    # Find all manifest files within the project scope
+    for manifest_filepath in Path(project_root_dir).rglob(f'**/{MANIFEST_FILENAME}'):
+        # Specifically target manifest files within their respective data directories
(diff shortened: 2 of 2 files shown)


quisquous:
Hmm, when I run this under mingw32, I get a giant pile of warnings like:
```
Warning: Found unlisted file 04-sb\eureka\eureka_pagos.js in data/ directory that does not exist within manifest.
Warning: Found unlisted file 02-arr\raid\t12.txt in data/ directory that does not exist within manifest.
Warning: Found unlisted file 03-hw\raid\a10s.js in data/ directory that does not exist within manifest.
Warning: Found unlisted file 05-shb\dungeon\qitana_ravel.txt in data/ directory that does not exist within manifest.
Warning: Found unlisted file 05-shb\dungeon\mt_gulg.txt in data/ directory that does not exist within manifest.
```

Also an error:
```
Error: Manifest entry for 04-sb/raid/o8n.js found within manifest, but no corresponding file found within data/ directory.
```

> Raidboss-specific warnings:
> * Trigger file exists within the manifest without timeline file
> * Timeline file exists within the manifest without trigger file

Just to clarify, this is trying to say that if a trigger file lists a timeline file, then the timeline file must also be in the manifest? And if a timeline file exists in the manifest, it must be referenced by a triggers file? (I think that plus the everything in the manifest exists on disk tests is really what needs to be tested?  I'm not quite sure I understand what these two bullet points mean otherwise).

panicstevenson:
> Hmm, when I run this under mingw32, I get a giant pile of warnings like:

Yep; thanks Windows. It should be fixed, I replaced the entries pulled from the manifest to use the operating system separator instead of forward slashes.

`return [manifest_entry.rstrip('\n').replace('/', os.sep) for manifest_entry in open(filepath)]`

quisquous:
Following up from discussion elsewhere:

I think it shouldn't be a warning if triggers don't have a timeline or vice versa.  I'd really mostly just want this to run perfectly cleanly on the existing stuff and trip ci if any of this was broken:

* all files in data/ need to be in the manifest (maybe move test/ to be a peer of data/)
* all files in the manifest need to exist in data/
* a timeline file in data/ should be referenced by a triggers file
* all trigger files that list a timeline file should have that file exist (this is already done)

panicstevenson:
Wrote a pretty big overhaul that should take care of the first two points (and nothing more); additionally, I wrote something for the third point as well:

https://gist.github.com/panicstevenson/f36508e533128f79c571dcd16b29b611

I updated the manifest integrity test on this PR, but I think that after writing the test to check that every timeline has a purpose that it makes even more sense to have all of these live within a similar suite of tests. I don't know how useful things like pytest or something would be here, considering the code itself isn't written in Python, but it might be useful to have something that can quickly run through each file and say:

✅ exists in manifest
✅ has a corresponding trigger file (if it's a timeline)
✅ has a timeline file that exists (if it's a trigger)
etc

Which could theoretically take in ManifestEntry, RaidbossTimeline, RaidbossTrigger, and OopsyraidsyTrigger(?) objects, and run their corresponding validations as well as their interdependent ones.

But I'm definitely overthinking it.

PR #563, opened 2019-08-23
Log line table goodness

JLGarber:
Alright, I believe this is all the markers that are currently referenced in Cactbot. I'm not 100% sure on the truth of column 4, as I didn't verify each one individually, but I'm reasonably confident on most of them. Now that I've gone through and done this, I'm less sure on the value to future contributors, but oh well, we have the data now. I know there are a few markers present in game that are missing from Cactbot and the guide, but I'll probably add that to the list of things I'm planning to revisit.

### DIFF
--- a/docs/LogGuide.md
+++ b/docs/LogGuide.md
@@ -681,13 +681,68 @@ Examples:
 1B:106F0213:Potato Chippy:0000:0EE3:0061:0000:0000:0000:
 ```
 
-The different headmarker types (e.g. `0027` or `0061` in the examples above) are consistent across fights in which marker they represent.  For example, `0039` is the meteor marker in Shinryu Ex adds phase and the Baldesion Arsenal Ozma fight.  The data following the type always appears to be zero in practice, although `Unknown1` and `Unknown2` infrequently have non-zero values.
+The different headmarker types (e.g. `0027` or `0061` in the examples above) are consistent across fights as far as which marker they *visually* represent. (Correct *resolution* for the marker mechanic may not be.)  For example, `0039` is the meteor marker in Shinryu Ex adds phase and the Baldesion Arsenal Ozma fight.  The data following the type always appears to be zero in practice, although `Unknown1` and `Unknown2` infrequently have non-zero values.
 
 Note: It's unclear when the head markers disappear.  Maybe `Unknown2` is a duration time? It's not clear what either of these unknown values mean.
 
 Also, this appears to only be true on later fights.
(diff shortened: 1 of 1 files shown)


docs/LogGuide.md
```diff
@@ -681,14 +681,69 @@ Examples:
...
 Note: It's unclear when the head markers disappear.  Maybe `Unknown2` is a duration time? It's not clear what either of these unknown values mean.
 
 Also, this appears to only be true on later fights.
 Turn 5 fireball and conflag headmarkers are actions from Twintania and not `1B` lines.
 It seems likely this was implemented later and nobody wanted to break old content by updating it to use newer types.
 
+Marker Code | Name | Sample Locations | Consistent meaning?
+--- | --- | --- | ---
+000[1-2, 4] | Prey Circle (orange) | O6s, The Burn boss 2 | Yes
+0005 | Missile Tether (white) | T8 n/s | N/A
```
quisquous: Can you move this tether (and other tethers) down to some new table in the `23` section?

JLGarber: I'm... not even sure why I added `0005` there. I just now did a quick unsynced run, and that's clearly not a `1B` line.

However, the `0061` orange tethers seen in the Vault and in Shinryu are the head markers, not the actual tethers. I just ran that quickly to verify. The head marker is orange chains, `0061`, and the chain tether itself appears to use the `23` log line:

`23:1067FF69:[PLAYER 1]:106E5E7F:[PLAYER 2]:2C30:0000:0009:106E5E7F:000F:776F6C52`

Anyway. I'll remove the `0005` tether. If you're okay with it, I'll also update the `23` section to state that the log line is not used for player-to-player tethers *using player skills* or something to that effect.

JLGarber:
Fair point on the casing. After a quick check through, it appears the guide uses lowercased references to raids, with no references to trials (that I didn't add) beyond "Shinryu Ex" and "Suzaku Ex". If you're comfortable with it, I'll change the raid references in my table to lowercase and any instance of [TRIAL] ex/Ex to [TRIAL] N/EX.

quisquous:
Sounds good to me, thanks!  Mostly, I just wanted to be consistent.

JLGarber:
Alright, I think this should resolve everything. I found a couple of markers I hadn't yet included, so those are in the table now. (Acceleration Bomb is probably useless given its mechanics, but at least it's documented.)

quisquous:
Thanks for all this!! It's great to have a better reference.

PR #564, opened 2019-08-23
jobs: add MP tick timer for BLM

haasn:
BLM is a job that relies heavily on exact MP tick timings, since getting
a lucky MP tick early on in the umbral ice phase allows BLMs to skip
needing a second filler spell for mana.

This commit adds support for tracking the MP tick timer. To detect MP
tick events, we hard-code the known constants for MP tick numbers (these
were known even before SHB, which is why I'm hard-coding them as a
percentage of max MP rather than using exact values, even though max MP
is now fixed at 10000).

To continue tracking the MP tick even when mana ticks are disabled (such
as in astral fire) or mana is high enough that excess mana from ticks is
cut off, we extend the timer-bar class to support seamlessly looping the
timer.

The user is able to control the display of this MP ticker as well as
hide it when out of combat, which is the default behavior.

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -16,6 +16,9 @@ let Options = {
   ShowHPNumber: ['PLD', 'WAR', 'DRK', 'GNB', 'BLU'],
   ShowMPNumber: ['PLD', 'DRK', 'BLM', 'AST', 'WHM', 'SCH', 'BLU'],
 
+  ShowMPTicker: ['BLM'],
+  ShowMPTickerOutOfCombat: false,
+
   MaxLevel: 80,
 
   JustBuffTracker: false,
@@ -43,6 +46,13 @@ let Options = {
   LowHealthThresholdPercent: 0.2,
   MidHealthThresholdPercent: 0.8,
 };
+
+let kMPNormalRate = 0.06;
+let kMPCombatRate = 0.02;
+let kMPUI1Rate = 0.30;
+let kMPUI2Rate = 0.45;
+let kMPUI3Rate = 0.60;
+let kMPTickInterval = 3.0;
 
 // Regexes to be filled out once we know the player's name.
 let kComboBreakers = null;
@@ -594,12 +604,14 @@ class Bars {
     this.hp = 0;
     this.maxHP = 0;
     this.mp = 0;
+    this.prevMP = 0;
     this.maxMP = 0;
     this.level = 0;
     this.distance = -1;
     this.whiteMana = -1;
     this.blackMana = -1;
     this.oath = -1;
+    this.umbralStacks = 0;
     this.inCombat = false;
     this.combo = 0;
     this.comboTimer = null;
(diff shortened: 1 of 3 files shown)


haasn:
Pushed and update to resolve the CI failures, but I don't have the game booted up atm so I can't check if the changes broke anything until tomorrow.

ui/jobs/jobs.js
```diff
@@ -1579,7 +1607,35 @@ class Bars {
...
+
+    let delta = this.mp - this.prevMP;
+    this.prevMP = this.mp;
+    if (!delta)
+      return;
+
+    let baseTick = this.inCombat ? kMPCombatRate : kMPNormalRate;
+    let mpTicks = [
+      baseTick,
+      baseTick + kMPUI1Rate,
```
quisquous: The umbral ice quantity is in jobDetail for black mage, if you want to store it.  Would that make this code easier, so you can just know what the rate is supposed to be?

haasn:
Good idea. Pushed an updated that made the MP tick calculations more robust (now also includes the disable during AF, and correctly rounds the MP ticks based on the pre-SHB logic).

I also made the MP ticker change color depending on the umbral state, since this more easily lets you see if you're going to get a tick or not.

Also, I swear, `eslint` makes it impossible to use ternary operators. It rejected this perfectly readable code (and various variants of it):
```javascript
    let umbralTick = this.umbralStacks == -1 ? kMPUI1Rate
                   : this.umbralStacks == -2 ? kMPUI2Rate
                   : this.umbralSTacks == -3 ? kMPUI3Rate
                   : 0;
```

e.g. in this case claiming that I should be indenting each line further than the previous (??? Has it ever *seen* a multi-line ternary?!)

Anyways, I'll just avoid the use of multiline ternaries. (I swear, some of eslint's recommendations are just outright unreasonable...)

haasn:
Incidentally, if we want to push this more in the direction of being an actual BLM job gauge, one improvement I can think of is to mark the point on the MP ticker where you can complete a fast F3/B3 cast before the MP ticker finishes, based on your current spell speed. (This is relevant for both B3 and F3 - in the case of B3, casting before this line means you will get the MP tick just after the cast finishes; and in the case of F3, casting *after* this line means you will get another MP tick during the F3 cast time)

But I've decided to leave it alone for now, also since I don't know of a good way to visually mark a certain portion of the timer bar.

quisquous:
Thanks, this is great!

Having a marker for the timer bar would be good, but right now there's no good way to know spellspeed.  You'd likely have to estimate that yourself from cast times?  (Or find memory signatures for character stats?)

Sorry about the ternary headaches.  If there are ternary-related eslint settings to make it do something more reasonable, feel free to send a pull request for that.  I mostly set up eslint to do what I liked and so I didn't have to think about it when looking at pull requests.

haasn:
This also doesn't work at all for classes with Piety (unless you're at base piety i.e. 380). If we could gain access to character stats we could compute the correct piety-enhanced MP tick.

PR #566, opened 2019-08-24
raidemulator: ignore npc objectIds

DrLippe:
Since all npc objectIds start with 4, they can be ignored.
SMN pets will no longer be detected as players.


DrLippe:
adding blumage icon & improve visibility of progress bar

quisquous:
Thanks for these fixes!!

PR #580, opened 2019-08-26
Add JSEvent for StatusEffects

opmega:
Reads player, target and focus status effects such as buffs and debuffs and exposes them via _onStatusEffectsUpdateEvent_ JavaScript event.

Unfortunately the game does not store the max buff durations so I added a dictionary to store the highest seen effect duration per ID.

Focus target offset was off.


quisquous:
Thanks for the pull request.  I should be able to take a look at this later this coming weekend.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -512,6 +573,16 @@ public class EntityData {
             data.debug_job += string.Format("{0:x2}", job_bytes[i]);
           }
         }
+
+        for (int i = 0; i < data.statusEffects.Length; i++) {
+          var currentBase = kEntityStructureOffsetCharacterDetails + kEntityStructureOffsetStatusEffects + i * StatusEffect.structSize;
+
+          var id = BitConverter.ToUInt16(bytes, currentBase + StatusEffect.offsetId);
```
quisquous: Would it be possible to not include all of the blank status effects with id zero?

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -512,6 +573,16 @@ public class EntityData {
...
+          var stackCount = bytes[currentBase + StatusEffect.offsetStackCount];
+          var duration = BitConverter.ToSingle(bytes, currentBase + StatusEffect.offsetDuration);
+          var casterId = BitConverter.ToUInt32(bytes, currentBase + StatusEffect.offsetCasterId);
```
quisquous: From a usability perspective, do you think `casterId` matters past "is this my effect"? I'm mostly just wondering if this should be a bool.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -438,6 +442,9 @@ public CactbotOverlay(CactbotOverlayConfig config)
       // The |focus| can be null when no focus target is selected.
       FFXIVProcess.EntityData focus = ffxiv_.GetFocusData();
 
+      // onStatusEffectUpdateEvent: Sends event with buff information for player, target and focus on every SendFastRateEvents
+      OnStatusEffectsUpdate(new JSEvents.StatusEffectsUpdateEvent(player,target,focus));
```
quisquous: style nit: please put spaces between parameters

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -368,6 +371,63 @@ public enum EntityJob {
...
+      private static Dictionary<int, float> maxDurations = new Dictionary<int, float>();
+      public ushort id = 0;
+      public byte stackCount = 0;
+      public float duration = 0;
+      public uint casterId = 0;
+      public float maxDuration = 0;
+
+      public static int offsetId = 0x0;
+      public static int offsetStackCount = 0x2;
+      public static int offsetDuration = 0x4;
```
quisquous: Sorry that the C# style is weird, but most of the C# code uses hacker_style_underscores and not camelCase.  The only time it uses camelCase is when it's for JSEvent structs that are serialized to Javascript, so that those can be in the right style.  It'd be nice if this could all be consistent.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -368,6 +371,63 @@ public enum EntityJob {
       return active_process_id == process_.Id;
     }
 
+    public class StatusEffect {
+      private static Dictionary<int, float> maxDurations = new Dictionary<int, float>();
```
quisquous: I think it'd be better to have this as a table in Javascript rather than sending this on every effect every frame.

CactbotOverlay/JSEvents.cs
```diff
@@ -72,6 +72,18 @@ public class PartyWipeEvent : JSEvent {
       public string EventName() { return "onPartyWipe"; }
     }
 
+    public class StatusEffectsUpdateEvent : JSEvent {
+      public StatusEffectsUpdateEvent(FFXIVProcess.EntityData player, FFXIVProcess.EntityData target, FFXIVProcess.EntityData focus) {
+        playerStatusEffects = player?.statusEffects;
+        targetStatusEffects = target?.statusEffects;
+        focusStatusEffects = focus?.statusEffects;
```
quisquous: Do you think it'd be worth optimizing the case where target and focus are the same?

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -512,6 +573,16 @@ public class EntityData {
             data.debug_job += string.Format("{0:x2}", job_bytes[i]);
           }
         }
+
+        for (int i = 0; i < data.statusEffects.Length; i++) {
+          var currentBase = kEntityStructureOffsetCharacterDetails + kEntityStructureOffsetStatusEffects + i * StatusEffect.structSize;
+
+          var id = BitConverter.ToUInt16(bytes, currentBase + StatusEffect.offsetId);
+          var stackCount = bytes[currentBase + StatusEffect.offsetStackCount];
+          var duration = BitConverter.ToSingle(bytes, currentBase + StatusEffect.offsetDuration);
```
quisquous: Oh wow, these are integers and not milliseconds remaining like on the job gauge.

opmega:
Thanks for your feedback, I really appreciate it, no hard feelings for criticism! 

My main goal with this was to have an alternative to FFXIV_ACT_Plugin combat log events for buff/debuff tracking, since calculating the durations in JS gave me inconsistent/off values in my custom job overlay (which most likely is a problem with my overlay, but after digging around in the games memory this approach of reading it directly seemed better to me).

I don't really see the use of having another _gainedStatus/lostStatus_ event, when the combat log events from the FFXIV_ACT_Plugin already do this. 
Therefore I decided to send all of the status effects with durations read from memory, which makes the JS overlay a pure presentation layer updating on events, instead of having to update/calculate an internal state of the status effects.

About the FFXIV_ACT_Plugin I don't know what you mean by digging into its code, do you mean by reversing it with dnSpy or something? I can't find any public source code. If it already has the capabilities to read entity status effects it would of course be redundant to reimplement this logic in cactbot.

Going forward I will try to

1. Convert the StatusEffect arrays to lists to only track actual buffs with an id, no "empty" StatusEffects
2. Instead of using the _caster_id_, check if the caster was the player and use something like a boolean _is_players_effect_ instead
3. Split event into seperate events for player, target and focus
4. Create those events only if a player, target or focus exists respectivley
5. Check if any of player, target or focus are the same to reuse the data
6. Update code format and style acccording to your feedback

Thank you so much for your time and this amazing piece of software, making FF14 actually playable for me :)

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -45,10 +45,11 @@ public class FFXIVProcess {
     // TargetStruct {
     //   0x00 bytes in: EntityStruct* target;
     //   ...
-    //   0x78 bytes in: EntityStruct* focus;
+    //   0x58 bytes in: EntityStruct* mousoverTarget
```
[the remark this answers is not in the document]
opmega: the focus offset is in line 49, I just added the mouseover offset for future reference since I stumbled on it

CactbotOverlay/FFXIVProcess.cs
opmega: This isn't sent with the JSEvent at all, it is only used to internally store the maximum durations by id for already seen status effects. Each status effect gets only its own max duration in the constructor.

quisquous:
> Thanks for your feedback, I really appreciate it, no hard feelings for criticism!

Ok, thanks! I am always reluctant to send folks who have worked hard on a large change feedback that's like "good idea, but could you just do this entiiiiirely differently??"

> My main goal with this was to have an alternative to FFXIV_ACT_Plugin combat log events for buff/debuff tracking, since calculating the durations in JS gave me inconsistent/off values in my custom job overlay (which most likely is a problem with my overlay, but after digging around in the games memory this approach of reading it directly seemed better to me).

Yeah, I hear that.  I think here are my priorities:
* make things possible in cactbot that aren't possible
* don't add memory locations when possible
* don't use too much cpu
* make it easy to write cactbot overlays

This patch doesn't make things possible that you couldn't do before.  It also adds memory locations.  I think the biggest issue is that it seems a bit of a cpu waste to read memory to constantly re-send a times and effects that Javascript already knows.  And, only sending those integer times means that you can't show like "5.2 seconds" remaining or a countdown bar that smoothly moves, which limits the kind of ui that you can make.

If you send every event every time, then that means on the javascript side you have limited choices. 
 It seems like you have to either (1) reconstruct all of the dom objects every frame from scratch or (2) you have to do a comparison of the effects that you were sent and the effects that you have (which seems like more work than just responding to gained/removed).  Maybe from a lines of code perspective, this is simpler, but it's a lot of work in the whole pipeline to save a bit of javascript.

That's just my intuition, sorry.

> I don't really see the use of having another _gainedStatus/lostStatus_ event, when the combat log events from the FFXIV_ACT_Plugin already do this.

This is true, but it's also a bit cumbersome to parse every log line ever to find status effects.  If ravahn's plugin is already tracking what's new and old, it seems not terrible to me to send a single event that's just like "here's a new status effect to track."  This fits in the "make it easier to write overlays" priority above.  But, you're also right that you can do this from log lines alone, and so maybe that's the better approach here.

> About the FFXIV_ACT_Plugin I don't know what you mean by digging into its code, do you mean by reversing it with dnSpy or something? I can't find any public source code. If it already has the capabilities to read entity status effects it would of course be redundant to reimplement this logic in cactbot.

If you add the FFXIV_ACT_Plugin SDK download as references, you can see its exposed public API.  It's got a DataRepository and a number of events to subscribe to.

> Thank you so much for your time and this amazing piece of software, making FF14 actually playable for me :)

:heart:

Thanks for the pull requests.  They're much appreciated!

CactbotOverlay/FFXIVProcess.cs
opmega: It is actually a single precision float

quisquous: When I printed these out, they sure looked like integers.  Are they getting converted at some point?

quisquous:
I think I am going to close this pull request for now.  As you say, this is still very do-able from JS by parsing log lines, and I think that's the best way to handle this for now.

Issue #576, opened 2019-08-25
jobs: show shield values


quisquous:
Currently jobs attempts to replace your health bar with something else.  It'd be cool if that also knew what shields were on you.  If somebody wants to figure out where in memory this is so that the plugin can report this back to javascript in the player details, that would be really neat.

See cheat engine instructions: https://github.com/quisquous/cactbot/blob/master/docs/MemorySignatures.md

opmega:
At `kEntityStructureOffsetCharacterDetails+0x5c` is a uint32 representing the percentage of max hp shielded, i.e. 16 for 16%.

I haven't found an absolute value but with actual HP values right above it can be calculated if needed.

haasn:
Oops, didn't see @opmega already found the address. I can confirm the same offset from my own findings, and it works for anything I tested it on (even when the game itself doesn't show the absorb amount). It does appear to be relative to maximum HP, and updates automatically as your max HP changes.

One thing I did notice was that it caps out at 200, no matter how big you make the shield. So I'm not sure if it's a uint32, we might as well treat it as a uint8 to be on the safe side.

haasn:
It also appears to be rounding unfavorably in some situations. For example, Manaward, which reads "absorbs 30% of max HP", only shows up as 29 in the client.

opmega:
Very valid points @haasn , if it caps out at 200 it might even be a ushort, but I will make it read as uint16 for now to be safe and update the PR. 

As for the rounding error, I don't see a way around it if there is no other, more accurate shield value stored somewhere. I guess the value at 0x5c is only used to determine the party shield bar length and therefore it doesn't need to be any more precise for the game itself.

quisquous:
Yeah, I agree that this value is just the party bar length and the real value is just kept on the server.  It's still better than nothing, even if it's a little inaccurate.

I changed the title on this issue to be more about updating the jobs module to show this in the hp bar now that the plugin issues got sorted by @opmega.

quisquous:
This got added in cfc71d724.

PR #581, opened 2019-08-26
Add shield value to entities

opmega:
Proposed implementation for #576


quisquous:
This is great, thanks!

Issue #569, opened 2019-08-25
oopsy: add entries for all the 5.x levelling dungeons


quisquous:
I added oopsy entries for the trials, raids, and level 80 dungeons, but I didn't do it for any of the leveling dungeons.  If anybody wants to do this, it'd be appreciated.  It should be straightforward, but I just haven't done it.

The process here is to run the dungeon, and then look at the logs in ACT (or import one of your old logs and look at that).  I usually search for `E00000000` and find all the actions that (hopefully) nobody stood in, and then add them to an oopsy file, like this: https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/data/amaurot.js

I think a first pass could just list "things you shouldn't stand in".  If there's things that should only hit one person or bad effects you shouldn't get, then add them like this: https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/data/e2s.js#L15-L33

JLGarber:
I can do this with my trusts this week, at least for Holminster. Should be able to get Malkiah's Well and Mt Gulg as well.

JLGarber:
As an update, for anyone passing by to see this, the only dungeons left for the first pass are Dohn Mheg and Qitana Ravel for Shadowbringers. I'll probably have those in tomorrow/Wednesday if there are no pull requests in at that point.

@quisquous did you want to add previous expansions to this as well? I probably won't bother running the dungeons specifically to find oopsy stuff, but if I get them I have no problem putting something together.

quisquous:
Yeah, previous expansions would be great if you want to add them!

JLGarber:
Finally getting back to this. What's your preferred method for obtaining the names of everyone in the party? Specifically, I'm trying to put together a warning if a player fails to step into the tethers for Growth on the second boss in Dohn Mheg. When it's done correctly it looks like this:

```
[02:00:17.048] 25:4000C388:Griaule:00002F34:3140535:3774000:10000:10000:0::7.156433:-339.1318:23.13342:-0.04491758:03E8:0:0:01:02000186:01:44160000:4000C38B:
[02:00:17.048] 1A:4000C388:Griaule gains the effect of Growing from Painted Sapling for 600.00 Seconds.
[02:00:17.048] 25:12345678:Player 1:00002F35:40761:57059:10000:10000:0::14.56638:-338.9952:23.05715:-1.556229:03E8:1700:0:01:0100017F:01:44160000:4000C38C:
[...]

PR #584, opened 2019-08-26
Oopsy holminster

JLGarber:
This is my first pass at this, per #569. It's not perfect, but given that I ran this as a trust I didn't completely test Fierce Beating to determine which component is which. (My untested assumption is that one element is the actual fist contact, and the other two are the Exaflare markers. Cat O' Nine Tails is the rear attack.)

I'm not sure whether it lines up with your philosophy, but I put Taphephobia (the purple spread circles) as a failure given that it's likely not just going to kill the player but their teammate as well. Aethersup (the frontal cone immediately after the gaol break) is also tagged as a failure because there's no other party damage around that time, so standing in it for the 4-5 ticks required to die is solely due to mechanical failure.

Flagellation might also need to be marked as a failure, but given that it's a faster mechanic and more dependent on teammates, I felt it was best to leave it as a warning.


quisquous:
Thanks!  Much appreciated.  :)

Issue #572, opened 2019-08-25
oopsy: rearrange files


quisquous:
oopsy doesn't have quite as many data files as raidboss, but it'd be nice to move them into the same folder structure that raidboss now has for consistency before it gets overwhelming.

Issue #548, opened 2019-08-17
Add more raidboss testing


quisquous:
There's a bunch of common things that could be checked with automatic (python?) testing that would prevent obvious bugs:

* multiple entries of the same type in the same dict (two `regex` instead of one `regex` and one `regexFr`) or multiple `triggers` entries
* missing timeline files
* trigger and timeline missing from manifest file
* timeline triggers translated (they never should be)
* translation block exists, but doesn't translate all the syncs / texts

This should probably also apply to oopsy files as well.

panicstevenson:
A very simple test can be made for checking the manifest integrity: making sure all the files listed in both the [raidboss manifest](https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/manifest.txt) and the [oopsyraidsy manifest](https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/data/manifest.txt) exist, and make sure the opposite is true as well: everything in the actual data/ folders exist within their respective manifest files.

The test can also be added to the [Travis CI run](https://github.com/quisquous/cactbot/blob/master/.travis.yml).

It should be a relatively easy test to add; I can add it if no one else gets around to it, but, if someone wants to pick it up, it's a good first task to get your feet wet.

quisquous:
Yes yes, I definitely intended all of these thoughts to be part of travis ci!

quisquous:
Probably should also add all the runtime checks from: https://github.com/quisquous/cactbot/blob/master/ui/raidboss/popup-text.js#L47

quisquous:
Another random testing thought that would be nice is that all the keys in a trigger should be listed in the order that they run.  That's a little obnoxious for linting, but at the same time, I think it would add a lot of readability and consistency.

quisquous:
I think this is mostly done.

PR #588, opened 2019-08-28
test: Update Cactbot Test Suite

panicstevenson:
Trying to make things a little more uniform for the test suite; ultimately, I'm not a huge fan of where this pull request leaves things, but I think it's a good place to build off of.

Realistically, I'd like to create a way to test individual mechanisms within each type of file, and have some process that gathers all the information before executing the tests. In essence, it'd be really nice to have a test suite that does as follows:

- Python script (probably) to gather all of the information necessary to execute the tests
  - Manifest entries for each module that has a manifest, currently raidboss and oopsyraidsy
  - Raidboss timelines
  - Raidboss triggers
  - Oopsyraidsy triggers (name?)
- Python script (probably) to run down each type of test
  - Manifest entries (string)
    - Validate manifest entries against their data directory counterparts
    - Well formed, see https://github.com/quisquous/cactbot/issues/572
  - Raidboss timelines (.txt)
    - Needs a trigger counterpart that references it as a `timelineFile`
  - Raidboss triggers (.js)
    - Compiles (?)
    - Individual, module-specific test files, written in JavaScript (probably), see https://github.com/quisquous/cactbot/issues/548
  - Oopsyraidsy triggers (.js)
    - Compiles (?)
    - Individual, module-specific test files, written in JavaScript (probably), see https://github.com/quisquous/cactbot/issues/548

With a file structure looking something like:
```
test/
   test1.py
   test2.py
   test3.py
   raidboss/
      test1.js
      test2.js
   oopsyraidsy/
      test1.js
```

And (ideally) this would utilize an actual test framework, such that a test runner script isn't necessary and that we can actually harness meaningful information from our TravisCI output.

Quick note on `(probably)`s:
I personally don't care what is used; to me, it seems better to have things that are beyond the scope of JavaScript-specifics (ie probably something to do with the filesystem) to be invoked with Python, and have the JavaScript specifics tested via JavaScript, but that's just me and I am easily swayed.


quisquous:
The testing """infrastructure""" in cactbot is mostly just things I've stuck in there, but haven't had much thought towards how it should be shaped.  So, I'm more than willing to take whatever input you have here.

One concern I have is that your change does seem to get rid of check_timelines.py, though, which makes sure that the timelines are well-formed? It's using javascript because then it can share the same code in timeline.js.

I think your overall thought of having something well-structured and using an actual test framework such that it could return better travis ci output sounds great.

panicstevenson:
I thought that everything from `check_timelines.js` was covered in the new `test_timelines.py`, but you're right that that's definitely not the case. I'll get that updated. 👍

quisquous:
Thanks! I like where this is going, and having a bit more organization about directories and modules creates a structure where it's clear where to follow up with more testing.

quisquous:
Did you want this merged at this point (it's a draft), or are you still working on it?

panicstevenson:
I did have plans to merge it; but, I meant to ask how you were feeling about the language split. Originally I really liked it because I feel like Python(3)'s ability to walk the filesystem is very powerful and pretty intuitive, but it feels a bit odd to me now that I've actually added a JavaScript test. Nothing's ever set in stone though, so we can iterate, but if you're good with the concept I'll mark it ready for review.

quisquous:
No no, I like this sort of module (python) -> file (python) -> individual tests (python or js) model.  I think it's a lot more clear than having sort of standalone tests that try to do everything.

panicstevenson:
Yea, I really like it conceptually. I think the line that threw me off was adding:

```
let inputFilename = String(process.argv.slice(2));
```

which gave off some sort of code smell to me. We can see how it plays out adding a few more tests from #548; worst case scenario we have more tests and we change it down the road.

test/timeline/test_timeline.js
```diff
@@ -0,0 +1,25 @@
...
+
+// This test loads an individual raidboss timeline and makes sure
+// that timeline.js can parse it without errors.
+
+let fs = require('fs');
+let Timeline = require('../../ui/raidboss/timeline.js');
+
+let exitCode = 0;
+
+let inputFilename = String(process.argv.slice(2));
```
quisquous: This doesn't bother me so much? It's the same thing you'd have in a Python script called from elsewhere.  It's a little weird, but this just feels like any other command line intermediary between scripts in two languages.

PR #577, opened 2019-08-25
Request for comments on OverlayPlugin port

ngld:
Sorry in advance for the big wall of text...

I've ported Cactbot to the new addon API for my OverlayPlugin fork.
These changes currently break compatibility with the old OverlayPlugin (0.3.4.0 and earlier). However, before I start working on that I'd like some feedback.

The main changes are that I've split the responsibilites of the older overlay types into two categories:
Overlay types are only responsible for rendering overlays. Events are handled by `EventSources` instead. During startup OverlayPlugin creates a new instance for each registered `EventSource`. This instance then registers all event types it will dispatch and "handlers" which are just functions which can be called from JavaScript (`RegisterEvents` and `RegisterHandler` calls in the constructor).
Since there's just one instance of each `EventSource`, it doesn't know which overlay it's talking to. Instead, each overlay can subscribe to events and will then receive those whenever they're dispatched.
This means that Cactbot overlays can subscribe to both Cactbot events as well as MiniParse events.

The communication between OverlayPlugin and JavaScript is mostly handled through a few functions in `resources/common.js`. Most importantly `addOverlayHandler` (which works mostly the way `document.addEventListener` did) and `callOverlayHandler` (which allows the scripts to call the handlers registed in `CactbotEventSource`).
These functions hide the difference between running in OverlayPlugin directly and running in a web browser and receiving data over a WebSocket connection. The idea is that if most code doesn't know whether it's running in OBS or an overlay, that feature is less likely to break (and easier to maintain).

I've had several issues where the web page loads before the load event arrives in C#. Probably because ACT is doing a lot during startup and processes the event too late (it's not as much of a problem if a an overlay is reloaded afterwards). To avoid race-conditions, etc. I've restructured the code a bit to wait until the necessary APIs are available.

Most of the C# code should be unchanged. The only major change is that I've split a few things out of `SendFastRateEvents` and turned them into handlers which the overlays can call as necessary (data files, configuration, notify state resets).
Some things aren't as nice as they could be (several JavaScript files just reset the notification state to trigger events instead of requesting the information they need directly (which would avoid a bit of overhead since events will be dispatched to all subscribed overlays and not just the one that requested them).

Either way, in the current state this works for me as well as a few beta testers. Both as a normal overlay as well as in OBS / browser / whatever. Let me know if you have any questions regarding this code or my changes to OverlayPlugin.

Oh, and why did you want to run the overlay with over 60 FPS? It runs perfectly smooth for me at 60. :D

PR #593, opened 2019-09-01
cactbot.dll: Adding filewatcher to user and resource directory

### LINKED ISSUE #603: New 'Reload overlay on file change' setting isn't saved
ACT doesn't remember this option like the others, when I close ACT it will be disabled.

DrLippe:
- [x] Listen to user directory based on custom user directory setting
- [x] Listen to directories based on `href=""` and `src=""` paths in overlay html
- [x] Moving code to `CactbotOverlay.cs`
- [x] Saving state
- [x] checked working

Resolves #603 . Can be merged.

### DIFF
--- a/CactbotOverlay/CactbotOverlay.cs
+++ b/CactbotOverlay/CactbotOverlay.cs
@@ -34,7 +34,7 @@ namespace Cactbot {
     private List<string> last_log_lines_ = new List<string>(40);
     private List<string> last_import_log_lines_ = new List<string>(40);
 
-    public FileSystemWatcher[] watchers;
+    private FileSystemWatcher[] watchers;
 
     // When true, the update function should reset notify state back to defaults.
     private bool reset_notify_state_ = false;
@@ -45,6 +45,7 @@ namespace Cactbot {
 
     // Temporarily hold any navigations that occur during construction
     private string deferred_navigate_;
+    private bool watcherState = false;
 
     private StringBuilder dispatch_string_builder_ = new StringBuilder(1000);
     JsonTextWriter dispatch_json_writer_;
@@ -59,7 +60,7 @@ namespace Cactbot {
     private FightTracker fight_tracker_;
     private WipeDetector wipe_detector_;
     private string language_ = null;
-    private System.Threading.SynchronizationContext main_thread_sync_;
+    private SynchronizationContext main_thread_sync_;
   
 
     public delegate void GameExistsHandler(JSEvents.GameExistsEvent e);
@@ -104,7 +105,7 @@ namespace Cactbot {
 
     public CactbotOverlay(CactbotOverlayConfig config)
         : base(config, config.Name) {
-      main_thread_sync_ = System.Windows.Forms.WindowsFormsSynchronizationContext.Current;
+      main_thread_sync_ = SynchronizationContext.Current;
     }
 
     void Initialize() {
@@ -223,6 +224,7 @@ namespace Cactbot {
       init_ = true;
       if (deferred_navigate_ != null) {
         this.Navigate(deferred_navigate_);
+        this.SetupFileWatcher(deferred_navigate_);
       }
     }
 
@@ -250,17 +252,33 @@ namespace Cactbot {
       base.Navigate(url);
       fast_update_timer_semaphore_.Release();
     }
- 
-    public void SetupFileWatcher(string url)
+
+    public void EnableWatchers(bool check)
     {
-      if (watchers.Length > 0)
-        foreach(var watcher in watchers)
+      if (!init_ || this.watchers == null)
+      {
+        this.watcherState = check;
+        return;
+      }
+      // I doubt that the following code is actually necessary
+      if (this.watchers.Length > 0)
+        foreach (var watcher in this.watchers)
+          watcher.EnableRaisingEvents = check;
+    }
+ 
+    private void SetupFileWatcher(string url)
+    {
+      if (!init_)
+        return;
+      if (this.watchers != null && this.watchers.Length > 0)
+        foreach(var watcher in this.watchers)
           watcher.Dispose();
       List<string> pathsList = new List<string>();
       
       // Overlay Dir
       if (url == "")
         return;
+
       var overlayFullPath = url;
       overlayFullPath = Regex.Replace(overlayFullPath, @"file:[\\\/]+", "");
       var overlayPath = Path.GetDirectoryName(url);
@@ -273,7 +291,8 @@ namespace Cactbot {
       Dictionary<string, string> localUserFiles = new Dictionary<string, string>();
       GetUserConfigDirAndFiles(out userPath, out localUserFiles);
       userPath = Regex.Replace(userPath, @"file:[\\\/]+", "");
-      pathsList.Add(userPath);
+      if(!userPath.Contains(basePath))
+        pathsList.Add(userPath);
 
       // Path from HTML
       string html = File.ReadAllText(overlayFullPath);
@@ -282,8 +301,8 @@ namespace Cactbot {
       foreach(var match in htmlPath.Groups) {
         string matchPath = match.ToString();
         if (matchPath.StartsWith(".."))
-          matchPath = Path.Combine(overlayFullPath, matchPath);
-        if (!pathsList.Contains(matchPath))
+          matchPath = overlayPath + "/" + matchPath;
+        if (!matchPath.Contains(basePath) && Directory.Exists(matchPath) && !pathsList.Contains(matchPath))
           pathsList.Add(matchPath);
       }
 
@@ -306,10 +325,16 @@ namespace Cactbot {
         watcher.Deleted += (o, e) => { Navigate(url); };
         watcher.Renamed += (o, e) => { Navigate(url); };
         watcher.Changed += (o, e) => { Navigate(url); };
-        watcher.EnableRaisingEvents = false;
+        watcher.EnableRaisingEvents = this.watcherState;
 
         watchers[i] = watcher;
       }
+      for (int i = 0; i < paths.Length; i++)
+      {
+        paths[i] = Regex.Replace(paths[i], @"[\\\/]\z", "");
+        paths[i] = Regex.Replace(paths[i], @".*[\\\/]", "");
+      }
+      LogInfo($"Setup file watchers for {paths.Length} directories: {String.Join(", ", paths)}");
     }
 
     private void OnLogLineRead(bool isImport, LogLineEventArgs args) {
--- a/CactbotOverlay/CactbotOverlay.cs
+++ b/CactbotOverlay/CactbotOverlay.cs
@@ -1,4 +1,4 @@
-﻿using Advanced_Combat_Tracker;
+using Advanced_Combat_Tracker;
 using Newtonsoft.Json;
 using RainbowMage.OverlayPlugin;
 using System;
@@ -255,7 +255,7 @@ namespace Cactbot {
 
     public void EnableWatchers(bool check) {
       this.watcher_state = check;
-    foreach (var watcher in this.watchers)
+      foreach (var watcher in this.watchers)
         watcher.EnableRaisingEvents = check;
     }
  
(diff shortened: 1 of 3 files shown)


CactbotOverlay/CactbotOverlayConfigPanel.cs
```diff
@@ -69,28 +69,47 @@ public partial class CactbotOverlayConfigPanel : UserControl {
...
-      {
-        this.overlay.LogError("Directory does not exist!" + path);
-        return;
+      var overlayPath = System.IO.Path.GetDirectoryName(config.Url);
+      overlayPath = System.Text.RegularExpressions.Regex.Replace(overlayPath, @"file:[\\\/]+", "");
+      pathsList.Add(overlayPath);
+
+      // User Dir
+      string basePath = System.Text.RegularExpressions.Regex.Match(overlayPath, @"(.*cactbot)").Groups[1].ToString();
+      string userPath = System.IO.Path.Combine(basePath, "user");
```
quisquous: The user directory can exist in a number of places.  It can be relative to the html, relative to the dll, or specified explicitly as an option.

DrLippe:
> Rather than listing all these separate paths, do you think it'd be worth just walking up and watching the entire cactbot directory? It does mean that plugins might reload more than they should, but it does also make this simpler too.

Sure. I thought maybe I should stick to this to prevent other overlays from reloading. But since this feature is off by default on every start this might not even be an issue. Will change that.

> The user directory can exist in a number of places. It can be relative to the html, relative to the dll, or specified explicitly as an option.

Yeah, I though of this. I might even have to parse the html file to get all directories involved.

ngld:
Please read my comment on #603.
Another unrelated comment: In my opinion, `SetupFileWatcher` should be in `CactbotOverlay` and not in `CactbotOverlayConfigPanel` since it has nothing to do with the user interface.

DrLippe:
⚠️ Changes are not tested yet!

DrLippe:
- [x] Listen to user directory based on custom user directory setting
- [x] Listen to directories based on `href=""` and `src=""` paths in overlay html
- [x] Moving code to `CactbotOverlay.cs`
- [x] resolved #603
- [x] checked working

Can be merged.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,90 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+      }
+      // I doubt that the following code is actually necessary
+      if (this.watchers.Length > 0)
+        foreach (var watcher in this.watchers)
+          watcher.EnableRaisingEvents = check;
+    }
+ 
+    private void SetupFileWatcher(string url)
+    {
+      if (!init_)
```
quisquous: init_ will always be true when you get here, just based on where this is called.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,90 @@ public CactbotOverlay(CactbotOverlayConfig config)
       fast_update_timer_semaphore_.Release();
     }
 
+    public void EnableWatchers(bool check)
+    {
+      if (!init_ || this.watchers == null)
```
quisquous: I think if you initialize watchers in its declaration to be [], then this function could be much shorter and look like this:
```
public void EnableWatchers(bool check) {
  this.watcher_state = check;
  foreach (var watcher in this.watchers)
    watcher.EnableRaisingEvents = check;
}
```

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,90 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+          watcher.Dispose();
+      List<string> pathsList = new List<string>();
+      
+      // Overlay Dir
+      if (url == "")
+        return;
+
+      var overlayFullPath = url;
+      overlayFullPath = Regex.Replace(overlayFullPath, @"file:[\\\/]+", "");
+      var overlayPath = Path.GetDirectoryName(url);
```
quisquous: This doesn't work when your url is remote, e.g. `https://quisquous.github.io/cactbot/ui/raidboss/raidboss.html`

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -345,14 +435,14 @@ public CactbotOverlay(CactbotOverlayConfig config)
         // is relative to the pointed to url and not the local file.
         if (url.StartsWith("file:///")) {
           var html = File.ReadAllText(new Uri(url).LocalPath);
-          var match = System.Text.RegularExpressions.Regex.Match(html, @"<meta http-equiv=""refresh"" content=""0; url=(.*)?""\/?>");
+          var match = Regex.Match(html, @"<meta http-equiv=""refresh"" content=""0; url=(.*)?""\/?>");
```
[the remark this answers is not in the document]
DrLippe: sure

CactbotOverlay/CactbotOverlay.cs
DrLippe: Ouch, yeah didn't think of that

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -219,6 +224,7 @@ public CactbotOverlay(CactbotOverlayConfig config)
       init_ = true;
       if (deferred_navigate_ != null) {
         this.Navigate(deferred_navigate_);
+        this.SetupFileWatcher(deferred_navigate_);
```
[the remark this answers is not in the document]
DrLippe: I guess, I'll hand this problem over to you?

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+        return;
+
+      foreach(var watcher in this.watchers)
+        watcher.Dispose();
+
+      List<string> path_list = new List<string>();
+      var overlay_full_path = url;
+      string base_path = null;
+
+      if (!url.Contains("://quisquous.github.io/cactbot/")) {
```
quisquous: Maybe this should check if it starts with `file://` instead? Folks could use a different location for remote urls.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+        // Path of overlay if overlay is not remote!
+        overlay_full_path = Regex.Replace(overlay_full_path, @"file:[\\\/]+", "");
+        var overlay_path = Path.GetDirectoryName(url);
+        overlay_path = Regex.Replace(overlay_path, @"file:[\\\/]+", "");
+        base_path = Regex.Match(overlay_path, @"(.*cactbot)").Groups[1].ToString();
+
+        path_list.Add(base_path);
+
+        // Path of HTML resources if overlay is not remote!
```
quisquous: I don't understand this section.  Are you expecting things outside the `base_path` to be included?

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+        var watcher = new FileSystemWatcher() {
+          Path = path_list[i],
+          NotifyFilter = NotifyFilters.LastWrite | NotifyFilters.FileName,
+          IncludeSubdirectories = true,
+        };
+
+        watcher.Created += (o, e) => { Navigate(url); };
+        watcher.Deleted += (o, e) => { Navigate(url); };
+        watcher.Renamed += (o, e) => { Navigate(url); };
+        watcher.Changed += (o, e) => { Navigate(url); };
```
quisquous: I don't know if it matters, but if I save a file, I definitely get two reloads instead of just one.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+        watcher.EnableRaisingEvents = this.watcher_state;
+
+        watchers[i] = watcher;
+      }
+      for (int i = 0; i < path_list.Count; i++) {
+        path_list[i] = Regex.Replace(path_list[i], @"[\\\/]\z", "");
+        path_list[i] = Regex.Replace(path_list[i], @".*[\\\/]", "");
+      }
+      LogDebug($"Setup file watchers for {path_list.Count} directories: {String.Join(", ", path_list)}");
```
quisquous: I think this is too noisy to print out in the log.

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+            match_path = overlay_path + "/" + match_path;
+          if (!match_path.Contains(base_path) && Directory.Exists(match_path) && !path_list.Contains(match_path))
+            path_list.Add(match_path);
+        }
+      }
+
+      // User Dir
+      string user_path;
+      Dictionary<string, string> local_UserFiles = new Dictionary<string, string>();
+      GetUserConfigDirAndFiles(out user_path, out local_UserFiles);
```
quisquous: This is pretty unfortunate, as it causes all the user files to get loaded twice.  Is there any way around that?

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
       fast_update_timer_semaphore_.Release();
     }
 
+    public void EnableWatchers(bool check) {
+      this.watcher_state = check;
+    foreach (var watcher in this.watchers)
```
quisquous: Can you indent this properly?

CactbotOverlay/CactbotOverlay.cs
```diff
@@ -247,6 +253,79 @@ public CactbotOverlay(CactbotOverlayConfig config)
...
+            path_list.Add(match_path);
+        }
+      }
+
+      // User Dir
+      string user_path;
+      Dictionary<string, string> local_UserFiles = new Dictionary<string, string>();
+      GetUserConfigDirAndFiles(out user_path, out local_UserFiles);
+      user_path = Regex.Replace(user_path, @"file:[\\\/]+", "");
+      if(base_path == null || !user_path.Contains(base_path))
```
quisquous: Can you add a space after if?

CactbotOverlay/CactbotOverlay.cs
DrLippe: I saw you were using this method somewhere else to determine whether its a local or remote url.
But doesn't the overlay accept local urls lacking the `file:///`-protocol?

CactbotOverlay/CactbotOverlay.cs
DrLippe: For this case:
```
Desktop/
  myOverlay/
    html/
      overlay.html
    css/
      style.css
      image.png
    js/
      script.js
      anotherscript.js
```
However I will delete this section and simply require the `html` to be in the root directory.

CactbotOverlay/CactbotOverlay.cs
DrLippe: Since the `user_path` is not assigned to an accessible variable, I guess not?

CactbotOverlay/CactbotOverlay.cs
DrLippe: For the same overlay? 
You may get reloads for every overlay that has the file watcher activated and is located within the same `base_path`.

quisquous: Hmm.  I have two watchers, one for cactbot/ and one for cactbot/user/.  If I add a LogDebug message into the Changed callback, and then change a file that's in cactbot/ui/, I see that watcher trigger twice.  Does that not happen for you?

CactbotOverlay/CactbotOverlay.cs
quisquous: Hmm, that's fair.  What about checking for http or https instead?

ngld:
Just a heads up from me regarding PR #577: As I pointed out in https://github.com/quisquous/cactbot/pull/577#issuecomment-528301355, this PR will mostly be wiped out because #577 removes the `CactbotOverlay` class and the `CactbotEventSource` class which replaces it has no access to overlays.
The best solution to this would be merging this logic into `MiniParseOverlay` but I'm hesitant to do so ~~since IMO the logic to detect local files isn't too robust~~ (sorry, just realized that part was removed).

Still, this PR adds nothing that's not already possible with something like Webpack's dev server or [livereload](https://github.com/napcs/node-livereload). Both of which offer more features and options than this implementation. And they work with WebSocket mode as well which this wouldn't...

CactbotOverlay/CactbotOverlay.cs
ngld: Depending on how your text editor saves files, it can trigger multiple FS events. For example, if you have `atomic_save` enabled in Sublime Text, it creates a new file each time you save and replaces the old file with the new one ([here's a more detailed explanation](https://stackoverflow.com/questions/20634684/what-is-sublime-text-doing-when-i-save-a-file)).
Basically, the only way around this is [debouncing](https://lodash.com/docs/4.17.15#debounce) the reloading.

quisquous:
Hmm, ok.  So, maybe the thing to do here is just to fix #603 (saving the setting properly) in both the master and the new ngldoverlay branches, and not try to be fancy with user directories.

ngld:
Watching the user directory should be fine since that's a global setting. We can even reload overlays after changes by sending an event (`onConfigChanged`?) which triggers `location.reload()` in the overlay. That part makes a lot of sense, too, since it's useful to users while the overlay files will most likely only be touched by devs which can use other tools.

The only problem is the dependency on the overlay's URL since we can't easily get that. There are ways to work around that part (send the URL from JavaScript, watch the `ui` and `resource` directories, send an event on file change and then filter in JavaScript to determine which overlay to reload) but that seems like more work than it's worth.

nenquelote:
Hello, I have a compatibility problem can you help me?

image

ngld:
Not the right place for this. However, you downloaded the hibiya build of Cactbot and used it with the ngld OverlayPlugin. Download the ngld build instead.
If you still have issues with that, do **not** post here instead open a new issue, please.

quisquous:
Sorry for all the conflicts with plugin changes.  Because I wanted to fix issue #603, I landed a small change in fa08dd8 just to do that bit.

I think if you'd like to add reload logic based on the user directory as ngld says in https://github.com/quisquous/cactbot/pull/593#issuecomment-541475174, I'd appreciate if you did it on top of the ngldoverlay branch.

For now, I'm going to close this, because it doesn't make sense on top of the ngldoverlay branch.  Sorry for all the hassle.

Issue #568, opened 2019-08-25
jobs: move icons to pngs


quisquous:
Right now, for whatever reason all the job icons for the big buff tracker are in data urls in job-icons.js.  This is super confusing.  They should probably all get turned into real pngs and get stored in `resources/icon/status` next to all the other job icons there.

PR #598, opened 2019-09-03
jobs: Convert Buff Icon Data URLs to PNG

panicstevenson:
Convert job-icons.js data URLs to individual buff icons in PNG format.

Closes https://github.com/quisquous/cactbot/issues/568


quisquous:
Would you mind also removing the icons js file and updating jobs to use these new files?

panicstevenson:
Done! I was trying to keep the changes as small as possible, but I guess it was probably too small to even be a functional change

panicstevenson:
Missed one; luckily it was pretty simple to catch in testing!

quisquous:
Thank you so much for taking care of this!

Issue #567, opened 2019-08-25
jobs: add other astro cards to listed buffs


quisquous:
Right now only balance is listed as a card, but since everything is a personal card, it'd probably be worth including them all as listed big buffs.

See: https://github.com/quisquous/cactbot/blob/master/ui/jobs/jobs.js#L434

PR #608, opened 2019-09-06
Added remaining cards

JLGarber:
I had a few minutes tonight in queue, so I went out and finished this bit up per #567. Colors may need adjustment, I matched them closely but not programmatically.


quisquous:
Thank you!

PR #619, opened 2019-09-13
Oopsy--Added remaining shb dungeons

JLGarber:
This should finish out #569. Qitana has entirely too many elements to everything, particularly Confession of Faith, but I think these are all correct.

Viper Poison 1 requires a full trigger, as the attack does damage to its primary target regardless of whether things are handled correctly. The puddle dropped by the attack is the same as the other puddles though.

### DIFF
--- a/resources/lang/de.js
+++ b/resources/lang/de.js
@@ -48,6 +48,9 @@ class CactbotLanguageDe extends CactbotLanguage {
       Sludge: 'Schlamm',
       Doom: 'Verhängnis',
       StoneCurse: 'Steinfluch',
+      Imp: 'Flusskobold',
+      Toad: 'Frosch',
+      FoolsTumble: 'Trügerischer Sturz', // 0x387
     });
 
     this.kUIStrings = Object.freeze({
--- a/resources/lang/de.js
+++ b/resources/lang/de.js
@@ -50,7 +50,7 @@ class CactbotLanguageDe extends CactbotLanguage {
       StoneCurse: 'Steinfluch',
       Imp: 'Flusskobold',
       Toad: 'Frosch',
-      FoolsTumble: 'Trügerischer Sturz', // 0x387
+      FoolsTumble: 'Trügerischer Sturz', // 0x183
     });
 
     this.kUIStrings = Object.freeze({
--- a/ui/oopsyraidsy/data/05-shb/dungeon/dohn_mheg.js
+++ b/ui/oopsyraidsy/data/05-shb/dungeon/dohn_mheg.js
@@ -22,23 +22,23 @@
   triggers: [
     {
       id: 'Dohn Mheg Imp Choir',
-      regex: / 1A:........:(\y{Name}) gains the effect of Imp/,
-      mistake: function(e, data, matches) {
-        return { type: 'warn', blame: matches[1], text: 'Shouldn\'t have looked!' };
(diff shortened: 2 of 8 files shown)


ui/oopsyraidsy/data/05-shb/dungeon/dohn_mheg.js
```diff
@@ -0,0 +1,45 @@
...
+    'Dohn Mheg Pinecone Bomb': '3DB1', // Circular ground AoE marker, Dohnfaust Rowans throughout instance
+    'Dohn Mheg Bile Bombardment': '34EE', // Ground AoE marker, boss 3
+    'Dohn Mheg Corrosive Bile': '34EC', // Frontal cone, boss 3
+    'Dohn Mheg Flailing Tentacles': '3681',
+
+  },
+  triggers: [
+    {
+      id: 'Dohn Mheg Imp Choir',
+      regex: / 1A:........:(\y{Name}) gains the effect of Imp/,
```
quisquous: Can you make these all gainsEffectRegex-style triggers like e2s? You'll need to also add translations for the effect name.  (I can dig it up for you if you need it.)  I think in general, it's also fine to use the effectName where possible for oopsy names, so that no extra translation is needed.

JLGarber: To clarify, you would prefer that I add the effects to the respective `lang/$language.js` files? I can do that, but I don't have translations for JP/KO. (DE/FR fortunately have both effects appearing elsewhere in the repository, so I can cheat using that, although Fool's Tumble will still need translation.) Should I simply add an empty string value for each effect property where I don't know the translation, or take a chance on machine translation, or just leave it blank and open a new issue for others who know those languages?

quisquous: Yeah, that's exactly it.  If you don't have a translation, just put a FIXME there.

Fool's Tumble is 387 (https://xivapi.com/Status/387?pretty=true)
Imp is 613, 1103, 1134, 1356
Toad is 439, 441, 1101

Hopefully all of those Imp and Toad translations are the same? I'm not sure why the game needs so many.  I have a local version of SaintCoinach and just look in Status.csv to find these, and then look up the id in xivapi.

JLGarber: Ahh, you're right, I totally skipped over checking XIVAPI. That makes it easy. I'll get those worked out tonight.

resources/lang/de.js
```diff
@@ -48,6 +48,9 @@ class CactbotLanguageDe extends CactbotLanguage {
       Sludge: 'Schlamm',
       Doom: 'Verhängnis',
       StoneCurse: 'Steinfluch',
+      Imp: 'Flusskobold',
+      Toad: 'Frosch',
+      FoolsTumble: 'Trügerischer Sturz', // 0x387
```
quisquous: Sorry, but if you're going to leave a comment on the id, it's not 0x387 (which is hex).  It's either 387 or 0x183.  (or no need for comment at all).

JLGarber: Oof. Counting is hard, you're right.

quisquous:
Thanks for all the oopsy changes!

Sorry that oopsy translations are so different than raidboss translations.  I think there's some inconsistency and indecision about how best to handle everything.

PR #621, opened 2019-09-17
test: Add Trigger Tests to Test Suite

panicstevenson:
Adding individual trigger tests to the testing suite. I know it's bad form to have obnoxious `TODO`'s in a new commit, but there are a lot of changes that would be necessary to capture the output of the tests; it might be best to deliver these in a separate update.

### DIFF
--- a/test/trigger/test_trigger.js
+++ b/test/trigger/test_trigger.js
@@ -1,0 +1,89 @@
+'use strict';
+
+// This test loads an individual trigger file and makes validates
+// the format and regex calls made.
+
+// TODO: Remove ` ?` before each hex value once global prefix `^.{14} ` is added.
+// JavaScript doesn't allow for possessive operators in regular expressions.
+
+let fs = require('fs');
+
+let exitCode = 0;
+
+let inputFilename = String(process.argv.slice(2));
+
+let testValidTriggerRegexLanguage = function(file, contents) {
+  let unsupportedRegexLanguage = /(?:regex|triggerRegex)(?!:|Cn|De|Fr|Ko|Ja).*?:/g;
+  let results = contents.match(unsupportedRegexLanguage);
+  if (results && results.length > 0) {
+    for (const result of results) {
+      console.error(`${file}: invalid regex language '${result}'`);
+      exitCode = 1;
+    }
+  }
+};
+
+let createTriggerRegexString = function(string) {
+  return new RegExp(`(?:regex|triggerRegex)(?:|\w{2}): \/${string}\/`, 'g');
+};
+
+let testWellFormedNewCombatantTriggerRegex = function(file, contents) {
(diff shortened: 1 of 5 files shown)


test/trigger/test_trigger.js
```diff
@@ -0,0 +1,89 @@
...
+// JavaScript doesn't allow for possessive operators in regular expressions.
+
+let fs = require('fs');
+
+let exitCode = 0;
+
+let inputFilename = String(process.argv.slice(2));
+
+let testValidTriggerRegexLanguage = function(file, contents) {
+  let unsupportedRegexLanguage = /(?:regex|triggerRegex)(?!:|Cn|De|Fr|Ko|Ja).*?:/g;
```
quisquous: This is neither here nor there, but I wonder a little bit about the inconsistency between infoText (and friends) returning a dict indexed by lang, and regex having different keys entirely.  <_<

quisquous:
I agree that it can be obnoxious to land things with a bunch of TODOs, but you have to start somewhere! I appreciate landing the tests and then improving things separately, especially if there's a lot of changes.

quisquous:
Thanks also for all the testing improvements! This is super appreciated.  :+1:

test/trigger/test_trigger.js
panicstevenson: I think the infoText approach is possibly better? If nothing else, it might clean up the JavaScript logic in `popup-text.js`; I tried adding a mock version of `triggerRegex` to the logic [found here](https://github.com/quisquous/cactbot/blob/master/ui/raidboss/popup-text.js#L119), and found that it got a little obnoxious looking for 4 potential variables (triggerRegexFr, triggerRegex, regexFr, regex in one example), with an inherent pecking order. Although, I can't say I've tried the opposite; and, I don't know how much better it would be dealing with 3 variables (triggerRegex[fr], regexFr, regex) over 4. But, I think if you were talking end-state, it might be better to go with the dict, how do you feel about it?

quisquous: Yeah, I think it would.  I think consistency would be nice here too to not have two ways to specify translations.  There is a pecking order inherent in the keys as well, so I don't think that would improve.

(I was commenting about consistency in oopsy vs raidboss translations in #619, which is another point of inconsistency.  I think I was trying something new there and I'm not sure that it works out in the end.  Translated effect names make a lot more sense in ui/jobs but maybe less sense in a triggers-esque file.)

I haven't really deprecated much in cactbot, but at some point I think it would make sense to switch from regexFr to regex['fr'].  Probably very few people have user triggers that are translated? (And if we're deprecating things maybe somebody we can change everything on `data` to have proper javascript camelCasing <_<)

This sort of thing doesn't really rise to the level of putting in a github issue (or maybe it does?), but I'm not quite sure where to put these sorts of thoughts for the long term, so it's clear like "I'm not happy about this, but it'd be a bunch of work to change".

PR #622, opened 2019-09-17
Tiny Holminster fix

JLGarber:
Not really a big deal, but it was an easy fix, so I figured I might as well toss it in while I was thinking on it.


quisquous:
Thanks! I was literally about to go do this, because I just ran this today!

PR #626, opened 2019-09-20
Fixed UCoB regex filter for JP clients.

LeonBlade:
Took me longer to realize this was the issue I was chasing, but here it is nonetheless.  The regex match for the JP client was missing one of the floats and wouldn't match properly which threw off calculating dragon positions.


quisquous:
Whoops! Sorry about this.  Thanks for the fix!

PR #630, opened 2019-09-23
Fractal Normal Raidboss/Oopsy

JLGarber:
We finished raid early tonight, so I put this together with the extra time. I have the data needed for timelines, but I'm still working through that.

I thought about adding uncleansed bombs on boss 3, but I wasn't sure whether that sort of mistake is something you wanted. (It would probably also need to track whether Paean is on cooldown or not, and that's a little more advanced than I wanted to go for something like this.)

I'm not 100% happy with the mines trigger, but I have only the one log and I didn't record the kill, so I'm not entirely sure what the actual cast is. (The adds can spawn them as well if they're left alone.) There's about 60s between primary spawns anyway, so it will function, but a 30s suppression feels messy. I'll clean that up when I do further testing runs.

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
+++ b/ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
@@ -22,7 +22,7 @@
     },
     {
       id: 'Fractal Slash/Swipe',
-      regex: / (14:F81:Minotaur starts using 11-Tonze Swipe|14:F83:Minotaur starts using 10-Tonze Slash)/,
+      regex: / 14:(F81:Minotaur starts using 11-Tonze Swipe|F83:Minotaur starts using 10-Tonze Slash)/,
       infoText: {
         en: 'Out of front',
       },
--- a/ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
+++ b/ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
@@ -21,8 +21,15 @@
       },
     },
     {
-      id: 'Fractal Slash/Swipe',
-      regex: / 14:(F81:Minotaur starts using 11-Tonze Swipe|F83:Minotaur starts using 10-Tonze Slash)/,
+      id: 'Fractal Slash',
+      regex: / 14:F83:Minotaur starts using 10-Tonze Slash/,
+      infoText: {
+        en: 'Out of front',
+      },
+    },
+    {
+      id: 'Fractal Swipe',
+      regex: / 14:F81:Minotaur starts using 11-Tonze Swipe/,
       infoText: {
         en: 'Out of front',
       },
@@ -43,7 +50,7 @@
     },
     {
       id: 'Fractal Aetherochemical Bomb',
-      regex: / 1A:........:(\y{Name}) gains the effect of Aetherochemical Bomb/,
+      regex: / 1A:(\y{ObjectId}):(\y{Name}) gains the effect of Aetherochemical Bomb/,
       condition: function(data) {
         return data.CanCleanse;
       },
@@ -53,7 +60,7 @@
     },
     {
       id: 'Fractal Alarums',
-      regex: / 03:........:Added new combatant Clockwork Alarum/,
+      regex: / 03:(\y{ObjectId}):Added new combatant Clockwork Alarum/,
       suppresSeconds: 5,
       infoText: {
         en: 'Kill adds',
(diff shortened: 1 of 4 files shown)


JLGarber:
Sorry about that. It looks like errors in the manifest cascade through a number of other files in the build failure report, and at this time of morning I didn't read the wall of text as well as I should have.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
```diff
@@ -0,0 +1,71 @@
...
+        if (data.role == 'healer') {
+          return {
+            en: 'Buster on ' + data.shortName(matches[1]),
+          };
+        }
+      },
+    },
+    {
+      id: 'Fractal Slash/Swipe',
+      regex: / 14:(F81:Minotaur starts using 11-Tonze Swipe|F83:Minotaur starts using 10-Tonze Slash)/,
```
quisquous: Could you make these separate regexes?

ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
```diff
@@ -0,0 +1,71 @@
...
+    {
+      id: 'Fractal Big Swing',
+      regex: / 14:F87:Minotaur starts using 1111-Tonze Swing/,
+      alertText: {
+        en: 'Use a cage',
+      },
+    },
+    {
+      id: 'Fractal Aetherochemical Bomb',
+      regex: / 1A:........:(\y{Name}) gains the effect of Aetherochemical Bomb/,
```
quisquous: Can you make this ` 1A:\y{ObjectId}: etc etc` instead of just `.`?

ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
```diff
@@ -0,0 +1,71 @@
...
+      condition: function(data) {
+        return data.CanCleanse;
+      },
+      infoText: {
+        en: 'Cleanse bomb',
+      },
+    },
+    {
+      id: 'Fractal Alarums',
+      regex: / 03:........:Added new combatant Clockwork Alarum/,
```
quisquous: Same thing here

ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
```diff
@@ -0,0 +1,71 @@
...
+      suppresSeconds: 5,
+      infoText: {
+        en: 'Kill adds',
+      },
+    },
+    {
+      id: 'Fractal Mines',
+      regex: / Added new combatant Aetherochemical Mine/,
```
quisquous: Can you add the `03:` here?

JLGarber:
Appreciate the feedback. Once I get the timelines in place I'll go back and add the healer AoE triggers on bosses 1 and 3, since those have no cast time.

(I'm not sure why the  check is failing on coinach.py, as I didn't touch the utilities. Just a hiccup, or did I poke something somewhere else that's cascading?)

panicstevenson:
Very doubtful it was your fault to begin with, since you're not even touching the Python files; but, I can assure you that it's not your fault. Nor was it a hiccup.

Looks like a new version of [pylint](https://pypi.org/project/pylint/#history) was recently pushed; it looks to be a minor version bump, so they must have added a new error that we have. Ideally, we'd fix this issue, but also ideally we'd be pinning our versions in Travis CI. The latter feels more appropriate a response, so I've opened https://github.com/quisquous/cactbot/pull/632.

Extremely interesting to note that, while 2.4.0 was the package recently released, 2.3.1, released in March and apparently the version we've been pulling down, does not work either; at least, not for me locally or for Travis.

JLGarber:
Ah, that makes more sense. I'm generally unfamiliar with build tooling except as a concept, and I didn't have time over lunch today to do actual reading, so defaulting to "I might have done it" felt reasonable. I appreciate the explanation.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.js
```diff
@@ -0,0 +1,78 @@
...
+      regex: / 14:F87:Minotaur starts using 1111-Tonze Swing/,
+      alertText: {
+        en: 'Use a cage',
+      },
+    },
+    {
+      id: 'Fractal Aetherochemical Bomb',
+      regex: / 1A:(\y{ObjectId}):(\y{Name}) gains the effect of Aetherochemical Bomb/,
+      condition: function(data) {
+        return data.CanCleanse;
```
quisquous: Sorry, one more thing: this is a function, and not just a property.

JLGarber: Oops, so it is.

PR #631, opened 2019-09-24
test: Fix Trigger Test Regexes

panicstevenson:
Fix incorrect regex string builder string and update tests to be more aggressive in their search criteria.

Update any trigger files found to no longer be in compliance.

### DIFF
--- a/ui/raidboss/data/04-sb/raid/o4s.js
+++ b/ui/raidboss/data/04-sb/raid/o4s.js
@@ -779,8 +779,8 @@
     },
     { // Final phase Addle warning when Reprisal is ending.
       id: 'O4S2 Reprisal',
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/04-sb/raid/o4s.js
```diff
@@ -779,8 +779,8 @@
     },
     { // Final phase Addle warning when Reprisal is ending.
       id: 'O4S2 Reprisal',
-      regex: /loses the effect of Reprisal from/,
-      regex: /loses the effect of Reflexion from/,
+      regex: / 1E:\y{ObjectId}:(\y{Name}) loses the effect of Reprisal from/,
```
quisquous: Oh these are some ancient triggers, that's for sure.

Issue #591, opened 2019-09-01
Option to increase texts time display


ddc:
Hi, is it possible to add an option to increase the time display on "Alarm Text", "Alert Text" and "Info text" in raidboss.js, they disapear real quick.
 I was thinking about adding to my "user/raidboss.js" file like:

Options.DisplayAlarmTextForSeconds = 5;
Options.DisplayAlertTextForSeconds = 5;
Options.DisplayInfoTextForSeconds = 5;

quisquous:
This is a great idea, thanks! Would you want different texts to have different lengths, or would a single option work just as well?

ddc:
Well lengths for each one, like my example above, would be better, so we can have more flexibility on them.

quisquous:
That's true, but I was just curious if you personally would set these to different values.

ddc:
+ Well, I know ACT has its own triggers system and cactbot have "jobs.js", but I also use catcbot to alert me about lots of things, because I can see the text on my screen, and I like to keep my screen "clean" :D
+ Maybe another option as well, would be a custom text to be displayed on screen instead of those three, maybe.
+ Here are some examples, maybe I should replace those "infoText" on my examples bellow with "userInfoText" and increase the timer on them so they can stay longer on screen and dont overlap with others texts from cactbot:

```js
Options.SpokenAlertsEnabled = true;
var swiftCastReady = 'Swiftcast is ready';
var exitOpen = 'Exit is Open';

Options.Triggers = [
	{
		zoneRegex: /.*/,
		triggers:
		[
			{
				regex: /^(You use Swiftcast.)$/,
				delaySeconds = 59;
				infoText: 'swiftCastReady',
			},
		],
	},

	{
		zoneRegex: /.+(Palace Of The Dead)$/,
		triggers:
		[
			{
				regex: /The Cairn of Passage is activated!/,
				infoText: exitOpen,
			},
		],
	},
];
```

quisquous:
Ah, in these cases you can use `durationSeconds` in these triggers to extend the length of the time it is on screen.  Does that help?

ddc:
Yes, `durationSeconds` will work just fine with those examples. I didnt know about that variable.

quisquous:
Oh, great! If you haven't seen it, this doc talks about all the trigger fields: https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/README.md

PR #634, opened 2019-09-25
Alarm/Alert/Info duration

Snsei987:
#591

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -293,10 +293,7 @@ class PopupText {
     let userDisabled = trigger.id && this.options.DisabledTriggers[trigger.id];
     let delay = 'delaySeconds' in trigger ? ValueOrFunction(trigger.delaySeconds) : 0;
     let duration = {
-      fromTrigger:
-        'durationSeconds' in trigger
-          ? ValueOrFunction(trigger.durationSeconds)
-          : undefined,
+      fromTrigger: ValueOrFunction(trigger.durationSeconds),
       alarmText: this.options.DisplayAlarmTextForSeconds,
       alertText: this.options.DisplayAlertTextForSeconds,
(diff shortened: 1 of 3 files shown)


ui/raidboss/popup-text.js
```diff
@@ -292,7 +292,15 @@ class PopupText {
 
     let userDisabled = trigger.id && this.options.DisabledTriggers[trigger.id];
     let delay = 'delaySeconds' in trigger ? ValueOrFunction(trigger.delaySeconds) : 0;
-    let duration = 'durationSeconds' in trigger ? ValueOrFunction(trigger.durationSeconds) : 3;
+    let duration = {
+      fromTrigger:
+        'durationSeconds' in trigger
+          ? ValueOrFunction(trigger.durationSeconds)
+          : undefined,
```
quisquous: Should undefined here be 3 as the default?

ui/raidboss/popup-text.js
```diff
@@ -355,7 +363,10 @@ class PopupText {
           let holder = that.alarmText.getElementsByClassName('holder')[0];
           let div = makeTextElement(text, 'alarm-text');
           addText.bind(that)(holder, div);
-          window.setTimeout(removeText.bind(that, holder, div), duration * 1000);
+          window.setTimeout(
+              removeText.bind(that, holder, div),
+              (duration.fromTrigger || duration.alarmText) * 1000
```
quisquous: Can this be max of duration.alarmText and duration.fromTrigger? If somebody specifies a really long alarmText option, a specified duration on the trigger should not make it shorter, I think?

Snsei987: Well my idea was as no durationSecond is specified by default, the user had to add it manually therefore he wants it to be the time for this specific trigger.

quisquous:
Thanks for this!!

PR #636, opened 2019-09-29
Fractal timeline additions

JLGarber:
I had thought Fractal Continuum would be an easy introduction to timeline creation, but it very definitely was not. Phantom Ray was the only straightforward part.

Minotaur's timeline is messy due to several interlocking possibilities. The basic Swipe > Swing > Groan > Slash > Swipe rotation is simple enough. The intended flow after this is that he starts a long 1111-Tonze Swing cast, which is then interrupted by opening a cage. This starts a long Feast cast, which leads back into the primary rotation. However, if the cage is not used, Minotaur jumps to the basic rotation immediately after using 1111-Tonze Swing. It's fiddly to get this to work, but it's functional in this release.

There is currently an issue with Minotaur in that it's possible to use a cage early. This will send him to where he uses Feast in the rotation, but the timeline is not set up to handle this. It will continue to display the standard rotation until Feast is used, at which point it will jump to the correct point. This is because I set it up to look for the cancellation message on 1111-Tonze Swing. The most correct way to handle it would be to look for the tether marker that indicates he's tethered to the bait. (This is the 1B overhead marker 0036, although it displays visually as a tether.) However, this would involve a regex similar to `/..../..../0036/` or `/1B:........:Minotaur:/`, and I wasn't sure whether long regex strings were okay in timelines.

The Curator is fully functional and correct so far as it goes, but is missing a third phase. Unlike Phantom Ray, Curator does not immediately go to the start of his next phase upon reaching his HP threshold. I wasn't able to figure out completely how this works, but I *think* it's based on when he reaches his threshold prior to casting The Educator. This holds true for the 1 > 2 transition, but unfortunately the phase 3 transition can enter anywhere during the rotation. There are 3 casts of Seed of the Rivers in phase 3, and he can enter at any of them. Worse, none of the abilities following these casts are safe attachment points for a wider sync window. Fortunately, the Seed of the River casts are added to the standard rotation without displacing anything, so the only deficiency in this version of the timeline is that they won't be displayed.

I also removed the trigger for mines. They're visually distinct on the field, and the trigger really didn't seem to add anything.

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
+++ b/ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
@@ -1,0 +1,156 @@
+hideall "--reset--"
+hideall "--sync--"
+hideall "--start--"
+
+0 "--reset--" sync /:Exhibit level III is no longer sealed/ window 10000 jump 0
+0 "--reset--" sync /:The high-level incubation bay is no longer sealed/ window 10000 jump 0
+0 "--reset--" sync /:The reality augmentation bay is no longer sealed/ window 10000 jump 0
+
+
+###Phantom Ray
+#-ii F7D
+
+0 "Start"
+0.0 "--start--" sync /Exhibit level III will be sealed off/ window 0,1
+10.3 "Rapid Sever" sync /:Phantom Ray:F7A:/
+16.6 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+21.4 "Double Sever 1" sync /:Phantom Ray:F7B:/
+25.6 "Double Sever 2" sync /:Phantom Ray:F7C:/
+27.4 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+
+41.5 "Rapid Sever" sync /:Phantom Ray:F7A:/
+47.7 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+52.5 "Double Sever 1" sync /:Phantom Ray:F7B:/
+56.5 "Double Sever 2" sync /:Phantom Ray:F7C:/
+58.3 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/window 8,10 jump 27.4
+
+72.4 "Rapid Sever" sync /:Phantom Ray:F7A:/ jump 41.5
+78.6 "Atmospheric Displacement"
+83.3 "Double Sever 1" sync
+87.4 "Double Sever 2" sync
+89.2 "Atmospheric Displacement"
+#Phase 2 at < 75% HP
+
+97.7 "--sync--" #landing for phase 2 loop
+100.0 "Damage Up" sync /:Phantom Ray:F7F:/  window 100,30 #Overclock
+104.8 "Double Sever 1" sync /:Phantom Ray:F7B:/
+107.6 "Atmospheric Compression" sync /:Phantom Ray:F80:/
+109.0 "Double Sever 2" sync /:Phantom Ray:F7C:/
+111.8 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+118.9 "Rapid Sever" sync /:Phantom Ray:F7A:/
+124.0 "Rapid Sever" sync /:Phantom Ray:F7A:/
+128.2 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+133.1 "Double Sever 1" sync /:Phantom Ray:F7B:/
+137.3 "Double Sever 2" sync /:Phantom Ray:F7C:/
+139.0 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+146.1 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ jump 97.7
+
+148.4 "Damage Up" sync
+153.2 "Double Sever 1"
+155.8 "Atmospheric Compression"
+157.4 "Double Sever 2"
+160.2 "Atmospheric Displacement"
+167.4 "Rapid Sever"
+172.5 "Rapid Sever"
+176.7 "Atmospheric Displacement"
+
+###Minotaur
+1000.0 "--start--" sync /:The high-level incubation bay will be sealed off/ window 1999,5
+1000.0 "Start"
+1003.4 "--sync--" /:Minotaur:366:/ window 3.4,0
+1008.8 "11-Tonze Swipe" sync /:Minotaur:F81:/
+1019.0 "111-Tonze Swing" sync /:Minotaur:F82:/
+1028.2 "Disorienting Groan" sync /:Minotaur:F84:/
+1029.4 "Zoom In x3" sync /:Minotaur:F86:/ duration 3.7
+1034.3 "10-Tonze Slash" sync /:Minotaur:F83:/
+1043.5 "11-Tonze Swipe" sync /:Minotaur:F81:/
+
+
+#Rotation desyncs here due to timing on cage use
+1048.0  "--sync--" sync /:F87:1111-Tonze Swing:Cancelled:/ window 10,15 jump 1151.0
+1064.2 "Feast?" sync /:Minotaur:F88:/ window 60,15 jump 1167.2
+1064.3 "1111-Tonze Swing?" sync /:Minotaur:F87:/ window 1,10 jump 1167.3
+
+1076.4 "11-Tonze Swipe" sync /:Minotaur:F81:/ jump 1179.4
+1086.7 "111-Tonze Swing"
+
+#Second rotation landing spots
+1151.0"--sync--" #landing for correct cage use
+1167.2 "Feast" sync /:Minotaur:F88:/ window 60,15
+1167.3 "--sync--" #landing for missed cage
+
+#Second rotation, same as the first
+1179.4 "11-Tonze Swipe" sync /:Minotaur:F81:/ window 15,15
+1189.7 "111-Tonze Swing" sync /:Minotaur:F82:/
+1199.0 "Disorienting Groan" sync /:Minotaur:F84:/
+1200.1 "Zoom In x3" sync /:Minotaur:F86:/ duration 3.7
+1204.7 "10-Tonze Slash" sync /:Minotaur:F83:/
+1214.0 "11-Tonze Swipe" sync /:Minotaur:F81:/
+1219.0 "--sync--" sync /:F87:1111-Tonze Swing:Cancelled:/ window 10,15 jump 1151.0
+1235.2 "Feast?" sync /:Minotaur:F88:/ window 50,15 jump 1167.2
+1235.3 "1111-Tonze Swing?" sync /:Minotaur:F87:/ window 1,10 jump 1167.3
+
+1247.4 "11-Tonze Swipe" jump 1179.4
+1263.6 "111-Tonze Swing"
+
+###The Curator
+# -ii F8C F8E F8F F90 F91 F92
+
+2000 "--start--" sync /:The reality augmentation bay will be sealed off/ window 2000,0
+2000.0 "Start"
+2003.0 "--sync--" /:The Curator:368:/ window 3,2
+
+2006.2 "Sanctification" sync /:The Curator:F89:/
+2011.4 "Unholy" sync /:The Curator:F8A:/
+2020.6 "Sanctification" sync /:The Curator:F89:/
+2028.8 "Aetherochemical Explosive" sync /:The Curator:F8B:/
+2033.0 "Unholy" sync /:The Curator:F8A:/
+2035.8 "Sanctification" sync /:The Curator:F89:/
+2042.0 "Unholy" sync /:The Curator:F8A:/
+2044.1 "Sanctification" sync /:The Curator:F89:/
+2053.2 "Aetherochemical Explosive" sync /:The Curator:F8B:/
+2066.4 "The Educator" sync /:The Curator:F8D:/
+
+2073.2 "Sanctification" sync /:The Curator:F89:/
+2078.4 "Unholy" sync /:The Curator:F8A:/
+2087.6 "Sanctification" sync /:The Curator:F89:/
+2095.8 "Aetherochemical Explosive" sync /:The Curator:F8B:/
+2100.0 "Unholy" sync /:The Curator:F8A:/
+2102.8 "Sanctification" sync /:The Curator:F89:/
+2109.0 "Unholy" sync /:The Curator:F8A:/
+2111.1 "Sanctification" sync /:The Curator:F89:/
+2120.2 "Aetherochemical Explosive" sync /:The Curator:F8B:/
+2133.4 "The Educator" sync /:The Curator:F8D:/ window 10,10 jump 2066.4
+
+2140.2 "Sanctification" sync /:The Curator:F89:/ jump 2073.2
+2145.4 "Unholy"
+2154.6 "Sanctification"
+2162.8 "Aetherochemical Explosive"
+2167.0 "Unholy"
+2169.8 "Sanctification"
+
+#Phase 2 at first Educator after HP < 69%?
+
+2191.7 "--sync--" #landing for phase 2 loop
+2196.1 "Sanctification" sync /:The Curator:F89:/
+2200.0 "Aetherochemical Mine" sync /:Repository Node:F8F:/ window 2200,10
+2201.3 "Unholy" sync /:The Curator:F8A:/
+2210.7 "Sanctification" sync /:The Curator:F89:/
+2218.8 "Aetherochemical Explosive" sync /:The Curator:F8B:/
+2221.1 "Aetherochemical Mine" sync /:Repository Node:F8F:/
+2222.9 "Unholy" sync /:The Curator:F8A:/
+2225.1 "Sanctification" sync /:The Curator:F89:/
+2231.1 "Unholy" sync /:The Curator:F8A:/
+2233.5 "Sanctification" sync /:The Curator:F89:/
+2238.1 "Aetherochemical Mine" sync /:Repository Node:F8F:/
+2242.6 "Aetherochemical Explosive" sync /:The Curator:F8B:/
+2255.8 "The Educator" sync /:The Curator:F8D:/ window 15,15 jump 2191.7
+
+2263.1 "Sanctification" sync /:The Curator:F89:/ jump 2196.1
+2267.0 "Aetherochemical Mine"
+2268.3 "Unholy"
+2277.7 "Sanctification"
+2285.8 "Aetherochemical Explosive"
+2288.1 "Aetherochemical Mine"
+2289.9 "Unholy"
+2292.1 "Sanctification"
--- a/ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
+++ b/ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
@@ -8,14 +8,14 @@ hideall "--sync--"
 #-ii F7D -p F7F:100
 
 0.0 "--sync--" sync /Exhibit level III will be sealed off/ window 0,1
-10.3 "Rapid Sever" sync /:Phantom Ray:F7A:/
-16.6 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ window 5,5
+10.3 "Rapid Sever" sync /:Phantom Ray:F7A:/ window 15,15
+16.6 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
 21.4 "Double Sever 1" sync /:Phantom Ray:F7B:/
 25.6 "Double Sever 2" sync /:Phantom Ray:F7C:/
 27.4 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
 
-41.5 "Rapid Sever" sync /:Phantom Ray:F7A:/
-47.7 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ window 5,5
+41.5 "Rapid Sever" sync /:Phantom Ray:F7A:/ window 15,15
+47.7 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
 52.5 "Double Sever 1" sync /:Phantom Ray:F7B:/
 56.5 "Double Sever 2" sync /:Phantom Ray:F7C:/
 58.3 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
--- a/ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
+++ b/ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
@@ -22,12 +22,12 @@ hideall "--start--"
 47.7 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
 52.5 "Double Sever 1" sync /:Phantom Ray:F7B:/
 56.5 "Double Sever 2" sync /:Phantom Ray:F7C:/
-58.3 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/window 8,10 jump 27.4
+58.3 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ window 8,10 jump 27.4
 
 72.4 "Rapid Sever" sync /:Phantom Ray:F7A:/ jump 41.5
 78.6 "Atmospheric Displacement"
-83.3 "Double Sever 1" sync
-87.4 "Double Sever 2" sync
+83.3 "Double Sever 1"
+87.4 "Double Sever 2"
 89.2 "Atmospheric Displacement"
 #Phase 2 at < 75% HP
 
@@ -57,7 +57,7 @@ hideall "--start--"
 ###Minotaur
 1000.0 "--start--" sync /:The high-level incubation bay will be sealed off/ window 1999,5
 1000.0 "Start"
-1003.4 "--sync--" /:Minotaur:366:/ window 3.4,0
+1003.4 "--sync--" sync /:Minotaur:366:/ window 3.4,0
 1008.8 "11-Tonze Swipe" sync /:Minotaur:F81:/
 1019.0 "111-Tonze Swing" sync /:Minotaur:F82:/
 1028.2 "Disorienting Groan" sync /:Minotaur:F84:/
@@ -75,7 +75,7 @@ hideall "--start--"
 1086.7 "111-Tonze Swing"
 
 #Second rotation landing spots
-1151.0"--sync--" #landing for correct cage use
+1151.0 "--sync--" #landing for correct cage use
 1167.2 "Feast" sync /:Minotaur:F88:/ window 60,15
 1167.3 "--sync--" #landing for missed cage
 
@@ -90,7 +90,7 @@ hideall "--start--"
 1235.2 "Feast?" sync /:Minotaur:F88:/ window 50,15 jump 1167.2
 1235.3 "1111-Tonze Swing?" sync /:Minotaur:F87:/ window 1,10 jump 1167.3
 
-1247.4 "11-Tonze Swipe" jump 1179.4
+1247.4 "11-Tonze Swipe" sync /:Minotaur:F81:/ jump 1179.4
 1263.6 "111-Tonze Swing"
 
 ###The Curator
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
+hideall "--reset--"
+hideall "--sync--"
+hideall "--start--"
+
+0 "--reset--" sync /:Exhibit level III is no longer sealed/ window 10000 jump 0
```
quisquous: Can you make this just "is no longer sealed"? It makes translations easier.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
...
+25.6 "Double Sever 2" sync /:Phantom Ray:F7C:/
+27.4 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+
+41.5 "Rapid Sever" sync /:Phantom Ray:F7A:/
+47.7 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+52.5 "Double Sever 1" sync /:Phantom Ray:F7B:/
+56.5 "Double Sever 2" sync /:Phantom Ray:F7C:/
+58.3 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ window 8,10 jump 27.4
+
+72.4 "Rapid Sever" sync /:Phantom Ray:F7A:/ jump 41.5
```
quisquous: Does this need a jump too? Or does the 58.3 cover it?

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
...
+78.6 "Atmospheric Displacement"
+83.3 "Double Sever 1"
+87.4 "Double Sever 2"
+89.2 "Atmospheric Displacement"
+#Phase 2 at < 75% HP
+
+97.7 "--sync--" #landing for phase 2 loop
```
quisquous: I think in general, I try to have abilities loop back to themselves rather than to an empty spot in time.  Can you just have the single jump back to 100? Instead of one jump to 97.7 and another jump to 100.0?

I guess I don't quite understand the double jumps everywhere.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
...
+hideall "--sync--"
+hideall "--start--"
+
+0 "--reset--" sync /:Exhibit level III is no longer sealed/ window 10000 jump 0
+0 "--reset--" sync /:The high-level incubation bay is no longer sealed/ window 10000 jump 0
+0 "--reset--" sync /:The reality augmentation bay is no longer sealed/ window 10000 jump 0
+
+
+###Phantom Ray
+#-ii F7D
```
quisquous: -ii F7D -p F7F:100

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
...
+2133.4 "The Educator" sync /:The Curator:F8D:/ window 10,10 jump 2066.4
+
+2140.2 "Sanctification" sync /:The Curator:F89:/ jump 2073.2
+2145.4 "Unholy"
+2154.6 "Sanctification"
+2162.8 "Aetherochemical Explosive"
+2167.0 "Unholy"
+2169.8 "Sanctification"
+
+#Phase 2 at first Educator after HP < 69%?
```
quisquous: This has two phases?!!

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
...
+155.8 "Atmospheric Compression"
+157.4 "Double Sever 2"
+160.2 "Atmospheric Displacement"
+167.4 "Rapid Sever"
+172.5 "Rapid Sever"
+176.7 "Atmospheric Displacement"
+
+###Minotaur
+1000.0 "--start--" sync /:The high-level incubation bay will be sealed off/ window 1999,5
+1000.0 "Start"
```
quisquous: This is a tiny nitpick, but no need for "Start" anywhere, and also maybe prefer "--sync--" over "--start--" just to match other timelines? I doesn't really matter, but it's mostly just a consistency thing.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -0,0 +1,156 @@
...
+111.8 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+118.9 "Rapid Sever" sync /:Phantom Ray:F7A:/
+124.0 "Rapid Sever" sync /:Phantom Ray:F7A:/
+128.2 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+133.1 "Double Sever 1" sync /:Phantom Ray:F7B:/
+137.3 "Double Sever 2" sync /:Phantom Ray:F7C:/
+139.0 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+146.1 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ jump 97.7
+
+148.4 "Damage Up" sync /:Phantom Ray:F7F:/ jump 100.0
```
quisquous: I wish this was automated (or maybe timelines should just have a loop keyword??), but can you make sure the times in your loop lookahead match the times before? If you run `python util/timeline_adjust.py --file=ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt --adjust=48.4`, then you'll get this output:

```
148.4 "Damage Up" sync /:Phantom Ray:F7F:/  window 100,30 #Overclock
153.2 "Double Sever 1" sync /:Phantom Ray:F7B:/
156.0 "Atmospheric Compression" sync /:Phantom Ray:F80:/
157.4 "Double Sever 2" sync /:Phantom Ray:F7C:/
160.2 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
167.3 "Rapid Sever" sync /:Phantom Ray:F7A:/
172.4 "Rapid Sever" sync /:Phantom Ray:F7A:/
176.6 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
```

...which doesn't quite line up with these times.  (48.4 is because you're jumping from 148.4 back to 100.0).  If the times line up perfectly, then when the loop happens, the timeline won't jump perceptively.  Sorry this is a pain.  <_<

This is what I was trying to get at in https://github.com/quisquous/cactbot/blob/master/docs/TimelineGuide.md#building-loops

quisquous:
> The most correct way to handle it would be to look for the tether marker that indicates he's tethered to the bait. (This is the 1B overhead marker 0036, although it displays visually as a tether.) However, this would involve a regex similar to `/..../..../0036/` or `/1B:........:Minotaur:/`, and I wasn't sure whether long regex strings were okay in timelines.

Long regex strings are fine.  They match against the exact same log lines that the trigger files do.  So, if you wanted to make a 1B marker for this, I think it'd make a lot of sense.

> The Curator is fully functional and correct so far as it goes, but is missing a third phase. Unlike Phantom Ray, Curator does not immediately go to the start of his next phase upon reaching his HP threshold. I wasn't able to figure out completely how this works, but I _think_ it's based on when he reaches his threshold prior to casting The Educator. This holds true for the 1 > 2 transition, but unfortunately the phase 3 transition can enter anywhere during the rotation. There are 3 casts of Seed of the Rivers in phase 3, and he can enter at any of them. Worse, none of the abilities following these casts are safe attachment points for a wider sync window. Fortunately, the Seed of the River casts are added to the standard rotation without displacing anything, so the only deficiency in this version of the timeline is that they won't be displayed.

Oh interesting.  It was certainly curious to me that both the first and the second phase are both exactly 67 seconds and that the rotation is the same other than adding new abilities.  Most bosses that add additional abilities sort of desync themselves when they add them in.

So if I'm understanding correctly, the boss has a set rotation in phase 1, and adds in an additional ability in phase 2 (but otherwise has the same rotation?).  And in phase 2, the boss always enters at the same spot in the rotation, but that's not true for phase 3?

Which ones are seeds of the river? Is that stack or spread? Is there a headmarker?

Also, can you leave a comment about them not being in the timeline here so future folks can remember these details?

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
JLGarber: I... think this was because of missed windows... somewhere? Might have been "let's just make SURE this jump happens by adding a backup". I'm not attached to it, so if you prefer to have only one jump per loop, I'm happy to remove it.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
JLGarber: Same as the earlier jump, in terms of why it's doubled.

This is also the start of phase 2, so I think I was trying to make it start on an even 100? That makes the exact phase length easy to see. The sync to an empty line allows that to happen. I did the same thing for Curator but not for Minotaur, so I'm unsure now what I was thinking. Regardless, it should be easy to change to whatever you prefer.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
JLGarber: Can you clarify that? Overclock is included in the timeline, I simply changed the display to "Damage Up" (or the equivalent localization) since there's never a direct cast and you never see an on-screen display as you do with the instant-cast tankbusters/AoEs.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
JLGarber: That's an artifact, I think. It looks like on this one I had run through a long encounter with Phantom Ray and simply adjusted the second phase once for both loops and tossed it in. I suppose I should verify the lookaheads for Minotaur and Curator as well.

JLGarber:
As far as long regexes, I should have said "stylistically okay". I couldn't recall seeing head markers anywhere in timelines, or really anything besides `ENEMYNAME:ABILITYCODE` or emotes/dialogue.

Your understanding is correct on Curator. I was extremely surprised by it myself. It's likely my timings are off, but the fixed rotation is definitely correct. (That, or I didn't test enough and somehow consistently entered at that point in an otherwise variable rotation.)

Seed of the Rivers is a ground-target circle AoE. It's very slow and very clear, so it doesn't need a trigger notification, and honestly not a lot of people are even going to see it in the first place. It's ability F91, but I can add an additional comment to detail why it's excluded.

Thanks for looking over all that!

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
quisquous: Yeah, I think I'd prefer one jump per loop to make the timeline easier to read.  I think if I'm worried about missed windows, I'll just make the window larger on the loop jump.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
quisquous: Sorry, I was just suggesting that if you have the second phase starting at time=100, that you add that in to the args as well.  `-p` is the phase arg, so `-p F7F:100` says take the first instance of skilll `F7F` and adjust it to be time=100.

JLGarber: Oh, I see. Yes, you're right, that does make sense.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
JLGarber: It looks like I handled the Minotaur/Curator loops "correctly" in that sense, so I'll just fix the Phantom Ray lookahead and that should take care of things.

quisquous:
> As far as long regexes, I should have said "stylistically okay". I couldn't recall seeing head markers anywhere in timelines, or really anything besides `ENEMYNAME:ABILITYCODE` or emotes/dialogue.

Yeah, part of the reason is that the `util/test_timeline.py` script reads Network log files and then tests timelines.  It does some of the same formatting transformations that the FFXIV_ACT_Plugin does when you import a Network file so that these lines will still match.  So, any log line type has to be specially handled there if you want that script to be able to use it: https://github.com/quisquous/cactbot/blob/master/util/test_timeline.py#L43

> Seed of the Rivers is a ground-target circle AoE. It's very slow and very clear, so it doesn't need a trigger notification, and honestly not a lot of people are even going to see it in the first place. It's ability F91, but I can add an additional comment to detail why it's excluded.

I see! Yeah, if it's just a ground aoe, I think a comment is perfect and not worth worrying too much about.

> Thanks for looking over all that!

Thanks for the timeline! Much appreciated, especially when it's this complicated.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
JLGarber: Actually, this is a question. It's currently set up as:

```
139.0 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
146.1 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ jump 97.7

148.4 "Damage Up" sync /:Phantom Ray:F7F:/ jump 100.0
153.2 "Double Sever 1"
```

Stylistically, I suppose it's better to jump from the start of the lookahead then? That is, leave the jump at 148.4 and remove the one at 146.1? That would remove the need for a landing point before the loop.

quisquous: I usually jump from the start of the lookahead, just because that's the usually the first skill you can pick where you can jump back to that same skill earlier in the timeline.  If that doesn't have an id or is something you don't want to sync against for various reasons, picking some other skill in the lookahead is fine too.

I think the important part for me is to jump from one skill to itself earlier.

JLGarber: Great. Thank you.

JLGarber:
Ahh, THAT explains why `test_timeline.py` wasn't picking up the cancellation log line and started warning of massive desyncs every time through the cage sequence even though the timeline had visibly functioned. I suppose I could have figured that out with a slightly more careful read through the script. I haven't worked with Python in years, but it doesn't look too hard to extend it to include head markers. I might do that next after completing the current pull requests.

I will have to do a live test on the timeline after all these changes, so it might be a day or two before this is done.

JLGarber:
Sooo, I did a brief test tonight, and I'm going to have to make some changes. It looks like the previous 10ish runs through were all with a "fast" instance, while tonight's was slow. I noticed desyncs during both phases of Phantom Ray, and during the first rotation on Minotaur. (Curator functioned perfectly for some reason.)

I don't have enough time tonight to work that out, but I think it might be necessary to add slightly wider sync windows on the first uses of each ability or something. Have you seen fast vs slow instances cause desyncs before, and how have you usually handled it?

quisquous:
> I don't have enough time tonight to work that out, but I think it might be necessary to add slightly wider sync windows on the first uses of each ability or something. Have you seen fast vs slow instances cause desyncs before, and how have you usually handled it?

Oh noo.  Does test_timeline show this as well for Phantom Ray?

Most fast vs slow instances are on the order of like 1 second a minute or so.  So on an ability by ability basis, there's no that much slippage.  I haven't had to change sync windows in the past.  I've mostly just written this off and not worried about it.

I don't always do this, but I think it's good practice to have the first non-auto ability have a much wider sync window.  I also will sometimes add in the first auto to start the timeline as well (but usually on fights that don't have a zone seal message).

JLGarber:
Unfortunately yes. Sorry about the text dumps, but it's very possible I'm misunderstanding the implications of some of this output:

```
0.000: Matched entry: 0.0 --sync-- (+0.000s)
9.869: Matched entry: 10.3 Rapid Sever (+0.431s)
16.434: Matched entry: 16.6 Atmospheric Displacement (+0.166s)
27.094: Matched entry: 27.4 Atmospheric Displacement (+0.306s)
    Missed sync: Double Sever 1 at 21.4 (last seen at 25.448)
    Missed sync: Double Sever 2 at 25.6 (last seen at 21.339000000000002)
41.481: Matched entry: 41.5 Rapid Sever (+0.019s)
47.641: Matched entry: 47.7 Atmospheric Displacement (+0.059s)
58.204: Matched entry: 58.3 Atmospheric Displacement (+0.096s)
    Missed sync: Double Sever 1 at 52.5 (last seen at 56.461)
    Missed sync: Double Sever 2 at 56.5 (last seen at 52.440000000000005)
    Jumping to 27.400
41.468: Matched entry: 41.5 Rapid Sever (+0.032s)
47.635: Matched entry: 47.7 Atmospheric Displacement (+0.065s)
52.432: Matched entry: 52.5 Double Sever 1 (+0.068s)
56.614: Matched entry: 56.5 Double Sever 2 (-0.114s)
58.141: Matched entry: 58.3 Atmospheric Displacement (+0.159s)
    Jumping to 27.400
41.453: Matched entry: 41.5 Rapid Sever (+0.047s)
47.611: Matched entry: 47.7 Atmospheric Displacement (+0.089s)
58.265: Matched entry: 58.3 Atmospheric Displacement (+0.035s)
    Missed sync: Double Sever 1 at 52.5 (last seen at 56.613)
    Missed sync: Double Sever 2 at 56.5 (last seen at 52.42400000000001)
    Jumping to 27.400
41.551: Matched entry: 41.5 Rapid Sever (-0.051s)
47.619: Matched entry: 47.7 Atmospheric Displacement (+0.081s)
52.606: Matched entry: 52.5 Double Sever 1 (-0.106s)
56.609: Matched entry: 56.5 Double Sever 2 (-0.109s)
58.145: Matched entry: 58.3 Atmospheric Displacement (+0.155s)
    Jumping to 27.400
35.812: Matched entry: 100.0 Damage Up (+64.188s)
```

I'm actually not sure why I saw desyncs during phase 2, as test_timeline shows matches for all of it. Basically, at various points, the timeline just stopped updating until a point with a larger sync window, at which time it correctly snapped to where it should be. It seemed random during phase 2, but phase 1  consistently stalled with the first Atmospheric Displacement never clearing, and every other ability piling up after it.

Minotaur was, well...

```
0.000: Matched entry: 1000.0 --sync-- (+1000.000s)
1002.484: Matched entry: 1003.4 --sync-- (+0.916s)
1066.123: Matched entry: 1064.3 1111-Tonze Swing? (-1.823s)
    Warning: 62.723s since last sync
    Missed sync: 11-Tonze Swipe at 1008.8 (last seen at 1049.96)
    Missed sync: 111-Tonze Swing at 1019.0 (last seen at 1022.005)
    Missed sync: Disorienting Groan at 1028.2 (last seen at 1031.271)
    Missed sync: Zoom In x3 at 1029.4 (last seen at 1035.528)
    Missed sync: 10-Tonze Slash at 1034.3 (last seen at 1040.622)
    Missed sync: 11-Tonze Swipe at 1043.5 (last seen at 1049.96)
    Missed sync: --sync-- at 1048.0
    Missed sync: Feast? at 1064.2
    Jumping to 1167.300
```
It went back to normal after the jump here, but nothing synced prior. I'm not sure why the timeline loses three seconds on Disorienting Groan. Is it possible that I didn't implement the `duration` option correctly?

For Minotaur, at least, I think it's probably fine to just add wide sync windows to 111-Tonze Swing, Disorienting Groan, and 10-Tonze Slash. (Not Swipe, since that occurs twice during the rotation.) This would be doubly beneficial, as it appears he doesn't actually reset his rotation if a cage is used early. (At least in tonight's testing, it looked like he paused the rotation to eat the bait, then resumed where he left off.) I did swap out the the 1111-Tonze Swing cancellation message in favor of the 1B line, and that seemed to work fine, so I will probably leave that instead of the cancellation line.

Finally, in tonight's testing, it looks like Curator's phase division is just "cast Educator once", with no HP threshold. I'm almost wondering whether this encounter has different rotations depending on the instance. I assume you haven't heard of something like this? It seems ridiculous, but nothing about Fractal would surprise me after all this nonsense. Not that it matters for Curator, as the timeline seems to have worked perfectly on every test.

JLGarber:
[I should have added logs when I posted the previous comment](https://www.dropbox.com/s/ip0vwq58tnzidog/fractallogs.zip?dl=0).  Times to look for:

Log 1
```
Phantom Ray: 
-s 23:36:25.000 -e 23:40:09.000, -s 23:42:32.000 -e 23:44:42.000,
-s 23:46:14.000 -e 23:49:27.000, -s 23:52:02.000 -e 23:54:12.000,
-s 23:57:28.000 -e 00:00:17.000, -s 00:02:04.000 -e 00:03:47.000,
-s 00:09:05.000 -e 00:13:23.000

Minotaur: -s 00:17:39.000 -e 00:22:16.000

Curator: -s 00:24:40.000 -e 00:28:14.000
```
Log 2
```
Phantom Ray: -s 00:44:04.000 -e 00:45:40.000

Minotaur: -s 00:48:07.000 -e 00:50:35.000

Curator:
-s 00:53:20.000 -e 00:55:46.000 , -s 00:56:41.000 -e 00:59:06.000,
-s 00:59:53.000 -e 01:02:31.000
```

test_timeline is going to report missing syncs on Minotaur regardless, of course, since that encounter relies on currently unsupported log line types.

(Aside, I wonder how feasible it would be to have test_timeline automatically search for encounters in log files? From my brief look through it, I think the information it needs is there in terms of necessary regexes, it's "just" a matter of extending the script to look for sealing/unsealing messages and then passing those timestamps automagically as the log file arguments if a new super-special-awesome -a/automagic flag were set. I probably wouldn't touch that for a while, since I would want to add support for more log lines first, but if it's realistic I could do that later.)

JLGarber:
After further testing, it looks like the major sync issues with Minotaur have been worked out in my local version. I'm still seeing the "ability pile-up" desync on Phantom Ray, but all the syncs are "matching", so I'm genuinely unsure what the issue could be:

```
0.000: Matched entry: 0.0 --sync-- (+0.000s)
9.463: Matched entry: 10.3 Rapid Sever (+0.837s)
16.476: Matched entry: 16.6 Atmospheric Displacement (+0.124s)
21.309: Matched entry: 21.4 Double Sever 1 (+0.091s)
25.576: Matched entry: 25.6 Double Sever 2 (+0.024s)
27.332: Matched entry: 27.4 Atmospheric Displacement (+0.068s)
41.713: Matched entry: 41.5 Rapid Sever (-0.213s)
47.673: Matched entry: 47.7 Atmospheric Displacement (+0.027s)
52.516: Matched entry: 52.5 Double Sever 1 (-0.016s)
56.592: Matched entry: 56.5 Double Sever 2 (-0.092s)
58.146: Matched entry: 58.3 Atmospheric Displacement (+0.154s)
72.325: Matched entry: 72.4 Rapid Sever (+0.075s)
    Jumping to 41.500
41.500: Matched entry: 41.5 Rapid Sever (+0.000s)
47.607: Matched entry: 47.7 Atmospheric Displacement (+0.093s)
52.511: Matched entry: 52.5 Double Sever 1 (-0.011s)
56.591: Matched entry: 56.5 Double Sever 2 (-0.091s)
58.231: Matched entry: 58.3 Atmospheric Displacement (+0.069s)
64.568: Matched entry: 100.0 Damage Up (+35.432s)
    Missed sync: Rapid Sever at 72.4 (last seen at 41.5)
104.905: Matched entry: 104.8 Double Sever 1 (-0.105s)
107.341: Matched entry: 107.6 Atmospheric Compression (+0.259s)
109.230: Matched entry: 109.0 Double Sever 2 (-0.230s)
111.743: Matched entry: 111.8 Atmospheric Displacement (+0.057s)
119.039: Matched entry: 118.9 Rapid Sever (-0.139s)
124.005: Matched entry: 124.0 Rapid Sever (-0.005s)
128.169: Matched entry: 128.2 Atmospheric Displacement (+0.031s)
133.047: Matched entry: 133.1 Double Sever 1 (+0.053s)
137.127: Matched entry: 137.3 Double Sever 2 (+0.173s)
138.952: Matched entry: 139.0 Atmospheric Displacement (+0.048s)
146.134: Matched entry: 146.1 Atmospheric Displacement (-0.034s)
148.287: Matched entry: 148.4 Damage Up (+0.113s)
    Jumping to 100.000
```

I'm a little concerned that the timing is off by as much as a quarter second on some uses, but that shouldn't be causing issues as far as rendering the timeline, correct?

Once we get this resolved, I'm comfortable with where the timeline is as far as release goes. I'd like better handling on Minotaur, but I really don't see a good way to handle the potential special cases where the party uses cages far too early. (And besides, the Feast/Swing syncs will pick things back up after if they actually do that.)

quisquous:
> I'm a little concerned that the timing is off by as much as a quarter second on some uses, but that shouldn't be causing issues as far as rendering the timeline, correct?

I think most people will have a hard time noticing a quarter second difference.  Timelines didn't use to be remotely that accurate and they still felt good enough.  I think that's quite ok.  Did it feel bad to you when you ran this dungeon?

I've been a little busy so haven't looked at the logs directly yet.  Would you like me to still take a look, or do you feel like they're in a good spot now?

> (Aside, I wonder how feasible it would be to have test_timeline automatically search for encounters in log files? From my brief look through it, I think the information it needs is there in terms of necessary regexes...

This would be amazingly useful.  I filed #646 about this.

> Once we get this resolved, I'm comfortable with where the timeline is as far as release goes. I'd like better handling on Minotaur, but I really don't see a good way to handle the potential special cases where the party uses cages far too early. (And besides, the Feast/Swing syncs will pick things back up after if they actually do that.)

I don't think there's anything to be done here about this.  This seems good enough!

> Sorry this whole thing has been so complicated. I genuinely didn't expect Fractal to have this many oddities about it.

No no no.  No need for apologies! Thanks for tackling a complicated timeline.  :)

JLGarber:
Well, here's what we have. I'm happy enough with Minotaur and Curator, as noted. I think rather than having you poke through the logs to start, we might be better served by you testing Phantom Ray when you get a chance and seeing what happens. There are no obvious errors in the timeline that I can see from a text perspective, and obviously it passed the linter, so whatever is causing the issues on my end probably isn't related to an actual problem with the timeline itself.

Other than that, I think this one is ready for release.

ui/raidboss/data/03-hw/dungeon/fractal_continuum.txt
```diff
@@ -1,37 +1,32 @@
...
 ###Phantom Ray
-#-ii F7D
+#-ii F7D -p F7F:100
 
-0 "Start"
-0.0 "--start--" sync /Exhibit level III will be sealed off/ window 0,1
+0.0 "--sync--" sync /Exhibit level III will be sealed off/ window 0,1
 10.3 "Rapid Sever" sync /:Phantom Ray:F7A:/
-16.6 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/
+16.6 "Atmospheric Displacement" sync /:Phantom Ray:F7E:/ window 5,5
```
quisquous: For what it's worth, the default window is 5 seconds, so `window 5,5` is a noop.

If you're worried about syncs, I'd say maybe just put like a 10-20 seconds window on Rapid Sever?

JLGarber: That's very odd, it's definitely not been operating that way. Is that what's defined in `timeline.js` at line 158? Both there and in DnT's guide it looks like the default is 2.5+/-.

quisquous: Sorry sorry, you are 100% correct.

quisquous:
Ok ok, I ran through all your phantom ray examples.  They all look perfect except for one issue:

`-f fractallog1.log -s 23:57:28.000 -e 00:00:17.000`

```
58.108: Matched entry: 58.3 Atmospheric Displacement (+0.192s)
    Missed sync: Double Sever 1 at 52.5 (last seen at 56.457)
    Missed sync: Double Sever 2 at 56.5 (last seen at 52.435)
```

This is what you commented about in https://github.com/quisquous/cactbot/pull/636#issuecomment-537818913.

I think that just means that sever 1 and sever2 abilities can come in either order.  So they should probably all be `F7[BC]`?

Other than that, it seems perfect.  There are some missed syncs listed, but those are all what happens when you jump to Damage Up and you miss all the syncs in between wherever you were and the jump target.  It might be nice to somehow mark those as being intended skips vs unintended ones for the test tool.

JLGarber:
THANK you. That's what I had missed. And of course now that you mention it I do remember noting early on that the order was random. That should be it.

Issue #385, opened 2019-03-14
Is it possible to change colors of text within the Timeline plugin?


Merkava2k15:
Is it possible to change the color of individual lines within the Cactbot Timeline raidboss plugin?

For example in the preview picture you posted:
https://raw.githubusercontent.com/quisquous/cactbot/master/screenshots/Raidboss.png

Would it be possible to change say... "Dualcast" to red instead of it being white?

Thanks for any support offered!

quisquous:
At the moment, there's no way to do this, sorry.  This is a good thought though, and I should be able to make it possible to add custom styling to entries (color, font size, etc).  I'll keep this open as a feature and hopefully I (or somebody else?) can get to it sometime in the next while.

Merkava2k15:
Thanks for the quick reply.

Still, Thanks for everything and keep up the great work!

Snsei987:
Can't you just do it within the `raidboss.css` file ?

quisquous:
This got fixed in #637.

PR #637, opened 2019-09-29
raidboss: change color/font of text

Snsei987:
#385

Could probably be better, tell me what do you think

Example :
```
  timelineStyles: [
    {
      regex: /Angry Dummy/,
      style: {
        color: 'red',
        'font-family': 'Impact', // If the CSS property has '-' in it, you must use quotation marks
      },
    },
  ],
```

Add this is any `.js` file from the `data` folder

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -640,10 +640,8 @@ class TimelineUI {
     bar.toward = 'right';
     bar.style = !channeling ? 'fill' : 'empty';
 
-    if (e.style) {
-      if (e.style.color) bar.color = e.style.color;
-      if (e.style.font) bar.font = e.style.font;
-    }
+    if (e.style)
+      bar.applyStyles(e.style);
 
     if (!channeling && e.time - fightNow > this.options.BarExpiresSoonSeconds) {
       bar.fg = this.barColor;
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -226,12 +226,12 @@ class Timeline {
       }
 
       if (styles) {
-      	for (const style of styles) {
-      		const m = e.name.match(style.regex);
-      		if (!m)
-      			continue;
-      		Object.assign(e, { style: style.style });
-      	}
+        for (const style of styles) {
+          const m = e.name.match(style.regex);
+          if (!m)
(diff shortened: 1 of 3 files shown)


ui/raidboss/timeline.js
```diff
@@ -119,6 +120,23 @@ class Timeline {
         continue;
       }
 
+      match = line.match(/^color\s+\"([^"]+)\"\s+(#[a-fA-F0-9]{6})$/);
```
quisquous: Hmm.  I think I would prefer if these were more like timeline triggers than new timeline line commands.

Maybe something like:
```
timelineStyles: [
  {
    regex: /something/,
    style: {
      color: 'papaya-whip',
      font-size: 'gigantic',
    },
  },
],
```

This would make it easier to have an array of styles, a more complicated regex to apply to lines, and to arbitrarily allow for any kind of styling rather than just a few whitelisted ones.

What do you think?

Snsei987: Seems better yes, i've no clue on how to access it tho. And help ?

Snsei987: It should work now

ui/raidboss/timeline.js
```diff
@@ -631,6 +640,11 @@ class TimelineUI {
     bar.toward = 'right';
     bar.style = !channeling ? 'fill' : 'empty';
 
+    if (e.style) {
```
Snsei987: This is where you should handle new properties in the future, with ofc the correct setup in `resources/timerbar.js`

quisquous: I was wondering if it was possible to just set an array of styles here and pass that through all the way to timerbar.js and apply all of them there.  That way anybody could override whatever css style they wanted.  Do you think that would work?

Snsei987: Should work like that now

quisquous:
Sorry about all the lint.  `npm run lintfix` should do most of it for you, if you're not already using that.

quisquous:
Thanks! I think this is super useful.  I wonder a little bit if being able to style left/center/right text matters here (for timelines right is always the time and left is the text) or if being able to style the box itself matters too? But maybe those are thoughts for the future.  :)

Issue #573, opened 2019-08-25
docs: update log guide for 5.x


quisquous:
There were a number of log changes in the 5.x version of the ffxiv plugin, in terms of object ids being included in the formatted log output for gains/loses effect and added combatant.

There's also a few new log lines that need to be added like 249 (settings) and 250 (current process information).

JLGarber:
Does Ravahn have this documented anywhere, or is this a matter of hunting through for things? I'm happy to help with this as I can, but I don't know enough about the log format to "just know what needs fixing". I suppose I can start with searching in the XIV Plugin Discord server...

quisquous:
No, this isn't documented anywhere.  You just have to kind of run stuff and figure it out based on reading log lines.  That's how the log guide got written in the first place.  <_<

Additional new log lines now include:
* 0xF9 (settings)
* 0xFA (process information)
* 0x0C (player stats)
* 0x24 (limit break)
* 0x25, 0x26, 0x27: new things added in 2.0.4.0 related to hp/mp/status??

Additionally: 
* 0x15 and 0x16 have heading and "sequence number" at the end of them.
* 0x14 has the caster id in it
* 0x1A and 0x1E have the caster object id at the front

JLGarber:
Oh, so lots of hunting for stuff I don't know. Fair enough, I can do that. No idea on how long it will take me to get any of that sorted, but are you okay with pull requests that update only one log line? This way if anyone also wants to pick up at some point, they can be reasonably sure they aren't duplicating effort for more than one line type.

quisquous:
Pull requests with one line are just fine.  No need to get everything done before updating documentation.

ngld:
0x25 is `NetworkEffectResult`, 0x26 is `NetworkStatusList`, 0x27 is `NetworkUpdateHp`

Unless I counted wrong somewhere, 0x1A is `NetworkBuff`, 0x1E is `NetworkBuffRemove` and they have the following format:
`BuffID|buffName|Duration|ActorID|ActorName|TargetID|TargetName|BuffExtra|TargetMaxHP|ActorMaxHP`
The parsed version is: `[timestamp] <type>:<TargetID>: <target> (loses|gains) the effect of <skillname> from <actor> ( for <duration> seconds)?\.( <stacks>)?`

Formats for `NetworkEffectResult`:
[...]

PR #639, opened 2019-09-30
Added lines 24-27 to log guide

JLGarber:
Per #573 I've *finally* made the time to sit down and add these lines to the guide. I've marked this as a draft because it's not as clean as I would like. The information in here is correct as far I can tell, but it seems minimal.

### DIFF
--- a/docs/LogGuide.md
+++ b/docs/LogGuide.md
@@ -861,6 +861,46 @@ There are also a number of examples where tethers are generated in some other wa
 * Suzaku Extreme birbs: who knows
 * player to player tethers (dragonsight, cover, fairy tether)
 
+## 24:LimitBreak
+
+This log line is recorded every server tick where limit break energy is generated while in combat in a light or full party. (Generation is not recorded while at cap.) It starts at 0x0000 at the beginning of the instance (or encounter in the case of a single-encounter instance,) and counts up by 0x00DC (220 decimal) until the limit break is used, or to a max of 0x7530 (30,000 decimal). This rate of increase is constant, but other actions taken can cause extra increments to happen independent of the base increase. (These other increments occur in the same packet as the base rate, but separately.) Each limit break bar is 0x2710 (10,000 decimal) units.
+
+## 25:NetworkEffectResult
+
+This log line appears to be recorded at any time an actor is targeted by a hostile action. It appears that the action must hit and must deal non-zero damage OR must inflict its effect in order to generate this log line. Individual DoT ticks do not appear to generate separate lines.
+
+This line's structure (parsed) is
+`25:Player Object ID:Sequence Number:Current HP:Max HP:Current MP:Max MP:Current TP:Max TP:Position X:Position Y:Position Z:Facing:[packet data thereafter]`
+
+A sample line would be
+` 25:12345678:PlayerOne:0000132A:33635:35817:10000:10000:0::0.3841706:-207.8767:2.901163:-3.00212:03E8:2500:0:01:03000000:0:0:E0000000:`
+
+## 26:NetworkStatusEffects
+
+For NPC opponents (and possibly PvP) this log line is generated alongside `18:NetworkDoT` lines. For non-fairy allies, it is generated alongside [1A: NetworkBuff](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#1e-networkbuffremove), [1E: NetworkBuffRemove](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#1e-networkbuffremove), and [25:NetworkEffectResult](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#25-networkeffectresult).
+
+This line's structure is:
+
+`26:Target Id:Target Name:Job Levels:Current HP:Max Hp:Current Mp:Max MP:Current TP:Max TP:Position X:Position Y:Position Z:Facing:<status list; each status is a uint as hex>`
+
+A sample line would be:
+
+`26:12345678:PlayerOne:3C503C1C:24136:24136:9045:10000:4:0:-0.4730835:-158.1598:-23.9:3.110625:03E8:45:0:020130:0:106501CA:0129:4172D113:106501CA:012A:4168C8B4:106501CA:012B:40919168:106501CA:0232:40E00000:E0000000:`
+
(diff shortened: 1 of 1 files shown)


docs/LogGuide.md
```diff
@@ -861,6 +861,46 @@ There are also a number of examples where tethers are generated in some other wa
...
+
+This line's structure is:
+
+`26:Target Id:Target Name:Job Levels:Current HP:Max Hp:Current Mp:Max MP:Current TP:Max TP:Position X:Position Y:Position Z:Facing:<status list; each status is a uint as hex>`
+
+A sample line would be:
+
+`26:12345678:PlayerOne:3C503C1C:24136:24136:9045:10000:4:0:-0.4730835:-158.1598:-23.9:3.110625:03E8:45:0:020130:0:106501CA:0129:4172D113:106501CA:012A:4168C8B4:106501CA:012B:40919168:106501CA:0232:40E00000:E0000000:`
+
+It seems likely that Square implemented this line in order to extend functionality for the `18`, `1A`, and `1E` log lines without breaking previous content.
```
quisquous: All the log lines in this guide (other than the network packets) are all output from ravahn's FFXIV_ACT_Plugin.  So, I think this is ravahn adding additional information for fflogs without breaking the triggers that people are using with other formatted log lines.

docs/LogGuide.md
```diff
@@ -861,6 +861,46 @@ There are also a number of examples where tethers are generated in some other wa
...
+
+## 27:NetworkUpdateHP
+
+It's not completely clear what triggers this log line, but it contains basic information comparable to `25` and `26`. It applies to allies and fairies.
```
quisquous: Does this not apply to other pets, just fairies?

docs/LogGuide.md
```diff
@@ -861,6 +861,46 @@ There are also a number of examples where tethers are generated in some other wa
 * Suzaku Extreme birbs: who knows
 * player to player tethers (dragonsight, cover, fairy tether)
 
+## 24:LimitBreak
+
+This log line is recorded every server tick where limit break energy is generated while in combat in a light or full party. (Generation is not recorded while at cap.) It starts at 0x0000 at the beginning of the instance (or encounter in the case of a single-encounter instance,) and counts up by 0x00DC (220 decimal) until the limit break is used, or to a max of 0x7530 (30,000 decimal). This rate of increase is constant, but other actions taken can cause extra increments to happen independent of the base increase. (These other increments occur in the same packet as the base rate, but separately.) Each limit break bar is 0x2710 (10,000 decimal) units.
```
quisquous: In github markdown at least, you can break up each sentence (or fragments) onto their own line.  It makes future edits easier to read.  Would you mind inserting some line breaks here?  The formatted markdown is equivalent, but the raw markdown is better.

You can see above near `The reason why there is such a discrepancy between ACT and fflogs` for an example.

docs/LogGuide.md
```diff
@@ -861,6 +861,46 @@ There are also a number of examples where tethers are generated in some other wa
...
+
+For NPC opponents (and possibly PvP) this log line is generated alongside `18:NetworkDoT` lines. For non-fairy allies, it is generated alongside [1A: NetworkBuff](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#1e-networkbuffremove), [1E: NetworkBuffRemove](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#1e-networkbuffremove), and [25:NetworkEffectResult](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#25-networkeffectresult).
+
+This line's structure is:
+
+`26:Target Id:Target Name:Job Levels:Current HP:Max Hp:Current Mp:Max MP:Current TP:Max TP:Position X:Position Y:Position Z:Facing:<status list; each status is a uint as hex>`
+
+A sample line would be:
+
+`26:12345678:PlayerOne:3C503C1C:24136:24136:9045:10000:4:0:-0.4730835:-158.1598:-23.9:3.110625:03E8:45:0:020130:0:106501CA:0129:4172D113:106501CA:012A:4168C8B4:106501CA:012B:40919168:106501CA:0232:40E00000:E0000000:`
```
quisquous: I'm not sure that each status is a uint as hex is quite precise here.  Like, it kinda looks like a collection of player ids (source?) and some effect ids too.  I'm not really sure I understand.  I can find effect ids, but they're not always in the same spot.   It might be worth just saying this is unknown data.

docs/LogGuide.md
JLGarber: Certainly. What's your preferred maximum, 110 characters? (I realize I can't just break at "word boundary nearest 110", but having that as a rough guideline is helpful.)

It also looks like this is something my previous additions to the log guide didn't respect, would you prefer that fixed here or in a separate update?

docs/LogGuide.md
JLGarber: Honestly, that was the one thing I didn't test in-game. ngld was kind enough to provide the structure for each of those log lines, and I verified the correctness (or at least appearance of correctness) for the other fields. I have no problems changing it to "unknown" for now.

docs/LogGuide.md
JLGarber: You're entirely correct, it does.

docs/LogGuide.md
JLGarber: Fair enough. I never really thought about it, so I merely assumed that the output is there in raw form already and that the plugin just parses it into a more human-readable format.

docs/LogGuide.md
quisquous: It's less about word boundaries and more a desire for semantic line breaks: https://sembr.org/.  If you made me pick a limit, I'd say 80 is probably reasonable for most things?

Also I assume that the max here of 0x7530 is lb3? and that if you only get lb1 or lb2 it stops at lower values?

JLGarber:
Alright, I think this should be better. I'll probably do more work on the other new log lines after I finish up the Fractal stuff and look into what it would take to update `test_timeline.py`. Let me know what else you think might be good to add to this request and I'll see what I can do. Thanks for looking over things!

quisquous:
> Thanks for looking over things!

No no, thanks for the pull request! Much appreciated!

quisquous:
I think this seems to good to merge in, if you want to?

JLGarber:
I think so, yes. I have nothing further to add right now on these log lines. Once my static starts clearing Titan on Tuesdays consistently I might revisit this and flesh these sections out a little more.

PR #641, opened 2019-10-01
Lots of FR improvements

mooondark:


Akurosia:
Just a short question for future addtions to regexes:
Is there no rule e.g. what is capitalized and what not?
As for me it seams everithing in a normal :14: message ( starts using) needs the [dD] element in front.

Depending on your answer  i could change my work flow a little bit when adding regex translations for the future.

mooondark:
As far as I know, there's no specific rule. Sometimes it's in capitals sometimes no ... I don't know why the text is wrong in fflogs.

Thanks anyway :)

JLGarber:
The names in Dohn Mheg were a known issue in English for 14 log lines, so I'm not surprised to see that it exists in French as well. There's no way to know whether the log line will capitalize those words or not when they appear as part of enemy names.

PR #643, opened 2019-10-05
raidboss: Fix Unreachable Code

panicstevenson:
Fix unreachable code block for Flatten target in Turn 13.

I came across this while adding additional tests. Also makes me wonder if BLU should be included into the healer role (or all roles? kind of gross) since this would be applicable to any fight in which BLU can participate.

### DIFF
--- a/ui/raidboss/data/02-arr/raid/t13.js
+++ b/ui/raidboss/data/02-arr/raid/t13.js
@@ -45,9 +45,9 @@
         }
       },
       infoText: function(data, matches) {
-        if (matches[1] == data.me) {
+        if (matches[1] == data.me)
           return;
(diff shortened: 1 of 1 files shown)


quisquous:
Yeahhhh, I don't know what to do about blu.  We could have a "cares about tankbusters" and "cares about aoe" function on `data`, in the same way that there's `CanStun` etc.  It might become more relevant with whatever they're doing with no-echo blu stuff coming out.  (I guess I should get working on those extreme primal timelines <_<).

Having blu be considered a healer at least covers the tankbuster and aoe cases pretty elegantly.  It's a super hack though.

quisquous:
Thanks also for this fix! Sorry my extensive trigger test suite and large qa team did not catch it earlier.

PR #645, opened 2019-10-05
test: Add Additional Tests for Regex Cleanliness

panicstevenson:
Make the regex matching for previous tests stricter. Ensure there are no invalid combinations of '.' operators within the trigger regexes. Inform the user when there are instances of matching a single character via the use of groups instead of picking a character from a set.

Note: The former isn't that smart, and can't tell if you've used something like `(?:ABC1|ABC2)`; only if you're using `ABC(?:1|2)`.


quisquous:
Thanks for all the additional testing.

PR #647, opened 2019-10-05
raidboss: Fix Trigger Syntax Issues

panicstevenson:
General fix for various syntax issues within Raidboss triggers.


JLGarber:
Oof. Thanks for finding those. I'll be more careful about that pattern going forward.

PR #651, opened 2019-10-08
Replace incorrect PLD reference to `blood` with `oath`

thomas-holmes:
PLD section was incorrectly referring to `blood` in if/else flow, change to oath instead.


quisquous:
Whoops, thank you.  /o\

PR #655, opened 2019-10-12
raidboss: Update zoneRegex

panicstevenson:


quisquous:
I love the naming consistency here.  I had to double check my own logs, but that's right.

PR #659, opened 2019-10-12
raidboss: List Regex Keys in Ascending Order

panicstevenson:
Update the regex keys for various triggers to be in ascending order.


quisquous:
Thanks! Is it possible to have a test for this?

panicstevenson:
> Thanks! Is it possible to have a test for this?

Yes, and no. No in the sense of not currently, but yes that we can eventually. We'd have to add the ability to test a trigger in its entirety first, which is a good thing to add anyway.

Alternatively, you can do what I did and turn on `sort-keys` in ESLint to find the culprits, but that lints... everything.

Issue #660, opened 2019-10-14
ngld overlayplugin is crashy


quisquous:
See my comments on #577.

Filing this issue just to have a place to have this conversation and for other people to chime in if they're having the same issue.

nicholasnjh:
(1) disable the hibiyasleep OverlayPlugin.dll in ACT, close ACT
(2) download the latest version of the ngld overlayplugin
(3) unblock and unzip that somewhere on disk
(4) load the OverlayPlugin.dll in that folder as an ACT plugin
(5) download the cactbot-0.11.1-ngld.zip file
(6) read and follow these new instructions carefully: https://github.com/quisquous/cactbot/tree/ngldoverlay#installing-cactbot
(7) you will need to re-add all of your overlays, sorry
(8) please also click the "Automatically Report Errors" checkbox in the general tab


I followed the steps, once i disabled hibiyasleep overlayplugin.dll everything stops working, my acts stops responding altogether. nvm , i force closed it, reopened it and loaded the new overlayplugin and it prompted with

"Could not load type 'rainbowmage.overlayplugin.registry' from assembly 'overlayplugin.common, version= 0.3.4.0. culture= netral, publickeytoken = null

nicholasnjh:
wait.. where is the timeline for it now? it doesnt show it anymore... or I cant browse to add in cactbot overlay helpppppppppppppppppppppppppppppppppppp
![image](https://user-images.githubusercontent.com/40110522/66763362-4df9ab80-eeda-11e9-9af6-ec3885d12763.png)

Akurosia:
Just tried to use ngld's version, but as soon as i add the raidboss overlay, i get following log output

> [14.10.2019 17:57:55] Info: InitPlugin: PluginDirectory = E:\FINAL FANTASY XIV - A Realm Reborn\AdditionalExtern\ACT\Plugins\OverlayPlugin_ngld
[14.10.2019 17:57:56] Info: InitPlugin: Initialized.
[14.10.2019 17:57:56] Error: Got a subscription for missing event "onPlayerChangedEvent"!
[14.10.2019 17:57:56] Error: Got a subscription for missing event "onLogEvent"!
[14.10.2019 17:57:56] Error: Got a subscription for missing event "onGameActiveChangedEvent"!
[14.10.2019 17:57:56] Error: JS Handler call failed: System.Exception: Tried to call missing handler "cactbotReadDataFiles"!
[...]

PR #672, opened 2019-10-16
raidboss: Eliminiate a race condition during raidboss initialization.

ngld:
If we attach the overlay listeners before gPopupText has been set, we risk triggering a "Cannot read property '...' of undefined" error. If we attach them in the constructor instead, we can be sure that the object exists.
Since only one instance of PopupText is ever created, we don't have to worry about detaching the overlay listeners.

This should fix the issue reported in https://github.com/quisquous/cactbot/issues/660#issuecomment-542819459.

I've also eliminated a duplicated call to `cactbotReadDataFiles` which was part of the modified code.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -14,6 +14,19 @@ class PopupText {
     this.kMaxRowsOfText = 2;
 
     this.Reset();
+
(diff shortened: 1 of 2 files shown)


quisquous:
Thanks!

Issue #673, opened 2019-10-16
raidboss alert ucob issue


Feanarus:
the announcements for the ""id: 'UCU Nael Your Doom'" (line 590-687) -part doesn't work 
copied the part from version 11.0 works for me 
Translation issue line 548 - 
old    regexDe: /Zur Herrschaft steige ich herab, auf umlodertem Pfadt/, 
new   regexDe: /Zur Herrschaft steige ich herab, auf umlodertem Pfad/, 
OverlayPlugin: Hibiya
Game Language: DE

Akurosia:
The line you are talking about was updated a year ago with #134 
So i am not sure how it can than work with the part from 11.0 tbh.

Will have a look at translation for this one maybe tomorrow.

Feanarus:
yes you're right i've made a diffchecker to unterstand the changes from 0.11.0 (green-right) to 0.11.1 (red-left)

https://www.diffchecker.com/nwMZQVp6

PR #674, opened 2019-10-18
raidboss: Fix 'gains effect' UCoB Triggers

panicstevenson:
Fixes https://github.com/quisquous/cactbot/issues/673


quisquous:
whoooops

PR #678, opened 2019-10-19
test: Add Testing for Regex Capturing Group Usage

panicstevenson:
Add validation for regex capturing group usage.

Test will fail on:

  - 'matches' in any trigger function, but no capturing group.
  - Capturing group present in regex, but no 'matches' in trigger
    functions.
  - Some languages have capturing group regex, but others do not.


test/trigger/test_trigger.js
```diff
@@ -108,6 +108,64 @@ let testUnnecessaryGroupRegex = function(file, contents) {
...
+    let containsMatches = false;
+    let triggerFunctions = [
+      'alarmText',
+      'alertText',
+      'condition',
+      'delaySeconds',
+      'infoText',
+      'preRun',
+      'run',
+      'suppressSeconds',
```
quisquous: durationSeconds, sound, soundVolume, disabled, groupTTS

Sorry for poor documentation.

It'd be nice to have this list somewhere such that it could also be used to verify that there aren't misspelled trigger entries.

test/trigger/test_trigger.js
```diff
@@ -108,6 +108,64 @@ let testUnnecessaryGroupRegex = function(file, contents) {
...
+      'tts',
+    ];
+
+    for (let j = 0; !containsMatches && j < triggerFunctions.length; j++) {
+      let currentTriggerFunction = currentTrigger[triggerFunctions[j]];
+      if (typeof currentTriggerFunction !== 'undefined' && currentTriggerFunction !== null)
+        containsMatches = currentTriggerFunction.toString().includes('matches');
+    }
+
+    let regexContainsCapturingGroups = /(\((?!\?:).*?\))/.test(currentTrigger.regex);
```
quisquous: *skeptical face*

I mean, I agree that this works for all triggers, but maybe you need a negative lookahead so that something like `\(real parens\)` doesn't match here too.

I actually wonder if the fact that javascript will yell at you for an unterminated group if all you need to do is verify that there's the beginning of a matching group?

test/trigger/test_trigger.js
```diff
@@ -108,6 +108,64 @@ let testUnnecessaryGroupRegex = function(file, contents) {
...
+      'infoText',
+      'preRun',
+      'run',
+      'suppressSeconds',
+      'tts',
+    ];
+
+    for (let j = 0; !containsMatches && j < triggerFunctions.length; j++) {
+      let currentTriggerFunction = currentTrigger[triggerFunctions[j]];
+      if (typeof currentTriggerFunction !== 'undefined' && currentTriggerFunction !== null)
```
quisquous: :innocent:

ui/raidboss/data/02-arr/raid/t9.js
```diff
@@ -162,10 +162,10 @@
     },
     {
       id: 'T9 Garotte Twist Lose',
-      regex: / 1E:\y{ObjectId}:(\y{Name}) loses the effect of Garrote Twist/,
-      regexDe: / 1E:\y{ObjectId}:(\y{Name}) loses the effect of Leicht Fixierbar/,
-      regexFr: / 1E:\y{ObjectId}:(\y{Name}) loses the effect of Sangle Accélérée/,
-      regexJa: / 1E:\y{ObjectId}:(\y{Name}) loses the effect of 拘束加速/,
+      regex: / 1E:\y{ObjectId}:\y{Name} loses the effect of Garrote Twist/,
```
quisquous: I think this is meant to have a conditional that this is you, because `T9 Garotte Twist Gain` only applies to you.

test/trigger/test_trigger.js
panicstevenson: Interesting... I thought I had solved exactly this, and I did [on this line](https://github.com/quisquous/cactbot/pull/678/files#diff-7e6ad728e47e4d5e6d9d7f6cfe5a7dd6R150). Time for variable.

panicstevenson: I've updated this, but we should discuss if it's good enough. For instance, it won't match on `\(hey\)` anymore, but it will match on `(hey)` and `(hey\)`.

ui/raidboss/data/02-arr/raid/t9.js
panicstevenson: I updated this to be what I thought logically made sense; ie the inverse of gain. Please check when you can 😄

test/trigger/test_trigger.js
ngld: If you just want to know how many captures the regex has, you can do something like this:
```javascript
let captures = new RegExp('(?:' + currentTrigger.regex.toString() + ')?').exec('').length - 1
```

quisquous: Ooh, that's nice.  :)

panicstevenson: Nice! This even found a case where there was two capturing groups for one language and one for the others.

ui/raidboss/data/04-sb/eureka/eureka_anemos.js
```diff
@@ -172,7 +172,7 @@
       regexDe: / 14:29C3:Nichtsschuppe starts using Eisball on (\y{Name})/,
       regexFr: / 14:29C3:Vidécailles starts using Boule De Glace on (\y{Name})/,
       regexJa: / 14:29C3:ヴォイドスケイル starts using 氷結球 on (\y{Name})/,
-      regexCn: / 14:29C3:虚无鳞龙 starts using (冰球|冻结) on (\y{Name})/,
+      regexCn: / 14:29C3:虚无鳞龙 starts using (?:冰球|冻结) on (\y{Name})/,
```
quisquous: !!!

PR #681, opened 2019-10-20
lint: Add ESLint Recommended Preset

panicstevenson:
Add the ESLint "recommended" preset which includes a few nice checks, most notably the checking of multiple spaces within a regular expression, which was updated in #677. Additional checking for JavaScript logic and syntax is included as well.

Compresses, reorders, and updates the package.json rules section to now exclude duplicate rules and to explicitly say "error" and "off", over 2 and 0, respectively.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,90 +19,74 @@
       "browser": true,
       "es6": true
     },
-    "extends": "google",
+    "extends": [
+      "eslint:recommended",
+      "google"
+    ],
     "rules": {
-      "no-else-return": 2,
+      "camelcase": "off",
       "curly": [
         "error",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -19,102 +19,102 @@
...
       "es6": true
     },
-    "extends": "google",
+    "extends": [
+      "eslint:recommended",
+      "google"
+    ],
     "rules": {
-      "no-else-return": 2,
+      "camelcase": "off",
```
quisquous: maybe some day

quisquous:
Thanks.  I looked through these rules and they seem reasonable to me.

PR #682, opened 2019-10-20
raidboss: Fix Suppress Trigger Function Spelling

panicstevenson:
Noticed that these triggers were a bit spammy when I ran this the other day, so I went back to add a `suppressSeconds` to the function, but noticed it was already there.


quisquous:
whoooooops

PR #691, opened 2019-10-23
oopsyraidsy: add cn language to oops html

Bluefissure:


Bluefissure:
I'm not sure about more localizations of other parts of oopsraidsy, will it work if I add translations to specific triggers in `oopsraidsy/data/.../*.js`?

quisquous:
Yeah, adding translations to specific triggers should do the trick.

I've tried to use effect and ability names where I could to limit the amount of translation needed.

PR #693, opened 2019-10-24
Initial Gubal (Hard) triggers and timelines

JLGarber:
Taking a break  from the (very) slow progress on `make_timeline`, this dungeon ended up being relatively simple.

I'm not terribly happy about the `00` game log lines for the phase changes on Liquid Flame, but as I noted in the timeline file, whichever phase comes second inevitably has multiple `gainsEffect` lines. I couldn't think of a way to clean that up with ACT log lines since the defining abilities (Sun/Moon seal, Magnetism, and Slosh) inside each phase are much too far in to be usable. (Maybe something could be done with `26`? But I didn't really have the time to untangle that one.)

Strix's ability names are too long. I cut "On The" from each to make things fit better, but if there's a better solution I'm much happier to do that. (Especially since this doesn't help at all for the FR localization.)

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
+++ b/ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
@@ -68,7 +68,7 @@
     },
     {
       id: 'Gubal Hard Ferrofluid Execute',
-      regex: / 1B:\y{ObjectId}:\y{Name}:....:....:0030|0031:/,
+      regex: / 1B:\y{ObjectId}:\y{Name}:....:....:(?:0030|0031):/,
       suppressSeconds: 5,
       delaySeconds: 0.5,
       infoText: function(data) {
@@ -177,7 +177,7 @@
       regex: /14:195D:Behemoth Ward starts using Ecliptic Meteor/,
       delaySeconds: 14, // Leaving about 10s warning to complete the LoS
       alertText: {
-        en: 'LoS Behemoth with boulder',
+        en: 'Hide behind boulder',
       },
     },
   ],
--- a/ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
+++ b/ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
@@ -1,0 +1,300 @@
+'use strict';
+
+// The Great Gubal Library--Hard
+[{
+  zoneRegex: /Great Gubal Library \(Hard\)/,
+  timelineFile: 'gubal_library_hard.txt',
+  timelineTriggers: [
+    {
+      id: 'Gubal Hard Triclip',
+      regex: /Triclip/,
+      beforeSeconds: 5,
+      condition: function(data) {
+        return data.role == 'healer' || data.role == 'tank';
+      },
+      infoText: {
+        en: 'Tank Buster',
+      },
+    },
+    {
+      id: 'Gubal Hard Searing Wind',
+      regex: /Searing Wind/,
+      beforeSeconds: 5,
+      infoText: function(data) {
+        if (data.role == 'tank') {
+          return {
+            en: 'Tank Cleave on you',
+          };
+        }
+        return {
+          en: 'Avoid tank cleave',
+        };
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Darkness',
+      regex: /Darkness \(buster\)/,
+      beforeSeconds: 5,
+      condition: function(data) {
+        return data.role == 'healer' || data.role == 'tank';
+      },
+      infoText: {
+        en: 'Tank Buster',
+      },
+    },
+  ],
+  triggers: [
+    {
+      id: 'Gubal Hard Bibliocide',
+      regex: / 14:1945:Liquid Flame starts using Bibliocide/,
+      condition: function(data) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      // Ferrofluid handling is shamelessly copied from Hades Normal
+      id: 'Gubal Hard Ferrofluid Collect',
+      regex: / 1B:\y{ObjectId}:(\y{Name}):....:....:(0030|0031):/,
+      run: function(data, matches) {
+        data.fluid = data.fluid || {};
+        if (matches[2] == '0030')
+          data.fluid[matches[1]] = 'positive';
+        if (matches[2] == '0031')
+          data.fluid[matches[1]] = 'negative';
+      },
+    },
+    {
+      id: 'Gubal Hard Ferrofluid Execute',
+      regex: / 1B:\y{ObjectId}:\y{Name}:....:....:0030|0031:/,
+      suppressSeconds: 5,
+      delaySeconds: 0.5,
+      infoText: function(data) {
+        if (data.fluid[data.me] == data.fluid['Liquid Flame']) {
+          return {
+            en: 'Close to boss',
+          };
+        }
+        return {
+          en: 'Away from boss',
+        };
+      },
+    },
+    {
+      id: 'Gubal Hard Ferrofluid Cleanup',
+      regex: / 1B:\y{ObjectId}:\y{Name}:....:....:003[01]:/,
+      delaySeconds: 5,
+      run: function(data) {
+        delete data.fluid;
+      },
+    },
+    {
+      id: 'Gubal Hard Slosh',
+      regex: / 23:\y{ObjectId}:Liquid Flame:\y{ObjectId}:(\y{Name}):....:....:0039/,
+      condition: function(data, matches) {
+        return data.me == matches[1];
+      },
+      infoText: {
+        en: 'Away from boss',
+      },
+    },
+    {
+      id: 'Gubal Hard Seals',
+      regex: / 1A:\y{ObjectId}:(\y{Name}) gains the effect of (Moon|Sun)seal/,
+      condition: function (data, matches) {
+        return data.me == matches[1];
+      },
+      infoText: function(data, matches) {
+        if (matches[2] == 'Moon') {
+          return {
+            en: 'Stand in blue',
+          };
+        }
+        return {
+          en: 'Stand in red',
+        };
+      },
+    },
+    {
+      // This inflicts a vulnerability stack on the tank if not interrupted
+      id: 'Gubal Hard Condensed Libra',
+      regex: / 14:198D:Mechanoscribe starts using Condensed Libra on \y{Name}/,
+      infoText: function(data) {
+        if (data.CanStun()) {
+          return {
+            en: 'Stun Mechanoscribe',
+          };
+        }
+        if (data.CanSilence()) {
+          return {
+            en: 'Interrupt Mechanoscribe',
+          };
+        }
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Quakes',
+      regex: / 14:1956:Strix starts using On The Properties Of Quakes/,
+      infoText: {
+        en: 'Stand in light circle',
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Tornadoes',
+      regex: / 14:1957:Strix starts using On The Properties Of Tornados/,
+      infoText: {
+        en: 'Stand in dark circle',
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Imps',
+      regex: / 14:1959:Strix starts using On The Properties Of Imps/,
+      infoText: {
+        en: 'Cleanse in green circle',
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Thunder',
+      regex: /14:195A:Strix starts using On The Properties Of Thunder III/,
+      infoText: {
+        en: 'Spread',
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Darkness II',
+      regex: / 14:1955:Strix starts using On The Properties Of Darkness II/,
+      condition: function(data) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      id: 'Gubal Hard Ecliptic Meteor',
+      regex: /14:195D:Behemoth Ward starts using Ecliptic Meteor/,
+      delaySeconds: 14, // Leaving about 10s warning to complete the LoS
+      alertText: {
+        en: 'LoS Behemoth with boulder',
+      },
+    },
+  ],
+  timelineReplace: [
+    {
+      'locale': 'de',
+      'replaceSync': {
+        'Demon of the Tome': 'Bücherdämon',
+        'Liquid Flame': 'flüssig[a] Flamme',
+        // Strix is the same for DE
+
+        'The Hall of Magicks': 'Halle der Magie',
+        'The Astrology and Astromancy Camera': 'Astrologisches und Astronomisches Gewölbe',
+        'The Rare Tomes Room': 'Abteilung für seltene Schriften',
+      },
+      'replaceText': {
+        'Triclip': 'Dreischnitt',
+        'Folio': 'Foliant',
+        'Book Drop': 'Publizieren',
+        'Issue': 'Publizieren',
+        'Frightful Roar': 'Furchtbares Brüllen',
+        'Discontinue': 'Druck einstellen',
+
+        'Searing Wind': 'Versengen',
+        'Bibliocide': 'Bibliozid',
+        'Sea Of Flames': 'Flammenmeer',
+        'Slosh': 'Durchbläuen',
+        'Seal Of Night And Day': 'Siegel',
+        'Magnetism': 'Magnetismus',
+        'Repel': 'Abstoßung',
+
+        'Check Out': 'Anthologie',
+        'Properties Of Darkness (buster)': 'Theorie der Dunkelung',
+        'Properties Of Quakes': 'Theorie des Seisga',
+        'Properties Of Darkness II': 'Theorie der Dunkelung II',
+        'Properties Of Tornados': 'Theorie des Tornado',
+        'Properties Of Imps': 'Über Flusskobolde',
+        'Properties Of Thunder III': 'Theorie des Blitzga',
+        'Meteor Impact': 'Meteoreinschlag',
+        'Ecliptic Meteor': 'Ekliptik-Meteor',
+      },
+    },
+    {
+      'locale': 'fr',
+      'replaceSync': {
+        'Demon of the Tome': 'Démon du Tome',
+        'Liquid Flame': 'flamme liquide',
+        // Strix is the same for FR
+
+        'The Hall of Magicks': 'Puits des magies',
+        'The Astrology and Astromancy Camera': 'Dôme d\'astrologie et d\'astromancie',
+        'The Rare Tomes Room': 'Dôme des manuscrits rares',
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+        data.fluid = data.fluid || {};
+        if (matches[2] == '0030')
+          data.fluid[matches[1]] = 'positive';
+        if (matches[2] == '0031')
+          data.fluid[matches[1]] = 'negative';
+      },
+    },
+    {
+      id: 'Gubal Hard Ferrofluid Execute',
+      regex: / 1B:\y{ObjectId}:\y{Name}:....:....:0030|0031:/,
```
panicstevenson: Should the or be within parentheses?

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      id: 'Gubal Hard Ecliptic Meteor',
+      regex: /14:195D:Behemoth Ward starts using Ecliptic Meteor/,
+      delaySeconds: 14, // Leaving about 10s warning to complete the LoS
+      alertText: {
+        en: 'LoS Behemoth with boulder',
```
panicstevenson: Maybe just `hide behind boulder`, capitalized however you prefer?

JLGarber: You're right, *what* over *how*. The meteor falling is obvious enough.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
JLGarber: Um. Wow. You're right, it should definitely be `(?:0030|0031)`. I'm a little surprised this didn't fail for me at least once, 25% potential failure rate isn't that low.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+      regex: / 14:1945:Liquid Flame starts using Bibliocide/,
+      condition: function(data) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      // Ferrofluid handling is shamelessly copied from Hades Normal
```
quisquous: This isn't really important to answer, but I'm not sure what you're referring to out of Hades Normal?

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+      condition: function(data, matches) {
+        return data.me == matches[1];
+      },
+      infoText: {
+        en: 'Away from boss',
+      },
+    },
+    {
+      id: 'Gubal Hard Seals',
+      regex: / 1A:\y{ObjectId}:(\y{Name}) gains the effect of (Moon|Sun)seal/,
```
quisquous: For the sake of translations, I think this would be better to break up into two separate triggers, one for Moon and one for Sun.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+      // This inflicts a vulnerability stack on the tank if not interrupted
+      id: 'Gubal Hard Condensed Libra',
+      regex: / 14:198D:Mechanoscribe starts using Condensed Libra on \y{Name}/,
+      infoText: function(data) {
+        if (data.CanStun()) {
+          return {
+            en: 'Stun Mechanoscribe',
+          };
+        }
+        if (data.CanSilence()) {
```
quisquous: Can you put the silence first before the stun? Sometimes things get stun resistant, so if you can silence, I think that's preferred.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+          data.fluid[matches[1]] = 'negative';
+      },
+    },
+    {
+      id: 'Gubal Hard Ferrofluid Execute',
+      regex: / 1B:\y{ObjectId}:\y{Name}:....:....:(?:0030|0031):/,
+      suppressSeconds: 5,
+      delaySeconds: 0.5,
+      infoText: function(data) {
+        if (data.fluid[data.me] == data.fluid['Liquid Flame']) {
```
quisquous: Ah, this is tricky because of the translation.  You can tell from the ObjectId whether something is an enemy or not, because it starts with '4'.

Does the boss always get listed first before the player or vice versa?

I'm mostly curious if this could be simplified somehow.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,300 @@
...
+    {
+      id: 'Gubal Hard Properties of Tornadoes',
+      regex: / 14:1957:Strix starts using On The Properties Of Tornados/,
+      infoText: {
+        en: 'Stand in dark circle',
+      },
+    },
+    {
+      id: 'Gubal Hard Properties of Imps',
+      regex: / 14:1959:Strix starts using On The Properties Of Imps/,
```
quisquous: What do you think about having a delaySeconds on this to account for the difference in time between when the cast starts and when you get the imp effect?

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.txt
```diff
@@ -0,0 +1,157 @@
...
+125.8 "Discontinue" sync /:Middle Shelf Tome:1940:/
+131.5 "Frightful Roar" sync /:Demon of the Tome:193B:/
+136.7 "Triclip" sync /:Demon of the Tome:193A:/
+144.7 "Folio" sync /:Demon of the Tome:193C:/
+150.8 "Issue"
+155.4 "Discontinue"
+156.8 "Issue"
+157.4 "Discontinue x3"
+
+# Ordinarily we would sync from the end of the last block
```
quisquous: Yeah, this is very reasonable, to sync from the start of the next block too.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.txt
```diff
@@ -0,0 +1,157 @@
...
+185.1 "Folio"
+191.8 "Book Drop"
+193.4 "Issue"
+199.0 "Discontinue"
+204.7 "Frightful Roar"
+
+# Liquid Flame
+# -ii 1946 1953 194B
+# We use game log lines for the phases because the ACT log lines are unreliable
+# Whichever intermission comes second ends up with phantom "Gains Effect" lines
```
quisquous: I'm not sure I understand the issue here.  Can you give an example of what they look like, or what you mean by phantom?

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.txt
```diff
@@ -0,0 +1,157 @@
...
+2126.1 "--sync--" sync /:Meteor:1A6A:/
+2126.7 "Meteor Impact" sync /:Behemoth Ward:195E:/
+2133.8 "Properties Of Darkness (buster)" sync /:Strix:1954:/
+2135.1 "--sync--" sync /:Meteor:1A6A:/
+2135.6 "Meteor Impact" sync /:Behemoth Ward:195E:/
+2143.7 "Ecliptic Meteor" sync /:Behemoth Ward:195D:/
+2144.7 "--sync--" sync /:Behemoth Ward:195C:/
+
+2155.0 "Check Out" sync /:Strix:194E:/ window 20,20
+2172.3 "Quakes/Tornados" sync /:Strix:195[67]:/
```
quisquous: I don't think I ever knew this was random!

quisquous:
Oh, thanks also for the oopsy changes too along with the timeline.  I find those to be pretty quick to do when you have a log, so hopefully it wasn't too much work.

> Strix's ability names are too long. I cut "On The" from each to make things fit better, but if there's a better solution I'm much happier to do that. (Especially since this doesn't help at all for the FR localization.)

This seems quite reasonable.  I think I tend to prefer to have the ability names in the timeline, but at the same time there's a balance to be had.  In that vein, I agree with you putting "buster" in the names here, as nobody's going to remember the difference between properties of darkness I and II.

> Taking a break from the (very) slow progress on make_timeline

I'm still excited for whenever this gets done.  :)

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
JLGarber: Oh, I was feeling too lazy to actually think about how to deal with Ferrofluid, so I just went to the closest encounter where I could remember a similar situation where multiple overhead markers interact. (During Hades, there's the situation where stack and spread markers come up at the same time, so it's the same concept of "do I want to stack on the target or not?")

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
JLGarber: You're right, that's definitely better.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
JLGarber: Good point, I hadn't considered that.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
JLGarber: It looked like it does? Or at least it seemed to the times I ran this. I was hesitant to rely on relative timing since I know I've seen encounters where that wasn't the case, hence the handling via collection and comparison. Do you have any suggestions on where similar simpler patterns might already exist in the repository?

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
JLGarber: The only issue I have with this is that it's *really* easy to end up too far from the cleanse point if the user isn't thinking about it, and if there's a delay the player might end up Imped for more than a GCD while they chase down a cleanse. As it is, if the user is in the center of the room, and they move as soon as the trigger fires, they will arrive at the cleanse point about 1.5-2s before the cast completes. Maybe "Green after Imp" would convey the intent better?

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.txt
JLGarber: Sure. Let's take Tornado:

```
1200.0 "--sync--" sync /00:....:The liquid flame gains the effect of Anemomorph/ window 200,0
1204.5 "Bibliocide" sync /:Liquid Flame:1945:/ window 10,10
1215.4 "Magnetism/Repel?" sync /:Liquid Flame:194[CD]:/
1226.9 "Bibliocide" sync /:Liquid Flame:1945:/
1237.8 "Magnetism/Repel" sync /:Liquid Flame:194[CD]:/
1246.1 "Form Shift"
```

Unfortunately, there are multiple `gainsEffect` lines in this block:

```
[23:26:48.082] 1A:4000C430:Liquid Flame gains the effect of Anemomorph from  for 1800.00 Seconds.
[23:26:50.354] 1A:4000C430:Liquid Flame gains the effect of Anemomorph from  for 1800.00 Seconds.
[23:26:54.011] 1A:4000C430:Liquid Flame gains the effect of Anemomorph from  for 1800.00 Seconds.
[23:26:56.744] 1A:4000C430:Liquid Flame gains the effect of Anemomorph from  for 1800.00 Seconds.
[23:27:00.010] 1A:4000C430:Liquid Flame gains the effect of Anemomorph from  for 1800.00 Seconds.
[23:27:02.919] 1A:4000C430:Liquid Flame gains the effect of Anemomorph from  for 1800.00 Seconds.
```

This seems like it would restrict us to lookbehind windows.  Otherwise we would end up with a constantly shifting timeline during the second phase. However, this phenomenon happens to both Tornado and Hand forms, whichever comes second.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
quisquous: No, that's fair.  I guess it's better to warn ahead of time, and so this is fine as it is.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.txt
JLGarber: I didn't either, I had always assumed it alternated. We learn SO much doing this.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.txt
quisquous: Ah, I see.  And every 2 seconds too is extremely rough.

I think the only way you could use the 1A lines here would be to put one sync per potential phase at the end of each phase group, and then jump to the right place.  In other words, N phases have N syncs at the end with small windows but big jumps instead of N phases with 1 sync at the beginning with big windows.

I'm not sure that's worth it to do.  This is probably fine, and since it's an effect it needs to be translated either way.

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
quisquous: I've depended on that for other things, but I think you're right that it feels like it might not be that robust.

I think a simpler trigger might look like this.  A single trigger that does this:
* condition: on me or the boss (by object id comparison)
* preRun: store the headmarker in a list (and create the list if it doesn't exist)
* infoText: if there aren't two things in the list, return.  If there are two things in the list, compare them and print an alert.  Also, delete the list.

JLGarber: Oh, I see where you're going with that. You're right, that would break in non-English locales. Okay. So maybe change over the collection regex to capture the ObjectID instead, then filter for which one starts with 4 and set that one as the primary. Something like this but better...

```
    {
      id: 'Gubal Hard Ferrofluid Collect',
      regex: / 1B:\(y{ObjectId}):\y{Name}:....:....:(0030|0031):/,
      run: function(data, matches) {
        if (matches[1].slice(0,1) == '4') {
          data.flameId = data.flameId || matches[1];
        }
        data.fluid = data.fluid || {};
        if (matches[2] == '0030')
          data.fluid[matches[1]] = 'positive';
        if (matches[2] == '0031')
          data.fluid[matches[1]] = 'negative';
      },
    },
    {
      id: 'Gubal Hard Ferrofluid Execute',
      regex: / 1B:(\y{ObjectId}):(\y{Name}):....:....:(?:0030|0031):/,
      delaySeconds: 0.5,
        return matches[2] == data.me;
      }
      infoText: function(data) {
        if (data.fluid[matches[1]] == data.fluid[data.flameId]) {
          return {
            en: 'Close to boss',
          };
        }
        return {
          en: 'Away from boss',
        };
      },
    },
```

JLGarber: Hmm, I need to refresh the page before commenting, it seems. Yes, your suggestion would definitely be easier than what I have there. I'll poke at that tomorrow.

JLGarber: Well, never mind, it's *technically* tomorrow!

```
    {
      id: 'Gubal Hard Ferrofluid',
      regex: / 1B:(\y{ObjectId}):(\y{Name}):....:....:(0030|0031):/,
        return data.me == matches[2] || matches[1].slice(0,1) == '4';
      },
      preRun: function(data, matches) {
        data.markers = data.markers || [];
        data.markers.push(matches[3]);
      },
      infoText: function(data) {
        if (data.markers.length < 2)
          return;
        if (data.markers[1] == data.markers[2]) {
          delete data.markers;
          return {
            en: 'Close to boss',
          };
        }
        delete data.markers;
        return {
          en: 'Away from boss',
        };
      },
    },
```

I appreciate the pointers on this.  It's definitely better than stacking three triggers and is locale-agnostic too. I don't like duplicating the deletion, but adding a `run` command would end up being just as long. Any further suggestions on it?

quisquous: Hmm! You could store the markers off into a separate variable, and then delete them?
```
let sameMarkers = data.markers[0] == data.markers[1];
delete data.markers;
if (sameMarkers) {
// etc
```

JLGarber: Definitely better. It is done!

quisquous:
Thanks!

(If you're not already using `npm run lintfix`, it will take care of most linting issues automatically.)

JLGarber:
Alright, I think that should take care of that. No, this was fine. When there aren't oddities the timeline work is a lot easier than I had thought when I first started contributing.

As far as labeling the buster, I don't like it not being localized, but I figured that it was helpful enough and wouldn't cause problems since the timeline is localized.

I hope to have an update to `make_timeline` ready sometime this week. It's been challenging but I'm pleased with what it should be able to do.

On Lintfix, there's a story behind that that involves not knowing about how `package.json` works. Not really sure how the previous rounds of searching didn't turn that up, but going forward things will be 100% lint-free (or as much as ESlint can make it.)

ui/raidboss/data/03-hw/dungeon/gubal_library_hard.js
```diff
@@ -0,0 +1,292 @@
...
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      id: 'Gubal Hard Ferrofluid',
+      regex: / 1B:(\y{ObjectId}):(\y{Name}):....:....:(0030|0031):/,
+      condition: function(data, matches) {
+        return data.me == matches[2] || matches[1].slice(0, 1) == '4';
+      },
```
panicstevenson: \+ @JLGarber @quisquous 
Not to revive dead PRs or anything, but can either of you explain the above to me? This is the only instance in all of cactbot where we're matching on the ObjectId. It doesn't look to be erroneous since we're matching on the first character.

JLGarber: [This was a deliberate choice to avoid translation issues](https://github.com/quisquous/cactbot/pull/693#discussion_r339323965). In [this revision](https://github.com/quisquous/cactbot/pull/693/commits/0d5e437dbce04c344fe7ac345ac10f6887eb08bf#diff-7bee985354c543175e0679847d231b98R68), we removed the hard-coded name for Liquid Flame, replacing it with a more generic test for "is this a non-player". This was the simplest(maybe?) solution we came up with to determine whether a particular overhead marker belonged to a player or the boss.

panicstevenson: Both? I guess I'm fine to work around it, but I just thought it was very odd and hadn't seen it before.

JLGarber: Fair enough. Let me know if further updates to this trigger file would be helpful

Issue #646, opened 2019-10-05
util: let test_timeline.py and make_timeline.py autoselect encounters from a file


quisquous:
Right now, it's a pain to select start and end times when building and testing timelines.  If no start and end times are selected, it'd be nice to list all of the potential encounters (seal/unseal, and detected wipes) by number, and then just let you pass in `-rf` to select a single fight as you can with fflogs.

Just as a note, I've seen some issues with picking times when the network log has a different timezone (you have to specify times with local times).  It'd be good to test this with this new feature.

This was mentioned in #636, so filing it as a feature request.

quisquous:
Also additionally it would be nice to not have to specify a timeline file, if it could somehow figure out which timelines applied based on zone.

JLGarber:
For `make_timeline`, how open are we to parsing the file twice if no start/end are specified, once to find encounters and the second pass as it currently exists? For minimum effort, it seems we could add  `find_encounters_in_file` that would quickly iterate through and pull out the start/end times to pass to `parse_file`. Something like:

```
        for log_line in file:
            # Ignore log lines that aren't game status info
            if not (log_line[39:42] == '0839'):
                continue

            # Build start/end time groupings
            log_line_fields = log_line.split('|')
            if (log_line_fields[4].endswith('sealed off in 15 seconds!')):
                encounter_starts.append(parse_line_time(log_line))
                continue
            if (log_line_fields[4].endswith('no longer sealed!')):
                encounter_ends.append(parse_line_time(log_line))
                continue
```

From here we can easily return a list of encounter start/end pairs. With a little extra work it should be not-difficult to add the first few NPC names that appear after the zone sealing message to accompany the timestamps. Or we could just add the zone name to the timestamp and let the user work it out from there.

(Apologies for the ugly. I'm vastly unfamiliar with Python's file parsing, so I don't know whether the performance benefits of not parsing twice would be worth the effort to extend `parse_file` cleanly. Nor do I know whether native string methods would be better/worse than regex for this purpose, it's just what I grabbed out of the standard library.)

One major benefit of the extra pass would be ease of dropping encounters that are under a certain (potentially configurable) length. If we try to do it all in-line, I feel like it's a little complex to enable listing encounters from within `parse_file`. Additionally, the extra pass would only need to be incurred if the user wants to avoid using start/end flags, which seems like an acceptable trade-off in that situation.

[...]

PR #700, opened 2019-10-31
Util: make_timeline automation

JLGarber:
It's here! Sort of.

(This is in response to #646)

Before we go into details, there's a great deal of room for improvement on this one. It's truly a draft and will need much more input before it's ready to go live. Most of the choices I made on handling various aspects of this had to do with "This was an obvious tool to use from the standard library" or "I just want this to work, we can always clean it up later". I haven't used Python for anything besides personal education before, so this was a bit of an adventure.

Now then. The changes I made involve the addition of an argument, `search_fights` (aliased to `-lf` currently,) which is (intended to be) mutually exclusive with `-start` + `-end` or `-report`. All other arguments should be compatible. 

`-search_fights` takes zero or one integer arguments, from 1 upward, indicating which encounter in the chosen log file to use. When called without, it lists the encounters found in the log file.

```
python make_timeline.py -f testlog.log -lf -1
0. 01:06:51 01:08:49 The Astrology and Astromancy Camera
1. 01:30:31 01:33:32 Repurposing
2. 01:37:26 01:40:05 Aetherial Observation
3. 01:43:36 01:47:45 The Cornice
4. 01:57:27 01:59:03 The Hall of the Tainted
5. 02:00:06 02:01:08 The Walk of Lament
6. 02:01:33 02:02:37 The Hall of the Deviant
7. 02:03:50 02:05:18 The Walk of Fire
8. 02:07:04 02:08:38.349000 The Hall of the Inexorable
9. 02:11:28.053000 02:18:08.061000 Eden's Gate: Resurrection
10. 14:47:31 14:49:25 Repurposing
11. 14:52:07 14:54:02 Aetherial Observation
12. 14:56:54 14:59:45.967000 The Cornice
13. 15:50:23.554000 15:57:28.049000 Eden's Gate: Resurrection
14. 16:08:05.251000 16:14:45.434000 Eden's Gate: Descent
15. 16:37:58.044000 16:44:57.727000 Eden's Gate: Inundation
16. 16:56:10.509000 17:02:34.354000 Eden's Gate: Sepulture
17. 17:55:52.948000 18:03:00.745000 Eden's Gate: Resurrection
18. 18:28:07.463000 18:35:45.697000 Eden's Gate: Descent
19. 18:37:48.648000 18:46:24.621000 Eden's Gate: Inundation
```

From here, the user can either repeat the command, replacing -1 with the index of the fight they want, or they may opt to copy those timestamps and run them through manually.

Known issues/deficiencies:

1. Double-pass file parsing. Because I'm not terribly familiar with Python's file handling, I opted to scan the file for encounters first before actually making the timeline. This does potentially double the run time.

2. Microsecond elision. By default, `datetime` objects elide zero-value microseconds.  If the user were to select one of the encounters without microseconds, as seen in 0-8 and 10-12, the command would fail to find any encounters. (The script would attempt to match, say, `01:06:51` in the generated encounter listing against `01:06:51.000` in the log.)  This isn't insurmountable, as there are a couple of places in the new code where we could check whether the timestamp format fits and pad/trim it as necessary. However, I didn't want to do that without input on how/where, since it would be even messier otherwise.

3. Unguarded arguments. `-search_fights` does not currently validate that the user's index input is -1 or greater. I wasn't sure where to handle that, and I didn't have the time to fully familiarize with `argparse`. In line with that, timestamp validation for `-start` and `-end` is currently disabled so that I could actually get the script to run.

4. Duplicated/excessive string handling. Every `00`, `01`, `33`, and `36` log line is  split, whether or not it's relevant. This made it easier for me to do comparisons in the `is_line_start`/`is_line_end` functions, but it could clearly be improved.

5. Large log files. As I've noted elsewhere, I default to 24 hours as the longest any of my log files grow. I have no idea what would happen if the user attempted this on a multi-day file.

The operation of `make_timeline` is unchanged as far as its internal workings. The entirety of the new functionality branches right after the log file is opened, at line 101/116. From there, the file is scanned for encounters, and if the user didn't specify one to return, the list of encounters is printed to the console in the same way timelines are. If they did, the start and end times at that index are set as the start/end times for `parse_file`, and the script continues just as if the user had used `-start` + `-end`.

Most of the new functionality is commented, but there are likely places that will need fuller explanation. I would be happy to receive suggestions on what could be improved.

All that being said, in its current state, the script would be "usable" for new work immediately, in the sense that it can quickly and accurately identify encounters within a log file. It's not merge-ready, but it's mostly functional.

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -48,7 +48,7 @@ def is_zone_unseal(line_fields):
 
 def is_line_attack(line_fields):
     # We want only situations where a friendly attacks an enemy
-    return (line_fields[0] in ('21', '22') and line_fields[6].startswith('4'))
+    return line_fields[0] in ('21', '22') and line_fields[6].startswith('4')
 
 def is_limit_reset(line_fields):
     return line_fields[4] == 'The limit gauge resets!'
@@ -142,7 +142,7 @@ def parse_file(args):
             encounter_sets = find_fights_in_file(file)
             # If all we want to do is list encounters, stop here and give to the user.
             if args.search_fights < 0:
-                return [f'{i + 1}. {e_info[0]} {e_info[1]} {e_info[2]}' for i, e_info in enumerate(encounter_sets)]
+                return [f'{i + 1}. {" ".join(e_info)}' for i, e_info in enumerate(encounter_sets)]
             elif args.search_fights > len(encounter_sets):
                 raise Exception('Selected fight index not in selected ACT log.')
         # Scan the file until the start timestamp
@@ -197,7 +197,7 @@ def find_fights_in_file(file):
     encounter_sets = []
     for line in file:
         # Ignore log lines that aren't game status info
-        if not (line[37:41] == '0839' or line[0:2] in ['00', '01', '21', '22', '33']):
+        if line[37:41] != '0839' and line[0:2] not in ['00', '01', '21', '22', '33']:
             continue
         line_fields = line.split('|')
 
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -29,11 +29,6 @@ def stringify_time(timestamp):
     Trims microseconds to milliseconds."""
     return timestamp.strftime('%H:%M:%S.%f')[:-3]
 
-def is_line_start(line_fields, started):
-    # Zone seal messages are guaranteed to be an encounter start
-    if is_zone_seal(line_fields):
-        return True
-    return is_line_attack(line_fields) and not started
 
 def is_line_end(line_fields):
     if is_zone_unseal(line_fields) or is_limit_reset(line_fields):
@@ -216,12 +211,12 @@ def find_fights_in_file(file):
             continue
 
         # Build start/end time groupings
-        if is_line_start(line_fields, encounter_in_progress):
-            # This nested IF is messy, but it feels like the clearest way to express this
-            if is_zone_seal(line_fields):
-                encounter_start_staging = [parse_line_time(line), line_fields[4].split(' will be sealed off')[0]]
-            else:
-                encounter_start_staging = [parse_line_time(line), current_instance]
+        if is_zone_seal(line_fields):
+            encounter_start_staging = [parse_line_time(line), line_fields[4].split(' will be sealed off')[0]]
+            encounter_in_progress = True
+            continue
+        elif not encounter_in_progress and is_line_attack(line_fields):
+            encounter_start_staging = [parse_line_time(line), current_instance]
             encounter_in_progress = True
             continue
         # If this fired regardless of an encounter being found, we would end up with phantom encounters.
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -32,20 +32,43 @@ def stringify_time(timestamp):
     return timestamp.strftime('%H:%M:%S.%f')[:-3]
 
 def is_line_start(line_fields, started):
-    # FIXME: May need localization?
-    if (line_fields[4].endswith('sealed off in 15 seconds!')):
+    # Zone seal messages are guaranteed to be an encounter start
+    if is_zone_seal(line_fields):
         return True
-    if line_fields[0] == '36' and line_fields[2] == '0000' and started == False:
+    # Except in trial instances, the 0000 limit break line indicates a start
+    if is_limit_break_zero(line_fields) and not started:
         return True
     return False
 
 def is_line_end(line_fields):
-    # FIXME: May need localization?
-    if line_fields[4].endswith('no longer sealed!') or line_fields[4] == 'The limit gauge resets!' or line_fields[4].endswith('has ended.'):
+    if is_zone_unseal(line_fields) or is_limit_reset(line_fields):
         return True
-    if line_fields[0] == '33'and (line_fields[3] == '40000010' or line_fields[3] == '40000003'):
+    if is_encounter_end_code(line_fields):
         return True
     return False
+
+def is_zone_seal(line_fields):
+    return line_fields[4].endswith('sealed off in 15 seconds!')
+
+def is_zone_unseal(line_fields):
+    return line_fields[4].endswith('no longer sealed!')
+
+def is_limit_break_zero(line_fields):
+    return line_fields[0] == '36' and line_fields[2] == '0000'
+
+def is_limit_reset(line_fields):
+    return line_fields[4] == 'The limit gauge resets!'
+
+def is_instance_begun(line_fields):
+    return line_fields[0] == '00' and line_fields[4].endswith('has begun.')
+
+def is_instance_ended(line_fields):
+    return line_fields[0] == '00' and line_fields[4].endswith('has ended.')
+
+def is_encounter_end_code(line_fields):
+    if not line_fields[0] == '33':
+        return False
+    return line_fields[3] == '40000010' or line_fields[3] == '40000003'
 
 def parse_report(args):
     """Reads an fflogs report and return a list of entries"""
@@ -124,21 +147,21 @@ def parse_file(args):
         if args.search_fights:
             encounter_sets = find_fights_in_file(file)
             # If all we want to do is list encounters, stop here and give to the user.
-            if args.search_fights[0] < 0:
-                return [f'{i}. {encounter_name}' for i, encounter_name in enumerate(encounter_sets)]
+            if args.search_fights < 0:
+                return [f'{i + 1}. {e_info[0]} {e_info[1]} {e_info[2]}' for i, e_info in enumerate(encounter_sets)]
         # Scan the file until the start timestamp
         for line in file:
             start_time = end_time = 0
             if args.search_fights:
-                start_time = (encounter_sets[args.search_fights[0]][0])
-                end_time = (encounter_sets[args.search_fights[0]][1])
+                start_time = (encounter_sets[args.search_fights - 1][0])
+                end_time = (encounter_sets[args.search_fights - 1][1])
             else:
                 start_time = args.start
                 end_time = args.end
-            if not started and not (line[14:26] == start_time or line[14:29] == start_time):
+            if not started and start_time not in (line[14:26], line[14:29]):
                 continue
 
-            if line[14:26] == end_time or line[14:29] == end_time:
+            if end_time in (line[14:26], line[14:29]):
                 break
 
             # We're at the start of the encounter now.
@@ -173,32 +196,32 @@ def find_fights_in_file(file):
     encounter_start_staging = 0 # Staged to avoid spurious starts such as limit break usage
     e_ends = []
     encounter_in_progress = False
-    current_instance = "none"
+    current_instance = None
     encounter_sets = []
     for line in file:
         # Ignore log lines that aren't game status info
-        if not (line[37:41] == '0839') and not (line[0:2]in ['00','01','33','36']):
+        if not (line[37:41] == '0839') and not (line[0:2] in ['00', '01', '33', '36']):
             continue
         line_fields = line.split('|')
 
         # Update current instance for returning alongside timestamps
         # Don't update if we're not entering a combat instance
-        if line_fields[0] == '00' and line_fields[4].endswith('has begun.'):
+        if is_instance_begun(line_fields):
             current_instance = line_fields[4].split(' has begun')[0]
             continue
-        if line_fields[0] == '00' and line_fields[4].endswith('has ended.'):
+        if is_instance_ended(line_fields):
             current_instance = None
             encounter_in_progress = False
             continue
-        
+
         # We don't want to look for encounters outside a combat instance
-        if current_instance == None:
+        if current_instance is None:
             continue
 
         # Build start/end time groupings
         if is_line_start(line_fields, encounter_in_progress):
             # This nested IF is messy, but it feels like the clearest way to express this
-            if line_fields[4].endswith('sealed off in 15 seconds!'):
+            if is_zone_seal(line_fields):
                 encounter_start_staging = [parse_line_time(line), line_fields[4].split(' will be sealed off')[0]]
             else:
                 encounter_start_staging = [parse_line_time(line), current_instance]
@@ -221,6 +244,7 @@ def find_fights_in_file(file):
         encounter_sets.append(encounter_info)
     file.seek(0)
     return encounter_sets
+
 
 def main(args):
     timeline_position = 0
@@ -264,7 +288,7 @@ def main(args):
         phases[ability] = float(time)
 
     # Get the entry list or return fight list
-    if args.search_fights and args.search_fights[0] == -1:
+    if args.search_fights and args.search_fights == -1:
         return parse_file(args)
     elif args.report:
         entries, start_time = parse_report(args)
@@ -377,7 +401,7 @@ if __name__ == "__main__":
     # Log file arguments
     parser.add_argument('-s', '--start', help="Timestamp of the start, e.g. '12:34:56.789")
     parser.add_argument('-e', '--end', help="Timestamp of the end, e.g. '12:34:56.789")
-    parser.add_argument('-lf', '--search_fights', nargs=1, type=int, help="Encounter whose start/stop times should be used, e.g. '1'. -1 to list all.")
+    parser.add_argument('-lf', '--search_fights', nargs='?', const=-1, type=int, help="Encounter in log to use, e.g. '1'. If no number is specified, returns a list of encounters.")
 
     # Filtering arguments
     parser.add_argument('-ii', '--ignore-id', nargs='*', default=[], help="Ability IDs to ignore, e.g. 27EF")
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -6,18 +6,16 @@ import fflogs
 import timeline_aggregator
 
 
-# def timestamp_type(arg):
-#     """Defines the timestamp input format"""
-#     if re.match(r'\d{2}:\d{2}:\d{2}\.\d{3}', arg) is None:
-#         raise argparse.ArgumentTypeError(
-#             "Invalid timestamp format. Use the format 12:34:56.789")
-#     return arg
-
+def timestamp_type(arg):
+    """Defines the timestamp input format"""
+    if arg and re.match(r'\d{2}:\d{2}:\d{2}\.\d{3}', arg) is None:
+        raise argparse.ArgumentTypeError(
+            "Invalid timestamp format. Use the format 12:34:56.789")
+    return arg
 
 def parse_time(timestamp):
     """Parses a timestamp into a datetime object"""
     return datetime.strptime(timestamp, '%Y-%m-%dT%H:%M:%S')
-
 
 def parse_line_time(line):
     """Parses the line's timestamp into a datetime object"""
@@ -143,12 +141,14 @@ def parse_file(args):
     started = False
     encounter_sets = []
     with args.file as file:
-        # If searching for encounters, divert and find start/end first
+        # If searching for encounters, divert and find start/end first3
         if args.search_fights:
             encounter_sets = find_fights_in_file(file)
             # If all we want to do is list encounters, stop here and give to the user.
             if args.search_fights < 0:
                 return [f'{i + 1}. {e_info[0]} {e_info[1]} {e_info[2]}' for i, e_info in enumerate(encounter_sets)]
+            elif args.search_fights > len(encounter_sets):
+                raise Exception('Selected fight index not in selected ACT log.')
         # Scan the file until the start timestamp
         for line in file:
             start_time = end_time = 0
@@ -399,8 +399,8 @@ if __name__ == "__main__":
     parser.add_argument('-rf', '--fight', type=int, help="Fight ID of the report to use. Defaults to longest in the report")
 
     # Log file arguments
-    parser.add_argument('-s', '--start', help="Timestamp of the start, e.g. '12:34:56.789")
-    parser.add_argument('-e', '--end', help="Timestamp of the end, e.g. '12:34:56.789")
+    parser.add_argument('-s', '--start', type=timestamp_type, help="Timestamp of the start, e.g. '12:34:56.789")
+    parser.add_argument('-e', '--end', type=timestamp_type, help="Timestamp of the end, e.g. '12:34:56.789")
     parser.add_argument('-lf', '--search_fights', nargs='?', const=-1, type=int, help="Encounter in log to use, e.g. '1'. If no number is specified, returns a list of encounters.")
 
     # Filtering arguments
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -275,7 +275,9 @@ def main(args):
         'Minfilia',
         'Thancred',
         'Urianger',
-        ''
+        '',
+        '2P',
+        'Crystal Exarch',
     ]
 
     # Format the phase timings
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -6,12 +6,12 @@ import fflogs
 import timeline_aggregator
 
 
-def timestamp_type(arg):
-    """Defines the timestamp input format"""
-    if re.match('\d{2}:\d{2}:\d{2}\.\d{3}', arg) is None:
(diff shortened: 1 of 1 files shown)


JLGarber:
Just to add a couple new issues I've come across that aren't showstoppers, but might be annoying:

5. Instance detection currently relies on the "[INSTANCE] has begun" log line, which is consistent and mostly fine. However, if the user does not have ACT open at the time this line *would* be sent, none of the encounters in the instance will be included in the listing. It does look like there are multiple `01` log lines that get sent during the instance, whenever the user crosses a zone boundary, so I think it should be possible to use that as a backup.

6. I haven't tested it extensively, but at least on Hades the limit break 0000 log line is NOT an indicator the fight has started. I'm not sure why, but it does mean we'll have to come up with something else for that fight. I did confirm that it's not a general new thing for single-encounter instances though. (It's going to be really sad if all single-encounter instances going forward do this nonsense.)

util/make_timeline.py
```diff
@@ -97,14 +112,28 @@ def parse_file(args):
...
-
+    encounter_sets = []
     with args.file as file:
+        # If searching for encounters, divert and find start/end first
+        if args.search_fights:
+            encounter_sets = find_fights_in_file(file)
+            if args.search_fights[0] < 0:
+                for j in range(0, len(encounter_sets)):
+                    encounter_sets[j] = str(j) + '. ' + (" ").join(encounter_sets[j])
+                return encounter_sets
```
panicstevenson: If you want to be slick:
```
return [f'{i}. {encounter_name}' for i, encounter_name in enumerate(encounter_sets)]
```

JLGarber: That's... huh. I had no idea F-string formatting was a thing. There really *is* a tool for everything in Python. That could prove very useful.

quisquous:
Ah, this is super exciting! Let me put in a little bit more work on Hades Ex stuff, and then I will get back to this this weekend.

JLGarber:
Not a problem, I won't have much more time to work on this before Sunday myself. I need to confirm it, but I'm reaching the conclusion that trials just don't handle limit break log lines in a way that's going to be usable. While awful, it's still light-years better than the possibility that Square changed limit break generation for all new instances.

Once we get this a little more stable, I'm reasonably confident that updating `test_timeline` should be mostly simple. The same technique I used here should just drop right in once it's tweaked to be cleaner. (I do plan to look into parsing timeline files for start/end timestamps though, since that would negate the heuristic problem for trials.)

quisquous:
> Microsecond elision. By default, datetime objects elide zero-value microseconds. If the user were to select one of the encounters without microseconds, as seen in 0-8 and 10-12, the command would fail to find any encounters. (The script would attempt to match, say, 01:06:51 in the generated encounter listing against 01:06:51.000 in the log.) This isn't insurmountable, as there are a couple of places in the new code where we could check whether the timestamp format fits and pad/trim it as necessary. However, I didn't want to do that without input on how/where, since it would be even messier otherwise.

I'd recommend: https://docs.python.org/3/library/datetime.html#datetime.date.strftime and add some function that converts a datetime into a string that would match a log time.

JLGarber:
That makes sense. I suppose it should have been obvious that there would be an inverse function. That shouldn't be too hard to implement then, now that I see it.

JLGarber:
Well, that works. That pushes this thing to functional for non-trial encounters. "Just" have to clean things up now. I'm *really* happy with its operation now. Auto-generation is noticeably slower than selecting explicit timestamps due to the double parse, but still *much* faster than manual timestamp entry.

```
cactbot\util>python make_timeline.py -f logfile.log -lf -1
0. 17:27:32.000 17:29:26.000 Rhalgr's Gate
1. 17:33:25.000 17:35:23.000 The Chamber of Knowledge
2. 17:39:48.000 17:43:37.178 The Hall of the Griffin
3. 21:11:11.000 21:13:55.000 The Martial Court
4. 21:16:40.000 21:19:26.000 The Font of Quintessence
5. 21:24:55.000 21:29:14.489 The Chamber of Celestial Song
6. 23:30:24.534 23:33:11.459 Alexander - The Burden of the Father
```
Trials are a PAIN. Limit break generation works very differently in trial instances, in that the 0000 line is sent immediately upon wipe rather than on pull. Additionally, the on-pull limit break line can be delayed up to two seconds, so it's totally useless. 

The only consistent thing I've found so far is the boss's initial emote. (In Hades' case, using that as an example since you're on that recently, that line would be *"Show me your “vaunted” strength. And I shall expose the lie of your fragmented existence!"*) I don't like the idea, but perhaps adding a list of known encounter lines that we regularly update could work. (Or maybe pull them in from a config file, although adding yet *another* helper file isn't great.) Not an ideal situation by any means, but a one-time  change to the script would save the time easily enough in not having to specify start/end times.

quisquous:
> Auto-generation is noticeably slower than selecting explicit timestamps due to the double parse, but still much faster than manual timestamp entry.

I think the double parse is the best option here, so I wouldn't overthink it too much.

> Trials are a PAIN. Limit break generation works very differently in trial instances, in that the 0000 line is sent immediately upon wipe rather than on pull. Additionally, the on-pull limit break line can be delayed up to two seconds, so it's totally useless.

I was thinking about this, and I think trial could work with just looking for the first attack (15/16 log line) after a zone change/wipe, as long as it is ended by a 21 wipe message (to differentiate from a striking dummy or some overworld mob)?

util/make_timeline.py
```diff
@@ -284,8 +375,9 @@ def main(args):
     parser.add_argument('-rf', '--fight', type=int, help="Fight ID of the report to use. Defaults to longest in the report")
 
     # Log file arguments
-    parser.add_argument('-s', '--start', type=timestamp_type, help="Timestamp of the start, e.g. '12:34:56.789")
-    parser.add_argument('-e', '--end', type=timestamp_type, help="Timestamp of the end, e.g. '12:34:56.789")
+    parser.add_argument('-s', '--start', help="Timestamp of the start, e.g. '12:34:56.789")
+    parser.add_argument('-e', '--end', help="Timestamp of the end, e.g. '12:34:56.789")
+    parser.add_argument('-lf', '--search_fights', nargs=1, type=int, help="Encounter whose start/stop times should be used, e.g. '1'. -1 to list all.")
```
quisquous: I can't remember how parseargs work, but is it possible to have `-lf` on its own (with no magic -1 value) do the automatic listing?

util/make_timeline.py
```diff
@@ -284,8 +375,9 @@ def main(args):
     parser.add_argument('-rf', '--fight', type=int, help="Fight ID of the report to use. Defaults to longest in the report")
 
     # Log file arguments
-    parser.add_argument('-s', '--start', type=timestamp_type, help="Timestamp of the start, e.g. '12:34:56.789")
-    parser.add_argument('-e', '--end', type=timestamp_type, help="Timestamp of the end, e.g. '12:34:56.789")
+    parser.add_argument('-s', '--start', help="Timestamp of the start, e.g. '12:34:56.789")
```
quisquous: I assume you're still going to look into what's going on with start/end/timestamp_type as well? I know you mentioned you had commented this out for testing.

util/make_timeline.py
JLGarber: From what I can see, it should be possible with `nargs='?', type=int, const=-1`. The brief change I made to permit that seems to work okay, but I'll want to test it more before turning it in.

util/make_timeline.py
```diff
@@ -134,9 +168,61 @@ def parse_file(args):
...
+def find_fights_in_file(file):
+    e_starts = []
+    encounter_start_staging = 0 # Staged to avoid spurious starts such as limit break usage
+    e_ends = []
+    encounter_in_progress = False
+    current_instance = "none"
+    encounter_sets = []
+    for line in file:
+        # Ignore log lines that aren't game status info
+        if not (line[37:41] == '0839') and not (line[0:2]in ['00','01','33','36']):
```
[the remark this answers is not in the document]
JLGarber: Oof, true, I haven't got pylint (or even pip) set up on this machine since I don't do much Python work. I'll ensure that's taken care of before finalizing the pull request.

util/make_timeline.py
```diff
@@ -25,6 +25,27 @@ def parse_line_time(line):
...
 
+def stringify_time(timestamp):
+    """
+    Parses a datetime object into a string for log comparisons.
+    Trims microseconds to milliseconds."""
+    return timestamp.strftime('%H:%M:%S.%f')[:-3]
+
+def is_line_start(line_fields, started):
+    # FIXME: May need localization?
+    if (line_fields[4].endswith('sealed off in 15 seconds!')):
```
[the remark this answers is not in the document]
JLGarber: That's a good point, I probably shouldn't have rushed on these quite so much. Helper functions absolutely would be the way to go here.

As for the translation, you're right, there's not much reason to account for localization when the timeline itself never will be localized.

util/make_timeline.py
JLGarber: Yes, although I'm a little more concerned about this than I am the rest of the changes. I had hoped to get through these updates without needing to completely familiarize with `argparse`, but I'm not seeing a clear path to enable validation as you had originally set it up. That most likely means I don't know enough, so I'll have to try to take some time this next week to dig through `argparse` and learn it properly.

util/make_timeline.py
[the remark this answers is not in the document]
quisquous: No need to self-deprecate here!  For what it's worth, this patch was super useful to me already, and if you had held off, I wouldn't have been able to use it.  Please take code review comments as suggestions to make code better together, and definitely not as passive aggressive hints that you should have written it differently in the first place.  :)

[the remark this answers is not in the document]
JLGarber: Oh, no, rest assured, I didn't take it that way. I work out problems by talking through them, and that carries over into some other things like this. I think we're on the same page with "the purpose of code reviews", and I'm *very* happy this has been useful in time for the 5.1 release.

JLGarber:
On using 15/16 log lines for encounter start times, I think that should work, but it will probably be necessary to build in some extra support for it when ending encounters. Otherwise trash pulls will end up in the listing. (I *think* I've already accounted for overworld mobs etc by only permitting encounters to start while in an instance, and immediately ending an active encounter if the user leaves the instance. If you saw overworld mobs in the generated listing, could I get a copy of that logfile?) The issue there is that with the current meta of "pull to the wall and AoE", it's very easy for encounters to last more than 1 minute. It could be that we should reconfigure that to 3 minutes or something though, since it's almost always impossible to get a usable timeline with a shorter encounter length.

util/make_timeline.py
JLGarber: Well, that wasn't too difficult. I'm not totally happy with the implementation, but it works and overall just makes more sense to the user.

util/make_timeline.py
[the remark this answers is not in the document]
JLGarber: Wow, that's a lot of helper functions! It definitely ends up more readable, and it made it easier to see where I could clean up the instance listing code. Maybe we don't need *that* many, but I think the abstraction makes sense in the line_start/line_stop functions.

JLGarber:
Just pushing the changes to-date for looking-over in case there are improvements to be made. I'm not expecting to get through trial compatibility or fixing timestamp validation tonight.

If anyone happens to have a 24+ hour log, I'd like to have a look at the output from that. Given that the original script doesn't have problems with this, I don't anticipate things breaking, but it would be nice to be sure.

JLGarber:
Alright, that ended up simpler than I thought, although it's a bit of a brute-force fix to validation. But realistically speaking, there's no reason to fail validation anymore unless the user actually enters timestamps that require it.

I'll put some more time this week into enabling trial parsing. Beyond that, I would appreciate any further suggestions for general improvements. If all goes well, I hope to submit a clean pull request at some point before the weekend.

util/make_timeline.py
```diff
@@ -97,14 +139,29 @@ def parse_file(args):
...
     started = False
-
+    encounter_sets = []
     with args.file as file:
+        # If searching for encounters, divert and find start/end first3
+        if args.search_fights:
+            encounter_sets = find_fights_in_file(file)
+            # If all we want to do is list encounters, stop here and give to the user.
+            if args.search_fights < 0:
+                return [f'{i + 1}. {e_info[0]} {e_info[1]} {e_info[2]}' for i, e_info in enumerate(encounter_sets)]
```
quisquous: Can you leave a comment somewhere (here or maybe line 156) that fight indices start at one?

util/make_timeline.py
```diff
@@ -97,14 +139,29 @@ def parse_file(args):
...
+                start_time = (encounter_sets[args.search_fights - 1][0])
+                end_time = (encounter_sets[args.search_fights - 1][1])
+            else:
+                start_time = args.start
+                end_time = args.end
+            if not started and start_time not in (line[14:26], line[14:29]):
                 continue
 
-            if line[14:26] == args.end:
+            if end_time in (line[14:26], line[14:29]):
```
quisquous: I'm not sure I understand what's going on with the 14:26 and 14:29, could you explain it?

util/make_timeline.py
```diff
@@ -134,9 +191,62 @@ def parse_file(args):
...
+            encounter_in_progress = False
+            continue
+
+        # We don't want to look for encounters outside a combat instance
+        if current_instance is None:
+            continue
+
+        # Build start/end time groupings
+        if is_line_start(line_fields, encounter_in_progress):
+            # This nested IF is messy, but it feels like the clearest way to express this
```
quisquous: I think this is reasonable!

util/make_timeline.py
```diff
@@ -134,9 +191,62 @@ def parse_file(args):
 
     return entries, last_ability_time
 
+def find_fights_in_file(file):
+    e_starts = []
+    encounter_start_staging = 0 # Staged to avoid spurious starts such as limit break usage
```
quisquous: Are `encounter_start_staging != 0` and `encounter_in_progress` redundant with each other? Just wondering if `encounter_in_progress` could get removed.

util/make_timeline.py
```diff
@@ -97,14 +139,29 @@ def parse_file(args):
 
     entries = []
     started = False
-
+    encounter_sets = []
     with args.file as file:
+        # If searching for encounters, divert and find start/end first3
```
quisquous: typo?

util/make_timeline.py
JLGarber: That's an artifact from before the elided microsecond issue was fixed. Definitely unnecessary now.

util/make_timeline.py
JLGarber: That definitely could be clearer. I don't think they're functionally identical, since `in_progress` is explicitly for binary state, while `staging` contains formatted data. 

An example of where this could be important is use of the limit break on trash between bosses. This would send a 0000 limit break line, triggering a fight start. This fight start would either show the name of the dungeon, not the boss zone, or the previous boss zone. The party would then get to the next boss, which would trigger a new fight start, since zone seals have priority over everything else. However, the spurious fight start would already have been placed in the `encounter_starts` list, meaning that we would have an imbalance between the start and stop lists. Having it staged means that the spurious fight gets discarded without any extra work on our part.

To put it another way, `in_progress` is used specifically in relation to parsing unknown data, while `staging` has to do with storage and presentation of already-known data. It's likely that we *could* change things around some to negate the need for two separate variables here, but that would then require us to do more manipulation of the `starts` list, and I'm not sure whether that's worth it.

I'm totally fine with combining it if there's a simpler way to discard trash pulls It's very possible that once I get things worked out with `15/16` line starts vs `21` ending messages, we could combine them. But as it stands, we do need the two separate.

util/make_timeline.py
JLGarber: Yes. Not sure why I would have been using the number pad for that, but 3 is directly above the period key there.

util/make_timeline.py
JLGarber: Perhaps something to the effect of "This is to make 1-based indices transparent to the user"? I don't like the hardcoded offsets, but I'm not sure whether there's a better solution that's a good user experience.

JLGarber:
Yes, "has begun" and "has ended" are in any situation where there's the starting circle. I used those because that's (what I thought was) the simplest way to ensure that we don't even bother to look for encounters outside a combat instance. Zone change messages might work, but since we can't rely on those being binary pairs, we could have overworld fights polluting the list.

(That being said, I need to test what happens to auto-parsing if the user logs out while in an instance. I put all this together on the assumption of no disconnection during the instance, but it's entirely possible I need to add the zone check anyway, depending on what the logout information says.)

In the same way, the 60-second timeout is to keep the returned fight list shorter. (I frequently have entered a zone-seal boss and then suddenly realized I need to back out before it actually seals for whatever reason. It's nice to have those situations automatically culled, rather than going to the ACT interface and deleting the encounter.) I can't think of a situation where the user *wants* to use a fight shorter than that as the basis for a timeline. That being said, the check is probably unnecessary in the strict sense.

JLGarber:
I thought about it a little more, and you were entirely correct, we can just use `15/16` lines. There's no need for all the complicated start/stop conditions, since only instances with zone seals have trash.  Any time we don't have trash, `15/16` lines are (or should be) 100% reliable indicators of a fight starting if it hasn't already. Any time there *is* trash, the zone seal message will ensure that the trash isn't included. Ensuring that we record encounters only while in instances guarantees no pollution from overworld enemies.

After redoing things a little this morning, I got this wonderful output:

```
make_timeline.py -f testlog2.log -lf
1. 00:57:37.000 00:59:56.000 The Beast
2. 01:00:27.000 01:01:53.000 The forward port ventilator
3. 01:02:37.000 01:04:52.000 The modification cell
4. 01:07:18.000 01:09:38.000 The conversion chamber
5. 01:10:13.000 01:13:00.682 The queen's bedchamber
6. 01:41:24.000 01:43:57.000 The Martial Court
7. 01:45:23.000 01:48:12.000 The Martial Court
8. 01:51:45.000 01:54:14.000 The Font of Quintessence
9. 01:00:03.000 01:03:22.000 The Chamber of Celestial Song
10. 01:05:30.000 01:08:43.000 The Chamber of Celestial Song
11. 01:10:35.000 01:14:25.000 The Chamber of Celestial Song
12. 16:38:00.125 16:42:35.169 Deltascape V4.0
13. 17:08:41.825 17:15:37.849 The Minstrel's Ballad: Hades's Elegy
14. 17:17:50.111 17:24:02.046 The Minstrel's Ballad: Hades's Elegy
15. 17:25:27.182 17:28:52.700 The Minstrel's Ballad: Hades's Elegy
16. 17:31:29.325 17:36:43.724 The Minstrel's Ballad: Hades's Elegy
17. 17:38:10.720 17:40:37.079 The Minstrel's Ballad: Hades's Elegy
18. 17:44:14.659 17:49:02.964 The Minstrel's Ballad: Hades's Elegy
19. 17:50:40.175 17:56:27.805 The Minstrel's Ballad: Hades's Elegy
20. 17:56:58.333 18:02:42.409 The Minstrel's Ballad: Hades's Elegy
21. 19:27:12.000 19:28:38.000 The 1st Chamber
22. 19:42:19.000 19:43:47.000 Condemnation
23. 20:14:42.000 20:15:59.000 The 1st Chamber
24. 20:23:59.000 20:25:09.000 The 1st Chamber
25. 20:40:56.000 20:42:00.000 The 1st Chamber
26. 21:19:23.000 21:20:26.000 The 4th Chamber
27. 21:21:20.000 21:22:44.000 Condemnation
28. 21:42:04.000 21:43:07.000 Condemnation
```
(That last set of encounters is from map farming. I'm thoroughly uninterested in trying to "fix" things to work for maps, although it might not be a bad idea to add known map zones to a default ignore list so they don't show up in the listed output. That wouldn't need updating more than once/twice per major patch.)

At this point, I think it's ready for release. I ran it against a number of logfiles from my archive and it seemed to perform nicely on all of them. Now that that's worked out, I'm confident we can get `test_timeline` functional in the same way within a week or two.

util/make_timeline.py
JLGarber: Well, the specific start line (0000 limit break) is gone now, but the reason for this remains true even after use of `15/16` log lines, so I think it's worth maintaining the two different variables here.

util/make_timeline.py
```diff
@@ -97,14 +135,30 @@ def parse_file(args):
...
+        if args.search_fights:
+            encounter_sets = find_fights_in_file(file)
+            # If all we want to do is list encounters, stop here and give to the user.
+            if args.search_fights < 0:
+                return [f'{i + 1}. {e_info[0]} {e_info[1]} {e_info[2]}' for i, e_info in enumerate(encounter_sets)]
```
panicstevenson: Do the items in encounter_set have more than 3 attributes? If not, you can do:
```python
" ".join(e_info)
```
instead of enumerating them. Or, we can make them their own class with separate attributes if you think that's cleaner, like Encounter with `start_time`, `end_time`, and `name`.

util/make_timeline.py
```diff
@@ -134,9 +188,62 @@ def parse_file(args):
...
+def find_fights_in_file(file):
+    e_starts = []
+    encounter_start_staging = 0 # Staged to avoid spurious starts such as limit break usage
+    e_ends = []
+    encounter_in_progress = False
+    current_instance = None
+    encounter_sets = []
+    for line in file:
+        # Ignore log lines that aren't game status info
+        if not (line[37:41] == '0839' or line[0:2] in ['00', '01', '21', '22', '33']):
```
panicstevenson: I think [De Morgan's laws](https://en.wikipedia.org/wiki/De_Morgan%27s_laws) finally come in handy!
I think
```python3
if line[37:41] != '0839' and line[0:2] not in ['00', '01', '21', '22', '33']:
```
should resolve slightly faster

panicstevenson:
Also, sorry if I'm super late to this party; is there a gain for listing the fights to be -1 instead of having a separate flag to list the fights? Just the ease of hitting up on the command line and subbing out the value?

JLGarber:
Good news on that last bit, I dropped the changes to date into `test_timeline`, and after a bit of fiddling to account for the slightly different program flow it seemed to work perfectly. Once we get `make_timeline` settled, I'll put together an additional pull request for that. Or, if you prefer, I could just fold it into this one.

util/make_timeline.py
JLGarber: That's a good point, I hadn't considered short-circuit evaluation. It's enough faster to be noticeable, thanks for pointing that out!

panicstevenson:
🙌 the ability to swap between make and test will be awesome

util/make_timeline.py
JLGarber: You're right, it's a guaranteed 3 elements. I appreciate this.

JLGarber:
I've had pipe dreams of consolidating them into one script, but I don't think that would really save the user much time, if any. Removing the need to hunt for timestamps should be enough of a time save there anyway, even if we do have to continue swapping between two terminal instances.

panicstevenson:
Just used this btw; very intuitive. I like it.

quisquous:
I think a second pull request would be good!

util/make_timeline.py
```diff
@@ -134,9 +188,62 @@ def parse_file(args):
...
+            current_instance = None
+            encounter_in_progress = False
+            continue
+
+        # We don't want to look for encounters outside a combat instance
+        if current_instance is None:
+            continue
+
+        # Build start/end time groupings
+        if is_line_start(line_fields, encounter_in_progress):
```
quisquous: I find bools passed to helper functions a little bit confusing.  What do you think about rewriting this block as:
```
if is_zone_seal(line_fields):
    encounter_start_staging = ...
    encounter_in_progress = True
    continue
elif not encounter_in_progress and is_line_attack(line_fields):
    encounter_start_staging = ...
    encounter_in_progress = True
    continue
```

and getting rid of is_line_start? In other words, hoist the logic out of is_line_start.  I think as a helper it maybe encapsulates too much logic and so is confusing, and it'd be better to be explicit at the callsite.  What do you think?

JLGarber: That's perfectly fair. Un-nesting that `if` is nice too.

quisquous:
\o/

PR #701, opened 2019-11-01
Raidboss: Ala Mhigo oopsy, timeline, and triggers

JLGarber:
This one was easy enough. I ended up doing some of the testing for the `make_timeline` changes on this. Magitek Scorpion is boring and a terrible boss.

Aulus Mal Asina is sort of interesting, in that the party members are cloned during the intermission, and everything then happens to the clones. It's fascinating how Square handles these things.

Zenos is just a well-designed battle overall. Each of the phases flows naturally into the next, and it's all put together so cleanly that I never noticed the divisions before deliberately looking for them. The action list within each block is also long enough to disguise the rotation once he is brought under 65% HP, especially combined with the randomizing of Storm and Swell. Square really outdid themselves with this one, and I wish I had appreciated it more at Stormblood's release.

### DIFF
--- a/ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
+++ b/ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
@@ -38,8 +38,8 @@
       id: 'Ala Mhigo Unmoving Troika',
       damageRegex: '2060',
       condition: function(e, data) {
-        // Non-tanks shouldn't be hit
-        return data.role != 'tank';
+        // Double taps only
+        return e.type != '15';
       },
       mistake: function(e, data) {
         return { type: 'warn', blame: e.targetName, text: e.abilityName };
--- a/ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
+++ b/ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
@@ -1,0 +1,80 @@
+'use strict';
+
+// Ala Mhigo
+[{
+  zoneRegex: /Ala Mhigo/,
+  damageWarn: {
+    'Ala Mhigo Magitek Ray': '24CE', // Line AoE, Legion Predator trash, before boss 1
+    'Ala Mhigo Lock On': '2047', // Homing circles, boss 1
+    'Ala Mhigo Tail Laser 1': '2049', // Frontal line AoE, boss 1
+    'Ala Mhigo Tail Laser 2': '204B', // Rear line AoE, boss 1
+    'Ala Mhigo Tail Laser 3': '204C', // Rear line AoE, boss 1
+    'Ala Mhigo Shoulder Cannon': '24D0', // Circle AoE, Legion Avenger trash, before boss 2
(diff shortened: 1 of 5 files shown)


ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
```diff
@@ -0,0 +1,80 @@
...
+      mistake: function(e, data) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
+      },
+    },
+    {
+      id: 'Ala Mhigo Unmoving Troika',
+      damageRegex: '2060',
+      condition: function(e, data) {
+        // Non-tanks shouldn't be hit
+        return data.role != 'tank';
```
quisquous: This doesn't work.  `data.role` is your role, not the person's role.  Given that party information is now an event when using ngld overlay, this could be hooked up in the future, but not now.

ui/raidboss/data/04-sb/dungeon/ala_mhigo.js
```diff
@@ -0,0 +1,206 @@
...
+      condition: function(data) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      id: 'Ala Mhigo Mana Burst',
+      regex: / 14:204F:Aulus [mM]al Asina starts using Mana Burst/,
```
quisquous: regexes are now case insensitive, so this isn't necessary anymore.  :)

ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
JLGarber: Rats, I knew I'd forgotten something. Okay. We have a couple choices on this one. I could just remove it entirely, although I don't like that option. I could comment it with a FIXME for when party information *is* available. (My preferred choice, although I can understand if you prefer not to merge commented code.) Or I could remove the tank condition and simply have all double taps be warnings. I don't like that one as much, but it's better than just removing it. What do you think?

ui/raidboss/data/04-sb/dungeon/ala_mhigo.js
JLGarber: Oh, did I miss that when it got merged? I don't remember seeing it specifically, but I might have missed it in the whirlwind panicstevenson submitted recently. Really awesome though, that's going to make things a LOT easier.

ui/oopsyraidsy/data/04-sb/dungeon/ala_mhigo.js
quisquous: I think having this be a double tap is probably correct even if we had party information.  Since if a tank is dead and it's on somebody else (and the tank gets rezzed?), the tank probably still shouldn't be hit by it, ideally? So, I think just warning on double taps seems the best.

JLGarber:
Well then, that should do it. Always good when the dungeons aren't difficult to put things together for.

quisquous:
Yeah, I'm glad this one was more straightforward!

PR #705, opened 2019-11-02
Fixed translations and trigger

Akurosia:
Added some missing translations for stormblood content + added timeline and trigger translation for hades-ex


quisquous:
Thanks for all your translation work.  <3

PR #713, opened 2019-11-03
raidboss: Add Tether Instructions for Grand Cosmos

panicstevenson:
Why do people have to burn the furniture?

### DIFF
--- a/ui/raidboss/data/05-shb/dungeon/the_grand_cosmos.js
+++ b/ui/raidboss/data/05-shb/dungeon/the_grand_cosmos.js
@@ -268,5 +268,25 @@
         fr: 'Droite',
       },
     },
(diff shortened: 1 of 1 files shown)


JLGarber:
It might be worth adding a special healer/ranged callout for the first Mortal Flames cast recommending they burn a curtain? If the curtains and piano + bench are the first things burned it's a lot easier to handle tethers since three sides of the room are available.

panicstevenson:
That sounds like a good idea. My only concern though is if there's two ranged and a healer. In that case, the trigger might be confusing when both curtains have burned and you're standing there with fire on your head.

JLGarber:
Good point. Something to return to later when we have full party member information available!

PR #716, opened 2019-11-04
raidboss: Lugus Timeline Updates

panicstevenson:
Apparently if you push Lugus, you can skip the first tankbuster. Also, it appears that if you fail the tether mechanic then you get hit with a different move, 475F, instead of 4760, which can mess up synchronization.

Also, you can skip a Sorching Left/Right if his HP is too low.


quisquous:
WHO KNEW

PR #719, opened 2019-11-05
Util: Test_timeline changes

JLGarber:
After all the work we put in on `make_timeline`, it really did drop in nicely here. I'll just go through its operation quickly so that this request can technically stand on its own. (This is in response to #646.)

This update adds an optional argument, `-search_fights` (aliased to `-lf`) that either takes no value or a positive integer as input. When called with this flag, the script will divert and build an encounter listing, and either return that list to the user or pull the selected encounter from the list and run the selected timeline against it.

Standard operation of the script is unaffected, so the user could (if they wanted to for whatever reason) list the fights and then enter the start/end times manually using `-start/-end`.

### DIFF
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -182,6 +182,46 @@ def parse_event_time(event):
     elif isinstance(event, dict):
         return event['time']
 
+def stringify_time(timestamp):
+    """
+    Parses a datetime object into a string for log comparisons.
+    Trims microseconds to milliseconds."""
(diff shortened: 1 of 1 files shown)


quisquous:
Is there a way to better share code between these two? Maybe like a timeline.py that both can import?

JLGarber:
That's not unreasonable. I'll look into it. I'm not sure what level of complexity I can currently handle with module production, but a simple helper module should be doable.

PR #720, opened 2019-11-05
Copied factory FR translations

mooondark:
There's a lot of names and action that contains ':', I hope it do not break everything.

### DIFF
--- a/ui/raidboss/data/05-shb/alliance/the_copied_factory.js
+++ b/ui/raidboss/data/05-shb/alliance/the_copied_factory.js
@@ -23,6 +23,7 @@
         return {
           en: 'Be Near Boss',
           de: 'sei in der Nähe des Bosses',
+          fr: 'Près du boss',
         };
       },
     },
@@ -33,6 +34,7 @@
       infoText: {
         en: 'Move to South Edge',
         de: 'zur südlichen Kante',
+        fr: 'Allez au Sud',
       },
     },
   ],
@@ -40,6 +42,7 @@
     {
       id: 'Copied Serial Forceful Impact',
       regex: / 14:48CF:Serial-Jointed Command Model starts using Forceful Impact/,
+      regexFr: / 14:48CF:Modèle multiarticulé : commandant starts using Forte Secousse/,
       condition: function(data) {
         return data.role == 'tank' || data.role == 'healer' || data.CanAddle();
       },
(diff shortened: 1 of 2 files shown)


quisquous:
Hmm, I merged this in but there appear to be different translations for some things (soldat vs commandant) in #722.  Can you take a look there and resolve what these should be?

Akurosia:
Here is 1 example where you can see that both the Comand-Model and Service model are translated the same as `Modèle multiarticulé : commandant`
```
      regex: / 14:48CF:Serial-Jointed Command Model starts using Forceful Impact/,
      regexFr: / 14:48CF:Modèle multiarticulé : commandant starts using Forte Secousse/,

      regex: / 14:48FA:Serial-Jointed Service Model starts using High-Caliber Laser/,
      regexFr: / 14:48FA:Modèle multiarticulé : commandant starts using Laser À Large Faisceau/,
```

Also from GameData the english line: `The wall-mounted right arm begins to move...` is translate to: `Le bras mural droit s'active !`. 
So i am not sure if the 3 dots in the end are correct here: https://github.com/quisquous/cactbot/pull/720/files#diff-f3d7c7562fdf339ef1d75715334ba3e3R172

mooondark:
I've done it manually with my log file, maybe there's a wrong copy/paste.
I've checked again and `Serial-Jointed Service` should be `Modèle multiarticulé : soldat` ... But I'm sure about that speels and actions are right.

I've no time right now, i'll have a look as soon as I can.
In the ReplaceSync section, in FR, text always begin with a capital letter. I don't know why the script always put a lower case.

Akurosia:
If you find the time, could you please upload the logfile so i could have a look whats not matching here as `The wall-mounted right arm begins to move...` is part of the GamesData Logmessage (i wile you could retriev by using SaintCoinach or view in Godbert.

The "issue" with capital letters can be ignored as regexes are now case insensitive. 
The lower case version btw is comming from games data while most capitalzation is done by the FFXIV_ACT plugin and this is why there was/is a lot of confusion with capitalized names or part of names that makes no sense (same issue is available in german for example)

Akurosia:
Thanks for the log.
Due to this i found an error in my converting script due to `</Indent>`
The line i could find was `00|2019-11-05T12:27:18.0000000+01:00|0839||Le bras mural droit s'active!|d88a577f7abf22f5dd4ece2148ec2a47
`

PR #722, opened 2019-11-05
add trigger translation for copied factory

Akurosia:
i forgot to include it to my last PR as it wasn't done when i started working on it... <.<


quisquous:
Would you mind fixing the conflicts? Sorry that I committed the other patch.  Maybe I should have started here, since this was the bigger patch.

Issue #322, opened 2019-01-12
Improve make_timeline.py


quisquous:
util/make_timeline.py is super useful for making timelines, but a few things that could be improved.

(1) delete abilities with the same name at the same time (or within <0.25 or some tolerance) to avoid spamming.
(2) if an ability has no name, label it --sync-- instead of blank
(3) handle overlapping syncs.  By default, everything without a window looks 2.5 seconds behind and 2.5 seconds ahead.  If there's an ability that triggers every second repeatedly (multiple explosions, etc), then the timeline will have multiple lines whose syncs overlap with each other.  I think in these cases, the syncs should be commented out and maybe a `#collision` comment added.

If anybody wants to contribute some small fixes, these sorts of things would be super appreciated! <3

quisquous:
* look up unknown abilities from xivapi
* have test_timeline.py find and auto-test against N fflogs fights so you don't have to specify them all individually
* have test_timeline.py auto-test against N fights and be able to print out average deviation of when things happen (and then be able to adjust the timeline to match them?)

quisquous:
* if a parameter is valid for make_timeline but not test_timeline (or vice versa) allow those parameters to silently (or with a warning?) be ignored, rather than not running at all (so that you can switch from make to test and vice versa without having to edit the command line)

JLGarber:
Any chance you would be willing to turn all this into a checklist? (Or, if you like, I guess I could open a new and more generic issue to cover all the timeline stuff.) I'll probably look at more of this in a month or so after I finish Heavensward's dungeons out. 

I think it might also be worth folding some of `check_translation`'s functionality into `translate_fight`, maybe expanding that a little in some way, and writing up some documentation on it (maybe similar to what I did for documenting trigger construction?) I know for me right now the translations are a bit of a pain because there's no built-in functionality to arrange/format them as I prefer, so there's a decent bit of (admittedly *technically* unnecessary) effort there.

And then I guess maybe a first-pass naive auto-generation of Oopsy files via a scan through the log for `E0000000::`?

quisquous:
- [ ] delete abilities with the same name at the same time (or within <0.25 or some tolerance) to avoid spamming.
[...]

PR #724, opened 2019-11-05
Util: Move common timeline functions into separate file

JLGarber:
Per #719, we have a new file containing most of the additional functionality that was introduced for the `make_timeline` updates. It's a fairly naive implementation as I'm not that familiar with Python's modularization potential, so I just went with what I could easily understand. I kept this solely as a function library rather than making it standalone, since there's *probably* not much reason to just list encounters in a file, and if the user actually does want that, the overhead from using that functionality within `make` or `test` is low enough to be negligible.

It's entirely possible we could move the fiddly bits hunting for the start/end timestamps into a helper function in `finder`, but that gets a little more complicated and would require passing state in as well.

### DIFF
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -440,12 +440,6 @@ def timeline_file(filename):
         return path.open()
 
 
-def timestamp_type(arg):
-    """Defines the timestamp input format"""
-    if arg and re.match(r'\d{2}:\d{2}:\d{2}\.\d{3}', arg) is None:
-        raise argparse.ArgumentTypeError("Invalid timestamp format. Use the format 12:34:56.789")
-    return arg
-
 
 if __name__ == "__main__":
     # Set up all of the arguments
@@ -471,8 +465,8 @@ if __name__ == "__main__":
     parser.add_argument('-rf', '--fight', type=int, help="Fight ID of the report to use. Defaults to longest in the report")
 
     # Log file arguments
-    parser.add_argument('-s', '--start', type=timestamp_type, help="Timestamp of the start, e.g. '12:34:56.789")
-    parser.add_argument('-e', '--end', type=timestamp_type, help="Timestamp of the end, e.g. '12:34:56.789")
+    parser.add_argument('-s', '--start', type=e_find.timestamp_type, help="Timestamp of the start, e.g. '12:34:56.789")
+    parser.add_argument('-e', '--end', type=e_find.timestamp_type, help="Timestamp of the end, e.g. '12:34:56.789")
     parser.add_argument('-lf', '--search_fights', nargs='?', const=-1, type=int, help="Encounter in log to use, e.g. '1'. If no number is specified, returns a list of encounters.")
 
     # Filtering arguments
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -5,6 +5,7 @@ import os
 import re
 
 import fflogs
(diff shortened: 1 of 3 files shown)


JLGarber:
It appears I somehow pulled in an older version of `make_timeline` from before it was merged for this branch. That's been resolved with this commit. I did end up splitting the start/end finding into the common functions, I'm not sure why I thought that would be a problem.

Also, slight commonality changes for terminology between `make` and `test`. (Both scripts involved a `parse_[THING]_ time` function, so I went with the broader version present in `test`, for instance. I'm sure there are others I missed, having those pointed out would be most welcome.

Finally, I added a step to fulfill one of the points from #322, which is to automatically make `Unknown` abilities `"--sync--"` entries instead. This might not be the correct pull request to do that, but I figured it was minor and simple enough that throwing it in wouldn't add too much extra time to the review process.

util/test_timeline.py
```diff
@@ -370,24 +347,36 @@ def run_file(args, timelist):
         'branch': 1,
         'timeline_stopped': True
     }
-    file_started = False
+    started = False
 
     with args.file as file:
+        # If searching for encounters, divert and find start/end first3
```
quisquous: :3

util/test_timeline.py
```diff
@@ -161,36 +162,13 @@ def parse_report(args):
 
     return entries, datetime.fromtimestamp((report_start_time + start_time) / 1000)
 
-
-def parse_time(timestamp):
-    """Parses a timestamp into a datetime object"""
-    return datetime.strptime(timestamp, '%Y-%m-%dT%H:%M:%S')
-
-
-def parse_event_time(event):
```
quisquous: Ah, this is nice cleanup too!

util/test_timeline.py
```diff
@@ -370,24 +347,36 @@ def run_file(args, timelist):
...
+        # If searching for encounters, divert and find start/end first3
+        if args.search_fights:
+            encounter_sets = e_find.find_fights_in_file(file)
+            # If all we want to do is list encounters, stop here and give to the user.
+            if args.search_fights == -1:
+                return [f'{i + 1}. {" ".join(e_info)}' for i, e_info in enumerate(encounter_sets)]
+            elif args.search_fights > len(encounter_sets) or args.search_fights < -1:
+                raise Exception('Selected fight index not in selected ACT log.')
+
+        start_time, end_time = e_find.choose_fight_times(args, encounter_sets)
```
quisquous: Like you do in `make_timeline`, I think you need to assign encounter_sets to be [] outside of the above conditional, or it might be unassigned here.

util/test_timeline.py
JLGarber: Gah. It keeps returning. An undead comment, of *all* things.

util/test_timeline.py
JLGarber: Do you think it's worth moving the timestamp definition as well? It is technically duplicated, but I feel a bit weird about having argument definitions defined outside the file.

quisquous: I think it'd be fine to move as they both really do share it.

quisquous:
Thanks!! These script changes are so great.  :)

JLGarber:
I'm glad it's something useful. It's been extremely good for me in terms of picking Python back up, so it's nice that's helpful as well.

PR #728, opened 2019-11-07
Raidboss: Temple of the Fist initial timelines and triggers

JLGarber:
Not much to say on this one, highly straightforward bosses.

### DIFF
--- a/ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
+++ b/ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
@@ -1,0 +1,36 @@
+'use strict';
+
+// Temple of the Fist
+[{
+  zoneRegex: /Temple Of The Fist/,
+  damageWarn: {
+    'Temple Fire Break': '21ED', // Conal AoE, Bloodglider Monk trash
+    'Temple Radial Blaster': '1FD3', // Circle AoE, boss 1
+    'Temple Wide Blaster': '1FD4', // Conal AoE, boss 1
+    'Temple Crippling Blow': '2016', // Line AoEs, environmental, before boss 2
+    'Temple Broken Earth': '236E', // Circle AoE, Singha trash
+    'Temple Shear': '1FDD', // Dual conal AoE, boss 2
+    'Temple Counter Parry': '1FE0', // Retaliation for incorrect direction after Killer Instinct, boss 2
+    'Temple Tapas': '', // Tracking circular ground AoEs, boss 2
+    'Temple Hellseal': '200F', // Red/Blue symbol failure, boss 2
+    'Temple Pure Will': '2017', // Circle AoE, Spirit Flame trash, before boss 3
+    'Temple Megablaster': '163', // Conal AoE, Coeurl Prana trash, before boss 3
+    'Temple Windburn': '1FE8', // Circle AoE, Twister wind, boss 3
+    'Temple Hurricane Kick': '1FE5', // 270-degree frontal AoE, boss 3
+    'Temple Silent Roar': '1FEB', // Frontal line AoE, boss 3
+    'Temple Mighty Blow': '1FEA', // Contact with coeurl head, boss 3
+  },
+  triggers: [
+    {
+      id: 'Temple Heat Lightning', // Purple spread circles, boss 1
+      damageRegex: '1FD7',
+      condition: function(e, data) {
+        // Double taps only
+        return e.type != '15';
+      },
+      mistake: function(e, data) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
+      },
+    },
+  ],
+}];
--- a/ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
+++ b/ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
@@ -24,11 +24,11 @@
     {
       id: 'Temple Heat Lightning', // Purple spread circles, boss 1
       damageRegex: '1FD7',
-      condition: function(e, data) {
+      condition: function(e) {
         // Double taps only
         return e.type != '15';
       },
-      mistake: function(e, data) {
+      mistake: function(e) {
         return { type: 'warn', blame: e.targetName, text: e.abilityName };
       },
     },
--- a/ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
+++ b/ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
@@ -3,7 +3,78 @@
 // The Temple of the Fist
 [{
   zoneRegex: /^The Temple Of The Fist$/,
+  timelineFile: 'temple_of_the_fist.txt',
+  timelineTriggers: [
+    {
+      id: 'Temple Pounce',
+      regex: /Pounce/,
+      beforeSeconds: 5,
+      condition: function(data) {
+        return data.role == 'healer' || data.role == 'tank';
+      },
+      infoText: {
+        en: 'Tank buster',
+      },
+    },
+    {
+      id: 'Temple Cardinal Shift',
+      regex: /Cardinal Shift/,
+      beforeSeconds: 5,
+      condition: function(data) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'AoE',
+      },
+    },
+  ],
   triggers: [
+    {
+      id: 'Temple Electric Burst',
+      regex: / 14:1FD6:Coeurl (?:Sruti|Smriti) starts using Electric Burst/,
+      condition: function(data) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'AoE',
+      },
+    },
+    {
+      id: 'Temple Fourfold Shear',
+      regex: / 14:1FD9:Arbuda starts using Fourfold Shear on (\y{Name})/,
+      infoText: function(data, matches) {
+        if (data.me == matches[1]) {
+          return {
+            en: 'Tank buster on YOU',
+          };
+        }
+        if (data.role == 'healer') {
+          return {
+            en: 'Tank buster on ' +data.shortName(matches[1]),
+          };
+        }
+      },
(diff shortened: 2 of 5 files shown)


ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
```diff
@@ -0,0 +1,36 @@
...
+    {
+      id: 'Temple Heat Lightning', // Purple spread circles, boss 1
+      damageRegex: '1FD7',
+      condition: function(e, data) {
+        // Double taps only
+        return e.type != '15';
+      },
+      mistake: function(e, data) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
+      },
```
panicstevenson: I don't think the `data` parameters are necessary here.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
```diff
@@ -50,5 +121,202 @@
...
+      },
+    },
+    {
+      id: 'Temple Touch Of Slaughter',
+      regex: / 14:1FE6:Ivon Coeurlfist starts using Touch Of Slaughter on (\y{Name})/,
+      alertText: function(data, matches) {
+        if (data.me == matches[1]) {
+          return {
+            en: 'Slaughter on YOU',
+          };
```
panicstevenson: If I remember correctly, there's nothing a player can do to avoid this mechanic. If that's the case, alerting the player seems like noise.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
```diff
@@ -50,5 +121,202 @@
...
+        } else if (data.role == 'healer') {
+          return {
+            en: 'Heal ' + data.shortName(matches[1]) + ' soon',
+          };
```
panicstevenson: While not a bad alert, I don't think we have anything like it currently. Does it make more sense for the move to resolve and _then_ alert? Or not alert at all, since the healer should have ample time to throw out a heal even if only reacting to mechanics as they happen?

Not that new triggers are bad, it just made me do a double-take. Feel free to ignore this one!

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
```diff
@@ -50,5 +121,202 @@
...
+        // Ivon summons invisible heads when he first appears. This avoids notifying.
+        if (data.firstHeads) {
+          return {
+            en: 'Avoid floating heads',
+          };
+        }
+      },
+      run: function(data) {
+        data.firstHeads = data.firstHeads || true;
+      },
```
panicstevenson: Does `data` get cleared upon wipe? This might have interesting side effects if not.

JLGarber: That's a good point, no, it doesn't. But, checking back through my logs, there's an instant cast that Ivon uses that *should* match this timing closely enough, so it might be possible to just avoid doing this altogether.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: There are ground effects at the time this goes out, and it's quickly followed by Hurricane Kick, the 270-degree cleave, so the target needs to be careful until the healer can get to them. Is it perhaps better to pre-alert the target for Hurricane Kick instead?

(I'm thinking of this as being sort of like delayed Hell Winds in E2s, where the target can't do anything about it directly, but *does* have to take different steps to resolve it.)

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: I'm not sure. The healer is guaranteed to be required to heal them, and we already pre-warn on tank busters and party AoE damage, so I feel like it's not terribly different in that sense. Obviously it's not a party-wipe level of danger like busters, but the logic feels the same to me.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
panicstevenson: I think Hell Winds are a great example actually; so, I agree with you that the healer probably should have something then that says to heal the slaughter target. However, with Hell Winds, the player has to explicitly change how they act in E2S, either by refusing to stack or not standing in the bleed pillar. Here I don't think the player does anything different than "avoid shit you'd usually avoid." so does that make the `on YOU` trigger valid still?

ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: It's possible, but this does follow conventions from elsewhere in the repository. (See [E2s](https://github.com/quisquous/cactbot/blob/26244749fac6e5d6d1ee187e423dc6b9f37ecce1/ui/oopsyraidsy/data/05-shb/raid/e2s.js#L30) or [Titania EX](https://github.com/quisquous/cactbot/blob/26244749fac6e5d6d1ee187e423dc6b9f37ecce1/ui/oopsyraidsy/data/05-shb/trial/titania-ex.js#L32) for examples of Quisquous' work matching this structure.) I'm not attached to it though, if it really is unnecessary I'm happy to leave it out.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: That is a fair point. Nothing does change for the target, just for the healer.

ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
panicstevenson: True; I only mention it because I started noticing the same, but I think they're remnants from copy/paste, since the functionality of just an event without data exists (see [O4S' oopsyraidsy entry](https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/data/04-sb/raid/o4s.js#L44-L46)).

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: Yes, this definitely is unnecessary. Coeurl Whisper is used invisibly to summon the heads, but I didn't even think to use it because it's instant-cast without much animation. Thanks for catching that!

ui/oopsyraidsy/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: I suppose that's true. I'll update that then.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
```diff
@@ -3,7 +3,78 @@
...
+      infoText: {
+        en: 'AoE',
+      },
+    },
+  ],
   triggers: [
+    {
+      id: 'Temple Electric Burst',
+      regex: / 14:1FD6:Coeurl (?:Sruti|Smriti) starts using Electric Burst/,
```
quisquous: For the purposes of making translations easier, can you make these two separate triggers?

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.txt
```diff
@@ -0,0 +1,137 @@
...
+
+98.1 "Wide Blaster"
+98.1 "Radial Blaster"
+104.2 "Pounce"
+112.3 "Pounce"
+123.9 "Wide Blaster"
+123.9 "Radial Blaster"
+
+# Phase change after one dies
+150.0 "Basic Instinct" sync /:Coeurl (?:Sruti|Smriti):1FD5:/ window 150,10
```
quisquous: Did `test_timeline` complain about the `?:`?

JLGarber: It did. I was just looking through that for another timeline. It's a *bit* of a long stack trace. Is it better to just use a capture there, or use an invisible second line, do you think?

quisquous: I think in the short term, better to use capture.  `test_timeline` could be updated to understand non-capturing groups in the future too.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
JLGarber: Sure. Is it possible that at some point we will get to a place where we can do direct translation replacements within the regex, such that we could do something like this?

`14:1FD6:(?:Coeurl Sruti|Coeurl Smriti) starts using Electric Burst`

I'll change it for now, of course, but it seems a shame to need two triggers for an identical mechanic.

quisquous: No, that's fair.  I think your proposed replacement will work with the translation scripts.  It's really about keeping the mob name together.

JLGarber: Okay. I'll keep that in mind for future work.

ui/raidboss/data/04-sb/dungeon/temple_of_the_fist.js
panicstevenson: Update: Data is cleared upon re-entering combat (as defined by the game's definition, not ACT's), see [oopsyraidsy's README](https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/data/README.md#data-fields).

JLGarber: Well then. Good to know that it *would* have worked, but it's just as well we went with something else in this situation, alerting based on ability use is *probably* better than alerting on added combatants.

PR #737, opened 2019-11-09
Oopsy: Minor fixes

JLGarber:
Or maybe that should be "extremely urgent". Who knows? Don't even look at Gubal Hard. It's awful. Just push the button on that one.

(More seriously, `console.log` might be worth adding to the test suite as a warning, even as blatantly obvious as it should be.)

I happened to be looking through the Oopsy folder for something else and noticed a couple of points where I had unthinkingly used `role == 'tank'` as a condition. As long as I was going to fix that, I went through all my Oopsy contributions and removed the unused `data` parameter from the triggers.

### DIFF
--- a/ui/oopsyraidsy/data/03-hw/dungeon/fractal_continuum.js
+++ b/ui/oopsyraidsy/data/03-hw/dungeon/fractal_continuum.js
@@ -17,10 +17,11 @@
     {
       id: 'Fractal Sanctification', // Instant conal buster, boss 3
       damageRegex: 'F89',
-      condition: function(data) {
-        return data.role != 'tank';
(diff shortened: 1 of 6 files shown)


quisquous:
Thanks for the followup.  Sorry for not catching these earlier.

I think we could warn on console.log, although sometimes it is useful.  For a while I was using the ucob dragon marking console output in case things got hectic and I missed it while it was on screen.

JLGarber:
I suppose that's true. Bahamut Ultimate was the only other non-utility file I saw it in, and of course it makes sense there. I'm just really surprised I forgot to remove that before the initial pull request.

PR #739, opened 2019-11-09
radar: init radar

Bluefissure:
Add the rotation address to dll and thus we can get an arrow to indicate the facing-based direction of mobs:

I'm sorry that I didn't follow the OOP when implementing the radar module, but we can just modify it to an OOP version afterward.


Bluefissure:
It seems the library of `jQueryRotate.js` failed in eslint check, do maybe I can move it so somewhere else where eslint doesn't check?
Or if there's any way to do the rotate without third library rotate... I'm just so naive at javascript.

JLGarber:
Do you have ESLint set up in your development environment? I went through pretty much the exact same thing when I started on contributing here, and after figuring out the setup it's been a lot easier to avoid lintfix commits.

That being said, I'm not sure what the views on third-party library code here are. The license shouldn't be a conflict, but I have no idea what sort of overhead the library would incur.

Bluefissure:
> Do you have ESLint set up in your development environment? I went through pretty much the exact same thing when I started on contributing here, and after figuring out the setup it's been a lot easier to avoid lintfix commits.
> 
> That being said, I'm not sure what the views on third-party library code here are. The license shouldn't be a conflict, but I have no idea what sort of overhead the library would incur.

Yeah, I've set up the eslint dev environment and fixed the problems of my code, just not sure how to fix the libraries lint problems.

JLGarber:
Oh. That's generally a lot easier then. Just open a terminal in the folder containing the library and run `eslint --fix [filename]`. It didn't look like there were any problems that couldn't be auto-fixed.

Your editor might have an ESLint integration, too, if you haven't done that already.

Bluefissure:
> Oh. That's generally a lot easier then. Just open a terminal in the folder containing the library and run `eslint --fix [filename]`. It didn't look like there were any problems that couldn't be auto-fixed.
> 
> Your editor might have an ESLint integration, too, if you haven't done that already.

I've tried, the auto-fix will ruin the functions and it cannot be used.

JLGarber:
Ah, I see. Poking through it in my own editor, things make more sense, it's definitely challenging. Not really sure what can be done about it, unfortunately, that's where my knowledge runs out too.

ui/radar/radar.js
```diff
@@ -0,0 +1,167 @@
...
+      img.setAttribute('class', 'radar-image-40');
+      th.appendChild(img);
+      th.setAttribute('style', 'max-width: 100px');
+      tr.appendChild(th);
+      th = document.createElement('th');
+      let text = document.createElement('div');
+      text.innerHTML = monster['name'];
+      th.appendChild(text);
+      tr.appendChild(th);
+      $('#radar-table').prepend(tr);
```
quisquous: I think I would rather not have the jquery dependency here.  (I know it's used in one dps file, but I'd like to keep it there or remove it.)  That would solve the eslint problem, although maybe not in the way you hoped.

It looks like you're only using it for this `prepend`, and the `rotate` below? You can use the built in `insertBefore` to do the equivalent of `prepend`.  You should also be able to add a css style for `transform: rotate` to do the rotation, I think.

ui/radar/radar.js
```diff
@@ -0,0 +1,167 @@
...
+  //   let Tdelta = Math.acos(theta(Vdelta, new Point2D(1, 0)));
+  //   if(Vdelta.y < 0)
+  //     Tdelta = -Tdelta
+  //   Tdelta += Math.PI - e.detail.rotation;
+  //   let t = -Tdelta*180/Math.PI;
+  //   $('#test-demo').rotate(t);
+  // }
+});
+
+document.addEventListener('onZoneChangedEvent', function(e) {
```
quisquous: Can you add a `UserConfig.getUserConfigLocation('radar', etc....` bit here, like there is elsewhere in the code? That will let users override this with `cactbot/user/radar.js` or `.css` files.

ui/radar/monsters.js
```diff
@@ -0,0 +1,787 @@
+'use strict';
+
+let Monsters = {
+  'Vogaal Ja': {
+    'name': {
+      'en': 'Vogaal Ja',
+      'cn': '丑男子 沃迦加',
+      'ja': '醜男のヴォガージャ',
+    },
+    'hp': 100000,
```
quisquous: It seems like everything has this hp value.  Would it make more sense to have a default minimum hp in code somewhere rather than specifying it here?

ui/radar/monsters.js
```diff
@@ -0,0 +1,787 @@
+'use strict';
+
+let Monsters = {
```
quisquous: Where did this list come from? It looks like the NotoriousMonster table in SaintCoinach (although xivapi doesn't like using this table) has all of this information in it, both names and rank.

Would it be possible to write a script like the weather or fishing script to automatically generate some `resources/hunts.js` file so I don't have to manually keep this up to date myself?

ui/radar/radar.js
Bluefissure: Ok I removed all the jquery libs.

ui/radar/radar.js
Bluefissure: Actually I'm not sure how the `UserConfig` works, I add a `UserConfig.getUserConfigLocation` to the end of `radar.js` but not sure whether this will overwrite files.

ui/radar/monsters.js
Bluefissure: No they have different hp values, we use this to determine whether a player can be detected if he has the same name as the hunt mob (which only happens in cn server since we don't have first/last names?).
I'm just lazy to dig out the true value of hp so I just add some lower bound of mob hp. For 2.0, it's 100000 for the A and 10000000 for the S and for 3.0+, 10000000 for the S and A.
It's kinda difficult to dig out all those HPs, maybe NotoriousMonster can help, as you said below.

ui/radar/monsters.js
Bluefissure: This resource will only be updated when New Expansion releases (I think), so it's won't be frequently updated maybe?
But sure, an auto-update script is much more convenient, I'll look into this later.

ui/radar/radar.js
quisquous: Calling getUserConfigLocation(radar) automatically includes `radar.css` and `radar.js` from the cactbot/user folder.  If there are things you don't want to set up or start until the full options have been set (like maybe the user wants to override something), most cactbot overlays do all of their setup (like adding event listeners, etc) inside the callback that it gets.  Pullcounter is a pretty straightforward example if you want to look at that.

You can try this out yourself by creating a `cactbot/user/radar.js` file and adding a `console.log('custom radar user file');` line in it.

I'm not sure what you mean about overwriting files.

ui/radar/monsters.js
quisquous: I think you could use the ids of the monsters (which always start with `4` instead of `1` for players) to differentiate whether a mob is a player or not, rather than using hp.

ui/radar/monsters.js
quisquous: I think mostly my concern here is that I am happy to put extremely useful modules like this into cactbot proper that a lot of people would find useful and keep it up to date as things change.  However, everything that gets added is also extra friction for me (see also not enough time to fix all the fishing module issues), and so it's really important for me that anything that gets added can be kept up to date as automatically as possible with little effort as new content is added.

An auto-update script would also remove the need to manually translate this into french and german as well.

Bluefissure:
Sorry for not catching up, got several dues lately and this is just an initial version of radar, I'll complete it after several days this week.

ui/radar/monsters.js
Bluefissure: I think Chocobo also starts with `4`, just like the monsters. And Chinese players can name their Chocobo names to hunt mobs lol.

ui/radar/radar.js
Bluefissure: Okay I add the user config file as `user/radar-example.js`

ui/radar/monsters.js
Bluefissure: You can use `util/get_radar_data.py` to auto-update now ~

Bluefissure:
Seems all good, if you have any further questions just let me know >,<

quisquous:
(Sorry, I'm out of town for a few days, but will try to review this next week when I have more time.)

util/get_radar_data.py
```diff
@@ -0,0 +1,76 @@
...
+    for locale in languages:
+        data = reader.exd('NotoriousMonster', lang=locale)
+        data = parse_data(data, lang=locale)
+        for m in data.keys():
+            monsters[m]['name'].update(data[m]['name'])
+            monsters[m]['rank'] = data[m]['rank']
+    return monsters
+
+
+CN_FFXIV_PATH = 'C:\Games\SDO\FFXIV'
```
quisquous: Do I need the chinese version to get the chinese names from Saint Coinach?

Bluefissure: You don't need to modify SaintCoinach to support Chinese client, it works well to export the files.
But yes, you need a Chinese version FFXIV client and use SaintCoinach to export.
We also have a git repo to do the data mining of CN client [ffxiv-datamining-cn](https://github.com/thewakingsands/ffxiv-datamining-cn).

quisquous: Hmmmmmmmmm.  How hard would it be to use xivapi and the chinese version of that instead of requiring keeping a second ffxiv client up to date on my computer?

Bluefissure: As you said before, xivapi has not yet hosted this csv file?
Or do I misunderstand what you mean?

quisquous: Maybe we're misunderstanding each other.  I was asking how hard it would be to use xivapi instead of saint coinach here.  That way updating the radar wouldn't involve installing multiple versions of the client.

It looks like it has these values: https://xivapi.com/NotoriousMonster/22?pretty=true

Bluefissure: lmao I didn’t use xivapi much so I thought your “xivapi doesn’t like to use it” means this csv is not hosted in xivapi.
I’ll change the script to a xivapi version tomorrow.

quisquous:
Sorry for a million comments.  To be super clear, I think this is a great overlay idea and I think a lot of people will be really happy to get their hands on this.  I just want to clean up the javascript a little bit and make sure I understand everything that's going on before this gets merged.  :)

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -148,6 +148,7 @@ public enum EntityJob {
     //   0xA0 bytes in: float32 pos_x;
     //   0xA4 bytes in: float32 pos_z;
     //   0xA8 bytes in: float32 pos_y;
+    //   0xB0 bytes in: float32 Rotation;
```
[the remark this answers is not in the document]
Bluefissure: It’s okay, I can fix the conflicts afterwards since radar just need one additional memory address.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+  return dot(a, b)/(length(a) * length(b));
+}
+
+function calc_damage(x) {
+  if (x.length < 8)
+    x = new Array(9 - x.length).join('0') + x;
+  x = '0x' + x;
+  let y = Number(x);
+  if (y & 0x00004000) {
+    // lot of damage
```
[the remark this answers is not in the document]
Bluefissure: Well actually I’m not updating the positions but the current hp of the detected hunt mobs.
I tried several methods to remove the mobs when they’re killed, but not moving farther (both of which will generate a “Removing combatant” log), and I finally decided to detect the current hp of mobs, only remove them when hp reaches 0 (or so).

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+    this.table = element;
+    this.options = Options;
+    this.monsters = Object.assign({}, gMonster, Options.CustomMonsters);
+  }
+
+  OnLogEvent(e) {
+    let lang = this.options.Language;
+    for (let i = 0; i < e.detail.logs.length; i++) {
+      // add new combatant
+      let r = e.detail.logs[i].match(/03:(........):Added new combatant (.*)\. (?:.*) Max HP: (\d+) (?:.*) Pos: \((-?\d+.\d+),(-?\d+.\d+),(-?\d+.\d+)\)/);
```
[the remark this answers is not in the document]
Bluefissure: Wow, I didn’t know the regex helpers before, I’ll take a look.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+      if (r) {
+        for (let name in this.monsters) {
```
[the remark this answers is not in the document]
Bluefissure: Sure, I didn’t focus on the efficiency of code when writing the first version cuz there’re only few testing mobs. But yes, I should make it more efficient by a hashmap/dictionary.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+          let r_id = r[1];
+          let r_name = r[2];
+          let r_hp = r[3];
+          let r_pos_x = r[4];
+          let r_pos_y = r[5];
+          let r_pos_z = r[6];
+          let monster = this.monsters[name];
+          let matchOrNot = (r_name.match(monster['name'][lang] || monster['name']['en']) != null);
+          matchOrNot &= (r[0].match(monster['regex'] || '') != null);
+          matchOrNot &= (Number(r_hp) >= (monster['hp'] || 0));
```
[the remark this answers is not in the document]
Bluefissure: Chinese players may want to use the minimum hp to filter out chocobos, which may have the same name as hunt mobs and also starts with 4 in the ObjectID of log.
I’m not sure whether this will happen in other servers so I didn’t remove the hp filter, but that’s compatible with current hunt resources (without any hp info) I think.
monster[‘regex’] is another leftover filter setting for user to custom, maybe I should mention it in the comment lol.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+          if (monster['rank'] in options.RankOptions)
+            options = Object.assign({}, this.options, options.RankOptions[monster['rank']]);
+          if (options.OnlyMobs)
+            matchOrNot &= r_id.startsWith('4');
+          if (matchOrNot) {
+            let mob_name = r_name;
+            let r2 = r_name.match(/(.*?)\((.*?)\)/); // character_name(world)
+            if (r2)
+              mob_name = r2[1];
+            let m = {
```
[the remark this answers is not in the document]
Bluefissure: Of cause, I did code a lot of duplicated code fragments, sorry.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+              this.targetMonsters.push(m);
+              if (options.TTS) {
+                OverlayPluginApi.overlayMessage(OverlayPluginApi.overlayName, JSON.stringify({ 'say': m['rank'] + ' ' + m['name'] }));
+              } else if (options.PopSoundAlert && options.PopSound && options.PopVolume) {
+                let audio = new Audio(options.PopSound);
+                audio.volume = options.PopVolume;
+                audio.play();
+              }
+            } else {
+              if (options.UpdateMonsterPos) {
```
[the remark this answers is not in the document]
Bluefissure: If user set SplitSameMonster to False, do we need to update mob positions with the same name when handling an “Adding combatant” log?
Maybe no? I’m confused here too.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+        let text = document.createElement('div');
+        text.innerHTML = monster['name'];
+        th.appendChild(text);
+        tr.appendChild(th);
+        this.table.insertBefore(tr, this.table.childNodes[0]);
+        monster['dom'] = tr;
+      } else {
+        tr = monster['dom'];
+      }
+      // caculate rotate
```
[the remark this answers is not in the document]
Bluefissure: Lol, will fix this afterwards.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+      r = e.detail.logs[i].match(instanceChangedRegex[lang] || instanceChangedRegex['en']);
+      if (r)
+        this.ClearTargetMonsters();
+      // removing new combatant
+      r = e.detail.logs[i].match(/04:(.{8}):Removing combatant (.*)\. (?:.*) Max HP: (\d+)\./);
+      if (r) {
+        for (let i = 0; i < this.targetMonsters.length; i++) {
+          let monster = this.targetMonsters[i];
+          if (monster['dom'] && r[1].match(monster['id']) &&
+                Date.now() / 1000 <= monster['battle_time'] + 60) {
```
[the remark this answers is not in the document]
Bluefissure: We also remove mobs when player get out of battle status if the monster
1. has a current hp less than a threshold 
2. was attacked in the past several minutes 
And battle_time is for recording the last battle time of the mob.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+            break;
+          }
+        }
+      }
+    }
+  }
+
+  OnPlayerChange(e) {
+    this.playerPos['x'] = e.detail.pos.x;
+    this.playerPos['y'] = -e.detail.pos.y;
```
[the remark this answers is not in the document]
Bluefissure: I think all y value are negated from e.detail.pos, because the y axis is upside down in the map, so, in order to doing math in a formal (high school mathematical) way, I just negated the y value.

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+      let monster = this.targetMonsters[i];
+      if (monster['dom'] && monster['current_hp'] / monster['hp'] * 100 < min_hp) {
+        monster['dom'].remove();
+        this.targetMonsters.splice(i, 1);
+      }
+    }
+  }
+}
+
+document.addEventListener('onLogEvent', function(e) {
```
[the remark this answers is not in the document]
Bluefissure: Okay, actually I haven’t read the ngld branch but I definitely need to, since it’ll become the default branch lol.

user/radar-example.js
```diff
@@ -0,0 +1,58 @@
...
+// Rename this file to `radar.js` and edit it to change the radar ui.
+// This file is Javascript.  Anything after "//" on a line is a comment.
+// If you edit this file, remember to reload ACT or click the "Reload overlay"
+// button on the raidboss CactbotOverlay.
+// If there are errors in this file, they will appear in the OverlayPlugin.dll
+// log window in ACT.
+
+
+// If true, monsters with the same names will be treated as separate ones.
+Options.SplitSameMonster = false;
```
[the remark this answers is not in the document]
Bluefissure: Well some little hunt mobs may have the same name, if you read all the hunt quests you may find some junior hunt quests have normal mobs, which are not S/A/B rank, just normal mobs.
However I don’t think ppl will use this to track them.
But what if there is? LOL

ui/radar/radar.js
```diff
@@ -0,0 +1,281 @@
...
+        let r_target_id = r[5];
+        let r_target_name = r[6];
+        let r_damage = r[8];
+        let damage = calc_damage(r_damage);
+        let r_target_hp = r[23];
+        for (let i = 0; i < this.targetMonsters.length; i++) {
+          let monster = this.targetMonsters[i];
+          if (monster['id'] == r_target_id) {
+            monster['battle_time'] = Date.now() / 1000;
+            monster['current_hp'] = Math.max(Number(r_target_hp) - damage, 0);
```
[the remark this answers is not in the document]
Bluefissure: No, just because I’m not familiar with js, sorry _(:з」∠)_

Bluefissure:
(Sorry for my recent busy schedule, got several sudden deadlines from my class and boss. Maybe I can update the code in thanksgiving holiday.)

Bluefissure:
Close this PR cuz it has tons of conflicts with the current master branch.

PR #740, opened 2019-11-09
Add canned regex helpers

### LINKED ISSUE #543: Simplifying Triggers for Locale Regex
I was wondering what the overall feeling would be to do away with matching the locale regex for logs and instead matching on broader regex. For example:

```
14:3FF1:Leviathan starts using Tidal Wave:
14:3FF1:Leviathan starts using Tidal Wave on person:143918:1092
14:3FF1:利维亚桑 starts using Tidal Wave:
14:3FF1:Leviathan starts using Flutwelle:
14:3FF1:Léviathan starts using Raz-De-Marée:
14:3FF1:リヴァイアサン starts using タイダルウェーブ:
```

can all be matched with `14:3FF1:.+? starts using .+?:` and can be extended to include `(on \y{Name})` as well, if we wanted. Mechanic names and boss names can similarly be substituted out if we wanted that functionality as well.

This, combined with a rudimentary dictionary of common Cactbot phrases plus their translation counterparts (I'm envisioning copy/paste, but it could ultimately live in a map within the code as well), could probably reduce the translation time and effort necessary for fights. An example would be standardizing all raidwide damage to be "aoe" and its non-English equivalents, etc.

The biggest drawback I immediately see is that the ids of the triggers themselves would have to be consistently good, since the mechanic name is no longer within the trigger regex itself. Additionally, anyone looking for `タイダルウェーブ` wouldn't find it since it's now just `.+?` and I'm expecting that the trigger ids will never be non-English just due to tradition at this point.

quisquous:
Adds a number of regex helper functions that all take a dictionary of optional fields.  Fields can be filled in optionally with strings/regular expressions/arrays of multiple values.  If field values are unspecified, it will match anything.

Each dictionary of fields can also specify `capture` as a boolean variable.  If not specified, capture is true.  If capture is true, then everything in the regex will be captured as a named group (with the same name as the field that would have specified it).

Examples:
* `Regexes.losesEffect({effect: '(?:Doom|Forked Lightning)'})`
* `Regexes.startsUsing({id: ['1C7F', '1C90'], name: 'Deathgaze Hollow', ability: 'Void Death'})`
* `Regexes.ability({name: 'Caduceus', id: '4B8', ability: 'Hood Swing'})`

Fixes #543.

### DIFF
--- a/resources/regexes.js
+++ b/resources/regexes.js
@@ -31,7 +31,7 @@ var Regexes = {
   },
 
   // fields: name, id, ability, capture
-  ability(f) {
+  ability: (f) => {
     let capture = trueIfUndefined(f.capture);
     let str = '\\y{TimeStamp} 1[56]:\\y{ObjectId}:' +
       Regexes.maybeCapture(capture, 'name', f.name, '.*?') + ':';
@@ -46,7 +46,7 @@ var Regexes = {
   },
 
   // fields: name, id, ability, target, flags, x, y, z, heading, capture
-  abilityFull(f) {
+  abilityFull: (f) => {
     let capture = trueIfUndefined(f.capture);
     let str = '\\y{TimeStamp} 1[56]:\\y{ObjectId}:' +
       Regexes.maybeCapture(capture, 'name', f.name, '.*?') + ':' +
@@ -219,7 +219,7 @@ var Regexes = {
     return str;
   },
 
-  parse: function(regexpString) {
+  parse: (regexpString) => {
     let kCactbotCategories = {
       TimeStamp: '^.{14}',
       LogType: '[0-9A-Fa-f]{2}',
--- a/resources/regexes.js
+++ b/resources/regexes.js
@@ -1,10 +1,202 @@
 'use strict';
 
+let trueIfUndefined = (value) => {
+  if (typeof(value) === 'undefined')
+    return true;
+  return !!value;
+};
+
 // Node loading shenanigans.  'var' lets other files require() this file inside of
 // Node and put Regexes as a global without conflicting when redefining.
 /* eslint-disable no-var */
 var Regexes = {
 /* eslint-enable */
+
+  // fields: name, id, ability, target, capture
+  startsUsing: (f) => {
+    let capture = trueIfUndefined(f.capture);
+    let str = '\\y{TimeStamp} 14:' +
+      Regexes.maybeCapture(capture, 'id', f.id, '\\y{AbilityCode}') + ':';
+
+    if (f.name || f.id || f.target || capture)
+      str += Regexes.maybeCapture(capture, 'name', f.name, '.*?') + ' starts using ';
+
+    if (f.ability || f.target || capture)
+      str += Regexes.maybeCapture(capture, 'ability', f.ability, '.*?') + ' on ';
+
+    if (f.target || capture)
+      str += Regexes.maybeCapture(capture, 'target', f.target, '.*?') + '\\.';
+
+    return Regexes.parse(str);
+  },
+
+  // fields: name, id, ability, capture
+  ability(f) {
+    let capture = trueIfUndefined(f.capture);
+    let str = '\\y{TimeStamp} 1[56]:\\y{ObjectId}:' +
+      Regexes.maybeCapture(capture, 'name', f.name, '.*?') + ':';
+
+    if (f.id || f.ability || capture)
+      str += Regexes.maybeCapture(capture, 'id', f.id, '\\y{AbilityCode}') + ':';
+
+    if (f.ability || capture)
+      str += Regexes.maybeCapture(capture, 'ability', f.ability, '.*?') + ':';
+
+    return Regexes.parse(str);
+  },
+
+  // fields: name, id, ability, target, flags, x, y, z, heading, capture
+  abilityFull(f) {
+    let capture = trueIfUndefined(f.capture);
+    let str = '\\y{TimeStamp} 1[56]:\\y{ObjectId}:' +
+      Regexes.maybeCapture(capture, 'name', f.name, '.*?') + ':' +
+      Regexes.maybeCapture(capture, 'id', f.id, '\\y{AbilityCode}') + ':' +
+      Regexes.maybeCapture(capture, 'ability', f.ability, '.*?') + ':\\y{ObjectId}:' +
+      Regexes.maybeCapture(capture, 'target', f.target, '.*?') + ':' +
+      Regexes.maybeCapture(capture, 'flags', f.flags, '.*?') + ':' +
+      '.*:' +
+      Regexes.maybeCapture(capture, 'x', f.x, '\\y{Float}') + ':' +
+      Regexes.maybeCapture(capture, 'y', f.y, '\\y{Float}') + ':' +
+      Regexes.maybeCapture(capture, 'z', f.z, '\\y{Float}') + ':' +
+      Regexes.maybeCapture(capture, 'heading', f.heading, '\\y{Float}') +
+      ':.*?:?$';
+    return Regexes.parse(str);
+  },
+
+  // fields: name, id, capture
+  headmarker: (f) => {
+    let capture = trueIfUndefined(f.capture);
+    let str = '\\y{TimeStamp} 1B:\\y{ObjectId}:' +
+      Regexes.maybeCapture(capture, 'name', f.name, '.*?') + ':....:....:' +
+      Regexes.maybeCapture(capture, 'id', f.id, '....') + ':';
+    return Regexes.parse(str);
+  },
+
+  // fields: name, capture
+  addedCombatant: (f) => {
+    let capture = trueIfUndefined(f.capture);
+    let str = '\\y{TimeStamp} 03:\\y{ObjectId}:Added new combatant ' +
+      Regexes.maybeCapture(capture, 'name', f.name, '.*?') + '\\.';
+    return Regexes.parse(str);
+  },
+
+  // fields: id, name, hp, x, y, z, capture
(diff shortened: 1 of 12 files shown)


ui/raidboss/popup-text.js
```diff
@@ -256,6 +256,11 @@ class PopupText {
     if ('disabled' in trigger && trigger.disabled)
       return;
 
+    // If using named groups, treat matches.groups as matches
+    // so triggers can do things like matches.targetName.
```
panicstevenson: Insanely minor, but the o5n example has `matches.target` not `targetName`

resources/regexes.js
```diff
@@ -1,10 +1,202 @@
...
+    let str = '\\y{TimeStamp} 1E:\\y{ObjectId}:' +
+      Regexes.maybeCapture(capture, 'name', f.name, '.*?') +
+      ' loses the effect of ' +
+      Regexes.maybeCapture(capture, 'effect', f.effect, '.*?') +
+      ' from ' +
+      Regexes.maybeCapture(capture, 'source', f.source, '.*?') + '\\.';
+    return Regexes.parse(str);
+  },
+
+  // fields: name1, name2, id, capture
```
panicstevenson: Should these be source, target? I don't know how these log lines usually work.

quisquous: I wish!  My experience is that these are a bit arbitrary who is the source and who is the target of a tether.  Sometimes it's boss->player and sometimes player->boss.  That's why I went with `name1` and `name2` even though that's inconsistent with everything else.  If you had alternative name suggestions, I'd love to hear them, because I'm not happy with this either.

ui/raidboss/popup-text.js
quisquous: Thanks! Insanely minor feedback still appreciated.  :)

quisquous:
> I have two nits, but they can be disregarded or argued: `s/TimeStamp/Timestamp/` and `s/headmarker/headMarker/`.

I'm amused that you want one camelcase and the other not.  I don't have strong feelings here.

> I have one other actual thought, which is about the user experience on the other end.
> 
> `startsUsing` is name (boss), target (player)
> `abilityFull` is name (boss), target (player)
> `headmarker` is name (player)
> `gainsEffect` is name (player), source (boss)

I think this is not the full picture.

`startsUsing` is name (boss, player), target (boss, player)
`abilityFull` is name (boss, player), target (boss, player)
`headmarker` is name (boss, player) // see o12s
`gainsEffect` is name (boss, player), source (boss, player)

I do agree that there is some inconsistency here in that `gainsEffect` has `source`+`name` and `ability` has `name`+`target`.  `name` is the target in one but the attacker/source in the other, which could be confusing.

My thinking here is that these ability lines are verbs and there are subjects and objects.  `name` is always the subject (who is gaining the effect, who is using the ability, who headmarker <_< ok maybe not that one) and then the object (who is the effect from, who is the ability on) gets named separately. 

I don't know if that explanation helps make things more consistent? Or if there's some better way to write these?

> `tether` is name1, name2

This is its own issue, which I left as a comment above.

> I don't know that it's a huge ordeal when creating the regexes, but I think on consuming them it's a little awkward if we can smooth out that language.

Yeah, agreed.  I think this point in particular (bikeshedding the parameter names) is why I put this up as a pull request to get some feedback on it.  These originally were much longer, e.g. `targetName` but I felt like there was some value in succinctness and dropped to the shorter values that are there now.  Maybe I went too far?

resources/regexes.js
panicstevenson: That's frustrating; does it change between uses or just random quirks between bosses/encounters? It might still make sense to have them labeled as source and target, even if the wording is weird and can be denoted via comments?

I guess my fear here is that these will, regardless of name, end up victims of copy/paste and will still not make sense. For an example, imagine that common practice would follow that the player is always name2 (I don't know if this is the actual case), but then somewhere in coil it wasn't standardized that early (or something) and the boss is actually name2. It's still confusing, but harder (for me) to parse?

```
if (matches.source == data.me) {
   // WE ARE ACTUALLY SOURCE THIS TIME
  doThing();
}
```
vs
```
if (matches.name1 == data.me) {
   // WE ARE ACTUALLY NAME1 THIS TIME
  doThing();
}
```

But this is also a pretty strawman-y example so I'm not steadfast in it, just explaining my thought process.

Also, FWIW I'm just keeping consistent with source/target as I wrote that in the first comment, if there's better names then I'm all for it.

quisquous: It doesn't change between uses.  It's just random quirks between bosses and encounters.  I'd be fine calling them `source` and `target`.  Everything right now is just ordered the same way that the lines appear in the logs.

As a conversational aside, one thing that might help in the future is that I was planned to make some canned  conditions (and responses).  So, we could have like `Conditions.onMe()` but maybe there could be a tether version like `Conditions.tetherOnMe` or maybe somehow we could magically handle `onMe` internally and do both??

panicstevenson:
> I'm amused that you want one camelcase and the other not. I don't have strong feelings here.

"headmarker" isn't a thing! 😛 

> I don't know if that explanation helps make things more consistent?

Definitely helps in understanding the pattern, for sure. It's also not a massive issue in the sense that the reference material will be any other trigger file for any other fight.

> Or if there's some better way to write these?

I'm mostly posting to hopefully spark the discussion or find out who is decidedly more clever than I am.

> These originally were much longer, e.g. targetName but I felt like there was some value in succinctness and dropped to the shorter values that are there now. Maybe I went too far?

I'm definitely in the camp of naming things long names, but I know some people feel strongly the other way. If I wrote this (Spoiler: I didn't) I would have definitely gone targetName, effectName, etc. but I can see the tediousness over time and brevity is the soul of wit, etc.

PS: cleverer and more clever are both correct, apparently

resources/regexes.js
JLGarber: Do you happen to know whether the source/target order seems to be tied to a specific tether type? I'll have to do some digging in my logs or possibly run some older stuff again, but just from a brief check through Grand Cosmos, both the "get away" and purple "okay" tethers are source --> target in the log lines, consistently. I'd be very interested to see whether the random tethers happen to be ones that are accompanied with overhead markers. (I'm thinking here of magnet tethers in particular, but this same thing is also true of the chain tethers in the Vault and Shinryu.)

JLGarber:
My view of the shed's paint job is that "headMarker" makes sense to me if we think of it as a contraction of "Overhead Marker". Or. Orrr, we could confuse everyone and use the game data name for it, `Lockon`

I tend to prefer source/target naming, myself. The way I see it, if the capture groups are going to be named now, rather than numbered, it *really* makes sense for the `matches[whatever]` keys to be consistent as to which "direction" the capture fields are pointing. When I'm putting triggers together, I'm usually thinking in terms of "who's getting hit?".

(Tethers being their own special beast makes perfect sense. I'll put investigation on how they work on my List Of Things To Do, but given how infrequently we've had to deal with them to date, I think the implementation as it is should be fine.)

I gotta agree with Quisquous on shorter vs longer identifiers though. I don't like it, but saving 5-10 characters horizontally is a win. In my case, since I'm not using ultrawide monitors, it's nice to minimize the need for horizontal scrolling when opening trigger and timeline files side by side in the editor.

Akurosia:
Looks good for me.
Only thing i would like to have is a Guide page with examples of ingame log lines and what a complete trigger for it would look like.

quisquous:
Hmm hmm! So, trying to be more consistent about source/target namings, what about this proposal:

```
source startsUsing on target
source uses ability on target
headmarker on target
target gainsEffect from source
target losesEffect from source
tether from source to target
```

I think that could make most of the "on me" conditions consistently be "am I target".

Thoughts?

resources/regexes.js
panicstevenson: There's some entries in [the log guide](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#23-networktether) for examples, apparently (good job @quisquous for documenting!).

I _think_ (read: extrapolating off minimal data) that if you're tethered to something that isn't a tangible thing then  you are the source of the tether, which, to me, makes sense. For instance, based off this, I would assume the fire and ice tethers in Syrcus Tower's first boss would also have you as the source, etc.

panicstevenson:
That makes more sense _to me_. I also think that it falls more in line with the verbiage of [the log guide's entries](https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#1a-networkbuff).

resources/regexes.js
JLGarber: Hah, I even edited that very section a while ago and forgot the listed format was a thing.

Poking through my logs, now that you've suggested that, it looks like some inanimate objects might not actually *get* tether log lines? The Shivering/Smoldering/Shuddering Souls in Syrcus Tower certainly don't, instead placing a head marker on the tethered players. It's not clear where the status drops, I couldn't find corresponding status lines to go with the markers. (But that was a bit of a cursory look, it might need a thorough inspection.) The Chaosphere tethers in ARF similarly just... don't seem to do anything besides attacking with a bleed.

That being said, there were a couple places where it looks like your theory is correct. (Flight units on 9S, for instance.) This is all way beyond scope at this point, but I'll definitely look into that later.

quisquous:
> Looks good for me.
> Only thing i would like to have is a Guide page with examples of ingame log lines and what a complete trigger for it would look like.

I'll add some comments in `resources/regexes.js` that point to the cactbot log guide entries for these.  I think that, combined with the comments about what fields are supported should be sufficient.  If you feel like it isn't, we can add more documentation to the triggers guide after this.

quisquous:
Ok, seems like everybody is on board with source/target.  I left name there for added/removing combatant.  If anybody feels strongly that this should be target, let me know.

PR #741, opened 2019-11-10
Update fishing data for 5.1

mabako:


quisquous:
Thanks for this! I'd been meaning to update it, but SaintCoinach kept crashing on me and I hadn't tried updating to a more recent version to get this in.

PR #743, opened 2019-11-10
lint: Initial Markdownlint Changes

panicstevenson:
Update various Markdown files in accordance with Markdownlint's
suggestions. Changes related to spacing were added via Markdownlint's
--fix option, while other changes were mostly implemented manually. As
for the code block language distinctions, I picked what seemed
reasonable for each type, opting for `bash` as the option for the
timeline output, seeing as that gives it _some_ color, but nothing too
distracting.

Spelling changes were done manually.

Markdownlint can be added to automation eventually, but there are still
some decisions that need to be made before that happens, namely:

* Unordered list convention is not consistent.
  * Markdownlint picks whichever is the first used unordered list
   character, and assumes that should be used everywhere. Some files
   begin with dashes, and others with asterisks, leaving inconsistent
   styling. I believe this is mainly due to the Table of Content
   generation in some files, which we can automate as well with
   whichever style we prefer.
* Line length.
  * Markdownlint wants to enforce line length to be 80 characters. I'm
   not too partial here other than 80 sounds too low. I'm open to 100 or
   120, but there are a lot of files that don't follow any line length
   conventions. Depending on what we decide, we might not want to
   enforce this.

### DIFF
--- a/AdvancedCactbot.md
+++ b/AdvancedCactbot.md
@@ -1,11 +1,13 @@
 # Advanced Cactbot Usage
 
 1. [Configuring UI Modules](#configuring-ui-modules)
-  1. [Customizing Appearance](#customizing-appearance)
-  1. [Customizing Behavior](#customizing-behavior)
-  1. [Text To Speech](#text-to-speech)
-  1. [Adding Custom Triggers](#adding-custom-triggers)
-  1. [Regular Expression Extensions](#regular-expression-extensions)
+    1. [User Directory](#user-directory)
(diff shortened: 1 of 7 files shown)


quisquous:
Thanks!!

I vaguely prefer asterisks as you have them for unordered list elements, as you've changed them here.

I think eslint is set to 100 characters, which seems reasonable to be consistent there? That said, going to 80 means people are more likely semantic linebreaks rather than just running to 100.  So, I would also be fine with 80.

Issue #683, opened 2019-10-21
White Mage lily timer is missing


attickdoor:
The amount of time until next lily is right next to the number of lilies in memory, but is not included in WhiteMageJobData. The commented out struct contains it, would be useful to accurately track how much time until the next lily.

quisquous:
Should be fixed by #745

PR #745, opened 2019-11-11
plugin: Rewrite FFXIVProcess.cs with FieldOffsets

KattTails:
Yikes. Sorry for the mess with the commits. I'm very new to using git and it shows.

### DIFF
--- a/CactbotOverlay/CactbotEventSource.cs
+++ b/CactbotOverlay/CactbotEventSource.cs
@@ -411,6 +411,16 @@ namespace Cactbot {
               OnPlayerChanged(e);
             }
           }
+        } else if (player.job == FFXIVProcess.EntityJob.GNB) {
+          var job = ffxiv_.GetGunbreaker();
+          if (job != null) {
+            if (send || !job.Equals(notify_state_.gnb)) {
+              notify_state_.gnb = job;
+              var e = new JSEvents.PlayerChangedEvent(player);
+              e.jobDetail = new JSEvents.PlayerChangedEvent.GunbreakerDetail(job);
+              OnPlayerChanged(e);
+            }
+          }
         } else if (player.job == FFXIVProcess.EntityJob.BRD) {
           var job = ffxiv_.GetBard();
           if (job != null) {
@@ -418,6 +428,16 @@ namespace Cactbot {
               notify_state_.brd = job;
               var e = new JSEvents.PlayerChangedEvent(player);
               e.jobDetail = new JSEvents.PlayerChangedEvent.BardDetail(job);
+              OnPlayerChanged(e);
+            }
+          }
+        } else if (player.job == FFXIVProcess.EntityJob.DNC) {
+          var job = ffxiv_.GetDancer();
+          if (job != null) {
+            if (send || !job.Equals(notify_state_.dnc)) {
+              notify_state_.dnc = job;
+              var e = new JSEvents.PlayerChangedEvent(player);
+              e.jobDetail = new JSEvents.PlayerChangedEvent.DancerDetail(job);
               OnPlayerChanged(e);
             }
           }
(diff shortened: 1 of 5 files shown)


CactbotOverlay/CactbotEventSource.cs
```diff
@@ -804,11 +834,14 @@ private class NotifyState {
...
       public FFXIVProcess.PaladinJobData pld = new FFXIVProcess.PaladinJobData();
+      public FFXIVProcess.GunbreakerJobData gnb = new FFXIVProcess.GunbreakerJobData();
       public FFXIVProcess.BardJobData brd = new FFXIVProcess.BardJobData();
+      public FFXIVProcess.DancerJobData dnc = new FFXIVProcess.DancerJobData();
       public FFXIVProcess.NinjaJobData nin = new FFXIVProcess.NinjaJobData();
       public FFXIVProcess.BlackMageJobData blm = new FFXIVProcess.BlackMageJobData();
       public FFXIVProcess.WhiteMageJobData whm = new FFXIVProcess.WhiteMageJobData();
-      public FFXIVProcess.SummonerAndScholarJobData smn_sch = new FFXIVProcess.SummonerAndScholarJobData();
+      public FFXIVProcess.SummonerJobData smn = new FFXIVProcess.SummonerJobData();
+      public FFXIVProcess.ScholarJobData sch = new FFXIVProcess.ScholarJobData();
```
KattTails: I've added new handlers for gunbreaker & dancer, and due to how different the classes are now in Shadowbringers I've split summoner and scholar into their own seperate handlers.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -789,17 +916,42 @@ public class NinjaJobData {
...
+      public ushort umbral_state; // Positive = Umbral Fire Stacks, Negative = Umbral Ice Stacks.
+
+      [FieldOffset(0x05)]
+      public byte umbral_hearts_count;
+
+      [FieldOffset(0x06)]
+      public byte foul_count;
+
+      [FieldOffset(0x07)]
+      public byte enochian_state; // Bit 0 = Enochian active. Bit 1 = Polygot active.
```
KattTails: A few of the classes job data needs further unpacking before it can be passed off which makes the JobData classes still necessary as far as I am aware.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -676,25 +632,95 @@ public class PaladinJobData {
...
+          PaladinJobData j = new PaladinJobData() {
+            oath = mem.oath,
+          };
+          return j;
+        }
+      }
+    }
+
+    [StructLayout(LayoutKind.Explicit)]
+    public struct GunbreakerJobMemory { // TODO: Needs a lot more research.
```
KattTails: I'm not familiar with gunbreaker's mechanics and I don't have it levelled to be able to work out how this data works. I've only been able to make a guess.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -703,34 +729,125 @@ public enum Song {
...
+            soul_gauge = mem.soul_gauge,
+            song_type = (BardJobData.Song)mem.song_type,
+          };
+          return j;
+        }
+      }
+    }
+
+    [StructLayout(LayoutKind.Explicit)]
+    public unsafe struct DancerJobMemory { // TODO: Needs more research.
```
KattTails: As with gunbreaker, I'm not 100% confident with these, as there could be more going on than I was able to see at level 60.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -831,108 +983,210 @@ public class BlackMageJobData {
...
-      j.lilies = bytes[kJobDataInnerStructOffsetJobSpecificData + 4];
-      return j;
+    public unsafe WhiteMageJobData GetWhiteMage() {
+      fixed (byte* p = GetJobSpecificData()) {
+        if (p == null)
+          return null;
+        else {
+          WhiteMageJobMemory mem = *(WhiteMageJobMemory*)&p[0];
+          WhiteMageJobData j = new WhiteMageJobData() {
+            lilies_ms = mem.lilies_ms,
```
KattTails: Closes #683

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -570,10 +485,19 @@ public class EntityData {
         // The pointer can be null when not logged in.
         return null;
       }
-
-      return Read8(job_inner_ptr, kJobDataInnerStructSize);
+      job_inner_ptr = IntPtr.Add(job_inner_ptr, kJobDataInnerStructOffset);
+      return Read8(job_inner_ptr, kJobDataInnerStructSize - kJobDataInnerStructOffset);
```
quisquous: Can you subtract this from the size in the declaration rather than here at the point of use?

CactbotOverlay/FFXIVProcess.cs
quisquous: I have gnb at 80 and can give this a look for you when the game isn't having maintenance.

CactbotOverlay/FFXIVProcess.cs
quisquous: I'm not totally sure I agree here, but I think that can be for a followup pull request and not done here.  This is all a really nice change as it is.  For what it's worth, I think these memory classes could just have a ToJObject() function that can do the unpacking as it needs.

CactbotOverlay/FFXIVProcess.cs
KattTails: Done.

quisquous:
I'm going to close this in favor of #750.

PR #746, opened 2019-11-11
add tug types from icykoneko/ff14-fish-tracker-app

mabako:
add tug types from icykoneko/ff14-fish-tracker-app (code for https://ff14fish.carbuncleplushy.com/)

Unfortunately, the current solution to detect tug types for fish doesn't detect all tug types correctly.

While the fish tracker mostly caters towards catching big fish, this is (presumably) what you most care about in terms of tugs. The data is presumably not complete, seeing as not all fish are relevant for the tracker - but it should include all time & weather-limited fish including mooching and their predators.

Old:

grafik

New:

grafik

This matches the packets with opcode 0x96 you receive for the fishing animation events more closely, although the current fishing implementation doesn't actually use network packets.


util/get_fisher_data.py
```diff
@@ -11,6 +12,7 @@
 locales = ['de', 'en', 'fr', 'ja']
 tackle_id = 30
 base = 'https://xivapi.com/'
+fishTrackerBase = 'https://raw.githubusercontent.com/icykoneko/ff14-fish-tracker-app/master/private/fishData.yaml'
```
quisquous: This seems like a lot of dependencies to get tug data.  Is the tug data in Saint Coinach or xivapi such that cactbot could look it up itself?  Or is this just user data that the yaml data contains?

quisquous:
Thanks for this!

util/get_fisher_data.py
mabako: As far a s I’m aware neither of those contains the tug data- I don’t think the game client knows at all, since it simply plays whatever animation the server sends. The only other way to know this is via network logs, but the relevant packet for this is only listed under 252|... so I don’t know if that can be used.

The tug data is manually tracked (in either the spreadsheets or the yml file).

util/get_fisher_data.py
```diff
@@ -181,6 +193,38 @@ def get_tackle():
...
+
+def get_tugs(fishes):
+    tugs = {}
+    data = fish_tracker()
+    for fish in data:
+        if 'tug' in fish and fish['tug']:
+            id = find_fish_by_name(fishes, fish['name'].lower())
+            tug = None
+
+            if fish['tug'] == 'light' or fish['tug'] == 'Light':
```
quisquous: Could you store `fish['tug'].casefold()` in a local here so you don't have to compare against both upper and lowercase, and can simplify this?

util/get_fisher_data.py
quisquous: Other information would be great if you wanted to add it!

util/get_fisher_data.py
mabako: done!

quisquous:
Thanks!!

PR #750, opened 2019-11-12
jobs: Serialization and code simplification

KattTails:
Hi!

I took my previous pull request and made it obsolete. The Gunbreaker, Dancer and Ninja memory bytes need confirming, I'm unable to test them properly on my character. I'm not really confident at all with the gunbreaker bytes.


CactbotOverlay/CactbotEventSource.cs
```diff
@@ -370,145 +370,97 @@ protected override void Update()
...
-            if (send || !job.Equals(notify_state_.sam)) {
-              notify_state_.sam = job;
-              var e = new JSEvents.PlayerChangedEvent(player);
-              e.jobDetail = new JSEvents.PlayerChangedEvent.SamuraiDetail(job);
-              OnPlayerChanged(e);
+        var job = ffxiv_.GetJobSpecificData(player.job);
+        if (job != null) {
+          if (send || !job.Equals(notify_state_.jobData)) {
+            notify_state_.jobData = job;
+            var e = new JSEvents.PlayerChangedEvent(player);
```
quisquous: So one thing you're losing with this change is that there's no longer an equals check, so we'll send job data constantly.  Is it possible to just compare the bytes that we read and turned into job specific data and see if those changed before sending?

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -570,522 +486,301 @@ public class EntityData {
...
-        return hash;
+      job_inner_ptr = IntPtr.Add(job_inner_ptr, kJobDataInnerStructOffset);
+
+      fixed (byte* p = Read8(job_inner_ptr, kJobDataInnerStructSize)) {
+        if (p == null)
+          return null;
+        else {
+          switch (job) {
+            case EntityJob.RDM: {
+                return JObject.FromObject(*(RedMageJobMemory*)&p[0]);
```
quisquous: What I wonder is if RedMageJobMemory could just be serialized itself, rather than going job memory -> jobject -> job detail classes.  Just, job memory, that's it.

You can add properties that are serialized, like this: https://github.com/quisquous/ACT_EnmityPlugin/blob/master/EnmityPlugin/FFXIVObject.cs#L51

You can also disallow properties to be serialized with `[NonSerialized]`.

Between the two of those, I think you should be able to make it so that the "job memory" classes serialize to the same as the "job detail" classes.  In most cases, they should be identical?

KattTails: Ahhh that's embarrassingly simple... I can make those changes and strip out all the JobDetail classes from JSEvents.

CactbotOverlay/CactbotEventSource.cs
KattTails: There is a check on line 375 still unless I'm misunderstanding?

CactbotOverlay/CactbotEventSource.cs
```diff
@@ -517,8 +469,8 @@ protected override void Update()
         }
       }
 
-      // onTargetChangedEvent: Fires when current target or their state changes.
-      if (target != notify_state_.target) {
+          // onTargetChangedEvent: Fires when current target or their state changes.
```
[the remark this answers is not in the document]
KattTails: Done. VS likes to mess with it..

CactbotOverlay/CactbotEventSource.cs
KattTails: Looking at it further I've replaced that line with `if (send || !JObject.DeepEquals(job, notify_state_.jobData)) {`.

CactbotOverlay/FFXIVProcess.cs
KattTails: I've pushed a new build with this implemented for you to review when you can :)

LethargyKaz:
This change looks awesome, the only things I found were code style-related / preexisting, like you could use string.Empty instead of "" in some places, but they felt like dumb comments / not related.

I tired to think of a way to shorthand the switch statement, but failed, which I think is fine, its very readable as is!

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -570,525 +504,445 @@ public class EntityData {
...
+    public struct GunbreakerJobMemory { // TODO: Needs a lot more research. 
+      [FieldOffset(0x00)]
+      public byte cartridges;
 
-      var j = new WarriorJobData();
-      j.beast = bytes[kJobDataInnerStructOffsetJobSpecificData];
-      return j;
-    }
+      [FieldOffset(0x02)]
+      public ushort continuationMilliseconds; // Notes: Jumps to 15000 (ms?) after using Gnashing Fang, doesn't decrease.
```
quisquous: continuationMilliseconds is 15000 if and only if continuationState is not zero.  I'm not sure it's milliseconds, and probably isn't worth including.

continuationState is correct.  It's 0 when not doing any combos, it's 1 after you hit Gnashing Fang, and 2 after you hit Savage Claw.  When you hit Wicked Talon it goes back to 0.

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -570,525 +504,445 @@ public class EntityData {
...
-      return j;
-    }
+      [FieldOffset(0x04)]
+      public byte ninkiAmount;
 
-    public class NinjaJobData {
-      public uint huton_ms = 0;
-      public uint ninki_amount = 0;
+      [FieldOffset(0x05)]
+      public byte hutonCount; //TODO: Confirm this.
```
quisquous: It sure seems like this is true, but what in the world???  I would say it's probably better to not expose this, as it's not useful.

CactbotOverlay/CactbotEventSource.cs
```diff
@@ -799,21 +666,8 @@ private class NotifyState {
       public bool? in_game_combat;
       public bool dead = false;
       public string zone_name = null;
+      public JObject jobData = new JObject();
```
quisquous: a tiny nit, but this should be `job_data` and not `jobData` to fit with the surrounding style.

quisquous:
Sorry this took so long for me to get back to reviewing.  All of this really looks great overall, and I love how much it cleans up the code.  I think I have just a few tiny things that I mentioned, and then I'm happy to land this.  It looks like it merges in cleanly even after the ngld overlayplugin changes, so that's great.  :)

Only tangentially related to this patch, but there have been a few questions about how to have multiple sets of offsets at once (#779) to support older chinese versions that are a version behind.  If you have any thoughts about how to cleanly handle multiple sets of memory offsets, I'd be curious to hear them.

CactbotOverlay/FFXIVProcess.cs
KattTails: Yeah I couldn't believe it was actually counting that. I guess it's an artefact from the numerous reworks ninja has had..

KattTails:
Hiya! I've pushed the changes you suggested, and no problem with the delay, I hope you had a nice holiday :)

With regards to the CN versions, would you be thinking of having one plugin for all versions or splitting for a separate CN build? Either way you'd be committing to following an extra release schedule so ultimately I don't think it really matters which way you went. 
That being said, either a preprocessor definition with separately defined offsets and signatures for the CN build, or, depending on how similar the CN client's memory is, you might be able to just use an extra offset and keep the same sigs? Maybe a mix of both, although this could change with each patch. After that and depending on job differences you'd just cast to a `JobMemory_CN` struct.
You wouldn't need any changes outside of `FFXIVProcess.cs` as far as I can tell at least, which is nice.

g-dove:
for CN you could make the structs you have right now implement an interface and make new CN structs that also implement the same interface, then you could return them to the rest of the code as interfaces but swap out the actual implementation based on it being CN or not?

CactbotOverlay/FFXIVProcess.cs
```diff
@@ -746,7 +746,7 @@ public struct NinjaJobMemory {
       public byte ninkiAmount;
 
       [FieldOffset(0x05)]
-      public byte hutonCount; //TODO: Confirm this.
+      private byte hutonCount; // Why though?
```
quisquous: haha this comment :)

quisquous:
Yay!  Thanks for all the cleanup here.  This is all great work.

> With regards to the CN versions, would you be thinking of having one plugin for all versions or splitting for a separate CN build? Either way you'd be committing to following an extra release schedule so ultimately I don't think it really matters which way you went.

Yeah, I think I'd rather have a single build.  That would make things a lot easier than trying to release two separate plugin DLLs.  That's already been enough of a pain.

I think ideally, the act language can get used to figure out which offsets and signatures to use.  Mostly I'm not sure what the best way to organize this in code.  Maybe the common helper memory functions in `FFXIVProcess.cs` could get moved into some other utility class, and we could have `FFXIVProcessNewest` and `FFXIVProcessCn` that both implemented some `FFXIVProcess` abstract interface with functions like GetJobSpecificData, etc etc.  So then you'd still use some `FFXIVProcess` object, but it might be one of those two different types.  Then when the chinese version upgraded, it should be pretty easy to just copy one file onto the other (or grab the right one out of version control).  How does that sound?

If you're interested in doing this, maybe it'd be worth continuing this conversation in a github issue? (If you're not interested, I can put it on my list of stuff to do.  It just seemed in line with what you've already been looking at, so I thought I'd ask.)

quisquous:
Oh, oh no, I didn't realize this was going into the ngldoverlay branch.  I'll merge it over, but @Bluefissure is going to be sad.

Bluefissure:
> Oh, oh no, I didn't realize this was going into the ngldoverlay branch. I'll merge it over, but @Bluefissure is going to be sad.

It's okay since I made a new PR https://github.com/quisquous/cactbot/pull/803 based on this.

PR #753, opened 2019-11-13
Add the Cascade enrage spell to Living Liquid phase of TEA

Legends0:
Timing is 8 seconds after Fluid Swing.


Legends0:
Here's a log with the event: https://www.fflogs.com/reports/qRWpDKHZr3wG7fy9#fight=62&type=damage-done&hostility=1&view=events&eventstart=12748499

PR #754, opened 2019-11-14
Add timers up to Final Sentence for TEA

Legends0:
Based on times from https://www.fflogs.com/reports/WthzBwZTmMbGXP9g#fight=110

Living Liquid and Living Rage appear to be interchangeable in this phase.


quisquous:
Thanks for this!

Issue #733, opened 2019-11-08
Fishing Timers showing nothing


Krojack:
## Question

I've been trying to get the Fishing timers to work but they never display any data.  I see the overlay and what bait I have selected but that's all.  The cast timer just stays at 0.0 and the @ location never changes to the fishing hole I'm at.  I do see the following in my chat log.
```
You cast your line on the Norvrandt Slope.
Something bites!
You land a stormfish measuring 44.3 ilms!
```
I have ACT installed in **N:\ACTv3\** and the **cache** folder is writable as date/time stamps are recent.

Other mods like RaidBoss work just fine.

KattTails:
Could you try catching a fish and then pasting the relevant network log lines for us to take a look at? It seems like a regexp error as the plugin is loading and displaying fine, just not parsing the fishing events.

Krojack:
Are you referring to the FFXIV parse logs? If so I've attached them below. I caught several fish but nothing showed up in the log however it will show me changing bait.  If there is a better way to get the network log data let me know.

[Network_2044_20191114.log](https://github.com/quisquous/cactbot/files/3847717/Network_2044_20191114.log)

I've changed my real characters name to "My Character" in the log just for this upload. Shouldn't change anything though.

KattTails:
Oh, it seems like you have the chat filter for gathering messages turned off. The overlay parses those, so you'll need to turn on 'own gatherer messages' in one of your log filters. It should work after that. I've attached an image for the option and where to find it below: 
![image](https://user-images.githubusercontent.com/12807478/68880497-47da3300-0703-11ea-9cf2-2d785656bb04.png)

Krojack:
Just checked and it was enabled. In fact it was enabled in 2 chat tab. I tried turning it off in one of them, no change.  I enabled other fishing messages and nothing there either.  I checked network logs again and no change there.

![ff14-ss](https://user-images.githubusercontent.com/908461/68882338-f919a280-06dc-11ea-9d45-b8fd1e352c52.jpg)

KattTails:
Do you have this option turned on inside ACT FFXIV plugin options? 
![image](https://user-images.githubusercontent.com/12807478/68882607-693d1e00-0707-11ea-96d1-ede45644842a.png)

Krojack:
That did it... Working now, thank you. Might wanna toss this into some FAQ or a Trouble Shooting list somewhere.

Thank you much.

KattTails:
[...]

PR #757, opened 2019-11-14
Add troubleshooting information for fisher module

KattTails:
Hi, as per #733 I've updated the readme.md with some troubleshooting for the fisher overlay. I'll try and keep the troubleshooting doc updated with more common problems if they arise.


README.md
```diff
@@ -237,9 +237,13 @@ When you cast your line at a fishing hole, this module keeps track of when you r
 
 ![fishing screenshot](screenshots/promo_fishing.png)
 
+You will need to ensure the chat filter for `[ ] Own Gatherer Messages` is enabled in at least one of your chat tabs and that the option `[ ] Hide Chat Log (for privacy)` is not enabled in ACT FFXIV Plugin's options. 
+
+See [here](https://gist.github.com/KattTails/a67527be56be64e22b9f2eb962d5cf4f#file-cactbot-fisher-troubleshooting-md) for instructions for this and other common problems.
```
quisquous: Do you want to include this inside the cactbot repository proper? Maybe just by adding a docs/readme-fisher.md or docs/fisher-troubleshooting-md file?

KattTails: You're right it would be more suited in such a file, not the main `readme.md`. I was just concerned people wouldn't find it/think to look there and you'd end up with the same number of issues created about it. 

The docs/ folder only really has 'advanced' documentation in it currently, perhaps it could be useful to have a general-troubleshooting.md with common cactbot issues similar to [ngld's FAQ](https://gist.github.com/ngld/e2217563bbbe1750c0917217f136687d#my-act-isnt-showing-any-numbers-what-can-i-do). You could then move the `Potential errors and workarounds` section from the readme.md too and link there instead.

quisquous: I think this documentation would fit in the docs/ folder, and I'm not sure I understand how linking to a gist is more findable than linking to a markdown file in the repository.  I definitely agree that there could also be a potential errors and workarounds doc too!

KattTails: Ah I meant on the rare occasions people go to check the readme before asking for help ^^
I thought if someone was having an issue with the fisher module for example, they'd go look in the fisher module section on the main page, not find anything and give up. I'll look into building a list of common problems & solutions and drafting up some F.A.Q style document.

quisquous: Sorry if I wasn't clear--I was trying to propose adding a doc in the docs/ folder and then linking it from both the main documentation talking about the fisher module as well as an FAQ.

quisquous: I think really my objection here is to have half the documentation about fishing in the cactbot repo and half in a gist.  I think it would make more sense to have it all in the cactbot repo, and instead of linking to a gist, instead link to some markdown file in docs/ so it can all be centrally located, if that makes sense?

I think the other changes to the main README.md are fine.

KattTails:
Ahh sorry for forgetting about this! I've added an FAQ into 'docs\` and added a link on the main `readme.md` to it. I've moved the `Common workarounds and solutions` section into the new FAQ and updated it for ngld's OverlayPlugin. I included a few of the common errors I've seen on the FFXIV ACT Discord too, and added the fishing module info underneath. Your thoughts on these changes?

README.md
```diff
@@ -215,6 +204,8 @@ Cast times are currently only logged as you fish, so there won't be any data unt
 
 [See here](https://www.youtube.com/watch?v=GHgWIA-Zhug) for examples of the different tug types.
 
+Check [here](docs/Fisher-Troubleshooting.md) for common troubleshooting tips.
```
quisquous: I assume you mean docs/FAQ-TroubleShooting.md#fisher-module?

KattTails: Oops. Yes, that would be correct. Originally it was two files but it felt unnecessary. I've changed it.

PR #778, opened 2019-11-20
Add canned responses and conditions

quisquous:


quisquous:
Just putting this here for discussion.  I left a bunch of TODOs for things I wasn't quite sure on.

resources/responses.js
```diff
@@ -0,0 +1,523 @@
...
+  },
+  stop: (sev) => {
+    let obj = {};
+    obj[defaultAlertText(sev)] = {
+      en: 'Stop',
+      de: 'Stopp',
+      fr: 'Stop',
+    };
+    return obj;
+  },
```
JLGarber: Would it be worth splitting this into "stop everything" vs "stop moving"? I'm thinking here of Acceleration Bomb as a "stop everything" mechanic, while there are various other mechanics (such as Scathach's Shadow Links or Guardian's Electric Pulse/Radar) that permit actions while forbidding movement.

resources/responses.js
```diff
@@ -0,0 +1,523 @@
...
+      ja: '<= 左/西へ',
+    };
+    return obj;
+  },
+  goEast: (sev) => {
+    let obj = {};
+    obj[defaultAlertText(sev)] = {
+      en: 'Get Right/East =>',
+      fr: 'Allez à Droite/Est =>',
+      de: 'Nach Rechts/Westen =>',
```
JLGarber: Should this be `'Nach Rechts/Osten =>'`?

JLGarber:
(Apologies for the odd line selections, I'm not sure why my browser doesn't like multi-line selections.)

I really like this. There's going to be somewhat of a learning curve in that we'll have to look up the exact names for the responses, but once we're past that I think it's going to be much more readable and less error-prone. (I know at least for myself, reducing the number of times I have to type out boilerplate responses reduces the need for a reviewer to catch mistakes in the trigger body.)

I'm not particularly familiar with spread syntax, but I'm not seeing any issues with it here. It feels reasonable to implement defaults this way.

resources/responses.js
Akurosia: I can confirm this should be `'Nach Rechts/Osten =>'`

resources/responses.js
quisquous: Yeah, good thought, thanks! I'll add both of those.

resources/responses.js
quisquous: Whoops, thanks!

quisquous:
Ok, wrote some tests, and fixed some tests, and this should be good to go.  Any last comments?

quisquous:
Oh no, apparently `...` does not seem to work inside an eval context, which is how all the triggers are loaded (right now, at least).

```
// ok
let x = {'a': 3, 'b': 4};
let y = {...x};
let x2 = eval("x = {'a': 3, 'b': 4}");

// not ok
let y2 = eval("{...x}");
```

panicstevenson:
Maybe something like caresAboutMagical, caresAboutPhysical, and caresAboutAOE? So we’re not tied to specific things

quisquous:
> Maybe something like caresAboutMagical, caresAboutPhysical, and caresAboutAOE? So we’re not tied to specific things

Done.

I do wonder if this will get noisy, but it is what it is.

I guess we need to magical tankbusters (if we want people to addle them) to have sort of infoText-y warnings for non-tanks and non-healers, since casters probably don't want yellow TANK BUSTER alerts.

quisquous:
Looks like nobody else really has any thoughts here, so I'm going to land this.

PR #791, opened 2019-11-23
Knockback and Cleave callouts for Limit Cut in TEA

Legends0:
delaySeconds calculated from duration from when player first gets their marker to when the charge or cleave for that player goes off - 5 seconds.

Does not necessarily mean that that player is the one getting hit... if they died then some other player will just be charged or cleaved.

Cleave on you callout is used... cleave behavior is slightly different from tank cleave as you must face away from the part versus tank cleave facing towards.


Legends0:
Comment should probably be updated to something, I took the comment from e3s, but this trigger includes cleave callout.

Legends0:
This does not work yet, need to convert string to hex to do binary comparison properly.

Legends0:
The timings don't work for the second limit cut.

quisquous:
Thanks! I got a log and a video through the end of the fight, so I'll see what this looks like and fix it up for wormhole if it's not good.

quisquous:
I really appreciate all the ultimate triggers and timelines! :+1: :+1: :+1:

PR #797, opened 2019-11-24
Remove oopsyraidsy warning message on ninja rabbit cast

anubiann00b:
This PR moves the oopsyraidsy warning message when ninjas mess up their mudras and cast rabbit medium out of the default messages in general.js and into the user example file.

My reasoning was no other class displays individual mechanical errors as a oopsyraidsy warning, and that a mudra error doesn't necessarily even mean a trick attack error--it could be from a raiton/suiton outside of the suiton for trick attack. Even if it was, no other class displays errors even for messing up raid buffs.

Let me know what you think!

### DIFF
--- a/ui/oopsyraidsy/data/00-misc/general.js
+++ b/ui/oopsyraidsy/data/00-misc/general.js
@@ -39,6 +39,6 @@
           return;
         delete data.lostFood[e.targetName];
       },
-    }
+    },
   ],
 }];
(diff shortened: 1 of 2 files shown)


anubiann00b:
so sorry--hit enter early, editing and adding description

quisquous:
Sorry, I really find this warning to be useful that trick attack might be delayed or something else might go awry.

If anything, in the future, I'd like to add more personal warnings to oopsy, like missing positionals, or breaking combos, or using aoe on things that don't hit enough targets.  Maybe at that point we could make those individual (and not show for everybody else).  However, I still think the bunny is a useful indicator for me in pugs.

quisquous:
I'm going to close this then, sorry to disagree.

PR #798, opened 2019-11-24
fisher: Add support for undiscovered fishing holes

KattTails:
Hiya, following a discussion on the ACT discord, the current fisher overlay doesn't handle undiscovered fishing holes very well.
I'm not very confident with the chinese regexes I've added, but I was able to test the en/fr/de/ja entries.

@xephero If you would like to look this over I'm open to any comments you may have.

Thanks!


quisquous:
Thanks for this fix!

PR #800, opened 2019-11-25
raidboss: large tea trigger update

quisquous:


quisquous:
There's a pile of obvious bugs here that I need to fix before this can land, but hopefully in the next day or two.

r4ha:
Japanese client has different Nisi symbol like "αβγδ". so how should I translate this? just put in to replaceSync or add ja to regex?

quisquous:
Those are also the English symbols, they just get capitalized in the effect messages.  Can you paste some `1A:` gains effect log lines with Nisi here for me?

PR #803, opened 2019-11-26
radar: add rotation to dll & test ui

Bluefissure:
I think adding this address in advance is a good idea since the new radar ui module is independent with further updates but the dll files are not.
I'll make another PR of radar ui after refactoring it to the ngld version.


quisquous:
Oh no, I realized that @KattTails's change to all of these memory locations got merged to the wrong branch, so this will conflict again.  Let me merge that in and maybe just fix up the conflicts here so you don't have to write this another time.  <_<

quisquous:
Thanks for this.  I fixed up the merge issues and landed this manually (with you as the author).  Let me know if I broke anything.

PR #804, opened 2019-11-26
raidboss: Fix e3s Smothering Waters regexes

KattTails:
The regexes are missing a `'for '`. It was brought up on the discord by Shira#9825.


Legends0:
The (?:.*) matches the for, but since this is a non-capturing group, the matches[1] still grabs just the number before Seconds.

quisquous:
Yeah, as @Legends0 says, this won't change the result of the regex, but it's still nice cleanup.  If something was failing in e3s (which I heard somewhere?) this won't be the fix, then.

PR #817, opened 2019-12-01
radar: ngld support

Bluefissure:
Based on https://github.com/quisquous/cactbot/pull/739

Updates:
1. Remove `SplitSameMonster` config to make it simple to retrieval the mob.
2. Add a `Puller` to the config file to record the earliest puller.
3. Use `19:` logs to remove mobs.

TODO:
- Resource generator of xivapi


Bluefissure:
Sorry I forget the lint again.

ui/radar/radar.js
```diff
@@ -0,0 +1,235 @@
...
+
+  OnLogEvent(e) {
+    let lang = this.options.Language;
+    for (let i = 0; i < e.detail.logs.length; i++) {
+      // add new combatant
+      let matches = e.detail.logs[i].match(Regexes.addedCombatantFull({}));
+      if (matches) {
+        let monster = this.nameToMonster[matches.groups.name];
+        if (monster)
+          this.AddMonster(e.detail.logs[i], monster);
```
quisquous: Here and in other function calls below like UpdateMonsterPuller, it seems like you could pass matches.groups to the function rather than passing the log and rematching.

ui/radar/radar.js
```diff
@@ -0,0 +1,235 @@
...
+      // removing new combatant
+      r = e.detail.logs[i].match(/19:(.*) was defeated by/);
+      if (r)
+        this.RemoveMonster(r[1]);
+    }
+  }
+
+  OnPlayerChange(e) {
+    this.playerPos['x'] = e.detail.pos.x;
+    this.playerPos['y'] = -e.detail.pos.y;
```
quisquous: I still think the negative y is confusing.  Is it possible to fix the math to operate on the real y coordinate?

ui/radar/radar.js
Bluefissure: removed rematching

ui/radar/radar.js
Bluefissure: y is negative because y position of in-game map is upside down, like

And negative y can transpose it to a standard [Cartesian Coordinate System](https://en.wikipedia.org/wiki/Cartesian_coordinate_system), which will make the math system more "standard".
Yes, we can fix the math, but that's only different in when you negative y coordinate (in the formula, pre-processing, etc).
Since `this.playerPos` is never shown to user so we can treat this as a variable used in the math system.

ui/radar/radar.js
```diff
@@ -0,0 +1,251 @@
...
+    let lang = this.lang;
+    let matchOrNot = (matches.groups.name.match(monster['name'][lang] || monster['name']['en']) != null);
+    matchOrNot &= (log.match(monster['regex'] || '') != null);
+    matchOrNot &= (parseFloat(matches.groups.hp) >= (monster['hp'] || 0));
+    let options = this.options;
+    if (monster['rank'] in options.RankOptions) // options overwrite
+      options = Object.assign({}, this.options, options.RankOptions[monster['rank']]);
+    if (options.OnlyMobs) {
+      matchOrNot &= matches.groups.id.startsWith('4');
+      matchOrNot &= !(log.match(/\(\d{0,5}00000\d{0,5}\)\.$/) === null);
```
quisquous: I added an `npcId` here for this to the addedCombatant regex so you don't have to match it manually.

ui/radar/radar.js
quisquous: I guess I don't think of that as upside down.  They're both valid cartesian coordinates, in my opinion.  The code is just confusing to me because sometimes y is negated (players) and sometimes it's not (monsters).

If you have x,y with y = down, then you can find the angle of the vector d = (monster x - player x, monster y - player y), with this equation: `let degrees = Math.atan2(d.y, d.x) / Math.PI * 180;`

e.g. a delta of (10, 0) is 0 degrees, a delta of (0, 10) is 90 degrees (pointing down).

I think this simplifies the rotation math some too, rather than using acos.

ui/radar/radar.js
```diff
@@ -0,0 +1,251 @@
...
+      matches = e.detail.logs[i].match(Regexes.abilityFull({}));
+      if (matches) {
+        let monster = this.targetMonsters[matches.groups.target];
+        if (monster)
+          this.UpdateMonsterPuller(monster, matches.groups.source);
+      }
+      // change instances
+      let r = e.detail.logs[i].match(instanceChangedRegex[lang] || instanceChangedRegex['en']);
+      if (r)
+        this.ClearTargetMonsters(10); // don't remove mobs lasting less than 10 seconds
```
quisquous: Does this mean they will stick around forever?

ui/radar/radar.js
```diff
@@ -0,0 +1,251 @@
...
+    for (let i = 0; i < e.detail.logs.length; i++) {
+      // add new combatant
+      let matches = e.detail.logs[i].match(Regexes.addedCombatantFull());
+      if (matches) {
+        let monster = this.nameToMonster[matches.groups.name];
+        if (monster)
+          this.AddMonster(e.detail.logs[i], monster, matches);
+      }
+      // network ability
+      matches = e.detail.logs[i].match(Regexes.abilityFull({}));
```
quisquous: This can just be abilityFull() with no {}

ui/radar/radar.js
```diff
@@ -0,0 +1,251 @@
...
+        let monster = this.targetMonsters[matches.groups.target];
+        if (monster)
+          this.UpdateMonsterPuller(monster, matches.groups.source);
+      }
+      // change instances
+      let r = e.detail.logs[i].match(instanceChangedRegex[lang] || instanceChangedRegex['en']);
+      if (r)
+        this.ClearTargetMonsters(10); // don't remove mobs lasting less than 10 seconds
+      // removing new combatant
+      r = e.detail.logs[i].match(/ 19:(.*) was defeated by/);
```
quisquous: I added a Regexes.wasDefeated for this.

ui/radar/radar.js
Bluefissure: got it, I just forget to remove this.

ui/radar/radar.js
Bluefissure: Yes, this is kinda confusing because the `pos_y` of monster is not negatived, it represents the true position of monsters, and thus the position shown in the overlay won't be affected.
When doing the math, what we actually use is the `pos` of monsters, it's negatived as I said, to be used in the math.
But yes, I should make the rotation calculating simple and clear, I'll fix this after.

ui/radar/radar.js
Bluefissure: cool!

ui/radar/radar.js
Bluefissure: At first, it will clear all monsters when changeZone or changeInstance happens. And  then there's a bug that because changeInstance is implemented by matching the `00` log by `instanceChangedRegex `, and such log will appear after several seconds when we enter a map. 

So, if radar detects monsters before the log shows up (which means the monster is just near the teleportation crystal or it's an S rank), it will clear the detected monster by the entering instance log.

ui/radar/radar.js
Bluefissure: cool!

Bluefissure: updated~

ui/radar/radar.js
Bluefissure: removed

ui/radar/radar.js
Bluefissure: got it

quisquous:
Thanks for all the changes and patience with my comments. This is looking good.  I'm going to play around with it some this weekend, see if I have any last comments, and will hope to merge it soon.

Bluefissure:
> Thanks for all the changes and patience with my comments. This is looking good. I'm going to play around with it some this weekend, see if I have any last comments, and will hope to merge it soon.

Sure, you can follow a hunt train to test it~

quisquous:
Seemed to work great for me.  I've got a couple more requests, then this is good to go:

* Did you forget to include the python script to generate hunt.js?

* The Forgiven Pedantry mob did not pop in the radar for me, although everything else did.  I think it's because it's being case sensitive when matching mob names.

Bluefissure:
> Seemed to work great for me. I've got a couple more requests, then this is good to go:
> 
> * Did you forget to include the python script to generate hunt.js?
> * The Forgiven Pedantry mob did not pop in the radar for me, although everything else did.  I think it's because it's being case sensitive when matching mob names.

- Yeah, I forget to put the script, but it's still using coinach, xivapi version may be implemented later this weekend.

- I lowercased the mobnames in the dict though I cannot test it in other servers, hope it works.

quisquous:
I think there's some issues with lowercasing, but I'll follow up myself.

Issue #821, opened 2019-12-02
plugin: dynamically support chinese memory offsets


quisquous:
Because the Chinese version lags a little bit behind, the memory offsets will also lag behind.  I think it'd be nice to figure out how to support both dynamically in a single plugin.  (I don't want to have to go to multiple builds.)

I think implementation-wise, we could just have another version of `FFXIVProcess` (e.g. `FFXIVProcessCn`?). At runtime, we could dynamically get the correct process based on language.  The language can be gotten via `FFXIVPlugin.GetLanguageId()`.  I think as versions update, we should be able to just copy one onto the other and use the correct classes and memory offsets.

If somebody adds a field to the newer version, we may need some help to get the correct one for the older Chinese version, however.

See also: #779, #787

quisquous:
Fixed, thanks!!

PR #825, opened 2019-12-03
plugin: Add dynamic support for CN version of FFXIV

KattTails:
Hiya,

As requested in #821 I've added support for the CN client, using memory offsets from the 0.11.2 release targetting FFXIV ver 5.01.

This also includes the changes from #824

### DIFF
--- a/CactbotOverlay/FFXIVProcessIntl.cs
+++ b/CactbotOverlay/FFXIVProcessIntl.cs
@@ -1,0 +1,646 @@
+﻿using System;
+using System.Runtime.InteropServices;
+using Newtonsoft.Json.Linq;
+
+namespace Cactbot {
+  public class FFXIVProcessIntl : FFXIVProcess {
+
+    [StructLayout(LayoutKind.Explicit)]
+    public unsafe struct EntityMemory {
+      public static int Size => Marshal.SizeOf(typeof(EntityMemory));
+
+      // Unknown size, but this is the bytes up to the next field.
+      public const int nameBytes = 68;
+
+      [FieldOffset(0x30)]
+      public fixed byte Name[nameBytes];
+
+      [FieldOffset(0x74)]
+      public uint id;
+
+      [FieldOffset(0x8C)]
+      public EntityType type;
+
+      [FieldOffset(0x92)]
+      public ushort distance;
+
+      [FieldOffset(0xA0)]
+      public Single pos_x;
+
+      [FieldOffset(0xA4)]
+      public Single pos_z;
+
+      [FieldOffset(0xA8)]
+      public Single pos_y;
+
+      [FieldOffset(0xB0)]
+      public Single rotation;
+
+      [FieldOffset(0x18B8)]
+      public CharacterDetails charDetails;
+    }
+
+    [StructLayout(LayoutKind.Explicit)]
+    public struct CharacterDetails {
+
+      [FieldOffset(0x00)]
+      public int hp;
+
+      [FieldOffset(0x04)]
+      public int max_hp;
+
+      [FieldOffset(0x08)]
+      public int mp;
+
+      [FieldOffset(0x0C)]
+      public int max_mp;
+
+      [FieldOffset(0x12)]
+      public short gp;
+
+      [FieldOffset(0x14)]
+      public short max_gp;
+
+      [FieldOffset(0x16)]
+      public short cp;
+
+      [FieldOffset(0x18)]
+      public short max_cp;
+
+      [FieldOffset(0x3C)]
+      public EntityJob job;
+
+      [FieldOffset(0x3E)]
+      public short level;
+
+      [FieldOffset(0x5F)]
+      public short shieldPercentage;
+    }
+    public FFXIVProcessIntl(ILogger logger) : base(logger) { }
+
+    public unsafe override EntityData GetEntityDataFromByteArray(byte[] source) {
+      fixed (byte* p = source) {
+        EntityMemory mem = *(EntityMemory*)&p[0];
+
+        EntityData entity = new EntityData() {
+          //dump '\0' string terminators
+          name = System.Text.Encoding.UTF8.GetString(mem.Name, EntityMemory.nameBytes).Split(new[] { '\0' }, 2)[0],
+          id = mem.id,
+          type = mem.type,
+          distance = mem.distance,
+          pos_x = mem.pos_x,
+          pos_y = mem.pos_y,
+          pos_z = mem.pos_z,
+          rotation = mem.rotation,
+        };
+        if (entity.type == EntityType.PC || entity.type == EntityType.Monster) {
+          entity.job = mem.charDetails.job;
+
+          entity.hp = mem.charDetails.hp;
+          entity.max_hp = mem.charDetails.max_hp;
+          entity.mp = mem.charDetails.mp;
+          entity.max_mp = mem.charDetails.max_mp;
+          entity.shield_value = mem.charDetails.shieldPercentage * entity.max_hp / 100;
+
+          if (IsGatherer(entity.job)) {
+            entity.gp = mem.charDetails.gp;
+            entity.max_gp = mem.charDetails.max_gp;
+          }
+          if (IsCrafter(entity.job)) {
+            entity.cp = mem.charDetails.cp;
+            entity.max_cp = mem.charDetails.max_cp;
+          }
+
+          entity.level = mem.charDetails.level;
+
+          byte[] job_bytes = GetRawJobSpecificDataBytes();
+          if (job_bytes != null) {
+            for (var i = 0; i < job_bytes.Length; ++i) {
+              if (entity.debug_job != "")
+                entity.debug_job += " ";
+              entity.debug_job += string.Format("{0:x2}", job_bytes[i]);
+            }
+          }
+        }
+        return entity;
+      }
+    }
+
+    internal override EntityData GetEntityData(IntPtr entity_ptr) {
+      if (entity_ptr == IntPtr.Zero)
+        return null;
+      byte[] source = Read8(entity_ptr, EntityMemory.Size);
+      return GetEntityDataFromByteArray(source);
+    }
+    public override EntityData GetSelfData() {
+      if (!HasProcess() || player_ptr_addr_ == IntPtr.Zero)
+        return null;
+
+      IntPtr entity_ptr = ReadIntPtr(player_ptr_addr_);
+      if (entity_ptr == IntPtr.Zero)
+        return null;
+      var data = GetEntityData(entity_ptr);
+      if (data.job == EntityJob.FSH)
+        data.bait = GetBait();
+      return data;
+    }
+
+    public override EntityData GetTargetData() {
+      if (!HasProcess() || target_ptr_addr_ == IntPtr.Zero)
+        return null;
+
+      IntPtr entity_ptr = ReadIntPtr(target_ptr_addr_);
+      if (entity_ptr == IntPtr.Zero)
+        return null;
+      return GetEntityData(entity_ptr);
+    }
+
+    public override EntityData GetFocusData() {
+      if (!HasProcess() || focus_ptr_addr_ == IntPtr.Zero)
+        return null;
+
+      IntPtr entity_ptr = ReadIntPtr(focus_ptr_addr_);
+      if (entity_ptr == IntPtr.Zero)
+        return null;
+      return GetEntityData(entity_ptr);
+    }
+
+    public unsafe override JObject GetJobSpecificData(EntityJob job) {
+      if (!HasProcess() || job_data_outer_addr_ == IntPtr.Zero)
+        return null;
+
+      IntPtr job_inner_ptr = ReadIntPtr(job_data_outer_addr_);
+      if (job_inner_ptr == IntPtr.Zero) {
+        // The pointer can be null when not logged in.
+        return null;
+      }
+      job_inner_ptr = IntPtr.Add(job_inner_ptr, kJobDataInnerStructOffset);
+
+      fixed (byte* p = Read8(job_inner_ptr, kJobDataInnerStructSize)) {
+        if (p == null)
+          return null;
+        else {
+          switch (job) {
+            case EntityJob.RDM: {
+                return JObject.FromObject(*(RedMageJobMemory*)&p[0]);
+              };
+            case EntityJob.MRD: {
+                return null;
+              };
+            case EntityJob.WAR: {
+                return JObject.FromObject(*(WarriorJobMemory*)&p[0]);
+              };
+            case EntityJob.DRK: {
+                return JObject.FromObject(*(DarkKnightJobMemory*)&p[0]);
+              };
+            case EntityJob.GLA: {
+                return null;
+              };
+            case EntityJob.PLD: {
+                return JObject.FromObject(*(PaladinJobMemory*)&p[0]);
+              };
+            case EntityJob.GNB: {
+                return JObject.FromObject(*(GunbreakerJobMemory*)&p[0]);
+              };
+            case EntityJob.BRD: {
+                return JObject.FromObject(*(BardJobMemory*)&p[0]);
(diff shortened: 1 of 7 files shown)


CactbotOverlay/FFXIVProcessIntl.cs
```diff
@@ -0,0 +1,465 @@
...
+
+      [FieldOffset(0x3E)]
+      public short level;
+
+      [FieldOffset(0x5F)]
+      public short shieldPercentage;
+    }
+    public FFXIVProcessIntl(ILogger logger) : base(logger) { }
+
+    [Serializable]
```
KattTails: FFXIVProcess.cs has dummy entries for all of the JobMemory structs and I've defined them in each version as although they're identical at the moment a future patch could change them and it saves work splitting them later on.

ShadyWhite:
I am curious how it recognizes the CN client. If it is through the option of FFXIV_ACT_Plugin, this may not work, because FFXIV_ACT_Plugin does not officially support the CN client. We are using a modified version, which cannot identify the client as the "cn".

KattTails:
The ACT plugin has an option for CN: 
image

ShadyWhite:
Yes, the Chinese option was added in version 2.0.4.2, it works for 5.1 clients, not 5.01.  We are using a modified version 2.0.4.0.  But I don't think this is a problem, there is no need to modify your code, because I think this code can work with version 2.0.4.2 ACT Plugin after CN client update to 5.1.  On behalf of the Chinese players, I thank you very much for your efforts!

quisquous:
Sorry, I merged the other one! Can you update this to not include it, and I can merge this one too?

I think the only other comment I have is that in the short term we could also check the ffxiv act dll version and if it's the old 2.0.4.0 then also use the CN version.

KattTails:
> Sorry, I merged the other one! Can you update this to not include it, and I can merge this one too?

I had to wrangle git a bit, but I *think* I've managed it. 

> I think the only other comment I have is that in the short term we could also check the ffxiv act dll version and if it's the old 2.0.4.0 then also use the CN version.

I've added that check too, good idea!

ShadyWhite:
I merged your codes from your repo to my fork and built a plugin.But still get the error of GetJobSpecificData.
image
image

KattTails:
I just pushed a new build that ouputs which version it's targetting. @ShadyWhite, would you be kind enough to try this new build and check the output log to see which version it has detected? 

If it's still using the 'Intl' version, would you possibly be able to upload a copy of the modified FFXIV_ACT_Plugin so I can see what the version string is?
If it *is* using the Cn version, then the memory offsets are wrong somewhere, which I can try and fix instead.


Also as an aside, I just noticed I've implemented a different version checking method to the `VersionChecker` class, which I didn't see before now. Oops. I can remove that, @quisquous, although mine is a bit cleaner than using regexes!

ShadyWhite:
The logs now shows like this.
image
The CN client mainly uses 3 versions of FFXIV_ACT_Plugin modified by 3 developers.
I have uploaded them to Google drive.https://drive.google.com/open?id=1Bzkso05VR9fJL_j2Y8-mafEr5A2Xc7LP

KattTails:
Okay, well the version matches up with what I was testing for on all of those dlls so that's all fine, and in your log it is using the cn version. From your previous screenshot showing the cactbot test ui, the id/pos/rot fields seem to be working, meaning it's only likely this offset is wrong: 
```C#
[FieldOffset(0x18A4)]
public CharacterDetails charDetails;
```
I've gone ahead and written a quick build that will dump the info I need to work this out, which I'll upload here: [CactbotOverlay.zip](https://github.com/quisquous/cactbot/files/3923436/CactbotOverlay.zip). 

Every 10 seconds it will print out a long string of hex that looks like this: 
image

If you could kindly run this version, and then paste that entire line here, along with your class and hp / max hp, I'll be able to update this to work properly.

Let me know if you need anything and thank you for your help!

ShadyWhite:
Strangely, your quick build works fine, it can read my job and HP correctly(ui/jobs works fine), and the data in test are all correct. I think the problem may be in my fork, sorry.
image
Then this is hex, but I think it may be useless for now.
Info: 78856fd4f77f00000000000000000000517203c30000a0411d4ef0400000000060a4cb3e0000000000000000000000005261697975690000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000899c1410000000000000000000000000000000e000002c000104000100010000bf41a0000000000000000000f7d908c3d277a8418537703f00000000806e513d0000803f0000a03fcdcc8c3f0000003f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000601308973f01000000000000000000000000000000000000000000000000000000000000c7c60f40d9cef7340000000000000000f2eaec3fd9cef7340000000000008037205caf3f00008036000000000000000000000000b0aa50a23f0100000000000000000000000000000000000000000000c7c60f40d9cef7340000000000000000f2eaec3fd9cef734000000000000000000000000000000000000000000886fd4f77f00000000000000000000e0e96dd4f77f000000000000000000005040729c3f0100000000000000000000d8e96dd4f77f00000000000000000000c0e96dd4f77f00000000000000000000000000000000000000000000000000000000000000000000ffffffff0000000000000000000000000000000000000000c8e96dd4f77f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d0e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000803f000000000000000000000000000000000000803f000000000000000000000000000000000000803f00000000000000000000000000000000000000000000803f000000000000000000000000000000000000803f000000000000000000000000000000000000803f0000000000000000000000000000000000000000b0e96dd4f77f00000000000000000000b0e96dd4f77f00000000000000000000b0e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f041729c3f0100000000000000000000000000000000000000000000000080bf000080bf000080bf000080bf000000000000000000000000000000000000000100000000000000000000000000000000d8e96dd4f77f00000000000000000000c0e96dd4f77f00000000000000000000000000000000000000000000000000000000000000000000ffffffff0000000000000000000000000000000000000000c8e96dd4f77f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d0e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000803f000000000000000000000000000000000000803f000000000000000000000000000000000000803f00000000000000000000000000000000000000000000803f000000000000000000000000000000000000803f000000000000000000000000000000000000803f0000000000000000000000000000000000000000b0e96dd4f77f00000000000000000000b0e96dd4f77f00000000000000000000b0e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f041729c3f0100000000000000000000000000000000000000000000000080bf000080bf000080bf000080bf010000000000000000000000000000000000000100000000000000000000000000000000d8e96dd4f77f00000000000000000000c0e96dd4f77f00000000000000000000000000000000000000000000000000000000000000000000ffffffff0000000000000000000000000000000000000000c8e96dd4f77f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000d0e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000803f000000000000000000000000000000000000803f000000000000000000000000000000000000803f00000000000000000000000000000000000000000000803f000000000000000000000000000000000000803f000000000000000000000000000000000000803f0000000000000000000000000000000000000000b0e96dd4f77f00000000000000000000b0e96dd4f77f00000000000000000000b0e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000f041729c3f0100000000000000000000000000000000000000000000000080bf000080bf000080bf000080bf02000000000000000000000000000000000000010000000000000000000000000000000036c26c3d030000000100000000000000105e6fd4f77f00000000000000000000385c6fd4f77f00000000000000000000305c6fd4f77f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000305c6fd4f77f00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002048729c3f01000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000b05c6fd4f77f0000000000000000000000000000000000000000000000000000000000000000000000000000000000002048729c3f010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000080bf000080bf000080bf000080bf000080bf00000000000000000000000000000000000000000000000000000000185c6fd4f77f0000000000000000000000000000000000000000000000000000185c6fd4f77f000031a7b23eea6ab13e418a963e9a26613e05000000000000000000000000000000485c6fd4f77f000000000000000000002048729c3f010000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000285c6fd4f77f00002048729c3f010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000405c6fd4f77f0000204b729c3f010000105d6fd4f77f0000204b729c3f010000205c6fd4f77f000005200000052000000000000000000000f85d6fd4f77f0000583b34d23f01000000000000000000002048729c3f0100005040729c3f010000000000000000000000000000000080bf0000803f00000000896e2c3f000000000000000000000001000000000000000080e96dd4f77f0000000000000000000078e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000803f000000000000000000000000000000000000000001000000ffffffffff0000000000000000000000706b42a23f0100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000700c0000000000000000000000000000000000000000000000000000700c00000000000000000000000000000000000000000000000000006f0c02000000000000000000000000005b0200000000000090007300700c00000000000000000000000000000000000000000000000000006f0c700c0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f0000803f000000005040729c3f010000d0f8c4af3f010000020000000100000000000000000000000000000000000000a8e96dd4f77f0000000000000000000078e96dd4f77f0000000000000000000000000000000000000000000000000000000000000000803f0000000000000000000000000000000000000000ffffffffffffffffff0000000000000000000000a0e96dd4f77f0000000000000000000000000f031703140300000000000000000000000000001b0319031a03180300005040729c3f010000000000000000000000000000000000000000000000000000c8696fd4f77f00005040729c3f010000000000006c000000000000e0000000000004000000000000000000000000000090fd83d4f77f000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000078fd83d4f77f00000000000000000000000000000000000000007043000070436bc2533f00000000000000e000000000000000e0000000000000000000000000000000000000017f0a00000000000000000000e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000f253729c3f010000400000000000000001000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000003000000000000006a54729c3f01000040000000000000000100000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000005040729c3f0100000000000000000000000000000000000000000000000000000000000000000000e9034c0001000000a08903973f010000e0d86dd4f77f0000a08903973f010000601308973f010000901893bb3f010000747332000900000005000000000000000100000001000000cdcc8c3fcdcc8c3f0000803f000000000000803f02000000030400000000000000000000000000000000000000000000e0d86dd4f77f000000000000000000000000000000000000000000000000000000000000ffffffff200000000000000000000000000000000000803f0000803f0000803f000000000000803f12000000010000000000000000000000000000000000000000000000e0d86dd4f77f000000000000000000000000000000000000000000000000000000000000ffffffff000000000000000000000000000000000000803f0000803f00000000000000000000803f12000000010000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000803f0000803f15020100cb17016515020100150201001502010055000100350001006e0001003500010035000100000000000000000000000000000000000000000000000000000f0000000000000000000000000000000000e00000000000000000000000000000000000000000000000e0000000009000849c3f01000000000000000000009000849c3f0100000000000000000000505d12c93f0100000000000000000000070001640d020380993ba06c7707033b02010202026404000135e8b5a4e9acbc000000000000000000000000000000001bb20f00940000000000000000000000000000e000000000000000e00000000000000000000000002870463f000000000000000060c47cbf0000000000000000000000000000000000000000000000000000000000000000000000000000803f0000000000000000000000000000000000000000000000000000000000000000000000e0000000001304130400000000fd5300005c1701005c170100102700001027000000005906000000000000ca01000000008e00ffffffff030022000000000000000000000000000100190050001b000000010002000000000000000000030b00ffff000000000000000100000000000000
BLM 71516/71516

KattTails:
Yes, it would seem the problem is in your fork if this build works fine. No worries, and thanks for all your help regardless!

ShadyWhite:
Thanks a lot for your awesome work!

quisquous:
Thanks for all this work.  :)

PR #828, opened 2019-12-05
Fixed eureka NM tracker names to match ffxiv-eureka.com

logarrhythmic:
Tried using the eureka overlay in Pagos today, it was nice but a few of the NMs on cooldown weren't picked up when I echo'd the list from the tracker. This commit fixes the tracker names to match what's currently output by the relevant button on ffxiv-eureka.com.


quisquous:
Thanks!

PR #837, opened 2019-12-07
plugin: level fix

KattTails:
Hiya,

Another bug introduced during #750. 😭 
Level should be an 8-bit int, not 16. Was causing undefined errors in jobs, and probably a lot more..


quisquous:
Thanks for the fix! I hadn't seen this causing undefined errors, but I could see how it might create some issues.

KattTails:
The new GCD Calcs in jobs.js were giving undefined errors, it's how I noticed it.

Issue #834, opened 2019-12-06
Eureka: log spammed after leaving eureka


KattTails:
## Description

The OverlayPlugin log is spammed with 
```js
Uncaught TypeError: Cannot read property 'primaryWeather' of undefined
    at EurekaTracker.UpdateTimes (eureka.js:2122)
    at EurekaTracker.<anonymous> (eureka.js:2074)
```
every second after leaving Eureka Pyros. I was in `Limsa Lominsa Lower Decks` when I first noticed it, but I'd been to various zones before then. 

## Additional information

I've been in and out of various Eureka zones all day and noticed this was being spammed in my log. I had this a few days ago too but wrote it off as a one time bug.

It seems like `updateTimesHandle` ends up assigned with the wrong handle (mine was assigned `61`). 
I was able to manually end the interval loop with `window.clearInterval(1)`, which was a lucky guess, it seems.

I'm testing a workaround, and if it appears to work I'll submit a PR with the change.

KattTails:
fix was merged :)

PR #841, opened 2019-12-10
eureka: fix log error `Uncaught TypeError: Cannot read property 'primaryWeather' of undefined`

KattTails:
Seems to mitigate #834..


quisquous:
Good catch, thanks!

PR #846, opened 2019-12-12
radar: add default monster name

Bluefissure:
"forgiven pedantry" doesn't have a cn name so take en for default.


quisquous:
Sorry about breaking this!

Issue #550, opened 2019-08-17
jobs: add buff countdowns


quisquous:
This is mostly a reminder to myself that I want to add countdowns for used buffs, and not just active buffs to the jobs module.

I'm thinking UI wise, that maybe they'll just exist on the same horizontal line, but with a little bit of separation.  Maybe the countdown is red instead of white? Maybe a different color box outline, like a thicker dark grey for "not ready yet" and a glowy "ready" one?  I might have to play around with this.

I've gotten feedback that nobody likes buffs on the left side of the hp (and this doesn't make sense for folks only using the buffs (via option).  I could add another horizontal line for other buffs, but maybe that would be visually confusing.  Maybe some experimentation needed here.

PR #855, opened 2019-12-14
Add BigBuffShowCooldownSeconds to jobs-example.js

anubiann00b:
Tiny change, adds the new option in the jobs module added in 6a0c7a4b to the jobs example file.

see #550

### DIFF
--- a/user/jobs-example.js
+++ b/user/jobs-example.js
@@ -41,7 +41,7 @@ Options.BigBuffTextHeight = 0;
 Options.BigBuffBorderSize = 1;
 
 // The minimum time on a cooldown before it is shown.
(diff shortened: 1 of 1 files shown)


user/jobs-example.js
```diff
@@ -40,6 +40,9 @@ Options.BigBuffTextHeight = 0;
 // Size of the big buff icon's colour border.
 Options.BigBuffBorderSize = 1;
 
+// The minimum time on a cooldown before it is shown.
+Options.BigBuffShowCooldownSeconds: 20;
```
quisquous: I think this should be `= 20` and not `: 20`?

quisquous:
Fixed the error and landed this.  Thanks!

PR #858, opened 2019-12-15
Add missing mid dot in the cast RegExp

beckyeeky:
China server supports the mid dot in ID.
Not sure whether we shall use "/u00b7" here instead.


quisquous:
> China server supports the mid dot in ID.
> Not sure whether we shall use "/u00b7" here instead.

Either are fine!

quisquous:
Thanks for the fix!

PR #865, opened 2019-12-18
raidboss: fixed regex for escape prohibition/detection

ffxiv-dev:


quisquous:
Ohmygosh, sorry about this! Thanks for the fix.

PR #869, opened 2019-12-21
plugin: add support for korean ffxiv v5.01

KattTails:
Hiya, 

Quick addition of an extra `FFXIVProcessKo` to support the recent Korean release. It's untested as I don't have access to a Korean build but I'm hoping the offsets are unchanged.

### DIFF
--- a/CactbotOverlay/CactbotEventSource.cs
+++ b/CactbotOverlay/CactbotEventSource.cs
@@ -189,7 +189,8 @@ namespace Cactbot {
         LogInfo("Language: {0}", language_);
       }
 
-      if (language_ == "cn") {
+      // Temporarily target cn if plugin is old v2.0.4.0
+      if (language_ == "cn" || ffxiv.ToString() == "2.0.4.0") {
         ffxiv_ = new FFXIVProcessCn(this);
         LogInfo("Version: cn");
       } else if (language_ == "ko") {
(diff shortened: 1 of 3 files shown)


CactbotOverlay/CactbotEventSource.cs
```diff
@@ -189,10 +189,12 @@ public override void SaveConfig(IPluginConfig config)
         LogInfo("Language: {0}", language_);
       }
 
-      // Temporarily target cn if plugin is old v2.0.4.0
-      if (language_ == "cn" || ffxiv.ToString() == "2.0.4.0") {
+      if (language_ == "cn") {
```
KattTails: I removed the hard check on the version string since new releases of the FFXIV Plugin have cn support.

quisquous: But does the ffxiv plugin that the cn and ko players are using have that yet? I thought they were all using an older version.

KattTails: Oh, I was falsely assuming they were backwards-compatible without really thinking it through. I'll re-add that line for cn for now.

I wonder if there's anyone who knows which version of the plugin the Korean players are using?

Bluefissure: @devunt Hi, can you help this about the Korean act plugin version?

Bluefissure: @devunt Hi~ Can you help with this Korean act plugin version problem?

quisquous: I talked to @hibiyasleep who said that they used the 2.0.4.0 version directly (with no translations).  I'm not sure that there's any way to differentiate the language then until the ffxiv plugin is new enough to include cn and ko language options.

So, maybe this is good enough for now.

PR #871, opened 2019-12-22
raidboss: Initial A6S Triggers

panicstevenson:
These could always be improved upon; they were taken from a combination
of triggers from guides, a solo run of A6S that I couldn't complete
alone, existing triggers, and video footage.


quisquous:
Thanks! I feel like the alexander raids are probably the area that could use the most timeline and trigger attention at the moment, so this is much appreciated.

Issue #749, opened 2019-11-12
Util: Improvements to translation tools


JLGarber:
Per some discussion on the ARF trigger additions in #744 , it looks like updates to the translation scripts would be in order. Given that the next major patch is coming up in a month or so, I'll probably prioritize this in preparation.

Primarily, I'm looking to update `translate_fight` to add a separate translation path that would look something like this:

1. Parse the timeline file for ability IDs and boss names.
2. Issue requests to XIVAPI to pull down the appropriate translations.
3. Consolidate the data and merge it into the existing JSON output stream. (Or maybe take a trigger file as optional input and automagically append the JSON output in the right place? I'm not as comfortable with that sort of file manipulation yet, but that might be ideal.)

The clear advantage to this over FFLogs translation is that this would just pull down the needed abilities based on the timeline, with no need to cull superfluous elements. The downside is that XIVAPI (so far as I'm aware) doesn't contain CN localizations, so we would have to pull those in from elsewhere.

I think the recent changes to regex handling will probably negate the need for the trigger updates portion of `check_translation` going forward.

I'm not sure what other features would be necessary or useful while doing these updates, so please let me know what else I should consider while planning this.

quisquous:
https://cafemaker.wakingsands.com/ is the xivapi equivalent with CN support: https://cafemaker.wakingsands.com/Action/3?pretty=true

JLGarber:
Oh, thank you. I didn't know about that. Same format and everything. Now to read a little on Python's handling of requests...

ngld:
If you're new to HTTP requests in Python, I'd very much recommend the [`requests` module](https://requests.kennethreitz.org/en/master/).
It makes your life so much easier compared to `urllib`.

JLGarber:
Ah, that would be significantly easier. I looked through `urllib` a little last night, and it was... not incomprehensible, but this is way better.

Thinking through things a little more, for parsing the timeline file, is there any reason not to just look for the sync regex per line? Obviously older timelines are a little messier, but I think it's safe to assume any future timelines will include syncs for every relevant ability. Given this, what other assumptions can we make about the sync regex? I'm sort of thinking we should start with this list:

1. `00`, usually as `00:0839:[ZONE] will be sealed off`.
2. `03`, as `03:........:Added new combatant [NAME]`
3. `15/16`, as `:[NAME]:[HEX ID]:`.

[...]

PR #872, opened 2019-12-22
UTIL: Updating translate_fight.py to use XIVAPI

JLGarber:
Alright, here's a draft for this. There's a lot of bad in here, but it's minimally functional and is faster than the FFLogs functionality that existed previously. As with the updates to `make_timeline` and `test_timeline`, this doesn't change previously-existing functionality except as necessary to add new stuff.

Currently, because I'm unsure of Cafemaker's search endpoint, this doesn't include support for Chinese translations. 

This is the German output for Copied Factory:

```
 {
   'locale': 'de',
   'replaceSync': {
     'small biped': 'klein[a] Zweibeiner',
     'serial-jointed service model': 'Modell[p] mit Omnigelenk',
     'serial-jointed command model': 'Befehlsmodell[p] mit Omnigelenk',
     'reverse-jointed Goliath': 'Goliath[p] mit Inversgelenk',
     'multi-leg medium model': 'mittelgroß[a] mehrbeinig[a] Modell',
     'medium exploder': 'mittelgroß[a] Selbstzerstörung',
     'flight unit': 'Flugeinheit',
     'Warehouse C': 'Warenlager C',
     'Warehouse B': 'Warenlager B',
     'Warehouse A': 'Warenlager A',
     'Quality Assurance': 'Warenkontrollhalle',
     'Marx': 'Marx',
     'Hobbes': 'Hobbes',
     'Goliath tank': 'Goliath-Panzer',
     'Engels': 'Engels',
     'Engage!': 'Start!',
     '9S-operated walking fortress': '9S\' mehrbeinig[a] Panzer',
     '9S-operated flight unit': '9S\' Flugeinheit',
   },
   'replaceText': {
     'Winds of Tartarus': 'Wind des Tartarus',
     'Wide-angle Diffuse Laser': 'Schwerer Diffusionslaser',
     'Undock': 'Abdocken',
     'Total Annihilation Maneuver': 'Offensive: Totale Vernichtung',
     'Throat Stab': 'Halsstich',
     'Systematic Targeting': 'Jagdformation',
     'Systematic Suppression': 'Artillerieformation',
     'Systematic Siege': 'Kesselformation',
     'Systematic Airstrike': 'Luftformation',
     'Surface Missile': 'Raketenschlag',
     'Spirit Dart': 'Geistpfeil',
     'Shrapnel Impact': 'Wrackteilregen',
     'Short-range Missile': 'Kurzstreckenrakete',
     'Shockwave': 'Schockwelle',
     'Ring Laser': 'Ringlaser',
     'Radiate Heat': 'Thermaloffensive',
     'Precision Guided Missile': 'Schwere Lenkrakete',
     'Oil Well': 'Ölaustritt',
     'Neutralization': 'Unterwerfung',
     'Marx Thrust': 'Marxscher Ansturm',
     'Marx Smash': 'Marxscher Schlag',
     'Marx Impact': 'Marxscher Sturz',
     'Marx Crush': 'Marxsche Offensive',
     'Marx Activation': 'Marx-Aktivierung',
     'Lightfast Blade': 'Lichtklingenschnitt',
     'Laser-resistance Test': 'Laserresistenztest',
     'Laser Turret': 'Hauptgeschützlaser',
     'Laser Sight': 'Laserbestrahlung',
     'Laser Saturation': 'Omnidirektionallaser',
     'Incendiary Saturation Bombing': 'Streubrandraketen',
     'Incendiary Bombing': 'Brandraketen',
     'High-powered Laser': 'Hochleistungslaser',
     'High-frequency Laser': 'Hochfrequenzlaser',
     'High-caliber Laser': 'Großkaliberlaser',
     'Hack Goliath Tank': 'Hacken: Goliath-Panzer',
     'Guided Missile': 'Lenkraketen',
     'Ground-to-ground Missile': 'Boden-Boden-Rakete',
     'Frontal Somersault': 'Sprungoffensive',
     'Freefire': 'Schwerer Beschuss',
     'Forceful Impact': 'Heftiges Beben',
     'Flash': 'Blitzlicht',
     'Fast Blade': 'Vortexschnitt',
     'Enrage': 'Finalangriff',
     'Engage Marx Support': 'Verstärkung: Marx',
     'Engage Goliath Tank Support': 'Verstärkung: Goliath-Panzer',
     'Energy Ring': 'Omnidirektionalenergie',
     'Energy Bombardment': 'Energiemörser',
     'Energy Blast': 'Energetische Explosion',
     'Energy Barrage': 'Energetisches Sperrfeuer',
     'Energy Assault': 'Energieschauer',
     'Diffuse Laser': 'Diffusionslaser',
     'Demolish Structure': 'Terraintilgung',
     'Crushing Wheel': 'Zangenradoffensive',
     'Convenient Self-destruction': 'Selbstsprengung',
     'Clanging Blow': 'Schwerer Angriff',
     'Ballistic Impact': 'Raketeneinschlag',
     'Arm Laser': 'Armlaser',
     'Area Bombing Maneuver': 'Offensive: Raketensalve',
     'Area Bombardment': 'Blindraketen',
     'Anti-personnel Missile': 'Antipersonenrakete',
     '360-degree Bombing Maneuver': 'Offensive: Raketenring',
     '--untargetable--': '--nich anvisierbar--',
     '--targetable--': '--anvisierbar--',
   },
   '~effectNames': {},
 },
```

Apart from the speed, another benefit here is that only relevant items will be translated. This has been one thing that's annoyed me about using FFLogs to translate, so this helps out in that regard.

Given how long this has been in progress, I'm not even sure what else to say about it. Criticize away and let's make this usable!

### DIFF
--- a/util/encounter_tools.py
+++ b/util/encounter_tools.py
@@ -142,7 +142,7 @@ def is_tl_line_begincast(poss_match):
     return re.search(r'^:([0-9A-F\(\)\|]+):([^:]+)', poss_match)
 
 def is_tl_line_buff(poss_match):
-    return re.search(r'1A:.......:(.+) gains the effect of (.+)( from)?', poss_match)
+    return re.search(r'1A:........:(.+) gains the effect of (.+)( from)?', poss_match)
 
 def is_tl_line_cast(poss_match):
     return re.search(r'^:([^:]+):([0-9A-F\(\)\|]+)', poss_match)
--- a/util/encounter_tools.py
+++ b/util/encounter_tools.py
@@ -132,8 +132,11 @@ def choose_fight_times(args, encounters):
 
 
 # Timeline test/translate functions
-def clean_and_split_tl_line(line):
-    return re.search(r'^(?P<time>[\d\.]+)\s+"(?P<label>[^"]+)"\s+(?P<options>.+)', line.split('#')[0])
+def clean_and_split_tl_line(line, split):
+    to_parse = line
+    if split:
+        to_parse = line.split('#')[0]
+    return re.search(r'^(?P<time>[\d\.]+)\s+"(?P<label>[^"]+)"\s+(?P<options>.+)', to_parse)
 
 def is_tl_line_syncmatch(line):
     return re.search(r'sync /([^/]+)/', line.group('options'))
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -21,7 +21,7 @@ def load_timeline(timeline):
             entry = {}
             # Remove trailing comment, if any,
             # then split the line into sections
-            match = e_tools.clean_and_split_tl_line(line)
+            match = e_tools.clean_and_split_tl_line(line, True)
             if not match:
                 continue
 
@@ -424,12 +424,12 @@ def timeline_file(filename):
     # Allow for just specifying the base filename, e.g. "o12s.txt" or "o12s"
     if not os.path.exists(filename):
         for root, dirs, files in os.walk(data_path):
-          if filename in files:
-            filename = os.path.join(root, filename)
-            break
-          if '%s.txt' % filename in files:
-            filename = os.path.join(root, '%s.txt' % filename)
-            break
+            if filename in files:
+                filename = os.path.join(root, filename)
+                break
+            if '%s.txt' % filename in files:
+                filename = os.path.join(root, '%s.txt' % filename)
+                break
 
     path = Path(filename)
     if not path.exists():
--- a/util/translate_fight.py
+++ b/util/translate_fight.py
@@ -5,15 +5,15 @@ from collections import defaultdict
 import json
 import re
 import sys
-import encounter_tools as e_tools
-import requests
 import time
-
+import requests
 import fflogs
+import encounter_tools as e_tools
+
 
 # 'en' here is 'www' which we consider the "base" and do automatically.
 # 'cn' exists on fflogs but does not have proper translations, sorry.
-languages = ['en', 'de', 'fr', 'ja', 'cn']
+fflogs_languages = ['en', 'de', 'fr', 'ja', 'cn']
 prefixes = {
     'en': 'www',
     'de': 'de',
@@ -43,7 +43,7 @@ def find_timeline_translatables(args):
     translatables = defaultdict(dict)
     # These keys do double duty as the indices for XIVAPI's search function.
     for element in ['action_ids', 'status', 'bnpcname', 'PlaceName']:
-        translatables[element] = []   
+        translatables[element] = []
 
     with args.timeline as file:
         for line in file:
@@ -52,7 +52,7 @@ def find_timeline_translatables(args):
                 continue
 
             # Boil each remaining line down to its sync regexes
-            cleaned_line = re.search(r'^(?P<time>[\d\.]+)\s+"(?P<label>[^"]+)"\s+(?P<options>.+)', line)
+            cleaned_line = e_tools.clean_and_split_tl_line(line, False)
 
             # If for some reason the line doesn't clean up properly, just skip it.
             if cleaned_line is None:
@@ -278,6 +278,7 @@ def main(args):
         languages = xiv_langs
 
     else:
+        languages = fflogs_languages
         # Get overall reports and enemies.
         enemies = defaultdict(dict)
         options = {
--- a/util/translate_fight.py
+++ b/util/translate_fight.py
@@ -5,6 +5,9 @@ from collections import defaultdict
 import json
 import re
 import sys
+import encounter_tools as e_tools
+import requests
+import time
 
 import fflogs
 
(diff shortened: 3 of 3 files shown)


Akurosia:
German will posible always require manual work for all cases where it contains the `[A]` inside.
For the `[P]` this could be just removed as i never saw a case so far where this needed to be replaced with any character.

JLGarber:
I'm almost entirely ignorant of German, is that the special-cased final characters, in the same way Greek has a different letter Sigma at the end of words? If so, is there consistency to it, or is it so context-dependent that it would require the manual work you mentioned?

Akurosia:
currently i could not point out a rule how to determin what the `[a]` will be translated to.
Best example are following to cases:
'mittelgroß[a] mehrbeinig[a] Modell', => 'mittelgroßes mehrbeiniges Modell',
'mittelgroß[a] Selbstzerstörung', => 'mittelgroße Selbstzerstörung',

As you can see, "mittelgroß[a]" can be results in two idfferent cases.
Might should get back to my old schoolbools to check if there is an easy to explain rule xD

util/encounter_tools.py
```diff
@@ -132,8 +132,11 @@ def choose_fight_times(args, encounters):
 
 
 # Timeline test/translate functions
-def clean_and_split_tl_line(line):
-    return re.search(r'^(?P<time>[\d\.]+)\s+"(?P<label>[^"]+)"\s+(?P<options>.+)', line.split('#')[0])
+def clean_and_split_tl_line(line, split):
+    to_parse = line
+    if split:
+        to_parse = line.split('#')[0]
```
quisquous: For my own understanding, why do you only conditionally remove the comment here?

(Also, this was wrong before your change here, but this code is probably inaccurate for things like `12.0 "Puddle #3" sync /thing/`)

util/test_timeline.py
```diff
@@ -424,12 +424,12 @@ def timeline_file(filename):
     # Allow for just specifying the base filename, e.g. "o12s.txt" or "o12s"
     if not os.path.exists(filename):
         for root, dirs, files in os.walk(data_path):
-          if filename in files:
-            filename = os.path.join(root, filename)
-            break
-          if '%s.txt' % filename in files:
-            filename = os.path.join(root, '%s.txt' % filename)
-            break
+            if filename in files:
```
quisquous: Huh! I would have thought that pylint would have been mad at this.  Thanks for fixing it.

util/translate_fight.py
```diff
@@ -5,12 +5,15 @@
...
 import sys
-
+import time
+import requests
 import fflogs
+import encounter_tools as e_tools
+
 
 # 'en' here is 'www' which we consider the "base" and do automatically.
 # 'cn' exists on fflogs but does not have proper translations, sorry.
```
quisquous: This comment is maybe no longer relevant?

util/translate_fight.py
```diff
@@ -28,6 +31,145 @@
...
+                continue
+
+            sync_match = e_tools.is_tl_line_syncmatch(cleaned_line)
+
+            # If we don't have a match for whatever reason, ignore the line.
+            if sync_match is None:
+                continue
+
+            # We don't want to catch the reset line.
+            if sync_match.group(1).startswith('is no longer sealed'):
```
quisquous: Why do you want to ignore this? Is there some larger category of things that should be ignored than just "is no longer sealed"? Also, why would you not include zone unsealing in terms of looking for zone names as you do with zone sealing below?

util/encounter_tools.py
```diff
@@ -142,7 +145,7 @@ def is_tl_line_begincast(poss_match):
     return re.search(r'^:([0-9A-F\(\)\|]+):([^:]+)', poss_match)
 
 def is_tl_line_buff(poss_match):
-    return re.search(r'1A:.......:(.+) gains the effect of (.+)( from)?', poss_match)
+    return re.search(r'1A:........:(.+) gains the effect of (.+)( from)?', poss_match)
```
quisquous: Will `(.+?)` fix this?

util/translate_fight.py
```diff
@@ -28,6 +31,145 @@
...
+
+            add_match = e_tools.is_tl_line_adds(sync_match.group(1))
+            if add_match:
+                if not add_match.group(1) in translatables['bnpcname']:
+                    translatables['bnpcname'].append(add_match.group(1))
+    return translatables
+
+def compile_and_send(translatables):
+    # XIVAPI/cafemaker allows for returning multiple items in one request if the item IDs are known.
+    IDstr = ",".join(translatables['action_ids'])
```
quisquous: naming: `IDstr` => `id_str`

util/translate_fight.py
```diff
@@ -28,6 +31,145 @@
...
+
+    # These columns will be used for every XIVAPI request, so we alias them now.
+    columns = 'Name_en,Name_de,Name_fr,Name_ja'
```
quisquous: Can you build this from `xiv_lang_map` values?

util/translate_fight.py
```diff
@@ -28,6 +31,145 @@
...
+    # TODO: Look up IDs in St Coinach so we can collapse all this into one request per type?
+    # Nested loops are just THE BEST
+    for element in translatables:
+        working_dict = finished_items[element]
+        if element is 'PlaceName':
+            working_dict = finished_items['bnpcname']
+        for item in translatables[element]:
+            # To avoid being timed out by the server, we manually slow down here.
+            # TODO: Allow for users to include their XIVAPI developer key to increase this limit.
+            if requestcount % 11 is 0:
```
quisquous: Is it possible to make an object that wraps `requests` and does this sleeping automatically?

util/encounter_tools.py
JLGarber: Because that specific function serves two purposes depending on that flag. It's also used in `test_timeline` to strip comments entirely, and for that purpose we specifically would want syncs to be ignored. Or at least that was the process that led me to setting it up that way. It's reasonable to re-split that function back out, I suppose.

That's a good point on in-line # symbols though. Ordinarily I would prefer to just discard that line since it's clearly an author add-in and not a game skill, but there could very well be abilities that contain the symbol. Perhaps check whether the symbol is inside quotes and then split based on that?

util/test_timeline.py
JLGarber: I wouldn't normally have done that in this sequence of requests, but given that I already had to make a mostly-cosmetic change, I figured it was reasonable to toss this in as well.

util/translate_fight.py
JLGarber: Perhaps? Given that we aren't actually removing the FFLogs functionality I would be hesitant to take this out as well.

util/translate_fight.py
JLGarber: That's a relic from when the 0.0 reset line was giving me problems. It looks like that shouldn't be an issue anymore, so you're definitely correct it's best to just remove it.

(In practice, since resets just sync to `/is no longer sealed/` now, it's moot, but you're right, there's no reason to avoid those lines in general, since we already de-duplicate entries in-line.)

util/encounter_tools.py
JLGarber: I don't think so? I might have an error somewhere then. As it stands, it looks like the current regex [captures "[EFFECT] from [SOURCE] for [FLOAT] seconds"]( https://regex101.com/r/LPsWtX/2/). If I'm understanding correctly, [that change would capture only the first letter of [EFFECT]](https://regex101.com/r/0o1aC8/3/).

It would definitely be nice to extract that cleanly with a regex rather than having to run an additional string operation!

util/translate_fight.py
JLGarber: Good point, one less configuration update location.

util/translate_fight.py
JLGarber: Probably? I've done nothing with Python wrappers before but I can always learn. (Given how things are going though, that might end up being something I add in later due to time constraints.)

JLGarber:
I had hoped to avoid special-casing Chinese when doing this, but it looks like that may be necessary, based on the discussion in the linked issue. (On the plus side, even if we do have to get the IDs from XIVAPI first, there won't be much of a speed penalty. It's only going to add three extra requests to the total, rather than doubling the number, since we can concatenate all the IDs per category as we did for Actions.) The primary downside would be needing to rework the request initialization and having to add a flag for whether or not to attempt Chinese translation.

util/encounter_tools.py
quisquous: It's because `from` is optional.  https://regex101.com/r/0o1aC8/4

util/translate_fight.py
quisquous: I meant wrapper in a colloquial sense.  I was thinking that you could make some Python class that either extends `requests` (or has-a `requests`) that you can call get on and hide the `requestcount` inside that class.

util/encounter_tools.py
quisquous: Is there a need to comment out translated syncs? Really my question comes down to why is there conditional behavior? It seems like you'd want to remove comments in all cases?

JLGarber: Oops, that's on me, I forgot to link the related issue when I created this request. [It's to ensure that even if there are abilities that don't get synced *in the timeline* we can still translate them.](https://github.com/quisquous/cactbot/issues/749#issuecomment-557830256) But if there's a better way to handle that situation I'd much prefer that.

util/encounter_tools.py
JLGarber: Hmm... that seems like it might work. I'll look a little more after TEA.

util/translate_fight.py
JLGarber: Mm. Possibly. I'm still not really sure what that would look like, but again I haven't touched that aspect of Python before. Always more to learn!

util/encounter_tools.py
quisquous: Oh right, we need to translate those so that we can find the names of the abilities in the timeline that they refer to.  Sorry for being forgetful.  (Maybe that can be a comment so that future me can remember.)

Style-wise, I'd rather have two differently named functions than one with a boolean parameter that adjusts logic.  It's just harder to follow the calling code to know what that boolean is doing, vs reading a more explanatory function name.

quisquous:
Also, just throwing out more complication, but @Akurosia linked me to https://github.com/Ra-Workspace/ffxiv-datamining-ko for Korean translations.  I don't think there's an xivapi similar thing, so this is more along the lines of a Saint Coinach dump.  Feel free to ignore this followup, because it adds complication, but just something to think about that I wanted to share.

util/encounter_tools.py
JLGarber: That's reasonable. I think I originally combined it when we made the changes to `test_timeline`, but you're right, it is better to separate them now that other stuff needs to use that regex.

util/encounter_tools.py
JLGarber: I think maybe I misunderstood where we were trying to go with this. What you linked there would match only case 1:
1. `1A:........:Thancred gains the effect of Regen from Alphinaud for 15 seconds`
2. `1A:........:Thancred gains the effect of Regen from Alphinaud`
3. `1A:........:Thancred gains the effect of Regen from`
4. `1A:........:Thancred gains the effect of Regen`

Ideally, (I think) we want to be able to match cases 2 and 4 here. (Case 3 should never show up in a timeline, and it's going to be rare enough to see case 1 in a timeline that I think it's worth assuming that will just be done manually.) If possible, it would be nice to extract the effect cleanly from 2 as well, but if we can't both match and extract cleanly in a single regex it's probably not too big a deal.

(Looking at this more, it seems that trying to do what I'm getting at here with matching both 2 and 4 would involve some unpleasant manipulation of lookaheads, and I'm thinking that probably isn't worth the effort.)

JLGarber:
That's a useful resource. I'm not familiar with programmatic access to St Coinach, but raw CSV is "easy enough". Given that we're already going to be pulling IDs for everything for Chinese translations, I can sort of visualize a path for the Korean items as well.

Seeing how drastically things have to change, I'll probably close this request sometime this week, send in another one to resolve the issues with `clean_and_split_line`, and continue on from there. I do still think it should be possible to get this ready for 5.2, or at least for the languages available on XIVAPI.

ngld:
FWIW XIVAPI uses `.csv` files internally as well. You can either [grab them](https://github.com/xivapi/ffxiv-datamining/tree/master/csv) and (assuming we can find the files for cafemaker as well) use `.csv` files for everything or use the XIVAPI source to provide an API to the korean `.csv` files.

JLGarber:
Hmm. Definitely something to consider. I think we'll have to go with a combination of data sources regardless, but it's nice to have options.

JLGarber:
Not dead yet! Things have been a lot busier than I expected. At this point I don't expect to get Chinese/Korean support completed until after the 5.2 patch, which should be fine given the delays those versions experience compared to EN/DE/FR/JA. Short-term, I'm hoping to integrate most of the review items sometime in the next week or two. We should still be on track for the 5.2 release if all goes well.

JLGarber:
Welp. That wasn't quite as much of a success as I had hoped for. Ultimate took my group all the way up to the final week, and then Savage has been an adventure this week. I'll be getting back to this in the next week, but obviously there have been changes to some things in the meantime so I'll have to sort that out. We'll close this for now and open a new one once I have a better idea where to go.

PR #880, opened 2019-12-25
Add config ui

quisquous:
This adds a config "module" to cactbot in `ui/config/config.html` and replaces the built in C# UI with an embedded browser that uses that page directly.

Any cactbot module that wants to specify config options creates a `ui/module/module_config.js` file.  This gets included in that module's html (to apply default options and help convert saved options into runtime `Options`) as well as in the `ui/config/data/manifest.txt` (so the config tool knows which modules exist and need to be loaded).  This creates at least a mild amount of separation between the config code and all of the different modules.

The config tool uses the `cactbotLoadData` and `cactbotSaveData` to store all of the options in the `%APPDATA%/Advanced Combat Tracker/Config/RainbowMage.OverlayPlugin.config.json` file under CactbotESConfig -> OverlayData -> options -> overlay name.  Thus, each overlay has its own namespace to write options in.

Each `module_config.js` file calls a new `UserConfig.registerOptions` function that registers its set of "option templates" (ids, information to build ui, descriptions, default values, and an optional `setterFunc` to convert from ui to `Options` if it's not simple) globally.  This function takes a dictionary with these properties:
* `options` (the option templates, mentioned above).  Most ids are the exact names of the fields on `Options`, which allows most of this to be automatic.
* `buildExtraUI` (optionally create extra ui, used for raidboss for triggers, and could be used for oopsyraidsy triggers in the future)
* `processExtraOptions` (used to turn any values stored from the extra ui into values in `Options`)

See: "DefaultAlertOutput" in `raidboss_config.js` or `FlagTimeoutSeconds` in `eureka_config.js` for good examples of `setterFunc`

In summary, the startup to set options for each module at runtime is:
* (new) `UserConfig.registerOptions` to let `UserConfig` know about option templates to apply (which happens automatically by including the `overlay_config.js` file)
* `UserConfig.getUserConfigLocation` is called for the module
* The language is set from the plugin
* Some default `Options` are set by the `Options` field in the `overlay.js` file itself.
* (new) `UserConfig.processOptions` is called to apply saved options to `Options`
* User files and skins are evaluated (and can override anything previously set)
* The callback passed to `getUserConfigLocation` is called to create the overlay.


quisquous:
I still need to sort out what to do about the user directory (thinking I should just make it always be inside the cactbot folder instead of arbitrarily specified).  I also didn't hook up the "reload files" option, but it appears that maybe it hasn't been hooked up for a while.

I also deliberately did not provide any options for group tts, as I kind of wonder if that should be removed and we should reconsider how to handle that use case.

quisquous:
image

ShadyWhite:
When this UI interface is released, can you add multi-language support？ I can provide Chinese translation.

quisquous:
> When this UI interface is released, can you add multi-language support？ I can provide Chinese translation.

Definitely! It's already there.  All of the text in that picture has localized strings.  If you look at the changes in this pull request, it will work like raidboss does.

ui/config/config.js
```diff
@@ -0,0 +1,280 @@
...
+        for (let j = 0; j < options.length; ++j) {
+          let opt = options[j];
+          if (!this.developerOptions && opt.debugOnly)
+            continue;
+          let buildFunc = {
+            checkbox: this.buildCheckbox,
+            select: this.buildSelect,
+            float: this.buildFloat,
+            integer: this.buildInteger,
+            directory: this.buildDirectory,
```
panicstevenson: Currently it looks like directory is unused; is the intent here for something like specifying custom sound files, or would that be an additional file/string option or is it a remnant of a user-specified user directory option?

quisquous: It's the remnant of a user-specified user directory option.

Today, I tried implementing a directory picker in C#, but didn't have any luck.  `FolderBrowserDialog`, even when I invoked it on the main act form, just didn't show anything when I used it.  `CommonOpenFileDialog` crashed, so I'm not sure what to do there.  @ngld suggested this approach, so I'll ask him for more suggestions.

quisquous:
Ok, added a directory dialog and also made the plugin depend on the settings from the config tool for watching reloaded files and the user directory.

This should be good to go.  Any other comments?

Issue #879, opened 2019-12-25
plugin: add proper ko version support


KattTails:
Just writing this here to keep track of it. Ravaan has released an official Korean version of the FFXIV Plugin, which uses its own versioning and has a different name `FFXIV_ACT_Plugin_Korean.dll`.

There's going to have to be a few changes in the version checking code to accommodate for the naming variation.

PR #884, opened 2019-12-26
plugin: Add new FFXIV_Plugin_Korean.dll support

KattTails:
As per #879 I've added compatibility with the new FFXIV_Plugin_Korean.dll released by Ravahn.

### DIFF
--- a/CactbotOverlay/CactbotEventSource.cs
+++ b/CactbotOverlay/CactbotEventSource.cs
@@ -173,7 +173,7 @@ namespace Cactbot {
       FFXIVPlugin plugin_helper = new FFXIVPlugin(this);
       language_ = plugin_helper.GetLocaleString();
 
-      var versions = new VersionChecker(this, language_);
+      var versions = new VersionChecker(this);
       Version local = versions.GetLocalVersion();
       Version remote = versions.GetRemoteVersion();
 
--- a/CactbotOverlay/CactbotEventSource.cs
+++ b/CactbotOverlay/CactbotEventSource.cs
@@ -172,7 +172,7 @@ namespace Cactbot {
       FFXIVPlugin plugin_helper = new FFXIVPlugin(this);
       language_ = plugin_helper.GetLocaleString();
 
-      var versions = new VersionChecker(this);
+      var versions = new VersionChecker(this, language_);
       Version local = versions.GetLocalVersion();
       Version remote = versions.GetRemoteVersion();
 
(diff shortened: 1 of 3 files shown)


CactbotOverlay/VersionChecker.cs
```diff
@@ -75,7 +77,10 @@ class VersionChecker {
     }
 
     private Advanced_Combat_Tracker.ActPluginData GetFFXIVPluginData() {
-      return GetPluginData("FFXIV_ACT_Plugin.dll");
+      if (language_ == "ko")
+        return GetPluginData("FFXIV_ACT_Plugin_Korean.dll");
+      else
+        return GetPluginData("FFXIV_ACT_Plugin.dll");
```
KattTails: If we have `FFXIV_ACT_Plugin_Korean.dll` but the plugin is falsely set to another language this will fail. Should we catch this?

quisquous: Yes, I think the korean dll check should come first, and if that dll is being used, it should always use the korean memory offsets.

KattTails: I've pushed another build that does this.

CactbotOverlay/CactbotEventSource.cs
```diff
@@ -195,7 +198,7 @@ public override void SaveConfig(IPluginConfig config)
       if (language_ == "cn" || ffxiv.ToString() == "2.0.4.0") {
         ffxiv_ = new FFXIVProcessCn(this);
         LogInfo("Version: cn");
-      } else if (language_ == "ko") {
+      } else if (ffxiv_plugin_region_ == "ko") {
```
quisquous: Should this come before the "cn" check?

CactbotOverlay/VersionChecker.cs
```diff
@@ -10,6 +10,7 @@ namespace Cactbot {
   // https://github.com/anoyetta/ACT.SpecialSpellTimer/blob/master/ACT.SpecialSpellTimer.Core/UpdateChecker.cs
   class VersionChecker {
     private ILogger logger_ = null;
+    private string language_ = "";
```
quisquous: Can this be removed too?

KattTails: Oops.

CactbotOverlay/CactbotEventSource.cs
KattTails: Good idea.

quisquous:
Thanks!!

PR #883, opened 2019-12-26
some kr translation of raidboss (e1n~e4n)

haka5e:
first time using github. sorry for any problems.

PR #887, opened 2019-12-28
add Korean supports to resources/lang, raidboss\data\00-misc

Bing-su:
Thanks for updating cactbot supports Korean. I have tested test.js and it worked well except playing tts and sound file.

and may I ask a question?
Dose it need in translating raidboss triggers to match exactly '~effectNames' with in-game data like 'Physical Vulnerability Up'? I saw translated titania and innocence triggers and find some difference.

### DIFF
--- a/resources/lang/ko.js
+++ b/resources/lang/ko.js
@@ -21,47 +21,47 @@ class CactbotLanguageKo extends CactbotLanguage {
       Medicated: '강화약', // tbc
       BattleLitany: '전투 기도', // 0x312
       Embolden: '성원', // 0x4d7
-      Arrow: 'FIXME', // 0x75c
+      Arrow: '오쉬온의 화살', // 0x75c
       Balance: '아제마의 균형', // 0x75a
-      Bole: 'FIXME', // 0x75b
-      Ewer: 'FIXME', // 0x75e
-      Spear: 'FIXME', // 0x75d
-      Spire: 'FIXME', // 0x75f
-      LadyOfCrowns: 'FIXME', // 0x755
-      LordOfCrowns: 'FIXME', // 0x754
+      Bole: '세계수의 줄기', // 0x75b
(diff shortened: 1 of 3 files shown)


Bing-su:
I think Travis says regex in general ready check has problem. but I don't figure out what... please help.

KattTails:
I think you can drop the brackets so it just reads 
``` js 
regexKo: Regexes.gameLog({ line: '\y{Name} 님이 준비 확인을 시작했습니다\.|준비 확인을 시작합니다\.', capture: false }),
```

Bing-su:
wow thanks a lot

quisquous:
> Dose it need in translating raidboss triggers to match exactly '~effectNames' with in-game data like 'Physical Vulnerability Up'? I saw translated titania and innocence triggers and find some difference.

If there are differences, feel free to adjust them.

Right now, the `timelineReplace` section is used in two ways:
* at runtime, to translate timeline syncs and text using `replaceSync` and `replaceText`
* via `util/check_translation.py`, using `replaceSync`, `replaceText`, `~effectNames` to create lines like `regexKo` that can then get committed.

In the future, I would like to remove `regexKo` and have the `timelineReplace` section do everything at runtime.  This will be easier once everything uses the `regex: Regexes.ability(etc)` style regexes rather than `regex: //` style.

I hope that helps explain.  Feel free to ask more questions if that's confusing or there's anything I can help with.

quisquous:
Also, one thing that definitely would help is to translate the `resources/responses.js` file.  I would like to move to replace more triggers to use those built in responses to help ease the translation burden.

quisquous:
I mentioned this on #883, but also which version of lord/lady of crowns is correct between this pull request and the one in #883?

Bing-su:
thank you for replying. mine is correct for those crowns. 883 guy writes that as ‘draw: lady of crown’.

PR #890, opened 2019-12-29
Config UI FR translations

mooondark:


quisquous:
Such fast translations! Thank you!!

PR #894, opened 2019-12-30
Config UI and responses.js KR translations

Bing-su:
I tried to do this tidily, but I didn't know the reverted also remains on the log haha...


quisquous:
Fixed a merge conflict here so I could get this landed.  Thanks for the translations.

No worries about the log.  When I am making pull requests, I usually try to fix things up with `git rebase -i` and clean up my commit history.

PR #895, opened 2019-12-30
kr translation of raidboss (e1s~e4s) (pre-update)

Jaehyuk-Lee:
v5.05 is not updated in KR server, but I've changed bnpcname/action/status names into Korean for those same strings are already updated in v5.01 or before. (like Titan or Eden)
Also, translated Trigger - player behavior information texts into Korean.

I think all timelines will work because all timeline sync strings are already named in v5.01. (based on DE/FR raidboss data)
Some unknown translations were translated Japanese into Korean for those have high possibility just write Kanji(or hanja in Korean) in Korean alphabet.

- some small changes
1. o8s Korean server compatibility update
2. Oopsyraidsy food buff lost translation into Korean.


quisquous:
Thanks!

PR #896, opened 2019-12-30
fix cn FIXME

ShadyWhite:


Bing-su:
at jobs_config.js, 10th and 21st lines are translated to same chinese. I think it also should be fixed.

ShadyWhite:
> at jobs_config.js, 10th and 21st lines are translated to same chinese. I think it also should be fixed.

Fixed，thanks.

quisquous:
Thanks!

PR #897, opened 2019-12-30
lint: Enable Markdown Linter for CI

panicstevenson:
Add [markdownlint-cli](https://github.com/igorshubovych/markdownlint-cli) as a step to be run within Travis CI's linting stage. Change includes bringing existing Markdown files to a state where they will not throw an error with the default configuration for Markdownlint. Changes can be made such that Markdownlint uses an externalized configuration, but it felt safe using the defaults for now.

Line wrapping was performed using the VSCode extension [Rewrap](https://marketplace.visualstudio.com/items?itemName=stkb.rewrap) and was done via the 'Rewrap Comment / Text at column...' option with 80 provided as the column to wrap at.

Various minor changes were made as well, mostly spelling errors caught while proofreading.

Table of contents were changed as well to follow a numbered format, to be later auto-generated as well.


CONTRIBUTING.md
```diff
@@ -20,102 +20,103 @@ The code of conduct for cactbot can be found here:
...
-than very large pull requests (that may conflict or take a while to land).
-Each pull request should be small and be self-contained in terms of what it is changing.
+It's vastly preferable to make a lot of small pull requests (that are easy to
+review) than very large pull requests (that may conflict or take a while to
+land). Each pull request should be small and be self-contained in terms of what
+it is changing.
 
-If your change is small, just send a pull request and we can have a conversation there.
-If your change is big, consider having a conversation before you embark on a lot of work.
+If your change is small, just send a pull request and we can have a conversation
```
quisquous: I think this reformatting here is not great in terms of removing  semantic linebreaks.

How hard would it be to add linting without line limits? (Or to increase the line limit to like 100 such that these lines don't get reformatted?) \*<_<*

panicstevenson: The level of effort isn't too high; the issue that I have is that to override the configuration we'd have to create a configuration file, and we can't use package.json. The ruby implementation lets you define your linting rules inline with the command-line tool, but the same does not exist for the Node version.

For example: https://github.com/eslint/eslint/blob/master/.markdownlint.yml

I don't have a huge issue with adding another file, but it's a little ugh in my opinion. Other than that, I have no issues; let me know what you think!

quisquous: An additional config file seems fine.  It's not clear that there's really another option.

PR #898, opened 2019-12-30
Fix the new user config for users who haven't used it, yet.

ngld:
C# `Dictionary`s throw an exception if a missing key is accessed. I used `TryGetValue` to avoid that.
The `JToken` objects behave as expected (a missing key just returns `null`).

Finally, I had to change `user_config.js` because it tried to access a property on a `null` value when `cactbotLoadData` returns `null`.


quisquous:
Thank you!!

PR #901, opened 2019-12-30
raidboss: Update Binding Coil Triggers

panicstevenson:
Update Binding Coil of Bahamut trigger files to utilize the new Regexes
and Responses format.

Additional changes were made to logic if it seemed incorrect.

### DIFF
--- a/ui/raidboss/data/02-arr/raid/t1.js
+++ b/ui/raidboss/data/02-arr/raid/t1.js
@@ -4,35 +4,35 @@
   zoneRegex: /^The Binding Coil Of Bahamut - Turn \(1\)$/,
   triggers: [
     {
-      id: 'T1 Silence',
-      regex: / 14:5A7:Ads starts using High Voltage/,
-      regexDe: / 14:5A7:Abwehrsystem starts using Hochstrom/,
-      regexFr: / 14:5A7:Sphère De Contrôle starts using Haute Tension/,
-      regexJa: / 14:5A7:制御システム starts using 高圧電流/,
+      id: 'T1 High Voltage',
+      regex: Regexes.startsUsing({ source: 'Ads', id: '5A7', capture: false }),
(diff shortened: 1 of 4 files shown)


ui/raidboss/data/02-arr/raid/t1.js
```diff
@@ -69,7 +69,7 @@
     },
     {
       id: 'T1 Slime Timer First',
-      regex: / 00:0839:The Allagan megastructure will be sealed off/,
+      regex: Regexes.gameLog({ line: '00:0839:The Allagan megastructure will be sealed off', capture: false }),
```
quisquous: This should be Regexes.message({line: 'The Allagan etc'})

ui/raidboss/data/02-arr/raid/t5.js
```diff
@@ -62,42 +47,42 @@
...
             en: 'Fireball on YOU',
           };
         }
       },
       infoText: function(data, matches) {
-        if (data.me != matches[1]) {
+        if (data.me != matches.target) {
           return {
-            en: 'Fireball on ' + data.ShortName(matches[1]),
+            en: 'Fireball on ' + data.ShortName(matches.target),
```
quisquous: I do really wonder if we need a "thing on X" style response, as this is a common pattern.  However, I'm not sure how to handle the translations on that.  This is just a side thought, and not something that needs to get done in this pull request.

panicstevenson: It might make things easier for translation? Since everything follows a similar pattern, `on X` for English, `sur X`, `auf X`, `→ X`, etc.
When in doubt, you could use the full name of the ability for initial translation passes.

quisquous: Yeah, we could just use `matches.ability`, pretty easily.

PR #903, opened 2020-01-02
plugin + eureka: Add fate tracking support

KattTails:
Following patch 5.18, NMs only render (and the client only receives the `Added Combatant` line) when the player is within 100y of them. Upon leaving the 100y radius `Removed Combatant` lines are sent. These changes made the overlay unable to usefully track FATEs in Hydatos.

This PR moves the eureka overlay from relying on `Added/Removed Combatant` loglines, and adds a new FateWatcher module to the plugin using FFXIV.Machina to watch network packets for FATE info. Korean/Chinese clients are also supported **(although I don't have the chinese client OpCode upon submission of this PR)**.

This also allows us to track the bunny FATEs! Yay!


CactbotOverlay/FateWatcherKo.cs
```diff
@@ -0,0 +1,35 @@
...
+  class FateWatcherKo : FateWatcher {
+    //
+    // for FFXIV KO version: 5.01
+    //
+    // Latest KO version can be found at:
+    // https://www.ff14.co.kr/news/notice?category=3
+    //
+    private const ushort Server_MessageTypeActorControl143 = 323;
+
+    public FateWatcherKo(CactbotEventSource client) : base(client) {
```
quisquous: I think for things like memory structrures, it makes sense to have different ffxivprocess classes, but I don't think the pattern really makes sense here, especially as it's just two values? I think this would be better if either:
* FateWatcher had constructor parameters for these values and they got passed in
* FateWatcher had a constructor value for the language, and it just set the values internally from the language

...and then we don't need multiple classes hanging around that duplicate a bunch of logic.

CactbotOverlay/FateWatcherKo.cs
```diff
@@ -0,0 +1,35 @@
...
+      if (message.Length < sizeof(Server_MessageHeader))
+        return;
+
+      fixed (byte* buffer = message) {
+        Server_MessageHeader serverMessageHeader = *(Server_MessageHeader*)buffer;
+        if ((ushort)serverMessageHeader.MessageType == Server_MessageTypeActorControl143)
+          ProcessMessage(*(Server_ActorControl143*)buffer);
+
+        // Sent upon client logout
+        if (serverMessageHeader.MessageType == (Server_MessageType)0x142) {
```
quisquous: Is this 0x142 value something that will change from patch to patch? If so, this is a little worrisome.

ui/eureka/eureka.js
```diff
@@ -33,6 +37,11 @@ let Options = {
     },
   },
   ZoneInfo: {
+    // Fate IDs
```
quisquous: Thanks for this comment!

CactbotOverlay/CactbotEventSource.cs
```diff
@@ -83,11 +84,19 @@ public class CactbotEventSource : EventSourceBase, ILogger {
...
+
+    public delegate void FateEventHandler(JSEvents.FateEvent e);
+    public event FateEventHandler OnFateEvent;
+
     public void Wipe() {
       Advanced_Combat_Tracker.ActGlobals.oFormActMain.EndCombat(false);
       OnPartyWipe(new JSEvents.PartyWipeEvent());
     }
 
+    public void DoFateEvent(JSEvents.FateEvent e) {
```
quisquous: Maybe calls to DoFateEvent should just go to OnFateEvent and this function shouldn't exist?

CactbotOverlay/CactbotOverlay.csproj
```diff
@@ -86,6 +94,16 @@
   <ItemGroup>
     <Compile Include="CactbotEventSource.cs" />
     <Compile Include="CactbotEventSourceConfig.cs" />
+    <Compile Include="CactbotEventSourceConfigPanel.cs">
+      <SubType>UserControl</SubType>
+    </Compile>
+    <Compile Include="CactbotEventSourceConfigPanel.Designer.cs">
```
quisquous: Were these intentionally added back in?

ui/eureka/eureka.js
```diff
@@ -1144,6 +924,34 @@ let Options = {
         en: 'Pyros Elemental',
         cn: '涌火元灵',
       },
+      bunnys: {
+        northbunny: {
+          label: {
+            en: 'Bunny',
```
quisquous: I wonder if it'd be worth defining a bunnyLabel object above, and then just having `label: bunnyLabel` to not have to repeat this.

ui/eureka/eureka.js
```diff
@@ -1604,22 +1328,16 @@ let Options = {
...
-            de: 'Kalamari',
-            fr: 'Khalamari',
-            ja: 'Khalamari',
-            cn: '卡拉墨鱼',
-          },
           trackerName: {
             en: 'Khalamari',
             de: 'Kalamari',
-            fr: 'Khalamari',
+            fr: 'Khalamar',
```
quisquous: Is this intentional?

ui/eureka/eureka.js
```diff
@@ -1704,22 +1404,16 @@ let Options = {
             ja: 'フロストメーン',
             cn: '老虎',
           },
-          mobName: {
+          trackerName: {
```
quisquous: Is this intentional, vs keeping the trackerName the same?

Similar comments on other trackerName vs mobName changes below.

ui/eureka/eureka.js
```diff
@@ -1949,13 +1590,38 @@ let gWeatherIcons = {
 let gNightIcon = '&#x1F319;';
 let gDayIcon = '&#x263C;';
 
+class EventQueue {
```
quisquous: I kind of think this is a little bit overkill.  I think you could just make fatequeue a `[]` that you push and shift from and save twenty lines of code.

ui/eureka/eureka.js
```diff
@@ -2,6 +2,10 @@
 
 let Options = {
   PopSound: '../../resources/sounds/PowerAuras/sonar.ogg',
+  PopVolume: 1.0,
+  TrackBunnyFates: true,
```
quisquous: I think it'd be fine to always track bunny fates and not give an option, for what it's worth.  If people don't like it, they can turn off the sound at least.

ui/eureka/eureka.js
```diff
@@ -2311,27 +2039,80 @@ class EurekaTracker {
...
+      for (let i = 0; i < this.nmKeys.length; ++i) {
+        let nm = this.nms[this.nmKeys[i]];
+        if (e.detail.fateID == nm.fateID) {
+          this.OnFateUpdate(nm, e.detail.progress);
+          return;
+        }
+      }
+      if (this.options.TrackBunnyFates) {
+        for (let i = 0; i < this.bunnyKeys.length; ++i) {
+          let bunny = this.bunnys[this.bunnyKeys[i]];
```
quisquous: One question I have is if the bunnies can be less special cased here.  I think removing the bunny option would help.

I think if you put the bunnies in the nm list proper (maybe with a bunny: true setting to change the sound) then this code could just be one loop.

Additionally, instead of having a bunch of code have to differentiate `.bunny-pop` from `.nm-pop`, you could add a `.bunny` class to the bunny nms, and treat bunnies as nms with an additional class.  In other words, bunny fates get `.nm-pop` and `.nm-down` but you just add another css rule for `.bunny.nm-pop` that changes the text color.  I think this would clean up some special casing code too.

And then OnFatePop could just be like "is this fate a bunny? pick a different sound" rather than having to duplicate some code in OnBunnyPop.

What do you think?

CactbotOverlay/CactbotEventSource.cs
ngld: You can't raise events from other classes in C#. Calling `OnFateEvent` from `FateWatcher` wouldn't work.
However, you could just scrap the `OnFateEvent` and call `DispatchToJS` from `FateWatcher`.

CactbotOverlay/CactbotOverlay.csproj
```diff
@@ -61,6 +61,14 @@
...
       <HintPath>ThirdParty\OverlayPlugin\libs\HtmlRenderer.dll</HintPath>
     </Reference>
+    <Reference Include="Machina, Version=2.1.1.2, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>ThirdParty\Machina.FFXIV\Machina.dll</HintPath>
+    </Reference>
+    <Reference Include="Machina.FFXIV, Version=2.1.3.10, Culture=neutral, processorArchitecture=MSIL">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>ThirdParty\Machina.FFXIV\Machina.FFXIV.dll</HintPath>
+    </Reference>
```
ngld: Ravahn never published Machina builds (as far as I know) so everyone would have to make their own Machina builds or unpack them from the FFXIV plugin. Both sound like a bad solution.
In OverlayPlugin I used reflection to work around this.

CactbotOverlay/FateWatcher.cs
```diff
@@ -0,0 +1,85 @@
...
+      // Fate End: 0x79
+      // param1: fateID
+      if ((int)packetMessage.category == 0x79) {
+        RemoveFate((int)packetMessage.param1);
+      }
+
+      // Fate Progress: 0x9B
+      // param1: fateID
+      // param2: progress (0-100)
+      if ((int)packetMessage.category == 0x9B) {
```
ngld: Wouldn't a `switch` be more readable and (marginally) faster than these `if`s?

CactbotOverlay/FateWatcherCn.cs
```diff
@@ -0,0 +1,36 @@
...
+      };
+    }
+    internal unsafe override void MessageReceived(byte[] message) {
+      if (message.Length < sizeof(Server_MessageHeader))
+        return;
+
+      fixed (byte* buffer = message) {
+        Server_MessageHeader serverMessageHeader = *(Server_MessageHeader*)buffer;
+        if ((ushort)serverMessageHeader.MessageType == Server_MessageTypeActorControl143)
+          ProcessMessage(*(Server_ActorControl143*)buffer);
```
ngld: You never check if the message is shorter than `sizeof(Server_ActorControl143)` which means you'll run into out-of-bound reads eventually since after the next opcode shuffle a different, possibly shorter message, could have this opcode.

CactbotOverlay/FateWatcherKo.cs
```diff
@@ -0,0 +1,35 @@
...
+
+namespace Cactbot {
+  class FateWatcherKo : FateWatcher {
+    //
+    // for FFXIV KO version: 5.01
+    //
+    // Latest KO version can be found at:
+    // https://www.ff14.co.kr/news/notice?category=3
+    //
+    private const ushort Server_MessageTypeActorControl143 = 323;
```
ngld: Why don't you rely on the FFIXV plugin here? It should always have the correct opcode for the current game version (that's what the korean build is for after all).

CactbotOverlay/FateWatcherKo.cs
ngld: Yes, it's the message's opcode. See also my other messages in reference to this. AFAIK you won't ever have to include an opcode unless you're looking for packets the FFXIV plugin doesn't care about.

CactbotOverlay/FateWatcherIntl.cs
```diff
@@ -0,0 +1,28 @@
...
+      if (message.Length < sizeof(Server_MessageHeader))
+        return;
+
+      fixed (byte* buffer = message) {
+        Server_MessageHeader serverMessageHeader = *(Server_MessageHeader*)buffer;
+        if (serverMessageHeader.MessageType == Server_MessageType.ActorControl143)
+          ProcessMessage(*(Server_ActorControl143*)buffer);
+
+        // Sent upon client logout
+        if (serverMessageHeader.MessageType == (Server_MessageType)0x142) {
```
ngld: This opcode remains the same between all three variants. Are you sure that's correct? I'd expect it to be shuffled like all the other opcodes.

CactbotOverlay/FateWatcherIntl.cs
```diff
@@ -0,0 +1,28 @@
+﻿using Machina.FFXIV;
+using Machina.FFXIV.Headers;
+
+namespace Cactbot {
+  class FateWatcherIntl : FateWatcher {
+
+    public FateWatcherIntl(CactbotEventSource client) : base(client) {
+      monitor = new FFXIVNetworkMonitor {
+        MessageReceived = (string connection, long epoch, byte[] message) => MessageReceived(message)
+      };
```
ngld: Why do you spin up an additional network monitor? This way every packet will be decoded twice: Once by the FFXIV plugin and once by Cactbot. Also, the list of active connections, etc., all will be duplicated.
The FFXIV plugin raises an event for each received packet which makes all of this unnecessary. [OverlayPlugin even provides a handy wrapper](https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/FFXIVRepository.cs#L155).

quisquous:
> Alright, I realize I don't have any authority here but...

No no...pull request comments from anybody are always much appreciated.  Thanks for taking the time to review.

KattTails:
Really thank you for your input here, everything you've suggested is good and I greatly appreciate it. I'll try and implement it over the next day or so.

ngld:
I just updated my own network parsing code to read the opcode from the FFXIV plugin. Feel free to use [this as a template](https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/NetworkParser.cs).

KattTails:
> 
> 
> I just updated my own network parsing code to read the opcode from the FFXIV plugin. Feel free to use [this as a template](https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/NetworkParser.cs).

Thank you! Sorry for the vanishing act, I'm back and working on this.

ui/eureka/eureka.js
KattTails: Yeah that sounds much, much cleaner.

CactbotOverlay/FateWatcherKo.cs
KattTails: I've consolidated this into a single FateWatcher class with help from @ngld's OverlayPlugin code.

CactbotOverlay/CactbotEventSource.cs
KattTails: Done.

ui/eureka/eureka.js
KattTails: I removed mobName because we don't care about the entities any more for tracking fates. The code relies only on the fate information.
I've updated all the trackerName entries with their current strings for all languages as of 15/04/2020.

ui/eureka/eureka.js
KattTails: It was a bit overkill, changed. 👍

ui/eureka/eureka.js
KattTails: Implemented as suggested.

KattTails:
I've tried implementing most of the suggestions here, thank you massively for all your input. I know it's been a long time..

AFAIK all I have left to do is add the new options to the config panel UI, add translations for the bunny fates (input welcome on that) and then a final review.

ui/eureka/eureka.js
KattTails: added.

ui/eureka/eureka.js
```diff
@@ -2280,7 +1937,6 @@ class EurekaTracker {
     let nextDay = findNextNight(nowMs);
     let nextNight = findNextDay(nowMs);
     let timeStr = '';
-    let timeVal;
```
KattTails, line 2283: removed an unused variable

quisquous:
Thank you!!!

I'll try to give this a whirl and take a look at it this weekend when I have a little bit more time.

plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -379,6 +394,7 @@ protected override void Update()
       // onZoneChangedEvent: Fires when the player changes their current zone.
       string zone_name = Advanced_Combat_Tracker.ActGlobals.oFormActMain.CurrentZone;
       if (notify_state_.zone_name == null || !zone_name.Equals(notify_state_.zone_name)) {
+        fate_watcher_.RemoveAndClearFates();
```
KattTails, line 397: onZoneChangedEvent fires _after_ we receive the network FATE data, causing the fates to be added -> zone change -> fate removed -> fate readded the next time an update event is sent.

KattTails: Pushed https://github.com/quisquous/cactbot/pull/903/commits/e30a31f283ee7d55ec495a52514e2ca11c9a41b7 to resolve this.

ui/eureka/eureka.js
```diff
@@ -650,23 +546,31 @@ let Options = {
         ko: '파고스 정령',
       },
       nms: {
+        northbunny: {
```
quisquous, line 549: Can you verify that importing from the tracker (click the red clipboard, past in game with an /echo) works properly? I get an error because northbunny (and other bunnies) don't have a trackerName entry.  I think the right fix is to adjust the function to handle entries with no trackerName (and not to add empty ones here).

ui/eureka/eureka.js
```diff
@@ -687,81 +593,63 @@ let Options = {
...
             fr: 'Taxim',
             ja: 'タキシム',
-            ko: '택심',
+            ko: '텍심',
             cn: '读书人',
           },
           x: 25.5,
           y: 28.3,
+          fateID: 1369,
+          bunny: false,
```
quisquous, line 607: I think "bunny: false" can just be implied when missing, and not included in every fate, if you don't mind removing them.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -0,0 +1,192 @@
...
+          }
+          if (fates[param1] != param2) {
+            UpdateFate(param1, param2);
+          }
+          break;
+        }
+      }
+    }
+
+    private void AddFate(int fateID) {
```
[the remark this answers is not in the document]
quisquous: On second thought, the old code didn't do this.  It's ok to not add this complication, and if anybody really needs it in the future we can add it.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -0,0 +1,192 @@
...
+    private int MessageType_Offset = 0;
+    private int Category_Offset = 0;
+    private int Param1_Offset = 0;
+    private int Param2_Offset = 0;
+    private ushort ActorControl143_Opcode = 0;
+
+    // fates<fateID, progress>
+    private static ConcurrentDictionary<int, int> fates;
+
+    public FateWatcher(CactbotEventSource client, string language) {
```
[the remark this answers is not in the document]
KattTails: no, just forgot to take it out 🙄

plugin/CactbotEventSource/FateWatcher.cs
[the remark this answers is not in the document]
KattTails: It would be neat to add this pre-emptively, I'll look into it either now or as another PR if this gets merged.

ui/eureka/eureka.js
KattTails: Good catch, fixed as you suggested.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -0,0 +1,192 @@
...
+      }
+    }
+
+    public void Stop() {
+      if (subscription != null)
+        subscription.NetworkReceived -= new NetworkReceivedDelegate(MessageReceived);
+    }
+
+    // GetOffset, GetEnumValue, GetOpcode copied directly from ngld's OverlayPlugin with deep gratitude
+    // Source: https://github.com/ngld/OverlayPlugin/blob/8f147d2cd11fa8c9713be2a8c2296a099bfa54ac/OverlayPlugin.Core/NetworkParser.cs
```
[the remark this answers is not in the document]
KattTails: Yeah I'd be open to moving this to OverlayPlugin, if it's cool with ngld. I don't really see any reason why onFateEvent needs to be specific to Cactbot.

quisquous:
Thanks for all this.  I'm going to merge this now to avoid conflicts on #1353, and because this all looks good and ready to go.  Other followups can be done separately if you want to do them.  :)

PR #904, opened 2020-01-02
plugin: add rogue to GetJobSpecificData

KattTails:
When I added the other missing classes I missed rogue..


panicstevenson:
>When I added the other missing classes I missed rogue..

Too stealthy

Issue #908, opened 2020-01-03
Error: GetJobSpecificData: 255


Jaehyuk-Lee:
## Description
```
[2020-01-03 오후 5:48:59] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:00] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:00] Info: [Cactbot] Raidboss: BrowserConsole: Unknown job role (Source: file:///C:/Advanced%20Combat%20Tracker/Plugins/cactbot/resources/common.js, Line: 82)
[2020-01-03 오후 5:49:00] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:00] Info: [Cactbot] Oopsyraidsy: BrowserConsole: Unknown job role (Source: file:///C:/Advanced%20Combat%20Tracker/Plugins/cactbot/resources/common.js, Line: 82)
[2020-01-03 오후 5:49:00] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:01] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:01] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:01] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:01] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:02] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:02] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:02] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:03] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:03] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:03] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:04] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:04] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:04] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:05] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:05] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:05] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:06] Error: GetJobSpecificData: 255
[2020-01-03 오후 5:49:06] Error: GetJobSpecificData: 255
[...]

Issue #906, opened 2020-01-03
cactbot Jobs doesn't work after 0.14.0 in Korean server


Bing-su:
When using 0.13.5, it worked well.

![jobs 작동안함](https://user-images.githubusercontent.com/37621276/71707067-3e2f9a80-2e2b-11ea-826e-d9ada719cab7.png)

overlay plugin logs
![녹화_2020_01_03_13_09_45_708](https://user-images.githubusercontent.com/37621276/71707085-5ef7f000-2e2b-11ea-94b1-93c1cb8a0758.gif)

Windows 10 64bit
FFXIV KR 5.01 DX11

FFXIV ACT plugin KR 2.1.0.1
Overlay plugin 0.12.1.0

KattTails:
Hi,

Can you check if this .dll works for you? 
[CactbotOverlay.zip](https://github.com/quisquous/cactbot/files/4018142/CactbotOverlay.zip)

Bing-su:
I checked, but it still not worked.

KattTails:
Can you paste your overlayplugin.dll log here? I'm interested in the lines shown here: 
![image](https://user-images.githubusercontent.com/12807478/71710397-ff322280-2df3-11ea-88ba-28ce3c97ffb8.png)

Bing-su:
here
![Cap 2020-01-03 16-32-11-955](https://user-images.githubusercontent.com/37621276/71711954-96c06100-2e46-11ea-8776-48ab531588c8.png)

KattTails:
Yep okay I see what the problem is. The korean offsets are only loaded if you're using the korean ACT plugin which has `Korean` in the file name, as it's given in the releases by Ravahn.

For now, can you rename your `FFXIV_ACT_Plugin.dll` to include `Korean`, eg `FFXIV_ACT_Plugin_Korean.dll`. It should work then and you should see `Version: ko` in the log. Please make sure you delete the dll I linked above and use the official release also.

Bing-su:
Thanks! It works cool.... with another small problem.
![Cap 2020-01-03 16-55-21-254](https://user-images.githubusercontent.com/37621276/71712747-fbc98600-2e49-11ea-8c7b-2da2602a0bf5.png)
![Cap 2020-01-03 16-55-47-992](https://user-images.githubusercontent.com/37621276/71712758-0421c100-2e4a-11ea-845d-d2c414294fe6.png)

anyway this issue is completed. I closed it.

KattTails:
[...]

PR #909, opened 2020-01-03
plugin: Revert korean region check and just use FFXIV_ACT_Plugin's language setting

KattTails:
Following the submission of #908 and #906, it's common that the Korean FFXIV_ACT_Plugin dll is being renamed. 

Using the FFXIV_ACT_Plugin's game language setting is also a reliable check, as although it isn't a true check of the region, the international client doesn't have Korean language support and so should only be set as such on a Korean client.

### DIFF
--- a/CactbotOverlay/CactbotEventSource.cs
+++ b/CactbotOverlay/CactbotEventSource.cs
@@ -45,7 +45,6 @@ namespace Cactbot {
     private FFXIVProcess ffxiv_;
     private WipeDetector wipe_detector_;
     private string language_ = null;
-    private string ffxiv_plugin_region_ = null;
     private List<FileSystemWatcher> watchers;
 
     public delegate void ForceReloadHandler(JSEvents.ForceReloadEvent e);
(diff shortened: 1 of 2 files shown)


Jaehyuk-Lee:
After merging this commits, please publish a new release as soon as possible. v0.14.0 is not compatible with HAERUBOT(which most Korean users are using for managing ACT & its plugins)
with HAERUBOT, cannot use v0.13.5 because HAERUBOT forces update before enabling it. (HAERUBOT may take some emergency action with this problem, but for now, it's impossible to use v0.13.5.)

KattTails:
> 
> 
> Haeru said she'll add "Korean" in plugin file name (maybe for emergency situation now, so might be removed after cactbot's new release)
> So, I think you don't have to publish a new release fast. It's okay now.

That's really cool, please send her my thanks!

I can't speak for quisquous directly but I think there will be another release quickly, there's another bug that's popped up in this release that will need addressing fast.

PR #911, opened 2020-01-03
plugin: hacky fix for null reference errors on FFXIV process exit

KattTails:
It isn't neat but it should catch all the attempts to access null variables with the new LimitedProcess wrapper.

Feel free to reject this :P I just thought I'd get something up that "works".


quisquous:
Fixed this differently.  Thanks for the speculative pull request though, thanks!

Issue #913, opened 2020-01-03
Radar: Some Hunts are not displayed (maybe DE specific issue)


Akurosia:
Language: DE
cacbot version: 0.14.0.0 (self compiled)
Overlayplugin version; 0.12.1.0 (self compiled)

Some Hunts are not displayed in the Radar plugin. 

An example name from an A rank in northern Thanalan: "Letzte Flamme Dalvags"

slashy:
This is indeed a german language specific issue. I made a [pull request](https://github.com/quisquous/cactbot/pull/916/files) for fixing those, however this is just a quick and dirty fix. Since the name for the Hunt mobs are taken from the game files directly with a python script that generates the hunts.js file where the names are found my fix will be overwritten once someone starts that script. 

Now usually parsing stuff from the game files is fine. However since german language is a b*tch, we do have all sorts of special cases for genders and what not. So in the game files "Letzte Flamme Dalvags" is noted as "Letzt[a] Flamme[p] Dalvags" 

Now I already know that [p] stands for plural and adds whatever letter(s) needs to be added to denote the plural version (in case of those names they are usually just removed as there is no plural). 

[a] as in **a**rticle (= Geschlechtswort) for gender based endings: Most of them are "e" (female) or "er" (male) depending on the gender of the name described by it.

There's also [t] which is the **t**itle (= Anrede) used, usually also denoted by the gender of whatever word/name it describes. 

**Example:** Sanu Vali [t] Tanzend[a] Schwinge would be "Sanu Vali die Tanzende Schwinge"

In this case [t] would be "die" because "Schwinge" is female and [a] would be "e" also because of Schwinge being of female origin. 

Now I don't know if the missing letters can be pulles from the game files aswell, which is why I did not modify the python-script generating the hunts.js, but instead opted for altering the actual file even though it says not to. 

I did not look too much into the radar.js, but if those names are used for regular expressions I assume one could come up with regexp-strings for replacement of each of those [*] tags. 

[p] could be removed
[t] would be /d(?>er|ie|as)/
[a] would be /er?/

So "Sanu Vali [t] Tanzend[a] Schwinge" would become "Sanu Vali d(?>er|ie|as) Tanzender? Schwinge" which would indeed match "Sanu Vali die Tanzende Schwinge".

"Letzt[a] Flamme[p] Dalvags" would be "Letzter? Flamme Dalvags" and would also match for "Letzte Flamme Dalvags".

Hope this helps to come up with something better than my pull request.

[...]

PR #916, opened 2020-01-03
Fixed incorrect german names for hunt targets

slashy:
I know it says not to edit this file directly, but the data gathered directly from the gamefiles includes placeholders for german grammar.

So for now this fixes the incorrect names so they are detected. 

However the underlying problem needs to be fixed in the python script parsing the game files, but I am not sure how to do it as there is no fixed grammar rules for each placeholder as they are determined by gender and other things regarding it's target.

### DIFF
--- a/resources/hunt.js
+++ b/resources/hunt.js
@@ -247,7 +247,7 @@ let gMonster = {
   'Dalvag\'s Final Flame': {
     'name': {
       'cn': '\u83f2\u5170\u5fb7\u7684\u9057\u706b',
-      'de': 'Letzt[a] Flamme[p] Dalvags',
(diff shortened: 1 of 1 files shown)


quisquous:
Ok, I can land this for now, but as you say the script definitely needs to be updated.

I think probably the thing to do would be to use regular expressions, so that [a] can become 'e' or 'er', etc.

Do you know what the mappings are? Right now for fishing, I have:
* [A] -> er
* [a] -> e, en, es, er
* [p] -> (nothing)
* [t] -> der

But it sounds like [t] can also be die? Can [p] be other things?

quisquous:
Oh, I see the conversation on #913, so I'll just continue it there.

PR #925, opened 2020-01-05
add a readme about using python util

Jaehyuk-Lee:
If this README.md file should be placed somewhere else, please tell me.

preview: https://github.com/Jaehyuk-Lee/cactbot/blob/master/util/README.md

### DIFF
--- a/util/README.md
+++ b/util/README.md
@@ -1,0 +1,17 @@
+# How to run python utils
+
+Download and install [python](https://www.python.org/).
+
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for filling in the missing documentation! Sorry about not writing any of this out.  It looks like @Akurosia is going to make this easier in #926.  I also ran into the same troubleshooting issue and sent an informal bug report to goats about it.  It's good to have that written out though until it gets fixed.

PR #926, opened 2020-01-05
Updated gen_hunt_data to avoid common issues

Akurosia:
added ability to specify saintcoinach.cmd and ffxiv path for reader and cacbot path for writer to avoid common user error

### DIFF
--- a/util/gen_hunt_data.py
+++ b/util/gen_hunt_data.py
@@ -4,11 +4,19 @@ import os
 import io
 import coinach
 
+# Path to FFXIV installation (None == Default location)
+_FFXIV_GAME_PATH = None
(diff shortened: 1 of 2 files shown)


util/gen_hunt_data.py
```diff
@@ -4,12 +4,20 @@
 import io
 import coinach
 
+# Path to FFXIV installation (None == Default location)
+_FFXIV_GAME_PATH = None
```
quisquous: Could you make these command line args or looking at environment variables? Or, something that folks don't have to edit files directly in order to adjust?

Akurosia: yeah just wanted a quick fix for the moment but will copy the style of the other tools to keep them consistent :)

PR #928, opened 2020-01-05
Update monster position

rreminy:
Quick change to update monster position when (re-)detected, mainly for B ranks but works for all monsters.

This fixes the situation where (example):
1. You're scouting and gets in range of a B rank.
2. Radar detects B rank and adds it to the list.
3. You fly out of range as you continue scouting.
4. B rank gets killed while you're out of range, causing it to respawn somewhere else.
5. You get in range of said monster's new position.
6. Radar doesn't update the position because the old one is "still alive".

This small change makes it so the radar update the B rank's position in example above. Updating the position in realtime would be suitable but I don't know enough to code it. This should however be enough as hunts don't move too far anyway.

### DIFF
--- a/ui/radar/radar.js
+++ b/ui/radar/radar.js
@@ -119,54 +119,64 @@ class Radar {
 
     let mobKey = matches.groups.name.toLowerCase();
     if (mobKey in this.targetMonsters) {
-      // Update monster position
-      this.targetMonsters[mobKey].pos =
+      // Get positions
+      let playerPos = new Point2D(this.playerPos.x, this.playerPos.y);
+      let curPos = this.targetMonsters[mobKey].pos;
+      let newPos =
         new Point2D(parseFloat(matches.groups.x), parseFloat(matches.groups.y));
-      this.targetMonsters[mobKey].posZ = matches.groups.z;
-      return;
-    }
 
-    // add dom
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks, this is a great idea.  One problem is that this doesn't work well with custom mobs that there may be more than one of, like adding Smilodon mobs for example:

```
Options.CustomMonsters = {
  'smilodon': {
    'name': {
      'en': 'Smilodon',
    },
    'rank': 'Custom',
  },
};
```

As you fly around in a number of these mobs, the distance kind of pops randomly back and forth.

One fix might be to only update the position if it's closer to the player than the current position?  Would that work or are there cases where that would cause problems?

(There are other larger solutions here too, like handling these mobs separately, or using ids instead of names, but that's maybe a larger fix than what you're looking to do here, so I'm trying to suggest something smaller.)

rreminy:
Added code to detect the distance from player to the new detected monster and only update if its closer.

I also added a bit of code to only make the sound if the monster is far from the original, previously detected monster position, otherwise we have the next problem of hearing the pop sound on every update too... hardcoded to 100m on line 141

quisquous:
Thanks for this!  100m seems really reasonable.

PR #932, opened 2020-01-06
Register the event source immediately after the addon is loaded

ngld:
With this users will no longer have to restart ACT after enabling Cactbot.
I've also added user-friendly error messages if OverlayPlugin is missing or outdated.

I'm not sure how stable the new load logic is so feel free to stress test this change.


quisquous:
Thanks! I'll play with this and see what it looks like.

quisquous:
I haven't forgotten about this, I just keep getting distracted when I mean to test this more.

quisquous:
Rebased: [0001-ngld-plugin-change-rebased.patch.txt](https://github.com/quisquous/cactbot/files/5009800/0001-ngld-plugin-change-rebased.patch.txt)

With this patch, disabling and then reenabling gives a C# exception when reregistering the event source (due to duplicate event handlers).  This doesn't happen before this patch, although arguably it never re-registered the event source when reloading in the past, and so it wasn't ever quite right either, although silently so?

I'm not quite sure what do about this.  I think the root of the situation is that you can't unregister an event source.

quisquous:
(Sorry also for taking half a year to get back to this.)

quisquous:
The other thing is that if cactbot is not enabled when ACT starts then all of its overlays throw errors about missing events.  When the cactbot plugin is enabled at that point, all of the overlays need a reload.  I can't use the `onForceReload` event because the MissingEventSubError early outs before subscribing to the (not yet existing) event.

Do you have a suggestion on how to handle this part of it, or is the idea that this only is an issue before overlays have been added the first time?

ngld:
I'm about halfway done merging Cactbot's event source into the my refactored version so I'd rather focus on that than fixing these issues.

The issue with unregistering event sources is that I'd also have to clear any events connected to them which would require tracking the registered events for each event source. There are also a few other bugs when you disable and re-enable OverlayPlugin for similar reasons. I, uh, never wrote proper cleanup logic since it seldom matters (i.e. only when you try to disable plugins which is pretty rare).

The issue around unregistered events has been present since I implemented the new event system. I suppose I could keep the error message but still register the event listeners that should allow you to send the reload event but I'm not sure if it breaks anything else (i.e. cached events).
In any case, I don't think it's a huge issue since this only happens if you reload overlays after disabling Cactbot or if you launch ACT with Cactbot disabled. In either case, restarting ACT or reloading the relevant overlays manually should fix the issue.

quisquous:
That seems reasonable! Thank you for looking into merging cactbot into OverlayPlugin! (Are you planning on releasing your ngrok work soon too?)

> In any case, I don't think it's a huge issue since this only happens if you reload overlays after disabling Cactbot or if you launch ACT with Cactbot disabled. In either case, restarting ACT or reloading the relevant overlays manually should fix the issue.

No, agreed that it's not a big deal, but I thought the intention of this patch was to figure out how to avoid unnecessary restarts.  Given that, maybe I should just close this for now?  Or, maybe I'm misunderstanding?

quisquous:
Given that you've got a beta out that merges cactbot event source into overlayplugin, I'll just go ahead and close this.  Sorry again for letting this drop for so long.

PR #933, opened 2020-01-06
TEA Japanese localization

r4ha:


ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
```diff
@@ -286,6 +303,8 @@
                   namesWithoutNisi.map((x) => data.ShortName(x)).join(', or '),
               de: 'Gebe ' + data.nisiNames[myNisi] + ' zu ' +
                   namesWithoutNisi.map((x) => data.ShortName(x)).join(', oder '),
+              ja: data.nisiNames[myNisi] + ' を ' +
+                  namesWithoutNisi.map((x) => data.ShortName(x)).join(', oder ') + ' に渡す',
```
quisquous: The join here is in case something goes awry and it finds multiple names.  So, in English, this could say "Pass Blue α to Person1, or Person2" if there are two people who need that color (somehow).  You probably want to change the `', oder'` german string to be maybe just `' か '`, I guess?

r4ha: oops, sorry i fixed it.

PR #934, opened 2020-01-06
WIP: raidboss: Fix formatting in e1/e3 timelines

amki:
I tried to parse cactbot's timeline files for a little tool of mine but found that naive parsing of spells from the timeline failed because of some formatting errors.

I guess these are unwanted and just happened to be there and not upset cactbot's parsing so I suggest to unify them.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e3s.txt
+++ b/ui/raidboss/data/05-shb/raid/e3s.txt
@@ -20,7 +20,7 @@ hideall "--sync--"
 69.7 "Freak Wave" sync /:Leviathan:3FE6:/
 73.7 "Temporary Current" sync /:Leviathan:(3FEB|3FEA):/
 87.9 "Maelstrom" sync /:Leviathan:3FFA:/
-90.0  "--untargetable--"
+90.0 "--untargetable--"
 99.3 "Spinning Dive 1" #sync /:Leviathan:3FFD:/
 101.3 "Spinning Dive 2" #sync /:Leviathan:3FFD:/
 106.0 "--targetable--"


ui/raidboss/data/05-shb/raid/e1s.txt
```diff
@@ -46,16 +46,16 @@ hideall "--sync--"
 500.0 "Eternal Breath" sync /:Eden Prime:3D8C:/ window 500,0
 510.0 "--targetable--"
 518.4 "Paradise Regained" sync /:Eden Prime:44FC:/
-528.8 "Vice And Virtue! (T)" sync /:Eden Prime:3D78:/
-542.5 "Vice And Virtue! (D)" sync /:Eden Prime:3D7A:/
+528.8 "Vice And Virtue (T)" sync /:Eden Prime:3D78:/
```
quisquous: These all have exclamation points to differentiate the powered up version from the original ones.

ui/raidboss/data/05-shb/raid/e3s.txt
```diff
@@ -20,7 +20,7 @@ hideall "--sync--"
 69.7 "Freak Wave" sync /:Leviathan:3FE6:/
 73.7 "Temporary Current" sync /:Leviathan:(3FEB|3FEA):/
 87.9 "Maelstrom" sync /:Leviathan:3FFA:/
-90.0  "--untargetable--"
+90.0 "--untargetable--"
```
quisquous: This is fine to change, but it might worth adding a test for it somewhere if you want it to be consistent across all timelines.

ui/raidboss/data/05-shb/raid/e1s.txt
amki: Oh you are right they even have different spell ids. I guess I just assumed wrong then

ui/raidboss/data/05-shb/raid/e3s.txt
amki: I'm not sure if it makes sense to enforce this format.

I think the timelines are staying compatible with "ACT Timeline"? https://github.com/quisquous/cactbot/blob/master/ui/raidboss/timeline.js#L14 suggests this.

quisquous: Yes, that's the idea, although there's no reason cactbot timelines couldn't enforce more consistent formatting that is also backwards compatible, if that makes sense?

quisquous:
I'd also be curious to know what you're trying to do with a tool if you wanted to share.  :)

amki:
My idea is to ease the planning of healing in boss fights.

Using the cactbot timeline feature I get an (for me) easy overview about the important spells happening during a boss fight so I parse the timeline and fetch how much damage a spell deals to the group/to the tanks on average. (Either put this in manually or maybe fetch from fflogs or parse some log myself?)

The idea is to then add healing spells to the timeline to counter the incoming damage to come to an optimal solution on how to heal with the least spells/gcds possible. (This could in deep dark theory be done automatically but I'm not really planning it because it's kinda komplex I think)

Afterwards the difference between timelines shall be exported and put into cactbot (user directory) as a timeline modification so that you'd automatically be reminded when it's your time to shine.

Issue #936, opened 2020-01-07
Jobs - KR Client job balance patch is different with Global Client


Jaehyuk-Lee:
## Description

Description of the issue you're seeing.

I don't think this is exactly a bug, but Korean server has different job balance
I found Trick Attack lasts 15 seconds on v5.1 patch. However, it lasts 10 seconds on v5.08 patch. So, I think jobs module need to divide Korean and other languages about cooldowns and duration times.

## Additional information

If available, please include any logs or screenshots depicting the issue you're experiencing.

Jaehyuk-Lee:
I've done with this myself, I'll push a pull request soon.

PR #937, opened 2020-01-07
Jobs - make KR job balance differ from Global

Jaehyuk-Lee:
Code change coping with #936.

**I didn't check other buff duration times.** I just applied only Trick Attack. So, some Korean Server Job balance written in code can be different from real situation.

I'm wondering if `en` representing SquareEnix's server is good. Because some contributors can think they should add `ja` `de` `fr` too, but it doesn't need it. Those four languages have same job balances all the time.


quisquous:
> I'm wondering if `en` representing SquareEnix's server is good. Because some contributors can think they should add `ja` `de` `fr` too, but it doesn't need it. Those four languages have same job balances all the time.

This is a really good point, and I have this concern as well.  I don't want to necessarily have `de` and `fr` etc in this case.  For translations in triggers, I think it's good (even if redundant) because it indicates whether there is translation work remaining.  In this case, there's no work remaining.

I think the other thing that might come up here is that cooldowns can be changed across versions as well, and not just durations.

Do you think it would be clearer instead to have a "buffOverrides" data structure that just gets merged in? Something like:

```
let buffOverrides = {
  ko: {
    trick: {
      duration: 10,
    },
    litany: {
      cooldown: 120,
    },
  },
};
```

...and then just merge that into the `buffInfo` variable? I think that would make it clearer?

Or, maybe you have another suggestion?

Jaehyuk-Lee:
I think something like this is good. (except const name `v508` looks bad, maybe `v5_08` is better?)
Since v5.08 job balance is applied to both KR, CN server. It uses less codes.
Also, it's easy to check whether this code is well applied to each server's balance patches. (We could use comment though...)

```
    const v508 = {
      trick: {
        durationSeconds: 10,
      },
      peculiar: {
        durationSeconds: 10,
      },
    }

    let buffOverrides = {
      ko: v508,
      cn: v508,
    };

    for (let key in buffOverrides[lang]) {
      for (let key2 in buffOverrides[lang][key])
        buffInfo[key][key2] = buffOverrides[lang][key][key2];
    }
```

quisquous:
Yeah, that looks great.  Thanks for the suggestion.

ui/jobs/jobs.js
```diff
@@ -771,7 +790,7 @@ class BuffTracker {
     if (b.gainRegex && !log.match(b.gainRegex))
       return;
 
-    let seconds = b.durationSeconds;
+    let seconds = b.durationSeconds[gLang.lang] || b.durationSeconds['en'];
```
quisquous: These two duration changes can be reverted now, right?

Jaehyuk-Lee:
Ah, I forgot it. Wait a second.

Jaehyuk-Lee:
Ah, I found another thing, `lang` should be changed. Wait a seconds...

Jaehyuk-Lee:
I've fixed bugs, tested those on real situation and worked well.

ui/jobs/jobs.js
```diff
@@ -761,6 +761,25 @@ class BuffTracker {
...
+        durationSeconds: 10,
+      },
+    };
+
+    let buffOverrides = {
+      ko: v5_08,
+      cn: v5_08,
+    };
+
+    for (let key in buffOverrides[gLang.lang]) {
```
quisquous: Instead of gLang.lang, can you use `this.options.Language`?

Jaehyuk-Lee: done with it.

quisquous:
Thanks!!

PR #939, opened 2020-01-08
xeno timer pulse effect

### LINKED ISSUE #571: jobs: add indication for blm when xeno is going to be overwritten
It'd be good to add some visual "omg you're going to lose a xeno if you don't push it now" indicator to the blm jobs ui.  I'm not sure if that means like flashing or transform animating the xeno counter, but something would be good, maybe in the last five seconds before it gets lost.

kuwoyuki:
Added a css transform animation to the xeno timer when the proc
is about to get overwritten, closes #571

### DIFF
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -581,6 +581,29 @@
   background-color: rgb(200, 50, 200);
 }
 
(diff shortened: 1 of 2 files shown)


quisquous:
Neat! Thanks for this.  :+1:

PR #942, opened 2020-01-10
small additions and fixes for gainsEffect

Akurosia:
added KO and CN to some TEA triggers and fixed generall incorrect escapings for gainsEffect


ui/raidboss/data/03-hw/alliance/dun_scaith.js
```diff
@@ -209,10 +209,10 @@
       id: 'Dun Scaith Debilitator Water',
       regex: Regexes.gainsEffect({ effect: 'Water Resistance Down Ii', capture: false }),
       regexDe: Regexes.gainsEffect({ effect: 'Wasserresistenz - \\(Stark\\)', capture: false }),
-      regexFr: Regexes.gainsEffect({ effect: 'Résistance À L\'Eau Réduite+', capture: false }),
+      regexFr: Regexes.gainsEffect({ effect: "Résistance À L\\'Eau Réduite\\+", capture: false }),
```
panicstevenson: As per Travis (via `npm lint`), this needs to have single quotes -- is that still a possibility given the escaping?

Akurosia: yeah, i could not do it as i had a question to @quisquous .
This got clarified (the ' does not need a double \\ just a single one) so the whole string can be in single quotes again. Will provide an update later today

PR #943, opened 2020-01-10
Change all 1E regexes to Regexes.losesEffect

Akurosia:


ui/raidboss/data/04-sb/trial/lakshmi-ex.js
```diff
@@ -28,10 +28,12 @@
       },
     },
     {
-      regex: / 1E:\y{ObjectId}:Lakshmi loses the effect of Chanchala/,
-      regexDe: / 1E:\y{ObjectId}:Lakshmi loses the effect of Chanchala/,
-      regexFr: / 1E:\y{ObjectId}:Lakshmi loses the effect of Chanchala/,
-      regexJa: / 1E:\y{ObjectId}:ラクシュミ loses the effect of チャンチャラー/,
+      regex: Regexes.losesEffect({ effect: 'Chanchala', capture: false }),
```
quisquous: I don't know that this matters because Lakshmi is the only thing getting the effect, but this should be target: Lakshmi here to match what was there before.

Akurosia: done

Issue #941, opened 2020-01-09
Raidboss ucob translation issue


Feanarus:
## Description

for the "UCU Nael Quote 12, 13 and 14 the german TTS-line mismatch with the TEXT-line
and in UCU Nael Quote 14 "...Pfadt/," without "t" -> "...Pfad/,"

## Additional information
![ucob](https://user-images.githubusercontent.com/38397799/72109834-03e86080-3337-11ea-9023-b092c5539421.PNG)

Akurosia:
just made a PR #945 for this

PR #945, opened 2020-01-11
fixed quote + tts for 12/13/14 (related to #941)

Akurosia:


quisquous:
Thanks.  I wonder if a bunch of explicit `tts` fields can just get removed now that `=>` turns into the right thing automatically, but I think this is the smaller fix for now.

PR #953, opened 2020-01-11
Change all 14 regexes to Regexes.startsUsing for arr

Akurosia:


Akurosia:
massed up some github stuff... will PR again after fixed

PR #958, opened 2020-01-12
added Regexes.hasHP to convert 0D trigger

Akurosia:


resources/regexes.js
```diff
@@ -283,6 +283,20 @@ var Regexes = {
     return Regexes.parse(str);
   },
 
+  // fields: name, hp, capture
+  // matches: https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#0d-combatanthp
+  hasHP: (f) => {
```
quisquous: Can you also add a few test lines for this in `test/unittests/regex_test.js`?

PR #967, opened 2020-01-14
Raidboss: KR update for UCU and minor changes on innocence, titania

Bing-su:
I don't want to try UCU... so I haven't tested it. but I think I understand all triggers correctly.


quisquous:
Thanks!

PR #970, opened 2020-01-15
radar: pass options to PlaySound

rreminy:
I spotted options wasn't being passed to PlaySound after the cleanup was reverted, this commit re-adds it.

Reason the cleanup broke PlaySound is because of lines 116-119, which creates a local version of the variable options, which is then used for PlaySound.


quisquous:
Whoops, thanks.  Sorry for breaking it twice.  :sob:

PR #974, opened 2020-01-17
fix issue when //FIXME is the last element

Akurosia:
This will prevent the `json.decoder.JSONDecodeError` if the last element of a timeline replacement ends with ` // FIXME`

### DIFF
--- a/util/check_translation.py
+++ b/util/check_translation.py
@@ -150,7 +150,7 @@ def parse_translations(triggers):
                 line = line.strip()
 
                 # remove comments (assume no comments in strings)
-                line = re.sub(r'//.*$', '', line)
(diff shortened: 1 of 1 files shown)


util/check_translation.py
```diff
@@ -150,7 +150,7 @@ def parse_translations(triggers):
                 line = line.strip()
 
                 # remove comments (assume no comments in strings)
-                line = re.sub(r'//.*$', '', line)
+                line = re.sub(r'\s?//.*$', '', line)
```
quisquous: Should this be `\s*`?

Akurosia: we can also do `\s*` but at least for all of my submissions it always matches `\s?//.*$` but will add your suggestion :3

PR #977, opened 2020-01-18
add replace of \d to avoid keyerror

Akurosia:
escape \d in timeline *.txt (e.g. o10s.txt) files to avoid KeyError/ re.error: bad escape \d

### DIFF
--- a/util/check_translation.py
+++ b/util/check_translation.py
@@ -221,6 +221,8 @@ def translate_timeline(line, trans):
     for old, new in trans['replaceSync'].items():
         did_work = did_work or re.search(old, text)
         text = re.sub(old, new, text)
+    # escape \d in timeline *.txt files to avoid KeyError/ re.error: bad escape \d
+    text = text.replace("\d", "\\\d")
     line = replace_sync_re.sub('sync /%s/' % text, line)
     if not did_work:
         line = line + ' #MISSINGSYNC'


util/check_translation.py
```diff
@@ -221,6 +221,8 @@ def translate_timeline(line, trans):
     for old, new in trans['replaceSync'].items():
         did_work = did_work or re.search(old, text)
         text = re.sub(old, new, text)
+    # escape \d in timeline *.txt files to avoid KeyError/ re.error: bad escape \d
```
quisquous: Should more things be escaped here?

Akurosia: this is currently the only issue i had (nearly done with stormblood)

quisquous: Which file was this failing on?

Akurosia: it was on o10s.js
the command i used was: `python check_translation.py -f o10s.js -t de -gm all`

quisquous: Thanks.  Sorry, I haven't had a chance to look into this more yet.  I wanted to understand the problem better and see if there was a more general solution.

quisquous:
Ok, I fixed this more generally in 82bf096.  Sorry for the delay, I just wanted to cover all the potential future issues of escaping characters.  Thanks for the pull request.

PR #978, opened 2020-01-20
ci: Add Python Formatting Check to CI

panicstevenson:
Add Python formatting check to Travis CI integration. The Python
formatter of choice is black, since it is unwaivering. For now, to make
pull request readability a thing, string normalization is skipped.

### DIFF
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -194,14 +194,21 @@ def get_type(event):
 
 def test_match(event, entry):
     # Normal case. Exclude begincast to avoid false positive match with cast events
-    if 'regex' in entry and re.search(entry['regex'], get_regex(event)) and not entry['special_type'] and get_type(event) != 'begincast':
+    if (
+        'regex' in entry
+        and re.search(entry['regex'], get_regex(event))
+        and not entry['special_type']
+        and get_type(event) != 'begincast'
+    ):
         return True
 
     # File parsing cases
     if isinstance(event, str) and entry['special_type']:
         # Begincast case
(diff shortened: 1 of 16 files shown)


util/test_timeline.py
```diff
@@ -300,15 +317,23 @@ def check_event(event, timelist, state):
             if not state['timeline_stopped']:
                 for other_entry in timelist:
                     if (
-                            'regex' in other_entry and
-                            other_entry['time'] > state['last_jump'] and
-                            other_entry['time'] < entry['time'] and
-                            other_entry['branch'] < entry['branch']
+                        'regex' in other_entry
+                        and other_entry['time'] > state['last_jump']
```
quisquous: I think this "operator at the start of the line" thing is the most personally distasteful style thing I see in this whole change, but I will take it at the expense of automatic formatting for everything else.

panicstevenson: Yea, some of it is gross.. We can still change courses and divert to something else if you have another preference

panicstevenson: In other news, I'm changing all my if statements from
```python
if x
  and y
```
to
```python
if x:
  if y:
```

quisquous: Yeah, I haven't investigated a ton of other ones.  Did you have anything else in mind, or should I go looking?

quisquous:
I looked at other formatters and I think this is good enough for me.  I think the only thing really missing is a convenient npm command to run black with the right args and reformat when there are errors, along the lines of `npm run lintfix`.

PR #979, opened 2020-01-20
Fix alertTexts for Dun Scaith

ngld:
This fixes a small issue with the `alertText` functions in two Dun Scaith triggers. During the regex overhaul the conditions were changed to check the wrong match groups.
After this change the alerts show up in the emulator again. Haven't done any further tests.

I'll see if the other two recently reported trigger issues are caused by similar mistakes.


ngld:
The Twintania issue is unrelated since that file hasn't been updated for the new regex helpers, yet.

I'm not sure about the E4S bug but it looks like the trigger is trying to compare strings and numbers. I don't know if that works but I'd stick a `parseInt` in there to be safe.

quisquous:
Oof.  Thanks for this fix.  I tried very hard to be meticulous reading through all the regex changes, but it's too easy to make mistakes.

PR #981, opened 2020-01-22
Add ability to install dependencies per script

Akurosia:
This is a copied versin of https://github.com/ngld/OverlayPlugin/blob/master/tools/fetch_deps.py so credits go to ngld

Please ignore the commits... my brain could not handle git yesterday... <.<


README.md
```diff
@@ -105,8 +105,15 @@ For more instructions about adding and configuring cactbot overlays, see the [Ad
 
 Follow all the steps above for installing cactbot first.
 
-1. Follow the instructions in the **dummy.txt** file in [plugin/ThirdParty/OverlayPlugin](plugin/CactbotOverlay/ThirdParty/OverlayPlugin).
-1. Follow the instructions in the **dummy.txt** file in [plugin/ThirdParty/ACT](plugin/CactbotOverlay/ThirdParty/ACT).
+##### Precondition:
+1. `curl` MUST be installed (this is used to download dependencies)
```
quisquous: What version of curl is required?

I get this error: ```curl: (35) error:1407742E:SSL routines:SSL23_GET_SERVER_HELLO:tlsv1 alert protocol version```

plugin/ThirdParty/FFXIV_ACT/dummy.txt
```diff
@@ -0,0 +1,15 @@
+Please download the latest SDK Zipfile from https://github.com/ravahn/FFXIV_ACT_Plugin/ (make sure the file says SDK in its name)
```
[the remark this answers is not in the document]
Akurosia: changed `Zipfile` to `Zip file`

tools/fetch_deps.py
```diff
@@ -0,0 +1,204 @@
+import sys
+import os.path
```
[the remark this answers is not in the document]
Akurosia: moved fetch_deps.py and DEPS.py to util

plugin/ThirdParty/FFXIV_ACT/dummy.txt
```diff
@@ -0,0 +1,15 @@
...
+Extract the "SDK" folder as well as the FFXIV_ACT_Plugin.dll
+
+OverlayPlugin
```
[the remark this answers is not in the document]
Akurosia: How can you even spot all these copy paste issues from me Q_Q

README.md
Akurosia: currently i use following `curl 7.59.0 (x86_64-w64-mingw32) libcurl/7.59.0 OpenSSL/1.0.2h (WinSSL) zlib/1.2.8 WinIDN libssh2/1.7.0_DEV
Release-Date: 2018-03-14`
i just put that info into there because i saw @ngld uses curl in his files, maybe he is can better answer, what version really is required or if there is a workaround

ngld: Any recent version should do. That error most likely indicates that your curl build doesn't support TLS 1.2. That seems odd... Did you get your build from [the official website](https://curl.haxx.se/download.html#Win64)?

If you have Windows 10, you can use the preinstalled curl so this shouldn't be an issue in the future.

ngld:
Why did you update the SDK version? In my experience, users have to use the same parser version you compiled your plugin with *or a newer version*. This means that it's very desirable to compile with an old SDK because that means your plugin will remain compatible with older parser versions which is very important for Korean and Chinese players.

Akurosia:
@ngld Ohh sry never knew that.
I knot that familiar with building stuff and what to look out for (my thoughts were, new SDK == new functions)
@quisquous if you want a specific version here, i can change it.

panicstevenson:
I don't really know where to put this discussion or what would be best, but I agree on having some sort of manifest of dependencies (I've brought this up before since it's dangerous to have automated builds pulling latest), but there's also a proliferation of dependencies that cross multiple languages. For instance, when discussing the CI-side of python, it's hard to put python linting/formatting into npm commands. I feel like this crosses that boundary in a similar way; it feels weird (to me) if I'm wanting to update the C# side of things to have to have any python knowledge (even so much as having it installed), or vice versa.

Not attempting to block on anything, just something I've had thoughts on for a while, and we're going to go deeper down this hole with changes like this so I thought I'd put something out there. I don't know how necessary it is to be friendly to new developers for the time being, but having that bleed between domains can cause friction.

ngld:
@Akurosia You accidentally commited the `DEPS.cache` file (probably because the entry in `.gitignore` is wrong).
@panicstevenson I honestly don't see the problem. All of the other utility scripts (even those that generate JSON) are written in Python.
Putting the linting commands for Python in the `package.json` makes much more sense to me than embedding those in shell scripts or elsewhere. It's much more intuitive (to me) to have all commands like this in a single place instead of separating them by language which seems pretty arbitrary to me.

Finally, you don't need any Python knowledge to update the C# dependencies since those are contained in the manifest. That file is a python script as well but JSON knowledge is enough to understand the structure and update/modify it.

True, you could rewrite the script in C# but how would that help anybody? It'd just make everything more complicated since then you'd have to compile it before you can install the dependencies which in turn are needed to compile the remaining C# code.

quisquous:
Yeah, I'm not particularly sure I see the issue here.  If I have reservations about dependencies, it's usually about adding new dependencies to the project as a whole.  I think I assume that anybody contributing cactbot code will likely install all the node / python / c# dependencies, and I don't really have an issue with making people do that.  I also think Python is just too useful to mandate pushing everything into node and Javascript, and so I'm personally fine with the current mix of Python/node that we have on the testing and automation side.

If things get more entangled and complex, please bring this up again and we can chat more about how to simplify the world.

plugin/ThirdParty/OverlayPlugin/dummy.txt
```diff
@@ -1,16 +1,13 @@
...
+Please download the latest Zip file from https://github.com/ngld/OverlayPlugin/releases/download/
+Extract the "libs" folder as well as the OverlayPlugin.dll
+
+OverlayPlugin
+|- libs
+|  |- HtmlRenderer.dll
+|  |- Newtonsoft.Json.dll
+|  |- OverlayPlugin.Common.dll
+|  |- OverlayPlugin.Core.dll
+|  |- OverlayPlugin.dll
```
Jaehyuk-Lee: `OverlayPlugin.dll` should be
`OverlayPlugin.Updater.dll`

Jaehyuk-Lee:
I see many conversations up there, but I think dummy.txt files under `ThirdParty` directory should be changed quickly. I had hard time making able to be built because of those wrong information.

plugin/ThirdParty/OverlayPlugin/dummy.txt
Akurosia: I fixed this and added a comment that the tree might be changed by future updates of the dependency.

README.md
quisquous: I apparently was running on an ancient version of git bash.  Updating fixed this unsurprisingly.

util/DEPS.py
```diff
@@ -0,0 +1,24 @@
+# If you're familiar with Chromium or other Google projects:
+#   Sorry, this file actually has nothing to do with gclient.
+#   I just borrowed the basic concept and naming scheme.
+
+deps = {
+    'ACT': {
+        'url': 'https://github.com/EQAditu/AdvancedCombatTracker/releases/download/3.4.5.265/ACTv3.zip',
+        'dest': 'plugin/ThirdParty/ACT',
+        'strip': 0,
+        'hash': ['sha256', 'adf13a38d0938ce90f8e674f8365b227d933b91636ddf72b26c85702f6e3b808'],
```
quisquous: What is this a hash of, and how do I update it when the version changes?

util/DEPS.py
```diff
@@ -0,0 +1,24 @@
+# If you're familiar with Chromium or other Google projects:
+#   Sorry, this file actually has nothing to do with gclient.
+#   I just borrowed the basic concept and naming scheme.
+
+deps = {
+    'ACT': {
+        'url': 'https://github.com/EQAditu/AdvancedCombatTracker/releases/download/3.4.5.265/ACTv3.zip',
```
quisquous: I think this needs to be 3.4.5.266 because that's the version the the 2.0.5.0 version of the FFXIV ACT plugin depends on.  Without bumping that, I get an `MSB3277: Found conflicts between different versions of "Advanced Combat Tracker" that could not be resolved.  These reference conflicts are listed in the build log when log verbosity is set to detailed.` warning in my log when building.

util/fetch_deps.py
```diff
@@ -0,0 +1,204 @@
...
+
+                    elif h.hexdigest() != meta['hash'][1]:
+                        print(h.hexdigest())
+                        print(meta['hash'][1])
+                        print('ERROR: %s failed the hash check.' % key)
+                        break
+
+                if os.path.isdir(dest):
+                    print('Removing old files...')
+                    safe_rmtree(dest)
```
quisquous: This also deletes the dummy files, unfortunately.  The dummy files existed to create the directories on disk that were needed, as you can't store an empty directory in git.

Maybe it would be better to put the instructions into the markdown file and delete the dummy text files? Alternatively, move it into a markdown file in the ThirdParty directory (which doesn't get deleted).

quisquous:
> Why did you update the SDK version? In my experience, users have to use the same parser version you compiled your plugin with _or a newer version_. This means that it's very desirable to compile with an old SDK because that means your plugin will remain compatible with older parser versions which is very important for Korean and Chinese players.

For better or for worse, I've been compiling against latest versions for a while now and nobody has complained, so I think this might be ok for now given the lack of complaints so far.  It'll only get more loose as time moves forward and these versions are pinned.

Also, the Korean version appears as if it is always ahead: `FFXIV_ACT_Plugin_Korean_2.1.0.3.zip`, and so that shouldn't be an issue.

util/DEPS.py
ngld: The hash is generated from the file content. It's just to make sure the downloaded file is what we expect.
You can update the hash by running the `fetch_deps.py` script with the `--update-hashes` option. The logic is a bit hacky (it does a simple string replace and replaces the old hash with the new hash). If you were to use JSON instead of Python as the format for the `DEPS.py` file (and rename it to `DEPS.json` or whatever), this could be improved (since you can just serialize the modified object again).

util/DEPS.py
ngld: This stuff was copied from OverlayPlugin and I'm building against an older version of the FFXIV plugin (2.0.4.0). I think Akurosia bumped the plugin version and didn't touch the ACT version.

util/fetch_deps.py
ngld: I think the instructions would be easier to find in the README file. The problem with the dummy files is that there's no reason to look at them until you get some kind of error about missing files in that directory.

The script was originally written with the assumption that the directories listed in the `DEPS` file would be completely managed by the script (including updating) and removing the whole directory is the safest way to get rid of old files from previous dependency versions.

util/DEPS.py
ngld: BTW here's the message from back then and why I haven't updated the FFXIV plugin, yet: https://github.com/ngld/OverlayPlugin/issues/43

Apparently that's no longer the case, though. I'll see if I can figure out what the current versions for CN players are since I'd expect them to run into the same issue.

quisquous: Mmm, that's interesting.  I am quite surprised that no Chinese players have complained about Cactbot dependencies, then.

quisquous: I think I will land this and then bump the ffxiv plugin version back to 2.0.4.0 to match OverlayPlugin.

Issue #57, opened 2018-02-19
Add fr localization


quisquous:
I think the first step is to add a locale file like hibiyasleep did for pull #23.

Timelines and triggers are their own thing.  The current trigger format looks like this:
```
{
  id: 'O8S Shockwave',
  regex: / 14:28DB:Graven Image starts using Shockwave/,
  delaySeconds: 5,
  alertText: 'Look for Knockback',
  tts: 'knockback',
},
```

I think we could do something like this:
```
{
  id: 'O8S Shockwave',
  regex: / 14:28DB:Graven Image starts using Shockwave/,
  regexFr: /something fr here/,
  regexKo: /something ko here/,
  delaySeconds: 5,
  alertText: {
    en: 'Look for Knockback',
    fr: 'fr text here',
    ko: 'ko text here',
  },
  tts: {
    en: 'knockback',
    fr: 'fr tts here',
    ko: 'ko tts here',
  },
},
```

For timelines, one option that I'm not really fond of is to just have entirely duplicate timelines for each language.  That seems less than great, because then if somebody ever resyncs times in a timeline or something, they'd have to adjust everything independently.  I was thinking of maybe being able to just replace ability names from the english version.

So, I was thinking of something like this, for o8s.txt:

```
{
  replaceNames: {
    'Graven Image': 'fr translation',
    'Mana Charge': 'fr translation2',
  },
  replaceSyncs: {
    'Kefka': 'fr kefka mob name',
    'Graven Image': 'fr graven image mob name',
  },
},
````

I was thinking of just applying them all as search and replace regex substitutions, with a different block for replacing the timeline text in the bars from the replacing the syncs.

Amiral-Benson:
I like the solution for Triggers, it's very clean and simple to work with.

As for the timeline solution, I too, would rather not have to deal with separate files if possible, especially with that kind of files, subject to many changes right after the release of a fight, and I find your solution of replacing names and syncs like this quite suitable.

quisquous:
I think this roughly should be enough for you to get started on raidboss translation.  Let me know if you have any questions or if any of this doesn't work.  I would suggest starting with the Middle La Noscea test.js/test.txt triggers and timeline to make sure this all works.  Please replace my "You poke the striking dummy in French" jokes with something more real.  ;)

[...]

PR #987, opened 2020-01-23
added timeline translation for sb

Akurosia:


ui/raidboss/data/04-sb/alliance/orbonne_monastery.txt
```diff
@@ -301,7 +301,7 @@ hideall "--sync--"
 6004.0 "--sync--" sync /:38C2:Ultima, the High Seraph/ window 1000,0
 6007.0 "Demi-Virgo Line" sync /:Ultima, the High Seraph:38C2:/
 6016.0 "East/West March" sync /:Ultima, the High Seraph:38A[01]:/
-6018.0 "Ray Of Light" sync /:Dominion:38B7:/
+6018.0 "Ray of Light" sync /:Dominion:38B7:/
```
quisquous: This does make me wonder if timeline replacements should also be case insensitive.

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.js
```diff
@@ -293,4 +293,461 @@
       },
     },
   ],
+  timelineReplace: [
+    {
+      'locale': 'de',
+      'replaceSync': {
+        'Added new combatant Archaeodemon': 'Added new combatant Archaeodemon', // FIXME
+        'Argath Thadalfus starts using Crippling Blow': 'Argath Thadalfus starts using Crippling Blow', // FIXME
```
quisquous: This seems incorrect.  I have plenty of logs if you need them to make better timelines from, but I would much rather have all of these be replaced with just `Argath Thadalfus` and to change all the syncs to be more properly something like `:SomeId:Argath Thadalfus` in the timeline sync.

Sorry for the bad timeline.  This was before better tools got made and I did this one by hand.

ui/raidboss/data/04-sb/eureka/eureka_hydatos.js
```diff
@@ -1075,108 +1075,479 @@
...
+        'Cloudlarker': 'クラウドラーカー',
+        'Ivory Palm': '白き手',
+        'Orlasrach': 'オールラスラッハ',
+        'Owain': 'オーウェン',
+        'Ozma': 'オズマ',
+        'Ozmashade': 'オズマの影',
+        'Proto Ozma': 'プロトオズマ',
+        'Raiden': 'ライディーン',
+        'Relative Virtue': 'レラティブヴァーチュー',
+        'Removing combatant Proto Ozma': 'Removing combatant Proto Ozma', // FIXME
```
quisquous: I see this and other Added combatants.  I don't think you need to include the Removing combatant part, just the name (which is translated above).

ui/raidboss/data/04-sb/raid/o11s.js
```diff
@@ -460,12 +461,15 @@
     {
       'locale': 'cn',
       'replaceSync': {
+        'Augmented Rocket Punch': '强化型火箭飞拳',
         'Engage!': '战斗开始！',
         'Level Checker': '等级检测仪',
         'Omega': '欧米茄',
-        'rocket punch': '火箭飞拳',
+        'Rocket Punch': '火箭飞拳',
```
quisquous: Does this need a ^ and $ on it, since it will conflict with Augmented Rocket Punch?

ui/raidboss/data/04-sb/raid/o11s.js
```diff
@@ -498,24 +504,25 @@
         'Repel': '相斥',
         'Reset': '初始化',
         'Rush': '突进',
-        'Starboard Wave Cannon Surge': '右舷齐射·零式波动炮',
         'Starboard Wave Cannon': '右舷齐射·波动炮',
+        'Starboard Wave Cannon Surge': '右舷齐射·零式波动炮',
```
quisquous: Similar questions here about `Starboard Wave Cannon` conflicting with `Starboard Wave Cannon Surge`.  I think your reordering breaks this.

Right now, all of these apply in order, so having Surge before non-Surge works properly.  However, alphabetizing breaks this.  I am not opposed to the idea that exactly one sync replacement should apply to a text, but it would be better to have testing for that and to add extra regex like ^ and $ where appropriate to ensure that.

ui/raidboss/data/04-sb/raid/o2s.js
```diff
@@ -415,8 +423,10 @@
         'Main Quake': '本震',
         'Maniacal Probe': '触手乱舞',
         'Paranormal Wave': '呪詛波',
+        'Probes': 'Probes', // FIXME
         'Tremblor': '地震',
-        'Unknown Ability': 'Unknown Ability',
+        'Unknown Ability': 'Unknown Ability', // FIXME
```
quisquous: :thinking:

ui/raidboss/data/04-sb/raid/o3s.js
```diff
@@ -562,5 +614,165 @@
...
+      'locale': 'ko',
+      'replaceSync': {
+        'Ancient Tome': '고대의 책',
+        'Apanda': '아판다',
+        'Engage!': '전투 시작!',
+        'Great Dragon': '거대 드래곤',
+        'Halicarnassus': '할리카르나소스',
+        'Soul Reaper': '영혼 수확자',
+        'White Flame': '하얀 불꽃',
+        'casts Haste III': 'casts Haste III', // FIXME
```
quisquous: Same comments as elsewhere about fixing the timelines to not require this.

ui/raidboss/data/04-sb/raid/o4s.js
```diff
@@ -1035,72 +1035,138 @@
...
         'Zombification': 'Zombie',
       },
     },
     {
       'locale': 'fr',
       'replaceSync': {
-        'Black Hole': 'Trou Noir',
-        'Engage!': 'À l\'attaque',
+        'Black Hole': 'Trou noir',
+        'Engage!': 'À l\'attaque !',
```
quisquous: I am a little bit skeptical about you adding this in everywhere.  I think it would be better to not have it.  My memory (maybe wrong?) is that this isn't a space character at the end, and so it's very easy to get wrong.

ui/raidboss/data/04-sb/raid/o4s.js
```diff
@@ -1035,72 +1035,138 @@
...
         'Exdeath': 'Exdeath',
+        'uses Almagest': 'uses Almagest', // FIXME
```
quisquous: More things that need to be fixed in the timeline.

quisquous:
Oof this pull request is a monster!

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.js
```diff
@@ -1952,71 +1954,67 @@
...
+        'Flames Of Rebirth': 'Flammen der Wiedergeburt',
         'Flare Breath': 'Flare-Atem',
         'Flatten': 'Einebnen',
         'Generate': 'Formung',
         'Gigaflare': 'Gigaflare',
         'Grand Octet': 'Großes Oktett',
         'Hatch': 'Austritt',
-        'Heavensfall(?! )': 'Himmelssturz',
         'Heavensfall Trio': 'Himmelssturz-Trio',
+        'Heavensfall(?! )': 'Himmelssturz',
```
quisquous: Correcting what I said above, about using `^` and `$` above, but I think that's not what we want.  I think something like this negative lookahead (or a negative lookbehind) regex is what we need to add to potentially conflicting replacements.

ui/raidboss/data/04-sb/alliance/orbonne_monastery.txt
Akurosia: this would be great

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.js
Akurosia: I agree, but currently i am not that familiar with creating timelines (need to look at your guides for this)
Should i remove this one? or should it stay until better timelines are there?

ui/raidboss/data/04-sb/raid/o11s.js
Akurosia: in this case it would work without a problem (alphabetical order, so Augmented Rocket Punch would always comes first).

ui/raidboss/data/04-sb/raid/o4s.js
Akurosia: In the game files it is a `<Indent/>` 
This is just to make them all be the same (see merged arr ones)
In my ARR PR, there was also the topic of my script not handling the indent very well.
So want me to replace all of them?

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.js
Akurosia: dou you have an example of the lookahead (e.g. for augmented rocket punch)?
Than i could add these to all fights where i know these happend (e.g. run, copy, paste which were not mentioned in the comments so far)

ui/raidboss/data/04-sb/raid/o2s.js
Akurosia: this is just a relic of old times...
i just starded with the existing timelines and this line was part of it :x

Akurosia:
@quisquous does currently only 1 timeline replacement happening? there might be some issues with `in` and `out` calls than

ui/raidboss/data/04-sb/eureka/eureka_hydatos.js
Akurosia: hope i catched all of them :)

ui/raidboss/data/04-sb/raid/o11s.js
Akurosia: done with lookahead/lookbehind

ui/raidboss/data/04-sb/raid/o2s.js
Akurosia: removed this old relik

ui/raidboss/data/04-sb/raid/o4s.js
Akurosia: the indent issue is also the part of #990 
is there a known rule how the indent is processed by the game?

quisquous:
> @quisquous does currently only 1 timeline replacement happening? there might be some issues with `in` and `out` calls than

Sorry for the confusion, but no.  Currently all replacements happen in key order.

I was trying to propose a hypothetical future feature where only 1 replacement happens so that we could better test (and prevent) issues of ordering and partial replacements that aren't desired.  I'll try to do a better job of differentiating between "this is how things work" and "here is a hypothetical feature if we wanted to make them work this way".

ui/raidboss/data/04-sb/raid/o4s.js
quisquous: Sorry to be forgetful, but I forgot that ACT strips the whitespace before the exclamation (or did in the past).  So, technically `À l\'attaque!` is correct.  However, I wonder if it would be safer and more future proof to just drop the space and exclamation point in case this ever changes in ACT.

See this huge thread: https://github.com/quisquous/cactbot/issues/57#issuecomment-371376181

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.js
quisquous: One option is to remove files that have this problem and follow up with other pull requests to make these timelines better.

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.txt
```diff
@@ -13,26 +13,26 @@ hideall "--start--"
 
 ### Mateus
 0.0 "--start--" sync /:The Crumbling Bridge will be sealed off/ window 0,1
-23.0 "Unbind" sync /:Mateus, The Corrupt starts using Unbind/ window 23,10 # start of cast
+23.0 "Unbind" sync /:2633:Mateus, The Corrupt:/ window 23,10 # start of cast
```
quisquous: This is close, but you need to drop the `:` at the end.  (I wonder if `npm run test` will complain at you about this.)

Here's an example of a sync for a `14` and a `15` line for the same ability:
https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/05-shb/raid/e4s.txt#L17-L18

For succinctness, timelines either are:
* 14 lines, `:ability id:name` (no trailing colon, because " starts using ability name" comes directly after)
* 15/16 lines, `:name:ability id:` (trailing colon)

Akurosia: `npm run test` will not throw an error

ui/raidboss/data/04-sb/raid/o4s.js
Akurosia: for sb content i removed the ` !` now.

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.js
```diff
@@ -298,24 +298,10 @@
...
         'Frostwave': 'Polarlanze',
-        'Hashmal, Bringer Of Order starts using Control Tower on Hashmal': 'Hashmal, Bringer Of Order starts using Control Tower on Hashmal', // FIXME
-        'Hashmal, Bringer Of Order starts using Extreme Edge': 'Hashmal, Bringer Of Order starts using Extreme Edge', // FIXME
-        'Hashmal, Bringer Of Order starts using Landwaster': 'Hashmal, Bringer Of Order starts using Landwaster', // FIXME
-        'Hashmal, Bringer Of Order starts using Summon': 'Hashmal, Bringer Of Order starts using Summon', // FIXME
         'I am (?:Revelation|the truth from which you run|the lies upon which you sup)': 'I am (?:Revelation|the truth from which you run|the lies upon which you sup)', // FIXME
-        'Mateus, The Corrupt starts using Rebind': 'Mateus, The Corrupt starts using Rebind', // FIXME
-        'Mateus, The Corrupt starts using Unbind': 'Mateus, The Corrupt starts using Unbind', // FIXME
-        'Mateus, the Corrupt': 'Mateus (?:der|die|das) Peiniger',
+        'Mateus, The Corrupt': 'Mateus (?:der|die|das) Peiniger',
```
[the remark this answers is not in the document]
Akurosia: as the timeline only contains the capitalized version, i dont think we need both version in the translation right?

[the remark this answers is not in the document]
quisquous: Yeah, you only need the one that matches the timeline.  (And it'd be nice to have a future feature to make this case insensitive too so we don't have to worry about it either.)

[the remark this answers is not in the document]
Akurosia: ok than this can be considered done (timline should only contain 1 form)

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.txt
```diff
@@ -41,17 +41,17 @@ hideall "--start--"
 1078.0 "Earth Hammer" duration 13.5 # start of cast -> damage
 1100.0 "Quake IV" # end of cast
 1109.0 "Rock Cutter" # end of cast
-1112.5 "--sync--" sync /:Command Tower starts using Command Tower/ window 100,100
+1112.5 "--sync--" sync /:266E:command tower/ window 100,100
```
quisquous: Is this lowercase?

ui/raidboss/data/04-sb/dungeon/ala_mhigo.js
```diff
@@ -134,112 +134,201 @@
...
-
-        'Art Of The Swell': 'Art de la tempête',
-        'Art Of The Storm': 'Art de l\'orage',
-        'Art Of the Sword': 'Art de l\'épée',
+        'Mana Burst': 'Explosion de mana',
+        'Mindjack': 'Contrainte mentale',
+        'Order To Charge': 'Ordre d\'attaquer',
+        'Order To Fire': 'Ordre d\'attaquer',
+        'Storm, Swell, Sword': 'Tempête, orage, épée',
+        'Storm?/Swell?': 'Tempête?/orage?',
```
quisquous: Aren't these question marks treated as regex question marks?

JLGarber: If they are, it's news to me. I was under the impression when I wrote this in English that the replacements are plain strings.

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.txt
Akurosia: Since i play in german i cant really verify this one.
But my script returned the lower case version (bnpcname > ID:6950)

ui/raidboss/data/04-sb/raid/o12s.js
```diff
@@ -1181,15 +1181,15 @@
         'Optimized Blade Dance': '쾌검난무: 오메가',
         'Optimized Blizzard III': '블리자가: 오메가',
         'Optimized Fire III': '파이라: 오메가',
-        'Optimized Meteor': '메테오: 오메가',
+        'Optimized Meteor': '메테오 : 오메가',
```
quisquous: This is what was in https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/04-sb/raid/o12s.js#L1171.  I am not sure this space is correct, but I have preserved it.

quisquous:
Thanks again for all this work.  My apologies to folks working on translations if this breaks something unintentionally.  I am hoping that having tests from #1008 and sorting timeline translations that things will only get better in the future.

I also wanted to mention that my recommendation for changes like this in the future is to break it apart into even smaller steps.  It's very hard to review something that has both moved and changed code.  If you ever want to sort something, it's better to do the sorting as a separate request from changing it so that the github diff tools work in your favor.

ui/raidboss/data/04-sb/dungeon/ala_mhigo.js
quisquous: Belatedly, the replacements on the right hand side are plain strings, but the left hand side keys are strings that are interpreted as regular expressions.

ui/raidboss/data/04-sb/alliance/royal_city_of_rabanastre.txt
quisquous: As with everything coming out of the FFXIV plugin, it is in proper case.  All regexes are case insensitive anyway, I just didn't understand why this needed a change.

ui/raidboss/data/04-sb/raid/o12s.js
Akurosia: I had a look into the "ko"-csv files and there all posible 4 ids (13092, 13093, 13134, 13135) are all displayed like followed: `메테오: 오메가`

PR #990, opened 2020-01-24
TEA Nisi & Inception French trigger fix + translations

1unie:
This should fix the Nisi and Inception formation RegEx that had ":" without a space beforehand. I've also added some extra translations!

### DIFF
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
@@ -2054,7 +2054,7 @@
             en: 'Purple, no tether: E->W',
             de: 'Lila, keine Verbindung: O->W',
             ja: '接触禁止, 線無し: 東から西へ',
-            fr: 'Violet, pas de lien: E->N',
+            fr: 'Violet, pas de lien: E->O',
           },
           // This person also has the shared sentence.
           '3': {
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
@@ -1060,6 +1060,7 @@
         en: 'avoid ray',
         de: 'Strahl ausweichen',
         ja: 'アポカリ避けて',
+        fr: 'Evitez le rayon',
       },
     },
     {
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
```diff
@@ -574,19 +594,21 @@
...
             ja: 'ノックバック チャージ; 中央向く',
+            fr: 'Poussée Charge; Regarde à l\'intérieur',
           };
         }
         if (isOddNumber) {
           return {
             en: 'Knockback Cleave on YOU',
             de: 'Rückstoß Cleave auf DIR',
-            fr: 'Cleave sur vous',
+            fr: 'Cleave sur TOI',
```
quisquous: Can you explain this change to me? Mostly because other folks doing french translations have been using vous instead of toi everywhere, and so it'd be nice to be consistent.  Maybe worth leaving as vous for now here (and elsewhere) and then chatting with @mooondark at least about what's best to follow up in the future?

mooondark: This is hard to explain. In french 'you' is splitted in two parts: 'tu' when used for a single person or familiar form and 'vous' for two (or more) persons or as the polite form.

I think 'vous' is far better for cactbot. So in order to be consistant, we should keep 'vous' when translating 'you'.

1unie: @quisquous @mooondark Both are correct, and using vous for consistency is absolutely valid. I actually prefer “vous” myself as well but could have sworn I saw “TOI” in other places and wanted to keep it consistent. Vous is better though so please overrule that one.

1unie: I’ll keep in mind to use vous for other translations!

quisquous:
Ok, glad you both are in agreement.  @1unie, do you want to add another commit to this pull request to fix up converting toi into vous?

Also, it looks like you have a few errors. 
 See this for details on how to investigate: https://github.com/quisquous/cactbot/blob/master/CONTRIBUTING.md#build-problems

1unie:
Got it @quisquous! I've corrected the errors and replaced the occurrences of "TOI" for "VOUS".

1unie:
I think that's it for this one @quisquous unless you want me to add/change anything. Let me know if you need me to do anything else!

quisquous:
Thanks! :+1:

PR #992, opened 2020-01-25
KR translation improve

Bing-su:
These are better!

If you have time, please update `static-data.js` in fisher. I'm unable to do that... Thanks.


quisquous:
I'll take a look at the fisher static data, thanks for mentioning it.

PR #994, opened 2020-01-26
localize 'party wipe' text on oopsyraidsy

Jaehyuk-Lee:


ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -543,7 +543,15 @@ class MistakeCollector {
     // wipe then (to make post-wipe deaths more obvious), however this
     // requires making liveList be able to insert items in a sorted
     // manner instead of just being append only.
-    this.OnFullMistakeText('wipe', null, 'Party Wipe');
+    const wipeLang = {
```
quisquous: Can you put this at the top with the `kEarlyPullText` / `kLatePullText` etc, so it isn't hidden down in this function?

quisquous:
Did you mean to include all of these other changes with the party wipe changes?

Also, it looks like you're making all your changes from your master branch, and then merging in upstream cactbot on top of that.  This is causing the list of commits for each pull request to get progressively longer.

Sorry if you know all of this already, but here's some advice:

The simple option is that instead of merging in upstream cactbot, is that if you only have one pull request outstanding and it gets merged in, then you just do a `git reset upstream/cactbot` (or whatever your remote name is) instead of `git merge upstream/cactbot`.  This will reset the log to be whatever is in upstream, rather than your local changes + merges (which turns out to be the same files, but looks different in the log).

The slightly more complicated option is to do all of your feature work in branches.  So before working, you'd do a `git checkout -tb partywipe` or something, do your work, upload it, and then do a pull request against your `partywipe` branch instead of your `master` branch.  Then once your change has been landed, you can delete the partywipe branch, maybe do a `git pull` on the master branch to update it cleanly, and then start the process again.  This will also let you have multiple changes in parallel.

quisquous:
I added some links and documentation to <https://github.com/quisquous/cactbot/blob/master/CONTRIBUTING.md#development-workflow> about feature branch workflow, since a number of people have had similar issues.

PR #999, opened 2020-01-28
TEA RegEx in French fix

1unie:
Seems like some spaces were either forgotten or reverted from the previous commit.


quisquous:
Looking at https://github.com/quisquous/cactbot/pull/990/files#diff-fbaf1c5b952fe90aa04f3bd82317317a, I think these were just not part of the last commit? I don't *think* I overwrote them based on what I'm seeing there.

PR #1000, opened 2020-01-28
force javascript calculate numbers

Jaehyuk-Lee:
someone got problem with this code, (maybe this problem is for everyone)
`this.options.BigBuffIconWidth + 2` was not calculating numbers. was attaching those two as strings.
I think this will fix this problem.

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -990,7 +990,7 @@ class Bars {
       this.o.leftBuffsList.rowcolsize = 7;
       this.o.leftBuffsList.maxnumber = 7;
       this.o.leftBuffsList.toward = 'left down';
-      this.o.rightBuffsList.elementwidth = parseInt(this.options.BigBuffIconWidth) + 2;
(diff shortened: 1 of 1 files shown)


quisquous:
Hmm! Do you know how this happened? Did somebody type letters into the config box, or did they set a file in `cactbot/user/`?

Mostly I'm wondering if it would be better to just do a parseInt/parseFloat during the options setting on everything vs this one variable.

Jaehyuk-Lee:
http://www.inven.co.kr/board/ff14/4953/3298
I don't know how this happened but post writer said string attachment was problem.

quisquous:
Thanks for the link, that's really helpful context. Sorry for writing my own patch here and not taking this pull request, but this seemed like a more general problem.  By catching it upstream, I think this will fix all issues, like BigBuffIconHeight as well as BigBuffIconWidth, and all integer/float values.

Jaehyuk-Lee:
okay, i understand

PR #1007, opened 2020-02-01
Silent version of raidboss for WS+OBS

g0shu:


ui/raidboss/popup-text.js
```diff
@@ -20,6 +20,16 @@ function onTriggerException(trigger, e) {
     console.error(lines[i]);
 }
 
+let params = new URLSearchParams(window.location.search);
```
quisquous: I think this logic seems reasonable, but I have a few small tweaks to suggest:

* is there a reason to have sounds and tts separate? It seems like maybe having a single `audio` param might be enough?
* maybe add a field on `Options` like `audioAllowed` that defaults to true and then if the url param is audio=0 or audio=1 then set it from there.  Then later on you can look up `this.options.audioAllowed` rather than have some free floating globals like `muteSounds`
* consolidate this URLSearchParams logic with the other code that uses it in the bottom of `ui/raidboss/raidboss.js`

quisquous:
Ah, I see now why this got more complicated with the per trigger overrides.  Thanks for pushing through and finishing this off.  :+1:

g0shu:
I don't think `audioAllowed` belongs in `raidboss_config.js`? Is there a better place to put it or just leave it where it is in `raidboss.js`?

quisquous:
> I don't think `audioAllowed` belongs in `raidboss_config.js`? Is there a better place to put it or just leave it where it is in `raidboss.js`?

Yeah, that's perfect.  Agreed that it shouldn't be in `raidboss_config.js`.

quisquous:
Thanks!!

PR #1014, opened 2020-02-05
e3s add arrow and tts

Jaehyuk-Lee:
add arrow and tts for better visibility


ui/raidboss/data/05-shb/raid/e3s.js
```diff
@@ -579,6 +579,14 @@
       regexCn: Regexes.startsUsing({ id: '3FEB', source: '利维亚桑', capture: false }),
       regexKo: Regexes.startsUsing({ id: '3FEB', source: '리바이어선', capture: false }),
       infoText: {
+        en: 'Front left / Back right',
+        de: 'Vorne Links / Hinten Rechts',
+        fr: 'Avant gauche / Arrière droite',
+        ja: '左前 / 右後ろ',
+        cn: '前左 / 后右',
+        ko: '↖ 앞 왼쪽 / 뒤 오른쪽 ↘',
```
quisquous: I think it would be better to ignore these arrows [here](https://github.com/quisquous/cactbot/blob/master/ui/raidboss/popup-text.js#L607) and not have a separate infoText/tts block that mostly duplicates each other.

Jaehyuk-Lee: Ah, there has tts replacer! I didn't know, it's very clever!

Jaehyuk-Lee:
I've added tts replacer rather than making specific tts
` ttsText = ttsText.replace('[↖-↙]', '');`

ui/raidboss/popup-text.js
```diff
@@ -604,6 +604,8 @@ class PopupText {
         ttsText = ttsText.replace(/[#!]/, '');
         // * slashes between mechanics
         ttsText = ttsText.replace('/', ' ');
+        // * arrows helping visually simple to understand e.g. ↖ Front left / Back right ↘
+        ttsText = ttsText.replace('[↖-↙]', '');
```
quisquous: Sorry, one more comment.  I think that maybe this should be `[↖↙↗↘]` with more arrows, and no `-` hyphen?

Jaehyuk-Lee: `[↖-↙]` and `[↖↙↗↘]` is the same. If you want to change this for code readability, I'll change it.

quisquous: Oh oh, I'm sorry.  I should not read code when I am so tired.  You're quite right.

PR #1016, opened 2020-02-05
build update

Jaehyuk-Lee:
I think that `dummy.txt` file is not properly written, so I've edited.
Also, there's no need to add finding `FFXIV_ACT_Plugin_Korean.dll` file.


Akurosia:
isn't this already covered in #981 ?

Jaehyuk-Lee:
@Akurosia I didn't saw it before. I've added a review there with one problem and closed this pull request.

PR #1017, opened 2020-02-05
FFXIV_ACT_Plugin_Korean.dll is not needed

Jaehyuk-Lee:
There is no need to find FFXIV_ACT_Plugin_Korean.dll anymore.


quisquous:
Does nobody use this anymore? Or everybody renames it?

Jaehyuk-Lee:
If you see latest Korean release, [here](https://github.com/ravahn/FFXIV_ACT_Plugin/tree/master/Releases_KR), it's original name does not contain "Korean".
It had been renamed by "해루봇" which is ACT maintainer for Korean users, but it's not renaming anymore.

quisquous:
Thanks for the explanation!

PR #1018, opened 2020-02-05
shb 71-79 dungeons KR

Bing-su:
thanks to purutu.

I don't know German at all, but do with refering to resources/lang/de.js


ui/raidboss/data/05-shb/dungeon/dohn_mheg.txt
```diff
@@ -76,7 +76,7 @@ hideall "--sync--"
 
 ### Aenc Thon, Lord of the Lengthsome Gate
 # -p 35A4:1011 34D2:1300 -ii 34EC 3681 3396
-1000.0 "--sync--" sync /00:0839:The throne room will be sealed off/ window 1000,0
+1000.0 "--sync--" sync /00:0839:The Throne Room will be sealed off/ window 1000,0
```
quisquous: Is this correct? I think this is lower case?

Bing-su: Cap 2020-02-06 19-24-44-700
I just followed the csv file 'PlaceName'. I think it is odd that all others are capital but only it is lower case. If lower case is correct, I'll fix.

Akurosia: aren't the sync caseinsensitive anyways?
As long as the timelineReplacments are the same case it shouldn't be a problem right?

quisquous: Yes, the sync is case insensitive, but it also seems odd to me to change the sync from something that had the correct case to something that does not.  In game for English, it is definitely `The throne room will be sealed off`.

Bing-su: that really odd.
sorry anyway.

ui/raidboss/data/05-shb/dungeon/malikahs_well.js
```diff
@@ -189,5 +190,36 @@
         'Armadillo': 'Armadillo',
       },
     },
+    {
+      'locale': 'ko',
+      'replaceSync': {
+        '00:0839:Terminus will be sealed off': 'Drehscheibe',
```
quisquous: Did you mean for these to not have a translation into Korean?

ui/raidboss/data/05-shb/dungeon/qitana_ravel.js
```diff
@@ -309,5 +310,37 @@
         'Viper Poison': 'Viperngift',
       },
     },
+    {
+      'locale': 'ko',
+      'replaceSync': {
+        'Batsquatch': '배츠콰치',
+        'Eros': '에로스',
+        'Lozatl': '로차틀',
+        '00:0839:The Divine Threshold will be sealed off': '초 후에 신에게 가는 문이 봉쇄됩니다',
```
quisquous: My only other comment here is that it'd be better not to have the `00:0839` as part of the replacement.  For both Korean and German, it'd be better to just have this style: `'The Divine Threshold will be sealed off': '초 후에 신에게 가는 문이 봉쇄됩니다',` so that the 00:0839 (which is part of the line) is still there.

ui/raidboss/data/05-shb/dungeon/malikahs_well.js
Bing-su: I missed that. thanks.

ui/raidboss/data/05-shb/dungeon/qitana_ravel.js
Bing-su: ex) Original EN sync:
`00:0839:The Divine Threshold will be sealed off **in 15 seconds**`

but in KO, DE
KO - 00:0839:**15**초 후에 신에게 가는 문이 봉쇄됩니다
DE - 00:0839:**Noch 15** Sekunden bis sich der Zugang zu Götterpforte schließt

like this, the seconds comes in front of the message.

I think if only back part after '00:0839:' is replaced, the changed message will be
`00:0839:초 후에 신에게 가는 문이 봉쇄됩니다`
and it is not synced with 00:0839:**15**초 후에 신에게 가는 문이 봉쇄됩니다

Writing this comment, I realize I was stupid and there was another good way...
I'll replace these sync texts to include second texts.

PR #1020, opened 2020-02-07
ShB(KR) replacement order problem fixed and about this issue

Jaehyuk-Lee:
Replacement order problem is not only here though. I've finished rearranging for Ultimates.
I think this is very important problem. When making replacement texts, everyone should keep in mind that replacement order MATTERS. I think most timeline replacer are made automatically, letters were ordered by 'abc' order. We need replacement order verification part.

for example,
If 'Burst' is earlier replaced than 'Downburst', 'Downburst' is not correctly changed.
'Burst' changes not only exact 'Burst', but 'Downburst' or 'Vulcan Burst' too.
'Burst' changes to '대폭발' in Korean.
So, 'Downburst' become 'Down대폭발', 'Vulcan Burst' become 'Vulcan 대폭발' in Korean.
So, even if there's 'Downburst' or 'Vulcan Burst' replacer after that, replacement does not happen. javascript can't find those.

This is not a critical problem for timeline texts, but it's very critical when timeline synchronization part is not translated well.

I though this was important issue, so I've made webpage to test replacement. But my programming skill sucks, I couldn't make it perfect, some are not replaced (I think this is regular expression problem).
[Test page link](https://jaehyuk-lee.github.io/cactbot-replace-test) and [Code repo](https://github.com/Jaehyuk-Lee/cactbot-replace-test)
If someone can help with this code, thank you for your help.

Example image, you can easily find something is not translated.
K-002

I think it's good to add replacement tester inside cactbot repository. I didn't add here, in the repository, because I couldn't make it perfect.


Akurosia:
Fixing the order would just temporarly solve the problem. There are already test implemented in another branch which catch these and new replacements are already in process which should catch these issues (https://github.com/quisquous/cactbot/tree/timeline_tests)

Also these changes should already be covered here: https://github.com/quisquous/cactbot/pull/987

Jaehyuk-Lee:
Oh, there's regular expression which does not catch others with more characters!
thanks for `(?<![\\w| ])` information.

I think I should resolve this collision after that pull request is accepted.
I'll not close this since this contains ShB timeline fixes.

Akurosia:
#987 has no open replacement issues based on the tests implemented in (https://github.com/quisquous/cactbot/tree/timeline_tests)

Jaehyuk-Lee:
yes, I mean, UCU and UWU change from this branch will be removed after #987 is accepted. Github will show conflicts and I'll change there.

Jaehyuk-Lee:
@quisquous please accept #987 first before accepting this pull request.

Jaehyuk-Lee:
@quisquous finished resolving conflicts

Jaehyuk-Lee:
Ah, I should start again from latest master branch closing this request.

PR #1021, opened 2020-02-08
WIP: Add defensive buffs to jobs.js

anubiann00b:
Hey! This is a proposed change to add raid buff indicators for troubadour, shield samba, tactician, reprisal, addle, and feint. I gave them a property `defensive`, and added a new option `HideDefensiveBuffs` that's true by default that blanket disables them to maintain current functionality.

I started this because I found the raid buffs mode of jobs.js suuuper helpful for prog and static purposes, and additionally wanted mitigation tracking. It's helpful for me to be able to keep track of cooldowns of party members to get good use out of yours when pugging, and it's extra helpful to keep track of everyone's in a static.

Todos:

 * [ ] Consider other buffs worth adding
 * [x] Add icons in `resources/icon/status` (I nabbed the 80x80 ones off gamerescape, seem equivalent to the lodestone ones mentioned at https://xivapi.com/docs/Icons unless I'm missing something)
 * [ ] Check and clean up colors
 * [ ] Validate basic functionality
 * [ ] Add effect name translations
 * [ ] Squash when things are looking finalized

Potential other buffs might be:
- Heart of Light/Dark Missionary/Shake It Off/Divine Veil
- Mantra/Nature's Minne possibly?
- Healer stuff (temperance, fey illumination, asylum) even? (might need to be a third category to avoid bloat)

(I did a little gainAbility -> useAbility refactor btw, since gainEffect/loseEffect/useAbility made more sense to me--let me know if that's not cool and I'll take that commit out)

Thoughts/feedback?


quisquous:
Thanks! This is a really neat idea.  I'll try to take a look this weekend or Monday and give some initial feedback.

Overall, I'm definitely supportive of tracking defensive buffs.  I might want to think a bit more about how to display them (do we need another row rather than everything mixed together), but that seems like a smaller detail.

anubiann00b:
Okay I tried it out a bit today--everything mixed together is an absolute mess, two rows sounds like a good way to go. I wonder if there'd be a more generic way of grouping buffs (maybe with a `group: "offensive|mitigation|healing"` attribute?) and like, giving each group it's own line? Don't think there's a huge use case but /shrug

quisquous:
Sorry for taking so long to get back to this.  I think all the code looks good, just not sure how to present all this information either.

Some miscellaneous replies: I think the longer tank cooldowns would be really useful to see.  Your code renaming seems fine too.  (I think my mind was thinking "this is the ability which causes the effect to be gained, so the gainAbility.  However, I think what you have is much more readable.)  Re: icons.  I don't think there's anything very consistent with where they come from, so as long as they don't look out of place, that's fine with me.

I like the idea of offensive / mitigation / healer categorization.  Does that mean we have three separate rows of icons potentially? I guess there's room for it vertically if we wanted to stack them? Would that approach make sense to you?

This is a side thought too, I have also thought about having people's initials on the cooldowns (since there will often be two reprisals/feints/troubadours etc depending on party comp) so you knew whose was up if you wanted to call somebody out to do throw something.  There's a parameter for text, but it just doesn't show anything when a skill is ready at the moment.

anubiann00b:
Ooh initials on cooldowns would be rad--I've definitely found I've needed that leading raids, multiple addles or feints is fairly common, and there's always two reprisals so. This might be out of scope, but would it be sensible to move the buff appearance to a template file somehow? (not actually sure how to do this in pure JS and imagine it's not worth a whole dependency but I can figure it out/see how it's done elsewhere in the project)

Stacking vertically sounds perfect 👍 

I'll try to make some progress on this tonight since 5.2 is right around the corner haha

Also, I'm not sure where to ask this, but is there a way I could help with making triggers/timelines for the new ex trials and savage raids? Not sure what the process for that is but I end up writing a mini guide and make cactbot trigger modifications for the group I teach anyway (we do the whole tts discord relay thing since we have people on console)

quisquous:
> Ooh initials on cooldowns would be rad--I've definitely found I've needed that leading raids, multiple addles or feints is fairly common, and there's always two reprisals so.

I think that should be pretty straightforward if you wanted to send a different pull request for it.  I'm happy to give you pointers if you need any.

> This might be out of scope, but would it be sensible to move the buff appearance to a template file somehow? (not actually sure how to do this in pure JS and imagine it's not worth a whole dependency but I can figure it out/see how it's done elsewhere in the project)

Do you mean the giant JSON-y blob of all the skills and cooldowns? I guess I'm not sure what you mean by "template" file here, so am not sure what you're proposing, sorry.

> Also, I'm not sure where to ask this, but is there a way I could help with making triggers/timelines for the new ex trials and savage raids?

I am usually fairly fast at getting savage done because I'm also doing it for myself.  This time around I think ACT is going to be very slow to update (maybe a week) and so we won't even have logs for a bit.  There's a [guide](https://github.com/quisquous/cactbot/blob/master/docs/TimelineGuide.md) for making timelines.  Triggers don't really have a guide other than to just follow examples from other triggers.  Feel free to write send pull requests for triggers if I'm being slow, although you may want to check with me in the FFXIV ACT discord if you don't want to spend time on something I'm in the middle of.

> Not sure what the process for that is but I end up writing a mini guide and make cactbot trigger modifications for the group I teach anyway (we do the whole tts discord relay thing since we have people on console)

Oh that's neat! If there's anything that could be made easier in terms of customization, I definitely want to hear about it.

I am sorry about the state of cactbot as a discord relay.  I use it pretty individually (maybe because we have a person who raid calls in my group) and so even though there's "groupTTS" that doesn't end up really getting used anywhere.  I'd definitely be up for discussing more about how to make cactbot better as a discord bot (maybe in another feature, if you wanted to file one)?  Partially I haven't spent time on that because it's just not as relevant to me personally (and I don't use tts on top of that!) and so I don't have a great sense of what people want vs what's too noisy.

quisquous:
Sorry for the incredible delay in doing anything about this patch.  I do think defensive buffs are a good idea, but I don't think having them in the buff list is the way to do it.  I'm going to close this for now, but I'd be happy to chat about alternative approaches.

https://www.youtube.com/watch?v=2ammUknoAAY is an interesting idea, and I think supporting something like that is an interesting idea.

PR #1026, opened 2020-02-09
KR v5.1 trigger translation

Jaehyuk-Lee:


ui/raidboss/data/05-shb/alliance/the_copied_factory.js
```diff
@@ -10,6 +10,7 @@
   zoneRegex: {
     en: /^The Copied Factory$/,
     cn: /^复制工厂废墟$/,
+    ko: /\(700\)/,
```
quisquous: ...what is the zone name here?

Jaehyuk-Lee: I tried `Unknown Zone (700)`
I first thought I need to add Zone ID first, because before updating FFXIV Parsing Plugin, ACT does not show its actual Zone name.
This problem happened when I was distributing `e1s.js` ~ `e4s.js` on Korean FFXIV community right after v5.05 pre-download before server opens for patch.
I added Zone Regex as real name, but FFXIV Parsing Plugin was not updated yet, so Cactbot couldn't recognize the Zone.
I fixed this problem with just adding zoneRegex with its ID.

However, there will be FFXIV Parsing plugin updates on every patches after v5.1. That, Opcodes moves randomly thing. So, this method won't work after v5.1. I didn't thought about this problem when I was writing this regex.

quisquous: Ah, I see.  This makes sense.  This makes me think that we really need a zone id and not a zone regex to avoid this in the future.

PR #1035, opened 2020-02-10
added common easy responses for HW content

Akurosia:


ui/raidboss/data/03-hw/alliance/dun_scaith.js
```diff
@@ -629,10 +604,7 @@
       condition: function(data) {
         return data.role == 'healer';
       },
-      alertText: {
-        en: 'Extreme raid damage!',
-        fr: 'Dommages de raid extrêmes !',
-      },
+      response: Responses.aoe(),
```
quisquous: Maybe we need a big aoe response here.

ui/raidboss/data/03-hw/dungeon/the_vault.js
```diff
@@ -90,17 +88,8 @@
...
-          return {
-            en: 'Spread marker on YOU',
-          };
-        }
-        return {
-          en: 'Avoid ' + data.ShortName(matches.target),
-        };
-      },
+      regex: Regexes.headMarker({ id: '0020', capture: false }),
+      response: Responses.spread(),
```
quisquous: Maybe we need a spreadOn (or antiStackOn?) like we have a stackOn

ui/raidboss/data/03-hw/dungeon/sohm_al.js
```diff
@@ -72,25 +54,8 @@
...
-        if (matches.target == data.me) {
-          return {
-            en: 'meteor',
-            de: 'meteor',
-            fr: 'météore',
-          };
-        }
-      },
+      regex: Regexes.headMarker({ id: '0007', capture: false }),
+      response: Responses.meteorOnYou(),
```
quisquous: I feel like this is not as good as what it is replacing.

ui/raidboss/data/03-hw/dungeon/the_vault.js
Akurosia: What is the difference between spread and spreadOn? Nobody should overlap in a spread anyway so i dont really see the benefit from spreadOn here?
And if you just want to avoid a single person, shouldn't be something similar to .goAwayFrom("x") be used?

ui/raidboss/data/03-hw/dungeon/sohm_al.js
Akurosia: I can understand that you would loose the "on edge" part of the trigger but i think it still better to have it generalized (as all other meteor trigger also dont include a "on X" position part, e.g. t9).

ui/raidboss/data/03-hw/alliance/dun_scaith.js
```diff
@@ -523,27 +474,21 @@
...
+      regexKo: Regexes.gainsEffect({ target: '디아볼로스', effect: '밤의 방패' }),
       suppressSeconds: 5,
       condition: function(data) {
         return data.role == 'tank' || data.role == 'healer';
       },
-      alertText: {
-        en: 'Boss hitting hard--Shield/Mitigate',
-        fr: 'Le boss frappe fort - Bouclier/Mitigation',
-      },
+      response: Responses.tankBuster(),
```
quisquous: This is different.  It's not really a tankbuster, it's more like hard autos?

ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.js
```diff
@@ -75,11 +45,7 @@
       regexJa: Regexes.startsUsing({ id: '10EB', source: 'ハルマキス', capture: false }),
       regexCn: Regexes.startsUsing({ id: '10EB', source: '赫鲁玛奇斯', capture: false }),
       regexKo: Regexes.startsUsing({ id: '10EB', source: '하르마키스', capture: false }),
-      infoText: {
-        en: 'Look away',
-        de: 'Wegschauen!',
-        fr: 'Regardez ailleurs',
-      },
+      response: Responses.lookAway(),
```
quisquous: lookAway('info')

ui/raidboss/data/03-hw/dungeon/the_vault.js
```diff
@@ -123,10 +84,7 @@
       condition: function(data) {
         return data.CanStun();
       },
-      infoText: {
-        en: 'Interrupt the Knight',
-        fr: 'Interrompez le chevalier',
-      },
+      response: Responses.interupt(),
```
quisquous: This loses the thing to interrupt?

ui/raidboss/data/03-hw/dungeon/sohm_al.js
quisquous: I guess all meteors go out, more or less.  Ok, I agree here, then!

ui/raidboss/data/03-hw/dungeon/the_vault.js
quisquous: And shouldn't this be stun and not interrupt?

ui/raidboss/data/03-hw/dungeon/the_vault.js
quisquous: Maybe this is awayFrom, then.

ui/raidboss/data/03-hw/alliance/dun_scaith.js
Akurosia: would be miniBuster here ok? or should i revert it?

ui/raidboss/data/03-hw/dungeon/the_vault.js
Akurosia: i think this will be a generall question how to deal with stun/interupts on targets (due to the languages)

should stun/interupt just be used for bosses? and adds will not have response lines?

Akurosia: is there a way to access the "source" from the trigger inside a response? Because than you could just relay on the Source itself (to call out stun target in all languages)

ui/raidboss/data/03-hw/alliance/dun_scaith.js
quisquous: I think the difference here is that this is over a longer period of time, so a buster call on its own might make somebody think it was fine and then over? I'm not sure.  Maybe revert for now and can revisit later.

Akurosia: ok it is now reverted

Akurosia: ok it is now reverted

ui/raidboss/data/03-hw/dungeon/the_vault.js
quisquous: Yes, matches.source should have what you want.

PR #1041, opened 2020-02-11
change to regex

Jaehyuk-Lee:
I am idiot. :(


quisquous:
Haha, whoops! I looked at that and didn't see anything wrong either, so it's not just you.  Maybe too many regexes that are strings elsewhere made me miss this.

PR #1049, opened 2020-02-16
add common replacement

Jaehyuk-Lee:
I've done some code with common replacement.

`ui/raidboss/timeline.js` part can be improved much, but I don't know how this code exactly works. I've just added new function, but it's having error.
Suggestions to this code is needed.
```
TypeError: Cannot read property 'Language' of undefined
    at Timeline.GetReplacedCommon (/home/travis/build/Jaehyuk-Lee/cactbot/ui/raidboss/timeline.js:61:31)
    at Timeline.GetReplacedText (/home/travis/build/Jaehyuk-Lee/cactbot/ui/raidboss/timeline.js:71:17)
    at Timeline.LoadFile (/home/travis/build/Jaehyuk-Lee/cactbot/ui/raidboss/timeline.js:166:20)
    at new Timeline (/home/travis/build/Jaehyuk-Lee/cactbot/ui/raidboss/timeline.js:37:10)
    at Object.<anonymous> (/home/travis/build/Jaehyuk-Lee/cactbot/test/timeline/test_timeline.js:22:16)
    at Module._compile (internal/modules/cjs/loader.js:956:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:973:10)
    at Module.load (internal/modules/cjs/loader.js:812:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.Module.runMain (internal/modules/cjs/loader.js:1025:10)
/home/travis/build/Jaehyuk-Lee/cactbot/ui/raidboss/timeline.js:61
    let locale = this.options.Language || 'en';
```
[error log on travis-ci](https://travis-ci.com/Jaehyuk-Lee/cactbot/jobs/287864890)

 * I think common replacements should include only those are really commonly used strings. Like, at least five times used.

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -60,9 +60,9 @@ class Timeline {
   GetReplacedCommon(text) {
     let locale = this.options.Language || 'en';
     let keys = Object.keys(commonReplacement);
-    for (let j = 0; j < keys.length; ++j){
-        let re = new RegExp(keys[j],"gi");
-        text = text.replace(re, commonReplacement[keys[j]][locale])
+    for (let j = 0; j < keys.length; ++j) {
+      let re = new RegExp(keys[j], 'gi');
+      text = text.replace(re, commonReplacement[keys[j]][locale]);
     }
     return text;
   }
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -57,12 +57,22 @@ class Timeline {
     return text;
   }
 
+  GetReplacedCommon(text) {
+    let locale = this.options.Language || 'en';
+    let keys = Object.keys(commonReplacement);
+    for (let j = 0; j < keys.length; ++j){
+        let re = new RegExp(keys[j],"gi");
+        text = text.replace(re, commonReplacement[keys[j]][locale])
+    }
+    return text;
+  }
+
   GetReplacedText(text) {
-    return this.GetReplacedHelper(text, 'replaceText');
+    return this.GetReplacedCommon(this.GetReplacedHelper(text, 'replaceText'));
   }
 
   GetReplacedSync(sync) {
-    return this.GetReplacedHelper(sync, 'replaceSync');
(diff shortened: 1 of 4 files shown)


ui/raidboss/timeline.js
```diff
@@ -57,12 +57,22 @@ class Timeline {
...
+    for (let j = 0; j < keys.length; ++j) {
+      let re = new RegExp(keys[j], 'gi');
+      text = text.replace(re, commonReplacement[keys[j]][locale]);
+    }
+    return text;
+  }
+
   GetReplacedText(text) {
-    return this.GetReplacedHelper(text, 'replaceText');
+    return this.GetReplacedCommon(this.GetReplacedHelper(text, 'replaceText'));
```
quisquous: Can GetReplacedCommon be called within GetReplacedHelper or GetReplacedCommon just be part of that function?

I think if you move this code after the `if (!this.replacements)` in GetReplacedCommon, then you won't have the test error.  The issue is that the test is just importing a few files, and `this.options` is undefined because no options is being passed to the timeline constructor.  It might be worth having the timeline constructor set options to {} if it's not defined to avoid this sort of thing in the future too.

Jaehyuk-Lee: Oh, yes. I first thought I need just one time to call this function, logically. So, I thought I need this separately. (like if I put inside of that, this function will called twice.)
but I don't know how this code works, so I changed to minimize impact of these changes.
So, I left everything same, but added new function but I remembered I decided not to add these code inside that. It doesn't have to be but I remembered only that so I made some... mess.

Jaehyuk-Lee:
For `test_timeline.js`... I cannot understand every code, at least for now. I should see many times to understand this...
This will take long time to do it myself..

quisquous:
> For `test_timeline.js`... I cannot understand every single code, at least for now. I should see many times to understand this...
> This will take long time to do it myself..

Ok, I"ll take a look myself.  Thanks for this patch!

Jaehyuk-Lee:
Thanks :smiley:

PR #1050, opened 2020-02-16
Include common translations in timeline tests

quisquous:
This is a followup to #1049.


quisquous:
@Jaehyuk-Lee @Akurosia @mooondark  @panicstevenson 

There's a lot of changes here, so pinging folks who care about translations if you want to take a look over this "tiny" +242 −1,535 patch.

This is the end result of fixing up everything to account for the new "common translations" from #1049.  I'm especially not sure about the French translations in terms of what's reasonable to include in the common half vs in the individual files.  I think what's there will work now, but it's maybe not...ergonomic.

Akurosia:
looks good from my side :)

ui/raidboss/data/04-sb/alliance/ridorana_lighthouse.js
```diff
@@ -846,14 +846,14 @@
         'Construct 7': 'Bâtisseur n°7',
         'Construct 7.1': 'Bâtisseur n°7 amélioré',
         'Dark Rain': 'Trombe d\'eau',
-        'Echoes from Time\'s Garden will be sealed off': 'Fermeture du Jardin d\'un autre temps dans',
+        'Echoes from Time\'s Garden': 'Jardin d\'un autre temps dans',
```
mooondark: No need to have _"dans"_ in this sentence I guess. The full sentence is _"Fermeture du Jardin d\'un autre temps dans 15 secondes."_ so _dans_ means _in_.

ui/raidboss/data/04-sb/alliance/ridorana_lighthouse.js
```diff
@@ -846,14 +846,14 @@
...
         'Famfrit, The Darkening Cloud': 'Famfrit le Nuage Ténébreux',
         '(?<! )Gigas': 'serviteur de Bélias',
         'Heart of the Dragon': 'cœur de Yiazmat',
         'Missile': 'Missile',
-        'The Cleft of Profaning Wind will be sealed off': 'Fermeture de la corniche des Vents distordants dans',
-        'The Clockwork Coliseum will be sealed off': 'Fermeture de l\'arène de Goug dans',
-        'The Spire\'s Bounds will be sealed off': 'Fermeture de l\'orée des Cieux dans',
+        'The Cleft of Profaning Wind': 'la corniche des Vents distordants dans',
```
mooondark: Same here (remove _dans_) and same for lines 855 & 856.

Jaehyuk-Lee:
I thought "attack" is okay to be in common_replacement because it's done after everything is translated from replaceText and replaceSync.
if "something attack" is translated "어떤 공격" by replaceText, "attack" common replacement does nothing since it's already translated to other.

ui/raidboss/common_replacement.js
```diff
@@ -1,18 +1,18 @@
...
 
+// TODO: maybe this should be structured identically to a timelineReplace section.
 let commonReplacement = {
   ':([0-9]{4}):(.*) will be sealed off': {
-    de: ':$1:Noch 15 Sekunden bis sich der Zugang zu $2 schließt',
-    fr: ':$1:Fermeture $2 dans',
+    de: ':$1:Noch 15 Sekunden bis sich (?:das|die|der Zugang zu[rm]?) $2 schließt',
+    cn: ':$1:距$2被封锁还有',
+    fr: ':$1:Fermeture d(?:e|u|es) $2 dans',
     ja: '$1:$2の封鎖まであと',
```
Jaehyuk-Lee: I think I made mistake here. `ja: ':$1:$2の封鎖まであと',` is right I think.

quisquous:
> I thought "attack" is okay to be in common_replacement because it's done after everything is translated from replaceText and replaceSync.
> if "something attack" is translated to "어떤 공격" by replaceText, "attack" common replacement does nothing since it's already translated to other.

I think the only time the word attack shows up in a timeline entry is O4S "Double Attack".  All of the other "attack" translations that show up are just translations (of nothing?) that were automatically generated by the translation scripts.  That's why I think they shouldn't be there.

Maybe you disagree with this approach, but my hope is to get the translations to all be at a point where there is not an ordering that has to be thought about.  Either they can be applied in any order and have the same result, or they will have an error in the "post translation" test in test_timeline.js.  I think it was too subtle to have the order of translations be the order they showed up in the file, and so that's partially why I wrote these tests.

ui/raidboss/data/04-sb/alliance/ridorana_lighthouse.js
quisquous: Done, thanks!

ui/raidboss/data/04-sb/alliance/ridorana_lighthouse.js
quisquous: Ah whoops! Thank you for catching my mistake.

ui/raidboss/common_replacement.js
quisquous: Fixed, and cleaned this up as well.

Jaehyuk-Lee:
okay 👍

PR #1054, opened 2020-02-21
KR v5.0 fix

Jaehyuk-Lee:
I think @quisquous changed signature in `FFXIVProcessKo.cs` to match with v5.2 patch mistakenly.


quisquous:
Ahhh whoops.

Issue #1019, opened 2020-02-05
Better visibility for Korean


Jaehyuk-Lee:
## Summary

Brief explanation of the feature you'd like to request.

I tried making raidboss' timeline better visible in Korean. But... those style codes are too complicated for me to control them myself.
I tried changing `resources\timerbar.js` line 170. But as you see it's too complicated so I gave up.
```javascript
    root.innerHTML = `
      <style>
        .timerbar-root {
          position: relative;
          border: 1px solid black;
          box-sizing: border-box;
        }
        .timerbar-bg {
          position: absolute;
          width: 100%;
          height: 100%;
          opacity: 0.8;
        }
        .timerbar-fg {
          position: absolute;
          width: 100%;
          height: 100%;
          opacity: 1.0;
          will-change: transform;
        }
        .text {
          position: absolute;
          font-family: arial;
          font-weight: bold;
          color: white;
          text-shadow: -1px 0 3px black, 0 1px 3px black, 1px 0 3px black, 0 -1px 3px black;
          will-change: content;
        }
        .text-container {
          position: absolute;
          left: 0px;
          top: calc(50% - 1.2ex);
          width: 100%;
          height: 100%;
          overflow: hidden;
        }
        .timerbar-lefttext {
          position: relative;
          text-align: left;
          padding: 0px 0.4em 0px 0.4em;
        }
        .timerbar-centertext {
          position: relative;
          text-align: center;
          padding: 0px 0.4em 0px 0.4em;
        }
        .timerbar-righttext {
          position: relative;
          text-align: right;
          padding: 0px 0.4em 0px 0.4em;
        }

        :host-context(.skin-lippe) .timerbar-root {
          border: none;
        }

        :host-context(.skin-lippe) .timerbar-bg {
          height: 5px !important;
          border-radius: 1px;
          background-color: #312008 !important;
          border: 1px solid #AA6E03 !important;
          box-shadow: 0 0 8px 0 #AA6E03;
          opacity: 1.0;
          z-index: 0;
        }

        :host-context(.skin-lippe) .timerbar-fg {
          height: 5px !important;
          top: 0px;
          left: 0px;
          background-color: rgba(255, 255, 255, 1) !important;
          box-shadow: 0 0 2px 0 rgba(255, 255, 255, 1) !important;
          text-align: center;
          margin: 1px;
          z-index: 1;
          opacity: 1.0;
        }

        :host-context(.skin-lippe) .text {
          text-shadow:
[...]

PR #1056, opened 2020-02-21
better visibility for Korean

Jaehyuk-Lee:
This solves #1019.


Jaehyuk-Lee:
before
이미지
after
K-005

don't focus on clear font, it's just difference of how I took screenshot.
bottom side of the text moved upper. so it looks middle aligned.

quisquous:
Thanks! Glad you were able to get that working, it looks so much better.

PR #1064, opened 2020-02-23
raidboss: Initial Ruby Weapon Extreme Triggers

panicstevenson:
Initial set of Ruby Weapon (Extreme) triggers.


ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
```diff
@@ -2,10 +2,130 @@
...
+      condition: Conditions.targetIsYou(),
+      infoText: {
+        en: 'Stack on YOU',
+      },
+    },
+    {
+      id: 'RubyEx Optimized Ultima',
+      regex: Regexes.startsUsing({ source: 'The Ruby Weapon', id: '4B2D', capture: false }),
+      alertText: {
+        en: 'Enrage!',
```
quisquous: Maybe timeline only?

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
```diff
@@ -2,10 +2,130 @@
...
+      },
+    },
+    {
+      id: 'RubyEx Pall of Rage',
+      regex: Regexes.gainsEffect({ effect: 'Pall of Rage' }),
+      condition: Conditions.targetIsYou(),
+      infoText: {
+        en: 'Attack Blue',
```
quisquous: Can you say which side this is on?

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
```diff
@@ -2,10 +2,130 @@
...
+    {
+      id: 'RubyEx Stamp',
+      regex: Regexes.startsUsing({ source: 'The Ruby Weapon', id: '4B03' }),
+      condition: function(data) {
+        return data.role == 'tank' || data.role == 'healer';
+      },
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'RubyEx Undermine',
```
quisquous: Sorry for being confused, but can you explain what's going on here? Why do you need the rubyCounter?

panicstevenson: Liquefaction is used when he does the dashes as well, so it's:

Undermine/Liquefaction
Liquefaction + Dashes
Undermine/Liquefaction
Liquefaction + Dashes

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
panicstevenson: We can; we'd have to have logic for saying that they swapped colors though. Thoughts?

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
panicstevenson: Only concern is that enrage can happen mid-rotation -- like if you push the boss to ~5% it will start enraging, even if there's still mechanics left in the timeline

quisquous: No, that's fair.  Can you leave a comment and say that this fight can have a ridiculously early enrage sometimes, so it needs a warning?  I don't want other people seeing this and making enrage triggers for everything.

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
quisquous: OHHHH right.  Thanks!  Can you leave a comment about this?

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
quisquous: I haven't looked at logs.  Would you get a second trigger when they swapped or just the initial?

I think it'd be ideal to have one "Attack Blue (east)" and "Swap to Orange (west)" second calls.  Or whatever side the colors are on, I can't remember.  <_<

ui/raidboss/data/05-shb/trial/ruby-weapon-ex.js
```diff
@@ -2,10 +2,165 @@
...
+          return true;
+      },
+      suppressSeconds: 1,
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'RubyEx Change of Heart',
+      regex: Regexes.ability({ source: 'The Ruby Weapon', id: '4AFC', capture: false }),
+      preRun: function(data) {
+        for (color of data.colors) {
```
quisquous: I'm a little surprised eslint doesn't yell at you for not having let here.

PR #1069, opened 2020-02-23
e5s triggers

Jaehyuk-Lee:
`E5S Stratospear Summons` can be changed to timelineTrigger, but there's no timeline now, so I added there. (Only the first *Stratospear Summons* (4BA5) is needed.)
There's something needed to be fixed. `E5S Stepped Leader` has problem, it only alerts `data.leaderCount %2 == 0` (ready spread). I believe this logic should do both. Strangely, `E5S Stepped Leader Spread` is working well. only when `data.leaderCount %2 == 0`, it says to Spread.
I don't know why this happens..
Other things are all working fine. I've tested so many times with melee DPS. (So, I didn't actually tested `E5S Crippling Blow` but, it should be fine.)

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e5s.js
+++ b/ui/raidboss/data/05-shb/raid/e5s.js
@@ -7,6 +7,206 @@
   },
   timelineFile: 'e5s.txt',
   triggers: [
+    {
+      id: 'E5S Stratospear Summons',
+      regex: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BA5', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BA5', source: '라무', capture: false }),
+      delaySeconds: 5,
+      preRun: function(data) {
+        data.firstSpear = data.firstSpear || 1;
+      },
+      infoText: function(data) {
+        if (data.firstSpear == 1) {
+          return {
+            en: 'Look for small spear',
+            ko: '작은 지팡이 확인',
+          };
+        }
+      },
+      run: function(data) {
+        data.firstSpear++;
+      }
+    },
+    {
+      id: 'E5S Tribunal Summons',
+      regex: Regexes.startsUsing({ id: '4BAC', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BAC', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BAC', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BAC', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BAC', source: '라무', capture: false }),
+      preRun: function(data) {
+        data.adds = data.adds || 0;
+      },
+      infoText: function(data) {
+        if (data.adds != 0) {
+          return {
+            en: 'Look for adds',
+            ko: '쫄 위치 확인',
+          };
+        }
+      },
+      run: function(data) {
+        data.adds++;
+      }
+    },
+    {
+      id: 'E5S Fury\'s Bolt',
+      regex: Regexes.startsUsing({ id: '4BAA', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BAA', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BAA', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BAA', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BAA', source: '라무', capture: false }),
+      infoText: {
+        en: 'Fury\'s Bolt',
+        ko: '라무 강화',
+      },
+    },
+    {
+      id: 'E5S Judgment Volts',
+      regex: Regexes.startsUsing({ id: '4BB5', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BB5', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BB5', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BB5', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BB5', source: '라무', capture: false }),
+      infoText: {
+        en: 'aoe',
+        de: 'AoE',
+        fr: 'Dégâts de zone',
+        cn: 'AOE',
+        ko: '전체 공격',
+      },
+    },
+    {
+      id: 'E5S Stepped Leader',
+      regex: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BC6', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BC6', source: '라무', capture: false }),
+      preRun: function(data) {
+        data.leaderCount = data.leaderCount || 0;
+      },
+      alertText: function(data) {
+        if (data.leaderCount %2 == 0) {
+          return {
+            en: 'Ready Spread',
+            ko: '산개 준비',
+          };
+        } else if (data.leaderCount %2 == 1) {
+          return {
+            en: 'donut AoE',
+            ko: '도넛 장판',
+          };
+        }
+      },
+      run: function(data) {
+        data.leaderCount++;
+      }
+    },
+    {
+      id: 'E5S Stepped Leader Spread',
+      regex: Regexes.ability({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BC6', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BC6', source: '라무', capture: false }),
+      alarmText: function(data) {
+        if (data.leaderCount %2 == 0) {
+          return {
+            en: 'Spread',
+            de: 'Verteilen',
+            fr: 'Ecartez-vous',
+            ja: '散開',
+            cn: '分散',
+            ko: '산개',
+          };
+        }
+      },
+    },
+    {
+      id: 'E5S Crippling Blow',
+      regex: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BCA', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BCA', source: '라무', capture: false }),
+      alertText: function(data, matches) {
+        if (matches.target == data.me) {
+          return {
+            en: 'Tank Buster on YOU',
+            de: 'Tankbuster auf DIR',
+            fr: 'Tank Buster sur VOUS',
+            ko: '탱버 대상자',
+          };
+        }
+      },
+    },
+    {
+      id: 'E5S Stormcloud Summons',
+      regex: Regexes.startsUsing({ id: '4BB8', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BB8', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BB8', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BB8', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BB8', source: '라무', capture: false }),
+      infoText: {
+        en: 'Position for Stormcloud',
+        ko: '번개 구름 위치',
+      },
+    },
+    {
+      id: 'E5S Centaur\'s Charge',
+      regex: Regexes.startsUsing({ id: '4BAD', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BAD', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BAD', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BAD', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BAD', source: '라무', capture: false }),
+      infoText: {
+        en: 'Be in your position',
+        ko: '자기 위치에 있기',
+      },
+    },
+    {
+      id: 'E5S Thunderstorm',
+      regex: Regexes.ability({ id: '4BBF', source: 'Ramuh', capture: false }),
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+    },
+    {
+      id: 'E5S Stepped Leader',
+      regex: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BC6', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BC6', source: '라무', capture: false }),
+      preRun: function(data) {
+        data.leaderCount = data.leaderCount || 0;
```
quisquous: My understanding is that Stepped Leader is different based on whether or not Ramuh has a stack from Fury's Bolt on him or not.  I thought it could be random later in the fight, but maybe not?  Mostly I'm wondering if it makes sense to track the stack on Ramuh as being there or not and then using that to determine what to calll for Stepped Leader.

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+    {
+      id: 'E5S Stratospear Summons',
+      regex: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BA5', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BA5', source: '라무', capture: false }),
+      delaySeconds: 5,
+      preRun: function(data) {
+        data.firstSpear = data.firstSpear || 1;
```
quisquous: If you only need the first stratosphere summons, I think it would be less code and more readable to use a boolean here, rather than also needing a conditional in infoText, e.g.
```
condition: function(data) { return data.seenFirstSpear; },
run: function(data) { data.seenFirstSpear = true; }
```

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+    },
+    {
+      id: 'E5S Fury\'s Bolt',
+      regex: Regexes.startsUsing({ id: '4BAA', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BAA', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BAA', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BAA', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BAA', source: '라무', capture: false }),
+      infoText: {
+        en: 'Fury\'s Bolt',
```
quisquous: Maybe this should say "pick up orb"?

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+      },
+    },
+    {
+      id: 'E5S Judgment Volts',
+      regex: Regexes.startsUsing({ id: '4BB5', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BB5', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BB5', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BB5', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BB5', source: '라무', capture: false }),
+      infoText: {
```
quisquous: `response: Responses.aoe()`

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+      },
+    },
+    {
+      id: 'E5S Levinforce',
+      regex: Regexes.startsUsing({ id: '4BCC', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '4BCC', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '4BCC', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '4BCC', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '4BCC', source: '라무', capture: false }),
+      infoText: {
```
quisquous: `response: Responses.knockback()`

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+      },
+    },
+    {
+      id: 'E5S Crippling Blow',
+      regex: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BCA', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BCA', source: '라무', capture: false }),
+      alertText: function(data, matches) {
```
quisquous: `response: Responses.tankBuster()`

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+        }
+      },
+    },
+    {
+      id: 'E5S Crippling Blow',
+      regex: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BCA', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BCA', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BCA', source: '라무', capture: false }),
```
quisquous: I think this should not be capture false if you are using matches below.  I'm surprised `npm run test` didn't complain.  (cc @panicstevenson)

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,206 @@
...
+    },
+    {
+      id: 'E5S Thunderstorm',
+      regex: Regexes.ability({ id: '4BBF', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BBF', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BBF', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BBF', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BBF', source: '라무', capture: false }),
+      infoText: {
+        en: 'Avoid AoE',
```
quisquous: I know you said this is just for melee, but I think this will be confusing to people who have to be hit by the thunderstorm.  Could you either adjust this to handle that case or remove it for now?

Jaehyuk-Lee:
I don't understand this error
`cactbot/ui/raidboss/data/05-shb/raid/e5s.js: Found 'matches' as a function parameter without regex capturing group for trigger id 'E5S Crippling Blow'.`
I just copy&paste this and changed to match e5s.
https://github.com/quisquous/cactbot/blob/ce0e9065e6710249a0706a82f9547a89c5040dec/ui/raidboss/data/03-hw/alliance/dun_scaith.js#L580-L603

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: ah, yes, you can determine whether Ramuh's Fury something or not. It is determinable by counts but I think your opinion is better. It is not random as far as I played.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: I wrote "Ramuh strengthened" in Korean, since one thing there donut AoE does not need orbs. So... I was wondering if which is better.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: Ah, that was the problem, I wrote `infoText: Responses.aoe()` before, it didn't worked. lol. I thought response was not working.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: Hmmmm... how do I change this to that?
```javascript
        if (matches.target == data.me) {
          return {
            en: 'Tank Buster on YOU',
            de: 'Tankbuster auf DIR',
            fr: 'Tank Buster sur VOUS',
            ko: '탱버 대상자',
          };
        }
      },
```
to
```javascript
        if (matches.target == data.me) {
          return Responses.tankBuster();
        }
      },
```
does this work?

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: a ha, capture is that thing. I didn't know.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: This is not for melee only. Everyone should avoid aoes.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: That complained.

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Ok, if it's not random, then doing it by count sounds fine!  I just wasn't sure.

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Ah that's true! This is good for now.  Maybe in the future there can be some smarter "pick up orb" trigger too.

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Sorry for not documenting this better.  `response` returns an object that can have all the same fields a trigger does.  So it's like one function that can return everything.  There's an example in the `00-misc/test.js` file.  #1064 also has some examples of using it.

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: I was just suggesting `response: Responses.tankBuster()` on its own, with no alertText.  Maybe add a condition that it is on you or that you're a healer, like https://github.com/quisquous/cactbot/pull/1064/files#diff-7cf55164a059021c14004927f7f84f4eR20

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Sorry for the confusion, I was mistaking this with the stormcloud abillity.

I think I'd say my feeling is that I think this trigger is not necessary, as it's not worth calling out obvious ground aoes.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: This is fine if this works... 😅 This is not working properly.

> There's something needed to be fixed. E5S Stepped Leader has problem, it only alerts data.leaderCount %2 == 0 (ready spread). I believe this logic should do both. Strangely, E5S Stepped Leader Spread is working well. only when data.leaderCount %2 == 0, it says to Spread.
I don't know why this happens..
I don't know why this is not working...

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: I was like so concentrated on dps-ing. forgot to avoid this, did jump and got hit by it lol.
After I become familiar with this, I was alright but as I experienced that, I put this trigger.
If you think this is not really needed, say it, I'll remove this.

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: Do you know how to add condition, Ramuh hasEffects id: '8B7'(Fury's Bolt)?
I can't find any examples.

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Yeah, my general rule of thumb is that visible ground aoes shouldn't be called out.  Things that are invisible or very fast (e.g. the cyclops in the new dungeon) are somewhat different.

This just sounds like drg problems.  :wink:

Jaehyuk-Lee:
I think this https://github.com/quisquous/cactbot/pull/1069/commits/5de0cb44fa566e0a3dedf3011e23465393f53dc7 (fix spread with ramuh fury) will fix problem I've said on the top.
I've looked up e2s spell-in-waiting.

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,181 @@
...
+    },
+    {
+      id: 'E5S Stratospear Summons',
+      regex: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexDe: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexFr: Regexes.ability({ id: '4BA5', source: 'Ramuh', capture: false }),
+      regexJa: Regexes.ability({ id: '4BA5', source: 'ラムウ', capture: false }),
+      regexKo: Regexes.ability({ id: '4BA5', source: '라무', capture: false }),
+      delaySeconds: 5,
+      condition: function(data) {
```
JLGarber: I may be misreading, but is this ever going to fire? I couldn't find another reference to `seenFirstSpear` outside this trigger. If I'm remembering the order of operations correctly, the condition won't fire, and without that the run() command to set the value won't either. I think it might instead need to be

`return !data.seenFirstSpear`

Jaehyuk-Lee: yes, your're right. if you haven't seen this, trigger should do some work. (when `seenFirstSpear` returns false

Jaehyuk-Lee:
Hmmm... does this responses needs to remove `capture: false`?
Maybe this is problem of difference between my branch and master branch here.
`This branch is 14 commits ahead, 7 commits behind quisquous:master.` this branch is forked from 0.16.2 release

Jaehyuk-Lee:
I'll just go with PR's check.

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -7,6 +7,194 @@
   },
   timelineFile: 'e5s.txt',
   triggers: [
+    {
+      id: 'E5S Fury\'s Bolt Gain',
+      regex: Regexes.gainsEffect({ target: 'Ramuh', effect: 'Fury\'s Bolt', capture: false }),
+      regexDe: Regexes.gainsEffect({ target: 'Ramuh', effect: 'Fury\'s Bolt', capture: false }),
```
quisquous: Sorry for one last comment, but can you remove translations for regexes that are not correct? I think it makes it much harder for translators to know if there's a missing translation or not.

Jaehyuk-Lee: I made mistake there. Ramuh was the same for en, de, fr. I don't know how exactly I was thinking, but that was the problem.

Jaehyuk-Lee:
in my head, I think it's logically fine, but it wasn't working. I don't know why....

quisquous:
> Hmmm... does this responses needs to remove `capture: false`?
> my branch shows an error with `capture: false` written.
> but this PR shows an error with `capture: false` not written.

Yeah, Responses.tankBuster() uses matches, so you need `capture: true` (which is there by default if you don't have `capture: false`).  This is correct as you have it written.

valarnin:
Testing these now, and the buster warning for crippling blow goes out after the hit on the tank.

Jaehyuk-Lee:
ah, wtf! it should be 'startsusing' not 'ability' sorry

Jaehyuk-Lee:
quisquous maybe fix this... I really have to sleep now. sorry and thanks :sweat:
It's 6AM here and I didn't sleep...

Jaehyuk-Lee:
I'll fix tomorrow if it's not fixed

valarnin:
Spread for Leader is also late, needs to be about 1.5s earlier roughly.

PR #1039, opened 2020-02-11
sync add - akadaemia_anyder

Jaehyuk-Lee:
When dps is good, "Arbor Storm" is skipped.
I wondered why timeline sync was not matching, this was the problem.
I don't know how exactly timeline works, so there can be better fix of this timeline.

PR #1071, opened 2020-02-23
Anyder timeline sync

Jaehyuk-Lee:
I tested this and worked fine.


Jaehyuk-Lee:
changed as KattTails said in #1039

Issue #1068, opened 2020-02-23
E6S.txt timeline remove leftover copy and paste code


BeIeriand:
## Description

leftover copy and paste code

https://github.com/quisquous/cactbot/blob/66394dc421d790c124237c9a2a8f3d9522491126/ui/raidboss/data/05-shb/raid/e6s.txt#L128-L130

quisquous:
@Legends0 Do you know what this should be?

BeIeriand:
They are copy's from line 106,122 and 123 and probably were forgotten there so they can be removed

Legends0:
Fixed in PR #1073. The lines were a leftover when adjusting the times as FFLogs had separate fights.

PR #1073, opened 2020-02-23
Fix for #1068 and Resync

Legends0:
- #1068, there were lines that were leftover from merging two timelines
- Resync timeline to https://www.fflogs.com/reports/RkPv8yn1WJK9TX4V/#fight=2.
- Commented a few additional syncs

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6s.txt
+++ b/ui/raidboss/data/05-shb/raid/e6s.txt
@@ -9,126 +9,123 @@ hideall "--sync--"
 ### Phase 1: Garuda
 0 "Start"
 0.0 "--sync--" sync /:Engage!/ window 0,1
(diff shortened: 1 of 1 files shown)


Legends0:
At some point in the future with DPS increases, we may find that windows may need to be added between these phases.

quisquous:
> At some point in the future with DPS increases, we may find that windows may need to be added between these phases.

Yeah, it makes sense.  For what it's worth, I usually add like 2-3 "big" syncs for one-time phase changing abilities into timelines, both for future proofing and also in case ACT crashes and you start it mid-fight (and it will catch up).  For example, https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/05-shb/raid/e7s.txt#L88

quisquous:
Thanks for the followup.  <3

PR #1077, opened 2020-02-24
Fix e5s crippling blow and stepped leader timing

anubiann00b:
Fix timing of e5s triggers based on the notes in #1069 

I haven't validated that these are correct yet but I will tonight when I run e5s.

@Jaehyuk-Lee pinging you in since this is an addition to your work on e5s triggers (thank you!)

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e5s.js
+++ b/ui/raidboss/data/05-shb/raid/e5s.js
@@ -125,7 +125,7 @@
       regexFr: Regexes.startsUsing({ id: '4BC6', source: 'Ramuh', capture: false }),
       regexJa: Regexes.startsUsing({ id: '4BC6', source: 'ラムウ', capture: false }),
       regexKo: Regexes.startsUsing({ id: '4BC6', source: '라무', capture: false }),
-      delaySeconds: 2.0,
+      delaySeconds: 3.0,
       condition: function(data) {
         return !data.fury;
       },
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -140,11 +140,11 @@
     },
     {
       id: 'E5S Crippling Blow',
-      regex: Regexes.ability({ id: '4BCA', source: 'Ramuh' }),
-      regexDe: Regexes.ability({ id: '4BCA', source: 'Ramuh' }),
-      regexFr: Regexes.ability({ id: '4BCA', source: 'Ramuh' }),
-      regexJa: Regexes.ability({ id: '4BCA', source: 'ラムウ' }),
-      regexKo: Regexes.ability({ id: '4BCA', source: '라무' }),
+      regex: Regexes.startsUsing({ id: '4BCA', source: 'Ramuh', capture: true }),
```
quisquous: can you remove capture: true? it's implicit and gets skipped everywhere else.

quisquous:
Thanks for the fix, I've had a really busy day and wasn't able to get back to the other pull request.

ui/raidboss/data/05-shb/raid/e5s.js
anubiann00b: fixed

anubiann00b:
actually, could we wait a sec before merging--I think that timing is wrong

quisquous:
Good now?

anubiann00b:
yep! I'mnot 100% sure abt 3.0 seconds because prog was a blur but if it's not right it's definitely closer

quisquous:
> yep! I'mnot 100% sure abt 3.0 seconds because prog was a blur but if it's not right it's definitely closer

I wonder if it would be better to just use the ability rather than a delay from the casting start, but I can try that in the future.

Issue #1037, opened 2020-02-11
Replace zoneRegex with zoneID


quisquous:
OverlayPlugin has a ChangeZone event that returns the ID, but no zone name string.  (Alternatively it could listen to the ZoneChangeDelegate from the ACT plugin instead of parsing the network log, which gives both the id and the zone name string).

It would be nice to hook cactbot up to something that returns both the zoneid and the regex and slowly switch over raidboss and oopsyraidsy to use zone ids.  I think it would be good to support both zoneID and zoneRegex until everything is swapped over, and then zoneRegex can be removed.

This would help remove translation needs for all zone names (especially if Ravahn ever decides to change whether zone names are translated) and will help with the start of a patch when the zone names are sometimes unknown.

quisquous:
Copying from #1528:

- [ ] land this
- [ ] get ngld/OverlayPlugin#138 merged
- [ ] wait for OverlayPlugin release
- [ ] switch to using the new event type
- [ ] add code in raidboss/oopsy to support zoneId
- [ ] add documentation
- [ ] switch raidboss/oopsy over to using zoneId
- [ ] switch various other bits of code using zone names (e.g. lang)
- [ ] remove use of zoneRegex from cactbot files (but don't remove support)

PR #1081, opened 2020-02-24
Add eden's verse to the default WellFedZones

anubiann00b:


quisquous:
You need to add these zone names to the lang files too.  Ideally we can switch everything to zone ids at some point.

quisquous:
Hmm.  It sure looks like this isn't translated anywhere.  I guess I didn't consider this when it turned out that ko/cn zone names were localized.  Ideally this could also be turned into a zone id in #1037.

Issue #1062, opened 2020-02-23
util/test_timeline.py throws errors 


quisquous:
```
$ python util/test_timeline.py -f e7s.log -lf 8                                 Traceback (most recent call last):
  File "util/test_timeline.py", line 552, in <module>
    main(args)
  File "util/test_timeline.py", line 438, in main
    timelist = load_timeline(args.timeline)
  File "util/test_timeline.py", line 14, in load_timeline
    with timeline as file:
AttributeError: __enter__
```

JLGarber:
Hmm, that should be throwing a "no timeline specified" error. Looking at the file as it stands though, there's no check for that during the argument parsing. Interesting. Probably add a check like...

```
    if args.search_fights and args.search_fights > -1 and not args.timeline:
        raise parser.error("You must specify a timeline file before testing against a specific encounter.")
```

The text could be cleaned up and it's probably better not to check relative values on the `search-fights` argument, but I think this should eliminate the error encountered here. (My feeling is that a timeline file should not be required unless an encounter is specified, as the user would otherwise not be able to list encounters without specifying a timeline.)

PR #1083, opened 2020-02-24
Util: Added missing parser error to test_timeline

JLGarber:
Per #1062 this should resolve the error if the user attempts to test against a fight without a timeline selected. I'm not 100% sure it's correct to order the errors like this, but I think it's clearer this way.


quisquous:
I think this seems very reasonable!

PR #1092, opened 2020-02-24
raidboss: Initial E6S Triggers

panicstevenson:
Adding some triggers; probably not overly helpful without video
reference to decipher things like Ifrit locations or Ferostorm
positioning.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6s.js
+++ b/ui/raidboss/data/05-shb/raid/e6s.js
@@ -36,18 +36,13 @@
         return data.handsOfFlame = true;
       },
       infoText: function(data, matches) {
-        if (data.me != matches.target || data.role != 'tank' || data.onGaruda)
-          return;
         if (data.me == matches.target) {
-          if (data.onIfrit == true) {
-            return {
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,157 @@
...
+        return {
+          en: 'Tethered to ' + matches.source,
+        };
+      },
+    },
+    {
+      id: 'E6S Conflag Strike Behind',
+      regex: Regexes.startsUsing({ source: 'Raktapaksa', id: '4C10', capture: false }),
+      delaySeconds: 31,
+      response: Responses.getBehind(),
```
quisquous: Maybe this was just what I've seen in practice, but I half wonder if this should be `getBehind('alarm')` just to be like really for real do this thing that's easy to forget.  Maybe that's silly.

quisquous: Eh, alert is probably ok for now.  I'll consider upping it later.

PR #1093, opened 2020-02-25
responses: add more options for multiple severities

quisquous:


quisquous:
I'm not super happy with this "unusedText" hack but I wanted to make it possible to hide "tankbuster on" for healers in alliance raids sometimes, since sometimes it can be very noisy.

Do folks have opinions on this? @panicstevenson @Akurosia @Jaehyuk-Lee

quisquous:
Maybe the cleanest solution here is to just add tankbusterTargetOnly() instead of hacky 'none'.

quisquous:
I'm just going to remove 'none' here.  I think I forgot you could just add a conditional if you only wanted it on the target.

PR #1094, opened 2020-02-25
raidboss: Initial E8S Triggers

panicstevenson:
An initial set of triggers; not complete, and not enough information to
go off of without video reference.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e8s.js
+++ b/ui/raidboss/data/05-shb/raid/e8s.js
@@ -6,7 +6,193 @@
     ko: /^희망의 낙원 에덴: 공명편\(영웅\) \(4\)$/,
   },
   timelineFile: 'e8s.txt',
+  timelineTriggers: [
+    {
+      id: 'E8S Shining Armor',
+      regex: /Shining Armor/,
+      beforeSeconds: 2,
+      response: Responses.lookAway(),
+    },
+  ],
   triggers: [
+    {
+      id: 'E8S Absolute Zero',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -6,7 +6,193 @@
...
+      },
+    },
+    {
+      id: 'E8S Wyrmclaw',
+      regex: Regexes.gainsEffect({ effect: 'Wyrmclaw' }),
+      condition: Conditions.targetIsYou(),
+      preRun: function(data, matches) {
+        if (data.wyrmsLament == 1) {
+          data.wyrmclawNumber = {
+            '14.00': 1,
```
quisquous: This is not quite good enough.  Can you make this a conditional for <=14, <=22, <=30, else to assign? Sometimes effect timers are slightly off because the ffxiv plugin returns you the time remaining and it may be delayed, so you might be 13.98s.  This is why Nael doom timers do this as well.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -6,7 +6,193 @@
...
+    },
+    {
+      id: 'E8S Icelit Dragonsong Cleanse',
+      regex: Regexes.ability({ source: 'Shiva', id: '4D70', capture: false }),
+      suppressSeconds: 1,
+      condition: function(data) {
+        return data.CanCleanse();
+      },
+      infoText: {
+        en: 'Cleanse',
```
quisquous: This should say a `Cleanse DPS only`

quisquous:
Thank youuuu

PR #1096, opened 2020-02-25
FR RegEx for E7S

1unie:
Work in progress. I think there's an error in the XIV files when it comes to adds name in French, which I expect to be fixed in a future patch. Blashemy and Idolatry currently have the same name in-game: https://i.imgur.com/fnkXoNP.jpg


1unie:
This is complete.

PR #1097, opened 2020-02-25
e5s trigger fixes

Jaehyuk-Lee:
1. show "enrage soon" not "grab an orb" on last cast of Fury's Fourteen.
2. undefined returns false, so I fixed it to work fine.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e5s.js
+++ b/ui/raidboss/data/05-shb/raid/e5s.js
@@ -86,10 +86,23 @@
       regexFr: Regexes.startsUsing({ id: '4BAB', source: 'Ramuh', capture: false }),
       regexJa: Regexes.startsUsing({ id: '4BAB', source: 'ラムウ', capture: false }),
       regexKo: Regexes.startsUsing({ id: '4BAB', source: '라무', capture: false }),
-      infoText: {
-        en: 'Grab an orb',
-        fr: 'Prenez un orb',
-        ko: '구슬 줍기',
+      infoText: function(data) {
+        if (data.fourteenCount < 3) {
+          return {
+            en: 'Grab an orb',
+            fr: 'Prenez un orb',
+            ko: '구슬 줍기',
+          };
+        } else if (data.fourteenCount == 3) {
+          return {
+            en: 'Enrage Soon',
+            ko: '곧 전멸기',
+          };
+        }
+      },
+      run: function(data) {
+        data.fourteenCount = data.fourteenCount || 0;
+        data.fourteenCount++;
       },
     },
     {
--- a/ui/raidboss/data/05-shb/raid/e5s.js
+++ b/ui/raidboss/data/05-shb/raid/e5s.js
@@ -85,6 +85,10 @@
       regexFr: Regexes.startsUsing({ id: '4BAB', source: 'Ramuh', capture: false }),
       regexJa: Regexes.startsUsing({ id: '4BAB', source: 'ラムウ', capture: false }),
       regexKo: Regexes.startsUsing({ id: '4BAB', source: '라무', capture: false }),
+      preRun: function(data) {
+        data.fourteenCount = data.fourteenCount || 0;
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -55,15 +55,15 @@
       regexJa: Regexes.startsUsing({ id: '4BAC', source: 'ラムウ', capture: false }),
       regexKo: Regexes.startsUsing({ id: '4BAC', source: '라무', capture: false }),
       condition: function(data) {
-        return !data.firstAdd;
```
quisquous: What's wrong with this? I think the code as written should be fine, as !undefined is true.

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -86,10 +86,23 @@
...
+      infoText: function(data) {
+        if (data.fourteenCount < 3) {
+          return {
+            en: 'Grab an orb',
+            fr: 'Prenez un orb',
+            ko: '구슬 줍기',
+          };
+        } else if (data.fourteenCount == 3) {
+          return {
+            en: 'Enrage Soon',
```
quisquous: Could you drop the enrage text? At least for consistency, nothing else shows the enrage except for Ruby ex which has weird timing where the enrage can suddenly appear.  If the enrage isn't here, then this could just be a condition on fourteenCount.

Jaehyuk-Lee: I tried to say, there will be an enrage so just don't grab orbs and focus on dps.
should I say 'don't grab orbs'?

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: ummmm, that's right. But why did this mis-functioned... It showed message on first add...

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Ah, I see.  Yeah, I'm not sure how to communicate all of that.  "Don't get orbs" seems like a "don't get orbs, or you'll die" kind of warning.  Maybe this is just an optimization and no need for a trigger.

I think usually I use the timeline as the "enrage" marker and not triggers, for what it's worth.

ui/raidboss/data/05-shb/raid/e5s.js
quisquous: Oh! You *don't* want it to run the first time? I would remove the condition then, and just have a 
```
infoText: function(data) {
  if (data.seenFirstAdd) {
    return { look for adds etc };
  }
  data.seenFirstAdd = true;
}
```

ui/raidboss/data/05-shb/raid/e5s.js
Jaehyuk-Lee: I just removed message on last Fury's fourteen.

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -86,6 +85,13 @@
       regexFr: Regexes.startsUsing({ id: '4BAB', source: 'Ramuh', capture: false }),
       regexJa: Regexes.startsUsing({ id: '4BAB', source: 'ラムウ', capture: false }),
       regexKo: Regexes.startsUsing({ id: '4BAB', source: '라무', capture: false }),
+      preRun: function(data) {
+        data.fourteenCount = data.fourteenCount || 0;
+        data.fourteenCount++;
+      },
+      condition: function(data) {
```
quisquous: Confusingly, I think preRun only happens if condition is met, and runs after the condition.  So maybe this needs to be: `return !data.fourteenCount || data.fourteenCount < 2`.  This makes me wonder if I should write a test that all the fields have to be in the order they run, for clarity.

quisquous:
Thanks!

PR #1098, opened 2020-02-25
Response for ShB

Akurosia:


Akurosia:
sometimes i hate git... will resubmit once sb/shb are split

PR #1103, opened 2020-02-26
Util: Make_timeline fix for when ACT wasn't open at start of instance

### LINKED ISSUE #1061: util/make_timeline.py -lf isn't finding all fights
`-lf` finds no e8s fights in this log file:
[Network_2052_20200222.zip](https://github.com/quisquous/cactbot/files/4240429/Network_2052_20200222.zip)

`-lf` finds only e8s and not e7s in this log file:
[Network_2051_20200222.zip](https://github.com/quisquous/cactbot/files/4240430/Network_2051_20200222.zip)

(CC @JLGarber)

JLGarber:
This should resolve #1061. We may have to add more `21` lines to `is_instance_restarted`, but from the More Than One logs I tested this on, it seems to be happy regardless.

### DIFF
--- a/util/encounter_tools.py
+++ b/util/encounter_tools.py
@@ -67,6 +67,10 @@ def is_instance_begun(line_fields):
     return line_fields[4].endswith("has begun.")
 
 
+def is_instance_updated(line_fields):
+    return line_fields[0] == "33" and line_fields[3] in ["40000001", "40000006"]
+
+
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for this! I've been using a bunch of user submitted logs so having this be more robust will be super helpful.  I can't live without -lf these days!

PR #1104, opened 2020-02-26
raidboss: Minor E8S Additions

panicstevenson:
Some small additions for anyone progging.


ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -72,6 +72,23 @@
...
+      infoText: {
+        en: 'Chain on YOU',
+      },
+    },
+    {
+      id: 'E8S Holy Light',
+      regex: Regexes.tether({ id: '0002' }),
+      condition: Conditions.targetIsYou(),
+      infoText: {
+        en: 'Orb on YOU',
```
quisquous: We could technically also call this out immediately after the chains go out.

PR #1105, opened 2020-02-26
added doritoStack and include source into sleep/stun/interupt

Akurosia:


resources/responses.js
```diff
@@ -584,36 +597,43 @@ let Responses = {
...
-      en: 'Sleep',
-      de: 'Schlaf',
-      fr: 'Sommeil',
-      ja: 'スリプル',
-      cn: '催眠',
-      ko: '슬리플',
+    obj[defaultAlertText(sev)] = (data, matches) => {
+      let source = getSource(matches);
+      return {
+        en: 'Sleep => ' + source,
```
quisquous: At least for English, I think all of these would be better without `=>`

Akurosia: yeah it was just to keep the old text (as i am also not sure how other language really handels it)

quisquous:
Hey, look at those tests catching things they should!

PR #1106, opened 2020-02-26
ocean fishing fix

### LINKED ISSUE #1075: Fishing Addon - Ocean Fishing
Now that the ACT plugin has been updated, is there any plans to extend the fishing cast timer to ocean fishing?

Currently doesn't seem to detecting ocean fish, but I assume it needs data added for location?

pikalala:
The fishing spot/place during ocean fishing changes while fishing. Set it on every cast regardless of whether it has already been set to handle this.

This should also fix #1075 as the ocean fish data has already been added.

Only tested in one voyage but it worked, and regular fishing looks like it still works.

### DIFF
--- a/ui/fisher/fisher.js
+++ b/ui/fisher/fisher.js
@@ -230,15 +230,13 @@ class Fisher {
       this.ui.startFishing();
       return;
     }
-    // Set place, if it's unset
-    if (!this.place || !this.place.id) {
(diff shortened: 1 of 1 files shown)


AdriaanDeVos:
Works on my end 👍

quisquous:
Thanks! I didn't test this myself, but the code looks good to my eyes.

PR #1114, opened 2020-02-26
Raidboss player name override features

valarnin:
Allow overriding player name and playing TTS through browser when using raidboss overlay with websocket. Example query string:
`&player=Some%20Person`

If that person is in your party, raidboss will display the same timeline as normal, but give callouts for that person and their job instead of the person hosting the ACT/Overlay/WS/Cactbot instance.

Text to Speech is done via google services as that was the simplest to implement. Audio should only be fetched once per TTS string per page load.

Initial tests for this functionality are good but haven't had a chance to do extensive testing yet.

### DIFF
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -42,7 +42,18 @@ class GoogleTTSItem extends TTSItem {
 
 class BrowserTTSEngine {
   constructor() {
-    this.engineType = null;
+    // figure out what TTS engine type we need
+    if (window.speechSynthesis !== undefined) {
+      let me = this;
+      window.speechSynthesis.onvoiceschanged = function() {
+        if (window.speechSynthesis.getVoices().length > 0) {
+          window.speechSynthesis.onvoiceschanged = null;
+          me.engineType = TTSEngineType.SpeechSynthesis;
+        }
+      };
+    }
+
+    this.engineType = TTSEngineType.GoogleTTS;
     this.ttsItems = {};
   }
 
@@ -51,27 +62,21 @@ class BrowserTTSEngine {
       if (this.ttsItems[text] !== undefined) {
         this.ttsItems[text].play();
       } else {
-        switch (this.engineType) {
-        case null:
-          // figure out what TTS engine type we need
-          if (window.speechSynthesis !== undefined &&
-            window.speechSynthesis.getVoices().length > 0)
-            this.engineType = TTSEngineType.SpeechSynthesis;
-          else
-            this.engineType = TTSEngineType.GoogleTTS;
-          // just recurse to avoid repeating code
-          this.play(text);
-          break;
-        case TTSEngineType.SpeechSynthesis:
-          this.playSpeechTTS(text);
-          break;
-        case TTSEngineType.GoogleTTS:
-          this.playGoogleTTS(text);
-          break;
-        }
+        this.playTTS(text);
       }
     } catch (e) {
       console.log('Exception performing TTS', e);
+    }
+  }
+
+  playTTS(text) {
+    switch (this.engineType) {
+    case TTSEngineType.SpeechSynthesis:
+      this.playSpeechTTS(text);
+      break;
+    case TTSEngineType.GoogleTTS:
+      this.playGoogleTTS(text);
+      break;
     }
   }
 
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -42,8 +42,8 @@ class GoogleTTSItem extends TTSItem {
 
 class BrowserTTSEngine {
   constructor() {
-    engineType = null;
-    ttsItems = {};
+    this.engineType = null;
+    this.ttsItems = {};
   }
 
   play(text) {
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -59,11 +59,10 @@ class BrowserTTSEngine {
 
   play(text) {
     try {
-      if (this.ttsItems[text] !== undefined) {
+      if (this.ttsItems[text] !== undefined)
         this.ttsItems[text].play();
-      } else {
+      else
         this.playTTS(text);
-      }
     } catch (e) {
       console.log('Exception performing TTS', e);
     }
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -44,11 +44,10 @@ class BrowserTTSEngine {
   constructor() {
     // figure out what TTS engine type we need
     if (window.speechSynthesis !== undefined) {
-      let me = this;
-      window.speechSynthesis.onvoiceschanged = function() {
+      window.speechSynthesis.onvoiceschanged = () => {
         if (window.speechSynthesis.getVoices().length > 0) {
           window.speechSynthesis.onvoiceschanged = null;
-          me.engineType = TTSEngineType.SpeechSynthesis;
+          this.engineType = TTSEngineType.SpeechSynthesis;
         }
       };
     }
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -4,20 +4,20 @@ class BrowserTTSEngine {
   static play(text) {
     try {
       // check to see if we've already gotten this tts before
-      let iframe = window.document.querySelector("[data-tts=\""+text+"\"]");
+      let iframe = window.document.querySelector('[data-tts="'+text+'"]');
       if (iframe === null) {
-        iframe = document.createElement("iframe");
+        iframe = document.createElement('iframe');
         // remove sandbox so we can modify contents/call play on audio element later
         iframe.removeAttribute('sandbox');
-        iframe.style.display = "none";
+        iframe.style.display = 'none';
         document.body.appendChild(iframe);
         let encText = encodeURIComponent(text);
-        iframe.contentDocument.body.innerHTML = "<audio src=\"https://translate.google.com/translate_tts?ie=UTF-8&client=tw-ob&tl=en&q=" + encText + "\" id=\"TTS\">";
-        iframe.dataset["tts"] = text;
+        iframe.contentDocument.body.innerHTML = '<audio src="https://translate.google.com/translate_tts?ie=UTF-8&client=tw-ob&tl=en&q=' + encText + '" id="TTS">';
+        iframe.dataset['tts'] = text;
       }
       iframe.contentDocument.body.firstElementChild.play();
-    } catch(e) {
-        console.log("Exception performing TTS", e);
+    } catch (e) {
+      console.log('Exception performing TTS', e);
     }
   }
 }
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -1,23 +1,87 @@
 'use strict';
 
+const TTSEngineType = {
+  SpeechSynthesis: 0,
+  GoogleTTS: 1,
+};
+
+class TTSItem {
+  play() {}
+}
+
+class SpeechTTSItem extends TTSItem {
+  constructor(text) {
+    super();
+    this.text = text;
+    this.item = new SpeechSynthesisUtterance(text);
+  }
+
+  play() {
+    window.speechSynthesis.speak(this.item);
+  }
+}
+
+class GoogleTTSItem extends TTSItem {
+  constructor(text) {
+    super();
+    this.text = text;
+    let iframe = document.createElement('iframe');
+    // remove sandbox so we can modify contents/call play on audio element later
+    iframe.removeAttribute('sandbox');
+    iframe.style.display = 'none';
+    document.body.appendChild(iframe);
+    let encText = encodeURIComponent(text);
+    iframe.contentDocument.body.innerHTML = '<audio src="https://translate.google.com/translate_tts?ie=UTF-8&client=tw-ob&tl=en&q=' + encText + '" id="TTS">';
+    this.item = iframe.contentDocument.body.firstElementChild;
+  }
+
+  play() {
+    this.item.play();
+  }
+}
+
 class BrowserTTSEngine {
-  static play(text) {
+  constructor() {
+    engineType = null;
+    ttsItems = {};
+  }
+
+  play(text) {
     try {
-      // check to see if we've already gotten this tts before
-      let iframe = window.document.querySelector('[data-tts="'+text+'"]');
-      if (iframe === null) {
-        iframe = document.createElement('iframe');
-        // remove sandbox so we can modify contents/call play on audio element later
-        iframe.removeAttribute('sandbox');
-        iframe.style.display = 'none';
-        document.body.appendChild(iframe);
-        let encText = encodeURIComponent(text);
-        iframe.contentDocument.body.innerHTML = '<audio src="https://translate.google.com/translate_tts?ie=UTF-8&client=tw-ob&tl=en&q=' + encText + '" id="TTS">';
-        iframe.dataset['tts'] = text;
+      if (this.ttsItems[text] !== undefined) {
+        this.ttsItems[text].play();
+      } else {
+        switch (this.engineType) {
+        case null:
+          // figure out what TTS engine type we need
+          if (window.speechSynthesis !== undefined &&
+            window.speechSynthesis.getVoices().length > 0)
+            this.engineType = TTSEngineType.SpeechSynthesis;
+          else
+            this.engineType = TTSEngineType.GoogleTTS;
+          // just recurse to avoid repeating code
+          this.play(text);
+          break;
+        case TTSEngineType.SpeechSynthesis:
+          this.playSpeechTTS(text);
+          break;
+        case TTSEngineType.GoogleTTS:
+          this.playGoogleTTS(text);
+          break;
+        }
       }
-      iframe.contentDocument.body.firstElementChild.play();
     } catch (e) {
       console.log('Exception performing TTS', e);
     }
+  }
+
+  playSpeechTTS(text) {
+    this.ttsItems[text] = new SpeechTTSItem(text);
+    this.ttsItems[text].play();
+  }
+
+  playGoogleTTS(text) {
+    this.ttsItems[text] = new GoogleTTSItem(text);
+    this.ttsItems[text].play();
   }
 }
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -1,0 +1,23 @@
+'use strict';
+
+class BrowserTTSEngine {
+  static play(text) {
+    try {
+      // check to see if we've already gotten this tts before
+      let iframe = window.document.querySelector("[data-tts=\""+text+"\"]");
+      if (iframe === null) {
+        iframe = document.createElement("iframe");
+        // remove sandbox so we can modify contents/call play on audio element later
+        iframe.removeAttribute('sandbox');
+        iframe.style.display = "none";
+        document.body.appendChild(iframe);
+        let encText = encodeURIComponent(text);
+        iframe.contentDocument.body.innerHTML = "<audio src=\"https://translate.google.com/translate_tts?ie=UTF-8&client=tw-ob&tl=en&q=" + encText + "\" id=\"TTS\">";
+        iframe.dataset["tts"] = text;
+      }
+      iframe.contentDocument.body.firstElementChild.play();
+    } catch(e) {
+        console.log("Exception performing TTS", e);
+    }
+  }
+}
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -632,8 +632,13 @@ class PopupText {
           ko: ' 그리고 ',
         };
         ttsText = ttsText.replace(/\s*(<[-=]|[=-]>)\s*/, arrowReplacement[lang]);
-        let cmd = { 'call': 'cactbotSay', 'text': ttsText };
-        window.callOverlayHandler(cmd);
+        // if we're overriding player name and browser tts engine is loaded, use that
+        if(Options.PlayerNameOverride !== null && typeof BrowserTTSEngine === "function") {
+          BrowserTTSEngine.play(ttsText);
+        } else {
+          let cmd = { 'call': 'cactbotSay', 'text': ttsText };
+          window.callOverlayHandler(cmd);
+        }
       } else if (soundUrl && playSounds) {
         let audio = new Audio(soundUrl);
         audio.volume = soundVol;
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -31,6 +31,19 @@ class PopupText {
     this.alertText = document.getElementById('popup-text-alert');
     this.alarmText = document.getElementById('popup-text-alarm');
 
+    if (Options.PlayerNameOverride !== null &&
+      typeof BrowserTTSEngine === 'function') {
+      this.ttsEngine = new BrowserTTSEngine();
+      this.ttsSay = function(text) {
+        this.ttsEngine.play(ttsText);
+      };
+    } else {
+      this.ttsSay = function(text) {
+        let cmd = { 'call': 'cactbotSay', 'text': ttsText };
+        window.callOverlayHandler(cmd);
+      };
+    }
+
     this.partyTracker = new PartyTracker();
     addOverlayListener('PartyChanged', (e) => {
       this.partyTracker.onPartyChanged(e);
@@ -635,13 +648,7 @@ class PopupText {
         };
         ttsText = ttsText.replace(/\s*(<[-=]|[=-]>)\s*/, arrowReplacement[lang]);
         // if we're overriding player name and browser tts engine is loaded, use that
-        if (Options.PlayerNameOverride !== null &&
-            typeof BrowserTTSEngine === 'function') {
-          BrowserTTSEngine.play(ttsText);
-        } else {
-          let cmd = { 'call': 'cactbotSay', 'text': ttsText };
-          window.callOverlayHandler(cmd);
-        }
+        this.ttsSay(ttsText);
       } else if (soundUrl && playSounds) {
         let audio = new Audio(soundUrl);
         audio.volume = soundVol;
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -78,12 +78,11 @@ class PopupText {
   OnPlayerChange(e) {
     // allow override of player via query parameter
     // only apply override if player is in party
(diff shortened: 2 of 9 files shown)


ui/raidboss/browser_tts_engine.js
```diff
@@ -0,0 +1,23 @@
...
+      // check to see if we've already gotten this tts before
+      let iframe = window.document.querySelector('[data-tts="'+text+'"]');
+      if (iframe === null) {
+        iframe = document.createElement('iframe');
+        // remove sandbox so we can modify contents/call play on audio element later
+        iframe.removeAttribute('sandbox');
+        iframe.style.display = 'none';
+        document.body.appendChild(iframe);
+        let encText = encodeURIComponent(text);
+        iframe.contentDocument.body.innerHTML = '<audio src="https://translate.google.com/translate_tts?ie=UTF-8&client=tw-ob&tl=en&q=' + encText + '" id="TTS">';
```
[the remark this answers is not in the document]
valarnin: Support for window.speechSynthesis is spotty for basically anything other than Windows Chrome. I only tested on my own android device and an android SDK emulator, for instance, but neither had any voices to use so invoking the speak function didn't actually cause any noise. Since at least two members of my static would be relying on this working via mobile phone, I decided that the HTML5 audio option was the better option.

The weird way that the audio embed itself is written is to work around the Google service not sending CORS headers, and sending a 404 response if there's a Referer header on the request. I'm certainly open to other TTS options, I just went with the easiest/quickest to implement that wouldn't be making a new request every time the same TTS was used.

For Player HP, is that not available in the client via memory for the party? Or is the concern more that the HP decrease might be delayed due to network differences? Same for party player positions.

ui/raidboss/popup-text.js
```diff
@@ -622,8 +636,14 @@ class PopupText {
           ko: ' 그리고 ',
         };
         ttsText = ttsText.replace(/\s*(<[-=]|[=-]>)\s*/, arrowReplacement[lang]);
-        let cmd = { 'call': 'cactbotSay', 'text': ttsText };
-        window.callOverlayHandler(cmd);
+        // if we're overriding player name and browser tts engine is loaded, use that
+        if (Options.PlayerNameOverride !== null &&
```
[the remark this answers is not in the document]
valarnin: That could certainly be done, yeah. I'll look into making that adjustment tomorrow.

ui/raidboss/browser_tts_engine.js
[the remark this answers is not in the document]
quisquous: https://codepen.io/matt-west/pen/wGzuJ worked fine on my Android device.  It looks like there's wide support for it on https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis.

Re: HP.  It's not that it's not available, it's just that it's not sent for every player.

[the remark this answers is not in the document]
valarnin: Maybe some fallback logic for this? Check for window.speechSynthesis and valid voices and use that if available, otherwise fall back to this logic?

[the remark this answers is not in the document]
valarnin: To clarify my previous point about speechSynthesis, consider this output:
```
typeof window.speechSynthesis
"object"
window.speechSynthesis.getVoices().length
0
```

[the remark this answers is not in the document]
quisquous: Huh! What browser has that?

ui/raidboss/browser_tts_engine.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+
+  play() {
+    this.item.play();
+  }
+}
+
+class BrowserTTSEngine {
+  constructor() {
+    engineType = null;
```
quisquous: `this.engineType` and `this.ttsItems`

ui/raidboss/popup-text.js
```diff
@@ -31,6 +31,19 @@ class PopupText {
     this.alertText = document.getElementById('popup-text-alert');
     this.alarmText = document.getElementById('popup-text-alarm');
 
+    if (Options.PlayerNameOverride !== null &&
+      typeof BrowserTTSEngine === 'function') {
+      this.ttsEngine = new BrowserTTSEngine();
+      this.ttsSay = function(text) {
+        this.ttsEngine.play(ttsText);
```
quisquous: s/ttsText/text/

ui/raidboss/raidboss.js
```diff
@@ -52,6 +54,18 @@ UserConfig.getUserConfigLocation('raidboss', function(e) {
...
   }
 
+  let PlayerNameOverride = params.get('player');
+  if (PlayerNameOverride !== null) {
+    Options.PlayerNameOverride = PlayerNameOverride;
+    console.log('Enabling player name override via query parameter, ' + PlayerNameOverride);
+    if (Options.audioAllowed) {
+      // include the tts engine script
+      // re-using this since it's already created
+      UserConfig.appendJSLink('browser_tts_engine.js');
```
quisquous: This gets parsed and run after gPopupText is created below, and so it's never included in time for the constructor.  Can you just include it always in the raidboss.html and emulator html?  I don't think load time is a serious concern here.

ui/raidboss/popup-text.js
```diff
@@ -31,6 +31,19 @@ class PopupText {
     this.alertText = document.getElementById('popup-text-alert');
     this.alarmText = document.getElementById('popup-text-alarm');
 
+    if (Options.PlayerNameOverride !== null &&
```
quisquous: This is a bit of a silly request, but can you add another option for "use act tts that defaults to true"? It's very likely that I will use this "overriding the player" logic for the raid emulator (which pretends to be a different player).  So, I'd like to remove the assumption that "using a player override" is the same case as "not using act tts".

ui/raidboss/browser_tts_engine.js
```diff
@@ -0,0 +1,87 @@
...
+    try {
+      if (this.ttsItems[text] !== undefined) {
+        this.ttsItems[text].play();
+      } else {
+        switch (this.engineType) {
+        case null:
+          // figure out what TTS engine type we need
+          if (window.speechSynthesis !== undefined &&
+            window.speechSynthesis.getVoices().length > 0)
+            this.engineType = TTSEngineType.SpeechSynthesis;
```
quisquous: Ok, sorry.  I led you astray, I think.  I didn't realize that `speechSynthesis` got caught up in the Chrome autoplay shenanigans and you can't use it before a user interacts with the document, so it's very likely unusable.  Maybe better to just remove the complexity.  I was hoping it would be easier to just lean on an existing web api instead of having to depend on a service (which is more likely to change or have errors, etc), but apparently that's not possible.

Sorry to waste your time with a half-baked suggestion.

ui/raidboss/browser_tts_engine.js
valarnin: This is fixed in my local git now, though it worked fine without the `this.` so it was more of a linting thing?

ui/raidboss/popup-text.js
valarnin: This is fixed in my local git now.

ui/raidboss/raidboss.js
valarnin: This is fixed in my local git now. It was originally conditional because the script had side effects, but that version of the script never made it into the git tree.

ui/raidboss/popup-text.js
valarnin: The only potential issue here is that if multiple people are using the same websocket for cactbot callouts, and any of them don't include the option to not `use act tts`, that player's alerts are going to route back through ACT to the person hosting causing them to get incorrect (dps vs healer, etc) or otherwise duplicate callouts with slight delay, causing a sort of echo effect.

ui/raidboss/browser_tts_engine.js
valarnin: They both suffer from the same problem vs autoplay. The reason for the fallback logic and external service is to address if the API is unavailable for some reason or if there are no voices available (which is what I've seen with my own mobile device and chromium/chrome/vivaldi on linux).

quisquous: Oh right, of course it would be the same autoplay issue.  I shouldn't do code review while exhausted.  This is all fine, then.

Ok! Sorry to push on this.  Maybe I had just hoped there could be an easier solution somewhere here.

ui/raidboss/popup-text.js
quisquous: Ah, I see the fear here.  I would be fine with an override *parameter* also setting the act tts option, and not requiring a second parameter.  My request was more around the code inside the constructor, etc.  Yeah, it's just the constructor so in practice it doesn't matter, but I just want to decouple the two options in case it proliferates more in the future.

ui/raidboss/browser_tts_engine.js
valarnin: The best I could think of is to detect the autoplay issue (there's google documentation on how to detect that interaction is still required) and prompt the user to click a button or something if we're using browser tts instead of ACT tts.

ui/raidboss/popup-text.js
valarnin: So add the option, default it to true, and if the player override is specified default it to false, and finally make sure the check for the parameter happens after the check for player override to allow that parameter to still take priority?

ui/raidboss/browser_tts_engine.js
quisquous: Yeah, that sounds like a good approach.  I wonder if the button should be at load time though, as it would not be great to be midfight and suddenly have a button to hit before tts works.

ui/raidboss/popup-text.js
quisquous: Yes, up to the "and finally" part and no after that.  I was thinking no parameter for act tts, just an option.  I just want to decouple the two behaviors of tts and player override in popuptext.

ui/raidboss/browser_tts_engine.js
valarnin: Maybe add the button (as a position absolute to avoid layout issues) in the `BrowserTTSEngine` constructor if interaction is still required? That should get triggered on page load.

ui/raidboss/popup-text.js
valarnin: Alright, I understand the intention now, but wouldn't that config just use the ACT host's user config, therefore causing the same problem as before (ACT host gets either all the callouts or none of the callouts)?

ui/raidboss/browser_tts_engine.js
quisquous: This was an error for me.  I don't know how that worked with 'use strict'.

ui/raidboss/popup-text.js
quisquous: If you set the parameter inside that same function where you're looking at parameters in raidboss.js, then that will override everything any user config happened to set.

ui/raidboss/browser_tts_engine.js
quisquous: Yeah, I'm not sure what makes the most sense.  Maybe adding a button would be a good followup patch and I can merge this without it to start so other people can start using it too.

ui/raidboss/browser_tts_engine.js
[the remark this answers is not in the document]
valarnin: Vivaldi (chromium) on Linux. Same thing with normal Chrome on my Linux distro, as well.

ui/raidboss/popup-text.js
valarnin: Please review the latest commits to make sure we're on the same page here for this functionality. It made more sense to me for it to be an opt-in-to-browser-tts option, rather than an opt-out-of-act-tts option.

ui/raidboss/browser_tts_engine.js
valarnin: I can code this up tomorrow sometime. Would you want the styles for this put in a stylesheet somewhere or should I just do the entire button inline?

quisquous: Probably the normal raidboss.css file is a good spot.

ui/raidboss/raidboss.js
```diff
@@ -52,6 +54,12 @@ UserConfig.getUserConfigLocation('raidboss', function(e) {
       console.log('Enabling audio via query parameter');
   }
 
+  let PlayerNameOverride = params.get('player');
+  if (PlayerNameOverride !== null) {
+    Options.PlayerNameOverride = PlayerNameOverride;
+    console.log('Enabling player name override via query parameter, ' + PlayerNameOverride);
```
quisquous: Sorry for any bad communication on my part.  What I wanted was to set `Options.BrowserTTS = false` here, and to have the line in the PopupText constructor just say `if (this.options.BrowserTTS) {`, so-as to not couple these concepts so closely.  Doing it here will always override any config or user javascript file.

ngld:
Regarding the autoplay issue: The same issue also applies to alert sounds (any sound really).
If you add the button, it'd probably be best to show it whenever the overlay is loaded over WebSocket and add an option to disable it because clicking anything in OBS' browser source is annoying.

anubiann00b:
Chiming in on the discord bot use case--I don't think this would solve the entire problem, but it definitely adds flexibility--someone could play dps, but set the callout person as a healer to get callouts for raidwides and and tankbusters and such?

I haven't messed around with OverlayPlugin's websocket server features--can you use it for the same purpose as ACT_DiscordTriggers?

(this change looks neat by the way, ty!)

quisquous:
> Chiming in on the discord bot use case--I don't think this would solve the entire problem, but it definitely adds flexibility--someone could play dps, but set the callout person as a healer to get callouts for raidwides and and tankbusters and such?

The idea would be that you wouldn't have a discord bot.  You'd have one person running ACT, and then everybody can connect to their ACT and open up a webpage locally on their other monitor or on their phone or whatever and it would do individual callouts for those people.

ngld:
> I haven't messed around with OverlayPlugin's websocket server features--can you use it for the same purpose as ACT_DiscordTriggers?

The WebSocket interface allows you to display your overlay on anything that has a web browser. You can use a second PC, laptop, smartphone, tablet, a smart TV, your fridge, your car, whatever. As long as it has a modern web browser, it can display your overlay (MopiMopi, Kagerou, Cactbot, etc.).

You can open the overlay on several devices which means that you can have one person in a group run ACT and everyone else can then open the overlay in their web browser.
However, since the overlays assume that the user is running ACT, all members of the group will get the same callouts or have the same person labeled as `YOU` on the DPS overlay. This PR fixes that issue (specifically the callouts for Cactbot).

ui/raidboss/raidboss.js
valarnin: Will setting `Options.BrowserTTS` here persist that change to the config file, or is it purely ephemeral?

valarnin:
> Regarding the autoplay issue: The same issue also applies to alert sounds (any sound really).
> If you add the button, it'd probably be best to show it whenever the overlay is loaded over WebSocket and add an option to disable it because clicking anything in OBS' browser source is annoying.

This is a good point, the non-TTS audio alerts also won't work in a browser without user interaction with the page, e.g. this code from `popup-text.js`:
```javascript
      } else if (soundUrl && playSounds) {
        let audio = new Audio(soundUrl);
        audio.volume = soundVol;
        audio.play();
      }
```

@quisquous does it make sense to write this check and button as something completely independent and trigger it on page load if audio alerts are enabled or browser tts is enabled, rather than having it tied to the browser TTS functionality directly?

valarnin:
> The WebSocket interface allows you to display your overlay on anything that has a web browser. You can use a second PC, laptop, smartphone, tablet, a smart TV, your fridge, your car, whatever. As long as it has a modern web browser, it can display your overlay (MopiMopi, Kagerou, Cactbot, etc.).

Just a quick note that Cactbot itself only works in Chromium-based browsers. Other overlays might work in other browsers.

Maybe once I get this PR finished and settled I'll look into getting it to work in other browsers as well, but when I was testing a week ago even after I fixed the JS errors stopping Cactbot from working in Firefox there were something like 50 regex errors in the trigger js files that I just didn't have time to look at.

ui/raidboss/raidboss.js
quisquous: Yeah, it's ephemeral.  Only changes on the config page itself get persisted to the json file.  Other settings (at runtime, or in user javascript files), are just temporary overrides.

quisquous:
Sorry, I did not mean to close this!

quisquous:
> > Regarding the autoplay issue: The same issue also applies to alert sounds (any sound really).
> > If you add the button, it'd probably be best to show it whenever the overlay is loaded over WebSocket and add an option to disable it because clicking anything in OBS' browser source is annoying.
>
> @quisquous does it make sense to write this check and button as something completely independent and trigger it on page load if audio alerts are enabled or browser tts is enabled, rather than having it tied to the browser TTS functionality directly?

Yeah, I think it would make sense to have this be independent.  I'm also not quite sure that I understand precisely which cases an interaction is needed for, as this hasn't been an issue in practice.

> Just a quick note that Cactbot itself only works in Chromium-based browsers. Other overlays might work in other browsers.
> 
> Maybe once I get this PR finished and settled I'll look into getting it to work in other browsers as well, but when I was testing a week ago even after I fixed the JS errors stopping Cactbot from working in Firefox there were something like 50 regex errors in the trigger js files that I just didn't have time to look at.

Yeah, Firefox isn't the primary use case, but if it's easy enough to fix, then I'm happy to take fixes, especially if there's any automated testing that can be done.

anubiann00b:
> > I haven't messed around with OverlayPlugin's websocket server features--can you use it for the same purpose as ACT_DiscordTriggers?
> 
> The WebSocket interface allows you to display your overlay on anything that has a web browser. You can use a second PC, laptop, smartphone, tablet, a smart TV, your fridge, your car, whatever. As long as it has a modern web browser, it can display your overlay (MopiMopi, Kagerou, Cactbot, etc.).
> 
> You can open the overlay on several devices which means that you can have one person in a group run ACT and everyone else can then open the overlay in their web browser.
> However, since the overlays assume that the user is running ACT, all members of the group will get the same callouts or have the same person labeled as `YOU` on the DPS overlay. This PR fixes that issue (specifically the callouts for Cactbot).

oh this is rad!!! yeah this totally solves the discord bot use case lol this would be wayyy more convenient

ngld:
> I'm also not quite sure that I understand precisely which cases an interaction is needed for, as this hasn't been an issue in practice.

I noticed these issues when I worked on the new CEF version for OverlayPlugin. I had to [explicitly disable the autoplay-requires-interaction rules](https://github.com/ngld/OverlayPlugin/blob/b71797f9c98eb02379a9a5f836d7210fc5cc76dd/HtmlRenderer/Renderer.cs#L437-L438). These rules are obviously still enabled in a real browser and as such Cactbot will behave differently depending on whether it's running in an overlay or a browser.

Anything that plays sound requires a user interaction. You have to assume that the whole page is muted until the user clicks or types inside the page. It doesn't matter if there's a button or not, even a click on an alert or a white background counts as an interaction.

valarnin:
Hopefully I did that rebase right, never had to rebase to an upstream project before.

ngld:
Uh, no. You merged in the recent changes instead of rebasing. See all of those commits above? Especially the merge commit? Those are a sign that the rebase didn't go as expected.

If you want to try again, use these commands:
```bash
# First, throw away the merge commit
git reset --hard 2f17609
# Then rebase
git rebase origin/master
# And finally, push
git push --force
``` 

You might want to double-check the commit hash above. It should be the last commut you made before the merge. If it's wrong, you'll lose everything between the specified commit and the merge so make sure I got it right.
The `--force` at the end is necessary because a rebase rewrites commits and thus overwrites existing history. A normal `push` doesn't allow that since it's a destructive process.

anubiann00b:
you can make a backup of your current branch state with `git checkout -b my-branch-backup` btw it'll make it easier to recover if you reset the wrong thing

valarnin:
Alright, try number two at rebase, looks much better now I hope. I was missing the `--force` on my first attempt which threw off the entire process.

ui/raidboss/popup-text.js
```diff
@@ -227,8 +244,22 @@ class PopupText {
   }
 
   OnJobChange(e) {
-    this.me = e.detail.name;
-    this.job = e.detail.job;
+    // allow override of player via query parameter
+    // only apply override if player is in party
+    if (Options.PlayerNameOverride !== null &&
+        this.partyTracker.inParty(Options.PlayerNameOverride)) {
```
quisquous: I'm a little bit not sure of this.  Sometimes "being in the party" is quite odd.  Like, a cross world party not in an instance doesn't have anybody in the party.  Or you and your chocobo aren't a party.  Maybe I'm overthinking a first version here, but I wonder a little if we need a "JobOverride" default rather than just not using the override at all if there's not in a party.  This can be punted to a future patch and fixed later if it ever becomes an issue, but just thinking out loud.  I'm fine to merge without changing this unless you have some easy idea that I don't see.

ui/raidboss/popup-text.js
```diff
@@ -227,8 +244,22 @@ class PopupText {
   }
 
   OnJobChange(e) {
-    this.me = e.detail.name;
-    this.job = e.detail.job;
+    // allow override of player via query parameter
+    // only apply override if player is in party
+    if (Options.PlayerNameOverride !== null &&
```
quisquous: I tested this out locally, and any time `OnPlayerChange` is called, then `data.me` no longer equals `e.detail.name` because it's set here, and so it reloads everything on every update.  I think it might make more sense to apply the override in `OnPlayerChange` and edit the event that gets passed to `OnJobChange`?

ui/raidboss/browser_tts_engine.js
```diff
@@ -0,0 +1,87 @@
...
+  play(text) {
+    try {
+      if (this.ttsItems[text] !== undefined) {
+        this.ttsItems[text].play();
+      } else {
+        switch (this.engineType) {
+        case null:
+          // figure out what TTS engine type we need
+          if (window.speechSynthesis !== undefined &&
+            window.speechSynthesis.getVoices().length > 0)
```
quisquous: Oddly, I found this extremely inconsistent.  If I open up raidboss in Chrome 82.0.4073.0, click the button, and then type `window.speechSynthesis.getVoices()` in the console, it comes up empty and then the second time it comes up with a list of voices.  So, in practice, this never works out.  Maybe there's a bug in Chrome, but something is awry here for sure.  I'm also fine to merge this as-is and fix this later, as both paths seem to work just fine once the voices get initialized.

ui/raidboss/popup-text.js
valarnin: Yeah, I think even out-of-dungeon-and-in-separate-overworld-zones cases have issues with not being in the party properly, that's what prompted me to add this in the first place. Maybe this should be moved to its own function and fired here where it was originally as well as in something like zone change?

ui/raidboss/popup-text.js
valarnin: Agreed. I'll see about making this adjustment tomorrow if I get time before raid.

ui/raidboss/browser_tts_engine.js
valarnin: So it seems like the voices list doesn't populate until after the first interaction with `speechSynthesis`, and even then it loads async. I'll think about how to better handle this, maybe default to the google service and fall back to browser tts based on if the event fires successfully and has voices afterwards...

valarnin:
Just as a heads up, the current version will cause load issues with `raidemulator.html` due to expecting the new js files to be loaded. I've fixed this locally but haven't pushed it up to github yet.

ui/raidboss/popup-text.js
valarnin: Added `job` parameter in latest changes.

ui/raidboss/popup-text.js
valarnin: Moved this logic up to `OnPlayerChange` in latest changes.

ui/raidboss/browser_tts_engine.js
valarnin: I've refactored this to a manner that _should_ work properly now with the latest changes, but I don't have a good way to test it on my end. Can anyone with a working `window.speechSynthesis` that has voices test the new logic? Any Chrome on Windows should work, just pop open the dev tools and breakpoint `browser_tts_engine.js` line 50 and see if it's ever hit on a page refresh.

quisquous: Works for me, thanks!

ui/raidboss/browser_tts_engine.js
```diff
@@ -0,0 +1,91 @@
...
+  play() {
+    this.item.play();
+  }
+}
+
+class BrowserTTSEngine {
+  constructor() {
+    // figure out what TTS engine type we need
+    if (window.speechSynthesis !== undefined) {
+      let me = this;
```
quisquous: If you change the `function() {` on the next line to `() => {`, then it won't rebind this and you can just say `this.engineType` inside of the function without the extra variable.

ui/raidboss/popup-text.js
```diff
@@ -59,6 +76,22 @@ class PopupText {
   }
 
   OnPlayerChange(e) {
+    // allow override of player via query parameter
+    // only apply override if player is in party
+    if (Options.PlayerNameOverride !== null &&
+        this.partyTracker.inParty(Options.PlayerNameOverride)) {
```
quisquous: Can you move this partyTracker check just before the `partyTracker.jobName` call? I had suggested the "job override" as a way to have this work when the party tracker doesn't have the person in the party.

Sorry for the comments here, but I just also filed #1153 which has issues with the party name.  So, this feels extra fragile.

ui/raidboss/browser_tts_engine.js
valarnin: Fixed in latest changes, thanks.

ui/raidboss/popup-text.js
valarnin: This has been updated in the latest changes. Please re-review this to make sure I understood your request properly.

quisquous:
Thanks for this!!

SolarMisae:
Hey there! It's been awhile since this feature was added, and I'm curious if we'll ever get a plebeians guide to setting this up? As someone who knows nothing about how any of this works I can't for the life of me figure out how to use this feature with cactbot so I feel pretty blind/dumb. I'm one of those people who currently runs a discord bot and would love to switch to this (as the bot causes me some issues these days), because getting my raid group to install their own cactbot is like herding cats; they're significantly more technologically inept than me, and I'm inept enough as it is. =X

quisquous:
There's https://github.com/quisquous/cactbot/blob/503dc66bf8391aed2c86cf9bab3a9097fd1c4098/docs/RemoteCactbot.md

I haven't merged it yet because ngld is working on a built-in version of this to OverlayPlugin, where you'd just need to hit one button to do most of the steps there.

SolarMisae:
Thanks so much for the quick response! And nice to know an even easier way of doing it is being worked on. :)

PR #1115, opened 2020-02-26
raidboss: Increase Alerts for Getting Orbs in E5S

panicstevenson:
Raise infoText level alerts for getting an orb to alertText if you don't
already have one.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e5s.js
+++ b/ui/raidboss/data/05-shb/raid/e5s.js
@@ -6,25 +6,37 @@
     ko: /^희망의 낙원 에덴: 공명편\(영웅\) \(1\)$/,
   },
   timelineFile: 'e5s.txt',
+  timelineTriggers: [
+    {
+      id: 'E5S Pre-Chain Lightning',
+      regex: /^Chain Lightning$/,
+      beforeSeconds: 4,
+      alertText: function(data) {
+        if (!data.surgeProtection) {
+          return {
+            en: 'Grab an orb',
+            fr: 'Prenez un orbe',
+            ko: '구슬 줍기',
+          };
+        }
+      },
+    },
+  ],
   triggers: [
     {
-      id: 'E5S Fury\'s Bolt Gain',
-      regex: Regexes.gainsEffect({ target: 'Ramuh', effect: 'Fury\'s Bolt', capture: false }),
-      regexDe: Regexes.gainsEffect({ target: 'Ramuh', effect: 'Wütender Blitz', capture: false }),
-      regexFr: Regexes.gainsEffect({ target: 'Ramuh', effect: 'Boules de foudre', capture: false }),
-      regexJa: Regexes.gainsEffect({ target: 'ラムウ', effect: 'チャージボルト', capture: false }),
+      id: 'E5S Surge Protection Gain',
+      regex: Regexes.gainsEffect({ effect: 'Surge Protection' }),
+      condition: Conditions.targetIsYou(),
       run: function(data) {
-        data.fury = true;
+        data.surgeProtection = true;
       },
     },
     {
-      id: 'E5S Fury\'s Bolt Lose',
-      regex: Regexes.losesEffect({ target: 'Ramuh', effect: 'Fury\'s Bolt', capture: false }),
-      regexDe: Regexes.losesEffect({ target: 'Ramuh', effect: 'Wütender Blitz', capture: false }),
-      regexFr: Regexes.losesEffect({ target: 'Ramuh', effect: 'Boules de foudre', capture: false }),
-      regexJa: Regexes.losesEffect({ target: 'ラムウ', effect: 'チャージボルト', capture: false }),
+      id: 'E5S Surge Protection Loss',
+      regex: Regexes.losesEffect({ effect: 'Surge Protection' }),
+      condition: Conditions.targetIsYou(),
       run: function(data) {
-        data.fury = false;
+        data.surgeProtection = false;
       },
     },
     {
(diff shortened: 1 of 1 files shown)


panicstevenson:
Since we already have working triggers, this pull request is meant to be more of a working proposal.

valarnin:
Orbs aren't required for chain lightning. They make it easier, but 100% not needed. Not sure if it makes sense to raise the alert level for orbs when they're an optional mitigation for the mechanic?

Jaehyuk-Lee:
If you grab an orb on chain lightning, you will have less orbs on field that could be bad for party members.

panicstevenson:
Wait, really? I always thought it was necessary. I can remove that then.

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -175,10 +236,15 @@
...
-        ko: '자기 위치에 있기',
+      infoText: function(data) {
+        if (!data.furysBoltActive) {
+          return {
+            en: 'Short Knockback',
+          };
+        }
+        return {
+          en: 'Big Knockback',
+        };
```
Jaehyuk-Lee: `E5S Centaur\'s Charge` is always Big Knockback.
It happens only after fury's fourteen.
I think you were thinking about the add pushing party members from the center to sides, which is completely not dealt with e5s trigger now.
Think of that, tank moves Ramuh during first pushing.

panicstevenson: Understood; yea, I had those confused then.

panicstevenson:
I wonder if it makes more sense to instead have a timelineTrigger for Stepped Leader that specifies that it's next, and on-cast wipe that information.

Jaehyuk-Lee:
I agree with making timeline trigger with Stepped Leader, too. I made it Trigger because there were no timeline when I was making this. + I don't know how to make it. 🤣

Jaehyuk-Lee:
I think
```javascript
data.furysBoltEnum = {
          1: 'Judgment Volts',
          2: 'Tribunal Summons',
          3: 'Thunderstorm',
          4: 'Stepped Leader',
          5: 'Thunderstorm',
          6: 'Judgment Volts',
          7: 'Judgment Volts',
          8: 'Judgment Volts',
        };
```
is better to run only once, by using Timeline Trigger.
It's not a big deal, but just.

valarnin:
> Wait, really? I always thought it was necessary. I can remove that then.

Yeah, for chain lightning the orbs reduce the damage taken by the lightning pass by I think 25% or something? It's really not needed unless the group has no coordination, the only risky parts are the tankbuster (just shield it) and the raidwide (reprisal+any two other mitigations, e.g. temperance/shields/soil/etc). I've cleared e5s twice now and both times we just didn't bother with orbs at all unless they were directly in the way of positioning for chain lightning.

Aside from the offtank add kite, Orbs are important for the following mechanics:
- Judgement Volts (only if fury bolt is up)
- Stormcloud Summons (melee needs orb while clearing debuff, orb always required)
- Tribunal Summons/whatever the charge ability is (only if fury bolt is up)
- Thunderstorm (orb always required for blue lightning aoe)
- Fury's Fourteen (orb always required for knockback)

That's all of the mechs that require orb I think, unless I'm forgetting one.

Since we're on the subject of orbs though, I wonder if anything can be done for offtank orb count? Can we detect main tank, and have off tank's get orb text be dependent on the max stack size of the rest of the party (e.g. 6 people have 1 stack, 1 person has 2, off tank has 2, so off tank gets red text to grab extra orb)? Might be a bit beyond the scope here, I'm not sure.

panicstevenson:
>I wonder if anything can be done for offtank orb count?

You can track all of the gains, the Fury's Bolts, and any losses. Unfortunately, since 'gains effect' only has the name of the status, they're all Surge Protection (unless you were to go over). Does Raiden's tank buster remove a stack?

ui/raidboss/data/05-shb/raid/e5s.js
```diff
@@ -6,25 +6,31 @@
...
-      regex: Regexes.losesEffect({ target: 'Ramuh', effect: 'Fury\'s Bolt', capture: false }),
-      regexDe: Regexes.losesEffect({ target: 'Ramuh', effect: 'Wütender Blitz', capture: false }),
-      regexFr: Regexes.losesEffect({ target: 'Ramuh', effect: 'Boules de foudre', capture: false }),
-      regexJa: Regexes.losesEffect({ target: 'ラムウ', effect: 'チャージボルト', capture: false }),
+      id: 'E5S Surge Protection Loss',
+      regex: Regexes.losesEffect({ effect: 'Surge Protection' }),
+      condition: Conditions.targetIsYou(),
       run: function(data) {
-        data.fury = false;
+        data.surgeProtection = false;
```
quisquous: Do these effect messages only show up when you go from 0 to 1? Is there any message for more than that? Also half wondering if pro level triggers should look at times to figure out if this will run out before the mechanic you need it for too...

valarnin:
> Does Raiden's tank buster remove a stack?

If I remember correctly, no. It shouldn't matter either way, IIRC, based on how the fight is laid out the offtank is going to require the extra stack to ensure the add is on them, and there's never a case where two mechanics requiring it would hit before the buff wears off.

ui/raidboss/data/05-shb/raid/e5s.js
panicstevenson: The gain messages show up on every orb pickup.
The loss messages show up every time you lose the entirety of the buff or if you overcharge.

quisquous: Ah interesting, so if we wanted to track timers, you could just reset the timer on pickup.

quisquous:
I used this during reclears tonight and it was great! Thanks!

PR #1116, opened 2020-02-26
Raidboss: E5n timeline and triggers

JLGarber:
Minimal testing, but should be largely accurate. (I haven't had any runs that lasted long enough to fully test.) Because we don't know all (or any) of the phase pushes, I've put aggressive syncs all over. Apologies for stylistic things that may not match after the last couple months' changes. Triggers and translations are shamelessly copied/pasted from E5s with the necessary modifications.

Potential improvements:

1. Math out the spawn points of the lateral adds, so we can warn the user which lanes are safe.
2. Warn the user about proximity if they get the Stormcloud marker. (During the rotation block, Stratospear Summons is combined with a 2x Stormcloud marker, and if they are dropped too close together it repeatedly inflicts a 30s damage down that *also* consumes orb stacks.)


quisquous:
I think one thing we could do is just say "drop clouds away from each other", either every time, or especially the one with the knockback.  Also, add an oopsy warning for the cloud tether.

PR #1118, opened 2020-02-27
Raidboss: Initial e6n timeline

JLGarber:
No triggers alongside this yet, I'll probably have those soon enough. Minimal testing went into this etc.

Ferostorm is not synced because there are multiple different ability IDs to go with it. (I believe each cardinal direction can be one, in much the same way Double Sever in Fractal Continuum, or Port/Star and Fore/Aft in Temple of the Fist have separate IDs for each direction.) Rather than try to find all the correct IDs for both Garuda and Raktapaksa, I figured it was better just not to sync at all. [(The syncs are commented out rather than deleted so that if we do translation work based on timelines in the future, we still have those ability IDs available.](https://github.com/quisquous/cactbot/issues/749#issuecomment-557830256))

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6n.txt
+++ b/ui/raidboss/data/05-shb/raid/e6n.txt
@@ -4,4 +4,101 @@
 hideall "--Reset--"
 hideall "--sync--"
 
+# -ii 4BDD 4BE1 4BEB 4BEA 4BE5 4BE9 4BEC 4BD9 4BE7 4E39 4C1D 4C20 4F98
+# -ic "Tumultuous Nexus" "Great Ball Of Fire"
+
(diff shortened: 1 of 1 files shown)


quisquous:
Sounds good for now.  I think ideally we sync both of them on multiple ids? You can see that in e8s.  It's not super critical, but it's also helpful for me to have all those ids documented where possible.

JLGarber:
You're right, I'll do that when I get the triggers set up. [XIVAPI](https://xivapi.com/search?indexes=action&string=Ferostorm&columns=Name_en,ID) seems to say there are 12(!) different Ferostorm abilities present across Normal and Savage. Off the top of my head, I assume there are 4x cardinals for each difficulty, plus the castbar, plus the added effect. But we'll see when I poke at it further.

PR #1121, opened 2020-02-27
Util: QoL adjustment for timeline_adjust

JLGarber:
It's the tiniest of tiny nits, but every time I go to adjust a timeline I forget that the arguments aren't aliased. No more forgetting!


quisquous:
Appreciate all the QoL you want to send :)

PR #1132, opened 2020-02-28
test: add test for field ordering in triggers

quisquous:
Functions run in a particular order, so enforce that order in how
triggers are written to avoid logic errors.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -459,6 +459,7 @@ class PopupText {
       // Otherwise, if multiple alarm/alert/info are specified
       // it will pick one sound in the order of alarm > alert > info.
       let soundUrl = ValueOrFunction(trigger.sound);
+      let triggerSoundVol = ValueOrFunction(trigger.soundVolume);
       let soundVol = 1;
 
       let defaultTTSText;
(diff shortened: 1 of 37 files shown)


quisquous:
Did anybody have any feelings about this? I know there was some confusion about when things run, and so having fields ordered by when they run seems like it would only help in trigger readability, even if it makes Travis be even more of a jerk.

(cc @panicstevenson)

panicstevenson:
It seems good to me; maybe I'll finally remember to run tests 100% of the time 🤞

ui/raidboss/popup-text.js
```diff
@@ -587,11 +588,8 @@ class PopupText {
         }
       }
 
-      if ('soundVolume' in trigger)
-        soundVol = ValueOrFunction(trigger.soundVolume);
-
       soundUrl = triggerOptions.SoundOverride || soundUrl;
-      soundVol = triggerOptions.VolumeOverride || soundVol;
+      soundVol = triggerOptions.VolumeOverride || triggerSoundVol || soundVol;
```
panicstevenson: Is this from a separate change?

quisquous: Reply for others, it's from the middle commit d5b001c.

PR #1140, opened 2020-02-29
Resources: Spelling fix for responses.js

JLGarber:
I did a few searches and couldn't find any other uses in the repository, so I think we're safe to change this. E6n will most likely be a near-future use once I get the triggers working for that.


quisquous:
Thanks for fixing this NEAR DISASTER

PR #1141, opened 2020-02-29
Raidboss: Initial E6n triggers and updated timeline

JLGarber:
As with E5n, most of this is shamelessly stolen from elsewhere in the repository. I think I caught all the changes necessary, but I can't test the tank/healer triggers due to not being comfortable in this content with those roles. Everything that I was able to test functions as expected. Timing on targetable/untargetable lines may need slight adjustments.

To do:

1. Math out Ifrit's jump after Strike Spark, so we can tell the user the direction instead of just a simple warning.

2. Warn the user of the surprise Eruptions alongside Enumeration + Hands of Hell. (It's going to be just complex enough to call that I'm leaving it out for now, but it's one of the MAJOR sources of deaths in pickup runs. I'm actually a little surprised Square included it in a normal-difficulty encounter.)

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6n.js
+++ b/ui/raidboss/data/05-shb/raid/e6n.js
@@ -6,8 +6,271 @@
     ko: /^희망의 낙원 에덴: 공명편 \(2\)$/,
   },
   timelineFile: 'e6n.txt',
+  timelineTriggers: [
+    {
+      // We warn the user here because the startsUsing warning gives only 3.5s or so.
+      id: 'E6N Downburst',
+      regex: /Downburst/,
+      beforeSeconds: 5,
+      response: Responses.knockback('info'),
+    },
+
+  ],
   triggers: [
+    {
+      id: 'E6N Superstorm',
+      regex: Regexes.startsUsing({ source: 'Garuda', id: '4BD7', capture: false }),
+      condition: Conditions.caresAboutMagical(),
+      response: Responses.aoe(),
+    }, 
+    {
+      id: 'E6N Ferostorm',
+      regex: Regexes.startsUsing({ source: ['Garuda', 'Raktapaska'], id: ['4BD[DEF]', '4BE[345]'] }),
+      infoText: {
+        en: 'Avoid green nails',
+      },
+    },
+    {
+      id: 'E6N Air Bump',
+      regex: Regexes.headMarker({ id: '00D3' }),
+      suppressSeconds: 1,
+      infoText: function(data, matches) {
+        if (data.me == matches.target) {
+          return {
+            en: 'Enumeration on YOU',
+            de: 'Enumeration aud DIR',
+            fr: 'Enumération sur VOUS',
+          };
+        }
+        return {
+          en: 'Enumeration',
+          de: 'Enumeration',
+          fr: 'Enumération',
+        };
+      },
+    },
+    {
+      id: 'E6N Inferno Howl',
+      regex: Regexes.startsUsing({ source: ['Ifrit', 'Raktapaksa'], id: '4BF1', capture: false }),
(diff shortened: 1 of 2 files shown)


quisquous:
If you want more precise targetable lines, you can use the https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#22-networknametoggle lines.  Usually this is where I start, although the tool doesn't generate them (yet!)

JLGarber:
I may take a look later on in the tier for how easy it would be to add that functionality to `make_timeline`. It would definitely be helpful for battles like this one!

ui/raidboss/data/05-shb/raid/e6n.js
```diff
@@ -6,8 +6,271 @@
...
+      id: 'E6N Superstorm',
+      regex: Regexes.startsUsing({ source: 'Garuda', id: '4BD7', capture: false }),
+      condition: Conditions.caresAboutMagical(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'E6N Ferostorm',
+      regex: Regexes.startsUsing({ source: ['Garuda', 'Raktapaska'], id: ['4BD[DEF]', '4BE[345]'], capture: false }),
+      infoText: {
+        en: 'Avoid green nails',
```
quisquous: Do you know which ways each of these abilities go? I wanted to call left/right on these, both here and in savage...

ui/raidboss/data/05-shb/raid/e6n.txt
```diff
@@ -25,19 +25,22 @@ hideall "--sync--"
 81.7 "Vacuum Slice" sync /:Garuda:4BD5:/
 88.8 "Occluded Front" sync /:Garuda:4BD2:/ window 30,30
 98.3 "Irresistible Pull" sync /:Garuda:4BD6:/
+111.5 "--untargetable--" sync /22:........:Garuda:........:Garuda:00/
 
 # Ifrit solo
-114.0 "Touchdown" sync /:Ifrit:4BE8:/ window 30,30
+114.0 "--targetable--" sync /22:........:Ifrit:........:Ifrit:01/ window 30,30
```
quisquous: This is maybe personal preference, but I wanted to sync on the abilities, but use the 22 lines to generate the targetable/untargetable.  I guess we could sync on them too, but nothing else does.  I think it's worth keeping Touchdown here as well as it does damage.

ui/raidboss/data/05-shb/raid/e6n.js
JLGarber: Unfortunately I don't. I'll check for that as I have time, but I don't record, so it's difficult without an organized group.

ui/raidboss/data/05-shb/raid/e6n.txt
JLGarber: Fair point on Touchdown, that's very true. Is the issue with syncing `22` lines mostly that `test_timeline` currently doesn't support them?

quisquous: Yeah, I think it's that test_timeline doesn't support them mostly.  I also don't know about how reliable they are? The 15 lines are very accurate and don't drift.  14 is a little bit iffy according to an issue folks filed about network vs game log lines.  I think it's also just consistency in terms of syncing to abilities where possible, because that's what everything else uses.

ui/raidboss/data/05-shb/raid/e6n.js
quisquous: Ok! You can also always upload to fflogs and look at the playback and see where people move.

JLGarber: I'll try that tomorrow, thanks for the suggestion.

PR #1143, opened 2020-02-29
jobs: add language info in pull-bar container

Jaehyuk-Lee:
This commit will apply #1056 changes also into jobs module.
(better legibility for Korean)


Jaehyuk-Lee:
class name should be `'lang-' + language` because of this css.
https://github.com/Jaehyuk-Lee/cactbot/blob/9ae1ccc87a5dfd771262b018d3b6724a52cfbb83/resources/timerbar.js#L269-L276

PR #1144, opened 2020-03-01
Raidboss: E7n timeline and triggers

JLGarber:
As with the previous Eden normal modes, most of the triggers and all the translation is copied from Savage.

I've noted it in a comment in-line, but I put in warnings for the intermission even though ground markers are visible. It's really easy to forget that there are ground markers on the boss when the user is either about to teleport or is evading 7 other people with spread markers. I don't think this is too much a violation of "don't warn on visible markers", but I'm fine removing it if desired.

The dark/light triggers aren't great, but I thought this was still probably a better option than trying to retrofit the Savage color map. Since the Normal mechanics are just a Dark/Light/Dark or Light/Dark/Light pattern for all color phases, this seemed simplest.

For the future, it would be nice to determine the ability codes for where the Decrees spawn. Since the red/blue portals are fixed, this would let us tell the user exactly where to go. (It's possible this was in the Savage triggers and I skipped over it. I've been working on this package too long today, however, so I figured it was best to just release it and come back to it later.)


ui/raidboss/data/05-shb/raid/e7n.txt
```diff
@@ -4,4 +4,99 @@
...
 0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+
+# Opening block
+0.0 "--sync--" sync /:Engage!/ window 0,1
+3.8 "--sync--" sync /:The Idol Of Darkness:4C50:/ window 3,1
+12.8 "Empty Wave" sync /:The Idol Of Darkness:4C52:/ window 12,30
+25.0 "Unshadowed Stake" sync /:The Idol Of Darkness:4C51:/
+37.8 "Words of Motion" sync /:The Idol Of Darkness:4C2B:/
+49.6 "Light's Course" sync /:Unforgiven Idolatry:4C3C:/
```
quisquous: Can you comment out these syncs on Light's Course because their so close to each other?

ui/raidboss/data/05-shb/raid/e7n.txt
```diff
@@ -4,4 +4,99 @@
...
+236.8 "Away with Thee" sync /:The Idol Of Darkness:(?:4C39|4E7E):/
+251.7 "Strength in Numbers" sync /:Idolatry:4C4[CD]:/
+262.3 "Silver Shot" sync /:The Idol Of Darkness:4E7[CD]:/
+270.3 "Silver Shot" sync /:The Idol Of Darkness:4E7[CD]:/
+271.7 "Strength in Numbers" sync /:Idolatry:4C4[CD]:/
+281.7 "Explosion enrage?"
+
+
+# Rotation block. 135.6 seconds
+400.0 "Empty Flood" sync /:The Idol Of Darkness:(?:4E5[46]|4C53):/ window 300,20
```
quisquous: window 400?

ui/raidboss/data/05-shb/raid/e7n.js
```diff
@@ -7,7 +7,253 @@
...
+        en: 'Teleport into donut',
+      },
+    },
+    {
+      // Ordinarily we might not warn on ground AoE markers. However, there are player-dropped
+      // markers just before this, so it might be difficult to see.
+      id: 'E7N Strength In Numbers Circle',
+      regex: Regexes.startsUsing({ source: 'Idolatry', id: '4C4D', capture: false }),
+      suppressSeconds: 1,
+      response: Responses.getOut(),
```
quisquous: Sounds good! I think "get out" is also a reasonable direction to give vs some sort of ambiguous "avoid those aoes on the ground" kind of thing.

JLGarber:
I'm mostly doing this alongside weekly normal clears, and then running 2-3 more times to get the data I need. Unfortunately, my group isn't at 7s yet (it's a different group from the one I cleared Ultimate with) and testing could be difficult. Once we get there, it's likely I'll look into those potential enhancements.

Is the styling on the timeline regexes okay? I wanted to be sure all the possible IDs were represented, and chose to minimize line length, but would it be better to instead do groups per single ID, as `(?:4C3C|4C3D|4C3E)`? This might take more horizontal space, but would be easier to extract programmatically in future, and it's easier for a human to see at a glance what the IDS are.

ui/raidboss/data/05-shb/raid/e7n.txt
```diff
@@ -4,4 +4,99 @@
...
+528.9 "Light's Course" # sync /:Unforgiven Idolatry:4C40:/
+532.0 "False Twilight" sync /:The Idol Of Darkness:4C59:/
+532.1 "Light's Course" # sync /:Unforgiven Idolatry:4C40:/
+536.0 "Stygian Sword" sync /:The Idol Of Darkness:4C55:/ window 30,30
+541.0 "Silver Sledge" sync /:The Idol Of Darkness:4C54:/
+
+550.0 "Unjoined Aspect" sync /:The Idol Of Darkness:4C3B:/
+555.2 "Betwixt Worlds" sync /:The Idol Of Darkness:4CF8:/
+560.2 "Words of Night" sync /:The Idol Of Darkness:4C2C:/ window 30,30
+572.4 "Dark's Course/Light's Course" # sync /:Unforgiven Idolatry:4C(?:3[CDE]|4[013]):/
```
quisquous: This seems fine! It would be very wordy to write all 6 of these out.

PR #1148, opened 2020-03-03
Raidboss: E7n minor updates

JLGarber:
As it turns out, counting is hard. The color triggers will no longer fail to warn the user on the third application of darkness/light. The color tracking will also clean up after itself now. (I deliberately added the ability name to the cleanup trigger so it was clear what ability was being used without doing a lookup.) Also a small but significant correction to re-targeting time after the intermission.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e7n.js
+++ b/ui/raidboss/data/05-shb/raid/e7n.js
@@ -130,13 +130,13 @@
       },
     },
     {
-      //Safety in case the user dies during Dark/Light Course.
+      // Safety in case the user dies during Dark/Light Course.
       id: 'E7N Color Cleanup',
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/raid/e7n.js
```diff
@@ -102,7 +102,7 @@
       condition: Conditions.targetIsYou(),
       suppressSeconds: 3,
       infoText: function(data) {
-        data.colorCount = data.colorCount + 1 || 1;
+        data.colorCount = data.colorCount + 1 || 0;
```
quisquous: Belatedly, I don't think I realized the NaN shenanigans (sheNaNigans??) going on here before to make this statement work out properly when data.colorCount is undefined.

JLGarber: Oof. I suppose I should have separated the counter from the definition, you're right. I'll fix that if/when there's other stuff I have to touch in here.

PR #1150, opened 2020-03-03
Update Chinese translation for e8s and responses.js

EBlaster:


quisquous:
Thanks! Just as a recommendation, you might want to create a separate git branch (e.g. `git checkout -tb e8s_translation`) to do your work in rather than doing it all on master.  If you merge into master (like there are twice) then you accumulate very large diffs against upstream.  One way to fix this is to `git reset upstream master`, or whatever your origin is called for github/quisquous/cactbot, to reset history.)

PR #1165, opened 2020-03-06
raidboss: e8s mirror 4 uptime knockback callout

Legends0:
- Players can dodge both knockbacks if action is used during boss' turn to attack target.
Example: https://clips.twitch.tv/CreativeDreamyAsparagusKlappa
Log with earliest: https://www.fflogs.com/reports/rYgBcTKLkNqHdPmF#fight=9
Log with latest: https://www.fflogs.com/reports/rYgBcTKLkNqHdPmF#fight=7

In this situation, tanks have to adjust for mirrors everyone else can stay put. There could be callouts for the tanks to adjust to mirror, but the timing for all this is very quick.

I suspect the latest possible reaction time is 1.4s, but the best log I have is a 1.337s.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e8s.js
+++ b/ui/raidboss/data/05-shb/raid/e8s.js
@@ -352,7 +352,7 @@
       regexCn: Regexes.startsUsing({ source: '希瓦', id: '4D77', capture: false }),
       regexKo: Regexes.startsUsing({ source: '시바', id: '4D77', capture: false }),
       // This gives a warning within 1.4 seconds, so you can hit arm's length.
-      delaySeconds: 3.6,
+      delaySeconds: 8.6,
       durationSeconds: 1.4,
       response: Responses.knockback(),
     },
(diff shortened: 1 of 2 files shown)


quisquous:
I've seen this, but it looks really tight.

I was wondering about having a time on the timeline vs a trigger, since it seems extremely precise.  In practice does the trigger give you enough reaction time to hit it in time?

Legends0:
The video did make it look fairly precise to pull off. Originally I thought it was a 0.5s window to pull off. However, after examining logs, the window is appears much larger. I have seen videos of players manually calling out the time for it as well. I will have to test it out in game to see how it feels as I have been going just by visual queue so far. From my experience, the visual queue needed seems to vary due to game latency.

There are some drawbacks to this callout that I am aware of:
It is expected that those unaware of the strategy or this callout's purpose will have a longer reaction time. It may take a miss to realize that the callout requires faster reaction.
The callout deviates from the behavior of other "knockback" callouts in that it is not ~5 seconds prior to the knockback and the callout is for two knockbacks rather than one.
Furthermore, missing the duration window will lead to knockback on first mirror. Which will likely lead to multiple player deaths. Possibly also issues reporting that we are missing a callout for the first mirror, but a later realization that it is actually for both.

quisquous:
Yeah, I agree with it throwing people off.  People doing the normal strategy may get really thrown.

What do you think about putting it in the timeline? I know that's a little bit unusual (not really any other examples of that), but it's kind of a low key way of indicating it.

If you test the trigger and it works fine, one other option is to add an option for this trigger, and not have it on by default.  Kind of like the cactbot wormhole strategy triggers.  (Or I could think about having off-by-default triggers as a general category of things too.)

Legends0:
In testing, current times are off. They happen too early.

Legends0:
The timing is good on this. Will add optional setting to it to turn on/off.

quisquous:
Thanks for adding the option, as well!

Issue #1161, opened 2020-03-05
E7S trigger errors


nebularg:
## Description

`E7S Betwixt Worlds Stack` (looks like a c/p error with the var used) and `E7S Silver Sledge` were spamming the log with errors (progression up to the add phase, so lots of p1 stuff)

## Additional information

Snippet from the log

```
[3/3/2020 8:00:04 PM] Info: InitPlugin: PluginDirectory = C:\Games\Advanced Combat Tracker\Plugins\OverlayPlugin
[3/3/2020 8:00:05 PM] Info: InitPlugin: Initialized.
[3/3/2020 8:00:11 PM] Info: cactbot: 0.16.6.0 C:\Games\Advanced Combat Tracker\Plugins\cactbot\CactbotOverlay.dll (dir: C:\Games\Advanced Combat Tracker\Plugins\cactbot)
[3/3/2020 8:00:11 PM] Info: OverlayPlugin: 0.13.5.0 C:\Games\Advanced Combat Tracker\Plugins\OverlayPlugin\OverlayPlugin.dll
[3/3/2020 8:00:11 PM] Info: FFXIV Plugin: 2.0.5.5 C:\Games\Advanced Combat Tracker\Plugins\FFXIV_ACT_Plugin\FFXIV_ACT_Plugin.dll
[3/3/2020 8:00:11 PM] Info: ACT: 3.4.7.268 C:\Games\Advanced Combat Tracker\Advanced Combat Tracker.exe
[3/3/2020 8:00:11 PM] Info: Language: en
[3/3/2020 8:00:11 PM] Info: Version: intl
[3/3/2020 8:00:11 PM] Info: LoadAddons: CactbotOverlay.dll: Initialized
[3/3/2020 8:00:14 PM] Info: Found enmity memory for 5.2.
...
[3/3/2020 10:59:10 PM] Info: raidboss: BrowserConsole: Error in trigger: E7S Betwixt Worlds Stack (05-shb/raid/e7s.js) (Source: file:///C:/Games/Advanced%20Combat%20Tracker/Plugins/cactbot/ui/raidboss/popup-text.js, Line: 16)
[3/3/2020 10:59:10 PM] Info: raidboss: BrowserConsole: TypeError: Cannot read property 'indexOf' of undefined (Source: file:///C:/Games/Advanced%20Combat%20Tracker/Plugins/cactbot/ui/raidboss/popup-text.js, Line: 20)
[...]

PR #1166, opened 2020-03-06
raidboss: Fix E7S Betwixt Worlds Tether Condition

panicstevenson:
The condition only runs if you're the target, leaving the tether list to
be undefined at Betwixt Worlds Stack markers.

Fixes https://github.com/quisquous/cactbot/issues/1161


quisquous:
Thanks for the fix!!

PR #1170, opened 2020-03-06
Raidboss: The Vault--knights should no longer trigger pre-battle

### LINKED ISSUE #1167: Issues with triggers in last encounter of The Vault
## Description

The last encounter of `The Vault` against `Ser Charibert` has numerous issues with the fight-specific triggers.

## Additional information

The `Evade marching knights` trigger fired as soon as the last trash pack before the boss was killed, before we even left that tower room. At least one other fight-specific trigger fired at the wrong time though I didn't manage to catch what the text was.

The generic AoE and tankbuster triggers seemed to be fine.

The actual fight timeline also seemed to be fine.

JLGarber:
This should resolve #1167. It's a *bit* of a hack, but oh well.

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/the_vault.js
+++ b/ui/raidboss/data/03-hw/dungeon/the_vault.js
@@ -138,6 +138,16 @@
       response: Responses.breakChains(),
     },
     {
+      id: 'The Vault Knights Activation',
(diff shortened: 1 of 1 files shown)


quisquous:
(Your commit messages always make me laugh!)

PR #1173, opened 2020-03-07
- Add async support for triggers

valarnin:
This PR adds support for async triggers. A trigger can define a `promise` property which should be a function which returns a `Promise`. This is in response to @quisquous comment here: <https://github.com/ngld/OverlayPlugin/pull/116#issuecomment-593735692>

An example trigger might look like this:

```javascript
{
  id: 'Test Trigger',
  regex: Regexes.ability({ id: '1234' }),
    return true;
  },
  promise: function() {
    let p = new Promise((res) => {
      window.setTimeout(() => {
        res({
          name: "Testing",
        });
      }, 1000);
    });
  },
  infoText: function(data, matches) {
    let name = matches.name;
    return {
      en: 'Test Trigger: ' + name,
    };
  },
},
```

With the `window.setTimeout` being replaced by some other async function as needed.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -426,6 +426,11 @@ class PopupText {
     let suppress = 'suppressSeconds' in trigger ? ValueOrFunction(trigger.suppressSeconds) : 0;
     if (trigger.id && suppress > 0)
       this.triggerSuppress[trigger.id] = now + suppress * 1000;
+    let promise = 'promise' in trigger ? promise : () => {
+      return new Promise((res) => {
+        res(matches);
+      });
+    };
 
     // FIXME: this is quite gross that PerTriggerOptions does not use the same fields as
     // options.  Ideally we should smush everything down into a single trigger object.
@@ -665,19 +670,28 @@ class PopupText {
         trigger.run(that.data, matches);
     };
 
-    // Run immediately?
-    if (!delay) {
-      f();
-      return;
-    }
+    // Allow promise to return a new set of matches
+    promise().then((pMatches) => {
+      matches = pMatches;
 
-    this.timers.push(window.setTimeout(() => {
-      try {
+      // Re-apply the groups logic above if we've got a new set of matches
+      if ((matches != undefined) && (matches.groups != undefined))
+        matches = matches.groups;
+
+      // Run immediately?
+      if (!delay) {
         f();
-      } catch (e) {
-        onTriggerException(trigger, e);
+        return;
       }
-    }, delay * 1000));
+
+      this.timers.push(window.setTimeout(() => {
+        try {
+          f();
+        } catch (e) {
+          onTriggerException(trigger, e);
+        }
+      }, delay * 1000));
+    });
   }
 
   Test(zone, log) {
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -426,11 +426,6 @@ class PopupText {
     let suppress = 'suppressSeconds' in trigger ? ValueOrFunction(trigger.suppressSeconds) : 0;
     if (trigger.id && suppress > 0)
       this.triggerSuppress[trigger.id] = now + suppress * 1000;
-    let promise = 'promise' in trigger ? trigger.promise : () => {
-      return new Promise((res) => {
-        res(matches);
-      });
-    };
 
     // FIXME: this is quite gross that PerTriggerOptions does not use the same fields as
     // options.  Ideally we should smush everything down into a single trigger object.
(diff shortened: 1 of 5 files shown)


ui/raidboss/popup-text.js
```diff
@@ -426,6 +426,11 @@ class PopupText {
     let suppress = 'suppressSeconds' in trigger ? ValueOrFunction(trigger.suppressSeconds) : 0;
     if (trigger.id && suppress > 0)
       this.triggerSuppress[trigger.id] = now + suppress * 1000;
+    let promise = 'promise' in trigger ? promise : () => {
```
quisquous: I think you're missing `trigger.`, i.e. `let promise = 'promise' in trigger ? trigger.promise : () => {`

ui/raidboss/popup-text.js
```diff
@@ -426,6 +426,11 @@ class PopupText {
...
+      return new Promise((res) => {
+        res(matches);
```
quisquous: My inclination is to push back on this, because I'm not sure I follow what the use case is for modifying matches.  Can you help me understand?

If you need to return something from the promise, maybe there could be a better way, like calling functions with (data, matches, promiseReturn)?

ui/raidboss/popup-text.js
valarnin: Yes, you're correct. Fixing that now.

quisquous: Also it seems like `trigger.promise` must always be a function that returns a promise.  (This seems reasonable! I don't think it makes sense here to have a raw promise as the value associated with the `promise` key.)  Can you verify that `trigger.promise` is a function and later on that what it returns is an instanceof Promise, and throw a trigger error if not?

valarnin: I assume that the errors in the latest commit are what you're referring to when you say 'trigger error'?

ui/raidboss/popup-text.js
valarnin: Maybe for this case the `trigger.promise` function should accept either the `this.data` attribute or an entirely separate `promiseData` local object. The latter would require modifying the rest of the trigger code inside the local `f` function to account for the new variable, though. If you're disinclined to replacing the `matches` object then probably the former (pass in `this.data` to `trigger.promise`) is the best option.

quisquous: Sorry, it just feels a little inconsistent from the way other trigger functions work.  Right now, passing data between trigger functions is by modifying the data param (which is why ordering is important too).  Modifying matches to return data (and knowing that if you return anything that it will clobber the matches variable) seems a bit fragile too.

Calling `trigger.promise(data, matches)` and having the promise modify `data` (like every other trigger function) to communicate with other trigger functions seems much more consistent/intuitive (even if it's kind of clunky at the same time, admittedly).  And then I guess in that world the promise function just doesn't return anything.

valarnin: Alright, I'll make these adjustments. Not sure when I'll get the chance to add examples/documentation, might be mid-next-week.

valarnin:
Remaining TODO items for this PR:

- [ ] Example trigger in `ui/raidboss/data/00-misc/test.js`
- [ ] Add to list of trigger functions in `resources/responses.js`
- [ ] Add documentation to `ui/raidboss/data/README.md`
- [ ] Add ordering test to `test/trigger/test_trigger.js`

How exactly do I trigger the tests in `test/trigger/test_trigger.js`? Is there an `npm` command for that which I can't find, or some other way?

quisquous:
`npm run test` will run through those tests on every trigger file.

ui/raidboss/popup-text.js
```diff
@@ -665,19 +665,42 @@ class PopupText {
...
 
-    // Run immediately?
-    if (!delay) {
-      f();
-      return;
+    let promise = null;
+
+    if ('promise' in trigger) {
+      if (typeof trigger.promise === 'function') {
+        promise = trigger.promise(data);
```
quisquous: Could you make this `trigger.promise(data, matches)`, so that `promise:` takes the same args as all the other trigger functions?

ui/raidboss/popup-text.js
```diff
@@ -665,19 +665,42 @@ class PopupText {
...
+      } else {
+        console.error('Trigger ' + trigger.id + ': promise defined but not a function');
+      }
     }
 
-    this.timers.push(window.setTimeout(() => {
-      try {
+    if (promise === null) {
+      promise = new Promise((res) => {
+        res(matches);
```
quisquous: This could just be `res()` and not `res(matches)`, right?

quisquous:
Thanks for all the extra documentation and testing.  Looks good in general, just two more small comments.

ui/raidboss/popup-text.js
valarnin: Added `matches` to `test.js` and `README.md` with latest commit.

ui/raidboss/popup-text.js
valarnin: Removed `matches` from `popup-text.js` with latest commit.

valarnin:
I'll be testing my e6s trigger that uses this functionality tonight, if it works as expected do you want me to add it to this PR, or submit a separate PR for it?

quisquous:
> I'll be testing my e6s trigger that uses this functionality tonight, if it works as expected do you want me to add it to this PR, or submit a separate PR for it?

Definitely separate, thanks!

PR #1177, opened 2020-03-08
update e8s kor line, bug fix

Una-P:


quisquous:
Thanks for the fix! Sorry about that!

PR #1178, opened 2020-03-08
Oopsy: E5N initial

JLGarber:
Nowhere near as many mistakes  exist as I had expected. I'm not 100% sure on the setup for Stormclouds that I used, but I'm (I hope understandably) reluctant to test that one deliberately in the duty finder, and I haven't seen that mistake today.

Known deficiencies:

1. Standing in the center for the Deadly Discharge knockback from Will of Ramuh. I don't know what the ability that goes off here is called when it hits a player.

2. I'm not sure whether we had figured out the "fell off platform" detection yet. If so, it would be possible to improve the knockback mechanics with triggers to check whether a player was pushed off.

### DIFF
--- a/ui/oopsyraidsy/data/05-shb/raid/e5n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e5n.js
@@ -22,7 +22,7 @@
       id: 'E5N Static Condensation',
       gainsEffectRegex: gLang.kEffect.StaticCondensation,
       mistake: function(e) {
-        return { type: 'warn', blame: e.targetName, text: e.abilityName };
+        return { type: 'warn', blame: e.targetName, text: e.effectName };
       },
     },
     {
@@ -42,7 +42,7 @@
         return !data.hasOrb;
       },
       mistake: function(e) {
-        return { type: 'fail', blame: e.targetName, text: e.abilityName };
+        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' (no orb)' };
       },
     },
     {
@@ -59,11 +59,11 @@
       damageRegex: '4B9D',
       suppressSeconds: 30,
       mistake: function(e, data) {
-        for (let m in data.cloudMarkers) {
+        for (let m of data.cloudMarkers) {
           return {
             type: 'fail',
             blame: data.cloudMarkers[m],
-            text: e.abilityName,
+            text: e.abilityName + '(clouds too close)',
           };
         }
       },
--- a/ui/oopsyraidsy/data/05-shb/raid/e5n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e5n.js
@@ -6,9 +6,74 @@
     ko: /^희망의 낙원 에덴: 공명편 \(1\)$/,
   },
   damageWarn: {
+    '4B8F': 'Judgment Jolt', // Stratospear explosions
+    '4E3A': 'Impact', // Stratospear landing AoE
+    '4B9C': 'Lightning Bolt', // Stormcloud standard attack
+    '4B97': 'Gallop', // Sideways add charge
+    '4BA1': 'Shock Strike', // Small AoE circles during Thunderstorm
+    '4CF2': 'Volt Strike', // Large AoE circles during Thunderstorm
+
   },
   damageFail: {
   },
   triggers: [
+    {
+      // This happens when a player gets 4+ stacks of orbs. Don't be greedy!
+      id: 'E5N Static Condensation',
+      gainsEffectRegex: gLang.kEffect.StaticCondensation,
+      mistake: function(e) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
+      },
+    },
+    {
+      // Helper for orb pickup failures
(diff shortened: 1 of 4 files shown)


ui/oopsyraidsy/data/05-shb/raid/e5n.js
```diff
@@ -6,9 +6,75 @@
     ko: /^희망의 낙원 에덴: 공명편 \(1\)$/,
   },
   damageWarn: {
+    '4B8F': 'Judgment Jolt', // Stratospear explosions
+    '4E3A': 'Impact', // Stratospear landing AoE
+    '4B9C': 'Lightning Bolt', // Stormcloud standard attack
+    '4B97': 'Gallop', // Sideways add charge
+    '4BA1': 'Shock Strike', // Small AoE circles during Thunderstorm
+    '4CF2': 'Volt Strike', // Large AoE circles during Thunderstorm
```
quisquous: Can you remind me if this is the same as the volt strike in e5s, where everybody gets a circle and always takes this damage?

ui/oopsyraidsy/data/05-shb/raid/e5n.js
```diff
@@ -6,9 +6,75 @@
...
   },
   damageFail: {
   },
   triggers: [
+    {
+      // This happens when a player gets 4+ stacks of orbs. Don't be greedy!
+      id: 'E5N Static Condensation',
+      gainsEffectRegex: gLang.kEffect.StaticCondensation,
+      mistake: function(e) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
```
quisquous: e.effectName?

ui/oopsyraidsy/data/05-shb/raid/e5n.js
```diff
@@ -6,9 +6,75 @@
...
+      },
+    },
+    {
+      id: 'E5N Divine Judgement Volts',
+      damageRegex: '4B9A',
+      condition: function(e, data) {
+        return !data.hasOrb;
+      },
+      mistake: function(e) {
+        return { type: 'fail', blame: e.targetName, text: e.abilityName };
```
quisquous: Do you think it'd be worth adding text more than the ability name here? like `e.abilityName + ' (no orb)'`.  Ability names on their own are usually "you got hit by something you shouldn't and everybody (technically) gets hit by this.

ui/oopsyraidsy/data/05-shb/raid/e5n.js
```diff
@@ -6,9 +6,75 @@
...
+        data.cloudMarkers.push(e.targetName);
+      },
+    },
+    {
+      // This ability is seen only if players stacked the clounds instead of spreading them.
+      id: 'E5N The Parting Clouds',
+      damageRegex: '4B9D',
+      suppressSeconds: 30,
+      mistake: function(e, data) {
+        for (let m in data.cloudMarkers) {
```
quisquous: `let m of data.cloudMarkers`?

ui/oopsyraidsy/data/05-shb/raid/e5n.js
```diff
@@ -6,9 +6,75 @@
...
+          return {
+            type: 'fail',
+            blame: data.cloudMarkers[m],
+            text: e.abilityName,
```
quisquous: Maybe this could say clouds too close directly?

ui/oopsyraidsy/data/05-shb/raid/e5n.js
JLGarber: It is not. That one doesn't appear on normal.

ui/oopsyraidsy/data/05-shb/raid/e5n.js
JLGarber: That's a fair point. I'm not happy about introducing stuff like that since it requires manual translation, but you're right that otherwise it's not clear on its own.

ui/oopsyraidsy/data/05-shb/raid/e5n.js
JLGarber: Same concern as before. Would the previous construction work here too? `e.abilityName + '(clouds too close)'`. It's six words long, but for stuff like this it might be acceptable?

ui/oopsyraidsy/data/05-shb/raid/e5n.js
JLGarber: I should also clarify, I'm not attached to this method of handling the Stormcloud mistake, this was just the first thing that came to mind. I would be much happier with something that doesn't involve looping over stored data, if that's at all possible.

quisquous: I tink this is fine as an approach.  I think it would be very tricky to figure out who dropped which clouds and then blame just those people, so blaming everybody dropping clouds seems fine.

ui/oopsyraidsy/data/05-shb/raid/e5n.js
quisquous: Sounds good to me.

PR #1179, opened 2020-03-08
Oopsy: E6N initial

JLGarber:
Standard stuff for the most part.

Known deficiencies:

1. Storm of Fury tethers. It's going to be a bit of work to figure out overlap tracking for this one, since checking for double-taps won't work. (It's not a mistake for multiple people to be hit by an individual cone.) Currently I'm thinking of something like pre-collecting the tether targets, then collecting all cone targets, seeing whether there are duplicates, and THEN seeing whether the tether targets are *also* in that list of duplicates. A lot of code for one thing, and there's gotta be a better way.

2. No handling for Hands of Hell. I'm not really sure what we want to do with it, so I left it alone for now.

3. Ferostorm IDs might be incomplete? I'm not sure. I'm wondering whether the actual cones have only one ID, and then the different directions are two different casts. Not sure.

### DIFF
--- a/ui/oopsyraidsy/data/05-shb/raid/e6n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e6n.js
@@ -31,9 +31,9 @@
         return e.type != '15';
       },
       mistake: function(e) {
-        // This is a failure because it will kill non-tanks
+        // Kills non-tanks who get hit by it.
         return { type: 'fail', blame: e.targetName, text: e.abilityName };
       },
-    }
+    },
   ],
 }];
--- a/ui/oopsyraidsy/data/05-shb/raid/e6n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e6n.js
@@ -6,9 +6,34 @@
     ko: /^희망의 낙원 에덴: 공명편 \(2\)$/,
   },
   damageWarn: {
+    '4BDA': 'Thorns', // AoE markers after Enumeration
+    '4BDD': 'Ferostorm 1',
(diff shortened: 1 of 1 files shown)


quisquous:
Is it not a mistake to stack the tethers? I guess I don't remember what the normal mode mechanic is.  Do you just take some light damage if you are in them? There's no like magic vuln, right?

Hands of Hell, I'm not sure either.

Re: Ferostorm. I haven't looked at ids, but a common pattern is two different ability ids, but one damage ability id if anybody gets hit.  So, that wouldn't surprise me.

JLGarber:
Storm of Fury is a bit weird. It's like Earth Shakers in the sense that you don't want to stack them, but because the boss is not in a pre-determined location, and because the cones are so wide, I don't think it's fair to assume that any given person is at fault for being hit *once*. (This is particularly true during Raktapaksa, when the Eruptions that drop at the same time can drastically restrict options for dodging. Also, if you think about Savage for a moment, the tethers do come up at one point at the same time as the wind orbs + knockback, and it's literally impossible for the tanks/healers to avoid getting hit.)

However, it's very definitely a mistake if either tether person get hit by the other one, or really if anyone gets hit by both. That doesn't take much effort to avoid regardless of the phase.

On Ferostorm, you're right, now that I think about it that's exactly the behavior of other multi-directional cone attacks like Double Sever in Fractal Continuum.

quisquous:
Ok, maybe this is a good pass for now, and if you or anybody else wants to follow up to be more fancy, then so be it.  :)

PR #1180, opened 2020-03-09
E6S Raktapaksa ferostorm

mooondark:


ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -23,10 +23,15 @@
     },
     {
       id: 'E6S Ferostorm',
-      regex: Regexes.startsUsing({ source: 'Garuda', id: '4BF[EF]', capture: false }),
+      regex: Regexes.startsUsing({ source: ['Garuda', 'Raktapaksa'], id: ['4BF[EF]', '4C0[45]'], capture: false }),
+      regexDe: Regexes.startsUsing({ source: ['Garuda', 'Raktapaksa'], id: ['4BF[EF]', '4C0[45]'], capture: false }),
+      regexFr: Regexes.startsUsing({ source: ['Garuda', 'Raktapaksa'], id: ['4BF[EF]', '4C0[45]'], capture: false }),
+      regexJa: Regexes.startsUsing({ source: ['ガルーダ', 'ラクタパクシャ'], id: ['4BF[EF]', '4C0[45]'], capture: false }),
+      regexCn: Regexes.startsUsing({ source: '迦楼罗', id: '4BF[EF]', capture: false }),
```
quisquous: Can you update the Cn and Ko versions as well?

mooondark: I wish I can but I don't know what is the translation for Raktapaksa in these 2 languages...

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -23,10 +23,15 @@
...
+      regexKo: Regexes.startsUsing({ source: '가루다', id: '4BF[EF]', capture: false }),
```
Una-P: Korea is still 5.0 so we don't know the official translation yet, literally translation is 락타팍사.
```suggestion
      regexKo: Regexes.startsUsing({ source: ['가루다', '락타팍사'], id: '4BF[EF]', capture: false }),
```

ui/raidboss/data/05-shb/raid/e6s.js
quisquous: Could you remove these lines then? I think it's much more obvious that this needs a translation when there's no regex at all.  Having a regex there that's half done seems like something that could easily be missed in the future when the name is known.

PR #1181, opened 2020-03-09
Oopsy: E7N Initial

JLGarber:
There aren't a lot of mistakes to be made in this encounter either. (Or, more properly, there are only a LOT of variations of a few categories.)

I'm really not happy with the handling for Dark's/Light's Course. There's gotta be a better way than spending five(!) triggers on two mechanics. However, I'm hesitant to move any of that functionality in-line since we tend to keep things separated elsewhere in the repository. Improvements to this mess would be welcome. (Also, anything we do here "should" be a suitable drop-in for Savage once the ability IDs are switched out.)

Another thing that could use improvement, as I mentioned in-line, is that I don't like the regex soup I used to check for damage events on Course casts. Should I have used the `regex: Regexes.ability()` construction here instead? Not sure how that all interacts with Oopsy. Or did we want literally a trigger per ID?

I'm also planning to improve this by having a check for each player's position when the stun from Away With Thee wears off. Unlike most "can take you off the platform" mechanics, it (seems to me it) should be definite whether the player's teleport left them on the platform or not. (I need to test further, but I think the platform coordinates are 80,80-120,120, with origin at northwest off the platform. Or I could have been totally misreading the log lines when I checked last night. It was late, and Daylight Saving is horrible.) Maybe someday the log lines can tell us "fell off" directly. A human can dream...

### DIFF
--- a/resources/lang/ja.js
+++ b/resources/lang/ja.js
@@ -66,6 +66,8 @@ class CactbotLanguageJa extends CactbotLanguage {
       Throttle: '窒息',
       StaticCondensation: '蓄電',
       DamageDown: 'ダメージ低下',
+      AstralEffect: '偏属性：光',
+      UmbralEffect: '偏属性：闇',
 
       // UWU
       Windburn: '裂傷',
--- a/ui/oopsyraidsy/data/05-shb/raid/e7n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e7n.js
@@ -37,7 +37,7 @@
       // If we don't set this up ahead of time, the Dark/Light Course triggers will fail
       // on the first two uses of Light's Course, since the Astral/Umbral properties wouldn't exist.
       id: 'E7N Debuff Setup',
-      regex: Regexes.startsUsing({ source: 'The Idol Of Darkness', id: '4C2C' }),
+      regex: Regexes.startsUsing({ source: 'The Idol Of Darkness', id: '4C2B' }),
       condition: function(e, data) {
         return !(data.hasAstral && data.hasUmbral);
       },
--- a/ui/oopsyraidsy/data/05-shb/raid/e7n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e7n.js
@@ -66,20 +66,20 @@
       id: 'E7N Light\'s Course',
       damageRegex: ['4C3E', '4C40', '4C22', '4C3C', '4E63'],
       condition: function(e, data) {
-        return data.hasAstral[e.targetName];
+        return !data.hasUmbral[e.targetName];
       },
       mistake: function(e) {
-        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong buff' };
+        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong/no buff' };
       },
     },
     {
       id: 'E7N Darks\'s Course',
       damageRegex: ['4C3D', '4C23', '4C41', '4C43'],
       condition: function(e, data) {
-        return !data.hasAstral[e.targetName];
+        return data.hasUmbral[e.targetName];
       },
       mistake: function(e) {
-        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong/no buff' };
+        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong buff' };
       },
     },
   ],
--- a/ui/oopsyraidsy/data/05-shb/raid/e7n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e7n.js
@@ -69,7 +69,9 @@
         return !data.hasUmbral[e.targetName];
       },
       mistake: function(e) {
-        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong/no buff' };
+        if (data.hasAstral[e.targetName])
+          return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong buff' };
+        return { type: 'warn', blame: e.targetName, text: e.abilityName + ' no buff' };
       },
     },
     {
--- a/ui/oopsyraidsy/data/05-shb/raid/e7n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e7n.js
@@ -65,13 +65,13 @@
     {
       id: 'E7N Light\'s Course',
       // Would be nice to use an array instead, as
-      // damageRegex: ['4C3E', '4C40', '4C22', '4C3C', '4E63']
+      damageRegex: ['4C3E', '4C40', '4C22', '4C3C', '4E63'],
       // rather than this awful cup of regex soup.
       // We could instead split this out into individual triggers for each ID,
       // but that seems ridiculous.
-      damageRegex: '(?:4C(?:22|40|3[CE])|4E63)',
+      // damageRegex: '(?:4C(?:22|40|3[CE])|4E63)',
       condition: function(e, data) {
-        return data.hasAstral[targetName];
+        return data.hasAstral[e.targetName];
       },
       mistake: function(e) {
         return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong buff' };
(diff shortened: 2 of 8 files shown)


JLGarber:
Actually. Re-reading the OopsyReadme, would Away With Thee and similar I-fell-off mechanics be appropriate uses for the `deathReason` property? It wouldn't be perfect, of course, but it would be way better than nothing.

ui/oopsyraidsy/data/05-shb/raid/e7n.js
```diff
@@ -6,9 +6,87 @@
...
+      gainsEffectRegex: gLang.kEffect.UmbralEffect,
+      losesEffectRegex: gLang.kEffect.UmbralEffect,
+      run: function(e, data) {
+        data.hasUmbral[e.targetName] = e.gains;
+      },
+    },
+    {
+      id: 'E7N Light\'s Course',
+      // Would be nice to use an array instead, as
+      // damageRegex: ['4C3E', '4C40', '4C22', '4C3C', '4E63']
```
quisquous: This seems like an easy fix.  Could change https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/oopsyraidsy.js#L993 to use `Regexes.Any(trigger.damageRegex)` (as well as probably everything around it?

ui/oopsyraidsy/data/05-shb/raid/e7n.js
```diff
@@ -6,9 +6,87 @@
...
+      // rather than this awful cup of regex soup.
+      // We could instead split this out into individual triggers for each ID,
+      // but that seems ridiculous.
+      damageRegex: '(?:4C(?:22|40|3[CE])|4E63)',
+      condition: function(e, data) {
+        return !data.hasUmbral[e.targetName];
+      },
+      mistake: function(e) {
+        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong/no buff' };
```
quisquous: Is it wrong to take this with no buff? Just wondering if this should only warn if hasAstral on this case, vs !hasUmbral.

resources/lang/ja.js
```diff
@@ -66,6 +66,8 @@ class CactbotLanguageJa extends CactbotLanguage {
       Throttle: '窒息',
       StaticCondensation: '蓄電',
       DamageDown: 'ダメージ低下',
+      AstralEffect: '偏属性：光',
```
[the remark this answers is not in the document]
JLGarber: I would, but neither of the Github pages hosting their CSV files has those debuffs added yet. Is there an easy way to find them from FFLogs without having to run the entire translation script?

ui/oopsyraidsy/data/05-shb/raid/e7n.js
JLGarber: You're right, it's not possible for Dark's Course to be taken without a debuff. Thanks for catching that!

ui/oopsyraidsy/data/05-shb/raid/e7n.js
JLGarber: Hmm. I'm actually not too familiar with the inner workings of `regexes.js`, I just use the exposed functions. From a brief look through it, am I understanding correctly that

```
        if ('damageRegex' in trigger) {
          trigger.idRegex = Regexes.anyOf('^' + trigger.damageRegex + '$');
          this.damageTriggers.push(trigger);
        }
        if ('abilityRegex' in trigger) {
          trigger.idRegex = Regexes.anyOf('^' + trigger.abilityRegex + '$');
          this.abilityTriggers.push(trigger);
        }
etc

```
should Just Work?

resources/lang/ja.js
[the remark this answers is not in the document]
JLGarber: Oh, wait, do you mean "add placeholders"? Sure, I can do that.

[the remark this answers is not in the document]
quisquous: Yeah, add placeholders.

ui/oopsyraidsy/data/05-shb/raid/e7n.js
quisquous: I think it needs to be more along the lines of '^' + Regexes.anyOf(trigger.damageRegex) + '$'.  `anyOf` can take a list.

JLGarber: Ah, that would make more sense. In weeklies right now, but once we get done I'll take another look.

JLGarber:
Okay, from the More Than One times I ran that through, things seem to have triggered correctly. I'm not at all sure the modification I made to OopsyRaidsy.js was  entirely correct/complete, but at least there aren't any console errors. Thanks so much for pointing me to the changes needed there, the new setup is *way* nicer than a random regex string.

ui/oopsyraidsy/data/05-shb/raid/e7n.js
```diff
@@ -6,9 +6,81 @@
...
+      },
+    },
+    {
+      id: 'E7N Light\'s Course',
+      damageRegex: ['4C3E', '4C40', '4C22', '4C3C', '4E63'],
+      condition: function(e, data) {
+        return !data.hasUmbral[e.targetName];
+      },
+      mistake: function(e) {
+        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong/no buff' };
```
quisquous: I'm confused at the `/ no buff` part here?

Also, why are there more ids in light's course vs dark's course? Is one of these the initial uncolored ones, and those should be up in the damageFail category?

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -986,15 +986,15 @@ class DamageTracker {
       for (let j = 0; j < set.triggers.length; ++j) {
         let trigger = set.triggers[j];
         if ('regex' in trigger) {
-          trigger.regex = Regexes.parse(trigger.regex);
+          trigger.regex = Regexes.parse(Regexes.anyOf(trigger.regex));
           this.generalTriggers.push(trigger);
         }
         if ('damageRegex' in trigger) {
-          trigger.idRegex = Regexes.parse('^' + trigger.damageRegex + '$');
+          trigger.idRegex = Regexes.parse('^' + Regexes.anyOf(trigger.damageRegex) + '$');
```
quisquous: Thanks! This is a great addition!!

ui/oopsyraidsy/data/05-shb/raid/e7n.js
JLGarber: The same IDs are used for Light's Course whether or not the players are inflicted with Astral/Umbral, at any phase of the battle. It is *technically* not a failure to be hit during a non-debuff phase, in the sense that the player will not auto-die. What if we did something like this?

```
    {
      id: 'E7N Light\'s Course',
      damageRegex: ['4C3E', '4C40', '4C22', '4C3C', '4E63'],
      condition: function(e, data) {
        return !data.hasUmbral[e.targetName];
      },
      mistake: function(e) {
        if (data.hasAstral[e.targetName])
          return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong buff' };
        return { type: 'warn', blame: e.targetName, text: e.abilityName + ' no buff' };
      },
    },
```

As for the disparate numbers of IDs, I'm not sure, but I have a theory, It *looks* like each Course ability hits two "lanes", while the actual areas hit on the battlefield are dependent on which Unforgiven Idolatry uses which ability. (I base this supposition on the fact that the first two sets of birds for the encounter are always fixed, with the outer lanes being hit before inner with Light's course. Both of these hits, inner and outer, use the `4C3C` cast.) The "extra" ID comes from the unbalanced set of Light's Course casts at the start.

(We could *probably* dig into that more and use it to figure out some useful positioning triggers for Savage? I don't know.)

A quick search at XIVAPI shows 13 different versions of Light's Course compared with 11 for Dark. I do know that across my logs I've only seen the IDs I've put in the oopsy file.

quisquous:
I think ideally somebody tracks down the network data for falling off the edge/ death wall and we could add that universally.

JLGarber:
Fair enough. I've been experimenting with `deathReason` for E8N for the exact same purpose, and it hasn't been going well, so I may just ignore that for now and get that pull request put together.

PR #1186, opened 2020-03-12
Raidboss: HW dungeon timelines--Sync windows

JLGarber:
This is a minor update that's going to be part of a larger overall project.  Looking over my older work, I wasn't at all as complete with sync windows as I should have been, and especially on more recent content I've been noticing how important good sync windows are. I'm hoping to go back over the next week or two and update older timelines to modern standards in that regard.

(Note that these are untested. I don't have any recent logs to test against, but purely from eyeballing the timelines things look okay. I understand if it's preferable that I hold on this one until I can get logs, but it could be a while.)

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.txt
+++ b/ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.txt
@@ -90,7 +90,7 @@ hideall "--sync--"
 392.7 "Bastardbluss" sync /:Regula van Hydrus:10DA:/
 399.8 "Judgment" sync /:Regula van Hydrus:10DD:/
 405.0 "Judgment" sync /:Regula van Hydrus:10DE:/
-414.9 "Magitek Spread" sync /:Regula van Hydrus:10DC:/ 15,15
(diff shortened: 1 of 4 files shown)


ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.txt
```diff
@@ -44,9 +44,9 @@ hideall "--sync--"
 193.0 "Magitek Slug" sync /:Regula van Hydrus:10DB:/
 
 # Rotation resumes here whether or not turrets are successful
-199.3 "Bastardbluss" sync /:Regula van Hydrus:10DA:/ window 100,5
+199.3 "Bastardbluss" sync /:Regula van Hydrus:10DA:/ window 100,2
```
quisquous: I think this is fine but part of me says don't add syncs for things this close, just to avoid weird timline resyncing.  5s seems safe, but 3 feels iffy.

JLGarber: I suppose you're right. Papering over old work for even older content will do that. I'll be very careful to watch for that if I ever do the ARR content, since it will probably come up there as well. Thanks for the heads up!

PR #1190, opened 2020-03-13
Fix issue with async trigger promise call not referencing this

valarnin:
This PR is to fix a bug found with the new async trigger functionality while testing my e6s football trigger. The bug only happens if a trigger has a `promise` function.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -669,7 +669,7 @@ class PopupText {
 
     if ('promise' in trigger) {
       if (typeof trigger.promise === 'function') {
-        promise = trigger.promise(data, matches);
+        promise = trigger.promise(this.data, matches);
         // Make sure we actually get a Promise back from the function
         if (Promise.resolve(promise) !== promise) {
           console.error('Trigger ' + trigger.id + ': promise function did not return a promise');


ui/raidboss/popup-text.js
```diff
@@ -669,7 +669,7 @@ class PopupText {
 
     if ('promise' in trigger) {
       if (typeof trigger.promise === 'function') {
-        promise = trigger.promise(data, matches);
+        promise = trigger.promise(this.data, matches);
```
quisquous: I think the function this is in is bound weirdly and I never fixed it.  I think this may need to be s/this/that/?

valarnin: The scoping of this statement is the same as the scoping of the `let that = this;` statement above, so I'm not sure there would be a difference here? If the scoping does need to be adjusted, it would also need to be adjusted for a bunch of other things that are obviously currently working, such as below on line 696 `this.timers.push...`, or above on lines 657 or 503.

quisquous: Oh, ok.  I should have scrolled all the way up.  I think at one point this was rebound in the function and it got changed to be an arrow function.  I'll clean this up after this to avoid future confusion (on my part).

PR #1194, opened 2020-03-14
Oopsyraidsy: Fixing E5n/E6N/E7N

JLGarber:
I'm deeply sorry I didn't catch this before submitting. I had been getting console errors for Oopsy for a while, and I assumed that I had broken something with my install. But no, I had just reversed things.

### DIFF
--- a/ui/oopsyraidsy/data/05-shb/raid/e6n.js
+++ b/ui/oopsyraidsy/data/05-shb/raid/e6n.js
@@ -6,21 +6,21 @@
     ko: /^희망의 낙원 에덴: 공명편 \(2\)$/,
   },
   damageWarn: {
-    '4BDA': 'Thorns', // AoE markers after Enumeration
-    '4BDD': 'Ferostorm 1',
-    '4BE0': 'Storm of Fury 1', // Circle AoE during tethers--Garuda
-    '4BE2': 'Explosion', // AoE circles, Garuda orbs
-    '4BE5': 'Ferostorm 2',
-    '4BE6': 'Storm Of Fury 2', // Circle AoE during tethers--Raktapaksa
-    '4BEC': 'Heat Burst',
-    '4BEE': 'Conflag Strike', // 270-degree frontal AoE
-    '4BF0': 'Spike Of Flame', // Orb explosions after Strike Spark
-    '4BF2': 'Radiant Plume',
-    '4BF4': 'Eruption',
+    'Thorns': '4BDA', // AoE markers after Enumeration
+    'Ferostorm 1': '4BDD',
+    'Storm of Fury 1': '4BE0', // Circle AoE during tethers--Garuda
+    'Explosion': '4BE2', // AoE circles, Garuda orbs
+    'Ferostorm 2': '4BE5',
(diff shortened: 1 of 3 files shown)


ui/oopsyraidsy/data/05-shb/raid/e6n.js
```diff
@@ -6,21 +6,21 @@
...
+    'Conflag Strike': '4BEE', // 270-degree frontal AoE
+    'Spike Of Flame': '4BF0', // Orb explosions after Strike Spark
+    'Radiant Plume': '4BF2',
+    'Eruption': '4BF4',
   },
   damageFail: {
-    '4BD5': 'Vacuum Slice', // Dark line AoE from Garuda
-    '4BDB': 'Downburst', // Blue knockback circle. Actual knockback is unknown ability 4C20
+    'Vacuum Slice': '4BD5', // Dark line AoE from Garuda
+    'Downburst': '4BDB', // Blue knockback circle. Actual knockback is unknown ability 4C20
```
quisquous: Sorry for not catching this either!

ui/oopsyraidsy/data/05-shb/raid/e7n.js
```diff
@@ -66,22 +66,20 @@
...
     {
       id: 'E7N Darks\'s Course',
       damageRegex: ['4C3D', '4C23', '4C41', '4C43'],
       condition: function(e, data) {
-        return data.hasUmbral[e.targetName];
+        return !data.hasAstral[e.targetName];
       },
       mistake: function(e) {
-        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong buff' };
+        return { type: 'fail', blame: e.targetName, text: e.abilityName + ' wrong/no buff' };
```
quisquous: I'm not sure I understand all the changes here? What was wrong with how it was before?

JLGarber: Ack. I think I pulled this from a development branch instead of Master. I'm so sorry.

ui/oopsyraidsy/data/05-shb/raid/e6n.js
JLGarber: How difficult would it be to lint for this? I'm not terribly good at writing rules myself, but maybe check whether the (presumed) IDs have a length of 4?

quisquous: I think you'd need to check for "are these hex characters".  Not everything has a length of 4, I say, literally in the middle of writing an a1s timeline.

Unfortunately, because of the weird way gLang is used in oopsy triggers, it's tricky to add unit tests.  I was going to move all the effects into some autogenerated `resources/effects.js` and that might solve it.

JLGarber: Fair point, I forgot that older content is still a thing. I don't *expect* to run into this issue again myself, now that I've had it jump out at me once, but I've said the same thing about tripping over my own feet. Gotta be extra careful for now, I suppose!

PR #1195, opened 2020-03-14
Oopsyraidsy: E8N initial

JLGarber:
This one is mostly complete. I've marked it as a draft due to the triggers not working, and I'm not sure what I'm misunderstanding. To begin with, both triggers that include `damageReason` cause this console error:

`[3/14/2020 2:03:29 AM] Info: Oops: BrowserConsole: Uncaught TypeError: this.Translate is not a function (Source: file:///[PATH TO]/oopsyraidsy.js, Line: 887)`

([Oopsyraidsy.js line 887](https://github.com/quisquous/cactbot/blob/bc44de9c9786554fb24b9f3b91ee75c2be37481c/ui/oopsyraidsy/oopsyraidsy.js#L887))

I think it might need to be written as `ret.reason = this.collector.Translate(ret.reason)`? I'm not sure.

The other thing currently non-functional is the Shining Armor trigger. I'm confident the `E` flag indicates failure on that mechanic, but am I misunderstanding how to use the flag? I know it's hex, but my assumption from the OopsyReadme was that it's string-formatted.

Other than that, this should be ready to go.


quisquous:
> To begin with, both triggers that include `damageReason` cause this console error:
> 
> `[3/14/2020 2:03:29 AM] Info: Oops: BrowserConsole: Uncaught TypeError: this.Translate is not a function (Source: file:///[PATH TO]/oopsyraidsy.js, Line: 887)`
> 
> ([Oopsyraidsy.js line 887](https://github.com/quisquous/cactbot/blob/bc44de9c9786554fb24b9f3b91ee75c2be37481c/ui/oopsyraidsy/oopsyraidsy.js#L887))
> 
> I think it might need to be written as `ret.reason = this.collector.Translate(ret.reason)`? I'm not sure.

Possibly! I'll take a look, thanks.

> The other thing currently non-functional is the Shining Armor trigger. I'm confident the `E` flag indicates failure on that mechanic, but am I misunderstanding how to use the flag? I know it's hex, but my assumption from the OopsyReadme was that it's string-formatted.

It's surprising to me that it would *just* be `E`.  Can you paste some log examples just for my own curiosity?

JLGarber:
Sure, a successful execution is:
`16:40004C19:Shiva:4DF1:Shining Armor:[PLAYER ID]:[PLAYER]:1:0:1B:8000:0:0:0:0:0:0:0:0:0:0:0:0:[ETC]`

A failed execution is:
`16:40004C19:Shiva:4DF1:Shining Armor:[PLAYER ID]:[PLAYER]:E:950000:1B:4DF18000:0:0:0:0:0:0:0:0:0:0:0:0:[ETC]`

quisquous:
> Sure, a successful execution is:
> `16:40004C19:Shiva:4DF1:Shining Armor:[PLAYER ID]:[PLAYER]:1:0:1B:8000:0:0:0:0:0:0:0:0:0:0:0:0:[ETC]`
> 
> A failed execution is:
> `16:40004C19:Shiva:4DF1:Shining Armor:[PLAYER ID]:[PLAYER]:E:950000:1B:4DF18000:0:0:0:0:0:0:0:0:0:0:0:0:[ETC]`

Ah, so it is! I think it's that the flags are 4 bytes of hex and leading zeroes are dropped.

However, one thought I've had is that flags are not the most future proof.  What about adding gaining the stun effect instead as an indication of failure?

JLGarber:
Good point on the flags, I hadn't considered that at all.

Using Stun could be a little complex. We could just do this:

```
    {
      id: 'E8N Shining Armor',
      gainsEffectRegex: gLang.kEffect.Stun,
      mistake: function(e) {
        return { type: 'warn', name: e.targetName, reason: e.effectName };
      }
    },
```
And it would be correct, since as you note the only stun present in this encounter is from Shining Armor. However, this would report just a bare Stun, not the actual "looked at the light" mistake. Maybe add Shining Armor to the translation tables as well and report that with the `gLang` construction? Something like:

```
      mistake: function(e) {
        return { type: 'warn', name: e.targetName, reason: gLang.kAbility.ShiningArmor };
      }
```
Is that even supported?

quisquous:
gLang.kAbility exists, but they are just mappings of English names to ids.  The reason here is the ability name string, so I don't think that works.

I think saying Stun on its own would be fine, personally? (It doesn't appear to exist in kEffect though.  I guess one more reason to move everything into some autogenerated resources/effect.js file at some point soon!)

JLGarber:
Oof. I really need to be more careful. Sorry about the extra noise there!

ui/oopsyraidsy/data/05-shb/raid/e8n.js
```diff
@@ -6,9 +6,57 @@
...
+      gainsEffectRegex: gLang.kEffect.Stun,
+      mistake: function(e) {
+        return { type: 'warn', blame: e.targetName, reason: gLang.kEffect.Stun };
+      },
+    },
+    {
+      id: 'E8N Heavenly Strike',
+      damageRegex: '4DD8',
+      deathReason: function(e) {
+        return { type: 'fail', name: e.targetName, reason: 'Pushed off!' };
```
quisquous: Can you make these reasons a `reason: { en: 'Pushed off!' }` style object to make it easier for other people to add translations (and for `find_missing_translations.py` to not miss it)?

JLGarber: I'll do that. I'm currently waiting on a queue to test a change I made to the Frost Armor mistake that might cut it down to one trigger. (I forgot that Frost Armor inflicts Thin Ice on all players, so that would be a MUCH cleaner way to handle this.)

resources/lang/en.js
```diff
@@ -50,6 +50,8 @@ class CactbotLanguageEn extends CactbotLanguage {
       Shifu: 'Shifu',
       CircleOfPower: 'Circle Of Power',
 
+      SurgeProtection: 'Surge Protection',
```
quisquous: Sorry, one last thing, but can you remove this?

PR #1196, opened 2020-03-14
fix raidboss CN zoneRegex,update oopsyraidsy E2S mistake CN translation

DieMoe233:


kyan001:
I already submitted a fix for all the raids CN zoneRegex problem. Can you move the other updates to another commit?

Issue #1193, opened 2020-03-14
incorrect cn zoneregex


DieMoe233:
He committed the wrong localization information to overwrite the correct Chinese localization content.

quisquous:
Hey, I get that you're angry, but the issue title here is uncalled for.  I don't know Chinese, so I trust folks submitting pull requests.

kyan001:
> 
> 
> Hey, I get that you're angry, but the issue title here is uncalled for. I don't know Chinese, so I trust folks submitting pull requests.

I'm sorry for the zoneRegex problem. The root cause of this is that I used the Chinese names in Wiki, but they missed quotes which the ACT actually needed. 

I already fixed it in [PR#1197](https://github.com/quisquous/cactbot/pull/1197)

PR #1197, opened 2020-03-14
Add quotes to raids CN names for all the raids from "02-arr" to "05-shb" (#1193 fix)

kyan001:
This commit only contains the fix of CN raids zoneRegex problem. All 02~05 raids are fixed.


kyan001:
This commit fix the issue #1193

quisquous:
Thanks for the quick fix!

PR #1203, opened 2020-03-15
misc: update CN translation,anti-jamming regex

DieMoe233:


quisquous:
Thanks! If you haven't seen it, I wanted to mention `python util/find_missing_translations.py -l cn` for tracking down missing translations.

PR #1204, opened 2020-03-15
Raidboss/Oopsyraidsy: O1N initial package

JLGarber:
I can actually run this one solo through to enrage now, so I ran through it a couple times and put this together. Since positioning isn't as important on Normal, I didn't bother recording which fireballs are in vs out. (And most of the time it's irrelevant due to Breath Wing anyway.)

The timeline is as long as it is because it's genuinely one single timeline just like most Savage encounters, and there are just enough changes from block to block to make it not worth trying to break down. I've run it through several times and it's 100% deterministic from what I can see.

Yes, the zone name has parentheses in the `01` lines. I was surprised by that myself.


ui/raidboss/data/04-sb/raid/o1n.txt
```diff
@@ -0,0 +1,148 @@
...
+# DELTASCAPE V1.0--O1N #
+#      ALTE ROITE      #
+#~~~~~~~~~~~~~~~~~~~~~~#
+
+# Enrage is approximately 13:40. Yes, Normal has an enrage.
+# -ii 1ED9 1EDA 23D7 23D9 2583
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+0 "Start" sync /:Engage!/ window 0,1
+1.8 "--sync--" sync /:Alte Roite:368:/
```
quisquous: Can you put a window 2,0 on this just to avoid resyncing?

ui/raidboss/data/04-sb/raid/o1n.txt
```diff
@@ -0,0 +1,148 @@
...
+756.6 "Wyrm Tail" sync /:Alte Roite:23D6:/
+762.7 "Wyrm Tail" sync /:Alte Roite:23D6:/
+769.8 "Charybdis" sync /:Alte Roite:23DB:/
+776.9 "Roar" sync /:Alte Roite:23DC:/ window 30,15
+783.0 "Wyrm Tail" sync /:Alte Roite:23D6:/
+790.7 "Twin Bolt x2" # sync /:Alte Roite:23D8:/
+795.2 "Clamp" sync /:Alte Roite:23E2:/
+801.3 "Roar" sync /:Alte Roite:23DC:/
+803.5 "Flame" sync /:Alte Roite:23DD:/ window 50,5
+820.7 "Burn Enrage" sync /:Ball Of Fire:23E4:/
```
quisquous: Yea, this seems reasonable to roll the whole loop out if it's a fixed time and there's a fixed enrage.

ui/raidboss/data/04-sb/raid/o1n.txt
```diff
@@ -0,0 +1,148 @@
+#~~~~~~~~~~~~~~~~~~~~~~#
+# DELTASCAPE V1.0--O1N #
```
quisquous: This is the tiniest of complaints, but maybe make this consistent with other timeline headers? I don't think anything uses ascii art boxes.

quisquous:
Thanks! People definitely get confused and ask where these few missing stormblood normal modes are, so it's really good to fill them in.

Issue #853, opened 2019-12-13
Have raid emulator store fights in indexeddb instead of local storage


quisquous:
Local storage is easier to use but has much lower memory limits.

We should probably store all fights as blobs in indexeddb to make it possible to store more fights.

valarnin:
This is probably irrelevant now :)

PR #1206, opened 2020-03-16
Raidboss: TEA head marker fix

### LINKED ISSUE #1199: Cactbot Limit Cut/Enumeration not working
With the latest ffxiv plugin, cactbot callout for Enumeration and Limit Cut is not working in TEA...

I'm not sure what would be the fix, have tried reloading overlay and plugin...

JLGarber:
This turned out to be significantly simpler than I expected after Quisquous pointed me in the right directions. Note that it is entirely untested as I wasn't able to get the raid emulator working over lunch. However, looking things over there's no reason it shouldn't work. 

Apologies for the ugly bits. I mostly wanted to get concepts set down so we could improve on them and get this out for people as quickly as possible.

(Should resolve #1199 )

### DIFF
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
@@ -34,6 +34,9 @@ let getHeadmarkerId = (data, matches) => {
     // The first 1B marker in the encounter is Limit Cut 1, ID 004F.
     data.decOffset = parseInt(matches.id, 16) - 79;
   }
+  // The leading zeroes are stripped when converting back to string, so we re-add them here.
+  // Fortunately, we don't have to worry about whether or not this is robust,
+  // since we know all the IDs that will be present in the encounter.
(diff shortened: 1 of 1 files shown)


quisquous:
I think this looks good.  Was there anything else you wanted to change before merging? Just ask because it's still marked as a draft.

quisquous:
And, I think the raid emulator can be pretty tetchy about TEA logs too.  They tend to be bigger than fit in local storage unless they're clipped to somewhere in the inception phase.  This is one reason why I want to move raid emulator to use indexeddb instead of local storage (#853).

JLGarber:
Oh, if there's nothing else to review, then it's ready.

Legends0:
I tried this out in the raid tonight... seems all numbers but the first will get a callout. If the player with cactbot gets the one on the first limit cut, cactbot doesn't make a callout for them.

JLGarber:
Interesting. Glad to know that it's mostly working. I don't have any way of testing it myself, so I'm relying on reports like this. I'll open an issue based on this report so we don't clutter this pull request and we can work things out from there?

PR #1207, opened 2020-03-17
Resources: The smallest of additions to en.js

JLGarber:


quisquous:
Could you add this effect to all languages? Oopsy behaves badly when something is undefined.

JLGarber:
Oops! I remembered that it had made its way to DE/FR/JA and didn't even think to check the others.

quisquous:
I really need to add some tests for this or rearrange how it works entirely.  I think the effects spread over multiple files is not really working (or matches the rest of the code).

JLGarber:
It's definitely less comfortable compared to having a single language config file, although I can certainly understand why it was originally set up this way.

That being said, a thing we could do that would require less restructuring is to place a "master elements object" into the EN language file. Something like:

```
masterObject: {
  actionsToTranslate: {
    Action1: 'ID1',
    Action2: 'ID2',
    // ETC
  },
  statusesToTranslate: {
    Status1: 'ID1',
    Status2: 'ID2',
    // ETC
  },
  // ETC
}
```

So we would just add new items to be translated in one place, and then script something to iterate over the object, find the ID on XIVAPI/Wakingsands/Coinach CSV dumps, and update the various language files. Probably cumbersome, but might be less work than restructuring how the core translation functions operate.

PR #1209, opened 2020-03-18
Raidboss: E7N additional timeline syncs

JLGarber:
I've noticed this one seems to be *very* inconsistent on whether or not it starts. I think it's because I initially put down slightly narrower starting windows than I should have. After the changes, it seems much happier. (The post-intermission rotation block is already fairly well covered for syncs, and I've never noticed any issues there.)

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e7n.txt
+++ b/ui/raidboss/data/05-shb/raid/e7n.txt
@@ -13,11 +13,11 @@ hideall "--sync--"
 0.0 "--sync--" sync /:Engage!/ window 0,1
 3.8 "--sync--" sync /:The Idol Of Darkness:4C50:/ window 3,1
 12.8 "Empty Wave" sync /:The Idol Of Darkness:4C52:/ window 12,30
-25.0 "Unshadowed Stake" sync /:The Idol Of Darkness:4C51:/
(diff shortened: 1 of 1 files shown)


quisquous:
Do you have any logs where this is inconsistent, just out of my own curiosity? I'm surprised that anything past the s/12/12.8/ is needed.

JLGarber:
I've had a look and sent you one that I think evidenced this behavior. Looking at it, it does seem that at least that first attack is unreliable at best, so it's 100% necessary to sync Empty Wave. I haven't yet tested with only the 12 --> 12.8 change, but I can do that later tonight or tomorrow.

ui/raidboss/data/05-shb/raid/e7n.txt
```diff
@@ -43,8 +43,8 @@ hideall "--sync--"
 
 # Intermission block
 226.3 "--targetable--"
-236.8 "Away with Thee" sync /:The Idol Of Darkness:(4C39|4E7E):/ window 30,30
-251.7 "Strength in Numbers" sync /:Idolatry:4C4[CD]:/
+236.8 "Away with Thee" sync /:The Idol Of Darkness:(?:4C39|4E7E):/ window 30,30
```
quisquous: Sorry, one last thing, but can you remove the ?: here?

PR #1211, opened 2020-03-18
Raidboss: E2S translation fr minor

VanSlanzar:
Translation fr minor


VanSlanzar:
sorry

quisquous:
No worries at all! I appreciate the contribution enthusiasm, and it's really deceptively non-obvious from just a reading the file that one thing needs a translation and another doesn't.

The best way to find missing translations is to run `python util/find_missing_translations.py -l fr`.

PR #1230, opened 2020-03-19
raidboss: cn old content (5.0)

Bluefissure:
- lot of 5.0 stuff


ui/raidboss/data/05-shb/dungeon/amaurot.js
```diff
@@ -363,9 +363,11 @@
         'Cosmic Shrapnel': '飞散',
         'Cosmic Kiss': '流星坠落',
         'Comet': '彗星',
-        'Burst': '爆炸',
+        'Burst': '大爆炸',
         'Apokalypsis': '启示录',
         'Aetherspike': '以太钉刺',
+        '--sync--': '--sync--',
```
quisquous: Can you remove these, here and elsewhere? In general these are always hidden, and so they shouldn't need to be translated anywhere.  It's not a big deal, but I don't want other people seeing this and then adding it too.

Bluefissure: removed!

Bluefissure:
BTW those who want to take part in the (responses) translation of old content can take a look at https://docs.qq.com/sheet/DSFRGa1V1d2pOZk1H?tab=BB08J2&c=A4A0A0

顺便说一句，那些想参加旧内容的（触发器提示的）翻译的人可以看看https://docs.qq.com/sheet/DSFRGa1V1d2pOZk1H?tab=BB08J2&c=A4A0A0

quisquous:
Ooh, that's interesting! I'll take a look at that project.  If you find things that should be in `resources/responses.js` because it's repeatedly translated, feel free to move anything there!

Bluefissure:
> Ooh, that's interesting! I'll take a look at that project. If you find things that should be in `resources/responses.js` because it's repeatedly translated, feel free to move anything there!

Ah, I know the commonly used responses in `resources/responses.js`, but for every js file they may have specific responses, this is what this sheet is for.

PR #1234, opened 2020-03-19
Docs: Updates to timeline guide

JLGarber:
This update contains three independent elements:

1. Changed the information on sync windows to reflect current usage.
2. A throwaway capitalization change if the word "cactbot" appears as the first word of a sentence. (It doesn't really matter, but I was editing the file, so in it went.)
3. Information on `-lf`. Best to have that documented in this file.

(Maybe we should expand the command line arguments information into its own section, since it cuts across both of the timeline utilities in the guide?)

### DIFF
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -5,7 +5,7 @@ primarily for cactbot.
 
 ![import screenshot](images/timelineguide_timeline.png)
 
-Cactbot uses the [raidboss module](https://github.com/quisquous/cactbot#raidboss-module)
+cactbot uses the [raidboss module](https://github.com/quisquous/cactbot#raidboss-module)
 for triggers and timelines.
 These are combined together so that you can make triggers that are based on actions
 or triggers that are based on timelines themselves.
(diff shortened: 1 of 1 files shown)


docs/TimelineGuide.md
```diff
@@ -5,7 +5,7 @@ primarily for cactbot.
 
 ![import screenshot](images/timelineguide_timeline.png)
 
-cactbot uses the [raidboss module](https://github.com/quisquous/cactbot#raidboss-module)
+Cactbot uses the [raidboss module](https://github.com/quisquous/cactbot#raidboss-module)
```
quisquous: This is a weird request, but can you leave this as lowercase everywhere to be consistent with everything else?

docs/TimelineGuide.md
```diff
@@ -211,10 +213,11 @@ This has two purposes.
 The first purpose is for tools, to autogenerate regular expression translations for triggers.
 
 The second purpose is for timelines at runtime.
-cactbot will use the `replaceSync` section to auto-replace anything inside a `sync /text`/ on a timeline line and the `replaceText` section to auto-replace anything inside the ability text.
+Cactbot will use the `replaceSync` section to auto-replace anything inside a `sync /text`/ on a timeline line,
+and the `replaceText` section to auto-replace anything inside the ability text.
 
-These do not match the entire line by default.
-So, some care is needed to make sure that replacements are not overzealous.
+These do not match the entire line (that is, they are non-greedy) by default.
```
quisquous: Maybe I am confused, but I don't follow what greedy means in this context.

JLGarber: I'm not completely sure what I meant to say there myself. I think the update I made should better capture the sense of what the section is trying to say?

quisquous:
Thanks! It's nice to have the `-lf` options documented, as it is *such* a timesaver.

PR #1241, opened 2020-03-19
Raidboss: E6S Fr

VanSlanzar:
Adjustement correction fr

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6s.js
+++ b/ui/raidboss/data/05-shb/raid/e6s.js
@@ -29,7 +29,7 @@
       regexJa: Regexes.startsUsing({ source: ['ガルーダ', 'ラクタパクシャ'], id: ['4BF[EF]', '4C0[45]'], capture: false }),
       infoText: {
         en: 'Avoid green nails',
-        fr: 'Evitez les clous',
+        fr: 'Evitez les griffes',
         ko: '초록 발톱 피하기',
       },
     },
--- a/ui/raidboss/data/05-shb/raid/e6s.js
+++ b/ui/raidboss/data/05-shb/raid/e6s.js
@@ -29,7 +29,7 @@
       regexJa: Regexes.startsUsing({ source: ['ガルーダ', 'ラクタパクシャ'], id: ['4BF[EF]', '4C0[45]'], capture: false }),
       infoText: {
         en: 'Avoid green nails',
-        fr: 'Evitez les AoE',
+        fr: 'Evitez les griffes vertes',
         ko: '초록 발톱 피하기',
       },
     },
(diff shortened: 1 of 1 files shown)


quisquous:
I think they look pretty green (or green/blue?) to my eyes: https://icdn6.digitaltrends.com/image/digitaltrends/ffxiv-eden-e2-boss-guide-01-500x500.jpg

@mooondark,  weren't you the one that wrote green, originally in #1164?

VanSlanzar:
griffe

I think it depends on people and screens. But according to the rgb it is almost the same between blue and green. I don't know what to do with it ^^. Maybe emerald green or emerald blue or just claws. Tell me what you think.

JLGarber:
Without going into the specifics of whether the nails are green, it's probably worth at least ensuring there's some reference to the nails. Would it be particularly wrong to put it as "Evitez les griffes"?

(Separate from that, where a direct translation is possible, as here, I kind of feel that translations should match across languages. In this sense, my vote would go to the change in this pull request.)

quisquous:
Yeah, that is my feeling as well, that translations should match if possible.  If "green" is contentious, then "avoid nails" (or "avoid nail cleaves"?) sound fine to me too.

VanSlanzar:
I will call them (nails-clous), with reference to the nails of Ifrit. I redo each content before each change, to be sure that it matches.
About that:
- I saw that there remains some (zones which I would supplement by zones on the ground).
- Some translations of (AOE in en) are (area damage in fr) and others (AOE in fr). I wonder what to do. Tell me your opinion.

As said JLGarber, I will try to be the closest to the translations in order to be harmonized.

VanSlanzar:
**in 1 we must place ourselves between the nails**
Ferostorm1

**in 2 are (AOE or cleaves)?**
ferostorm2


suddenly what do you propose to harmonize and be faithful to the mechanics?

mooondark:
Sorry for answering so late.
For me the nails are more blue, but anyway, I agree with the @JLGarber translation "Evitez les griffes". "Clous" are more for Ifrit.

For AoE, I think we should let it has it is on old translations but for new ones, we should use AoE.
For puddle, hmmm, I really depends of the fight. Maybe we should use "flaque" for liquids & "zone" for the others ?

VanSlanzar:
I agree to the claws. I'll let you do it in English if you want to change.
I also agree with AoE.
And (ground area) or (puddle), I would be very careful depending on the fight.
Thank you all.

quisquous:
> quisquous, would it be possible to create a file or document, for English-French translations. It would be in a cacbot file, not linked to its functioning. It would be fair with basic verbs and expressions, to harmonize everything, nothing more. I would fill it with MooonDark, If he wishes. It could only be changed by you. It will be available to everyone, and especially to you, to avoid repeated changes. And we can make requests there, or debate a subject, when there is need.

Yes, feel free to start a `docs/translations.md` file or something, if you think it would be helpful for consistency.  If you think we need more things in `resources/responses.js`, we can move more there to avoid the need for redundant translations as well.

PR #1256, opened 2020-03-21
fix #1235

### LINKED ISSUE #1235: Radar. Provoke makes you the puller.
Provoke doesn't work on S Rank hunts, yet if someone uses provoke before the pull, you will see their name as puller instead of the one who actually pulls it later.

Bluefissure:
<del>Haven't tested it yet lol.</del>
Good for me.
Actually all hunt mobs won't be affected by provoke I think?

### DIFF
--- a/ui/radar/radar.js
+++ b/ui/radar/radar.js
@@ -257,8 +257,15 @@ class Radar {
       matches = e.detail.logs[i].match(Regexes.abilityFull());
       if (matches) {
         let monster = this.targetMonsters[matches.groups.target.toLowerCase()];
-        if (monster)
-          this.UpdateMonsterPuller(monster, matches.groups.source);
+        if (monster) {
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks!  Give it a test and let me know how it works.  😉

Bluefissure:
> Thanks! Give it a test and let me know how it works. 😉

I tested in CN server and it worked well.

quisquous:
Thanks for giving it a test!

quisquous:
Actually wait, regarding your previous comment, is it the case that you can't provoke A or B ranks too? Or is it just S?

Bluefissure:
I tested again for two B rank mobs, and both have the "(fully resisted)" on provoke, like

Bluefissure:
However, if you're close enough you'll start auto-attacking the hunt mobs after provoking, which will make you the puller.
Anyway, we can filter the provoke out and match the attack logs.

PR #1259, opened 2020-03-21
Raidboss: T9 Fr

VanSlanzar:
exact translation of the name of the skill, adjustment and correction of spelling.
- Line 157, adjustment as this is more understandable for mechanics.
- line 143: I readjust because silence no longer exists as with ARR. Now, there is an interrupts in the French description of the skill. There too, I will be careful because, there is also the stun.
- Line 473, name of skill.
- Line 194, I would like to make a difference:
 -between dispell an alteration  with a mechanism like on Nael, or with esuna. (Dissiper).
-Heal an alteration with Cure as doom (Soigner).

what do you think ? or just dispell for all (dissiper) .


quisquous:
> * Line 194, I would like to make a difference:
>   -between dispell an alteration  with a mechanism like on Nael, or with esuna. (Dissiper).
>   -Heal an alteration with Cure as doom (Soigner).

Yes, maybe it makes senses to say "esuna X" vs "cleanse X".  I think there are some "heal to full" responses for doom as well in other files.  In this case in English, maybe this response should stay "cleanse" but there are other places that maybe could change to "esuna" to disambiguate.

(I think I avoided using the word esuna in the past because not every healer had a skill named esuna.  At least they all do now.)

ui/raidboss/data/02-arr/raid/t9.js
```diff
@@ -154,7 +154,7 @@
       alertText: {
         en: 'Heavensfall',
         de: 'Himmelssturz',
-        fr: 'Chutes du ciel',
+        fr: 'Chute du ciel',
```
quisquous: Why is this heavensfall translated differently than the one in the timeline?

VanSlanzar:
I'm agree. 
- Dispell (disspier) for Esuna
- Heal to Full (Soignez à fond) i think. except, if you have better offer me.

**I't's me in T9 with my screen.**
Destruction

**And here dalamud, Heavensfall.**
Dalamud

exact translation of the name of the skill (Heavensfall) is (Destruction universelle).

**This is an indication of what will happen or what you should do. I found it more logical to leave it like that rather than putting (Destruction universelle),  which is already in 2 places.**
Chute

What do you think ?

PR #1267, opened 2020-03-22
Raidboss: O2N initial package

JLGarber:
No enrage on this one, at least within the first 30 minutes. I don't care to try for an hour just to see.


JLGarber:
Hmm, I'm sorry, I don't remember what the resolution for translation collisions was?

Akurosia:
`(?<! )Petrosphere` and `(?<! )Explosion`

JLGarber:
I suppose I should have figured that out from the translations I copyasta'd. Thanks for jogging my Sunday-nap brain.

ui/oopsyraidsy/data/04-sb/raid/o2n.js
```diff
@@ -0,0 +1,49 @@
...
+        return { type: 'warn', name: e.targetName, text: e.effectName };
+      },
+    },
+    {
+      // Instant tank cleave
+      id: 'O2N Paranormal Wave',
+      damageRegex: '250E',
+      condition: function(e) {
+        // Double taps only.
+        return e.type != '15';
```
quisquous: I really do think it'd be nice to have a doubleTapWarn helper for oopsy.  That seems to be the most common manual trigger these days.

ui/raidboss/data/04-sb/raid/o2n.js
```diff
@@ -0,0 +1,411 @@
...
+  timelineFile: 'o2n.txt',
+  timelineTriggers: [
+    {
+      id: 'O2N Paranormal Wave',
+      regex: /Paranormal Wave/,
+      beforeSeconds: 5,
+      infoText: function(data) {
+        if (data.activeTank == data.me) {
+          return {
+            en: 'Tank cleave on YOU',
```
quisquous: What do you think about using response: Responses.tankCleave() here? It doesn't have the "activeTank" part of it, but it's close?

ui/raidboss/data/04-sb/raid/o2n.js
```diff
@@ -0,0 +1,411 @@
...
+      regexKo: Regexes.ability({ id: '24E8', source: '카타스트로피' }),
+      run: function(data, matches) {
+        data.activeTank = matches.target;
+      },
+    },
+    {
+      id: 'O2N Gravitational Manipulation',
+      regex: Regexes.headMarker({ id: '0071' }),
+      condition: function(data) {
+        // The active tank shouldn't participate in the stack, as the boss would turn.
```
quisquous: I think this is maybe overthinking it.  The active tank can still participate without turning the boss if folks place it close.  Or they can hear the trigger and decide not to do it.

ui/raidboss/data/04-sb/raid/o2n.js
```diff
@@ -0,0 +1,411 @@
...
+        return !data.me == data.activeTank;
+      },
+      alertText: function(data, matches) {
+        if (data.me == matches.target) {
+          return {
+            en: 'Stack marker on YOU',
+          };
+        }
+        return {
+          en: 'Levitate--Stack on ' + data.shortName(matches.target),
```
quisquous: Maybe "Levitate and Stack on" etc vs the double dash? (Or alternatively use Responses.stackOn('alert') and then have a second "levitate" info callout for anybody not levitating?)

ui/raidboss/data/04-sb/raid/o2n.js
```diff
@@ -0,0 +1,411 @@
...
+      regex: Regexes.startsUsing({ id: '2512', source: 'Catastrophe', capture: false }),
+      regexDe: Regexes.startsUsing({ id: '2512', source: 'Katastroph', capture: false }),
+      regexFr: Regexes.startsUsing({ id: '2512', source: 'Catastrophe', capture: false }),
+      regexJa: Regexes.startsUsing({ id: '2512', source: 'カタストロフィー', capture: false }),
+      regexCn: Regexes.startsUsing({ id: '2512', source: '灾变者', capture: false }),
+      regexKo: Regexes.startsUsing({ id: '2512', source: '카타스트로피', capture: false }),
+      alertText: function(data) {
+        if (!data.levitating) {
+          return {
+            en: 'Earthquake: Levitate',
```
quisquous: The original o2s triggers were one of the first and doesn't match all the other triggers in some ways.  I'd make this just "levitate" for brevity.

quisquous:
> No enrage on this one, at least within the first 30 minutes. I don't care to try for an hour just to see.

:sob: :sob: :sob:

That's some dedication.

ui/raidboss/data/04-sb/raid/o2n.js
JLGarber: I suppose so. I may have been *slightly* upset from seeing other tanks doing exactly this thing with melee in the party.

ui/raidboss/data/04-sb/raid/o2n.js
JLGarber: I suppose that's true. I was trying to get around the lack of `matches` elements in timeline triggers for... something, but I can't think what it was now. (It was very clear when I wrote it, of course, so clear that a comment explaining was *completely* and laughably unnecessary.)

ui/raidboss/data/04-sb/raid/o2n.js
JLGarber: That second one should be better. I was kind of trying to avoid overlapping callouts, but you're right, as it's written it calls too much if the user is already levitating.

PR #1269, opened 2020-03-22
Added E6S Strike Spark Trigger Safezone callout

mrdregan:
I tried to port my Triggernometry Strike Spark Trigger into cactbot (see https://github.com/mrdregan/triggernometry_triggers/blob/master/E6S/E6S_Strike_Spark.xml for reference).

This is still untested and I used the implementation of @valarnin as base for this implementation - but with my approach and not the calculcations, since you only need to check which Actor has the highest HEX ID, since that one will always be the safe zone.

I will run most likely tuesday E6S so I can test it then, if anybody else wants to give it a go earlier or has some feedback, feel free to comment! Also I guess I will need help with localizing the callout? I can add german later on, but not the other languages.


ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,100 @@
...
   triggers: [
+    {
+      id: 'E6S Strike Spark',
+      regex: Regexes.startsUsing({ source: ['Ifrit', 'Raktapaksa'], id: '4BD3', capture: false }),
+      regexDe: Regexes.startsUsing({ source: ['Ifrit', 'Raktapaksa'], id: '4BD3', capture: false }),
+      regexFr: Regexes.startsUsing({ source: ['Ifrit', 'Raktapaksa'], id: '4BD3', capture: false }),
+      regexJa: Regexes.startsUsing({ source: ['イフリート', 'ラクタパクシャ'], id: '4BD3', capture: false }),
+      promise: function(data) {
+        let p = new Promise(async (res) => {
+          // helper function to delay the promise execution for the given time
```
quisquous: Hmm.  I wonder if delaySeconds should come before promise, so that promises can lean on delaySeconds to do this.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,100 @@
...
+          const sleep = m => new Promise(r => setTimeout(r, m));
+
+          await sleep(10000);
+
+          let combatantNames = null;
+
+          // select the 4 most recent Ifrit or Raktapaksa's depending on phase
+          if (data.phase = 'ifrit') {
+              combatantNames = { names: ['Ifrit', 'イフリート', '伊弗利特', '이프리트'] };
```
quisquous: `data.lang` does have the language in it, if you wanted to select only one.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,100 @@
...
+                } else {
+                  return id2;
+                }
+              };
+
+              let currentHighestCombatant = null;
+
+              // we need to filter for the Ifrit with the highest HEX ID
+              // since that one is always the safe spot.
+              combatantData.combatants.forEach(combatant => {
```
quisquous: I know it's less efficient, but what about something like this for brevity:
`let currentHighestCombatant = combatantData.combatants.sort((a, b) => a.ID.localeCompare(b.ID)).pop();`

I think lexicographically, hex ids sort in order, so no need to parseInt them.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,100 @@
...
+
+                  if (newHighestID !== combatant.ID) {
+                    currentHighestCombatant = combatant;
+                  }
+                } else {
+                  currentHighestCombatant = combatant;
+                }
+              });
+
+              // all variation ranges for all the 9 ball positions for the kicking actors
```
quisquous: Thanks for the comment!

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,100 @@
...
+              // @TODO: Need the CN/KR names for Raktapaksa
+              combatantNames = { names: ['Raktapaksa', 'ラクタパクシャ'] };
+          }
+            
+          let combatantData = await window.callOverlayHandler({
+            call: 'getCombatants',
+            data: combatantNames
+          });
+
+          if (combatantData !== null &&
```
quisquous: style nit: could just call res() and early return here if these aren't true to avoid having to indent all of this

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,100 @@
...
+                safeZoneString1 = "south";
+              }
+
+              if (currentHighestCombatant.posX > 84 && currentHighestCombatant.posX < 94) {
+                safeZoneString2 = "west";
+              } else if (currentHighestCombatant.posX > 106 && currentHighestCombatant.posX < 116) {
+                safeZoneString2 = "east";
+              }
+
+              data.safeZone = safeZoneString1 + safeZoneString2;
```
quisquous: Is it possible to make data.safeZone a localized object, i.e `data.safeZone = { en: '', de: '', fr: '' etc etc }`? I think this probably would involve a little bit of logic change above.

ui/raidboss/data/05-shb/raid/e6s.js
mrdregan: That'd be great. All triggers for positional data receiving I wrote this tier, are related to retrieving combatatant data after a delay. Also I assume the usage of promises are fairly new in cactbot, which makes changing it rather non-impactful, or?

mrdregan:
Adjusted everything that was mentioned! :)

ui/raidboss/data/05-shb/raid/e6s.js
quisquous: Yeah, promises *just* got added by @valarnin and there's not any use of them yet other than in a test.  So, very easy to change.  I'll try to do that later tonight.

valarnin:
Thanks for this. I gave up on my logic/position-based implementation because my static switched to using the uptime strat for this mechanic.
The only thing I can think of that you might want to adjust is the timing of the direction. Right now I think with the 10s delay the alert should pop at the exact same time as the eruption circles drop, but it might be helpful to know what direction to run beforehand. Maybe there should be a 'run' alert for when the eruptions hit and the direction should be shown at 7.5s instead? @quisquous, opinions?

mrdregan:
The `startsUsing` condition means this gets triggered when the cast starts, no? If you look at my triggernometry trigger, I use the logline for the finish of the cast. So you need to add the cast time of 3 seconds to my 7 seconds delay, which results in the 10 seconds I implemented here. 3 seconds later the eruption circles drop.

valarnin:
Yeah, my timing for the jump was off, but the original point still stands. Not sure if there should be a Run alert when the eruption circles hit.

mrdregan:
My timing is ~2 seconds after actors are visible. I guess if you want to stretch it a bit further you could reduce it by that time. I was just adding some time, since they don't have the position when they spawn, they get moved to their destination delayed - so i was adding some buffer to be on the safe side.

quisquous:
Hmm, I did some e6s last night with this applied and saw no callouts (and forgot to look for errors, etc). I'm going to add more logging before reclears tonight and see if I can debug this further.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,113 @@
...
+          else
+            combatantNames = { names: [raktapaksaLocaleNames[data.lang]] };
+
+
+          let combatantData = await window.callOverlayHandler({
+            call: 'getCombatants',
+            data: combatantNames,
+          });
+
+          // if we could not retrieve combatant data, the
```
valarnin: Hmm, shouldn't this return from the promise handler after calling `res();`?

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,113 @@
...
+          // southeast  x: 107-115  y: 107-115
+          // southwest  x: 85-93    y: 107-115
+          let safeZoneObj1 = { en: '', de: '' };
+          let safeZoneObj2 = { en: '', de: '' };
+
+          // don't need to go through all the posibilities,
+          // only those 4 ifs do reflect the above positions
+          if (currentHighestCombatant.posY > 84 && currentHighestCombatant.posY < 94) {
+            safeZoneObj1 = {
+              en: 'north',
```
valarnin: It might make sense to have common phrases like north/south/east/west/etc be in a new helper class in the same vein as the `Regexes` class? At the least, it makes more sense to have this be a const struct like this instead, which can have the correct return value looked up from, so that all the localization is in one place:
```javascript
const directions = {
  nw: {
    en: 'northwest',
    de: 'nordwesten',
    ...
  },
  n: {
    ...
  }
  ...
};
```

valarnin:
Here's the error:
```
[3/24/2020 11:18:15 PM] Info: Cactbot: BrowserConsole: Uncaught (in promise) TypeError: a.ID.localeCompare is not a function (Source: file:///C:/users/steamuser/Application%20Data/Advanced%20Combat%20Tracker/Plugins/Cactbot/ui/raidboss/raidboss.html, Line: 63)
```

ui/raidboss/data/05-shb/raid/e6s.js
quisquous: I need to think about where this would go.  I think `responses.js` itself is meant to be more of a "this is the entire trigger's response, with maybe some logic", but maybe it'd be good to have some file of localized phrases on their own. What would I call such a file... :thinking:

valarnin: Hmm... `language.js`? Could export a module such that:
```javascript
let Language = {
  directions: {
    nw: ...
  },
};
if (typeof module !== 'undefined' && module.exports) {
  module.exports = {
    language: Language,
  };
}
```

I'm just not sure what else would go in the file, given that a lot of the localization is already present in `responses.js` as it is.

quisquous:
Huh, does that mean there's no ID field? Or that it's not a string?

valarnin:
Combatant ID is an int.
image
As such, this should work:
```javascript
let currentHighestCombatant =
  combatantData.combatants.sort((a, b) => a.ID > b.ID ? a.ID : b.ID).pop();
```
image

quisquous:
> ```js
> let currentHighestCombatant =
>   combatantData.combatants.sort((a, b) => a.ID > b.ID ? a.ID : b.ID).pop();
> ```

Sorry, but I am a little skeptical of this.  Javascript custom sort functions expect a return of positive number, negative number, or zero depending on the comparison result.  In this case, I think maybe you're looking for `(a, b) => b.ID - a.ID`?

valarnin:
> In this case, I think maybe you're looking for `(a, b) => b.ID - a.ID`?

Probably, yeah. I only did a quick check to make sure it didn't error.

mrdregan:
Ah ok, I was assuming that the returning ID was still a hex string. Will test the change today and push the fix if it worked out.

mrdregan:
I adjusted and checked it now again: it returns all combatants and doesn't filter by name only.

I think when I fixed that one, it should work.

mrdregan:
Ok, this version works. I tested it. I didn't get to the delaySeconds change though and gotta check for comments. But I am short on time rn.

quisquous:
I can wait for Tuesday, no rush.  :)

FlippFuzz:
I'm sorry, been trying to test this.
However, this trigger don't seem to work for me. 

Here's an example.
Trigger calls out nothing.
However, the actual safe spot is northeast. 

I've added a simple console.log() to dump out the values of PosX and PosY and got the follow.
```
[29/3/2020 9:28:17 pm] Info: Raidboss: BrowserConsole: currentHighestCombatant.PosX: 99.98825 (Source: , Line: 1056)
[29/3/2020 9:28:17 pm] Info: Raidboss: BrowserConsole: currentHighestCombatant.PosY: 95.61156 (Source: , Line: 1057)
```
They don't match the conditionals.

What is window.callOverlayHandler({call: 'getCombatants', names: combatantNames, }); doing?
Is it just pulling the "Added new combatant Ifrit" log entries?
I do have varying IDs, but the positions of all of them are the same.
```
[21:28:10.663] 03:4000516A:Added new combatant Ifrit.  Job: N/A Level: 80 Max HP: 19240000 Max MP: 10000 Pos: (99.98825,95.61156,0) (92880000011559).
[21:28:10.663] 03:4000516B:Added new combatant Ifrit.  Job: N/A Level: 80 Max HP: 19240000 Max MP: 10000 Pos: (99.98825,95.61156,0) (92880000011559).
[21:28:10.663] 03:4000516C:Added new combatant Ifrit.  Job: N/A Level: 80 Max HP: 19240000 Max MP: 10000 Pos: (99.98825,95.61156,0) (92880000011559).
[21:28:10.779] 03:4000516D:Added new combatant Ifrit.  Job: N/A Level: 80 Max HP: 19240000 Max MP: 10000 Pos: (99.98825,95.61156,0) (92880000011559).
```

valarnin:
> What is window.callOverlayHandler({call: 'getCombatants', names: combatantNames, }); doing?
> Is it just pulling the "Added new combatant Ifrit" log entries?

That calls the `getCombatants` handler in OverlayPlugin, which pulls active combatant data from the FFXIV client memory. See these for reference:
https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/EventSources/MiniParseEventSource.cs#L103-L139
https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/EventSources/MiniParseEventSource.cs#L185-L255

I think the issue is the sort, it should be `b.ID - a.ID` but it's swapped? So it's getting the combatant with the lowest ID right now instead of the highest ID.

mrdregan:
The trigger works just fine for me. It does not call out anything though, did you look at the green text alert? Also where did you add the logging? On NPC spawn? Or within the code where the if clause is? If the text alert really is not showing up, then it might be a timing issue I would assume - but I doubt it. Just in case you could set the 10 second delay to 11 seconds and try it again.

FlippFuzz:
> Also where did you add the logging? On NPC spawn? 

Right before res() in the promise.

My issue here is that PosX: 99.98825, PosY: 95.61156 is outside all of these conditionals, which is why it doesn't trigger.

```
          if (currentHighestCombatant.PosX > 84 && currentHighestCombatant.PosX < 94) {
          } else if (currentHighestCombatant.PosX > 106 && currentHighestCombatant.PosX < 116) {

          if (currentHighestCombatant.PosY > 84 && currentHighestCombatant.PosY < 94) {
          } else if (currentHighestCombatant.PosY > 106 && currentHighestCombatant.PosY < 116) {
```

> Just in case you could set the 10 second delay to 11 seconds and try it again.

I did reduce the delay to 5 seconds.


Well, I guess what I can do is to just dump out the contents of "combatantData.combatants.sort" just to make sure that I can see all 4 Ifrit spawns and their sort order.

mrdregan:
Did you reduce the delay to 5 seconds before you tested it for the first time?

There's a simple reason for the "huge" delay: The position is delayed updated. 5 seconds is way too early. That would easily explain, why you did not get any message.

FlippFuzz:
> Just in case you could set the 10 second delay to 11 seconds and try it again.

This solved the issue for me.
For some reason, I needed 11 seconds and not 10.

valarnin:
Might make sense to put the coords check in a loop and re-check every 100ms? Something like this:
```javascript
let direction = '';
for (let i = 0; i < 20; ++i) {
  //Check coords
  //If a match is found, break the loop
  //Otherwise, await 100, re-query getCombatants for just the highest combatant's ID and update that object
}
//Map direction string to return object
```

Otherwise this will probably continue to be a timing issue for anyone with a higher ping (e.g. EU players connecting to US data centers, or US players connecting to JP DCs).

mrdregan:
@FlippFuzz Are you really sure you tried it with the 10 seconds delay? Because you were talking about a modification to 5 seconds before.

So far that has not been an issue for people using the triggernometry version. The only thing is that this here actually calls it a bit earlier than the triggernometry one. I was thinking to be able to get away with that. Apparently I don't, if FlippFuzz really tried it with the 10 seconds and not 5 seconds as he stated. If we can find somebody else to verify that, then I'll increase to to 11 seconds in order to make it "less strict".

@valarnin The thing you are suggesting will happen only most likely in case of a huge lag spike (because if somebody has a "high ping" the causing log line will be delayed already and it nullifies the effect). I'm not sure that's the proper way to handle that, especially for something that could be applied/valid to *all* async/delayed trigger logic. Also to normal triggers aswell, if you have a lag spike, then it all gets messy anyways.

FlippFuzz:
> @FlippFuzz Are you really sure you tried it with the 10 seconds delay? Because you were talking about a modification to 5 seconds before.

Yes, I had modified to 5 seconds the first time that I've reported the issue, but I've re-tested with 10 and 11 yesterday.
That's literally all I've been doing yesterday and dumping out the contents of combatantData.combatants.sort() into console.log().

I think @valarnin solution is actually more elegant because it's potentially the fastest and doesn't depend on a fixed waiting time.
Basically check that the last four elements in the list are reasonable. If so, return immediately. 

Anyways, I think for now, you guys can still commit at 10 seconds wait if nobody else reports the problem.
I'm perfectly happy with modifying the trigger via cactbot-user.
(I'm already adding a few lines to convert compass directions (North/South/etc) to map markers positions anyway(A/B/etc)).

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,112 @@
...
+          // southwest  x: 85-93    y: 107-115
+          let safeZoneObj1 = { en: '', de: '' };
+          let safeZoneObj2 = { en: '', de: '' };
+
+          // don't need to go through all the posibilities,
+          // only those 4 ifs do reflect the above positions
+          if (currentHighestCombatant.PosY > 84 && currentHighestCombatant.PosY < 94) {
+            safeZoneObj1 = {
+              en: 'north',
+              de: 'nord',
```
VanSlanzar: fr: 'nord',

If you can add that, thank you.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,112 @@
...
+            };
+          } else if (currentHighestCombatant.PosY > 106 && currentHighestCombatant.PosY < 116) {
+            safeZoneObj1 = {
+              en: 'south',
+              de: 'süd',
```
VanSlanzar: fr: 'sud',

If you can add that, thank you.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,112 @@
...
+
+          if (currentHighestCombatant.PosX > 84 && currentHighestCombatant.PosX < 94) {
+            safeZoneObj2 = {
+              en: 'west',
+              de: 'west',
+            };
+          } else if (currentHighestCombatant.PosX > 106 && currentHighestCombatant.PosX < 116) {
+            safeZoneObj2 = {
+              en: 'east',
+              de: 'ost',
```
VanSlanzar: fr: 'est',

If you can add that, thank you.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,112 @@
...
+            safeZoneObj1 = {
+              en: 'south',
+              de: 'süd',
+            };
+          }
+
+          if (currentHighestCombatant.PosX > 84 && currentHighestCombatant.PosX < 94) {
+            safeZoneObj2 = {
+              en: 'west',
+              de: 'west',
```
VanSlanzar: fr: 'ouest',

If you can add that, thank you.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,112 @@
...
+          } else if (currentHighestCombatant.PosX > 106 && currentHighestCombatant.PosX < 116) {
+            safeZoneObj2 = {
+              en: 'east',
+              de: 'ost',
+            };
+          }
+
+          data.safeZone = {
+            en: safeZoneObj1.en + safeZoneObj2.en,
+            de: safeZoneObj1.de + safeZoneObj2.de,
```
VanSlanzar: fr: safeZoneObj1.fr + safeZoneObj2.fr,

If you can add that, thank you.

FlippFuzz:
> That's literally all I've been doing yesterday and dumping out the contents of combatantData.combatants.sort() into console.log().

I might as well provide the output here for future reference.
Ifrit's Strike Spark: https://pastebin.com/gjwG2dL7
Raktapaksa's Strike Spark: https://pastebin.com/H9D8L9iQ

Changes that I made to create those log files.
Modified Line 61 to store the result of sort in a variable for printing later:
```
let sortedCombatants = combatantData.combatants.sort((a, b) => a.ID - b.ID);
let currentHighestCombatant = sortedCombatants.pop();
```
Added the following to line 106, right before res():
```
console.log("Raw Data for " + data.phase);
sortedCombatants.forEach(function (item, index) {
    console.log("Index: " + index + " ID: " + item.ID + " PosX: " + item.PosX + " PosY: " + item.PosY);
});
console.log("Final Result");
console.log("currentHighestCombatant.PosX: " + currentHighestCombatant.PosX);
console.log("currentHighestCombatant.PosY: " + currentHighestCombatant.PosY);
console.log("safeZone: " + data.safeZone.en);
```

So, if we really want to implement @valarnin idea of not waiting for a fixed time, it's going to look something like this:

```
for (let i = 0; i < 20; ++i) {
  // Grab combatantData
   combatantData = await window.callOverlayHandler({
    call: 'getCombatants',
    names: combatantNames,
  });

    // Check length
    if (data.phase === 'ifrit' && combatantData.length >= 24)
      break;
    else if (data.phase === 'raktapaksa' && combatantData.length >= 47)
      break;
	  
    //Otherwise, await 100, re-query getCombatants
    await sleep(100);
}
```

I'm not too sure how getCombatants works, but the number of results from getCombatants seem constant throughout my tests.

valarnin:
Checking the count of combatants wasn't my intention. I was suggesting that you sort and check the highest ID's position, and break if it was in one of the valid locations for a safe spot.

mrdregan:
Checking against count is pointless, since the combatant count does not change when we need it to. Pretty much what @valarnin said would be the only way.

But I am still against it, I am not a friend of overcomplicating things for things such as lag spikes - and if we do I would vote for something generic, that has reuse potential. I've increased it by another 500ms - if we don't have positions by then, the callout will be too late aswell and therefore will be pointless anyways, since you won't have enough time to react (this new delay amounts to ~3 seconds reaction time).

ui/raidboss/data/05-shb/raid/e6s.js
mrdregan: This is the last bit of the PR I didn't adjust yet. Any more input or should I ignore it for now?

mrdregan:
I did my adjustments, will have another test run tomorrow to check the last changes.

ui/raidboss/data/05-shb/raid/e6s.js
valarnin: I'm still of the opinion that it makes more sense to have all the language stuff in one spot in the trigger and then just map to it for the direction, and it would make it easier to relocate later to a common module.

mrdregan: Maybe we can just go with `responses.js` and add entries such as `Responses.goNorth()` etc? Would create a bit more writing work in the trigger, but yeah, having the language stuff in one spot sounds better to me aswell.

valarnin: This PR seems relevant to the conversation.
https://github.com/quisquous/cactbot/pull/1303

quisquous: Belatedly, I think this is fine to ignore and fix in a followup.

I agree with valarnin that translation-wise, it's almost always better to not do any string concatenation.  To just add logic that figures out that something is northeast, and then lookup "northeast" in some map.

Re: where that map should live, I think the Responses object itself is not the best fit, because Responses.foo is a function that returns an object meant to be attached to the `response` field of a trigger, i.e. `{ alertText: { en: 'Buster on YOU' } }` where this is more something where you just want `Trans.north` to literally be the value `{ en: 'North', etc }`.  If that makes sense?  Responses is good for simple and complicated trigger responses, but I think there's room for something simpler.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -7,6 +7,114 @@
...
+          // all variation ranges for all the 9 ball positions for the kicking actors
+          // north      x: 96-104   y: 85-93
+          // northeast  x: 107-115  y: 85-93
+          // northwest  x: 85-93    y: 85-93
+          // east       x: 107-115  y: 96-104
+          // west       x: 85-93    y: 96-104
+          // south      x: 96-104   y: 107-115
+          // southeast  x: 107-115  y: 107-115
+          // southwest  x: 85-93    y: 107-115
+          let safeZoneObj1 = { en: '', de: '' };
```
quisquous, line 70: I think I only have one other cleanup comment here, and that's that you should either fill out every language here (just to avoid future bugs) or to change the logic such that you initialize these to null and do something different later.

quisquous:
Overall looks good though, and worked just fine for me last week, at least in the sense that it called out west while we all stacked north for uptime.  :innocent:

Thanks for all the work here.  Much appreciated!

ui/raidboss/data/05-shb/raid/e6s.js
quisquous: Filed #1307

quisquous:
> I did my adjustments, will have another test run tomorrow to check the last changes.

What's the verdict?

quisquous:
This worked for me locally the other week, so I'm going to merge it.

PR #1273, opened 2020-03-23
Raidboss: T9 Fr

VanSlanzar:
Link conversation.
when I read it again, I doubt. Should I put the real name or a better indication.

https://github.com/quisquous/cactbot/pull/1259#issuecomment-602537597

I await your answer. Ty for your help.

### DIFF
--- a/ui/raidboss/data/02-arr/raid/t9.js
+++ b/ui/raidboss/data/02-arr/raid/t9.js
@@ -154,7 +154,7 @@
       alertText: {
         en: 'Heavensfall',
         de: 'Himmelssturz',
-        fr: 'Chute du ciel',
+        fr: 'Destruction universelle',
       },
     },
     {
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/02-arr/raid/t9.js
```diff
@@ -154,7 +154,7 @@
       alertText: {
         en: 'Heavensfall',
         de: 'Himmelssturz',
-        fr: 'Chutes du ciel',
+        fr: 'Chute du ciel',
```
quisquous: https://github.com/quisquous/cactbot/blob/master/CONTRIBUTING.md#trigger-text are my general guidelines.

In particular:
> Tell the player what to do rather than the mechanic name, i.e. prefer Get Out vs Iron Chariot
> If multiple strategies exist, tell the player the mechanic (Jail on YOU) instead of dictating a strategy.

Or, in this case, it's "if the directions are too confusing / complicated to put in a trigger, tell the player the mechanic".  Heavensfall means "get out of the middle but not too far out of the middle", but that's too long.  So, I think better to call it the mechanic name itself rather than describing what the mechanic does, and make this `Destruction universelle`.

VanSlanzar: If these are your guidelines, I follow them.
It disturbed me and I was only having doubts.
Thank you for your help.

PR #1274, opened 2020-03-24
Raidboss: O3N initial package

JLGarber:
There's a *lot* going on for a normal mode here! This was my favorite Savage encounter in Stormblood, so I choose to believe the complexity of normal mode is a testament to the excellence of that encounter.

There's probably a little more overthinking than necessary, particularly when this battle rarely even gets to the intermission. (One thing that does need to be added, speaking of the intermission, is a check for where the player spawns in the desert. It's easy to execute, but the timing to get the bonus seems *really* tight.)


JLGarber:
Oof. And now that I look at it, I totally forgot timeline replacements. A moment...

ui/oopsyraidsy/data/04-sb/raid/o3n.js
```diff
@@ -0,0 +1,94 @@
...
+      regexFr: Regexes.ability({ id: '367', source: 'Halicarnasse', capture: false }),
+      regexJa: Regexes.ability({ id: '367', source: 'ハリカルナッソス', capture: false }),
+      regexCn: Regexes.ability({ id: '367', source: '哈利卡纳苏斯', capture: false }),
+      regexKo: Regexes.ability({ id: '367', source: '할리카르나소스', capture: false }),
+      condition: function(e, data) {
+        return !data.initialized;
+      },
+      run: function(e, data) {
+        data.gameCount = 0;
+        // Indexing phases at 1 so as to make phases match what humans expect.
```
quisquous: What is this, Lua? ;)

If you know it offhand, can you leave a comment for what each of the phases are and what number they are, just for future reference?

ui/raidboss/data/04-sb/raid/o3n.js
```diff
@@ -0,0 +1,474 @@
...
+    en: /^Deltascape \(V3\.0\)$/,
+  },
+  timelineFile: 'o3n.txt',
+  timelineTriggers: [
+    {
+      id: 'O3N Frost Breath',
+      regex: /Frost Breath/,
+      beforeSeconds: 4,
+      condition: function(data) {
+        return data.role == 'tank'&& !data.activeTank == data.me;
```
quisquous: I can't find an eslint rule for this, but can you add a space before the && operator?

ui/raidboss/data/04-sb/raid/o3n.js
```diff
@@ -0,0 +1,474 @@
...
+      // No point in checking whether the user has the frog debuff,
+      // if they didn't get it, or got it when they shouldn't have, there's no fixing things.
+      infoText: function(data) {
+        if (data.phaseNumber == 3 && data.gameCount % 2 == 0) {
+          return {
+            en: 'Stand on frog tile',
+          };
+        }
+        return {
+          // Maybe there's a cleaner way to do this than just enumerating roles?
```
quisquous: This seems reasonable.  I wonder what will happen when 8 blu do this <_<

ui/raidboss/data/04-sb/raid/o3n.txt
```diff
@@ -0,0 +1,134 @@
...
+377.1 "Ribbit (take)" sync /:Halicarnassus:2466:/ window 30,30
+384.2 "The Game (toad)" sync /:Halicarnassus:246D:/
+394.3 "Dimensional Wave" sync /:Halicarnassus:2465:/
+403.5 "Spellblade Holy" sync /:Halicarnassus:22EF:/
+407.6 "Spellblade Fire III" sync /:Halicarnassus:2460:/
+412.5 "Holy Blur" sync /:Halicarnassus:2464:/
+412.7 "The Queen's Waltz (Sword Dance)" sync /:Halicarnassus:246F:/ window 30,30
+421.8 "Dimensional Wave" sync /:Halicarnassus:2465:/
+
+# The blocks don't change, but if we don't track them we can't note the enrage.
```
quisquous: What does this mean?

ui/raidboss/data/04-sb/raid/o3n.js
```diff
@@ -0,0 +1,474 @@
...
+      regexDe: Regexes.addedCombatant({ name: 'Riesendrache', capture: false }),
+      regexFr: Regexes.addedCombatant({ name: 'dragon suprême', capture: false }),
+      regexJa: Regexes.addedCombatant({ name: 'ドラゴングレイト', capture: false }),
+      regexCn: Regexes.addedCombatant({ name: '巨龙', capture: false }),
+      regexKo: Regexes.addedCombatant({ name: '거대 드래곤', capture: false }),
+      condition: function(data) {
+        return data.role == 'tank';
+      },
+      infoText: {
+        en: 'Grab dragon--point away',
```
quisquous: I do kind of feel like "point away" is implied for pretty much everything tanked.  It may just be enough to say "Dragon Add" for brevity.

ui/raidboss/data/04-sb/raid/o3n.js
```diff
@@ -0,0 +1,474 @@
...
+        for (let i = 1; i < 3; i++) {
+          if (data.holyTargets[i] == data.me) {
+            return {
+              en: 'Get out',
+              de: 'Raus da',
+            };
+          }
+        }
+        return {
+          en: 'Stack on ' + data.holyTargets[0],
```
quisquous: Can you fill in the other "stack on" translations from responses.js? Same with Get Out and Spread?

One other option here is to make this a `response:` instead of an `alertText` and do something like,

```
if (data.holyTargets[i] == data.me)
  return Responses.getOut()(data, matches);
```

I want to think of a better way to handle calling Responses from inside a response function, but I think this is still an option.  I don't think that there's any great way at the moment to handle stack, unless you wrote your own matches object and then called Responses.stackOn()(data, {target: data.holyTaragets[0]}), but that seems a bit invasive too.  I'm still trying to figure out the best way to handle all of this.

ui/oopsyraidsy/data/04-sb/raid/o3n.js
```diff
@@ -0,0 +1,94 @@
...
+      // Why overthink Normal mode, however?
+      id: 'O3N The Game',
+      // Guess what you just lost?
+      abilityRegex: '246D',
+      condition: function(e, data) {
+        // If the player takes no damage, they did the mechanic correctly.
+        return e.damage > 0;
+      },
+      mistake: function(e) {
+        return { type: 'warn', blame: e.targetName, text: { en: e.abilityName } };
```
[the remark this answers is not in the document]
JLGarber: Oops. That's a remnant from a prior version where there was additional text.

ui/raidboss/data/04-sb/raid/o3n.js
JLGarber: I'll be careful in future. That being said, this condition was probably unnecessary altogether here.

ui/raidboss/data/04-sb/raid/o3n.txt
JLGarber: There's no way to differentiate the enrage cast without doing wonky stuff with junction blocks. (Her cast for the enrage is the same ID as non-enrage uses.) It may not be really necessary, since nobody sees it in the duty finder, but I enumerated the blocks as opposed to looping since there is an actual enrage.

ui/raidboss/data/04-sb/raid/o3n.js
JLGarber: That's fair. I think for this specific case I'll just enumerate the translations from `responses.js, since I don't have time over lunch to test and bugfix using `response`. I'll probably follow that up with a proper fix later. "Sometime".

quisquous: That's fair for now.  There was a similar conversation in #1269, so maybe it'd be good to just have some set of functions that you can call that just return a `{ en: 'stuff here' }` object vs a nested `{ alertText: { en: 'stuff here' } }` style object.

JLGarber: (Commenting here so as to avoid cluttering the discussion on the other request, and I don't think this really warrants a separate issue from me.)

Naively, what about adding `ResponseText` to `responses.js`?

```
let Responses = {
  stackOn: (sev) => {
    let obj = {};
    obj[defaultAlertText(sev)] = (data, matches) => {
      let target = getTarget(matches);
      if (target == data.me) {
        return responseText.stackOnYou();
      }
      return responseText.stackOnTarget(data, target);
    };
    return obj;
  },
}
let responseText = {
  stackOnYou: () => {
    return {
      en: 'Stack on YOU',
      de: 'Auf DIR sammeln',
      fr: 'Package sur VOUS',
      ja: '自分にスタック',
      cn: '集合点名',
      ko: '쉐어징 대상자',
    };
  },
  stackOnTarget: (data, target) => {
    return {
      en: 'Stack on ' + data.ShortName(target),
      de: 'Auf ' + data.ShortName(target) + ' sammeln',
      fr: 'Package sur ' + data.ShortName(target),
      ja: data.ShortName(target) + 'にスタック',
      cn: '靠近 ' + data.ShortName(target) + '集合',
      ko: '쉐어징 → ' + data.ShortName(target),
    };
  },
},
```
(I'm not *terribly* happy about layering Yet More Functions over things? But it would maintain consistency with the rest of the helper constructs.)

I feel like it kind of makes sense to have the translation objects live in the same file as the response functions. If not, we could make `common-translations.js` or something and dump all the `responseText` objects in there.

PR #1277, opened 2020-03-25
Add monsters id to resources/hunt.js

rreminy:
Modified gen_hunt_data.py to also store the monsters id into hunt.js

Updated hunt.js included

### DIFF
--- a/util/gen_hunt_data.py
+++ b/util/gen_hunt_data.py
@@ -24,6 +24,8 @@ def update_german(list, search, replace):
 
 
 def parse_data(monsters, csvfile, lang="en", name_map=None):
+    print(f"Processing {lang} hunt names language...")
+
     reader = csv.reader(csvfile)
     # skip the first three header lines
     next(reader)
@@ -66,18 +68,22 @@ def parse_data(monsters, csvfile, lang="en", name_map=None):
             rank = "B"
 
         if nm_id in monsters:
+            assert monsters[nm_id]['id'] == name_id
             assert monsters[nm_id]["rank"] == rank
         else:
-            monsters[nm_id] = {"rank": rank, "name": {}}
+            monsters[nm_id] = {"id": name_id, "rank": rank, "name": {}}
 
         monsters[nm_id]["name"][lang] = name
 
 
 def update_coinach(monsters, reader):
+    print("Reading Notorious Monster list...")
+
+    nm_csv = reader.rawexd("NotoriousMonster")
     languages = ["en", "de", "fr", "ja"]
     for locale in languages:
-        exd = reader.exd("NotoriousMonster", lang=locale)
-        parse_data(monsters, exd, lang=locale, name_map=None)
+        name_map = process_npc_names(reader.exd("BNpcName", lang=locale))
+        parse_data(monsters, nm_csv, lang=locale, name_map=name_map)
     return monsters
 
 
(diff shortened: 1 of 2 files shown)


util/gen_hunt_data.py
```diff
@@ -66,18 +68,22 @@ def parse_data(monsters, csvfile, lang="en", name_map=None):
...
 
 def update_coinach(monsters, reader):
+    print("Reading Notorious Monster list...")
+
+    nm_csv = reader.rawexd("NotoriousMonster")
     languages = ["en", "de", "fr", "ja"]
     for locale in languages:
-        exd = reader.exd("NotoriousMonster", lang=locale)
-        parse_data(monsters, exd, lang=locale, name_map=None)
+        name_map = process_npc_names(reader.exd("BNpcName", lang=locale))
```
quisquous: Why do you need to read both the NotoriousMonster and BNpcName tables here? Isn't the id in the NotoriousMonster table as `base` in that function above?

rreminy: I changed the function call from "reader.exd" to "reader.rawexd" when getting the NotoriousMonsters table, which allowed me to get their raw ids, however that meant that as a result I needed the BNpcName table as well now. I then used it to create a name map using process_npc_names function, as was done for the cn and ko languages.

quisquous: Yes, my confusion is: isn't `base` in the `parse_data` function the npc id already?  Is that not correct?

rreminy: I am a bit confused: I am basing this on what I see on hunts, but when I went in-game to figure out what the game uses for npc id, in the logs I found this for Sum and Kurma:

`[00:57:44.829] 03:4004FCEE:Added new combatant Sum.  Job: N/A Level: 70 Max HP: 8471855 Max MP: 10000 Pos: (-277.9004,443.0309,6.004814) (60010000006850).
[01:14:03.207] 03:4004CA60:Added new combatant Kurma.  Job: N/A Level: 70 Max HP: 250755 Max MP: 10000 Pos: (-490.8055,353.6587,14.59056) (60060000006855).`

I know Sum is identified as 6001 in the BNpcName table but here its 60010000006850, and likewise for Kurma.

It seems I have assumed the wrong thing when I thought how npc id are identified ingame: It seems to have both the BNpcName and BNpcBase merged into its number, which correlates with the data found in the NotoriousMonsters table. The BNpcName contains the names while the BNpcBase contain their characteristics, which includes their rank. The BNpcName and BNpcBase ids are not the same.

While working with the hunt trackers, discord bots and how the data is represented, I learned that we identify them just by `id`, from which I correlated to coincide with the BNpcName table. I learned today how to identify their ranks automatically (thanks to this script), and wanted to contribute by adding this `id` into the structure, would renaming the `id` field I just added to `nameId` help?

rreminy: I sent you some additional information in DMs

rreminy: I realize I got myself confused again... The rank information is in the NotoriousMonster table, not the BNpcBase table, no clue what the BNpcBase table does but at least I learned we can use their IDs from the NotoriousMonster table to identify the SS rank monsters

quisquous: Ok thanks, that helps a lot for me to understand what the difference is between these two tables.

PR #1278, opened 2020-03-25
Raidboss: TEA headmarker 1 fix

### LINKED ISSUE #1266: Raidboss: TEA head marker fix wasn't perfect
[Per a comment on the pull request](https://github.com/quisquous/cactbot/pull/1206#issuecomment-602139160), the first number on the first Limit Cut does not get a callout. It's probably a race condition, so the easy solution is *probably* to just delay the Limit Cut callout by .25 seconds or something.

JLGarber:
Such an obvious thing looking at it now. Computers definitely do *exactly* what we tell them...

(Should resolve #1266.)


quisquous:
I didn't notice it the first time around either.  Very subtle.

PR #1279, opened 2020-03-25
KR v5.1 update

Jaehyuk-Lee:
1. changed `FFXIVProcessKo.cs` to match v5.1 offset.
2. changed `jobs` module to match v5.1 patch skill cooldown and duration time. (in KR and CN release)  
This is actually identical to v5.2, so I removed buffOverrides but remained as comment for future changes.
3. added official Korean translation to v5.1 update contents.

+ minor patch
'Scorching Right' is not used in timeline(Grand Cosmos), so removed it.

By the way, 
'Diffuse Laser' is translated to '확산 레이저' and '광확산 레이저'. I think this is like, (I haven't really seen this mechanic in game)
A B C is all called 'Diffuse Laser' in English client, 
but Korean client uses A B '확산 레이저' but C is '광확산 레이저'.
I've set js file to translate `Diffuse Laser` to `확산 레이저` and it would not be a big deal since those skill names ('확산 레이저' & '**광**확산 레이저') are similar.


JLGarber:
Diffuse Laser [looks *really* weird on XIVAPI](https://xivapi.com/search?indexes=action&string=diffuse+laser&columns=Name,Name_de,Name_fr,Name_ja,ID), so much so that I wonder whether this is a mistake by the localizers. Each of the different languages refers to a different version of the mechanic as "super/empowered/etc". 

For what it's worth, the one that's used in the trigger file is [18261/4755](https://xivapi.com/action/18261).  [18240/4740](https://xivapi.com/action/18240) also appears in the timeline as "Wide-Angle Diffusion Laser. (To be clear, @Jaehyuk-Lee, I know you probably already went through all that, but to save time for anyone else looking at this I thought I would compile the English links.)

plugin/CactbotEventSource/FFXIVProcessKo.cs
```diff
@@ -42,7 +42,7 @@ public unsafe struct EntityMemory {
       [FieldOffset(0xB0)]
       public Single rotation;
 
-      [FieldOffset(0x18A4)]
+      [FieldOffset(0x18B8)]
```
quisquous: Can you update the comment at the top as well for which version this applies to?

Jaehyuk-Lee:
@JLGarber so... you meant to add comment here? the link?
https://github.com/quisquous/cactbot/blob/e397de168f870d1182868a09345a0caa39249e1d/ui/raidboss/data/05-shb/alliance/the_copied_factory.txt#L261
I think I understood first paragraph fine, but I understood second as you're telling me there are "Diffuse Laser" and "Wide-Angle Diffuse Laser" on trigger side. However, there is no id: '4740' in `the_copied_factory.js` triggers.

JLGarber:
Oh! My apologies for the confusion. No, only "Diffuse Laser" appears in the triggers. "Wide Angle Diffuse Laser" is only in the timeline, as you said. I was mostly trying to save time for other English-speaking reviewers in terms of understanding why you translated those two terms the way you did.

quisquous:
Thanks for all of this!  I'm going to go ahead and do a release from this so that folks can use the memory updates and all these translations.  If there's more changes to make from here, just send more pull requests.  :)

PR #1280, opened 2020-03-25
Raidboss: A1S Fr

VanSlanzar:
Translation Fr missing.

for bait, I make a difference between:
-  attract a laser that follows you (attirez)
- place an area on the ground or other (placez-vous)

I don't know about the English client, but about the French client:
- This line
Liftoff
- This line
Resin

and other cast bar do not appear. On Faust yes, but on Oppressor no. I don't know, if it's normal.


VanSlanzar:
I just created the boss skill names. My skills are limited to this level of software.

VanSlanzar:
Kaltstrahl

I wanted to delete them thinking that the translation was useless. Bad idea, the check didn't go through any more, and that created other problems for me. I restored everything.

quisquous:
> Kaltstrahl
> 
> I wanted to delete them thinking that the translation was useless. Bad idea, the check didn't go through any more, and that created other problems for me. I restored everything.

Yes, sorry.  It seems kind of silly and useless to have them there, but without them there's no way to differentiate between "this hasn't been translated yet" and "this has been translated but it is the same".

For what it's worth, you can also run `npm run test` and `npm run lint` locally to do the same thing that the bot would tell you to do.

quisquous:
> Translation Fr missing.
> 
> for bait, I make a difference between:
> 
> * attract a laser that follows you (attirez)
> * place an area on the ground or other (placez-vous)

That makes sense to me.  I think in English the second one is "bait".  In copied factory `Copied Engels Guided Missile`, the first one is `Get Out + Dodge Homing AoE` which is a bit of an awkward phrase.

> I don't know about the English client, but about the French client:
> 
> * This line
>   Liftoff
> * This line
>   Resin
> 
> and other cast bar do not appear. On Faust yes, but on Oppressor no. I don't know, if it's normal.

Sorry, I don't know what you mean that they appear on Faust.  Are you saying that timelines in general don't appear? Or those particular timeline triggers about resin and liftoff aren't working?

I think this is maybe separate from this patch, so I will go ahead and merge this one.

quisquous:
Also also, thanks for the translations! :)

VanSlanzar:
Kaltstrahl

or and how should I make this order, it would actually help me.

quisquous:
You can install npm via https://nodejs.org/en/

I run `npm run test` from the root of my cactbot directory in a console. (I use MINGW64, which is what "git shell for windows" gives me.)

VanSlanzar:
**This is cast bar on Faust:**
Cast bar

**THis is on Oppressor: written warning, just for Hydrothermal OK, but no cast bar**
Hydro

**This is this is valid for all rest Oppressor: No warning written and no cast bar on fight.**
Resin

this is for the frenchg version, i don't know for the english version.

VanSlanzar:
Same for:
- A2S, A3S, i try other.

quisquous:
You need a `replaceSync` for the seal message.  In particular, you need a `replaceSync` entry for both `Machinery Bay 44` and `Hanger 8`.  Faust works because there's a sync on the first cleave.  There's no such sync on the first Royal Fount.  Let me add one for safety.

VanSlanzar:
- this solved the cast bar problem, but not the written warning (resin bomb). 

- for the cast bars, I would correct on the raid or trial where it does not work because I understood the sync, thank you.

quisquous:
> * this solved the cast bar problem, but not the written warning (resin bomb).

This is because there's no `regexFr` translation.  Take a look at other non-Alexander trigger files to see.  The timelineReplace section only applies to timelines right now.  I would like them to apply to the `regex` automatically in the future, but they do not yet.

> * for the cast bars, I would correct on the raid or trial where it does not work, I think I understood.
>   I hope.

Send a pull request, and we can chat about it there.  I'm happy to help you through it.

> * no sync here, it's normal ?
>   no sync a1s

Yes, that's a loop, where the phase repeats and so it is providing skills that look like they're ahead but then will match up with where it jumps to.

https://github.com/quisquous/cactbot/blob/master/docs/TimelineGuide.md#making-loops-loop

VanSlanzar:
For the announcement of resin bomb, I tried many combinations, but I did not succeed. Neither regexFr works, I look in other files, but I don't see how.

quisquous:
What does your `regexFr` line look like? What is the line in the log for resin bomb?

VanSlanzar:
regeFrx: Regexes.startsUsing({ id: 'E46', source: 'Oppresseur' }),

i tried this, like on a2s. But it doesn't work. I changed condition to ability and other things, to no avail.

quisquous:
I assume you mean `regexFr: Regexes.startsUsing({ id: 'E46', source: 'Oppresseur' }),`, right?

When you view the logs in ACT, what does the log line say? See: https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#viewing-logs-after-a-fight

quisquous:
This is not well-documented, but you can also use `ui/raidboss/raidemulator.html`.  If you open that up, then import an a2s log file, you can see which triggers fired in the timeline (by mousing over them) or by hitting play.  You can also hit "store" to store it in the browser storage so you don't have to import it again, make edits to your trigger file, reload the overlay, and then try again without having to go into an instance.

quisquous:
Oh! The issue is that Resin Bomb is E47 not E46.  Sorry.

VanSlanzar:
Thanks to you for finding. I think there is really a lot of possible things
to do, it's very interesting. But since I suck in English, I really hurt.
More documentation would help, indeed. But don't add extra workload.

Le sam. 4 avr. 2020 à 18:58, Adrienne Walker <notifications@github.com> a
écrit :

VanSlanzar:
You will laugh, but although this error is resolved, it is another problem
which must be the cause. Neither the infotext trigger for (gunnery pod) and
(resin bomb) works.
I am horrified.
Does it work in English?

Le dim. 5 avr. 2020 à 00:15, Jeremy bauer <gamerbauer1985@gmail.com> a
écrit :

PR #1281, opened 2020-03-26
Raidboss: O4N trigger updates

JLGarber:
Nothing major here, just some necessary cleanup. The Thunder 3 cleave warning is returning "avoid tank cleave" even for tanks, which is kind of weird. I haven't been able to figure out exactly why though, looking through `responses.js`, it *looks* like tanks are explicitly special-cased to just warn about cleaves.

### DIFF
--- a/ui/raidboss/data/04-sb/raid/o4n.js
+++ b/ui/raidboss/data/04-sb/raid/o4n.js
@@ -45,6 +45,7 @@
       regexJa: Regexes.startsUsing({ id: '24BA', source: 'エクスデス', capture: false }),
       regexCn: Regexes.startsUsing({ id: '24BA', source: '艾克斯迪司', capture: false }),
       regexKo: Regexes.startsUsing({ id: '24BA', source: '엑스데스', capture: false }),
+      suppressSeconds: 5,
       response: Responses.spread(),
     },
     {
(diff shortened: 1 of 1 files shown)


quisquous:
> Nothing major here, just some necessary cleanup. The Thunder 3 cleave warning is returning "avoid tank cleave" even for tanks, which is kind of weird. I haven't been able to figure out exactly why though, looking through `responses.js`, it _looks_ like tanks are explicitly special-cased to just warn about cleaves.

I think `tankCleave` should only do that if the cleave has a target and it's not you.  The logic (rewritten) should be:
```
if you're a tank, and there's no target at all:
  'tank cleave'
else if you're a tank and you're the target:
  'tank cleave on you'
else
  'avoid tank cleave'
```

...with the idea being that if it's on somebody else, you shouldn't be there.

JLGarber:
`14:24BD:Exdeath starts using Thunder III on Exdeath`. The actor ID in the network log is the same. (That is, the Exdeath actor casting is the same as the target actor, for both the `14` and the associated `15` lines.)

I'm mostly confused because the *very first condition* in `tankCleave()` is "are you a tank?". The logic of `tankCleave()` *looks* like it should be warning just a basic "Tank Cleave" for tanks, not the non-tank "Avoid".

quisquous:
> `14:24BD:Exdeath starts using Thunder III on Exdeath`. The actor ID in the network log is the same. (That is, the Exdeath actor casting is the same as the target actor, for both the `14` and the associated `15` lines.)

Ahhh yeah, I think the logic is bogus here.  I think instead of `if (matches)` it should be `if (target)`, as `getTarget` handles boss casting on itself.

> I'm mostly confused because the _very first condition_ in `tankCleave()` is "are you a tank?". The logic of `tankCleave()` _looks_ like it should be warning just a basic "Tank Cleave" for tanks, not the non-tank "Avoid".

Yeah, the implicit fall-through case is confusing.  Feel free to rewrite it for clarity? Sorry for the confusion.

JLGarber:
Oh, okay, you're right, I see where that is a mistake. Okay. I don't have time to work through that over lunch, I'm already putting together something else, but in the meantime, I think that means this request should be fine then. The error isn't in the O4N trigger file, so once I get around to rewriting `tankCleave()`, the Thunder 3 trigger should Just Work.

PR #1283, opened 2020-03-26
raidboss: e7s tornado spot callouts

nebularg:


nebularg:
Awkward, I've been running that for a few weeks with the wrong delay key without noticing a problem >.>

quisquous:
Thanks, this is great!

PR #1285, opened 2020-03-27
raidboss: have delaySeconds in triggers run before promise

quisquous:
In particular, delaySeconds runs, and the promise trigger field itself is not evaluated until after the delaySeconds.

Requested in #1269.

### DIFF
--- a/test/trigger/test_trigger.js
+++ b/test/trigger/test_trigger.js
@@ -311,11 +311,12 @@ let testTriggerFieldsSorted = function(file, contents) {
     'beforeSeconds',
     'condition',
     'preRun',
-    'promise',
     'delaySeconds',
     'durationSeconds',
     'suppressSeconds',
     // This is where the delay happens.
+    'promise',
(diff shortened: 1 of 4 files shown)


quisquous:
Does anybody have any comments on this? (or on trigger field ordering in general?) If no, that's also fine.  Just wanted to put this up for comment in case anybody did.

JLGarber:
Oh, I missed that you had updated certain elements of the triggers readme in this request. I'll have to make some changes to mine once you land this. (Nothing major, I had also shuffled the trigger elements listing for the same reason you did.)

As far as this pull request, I think the change makes sense. `delaySeconds` definitely should have priority over everything else once `preRun` finishes.

quisquous:
Thanks for the feedback.  I'll land this so you can sort out the merge issues (sorry!)

JLGarber:
Oh, no, I didn't mean to rush things for you, just that there was overlap that will need sorting out after. It's not a problem!

PR #1290, opened 2020-03-28
Raidboss: O4N Oopsy

JLGarber:
Nothing special about this one. There aren't a lot of mistakes to make in this fight! It's still missing Pyretic, since as I note in-line that isn't currently possible without some excessive effort, but other than that it should be basically complete.

### DIFF
--- a/ui/oopsyraidsy/data/04-sb/raid/o4n.js
+++ b/ui/oopsyraidsy/data/04-sb/raid/o4n.js
@@ -1,0 +1,65 @@
+'use strict';
+
+// O4N - Deltascape 4.0 Normal
+[{
+  zoneRegex: {
+    en: /^Deltascape \(V4\.0\)$/,
+  },
+  damageWarn: {
+    'Blizzard III': '24BC', // Targeted circle AoEs, Exdeath
+    'Empowered Thunder III': '24C1', // Untelegraphed large circle AoE, Exdeath
+    'Zombie Breath': '24CB', // Conal, tree head after Decisive Battle
+    'Clearout': '24CC', // Overlapping cone AoEs, Deathly Vine (tentacles alongside tree head)
+    'Black Spark': '24C9', // Exploding Black Hole
+  },
+  triggers: [
+    {
+      id: 'O4N Doom', // Kills target if not cleansed
+      gainsEffectRegex: gLang.kEffect.Doom,
+      deathReason: function(e) {
+        return { type: 'fail', name: e.targetName, reason: { en: 'Cleansers missed Doom!' } };
+      },
+    },
+    {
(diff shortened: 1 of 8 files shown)


ui/oopsyraidsy/data/04-sb/raid/o4n.js
```diff
@@ -0,0 +1,65 @@
...
+        return { type: 'fail', name: e.targetName, reason: { en: 'Pushed off!' } };
+      },
+    },
+    {
+      // Empowered Fire III inflicts the Pyretic debuff, which deals damage if the player
+      // moves or acts before the debuff falls. Unfortunately it doesn't look like there's
+      // currently a log line for this, so the only way to check for this is to collect
+      // the debuffs and then warn if a player takes an action during that time. Not worth it
+      // for Normal.
+      id: 'O4N Standard Fire',
```
quisquous: Does this do enough damage such that this is something people won't do? I know in savage you definitely had to stop, but I can't remember if this is a "heal through it" pyretic in normal mode.

ui/oopsyraidsy/data/04-sb/raid/o4n.js
```diff
@@ -0,0 +1,65 @@
...
+      condition: function(e) {
+        // Double taps only
+        return e.type != '15';
+      },
+      mistake: function(e) {
+        return { type: 'warn', name: e.targetName, text: { en: e.abilityName + ' (Buster)' } };
+      },
+    },
+    {
+      id: 'O4N Empowered Blizzard', // Room-wide AoE, freezes non-moving targets
```
quisquous: Everybody gets his with this, so you have to use the effect?

JLGarber: Correct. There is damage regardless, although I guess we could delay for two seconds while collecting vulnerability up stacks, then blame anyone who gets a stack in that interval. But Deep Freeze is unique to Empowered Blizzard, so I thought it might be an acceptable alternative.

ui/oopsyraidsy/data/04-sb/raid/o4n.js
JLGarber: Not for tanks, and for Shadowbringers-geared players, it's probably irrelevant. (Queues are too long currently for me to test it as damage.) I do know that even in the later tiers of Stormblood it was dangerous to ignore the debuff if we didn't know the healers would be on top of things.

quisquous: Yeah, I think triggers should generally assume synced ilvl.  If folks are unsynced and rolling through it, they can ignore triggers as they see fit.

PR #1291, opened 2020-03-28
Jobs: Add GP Ready Alarm

Jaehyuk-Lee:


Jaehyuk-Lee:
I was satisfied with this code, but I need to select a good "sound" for gp alarm. I never missed gp alarm with `phone.ogg` but I don't think this matches the situation, I'm okay but this but will be distributed to all of users, so.....
Is there someone who can select a good sound file for me? I'm very bad at selecting something.

Jaehyuk-Lee:
plus, this alarm will happen when you use GP Potion, too, with this code.
I think this would be okay, so I made it like that.

ui/jobs/jobs.js
```diff
@@ -2128,6 +2128,16 @@ class Bars {
     if (!this.o.gpBar) return;
     this.o.gpBar.value = this.gp;
     this.o.gpBar.maxvalue = this.maxGP;
+
+    // GP Alarm
+    if (this.gp < this.options.GpAlarmPoint) {
+      this.gpAlarmReady = true;
+    } else if (this.gpAlarmReady && this.gp >= this.options.GpAlarmPoint) {
```
quisquous: Re: potion.  Do you think it'd be possible to delay the gp alarm for 0.5-1 second, and cancel it if a cordial is used in the log?

quisquous:
This is neat!

> Is there someone who can select a good sound file for me? I'm very bad at selecting something.

`sonar.ogg` is what I used in Eureka for NMs popping.  That might be ok? I could also see `kaching.ogg`?

ui/jobs/jobs.js
Jaehyuk-Lee: Hmmm. I don't know how to read log...
I was thinking about checking GP different.

`if (this.gp < this.options.GpAlarmPoint) {`
to
`if (this.gp < this.options.GpAlarmPoint && this.gp >= this.options.GpAlarmPoint - 20) {`

so `this.gpAlarmReady = true;` won't happen before gp gets close to target GP point.

What you'd said is logically better I think, but I don't know how to read log.

Jaehyuk-Lee:
@quisquous I have a question, does cactbot object to using FFXIV sound effect? <se.1> thing.

Jaehyuk-Lee:
`sonar.ogg` is what I always hear when `/countdown`. Those two are always in different situation, so I think it's not bad, but I think this is not like "alarm" thing. It does not make you surprise, this feature does not necessarily make you surprise though. In my opinion, this is really... not making surprised.
`kaching.ogg` made me laugh, it makes sense yes, you saved enough GP 😆

ui/jobs/jobs.js
quisquous: jobs reads the log in `OnLogEvent`, for what it's worth.

quisquous:
> @quisquous I have a question, does cactbot object to using FFXIV sound effect? <se.1> thing.

I don't have a strong feeling here, but I would lean towards not.  I think it might be better to not be confusing about what's coming from in-game vs out-of-game?

Jaehyuk-Lee:
I'll go for `kaching.ogg` for now.
If you have better idea, make any comment please

Jaehyuk-Lee:
yeah, I always had to think of in game used se-chat macros. Those all using FFXIV sound effects were configurable, but cactbot is not, for now.
Some opinion was "why not uses FFXIV sound effect" from Korean FFXIV community.

ui/jobs/jobs.js
Jaehyuk-Lee: okay, I'll do some research tomorrow, I should sleep now. I'll do tomorrow.

ui/jobs/jobs_config.js
```diff
@@ -158,5 +158,23 @@ UserConfig.registerOptions('jobs', {
       type: 'integer',
       default: 1,
     },
+    {
+      id: 'GpAlarmPoint',
+      name: {
+        en: 'GP ready alarm (if 0, no alarm)',
```
g0shu: I'd suggest `GP alarm threshold (0 to disable)`

ui/jobs/jobs.js
g0shu: here's the regex I was using for detecting cordials when making a similar trigger for personal use, maybe it'll be helpful
`cordialRegex = new RegExp(e.detail.name + ':20(017FD|F5A3D|F844F|0420F|0317D):');`

Jaehyuk-Lee: I've found the log, and I can make this, but I don't know if this log is differ by game client version. 
```
[18:22:06.761] 15:1008A40B:블러드트레일:200317D:Item_317D:1008A40B:블러드트레일:E:1900000:0:0:0:0:0:0:0:0:0:0:0:0:0:0:18831:18831:0:0:0:1000:0.2368442:-0.08807932:3.814697E-06:-3.141593:18831:18831:0:0:0:1000:0.2368442:-0.08807932:3.814697E-06:-3.141593:00000625
[18:22:07.000] 00:082c:블러드트레일이 고성능 강장제를 사용했습니다.
[18:22:07.000] 00:08ad:⇒ 블러드트레일이 GP를 400 회복했습니다.
```
I am thinking about opcode part, which changes in every patches, on second line `082c` and on thrid line `08ad` is opcode I guess.
but is there no opcode on the first line?
If so, I can make that log for international use.

Jaehyuk-Lee: I've found two cordials id is `17FD` (NQ: 300 GP / HQ: 350GP) and `317D` (400 GP).
I think this trigger will happen when you eat food, too. I have never eat food for GP, but as far as I know, there is GP food.
```
configured GP alarm for 800
1. now GP : 790
2. eat food
3. now GP: 810
-> alarm rings
```
like this..
I don't think it's good to add all the food's id in regex. That makes developers add new ids on new patches.
So.... maybe make this problem just left in program.

Jaehyuk-Lee: or if opcode has no problem, it would be better to add chat regex.
`블러드트레일이 GP를 400 회복했습니다.` This is Korean battle chat log with GP healed(?).
So, 
add 
```
this.GpPotionRegex = function() {
  return /(이|가) GP를 (\y{Float}) 회복했습니다./;
};
```
on `resources\lang\ko.js` (and other languages)
```
if (log.search(e.detail.name + gLang.GpPotionRegex()) >= 0) {
  cancelGpAlarm();
}
```
on `ui\jobs\jobs.js`

Jaehyuk-Lee:
I've applied non-alarm on GP Potion usage (for next 2 seconds). regex referenced from @g0shu's code,  `cordialRegex = new RegExp(e.detail.name + ':20(017FD|F5A3D|F844F|0420F|0317D):');`

ui/jobs/jobs.js
```diff
@@ -2396,6 +2408,14 @@ class Bars {
             if (m)
               this.buffTracker.onUseAbility(m.groups.id, log);
           }
+          // use of GP Potion
+          let cordialRegex = new RegExp(this.me + ':20(017FD|F5A3D|F844F|0420F|0317D):');
```
quisquous: Can you make this Regexes.ability({source: this.me, id: '20(017FD|F5A3D|F844F|0420F|0317D)'})?

I agree that this will need to be updated in the future, but this also doesn't need to be translated.  Arguably, the game text for gaining GP might also change in the future (although probably less likely than adding new cordials)?

ui/jobs/jobs.js
```diff
@@ -2396,6 +2408,14 @@ class Bars {
...
+          if (log.match(cordialRegex)) {
+            this.GpPotion = true;
```
quisquous: (Can you name this `gpPotion` or `usedCordial` or some variable name that doesn't start with a capital letter? I know the options are all capitals, and I'll fix that eventually, but other new variables should definitely be lower case where possible.)

ui/jobs/jobs.js
Jaehyuk-Lee: I don't think game text will change. However, if you want this way, I'll do that way.

ui/jobs/jobs.js
Jaehyuk-Lee: ah, I was confused. That's a mistake.

PR #1292, opened 2020-03-28
Resources: Minor fix to Responses.tankCleave()

JLGarber:
Per [some previous discussion](https://github.com/quisquous/cactbot/pull/1281#issuecomment-604507116), this should remove any ambiguity in how the `tankCleave()` response works. Do we need to explicitly check that `target` is defined? I feel like no, but there might be something I'm not thinking of.


quisquous:
> Do we need to explicitly check that `target` is defined? I feel like no, but there might be something I'm not thinking of.

I think checking if the target is you is also sufficient for making sure it is defined.

Issue #1286, opened 2020-03-27
Please update information on writing custom trigger to reflect recent changes


Macapple96:
I love the recent changes to triggers, we now have functions to call such as `Regexes.startsUsing` or `Regexes.gainsEffect`, it will be easier to write triggers, but I couldn't find any updated information or list of these new functions; could you please update the examples? 💯

JLGarber:
Not a bad point, there is definitely some updating that could be done in the raidboss readme. I'll see about that this weekend.

PR #1293, opened 2020-03-29
Docs: Raidboss readme updates (helper function info)

JLGarber:
This should move us closer to resolution on #1286. This request as it stands does two things: It moves the order of the trigger elements in the document into conformity with the order found in the build tests, and it adds a small amount of information on helper functions and where to find them.

~~The edits to order are in a separate commit for ease of reversion if that's not desired at this time. I personally think it's best to ensure that all references to trigger elements are uniform in order, but I can see that it's slightly odd to have the `response` property before the popup text properties.~~ Rendered moot in #1285.

On the helper/replacement functions, I wasn't sure how much information we wanted to place in the triggers document. My thought was that maybe we could have a small amount of information here, and then at some later point a new separate unified document that covers all the helper functions could be created and linked to from here.

That all being said, I'm not sure about a lot of how I've presented this stuff, hence the draft status.


JLGarber:
After a little more looking at this, I think it's probably fine as it is. Unless there's a particular desire to keep the information on replacement functions here, I'll start on a separate document that will probably live in `cactbot/docs`.

quisquous:
I was actually thinking this doc should move to the docs/ folder.  It predates that folder.  Would it make sense to move it there and combine it with whatever you were thinking of writing separately?

JLGarber:
Oh, okay. I've wanted that myself, but thought there might have been a reason the triggers readme lived where it did. In that case, I think I'll put this one in storage briefly and instead do a separate request that just handles the location shuffling? This file isn't the only one that needs it, Oopsy also comes to mind, and there are probably a few others.

quisquous:
Yeah, doing this for oopsy would be good as well.  I think it's hard enough to find the docs/ folder, let alone look for some markdown deep in a folder tree.

JLGarber:
Okay, I'll work on fleshing out the replacement information for now. Maybe create a new section in `RaidbossGuide.md` dedicated to that?

docs/RaidbossGuide.md
```diff
@@ -63,29 +63,28 @@ Boolean, defaults to true. If true, timelines and triggers will reset automatica
...
+  durationSeconds: 3,
+  suppressSeconds: 0,
+  sound: '',
+  soundVolume: 1,
+   response: {
     alarmText: {'Alarm Popup'},
+    alertText: {'Alert Popup'},
+    infoText: {'Info Popup'},
     tts: {'TTS text'},
   },
```
quisquous: I know this was here before but `alarmText: {'Alarm Popup'}` is a bit ambiguous.  Maybe it should be either a string, or a function returning a string, but maybe not an invalid object.

(Also I think response shouldn't be indented?)

docs/RaidbossGuide.md
```diff
@@ -94,60 +93,75 @@ Boolean, defaults to true. If true, timelines and triggers will reset automatica
...
 Response has less priority than an explicity specified text or tts,
 and so can be overridden.
+(As with `regex` and `condition`, "canned" responses are available within [responses.js](https://github.com/quisquous/cactbot/blob/master/resources/responses.js).)
 
-**sound**
-Sound file to play, or one of 'Info', 'Alert', 'Alarm', or 'Long'. Paths to sound files are relative to the ui/raidboss/ directory.
-
-**soundVolume**
-Volume between 0 and 1 to play the sound associated with the trigger.
+**alarmtext**
```
quisquous: alarm*T*ext

docs/RaidbossGuide.md
```diff
@@ -63,29 +63,28 @@ Boolean, defaults to true. If true, timelines and triggers will reset automatica
...
   preRun: function(data, matches) { do stuff.. },
   promise: function(data, matches) { return promise to wait for resolution of },
-  infoText: {'Info popup'},
-  alertText: {'Alert Popup'},
-  alarmText: {'Alarm Popup'},
-  tts: {'TTS text'},
-  response: {
-    infoText: {'Info popup'},
-    alertText: {'Alert Popup'},
+  delaySeconds: 0,
```
quisquous: Can this function match the list at the bottom of this doc?

docs/RaidbossGuide.md
JLGarber: Looking at this more, do we really want to use a full object for `response` as opposed to a `Responses.foo()` function? Maybe 

```
response: Responses.doSomething(triggerSeverity),
alarmText: {en: 'Alarm Popup'},
alertText: {en: 'Alert Popup'},
infoText: {en: 'Info Popup'},
```
I'm not sure how often it's really going to happen that we return a localized object directly from `response` instead of a function? Or is there a more common use case I'm not thinking of?

(Someday copy/paste will be able to read our minds with respect to indents. Someday!)

JLGarber:
You're right, there's probably not a lot of point in a full document just for the helpers. I'll put a section at the end of the guide with a brief explanation and a few examples then.

PR #1294, opened 2020-03-29
Docs: Relocating the Oopsy and Raidboss readmes

JLGarber:
After a little poking around, it looks like these are the only two files that actually need to move to /docs. Happy to add any others that might be floating around while we're at it.


quisquous:
> I like to move it! Move it!

:P

JLGarber:
I probably should tighten up the levity on commit messages, but that one was too good to pass up, particularly on a 100% self-explanatory one.

quisquous:
The levity makes me laugh.  I don't think there's anything wrong with it.  :)

PR #1296, opened 2020-03-30
Docs: Small addition to Raidboss readme

JLGarber:
I'm so sorry, that bit about handling the `response` element didn't actually make it in, I was assuming it was still in question. This actually adds it, as well as localizing the `tts` element.


quisquous:
Whoops, sorry for misreading that comment on the other pull request.  Thanks for the followup.

PR #1302, opened 2020-03-30
fixed/added translations and regexes for shb raids

Akurosia:


ui/raidboss/data/05-shb/raid/e1s.js
```diff
@@ -437,7 +437,7 @@
       regexFr: Regexes.startsUsing({ id: '3D8D', source: 'Gardien Du Jardin' }),
       regexJa: Regexes.startsUsing({ id: '3D8D', source: 'エデン・ガーデナー' }),
       regexCn: Regexes.startsUsing({ id: '3D8D', source: '伊甸守护者' }),
-      regexKo: Regexes.startsUsing({ id: '3D8D', source: '에덴의 수호자' }),
+      regexKo: Regexes.startsUsing({ id: '3D8D', source: '에덴 정원사' }),
```
Una-P: I'm sorry but may I ask why you change this? They are translated to `'에덴의 수호자'`.

JLGarber: "에덴 정원사" is what appears in the dumped .csv file for NPC names located at https://github.com/Ra-Workspace/ffxiv-datamining-ko/blob/master/csv/BNpcName.csv. This is (so I understand) directly from the game files. Is the file mistaken?

Una-P: I,,,I'm not sure what happened, is it possible the name be different from datafile?
제목 없음
also '정원사' is misstranslated. '정원사'is gardener, '수호자' is guardian.. and we can't find nowhere '정원사' at e1s/e1n.
i'm so confused..

Una-P: tho I dont have data-mining technic,
https://github.com/Ra-Workspace/ffxiv-datamining-ko/blob/master/csv/BNpcName.csv is better correct than me..

JLGarber: It definitely does look like there's a mistake in the data mine, based on that screenshot. Is that a mistake someone would make easily? I think I would catch the difference between "gardener" and "guardian" in English, but I know nothing about Korean to know whether that is an easy mistake. Unfortunately it looks like the maintainers of that repository don't have issues enabled, so we can't ask them. 

@Jaehyuk-Lee, do you have any insights here?

Una-P: sss
Found this situation,translation has been modified at 5.1. `'정원사' `is right one NOW.

Una-P: To add a few words, we can also catch the difference between  '정원사'[jeong-won-sa] and '수호자'[su-ho-ja]  in Korea. but JP's original text is 'ガーデナー' [ga-de-na], This can be seen as gardner. And I think translation team decided to change his name as a gardener from a guardian at 5.1(last week).
But we didn't know, It is not specified on 5.1 patchnote, and you know, it's no important to clear dungeon.

Akurosia: Ok if i got irt correctly there is no need to change this one back right?

Una-P: yeap. Your line is correct. Tnx for change!

PR #1303, opened 2020-03-31
Common replacement many added all lang

VanSlanzar:
I have this to submit to you. I worked on it locally, and I managed to integrate everything. I noticed some lines integrated or not in some (js), and I don't understand why. I thought it would be easier to have them here. To do this, I need to delete these lines in about 27 files. I hope this will be useful to you. If you agree I would make these requests.

Affected files:
- arr: ifrit nm.
- Hw : fractal continuum, the vaults.
- Sb: eureka hydatos, royal city rabanastre, yojimbo, seiryu ex.
-Shb: amaurot, akademia anyder, holminster, twinning, the copied factory, innocence nm and ex, titania ex, hades ex, ruby weapon ex, e1n, e2n, e3n, e4n,e1s, e2s, e3s, e4s, e8s.


JLGarber:
I think this makes a lot of sense. I almost wonder whether it would be worth expanding this to include all the text from `responses.js` , so that we can separate the localized text from the response functions. I know we had discussed this a little in a couple of different pull requests, but expanding this makes sense to me.

quisquous:
This all looks great, thanks! I think it makes a lot of sense to put these in the common translations.

> I think this makes a lot of sense. I almost wonder whether it would be worth expanding this to include all the text from `responses.js` , so that we can separate the localized text from the response functions. I know we had discussed this a little in a couple of different pull requests, but expanding this makes sense to me.

That particular change to handle common response text seems better for a followup pull request, for sure.

JLGarber:
Oh, no question, I should have put it as "expanding it in future". When/if this lands, I'll take a look at what would be necessary to make that happen.

VanSlanzar:
If I understand correctly @JLGarber takes care of the files or do I have to make the requests?
Glad you are interested in my work.

ui/raidboss/common_replacement.js
```diff
@@ -28,18 +48,87 @@ let commonReplacement = {
...
+    ko: '--남동--',
+  },
+  '--southwest--': {
+    de: '--südwesten--',
+    fr: '--sud-ouest--',
+    ja: '--南西--',
+    cn: '--西南--',
+    ko: '--남서--',
+  },
+  '--sync--': {
```
Jaehyuk-Lee: As I know, --sync-- never show on timeline, so this does not need translation.

ui/raidboss/common_replacement.js
```diff
@@ -28,18 +48,87 @@ let commonReplacement = {
...
+    fr: '--saut--',
+    ja: '--ジャンプ--',
+    cn: '--跳--',
+    ko: '--점프--',
+  },
+  '--middle--': {
+    de: '--mitte--',
+    fr: '--milieu--',
+    cn: '--中间--',
+    ko: '--중간에--',
```
Jaehyuk-Lee: `ko: '--중간에--',` -> `ko: '--중앙--',`
is better I think

Jaehyuk-Lee: I thought this would be better `'--중앙--'`, but there was `'--center--'` upper there, which translates to `'--중앙--'` the same.
what's difference between 'middle' and 'center'? any examples?

mooondark: I think it depends of the shape of the arena. If it is a circular area, then _center_ is more accurate than _middle_.

Jaehyuk-Lee: then, this should be `'--중앙--'` rather than `'--중간에--'`.

VanSlanzar: - 중앙 for middle, right ? 
- for --sync, I suspected that it was not useful, but it will prevent it from being rewritten.

Jaehyuk-Lee: * 중앙 for middle **AND** center
* I think just removing all the "--sync--" thing written before is better, and tell people to not write it.

Jaehyuk-Lee: I didn't know center and middle is differently used in English.
Screenshot 2020-04-01 at 01 34 33
중앙 means both center and middle in Korean.

VanSlanzar: There is the same problem for the --reset-- lines, but I cannot integrate it because it exists in (skill-Boss) on O11. This line cannot be deleted and creates a conflict.

I correct for Ko lang.

VanSlanzar: For the rest, I think all the repetitive lines useful or not should be here. So we will be sure not to see them appear elsewhere.

JLGarber:
That's correct, @VanSlanzar, once this current request you're working on is done I'll move on to the next part.

VanSlanzar:
@JLGarber, to finish this request, i wonder how to do it. Do I have to make 27 different requests? or can I add them to this query, but I don't know how.

VanSlanzar:
well, i'm done XD. It was hard, I'm glad it worked ^^. I hope this will be of great help to you. Thank you all for your help.

ui/raidboss/common_replacement.js
```diff
@@ -15,6 +15,26 @@ let commonReplacement = {
     ja: 'の封鎖が解かれた',
     ko: '의 봉쇄가 해제되었습니다',
   },
+  '--center--': {
```
quisquous, line 18: `de: '--mitte-',`

ui/raidboss/common_replacement.js
```diff
@@ -28,18 +48,87 @@ let commonReplacement = {
...
-    cn: '--无法选中--',
-    ko: '--타겟 불가능--',
+  '--jump--': {
+    de: '--sprung--',
+    fr: '--saut--',
+    ja: '--ジャンプ--',
+    cn: '--跳--',
+    ko: '--점프--',
+  },
+  '--middle--': {
```
quisquous, line 58: Could you add something for Japanese here? Probably `中央` from `--center--` is fine.  I don't know that there's really a difference between center and middle in any timeline.

ui/raidboss/common_replacement.js
VanSlanzar: it's ok.

ui/raidboss/common_replacement.js
VanSlanzar: It's ok . 
In French, this is a synonym. It depends on the shape of the arena (round or rectangular).

VanSlanzar:
it's ok for 2 suggestions.
In French, this is a synonym. It depends on the shape of the arena (round or rectangular).
This is not easy for other languages because I do not know if you have this same difference. Thank you.

Akurosia:
the german cardianls/intercardinals are not consistent.
they are (starting north and going clockwise): norden, nordosten, osten, südosten, süden, südwesten, westen, nordwesten

otherwise the rest of german looks ok (also we do not make a difference if the arena is square or a circle. "Mitte" is correct in both cases

VanSlanzar:
French is so complicated, especially when you are picky like me. (as @mooondark  told me), and it becomes a headache.
TY very much @Akurosia.

VanSlanzar:
If no one else has any suggestions, i think the query may be (merge).

PR #1305, opened 2020-04-02
WIP: E8S Hallowed Wings

mrdregan:
Another day, another glorious port from triggernometry (not sure if anybody gets that reference):

See https://github.com/mrdregan/triggernometry_triggers/blob/master/E8S/Shiva_Hallowed_Wings.xml for source.

This here still misses the clockwise/counter-clockwise information (and is obviously still untested), but I wanted to start a discussion already on 2 things:
1. as of now it calls numbers. We could go for directions, but the timing is pretty tight and it would make the callout too long. Any ideas or stick to directons?
2. how do we explain "left" "right" announcements? Explanation what to do: When mirrors spawn, you need to place your camera TOWARDS red mirror and have the blue mirror BEHIND your camera. Then just follow the instruction based on your camera.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e8s.js
+++ b/ui/raidboss/data/05-shb/raid/e8s.js
@@ -385,6 +385,97 @@
       },
     },
     {
+      id: 'E8S Mirror, Mirror',
+      regex: Regexes.startsUsing({ source: 'Shiva', id: '4D5A', capture: false }),
+      regexDe: Regexes.startsUsing({ source: 'Shiva', id: '4D5A', capture: false }),
+      regexFr: Regexes.startsUsing({ source: 'Shiva', id: '4D5A', capture: false }),
+      regexJa: Regexes.startsUsing({ source: 'シヴァ', id: '4D5A', capture: false }),
+      regexCn: Regexes.startsUsing({ source: '希瓦', id: '4D5A', capture: false }),
+      regexKo: Regexes.startsUsing({ source: '시바', id: '4D5A', capture: false }),
+      condition: function(data) {
+        data.mirrorCast = data.mirrorCast || 0;
+        data.mirrorCast++;
+        data.isHallowedWingsMirrorMirror = data.mirrorCast === 3;
+
+        return data.isHallowedWingsMirrorMirror;
+      },
+      delaySeconds: 12,
+      promise: function(data) {
+        let p = new Promise(async (res) => {
+          const shivaLocaleNames = {
+            en: 'Shiva',
+            de: 'Shiva',
+            fr: 'Shiva',
+            ja: 'シヴァ',
+            cn: '希瓦',
+            ko: '시바',
+          };
+
+          let combatantData = await window.callOverlayHandler({
+            call: 'getCombatants',
+            names: [shivaLocaleNames[data.lang]],
+          });
+
+          // if we could not retrieve combatant data, the
+          // trigger will not work, so just resume promise here.
+          if (!(combatantData !== null &&
+            combatantData.combatants &&
+            combatantData.combatants.length)) {
+            data.safeZone = null;
+            res();
+            return;
+          }
+
+          // if we dont have information about which wing was
+          // casted, then the trigger will not work (safety measure)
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -384,6 +384,95 @@
...
+            sideResponse = Responses.goRight().alertText;
+            if (shivaCombatant.Heading > 3)
+              ttsNumbers = '1 4 ';
+          }
+
+          if (data.rightWing) {
+            sideResponse = Responses.goLeft().alertText;
+            if (shivaCombatant.Heading > 3)
+              ttsNumbers = '2 3 ';
+          }
```
valarnin, line 456: Might be a good idea to `Math.abs` the Heading value before comparison, e.g. `if (Math.abs(shivaCombatant.Heading) > 3)`. Might not be 100% necessary but heading ranges from -π to π (-3.14... to 3.14...) and the direct north/south cardinals can vary depending on rounding errors.

mrdregan: Nice catch, got that in the triggernometry trigger but forgot that to implement here aswell.

quisquous:
> Another day, another glorious port from triggernometry (not sure if anybody gets that reference):

(what's it from? internet searching did not help me)

> 1. as of now it calls numbers. We could go for directions, but the timing is pretty tight and it would make the callout too long. Any ideas or stick to directons?

I've tried to avoid number callouts, because everybody puts down markers differently.  I did dragon drives for A, B, C but that felt pretty universal (at least to me).  What about left forward / left back / right forward / right back?

> 2. how do we explain "left" "right" announcements? Explanation what to do: When mirrors spawn, you need to place your camera TOWARDS red mirror and have the blue mirror BEHIND your camera. Then just follow the instruction based on your camera.

Yeah, that's my thought too.  "Face away from blue mirror" or "Face red mirror" and then the other calls after that.

mrdregan:
> (what's it from? internet searching did not help me)

It's a voice line from Brutallus (a wow boss): Another day, another glorious battle!

> Yeah, that's my thought too. "Face away from blue mirror" or "Face red mirror" and then the other calls after that.

That's a good idea. I will add that.

> I've tried to avoid number callouts, because everybody puts down markers differently. I did dragon drives for A, B, C but that felt pretty universal (at least to me). What about left forward / left back / right forward / right back?

We turn the camera around beforehand, so generally this can get quite messy. Also the boss is turned either north or south. Maybe we could make something like: "left side of boss" (is safe) + "go right/left quadrant" (based on camera)?

quisquous:
Sorry for the belated review speed here.  It's been a week.

> > (what's it from? internet searching did not help me)
> 
> It's a voice line from Brutallus (a wow boss): Another day, another glorious battle!

Aha! I uh...never played wow <_<

> > I've tried to avoid number callouts, because everybody puts down markers differently. I did dragon drives for A, B, C but that felt pretty universal (at least to me). What about left forward / left back / right forward / right back?
> 
> We turn the camera around beforehand, so generally this can get quite messy. Also the boss is turned either north or south. Maybe we could make something like: "left side of boss" (is safe) + "right/left quadrant" (based on camera)?

What about:
(1) face away from blue mirror, and then...
(2) left and forward / left and back / right and forward / right and back

...and you can take into account the facing of the boss when making that call?

I would argue that the existing trigger already does what you are proposing.  If you face away from the blue mirror, then the existing trigger will say "right" and this means both "right quadrant is safe" and "right side of the boss is safe".  I mean, this isn't maybe the most obvious, but that's how I use it, in the absence of smarter triggers.

quisquous:
Sorry for the incredible delay in getting back to this, but I think I just was never quite happy with what this added over the trigger that existed that just said "left" (since it's your left + shiva's left if your back was to the blue mirror) and didn't feel like "go forward and right" would be easy to understand.

I appreciate the triggernometry trigger port.  Please send more of these in the future if you have them.  However, at this point (and especially give that this tier is pretty over), I think I'm just going to close this patch without merging.

PR #1312, opened 2020-04-05
e8s: Add "Sharingan/Ayatori" Light Rampant strat

valarnin:
This PR:
- Switches the regex for Refulgent Chain to the tether
- Adds an option to config for which Light Rampant strat to use
- Adds strat for sharingan/ayatori (bowtie, orbs opposite of spawn location)

This isn't fully tested due to my group not running today. I won't be able to properly test this until maybe Thursday depending on how re-clears go.

This should be extensible to allow for other strats to be added, I'm just not familiar with any other strats to add them.

`006E` is the initial tether. `006F` is the second tether that snaps in place partway into the mechanic when the distance between tethers snapshots.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e8s.js
+++ b/ui/raidboss/data/05-shb/raid/e8s.js
@@ -322,11 +322,12 @@
       id: 'E8S Refulgent Chain',
       regex: Regexes.tether({ id: '006E' }),
       condition: function(data, matches) {
-        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'None')
+        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'none')
           return data.me == matches.source;
 
         if (data.options.cactbote8sLightRampantStrat === 'sharingan') {
-          data.rampant = data.rampant || {
+          data.rampant = data.rampant || {};
+          data.rampant.chain = data.rampant.chain || {
             chains: {},
             total: 0,
             meChain: false,
@@ -334,44 +335,44 @@
             chainSwapDps: 'Stay',
           };
           if (data.party.isTank(matches.source)) {
-            data.rampant.chains.tank = matches;
+            data.rampant.chain.chains.tank = matches;
           } else {
             if (data.party.isHealer(matches.source)) {
-              data.rampant.chains.healer = matches;
+              data.rampant.chain.chains.healer = matches;
             } else {
-              if (data.rampant.chains.dps1 === undefined)
-                data.rampant.chains.dps1 = matches;
+              if (data.rampant.chain.chains.dps1 === undefined)
+                data.rampant.chain.chains.dps1 = matches;
               else
-                data.rampant.chains.dps2 = matches;
+                data.rampant.chain.chains.dps2 = matches;
             }
             if (data.party.isTank(matches.target))
-              data.rampant.chains.tankReverse = matches;
+              data.rampant.chain.chains.tankReverse = matches;
           }
-          ++data.rampant.total;
-          data.rampant.meChain = data.rampant.meChain || data.me == matches.source;
-          if (data.rampant.total===4) {
+          ++data.rampant.chain.total;
+          data.rampant.chain.meChain = data.rampant.chain.meChain || data.me == matches.source;
+          if (data.rampant.chain.total===4) {
             // First figure out which DPS is in which corner
             // So we can ensure dps1 is southwest and dps2 is southeast
             let meleePriority = ['MNK', 'DRG', 'NIN', 'SAM'];
             let rangedPriority = ['BRD', 'MCH', 'DNC', 'BLM', 'SMN', 'RDM'];
-            let dps1Role = data.party.jobName(data.rampant.chains.dps1.source);
-            let dps2Role = data.party.jobName(data.rampant.chains.dps2.source);
+            let dps1Role = data.party.jobName(data.rampant.chain.chains.dps1.source);
+            let dps2Role = data.party.jobName(data.rampant.chain.chains.dps2.source);
             let dps1Priority = meleePriority.indexOf(dps1Role);
             let dps2Priority = meleePriority.indexOf(dps2Role);
             if (dps1Priority >= 0 && dps2Priority >= 0) {
               // Both melee, so go on priority
               if (dps1Priority > dps2Priority) {
                 // Swap, they're out of order
-                let tmp = data.rampant.chains.dps1;
-                data.rampant.chains.dps1 = data.rampant.chains.dps2;
-                data.rampant.chains.dps2 = tmp;
+                let tmp = data.rampant.chain.chains.dps1;
+                data.rampant.chain.chains.dps1 = data.rampant.chain.chains.dps2;
+                data.rampant.chain.chains.dps2 = tmp;
               }
             } else if (dps1Priority >= 0 || dps2Priority >= 0) {
               if (dps2Priority >= 0) {
                 // Swap, they're out of order
-                let tmp = data.rampant.chains.dps1;
-                data.rampant.chains.dps1 = data.rampant.chains.dps2;
-                data.rampant.chains.dps2 = tmp;
+                let tmp = data.rampant.chain.chains.dps1;
+                data.rampant.chain.chains.dps1 = data.rampant.chain.chains.dps2;
+                data.rampant.chain.chains.dps2 = tmp;
               }
             } else {
               // Both ranged, so go on priority
@@ -379,41 +380,41 @@
               dps2Priority = rangedPriority.indexOf(dps2Role);
               if (dps1Priority > dps2Priority) {
                 // Swap, they're out of order
-                let tmp = data.rampant.chains.dps1;
-                data.rampant.chains.dps1 = data.rampant.chains.dps2;
-                data.rampant.chains.dps2 = tmp;
+                let tmp = data.rampant.chain.chains.dps1;
+                data.rampant.chain.chains.dps1 = data.rampant.chain.chains.dps2;
+                data.rampant.chain.chains.dps2 = tmp;
               }
             }
             // All positions based on tank tethers
             let tankTether1 = (
-              data.rampant.chains.tank.target === data.rampant.chains.healer.source ? 'NE' : (
-                data.rampant.chains.tank.target === data.rampant.chains.dps1.source ? 'SW' : 'SE'
+              data.rampant.chain.chains.tank.target === data.rampant.chain.chains.healer.source ? 'NE' : (
+                data.rampant.chain.chains.tank.target === data.rampant.chain.chains.dps1.source ? 'SW' : 'SE'
               )
             );
             let tankTether2 = (
-              data.rampant.chains.tankReverse.source === data.rampant.chains.healer.target ? 'NE' : (
-                data.rampant.chains.tankReverse.source === data.rampant.chains.dps1.target ? 'SW' : 'SE'
+              data.rampant.chain.chains.tankReverse.source === data.rampant.chain.chains.healer.target ? 'NE' : (
+                data.rampant.chain.chains.tankReverse.source === data.rampant.chain.chains.dps1.target ? 'SW' : 'SE'
               )
             );
             if ((tankTether1 === 'NE' && tankTether2 === 'SW') ||
                 (tankTether1 === 'SW' && tankTether2 === 'NE')) {
               // If the tank is tethered to NE and SW (makes a box)
               // Tank swaps with SW to make bowtie
-              data.rampant.chainSwapTank = 'SW';
-              data.rampant.chainSwapDps = data.rampant.chains.dps1.source;
+              data.rampant.chain.chainSwapTank = 'SW';
+              data.rampant.chain.chainSwapDps = data.rampant.chain.chains.dps1.source;
             } else if ((tankTether1 === 'NE' && tankTether2 === 'SE') ||
                 (tankTether1 === 'SE' && tankTether2 === 'NE')) {
               // If the tank is tethered to NE and SE (makes an hourglass)
               // Tank swaps with SE to make bowtie
-              data.rampant.chainSwapTank = 'SE';
-              data.rampant.chainSwapDps = data.rampant.chains.dps2.source;
+              data.rampant.chain.chainSwapTank = 'SE';
+              data.rampant.chain.chainSwapDps = data.rampant.chain.chains.dps2.source;
             }
           }
-          return data.rampant.total===4;
+          return data.rampant.chain.total===4;
         }
       },
       infoText: function(data, matches) {
-        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'None') {
+        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'none') {
           return {
             en: 'Chain on YOU',
             de: 'Kette auf DIR',
@@ -425,15 +426,15 @@
 
         if (data.options.cactbote8sLightRampantStrat === 'sharingan') {
           let ret = {};
-          if (data.rampant.meChain) {
+          if (data.rampant.chain.meChain) {
             let extra = '';
             if (data.rampant.chainSwapTank!=='Stay') {
-              if (data.rampant.chains.tank.source == data.me)
-                extra = ', Swap '+data.rampant.chainSwapTank;
+              if (data.rampant.chain.chains.tank.source == data.me)
+                extra = ', Swap '+data.rampant.chain.chainSwapTank;
               else if (data.rampant.chainSwapDps == data.me)
                 extra = ', Swap NW';
               else if (data.party.isHealer(data.me))
-                extra = ', Tank swap with '+data.rampant.chainSwapTank;
+                extra = ', Tank swap with '+data.rampant.chain.chainSwapTank;
             }
             ret = {
               en: 'Chain on YOU'+extra,
@@ -453,7 +454,7 @@
       regex: Regexes.tether({ id: '0002' }),
       condition: Conditions.targetIsYou(),
       promise: function(data, matches) {
-        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'None')
+        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'none')
           return null;
 
         let p = new Promise(async (res) => {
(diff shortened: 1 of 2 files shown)


valarnin:
This is still in testing. I wanted to add triggers for Lightsteeped stacks but it doesn't look like all the required information is available in the version of the `1A:`/`26|` log line that gets passed to the overlay.

These are the lines that gets written to the Network file, starred out player ID/name/hp, first shows a 1-stack and second shows a 2-stack:
```
26|2020-04-02T23:25:51.2690000-04:00|8d1|Lightsteeped|39.95|E0000000||********|**** ****|01|******|||0df4cd7b1b08c32eb86e6a5307039a64
26|2020-04-02T23:25:51.2690000-04:00|8d1|Lightsteeped|39.95|E0000000||********|**** ****|02|******|||58f1bbbcaa76979d19e147bfaf9e3b71
```

Actual log lines that get passed to overlays seem to be textual instead and don't contain the stack count, at least according to this:

https://github.com/quisquous/cactbot/blob/master/resources/regexes.js#L171-L189

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+      run: function(data, matches) {
+        if (data.options.cactbote8sLightRampantStrat === undefined ||
+          data.options.cactbote8sLightRampantStrat === 'none' ||
+          data.rampantOrb === null)
+          return;
+
+        if (data.options.cactbote8sLightRampantStrat === 'sharingan') {
+          // Take orb to opposite side of spawn
+          if (data.rampantOrb.PosX > 105)
+            data.rampantOrbText = 'W';
```
quisquous, line 505: I think 'W' on its own is maybe unclear.  I think also this makes translations tricky to append raw text like this.  Could this section be combined with infoText above, and maybe do something akin to:

```
if (no strat) {
  return { orb on you }
} else if (data.rampantOrb.PosX > 105) {
  return { bait orb W }
} else if (etc) {
} else {
  return { Bait Orb (???) }
}
```

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'none')
+          return null;
+
+        let p = new Promise(async (res) => {
+          let combatantData = await window.callOverlayHandler({
+            call: 'getCombatants',
+            ids: [matches.sourceId],
+          });
+          data.rampant = data.rampant || {};
+          if (!(combatantData !== null &&
```
quisquous, line 463: Can you apply this "not" through and make it combatantData === null || !combatantData.combatants || !combatantData.combatants.length to make this easier to read?

quisquous:
> This is still in testing. I wanted to add triggers for Lightsteeped stacks but it doesn't look like all the required information is available in the version of the `1A:`/`26|` log line that gets passed to the overlay.

Yeah, sorry.  This is because cactbot uses act log lines and not network log lines.  I am not sure what to do about this.  I probably should file an issue, because this needs to get sorted eventually, but the options look like:

(1) provide network log lines in addition to act log lines (cons: now you're parsing 2x as much data for every trigger
(2) replace act log lines with network log lines (cons: a bunch of work, also risky, maybe confusing to developers)
(3) implement hypothetical new log injection technique to add this info in, continue to use act log lines (cons: farther in the future, also a bunch of work)

I think #2 is plausible.  Timelines use raw act log lines, but they're all faiiiirly normalized and we could...convert them magically into network log lines, and use tests to verify that they can be converted as such.  Most other things use regex helpers, which could be converted to network log lines too.

This would also solve the zone id issue and effect id issues (to avoid translations on both of them), would provide positioning information on more lines, and some other good stuff.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+            if (data.party.isHealer(matches.source)) {
+              data.rampant.chain.chains.healer = matches;
+            } else {
+              if (data.rampant.chain.chains.dps1 === undefined)
+                data.rampant.chain.chains.dps1 = matches;
+              else
+                data.rampant.chain.chains.dps2 = matches;
+            }
+            if (data.party.isTank(matches.target))
+              data.rampant.chain.chains.tankReverse = matches;
```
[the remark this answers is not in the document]
JLGarber: From a quick check through my logs, this is the case, but the order is not consistent between the initial `006E` warning tethers and the `006F` danger tethers, if that matters.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+              // Tank swaps with SE to make bowtie
+              data.rampant.chain.chainSwapTank = 'SE';
+              data.rampant.chain.chainSwapDps = data.rampant.chain.chains.dps2.source;
+            }
+          }
+          return data.rampant.chain.total===4;
+        }
+      },
+      infoText: function(data, matches) {
+        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'none') {
```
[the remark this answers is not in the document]
JLGarber: From a 2- minute look, no. This info text should display only if there's no strategy defined, and in that event, the beginning of the condition *should* prevent the entire trigger activating more than once. (We've already established that we're confident there's exactly 1 tether that any chained player is the source for.)

(That being said, I'm not sure why we're checking both `undefined` and `none`, rather than doing something like `if (!data.options.cactbote8sLightRampantStrat)`.)

Also, maybe instead of stuffing all this into the condition we could do it as a `preRun`?

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
-        de: 'Kette auf DIR',
-        fr: 'Chaine sur VOUS',
-        ko: '사슬 대상자',
-        cn: '连线',
+      regex: Regexes.tether({ id: '006E' }),
+      condition: function(data, matches) {
+        if (data.options.cactbote8sLightRampantStrat === undefined || data.options.cactbote8sLightRampantStrat === 'none')
+          return data.me == matches.source;
+
+        if (data.options.cactbote8sLightRampantStrat === 'sharingan') {
```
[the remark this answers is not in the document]
valarnin: I'm not up to speed on the various chain and orb strats going around. For chain, this could be referred to as "hourglass", but I'm not sure what to call the orb strat then. And given that, does it make sense as well to have a separate setting for the conals/earthshakers/whatever we're calling them these days?

ui/raidboss/data/05-shb/raid/e8s.js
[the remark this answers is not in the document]
valarnin: > (That being said, I'm not sure why we're checking both `undefined` and `none`, rather than doing something like `if (!data.options.cactbote8sLightRampantStrat)`.)

`undefined` was a 'just in case that option hasn't been set' check, since I'm not sure exactly how Cactbot handles initializing those values if there's already a config present. The first version of this code block was a switch/case with the default being to return the 'chain on you' object, but that was ugly from a code readability standpoint.

> Also, maybe instead of stuffing all this into the condition we could do it as a preRun?

I don't know how feasible it would be to do this in a `preRun`. We need to do this logic regardless but we don't want to run the trigger fully unless we're at 4 triggers, and there's no way to abort a trigger after the `condition` that I've seen thus far.

I wish there was a way to move logic like this out to its own file, honestly.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+                data.rampant.chain.chains.dps1 = matches;
+              else
+                data.rampant.chain.chains.dps2 = matches;
+            }
+            if (data.party.isTank(matches.target))
+              data.rampant.chain.chains.tankReverse = matches;
+          }
+          ++data.rampant.chain.total;
+          data.rampant.chain.meChain = data.rampant.chain.meChain || data.me == matches.source;
+          if (data.rampant.chain.total===4) {
```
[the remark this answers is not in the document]
valarnin: These changes have been made in my local git.

ui/raidboss/data/05-shb/raid/e8s.js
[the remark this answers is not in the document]
valarnin: I observed that the order is random as well, that's why the code is written in the way it is (check the source class for the chain, and check the target to figure out where the tank's other tether is from). The one source per person pattern held true across the 25 pulls I checked through hoping for a better pattern to emerge.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+            // First figure out which DPS is in which corner
+            // So we can ensure dps1 is southwest and dps2 is southeast
+            let meleePriority = ['MNK', 'DRG', 'NIN', 'SAM'];
```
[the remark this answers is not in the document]
valarnin: I like this a lot more than my current code, I'll do some testing. Not sure about the change of return text, since not everyone can run Cactbot (we have a tank and a DPS both on PS4, if they have to swap then there's no way for the person doing callouts to know that from Cactbot itself).

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
       id: 'E8S Holy Light',
       regex: Regexes.tether({ id: '0002' }),
       condition: Conditions.targetIsYou(),
-      infoText: {
-        en: 'Orb on YOU',
-        de: 'Orb auf DIR',
-        fr: 'Orbe sur VOUS',
-        ko: '구슬 대상자',
-        cn: '拉球',
+      promise: function(data, matches) {
```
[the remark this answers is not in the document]
valarnin: Comment added in my local git.

ui/raidboss/data/05-shb/raid/e8s.js
valarnin: Adjusted in local git.

ui/raidboss/data/05-shb/raid/e8s.js
valarnin: I've reworked this but I'm not sure how to handle the translations. Should I just leave other languages out, or default it to the "orb on you" translation? Or run the text through google translate for the languages that were already present for the "orb on you" text?

ui/raidboss/data/05-shb/raid/e8s.js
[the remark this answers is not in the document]
JLGarber: I'm probably missing something obvious, but in that case, I feel like the logic could be split off into the pre-run after incrementing the chain counter. It looks like the bulk of the logic there doesn't actually run anyway unless the condition would be met, so I'm not sure what the loss would be. (Obviously this has already been written as it is, but if it's "just" a matter of pulling lines 350-409 into the pre-run, I feel like that would be worthwhile.)

Good point on the config being present though, I forgot that this isn't guaranteed to be falsy if there's no strategy selected.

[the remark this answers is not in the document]
valarnin: The trigger execution process is essentially this:
```
if (condition returns true) {
    execute preRun
    execute infoText
    execute run
}
```
Therefore, we'd still need to store off the info into data, then increment the chain counter, and then in preRun/infoText/run/whatever we'd need to early out for cases where it isn't this strat. Either way it's irrelevant with the latest commit where the logic has been moved into response (which replaces infoText).

ui/raidboss/data/05-shb/raid/e8s.js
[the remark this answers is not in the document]
valarnin: This has been rewritten based on your proposal, fixed some bugs and adjusted the return conditions.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,201 @@
...
+
+        return {
+          en: 'Orb on YOU, '+data.rampantOrbText,
+          de: 'Orb auf DIR, '+data.rampantOrbText,
+          fr: 'Orbe sur VOUS, '+data.rampantOrbText,
+          ko: '구슬 대상자, '+data.rampantOrbText,
+          cn: '拉球, '+data.rampantOrbText,
+        };
+      },
+      run: function(data, matches) {
```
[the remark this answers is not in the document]
valarnin: This is resolved in the latest commit.

quisquous:
Thanks, I'll take a look this weekend when I have time to look a little bit more in depth.

Do you think a promise returning null should just cancel the trigger?

valarnin:
> Do you think a promise returning null should just cancel the trigger?

I'm a bit torn on this. I'd almost want to say that a promise returning anything other than a promise should cancel the trigger, but then the "default state" case here becomes a hassle as seen in the code. Maybe it should be a truthy check, something along the lines of this?

```
if (return value is a promise)
  attach to the promise
else if (return value is truthy)
  continue trigger execution
else
  abort trigger
```

Or maybe even a hard check against `true` and have the default state be to abort.

ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -317,30 +317,173 @@
...
+          // If we need to move, alert us, otherwise just show info
+          if (alertFor.includes(data.me)) {
+            return {
+              alertText: ret,
+            };
+          }
+          // This is intentionally showing for all players
+          // not just players with chains, for callout purposes
+          return {
+            infoText: ret,
```
quisquous, line 395: I worry that this could be confusing, because calling out "Chain: Tank NW" or something for a non-chain player might make an orb player thing that they were a chain player?

Would it make sense to have something like:
- if you have a chain, alert Chain on YOU: Go/Swap NW
- if you don't have a chain, info "Chain Tank: NW"

valarnin: I went with just alerting on the tank chain position because they're the position that's always going to move to adjust to bowtie formation if an adjustment is needed. Everyone else should already know where they go for this mechanic (healer is always northeast, etc).

I could adjust the output text to show the swap text for the DPS that gets chained, my concern is if the healer is the one doing the callouts for the fight they won't see the tank position notification.

quisquous: It's not just the tank that needs to move though, it's a dps who needs to know if they are staying or swapping as well as the tank.  I'll change my proposal to:

- if you have a chain, alert Chain on YOU: Go/Swap NW/NE/SE/SW
- everybody, info "Chain Tank: NW"

And maybe disable tts for the "chain on you" one so they don't overlap and you still get the chain tank callout.

valarnin: If I return both `infoText` and `alertText` from the same `response` function, do they both get shown?

quisquous: Yeah, they both get shown.  Only the alert noise gets played.  The tts will be taken from the alert unless specified explicitly.

quisquous:
> > Do you think a promise returning null should just cancel the trigger?
> 
> I'm a bit torn on this. I'd almost want to say that a promise returning anything other than a promise should cancel the trigger, but then the "default state" case here becomes a hassle as seen in the code. Maybe it should be a truthy check, something along the lines of this?
> 
> ```
> if (return value is a promise)
>   attach to the promise
> else if (return value is truthy)
>   continue trigger execution
> else
>   abort trigger
> ```

The truthy check is a bit iffy to me.  I think I'd rather have truthy values throw an error (and also abort).  Promise or nothing are really the only two cases here.  If a developer wanted to just continue trigger execution, they could just resolve the promise.  Having a second path to do that seems confusing, at best.

valarnin:
Just rebased this to resolve the merge conflicts. Going to make the adjustments requested tomorrow most likely.

ui/raidboss/data/05-shb/raid/e8s.js
liaoarden: Does ayatori actually make the tanks do everything? 

Looking at the top [ayatori video](https://www.youtube.com/watch?v=pmYgfT3Dbno) they do all swaps by the southwest. We stuck the ranged there for our group so we always make them adjust (it's free real estate!).

image

(Also i'm crying in being rDPS and doing all the callouts...)

valarnin: The three different explanations I saw (toolbox link, text guide that I've lost, and YT video that is probably somewhere in my YT history) all had tank swapping with SW or SE DPS.

liaoarden: Hmm, this seems like a sharingan/NA specific thing....  And i guess the strat config is named sharingan so I shouldn't complain too much. The two ayatori videos I just perused (in addition to the one I just linked) swap the DPS at SW with everyone to form the correct shape.

valarnin: So you're saying we need an NA and a JP variation? :shrug:

liaoarden: Nah. I was just perusing this PR since I was hoping to make my life slightly easier for callouts haha. This does make me +1 quisquous's comment about keeping them separate and toggleable between orbs and chains.

JLGarber: Do note that having the tank swap 100% means the healer never has to, increasing their uptime slightly. Since PLD no longer has cast times during any cooldown block, this is a strict gain over "hope your physical ranged gets it".

liaoarden: We're running a rphys, double caster {RDM, SMN}, and the two of us casters have instant casts up in time for the bowtie movement so no need to hope for phys ranged.

Our healer can be sad about their ruin2 or dropped GCD if they get unlucky 🤣 .

valarnin: It wouldn't be that difficult to separate it out into settings like this:

Strat: [none, sharingan]
Who moves: [Tank, melee DPS]

But I'm going to keep moving forward with the tank moves strat for now.

What should the orb strat be called? `sharingan` as well?

quisquous: "Orbs Through Middle"?

liaoarden: 👍 on separating in the future.

Yeah I think I started seeing orb strat on opposite side more with `sharingan` than with `ayatori` (though we also do it since it simplifies the movement for which E/W cardinal you pull into), so it seems reasonable to also call it `sharingan`. If Xenosys(?) uptime bowtie didn't rotate orbs CW twice (you end up in the same sides as this strat) it would have been nice to call this `uptime` orbs...

valarnin:
I'm going to close this PR and abandon this branch. Given the many different implementations of the LR strat by the same name, I don't think it's feasible to account for all of them in a way that is user-friendly/configurable, and given that I'm no longer running this raid tier due to having gotten all my clears/gear, so I can't test this anymore. If someone else wants to take this up instead, feel free.

PR #1316, opened 2020-04-07
Oopsy: E7N Light's Course fix

JLGarber:
I'm not sure how this happened, but I'm guessing it was a copy/paste error when we redid the oopsy triggers to use regex replacements.


quisquous:
Thanks!

PR #1317, opened 2020-04-07
Raidboss: Added common all lang + A3S Fr

VanSlanzar:
- Common: Don't hesitate to tell me if the translations are wrong, I'm not sure with google for Asian languages

### DIFF
--- a/ui/raidboss/data/03-hw/raid/a3s.js
+++ b/ui/raidboss/data/03-hw/raid/a3s.js
@@ -35,6 +35,7 @@
       infoText: {
         en: 'Move Bosses',
         de: 'Bosse bewegen',
+        fr: 'Déplacez les Boss',
         ja: 'ボス動かして',
         ko: '보스 이동 주차',
         cn: '移动Boss',
@@ -48,6 +49,7 @@
       condition: Conditions.targetIsYou(),
       alertText: {
         en: 'Sluice on YOU',
+        fr: 'Eclusage sur Vous',
         cn: '蓝点名',
       },
     },
@@ -57,6 +59,7 @@
       condition: Conditions.targetIsYou(),
       infoText: {
         en: 'Tank Debuff',
+        fr: 'Debuff vulnérabilité',
         cn: '坦克 Debuff',
       },
     },
@@ -66,6 +69,7 @@
       condition: Conditions.targetIsYou(),
       infoText: {
         en: 'Healer Debuff',
+        fr: 'Debuff soins',
         cn: '奶妈 Debuff',
       },
     },
@@ -76,6 +80,7 @@
       condition: Conditions.targetIsYou(),
       infoText: {
         en: 'Damage Debuff',
+        fr: 'Debuff dégats',
         cn: 'DPS Debuff',
       },
     },
@@ -84,6 +89,7 @@
       regex: Regexes.ability({ source: ['Liquid Limb', 'Living Liquid'], id: 'F09', capture: false }),
       infoText: {
         en: 'Burn Higher HP Hand',
+        fr: 'Burn sur la main au PV le plus élevée',
         cn: '转火血多手',
       },
     },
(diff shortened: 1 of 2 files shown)


VanSlanzar:
- There are no regexFr lines on (a1s), but it works well.
- On (a2s) I created it, but not on (a3s)
Questions: do I have to create them on all raids (alex savage) or it is not useful?

ui/raidboss/data/03-hw/raid/a3s.js
```diff
@@ -144,12 +152,14 @@
...
           return {
             en: 'Repel: close to ' + data.ShortName(partner),
+            fr: 'Repoussement: Rapprochez-vous de ' + data.ShortName(partner),
             cn: '同极：靠近' + data.ShortName(partner),
           };
         }
 
         return {
           en: 'Attract: away from ' + data.ShortName(partner),
+          fr: 'Attirement: Eloignez-vous de ' + data.ShortName(partner),
```
mooondark, line 162: Attraction. _Attirement_ does not exist.

ui/raidboss/data/03-hw/raid/a3s.js
```diff
@@ -203,8 +216,44 @@
       },
       infoText: {
         en: 'Embolus: Move Boss',
+        fr: 'Caillot: Déplacez le Boss',
```
mooondark, line 219: General remark. In French, you have to put a space before a punctuation mark than contains two parts (like : ; ? !)
So in that case it will be `fr: 'Caillot : Déplacez le boss',` (Ah oui et en français il n'y a pas de majuscule à chaque mot (sauf nom propre).

ui/raidboss/data/03-hw/raid/a3s.js
```diff
@@ -144,12 +152,14 @@
         if (marker1 == marker2) {
           return {
             en: 'Repel: close to ' + data.ShortName(partner),
+            fr: 'Repoussement: Rapprochez-vous de ' + data.ShortName(partner),
```
mooondark, line 155: Répulsion. _Repoussement_ does not exist.

ui/raidboss/data/03-hw/raid/a3s.js
VanSlanzar: je corrige

ui/raidboss/data/03-hw/raid/a3s.js
VanSlanzar: Je corrige

ui/raidboss/data/03-hw/raid/a3s.js
VanSlanzar: je corrige

VanSlanzar:
split

this is on T1, should I take it as a translation for (--split--)?

Jaehyuk-Lee:
I don't know T1 mechanics. so, I don't know what it exactly means... sorry

Akurosia:
is there more than one --split-- in game? can't really think of any other enemy split.
In another note split can mean different things in german and "Zerteilung" would not always fit tbh.
"teilen" or "trennen" would match more cases but still not be always accurate

VanSlanzar:
This means that 1 monster becomes 2.  Same mechanic (t1) and (a1s)

Akurosia:
for the common_replacement file i would than translate --split-- into --teilen--

VanSlanzar:
I think this better.

Una-P:
I don't know how this code will work, too... I'll make a suggestion about choosing kor.
T1's split calls '분열', but '분리'is same meaning.
But '분열' has more meaning biologycal division like amoeba, '분리' is just separation for object(brute justice become 4 robot ,etc..)
'나누다' is a verb. I think '분열' or '분리' is more better one.

VanSlanzar:
it's ok.

quisquous:
Thanks! I agree that `--split--` is probably common enough to move to common.  If anybody wants to follow up and adjust other languages, they are welcome to submit another pull request.

PR #1319, opened 2020-04-07
add regexes +translation for new hw

Akurosia:


Akurosia:
@mooondark could you please have a look at the FR translation in these file (ii think there are some with colon in it where i am not sure if they are 100% correct)

quisquous:
Thanks as always!!

PR #1320, opened 2020-04-09
Initial Varis Triggers

KiroKai:
Couldn't quite get down Loaded Gunshield since it has varying timers.

### DIFF
--- a/ui/raidboss/data/05-shb/trial/varis-ex.js
+++ b/ui/raidboss/data/05-shb/trial/varis-ex.js
@@ -10,7 +10,6 @@
       regex: Regexes.startsUsing({ source: 'Varis Yae Galvus', id: '4CCC', capture: false }),
       run: function(data) {
         data.phase = 'phase2';
-        data.resetState();
       },
     },
     {
@@ -23,7 +22,7 @@
       id: 'VarisEx Ventus Est',
       regex: Regexes.startsUsing({ source: 'Ventus Est', id: '4CC7', capture: false }),
       delaySeconds: 2,
-      response: Responses.getIn(),
+      response: Responses.getIn('info'),
     },
     {
       id: 'VarisEx Altius',
@@ -35,13 +34,15 @@
     {
       id: 'VarisEx Citius',
       regex: Regexes.startsUsing({ source: 'Varis Yae Galvus', id: '4CF0' }),
-      condition: function(data) {
-        return data.role == 'tank' || data.role == 'healer';
-      },
       alertText: function(data, matches) {
         if (data.me == matches.target) {
           return {
             en: 'Buster on YOU',
+          };
+        }
+        if (data.role == 'dps') {
+          return {
+            en: 'Avoid Tanks',
           };
         }
         return {
@@ -52,8 +53,13 @@
     {
       id: 'VarisEx Alea lacta Est',
       regex: Regexes.startsUsing({ source: 'Varis Yae Galvus', id: '4CD2', capture: false }),
-      alarmText: {
-        en: 'Get Behind then Front',
+      response: Responses.getBehind('alarm'),
+    },
+    {
+      id: 'VarisEx Alea lacta Est Front',
+      regex: Regexes.ability({ source: 'Varis Yae Galvus', id: '4CD5', capture: false }),
+      alertText: {
+        en: 'Go Front',
       },
     },
     {
@@ -68,7 +74,15 @@
       condition: (data) => data.phase != 'phase2',
       delaySeconds: 10,
       alarmText: {
-        en: 'Stop Attacking',
+        en: 'Stop',
+      },
+    },
+    {
+      id: 'VarisEx Loaded Gunshield',
+      regex: Regexes.ability({ source: 'Varis Yae Galvus', id: '4CD8', capture: false }),
+      delaySeconds: 10,
+      alarmText: {
+        en: 'Spread soon',
       },
     },
     {
@@ -77,7 +91,14 @@
       condition: (data) => data.phase == 'phase2',
       delaySeconds: 20,
       alarmText: {
-        en: 'Stop Attacking',
+        en: 'Stop',
+      },
+    },
+    {
+      id: 'VarisEx Terminus Est',
+      regex: Regexes.startsUsing({ source: 'Terminus Est', id: '4CB4', capture: false }),
+      infoText: {
+        en: 'Dodge Clones',
       },
     },
     {
--- a/ui/raidboss/data/05-shb/trial/varis-ex.js
+++ b/ui/raidboss/data/05-shb/trial/varis-ex.js
@@ -20,14 +20,6 @@
       },
     },
     {
-      id: 'VarisEx Clones Cleanup',
-      regex: Regexes.ability({ source: 'Phantom Varis', id: '4CB3', capture: false }),
-      delaySeconds: 20,
-      run: function(data) {
-        delete data.clones;
-      },
-    },
-    {
       id: 'VarisEx Ignis Est',
       regex: Regexes.startsUsing({ source: 'Ignis Est', id: '4CB6', capture: false }),
       delaySeconds: 2,
@@ -123,6 +115,9 @@
       condition: (data) => data.clones == 'active',
       infoText: {
         en: 'Dodge Clones',
+      },
+      run: function(data) {
+        delete data.clones;
       },
     },
     {
--- a/ui/raidboss/data/05-shb/trial/varis-ex.js
+++ b/ui/raidboss/data/05-shb/trial/varis-ex.js
@@ -1,0 +1,103 @@
+'use strict';
+[{
+  zoneRegex: {
+    en: /^Memoria Misera \(Extreme\)$/,
+  },
+  timelineFile: 'varis-ex.txt',
+  triggers: [
+    {
+      id: 'VarisEx Phase 2',
+      regex: Regexes.startsUsing({ source: 'Varis Yae Galvus', id: '4CCC', capture: false }),
+      run: function(data) {
+        data.phase = 'phase2';
+        data.resetState();
+      },
+    },
+    {
+      id: 'VarisEx Ignis Est',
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,103 @@
...
+    en: /^Memoria Misera \(Extreme\)$/,
+  },
+  timelineFile: 'varis-ex.txt',
+  triggers: [
+    {
+      id: 'VarisEx Phase 2',
+      regex: Regexes.startsUsing({ source: 'Varis Yae Galvus', id: '4CCC', capture: false }),
+      run: function(data) {
+        data.phase = 'phase2';
+        data.resetState();
```
JLGarber, line 13: I'm a little confused by the `resetState()` here. What am I missing?

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,103 @@
...
+      id: 'VarisEx Ignis Est',
+      regex: Regexes.startsUsing({ source: 'Ignis Est', id: '4CB6', capture: false }),
+      delaySeconds: 2,
+      response: Responses.getOut(),
+    },
+    {
+      id: 'VarisEx Ventus Est',
+      regex: Regexes.startsUsing({ source: 'Ventus Est', id: '4CC7', capture: false }),
+      delaySeconds: 2,
+      response: Responses.getIn(),
```
JLGarber, line 26: For mechanics that are 50/50 (Ignis/Ventus, Chariot/Dynamo, etc,) best practice is to assign one an alert and one an info, so they are sound-distinguishable. (`getIn()` and `getOut()` both default to alert.) I think most other examples in the repository have `getIn()` be info level.

JLGarber:
This looks nice! Good to see this get taken care of so quickly. I was more looking at using the `21` log line values to determine trigger points for the Gunshields, but I think you're right, just adding a fixed delay is probably good enough.

It might also be worth adding a trigger warning of the clone dashes. I think that's Terminus Est, `4CB4`.

sethi12345:
Should have resolved most of the issues I believe, and just added an approximate trigger to avoid the Loaded spread.

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,124 @@
...
+        }
+        return {
+          en: 'Buster on ' + data.ShortName(matches.target),
+        };
+      },
+    },
+    {
+      id: 'VarisEx Alea lacta Est',
+      regex: Regexes.startsUsing({ source: 'Varis Yae Galvus', id: '4CD2', capture: false }),
+      response: Responses.getBehind('alarm'),
```
quisquous, line 56: Thanks for changing this to differ getBehind/goFront, but I think there are too many "alarms" in this fight.  Can getBehind be info (because most people already behind) and go front be alert (since more unusual)?

See also: https://github.com/quisquous/cactbot/blob/master/CONTRIBUTING.md#trigger-severity

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,124 @@
...
+      id: 'VarisEx Alea lacta Est Front',
+      regex: Regexes.ability({ source: 'Varis Yae Galvus', id: '4CD5', capture: false }),
+      alertText: {
+        en: 'Go Front',
+      },
+    },
+    {
+      id: 'VarisEx Electrified Gunshield',
+      regex: Regexes.ability({ source: 'Varis Yae Galvus', id: '4CD7', capture: false }),
+      delaySeconds: 22,
```
quisquous, line 68: What's the exact timing between 4CD7 and the knockback in the log? If possible, I want to give exactly a 5 second warning.  #1322 has 20 seconds, so I'm just curious.

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,124 @@
...
+      response: Responses.knockback(),
+    },
+    {
+      id: 'VarisEx Reinforced Gunshield',
+      regex: Regexes.ability({ source: 'Varis Yae Galvus', id: '4CD9', capture: false }),
+      condition: (data) => data.phase != 'phase2',
+      delaySeconds: 10,
+      alarmText: {
```
quisquous, line 76: Can you change this to `alertText: 'Shields Soon'`.  alarm is very loud, and "stop" on its own is ambiguous (stop moving? stop everything?)

Is there a difference between this and the gunshield below? I'm not sure I understand why there are two triggers.  (Sorry, I haven't run this fight yet, so my knowledge is a bit limited.)

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,124 @@
...
+        en: 'Stop',
+      },
+    },
+    {
+      id: 'VarisEx Loaded Gunshield',
+      regex: Regexes.ability({ source: 'Varis Yae Galvus', id: '4CD8', capture: false }),
+      delaySeconds: 10,
+      alarmText: {
```
quisquous, line 84: What about `response: Responses.spread('info')` ?

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -0,0 +1,124 @@
...
+            en: 'Avoid Tanks',
+          };
+        }
+        return {
+          en: 'Buster on ' + data.ShortName(matches.target),
+        };
+      },
+    },
+    {
+      id: 'VarisEx Alea lacta Est',
```
quisquous, line 54: small nit, `lacta` should be `Iacta` (it's a capital-I not a lowercase-l).

ui/raidboss/data/05-shb/trial/varis-ex.js
KiroKai: I think at least alert is warranted for get behind, as this attack happens immediately after the boss turns towards a random player, so most people are actually in front when this attack starts.

ui/raidboss/data/05-shb/trial/varis-ex.js
KiroKai: I have it set to 2 different attacks as the delay before it goes off is longer in phase 2, unless there is a better way to do it. I feel like 'Shields soon' might be too weak since this attack can easily kill you if you aren't paying attention. (It calls it out about 1-2 seconds before the shields spawn.) Maybe I could make it say 'Stop attacking'.

ui/raidboss/data/05-shb/trial/varis-ex.js
KiroKai: I've used these triggers and I believe a 20 second warning would have you use your kb mitigation too early.

ui/raidboss/data/05-shb/trial/varis-ex.js
KiroKai: Would this work in this case?
```
durationSeconds: function(data, matches) {
        if (data.phase == 'phase2') {
          return 20
        };
          return 10;
        },

**Comment by KiroKai** [comment]
Fixed a few things mentioned and figured out a way to do the Terminus Est clone callout.

File: ui/raidboss/data/05-shb/trial/varis-ex.js [hunk 8]
```diff
@@ -0,0 +1,148 @@
...
+      run: function(data) {
+        data.clones = 'active';
+      },
+    },
+    {
+      id: 'VarisEx Clones Cleanup',
+      regex: Regexes.ability({ source: 'Phantom Varis', id: '4CB3', capture: false }),
+      delaySeconds: 20,
+      run: function(data) {
+        delete data.clones;
```
JLGarber, line 27: It looks like the `clones` property is used only for the dodge trigger? If so, it might be better to move the `delete` logic to the end of the dodge trigger, rather than having an entirely separate trigger for cleanup.

ui/raidboss/data/05-shb/trial/varis-ex.js
JLGarber: I agree with alert for `getBehind()`. The boss can be baited into facing a particular direction, but failure to move means the user will most likely die. (If they are not already moving by the time the animation starts, they will be hit at least twice.)

ui/raidboss/data/05-shb/trial/varis-ex.js
JLGarber: Note also that at least one knockback is paired with a Chariot mechanic (Ignis Est.) It is probably important to make sure that the knockback warning is displayed after `getOut()` in that situation. (Speaking from experience playing Bard in Stormblood, not everyone uses anti-knockback abilities consistently.)

ui/raidboss/data/05-shb/trial/varis-ex.js
JLGarber: "Stop Attacking" or "Pause Attacking" are definitely better.

ui/raidboss/data/05-shb/trial/varis-ex.js
maiko-tan-coding: I assume it is 26.6 seconds, why I set the `delaySeconds` as 20 sec just I played Black Mage, so I need an earlier notice so that I can plan which GCD can I insert my anti-knockback. (Although most of the time I have a `Fire 3 proc` or `Xenoglossy` allowed me to "Surecast")
I think 5 sec was still fine. But I want to announce player with timeline text too.

DreadwyrmBahamut:
I have nothing to contribute but I just wanted to say I appreciate all the hard work you guys put into this, just know you are all appreciated very much. It makes learning fights so much easier and honestly funner in some respects because it gives you the feel of confidence when in parties and you have only just started learning the fight (toxicity reducer if you will haha)

quisquous:
Thank you so much for this!  It's super appreciated, and lots of people have been asking for it.

PR #1324, opened 2020-04-09
ARR translation and added minor Fr

VanSlanzar:


VanSlanzar:
- t1 correction
- t4, t9, t12 added
- t5, t6, t7, t8, t11 correction and remove area name useless
- t10 adjustment for timeline Fr and remove area name useless
t10 timeline Fr Add must be before name monster in french. Screen of the corrected version
- ifrit nm, titan ex, remove double symbol (( ))
Ifrit no before
ifrit yes after

ui/raidboss/data/02-arr/raid/t10.js
```diff
@@ -91,7 +91,6 @@
       'locale': 'de',
       'replaceSync': {
         'Imdugud': 'Imdugud',
-        'The Alpha Concourse': 'Alpha-Emergenzzone',
```
quisquous, line 94: Why are you removing this? https://github.com/quisquous/cactbot/blob/master/ui/raidboss/data/02-arr/raid/t10.txt#L12

ui/raidboss/data/02-arr/raid/t10.js
```diff
@@ -111,18 +110,18 @@
       'locale': 'fr',
       'replaceSync': {
         'Imdugud': 'Imdugud',
-        'The Alpha Concourse': 'secteur des croyants',
       },
       'replaceText': {
+        '1x Son': 'Adds 1x Fils',
```
quisquous, line 115: This is confusing to read the replacements.  If you're going to replace more than just "son" and "daughter", maybe replacing the whole line would be easier to read, e.g. `'1x Son / 1x Daughter Adds': 'Adds 1x Fils / 1x Fille',`

ui/raidboss/data/02-arr/trial/ifrit_nm.js
```diff
@@ -42,8 +42,8 @@
         'Surrender': 'Surrender', // FIXME
       },
       'replaceText': {
-        '(inner)': '(innen)',
-        '(outer)': '(außen)',
+        'inner': 'innen',
```
quisquous, line 45: For your own knowledge, the reason this caused double parentheses is that the `(inner)` is treated as a regular expression.  I think it would need to be `'\\(inner\\)': '(innen)'` to work properly as it was intended to before.  I think this is fine, because there is no other "inner" or "outer" things to replace, however I think other files might have this same issue.

ui/raidboss/data/02-arr/raid/t10.js
VanSlanzar: On the Alex raid, there are these lines also in the files (txt), but they are not written in the files (js) on (replaceSync).  I did not check for the other raids.
- Either it is useless because the check is ok. 
- Or it is an oversight on the alex files.

Isn't it just useful to sync the bosses? Tell me what to do and I'll do it ^^.

ui/raidboss/data/02-arr/raid/t10.js
VanSlanzar: I tried a lot of things, each time either (add) is positioned badly, or (x1, x2). I will try again.

VanSlanzar: - I changed, but I really need the 2 lines with different numbers for it to work.
- Thank you, what I did was complicated for nothing.

ui/raidboss/data/02-arr/trial/ifrit_nm.js
VanSlanzar: In order to understand us well:
- do you want me to put '\\ (inner \\)': '(innen)'?
- or do I leave as I put?

quisquous: I think either is fine here.  I was mentioning it as information for other timelines.  Sometimes not having the parentheses can cause conflicts where two translations apply to the same log.  The `npm run test` step will definitely tell you when this is the case.  So if the test passes, it is not needed.

ui/raidboss/data/02-arr/raid/t10.js
quisquous: I think this is an oversight on the alex files.  Many timelines have extra syncs in case there is not a translation (so that it will still start), but having the zone translation will allow the timeline to start immediately.

ui/raidboss/data/02-arr/raid/t10.js
```diff
@@ -111,18 +110,17 @@
       'locale': 'fr',
       'replaceSync': {
         'Imdugud': 'Imdugud',
-        'The Alpha Concourse': 'secteur des croyants',
       },
       'replaceText': {
+        '1x Son / 1x Daughter Adds': 'Adds 1x Fils / 1x Fille',
```
quisquous, line 115: Thanks, this is much easier to read!

ui/raidboss/data/02-arr/raid/t10.js
VanSlanzar: All right, I will correct that.

ui/raidboss/data/02-arr/trial/ifrit_nm.js
VanSlanzar: Very good

ui/raidboss/data/02-arr/raid/t10.js
VanSlanzar: Yes, sometimes I complicate my life ^^.

VanSlanzar:
Everything is corrected with the names of the games, I redid the instances.
- t6
t6
- t7
t7
- t8
t8
- in t9 and other, I see this, how sync is this supported?
t9
- t11, As for t6, I corrected so that the timeline Fr is correct.
t11 before Before
t11 after After

Thank you for your corrections, help, I am learning a lot from you.

quisquous:
> Everything is corrected with the names of the games, I redid the instances.

Thanks!

> * in t9 and other, I see this, how sync is this supported?
>   t9

Via https://github.com/quisquous/cactbot/blob/master/ui/raidboss/common_replacement.js#L46, which applies to both text and syncs.

VanSlanzar:
Ah, I seemed to have seen it. Thank you.

PR #1329, opened 2020-04-11
Respons correction minor all lang

VanSlanzar:
- I saw this on the new raid of shadowbringer, so I adjust the respons file.
Respons Tank buster

- Which is the correct spelling (Tankbuster) or (Tank Buster) ? 
- Is there a difference between (tank Buster) and (Tank Busters)?
- Can we consider, as on E4S, that a (tank buster skill) even if it gives several shots, remains only one skill, and does not need to be in the plural?

The name would therefore be (Tank Buster) all the time.
I ask for to avoid errors because I will resume the fixes on Shadowbringer, @JLGarber, @mooondark, @quisquous.

- a friend english say me (Tank Buster) is correcty), i wait your reply.


VanSlanzar:
Sorry for the spelling mistakes, I work surrounded by my children during confinement. This is difficult, I try to do better. But this is no excuse, I keep learning on my own language.

JLGarber:
As far as the difference between "tank buster" and "tankbuster", it looks to me like general usage in the repository depends on whether we are talking about code or human-readable text. Anywhere that humans would read output, such as timelines or localized response objects, it should probably be "tank buster". Anywhere it will only be read by a developer or the computer, it should be `tankbuster`.

I agree that no matter what we end up with, it should always be singular. If the buster is multiple hits, and if it *matters* that it is multiple hits, we can always put it as `Tank Buster x3`

 (To be clear, this is all my opinion, not The Way Things Are.)

quisquous:
My subjective opinion here is that I think it should be "Tank Busters" (plural) when there are multiple targets for the tankbuster, like in e2s, but agree that a multi-hit tankbuster on a single target is still generally just "Tank Buster" (singular).  And if the multi-hit part matters, probably the timeline is the right place to put it as "x3", as mentioned above.  I'm not sure it's ever been critical enough to include a trigger (yet).

resources/responses.js
```diff
@@ -444,9 +444,9 @@ let Responses = {
   getBehind: (sev) => {
     let obj = {};
     obj[defaultAlertText(sev)] = {
-      en: 'Get Behind',
-      de: 'Hinter ihn',
-      fr: 'Passez derrière',
+      en: 'Get Behind Boss',
```
quisquous, line 447: Subjectively, I think "boss" is implied here and on many other triggers, and I prefer the original.  I think if there's something else to "get behind" (or look towards, etc, etc) then that can be specified differently.

JLGarber:
That's a good point on the multi-busters, I totally forgot about guaranteed splits like that. (I play Bard, I'm ~never~ not supposed to be close by for those!)

resources/responses.js
VanSlanzar: ok i correct

VanSlanzar:
that's good.

VanSlanzar:
For tank buster, i'm DRK. That's why I think about it, I suffer them all the time XD.

PR #1333, opened 2020-04-12
raidboss: update ja translation for varis-ex.js

maiko-tan-coding:


ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -385,15 +384,17 @@
         'Ventus Est': 'ウェントゥス・エスト',
       },
       'replaceText': {
+        '\\?': ' ?',
+        '--clones appear--': '--幻影が現れる--',
         'Aetherochemical Grenado': '魔導榴弾',
         'Alea Iacta Est': 'アーレア・ヤクタ・エスト',
         'Altius': 'アルティウス',
+        'Blade\'s Pulse': '攻撃を受け止める',
```
maiko-tan-coding, line 392: Although I wrote the Japanese name as the game client shows exactly, in Japanese, `攻撃を受け止める` literally means "Take this damage so that party members would not be hurt.", already can it be a notice text for players.
So, I don't think it must be displayed in the timeline when player using Japanese client.

quisquous: Yeah, this time in the timeline is for when those objects appear.  If there's a better way to describe it than their literal descriptions, feel free to write that too.

PR #1336, opened 2020-04-12
Raidboss : Update fr varis-ex.js

VanSlanzar:


VanSlanzar:
To be consistent with changes to the response file

PR #1337, opened 2020-04-13
update ko translation, e8s miner fix

Una-P:
When shiva starts using holy, trigger notice me .getOut(). But when holy divided, .getIn() doesn't work. I think this line makes problem.


ui/raidboss/data/05-shb/raid/e8s.js
```diff
@@ -599,7 +600,6 @@
       regexJa: Regexes.startsUsing({ source: 'シヴァ', id: '4D83', capture: false }),
       regexCn: Regexes.startsUsing({ source: '希瓦', id: '4D83', capture: false }),
       regexKo: Regexes.startsUsing({ source: '시바', id: '4D83', capture: false }),
-      condition: (data) => data.role == 'tank',
```
quisquous, line 602: Whoops, thanks!

PR #1344, opened 2020-04-15
Respons.js correction Fr

VanSlanzar:
- Several friends who speak good English have told me about my mistakes. So we are meeting to discuss this.
After a very long debate on the mechanics and announcements to be made, we decided to make these changes.
 - Some words like (spread) or (stack), they complicated, because there are different lines with these words. In order to keep the same word for each line using it, some have been conjugate and others have been transformed.
- (package, packez) have been deleted because the definition is not correct. (package) Only concerns objects or files, but I could accept it.
- the rest has been simplified while remaining logical.
- I think that a party of French players who raid put FFXIV in English, and I would like those who put it in French to benefit from the calls in French. If (spread, stack) still causes problems for someone on the translation. I would simply suggest not the translates.
- I would use it as a basis (respons) for all my other translations (ARR, Hw, St, Shb) and i will correct all the files where I made these errors, once this request has been validated.
- Would sorting alphabetically in this file for the research be helpful?
- I remain open to any suggestions, And do you approve of these changes ? @mooondark .

**My apologies to all of you for these new changes.** @quisquous (wait for a response before merge).

### DIFF
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -218,7 +218,7 @@ let Responses = {
       return {
         en: 'Avoid tank cleave',
         de: 'Tank Cleave ausweichen',
-        fr: 'Évitez le cleave sur le tank',
+        fr: 'Évitez le tank cleave',
         ja: '前方範囲攻撃を避け',
         ko: '광역 탱버 피하기',
         cn: '远离顺劈',
@@ -267,7 +267,7 @@ let Responses = {
     obj[defaultInfoText(sev)] = {
       en: 'Spread',
       de: 'Verteilen',
-      fr: 'Écartez-vous',
+      fr: 'Dispersion',
       ja: '散開',
       cn: '分散',
       ko: '산개',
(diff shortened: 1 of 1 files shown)


mooondark:
Hello there.
Here's my opinion on two things :
- _Regroupement_ is really awful, I've never heard that term being used in every MMO I played (Wow/SWTOR/GW2). I pugged a lot on Wow and pack/package were always used to stack.
- I think _Éloignez-vous_ is better than _Loin_, because it include movement (like move away from front).

For the rest, I've nothing to say, it's fine !

VanSlanzar:
As written above, we had a long debate.
- And I put that (package) was acceptable. If it suits you, I reinstate package. @mooondark 
- I'm ok for the second point.

VanSlanzar:
J'ai inclus tes préférence. Et la ou tu n'en n'avais pas, j'ai mis comme si la mécanique était solo (packez-vous sur) et quand c'est une succesion j'ai fait court (dispersion => package). Dis moi, si tu valide tous ca ^^. @mooondark 

Sorry for the inconvenience to all, I will check the files (ARR and Shadowsbringers) and I will be corrected for the translations.

When @mooondark  has validated, you can (merge) @quisquous.

mooondark:
Everything looks fine for me.

VanSlanzar:
you can merge

quisquous:
Thanks for all the discussion to come to an agreement on these translations (and others).  It's hard for me to know what is best for anything that is not English.  :)

PR #1347, opened 2020-04-17
fix mixed up startsUsing and ability

Akurosia:


quisquous:
There's no startsUsing for this ability.  Am I missing something?

Akurosia:
tbh i just found they are inconsistent between Regex and e.g. RegexDe and usually i think regex is correct and change the others accordingly

quisquous:
Whoops, sorry for the misunderstanding.  :eyes:

PR #1351, opened 2020-04-19
Raidboss: Fix for broken TEA headmarker fix

### LINKED ISSUE #1346: v5.11 TEA head marker support
## Description

head markers are shown wrong on cactbot overlay with v5.11 client.
|exact number|shown number|
|---|---|
|1|1|
|2|1|
|3|2|
|...|...|
|7|6|
|8|7|

## Additional information

I don't have network log file, but there's working js file now distributed on Korean community.
[link](https://drive.google.com/file/d/1DPmuGLTV-j9Jp8kWR5t95t9HdmyP9x0T/view)

FYI - I don't do TEA.

JLGarber:
This should resolve #1346 . The fix introduced for the 5.2 scrambling of the TEA markers had a ridiculously subtle bug. As the comment notes, any time the initial head marker is `004F`, the offset is zero. Since zero is falsy, this collides with the naive check for "did we find the offset yet?" This wasn't a problem in 5.2 content, since the odds of hitting That One Instance where the offset was actually 0 are extremely low. No surprise it wasn't noticed until people had to do this on 5.1! 

Many thanks to quisquous for taking the time to look through this and point me in the right direction. Thanks Jaehyuk-Lee for the bug report!

### DIFF
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
@@ -30,7 +30,9 @@
 // added to their ID. This offset currently appears to be set per instance, so
 // we can determine what it is from the first overhead marker we see.
 let getHeadmarkerId = (data, matches) => {
-  if (!data.decOffset) {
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for the fix!

PR #1353, opened 2020-04-19
eureka: add Korean server support

Jaehyuk-Lee:
This does not completely support Korean, I mean Korean language. (some characters are still shown in English.)
It seems every function is working fine on Korean server with this code.


ui/eureka/eureka.js
```diff
@@ -18,10 +18,27 @@ let Options = {
...
+      'Eureka Pagos': '파고스편',
+      'Eureka Pyros': '피로스편',
+      'Eureka Hydatos': '히다토스편',
+    },
+  },
+  timeString: {
+    'for ([0-9]{1,3})m': {
+      'ko': '$1분 동안',
+    },
+    // 'for \?\?\?': {
```
Jaehyuk-Lee, line 31: I get `Uncaught SyntaxError: Invalid regular expression: /for ???/: Nothing to repeat` error here. I don't know why this is invalid.

ui/eureka/eureka.js
```diff
@@ -18,10 +18,30 @@ let Options = {
...
+    'in ([0-9]{1,3})m': {
+      'ko': '$1분 후',
+    },
+    // 'in \?\?\?': {
+    //   'ko': '??? 후',
+    // },
+    '([0-9]{1,3})m': {
+      'ko': '$1분',
+    },
+    '(;[0-9]{1,3})m': {
```
Jaehyuk-Lee, line 43: I've added `;` in front, because it makes clear that there's no another string between weatherIcon and minute number.

Jaehyuk-Lee:
does flag on map always shown with Anemos Elemental, Pagos Elemental things? no other things makes flags? 
I think `Duration of flags on the map (seconds)`'s Korean translation is better says Elemental not flag if so.

ui/eureka/eureka.js
quisquous: It's a string, so you have to double escape if you want to , `'for \\?\\?\\?'`

ui/eureka/eureka.js
```diff
@@ -2256,7 +2293,7 @@ class EurekaTracker {
       } else {
         weatherStr += ' for ???';
       }
-      document.getElementById('label-weather0').innerHTML = weatherStr;
+      document.getElementById('label-weather0').innerHTML = this.strLocalization(weatherStr);
```
quisquous, line 2296: Rather than doing localization by string replacement on strings that are appended together, could you localize the building of these strings, so that it's explicit?
```
let timeStrings = {
  weatherFor: {
    en: (weather, nowMs, stopTime) => {
      let weatherStr = gWeatherIcons[currentWeather];
      if (stopTime) {
        let min = (stopTime - nowMs) / 1000 / 60;
        weatherStr += ' for ' + Math.ceil(min) + 'm';
      } else {
        weatherStr += ' for ???';
      }
      return weatherStr;
    },
    ko: (weather, stopTime) => {
      // etc
    },
  },
  remainingTime: {
    en: (remainingMs) => {
    },
  },
},
```

That way if anybody ever changes the format of what it says in English then it won't break other languages.

ui/eureka/eureka.js
```diff
@@ -82,7 +108,7 @@ let Options = {
             de: 'Sabo',
             fr: 'Pampa',
             ja: 'サボテン',
-            ko: '사보텐더',
+            ko: '사보',
```
quisquous, line 111: This looks like the changes here to the mob list will conflict with Katt's #903.  Can you verify that Katt's changes are identical so that this won't conflict?

Jaehyuk-Lee: That looks identical with mine (for Korean 'trackerName's)

ui/eureka/eureka.js
Jaehyuk-Lee: finished changing to this method.

ui/eureka/eureka.js
Jaehyuk-Lee: I didn't understand actually, putting `for \?\?\?` into Regexp changes to `for ???`?

ui/eureka/eureka.js
```diff
@@ -37,6 +83,12 @@ let Options = {
       'gImportRegex': Regexes.parse(/00:00..:(.*)冷却中的NM: (\S.*\))/),
       'gTimeRegex': Regexes.parse(/(.*) \((\d*)分(钟*)\)/),
     },
+    ko: {
+      'gFlagRegex': Regexes.parse(/00:00..:(.*)에우레카 (?:아네모스|파고스|피로스|히다토스) \( (\y{Float})\s*, (\y{Float}) \)(.*$)/),
```
Jaehyuk-Lee, line 87: I don't get what `gFlagRegex` is for

quisquous: gFlagRegex is for when people link a map location in chat.  If I hold down ctrl on the keyboard and right click on the map, then I get a red flag and in my chat window, it says `<flag>` which turns into coordinates if I chat with that.

ui/eureka/eureka.js
```diff
@@ -18,10 +18,56 @@ let Options = {
...
-      'Eureka Hydatos': '히다토스 지대',
+      'Eureka Anemos': '아네모스편',
+      'Eureka Pagos': '파고스편',
+      'Eureka Pyros': '피로스편',
+      'Eureka Hydatos': '히다토스편',
+    },
+  },
+  timeStrings: {
+    weatherFor: {
+      en: (nowMs, stopTime) => {
```
quisquous, line 41: Can you include the weather in here too, in case some other language wants the time before the "for" part? Sorry for all the requests, I just think that localized text should be entirely localized and not partially constructed by appending strings.

ui/eureka/eureka.js
Jaehyuk-Lee: ah, so this function is to make this plugin works similar to original game client's map.

ui/eureka/eureka.js
Jaehyuk-Lee: I don't know what you're saying about.
Only I can think of is what I've already written below `weatherIn`.

ui/eureka/eureka.js
quisquous: Yeah, except in the original game client, a player has to click on the linked location to have it show up on their map.  This should show up automatically (along with any text around it).  So, if somebody says "train going to (X, Y)" then there'll be a flag with the word "train" on it on the map.

Jaehyuk-Lee: I've changed code to work that way on Korean server.
스크린샷(6)
But, is this square thing intended or something I have to fix?

ui/eureka/eureka.js
Jaehyuk-Lee: for "entirely localized and not partially consturcted", I can only think of `minute` part, which only changes word, not word's location.

ui/eureka/eureka.js
quisquous: I'm not sure.  I'd debug what `AddFlag` is doing and see if there's something going on.  It's intended that `asd <flag> qwe` would just say `(flag icon) asd qwe` on the map.

ui/eureka/eureka.js
quisquous: Sorry for the confusion.  What I would like to avoid is partial/incomplete localized strings being appended to other partial strings to create a final result.  Appending strings together implicitly creates an ordering for grammar that doesn't apply consistently to all languages.  In this case, you have `weatherFor` which returns a string like `for 3m` (a partial string) and is appended to some weather emoji (the rest of the string).  What happens if some other language wants to put the weather emoji after the time and not before? Then this needs to be rewritten.

I think it would be better to have `weatherFor` return the complete string, e.g. `☃️ for 3m`.  That's why I was trying to suggest that `weatherFor` look like `en: (weather, nowMs, stopTime) => {` instead.

Sorry for all the changes.  Mostly I wanted to communicate this, as it seems likely it will come up again for other cases.

ui/eureka/eureka.js
Jaehyuk-Lee: ah, this character is this: 
square is font problem but, ffxiv_20200420_074837_713 this special character in ffxiv which is always shown in front of zone name.

Jaehyuk-Lee: This is U+E0BB.. replacing this character would fix this, but I don't know if this is the best solution. I'm looking at regex part now.

Jaehyuk-Lee: adding the problem character, "" in front of zone name fix this.
`'gFlagRegex': Regexes.parse(/00:00..:(.*)에우레카: (?:아네모스|파고스|피로스|히다토스) 지대 \( (\y{Float})\s*, (\y{Float}) \)(.*$)/),`
But I'm wondering why English client doesn't have this problem. English client has that special character in front of zone/area name.

ui/eureka/eureka.js
Jaehyuk-Lee: ah, you meant some condition with weather icon. I was actually thinking about separating weather icon part and text part to align text.
because weather icon width is different themselves.
K-002
Those were something I was thinking about this.

The real response with your comment.
I don't think those icons will move front or back, because people don't read icon, they just see the images.

For Korean, it's better to say
"3분 동안 눈"
first, time part
next, means "for"
lastly, means "snow"

but as this is an icon, it does not follow grammar.
It's like... hmmm
just indicating. you might see images for order, not thinking about time thing.
I cannot explain accurately because of my lack of English.

Jaehyuk-Lee: maybe for right-align written languages can be better with back side icons.
but hmmm.... I think separating text part and icon part, then fixing icon on the left is better.

Jaehyuk-Lee: Oh, I was thinking about right-to-left written languages, but they are not written backward.
like not "English", "hsilgnE" with right sided.
So, icons with in front is better.

Jaehyuk-Lee:
since UCU fix is very minor thing, just added commit here, not making another branch.

quisquous:
> since UCU fix is very minor thing, just added commit here, not making another branch.

I think I'd prefer a separate pull request, if you don't mind! Small pull requests are very easy to review and get merged quickly.  It's better to have pull requests be about separate topics.

ui/eureka/eureka.js
quisquous: Separating sounds nice and would align better.  However, how do you differentiate between "current weather for X minutes" vs "new weather in X minutes"?

quisquous: And yeah, in English, the icon is intended to be part of the sentence.  It's "snow for 20 minutes" etc.

ui/eureka/eureka.js
quisquous: Actually, I think English may have this problem as well.  There used to be garbage characters due to the ffxiv plugin not handling unicode properly.  These got removed in c529b4807db196b0dd702183a49c646776ec2038, and maybe there's a box now as well.  I haven't done much Eureka in a while.  I can test later.

ui/eureka/eureka.js
Jaehyuk-Lee: I was thinking about making separate div.
I didn't think deeply though.

Jaehyuk-Lee: K-002

quisquous: Ah, separate div seems good.  Maybe the text can be the same in English, and that would be fine.

quisquous: Ooh, I like this a lot, thanks!

Jaehyuk-Lee:
oh, don't merge this. There is conflict with #903.

Jaehyuk-Lee:
It's okay for now to merge.

quisquous:
Thanks for making this look better!

PR #1362, opened 2020-04-20
plugin: Add cn/kr FATE opcodes

### LINKED ISSUE #1359: eureka: fate support in Korean server broken
## Description

simple, fate support is not available in Korean server.
yet, Korean server is not v5.18 so "Added Combatant" worked before.

## Additional information

스크린샷(7)

KattTails:
Add the missing OPCodes for KR/CN regions.
I've confirmed the KR opcode with @Jaehyuk-Lee, I'm assuming the CN opcode is the same until 5.2.

This should fix #1359.

### DIFF
--- a/plugin/CactbotEventSource/FateWatcher.cs
+++ b/plugin/CactbotEventSource/FateWatcher.cs
@@ -49,10 +49,10 @@ namespace Cactbot {
       // Fate start
       // param1: fateID
       // param2: unknown
-
+      //
       // Fate end
       // param1: fateID
-
+      //
       // Fate update
       // param1: fateID
       // param2: progress (0-100)
@@ -193,10 +193,8 @@ namespace Cactbot {
       if (a == opcodes[region_].add) {
         AddFate(*(int*)&buffer[Param1_Offset]);
       } else if (a == opcodes[region_].remove) {
-
         RemoveFate(*(int*)&buffer[Param1_Offset]);
       } else if (a == opcodes[region_].update) {
-
         int param1 = *(int*)&buffer[Param1_Offset];
         int param2 = *(int*)&buffer[Param2_Offset];
         if (!fates.ContainsKey(param1)) {
@@ -209,7 +207,6 @@ namespace Cactbot {
     }
 
     private void AddFate(int fateID) {
-
       if (!fates.ContainsKey(fateID)) {
         fates.TryAdd(fateID, 0);
         client_.DispatchToJS(new JSEvents.FateEvent("add", fateID, 0));
@@ -217,7 +214,6 @@ namespace Cactbot {
     }
 
     private void RemoveFate(int fateID) {
-
       if (fates.ContainsKey(fateID)) {
         client_.DispatchToJS(new JSEvents.FateEvent("remove", fateID, fates[fateID]));
         fates.TryRemove(fateID, out _);
@@ -225,7 +221,6 @@ namespace Cactbot {
     }
 
     private void UpdateFate(int fateID, int progress) {
-
       fates.AddOrUpdate(fateID, progress, (int id, int prog) => progress);
       client_.DispatchToJS(new JSEvents.FateEvent("update", fateID, progress));
     }
--- a/plugin/CactbotEventSource/FateWatcher.cs
+++ b/plugin/CactbotEventSource/FateWatcher.cs
@@ -5,11 +5,11 @@ using FFXIV_ACT_Plugin.Common;
 using System.Linq;
 using System.Reflection;
 using System.Runtime.InteropServices;
-using System.Diagnostics;
 
 namespace Cactbot {
   public class FateWatcher {
     private CactbotEventSource client_;
+    private string language_;
     private IDataSubscription subscription;
 
     private Type MessageType = null;
@@ -26,8 +26,9 @@ namespace Cactbot {
     // fates<fateID, progress>
     private static ConcurrentDictionary<int, int> fates;
 
-    public FateWatcher(CactbotEventSource client) {
+    public FateWatcher(CactbotEventSource client, string language) {
       client_ = client;
+      language_ = language;
       fates = new ConcurrentDictionary<int, int>();
 
       var FFXIV = ActGlobals.oFormActMain.ActPlugins.FirstOrDefault(x => x.lblPluginTitle.Text == "FFXIV_ACT_Plugin.dll");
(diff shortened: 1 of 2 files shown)


Jaehyuk-Lee:
hmmm... how did you guess CN opcode does not change until 5.2?
Also, I was thinking, Chinese server won't work with 0.17.0 too. Did anyone checked it?

KattTails:
The opcode for our version was the same as the KR regions opcode is now, and was only changed when 5.2 was released. It was static prior to that. Since CN is on v5.15 I'm assuming their opcode did not change yet either.
I don't think there have been any reports from CN users as to whether 0.17.0/1 works for them. @ShadyWhite @Bluefissure sorry for the pings, do you know if FATE tracking/the eureka overlay works in your region?

Jaehyuk-Lee:
oh really? While I was maintaining DFAssist, duty finder related opcodes were different with every patches and every region. (after v5.1)
Fates doesn't? oh

KattTails:
The FATE codes are a category within the ActorControl_143 message. The ActorControl_143 was shuffled every update, but the categories themselves were not. That seems to have changed with 5.2 however.

Bluefissure:
> The opcode for our international version was the same as the KR regions opcode is now, and was only changed when 5.2 was released. It was static prior to that. Since CN is on v5.15 I'm assuming their opcode did not change yet either.
> I don't think there have been any reports from CN users as to whether 0.17.0/1 works for them. @ShadyWhite @Bluefissure sorry for the pings, do you know if FATE tracking/the eureka overlay works in your region?

Neither 5.1 nor 5.25 opcodes work for CN 5.15, I'm currently busy at my homework and I don't have a good way to dig into the network packages so I'm hoping to fix the opcodes in CN later tonight.

BTW I added some localizations and default language in the `eureka.js` (see https://github.com/quisquous/cactbot/pull/1363), without which missing localization will cause exceptions.

KattTails:
> Neither 5.1 nor 5.25 opcodes work for CN 5.15, I'm currently busy at my homework and I don't have a good way to dig into the network packages so I'm hoping to fix the opcodes in CN later tonight.

Thanks for looking into it. If you do have time after your homework can you try out what is posted [here](https://github.com/quisquous/cactbot/issues/1359#issuecomment-616828933). If you can get the opcodes from that it would be very much appreciated.

Jaehyuk-Lee:
Oh, while I was looking at DFAssist about Fate alarm, I found there's never been any change to 
https://github.com/quisquous/cactbot/blob/2aec6e8669de1caff7ae6c8ac1b174265926435d/plugin/CactbotEventSource/FateWatcher.cs#L139
https://github.com/quisquous/cactbot/blob/2aec6e8669de1caff7ae6c8ac1b174265926435d/plugin/CactbotEventSource/FateWatcher.cs#L146
https://github.com/quisquous/cactbot/blob/2aec6e8669de1caff7ae6c8ac1b174265926435d/plugin/CactbotEventSource/FateWatcher.cs#L154
and... they are not opcodes (maybe I'm confusing, actually I don't know how FateWatcher.cs is working, I don't know about ACT plugin thing).
It looks `switch~case` part won't change afterwards if these three cases are only thing you were concerning about.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -124,35 +125,116 @@ public class FateWatcher {
...
+          //
+          // for FFXIV KO version: 5.11
+          //
+          // Latest KO version can be found at:
+          // https://www.ff14.co.kr/news/notice?category=3
+          //
+          // Fate Start: 0x74
+          // param1: fateID
+          // param2: unknown
+          case 0x74: {
```
quisquous, line 139: Would it be possible to just have a small lookup table of these offsets by language, rather than repeating all the switch statement logic?

In other words, something like if (offset == offsets[lang].add) { } etc

KattTails: Yes that is much simpler.

Jaehyuk-Lee:
I don't know why but AddFate() is not working with this code in Korean server.
I've pull https://github.com/KattTails/cactbot/tree/korean-support and built it with log written by myself (client_.LogInfo("some text here"))
```cs
          case 0x74: {
            AddFate(*(int*)&buffer[Param1_Offset]);
            client_.LogInfo("Add FATE");
            break;
          };

          // Fate End: 0x79
          // param1: fateID
          case 0x79: {
            RemoveFate(*(int*)&buffer[Param1_Offset]);
                            client_.LogInfo("End FATE");
                            break;
          };

          // Fate Progress: 0x9B
          // param1: fateID
          // param2: progress (0-100)
          case 0x9B: {
            int param1 = *(int*)&buffer[Param1_Offset];
            int param2 = *(int*)&buffer[Param2_Offset];
            if (!fates.ContainsKey(param1)) {
              AddFate(param1);
              client_.LogInfo("update FATE");
                            }
            if (fates[param1] != param2) {
              UpdateFate(param1, param2);
            }
            break;
          }
```
"update FATE" was logged.
"End FATE" was also logged.
but "Add FATE" is never logged.

KattTails:
It's possible the 0x9B packet was received before the 0x74 add, and your code will print `update FATE` if an 0x9B packet is received while the fate isn't in the fates dictionary. You'll want to move the code around such as:
```cs
          // Fate Progress: 0x9B
          // param1: fateID
          // param2: progress (0-100)
          case 0x9B: {
            int param1 = *(int*)&buffer[Param1_Offset];
            int param2 = *(int*)&buffer[Param2_Offset];
            if (!fates.ContainsKey(param1)) {
              AddFate(param1);
              client_.LogInfo("add FATE");
            }
            if (fates[param1] != param2) {
              UpdateFate(param1, param2);
              client_.LogInfo("update FATE");
            }
            break;
          }
```

Jaehyuk-Lee:
Ah, I understand the problem, but `case 0x74:` is not used?

KattTails:
It is, it's just the 0x9B packet is sent twice per second and might be received before 0x74. It isn't *strictly* necessary to have 0x74.

Jaehyuk-Lee:
ah now I got it. okay.

KattTails:
I wrote up a small documentation of the fate packets [here](https://github.com/KattTails/cactbot/commit/4de2d02e0634a30968de59fe7655f12324c9fad3#r36607199)

Bluefissure:
> > Neither 5.1 nor 5.25 opcodes work for CN 5.15, I'm currently busy at my homework and I don't have a good way to dig into the network packages so I'm hoping to fix the opcodes in CN later tonight.
> 
> Thanks for looking into it. If you do have time after your homework can you try out what is posted [here](https://github.com/quisquous/cactbot/issues/1359#issuecomment-616828933). If you can get the opcodes from that it would be very much appreciated.

It was my fault, I built cactbot with an old ffxiv plugin, which has an outdated `ActorControl143` opcode.
I rebuilt one with the latest ffxiv plugin and it works well.

Jaehyuk-Lee:
I think first make a set those which varies by client version, then applying to regions would better to maintain.
https://github.com/quisquous/cactbot/blob/e3154ddcee1c85ffba05f6f9b2e75dba97b7ad14/ui/jobs/jobs.js#L765-L783
`const v5_08` - version
```
let buffOverrides = { 
     ko: v5_08, 
     cn: v5_08, 
   };
```
- apply version to region.

For this PR, `v5.1` and `v5.2` would be used for version. `ko`, `cn`, `intl` for region.

KattTails:
> I think first make a set those which varies by client version, then applying to regions would better to maintain.

This assumes each patches OPCodes will be the same across all regions which may not always be the case.

I think it's best to separate them for now.

Jaehyuk-Lee:
Those category had been same across all regions before v5.1. (This is verified) I'm assuming they doesn't vary by region after v5.1 too.

Jaehyuk-Lee:
|      | ~v5.1 | v5.1       | v5.2      | 
|------|-------|------------|-----------|
| intl | **same**  | don't know | different | 
| ko   | **same**  | **same**       | TBD       | 
| cn   | **same**  | **same**       | TBD       |

KattTails:
I've gone ahead and implemented the version definitions as you suggested.

quisquous:
The code looks fine to me, but is this good to merge? Have folks using the cn or ko versions given this a try?

Bluefissure:
> The code looks fine to me, but is this good to merge? Have folks using the cn or ko versions given this a try?

Good for cn.

KattTails:
It's correct for kr too. I got the correct opcodes from @Jaehyuk-Lee in #1359

quisquous:
Thank you!!

Issue #1364, opened 2020-04-21
How can i use chinese text and alter in EN client?


FrankBlue453:
Alter language i can find easily
but cant find how to show chinese text in raidboss
timeline and trigger

Una-P:
![image](https://user-images.githubusercontent.com/39465903/79813929-f1f76e80-83b7-11ea-9302-01bdc60deb85.png)
Did you change Alerts language to CN?

FrankBlue453:
> ![image](https://user-images.githubusercontent.com/39465903/79813929-f1f76e80-83b7-11ea-9302-01bdc60deb85.png)
> Did you change Alerts language to CN?

sure， nothing change....

Bluefissure:
Do you mean text in trigger alerts or timeline?

IIRC the text in trigger alerts can be changed in the config panel, not sure about the timeline.

Jaehyuk-Lee:
timeline is fixed to your game language, it wouldn't be changed since timeline mostly consists of skill names. You would see skill bar charging in game and same on timeline.
For Plugin UI language, I think it's possible to change, but I'm not doing it right now.

FrankBlue453:
maybe we need some function to do it....

Bluefissure:
> maybe we need some function to do it....

do what?

quisquous:
I think you're looking for #1074

quisquous:
If you're looking for timelines, #1247 will have fixed it.

Issue #1247, opened 2020-03-20
English Timers with German Client


olibeniron:
## Summary
Greetings. 
I would like to ask for a function that would allow changing Timers' language in the same way as Alerts' language can be changed independently of the client's language.
With all due respect, Touya.

quisquous:
Fixed in #1367.

PR #1367, opened 2020-04-22
raidboss: add timeline language config

Bluefissure:
This is probably what https://github.com/quisquous/cactbot/issues/1247 and https://github.com/quisquous/cactbot/issues/1364 want.


Jaehyuk-Lee:
I thought I would use this function if there is.
However, I didn't mind if there isn't, so I was not trying to make this function.
thanks :)

quisquous:
Thank you! :)

PR #1370, opened 2020-04-23
raidboss: better Korean css

Jaehyuk-Lee:


quisquous:
Oof, I'm sorry I broke the way this worked before when I updated the raidboss CSS.  Thanks for the fix.

PR #1371, opened 2020-04-23
eureka: remove character name on map

Jaehyuk-Lee:
remove these characters on map ffxiv_20200423_211302_148
party member's \<flag\> chat -> those party member's number also printed on map

one better Korean label name.

### DIFF
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -2056,9 +2056,9 @@ class EurekaTracker {
       let gRegex = this.Translate(this.options.Regex);
       let gFlagRegex = gRegex['gFlagRegex'];
       let match = log.match(gFlagRegex);
-      if (match){
+      if (match) {
         // remove character name on Korean server (if not /echo)
-        if(this.options.Language == 'ko' && !log.match(/00:0038:/))
+        if (this.options.Language == 'ko' && !log.match(/00:0038:/))
           match[1] = match[1].replace(/(.*):/, '');
         this.AddFlag(match[2], match[3], match[1], match[4]);
       }
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -2171,6 +2171,8 @@ class EurekaTracker {
       afterText = '';
     }
     beforeText = beforeText.replace(/(?: at|@)$/, '');
+    // remove party member number in the front
+    beforeText = beforeText.replace(/\uE090-\uE097/g, '');
 
     let container = document.getElementById('flag-labels');
     let label = document.createElement('div');
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -2056,8 +2056,12 @@ class EurekaTracker {
       let gRegex = this.Translate(this.options.Regex);
       let gFlagRegex = gRegex['gFlagRegex'];
       let match = log.match(gFlagRegex);
-      if (match)
+      if (match){
+        // remove character name on Korean server (if not /echo)
+        if(this.options.Language == 'ko' && !log.match(/00:0038:/))
+          match[1] = match[1].replace(/(.*):/, '');
         this.AddFlag(match[2], match[3], match[1], match[4]);
+      }
       let gTrackerRegex = gRegex['gTrackerRegex'];
       match = log.match(gTrackerRegex);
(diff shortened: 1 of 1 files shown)


Jaehyuk-Lee:
character's server(world) name is not removed in Korean server. I'm checking it.

Jaehyuk-Lee:
`00|2020-04-23T21:40:13.0000000+09:00|000b|(CharacterName)(WorldName)|(Chat)|38f38beb0d62f99066aba265c263b36e`
Is it impossible to remove world name? It seems there's no delimiter between Character name and its World name.

quisquous:
> `00|2020-04-23T21:40:13.0000000+09:00|000b|(CharacterName)(WorldName)|(Chat)|38f38beb0d62f99066aba265c263b36e` - from parser log
> Is it impossible to remove world name? It seems there's no delimiter between Character name and its World name.

Yeah, it's the same in English.  In English at least, all the names are stripped from the flag text, so it doesn't get shown.

Jaehyuk-Lee:
I thought showing character name (who said) was intended.
Should character name not be shown on map?

quisquous:
In English it gets too long, so I'd say no? Really the goal of flags on the map was that people post a ton of flags about where the train is, where rezes are needed, etc and I just wanted a way to show those easily.  Who is saying those things seems less important.

Jaehyuk-Lee:
okay, I'll try removing character name after short trip I'm having right now, maybe the day after tomorrow?

quisquous:
(oops sorry, wrong button)

Sounds good! We probably don't need the number replacement if the party name is missing, I think, so I'll wait for you to get back.  Have a nice trip!

Jaehyuk-Lee:
is there any other variables that gets FFXIV parsing plugin's language? I think `this.options.Language` is not this. I think some codes are using this variable mixed with server determine thing and language determine thing.

In this code, I'm using `this.options.Language` for determining Korean server, but I should take a look for server determining.

ui/eureka/eureka.js
```diff
@@ -2056,8 +2056,12 @@ class EurekaTracker {
       let gRegex = this.Translate(this.options.Regex);
       let gFlagRegex = gRegex['gFlagRegex'];
       let match = log.match(gFlagRegex);
-      if (match)
+      if (match) {
+        // remove character name on Korean server (if not /echo)
```
quisquous, line 2060: Hmm, it is surprising to me that there is different logic for different languages.  Is it possible to make them all behave the same?

quisquous:
> is there any other variables that gets FFXIV parsing plugin's language? I think `this.options.Language` is not this. I think some codes are using this variable mixed with server determine thing and language determine thing.

`this.options.Language` should be it.  That is what is sent from the C# code based on what your parsing language is.  Which code is using it wrong?

Jaehyuk-Lee:
Oh, I thought different thing.
alertsLanguage and timelineLanguage are configurable to other languages, so I thought some mistakenly.

ui/eureka/eureka.js
Jaehyuk-Lee: > Yeah, it's the same in English. In English at least, all the names are stripped from the flag text, so it doesn't get shown.

As you said before, it seems only Korean (or Chinese too) server is showing character name on map, so I've set only to work on Korean server.
This logic is removing string ahead of `:`, which always logged except /echo.
so, if this logic is also applied to English version, 
`some place name: <flag>` will be replaced to ` <flag>`.

quisquous: The English flag regex is:
`'gFlagRegex': Regexes.parse(/00:00..:(?:[^:]*:)?(.*)Eureka (?:Anemos|Pagos|Pyros|Hydatos) \( (\y{Float})\s*, (\y{Float}) \)(.*$)/),`

Yeah, I think you're correct that if somebody did `/say text: <flag>` then the `text` part would be removed.

Maybe `/00:00(?:38:|..:[^:]*:)(.*)Eureka ... etc etc` would remove the name from non-echoes but work for echo, and we could apply that to all languages?

Jaehyuk-Lee: This is working!

quisquous:
Thanks for the patch and for the testing!

PR #1373, opened 2020-04-24
raidboss: add ja translations in shb/trial

maiko-tan-coding:
Add / Fix Japanese translation in:
- raidboss_config.js
- shb/titania
- shb/ruby-weapon
- shb/titania-ex
- shb/innocence
- shb/innocence-ex
- shb/hades
- shb/hades-ex

(´。＿。｀) Since I'm not a native Japanese, some translation may not be the best.


ui/raidboss/data/05-shb/trial/hades-ex.js
```diff
@@ -403,7 +403,7 @@
         en: 'Tank Busters',
         de: 'Tankbuster',
         fr: 'Tankbuster',
-        ja: 'タンクにTB',
+        ja: 'タンクにタンクバスター',
```
quisquous, line 406: Sorry that I didn't notice that the old text was like this, but where possible can you use the same text from `resources/responses.js` for consistency?

For example, this is maybe `タンクバスター`

ui/raidboss/data/05-shb/trial/hades-ex.js
```diff
@@ -555,15 +555,15 @@
...
+            ja: data.ShortName(matches.target) + 'にタンクバスター',
             cn: '死刑点 ' + data.ShortName(matches.target),
             ko: '"' + data.ShortName(matches.target) + '" 탱버',
           };
         }
         return {
           en: 'Away from ' + data.ShortName(matches.target),
           fr: 'Eloignez-vous de ' + data.ShortName(matches.target),
-          ja: data.ShortName(matches.target) + 'から離れて',
+          ja: data.ShortName(matches.target) + 'から離れ',
```
quisquous, line 566: The previous text is what is in responses.js as well.

ui/raidboss/data/05-shb/trial/hades-ex.js
maiko-tan-coding: OK, I will fix it as the common translation in response.js.

ui/raidboss/data/05-shb/trial/hades-ex.js
maiko-tan-coding: Well, I have another opinion about this translation.
I thought `離れ` would be more command-style, just like `Hurry up! Go away!` in English, and `離れて` would be more gently (request-style), like `Please move yourself away from someone`. I prefer the previous one to be the alert text.

maiko-tan-coding:
Merged the newest branch to avoid CI check failed.

ui/raidboss/data/05-shb/trial/hades-ex.js
quisquous: That seems reasonable.  Thanks for the explanation!

quisquous:
Thanks!!

PR #1375, opened 2020-04-25
Translation fix based on find_missing_translations

Akurosia:
Most fixes in this PR are based on the entries, when running `python util/find_missing_translations.py -l de`

The fixes targets averything non shb related.
Missing elmenets:
- Old HW trials (as they still use the old format)
- few trigger in SB where most likly the tools is not correct

Also there is a strange entrie for the test file `.\ui\raidboss\data\00-misc\test.js:102 Test Poke:` which also seams to be a bug of the script

Some other adjustments:
- If the whole timelineReplace element was missing, elements for other languages was also added.
- 'missingTranslations': true, was added to severall non German elements
- fix the strange SE change from `6 Fulms Under` to `Six Fulms Under` in o2n and o2s


Akurosia:
there seams also to be another issue for example this is reported: `.\ui\raidboss\data\04-sb\alliance\ridorana_lighthouse.js:190 Ridorana Belias Gigas: missing timelineReplace for regex '^.{14} 03:[0-9a-f]{8}:added new combatant gigas\.' (but have regexDe)` while this `'(?<! )Gigas': 'Diener von Belias',` is part of the replaceSync

quisquous:
> there seams also to be another issue for example this is reported: `.\ui\raidboss\data\04-sb\alliance\ridorana_lighthouse.js:190 Ridorana Belias Gigas: missing timelineReplace for regex '^.{14} 03:[0-9a-f]{8}:added new combatant gigas\.' (but have regexDe)` while this `'(?<! )Gigas': 'Diener von Belias',` is part of the replaceSync

`(?<! )Gigas` is "match the word Gigas that has no space before it.  So it's not going to match `Added new combatant Gigas`.  It's probably trying to not match `Belias, The Gigas` so maybe it should be `(?<! The Gigas`?

quisquous:
Thanks for this giant patch!  Much appreciated.  :)

Issue #1328, opened 2020-04-11
Support network log lines in cactbot


quisquous:
Mentioned in #1312.

cactbot currently only operates on act log lines, but this has some limitations.  There's more data in the network log lines that would be nice to write triggers against (like lightsteeped stacks, or zone ids, or effect ids, or cast bar lengths, etc).

Options look like this:

(1) provide network log lines in addition to act log lines (cons: now you're parsing 2x as much data for every trigger
(2) replace act log lines with network log lines (cons: a bunch of work, also risky, maybe confusing to developers)
(3) implement hypothetical new log injection technique to add new act log lines with this info, continue to use act log lines (cons: farther in the future, also a bunch of work)

I think #2 is plausible. Timelines use raw act log lines, but they're all faiiiirly normalized and we could...convert them magically into network log lines, and use tests to verify that they can be converted as such. Most other things use regex helpers, which could be converted to network log lines too.

This would also solve the zone id issue and effect id issues (to avoid translations on both of them), would provide positioning information on more lines, and some other good stuff.

FabulousCupcake:
Has anyone observed/measured the actual performance impact of option 1?  
Doubling parsing load sounds bad, but maybe it's still well within the acceptable range

JLGarber:
Whee, yes please! This would also mean no more need for supporting both types of log lines in the timeline utilities. That was a major source of confusion for me when I first started the updates to them.

Off the top of my head, I want to say that 1 isn't an option. It's a guarantee that it will double processing time, and that's only going to get worse as the plugin gets older. That being said, FabulousCupcake is right that we should get some hard numbers on it. Didn't panicstevenson do some benchmarking for us when we moved regexes to global case insensitivity? Maybe we could ~bribe~ persuade them to whip something up for us here too?

Honestly, I don't think it's *too* risky to switch to network lines. Just about all the timelines use `15` lines for everything, so in theory automatic conversion there should Just Work. I know there are a few here and there that use other lines (Fractal Continuum comes to mind) but the majority should be fine.

quisquous:
Re: Option 1.  Yeah, agreed that maybe I'm overthinking the performance impact, but it sure *feels* bad.  There's also (unmeasured) overhead for the plugin having to turn all the logs into strings and send them to javascript too.  An additional negative that there's also just added developer burden to have multiple kinds of log lines to match against, and some fields that would only work in certain cases.  It'd be simpler to think about everything if there was only one kind of log line.

[...]

PR #1377, opened 2020-04-25
Raidboss: Sephirot EX timeline update

JLGarber:
Per #1328, this is the first of some updates that will be needed before that can go through.

Shasta and DnT did great work at the time, but the Heavensward timelines they graciously put together are extremely outdated. With some help from quisquous, we now have a modernized Sephirot EX timeline.

(Note that there are two abilities, `Gevurah Chesed|1578` and `Chesed Gevurah|1579`. I'm not 100% sure of the difference, but I'm relatively sure it has to do with who gets what color debuffs. Regardless, because having multiple different variants of "Chesed" in the translation object is painful, I opted to have all instances of either be displayed as "Chesed Gevurah". If someone who knows Hebrew has improvements I would gladly integrate that.)

Potential deficiencies:

1. Zero live testing. I've run this against More Than One log and it matches closely, but because of the stack markers in phase 1 I can't run this solo to test myself.

2. Uncertain enrage. It's entirely possible that a group burning through phase 1 and the intermission could get an extra rotation in phase 2 before the enrage. I don't know what happens there, and I didn't feel like spending the afternoon hunting through FFLogs for a unicorn. (From the example we see in E8n, it's possible Sephirot might break the pattern to go straight to spawning the Storm of Words, leading into the enrage sequence that way. I don't know.)

3. Obviously, we still don't have triggers. That will probably be forthcoming in [6-to-8 weeks](https://meta.stackexchange.com/a/19514).

### DIFF
--- a/ui/raidboss/data/03-hw/trial/sephirot-ex.txt
+++ b/ui/raidboss/data/03-hw/trial/sephirot-ex.txt
@@ -69,7 +69,7 @@ hideall "--sync--"
 1029.0 "Life Force" sync /:Sephirot:157A:/
 1029.0 "Spirit" sync /:Sephirot:157B:/
 1034.6 "Fiendish Wail" sync /:Sephirot:1576:/
-1042.7 "Da'at" # sync /:Sephirot:1574:/ duration 3.2
+1042.7 "Da'at Tethers" # sync /:Sephirot:1574:/ duration 3.2
 1053.3 "Fiendish Wail" sync /:Sephirot:1576:/
 1060.3 "Chesed Gevurah" sync /:Sephirot:157[89]:/
 1060.9 "Life Force" sync /:Sephirot:157A:/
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/03-hw/trial/sephirot-ex.txt
```diff
@@ -1,234 +1,155 @@
...
+1369.5 "Chesed Gevurah" sync /:Sephirot:157[89]:/
+1370.1 "Life Force" sync /:Sephirot:157A:/
+1370.1 "Spirit" sync /:Sephirot:157B:/
+1385.7 "Chesed Gevurah" sync /:Sephirot:157[89]:/
+1386.3 "Life Force" sync /:Sephirot:157A:/
+1386.3 "Spirit" sync /:Sephirot:157B:/
+1388.9 "Pillar of Severity" sync /:Sephirot:1585:/
+1392.6 "Impact of Hod" sync /:Sephirot:172C:/
+1397.0 "Ascension" sync /:Coronal Wind:1584:/
+1403.9 "Pillar of Severity Enrage" sync /:Sephirot:1585:/
```
quisquous, line 155: Yeah, I think my memory is that this is a "rotational" hard enrage.  I don't think I've ever had a slow enough group in phase 1 to know if there's a time-based one.

ui/raidboss/data/03-hw/trial/sephirot-ex.txt
```diff
@@ -1,234 +1,155 @@
...
-992 "Fiendish Wail"
-995 "STACK"
-1000 "Color Slam"
-1002 "STACK"
-1010 "Yesod, MOVE"
-1016 "Color Slam"
-1022 "Impact of Hod" #Small knockback
-1022 "JUMP" #Indicator for jump
-1027 "Ascension" #Damage from jump
+# Containment Bay S1-T17 (Extreme)
```
quisquous, line 1: Should this be S1T7?

JLGarber: It should, yes.

Issue #1378, opened 2020-04-25
Enforce JavaScript style


quisquous:
I would like for all variables, properties, and functions to have camelCase names.  Classes should have ProperCase names.

Goal: make the following pass `npm run lint`:
```diff
diff --git a/package.json b/package.json
index cf3f5dde..a359f93b 100644
--- a/package.json
+++ b/package.json
@@ -30,7 +30,6 @@
       "google"
     ],
     "rules": {
-      "camelcase": "off",
       "curly": [
         "error",
         "multi-or-nest",
@@ -70,14 +69,6 @@
           "tabWidth": 2
         }
       ],
-      "new-cap": [
-        "error",
-        {
-          "capIsNew": false,
-          "newIsCap": false,
-          "properties": false
-        }
-      ],
       "no-cond-assign": [
         "error",
         "always"

```

This implicitly makes both new-cap and camel-case be "error".

For now, before these style rules are turned on, please:
* keep local variables and parameters as camelCase
* be consistent with surrounding code
* don't make unrelated style changes to surrounding code in other pull requests

This will probably cause some headaches because of some "user-exposed" functions on `data` in triggers, and in those cases let's just alias the ProperCase functions to new renamed camelCase ones and convert triggers to use camelCase.

quisquous:
cc #1366

valarnin:
Request to add babel-eslint as the parser so we can support some of the newer syntax that works in Chrome? For instance, I'm using this in my raidemulator redesign:
```javascript
class A {
  static property = {};
}
```
That syntax errors in eslint without babel-eslint parsing. However, enabling babel-eslint also drops a bunch of extra linting errors probably due to inheriting some other rules by default.

Also the eslint version used in cactbot is pretty outdated, might be a good idea to update it in general as part of this change even if you can't enable the other parser for it.

quisquous:
I updated the eslint version in #1379, mostly because it caught some additional (camelcase) errors that the older version did not.

I looked up babel-eslint, but its documentation says `Note: You only need to use babel-eslint if you are using Babel to transform your code. If this is not the case, please use the relevant parser for your chosen flavor of ECMAScript (note that the default parser supports all non-experimental syntax as well as JSX).` But eslint also says "if you need experimental features, this isn't where you get them".  So, I'm not totally sure.  Can you help explain (either in words or patch form) what's adding babel-eslint would do?

valarnin:
[...]

PR #1379, opened 2020-04-25
eslint: turn on camelcase errors

quisquous:
This is part of #1378.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6s.js
+++ b/ui/raidboss/data/05-shb/raid/e6s.js
@@ -14,104 +14,97 @@
       regexFr: Regexes.startsUsing({ source: ['Ifrit', 'Raktapaksa'], id: '4BD3', capture: false }),
       regexJa: Regexes.startsUsing({ source: ['イフリート', 'ラクタパクシャ'], id: '4BD3', capture: false }),
       delaySeconds: 11,
-      promise: function(data) {
-        let p = new Promise(async (res) => {
-          const ifritLocaleNames = {
-            en: 'Ifrit',
-            de: 'Ifrit',
-            fr: 'Ifrit',
-            ja: 'イフリート',
+      promise: async (data) => {
+        const ifritLocaleNames = {
+          en: 'Ifrit',
+          de: 'Ifrit',
+          fr: 'Ifrit',
+          ja: 'イフリート',
+        };
+
+        const raktapaksaLocaleNames = {
+          en: 'Raktapaksa',
+          de: 'Raktapaksa',
+          fr: 'Raktapaksa',
+          ja: 'ラクタパクシャ',
+        };
+
+        // select the 4 most recent Ifrit or Raktapaksa's depending on phase
+        let combatantName = null;
+        if (data.phase === 'ifrit')
+          combatantName = ifritLocaleNames[data.lang];
+        else
+          combatantName = raktapaksaLocaleNames[data.lang];
+
+        let combatantData = null;
+        if (combatantName) {
+          combatantData = await window.callOverlayHandler({
+            call: 'getCombatants',
+            names: [combatantName],
+          });
+        }
+
+        // if we could not retrieve combatant data, the
+        // trigger will not work, so just resume promise here.
+        if (!(combatantData !== null &&
+          combatantData.combatants &&
+          combatantData.combatants.length)) {
+          data.safeZone = null;
+          return;
+        }
+
+        // we need to filter for the Ifrit with the highest ID
+        // since that one is always the safe spot.
+        let currentHighestCombatant =
+          combatantData.combatants.sort((a, b) => a.ID - b.ID).pop();
+
+        // all variation ranges for all the 9 ball positions for the kicking actors
+        // north      x: 96-104   y: 85-93
+        // northeast  x: 107-115  y: 85-93
+        // northwest  x: 85-93    y: 85-93
+        // east       x: 107-115  y: 96-104
+        // west       x: 85-93    y: 96-104
+        // south      x: 96-104   y: 107-115
+        // southeast  x: 107-115  y: 107-115
+        // southwest  x: 85-93    y: 107-115
+        let safeZoneObj1 = { en: '', de: '' };
+        let safeZoneObj2 = { en: '', de: '' };
+
+        // don't need to go through all the posibilities,
+        // only those 4 ifs do reflect the above positions
+        if (currentHighestCombatant.PosY > 84 && currentHighestCombatant.PosY < 94) {
+          safeZoneObj1 = {
+            en: 'north',
+            de: 'nord',
+            fr: 'nord',
           };
-
-          const raktapaksaLocaleNames = {
-            en: 'Raktapaksa',
-            de: 'Raktapaksa',
-            fr: 'Raktapaksa',
-            ja: 'ラクタパクシャ',
+        } else if (currentHighestCombatant.PosY > 106 && currentHighestCombatant.PosY < 116) {
+          safeZoneObj1 = {
+            en: 'south',
+            de: 'süd',
+            fr: 'sud',
           };
+        }
 
-          // select the 4 most recent Ifrit or Raktapaksa's depending on phase
-          let combatantName = null;
-          if (data.phase === 'ifrit')
-            combatantName = ifritLocaleNames[data.lang];
-          else
-            combatantName = raktapaksaLocaleNames[data.lang];
-
-          let combatantData = null;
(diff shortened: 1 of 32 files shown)


quisquous:
cc @valarnin for extra eyes, as I am not the best JavaScript asynchronous expert.  I think the logic in https://eslint.org/docs/rules/no-async-promise-executor makes sense to me as to why this is not desired in general, but it sure makes some things more awkward too.

Any other eyes to make sure I didn't mess anything up, or complaints about camelCase, feel free to chime in as well.

ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -39,7 +39,11 @@
 
           let combatantData = null;
           if (combatantName) {
-            combatantData = await window.callOverlayHandler({
+            // TODO: maybe it'd be convenient to define at the top level.
+            let callOverlayHandlerSync = async (params) => {
+              return await window.callOverlayHandler(params);
+            };
+            combatantData = callOverlayHandlerSync({
```
valarnin, line 46: Sorry for the wall of code here, but the fix for this linting change isn't simple.

I tested this with new raidemulator and it works properly, but I can't check that this passes the eslint error for async that you were seeing before.

```javascript
      promise: async (data) => {
        const ifritLocaleNames = {
          en: 'Ifrit',
          de: 'Ifrit',
          fr: 'Ifrit',
          ja: 'イフリート',
        };

        const raktapaksaLocaleNames = {
          en: 'Raktapaksa',
          de: 'Raktapaksa',
          fr: 'Raktapaksa',
          ja: 'ラクタパクシャ',
        };

        // select the 4 most recent Ifrit or Raktapaksa's depending on phase
        let combatantName = null;
        if (data.phase === 'ifrit')
          combatantName = ifritLocaleNames[data.lang];
        else
          combatantName = raktapaksaLocaleNames[data.lang];

        let combatantData = null;
        if (combatantName) {
          combatantData = await window.callOverlayHandler({
            call: 'getCombatants',
            names: [combatantName],
          });
        }

        // if we could not retrieve combatant data, the
        // trigger will not work, so just resume promise here.
        if (!(combatantData !== null &&
          combatantData.combatants &&
          combatantData.combatants.length)) {
          data.safeZone = null;
          return;
        }

        // we need to filter for the Ifrit with the highest ID
        // since that one is always the safe spot.
        let currentHighestCombatant =
          combatantData.combatants.sort((a, b) => a.ID - b.ID).pop();

        // all variation ranges for all the 9 ball positions for the kicking actors
        // north      x: 96-104   y: 85-93
        // northeast  x: 107-115  y: 85-93
        // northwest  x: 85-93    y: 85-93
        // east       x: 107-115  y: 96-104
        // west       x: 85-93    y: 96-104
        // south      x: 96-104   y: 107-115
        // southeast  x: 107-115  y: 107-115
        // southwest  x: 85-93    y: 107-115
        let safeZoneObj1 = { en: '', de: '' };
        let safeZoneObj2 = { en: '', de: '' };

        // don't need to go through all the posibilities,
        // only those 4 ifs do reflect the above positions
        if (currentHighestCombatant.PosY > 84 && currentHighestCombatant.PosY < 94) {
          safeZoneObj1 = {
            en: 'north',
            de: 'nord',
            fr: 'nord',
          };
        } else if (currentHighestCombatant.PosY > 106 && currentHighestCombatant.PosY < 116) {
          safeZoneObj1 = {
            en: 'south',
            de: 'süd',
            fr: 'sud',
          };
        }

        if (currentHighestCombatant.PosX > 84 && currentHighestCombatant.PosX < 94) {
          safeZoneObj2 = {
            en: 'west',
            de: 'west',
            fr: 'ouest',
          };
        } else if (currentHighestCombatant.PosX > 106 && currentHighestCombatant.PosX < 116) {
          safeZoneObj2 = {
            en: 'east',
            de: 'ost',
            fr: 'est',
          };
        }

        data.safeZone = {
          en: safeZoneObj1.en + safeZoneObj2.en,
          de: safeZoneObj1.de + safeZoneObj2.de,
          fr: safeZoneObj1.fr + safeZoneObj2.fr,
        };
      }
```

valarnin:
I enabled these package.json changes on the PopupText refactor branch and tested them against that file, and it's only showing problems with the function names. You probably want some other eslint rules as well. I'm not sure what the setting would be though. Here's what I'm getting back from my refactor for linting issues with these changes:

```
/home/valarnin/Source/cactbot/ui/raidboss/popup-text.js
  436:3  error  Identifier '_OnTriggerInternal_GetHelper' is not in camel case        camelcase
  482:3  error  Identifier '_OnTriggerInternal_CheckSuppressed' is not in camel case  camelcase
  492:3  error  Identifier '_OnTriggerInternal_Condition' is not in camel case        camelcase
  503:3  error  Identifier '_OnTriggerInternal_HelperDefaults' is not in camel case   camelcase
  537:3  error  Identifier '_OnTriggerInternal_PreRun' is not in camel case           camelcase
  542:3  error  Identifier '_OnTriggerInternal_DelaySeconds' is not in camel case     camelcase
  562:3  error  Identifier '_OnTriggerInternal_DurationSeconds' is not in camel case  camelcase
  571:3  error  Identifier '_OnTriggerInternal_SuppressSeconds' is not in camel case  camelcase
  577:3  error  Identifier '_OnTriggerInternal_Promise' is not in camel case          camelcase
  599:3  error  Identifier '_OnTriggerInternal_Sound' is not in camel case            camelcase
  603:3  error  Identifier '_OnTriggerInternal_SoundVolume' is not in camel case      camelcase
  608:3  error  Identifier '_OnTriggerInternal_Response' is not in camel case         camelcase
  622:3  error  Identifier '_OnTriggerInternal_AlarmText' is not in camel case        camelcase
  626:3  error  Identifier '_OnTriggerInternal_AlertText' is not in camel case        camelcase
  630:3  error  Identifier '_OnTriggerInternal_InfoText' is not in camel case         camelcase
  634:3  error  Identifier '_OnTriggerInternal_GroupTTS' is not in camel case         camelcase
  645:3  error  Identifier '_OnTriggerInternal_TTS' is not in camel case              camelcase
  658:3  error  Identifier '_OnTriggerInternal_PlayAudio' is not in camel case        camelcase
  710:3  error  Identifier '_OnTriggerInternal_Run' is not in camel case              camelcase
```

It's not catching, for example, the last line here:

```javascript
class PopupText {
  constructor(options) {
    this.options = options;
    this.triggers = [];
    this.timers = {};
    this.CurrentTriggerID = 0;
```

quisquous:
In eslint terms, "camel case" includes what you call "proper case".  In its definition, camel case means not using underscores in the middle of a word.  https://eslint.org/docs/rules/new-cap is the one that differentiates `CurrentTriggerID` from `currentTriggerID`.  There's a LOT more to adjust with that one, which is why I did "camel case" first.

ui/raidboss/data/05-shb/raid/e6s.js
quisquous: Ah this is even better, thanks!

valarnin:
Sorry, I thought those were enabled with your changes to package.json, but it turns out I'm blind. However, even after adjusting package.json such that:

```json
      "new-cap": [
        "error",
        {
          "capIsNew": true,
          "newIsCap": true,
          "properties": true
        }
      ],
```

eslint still isn't catching the case I outlined above (that's on line 28 of popup-text.js, but the first error returned is on line 50).

```
/home/valarnin/Source/cactbot/ui/raidboss/popup-text.js
   50:22  error  A function with a name starting with an uppercase letter should only be used as a constructor  new-cap
```

quisquous:
> eslint still isn't catching the case I outlined above (that's on line 28 of popup-text.js, but the first error returned is on line 50).

Yeah, I think it's only handling functions.  It doesn't really look like eslint has anything for this (weirdly??).  Do you see a rule that I'm missing?

https://github.com/farzonl/eslint-lowerCase-plugin is a plugin, but it only looks half-baked and not really usable.

valarnin:
I think it expects you to define the properties at the class level instead of just in the constructor.

e.g. This:

```javascript
class A {
  someProperty;
  constructor() {
    this.someProperty = 'test';
  }
}
```
Otherwise it doesn't actually treat it as a class property?

quisquous:
> I think it expects you to define the properties at the class level instead of just in the constructor.

Foof, that seems like its own pass through the code to change all the classes to behave like that.  But, probably something that should happen too.

quisquous:
I ran with this through reclears and some other content, and don't feel like I broke anything.  So, here goes.

PR #1383, opened 2020-04-26
small translation script update

Akurosia:
- check_translation had some issues with cases on replacementText
- find_missing_translation: added ability to filter results (path and files) to make it easier to go through all translations step by step

### DIFF
--- a/util/find_missing_translations.py
+++ b/util/find_missing_translations.py
@@ -22,24 +22,23 @@ zoneregex_locales = set(["en", "cn", "ko"])
 non_zoneregex_locales = all_locales - zoneregex_locales
 
 # Where to start looking for files.
-
-
 def base_path():
     return os.path.relpath(os.path.join(os.path.dirname(__file__), "..\\"))
 
 
 # Return a list of all javascript filenames found under base_path()
-def find_all_javascript_files(filefilter):
+def find_all_javascript_files(filter):
     python_files = []
     for root, dirs, files in os.walk(base_path()):
         dirs[:] = [d for d in dirs if d not in ignore_dirs]
 
         for file in files:
-            if filefilter:
-                if filefilter not in root and filefilter not in file:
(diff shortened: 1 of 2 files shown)


util/find_missing_translations.py
```diff
@@ -22,17 +22,22 @@
...
     python_files = []
     for root, dirs, files in os.walk(base_path()):
         dirs[:] = [d for d in dirs if d not in ignore_dirs]
 
         for file in files:
+            if filefilter:
+                if filefilter not in root and filefilter not in file:
+                    continue
             if file.endswith(".js"):
                 python_files.append(os.path.join(root, file))
```
quisquous, line 42: Because you're checking root and file separately, you can't filter for something that includes both.  What about:

```
for file in files:
  if not file.endswith(".js"):
    continue
  full_path = os.path.join(root, file)
  if file_filter in full_path:
    continue
  python_files.append(full_path)
```

util/find_missing_translations.py
```diff
@@ -22,17 +22,22 @@
 non_zoneregex_locales = all_locales - zoneregex_locales
 
 # Where to start looking for files.
+
```
quisquous, line 25: Can you remove these additional newlines?

Akurosia: sry this was auto-formatting of my sublime Q_Q

util/find_missing_translations.py
```diff
@@ -22,17 +22,22 @@
...
+
 def base_path():
     return os.path.relpath(os.path.join(os.path.dirname(__file__), "..\\"))
 
 
 # Return a list of all javascript filenames found under base_path()
-def find_all_javascript_files():
+def find_all_javascript_files(filefilter):
```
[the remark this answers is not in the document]
Akurosia: i will use filter than

PR #1384, opened 2020-04-26
Add NetRegexes

quisquous:
This is for #1328.


JLGarber:
Glad to see this coming together. I'm really happy about having all the positional data available.

Is it worth adding support for `11` (party member IDs) and `36` (limit break gauge) as well? Maybe not for triggers directly, but for Some Future Something jobs-related?

quisquous:
I hooked this up for one test trigger too.

I think the only thing I am not sure about is this very last config, where I deliberately make `gameLog` (and friends) not match the entire line, so that partial lines are ok.

The alternative to this, is that I could also add a `.*?` to the end of every existing gameLog right now for safety.  This would also mean that other gameLogs in the future would also need to do this if they were partial lines.  That said, it does make this one field inconsistent with all other fields.  I am a little bit torn on what the right thing to do here is.

Maybe simplicity / consistency is better?

> Is it worth adding support for `11` (party member IDs) and `36` (limit break gauge) as well? Maybe not for triggers directly, but for Some Future Something jobs-related?

I think this is probably a thing to do for the future.  Ideally, I would really like to have some way to inject log lines, so we could get all of the detailed information from the party event.  The party events are not great on their own.

JLGarber:
> I think the only thing I am not sure about is this very last config, where I deliberately make gameLog (and friends) not match the entire line, so that partial lines are ok.

Given that most of the time we can avoid using `gameLog` in the first place, I think the inconsistency compared to the other fields is fine.

I suppose waiting for line injection does make sense.

panicstevenson:
> The alternative to this, is that I could also add a .*? to the end of every existing gameLog right now for safety. This would also mean that other gameLogs in the future would also need to do this if they were partial lines. That said, it does make this one field inconsistent with all other fields. I am a little bit torn on what the right thing to do here is.

To me, it makes more sense to append `.*?` to each gameLog in the transition to NetRegexes. It may look messier, but I think keeping the mechanics of the code consistent is more important than the semantics, especially since the majority of triggers are added either by referencing older triggers or simply copy/pasting old ones and transforming them. If everything already has `.*?` at the end, there's a good chance that any new triggers would include that as well, even from someone approaching the codebase for a first time.

I think the only concern would be for people adopting the NetRegex change for their custom existing gameLog triggers, but that feels very niche to me since the old functionality is sufficient, and/or anyone adopting NetRegex _hopefully_ is in the loop enough to figure out what needs to happen to make their triggers work in the new format.

But also, like JLGarber said, I think that gameLog is already used so infrequently (for good reason) that it doesn't ultimately matter too much.

JLGarber:
> If everything already has .*? at the end, there's a good chance that any new triggers would include that as well, even from someone approaching the codebase for a first time.

Fair point. I *really* dislike situations where the user has to remember to do a thing, but I suppose that's not a big deal with robust testing, combined with the minimal number of situations this will come up in the first place.

@quisquous just to clarify, are we planning to deprecate `Regexes` for all new trigger files immediately once this is in place? Or is that intended to be a second step after some later milestone is reached? (I know we had planned to wait on explicit conversion of timelines for now.)

quisquous:
> Fair point. I _really_ dislike situations where the user has to remember to do a thing, but I suppose that's not a big deal with robust testing, combined with the minimal number of situations this will come up in the first place.

Yeah, this is my feeling as well.  It's just a question of "what is remembering to do a thing".  I think "every field matches the entire field" is a simpler thing to remember than "every field matches the entire field except for `line` which is special.

> @quisquous just to clarify, are we planning to deprecate `Regexes` for all new trigger files immediately once this is in place? Or is that intended to be a second step after some later milestone is reached? (I know we had planned to wait on explicit conversion of timelines for now.)

I think these can all be separate steps.  I think new trigger files should ideally be written with NetRegexes if possible.  It'd be nice to convert old ones after this.  Timelines need some "automatic" converter to still be written in the background.  Because this isn't breaking user code, I think we can procede at whatever speed we can get code written in.

PR #1387, opened 2020-04-27
Raidemulator redesign

valarnin:
## Summary

This is a draft PR for a redesigned raidemulator. There are still some outstanding issues to resolve, but I want to get this PR open for feedback on the visual and functional side of the redesign.

## Remaining issues to resolve:

- [x] When importing logs, the encounters displayed in the bottom tab don't always refresh and can require a page reload to display properly
- [x] When switching current encounters the selected perspective isn't set correctly
- [ ] A lot of linting cleanup needs to happen
- [x] Static class properties aren't supported by eslint's default parser, so need to pull those outside of the class defintions
- [ ] Need to test this in the actual overlay display, all testing so far has been in browser directly
- [x] Displayed trigger text when seeking isn't cleared properly
- [ ] Languages other than English need tested, both network logs and imported ACT logs
- [x] Documentation for new classes
- [x] Remove jQuery dependency
- [x] Use `Regexes` class to reduce duplicate Regexes
- [x] Intro modal
- [ ] Automatically trim encounter to pre-pull countdown or beginning of encounter, whichever is earlier
- [x] Performance improvements for seeking
- [x] Show selected player's name
- [ ] Make time in encounter bar editable
- [x] Fix regex cloning for trigger display
- [ ] Collapse duplicate triggers that all trigger at once

## Features brought by this redesign:

- New UI
- Tracking triggers per player
- Trigger debug info tracked and displayed
- Can pause, resume, and seek during encounter playback
- Encounters are stored in IndexedDB and can be loaded again later without having to re-import. A 116mb network log uses approximately 60mb of IndexedDB storage
- Encounter parsing now captures pre-pull actions by default, with an option to prune the encounter to the first hostile action or to the Engage line
- Can import Network logs directly without having to go through ACT

## Usage:

1. Open `ui/raidboss/raidemulator.html?OVERLAY_WS=ws://127.0.0.1:10501/ws` with ACT open and listening on a websocket. This is required still due to the way cactbot manifests and timelines are stored.
2. Import logs via ACT or import logs via the web page UI, or even by dragging and dropping a log file on to the web page
3. Expand the encounter tab with the button bottom-right and drill down to the encounter you want to view, then click Load Encounter
4. Select a perspective by clicking a player on the right (first tank's perspective is selected by default)
5. Click play, or click on the progress bar at the top to seek to a spot in the encounter, or hover over the indicators in the bar to see what triggers fire where.
6. Expand the trigger info with the buttons in the trigger info pane. If a trigger didn't pass its condition, it is hidden by default but can be displayed by unchecking the checkbox in the top right.

## Screenshots

image
image
image

## Version 2

Eventually I'll start working on a v2 of this redesign.

### Planned features for v2

- Track 'in party' status based off party update log lines
- Encounter map showing player positions during the fight
- Cache calls to cactbot ACT overlay handlers so that WS connection is only required the first time

### Potential features for v2

- Interactive timeline debugging tool
- Interactive trigger debugging tool


ui/raidboss/emulator/data/Combatant.js
```diff
@@ -0,0 +1,76 @@
...
+    else if (index === lastSignificantStateindex ||
+        timestamp > this.significantStates[lastSignificantStateindex])
+      return this.states[this.significantStates[lastSignificantStateindex]];
+
+    let i = 0;
+    for (; i < this.significantStates.length; ++i) {
+      if (this.significantStates[i] > timestamp)
+        return this.states[this.significantStates[i]];
+    }
+    return this.states[this.significantStates[i]];
```
quisquous, line 36: I'm not sure I understand this loop variable.  Is this just a synonym for "return;" or do you mean to be returning something else?

ui/raidboss/emulator/data/AnalyzedEncounter.js
```diff
@@ -0,0 +1,271 @@
...
+        else
+          ret[i] = jQuery.extend(true, {}, data[i]);
+      } else {
+        ret[i] = data[i];
+      }
+    }
+    return ret;
+  }
+
+  JobToJobEnum(Job) {
```
quisquous, line 266: Maybe this should live next to `Util.jobEnumToJob`?

ui/raidboss/raidemulator.html
```diff
@@ -1,88 +1,195 @@
...
+          </div>
+          <div class="playerTriggers">
+            <div class="triggerBar player0"></div>
+            <div class="triggerBar player1"></div>
+            <div class="triggerBar player2"></div>
+            <div class="triggerBar player3"></div>
+            <div class="triggerBar player4"></div>
+            <div class="triggerBar player5"></div>
+            <div class="triggerBar player6"></div>
+            <div class="triggerBar player7"></div>
```
quisquous, line 108: What about alliance raids?

ui/raidboss/emulator/data/Combatant.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+class Combatant {
+  constructor(id, name) {
+    this.id = id;
+    this.name = name.split('(')[0];
+    this.states = {};
+    this.significantStates = [];
+    this.latesttimestamp = null;
```
quisquous, line 9: I'm trying hard to not leave bikeshed comments about naming things, but ... maybe `latestTimeStamp`?

ui/raidboss/emulator/EmulatorCommon.js
```diff
@@ -0,0 +1,64 @@
...
+  ja: /\[(?<lineTimestamp>[^\]]+)\] 00:.{4}:戦闘開始まで(?<Time>\d+)秒！/,
+  en: /\[(?<lineTimestamp>[^\]]+)\] 00:.{4}:Battle commencing in (?<Time>\d+) seconds!/,
+  de: /\[(?<lineTimestamp>[^\]]+)\] 00:.{4}:Noch (?<Time>\d+) Sekunden bis Kampfbeginn!/,
+  fr: /\[(?<lineTimestamp>[^\]]+)\] 00:.{4}:Début du combat dans (?<Time>\d+) secondes !/,
+  cn: /\[(?<lineTimestamp>[^\]]+)\] 00:.{4}:距离战斗开始还有(?<Time>\d+)秒！/,
+  ko: /\[(?<lineTimestamp>[^\]]+)\] 00:.{4}:전투 시작 (?<Time>\d+)초 전!/,
+};
+
+EmulatorCommon.EventDetailsRegexes = {
+  '03': /\[(?<lineTimestamp>[^\]]+)\] 03:(?<source_id>[^:]+):Added new combatant (?<source_name>[^:]+)\. {2}Job: (?<job>[^:]+) Level: (?<level>[^:]+) Max HP: (?<source_maxHP>\d+) Max MP: (?<source_maxMP>\d+) Pos: \((?<source_posX>[^,)]+),(?<source_posY>[^,)]+),(?<source_posZ>[^,)]+)\)/i,
```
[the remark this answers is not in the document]
valarnin: `resources/regexes.js` doesn't tend to capture all the fields. `03` in particular it only captures the ID and name.

ui/raidboss/emulator/EmulatorCommon.js
```diff
@@ -0,0 +1,64 @@
...
+class EmulatorCommon {
+  static getTimestampFromLogLine(day, line) {
+    let result = EmulatorCommon.logLineRegex.exec(line);
+    return +new Date(day + ' ' + result.groups.lineTimestamp);
+  }
+}
+
+EmulatorCommon.logLineRegex = /^\[(?<lineTimestamp>[^\]]+)\] ([0-9A-Z]+):(.*)$/i;
+
+EmulatorCommon.SealRegexes = {
```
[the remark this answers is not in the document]
valarnin: Same problem as the `EventDetailsRegex`, the existing regexes don't capture all the fields. In this case it's the timestamp. Since the emulator needs to chunk through thousands of lines all at once, every additional regex we need to run (to e.g. pull the timestamp from the line) means a significant increase in processing time.

ui/raidboss/emulator/EmulatorCommon.js
```diff
@@ -0,0 +1,64 @@
...
+  '16': /\[(?<lineTimestamp>[^\]]+)\] 16:(?<source_id>[^:]*?):(?<source_name>[^:]*?):(?<AbilityID>[^:]*?):(?<AbilityName>[^:]*?):(?<target_id>[^:]*?):(?<target_name>[^:]*?):[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:[^:]*?:(?<target_HP>[^:]*?):(?<target_maxHP>[^:]*?):(?<target_MP>[^:]*?):(?<target_maxMP>[^:]*?):[^:]*?:[^:]*?:(?<target_posX>[^:]*?):(?<target_posY>[^:]*?):(?<target_posZ>[^:]*?):(?<target_heading>[^:]*?):(?<source_HP>[^:]*?):(?<source_maxHP>[^:]*?):(?<source_MP>[^:]*?):(?<source_maxMP>[^:]*?):[^:]*?:[^:]*?:(?<source_posX>[^:]*?):(?<source_posY>[^:]*?):(?<source_posZ>[^:]*?):(?<source_heading>[^:]*?):/i,
+  '26': /\[(?<lineTimestamp>[^\]]+)\] 26:(?<source_id>[^:]+):(?<source_name>[^:]*?):[^:]+:(?<source_HP>[^:]+):(?<source_maxHP>[^:]+):(?<source_MP>[^:]+):(?<source_maxMP>[^:]+)/i,
+};
+
+EmulatorCommon.petNames = [
+  'Emerald Carbuncle', 'Topaz Carbuncle', 'Moonstone Carbuncle',
```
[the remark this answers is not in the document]
valarnin: Yeah, it probably makes sense to put this array in `resources/lang/<langcode>.js`.

ui/raidboss/emulator/EmulatorCommon.js
```diff
@@ -0,0 +1,64 @@
...
+  'Ifrit-Egi', 'Titan-Egi', 'Garuda-Egi',
+  'Eos', 'Selene', 'Seraph',
+  'Rook Autoturret', 'Bishop Autoturret', 'Automaton Queen',
+  'Demi-Bahamut', 'Demi-Phoenix',
+  'Esteem', 'Bunshin',
+];
+
+EmulatorCommon.Languages = [
+  'ja',
```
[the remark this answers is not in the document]
valarnin: Yeah, I can move it there.

ui/raidboss/emulator/data/AnalyzedEncounter.js
```diff
@@ -0,0 +1,271 @@
...
+            postData: AnalyzedEncounter.cloneData(data),
+            matches: matches,
+            status: status,
+          });
+        }
+      }
+    }
+  }
+
+  async runTriggers(timestamp, trigger, data, matches, suppressed) {
```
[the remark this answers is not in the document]
valarnin: I wrote this before I did the refactor of PopupText as basically a lift-and-shift of the existing raidemulator logic. It's actually on my todo list for the code cleanup I mentioned in the PR. Now that PopupText has been refactored I can inherit the existing logic and intercept the functions from the `OnTriggerInternal` logic and intercept them there for tracking.

ui/raidboss/emulator/data/Combatant.js
valarnin: Unless I've miscalculated this should return the last significant state, though I haven't put a breakpoint in and double checked that yet.

ui/raidboss/emulator/data/AnalyzedEncounter.js
valarnin: Probably. I was avoiding making changes to the existing files. I'll move this over.

ui/raidboss/raidemulator.html
valarnin: Alliance raid stuff is going to depend on the inParty tracking code. I'm also not sure how to tell alliance membership from the party log lines.

ui/raidboss/emulator/data/Combatant.js
valarnin: This one's actually a mistake, it got caught in my renaming of `Timestamp` => `timestamp` :blush:

ui/raidboss/raidemulator.html
valarnin: Should I move the party tracking stuff up into v1?

quisquous: Ah no, sorry.  It's not clear to me where the v1/v2 line is.  So, maybe leave a TODO for more players and can look in the future.

ui/raidboss/emulator/data/AnalyzedEncounter.js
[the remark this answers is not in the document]
quisquous: Gotcha, sorry, I missed that this is what you meant by that.  :+1:

ui/raidboss/emulator/data/Combatant.js
quisquous: I guess I'm just reading this as, if you get to the end of the loop and don't find a significant state, then `i === this.significantStates.length` and so `return this.states[this.significantStates[this.significantStates.length]];` is a phrase that means `return;`

ui/raidboss/emulator/EmulatorCommon.js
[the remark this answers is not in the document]
quisquous: What do you think about adding them to `addedCombatantFull`?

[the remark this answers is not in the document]
quisquous: Also, if you need the timestamp on all of them, what about an additional `captureTimestamp` option? Or make `capture` return the timestamp as well?

ui/raidboss/emulator/data/Combatant.js
valarnin: Yeah, that's probably the case. Like I said, I didn't sit down with a breakpoint in the code to check this. It never threw any error/exception when it hit. I'll just modify the end return to return the last significant state instead.

ui/raidboss/raidemulator.html
valarnin: The goal would be an alliance selector above or below the party list in the UI, rather than trying to cram 24 player bars into the UI in a stack there.

ui/raidboss/emulator/EmulatorCommon.js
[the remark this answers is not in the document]
valarnin: Probably the existing regexes should just have timestamp added to them. But right now I rely on some automatic mapping (pull log event type, then call `this['convert'+event]` if it exists, otherwise invoke a generic convert method) and I'd have to map every event ID to its `Regexes` entry.

There's also the issue of not knowing what language is in use ahead of time, which means for example having to have a copy of every `CactbotLanguage<code>` instantiated and hitting each of those separately for the seal/unseal/etc regexes, which is going to be pretty ugly.

ui/raidboss/emulator/ui/EncounterTab.js
```diff
@@ -0,0 +1,167 @@
...
+    this.$encounterColumn = $('#encountersTab .encounterList').on('click', '.selectorRow', (ev) => {
+      let t = $(ev.currentTarget);
+      t.parent().children('.selectorRow.selected').removeClass('selected');
+      t.addClass('selected');
+      this.currentEncounter = t.data('index');
+      me.RefreshUI();
+    });
+
+    // @TODO: Probably a better way to do this...
+    this.$infoColumn = $('#encountersTab .encounterInfo').on('click', '.encounterLoad', (ev) => {
```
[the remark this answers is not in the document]
valarnin: jQuery handles binding events to elements further down the DOM tree that match the second argument. This saves having to re-bind these events ourselves every time we re-build the encounters list.

ui/raidboss/emulator/data/AnalyzedEncounter.js
```diff
@@ -0,0 +1,271 @@
...
+    this.popupText = null;
+    this.perspectives = {};
+    this.encounter = encounter;
+  }
+
+  selectPerspective(ID) {
+    let partyMember = this.encounter.combatantTracker.combatants[ID];
+    this.popupText.OnPlayerChange({
+      detail: {
+        name: partyMember.name.split('(')[0],
```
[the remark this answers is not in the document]
valarnin: Sorry, this is server name split in cross-server content. Some lines have it and some don't, and right now name is only read once. It probably should be moved to assigning the combatant's name instead, with server moved to its own property when it's available.

ui/raidboss/emulator/EventBus.js
```diff
@@ -0,0 +1,38 @@
+'use strict';
+// Simple event bus implementation to decouple things
```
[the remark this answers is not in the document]
valarnin: Yeah, I'll add some more documentation for this.

valarnin:
@quisquous Based on the feedback already given, I'd like to propose a couple things:

1. Drop jQuery, it'll increase the code for everything UI-related by about 200 lines.
2. Pull in Parallel.js for webworkers, for the analysis of data. <https://parallel.js.org/>
3. Modify the existing Regexes functions to capture everything required for the actual encounter analysis
4. Move party/alliance tracking into v1

Thoughts/opinions? Should #3 be its own separate PR along with the other non-raid-emulator changes that have been required?

quisquous:
> @quisquous Based on the feedback already given, I'd like to propose a couple things:
> 
> 1. Drop jQuery, it'll increase the code for everything UI-related by about 200 lines.

:+1:

> 2. Pull in Parallel.js for webworkers, for the analysis of data. https://parallel.js.org/

I'd say this sounds like a version 2 sort of thing.

> 3. Modify the existing Regexes functions to capture everything required for the actual encounter analysis

Separate pull request here sounds perfect.  Thanks for understanding.  It's that sort of duplication that easily gets out of date, especially for something like a raid emulator that not everybody is using all the time.

> 4. Move party/alliance tracking into v1

This seems like also ok to push into version 2.  Or, ideally if it gets done by the next alliance raid.  :)

valarnin:
> > 2. Pull in Parallel.js for webworkers, for the analysis of data. https://parallel.js.org/
> 
> I'd say this sounds like a version 2 sort of thing.

The reason that I was thinking about bringing it into v1 is because using that library makes it really easy to implement.

> > 4. Move party/alliance tracking into v1
> 
> This seems like also ok to push into version 2. Or, ideally if it gets done by the next alliance raid. :)

Next alliance raid is exactly why I was thinking of pulling this into v1.

ui/raidboss/emulator/data/AnalyzedEncounter.js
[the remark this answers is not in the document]
valarnin: This is resolved now, with latest commit.

quisquous:
# UI

Could you add some instructions when you first open it up if nothing is loaded?

Can the arrow menu in the bottom right be open by default when the page loads? It seems like everything you want to do to start is in there.

I still kind disagree with your time decision.  I think t=0 should be the pull.  It might be nice to trim negative time to be the first event with a trigger as well, since three minutes of standing around squishes everything else over.  What is your goal around showing that "before pull" time so prominently?  I feel like maybe I'm just not understanding the use case.

I think the "date" selector column in the bottom UI should be narrower.  The width is about three times as long as the date, and the final box gets very crowded at smaller window sizes.

## UI nice-to-haves:

It's reaally slow (~3 seconds) to click on the time slider.  It's also reaaaaaally slow (~6 seconds) to click on a person's name.  Can any of this work be deferred?  Saying "performance is a version 2 thing" is also ok.  It's always easier to do synchronous work.

It'd be nice to have the selected player's name above the HP bars, maybe? As an addition to the green box around the job.  I see that there's a mouseover there, but that's not quite visible enough.

Is it possible to collapse multiple triggers with the same log line? For example, something like `E7S Insatiable Light` which hits multiple people, but the log line is the same with the same time and data is the same afterwards.  I suppose I could add a suppression on this, but it's just noisy in the trigger view.

Could the time on the time slider be editable?

Could the last loaded fight be stored, and then reloaded by default on page load?

# Implementation

HP Display has some errors.  Sometimes Max MP is zero and sometimes it's 10k.  I think you just need to always set it to 10k.  There's also an issue where sometimes your hp values are incorrect.  I've seen it say "-0.00004792213/81950" (81950 is also not the max hp for this player).  I see that on the e7s log linked below.

image

None of the regexes appear to have anything listed.  If I click on a trigger in the trigger view, I see something like:
```
{
  "triggerHelper": {
    "trigger": {
      "id": "E7S Silver Shot",
      "regex": {},  // <- shouldn't there be something here?
      "filename": "05-shb/raid/e7s.js",
      "localRegex": {}
    },
```

Upon opening the first time, I got a ton of these errors repeatedly (but I couldn't repro):
```
RaidEmulatorTimelineController.js:42 Uncaught TypeError: Cannot read property '0' of undefined
    at RaidEmulatorTimelineController.OnLogEvent (RaidEmulatorTimelineController.js:42)
    at raidemulator.js:127
    at processEvent (common.js:239)
    at WebSocket.<anonymous> (common.js:189)
```

Importing alex_20510_20200429.log
* Why is there an Alexander Prime fight in the Mist zone?
* Looking at the fight names, a8s has a lot of different fight names.  Maybe it'd be worth finding the mob with the most hit points in the zone / sealed area for consistency?
* Console error: `Timeline file not found: 03-hw/raid/a8s.txt`

Importing e7s.log
* Same issue with a 4 minute fight in the Mist zone.
* When playing back the longest e7s fight, many many missing triggers on the leftmost view.  Hallowed shows up, but the first aoe does not, and most others do not.
* Same console errors about the timeline not being found.

I'll send you these logs in DM because they're too large to attach here.

valarnin:
> # UI
> Could you add some instructions when you first open it up if nothing is loaded?
> 
> Can the arrow menu in the bottom right be open by default when the page loads? It seems like everything you want to do to start is in there.

These two are doable. Do you want the check to be 'nothing in the indexeddb', or 'first time loading'?

> I still kind disagree with your time decision. I think t=0 should be the pull. It might be nice to trim negative time to be the first event with a trigger as well, since three minutes of standing around squishes everything else over. What is your goal around showing that "before pull" time so prominently? I feel like maybe I'm just not understanding the use case.

The goal wasn't to show the three minutes before the pull, but the immediate pre-pull opener/positioning/etc. I agree as well that displaying 3 minutes is a bit useless. Maybe it makes sense to by default trim to start of countdown, or 5 seconds before pull if there's no countdown? Things also get weird detecting player/enemy/ally/pet state with missing 03 events if those lines are trimmed out, which is why I left them in originally.

> I think the "date" selector column in the bottom UI should be narrower. The width is about three times as long as the date, and the final box gets very crowded at smaller window sizes.

I'll adjust the date column to be fixed-width, since it's always got the same width for content.

> ## UI nice-to-haves:
> It's reaally slow (~3 seconds) to click on the time slider. It's also reaaaaaally slow (~6 seconds) to click on a person's name. Can any of this work be deferred? Saying "performance is a version 2 thing" is also ok. It's always easier to do synchronous work.

The further into the fight you're seeking, the slower it is, since it has to play back the encounter from the start to that position. There's two ways to optimize this:

1. When seeking forward, play from where you were previously instead of the beginning of the encounter. This should be a relatively easy adjustment, but it only addresses if you seek forward.
2. Pre-cache the popup state at every possible log line timestamp for every player in the encounter and load that state when seeking. This would be really memory intensive and cause the initial encounter load to be really slow (an 11 minute encounter with 8 players would take upwards of 50 seconds to do this).

> It'd be nice to have the selected player's name above the HP bars, maybe? As an addition to the green box around the job. I see that there's a mouseover there, but that's not quite visible enough.

The area above the HP bars is where the map will go in V2. I could shift the bars down a bit to put the selected player's name in place.

> Is it possible to collapse multiple triggers with the same log line? For example, something like `E7S Insatiable Light` which hits multiple people, but the log line is the same with the same time and data is the same afterwards. I suppose I could add a suppression on this, but it's just noisy in the trigger view.

It should only show the trigger if it fired for the given player. In the case of `Insatiable Light` stacks, it's two separate log lines, one for each stack target. Would it make sense to hide the trigger by default if there's no displayed text as well, behind the same checkbox as the `didn't actually run` hide?

> Could the time on the time slider be editable?

I'm not sure what the UX expectation would be for this. Would you enter a custom time and it would seek to that time on blur? Or on enter? If it's an invalid time what should the result be?

> Could the last loaded fight be stored, and then reloaded by default on page load?

Yes, this is doable.

> HP Display has some errors. Sometimes Max MP is zero and sometimes it's 10k. I think you just need to always set it to 10k. There's also an issue where sometimes your hp values are incorrect. I've seen it say "-0.00004792213/81950" (81950 is also not the max hp for this player). I see that on the e7s log linked below.

I'll look into the logs you DM'd me. Looks like maybe something is off by one field (e.g. it's reading current hp from the heading field, and max hp from the current hp field).

I avoided any sort of hard-coding of values just in case next expac they decide 'oh hey, we want max mp to be 5k instead of 10k'.

> None of the regexes appear to have anything listed. 

I think this is a problem with my custom clone code. Probably, I need to add a specific case for RegExp classes.

> Upon opening the first time, I got a ton of these errors repeatedly (but I couldn't repro):

I think this can happen if you're in-game. I need to modify the `RaidEmulatorPopupText` class to avoid the `PopupText` constructor so it doesn't bind to the `OnLogEvent` handler, otherwise it's going to do weird stuff if you're in-game and log events are firing.

> Importing alex_20510_20200429.log
> Importing e7s.log

Will check these logs and track down the problems.

quisquous:
> > Could you add some instructions when you first open it up if nothing is loaded?
> > Can the arrow menu in the bottom right be open by default when the page loads? It seems like everything you want to do to start is in there.
> 
> These two are doable. Do you want the check to be 'nothing in the indexeddb', or 'first time loading'?

I think "nothing in the indexeddb" is better, since that's really the state where somebody will need to know what to do.  Especially since somebody might open it, not load anything, and then need instructions the second time.

> > I still kind disagree with your time decision. I think t=0 should be the pull. It might be nice to trim negative time to be the first event with a trigger as well, since three minutes of standing around squishes everything else over. What is your goal around showing that "before pull" time so prominently? I feel like maybe I'm just not understanding the use case.
> 
> The goal wasn't to show the three minutes before the pull, but the immediate pre-pull opener/positioning/etc. I agree as well that displaying 3 minutes is a bit useless. Maybe it makes sense to by default trim to start of countdown, or 5 seconds before pull if there's no countdown? Things also get weird detecting player/enemy/ally/pet state with missing 03 events if those lines are trimmed out, which is why I left them in originally.

To me there's a difference between storing that information somewhere and displaying that information so prominently.  I would be fine with you doing everything you're already doing internally, but I just don't need to know anything in the UI before t=0.

My other reasons here are: including that time also skews fight lengths (which then will differ even farther from fflogs and ACT encounters).  It also makes "this trigger happens at t=2m30s" much harder to parse when it's got some prepull time on it.  It seems unlikely that anybody wants to know where somebody was standing two minutes before pull even if it's useful for you to have that state internally.

> > ## UI nice-to-haves:
> > It's reaally slow (~3 seconds) to click on the time slider. It's also reaaaaaally slow (~6 seconds) to click on a person's name. Can any of this work be deferred? Saying "performance is a version 2 thing" is also ok. It's always easier to do synchronous work.
> 
> The further into the fight you're seeking, the slower it is, since it has to play back the encounter from the start to that position. There's two ways to optimize this:
> 
> 1. When seeking forward, play from where you were previously instead of the beginning of the encounter. This should be a relatively easy adjustment, but it only addresses if you seek forward.
> 2. Pre-cache the popup state at every possible log line timestamp for every player in the encounter and load that state when seeking. This would be really memory intensive and cause the initial encounter load to be really slow (an 11 minute encounter with 8 players would take upwards of 50 seconds to do this).

It doesn't have to be all or nothing.  You could do something like video compression and lazily add keyframes every 30 seconds or 60 seconds and then seek from there? I think it's also a question about page interactivity when you click.

> > It'd be nice to have the selected player's name above the HP bars, maybe? As an addition to the green box around the job. I see that there's a mouseover there, but that's not quite visible enough.
> 
> The area above the HP bars is where the map will go in V2. I could shift the bars down a bit to put the selected player's name in place.

I don't feel strongly that "above" is the right answer, but "show the selected player's name somewhere" seems useful.

> > Is it possible to collapse multiple triggers with the same log line? For example, something like `E7S Insatiable Light` which hits multiple people, but the log line is the same with the same time and data is the same afterwards. I suppose I could add a suppression on this, but it's just noisy in the trigger view.
> 
> It should only show the trigger if it fired for the given player. In the case of `Insatiable Light` stacks, it's two separate log lines, one for each stack target. Would it make sense to hide the trigger by default if there's no displayed text as well, behind the same checkbox as the `didn't actually run` hide?

Hmm, when debugging, some triggers that don't show text are still worth knowing that they run, just not they ran redundantly 8 times.  That's why I was trying to get at "this is a no-op duplicate" and not "this is a trigger than doesn't show anything".

> > Could the time on the time slider be editable?
> 
> I'm not sure what the UX expectation would be for this. Would you enter a custom time and it would seek to that time on blur? Or on enter? If it's an invalid time what should the result be?

Hrm.  If the recalculation of triggers was asynchronous (and thus it's very fast to just update the position in the time slider), then I would say move the time on every keypress and then do the update for real on blur.  If invalid then use the original time on blur?

I don't know.  Feel free to punt to V2.  I was just thinking about how I would want to use it. :)

valarnin:
@quisquous I've been working through removing the jQuery dependency, and I've got a good method for it that doesn't result in a ton of extra js code. The only snag now is the tooltips. Bootstrap implements its tooltips through jQuery + Popper (see https://getbootstrap.com/docs/4.5/components/tooltips/ and https://getbootstrap.com/docs/4.5/getting-started/javascript/#dependencies).

I'm not sure how to resolve this. Any thoughts?

The two options I can see are:

1. Leave jQuery in, but only use it for tooltips
2. Interface with Popper directly or otherwise replicate the Bootstrap tooltip js code

quisquous:
> The two options I can see are:
> 
> 1. Leave jQuery in, but only use it for tooltips
> 2. Interface with Popper directly or otherwise replicate the Bootstrap tooltip js code

I think popper looks pretty straightforward, but if it's a pain, then go with #1.

> Edit to add: Modals also rely on jQuery.

There are modals??

valarnin:
> There are modals??

The Yes/No dialog to confirm clearing the indexeddb is in a modal, and the instructions on empty indexeddb when you load the page are going to be a modal.

Also probably going to add a loading modal of some kind since it's not obvious when it's doing something that takes a while to process.

I can replicate the modal code outside of jQuery, that's fairly easy. The annoying part is the Popper stuff. I'll poke at it when I get the chance again, maybe Tuesday, and see if I can get it working without jQuery.

valarnin:
Alright, jQuery fully removed. Had to ditch Popper as well, since there was too much intertwined logic and styling between jQuery/Bootstrap/Popper to easily replicate the functionality. Replaced that with a Tooltip class that's pretty simple.

ui/raidboss/emulator/EventBus.js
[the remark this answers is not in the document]
valarnin: @quisquous I've added documentation to EventBus. Is this sufficient or do you want more documentation for this class? I plan on adding more documentation to the rest of the new classes as well.

ui/raidboss/raidemulator.html
quisquous: Belated thought, but ... is there actually a way to know which alliance somebody is in other than just "not in your immediate party"?

ui/raidboss/emulator/EventBus.js
```diff
@@ -0,0 +1,53 @@
+'use strict';
+
+/**
+ * This is a base class that classes can extend to inherit event bus capabilities.
+ * This allows other classes to listen for events with the `on` function.
+ * The inheriting class can fire those events with the `dispatch` function.
```
quisquous, line 6: Thanks, perfect! Sorry for my ignorance in not knowing what this was.

quisquous:
Also, a random followup, I think you'll need to add support for `NetRegexes` too.

(As a bonus, this would let you drop all of the "figure out what job people are on" code because you can just read it right out of the added combatant line.)

quisquous:
Hey, I just wanted to check in and understand what your plans were here.  It's 10000% fine if "I have no time or brain for this, sorry" is the answer, but I just wanted to check in and see where you were.  Mostly in a logistics sense, the old raid emulator is broken and this new one isn't ready yet, and I'm trying to figure out what the best way forward is, since a couple of folks have asked about it at this point.

valarnin:
This is 100% on my todo list still, and I've got more free time for it now that my static is done raiding for the tier. I'll see about getting the requisite git management stuff done ASAP and get some motion on this this weekend.

valarnin:
Alright, origin/master renamed to main and this has been rebased to account for changes (properly I hope) per #1623. Struggled with git like always to get it to cooperate but finally getting back to this.

Given that old raidemulator is currently broken, what are the must-haves from the first comment to get this merged into trunk to resolve the issue?

Also, I haven't been following the source-level changes to triggers since the conversion to use `NetRegexes` for everything. Is there anything still using `Regexes` that I need to do a conversion of network log line to game log line for still, or can I drop that part of the code entirely?

quisquous:
> Alright, origin/master renamed to main and this has been rebased to account for changes (properly I hope) per #1623. Struggled with git like always to get it to cooperate but finally getting back to this.

Thanks!  Glad you got it sorted.  I'm happy to try to assist in this sort of thing too if you need it.

> Also, I haven't been following the source-level changes to triggers since the conversion to use `NetRegexes` for everything. Is there anything still using `Regexes` that I need to do a conversion of network log line to game log line for still, or can I drop that part of the code entirely?

There are still some for HP, and I suspect that users probably want to test their own, and they might still be using Regexes.  I would say since you've already written the conversion, you might as well keep that code? I think it could be a v2 if you wanted it, but I think it's important.

> Given that old raidemulator is currently broken, what are the must-haves from the first comment to get this merged into trunk to resolve the issue?

I think my feelings are, if I can drag a Network log in, and I can play it back, and I can see/investigate triggers in English, even if it's janky or seeking is painful (previously seeking didn't exist), then that's good enough to get a v1 done.

### done

> - [x] When importing logs, the encounters displayed in the bottom tab don't always refresh and can require a page reload to display properly
> - [x] When switching current encounters the selected perspective isn't set correctly
> - [x] Static class properties aren't supported by eslint's default parser, so need to pull those outside of the class defintions
> - [x] Remove jQuery dependency
> - [x] Use `Regexes` class to reduce duplicate Regexes
> - [x] Intro modal
> - [x] Fix regex cloning for trigger display

### must have

* support NetRegexes
* support ZoneId (but keep zoneRegex support)

> - [ ] Show selected player's name

I think this is important.

> - [ ] Documentation for new classes
> - [ ] A lot of linting cleanup needs to happen

These two are good to the extent that it will help me understand the code? It doesn't have to be super extensive, but anything to help me find my way around would be helpful.

### v2 / later / not required

> - [ ] Displayed trigger text when seeking isn't cleared properly

I'm not sure what this means.

> - [ ] Need to test this in the actual overlay display, all testing so far has been in browser directly
> - [ ] Languages other than English need tested, both network logs and imported ACT logs
> - [ ] Automatically trim encounter to pre-pull countdown or beginning of encounter, whichever is earlier
> - [ ] Performance improvements for seeking
> - [ ] Make time in encounter bar editable
> - [ ] Collapse duplicate triggers that all trigger at once

valarnin:
> > Also, I haven't been following the source-level changes to triggers since the conversion to use NetRegexes for everything. Is there anything still using Regexes that I need to do a conversion of network log line to game log line for still, or can I drop that part of the code entirely?

> There are still some for HP, and I suspect that users probably want to test their own, and they might still be using Regexes. I would say since you've already written the conversion, you might as well keep that code? I think it could be a v2 if you wanted it, but I think it's important.

I'm going to have to refactor the code that's storing the logs in local storage regardless, since it's only storing the converted logs right now, so I wanted to know what should be required before I start on that.

> > [ ] Displayed trigger text when seeking isn't cleared properly

> I'm not sure what this means.

When you seek and it displays the alert/info/etc text, it doesn't clear it out properly. So if you seek to 60s into the pull, it'll show the text for the entire first 60 seconds for the next 3 seconds.

valarnin:
@quisquous Right now, when I'm doing the conversion from network log, I have to build a list of abilities so that the ability name can be mapped for HoT/DoT entries for certain edge cases. Should I instead add a new python script to pull the abilities from XIVAPI?

See the calls to addAbility here:
https://github.com/valarnin/cactbot/blob/raidemulator-redesign/ui/raidboss/emulator/data/NetworkLogConverter.js#L81-L107

And where it's actually used here:
https://github.com/valarnin/cactbot/blob/raidemulator-redesign/ui/raidboss/emulator/data/NetworkLogConverter.js#L209-L216

And the list of abilities that I've found need their name shown to match the normal game logs here:
https://github.com/valarnin/cactbot/blob/raidemulator-redesign/ui/raidboss/emulator/data/NetworkLogConverter.js#L467-L478

quisquous:
> > > Also, I haven't been following the source-level changes to triggers since the conversion to use NetRegexes for everything. Is there anything still using Regexes that I need to do a conversion of network log line to game log line for still, or can I drop that part of the code entirely?
> 
> > There are still some for HP, and I suspect that users probably want to test their own, and they might still be using Regexes. I would say since you've already written the conversion, you might as well keep that code? I think it could be a v2 if you wanted it, but I think it's important.
> 
> I'm going to have to refactor the code that's storing the logs in local storage regardless, since it's only storing the converted logs right now, so I wanted to know what should be required before I start on that.

Ah, yeah.  I suspect you're going to need to store the network logs instead of the parsed logs, and then parse them afterwards.

> > > [ ] Displayed trigger text when seeking isn't cleared properly
> 
> > I'm not sure what this means.
> 
> When you seek and it displays the alert/info/etc text, it doesn't clear it out properly. So if you seek to 60s into the pull, it'll show the text for the entire first 60 seconds for the next 3 seconds.

Ah, I see! Making seeking better sounds like a v2, then.

> @quisquous Right now, when I'm doing the conversion from network log, I have to build a list of abilities so that the ability name can be mapped for HoT/DoT entries for certain edge cases. Should I instead add a new python script to pull the abilities from XIVAPI?
> 
> See the calls to addAbility here:
> https://github.com/valarnin/cactbot/blob/raidemulator-redesign/ui/raidboss/emulator/data/NetworkLogConverter.js#L81-L107
> 
> And where it's actually used here:
> https://github.com/valarnin/cactbot/blob/raidemulator-redesign/ui/raidboss/emulator/data/NetworkLogConverter.js#L209-L216
> 
> And the list of abilities that I've found need their name shown to match the normal game logs here:
> https://github.com/valarnin/cactbot/blob/raidemulator-redesign/ui/raidboss/emulator/data/NetworkLogConverter.js#L467-L478

Can you help me understand why the HoT/DoT lines are important to the raid emulator? I don't think anything matches on these, and so I kind of expected them to be dropped.

valarnin:
> Can you help me understand why the HoT/DoT lines are important to the raid emulator? I don't think anything matches on these, and so I kind of expected them to be dropped.

The main reason is for the combatant state tracking, since they contain the combatant's current/max hp/mp and x/y/z/heading. The only reason I cared about ability name when it's not in the network log for the packet is because it's in the converted format, and even then only for certain abilities. I could just not worry about that piece and ignore the ability name portion, sacrificing that bit of accuracy.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x18.js
```diff
@@ -32,9 +32,9 @@ class LineEvent0x18 extends LineEvent {
   }
 
   convert(repo) {
-    this.abilityName = repo.Abilities[this.abilityId.toUpperCase()];
+    this.abilityName = LineEvent0x18.showAbilityNamesFor[this.abilityId.toUpperCase()];
```
quisquous, line 35: Could you also s/Ability/Effect/ to make it clearer what these ids are?

valarnin: I can rename the `Ability` portions without any issue, but renaming `this.abilityId` will impact `CombatantTracker` here:
https://github.com/valarnin/cactbot/blob/57721fc884569564dc7310f894e4d8e1831602f9/ui/raidboss/emulator/data/CombatantTracker.js#L134-L135

I'm not sure if it makes sense to add `effectId` to the CombatantTracker code, or exclude it altogether since it's probably not accurate and has just been working because it should never theoretically fire against a 0x18 line.

quisquous: I think it will impact `CombatantTracker`, but in a positive manner towards a more correct state.  Currently, `this.abilityId` for this line type is an effect id and not an ability id.  I haven't looked if there's any collisions between effect ids in the 0x18 list and ability ids in the combatant search list, but it seems like a potential risk.

I think it's fine to only use abilities to figure out who combatants are.  Abilities are probably sufficient and it's fine to exclude effects.

valarnin: Should be good now.

valarnin:
Alright, the only thing still outstanding as far as I know is a linting cleanup pass, to adjust variable and class property/method names. What's the preferred naming scheme, because eslint doesn't seem to be consistently enforcing it? Something like this?

```javascript
class SomeClass {
  constructor() {
    this.someProperty = '';
    this._someInternalProperty = '';
  }
  SomeFunction(someParameter) {
    let someVar = '';
  }
  _someInternalFunction() {
  }
}

SomeClass.someStaticProperty = '';
SomeClass.someStaticFunction = () => {};
```

quisquous:
> Alright, the only thing still outstanding as far as I know is a linting cleanup pass, to adjust variable and class property/method names. What's the preferred naming scheme, because eslint doesn't seem to be consistently enforcing it? Something like this?

Re: eslint.  Nothing has changed since our discussion in #1379 and #1378.  eslint does not enforce new-cap yet because I haven't gotten around to enabling it.  I'm sorry the code is inconsistent.

> ```js
> class SomeClass {
>   constructor() {
>     this.someProperty = '';
>     this._someInternalProperty = '';
>   }
>   SomeFunction(someParameter) {
>     let someVar = '';
>   }
>   _someInternalFunction() {
>   }
> }
> 
> SomeClass.someStaticProperty = '';
> SomeClass.someStaticFunction = () => {};
> ```

Yes, other than `SomeFunction` as `someFunction`.  This is probably the most inconsistent cactbot style bit, and so I'm not going to complain about either one as long as you're consistent with surrounding code.  I am aiming for `someFunction` as the eventual target to land on.

valarnin:
I should be fine to enable new-cap, fix the files for this PR, and then disable new-cap, correct? If so that should be sufficient from the linting side from my perspective.

quisquous:
Yeah, that seems sufficient to me as well.  Thanks!

valarnin:
Alright, linting done. Fixed some stuff that wasn't caught by new-cap as well, via regex searching. Were there any other changes/documentation you wanted for this, or is it ready for merge now?

ui/raidboss/emulator/overrides/RaidEmulatorTimelineUI.js
```diff
@@ -0,0 +1,143 @@
...
+    this.init = true;
+    this.timerlist = { clear: () => { } };
+
+    this.activeBars = {};
+    this.expireTimers = {};
+  }
+
+  // Overrride
+  InitDebugUI() {
+    this.debugElement = {};
```
quisquous, line 97: Ahhh, I misunderstood what this bug was.  `debugElement` expects to be an element.  I think this needs to be `this.debugElement = document.createElement('div');` otherwise you get a bunch of errors when `Options.Debug` is true.

quisquous:
That's my only concern, though. Let's get this merged.  :tada:

(I think I am surprised a little that new-cap didn't catch things like `ID` as parameter names, but that can get fixed later.  Maybe there's more eslint settings I need to figure out.)

ui/raidboss/emulator/overrides/RaidEmulatorTimelineUI.js
valarnin: I've made that adjustment but realistically we should probably display that div in some manner in the future I think.

quisquous:
:confetti_ball:

PR #1389, opened 2020-04-28
raidboss: default css - infoText color to white

### LINKED ISSUE #1352: raidboss: default css - infoText color
## Summary

I'm using white color for infoText, green color wasn't standing out against green grass platform while Titania EX.
K-001

I've applied some other slight changes to css, but I don't know if those changes are good for everyone.
However, I think white color for infoText is definitely better than green one.
+I'm not a designer so I'm not good at coloring, so I didn't tried changing text-shadow css, my best is black, simple.
If comments says this change would be good, I'll pull request here.

Jaehyuk-Lee:
Fixes #1352

### DIFF
--- a/ui/raidboss/raidboss.css
+++ b/ui/raidboss/raidboss.css
@@ -57,7 +57,7 @@
 }
 
 .info-text {
-  color: #2b2;
+  color: #fff;
   font-size: 140%;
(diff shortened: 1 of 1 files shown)


quisquous:
Let's do it.  The time for comments has passed.

PR #1391, opened 2020-04-28
eureka: remove unused code

Jaehyuk-Lee:
I think those code make volume not customizable.


Jaehyuk-Lee:
Ah, I remember bunny fate didn't alarm with 0 configured. I think that configure is working, but those codes are still useless.

PR #1392, opened 2020-04-29
Add Ja translations for configs / 様々な設定項目を日本語へ通訳しました

maiko-tan-coding:
But I don't know if these translations were the best, so I would make this Pull Request as a draft so that others could provide some advice.

最適な通訳かどうかわかりませんので、一応DraftとしてPull Request出します。皆さんのご意見はコメントで自由にお話ししてください！<( _ _ )>


ui/radar/radar_config.js
```diff
@@ -44,6 +47,7 @@ UserConfig.registerOptions('radar', {
         en: 'Volume to play pop sound (0-1)',
         de: 'Lautstärke für den Popsound (0-1)',
         fr: 'Volume du son (0-1)',
+        ja: '知らせ音声の音量（0-1）',
```
r4ha, line 50: This should be "お知らせ音声の音量 (0-1)" I guess.

ui/oopsyraidsy/oopsyraidsy_config.js
```diff
@@ -21,6 +22,7 @@ UserConfig.registerOptions('oopsyraidsy', {
         en: 'Number of mistakes to show in combat',
         de: 'Anzahl der Fehler, die während des Kampfes angezeigt werden',
         fr: 'Nombre de fautes à afficher en combat',
+        ja: '戦闘の時に表示する最大の間違い回数',
```
r4ha, line 25: This should be "戦闘中ミスをした回数".

ui/jobs/jobs_config.js
```diff
@@ -56,6 +60,7 @@ UserConfig.registerOptions('jobs', {
         en: 'Show MP ticker out of combat',
         de: 'Zeige MP-Ticker auserhalb des Kampfes',
         fr: 'Afficher symbole MP si hors combat',
+        ja: '非戦闘の時もMPを表示する',
```
r4ha, line 63: This should be "非戦闘時にもMPを表示する".

ui/jobs/jobs_config.js
```diff
@@ -32,6 +34,7 @@ UserConfig.registerOptions('jobs', {
         en: 'Opacity of ui when out of combat',
         de: 'Deckkraft der UI auserhalb des Kampfes',
         fr: 'Opacité de l\'UI si hors combat',
+        ja: '非戦闘の時UIの透過度',
```
r4ha, line 37: This should be "非戦闘時のUI透過度".

ui/jobs/jobs_config.js
```diff
@@ -20,6 +21,7 @@ UserConfig.registerOptions('jobs', {
         en: 'Lower ui opacity when out of combat',
         de: 'Veringere die Deckkraft auserhalb des Kampfes',
         fr: 'Diminiuer l\'opacité de l\'UI si hors combat',
+        ja: '非戦闘の時にUIを透過する',
```
r4ha, line 24: This should be "非戦闘時にUIを透過する".

r4ha:
Hi! I'm a native Japanese speaker. Just dropped some suggestions. Overall, unify brackets with single-byte is better than 2-bytes bracket I guess.

ui/oopsyraidsy/oopsyraidsy_config.js
maiko-tan-coding: Do you think ミス is better than 間違い to describe mistakes in this situation?

I have written that before but one of my Japanese friends told me that ミス was ambiguity sometimes. What do you think about that?

maiko-tan-coding:
> Hi! I'm a native Japanese speaker. Just dropped some suggestions. Overall, unify brackets with single-byte is better than 2-bytes bracket I guess.

Thanks for your help, I would change it soon.
ありがとうございます。ヾ(•ω•`)o

ui/oopsyraidsy/oopsyraidsy_config.js
maiko-tan-coding: I think this option refers to the display-able maximum number of mistakes that players did in combat. If the number of their mistakes was more than this value, then the over part would not be displayed. So it could not be simply 戦闘中ミスをした回数.

it is more like 戦闘中に記録したミス回数がその数値を超えた場合は表示できません。

r4ha: I think those two words mean the same thing. If you want to use 間違い," 戦闘中に表示する間違えた回数". Most of Japanese use ミス here imo. so I suggested.

maiko-tan-coding: Oh sorry, I misunderstand your suggested translation. Yours is right.
And if most Japanese use ミス, I would use that, too.
Thank you very much for your help again.

r4ha: Nah maybe you saw my comment before I edited. My bad!

PR #1400, opened 2020-05-01
plugin: separate language support for localizable texts

### LINKED ISSUE #1074: Set cactbot config shown language to PC's locale.
## Summary

Brief explanation of the feature you'd like to request.

Like this, shown language is not determined by FFXIV parsing plugin's language, I think this is based on PC's locale setting.
K-001

cactbot is in English (FFXIV parsing plugin's language)
K-002

I think cactbot can determine in which language user is using their PC.
I don't know how though.

Also, I've tried to match config's language setting to `Alert Language`'s language setting. 
Example,
If `Use FFXIV Plugin Language` - follow FFXIV Plugin's language
If `English` - fix to English whatever user's PC language is.

But... I couldn't do that.

Jaehyuk-Lee:
Fixes #1074

## config screen with updated dll file
K-001

## Important part
K-001-2

Before: There were 
```
Langauge: en
Version: Intl
```

After: With my change
```c
Parsing Plugin Langauge: en // was 'Language' before
PC Locale: ko-KR // new added information here
Version: Intl
```

PC Locale now used for config language, not parsing plugin's language.

## Enforce 'zh' to 'cn'
```javascript
Options.ShortLocale = e.detail.locale.substring(0, 2);
if (Options.ShortLocale == 'zh')
  Options.ShortLocale = 'cn';
```
Chinese locale is `zh-**`, but cactbot is using 'cn' for Chinese locale (or language).
So I've enforce to change to 'cn'.

## Why do I check supportedLanguage?
```javascript
const supportedLanguage = ['en', 'de', 'fr', 'ja', 'cn', 'ko'];
if (supportedLanguage.indexOf(shortLocale) == -1)
  shortLocale = 'en';
this.lang = shortLocale;
```
Most localized text are shown in English if not user's language is supported. But sometimes, there can be a mistake forgot to add `|| 'en'` which is used to enforce use English if user's language is not supported.
Upper code will prevent this problem for other regions.
Example, `ru-RU`. (There were no 'ru' thing coming before since there's no FFXIV Parsing Plugin Language for Russian.)
Config text becomes `[Object object]` without `|| 'en'`.

## download for test
[CactbotEventSource.zip](https://github.com/quisquous/cactbot/files/4564783/CactbotEventSource.zip)
* config screen is not changed to your locale before you apply my change with `resources/user_config.js` and `ui/config/config.js`, you can only check `PC Locale: ko-KR` thing on Overlay plugin Log with this dll file.

## other helpful link
Langauge Code Table: http://www.lingoes.net/en/translator/langcode.htm

## minor change with html class about language
After timeline supports changing language, language info in html class should be timelineLanguage.


plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -47,6 +47,7 @@ public class CactbotEventSource : EventSourceBase, ILogger {
     private FateWatcher fate_watcher_;
 
     private string language_ = null;
+    private string pc_locale = null;
```
quisquous, line 50: Can you name this `pc_locale_` with an underscore, to match the other member variables?

ui/raidboss/timeline.js
```diff
@@ -583,7 +583,7 @@ class TimelineUI {
 
     this.root = document.getElementById('timeline-container');
     if (Options.Language)
-      this.root.classList.add('lang-' + Options.Language);
+      this.root.classList.add('lang-' + Options.TimelineLanguage);
```
quisquous, line 586: Oh, thanks for this fix!

ui/config/config.js
```diff
@@ -55,7 +55,11 @@ class CactbotConfigurator {
       'jobs': [],
     };
     this.configOptions = configOptions;
-    this.lang = configOptions.Language || 'en';
+    let shortLocale = configOptions.ShortLocale;
+    const supportedLanguage = ['en', 'de', 'fr', 'ja', 'cn', 'ko'];
+    if (supportedLanguage.indexOf(shortLocale) == -1)
```
quisquous, line 60: Can you put this code to validate the locale into one of the valid languages up in user_config.js? That way anybody else using it can just use variable directly without also having to do this.

plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -667,7 +675,9 @@ protected override void Update()
       var result = new JObject();
       result["userLocation"] = config_dir;
       result["localUserFiles"] = user_files == null ? null : JObject.FromObject(user_files);
+
       result["language"] = language_;
+      result["locale"] = pc_locale;
```
quisquous, line 680: Sorry for other naming suggestions, but I think "locale" and "language" are a bit too synonymous.

What do you think about:
```
result["parserLanguage"] = language_;
result["systemLanguage"] = pc_locale_;
// For backwards compatibility:
result["language"] = language_;
```

Jaehyuk-Lee: I understand some people can be confused with two words, but locale is actually not meaning language.
I sent full locale information because there can be different locale for same language.
example)
`zh-CN` and `zh-TW` use different Chinese character, Simplified one and Traditional one.
I don't know western languages well, so I can't give you another example.

So, to compromise two ideas, I'd better suggest
```
result["parserLanguage"] = language_;
result["systemLocale"] = pc_locale_;
// For backwards compatibility:
result["language"] = language_;
```

Jaehyuk-Lee: I was afraid of breaking something by changing variable names, but that backwards compatibility part will make okay with changes.
But I couldn't figure out how javascript codes get variable from c# code, only `resources/user_config.js` is coping with this? And javascript files are so many, and I don't know most of them, so I couldn't change "language" variable name to "parserLanguage" I will break something if I try to do it.
So, those codes are somehow confusing now, if language variable is using Parser's language or System language or even user's setting. I think this should be made clear.

quisquous: My understanding is that locale is a combination of language plus other things.  So, they're not quite the same (and cactbot code interchangeably uses the words locale and language in some places, sorry <_<) but without extra words it's not clear which is the parser and which is the system.  Your suggestion of parserLanguage vs systemLocale sounds great, thanks.

That `result` object automatically becomes `e.detail` here https://github.com/quisquous/cactbot/blob/master/resources/user_config.js#L42.  So if you add more fields, you can just access them by name on `e.detail.systemLocale`.  I think that's the only place that language gets accessed.

Jaehyuk-Lee: better understanding of locale is language plus country.
`en-GB` - English, Great Britain
`en-US` - English, United States
`fr-FR` - French, France
`ru-RU` - Russian, Russia
`zh-cn` - zhōngwén(maybe), China

I don't know exact way of making locale string, but it's simply `language-country(or sometimes maybe location)`

Jaehyuk-Lee:
I had this error so I changed some code.
```
/home/travis/build/Jaehyuk-Lee/cactbot/ui/raidboss/popup-text.js
  238:1  error  This line has a length of 102. Maximum allowed is 100  max-len
```

I think both line works the same. If doesn't please tell me. I had to make this line shorten.
```javascript
            let netRegex = trigger[netRegexParseLang] ? trigger[netRegexParseLang] : trigger.netRegex;
            let netRegex = trigger[netRegexParseLang] || trigger.netRegex;
```

Jaehyuk-Lee:
I've removed manual language setting on `fisher.js`, it worked well without it.
(Tested on English client and Japanese client, thanks to free login campaign :smiley:)
9e1390d

Jaehyuk-Lee:
I've changed comment and text saying "locale" to "parser language" thing. I don't understand exactly what that means, so some transition might wrong. Please check this. thanks!
d9f469e

resources/user_config.js
```diff
@@ -39,8 +39,20 @@ let UserConfig = {
...
+        // Backward compatibility, everything "Language" should be changed to "ParserLanguage"
+        Options.Language = e.detail.parserLanguage;
+      }
+      if (e.detail.systemLocale) {
+        Options.SystemLocale = e.detail.systemLocale;
+        Options.ShortLocale = e.detail.SystemLocale.substring(0, 2);
+        if (Options.ShortLocale == 'zh')
+          Options.ShortLocale = 'cn';
+        const supportedLanguage = ['en', 'de', 'fr', 'ja', 'cn', 'ko'];
+        if (supportedLanguage.indexOf(Options.ShortLocale) == -1)
```
quisquous, line 53: Funnily enough, #1408 is adding this list.  I think this is fine for now, and we can merge this to use the other later.

Can you also make this `if (Options.ShortLocale.includes(supportedLanguage))` instead of using `indexOf`?

ui/eureka/eureka.js
```diff
@@ -2067,7 +2071,7 @@ class EurekaTracker {
 
         let remainingMinutes = Math.ceil(remainingMs / 1000 / 60);
         let nmString = respawnIcon + remainingMinutes +
-            this.Translate(this.options.timeStrings.minute);
+            this.TransByLocale(this.options.timeStrings.minute);
```
quisquous, line 2074: Why are these translated by locale vs by parser language? Can you leave this as `Translate` for now, and if you want to follow up in another pull request with adding a "display language" for everything (vs just assuming that somebody wants the system locale), we can add more options (maybe a top level one?), like raidboss has alert/timeline language options.

ui/raidboss/timeline.js
```diff
@@ -43,11 +43,11 @@ class Timeline {
       return text;
 
     let orig = text;
-    let locale = this.options.Language || 'en';
+    let parseLang = this.options.PraserLanguage || 'en';
```
quisquous, line 46: typo here and elsewhere

ui/raidboss/popup-text.js
```diff
@@ -228,14 +228,14 @@ class PopupText {
...
-            let regex = trigger[regexLocale] ? trigger[regexLocale] : trigger.regex;
+            // parser-language-based regex takes precedence.
+            let regex = trigger[regexParseLang] || trigger.regex;
             if (regex) {
               trigger.localRegex = Regexes.parse(regex);
               this.triggers.push(trigger);
             }
 
-            let netRegex = trigger[netRegexLocale] ? trigger[netRegexLocale] : trigger.netRegex;
+            let netRegex = trigger[netRegexParseLang] || trigger.netRegex;
```
quisquous, line 238: This seems fine!

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -552,7 +552,7 @@ class MistakeCollector {
     // wipe then (to make post-wipe deaths more obvious), however this
     // requires making liveList be able to insert items in a sorted
     // manner instead of just being append only.
-    this.OnFullMistakeText('wipe', null, kPartyWipeText[this.options.Language || 'en']);
+    this.OnFullMistakeText('wipe', null, kPartyWipeText[this.options.ShortLocale || 'en']);
```
quisquous, line 555: For the same reason as the eureka ones above, I think this would be better served as "display language" rather than assuming that the user wants the system locale.  Can you undo this to use the parse language again?

resources/user_config.js
valarnin: I think this is supposed to be `if (supportedLanguage.includes(Options.ShortLocale))`?

quisquous: Oops thanks, sorry.  I really just meant "use includes" here.

docs/FAQ-Troubleshooting.md
```diff
@@ -53,16 +53,9 @@ You need to have at least one chat log set to show `Own Gathering Messages` You
...
-- **fisher.js language setting**
-If your game has a text language setting other than english you will need to open /ui/fisher/fisher.js and change the language option to your chosen language:
-  - `en` - English
-  - `de` - German
-  - `fr` - French
-  - `jp` - Japanese
-  - `cn` - Chinese
-
+- Supported language
 Korean is not supported at this time.
```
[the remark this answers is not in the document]
Jaehyuk-Lee: Korean is not supported since there's no Korean data on `static.js`

resources/user_config.js
Jaehyuk-Lee: Is there any reason to use `includes`?
checking `indexOf` is way more faster than `includes`
Screenshot 2020-05-06 at 13 05 49
tested from: http://jindo.dev.naver.com/jsMatch/index.html

docs/FAQ-Troubleshooting.md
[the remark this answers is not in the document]
quisquous: Haha whoops, thanks.  I forgot that there were still missing fish there.

Jaehyuk-Lee:
I wrote wrong code there and two codes actually had same speed, so removed that comment but you've seend it :sweat_smile:
I thought extreme performance improvement worth than readability.

quisquous:
Ha, thanks for the info.  I'm glad to know!

quisquous:
Thanks for all the changes.  I think this all looks good.  I want to try this out myself in game (and hopefully I should have time tomorrow or Friday), but I don't have any other feedback at this point.

As a followup, do you think we should remove alerts language / timeline language and have those use DisplayLanguage as well?

Jaehyuk-Lee:
I think it's better to be remained. Skill names are like "proper noun" thing. Skill name's translation to each language client is not far different but rarely, their names are very different. So, this means people can't understand timeline with their display language since they see client's skill name.
Ummm... Someone would want to see as parsing plugin's language. I don't know if we should change to selection of "parsing plugin language" and "display language". It's okay for remaining as now.

resources/user_config.js
```diff
@@ -39,8 +39,25 @@ let UserConfig = {
 
       // The plugin auto-detects the language, so set this first.
       // If options files want to override it, they can for testing.
-      if (e.detail.language)
-        Options.Language = e.detail.language;
+      if (e.detail.parserLanguage) {
+        Options.ParserLanguage = e.detail.parserLanguage;
+        // Backward compatibility, everything "Language" should be changed to "ParserLanguage"
+        Options.Language = e.detail.parserLanguage;
```
quisquous, line 45: Could you add a section at the top with this code, that would then be overwritten by parserLanguage and systemLocale and displayLanguage?

```
if (e.detail.language) {
  Options.ParserLanguage = e.detail.language;
  Options.ShortLocale = e.detail.language;
  Options.DisplayLanguage = e.detail.language;
}
```

Mostly so that folks who are using the latest on github won't break when this lands? I don't normally care that much about breaking people on plugin updates, but this seems like an easy fix.

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -458,8 +458,8 @@ class MistakeCollector {
   Translate(obj) {
     if (obj !== Object(obj))
       return obj;
-    if (this.lang in obj)
-      return obj[this.lang];
+    if (this.parserLang in obj)
```
quisquous, line 461: This function looks like it is only used for display language.  Can you maybe rename this to TransByDispLang?

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -493,7 +493,7 @@ class MistakeCollector {
     }
     let seconds = ((Date.now() - this.startTime) / 1000);
     if (this.firstPuller && seconds >= this.options.MinimumTimeForPullMistake) {
-      let text = kEarlyPullText[this.lang] + ' (' + seconds.toFixed(1) + 's)';
+      let text = kEarlyPullText[this.parserLang] + ' (' + seconds.toFixed(1) + 's)';
```
quisquous, line 496: Display language?

resources/user_config.js
```diff
@@ -39,8 +39,25 @@ let UserConfig = {
...
+      const supportedLanguage = ['en', 'de', 'fr', 'ja', 'cn', 'ko'];
+      if (e.detail.systemLocale) {
+        Options.SystemLocale = e.detail.systemLocale;
+        Options.ShortLocale = e.detail.systemLocale.substring(0, 2);
+        if (Options.ShortLocale == 'zh')
+          Options.ShortLocale = 'cn';
+        if (!supportedLanguage.includes(Options.ShortLocale))
+          Options.ShortLocale = Options.ParserLanguage;
+      }
+      if (e.detail.displayLanguage) {
```
quisquous, line 56: Can you add some code such that if `e.detail.displayLanguage` is null, that `Options.DisplayLanguage` still gets set?

ui/raidboss/timeline.js
```diff
@@ -43,11 +43,11 @@ class Timeline {
       return text;
 
     let orig = text;
-    let locale = this.options.Language || 'en';
+    let parserLang = this.options.PraserLanguage || 'en';
```
quisquous, line 46: typo, here and below

also maybe this variable should just be called "replaceLang" or just "lang" since it's not necessarily the parser language.

ui/raidboss/popup-text.js
```diff
@@ -174,11 +174,11 @@ class PopupText {
       }
     }).bind(this);
 
-    let locale = this.options.Language || 'en';
+    let parserLang = this.options.ParserLanguage || 'en';
     // construct something like regexEn or regexFr.
-    let localeSuffix = locale.charAt(0).toUpperCase() + locale.slice(1);
-    let regexLocale = 'regex' + localeSuffix;
-    let netRegexLocale = 'netRegex' + localeSuffix;
+    let localeSuffix = parserLang.charAt(0).toUpperCase() + parserLang.slice(1);
```
quisquous, line 179: Maybe `langSuffix` and not `localeSuffix`?

ui/raidboss/timeline.js
```diff
@@ -43,11 +43,11 @@ class Timeline {
...
     if (replaceKey === 'replaceText')
-      locale = this.options.TimelineLanguage || this.options.Language || 'en';
+      parserLang = this.options.TimelineLanguage || this.options.PraserLanguage || 'en';
```
quisquous, line 48: I wonder a little if the default for TimelineLanguage and AlertsLanguage should be DisplayLanguage and not ParserLanguage.  What do you think?

ui/raidboss/timeline.js
```diff
@@ -43,11 +43,11 @@ class Timeline {
...
     for (let r of this.replacements) {
-      if (r.locale && r.locale != locale)
+      if (r.parserLang && r.parserLang != parserLang)
```
quisquous, line 50: This line is a bug that I only caught because I tried this out in game.  This needs to be `if (r.locale && r.locale != parserLang)` (sorry).  This is referring to fields called `locale` inside the replacement block in trigger files.

ui/raidboss/timeline.js
```diff
@@ -582,8 +582,8 @@ class TimelineUI {
     this.init = true;
 
     this.root = document.getElementById('timeline-container');
-    if (Options.Language)
-      this.root.classList.add('lang-' + Options.Language);
+    if (Options.TimelineLanguage)
```
quisquous, line 585: Hmm.  This isn't quite true either, because `TimelineLanguage` could be unset.

Maybe it'd be worth doing something like:
`this.options.TimelineLanguage = this.options.TimelineLanguage || this.options.ParserLanguage || 'en';`

(or storing that in some other `this.timelineLanguage` variable if you don't want to modify options?)

And then TimelineLanguage will be set correctly.  Without this, if you don't set timeline language explicitly, then you won't get the CSS.  (Previously you only got the CSS for the parser language, which was also wrong.)

ui/oopsyraidsy/oopsyraidsy.js
Jaehyuk-Lee: isn't this `TransByParserLang`?

Jaehyuk-Lee:
I actually understood you wanted separate pull request but I was like easier to make changes in this branch because I have done some work for that displayLanguage thing. I was thinking that would be better for manual language selection which has higher priority of applying language than auto selected system language before you say about adding displayLanguage setting.

ui/oopsyraidsy/oopsyraidsy.js
quisquous: I don't think so.  It's only used for text that is displayed (like reasons for death, etc) and not for regexes.  Am I reading the code incorrectly?

quisquous:
> I actually understood you wanted separate pull request but I was like easier to make changes in this branch because I have done some work for that displayLanguage thing.

Ah, thanks for clarifying.  I think I just don't want you to feel like this pull request is taking forever to merge because there keeps being more things.  Sometimes it feels (for my own pull requests, at least?) simpler to merge one thing, and then do the next, rather than keep trying to append more and more into the same pull request.  Sometimes though it is harder and more complicated to break things up, so if you were already in the middle of it, then that's quite fine.

Mostly I am just apologizing for leaving so many comments here.  I just really think this is a nice feature and want to make sure it applies everywhere and makes sense with everything else that already exists.  :)

> I was thinking that would be better for manual language selection which has higher priority of applying language than auto selected system language before you say about adding displayLanguage setting.

Ok, so to reiterate what you are saying, you're saying that you also think that the priority for languages should go:

AlertLanguage/TimelineLanguage > DisplayLanguage > ParserLanguage > English

(And for things that don't have AlertLanguage/TimelineLanguage, then just start with DisplayLanguage, if set.)

> I gathered all changes about displayLanguage into one commit by resetting and pushing with force option, I thought this would be enough to review because reviewer have to see only one commit.

Yeah, this is fine! I think extra commits can make it easier to see what has changed, but it is not a big deal for me either way.

ui/oopsyraidsy/oopsyraidsy.js
Jaehyuk-Lee: ah, yes this is displayLanguage thing. 
ninja rabbit thing was translated by this.
I thought this was translating skill names of boss, but it isn't.

ui/raidboss/timeline.js
Jaehyuk-Lee: TimelineLanguage > ParserLanguage > 'en'
AlertsLanguage > DisplayLanguage > 'en'
looks good for me, as I said before, 
> Skill names are like "proper noun" thing.

Timeline contains mostly skill names, so it's better to be same as client's language.
However, Alerts doesn't, so display language would be better.

ui/raidboss/timeline.js
Jaehyuk-Lee: remove `if (Options.TimelineLanguage)` and adding 
```javascript
this.root.classList.add('lang-' + Options.TimelineLanguage || Options.ParserLanguage || 'en');
```
would fix this.

Jaehyuk-Lee:
> Ok, so to reiterate what you are saying, you're saying that you also think that the priority for languages should go:
>AlertLanguage/TimelineLanguage > DisplayLanguage > ParserLanguage > English

I want
TimelineLanguage > ParserLanguage > English
(AlertLanguage > )DisplayLanguage > ParserLanguage > English
So, no display language for timeline

resources/user_config.js
Jaehyuk-Lee: I think removing `if` and adding ` || 'en'` would be okay.

quisquous:
> > Ok, so to reiterate what you are saying, you're saying that you also think that the priority for languages should go:
> > AlertLanguage/TimelineLanguage > DisplayLanguage > ParserLanguage > English
> 
> I want
> TimelineLanguage > ParserLanguage > English
> (AlertLanguage > )DisplayLanguage > ParserLanguage > English
> So, no display language for timeline
> This is not actually a big deal to remove display language part on timeline language, because users can select manually with timeline language setting.

I think I disagree a tiny bit, but not very strongly.  It feels kind of inconsistent, and I worry a bit about the future bug report where somebody says "I put my display language in French, so why are the timeline entries in English??" However! I think you are much more in touch with what people are expecting when playing with multiple languages, and if you think this will make much more sense, then I will listen to your advice that timeline language should not use display language.

In other words, I am in agreement with what you said above.  Thank you for all the discussion.

Does that mean that the option for alert language should say "Use display language by default" instead of "use plugin language" or whatever it says now?

Jaehyuk-Lee:
oh yes, I think "Use display language by default" is better for Alerts.

Jaehyuk-Lee:
I don't know other language translation, so ... I didn't make any changes on alert language setting, someone should fix this.

ui/raidboss/timeline.js
```diff
@@ -43,11 +43,11 @@ class Timeline {
       return text;
 
     let orig = text;
-    let locale = this.options.Language || 'en';
+    let replaceLang = this.options.PraserLanguage || 'en';
```
quisquous, line 46: typo here, and below

ui/raidboss/raidboss_config.js
```diff
@@ -907,7 +907,7 @@ UserConfig.registerOptions('raidboss', {
           '朝鲜语 (ko)': 'ko',
         },
         ko: {
-          'FFXIV Plugin 언어 사용': 'default',
+          '주 사용 언어 사용': 'default',
```
quisquous, line 910: Can you remove the de/fr/cn sections above just so folks know that they need to be translated? I'd rather them be missing than be incorrect, because nobody will notice if they are just incorrect.

Alternatively, maybe @Akurosia @VanSlanzar @maiko-tan-coding @ShadyWhite can chime in with translations before this gets merged? (Sorry for ping!)

ui/config/data/general_config.js
```diff
@@ -41,5 +41,76 @@ UserConfig.registerOptions('general', {
       type: 'checkbox',
       default: false,
     },
+    {
+      id: 'DisplayLanguage',
+      name: {
+        en: 'Display language',
```
VanSlanzar, line 47: Thank you for mentionned me, here :
`fr: 'Langue d\'affichage',`

ui/config/data/general_config.js
```diff
@@ -41,5 +41,76 @@ UserConfig.registerOptions('general', {
...
+          'Benutze FFXIV Plugin Sprache': 'default',
+          'Englisch (en)': 'en',
+          'Chinesisch (cn)': 'cn',
+          'Deutsch (de)': 'de',
+          'Französisch (fr)': 'fr',
+          'Japanisch (ja)': 'ja',
+          'Koreanisch (ko)': 'ko',
+        },
+        fr: {
+          'Utiliser la langue du Plugin FFXIV': 'default',
```
VanSlanzar, line 71: Just correction default in french, here :
`'Utiliser la langue du Plugin FFXIV' : 'par défaut',`

ui/raidboss/raidboss_config.js
VanSlanzar: need correction line 883 
` 'Utiliser la langue d\'affichage : 'par défaut',`
For the rest, it's good.

ui/config/data/general_config.js
Jaehyuk-Lee: latter is not localized I think, it's used as common thing.
like `if(language == "ko")` thing?

VanSlanzar: I did not understand.
If you don't think it's useful to translate it, don't do it. I would look on the software after its update.

quisquous: The values on the right hand side here are not shown to users and is only used by code.  The only valid values are 'default' or something like 'en' or 'fr' etc.  They should not be translated or the code will break.

ui/raidboss/raidboss_config.js
quisquous: Other folks can follow up with pull requests for translations in other languages (sorry).

ui/config/data/general_config.js
Jaehyuk-Lee: yes, that's what I was to say. my bad English :(

quisquous: Ah no, I understood what you meant.  I was trying to say it differently in the hopes that our comments combined would communicate clearly here.  :)

PR #1406, opened 2020-05-03
Raidboss: Sophia EX timeline update

JLGarber:
Oof. Sephirot was relatively simple. Sophia was not. There are hidden skips all over before getting to the rotation block. I'm reasonably confident of the timings but it was a struggle to get it all worked out. This final iteration successfully made it through `test_timeline` with More Than Five different log files.

Arms Of Wisdom seems to be on its own sort-of-separate-but-not-really timer during the post-intermission block. It's definitely in the rotation, but it can be skipped or not sometimes based on DPS. I wasn't able to get it closer than about 0.5 seconds of accuracy on the syncs. (It's likely possible to get closer, but there's a point of diminishing returns.)

Because of all the issues with awful timings, there aren't a lot of safety windows included. Syncing to the clones appearing and to the Gnosis knockback is relatively safe, but beyond that I've just left it.

I think tilt severity is fixed at each point in the rotation, but for the timeline I wasn't too concerned about it. (It's going to matter a good deal for triggers.) Preliminarily, `196E` seems to be associated with unequal tethers (not necessarily unbalanced!) while `19A7` seems to be for equal tether counts. It's all very annoying and I wish Square had arranged things more sensibly.

The enrage is a repeated 99,999 `1AE0` Cintamani that pierces tank invulnerabilities, at ~11:10. Unfortunately, there's no way whatsoever to warn the user via the timeline. The only indication that the enrage is occurring is that all eight players get the same color of Sin/Suffering markers. As soon as those resolve, the first Cintamani goes off and kills anyone who's still alive.


JLGarber:
I knew I shouldn't have pushed this directly before bed. Totally forgot about the translations.

There may be other stuff to add once we get triggers in place, this is just the minimum necessary to get the timeline integrated.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
-888	"Arms of Wisdom"
-899	"Scale Tethers"
-903	"Dischordant Damnation"
-910	"Tilt"
+hideall "--Reset--"
+hideall "--sync--"
+
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+0 "Start"
```
quisquous, line 14: Can you add an Engage! sync as well?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+1.8 "--sync--" sync /:Sophia:19C3:/ window 1.8,2
+12.0 "Thunder II" sync /:Sophia:19B0:/ window 12,5
+22.4 "Gnosis" sync /:Barbelo:19C2:/
+41.4 "Arms Of Wisdom?" sync /:Sophia:19C4:/ # May be skipped depending on DPS
+
+# This section will be reached either at 56.0, or at the time she is pushed to 90%
+56.0 "--Clones Appear--" sync / 03:........:Added new combatant Aion Teleos/ window 56,5
```
quisquous, line 21: Don't these always appear in the same places and are not targetable? This might be worth just making ` --sync--` and not including in the visible timeline.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+69.7 "Duplicate" sync /:Aion Teleos:19AB:/
+79.8 "Execute" sync /:Sophia:19AA:/
+#79.8 "Execute" sync /:Aion Teleos:19AF:/
+90.2 "Arms Of Wisdom" sync /:Sophia:19C4:/
+
+# Intermission.
+# Divine Spark can be pushed and thus is not synced.
+# Gnostic Rant and Infusion are purely HP-based pushes.
+110.6 "Cloudy Heavens" sync /:Sophia:19BE:/ window 110.6
+111.6 "--Untargetable--"
```
quisquous, line 34: I know you like capital letters, but lower case untargetable for consistency.

Can you add a --targetable-- line for the adds as well?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+
+0 "Start"
+1.8 "--sync--" sync /:Sophia:19C3:/ window 1.8,2
+12.0 "Thunder II" sync /:Sophia:19B0:/ window 12,5
+22.4 "Gnosis" sync /:Barbelo:19C2:/
+41.4 "Arms Of Wisdom?" sync /:Sophia:19C4:/ # May be skipped depending on DPS
+
+# This section will be reached either at 56.0, or at the time she is pushed to 90%
+56.0 "--Clones Appear--" sync / 03:........:Added new combatant Aion Teleos/ window 56,5
+63.6 "Thunder II/Thunder III" sync /:Sophia:19(AC|B0):/
```
quisquous, line 22: `II/III` for brevity?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+129.0 "Ring Of Pain" sync /:The Third Demiurge:19BA:/
+135.1 "Horizontal Kenoma/Vertical Kenoma" sync /:The First Demiurge:19B[BC]:/
+147.3 "Gnostic Spear?" sync /:The Third Demiurge:19B9:/
+149.2 "Horizontal Kenoma/Vertical Kenoma" sync /:The First Demiurge:19B[BC]:/
+191.6 "Zombification Enrage?"
+
+# There's an introductory block leading into a rotation.
+250.0 "The Scales Of Wisdom" sync /:Sophia:1981:/ window 250,5
+267.5 "The Scales Of Wisdom" sync /:Sophia:1AE1:/
+272.9 "--Targetable--"
```
quisquous, line 47: :eyes:

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+280.8 "Quasar (Meteor Snapshot)" sync /:Sophia:19(6E|A7):/
```
quisquous, line 48: What about `Quasar (Snapshot)` vs `Quasar (Meteor)` vs `Quasar (Tilt)` for brevity? Feel free to disagree.  :)

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+393.3 "Duplicate" sync /:Aion Teleos:19AB:/
+393.3 "Thunder II/Thunder III" sync /:Sophia:19(AC|B0):/
+399.4 "Duplicate" sync /:Aion Teleos:19AB:/
+399.4 "Aero III" sync /:Sophia:19AE:/ window 30,5
+408.5 "Arms Of Wisdom" sync /:Sophia:19C4:/
+419.8 "Light Dew" sync /:Barbelo:19BF:/
+420.5 "Execute" sync /:Sophia:19AA:/ window 10,10
+
+# Rotation block.
+# Technically it can loop, but unless party DPS is VERY high,
```
quisquous, line 77: Is it worth just making it loop to avoid writing it out a second time?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+361.5 "Cintamani x2" duration 3 # sync /:Sophia:19C5:/
+370.8 "Arms Of Wisdom" sync /:Sophia:19C4:/ window 30,5
+379.0 "--Clones Appear--" sync / 03:........:Added new combatant Aion Teleos/ window 60,20
+382.0 "Cintamani x3" duration 6 # sync /:Sophia:19C5:/
+393.3 "Duplicate" sync /:Aion Teleos:19AB:/
+393.3 "Thunder II/Thunder III" sync /:Sophia:19(AC|B0):/
+399.4 "Duplicate" sync /:Aion Teleos:19AB:/
+399.4 "Aero III" sync /:Sophia:19AE:/ window 30,5
+408.5 "Arms Of Wisdom" sync /:Sophia:19C4:/
+419.8 "Light Dew" sync /:Barbelo:19BF:/
```
quisquous, line 73: What is this ability, just for my own curiosity?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -1,157 +1,147 @@
...
+12.0 "Thunder II" sync /:Sophia:19B0:/ window 12,5
+22.4 "Gnosis" sync /:Barbelo:19C2:/
+41.4 "Arms Of Wisdom?" sync /:Sophia:19C4:/ # May be skipped depending on DPS
+
+# This section will be reached either at 56.0, or at the time she is pushed to 90%
+56.0 "--Clones Appear--" sync / 03:........:Added new combatant Aion Teleos/ window 56,5
+63.6 "Thunder II/Thunder III" sync /:Sophia:19(AC|B0):/
+63.6 "Duplicate" sync /:Aion Teleos:19AB:/ window 30,2.5
+69.7 "Aero III" sync /:Sophia:19AE:/
+69.7 "Duplicate" sync /:Aion Teleos:19AB:/
```
quisquous, line 25: Do you think it'd be worth hiding Duplicate from the timeline?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: Normally I would agree, but here I think it's probably better to sync to it and to have it visible. For at least the first two sets, it's a landing point if DPS is pushed, but if it's not pushed, players should still be told "hey, the clone set is coming up".

Also no, the clones have 10 fixed points where they can appear. I have a short explanation in my preliminary trigger file, but I'll just copy that in here:

>The Aion Teleos clones have 10 fixed points where they can appear, but not all combinations of points are valid:
      (0,9), (0,-9), (-10,9), (-10,-9), (10,9), (10,-9), (-15,9), (-15,-9), (15,9), (15,-9).
      Each quadrant can contain 0 or 1 clones, and the center can have 0-2.
      There will always be 4 or 5 clones.
      There can never be more than 3 clones north or south.

I think it's at least a little helpful for tanks who might want to reposition her to the center before the clones go off? Once they spawn, she goes into a long chain of casts and it's really hard to move her.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: Sure thing. As for the adds, those are targetable at the same time as Sophia's disappearance. Maybe put an identically-timed `--adds spawn--` line alongside?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: Very happy that first Scales Of Wisdom is there so we can sync as soon as the intermission ends!

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: I thought about it, but not all languages use the same convention for elemental names. (German, for instance, has no numbers.) If we do that, we will not be displaying the game name for these abilities on the timeline. I'm okay with that, but just be aware that that's what will happen.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: We can do that, sure. I did this because it made the jump for the loop clean, but we could do a fake lookahead and sync from there. I typically avoid doing that but we can here.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: The daughter's orange laser shot.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: You're right, that does always follow a `Quasar (Meteor)` line, so it wouldn't be ambiguous.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: Can do. Is there a particular reason I couldn't do a small modification to `make_timeline` so that's automatic? I feel like I've forgotten that in a number of recent timelines. >_>

ui/raidboss/data/03-hw/trial/sophia-ex.txt
quisquous: :+1:

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: I suppose we could. I left it in as a "hey, look at the tethers" bit. Maybe we could leave it in for now with a note to remove once triggers are in place?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
quisquous: I think the translation blocks can include a `'Thunder II/III': 'etc'` line there, in case they want to simplify the output in some language-specific way? This is along the lines of simplifications in other timelines, I think.  Or, do you think this is too cumbersome?

ui/raidboss/data/03-hw/trial/sophia-ex.txt
quisquous: Sounds reasonable!

ui/raidboss/data/03-hw/trial/sophia-ex.txt
quisquous: :+1:

ui/raidboss/data/03-hw/trial/sophia-ex.txt
quisquous: I think either removing it (esp since there's already a sync on that line) or leaving a note for the future would be fine.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
JLGarber: I can do that. To be clear, this means "put the [language] translation for the base skill in front of the numbers, and then let the [language] community figure it out? As `'Thunder II/III': 'Blitz II/III'`, and then leave it for them? (The German progression would normally be `Blitz/Blitzra/Blitzga`.)

JLGarber: Okay, I think that's what we were going for.

PR #1407, opened 2020-05-03
raidboss: fill Korean responses

Jaehyuk-Lee:
I believe Dorito Stack means to stack with person who's having this marker. (image below)
(I don't know what Dorito means, but translator says it's something "marker")
doritoStack
If not, please tell me.


Akurosia:
afaik Dorito stack are the dark red triangles which may turn you into a frog or explode when you stand alone (e.g. https://crystaluniverse.de/wp-content/uploads/2018/08/Stackmarks.png=

Jaehyuk-Lee:
ah, that thing! thanks!

PR #1409, opened 2020-05-03
Added --rotate-- line raidboss/common_replacement.js

VanSlanzar:
-  (Rotate), this is a line on Seiryu, i think it can be in (common).
@Akurosia @JLGarber  @Jaehyuk-Lee  @maiko-tan-coding, can you help me fix missing translations, I'm afraid google translation is not appropriate or accurate. Sorry for the inconvenience. Thank You all.

- (d(?:e|u|es) $1 dans), I would like to add this because I think this is missing. Is usefull ? @quisquous 
ffxiv_03052020_183950_812 **Ouverture du**
ffxiv_03052020_190950_668**Ouverture de**

### DIFF
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -11,7 +11,7 @@ let commonReplacement = {
   },
   'is no longer sealed': {
     de: 'öffnet sich erneut',
-    fr: 'Ouverture ',
+    fr: 'Ouverture d(?:e|u|es) $1 dans',
     ja: 'の封鎖が解かれた',
     ko: '의 봉쇄가 해제되었습니다',
   },
@@ -90,6 +90,13 @@ let commonReplacement = {
     ja: '--北西--',
     cn: '--西北--',
     ko: '--북서--',
+  },
+  '--rotate--': {
+    de: '--rotate--', // FIXME
+    fr: '--tounez--',
+    ja: '--rotate--', // FIXME
+    cn: '--龙回转--', // FIXME
+    ko: '--rotate--', // FIXME
   },
   '--south--': {
     de: '--süden--',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -96,7 +96,7 @@ let commonReplacement = {
     fr: '--tounez--',
     ja: '--rotate--', // FIXME
     cn: '--龙回转--', // FIXME
-    ko: '--rotate--', // FIXME
+    ko: '--회전--', // FIXME
   },
   '--south--': {
     de: '--süden--',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -3,7 +3,7 @@
 // TODO: maybe this should be structured identically to a timelineReplace section.
 let commonReplacement = {
   '(?<=00:0839:)(.*) will be sealed off(?: in (?:[0-9]+ seconds)?)?': {
-    de: 'Noch 15 Sekunden, bis sich (?:(?:der|die|das) )?(?:Zugang zu(?:[rm]| den)? )?$1 schließt',    
+    de: 'Noch 15 Sekunden, bis sich (?:(?:der|die|das) )?(?:Zugang zu(?:[rm]| den)? )?$1 schließt',
     fr: 'Fermeture d(?:e|u|es) $1 dans',
     ja: '$1の封鎖まであと',
     cn: '距$1被封锁还有',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -3,15 +3,15 @@
 // TODO: maybe this should be structured identically to a timelineReplace section.
 let commonReplacement = {
   '(?<=00:0839:)(.*) will be sealed off(?: in (?:[0-9]+ seconds)?)?': {
-    de: 'Noch 15 Sekunden, bis sich (?:(?:der|die|das) )?(?:Zugang zu(?:[rm]| den)? )?$1 schließt',
-    cn: '距$1被封锁还有',
+    de: 'Noch 15 Sekunden, bis sich (?:(?:der|die|das) )?(?:Zugang zu(?:[rm]| den)? )?$1 schließt',    
     fr: 'Fermeture d(?:e|u|es) $1 dans',
     ja: '$1の封鎖まであと',
+    cn: '距$1被封锁还有',
     ko: '15초 후에 $1(이|가) 봉쇄됩니다',
   },
   'is no longer sealed': {
     de: 'öffnet sich erneut',
-    fr: 'Ouverture d(?:e|u|es) $1 dans',
+    fr: 'Ouverture ',
     ja: 'の封鎖が解かれた',
     ko: '의 봉쇄가 해제되었습니다',
   },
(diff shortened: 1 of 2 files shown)


Jaehyuk-Lee:
https://github.com/quisquous/cactbot/blob/e3098f5214a121c30de0e128ad14dc9459800cb9/ui/raidboss/data/04-sb/trial/seiryu-ex.js#L616
rotate Korean just go '회전'

VanSlanzar:
I just realized that I had missed it thanks to you. @Jaehyuk-Lee

JLGarber:
Your changes to `--rotate--` are fine. I personally would not add something to `common` that is used in only one place, but I can agree that there could be other uses for it in the future. Of course quisquous will have to make the final determination there.

I am a little confused by the necessity for the added regex on zone seal/unseal. What are you trying to do with `$1 dans`? And is there a situation where `ouverture` does not  correctly match the zone unsealing?

ui/raidboss/common_replacement.js
```diff
@@ -11,7 +11,7 @@ let commonReplacement = {
   },
   'is no longer sealed': {
     de: 'öffnet sich erneut',
-    fr: 'Ouverture ',
+    fr: 'Ouverture d(?:e|u|es) $1 dans',
```
quisquous, line 14: I'm not sure I understand why? I think this was fine before?

ui/raidboss/common_replacement.js
```diff
@@ -91,6 +91,13 @@ let commonReplacement = {
     cn: '--西北--',
     ko: '--북서--',
   },
+  '--rotate--': {
+    de: '--rotate--', // FIXME
```
quisquous, line 95: Sorry, but can you remove the FIXME lines? This makes them not show up in `find_missing_translations.py`.

VanSlanzar:
I hope we understand each other.:
- French punctuation requires spaces, removing a word can completely change the meaning of the sentence and not appear correct in terms of syntax or semantics.
- On Baldesion:
 beginning of the sentence

 name of the area

 complete sentence

if we remove (de, du, dans) it no longer makes sense.

ui/raidboss/common_replacement.js
VanSlanzar: I answered below

ui/raidboss/common_replacement.js
VanSlanzar: I asked several people I am waiting for their answers to correct this

quisquous:
Are all of the Baldesion Arsenal zones `Ouverture de` and `Fermeture de` or are they wrong and need to just be `Ouverture` like the text in `common_replacement.js`?

I don't think this means that you need to add `de` or `d(?:e|u|es)` to `common_replacement.js`.  The only reason to add more text is if there is other use of `Ouverture` in game text that is not a zone being unsealed.  (This is what @JLGarber was asking.) It doesn't sound like there is though?

ui/raidboss/common_replacement.js
Akurosia: `--rotate--` would be `--rotieren--` in german

VanSlanzar:
us understanding is not easy ^^. 
all instances in French, when the area opens, say (ouverture de, ouverture du).
if it is not useful for (ouverture) why it is for (fermeture).
forgive me i'm just trying to figure out if necessary i will delete it.

quisquous:
> if it is not useful for (ouverture) why it is for (fermeture).
> forgive me i'm just trying to figure out if is necessary i will delete it.

`common_replacement.js` is about automating timeline and sync translations.

This is what the "unsealing" message looks like in timelines:
```
0.0 "--Reset--" sync / 00:0839:.*is no longer sealed/ window 10000 jump 0
```

This is what the "will be sealed" message looks like in timelines:
```
0.0 "--sync--" sync / 00:0839:The Alpha Concourse will be sealed off/ window 5,5
4000.0 "--sync--" sync /00:0839:The Realm of the Thunder God will be sealed off/ window 4000,0
2000.0 "--sync--" sync / 00:0839:Doxa will be sealed off/ window 2000,0
```

To translate the "unsealed" message, you just need to replace "is no longer sealed" with some phrase that appears in the "unsealing" phrase, like `Ouverture`.

The "will be sealed" translations in `common_replacement.js` replace everything but the zone name, so that the translations in each file itself can translate just the zone name.  For example, `'Doxa': 'Platz der Doxa',`  It's the "common" part of the "will be sealed" message, so that this doesn't have to be repeated in every timeline file.

I hope that helps?

ui/raidboss/common_replacement.js
VanSlanzar: thank you @Akurosia

VanSlanzar:
I understand what you mean, so it should be:
- ( fermeture) only and (ouverture) only. And it will work even if there are words missing.

quisquous:
> * ( fermeture) only and (ouverture) only. And it will work even if there are words missing.

Hmm.  I do not know what you mean by "only" here.  To restate, it should be `Fermeture` with the the zone name replacement and the $1 etc, and it should be `Ouverture` on its own, as a partial line.  In other words, it should match what the other languages are doing.

VanSlanzar:
Is this correct? If it is not the case I just restore (opening), and I do not touch closing. Sorry for the inconvenience.

- Fermeture d(?:e|u|es) $1 dans 15 secondes
-  Ouverture d(?:e|u|es)

ui/raidboss/common_replacement.js
```diff
@@ -11,7 +11,7 @@ let commonReplacement = {
   },
   'is no longer sealed': {
     de: 'öffnet sich erneut',
-    fr: 'Ouverture ',
+    fr: 'Ouverture d(?:e|u|es) $1 dans',
     ja: 'の封鎖が解かれた',
     ko: '의 봉쇄가 해제되었습니다',
```
VanSlanzar, line 16: Chinese translation is missing here, no ?

VanSlanzar, line 16: Chiness translation is missing here, no ?

quisquous:
> * If it is not the case I will restore (opening), and I do not touch (fermeture). Sorry for the inconvenience.

Unless they are not working, I think they are both fine as-is and there is not a need to change these.

VanSlanzar:
Fix  it, sorry for the inconvenient.

ui/raidboss/common_replacement.js
maiko-tan-coding: the ja translation is correct, but cn is not, it should be `--回转--`

quisquous:
Thanks! Sorry for the confusion!

ui/raidboss/common_replacement.js
maiko-tan-coding: cn it should be `的封锁解除了……`

maiko-tan-coding:
I think i have understood what quisquous said. Let me try to explain it to you.

You can see, `<place> is no longer sealed` is the "unsealing" message. But we only care about `no longer sealed`. What place? It does not matter.

But for the "sealed off" message, we try to sync them **with** place name. So it should goes `$1 will be sealed off`, while `$1` is the place name we try to add into this regex.

VanSlanzar:
You are absolutely understood. I was just trying to put the French prepositions (de, du, des) in the translation. Thank you for these explanations.

PR #1422, opened 2020-05-10
translation fixes for HW

Akurosia:
HW translation fixes, removed FIXME and added missingTranslations if necassary
this excludes changes for a1s and a11s as they are currently throwing errors with the changes made


quisquous:
Thanks for this.  I'll follow up about the a111s issues

PR #1428, opened 2020-05-12
Translation tools update

Akurosia:
Fixed an issue in `find_missing_timeline_translation` where 1b: and 23: would not match the regex in timeline.js as well as where only the first matching occurence of a match would be replaced to the corresponding language

Added the ability to `check_translation.py` where douplicated lines and not used translations are printed.


ui/raidboss/timeline.js
```diff
@@ -78,10 +78,8 @@ class Timeline {
       '--Reset--',
       '--sync--',
       'Start',
-      '^ ?21:',
-      '^(\\^\\.\\{14\\})? ?1B:',
-      '^(\\^\\.\\{14\\})? ?21:',
-      '^(\\^\\.\\{14\\})? ?23:',
+      '^(\\(\\?\\<timestamp\\>\\^\\.\\{14\\}\\)) (1B|21|23):',
```
quisquous, line 81: Oof.  Thank you.  I would be happy to land this bit on its own, since it's kind of a bugfix vs the feature that the rest of the pull request is.

util/check_translation.py
```diff
@@ -256,6 +293,58 @@ def filter_print_timeline(text, missing_filter):
...
+                else:
+                    remove.append(old)
+                    print(f'Found duplicate entry for "{old}" in "{category}"!')
+            for entry in remove:
+                del trans[locale][category][entry]
+        print()
+    return trans
+
+
+def print_not_used_translations(not_used_trans):
```
quisquous, line 313: I think it would be better to do this in a test rather than in check_translation.py, where you are already in javascript, and we could (potentially?) throw an error if a translation isn't used.

I feel like a little bit that the tests and find_missing_translations script have superseded check_translation.py.  I have been wondering a little if I should remove it.

What do you think about all of this?

Akurosia: I like the output of check_translations a little bit more.
i usually used the check_translation part to do all my translation checks and only controlled them in the end by the find_missing_timeline_translation script.
I think mostly because i am not that familiar with adding my crude hacks as real tests.

ui/raidboss/timeline.js
Akurosia: ok i will make seperate PRs than :)

Akurosia:
ok i made now a seperate PR #1429 for the find_missing_translation part (timeline.js included)
I would leave this PR still open to discuss the check_translation part

PR #1429, opened 2020-05-13
fix find_missing_timeline_translation

Akurosia:
as discussed i made a seperate PR for just this part of the translation updates


util/find_missing_timeline_translations.js
```diff
@@ -88,7 +88,7 @@ function findMissingRegex() {
         let replace = Regexes.parse(regex);
         if (transRegex.match(replace))
           foundMatch = true;
-        transRegex = transRegex.replace(replace, set[regex]);
+        transRegex = transRegex.split(replace).join(set[regex]);
```
quisquous, line 91: Can you explain to me what this part is doing, sorry?

Akurosia: This is an equivalent of .replaceAll() in other languages

The old version only replaced the first occurence and not all (e.g. shinryus wings were not correctly handeled)

PR #1430, opened 2020-05-13
Fix job and level capture missing from Regexes.addedCombatantFull

valarnin:
Missed `job` and `level` capturing groups for Regexes.addedCombatantFull.

I'm still working through converting the regexes for #1387 to use `Regexes` functions.

### DIFF
--- a/resources/regexes.js
+++ b/resources/regexes.js
@@ -71,6 +71,21 @@ var Regexes = {
       'targetId',
       'target',
       'flags',
+      'flag0',
+      'flag1',
+      'flag2',
+      'flag3',
+      'flag4',
+      'flag5',
+      'flag6',
+      'flag7',
+      'flag8',
+      'flag9',
+      'flag10',
+      'flag11',
+      'flag12',
+      'flag13',
+      'flag14',
       'targetHp',
       'targetMaxHp',
       'targetMp',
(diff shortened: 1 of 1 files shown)


valarnin:
Another thing I'm running into issues with is this:
```javascript
      AbilityCode: '[0-9A-Fa-f]{1,4}',
```

This regex doesn't match item uses, e.g. this line:
```
[23:00:59.384] 15:12345678:Some One:20FB57E:Item_Fb57e:12345678:Some One:E800010E:300000:0:0:0:0:0:0:0:0:0:0:0:0:0:0:117462:117462:10000:10000:0:1000:99.71765:111.1925:-5.960464E-08:-2.182744:117462:117462:10000:10000:0:1000:99.71765:111.1925:-5.960464E-08:-2.182744:0000501C
```

quisquous:
Re: ability code.  Maybe change all the captures to be `.*?` in `regexes.js`?  I think I did that for almost everything in `netregexes.js`.  I tried to make `\y{Name}` be just `.*?` in general rather than the monster that it is, but it broke a few things.

Alternatively, make it more complicated to support Item_etc as well?

What do you think?

valarnin:
It could just be `[0-9a-zA-Z]{1,8}` instead. I'm just not sure if it was done with 1-4 for a reason.

quisquous:
> It could just be `[0-9a-zA-Z]{1,8}` instead. I'm just not sure if it was done with 1-4 for a reason.

Yeah, that's also reasonable! Probably the most minimal change.

valarnin:
> I tried to make \y{Name} be just .*? in general rather than the monster that it is, but it broke a few things.

That should work without issues as long as the regexes using it do a full line match. Do you have specific examples of where it broke?

valarnin:
I tracked this back to https://github.com/quisquous/cactbot/commit/ea1cff1a798d59ce228354ed248145b70f7f6d86 which seems like it should be fine to un-revert now, since everything's using Regexes that references `\y{Name}` as far as I can see, with the exception of https://github.com/quisquous/cactbot/blob/master/ui/raidboss/raidemulator.js#L62 which shouldn't have any issues with `.*?`. I don't think it's in scope for this PR, but might be something you want to look at or it's something I can submit a separate PR for.

Either way, I've amended the commit with the 1-8 change, so this should be good to merge now.

valarnin:
Alright. Hopefully now I've actually found all the missing capture groups and such. I've passed these through about 6.1 million log lines now, for content from 2.0-latest, normal dungeons, trials, and raids.

Only thing I couldn't really test was open world content since I don't usually bother running ACT for that.

If I find anything else missing I'll just open a new PR for it.

resources/regexes.js
```diff
@@ -71,6 +71,21 @@ var Regexes = {
       'targetId',
       'target',
       'flags',
+      'flag0',
+      'flag1',
+      'flag2',
+      'flag3',
+      'flag4',
+      'flag5',
+      'flag6',
```
quisquous, line 80: Are these flags useful to capture, vs just a `(?:[^:]*:){15}`?

valarnin: Not sure how many of the flags would actually be useful to capture, or even what to call them. Based on <https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#15-networkability> some of them might be useful as capture targets at least, but it would be on a per-trigger basis.

quisquous: I guess? There just haven't been any examples yet where this was something useful to have.  That's why I asked if you had a use in mind here.

PR #1431, opened 2020-05-13
Util: Basic support for 34 network log lines in make_timeline

JLGarber:
This permits the user to generate timelines with precisely-timed  `targetable/untargetable` entries baked in. It's rather a pain to hunt through the log and manually add these, so based on an idle suggestion from quisquous the other day I went and added it.

I played with it a little, but I don't like working with `argparse` much, so for now `include_targetable` is only an explicit allow list. (Passing as a bare flag has no effect, and honestly unless Square comes up with something *really* out there, I don't think it's necessary to just display all nameplate changes.) Suggestions welcome, since I mostly slapped these changes together over lunch earlier today.

Sample output from Sophia's intermission:

```
102.4 "Cloudy Heavens" sync /:Sophia:19BE:/
103.6 "--untargetable--"
112.7 "Vertical Kenoma" sync /:The First Demiurge:19BB:/
113.7 "Gnostic Spear" sync /:The Third Demiurge:19B9:/
117.7 "Divine Spark" sync /:The Second Demiurge:19B6:/
120.9 "Ring Of Pain" sync /:The Third Demiurge:19BA:/
126.9 "Vertical Kenoma" sync /:The First Demiurge:19BB:/
137.7 "Gnostic Rant" sync /:The Third Demiurge:19B8:/
141.1 "Horizontal Kenoma" sync /:The First Demiurge:19BC:/
145.2 "Gnostic Spear" sync /:The Third Demiurge:19B9:/
168.7 "Infusion" sync /:The First Demiurge:1988:/
182.1 "The Scales Of Wisdom" sync /:Sophia:1981:/
190.1 "--sync--" sync /:Sophia:1A4C:/
190.1 "--sync--" sync /:Sophia:1ABE:/
195.1 "--sync--" sync /:Sophia:1A4C:/
195.1 "--sync--" sync /:Sophia:1ABE:/
199.5 "The Scales Of Wisdom" sync /:Sophia:1AE1:/
205.0 "--targetable--"
212.8 "Quasar" sync /:Sophia:196E:/
```

Currently I've left these unsynced just like we do in modern use, but I'm happy to add them as a sync. (This would also permit us to add them to `test_timeline` at some point in the future.)

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -112,7 +112,7 @@ def parse_file(args):
                 continue
             line_fields = line.split("|")
             # We aren't including targetable lines unless the user explicitly says to.
-            if line[0:2] == '34' and not line_fields[3] in args.includetargetable:
+            if line[0:2] == '34' and not line_fields[3] in args.include_targetable:
               continue
 
             # At this point, we have a combat line for the timeline.
@@ -219,7 +219,7 @@ def main(args):
           # Ignore aoe spam
           if entry["time"] == last_entry["time"] and entry["ability_id"] == last_entry["ability_id"]:
               continue
-        elif entry["line_type"] == "34" and not args.includetargetable:
+        elif entry["line_type"] == "34" and not args.include_targetable:
           continue
 
         # Find out how long it's been since our last ability
@@ -265,7 +265,7 @@ def main(args):
         entry["position"] = timeline_position
 
         # Write the line
-        if entry["line_type"] == "34" and args.includetargetable:
+        if entry["line_type"] == "34" and args.include_targetable:
           print(entry)
           output_entry = '{position:.1f} "{targetable}"'.format(
               **entry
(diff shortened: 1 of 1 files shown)


util/make_timeline.py
```diff
@@ -243,7 +254,7 @@ def main(args):
             pass
 
         # Set the time, possibly adjusting to specified phase
-        if entry["ability_id"] not in phases:
+        if entry["line_type"] == "34" or entry["ability_id"] not in phases:
```
quisquous, line 257: Probably this should be a check for not 21,22 rather than is 34, for if/when we add 03 lines.

JLGarber:
Much appreciated! I agree, I'm not 100% happy with how it's set up either, but I think it's acceptable. We'll (at most) be adding support for two more log lines, and `03/04` would be handled pretty much identically.

After this I might make make another pass through and try to extract more stuff to `encounter_tools` from `make/test`. I think that might make it slightly easier to make changes in future? (I also need to split some stuff in `encounter_tools` that I mistakenly combined a while ago anyway.)

quisquous:
Thank you so much for this.  Even looking at a7s timelines right now, this is just such a relief to have these lines automatically added.

PR #1439, opened 2020-05-17
removed all fixme lines for SB and ShB content

Akurosia:


quisquous:
Thank you, this was way easier to parse when it's just that on its own.

PR #1440, opened 2020-05-17
fix: regex escape for unending coil ultimate

oott123:
this should fix chinese and japanese dialog matches


quisquous:
Ah, thank you! I missed that when we switched to `Regexes.dialog`.

PR #1427, opened 2020-05-12
translation fixes for SB

Akurosia:
same as for ARR and HW, here a fix of all SB translations.

PR #1441, opened 2020-05-17
sort all sb content alphabeticly

Akurosia:


VanSlanzar:
I noticed that the ARR files are not all sorted. Are you going to do it or do you want me to do it? Before I sorted manually, now I use the sort function of (sublime text).

Akurosia:
first i will finish my open tass for sb/shb (more relevant for me right now).
After this i can have a look again at ARR / HW

VanSlanzar:
With pleasure, I let you do it at your own pace. Thank you.

quisquous:
After looking at this, I'm not really convinced that sorting is what I want. I think my feelings come from:
* sorting regexes is awkward, and so it doesn't make it easier to find things, as `(?! )Zzzzzz` is at the beginning and not the end
* Nael quotes get split all over the place
* there's no testing for this, so it's not going to stick, but that said, I'm not sure that I'm convinced that I want to enforce this with testing either

Can I ask what you get out of this? Do you (or others) find these files easier to read with these changes?

Akurosia:
As i do all my translations via scripts with minimal manuall editing in the end it will always be sorted alphabeticly. The seperation of the PRs was discussed as par of: #1427

quisquous:
> As i do all my translations via scripts with minimal manuall editing in the end it will always be sorted alphabeticly. The seperation of the PRs was discussed as par of: #1427

Is it a requirement that your scripts have to sort them?

I know we discussed separating this out earlier, but when I see this on its own, I'm a bit hesistant based on the reasons I stated above.

Akurosia:
> > As i do all my translations via scripts with minimal manuall editing in the end it will always be sorted alphabeticly. The seperation of the PRs was discussed as par of: #1427
> 
> Is it a requirement that your scripts have to sort them?
> 
> I know we discussed separating this out earlier, but when I see this on its own, I'm a bit hesistant based on the reasons I stated above.

As it is handeled by a script, the script will either save it in a random order every time i ran it or i can specify the sort=true flag and i rather have a sorted list then a completly random one

VanSlanzar:
I personally use the sort function of the sublime text software. Everything that starts with symbols is listed at the top = at the top of the list. If you want to take care of other things, I can sort the files.

quisquous:
> As it is handeled by a script, the script will either save it in a random order every time i ran it or i can specify the sort=true flag and i rather have a sorted list then a completly random one

I guess I'm asking if those are the only options? Can your script keep them in the same order? What if there are comments on separate lines?

Akurosia:
comments needs to be removed before using the script as the demjson library will throw an error otherwise.
from a fes test i did, i cant seam to find a way to keep a valid ordering as demjson will always order alphabeticly when reading elements.

an easy exmaple is followign code 
```
import demjson
var = {'replaceSync': {'Guardian of Paradise': 'Hüter von Eden','Eden Prime': 'Prim-Eden',},}
print(demjson.encode(var)) 
```
which will result in a printed output of `{"replaceSync":{"Eden Prime":"Prim-Eden","Guardian of Paradise":"H\u00fcter von Eden"}}`.

When i try loading the js-json with the standard json lib, i receive following erro:
```
>>> print(json.loads(var))
  File "<stdin>", line 1, in <module>
  File "C:\Program Files (x86)\Python37-32\lib\json\__init__.py", line 341, in loads
    raise TypeError(f'the JSON object must be str, bytes or bytearray, '
TypeError: the JSON object must be str, bytes or bytearray, not dict
```

quisquous:
What about https://github.com/dmeranda/demjson/issues/14#issuecomment-64242490

Akurosia:
ok after a couple of tries with how my translation setup worked, i am only limited to 1 of those options:
1. only translate and modify the german section to avoid sorting entries etc
2. modify all sections with known skillnames (which would reorder at least all non german elements)

i still prefere sorted translations tbh but i will adjust if you thinks its not needed, but than i could not fix the non german languages for old content (as it would just take to much time as i need +4x the effort)

quisquous:
Did `SORT_PRESERVE` not work?

Akurosia:
the problem is, that with my translation setup all non DE languages are "newly created" elements on that dict and they will always be sorted alphabeticly.
its just to much of a hassle to to a full rework of the translation script just to keep the order for me

i also got confused of some different parts on my script as i just needed the demjson for the pure sorting of the existing js translation data...

quisquous:
Newly created ones being sorted is fine! It's more, I don't want every time you add a translation to have to also re-sort the file.

If somebody else rearranges translations or groups them or puts them in a different order, then your script would undo that.  If there's nothing there, then sorted is a reasonable order to put them in by default.

Akurosia:
i usually just do the script usage of this if many changes are there (like in this case where i want to fix ALL of the reported errors by the find_missing script) usually for small translations i just look the translations up and put them in

quisquous:
So, I just wanted to check back here.  I'm sorry for the push back, I think it's just that I don't know that I want sorting if there's a choice, and sorting existing things when touching them also makes it harder to read.

In the end of things, your contributions are extremely appreciated, and I would hate to lose them or especially to add 4x the amount of work that you'd have to do.  Sorting is definitely not even close to that important to me.

I think with all of this, I just wanted push back a tiny bit and ask "hey, how hard would it be to do this thing" and understand what that looks like.  If you want me to take a look at your script, I can do that too, if it'd be helpful.

PR #1444, opened 2020-05-17
Correction Fr ARR

VanSlanzar:
Change of incorrect translations, correction of capital letters. Adding a missing translation.


ui/raidboss/data/02-arr/raid/t9.js
```diff
@@ -485,7 +485,7 @@
       'replaceSync': {
         'Astral Debris': 'Débris astral',
         'Dalamud Fragment': 'Débris de Dalamud',
-        'Dalamud Spawn': 'golem de Dalamud',
+        'Dalamud Spawn': 'Golem de Dalamud',
```
quisquous, line 488: Just to mention it again to save you work, the capitalization in `replaceSync` does not matter.  This will never be shown to a user, and these are only used in a case-insensitive way.

VanSlanzar: I know it's not useful, it's just that I want to respect capital letters personally. It doesn't give me more work, there are very few cases like that. Sorry to be maniac. Thank you for your work.

PR #1445, opened 2020-05-18
Raidboss: Shinryu normal package

JLGarber:
Not too much to be said about this one. I threw it together on a whim to test the effectiveness of the changes to `make_timeline` as far as `targetable/untargetable`.

Oopsy is missing a few things though. Because I'm lazy, we don't currently have triggers set up to check whether users were correctly in/out of puddles based on mechanics combined with debuffs. Once Somebody:tm: expands `oopsyraidsy.js` to be compatible with `netRegexes` this whole thing will be a lot easier. (We'll no longer have to modify each and every one of the language files, because of effect IDS.)

As usual with a number of these encounters, the triggers are largely copied over from Extreme.


ui/raidboss/data/04-sb/trial/shinryu.js
```diff
@@ -0,0 +1,549 @@
...
+      'missingTranslations': true,
+      'replaceSync': {
+        'Cocoon': '빛의 고치',
+        'Left Wing': '왼쪽 날개',
+        'Right Wing': '오른쪽 날개',
+        'Shinryu': '신룡',
+        'Tail(?! )': '신룡의 꼬리',
+      },
+      'replaceText': {
+        '--adds spawn--': '--adds spawn--',
```
Jaehyuk-Lee, line 507: `'--adds spawn--': '--쫄 소환--',`

ui/raidboss/data/04-sb/trial/shinryu.js
```diff
@@ -0,0 +1,549 @@
...
+        'Aerial Blast': '대기 폭발',
+        'Akh Morn': '아크 몬',
+        'Akh Rhai': '아크 라이',
+        'Dark Matter': '암흑물질',
+        'Diamond Dust': '다이아몬드 더스트',
+        'Dragonfist': '용의 손바닥',
+        'Dragonflight': '용의 비행',
+        'Elemental Attack': 'Elemental Attack',
```
Jaehyuk-Lee, line 515: `'Elemental Attack': '원소 공격',`

ui/raidboss/data/04-sb/trial/shinryu.js
```diff
@@ -0,0 +1,549 @@
...
+        'Lightning Resistance Down II': '雷属性耐性大幅降低',
+        'Magic Vulnerability Up': '魔法受伤加重',
+        'Physical Vulnerability Up': '物理受伤加重',
+        'Stun': '眩晕',
+        'Thin Ice': '冰面',
+      },
+    },
+    {
+      'locale': 'ko',
+      'missingTranslations': true,
```
Jaehyuk-Lee, line 498: remove after my suggestion is applied.

ui/raidboss/data/04-sb/trial/shinryu.js
```diff
@@ -0,0 +1,549 @@
...
+    },
+    {
+      // There doesn't really seem to be any verified information about this marker.
+      // It usually appears around Burning Chains, but it's wildly inconsistent.
+      // However, it *seems* that the tail attacks from the side the target is on.
+      id: 'Shinryu Normal Slap Marker',
+      netRegex: NetRegexes.headMarker({ id: '0062' }),
+      condition: Conditions.targetIsYou(),
+      infoText: {
+        en: 'Bait onto unbroken squares',
```
Jaehyuk-Lee, line 284: `ko: '꼬리 내리치기 유도',`

JLGarber:
Thanks so much for the additions! At some point, I think we will move `--adds spawn--` to one of the common translation files, but I will leave it here for now, at least until after this pull request is merged.

ui/raidboss/data/04-sb/trial/shinryu.js
JLGarber: Thanks! Do note, it is possible we may change the English on this one, depending on what other reviewers have to say. For reference, this mechanic is one of six possible attacks:
```
Aerial Blast
Judgment Bolt
Diamond Dust
Hellfire
Earthen Fury
Tidal Wave
```
We can't list all of them in the timeline, so I shortened it here. (You can see that because the CN translations for these attacks are all four characters long, it was easy to just put all six into that replacement.)

Jaehyuk-Lee: I saw that in timeline file, and I think this translation is okay with it, better than just listing all the skill names.

ui/oopsyraidsy/data/04-sb/trial/shinryu.js
```diff
@@ -0,0 +1,58 @@
...
+      mistake: function(e) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
+      },
+    },
+    {
+      // Icy floor attack.
+      id: 'Shinryu Normal Diamond Dust',
+      gainsEffectRegex: gLang.kEffect.ThinIce,
+      deathReason: function(e) {
+        return { type: 'fail', name: e.targetName, reason: { en: 'Slid off!' } };
```
quisquous, line 38: We really need to find some network data for falling to your death.  Some day~

ui/raidboss/data/04-sb/trial/shinryu.js
```diff
@@ -0,0 +1,549 @@
...
+      alertText: {
+        en: 'out of water',
+        de: 'Raus aus dem Wasser',
+        cn: '离开水圈',
+        ko: '물 장판 밖으로',
+      },
+    },
+    {
+      id: 'Shinryu Normal Levinbolt',
+      netRegex: NetRegexes.startsUsing({ id: '1F9B', source: 'Right Wing', target: 'Right Wing', capture: false }),
```
quisquous, line 149: What does the target specifier get you here? Mostly curious because nothing else uses this.

ui/raidboss/data/04-sb/trial/shinryu.txt
```diff
@@ -0,0 +1,84 @@
...
+59.2 "Summon Icicle" sync /:Left Wing:1F9F:/
+59.9 "Icicle Impact x2" # sync /:Icicle:1FA0:/
+64.4 "Spikesicle x2" # sync /:Icicle:1FA1:/
+75.2 "Ice Storm" sync /:Left Wing:1FA2:/
+88.3 "Earth Breath" sync /:Shinryu:1F9D:/ # May be skipped with high DPS
+115.8 "Hellfire/Judgment Bolt"sync /:Shinryu:1FA[BC]:/ window 30,30
+118.1 "--untargetable--"
+
+# Intermission.
+# The given times for Meteor Impact are the latest they can occur.
```
quisquous, line 25: Are there start casting times for those that you could sync to instead?

ui/oopsyraidsy/data/04-sb/trial/shinryu.js
JLGarber: Someday.

ui/raidboss/data/04-sb/trial/shinryu.js
JLGarber: Nothing, that's a copypasta leftover from Extreme. No idea why it was used there, honestly.

ui/raidboss/data/04-sb/trial/shinryu.txt
JLGarber: It looks like it? I'm having trouble getting that to sync, just trying a couple more things here...

JLGarber: That seems to work. I wish there were a good way to include those by default at timeline generation, but it's not a simple thing to filter like `targetable/untargetable`.

quisquous: Yeah, I think this is harder.  I mean, one option would be to have a timeline option where you could pass a flag and it would always print them, and then you could use that and splice them in? It doesn't get needed all that many times, I think, so it's not a huge missing feature either.

PR #1455, opened 2020-05-21
Raidboss: fix error when AlertText is number

### LINKED ISSUE #1268: UWU Titan Gaols issues
As title says, the callout for the titan gaols works only when you are not one of the gaols, otherwise it just calls nothing :(
I cleared uwu months ago, and back then the gaol callout worked perfectly, but right now it works only when I'm not one of the gaols.
Thanks for the patience!

elsorume:
Fixes #1268

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -5,6 +5,8 @@
 function triggerUpperCase(str) {
   if (!str)
     return str;
+  if (typeof(str) === 'number')
+    return str;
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks, it looks like you're failing a lint error.  See https://github.com/quisquous/cactbot/blob/master/CONTRIBUTING.md#build-problems for more details.  Would you mind fixing it and adding another commit?

I suspect it's because it should be `typeof str` and not `typeof(str)` since `typeof` is an operator and not a function.

quisquous:
Thanks!!

Issue #1425, opened 2020-05-11
Missing timelineReplace for regex


VanSlanzar:
After good work @Akurosia on Arr and Heaven, I noticed that this was missing. I'm not talking about missing txt, but (missing timelineReplace for regex). Is this a bug, or like on my last request for missing translations in effect names? Or it's normal, if it's case my apologies. It's only for arr, i have not looked for HW. 
arr zip for all language.
[Arr missing translations.zip](https://github.com/quisquous/cactbot/files/4608902/Arr.missing.translations.zip)

Akurosia:
This is in my eyes not a bug and more of an enhancement.
All languages where translations are missing should be flaged with `'missingTranslations': true,` and needs to be translated by members of the community who speak the corresponding language.

Most of all skills, effects and enemy names are translated and usually only the replacementText part is missing as this can have non game related text (nothing you could easly look up in the game files)

Generally it would be good to have someone who could maintain a givin language like i do for the DE part or @mooondark for most of the FR part

VanSlanzar:
I was speaking exclusively for this : missing incorrect timelineReplace For regex. #1419 

$ python util/find_missing_translations.py -l fr -f arr
.\ui\raidboss\data\02-arr\raid\t10.js:31 T10 Wild Charge: missing timelineReplace for regex '(?<timestamp>^.{14}) 1b:(?<targetid>[0-9a-f]{8}):(?<target>[^:]*?):....:....:(?<id>001f):'
.\ui\raidboss\data\02-arr\raid\t10.js:55 T10 Prey: missing timelineReplace for regex '(?<timestamp>^.{14}) 1b:(?<targetid>[0-9a-f]{8}):(?<target>[^:]*?):....:....:(?<id>001e):'
.\ui\raidboss\data\02-arr\raid\t12.js:89 T12 Whitefire: missing timelineReplace for regex '(?<timestamp>^.{14}) 1b:(?<targetid>[0-9a-f]{8}):(?<target>[^:]*?):....:....:(?<id>0020):'
.\ui\raidboss\data\02-arr\raid\t12.js:102 T12 Bluefire: missing timelineReplace for regex '(?<timestamp>^.{14}) 1b:(?<targetid>[0-9a-f]{8}):(?<target>[^:]*?):....:....:(?<id>0021):'
[...]

PR #1458, opened 2020-05-22
Raidboss: t7 Change regex to netRegex 

VanSlanzar:
Change regex to netRegex for cursed and shriek voice.

### DIFF
--- a/ui/raidboss/data/02-arr/raid/t7.js
+++ b/ui/raidboss/data/02-arr/raid/t7.js
@@ -82,12 +82,12 @@
     },
     {
       id: 'T7 Voice',
-      netRegex: NetRegexes.gainsEffect({ effectId: '1C3', source: 'Melusine' }),
-      netRegexDe: NetRegexes.gainsEffect({ effectId: '1C3', source: 'Melusine' }),
-      netRegexFr: NetRegexes.gainsEffect({ effectId: '1C3', source: 'Mélusine' }),
-      netRegexJa: NetRegexes.gainsEffect({ effectId: '1C3', source: 'メリュジーヌ' }),
-      netRegexCn: NetRegexes.gainsEffect({ effectId: '1C3', source: '美瑠姬奴' }),
-      netRegexKo: NetRegexes.gainsEffect({ effectId: '1C3', source: '멜뤼진' }),
+      netRegex: NetRegexes.gainsEffect({ effectId: '1C3', }),
+      netRegexDe: NetRegexes.gainsEffect({ effectId: '1C3', }),
+      netRegexFr: NetRegexes.gainsEffect({ effectId: '1C3', }),
+      netRegexJa: NetRegexes.gainsEffect({ effectId: '1C3', }),
+      netRegexCn: NetRegexes.gainsEffect({ effectId: '1C3', }),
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/02-arr/raid/t7.js
```diff
@@ -82,12 +82,12 @@
     },
     {
       id: 'T7 Voice',
-      regex: Regexes.gainsEffect({ effect: 'Cursed Voice' }),
-      regexDe: Regexes.gainsEffect({ effect: 'Stimme Der Verwünschung' }),
-      regexFr: Regexes.gainsEffect({ effect: 'Voix Du Maléfice' }),
-      regexJa: Regexes.gainsEffect({ effect: '呪詛の声' }),
-      regexCn: Regexes.gainsEffect({ effect: '诅咒之声' }),
-      regexKo: Regexes.gainsEffect({ effect: '저주의 목소리' }),
+      netRegex: NetRegexes.gainsEffect({ effectId: '1C3', source: 'Melusine' }),
```
quisquous, line 85: Why are you adding `source` here? From your log line pictures that you linked to me, it doesn't look like there's a source.

VanSlanzar: Ah, https://github.com/quisquous/cactbot/blob/master/docs/LogGuide.md#1a-networkbuff
`1A:[ObjectId]:[Target Name] gains the effect of [Status] from [Source Name] for [Float_Value] Seconds`.

I believed in an oversight on my part, if it is useless, I correct that.

quisquous: It's not that it's useless--it's that it's incorrect.  Looking at the log lines you sent me, there is no `source` from those gains effect lines.  It is empty.

VanSlanzar: Yes, i understand. It's corrected.

ui/raidboss/data/02-arr/raid/t7.js
```diff
@@ -82,12 +82,12 @@
     },
     {
       id: 'T7 Voice',
```
panicstevenson, line 84: Can this be T7 Cursed Voice? I know it's kind of pedantic, but we lose some context by moving to the effectId

quisquous: And `T7 Cursed Shriek` elsewhere too!

VanSlanzar: This was done to resolve this issue. #1425 
As I did not understand the meaning of @panicstevenson  what should I do? @quisquous

JLGarber: What they are saying is that "voice" as a trigger ID in English does not really explain what is going on with this trigger. Instead, it should be:

```
 {
      id: 'T7 Cursed Voice',
      // trigger body etc
  }
```

This way, we know what the mechanic is without having to look for clues or looking it up on XIVAPI.

(This is not your fault, the trigger was named this a very long time ago when we didn't have as much standardization for things. However, now that we are using bare effect IDs, we do need to have the name for it *somewhere* in the trigger.)

VanSlanzar:
Is this better? thanks to @JLGarber for his explanations, I understand better.

quisquous:
Thanks, perfect!

PR #1462, opened 2020-05-23
Docs: Added recommendation for contextual trigger names in Raidboss

### LINKED ISSUE #1460: Raidboss: Netregexes and names--Ability/Effect names should be indicated somewhere
panicstevenson [noted](https://github.com/quisquous/cactbot/pull/1458#discussion_r429465012) that we can lose context more easily when we are using bare ability/effect IDs. That is, we have no way of knowing what these lines are, really:

`regex: NetRegexes.ability({ id: '156F', source: 'Sephirot', capture: false }),`
`regex: NetRegexes.gainsEffect({ id: ['3ED', '3EE'], source: 'Sephirot' }),`

The first is Ein Sof, the green puddles, and the second is Force Against Magic/Might, but without context, such as descriptive trigger names, we don't know that.

Currently, the `raidboss` doc has this to say about the trigger ID:

> An id string for the trigger, used to disable triggers. Every built-in trigger that has a text/sound output should have an id so it can be disabled. User-defined triggers need not have one.

I would propose adding to it so that it reads something like this:

> An id string for the trigger, used to disable triggers. Every built-in trigger that has a text/sound output should have an id so it can be disabled. User-defined triggers need not have one.

> Because the current structure for `Regexes/NetRegexes` no longer requires that the ability/effect/whatever name be present as part of the expression, it is extremely important that that information is somewhere alongside. Recommended practice is either to have the effect/ability/NPC name in the trigger ID itself, or in an explanatory comment alongside. Context solely from the trigger body is not necessarily sufficient!

I haven't seen this as being a big problem, necessarily, since most of us are already careful to give triggers descriptive IDs. However, I feel like it should be In The Documentation Somewhere. We lose nothing by making descriptive IDs/comments a standard, and we potentially gain from it.

JLGarber:
(This will resolve #1460.)

Per panicstevenson's [comment](https://github.com/quisquous/cactbot/pull/1458#discussion_r429465012) and #1460, I've added a short paragraph indicating that context is now recommended for triggers included in the cactbot repository.

I added this alongside the id field information, since that's usually where *I* put it, but if we think it should be moved elsewhere, we could put it after the description of the regex fields.

Wording advice/changes welcome!

(I'll probably do something similar for Oopsy, but given that only 2-3 people seem to be actively working with those files right now, it's less of a priority.)

### DIFF
--- a/docs/RaidbossGuide.md
+++ b/docs/RaidbossGuide.md
@@ -89,7 +89,14 @@ Boolean, defaults to true. If true, timelines and triggers will reset automatica
 ### Trigger Elements
 
 **id string**
- An id string for the trigger, used to disable triggers. Every built-in trigger that has a text/sound output should have an id so it can be disabled. User-defined triggers need not have one.
+ An id string for the trigger, used to disable triggers. Every built-in trigger that has a text/sound output should have an id so it can be disabled.
+(User-defined triggers not intended for inclusion in the cactbot repository need not have one.)
+
+The current structure for `Regexes/NetRegexes` does not require that the ability/effect/whatever name be present as part of the expression.
+Because of this, it is extremely important that that information is somewhere close by.
+Recommended practice is either to have the effect/ability/NPC name in the trigger ID itself,
+or in an explanatory comment alongside. Context solely from the trigger body is not necessarily sufficient!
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks!

PR #1465, opened 2020-05-24
misc: improve wsUrl capture

DieMoe233:
This pull request changed resources/common.js to improve wsUrl capture.

### DIFF
--- a/resources/common.js
+++ b/resources/common.js
@@ -136,7 +136,7 @@ let Util = {
 };
 
 (function() {
-  let wsUrl = /[\?&]OVERLAY_WS=([^&]+)/.exec(location.href);
+  let wsUrl = /[\?&](OVERLAY_WS|HOST_PORT)=([^&]+)/.exec(location.href);
   let ws = null;
   let queue = [];
   let rseqCounter = 0;
@@ -153,7 +153,7 @@ let Util = {
     };
 
     let connectWs = function() {
-      ws = new WebSocket(wsUrl[1]);
+      ws = new WebSocket(wsUrl[2]);
 
       ws.addEventListener('error', (e) => {
         console.error(e);


quisquous:
Sorry, but I think this is incorrect.  My understanding is that `HOST_PORT` is used for overlays that work with both ACT WebSocket and OverlayPlugin.  `OVERLAY_WS` is only for OverlayPlugin.  As cactbot only works with OverlayPlugin, I think it is correct to only support `OVERLAY_WS`.

The OverlayPlugin websocket tab (ACT -> Plugins -> OverlayPlugin WSServer) knows which is supported and will generate urls with the correct key.

DieMoe233:
> The OverlayPlugin websocket tab (ACT -> Plugins -> OverlayPlugin WSServer) knows which is supported and will generate urls with the correct key.

WS server of ngld's overlayplugin will automatically add Host_Port parameter when using server url on the internet, no matter what overlay it is.
So an improvement should be launch.

quisquous:
> WS server of ngld's overlayplugin will automatically add Host_Port parameter when using server url on the internet, no matter what overlay it is.

I am not sure that I understand this.  When I look at the URL in the `OverlayPlugin WSServer` tab, it has `OVERLAY_WS` for cactbot urls.  I do know that the overlay has to be enabled for this to be correct, although I think @ngld said something about changing this lists to include presets, so that it will always be correct.

MaikoTan:
@DieMoe233 and I are both in a Chinese chat server, sometimes we would talk about ACT things. Based on her given information in that chat server, let me try to explain to you what her thought in English in this PR.

>When I look at the URL in the `OverlayPlugin WSServer` tab, it has `OVERLAY_WS` for cactbot urls.

As DieMoe said, when you try to add any non-loaded overlay (or webpage) in `Overlay WSServer`, it would be added the wrong parameter `HOST_PORT`.
Non-loaded overlay means the overlay which have not loaded in `OverlayPlugin`, and/or have not registered the EventListener (or subscribed the EventSouces) yet.

So, only if you try to add a "loaded" overlay (added in OverlayPlugin as well as registered EventListener), `OverlayPlugin WSServer` would generate the correct parameter `OVERLAY_WS`.

For a English speaker, when he pay attention on ngld's FAQ or document, he would know which the true parameter is. So he would change this parameter to the correct one.
But many of we east Asian don't understand English, when overlay have not displayed expected, they would only come up to search this problem in their own language via search engine.
(Besides, google is not accessible in mainland China, the local search engine Baidu would not show github result on the top. So they Chinese contributors always be disturbed by Chinese users on some stupid problems.)
So, rather than let user has ants in his pants, why don't we directly fix this problem in code?

quisquous:
Hi! Thank you for chiming in.  I am sorry if what I said came off as rude or dismissive, that really wasn't my intention at all.  Maybe there is some solution here to help users not be frustrated or confused, but I am not sure what that is, and I still don't think that it is this pull request.

These are the two options that could print out in `Overlay WSServer`, depending on whether the overlay has been loaded or not:
1) `ui/raidboss/raidboss.html?OVERLAY_WS=wss://127.0.0.1:10501/ws`
2) `ui/raidboss/raidboss.html?HOST_PORT=wss://127.0.0.1:10501`

(It may say `ws://` instead of `wss://` depending on whether `Enable SSL` is clicked, but the result is the same.)

#2 above there will not work correctly.  So, I think this will not solve the problem.

Therefore, I think it is just that there are deeper issues here than just a parameter name.  I agree with both of you that when you have a non-loaded overlay in `Overlay WSServer` it has the wrong parameter, but more than that, it has the wrong url as well.  You need the `/ws` on the end of the url to get the "[modern api](https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/WSServer.cs#L102)".

Here are a couple of suggestions of other ways to handle this:
1) Translated error messages, if somebody includes HOST_PORT for cactbot, possibly also add this to ngld/overlayplugin's [common.js](https://github.com/ngld/OverlayPlugin/blob/master/docs/assets/shared/common.js)
2) Maybe a pull request to ngld/overlayplugin to include both parameters in the `Overlay WSServer` ui, e.g. `ui/raidboss/raidboss.html?HOST_PORT=wss://127.0.0.1:10501&OVERLAY_WS=wss://127.0.0.1:10501/ws` in the case that the plugin isn't loaded.

PR #1466, opened 2020-05-25
util: Add log splitter and anonymizer

quisquous:
This is just a work in progress.


JLGarber:
Very nice! I like this approach overall.

You've probably already thought through some of this, but what could we do to mitigate the effects of changes to log lines over time? One significant one that jumped out at me was the changes to 03/04 log lines from 2018 to now:

`03|2018-01-10T22:39:30.8360000-05:00|12345678|JLGarber|17|46|94ce|2508|0|9d676b7c08ce13221483fd7dd306acae`

`03|2020-05-25T12:22:09.8540000-04:00|12345678|JLGarber|17|50|0|29|Server|0|0|100000|100000|10000|10000|0|0|-1.653331|18.39634|-2.37337|3.126688||Hash`

For summons, there's a slightly different difference:

`03|2018-01-11T21:07:29.6550000-05:00|4000afcd|Garuda-Egi|0|3c|4b72|2288|OWNERID|HASH`

`03|2020-05-25T12:22:09.8540000-04:00|40080d26|Topaz Carbuncle|0|1a|OWNERID|0||1400|1011|590|590|10000|10000|0|0|33.21568|39.45135|1.2|2.956841||HASH`

Other than that, the only other thing I would note for now is that it might be nice to have the option to anonymize players as "Player1-PlayerN" rather than "partially-random-cat-name string", for human-readability purposes.

quisquous:
> You've probably already thought through some of this, but what could we do to mitigate the effects of changes to log lines over time? One significant one that jumped out at me was the changes to 03/04 log lines from 2018 to now:

Yeah, this is something that worries me a little.  It's along the same lines as hashes not appearing in older ones.

I wonder a little about having a version parameter that would attempt to autodetect from log lines or could be specified directly.  Then if we knew the version, we could have some set of `logDefinitions` broken out by version, and include only the valid ones.  Maybe they could be sparse sets, such that they'd only include elements that changed from a previous version.

> Other than that, the only other thing I would note for now is that it might be nice to have the option to anonymize players as "Player1-PlayerN" rather than "partially-random-cat-name string", for human-readability purposes.

Except `1` isn't a valid character in a player name.  I mean, yes, maybe it doesn't matter, but it would break anybody with a regex that was trying to match player names and didn't include (invalid) numbers.  I would maybe also argue that player ids also cover the "Player N" angle, since they start from `10FF0000` and count up.

JLGarber:
> I wonder a little about having a version parameter that would attempt to autodetect from log lines or could be specified directly. Then if we knew the version, we could have some set of logDefinitions broken out by version, and include only the valid ones. Maybe they could be sparse sets, such that they'd only include elements that changed from a previous version.

I think that's reasonable. Obviously we can't 100% rely on the header containing the log version to be present, but I think for the vast majority of cases it would be a safe assumption. The `253` line doesn't really seem to have changed much:

`253|2018-01-11T19:15:10.5906001-05:00|FFXIV PLUGIN VERSION: 1.6.1.8, DELAY MODE: FastSleep|HASH`

`253|2020-05-25T12:22:05.4887389-04:00|FFXIV PLUGIN VERSION: 2.0.5.10|HASH`

> Except 1 isn't a valid character in a player name. I mean, yes, maybe it doesn't matter, but it would break anybody with a regex that was trying to match player names and didn't include (invalid) numbers. I would maybe also argue that player ids also cover the "Player N" angle, since they start from 10FF0000 and count up.

I'm skeptical that that first use case matters. I guess someone *might* search for player names with a negative-lookup for digits? But I feel like most people in a position to be using both regex and this anonymizer would be able to search by line field instead.

As for the "player IDs cover it" point, it's possible that it's just me, but I find it *really* hard to differentiate random-looking information visually. There have been plenty of cases, for me (Sophia comes to mind,) where it's a lot easier to inspect things for patterns visually rather than work out how to do it programmatically. 

What if instead we were to do something like `PlayerAAA-PlayerZZZ`? I don't know, I find this *much* more difficult:

```
04|2018-01-11T22:45:23.1120000-05:00|10FF012D|X'xzzrtd Tia|6|a|a8|ea|0|01234567012345670123456701234567
04|2018-01-11T22:45:23.1120000-05:00|10FF00E6|X'xzzrqk Tia|5|10|174|d1|0|01234567012345670123456701234567
04|2018-01-11T22:45:23.1120000-05:00|10FF0131|X'xzzrth Tia|18|46|8402|4aad|0|01234567012345670123456701234567
```

As compared to:

```
04|2018-01-11T22:45:23.1120000-05:00|10FF012D|PlayerAAA|6|a|a8|ea|0|01234567012345670123456701234567
04|2018-01-11T22:45:23.1120000-05:00|10FF00E6|PlayerAAB|5|10|174|d1|0|01234567012345670123456701234567
04|2018-01-11T22:45:23.1120000-05:00|10FF0131|PlayerAAC|18|46|8402|4aad|0|01234567012345670123456701234567
```

It's entirely possible that it's mostly a me problem, and if that's the case, no big deal, I'll just modify it for myself. If it's not *too* much effort to make it an option, though, I think it's worth considering.

panicstevenson:
>What if instead we were to do something like PlayerAAA-PlayerZZZ?

I think you want both a forename and surname here, and preferably want some way to keep the names valid but also decently identifiable. I think `Player A` -> `Player Z` doesn't buy enough potential names, and `Aa` -> `Zz` would be unfortunate if you incremented on the right-most character then incremented the left-most character once you've exhausted all options, since `Player Aa` and `Player Ab` aren't _super_ distinguishable? I'd say you could get around this by having it randomly give you one of the 256 potential options, but you'd have to store them in a dictionary or something. Overall, pretty low cost, but also kind of bleh?

There's also the fact that if you used alpha-only, you have... not great... combinations at both 2 and 3-letter combinations, and at 4-letters you're getting into a slew of interesting words that could litter your code.

I do potentially agree though that while player names are pretty fun, it does make it less legible; but, I wonder how much that matters?

quisquous:
Lalafell name generator incoming

DrLippe:
> Lalafell name generator incoming

Helping wherever i can 😉 
```js
Array.prototype.random = function() { return this[Math.floor(Math.random() * this.length)]; }

function generateLalafellName() {
    const consonants = 'B,C,D,F,G,H,J,K,L,M,N,P,Q,R,S,T,V,W,X,Z'.split(',')
    const vowels = 'A,E,I,O,U,Y'.split(','); // I know, Y is not technically a vowel

    let firstConsonant = consonants.random().toLowerCase();
    let secondConsonant = consonants.random().toLowerCase();
    
    let firstVowel = vowels.random().toLowerCase();
    let secondVowel = vowels.random().toLowerCase();

    let fname = firstConsonant.toUpperCase() + firstVowel + firstConsonant + firstVowel + secondConsonant + secondVowel;
    let lname = fname.charAt(2).toUpperCase() + fname.slice(3); 

    return `${fname} ${lname}`;
}
```

util/logtools/fake_name_generator.js
```diff
@@ -0,0 +1,136 @@
...
+
+  makeLalaSyl2() {
+    return this.makeLalaSyl1() + this.makeLalaSyl1();
+  }
+
+  makeLalaSyl1Or2() {
+    return randomFromArray([this.makeLalaSyl1(), this.makeLalaSyl2()]);
+  }
+
+  makeRandomLalafellName() {
```
quisquous, line 82: Fine! <_<

quisquous:
Ok, I think this is looking pretty reasonable at the moment.  I need to do a bit more checking to make sure these anonymized logs can still be imported.  I think I need to include all the combatants in a split log as well for ACT to be able to figure out where the combat starts, and there's also a few issues where parsed messages don't end up having names correctly, e.g. `[18:23:18.663] 1E:10FF000C: loses the effect of Dualcast from .`

Output looks like:
```
$ node split_log.js -f /d/ffxivlogs/log/Network_20510_20200716.log -lf
2020-07-17
   1) 01:08:02  1m53s Mist                      [Zone]
                      ~Syrcus Tower~
   2) 01:11:07    45s The Ladder of Ascension   [Unseal]
   3) 01:12:27  1m56s The Gathering             [Unseal]
   4) 01:14:51  1m10s The Braid                 [Unseal]
   5) 01:16:24  1m24s The Ring of the Protector [Unseal]
   6) 01:18:16    56s The Golden Sacristy       [Unseal]
   7) 01:19:38   2m8s The Final Curtain         [Unseal]
   8) 01:22:15   3m7s The Emperor's Throne      [Unseal]
   9) 01:29:07  4m17s Il Mheg                   [Zone]
  10) 01:34:01  5m33s Containment Bay (S1T7)    [Win]
  11) 01:40:43    44s Il Mheg                   [Zone]
  12) 01:42:31 11m20s Lakeland                  [Zone]
                      ~Castrum Abania~
  13) 01:58:14  1m52s Terrestrial Weaponry      [Unseal]
  14) 02:03:31  1m36s Project Aegis             [Unseal]
  15) 02:09:01  2m21s The assessment grounds    [Win]
  16) 02:13:04  2m41s Lakeland                  [Zone]

$ node split_log.js -f /d/ffxivlogs/log/Network_20510_20200716.log -lf 8
Wrote: SyrcusTower_TheEmperorsThrone_20200717_012215_3m7s.log

$ node split_log.js -f /d/ffxivlogs/log/Network_20510_20200716.log -lz
2020-07-17
  1) 01:05:53   4m2s Mist
  2) 01:09:54 16m36s Syrcus Tower
  3) 01:26:30  2m25s Mist
  4) 01:28:55  4m29s Il Mheg
  5) 01:33:23  6m37s Containment Bay (S1T7)
  6) 01:40:00  1m27s Il Mheg
  7) 01:41:27 12m23s Lakeland
  8) 01:53:50 18m12s Castrum Abania
  9) 02:12:02  3m43s Lakeland

$ node split_log.js -f /d/ffxivlogs/log/Network_20510_20200716.log -lz 8
Wrote: CastrumAbania_20200717_015350_18m12s.log
```

I think there could be some more cleanup to skip overworld zones, but I think this is fine for now.  I think it's plausible to reshare the `encounter_tools.js` with #1387 and not duplicate that logic in a third place.

The follow-up from this is to make a webpage where folks can drop a log in and do this logic, but I'll probably land this first.

The real follow-up is to then create a log repository of various logs across all content that people can refer to for trigger, timeline, and reference purposes.

quisquous:
Ok, fixed the ACT issues.  It really needs all of the combatant lines there to know what to do with players and encounter starts.

I think this is ready to go, if anybody has any other feedback?

quisquous:
Good enough.  We can fix it in post if anybody has thoughts.

PR #1471, opened 2020-06-06
oopsy: update e7n/e7s light dark course code

quisquous:


ui/oopsyraidsy/data/05-shb/raid/e7s.js
```diff
@@ -1,15 +1,24 @@
 'use strict';
 
-// TODO: when it's ok to take a light's course or dark's course
-// TODO: figuring out when courses are colored or not
 // TODO: missing an orb during tornado phase
 // TODO: jumping in the tornado damage??
 // TODO: taking sungrace(4C80) or moongrace(4C82) with wrong debuff
 // TODO: stygian spear/silver spear with the wrong debuff
 // TODO: taking explosion from the wrong Chiaro/Scuro orb
```
JLGarber, line 7: For my own edification, is there any missing knowledge for these, or are they just planned for the next version when there's time?

JLGarber:
I appreciate the changes from my initial attempt at this, the revision is definitely cleaner than devoting an entire extra trigger to initialization. Would it perhaps be possible to include a `deathReason` trigger for the tornado pushback?

ui/oopsyraidsy/data/05-shb/raid/e7s.js
quisquous: I don't know what missing an orb does or jumping in the tornado.  I know I have an example in my logs of each of these if needed, but I don't know offhand at all what they are at all.

Also, these TODOs are definitely "if anybody else has time".  I'm sure I will get distracted by other things, and so it's more of a "patches welcome" and "this is what's missing" rather than "this is my personal roadmap".

JLGarber: Do we have precedent for a "group failure" mechanic like missing the orbs, now that I think about it? I'm trying to think of how to map that out, but without knowing party roles, it's really tough to isolate anything on that.

As for the rest of it, I should be able to look at that that At Some Point. Can't be that difficult.

quisquous:
> I appreciate the changes from my initial attempt at this, the revision is definitely cleaner than devoting an entire extra trigger to initialization. Would it perhaps be possible to include a `deathReason` trigger for the tornado pushback?

Yeah, can do.  It always makes me sad to add stuff like that, but maybe some day the network data for that can be tracked down and it can be done "properly".

JLGarber:
We were able to clean up "wipe detection via boss HP/zone seals/whatever", we can clean it up here in the future. Nothing wrong with doing the best we can for now!

ui/oopsyraidsy/data/05-shb/raid/e7s.js
quisquous: I think missing interrupts are a "group failure" as well.  I think this is not handled that well at the moment.  I think in some cases it probably also blames it on the target (which is probably the mob casting the thing that needs to be interrupted).

Maybe it'd be better to leave the blame empty in these cases, unless you have a better idea.

PR #1483, opened 2020-06-16
Raidboss: Varis ex-fr+ common.js all language

VanSlanzar:
added corrections and missing translations fr on varis-ex.
removed translations on varis-ex, sophia-ex, shinryu for added on common.js all language.

### DIFF
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -15,12 +15,26 @@ let commonReplacement = {
     ja: 'の封鎖が解かれた',
     ko: '의 봉쇄가 해제되었습니다',
   },
+  '--adds spawn--': {
+    de: '--adds spawn--',
+    fr: '--apparition d\'adds--',
+    ja: '--adds spawn--',
+    cn: '--adds spawn--',
+    ko: '--쫄 소환--',
+  },
   '--center--': {
     de: '--mitte--',
     fr: '--centre--',
     ja: '--センター--',
     cn: '--中央--',
     ko: '--중앙--',
+  },
+  '--clones appear--': {
+    de: '--klone erscheinen--',
+    fr: '--apparition des clones--',
+    ja: '--幻影が現れる--',
+    cn: '--clones appear--',
+    ko: '--clones appear--',
   },
   '--corner--': {
     de: '--ecke--',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -29,12 +29,12 @@ let commonReplacement = {
     cn: '--中央--',
     ko: '--중앙--',
   },
-  '--clones appear(\\?)?--': {
-    de: '--klone erscheinen$1--',
-    fr: '--apparition des clones$1--',
-    ja: '--幻影出現$1--',
-    cn: '--克隆 体 出现$1--',
-    ko: '--분신 소환$1--',
+  '--clones appear--': {
+    de: '--klone erscheinen--',
+    fr: '--apparition des clones--',
+    ja: '--幻影出現--',
+    cn: '--克隆 体 出现--',
+    ko: '--분신 소환--',
   },
   '--corner--': {
     de: '--ecke--',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -18,7 +18,7 @@ let commonReplacement = {
   '--adds spawn--': {
     de: '--adds spawn--',
     fr: '--apparition d\'adds--',
-    ja: '--adds spawn--',
+    ja: '--雑魚出現--',
     cn: '--adds spawn--',
     ko: '--쫄 소환--',
   },
@@ -32,9 +32,9 @@ let commonReplacement = {
   '--clones appear--': {
     de: '--klone erscheinen--',
     fr: '--apparition des clones--',
-    ja: '--幻影が現れる--',
+    ja: '--幻影出現--',
     cn: '--clones appear--',
-    ko: '--clones appear--',
+    ko: '--분신 소환--',
   },
   '--corner--': {
     de: '--ecke--',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -29,12 +29,12 @@ let commonReplacement = {
     cn: '--中央--',
     ko: '--중앙--',
   },
-  '--clones appear--': {
-    de: '--klone erscheinen--',
-    fr: '--apparition des clones--',
-    ja: '--幻影出現--',
-    cn: '--克隆 体 出现--',
-    ko: '--분신 소환--',
+  '--clones appear(\\?)?--': {
+    de: '--klone erscheinen$1--',
+    fr: '--apparition des clones$1--',
+    ja: '--幻影出現$1--',
+    cn: '--克隆 体 出现$1--',
+    ko: '--분신 소환$1--',
   },
   '--corner--': {
     de: '--ecke--',
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -19,7 +19,7 @@ let commonReplacement = {
     de: '--adds spawn--',
     fr: '--apparition d\'adds--',
     ja: '--雑魚出現--',
-    cn: '--adds spawn--',
+    cn: '--小怪出现--',
     ko: '--쫄 소환--',
   },
   '--center--': {
@@ -33,7 +33,7 @@ let commonReplacement = {
     de: '--klone erscheinen--',
     fr: '--apparition des clones--',
     ja: '--幻影出現--',
-    cn: '--clones appear--',
+    cn: '--克隆 体 出现--',
     ko: '--분신 소환--',
   },
   '--corner--': {
--- a/ui/raidboss/data/05-shb/trial/varis-ex.js
+++ b/ui/raidboss/data/05-shb/trial/varis-ex.js
@@ -54,6 +54,7 @@
             alertText: {
               en: 'Dodge Clones + Stack',
               de: 'Klonen ausweichen und Sammeln',
+              fr: 'Évitez les Clones + packez-vous',
               ja: 'ターミナス・エストを避け／頭割り集合',
               cn: '躲避剑气 + 集合分摊',
               ko: '분신 피하기 + 집합',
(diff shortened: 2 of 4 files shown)


ui/raidboss/common_replacement.js
```diff
@@ -15,13 +15,27 @@ let commonReplacement = {
     ja: 'の封鎖が解かれた',
     ko: '의 봉쇄가 해제되었습니다',
   },
+  '--adds spawn--': {
+    de: '--adds spawn--',
+    fr: '--apparition d\'adds--',
+    ja: '--adds spawn--',
+    cn: '--adds spawn--',
+    ko: '--쫄 소환--',
```
VanSlanzar, line 23: @MaikoTan, @Jaehyuk-Lee, @Akurosia, @r4ha.
I need your help translating these lines, if there is any need. They were not translated in the other files, sorry to bother you again.

ui/raidboss/common_replacement.js
```diff
@@ -15,13 +15,27 @@ let commonReplacement = {
...
     ja: '--センター--',
     cn: '--中央--',
     ko: '--중앙--',
   },
+  '--clones appear--': {
+    de: '--klone erscheinen--',
+    fr: '--apparition des clones--',
+    ja: '--幻影が現れる--',
+    cn: '--clones appear--',
+    ko: '--clones appear--',
```
VanSlanzar, line 37: @MaikoTan, @Jaehyuk-Lee, @Akurosia, @r4ha.
I need your help translating these lines, if there is any need. They were not translated in the other files, sorry to bother you again.

ui/raidboss/common_replacement.js
Akurosia: `de: '--Adds erscheinen--'`

ui/raidboss/common_replacement.js
Akurosia: `de: '--Klone erscheinen--',`

ui/raidboss/common_replacement.js
MaikoTan: I am not very sure what is "spawn" meanning.
If it means "appear / turn to visible", I think this can be translate as "--雑魚出現--".
@r4ha what do you think?

ui/raidboss/common_replacement.js
MaikoTan: `ja: '--幻影が現れる--',`
or
`ja: '--幻影出現--',`

@r4ha what do you think? In my opinion, I prefer the pure Kanji version.

ui/raidboss/common_replacement.js
Una-P: no problem at kor.

ui/raidboss/common_replacement.js
Una-P: `ko: '--분신 소환--',`

r4ha: > `ja: '--幻影が現れる--',`
> or
> `ja: '--幻影出現--',`
> 
> @r4ha what do you think? In my opinion, I prefer the pure Kanji version.

Yea I think you're right!

ui/raidboss/common_replacement.js
r4ha: > I am not very sure what is "spawn" meanning.
> If it means "appear / turn to visible", I think this can be translate as "--雑魚出現--".
> @r4ha what do you think?

Agree!

VanSlanzar: No one for cn translation? Otherwise I will leave the English words. Thank you all.

ui/raidboss/common_replacement.js
VanSlanzar: No one for cn translation? Otherwise I will leave the English words. Thank you all.

ui/raidboss/common_replacement.js
MaikoTan: > No one for cn translation? Otherwise I will leave the English words. Thank you all.

Oh sorry I forgot it!
I think it should be '--小怪出现--' in cn.

ui/raidboss/common_replacement.js
MaikoTan: > No one for cn translation? Otherwise I will leave the English words. Thank you all.

This word was in Varis-ex at the first time, right? The CN server was now remain in 5.15 version because coronavirus, so there is no a common translation for now.
I think you can temporary leave it as '--克隆体出现--' for now.

As the ja translation, '--幻影出現--' was the best, thanks @r4ha !.

ui/raidboss/common_replacement.js
VanSlanzar: Thank so much.

ui/raidboss/common_replacement.js
VanSlanzar: Thank so much.

VanSlanzar:
I want to thank everyone for the missing translations. this should be good for this request.

ui/raidboss/data/05-shb/trial/varis-ex.js
```diff
@@ -359,10 +360,9 @@
         'Ventus Est': 'Ventus Est',
       },
       'replaceText': {
+        '\\?': '?',
```
quisquous, line 363: Sorry, can you explain what this line is doing?

VanSlanzar: Hi sure. 

This lines are in text file:
`519.7 "--clones appear--" sync /:Phantom Varis:4CB3:/`
This line is listed in the `common.js` file

`294.0 "--clones appear?--" #sync /:Phantom Varis:4CB3:/`
this line has`?` more. If I delete this line for example in de: `'\\?': '?',`, that gives an error.
varys-ex

Having it corrects this error. I don't know of any other solution.

MaikoTan: I think this text `--clones appear?--` here is just a fake syncing text for placeholding, maybe the `?` can be deleted?


And a little suggestion here:
when you try to share your cmd output information, you can just copy those text to comment, and quote them with \`\`\` and \`\`\`.

just like this:

>\`\`\`
>// your output info
>\`\`\`

then it comes to be:

```
// your output info
```

quisquous: The problem is that this translation does not do anything.

If the original text is `--clones appear?--` then `'\\?': '?'` translates `?` into `?` leaving you with the original unchanged `--clones appear?--`.  Therefore, that line does not translate anything, it only suppresses the error from testing.  That is my concern.

I see two options:
 1) translate `--clones appear\\?--` as its own line
 2) Change the common translation like this:

```
  '--clones appear(\\?)?--': {
    de: '--klone erscheinen$1--',
    fr: '--apparition des clones$1--',
    ja: '--幻影出現$1--',
    cn: '--克隆 体 出现$1--',
    ko: '--분신 소환$1--',
  },
```

Unfortunately, I don't think there's any way to add a space before the optional question mark here for French.  So if you want it to say `--apparaition des clones ?--` then option 1 is the only approach that would work.

Similarly, I think for the targetable one in #1484 you could do somethinglike:

```
  '--([0-9]+x )?targetable--': {
    de: '--$1anvisierbar--',
    fr: '--$1ciblable--',
    ja: '--$1ターゲット可能--',
    cn: '--$1可选中--',
    ko: '--$1타겟 가능--',
  },
```

And this would allow `--targetable--` to translate as `--ciblable--` and `--2x targetable--` to be `--2x ciblable--` with the same line.

quisquous:
I also wanted to add that in the future, it'd be better not to ping everybody for translations.  There's plenty of translations missing in cactbot and the `find_missing_translations` script will find any new ones.  Adding common translations doesn't increase the amount of missing translations, and so I think it's fine to land without them and have people follow up individually as they have time.

VanSlanzar:
As there are other lines with `?`, This creates collisions. Here I took the first solution, create a separate line, directly on the Varis file for de and ja. For fr `?` with space wokrs well for everything translations fr.

quisquous:
> As there are other lines with `?`, This creates collisions in fr. Here I took the first solution, create a separate line, directly on the Varis file for de and ja. For fr `?` with space works well for everything translations fr. Thank you.

Unfortunately, this will not work, I don't think?  You are translating `--clones appear?--` with the replacement: `'\\?': ' ?'`.  This will turn it into `--clones appear ?--`, which does not seem correct.

You may want to do something like this combination of lines:
```
'\\?(?!--)': ' ?',
'--clones appear\\?--': 'apparition des clones ?',
```

VanSlanzar:
I did not understand that the translation would not be done. I try your lines, but it doesn't work. Can you help me ? thank you.

quisquous:
Looking at https://github.com/quisquous/cactbot/runs/793028199?check_suite_focus=true, the error is: `/home/runner/work/cactbot/cactbot/ui/raidboss/data/05-shb/trial/varis-ex.js:locale fr: post-translation collision on replaceText '--clones appear?--' for '--clones appear\?--' => 'apparition des clones ?', then '\?(?!--)'`

This means that `'\\?(?!--)'` is matching `'apparition des clones ?'` and trying to translate that question mark, when it shouldn't.  I think the way to fix this is to make it not match against spaces beforehand, e.g. change it to `'(?<! )\\?(?!--)'` In other words, match a literal question mark as long as there's not a space before it or two hyphens after it.

quisquous:
Out of curiosity, are you able to run `npm run test` and `npm run lint` locally?

VanSlanzar:
I tried again, without success. I had tried several combinations, but I could not find the right one. `npm run test` yes. i do all time. `npm run lint` no. I will add it to my list, I did not remember this line.

quisquous:
Sorry, I don't understand your last comment.  What do are you talking about in terms of "trying"? Are you able to see the exact error from your last commit?

VanSlanzar:
I use all the time `npm run test` . I see errors don't worry about that, I got it right. But, I tried several combinations to resolve this conflict. Without success. I post it even if it does not work to show you the problems encountered. I hope you understand me better.

quisquous:
This is your patch:
```diff
-        '\\?(?!--)': ' ?',
+        '(?<! )\\?(?!--)',
```

This is the error:
```
/home/runner/work/cactbot/cactbot/ui/raidboss/data/05-shb/trial/varis-ex.js
  405:26  error  Parsing error: Unexpected token ,
```

Each line in the translation block needs to look like `'x': 'y',` for various values of `x` and `y`.  In this case, JavaScript is complaining because there's no colon or `'y'`.

I think this instead should be: `'(?<! )\\?(?!--)': ' ?',`.  Does this make more sense?

VanSlanzar:
................., my sincere apologies. I worked surrounded by my children today. And with all these symbols, I did not see that my line was not correct. This must have been the case several times during my tests. Thank you for your patience and your help.

PR #1485, opened 2020-06-17
removed all non used translations in stormblood + 2 small fixed …

Akurosia:
for incorrect comparisons:

o2s - fix for 6 fulms under
hydatos - removed not used backslash for proto ozma

NOTE: this is a preperation step to add the missing translations


VanSlanzar:
This is great, there are a lot of unnecessary lines in `replaceSync` or `replaceText`. I've removed a lot of them from ShB, but if you're using a script, I'd be glad you all double-check ShB, because i do manuelly.

Akurosia:
Yeah ShB is next to make sure everything is clean before doing any additional translation

ui/raidboss/data/04-sb/alliance/orbonne_monastery.js
```diff
@@ -707,7 +707,6 @@
       'replaceSync': {
         'Agrias': 'Agrias',
         'Aspersory': 'Aspersorium',
-        'Auracite Shard': 'Auracitenscherbe',
```
quisquous, line 710: Sorry, this is hard for me to review deletions when `missingTranslations` is true.  It is not clear to me without a lot more work whether you are removing translations that are required or translations that are unused.

I am sorry to request something tedious, but can you not remove translations from any section that is still missing translations? If it would be less work for you, I would be happy to file an issue and add some script code that would give warnings or errors on unused translations in these cases so you could find them automatically rather than re-doing this patch by hand.

If it would be helpful to prioritize translation issues that are most useful to cactbot, I would suggest looking at it with this view:
* highest priority: anything missing from `find_missing_translations.py` (other than a7s)
* medium priority: removing the `missingTranslations: true` from any section
* lowest priority: removing unused translations
* please don't do at all: changing capitalization of syncs

Akurosia: All of the removals are elements that are neither part of a regex or are not in the txt version of the fight.

The reason behind it, is to avoid translating to much unaccassary things as some of them (not all) have a lot of incorrect stuff (e.g. elements from other fights)

quisquous: I don't disagree with removing.

However, my primary worry is that I don't want to remove an element that is currently in use, because that would make the state of translations worse.  When you are removing a bunch of lines from a timeline that has `missingTranslations: true` that it is impossible for me to know without verifying each of them individually whether or not that element is truly in use or not.  Even if you are doing it with a script, how do I know your script is correct? For sections where they are `missingTranslations: false`, then it is very easy to verify because the test would complain if it was in use.

I think it would be strictly better (for me, reviewing the pull request) to remove `missingTranslations: true` first, and then after that come back and remove unnecessary lines.

Sorry to be so stubborn on pull requests lately.  I just only have so much energy and time for reviewing, and so trying to figure out how to make things be as easy to read and verify as possible is important.

Akurosia: Ok than i can't support other languages anymore since i can't fix the missing translations for other languages (if they are not from the gamefiles directly). 

If it might help you (thats what i did to somehow veryfi the changes in the end) i just removed all `'missingTranslations': true,` from all files in 04-sb (easy to do with sublime search replace just in a specified folder) and compare the reported results with the removals.

quisquous: Hmm.  That approach seems plausible but it will still take me a while to go through them all to verify.  I can still do it, it's just a lot.

(The problem here is that your script is taking all the translation lines that exist and translating them, and so if there's unnecessary ones then it can't do it? Could it just skip things it can't find from the data files? I guess it would help me to understand what the issue with leaving these lines in is.)

quisquous: Sorry also for all the push back on pull requests lately.  I hope you understand how much your contributions are appreciated.  This is another case where I mostly don't have enough information about how your process and script works and so I'm trying to find some middle ground where you can automatically translate things and it is not hard for me to review those translations.

Akurosia: the script i am using is a modification of your `check_translation.py`.

It is not 100% optimal because in very, very rare cases it has false positives in it (i think there was a case with infernal fetters and fetters, which i had to look up in more detail)

quisquous:
I fixed the conflicts on your patch and rebased on top of latest.  Then, I did this locally:

```bash
cat > all_translations.sh
python util/find_missing_translations.py -f 04-sb -l de
python util/find_missing_translations.py -f 04-sb -l fr
python util/find_missing_translations.py -f 04-sb -l ja
python util/find_missing_translations.py -f 04-sb -l cn
python util/find_missing_translations.py -f 04-sb -l ko
^D

source all_translations.sh > missing.master
git checkout Akurosia-new_remove_try
source all_translations.sh > missing.akurosia
diff missing.master missing.akurosia | less
```

The differences can be summed up as such:

1) o9n

Your patch removes Knock when it shouldn't.  (This is what I was worried about.  It makes me wish that we had better testing.  Would you prefer that there were errors when there were translation entries that don't apply anywhere?)

2) ucob

I re-did the ucob timeline (sorry) and so you need to re-run your script, as many things that were removed are now present.

3) Backslash woes

(spot the trailing `+` on `positives magnetfeld`)

`ui\raidboss\data\04-sb\alliance\ridorana_lighthouse.js:483 Ridorana Yiazmat Magnetic Positive: incorrect timelineReplace for regex, got '(?<timestamp>^.{14}) 1a:(?<targetid>[0-9a-f]{8}):(?<target>.*?) gains the effect of (?<effect>positives magnetfeld+) from (?<source>.*?) for (?<duration>-?[0-9]+(?:[.,][0-9]+)?(?:e-?[0-9]+)?) seconds\.', needed regexDe '(?<timestamp>^.{14}) 1a:(?<targetid>[0-9a-f]{8}):(?<target>.*?) gains the effect of (?<effect>positives magnetfeld) from (?<source>.*?) for (?<duration>-?[0-9]+(?:[.,][0-9]+)?(?:e-?[0-9]+)?) seconds\.'`

...whereas previously it was:
`ui\raidboss\data\04-sb\alliance\ridorana_lighthouse.js:483 Ridorana Yiazmat Magnetic Positive: missing timelineReplace for regex '(?<timestamp>^.{14}) 1a:(?<targetid>[0-9a-f]{8}):(?<target>.*?) gains the effect of (?<effect>magnetic lysis \+) from (?<source>.*?) for (?<duration>-?[0-9]+(?:[.,][0-9]+)?(?:e-?[0-9]+)?) seconds\.' (but have regexDe)`

So changing to `'Magnetic Lysis \\\\+': 'Positives Magnetfeld',` doesn't quite match `Regexes.gainsEffect({ effect: 'Magnetic Lysis \\+' }),`.  

You, uh, need another two backslashes.  <_<

One alternative way to """"""fix"""""" this would be to switch to `netRegex: NetRegexes.gainsEffect({ effectId: '60E' }),`.

4) Bonus: you've fixed a bunch of regex replacement errors (thanks!)

quisquous:
I think also I would suggest for the future not combining "small fix" patches with "cleanup removal" patches.  It may seem silly to have to separate them out, but it's easier for me to review something if all a patch is doing is doing exactly one thing.

ui/raidboss/data/04-sb/raid/o9n.js
```diff
@@ -190,44 +127,23 @@
...
+      'missingTranslations': true,
       'replaceSync': {
         'Chaos': '카오스',
-        'Chaosphere': '혼돈의 구체',
-        'dark crystal': '흑수정',
         'YOU DARE!': '네 이노오오옴',
       },
       'replaceText': {
         'Big Bang': '돌출',
-        'knock': '착탄',
```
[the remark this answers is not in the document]
Akurosia: I readded the knock line (also capitalized the K, as i think it is necassary in repalceText)

Akurosia:
yeah i am sry for the small patches but i would have forgotten them otherwise (this is way i have listed them)
I also like the idea of the netRegex tbh, as i dont like to put a billion backslases on everything x.x

Also dont worry with the ucob timeline, should be easy to redo this one

Akurosia:
wtf... i did a rebase and it just destroyed everything... u_u

@quisquous i think i will need to make a clean PR without the unrelated stuff...

Akurosia:
opened #1508 without the clutter of the rebase issue

PR #1486, opened 2020-06-17
Raidboss: Bahamut fr

VanSlanzar:
I just noticed that I was not always looking at the right line on the logs. There are several, but I understood which one is taken into account. So I correct those that I missed or wrote poorly. 
My apologies.


quisquous:
I'll merge this one change, but I will say it again that capitalization does not affect how the syncs work.  These all worked before.  These will all still work after.  I don't think it's worth your time to change these.

PR #1493, opened 2020-06-19
Raidboss: Eden's verse consistency

VanSlanzar:
I have modified names on `replaceSync`  fr and ja for more consistency with `replaceSync` de, text file and networks logs. 
I know it is not particularly useful, it seems to me. But I find it more consistent that the names are the same and with their punctuation everywhere. Sorry again.


quisquous:
Sorry, I do not think this is worth doing.  I think there is inconsistency between the way the game displays things and the way network logs display things (and these also change over time).  As `replaceSync` is case insensitive, it is not worth changing these.

I do not mean to be rude here, but I think this is just not worth the time to change.

Issue #1255, opened 2020-03-21
Manage localization via cooperation system


Bluefissure:
## Summary

We can use a cooperation system like [crowdin](https://crowdin.com/) or [transifex](https://www.transifex.com/) to cooperate with the localizations of responses and timeline texts of **raidboss** module. (Or even the whole project?)

Then we can make it easier for those who cannot code well to help with the translation (without modifying the js file).

These two system are both free for open-source projects ([crowdin](https://crowdin.com/page/open-source-project-setup-request), [transifex](https://docs.transifex.com/projects/open-source-project)).

## Current Progress

I have a demo project to export/import the responses/timelines from/to the js files, which is mostly designed for CN localizations:
- Project: [raidboss_translator](https://github.com/Bluefissure/raidboss_translator)
- Sheet: [exported sheet](https://docs.qq.com/sheet/DSFRGa1V1d2pOZk1H?tab=qwmxvl&c=E42A0RF0)

Do you think it's necessary and more efficient for later maintenance?

quisquous:
Sorry for the delay in responding here.

I have to admit that I am a little skeptical of the overhead of this approach.  Do you feel that there are many people who would be contributing to cactbot translations but aren't because they don't know JavaScript? I would hope that it is not too hard to translate things in JavaScript files, but I am somewhat biased here because I am a programmer.  But, I think the presence of somebody like VanSlanzar who is not a programmer speaks well to other people being able to pick it up and copy surrounding code.

One very big downside to having things in a file is that there is already confusion about what a particular mechanic means, and having translations in separate sheets and projects farther away from the code will make it harder to understand what "dorito stack" or "pepperoni" mean.

I think the hard part of cactbot translations is the replacement restrictions (collisions, tests, etc) and general regex knowledge.  I don't think a separate translation sheet will help with either of these problems.

One thing that I think has helped translations is to have `resources/responses.js` as well as `ui/raidboss/common_translation.js`.  I think another thing that will help along the lines of what @MaikoTan asked for is to do something like this https://github.com/quisquous/cactbot/issues/1307#issuecomment-608283837 for common translations.  I personally don't think it's worth doing this for absolutely everything, but just things that get used multiple times.  That will just cut down on the total number of translations that are needed overall.

Sorry for the skepticism, but I am just not convinced that an approach like this solves enough problems in my mind to be worth the extra cost to do it.

Bluefissure:
[...]

PR #1494, opened 2020-06-20
Raidboss: anamnesis_anyder

VanSlanzar:
added missing translations fr
corrections translations de and fr on `replaceText`
remove useless lines on `replaceSync`
@Akurosia, I have corrected a German translation error, can you confirm that it is correct, please.
@mooondark, `'Hammer/Blade Mark': 'Marque Marteau/Lame',`, c'est correct ou je dois mettre `Marque de Marteau` ?


ui/raidboss/data/05-shb/dungeon/anamnesis_anyder.js
```diff
@@ -305,27 +278,13 @@
       'locale': 'ja',
       'missingTranslations': true,
       'replaceSync': {
-        'trench yovra': 'トレンチ・ヨヴラ',
```
quisquous, line 308: Does this section still need `missingTranslations`?

VanSlanzar: This is missing, on `replaceSync`:
`'Depth Grip': `
`'Doxa': '`
`'Katharsis': `
`'Noesis':` 
`'Sinister Bubble':`

JLGarber:
```
'Depth Grip': 'ハンド・オブ・オーシャン',
'Doxa': 'ドクサの広間',
'Katharsis': 'カタルシスの広場',
'Noesis': 'ノエシスの間',
```

VanSlanzar: Thank you. I am obliged to leave `missingTranslations:true`, because translations are again missing after `npm run test`.
```
C:\Users\Bauer\Documents\GitHub\cactbot\ui\raidboss\data\05-shb\dungeon\anamnesis_anyder.js:locale ja:no translation for replaceText 'Unknown Add'
C:\Users\Bauer\Documents\GitHub\cactbot\ui\raidboss\data\05-shb\dungeon\anamnesis_anyder.js:locale ja:no translation for replaceText 'Swing/Swipe/Cyclone'
C:\Users\Bauer\Documents\GitHub\cactbot\ui\raidboss\data\05-shb\dungeon\anamnesis_anyder.js:locale ja:no translation for replaceText 'Hammer/Blade Mark'
C:\Users\Bauer\Documents\GitHub\cactbot\ui\raidboss\data\05-shb\dungeon\anamnesis_anyder.js:locale ja:no translation for replaceText 'Blade/Hammer Mark'
C:\Users\Bauer\Documents\GitHub\cactbot\ui\raidboss\data\05-shb\dungeon\anamnesis_anyder.js:locale ja:no translation for replaceText 'Pyre/Hearth'
```

quisquous:
In the future, removing unused lines causes me a lot of work and is probably the last priority in terms of translations.  This is why it took me a while to get to this patch.  I would recommend against doing that manually in the future if you can avoid it.  Akurosia doing it with a script still is a lot of work, but at least it is done in one big patch.

ui/raidboss/data/05-shb/dungeon/anamnesis_anyder.js
MaikoTan: Yes these lines should be translated. But I have time on weekend only, so I'll do it later. Thanks for mentioning. (ㆁωㆁ*)

MaikoTan:
I have a suggestion here.
Maybe we can make trigger i18n block as (for example a tankbuster on one tank):
```javascript
return translate('Tank Buster on {0}', data.ShortName(matches.target));
```
Then we have a separate file named `dungeon-name.json`, which include:
```javascript
"Tank Buster on {0}": {
  "de": "Tank buster auf {0}" ,
  "fr": "Tank buster sur {0}",
  "ja": "{0}にタンクバスター",
  "cn": "死刑 -> {0}",
  "ko": "\"{0}\" 탱버",
}
```

`translate` function would read out this json file, and find out which localized string to be replaced, based on user's locale settings. And then replace `{0}` and `{1}` (and so on) in the localized string to its second and third (and so on) parameters.

( I learned this approach first time in `react-i18next`, which is an i18n library for react. ( Although I have not seen its source code. ) They also named `translate` function as `t` or `__`, so make it easier to call. )

In this way, we can separate localization and code mechanic into two different files. We can also create a script or GUI tool to convert `json` file to csv or excel sheet (and vice versa), so even someone who doesn't know any js things can contribute translations.

quisquous:
This sounds like the same approach that was talked about in https://github.com/quisquous/cactbot/issues/1255.  Maybe we can follow up more there?

Issue #1487, opened 2020-06-18
oopsy: Use (Net)Regexes instead of custom "events"


quisquous:
oopsy came before Regexes and NetRegexes, which have named matching groups.  To solve this same problem (but in a different way!) oopsy passes an "event" to the first parameter of every function (oopsy does [event, data, matches] vs raidboss doing just [data,matches]).

[These events](https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/oopsyraidsy.js#L800-L832) are more or less the equivalent of Regexes.abilityFull.  There's a few added things, but it's arguable that none of them get used and could get skipped? Or we could turn them into helper functions on `data` rather than calculating them every time.

The reason this is an issue is that Regexes and oopsy events use different names and worse, it's possible to use a `regex: Regexes.headMarker()` (for example) as an oopsy regex and then instead of saying `e.targetName` you have to remember to say `matches.target`.  See: f3cc650c7

It would be better to:
* remove events entirely and make oopsy functions take [data, matches] as parameters
* return all the special information that oopsy wants in matches by using (Net)Regexes
* if there's any additional oopsy-specific information, make helper functions on `data`
* convert gainsEffectRegex/losesEffectRegex into NetRegexes.gainsEffect
  * also use effect ids to avoid translations
* convert abilityRegex into NetRegexes.abilityFull
* consider removing damageRegex / healRegex and converting to NetRegexes.abilityFull
  * at the very least, these should be NetRegexes.abilityFull in terms of the matches they return even if they have more conditions on how they match

cc @JLGarber

JLGarber:
Don't have time right now to fully analyze this, but I'm definitely going to second changing everything over to use NetRegexes. Realistically, the only reason `event` exists is that NetRegexes didn't at the time oopsy was put together. Special-case stuff dedicated for Oopsy should definitely be moved to helper functions.

If it's not a conversion that can be done in one pass, my vote for priority is changing `gainsEffect/losesEffect`. That's the single biggest pain point for me when I'm putting oopsy packages together.

I *think* it might be nice to retain `damageRegex/healRegex` as a convenient shorthand rather than having to write out the `NetRegexes` line in full. Or at least I feel that way. I may not be thinking this through, and it could be better to have One True Way.

quisquous:
[...]

PR #1502, opened 2020-06-21
oopsy: add NetRegex support via `netRegex` field

quisquous:


quisquous:
@JLGarber I think this gets us on the way to #1487.  Do you have any thoughts about this?

JLGarber:
No, nothing to add, it looks good.

PR #1504, opened 2020-06-21
oopsy: add shareWarn/shareFail helpers

### LINKED ISSUE #1488: oopsy: add helper function for double taps
There's a lot of custom code that checks for "is this ability a double tap".  It'd be nice to have some helpers like `damageWarn` or `damageFail` that are like `doubleTapWarn` or `doubleTapFail`.  There are some that also check for things that need to be shared, but there's aren't enough to know for sure if a helper is warranted in the same way.

quisquous:
Fixes #1488.


quisquous:
23 files changed, 140 insertions(+), 663 deletions(-)

cc @JLGarber

JLGarber:
Ahh, that is so good. Can't wait to use it!

PR #1508, opened 2020-06-23
reapplying changes from #1485 without the other commits clunk.

Akurosia:
since #1485 was a bit overloaded with non related changes due to an issue with git rebase, i opened this PR to make it clean again.

I applied the requests from the old ticket:
- o9n = added KR Knock entry (with fixed capitalization)
- ridorana = changed Magnetic Lysis + to netRegex (to avoid using a billin backslashes)
- fixed ucob entries

In regards to the UCOB entries, there are 3 sklls (Lunar Dynamo, Iron Chariot and Raven Dive) that are always commented out in the timeline file. I left them included in the replacemnet section but the question is, are they needed or can they be removed as well?


quisquous:
> In regards to the UCOB entries, there are 3 sklls (Lunar Dynamo, Iron Chariot and Raven Dive) that are always commented out in the timeline file. I left them included in the replacemnet section but the question is, are they needed or can they be removed as well?

Those are all the "nael rp" skills, where nael says something and then does 2 or 3 skills in a row that depend on the random quote that he says.  I left the times in for reference and in case anybody ever wanted to add exact times for the abilities vs just "nael quote".  I don't plan on including them, so if you wanted to remove them, that would be fine.

PR #1510, opened 2020-06-23
config: fix; parser language is set to English

### LINKED ISSUE #1509: parser language is set to English
default language was intended to match FFXIV plugin language, but it's set to English. I don't have computer to check this now, but I see many complaints about why cactbot is shown in English after update.
I'll check this after I go home.

Jaehyuk-Lee:
Fixes #1509


quisquous:
Thanks for the debugging and the quick fix!

PR #1511, opened 2020-06-24
Add Machinist abilities

kairix-gh:
This adds machinist abilities to lang.js and lang/en.js, combo tracking to jobs.js so they can be used in the jobs module. Also adds Peloton to the buff tracking.


ui/jobs/jobs.js
```diff
@@ -792,6 +802,14 @@ class BuffTracker {
         sortKey: 13,
         cooldown: 120,
       },
+      peloton: {
```
quisquous, line 805: What's the goal here of tracking Peloton in the buff tracker? It's only used out of combat, and has a 5 second cooldown, so it's almost always ready to be used.

kairix-gh: I personally just like to keep track of the time left (as opposed to the cooldown) when running lower level dungeons and I have the buff tracker in a central location on the screen.

In any case, since it seems I'll need to write my own module to prevent future updates from overwriting layout changes, you can probably remove it.

quisquous: Belatedly, for what it's worth, you can add the file `cactbot/user/raidboss.css` to change CSS layout, and that file will not get overwritten in future updates.  See: https://github.com/quisquous/cactbot/blob/master/docs/AdvancedCactbot.md#customizing-appearance

By "you can probably remove it" are you saying I should just close this pull request?

kairix-gh: Hey there - sorry I must have missed the notification on your message. Yep - you can close out the PR.

PR #1512, opened 2020-06-24
E6S FR fix for directions

mooondark:
We need a space in French because direction are separate words (Nord Est, Sud Ouest, and so on).

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e6s.js
+++ b/ui/raidboss/data/05-shb/raid/e6s.js
@@ -76,14 +76,14 @@
           safeZoneObj1 = {
             en: 'north',
             de: 'nord',
-            fr: 'nord',
+            fr: 'nord ',
             ko: '북',
           };
         } else if (currentHighestCombatant.PosY > 106 && currentHighestCombatant.PosY < 116) {
           safeZoneObj1 = {
             en: 'south',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e6s.js
```diff
@@ -76,14 +76,14 @@
           safeZoneObj1 = {
             en: 'north',
             de: 'nord',
-            fr: 'nord',
+            fr: 'nord ',
```
quisquous, line 79: Can you add the space only if necessary, maybe in the `concatSafeZones` function below? Otherwise if the same zone is north, then this will return `nord ` (with a trailing space).

mooondark: Sorry but I'm not good enough in prog. to do that. But if you tell me how to add a space or a "-" between Obj1&Obj2 I'll do it with great pleasure.

quisquous: Sorry for going ahead and just doing this myself, but I did this in ac0987a if you wanted to see how to do it.

mooondark: Np & Thanks a lot for the lesson :)

PR #1515, opened 2020-06-26
ci: Update Markdownlint Command to Work Universally

panicstevenson:
This was an (un)interesting issue. The original command:
```
node_modules/.bin/markdownlint **/*.md --ignore node_modules --ignore publish --ignore CactbotOverlay/ThirdParty
```
when run through the command line directly would not pick up root level changes to Markdown files. However, running the npm wrapper command:
```
npm run markdownlint
```
would pick it up.

However, neither the original command nor the npm wrapper command would work on my MacOS command line when attempting to find issues to files in the root directory. This updated command should work for all variations, Unix/Windows, command line directly, and npm wrapper command.


quisquous:
Thanks for the fix!

PR #1520, opened 2020-06-27
ci: Update GitHub Workflows

panicstevenson:
- Update names for steps within individual workflow jobs.
- Add problem matchers for markdownlint, pylint, and stylelint.

This should hopefully make error marking in pull requests more visible
and help with troubleshooting any failed action runs.

### DIFF
--- a/.github/matchers/markdownlint.json
+++ b/.github/matchers/markdownlint.json
@@ -1,0 +1,17 @@
+{
+  "problemMatcher": [
+    {
+      "owner": "markdownlint",
+      "pattern": [
+        {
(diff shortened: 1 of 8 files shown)


JLGarber:
Thanks for doing this. I'm not familiar enough with CI to point out where the changes to Github actions were confusing, but this looks like it should be a lot more understandable without having to chase down documentation elsewhere.

quisquous:
(For what it's worth, this won't help with black in particular, as its error is just "your files are not formatted" vs "you need a space between these operators in this file on this line".)

PR #1522, opened 2020-06-28
Util: Update make_timeline to restore FFLogs functionality

### LINKED ISSUE #1456: Util: Error in make_timeline when creating timeline from FFLogs
When attempting to create a timeline from an FFLogs report, the script fails due to lack of support for the `line_type` property on each entry.

```
python make_timeline.py -k [KEY] -r kvPbqZXchD7njwTV -rf 3
  File "make_timeline.py", line 400, in <module>
    print("\n".join(main(args)))
  File "make_timeline.py", line 201, in main
    if entry["line_type"] in ["21", "22"]:
KeyError: 'line_type'
```

This arises because I completely overlooked the FFLogs support when updating the script to allow for `targetable/untargetable` lines.

For my personal use, I just hardcoded all lines from FFLogs as being `21`, but I wonder whether it might be better to do something like
`if not entry["line_type"] or entry["line_type"] in ["21", "22"]`

The alternative would be to map from the `type` property in each FFLogs entry:

```
{ 
  'timestamp': 8794602,
  'type': 'cast',
  'sourceID': 2,
  'sourceIsFriendly': False,
  'targetID': 6,
  'targetIsFriendly': True,
  'ability': {
    'name': 'attack',
    'guid': 5092,
    'type': 1,
    'abilityIcon': '000000-000405.png'
  },
  'pin': '0' 
}
```

(Although I'm not sure whether anything other than 21/22 lines would be represented here anyway.)

JLGarber:
This should resolve #1456.

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -9,6 +9,31 @@ import timeline_aggregator
 import encounter_tools as e_tools
 
 
+"""FFLogs returns battle events in a list of dicts that looks something like this:
+    {
+      "timestamp": 4816719,
+      "type": "cast",
+      "sourceID": 113,
+      "sourceIsFriendly": false,
+      "targetID": 95,
+      "targetIsFriendly": true,
+      "ability": {
+        "name": "attack",
+        "guid": 870,
+        "type": 128,
+        "abilityIcon": "000000-000405.png"
+      },
+      "pin": "0"
+    },
+We map from the type property here to ACT network log line numbers.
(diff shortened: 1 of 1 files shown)


util/make_timeline.py
```diff
@@ -9,6 +9,31 @@
 import encounter_tools as e_tools
 
 
+"""FFLogs returns battle events in a list of dicts that looks something like this:
```
quisquous, line 12: This comment is great, thank you!!

PR #1523, opened 2020-06-28
Raidboss: dohn_mheg fr

VanSlanzar:
added missing translations fr
one is incorrect, i fix after.


quisquous:
Thanks! Sorry again about the husky troubles.

PR #1526, opened 2020-06-28
npm: Update Hooks to Only Affect Code Changes

panicstevenson:
Update Husky to run hooks more frequently, but to take a significantly smaller amount of time by only running on staged code changes.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,6 +16,8 @@
     "chai": "^4.2.0",
     "eslint": "^6.8.0",
     "eslint-config-google": "^0.9.1",
+    "husky": "^4.2.5",
+    "lint-staged": "^10.0.8",
     "markdownlint-cli": "^0.20.0",
     "stylelint": "^12.0.1",
     "stylelint-config-standard": "^19.0.0"
@@ -156,12 +158,13 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "node_modules/.bin/npm-run-all lint stylelint markdownlint test"
+      "pre-commit": "lint-staged"
     }
   },
-  "dependencies": {
-    "husky": "^4.2.5",
-    "markdownlint": "^0.18.0",
-    "npm-run-all": "^4.1.5"
+  "lint-staged": {
+    "*.css": "stylelint",
+    "*.js": "eslint",
+    "*.md": "markdownlint",
+    "*.py": "python -m pylint --errors-only"
   }
 }


package.json
```diff
@@ -156,12 +158,13 @@
   },
   "husky": {
     "hooks": {
-      "pre-push": "node_modules/.bin/npm-run-all lint stylelint markdownlint test"
+      "pre-commit": "lint-staged"
```
quisquous, line 161: What do you think about adding `"pre-push": "npm run test"`?

Also, I'm skeptical about wanting this on every commit, because I make a lot of temporary one-off commits that I don't want to bother formatting / testing / etc and eventually squash / revert / etc.

panicstevenson: > What do you think about adding "pre-push": "npm run test"?

This feels similar to what we already have existing, just without the linting?

> I'm skeptical about wanting this on every commit

I think that inherently [lint-staged](https://github.com/okonet/lint-staged) only can work with pre-commit, seeing as by the time you're pushing there's nothing staged within git.

My test output seems to confirm that:

```
husky > pre-push (node v10.16.3)
i No staged files found.
```

If we want, we can table this and find an alternative solution?

quisquous: > > What do you think about adding "pre-push": "npm run test"?
> 
> This feels similar to what we already have existing, just without the linting?

Do you think we should just run tests remotely?

> > I'm skeptical about wanting this on every commit
> 
> I think that inherently [lint-staged](https://github.com/okonet/lint-staged) only can work with pre-commit, seeing as by the time you're pushing there's nothing staged within git.

I seeeeeeeee.  Ok sure, let's give this a go.  I'll make a `git commit -a -m OOPS --skip-lint-staged-junk` alias.  :P

panicstevenson: > > > What do you think about adding "pre-push": "npm run test"?
> > 
> > 
> > This feels similar to what we already have existing, just without the linting?
> 
> Do you think we should just run tests remotely? (Or did you want to wait until they were per-file?)

Ideally people would still run tests themselves, but maybe forcing them is a little much? I guess we could add it as well, it'd be nice to somehow be like "only if you did X changes" such that a Markdownlint change doesn't require you to have npm and python. Theoretically we could add them back to the individual file you changed if we get it working per-file?

> > > I'm skeptical about wanting this on every commit
> > 
> > 
> > I think that inherently [lint-staged](https://github.com/okonet/lint-staged) only can work with pre-commit, seeing as by the time you're pushing there's nothing staged within git.
> 
> I seeeeeeeee. Ok sure, let's give this a go. I'll make a `git commit -a -m OOPS --skip-lint-staged-junk` alias. :P

The Husky `--no-verify` flag should still work here,

PR #1529, opened 2020-06-29
util: fix make_timeline errors with -it

quisquous:
Targetable/untargetable lines don't have "ability_name" or
"ability_id" on their entry, and so it's a key error to check for them.

e.g. `{'line_type': '34', 'time': datetime.datetime(2020, 6, 28, 18, 34, 10, 608000), 'combatant': 'Titan', 'targetable': '--untargetable--'}`

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -226,12 +226,23 @@ def main(args):
     output.append('0 "Start" sync /Engage!/ window 0,1')
 
     for entry in entries:
-
-        # Ignore targetable/untargetable while processing ignored entries
+        # Ignore targetable/untargetable while processing ability checks
         if entry["line_type"] in ["21", "22"]:
             # First up, check if it's an ignored entry
             # Ignore autos, probably need a better rule than this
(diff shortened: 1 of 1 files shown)


quisquous:
The other thought is maybe all events should fill in all keys with something None, at least?  I'm not sure if that would lead to incorrect conditionals somewhere else, though.

JLGarber:
Thanks for catching that. I'm genuinely unsure how I never saw a key error while testing, I threw (what I thought was) a good variety of logs at it!

I think you're right, it would be more robust to ensure all fields are accounted for all the time. It *probably* won't cause any breakage, based on a quick look over every occurrence of `if entry/entries`. Half of them are explicit comparisons to hardcoded values, and everywhere else is compared to argument lists, where we shouldn't see `None` in the first place. (Insert flippant comment about user error.)

quisquous:
What about this instead?

JLGarber:
Ahh, that's *very* good. It's possible we won't actually need to use it, but that's going to be very simple and clear if we ever do.

PR #1530, opened 2020-06-29
Raidboss: malikahs_well fr

VanSlanzar:
added missing translations fr


JLGarber:
Out of curiosity, is there a particular reason you did not move the KO translation block to the end? I ask because you fixed the order at line 42, but not in `timelineReplace`. It's not a big deal at all, but I think it would be fine to do alongside here unless quisquous has asked that we not do that generally.

quisquous:
> Out of curiosity, is there a particular reason you did not move the KO translation block to the end? I ask because you fixed the order at line 42, but not in `timelineReplace`. It's not a big deal at all, but I think it would be fine to do alongside here unless quisquous has asked that we not do that generally.

I think reordering is ok to do.  It makes diffs hard to read, but there is some documentation about recommended trigger order.  I think if we wanted to be more strict about it, we could add some custom eslint code that checks for object keys and their order.

I guess I'm saying I wouldn't go out of my way to reorder things elsewhere in a file, but cleaning up as you pass by something seems ok?

quisquous:
That said, I think these pull requests are all an improvement, so I'm going to merge them.  I think if there's some desire to go back and reorder `timelineReplace` blocks, that's easy enough to do separately.

VanSlanzar:
Hello, I did not rearrange the block, as I remember that @quisquous  asked that this be done on different requests for more understanding. So since this was just a french addition, it wouldn't have made sense to do that now. I think and I talk about it with @quisquous , make an alphabetical and coherent sorting in all the files, but later and on small packages, because I understood well that the more the package is big the more difficult it is.
@quisquous, @JLGarber

PR #1534, opened 2020-06-30
Name data files more consistently

quisquous:


docs/TimelineGuide.md
```diff
@@ -159,6 +159,21 @@ These are guidelines that cactbot tries to follow for timelines.
...
+separate with a hyphen.
+Dungeons with hard in the name can spell out "Hard" as a full word.
+Articles like `The` can be dropped.
+
+Examples to be consistent with:
+
+* The Grand Cosmos: `grand_cosmos`
+* Titan Extreme: `titan-ex`
+* Ruby Weapon Extreme: `ruby_weapon-ex`
+* The Great Gubal Library (Hard): `great_gubal_library_hard`
```
quisquous, line 175: :innocent:

quisquous:
@JLGarber @panicstevenson (and anybody else too!)

yes / no / hate it / don't care?

panicstevenson:
It feels a little confusing at first, but I think I can get behind it. I think it's a little bit confusing that we have trials named around the boss name, and raids and dungeons around the instance name. Not that I'm necessarily against that, or really any piece, but maybe that should be documented here as well? Additionally, we don't mention anything about tacking on an `s` for savage raids.

JLGarber:
I agree that from a "total correctness" standpoint, it's not so nice that the trials use a different naming scheme. However, I *really* don't like moving/renaming files unless we *really* have to, and the scheme quisquous lays out here would minimize the number of renamed files.

Apart from that, I'm fully behind the new scheme. Panic is correct that Normal/Savage versions need to be differentiated. (Having said that, Square is now certain to name the 6.0 raid series something that doesn't have a good colloquial community name.)

quisquous:
I think mostly I'm trying to more match "party finder" naming conventions.  It's "t6" not "Second Coil of Bahamut - Turn 1".  It's "Titan Ex", not "The Navel (Extreme)".  Dungeons seem to just get called by their name by everybody I know, but I'm not sure what else they would be named.

panicstevenson:
> but I'm not sure what else they would be named

I don't necessarily think they need to be named something else, just documented that they are named "[raid prefix character][turn #][potential s for savage]" if you're whipping up new documentation anyway.

Akurosia:
I like the given nameing schemas + i also agree with quisquous about the naming of e.g. t6 (since the abbreviated names are also used e.g. in Germany)

i would still have a question in regards to the edge case of turn 6-9: Technicly they should be t6n - t9n as they have also a savage version, but i think for consisteny with the others we wil llet them be just t6 - t9 and the savage versions t6s - t9s right?

Una-P:
In KO/JP/CN, trial and raids haven't instance name. It's literally 'War of ifrit' , 'Extreme war of hades' and raid is 'Eden's gate 1,2,3,4(savage)' 'Alexander maneuver 1,2,3,4'(OO of the father). So in our case, if it change to EN's instance name, it'll be hard to recognize at a glance.

quisquous:
Yeah, sounds like everybody is strongly agreed on "name things what people call them, not what zone it is".

I do think the mixed underscore/hyphen is a little weird, but it makes sense to me, and is mostly what was done already.  I could also see doing all underscores or all hyphens too if that seemed better?

quisquous:
> i would still have a question in regards to the edge case of turn 6-9: Technicly they should be t6n - t9n as they have also a savage version, but i think for consisteny with the others we wil llet them be just t6 - t9 and the savage versions t6s - t9s right?

Yeah.  I think nobody calls them "turn 6 normal" because they're not a normal mode in the same sense that like e6n is a normal mode.

PR #1536, opened 2020-07-01
Raidboss: twinning fr

VanSlanzar:
corrected 1 lines fr and added 1 lines fr.


quisquous:
I'm sorry to ask this, but can you not remove unused lines from the timeline?  It's a lot of work for me to check that this removing lines is correct or not.  I think @Akurosia was going to convert all of these at once and I can do one check for the whole PR like I did for the shadowbringers PR.

VanSlanzar:
Everything has been restored except the correction and the French add

quisquous:
Thank you! Sorry!

PR #1541, opened 2020-07-05
Util: Potentially fragile fix for key errors in make_timeline

### LINKED ISSUE #1538: Util: Make_timeline key failures when generating from FFLogs
For some encounters, when generating a timeline from FFLogs, `make_timeline.py` will fail with a key error:

```
  File "make_timeline.py", line 440, in <module>
    print("\n".join(main(args)))
  File "make_timeline.py", line 234, in main
    entries, start_time = parse_report(args)
  File "make_timeline.py", line 106, in parse_report
    "combatant": enemies[event["sourceID"]],
KeyError: 24
```

You can replicate this with:
`python make_timeline.py -k KEY -r z6m4AX1qMVLRFDHk -rf 3`
`python make_timeline.py -k KEY -r MjRzayFgmHZA3Bxk -rf 7`
`python make_timeline.py -k KEY -r J4vzBkLgT7wcPAGQ -rf 24`

In each case, it seems to fail on an event where the ability used is an untargeted AoE, rom-wide or not. It looks like the list of enemies generated here:

```
    for enemy in report_data["enemies"]:
        enemies[enemy["id"]] = enemy["name"]
```
may not be complete. That may be a bug in what FFLogs is returning? Could be worth poking Kihra over, I'm not sure.

JLGarber:
(This should resolve #1538)

This isn't the cleanest way to do it, but for the interim, until we update `make_timeline` to use V2 of the API, I think this is close enough. I ran a number of different encounters through it, specifically including the 3 failing ones I mentioned in the linked issue, and all of them went off successfully.

(I call this fix fragile because it just assumes that any events that are missing `sourceID` will include `event.source.id` and `event.source.name`. This is how it seems to work, but it is an assumption. I poked at explicitly checking for these elements a little better, but I didn't get anywhere.)


JLGarber:
Mmph. @panicstevenson, sorry to tag you in here, but I'm out of my depth here. [The closest thing I could find](https://bugzilla.redhat.com/show_bug.cgi?id=1710356) was from two years ago, but I'm not really sure whether this is a problem on my end or in the CI configuration.

panicstevenson:
Looks to be a CI issue, specifically [pylint](https://github.com/PyCQA/pylint/blob/2.2/pylint/__pkginfo__.py#L37) for 2.2.X has a dependency on isort that can take on any version above 4.2.5. [isort](https://pypi.org/project/isort/#history) released a new version recently, which appears to have breaking changes in pylint. This is probably obvious, since newer versions of pylint have [capped the version of isort to anything less than 5.X](https://github.com/PyCQA/pylint/blob/master/pylint/__pkginfo__.py#L41) and there is an open issue denoting the [same](https://github.com/PyCQA/pylint/issues/3722). I'm assuming that your system (as with mine) already has this requirement for isort established and has no reason to go out and grab a more recent version; however, it looks like on a fresh-install this is broken.

JLGarber:
Thanks for looking into that. Glad it's a relatively simple fix for now.

quisquous:
I think if you merge in the latest changes from master, these will now pass.

panicstevenson:
🎉

JLGarber:
Wonderful. Glad that's sorted.

quisquous:
Thanks for your patience!!!

PR #1542, opened 2020-07-05
ci: Pin isort Dependency

panicstevenson:
Pin CI system isort dependency until pylint is upgraded to a version where the dependency is properly capped. Addresses the dependency issues found in #1541.


panicstevenson:
@JLGarber @quisquous I feel like this is the best immediate move to get passed the dependency issues with pylint. There's probably zero harm in pinning pylint to a more recent version (and unpinning isort), but I'd like to guarantee that people are unblocked before moving forward with something that might have potential errors.

PR #1546, opened 2020-07-09
oopsy: warn when buffs don't hit the whole party

### LINKED ISSUE #1489: oopsy: warn when buffs don't hit the whole party
One thing oopsy could do better would be to show when heals / shields / offensive buffs miss people in the party.  Like when your BLM is standing WAY out there (again!!) and misses battle litany somehow.  Or you're pulling shiva and somebody missing wings and dies because they early pulled and took the first auto.  Or somebody is standing outside the range of your cure3 and dies to an aoe.

One UI possibility here is to just add more icons, like a dps (sword?) or heal (?) icon.  One other consideration is to add a warning type for abilities that give "damage down" (like Shiva puddles) and so that might fit with the dps icon as well as something to be warned about where somebody is doing less damage that pull than you would expect.

https://github.com/ngld/OverlayPlugin/issues/123 needs to get fixed before depending on too many more party things.  I think one workaround would be if your party ever hits 8 people, to consider all of those people in your party until you change zones even if you get a message saying one of them isn't in your party.

cc @JLGarber

quisquous:
Fixes #1489.

### DIFF
--- a/ui/oopsyraidsy/data/00-misc/general.js
+++ b/ui/oopsyraidsy/data/00-misc/general.js
@@ -1,5 +1,123 @@
 'use strict';
 
+// Abilities seem instant.
+let abilityCollectSeconds = 0.5;
+// Observation: up to ~1.2 seconds for a buff to roll through the party.
+let effectCollectSeconds = 2.0;
+
+// args: triggerId, netRegex, field, type, ignoreSelf
+let missedFunc = (args) => {
+  return {
+    id: args.triggerId,
+    netRegex: args.netRegex,
+    condition: function(evt, data, matches) {
+      return data.party.partyNames.includes(matches.source);
+    },
+    collectSeconds: args.collectSeconds,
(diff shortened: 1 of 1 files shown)


quisquous:
I would definitely take anybody who wanted to test this and give any feedback on correctness / usefulness.  It's definitely gone through quite a few testing iterations.  <_<

Also:
* any missing abilities I should add?
* should this get its own file instead of general.js?
* any other feedback?

JLGarber:
Sorry it's been so long on this, but it does appear to work as expected. Definitely a good decision to cap the listed missed players at 3 though, even at that there were one or two situations where the last player's name was clipped on the right side.

I think it's probably safe to merge as it is, but it might be nice to have a configuration option to disable if that's at all possible.

quisquous:
> I think it's probably safe to merge as it is, but it might be nice to have a configuration option to disable if that's at all possible.

Folks can disable them by trigger ids if they really don't like them, which seems sufficient to me?

PR #1548, opened 2020-07-10
ci: Unpin Unexecuted Dependencies

panicstevenson:
PyYAML and requests dependencies aren't ever directly executed and are
only pulled within CI systems to avoid import errors in pylint. Keeping
pinned dependencies for Python packages executed directly to avoid
randomly picking up breaking changes.


quisquous:
(Also I wonder if workflow changes should trigger actions?)

panicstevenson:
I had added that myself as well, was going to ask your input on that

PR #1550, opened 2020-07-11
ci: Update pylint Version

panicstevenson:
Update pylint version installed and remove temporary isort dependency
pin.


quisquous:
Thanks!!

PR #1552, opened 2020-07-11
Raidboss: Sohm Al (hard) dungeon package

JLGarber:
Mostly standard, but unfortunately we can't do a timeline for boss 2. Gowrow works slightly differently from any other boss I've come across. It's all "count to 3":

`Auto, auto, special`

Unfortunately, the special attacks all have different timers, both in terms of when they resolve and in how long the animation blocks auto-attacks thereafter. This is *further* dependent on whether or not Gowrow has self-buffed.

All of this is to say that a timeline would not only be difficult, but also largely useless. The best we could really hope for is to sync to Volcanic Howl every time it comes up in the rotation, but that wouldn't gain us much. The most we can tell the user is "okay, a special of some kind is coming up". That's honestly not worth it, at least for now.

Separate from that, we're missing the Korean translations because I don't have that scripted nicely as I do the others. (API access is *so* much nicer than dealing directly with CSV, but CSV is the only option I know of for Korean right now.) I'm working through that, and if this hasn't been fixed by then I'll come back and take care of it.


ui/oopsyraidsy/data/03-hw/dungeon/sohm_al_hard.js
```diff
@@ -0,0 +1,44 @@
...
+    'Molten Silk 3': '1C42', // Ring AoE, Lava Scorpion, boss 3
+    'Realm Shaker': '1C41', // Circle AoE, Lava Scorpion, boss 3
+  },
+  triggers: [
+    {
+      // Warns if players step into the lava puddles. There is unfortunately no direct damage event.
+      id: 'Sohm Al Hard Burns',
+      netRegex: NetRegexes.gainsEffect({ effectId: '11C' }),
+      mistake: function(e) {
+        return { type: 'warn', blame: e.targetName, text: e.abilityName };
```
quisquous, line 40: Just to save future work, can you use `matches` here instead of `e`? It's `mistake: function(e, data, matches)`.

ui/raidboss/data/03-hw/dungeon/sohm_al_hard.js
```diff
@@ -0,0 +1,314 @@
...
+      netRegexJa: NetRegexes.startsUsing({ id: '1C32', source: 'スポアサック', capture: false }),
+      suppressSeconds: 5,
+      infoText: {
+        en: 'Away from large pod',
+      },
+    },
+    {
+      // The actual effect being checked here is Heavy.
+      id: 'Sohm Al Hard Excretion',
+      netRegex: NetRegexes.gainsEffect({ effectId: '0e' }),
```
quisquous, line 50: For consistency with other ids, can you make this `0E`.  Names might be capitalized inconsistently, but ids are capital everywhere else.

quisquous:
Re: Gowrow.  It seems very similar to the cyclops boss in Amanesis Anyder.  Aren't the spawning of the explosive adds at a regular time, even if the individual attacks were not?

ui/oopsyraidsy/data/03-hw/dungeon/sohm_al_hard.js
JLGarber: Fair point, I'll be more careful about that going forward.

JLGarber:
Leightonward is the boss with the spawning adds. Gowrow is "will the melee remember to get out and not get hit with the healing debuff THIS time?"

quisquous:
> Leightonward is the boss with the spawning adds. Gowrow is "will the melee remember to get out and not get hit with the healing debuff THIS time?"

:innocent:

I thought there were spore pods and explosions? Maybe they're not targetable and calling them adds is a bit much.

JLGarber:
No, Leightonward, the first boss, is the one that has those, hence the whole thing with me forgetting the syntax for "Spore Sac" vs "Small Spore Sac". Gowrow has only itself.

quisquous:
Haha, sorry, I apparently totally forgot how this dungeon works.  I watched a video, and I think this makes total sense to not have a timeline here.  Sorry for my bad memory, it's been a while.

PR #1554, opened 2020-07-11
Missing raidboss translations german (part1)

Akurosia:


ui/raidboss/data/00-misc/test.js
```diff
@@ -230,7 +230,7 @@
         'cactbot lang': 'cactbot sprache',
         'cactbot test response': 'cactbot test antwort',
         'You clap for the striking dummy': 'Du klatschst begeistert Beifall für die Trainingspuppe',
-        'You psych yourself up alongside the striking dummy.*': 'Du willst wahren Kampfgeist in der Trainingspuppe entfachen',
+        'You psych yourself up alongside the striking dummy\\.\\*': 'Du willst wahren Kampfgeist in der Trainingspuppe entfachen',
```
quisquous, line 233: I suspect that both of these are incorrect.  Does this trigger work? My suspicion is that you should just make it `You psych yourself up alongside the striking dummy` so that the `.*` is left unreplaced.

Akurosia: `'You psych yourself up alongside the striking dummy.*.*?'` should this be only `'You psych yourself up alongside the striking dummy.*?'` in ` id: 'Test Poke',`?

quisquous: Ah yeah, that's why it would work, if it was working.  It should be exactly as you wrote it, with just one `.*?`

Akurosia: done

quisquous:
Thanks!

PR #1555, opened 2020-07-11
small fixes in raidboss

Akurosia:
common replacement change is in logmessages:2014

### DIFF
--- a/ui/raidboss/common_replacement.js
+++ b/ui/raidboss/common_replacement.js
@@ -10,7 +10,7 @@ let commonReplacement = {
     ko: '15초 후에 $1(이|가) 봉쇄됩니다',
   },
   'is no longer sealed': {
-    de: 'öffnet sich erneut',
+    de: 'öffnet sich wieder',
     fr: 'Ouverture ',
     ja: 'の封鎖が解かれた',
     cn: '的封锁解除了',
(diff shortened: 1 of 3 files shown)


ui/raidboss/common_replacement.js
```diff
@@ -10,7 +10,7 @@ let commonReplacement = {
     ko: '15초 후에 $1(이|가) 봉쇄됩니다',
   },
   'is no longer sealed': {
-    de: 'öffnet sich erneut',
+    de: 'öffnet sich wieder',
```
quisquous, line 13: What about both?

Akurosia: i can make it both tbh but in logmessages:2014 it is `wieder`. but will put both in just to be sure

ui/raidboss/common_replacement.js
```diff
@@ -10,7 +10,7 @@ let commonReplacement = {
     ko: '15초 후에 $1(이|가) 봉쇄됩니다',
   },
   'is no longer sealed': {
-    de: 'öffnet sich erneut',
+    de: 'öffnet sich (wieder|erneut)',
```
quisquous, line 13: Sorry, one more thing, but this should be `(?:wieder|erneut)`

PR #1557, opened 2020-07-11
Missing german translations for hw

Akurosia:


ui/raidboss/data/03-hw/raid/a11s.js
```diff
@@ -366,6 +366,58 @@
     },
   ],
   timelineReplace: [
+    {
+      'locale': 'de',
+      'missingTranslations': true,
```
quisquous, line 371: Are there still missing translations, here, below, and in other files you're modifying?

Akurosia: shoot... forgot to remove the missing flag in the HW files, give me a sec

Akurosia: ok it was only a11s missing from what i saw (for ravana i left it, as i need to clean this one up in a seperate PR before removing the flag)

PR #1558, opened 2020-07-12
raidboss: convert Regexes to NetRegexes

quisquous:
Converted via:
`git ls-files ui/raidboss/data/ | xargs -i sed -i "s/\(^\s\)*regex\(..\)\?: Regexes.\(\(he\|[^h]\)\w*\)({/\1netRegex\2: NetRegexes.\3({/" {}`

Currently, there's no NetRegexes.hasHP, and so those are left behind.
Maybe we can figure out some better way to handle them.

This will nearly finish off raidboss for #1328.


quisquous:
I'm going to do some local testing with this tiny `114 files changed, 7848 insertions(+), 7848 deletions(-)` patch, but will plan to land after closer review and testing.  Don't worry about conflicts, as this was generated with a script.

PR #1559, opened 2020-07-12
add missing translations and removes missing flag for germany

Akurosia:
all german translations should be done by now.
translation section still might contain not used data or douplicate entries.
After this is done, i will see what i can convert for other languages


ui/raidboss/data/03-hw/raid/a7s.js
```diff
@@ -175,4 +175,118 @@
       },
     },
   ],
+  timelineReplace: [
+    {
+      'locale': 'de',
+      'missingTranslations': true,
```
quisquous, line 181: Are there missing translations here? Maybe `find_missing_translations` should warn about any of these too.

ui/raidboss/data/03-hw/trial/ravana-ex.txt
```diff
@@ -12,7 +12,7 @@ hideall "--sync--"
 # Any displayed phase lengths are the maximum that could occur with no "extra" Bloodlust.
 
 0 "Start" sync /Engage!/ window 0,1
-1.2 "--sync--" sync /:Ravana/13E4:/ window 1.2,1
+1.2 "--sync--" sync /:Ravana:13E4:/ window 1.2,1
```
quisquous, line 15: !!!

ui/raidboss/data/03-hw/raid/a7s.js
Akurosia: i thought i checked all the flags and removed them x.x

Akurosia: ohh now i know, 
i left it in just for a7s, as it contains old log lines e.g. `suffers the effect of Pyretic` or even an empty elmenet `no translation for replaceText ''`

quisquous: Oh yeah, I marked a7s as `timelineNeedsFixing: true` and so those won't show up in `find_missing_timelines`.  I forgot that this is an old one.

PR #1560, opened 2020-07-12
Adds Oospyraidsy + ui

VanSlanzar:
adds missing translations fr.
adds missing regex de and fr for e4s.
removed fixme lines.
The lines added in French which are not translated is because we use the English words (early pull, late pull, party wipe).


quisquous:
Thanks for the followup!

PR #1567, opened 2020-07-14
oopsy: Added gainsEffect helper

JLGarber:
In line with #1504 , this adds support for `gainsEffect` mistakes without requiring dedicated triggers in a given instance's oopsy file. This is support only, with no changes to existing instance oopsy files.

Note that this helper uses the `matches` parameter rather than `event` in the returned `mistake` object. This is primarily because I don't have a firm enough grasp of `event` to have it work correctly with `netRegex` in the trigger body. This works, and is what we want to move toward in future anyway, so I feel that it's relatively acceptable here even if it doesn't exactly match the surrounding helpers.

I'm rather unhappy with 
```
      this.AddSimpleTriggers('warn', set.damageWarn);
      this.AddSimpleTriggers('fail', set.damageFail);
      this.AddGainsEffectTriggers('warn', set.gainsEffectWarn);
      this.AddGainsEffectTriggers('fail', set.gainsEffectFail);
      this.AddShareTriggers('warn', set.shareWarn);
      this.AddShareTriggers('fail', set.shareFail);
```
but realistically rewriting it to be "less bad" wouldn't save us much in terms of space or readability. Suggestions welcome!

### DIFF
--- a/ui/oopsyraidsy/oopsyraidsy.js
+++ b/ui/oopsyraidsy/oopsyraidsy.js
@@ -964,6 +964,23 @@ class DamageTracker {
     }
   }
 
+  AddGainsEffectTriggers(type, dict) {
+    if (!dict)
+      return;
+    let keys = Object.keys(dict);
+    for (let key of keys) {
+      let id = dict[key];
+      let trigger = {
+        id: key,
+        netRegex: NetRegexes.gainsEffect({ effectId: id }),
+        mistake: function(e, data, matches) {
+          return { type: type, blame: matches.target, text: matches.effect };
+        },
+      };
+      this.netTriggers.push(trigger);
+    }
+  }
+
   // Helper function for "double tap" shares where multiple players share
   // damage when it should only be on one person, such as a spread mechanic.
   AddShareTriggers(type, dict) {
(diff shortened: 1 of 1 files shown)


ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -964,6 +964,23 @@ class DamageTracker {
...
+    if (!dict)
+      return;
+    let keys = Object.keys(dict);
+    for (let key of keys) {
+      let id = dict[key];
+      let trigger = {
+        id: key,
+        netRegex: NetRegexes.gainsEffect({ effectId: id }),
+        mistake: function(e, data, matches) {
+          return { type: type, blame: matches.target, text: matches.effect };
```
quisquous, line 977: I think this is exactly correct.  `effect` is not filled out for `netRegex`, only for `damageRegex` / `healRegex`, so using `matches` is the right thing here.  For `netRegex`, the only thing in `event` is `line`: https://github.com/quisquous/cactbot/blob/master/ui/oopsyraidsy/oopsyraidsy.js#L647

I think we'd need to switch those special regexes over to using `netRegex` so that matches was filled out more fully.

This `EventFromFields` is a duplicate way of doing what `Regexes` does: https://github.com/quisquous/cactbot/blob/875b50b8110922c1accf8b9b7c66dce15f4436b1/ui/oopsyraidsy/oopsyraidsy.js#L806-L829.  It just predates `Regexes` (oops) and so needs to get rolled into it.

PR #1569, opened 2020-07-17
Docs: More accurate description for type 25 log lines

JLGarber:
Ravahn happened to give us some information on the `25` log line that was more accurate than what was currently noted in the guide. I cleaned it up and spliced it in as best I could.

The [] brackets indicate changes in the quote, much as you would expect to see in a news story. You can compare it with Ravahn's [original exact quote](https://discordapp.com/channels/551474815727304704/551476873717088279/733336512443187231):

>line 37 is effectively a 'sync' packet where the server tells the client that the previous action should now be rendered on the ui / chatlog.
or something along those lines, thats how i think about it.  others may have a better understanding.
example - if I cast a spell, i will get an effectresult packet (line type 21/22) showing the damage amount, but the target isnt expected to actually take that damage yet.  the line 37 has a unique identifier in it which refers back to the 21/22 line and indicates that the damage should now take effect on the target.
FFXIV plugin doesnt use these lines currently, they are used by FFLogs.  it would help though if I did, but ACT doesnt do multi-line parsing very easily so I would need to do a lot of work-arounds.

If that's not stylistically appropriate for the guide, I'm okay taking it out, but I'm leaving it all in now for accuracy reasons.

### DIFF
--- a/docs/LogGuide.md
+++ b/docs/LogGuide.md
@@ -64,7 +64,7 @@ With updates for:
   * [22: NetworkNameToggle](#22-networknametoggle)
   * [23: NetworkTether](#23-networktether)
   * [24: LimitBreak](#24-limitbreak)
-  * [25: NetworkEffectResult](#25-networkeffectresult)
+  * [25: NetworkActionSync](#25-NetworkActionSync)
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for adding this in!

PR #1570, opened 2020-07-17
npm: Update Lint-Staged Command to Run Via npx

panicstevenson:
Changes `lint-staged` to run via `npx` rather than being invoked
directly. This change sets up the PATH for various programs like VSCode
or GitHub Desktop that might be running into issues where `lint-staged`
is attempting to be directly executed, outside of Node.


quisquous:
Thank you! Would you mind adding a note to the install or contributing docs somewhere about global for speed?

JLGarber:
Much appreciated!

PR #1585, opened 2020-07-20
raidboss: cn-5.2 with netRegex

Bluefissure:
This is the timeline/trigger update of `raidboss` for cn patch 5.2.
The [FFXIVProcessCn](https://github.com/quisquous/cactbot/blob/master/plugin/CactbotEventSource/FFXIVProcessCn.cs) will need to be updated as well, which should be tested in-game.
I'll update it asap after the patch release.


quisquous:
Thank you!!

PR #1589, opened 2020-07-21
plugin: cn 5.2 memory & data structure

Bluefissure:
I just copy&paste from `FFXIVProcessIntl` and it works well.
Since the target info is from OverlayPlugin's Enmity stuff, I made https://github.com/ngld/OverlayPlugin/pull/141 to ngld's OverlayPlugin.

### DIFF
--- a/plugin/CactbotEventSource/FFXIVProcessCn.cs
+++ b/plugin/CactbotEventSource/FFXIVProcessCn.cs
@@ -3,759 +3,672 @@ using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using Newtonsoft.Json.Linq;
 
-namespace Cactbot
-{
-    public class FFXIVProcessCn : FFXIVProcess
-    {
-        // Last updated for FFXIV 5.2
+namespace Cactbot {
+  public class FFXIVProcessCn : FFXIVProcess {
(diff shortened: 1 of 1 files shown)


plugin/CactbotEventSource/FFXIVProcessCn.cs
```diff
@@ -3,677 +3,759 @@
...
-    // for FFXIV CN version: 5.1
-    //
-    // Latest CN version can be found at:
-    // http://ff.sdo.com/web8/index.html#/patchnote
-    //
-    [StructLayout(LayoutKind.Explicit)]
-    public unsafe struct EntityMemory {
-      public static int Size => Marshal.SizeOf(typeof(EntityMemory));
+namespace Cactbot
+{
```
quisquous, line 7: Hmm, why are you reformatting this at the same time? If this is the same as FFXIVProcessIntl, then it should be the identical file other than FFXIVProcessCn vs FFXIVProcessIntl, if you don't mind.

Bluefissure: Visual Studio automatically formatted it for me and I didn't notice that ><

quisquous:
Thanks! I'll do another release with this.

PR #1593, opened 2020-07-22
Update FateWatcher for CN 5.2 release

KattTails:
Untested but from the looks of things it should match the intl version.


quisquous:
Oops, thanks for catching this.

PR #1597, opened 2020-07-24
CI: build cactbot with github actions

MaikoTan:
This Github Actions workflow would:
- Build cactbot on every `push` or `pull request`
- Generate artifact with `./util/publish.sh`
- Upload artifact

It would make us easy to test cactbot, without download source code and build it locally.

Note: I am not familiar with `MSBuild`, but it runs well in my repo. If it lacks anything, please tell me.


panicstevenson:
Heh this looks similar:
https://github.com/panicstevenson/cactbot/blob/release-artifact-actions/.github/workflows/release.yml
😉

quisquous:
This is great!!

A bunch of questions:
* would it make sense to add a `paths:` restriction here to only run on changes that touch `plugin/` or this file?
* does it make sense to zip up the artifact in the same way that @panicstevenson's does? or, alternatively to only consider the built DLLs as the artifact? are there space limits on uploaded artifacts? I don't know anything about this.
* what happens when there's a build error?
* Is it possible to add a custom problem matcher like markdown-lint.yml does to surface errors and make the pull request / push turn red? (mostly curious, I'm happy to land this now, and think about this question later)

quisquous:
From https://docs.github.com/en/actions/configuring-and-managing-workflows/persisting-workflow-data-using-artifacts, it looks like:
* Storing artifacts uses storage space on GitHub. GitHub Actions usage is free for public repositories.
* Artifacts automatically expire after 90 days, but you can always reclaim used GitHub Actions storage by deleting artifacts before they expire on GitHub.

So maybe there's really no size limit here that matters.

MaikoTan:
> Heh this looks similar:
> https://github.com/panicstevenson/cactbot/blob/release-artifact-actions/.github/workflows/release.yml

Wow, it looks like yours is more complicated and robust, maybe we can merge it into one file in this PR?

> would it make sense to add a paths: restriction here to only run on changes that touch plugin/ or this file?

Well, in my opinion, I would let it build on when every code changes. So when we want to test someone's code, we can only download the artifact and unzip, instead of `vs build`, `git pull` and `copy-paste`.

> what happens when there's a build error?

Ouch, I don't know.
Maybe failed in workflow?

> Is it possible to add a custom problem matcher like markdown-lint.yml does to annotate lines with errors? (mostly curious, I'm happy to land this now, and think about this question later)

If `MSBuild` would output all the problem in `stderr`, it is not hard to do this.
But as I said above, I am not familiar with it, so I have to spend time to study `MSBuild` and Github workflows.

MaikoTan:
> does it make sense to zip up the artifact in the same way that @panicstevenson's does? or, alternatively to only consider the built DLLs as the artifact? are there space limits on uploaded artifacts? I don't know anything about this.

When specify the folder as artifact, github actions would automatically archive it as a zip file. So I don't think necessary to zip it manually.

panicstevenson:
As an additional piece to remember, if you're building on every commit you will be causing builds in each contributor's fork each time they commit to any branch on their fork.

panicstevenson:
> > does it make sense to zip up the artifact in the same way that @panicstevenson's does? or, alternatively to only consider the built DLLs as the artifact? are there space limits on uploaded artifacts? I don't know anything about this.
> 
> When specify the folder as artifact, github actions would automatically archive it as a zip file. So I don't think necessary to zip it manually.

Yep, this is for renaming it to match the cactbot-v{tag} format. I don't know if it's necessary here assuming you have all the necessary files in that archive.

MaikoTan:
> As an additional piece to remember, if you're building on every commit you will be causing builds in each contributor's fork each time they commit to any branch on their fork.

Yes, you're right.
So we should add an condition to limit it only run in this current repo?

MaikoTan:
> Yep, this is for renaming it to match the cactbot-v{tag} format. I don't know if it's necessary here assuming you have all the necessary files in that archive.

Actually you can specify the name with..

```yaml
uses: actions/upload-artifact@v2
with:
  name: your-fancy-name #
  path: ...
```

`with.name` property.

panicstevenson:
> > As an additional piece to remember, if you're building on every commit you will be causing builds in each contributor's fork each time they commit to any branch on their fork.
> 
> Yes, you're right.
> So we should add an condition to limit it only run in this current repo?

I had previously done:
https://github.com/quisquous/cactbot/blob/e52dc1979d74b0476b1e4b8b975667c6ac7568a9/.github/workflows/css-lint.yml#L3-L15

Which still builds off forks, but not as frequently.

I had done the repository check because I plan on updating my repository check to ensure that we don't create any tagged releases for any repo that's not the main one.

panicstevenson:
> Actually you can specify the name with..

True! I think I was doing that because I'm doing `actions/upload-release-asset@v1`, not `actions/upload-artifact@v2`, but maybe I should double check if it's necessary on my end.

quisquous:
> > would it make sense to add a paths: restriction here to only run on changes that touch plugin/ or this file?
> 
> Well, in my opinion, I would let it build on when every code changes. So when we want to test someone's code, we can only download the artifact and unzip, instead of `vs build`, `git pull` and `copy-paste`.

I am not sure I understand your argument here.  I am suggesting to only run this build step if plugin files change.  If somebody has a pull request that touches the plugin, then you would be able to download its artifact and use it.

For most pull requests that don't touch the plugin code, you can make a local branch for the pull request via `git fetch origin pull/ID/head:BRANCHNAME` and use your existing plugin.  I don't think there's any need for a built plugin in that case more than the one you are already using locally.

Am I not understanding something here?

> > what happens when there's a build error?
> 
> Ouch, I don't know.
> Maybe failed in workflow?

Is it possible to test this in your fork before landing this and see what happens?

MaikoTan:
So maybe we can split the building and packaging in two workflows?
Only `plugin/` changes call building workflow, and packaging workflow runs on every push and pull request.

> Is it possible to test this in your fork before landing this and see what happens?

Sure, of cource.
I would try it later.

quisquous:
> So maybe we can split the building and packaging in two workflows?
> Only `plugin/` changes call building workflow, and packaging workflow runs on every push and pull request.

Sorry to be opinionated here, but I think only plugin/ changes should both build and package.  I think it's a waste of resources to build the plugin and upload the repository for every commit, although arguably those resources are "free".  I think it's also confusing, to have an artifact for something that doesn't need rebuilding the plugin.  I think generally it's just good practice to have workflow steps only apply to requests that need to trigger them.

Is it burdensome to download a pull request that does not require building the plugin?  There are already `You can also open this in GitHub Desktop or view command line instructions` links on every pull request.  You can also download the repository at any commit via going to the commit, e.g. https://github.com/quisquous/cactbot/tree/2e22b5fef2b49c0289c6780b4f11359d1274c783 and then clicking Code -> Download Zip.

I think if this is something that would be useful for you personally, you could commit some yaml to your own cactbot fork such that it runs on every commit, but in this pull request only have it run in the main repository on changes to plugin and the workflow itself.

panicstevenson:
Additionally to the above; if you haven't already, you can setup GitHub pages to publish on your fork and any changes you make to your fork's master branch can be accessed via the remote URL, allowing you to use:
https://quisquous.github.io/cactbot/ui/raidboss/raidboss.html
or
https://panicstevenson.github.io/cactbot/ui/raidboss/raidboss.html
or
https://maikotan.github.io/cactbot/ui/raidboss/raidboss.html
Which I think adds to the emphasis that non-core engine changes don't _necessarily_ need to be built, packaged, and released.

MaikoTan:
Alright. I should admit that you are correct on this issue.
At the beginning, I wanted to let someone who doesn't know how to build a plugin with VS, also be able to download the pre-release version file, and simply unzip to use & test.
But you're right. It is no need to build a non-modified plugin. I would change the trigger as you said.

However, I got trouble when writting a matcher to detect build error.

https://github.com/MaikoTan/cactbot/blob/32447cb32328cbab1875cba021f4587599ed0ce4/.github/matchers/msbuild.json#L7-12

and the actions output:

https://github.com/MaikoTan/cactbot/actions/runs/180808163

I don't know why doesn't it match the file name. Am I wrong?

MaikoTan:
> Additionally to the above; if you haven't already, you can setup GitHub pages to publish on your fork and any changes you make to your fork's master branch can be accessed via the remote URL, allowing you to use:

Excellent! I have not thought about it.
Thank you!

panicstevenson:
> > Additionally to the above; if you haven't already, you can setup GitHub pages to publish on your fork and any changes you make to your fork's master branch can be accessed via the remote URL, allowing you to use:
> 
> Excellent! I have not thought about it.
> Thank you!

Sounds like something we can look to add in the Contribution guides.

> I don't know why doesn't it match the file name. Am I wrong?

I think the regex is `^##\[(.*)\](.*)\((\d*),(\d*)\):\s\w*\s(.*):\s(.*)$`; you'd have to translate that to be okay with a string, and the problem matcher fields would be:
```
"severity": "1",
"file": "2"
"line": "3",
"column": "4",
"code": "5".
"message": "6",
```

MaikoTan:
> I think the regex is `^##\[(.*)\](.*)\((\d*),(\d*)\):\s\w*\s(.*):\s(.*)$`; you'd have to translate that to be okay with a string, and the problem matcher fields would be:
> 
> ```
> "severity": "1",
> "file": "2"
> "line": "3",
> "column": "4",
> "code": "5".
> "message": "6",
> ```

I do not very understand the mechanic for now, but I copied the file content from [official dotnet action](https://github.com/actions/setup-dotnet/blob/master/.github/csc.json), which runs well in my repo.

MaikoTan:
- Add 8-bytes SHA in zip file name
image
Looks well.

.github/workflows/build-artifact.yml
```diff
@@ -0,0 +1,59 @@
+name: Build artifact
+
+on:
+  push:
+    paths:
+      - 'plugin/**'
```
panicstevenson, line 6: Can we add
```'.github/workflows/build-artifact.yml'```
to this list and the pull_request paths list as well?

File: .github/workflows/build-artifact.yml [hunk 2]
```diff
@@ -0,0 +1,59 @@
...
+    - name: Fetch dependencies
+      run: |
+        python .\util\fetch_deps.py
+
+    - name: Set up msbuild
+      uses: microsoft/setup-msbuild@v1.0.1
+
+    - name: Add Custom Problem Matcher
+      run: |
+        echo "::add-matcher::.github/matchers/msbuild.json"
```
panicstevenson, line 45: Can this be:
```
echo "::add-matcher::.github/matchers/${{ github.job }}.json"
```

.github/workflows/build-artifact.yml
```diff
@@ -0,0 +1,59 @@
...
+      - reopened
+      - synchronize
+    paths:
+      - 'plugin/**'
+
+jobs:
+  build:
+    runs-on: windows-latest
+
+    strategy:
```
panicstevenson, line 19: I don't necessarily think it matters, but I don't think you need this to have a testing matrix; I think you could specify the job to run with setup-python at your preferred version and run that way. I don't foresee a scenario where we will test the MSBuild against multiple versions of Python

quisquous:
Yay, build errors: https://github.com/MaikoTan/cactbot/pull/1/files

.github/workflows/build-artifact.yml
quisquous, line 6: Can you also include these files in `paths`:
* `util/publish.sh`
* `.github/workflows/build-artifact.yml`
* `util/DEPS.py`
* `util/fetch_deps.py`

.github/workflows/build-artifact.yml
MaikoTan: I see.
So this option is used for testing python program in different python environments, since this project only needs one runnable python environment, we don't need a matrix.

.github/workflows/build-artifact.yml
MaikoTan: Sure!
Then we should change the job name as `msbuild` too?

.github/workflows/build-artifact.yml
MaikoTan: Okay!

MaikoTan: Okay!

.github/workflows/build-artifact.yml
panicstevenson: Sure; either change the name to msbuild within the job or change the problem matcher to build; I think both options are viable!

panicstevenson:
> > I think the regex is `^##\[(.*)\](.*)\((\d*),(\d*)\):\s\w*\s(.*):\s(.*)$`; you'd have to translate that to be okay with a string, and the problem matcher fields would be:
> > ```
> > "severity": "1",
> > "file": "2"
> > "line": "3",
> > "column": "4",
> > "code": "5".
> > "message": "6",
> > ```
> 
> Thank you for pointing out.
> I do not very understand the mechanic for now, but I copied the file content from [official dotnet action](https://github.com/actions/setup-dotnet/blob/master/.github/csc.json), which runs well in my repo.

Here's the "official" documentation, if you're interested but I think what you have works!
https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

MaikoTan:
> Here's the "official" documentation if you're interested, but I think what you have works!
> https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md

Aha, here it is!
I googled before but I found nothing. Thanks for providing!

quisquous:
Thank you for all of this! This is really great!

PR #1600, opened 2020-07-25
oopsy: add blue mage heal skills to missed heal tracker

Bing-su:

PR #1601, opened 2020-07-25
raidboss: a1s ~ a12s korean

Bing-su:


quisquous:
It looks like you removed `missingTranslations` but there are still some additional missing translations for some of these fights when you run `npm test`.

Bing-su:
I think there is same problem occurred in #1600. Running `npm test` locally doesn't make any error.

Cap 2020-07-26 09-25-16-022

quisquous:
Ah thanks, it looks like this runs cleanly locally for me too.  Sorry for all of the hassle!

PR #1602, opened 2020-07-25
CI: add dependencies cache in build-artifact.yml

MaikoTan:
- Add dependencies (`plugin/ThirdParty/` directory) cache to accelerate building speed.

image

_According to the [official doc](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows), cache files would be kept up to 7 days if no any change._

I think it is useful when a changing plugin code frequently.


.github/workflows/build-artifact.yml
```diff
@@ -33,12 +33,24 @@ jobs:
         sha=$(echo ${{ github.sha }} | cut -c 1-8)
         echo "::set-env name=artifact_sha::$sha"
 
+    - name: Check dependencies cache
+      id: cache-dependencies
+      uses: actions/cache@v2
+      with:
+        path: ./plugin/ThirdParty
+        key: ${{ runner.os }}-cactbot-${{ hashFiles('./util/fetch_deps.py', './util/DEPS.py') }}
+        restore-keys: |
```
quisquous, line 42: I'm not quite sure I understand what `restore-keys` is doing here.  Can you explain?

.github/workflows/build-artifact.yml
```diff
@@ -33,12 +33,24 @@ jobs:
         sha=$(echo ${{ github.sha }} | cut -c 1-8)
         echo "::set-env name=artifact_sha::$sha"
 
+    - name: Check dependencies cache
+      id: cache-dependencies
+      uses: actions/cache@v2
+      with:
+        path: ./plugin/ThirdParty
+        key: ${{ runner.os }}-cactbot-${{ hashFiles('./util/fetch_deps.py', './util/DEPS.py') }}
```
panicstevenson, line 41: `runner.os` will always resolve to `windows`, right? Doesn't feel like we need it.

.github/workflows/build-artifact.yml
panicstevenson, line 42: Looking at the [docs](https://docs.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#example-using-multiple-restore-keys) they include the `${{ hashFiles }}` in their restore keys as the first hit. I haven't used this feature before, but is there any reason for us not to do the same?

MaikoTan: When cache `key` fully match the `key`property here, then its `outputs.cache-hit` would be `true`, so step "setup python" and "fetch dependency" would be skipped.

When no any cache `key` fully match the `key` property, then its `outputs.cache-hit` would be always `false`. But the most recent cache would be recovered if any `restore-keys` match its prefix.

Maybe this feature is not quite useful for `plugin` building here, but as for `npm` or `pip`, even the cache key was not exactly the same, but there might be some libraries already in that cache, so when running `npm install` or `pip install`, those libraries would be skipped, a little faster, right?

.github/workflows/build-artifact.yml
MaikoTan: Yes, but I think it remain `runner.os` would not bad.
Maybe in the future, cactbot can be built up on `ubuntu` or other os, then this key should be re-usable?

.github/workflows/build-artifact.yml
MaikoTan: I think they write `hashFiles` in `restore-keys` for explaining the mechanic of cache matching.
As far as I saw in the [official examples](https://github.com/actions/cache/blob/main/examples.md) they don't do that.

MaikoTan: For example in [this run](https://github.com/MaikoTan/cactbot/runs/909896093?check_suite_focus=true) it matched prefix only.

panicstevenson: Interesting; the docs say they also pick the latest on everything that matches, so maybe they're relying on that mechanic. Nevermind!

quisquous: I get it! Thanks for the example.

PR #1606, opened 2020-07-25
ci: Add Workflow to Update gh-pages From main

panicstevenson:
Add GitHub Actions workflow to update gh-pages with any commits added to
main.

### DIFF
--- a/.github/workflows/update-gh-pages.yml
+++ b/.github/workflows/update-gh-pages.yml
@@ -1,7 +1,9 @@
 name: Update gh-pages
+
 on:
   push:
     branches: [ master ]
+
 jobs:
   update-gh-pages:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/update-gh-pages.yml
```diff
@@ -0,0 +1,27 @@
+name: Update gh-pages
+on:
+  push:
+    branches: [ main ]
```
quisquous, line 4: This should be master here, at this point? And then followup to change to main?

panicstevenson: It can be! I wasn't sure if we were ready for all of this to happen yet, so I was letting that happen at your leisure.

quisquous: Ah, ok! I think this was the part that I was least sure about.  So, maybe let's land this as master first, verify switching the pages source works just fine, and then I can look at switching main and master later today.

Issue #1221, opened 2020-03-18
Resources: Improvements to the lang.[language] structure


JLGarber:
As noted on #1207 , the current setup of having a separate language file for each supported language is rather cumbersome and prone to forgetfulness from maintainers. My typical reaction to these sorts of situations is "let's build around it rather than restructuring":

> That being said, a thing we could do that would require less restructuring is to place a "master elements object" into the EN language file. Something like:

> ```
> masterObject: {
>   actionsToTranslate: {
>     Action1: 'ID1',
>     Action2: 'ID2',
>     // ETC
>   },
>   statusesToTranslate: {
>     Status1: 'ID1',
>     Status2: 'ID2',
>     // ETC
>   },
>   // ETC
> }
> ```

From here we could just update a file in one place, and then run a script to pull in translations and update the relevant language files.

A more expansive solution that also doesn't make structural changes would be a script to scan trigger files for `k[whatever]` elements, grabbing those translations from the appropriate resource sites, and automagically dropping those into the relevant translation objects in each `[language].js` file.

I'm not familiar enough with the internals of the translation engine to speculate about how we might change that if we're going to do something structural, unfortunately. When I get a little more time this summer I would be happy to look into an implementation if we come up with something as a group.

quisquous:
Yeah, sorry that the language system is so half-baked.  I wasn't quite sure what I needed when I wrote this, and then raidboss turned into something different and diverged from oopsy and jobs.  It just hasn't been the highest priority (and probably isn't, even now, but it's causing bugs, so here we are.)

I think my reaction to these sorts of problems is to identify what all the issues are, see if there's something that can solve all/most of them, and see what steps towards that would look like.

Here's what I see the problems being:
- translations spread across multiple files are error prone
- not everything in cactbot uses these consistently (raidboss translates effects inline whereas oopsy leans on `lang.js` to translate effects)
- some things in language files aren't even translated (e.g. zone names)
- there's regexes in language files and in `regexes.js`
- initialization is kind of awkward (you have to do some init before using gLang, so can't use it directly in an oopsy line)
- oopsy originally used these a lot, but then got some helper functions that removed the need for it

[...]

Issue #725, opened 2019-11-06
Suggestions: Job Frame for PLD Atonement stacks


Git3746:
Just a lil' suggestion...

quisquous:
Seems reasonable! I haven't played paladin this expansion, so I haven't spent as much time on its ui.

PR #1610, opened 2020-07-26
oopsy: use netRegex for all effects

quisquous:
This is a very indirect patch which is getting rid of non-jobs usages of
gLang.kEffect, so I can use netRegex for effects in jobs, remove
gLang.kEffect entirely, in order to track atonement stacks for #725.

This also contributes to several other ongoing nebulous efforts like
using NetRegexes for oopsy instead of events (#1487) and also cleaning
up the unfortunate gLang data structures (#1221).

### DIFF
--- a/ui/oopsyraidsy/data/00-misc/general.js
+++ b/ui/oopsyraidsy/data/00-misc/general.js
@@ -137,21 +137,22 @@ let missedMitigationAbility = missedHeal;
     },
     {
       id: 'General Food Buff',
-      losesEffectRegex: gLang.kEffect.WellFed,
-      condition: function(e, data) {
+      // Well Fed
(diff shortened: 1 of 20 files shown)


quisquous:
@JLGarber This is the patch in which I'm missing the White Hole / Cursed Shriek-related Petrification effect ids in O4S.

quisquous:
This also suggests that maybe we need a trackEffect(id, field name on data) helper.  Or a deathReasonIfNotCleansed(id).

JLGarber:
`trackEffect` might be nice, yes. Handling effects all in one trigger was one nice thing about `kEffect`.

Other than that, it all looks great. It's so nice not to have to modify 6 files just to add one trigger!

quisquous:
Thanks for the reviews, both of you!

PR #1611, opened 2020-07-26
Patch 1

owenwalker92:
Updated timeline to resync on Ultima Transition.
:The Ultima Weapon:2CF5: is the actor that casts freefire vs  :Titan:2CF5


quisquous:
Thank you! Fixes for older timelines that have been changed over time are much appreciated.  :)

Issue #1306, opened 2020-04-03
Replace localized regexes with timelineReplace


quisquous:
Right now, it's redundant to have to translate a timeline once and then every regex separately.  `timelineReplace` handles timeline translation at runtime, but only does regexes via scripts in `util/`.  It would be better to have `timelineReplace` cover both timelines and regexes at runtime.

Unfortunately, there are currently some issues before this could switch over:
(1) Missing translations (translations in `timelineReplace`but not in regexes or vice versa)
(2) Inconsistent translations (translations in both, but they're not the same)

To help with this, `util/find_missing_translations.py` complains about missing `timelineReplace`for regexes and also about mismatches beteween `timelineReplace` and localized regexes right now. 

When these are all fixed, and code is written to use `timelineReplace` at runtime, then it would be safe to switch over, and delete all localized regex.

Example errors:
```
.\ui\raidboss\data\04-sb\raid\o4s.js:925 O4S Neo Beyond Death Delta Initial: missing timelineReplace for regex '^.{14} 1a:(?<targetid>[0-9a-f]{8}):(?<target>.*?) gains the effect of (?<effect>beyond death) from (?<source>.*?) for (?<duration>-?[0-9]+(?:[.,][0-9]+)?(?:e-?[0-9]+)?) seconds\.' (but have regexDe)
.\ui\raidboss\data\04-sb\raid\o12s.js:613 O12S Archive Peripheral: incorrect timelineReplace for regex, got '^.{14} 1b:[0-9a-f]{8}:recht(?:e|er|es|en) arm:....:....:(?:009c|009d):', needed regexDe '^.{14} 1b:[0-9a-f]{8}:rechter arm:....:....:(?:009c|009d):'
.\ui\raidboss\data\04-sb\eureka\eureka_hydatos.js:299 BA Owain Ice Element: missing timelineReplace for regex '^.{14} 00:0044:.*?:[^:]*:munderg, turn blood to ice'
```

valarnin:
Copy/pasting my comment from 3120.

---

Wouldn't it be better to do either of these instead?
- Keying triggers with enemy names off of bNpcId
- Pulling translation info from XIVAPI and looking up from that

It seems a bit odd to me that a lot of triggers keep on keying off of name instead of ID in general, and something like one of the following would be better all around:

```typescript
    {
      id: 'T1 High Voltage',
      // `npcName` in this case is { [englishNpcName: string]: LocaleText }
[...]

PR #1613, opened 2020-07-26
fixed replacements in eureka hydatos

Akurosia:


ui/raidboss/data/04-sb/eureka/eureka_hydatos.js
```diff
@@ -1011,6 +1011,9 @@
         'The memories of heroes past live on again': 'Das Vermächtnis vergangener Helden lebt von Neuem auf',
         'Munderg, turn flesh to ash': 'Munderg, entfessele den Flammeneid',
         'Munderg, turn blood to ice': 'Munderg, das Eis der Ewigkeit soll sie für Äonen bannen',
+        'The Shin-Zantetsuken Containment Unit': 'Shin-Zantetsuken-Quarantäneblock',
+        'The Lance of Virtue Containment Unit': 'Lanze der Tugend-Quarantäneblock',
+        'The Proto-Yadis Containment Unit': 'Proto-Yadis-Quarantäneblock',
```
VanSlanzar, line 1016: it's `The Proto Ozma Containment Unit` in English on file txt

Akurosia:
i also noticed that there might be a problem with some of the replacements e,g, `The Proto Ozma Containment Unit` with `'Proto Ozma(?! containment)'` as i get `5000.0 "--sync--" sync /00:0839:The Proto-Yadis Containment Unit will be sealed off/ window 5000,0 #MISSINGSYNC`

quisquous:
> i also noticed that there might be a problem with some of the replacements e,g, `The Proto Ozma Containment Unit` with `'Proto Ozma(?! containment)'` as i get `5000.0 "--sync--" sync /00:0839:The Proto-Yadis Containment Unit will be sealed off/ window 5000,0 #MISSINGSYNC`

Here's what I get with your patch applied:
```
$ python util/check_translation.py -f ui/raidboss/data/04-sb/eureka/eureka_hydatos.js -t de | grep -a 5000
5000.0 "--sync--" sync /00:0839:Proto-Yadis-Quarant▒neblock will be sealed off/ window 5000,0
```

I'm not sure why yours is different, to be honest.  Unfortunately, `check_translation.py` doesn't know about `common_replacement.js`, but I think this is otherwise correct.

For what it's worth, I'm not inclined to fix `check_translation.py` to include `common_replacement.js`.  I feel like having Python tools trying to replicate JavaScript logic was a big mistake in the past.  I think my ideal vision here would be:
* remove the translate timeline functionality from `check_translation.py` and try to reuse `timeline.js`'s real `GetReplacedEtc` helpers to do the translations
* fix #1306, and then once that's done then `check_translation.py` can be removed entirely
* maybe think about how to combine `find_missing_translations.js` with the logic in `timeline.js` so that's not in two places as well

PR #1624, opened 2020-07-28
Remove lang files

quisquous:
The problems this is trying to solve is laid out here: https://github.com/quisquous/cactbot/issues/1221#issuecomment-602005392

I broke these up into separate commits, which hopefully should be easier to follow.  Probably the most interesting commit is 0d69d57 which creates a new `resources/translations.js` file (and might be a good home for some new `LocaleText` in the future) and combines all of the lang regexes in both Regex and NetRegex varieties.

The only thing really left to do is to remove the zones, and that's waiting on #1037.  I moved them out of the lang files just so that I could remove all the includes and variables.

Comments / nits / opinions / bikeshedding welcome.

(`27 files changed, 526 insertions(+), 1219 deletions(-)` isn't too shabby either.)


resources/translations.js
```diff
@@ -0,0 +1,76 @@
+'use strict';
+
+let LocaleRegex = {};
+let LocaleNetRegex = {};
+
+// TODO: figure out some non-var way to require files for node, but not break html includes.
+/* eslint-disable no-var */
+var Regexes;
+var NetRegexes;
+/* eslint-enable */
```
quisquous, line 43: Also, I still don't know how to fix this.  When you have files that you want to include when running in node, you need to `require` them.  When you include them in html, some other file will define it.  I don't see a way to handle both without `var`, but maybe there's some more obvious solution that I haven't been able to find while searching.

ui/jobs/jobs.js
```diff
@@ -1021,11 +1126,11 @@ class Bars {
 
     this.o.pullCountdown.width = window.getComputedStyle(pullCountdownContainer).width;
     this.o.pullCountdown.height = window.getComputedStyle(pullCountdownContainer).height;
-    this.o.pullCountdown.lefttext = gLang.kUIStrings.Pull;
+    this.o.pullCountdown.lefttext = kPullText[this.options.DisplayLanguage] || kPullText['en'];
```
quisquous, line 1129: I think being able to use the display language here is also a really nice change.

quisquous:
@JLGarber @Akurosia @valarnin FYI if you have any thoughts about this

JLGarber:
Nothing from me, no. I'm very happy to see these going away. `NetRegexes` have made oopsy work significantly less frustrating.

Akurosia:
Fine from my side as well :)

valarnin:
This looks good to me in general. The only thing I'm not really a fan of is this syntax:

```js
let x = langArray[displayLang] || langArray['en'];
```

I wonder if it would be better to use a [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) and have a default return, like this:

```js
const defaultLangHandler = {
  get: function(obj, prop) {
    return prop in obj ?
      obj[prop] :
      obj['en'];
  }
};

const langArray = new Proxy({
  'en': ...,
  'fr': ...,
}, defaultLangHandler);
```

And then it could be invoked directly as this instead:
```js
let x = langArray[displayLang];
```

quisquous:
The Proxy stuff is a great suggestion, thank you!.  I didn't know about that.  I think there are a million places in the code where that pattern occurs, so I should probably add some `TranslateProxy` in a common place and have everything use it.  I'll look at doing that in a follow-up.

quisquous:
Ok, fixed @panicstevenson's thoughts about an unnecessary global in 66a6170.

Thanks for all the feedback.  I wanted to make sure folks who cared about translations knew about this and thought it was good.  :+1:

PR #1631, opened 2020-07-31
oopsy: add The Weeping City of Mhach

quisquous:


quisquous:
This needs a test run, but I thought I'd put it up for perusal.

JLGarber:
Looks great overall, thanks for taking care of that!

Silken Spray, 1824, is missing. It's a `damageWarn` telegraphed rear cone attack from Arachne Eve.

Holy, 182E, could be a knockback failure on Ozma. It comes up twice during the Cube block and can push players off the platform.

I would argue that Feint Particle Beam should be a simple warning rather than a share warning. Any players not targeted by it made a mistake, and the person with the marker can avoid getting hit if they move. (It's not uncommon for the marker people to avoid it entirely while others get clipped, and it seems odd that this wouldn't get reported.)

Penetration is a gaze mechanic on Calofisteri. The failure is a stun, effect 95. Or, if you prefer to use flags, a failed Penetration is:

`22|TIMESTAMP|400004CC|Calofisteri|1822|Penetration|PLAYERID|PLAYER|750003|C760000|21|440000|E|950000|1B|18228000|0|0|0|0|0|0|0|0`

Successfully, it's:

`22|TIMESTAMP|400004CC|Calofisteri|1822|Penetration|PLAYERID|PLAYER|750003|12BD0000|20|560000|1B|18228000|0|0|0|0|0|0|0|0|0|0|`

PR #1633, opened 2020-08-01
Util: Force case insensitivity in zone seal regex

JLGarber:
When working with zoneseal messages where casing is not guaranteed, `test_timeline` can break if there are no close-by syncs:

`0 "--sync--" sync /00:0839:The Main Deck will be sealed off/`

This fails if the combat log has `the main deck` instead. Current practice is to title case the "name" portion of the zone seal message. Square's localization team is *very* good, but they seem to have real trouble with casing things, and that collides here.

Probably this logic should be moved to `encounter_tools`, but for now if nothing else this will *work*.


quisquous:
asdf indeed

PR #1642, opened 2020-08-02
raidboss: differentiate lookAwayFrom(Target|Source)

quisquous:


quisquous:
@JLGarber: I'm curious what you think about the short name vs not short name? Is this too magical? Should we be looking at ids to determine if something is a mob to use the short name?

JLGarber:
I think you're right that we should check ID for mob/not-mob. Given that Square hasn't done something so horrible as to make carbuncles/fairies experience mechanics like that, I think it's safe enough. And it's definite that Square is willing to do all sorts of variations on gazes like this, so it's most likely best to future-proof it as much as we can.

(Also, Cursed Shriek in O4s might be an appropriate addition? Haven't looked at how Square implemented that one, might not be as doable.)

JLGarber:
Oops, sorry I never followed this one up after the changes. I think this is probably okay now. Given the existence of a bare `lookAway` as well, I think this has removed any potential pitfalls for users.

PR #1645, opened 2020-08-03
job: small fix about raidbufftracker's player name in CN

Echoring:
QQ截图20200803143238
These icon can hold only 2 CN/Kor charactors. 3 will escape out of box.
(If I'm not misunderstand, this line is for CN/KR player, because only these name can be `length != 2`)

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -499,7 +499,7 @@ class Buff {
     else
       // These icon can hold only 2 CN/KR charactors. 3 will escape out of box
       txt = initials[0].slice(0, 2);
-      
+
     let color = this.info.borderColor;
 
     let readyKey = 'r:' + this.name + ':' + source;
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -494,7 +494,7 @@ class Buff {
     // TODO: could consider looking at the party list to make initials unique?
     let txt = '';
     let initials = source.split(' ');
-    if (initials.length == 2)
+    if (initials.length == 2) 
       txt = initials[0][0] + initials[1][0];
     else {
       const pattern = /[\u21-\u7e]+/g;
@@ -502,7 +502,7 @@ class Buff {
         txt = initials[0].slice(0, 3);
       else
         txt = initials[0].slice(0, 2);
-      }
+    }
 
     let color = this.info.borderColor;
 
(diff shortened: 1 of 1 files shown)


ui/jobs/jobs.js
```diff
@@ -497,7 +497,8 @@ class Buff {
     if (initials.length == 2)
       txt = initials[0][0] + initials[1][0];
     else
-      txt = initials[0].slice(0, 3);
+      // These icon can hold only 2 CN/KR charactors. 3 will escape out of box
+      txt = initials[0].slice(0, 2);
```
quisquous, line 501: Is the code on line 496 is doing the right thing here for Chinese and Korean names? Or, is there some better way to shorten to two characters?

Echoring: line 496 `let initials = source.split(' ');` do not work for Chinese names.(I'm not very sure about Korean)
in JP/NA/EU, player's name is like `Suzune Tsukuyomi`, a space separates the last name from the first name.
But in CN, there is **no last name**, at most 6 charactors allowed in name, and space is not allowed.

There is a problem that CN player's name could also have English charactors. Look like this box can hold at most 4 English charactors/1 chinese charactor+2 English charactors.
Can it be like this: 
```
var pattern = new RegExp("[\u4E00-\u9FA5]+"); // A RegExp that matches Chinese charactors.
if (pattern.test(initials[0][0]) && pattern.test(initials[0][1]) && pattern.test(initials[0][2]))
  txt = initials[0].slice(0, 2); // If the first 3 charactor is both Chinese charactor.
else
  txt = initials[0].slice(0, 3);
```

MaikoTan: Instead of matching Chinese characters, it is better to filter English characters in Regex, I think.
```javascript
const pattern = /[\u21-\u7e]+/g;
if (pattern.test(initials[0].substr(0, 3))
   txt = initials[0].slice(0, 3);
else
  txt = initials[0].slice(0, 2);
```

May it be better?

Una-p: Korea haven't second part of name, too. just have 1 input line and it could be korean or english and NOT allowed name to short (like Una Chron --> U.C).
I'm not sure about Latin characters, but I've never seen before.

quisquous:
Sounds good to me! Thanks! I really was asking in order to know if there was some better way to abbreviate here, but this seems like a good solution to cover all languages.

PR #1646, opened 2020-08-04
job: bug hotfix on #1645

Echoring:
Oops sorry, the regex has a little bug so it not work as expected.

### DIFF
--- a/resources/timericon.js
+++ b/resources/timericon.js
@@ -147,6 +147,8 @@ class TimerIcon extends HTMLElement {
         }
         #text {
           position: absolute;
+          overflow: hidden;
+          word-break: break-all;
         }
       </style>
       <div id="root" style="position: relative">
@@ -284,7 +286,8 @@ class TimerIcon extends HTMLElement {
     textStyle.top = iconTop + (iconHeight - textHeight) / 2;
     textStyle.left = iconLeft;
     textStyle.width = iconWidth;
-    textStyle.height = textHeight;
+    // Other languages' character can be higher, +5 to make them display completely.
+    textStyle.height = textHeight + 5;
     textStyle.fontSize = textHeight;
     textStyle.textAlign = 'center';
     textStyle.fontWeight = 'bold';
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -499,7 +499,7 @@ class Buff {
     if (initials.length == 2) {
       txt = initials[0][0] + initials[1][0];
     } else {
-      const pattern = /[\u21-\u7e]+/g;
+      const pattern = /^[\u21-\u7e]{2,}.*$/g;
       if (pattern.test(initials[0].substr(0, 3)))
         txt = initials[0].slice(0, 3);
       else
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -499,7 +499,7 @@ class Buff {
     if (initials.length == 2) {
       txt = initials[0][0] + initials[1][0];
     } else {
-      const pattern = /[\u21-\u7e]{2}/g;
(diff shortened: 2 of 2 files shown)


quisquous:
Would you mind explaining what the bug is? It looks like maybe there are sometimes one ASCII character but not two?

Can you leave a comment about why there is a `{2}` specifically?

Echoring:
The early one can match all name that have English charactors in first 3 chatactors. But if **a name include 2 chinese charactor and 1 English charactor**, it will also be matched and escape out of box.
The latest one only matches names which have more than 2 English characters in first 3 charactors, so the name above will not be judged True and escape out of box.

MaikoTan:
Yes, it should be `{2}`, my bad...

quisquous:
I feel like I'm still a little confused.  Is it not possible for somebody to have two English characters and 1 Chinese character? Would it be better for the check to say "if there are characters that are *not* English characters, use 2, otherwise 3" vs the current check that says "if there are two English characters in a row, then use 3, otherwise 2"?

Echoring:
"if there are characters that are not English characters, use 2, otherwise 3" will miss the name that have 2 English characters and 1 Chinese character.
I make a check below:
QQ截图20200805104834
Red line is box width.
I think this Regex will work well.

quisquous:
Sorry, but I am skeptical about counting English letters in this way.  What about letters like W that are very wide?

MaikoTan:
Well, well. Let's go back from the problem itself.
The first issue that @Echoring want to fix is "player name text in the box would wrap and overflow out of the box", then she found out the code above, so she tries to cut the best position of the text, so let it fit in the box.
But what about modifying the box itself? Is it possible to change CSS or JS code in timer-icon( I guess? ) the custom web component, let these text would not overflow?

quisquous:
Yeah, I was thinking maybe we'd need some CSS too.  What about some combination of `word-break: break-all` to break in the middle of a word and wrap to the next line, and then use `overflow: hidden` to hide it?

Echoring:
I have tried these CSS. `overflow: hidden` will cause charactors Y-axis down and make the bottom of charactors disappear.
But if I use `overflow-x: hidden`, a scrollbar will appear in icon. If we have a way to hidden this scrollbar, that is the solution.

Echoring:
@quisquous Done!
By this CSS added, it's not needed to slice the name.
Y-axis overflow of CN charactars can be solved by height +5(I'm not sure is there any better way to solve this)

MaikoTan:
I tink it is not neccesary to add comments to every code block you changed, usually, we wrote inline comments for some verbosity workflows, difficult algorithm or [magic numbers](https://en.wikipedia.org/wiki/Magic_number_(programming)).

Sometimes leave a comprehensive commit message would be better.

- [Git Book](https://git-scm.com/book/zh/v2/Git-%E5%9F%BA%E7%A1%80-%E8%AE%B0%E5%BD%95%E6%AF%8F%E6%AC%A1%E6%9B%B4%E6%96%B0%E5%88%B0%E4%BB%93%E5%BA%93#_committing_changes) if using git.exe CLI tool
- [Github Desktop Doc](https://docs.github.com/en/desktop/contributing-and-collaborating-using-github-desktop/committing-and-reviewing-changes-to-your-project#3-write-a-commit-message-and-push-your-changes) if using Github Desktop

For some commit message, not only write the title but also description to tell others what have you done.

PS: You can call these characters as [CJK characters](https://en.wikipedia.org/wiki/CJK_characters) (Chinese, Japanese, and Korean). Unless in this issue it only happens on CJK.

quisquous:
I think the +5 is fine, especially because there's a comment there explaining it.  Thanks!

PR #1649, opened 2020-08-05
Corrected some translations (Part1)

Akurosia:
There were some translation errors as the same englich name is translated to different names in other languages. 
E.g. `Earth Shaker` can be translated to `Erdstoß` or `Erderschütterer` depending of the fight (Most of the time it affects DE, a little bit less FR and rarely JP/CN/KO)

This PR is the first batch that should fixes these kind of mistakes (i adapted my scripts to point these translations out so it can be avoided in the future)

On some commits, i had used initally the incorrect ID e.g.:
t9 `Heavensfall` new id `7E4`, old id  `83B`
a10s `Impact` new  id `1A8B`, old id `1AC6`
a3s uses now following ids: `F01|EFE|F15`


ui/raidboss/data/03-hw/raid/a3s.js
```diff
@@ -345,7 +345,7 @@
         'Protean Wave': 'Vague inconstante',
         'Sluice': 'Éclusage',
         'Splash': 'Éclaboussement',
-        'Throttle': 'Geôle liquide',
+        'Throttle': 'Suffocation',
```
VanSlanzar, line 348: Don't touch plz. it's correct.
fil text a3s:
```
# Note: "Liquid Gaol" renamed to "Throttle" to disambiguate from "Liquid Claw",
# aka the healer gaol.  "Pressurize" also changed to "Embolus".
```

VanSlanzar:
Thanks for your corrections. I saw very little error fr, that reassures me.

PR #1651, opened 2020-08-06
raidboss: a4s + a5s fr

VanSlanzar:
a4s: minor correction.
a5s: add missing translations fr + removal missingTranslation=true.
I'm not doing at big pack anymore, I have not time, maybe during the holidays.
@mooondark je viens de me rendre compte que sur 3 ou 4 fichier j'ai traduit Dash par Ruée. De souvenir c'est Rush qui est traduit par ruée, du coup je me dis qu'un Dash c'est plutôt une Charge, qu'en penses tu ? cela concerne ( la voute, a5s, ravana et sophia ex).

### DIFF
--- a/ui/raidboss/data/03-hw/raid/a4s.js
+++ b/ui/raidboss/data/03-hw/raid/a4s.js
@@ -20,7 +20,7 @@
           return {
             en: 'Orbs on YOU',
             de: 'Orbs auf DIR',
-            fr: 'Orbes sur Vous',
+            fr: 'Orbes sur VOUS',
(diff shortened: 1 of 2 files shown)


quisquous:
I think this looks fine to me.  Should I wait for mooondark's answer about Dash before merging this?

mooondark:
Yeah Dash is better translated by Charge. But since it is old content, maybe he can put a PR later ?

VanSlanzar:
yes you can merged i would do a pr for dash. Thank you.

PR #1657, opened 2020-08-08
Raidboss: Alexander 12 normal package

JLGarber:
Nothing particularly special about this one. I was a little surprised at how few triggers there actually were. It feels like there's so much going on, and yet it's mostly mechanically simple. (Probably an artifact of how limited movement was during Heavensward.)

As with my other Normal submissions, the trigger file is shamelessly copied and modified from Savage. Note that I didn't pare down the translation object when stealing it from A12S. I'm assuming this is something that Akurosia or others are better-equipped to handle than I am, since I don't have that scripted. Since it doesn't affect functionality, I didn't see it as a priority for this submission.

I'm a little uncertain on the limit break trigger condition for Divine Judgment. Is that going to persist across pulls? Should I do some initialization somewhere between pulling and Divine Judgment?


ui/oopsyraidsy/data/manifest.txt
```diff
@@ -11,6 +11,7 @@
 03-hw/dungeon/fractal_continuum.js
 03-hw/dungeon/gubal_library_hard.js
 03-hw/dungeon/sohm_al_hard.js
+03-hw/raid/a12n.js
```
quisquous, line 14: The only HW raid.  Foof.  I guess I should have thought about doing these when I did the savages.  I still need to finish off a7s too <_<

ui/raidboss/data/03-hw/raid/a12n.js
```diff
@@ -0,0 +1,371 @@
...
+      beforeSeconds: 5,
+      condition: function(data) {
+        // Let's not spam tanks if they are unlucky enough to see enrage.
+        return !data.judgment && data.role == 'tank';
+      },
+      alarmText: {
+        en: 'Limit break now!',
+      },
+      run: function(data) {
+        data.judgment = true;
```
quisquous, line 19: Data is cleared between pulls, so this works.

Two other options you could choose that would eliminate the need for a variable:
1) `suppressSeconds: 9999`.
2) `regex: /^Divine Judgment$/,`

ui/raidboss/data/03-hw/raid/a12n.txt
```diff
@@ -0,0 +1,140 @@
...
+415.4 "--sync--" sync /:Alexander Prime:1AEA:/
+421.4 "Gravitational Anomaly" sync /:Alexander Prime:1AEB:/
+422.6 "Sacrament" sync /:Alexander Prime:1AE5:/
+429.8 "Incinerating Heat" sync /:Alexander Prime:1AF3:/
+
+# Summon Alexander seems to push at 30% HP.
+# It appears that Alexander has to complete at least a part of the Timegate section first?
+# (Pushes of anywhere from 5% to 30% have been seen.)
+# Technically everything immediately after is a loop.
+# However, there's a fixed enrage after a set number of loops.
```
quisquous, line 83: Oh wow, a timeline-based enrage and not a time-based one??

ui/raidboss/data/03-hw/raid/a12n.js
JLGarber: Ah, option 1 is definitely just what we want here.

ui/raidboss/data/03-hw/raid/a12n.txt
JLGarber: Sort of. To be clear, this is where the enrage happens when there are no pushes. I sort of assume that a group that ends up pushing is *probably* not going to see enrage, but it is just possible that if all previous phases are pushed, Alexander would go into a fifth loop, either completing it or breaking at the 12:15 mark to perform the enrage. (Suzaku works this way, for instance, or did before Square fixed the slow/fast instance situation.)

ui/oopsyraidsy/data/manifest.txt
JLGarber: I was slightly surprised about that too. Not to worry though, a lot of it is reconstructable from logs and recordings!

ui/raidboss/data/03-hw/raid/a12n.txt
quisquous: Would you mind rolling the loop back up and not unrolling it, then? I think if it technically is an enrage time, then it'd be better to just have the loop run until that gets hit.

I think if the goal is to indicate what the enrage timer is, then maybe there's some way to show this separately than an  entry in the timeline that's not always true.

JLGarber: Fair enough, we can do that.

ui/raidboss/data/03-hw/raid/a12n.js
quisquous: Could you also delete the `run` section too?

Issue #361, opened 2019-02-22
Feature request: Better oopsyraidsy tracking


sparkymotive:
It would be nice to have the ability to scroll through each pull, like with < and > buttons or something along those lines.

Also the ability to export said pulls to a .txt file would be tops.

quisquous:
Mmm, thanks for this idea!

One other thought I had was maybe making it possible to open up the results in a browser.  Possibly also being able to paste in a twitch link and autolinking to time stamps in a video?

sparkymotive:
Both of those sound :italian_hand:

We use twitch clips alot for mistake tracking.

quisquous:
I think if I move to ngld's overlay plugin where you can also use websocket, this idea of being able to open up results in a browser will make a ton more sense.

quisquous:
I think the summary view addresses this.

PR #1660, opened 2020-08-09
oopsy: add summary view when opened in a browser

quisquous:
This was asked for in #361.


quisquous:
Looks something like: image

I'm not sure if it makes sense to use "am I opened via websocket in a browser vs in an overlay" as the determining factor for which view to use, but I think this is a way better view if you have the room for it.  I could also make a separate html file here, even if it uses the same js underneath.

I think this creates a lot of room for followups:
* summary view over a zone / entire night
* loading old logs into it (like raid emulator, but for oopsy)
* shortcut for creating links to timestamps in twitch video? (provide a video link and a start time?)

quisquous:
> > I'm not sure if it makes sense to use "am I opened via websocket in a browser vs in an overlay" as the determining factor ... I could also make a separate html file here, even if it uses the same js underneath.
> 
> I think the latter makes more sense here; something like oopsy_report.html/oopsy_summary.html, but I wonder which option drives feature adoption more -- I can see an argument for both, and I can see making either option fairly painless for the user. I wonder if we should write something out to the console when oopsy loads saying, "Summary report available at <url>" or something similar; that way we could realistically do whatever made the most sense for us from a technical standpoint.

Unfortunately, it'd be hard to say "summary at url".  You'd need to know what port the OverlayPlugin WSServer was running on.  I think documentation is probably the best bet here?

Ok, so assuming I make a oopsy_summary.html page.  The other wrinkle to making this more accessible is that the OverlayPlugin WSServer tab will give you such URLs, but it currently only does so for overlays you have enabled.  (ngld has mentioned maybe making this use presets instead of the set of current overlays, but it's the same problem.) In either case, there won't be an oopsy_summary preset or overlay, because it's not meant to be added as an overlay.  So it will still be somewhat difficult to find this url.

I think for backwards compatibility, it'll be tough to rename oopsyraidsy.html proper.  So the questions are, should oopsyraidsy.html behave magically and check WSServer status? (This would help with the use via the OverlayPlugin-provided url?) Should we have separate html for summary? and live list too?

Options:
1) html proliferation: have oopsyraidsy (can be both) / oopsy_summary (only summary) / oopsy_live (only live).
2) somewhat magic: have oopsyraidsy (can be both) / oopsy_summary (only summary)
3) entirely magic: only oopsyraidsy (can be both)
4) no magic: oopsy_live (only live) and oopsy_summary (only summary)

quisquous:
Ok, I ended up going with the "no magic" option.  `oopsyraidsy.html` is the live list and `oopsy_summary.html` is the summary view and there's no "do something different if you're loading via websocket" shenanigans (which I think would cause issues with people loading oopsy as an OBS overlay anyway).

PR #1661, opened 2020-08-09
Raidboss: Dun scaith trigger cleanup

### LINKED ISSUE #549: Trigger issues in Dun Scaith
I could use some community help for this one. After writing the initial version of the Dun Scaith triggers, I've had some time this week to revisit that. I had a list of items that needed fixing in the [initial pull request](https://github.com/quisquous/cactbot/pull/512#issue-304035684). Several are still outstanding:

1.Particle Beam Circles
2. Hand Tethers
3. Flameflow
4. Atomos Detection

On Particle Beams, I've isolated it to the log line
```[01:23:40.002] FB:Warning: 01:23:48.978 Ground AOE tick but cannot find buff ID [739]```
This consistently appears alongside the Particle Beam circles during both Scathach and Diabolos. My guess is this is something not currently supported in the ACT plugin?

On hand tethers, I'm totally stumped. No 23 log lines appear *unless the targeted person looks at the hand*. Once they look at it, then the tether will change from 0001 to 0002 and back when they look away and back at it. It's baffling.

Flameflow and Atomos are regex matching issues. On Atomos, the way things work is that Ferdiad targets himself and casts Wormhole. As he initiates the cast, the game summons 3 (in phase 1) or 4 (in phase 2) Atomos. Additionally, in phase 1 the game summons either an Aether or an Aetherial Chakram, and on phase 2 the game will summon 2 in any combination. This Aether/Chakram immediately tethers to the type of Atomos that has an even number and starts casting Juggling Sphere. At the end of the cast, the Aether/Chakram will swoop in, reappear at the untethered Atomos, and explode in either a circle or a donut.

 The relevant regex for Atomos is:
```/ 14:(\y{AbilityCode}):\y{Name} starts using Juggling Sphere on Wailing Atomos/```
And the trigger function is:
```if (matches[1] == '1C9F')``` or ```if (matches[1] == '1CA0')```

I feel like the problem is that I'm fundamentally misunderstanding exactly what the ```matches[x]``` syntax is returning. What I understood it to do is return the entire matched log line, in the case of `matches[0]`, or the matched substring in the relevant match group, in the case of `matches[1]`, as noted in the [triggers readme](https://github.com/quisquous/cactbot/blob/ca6a8bfe113d29091889548f2264c7726ce2dc48/ui/raidboss/data/triggers/README.txt#L55). Clearly this is wrong, but I haven't had much luck trying to determine what *is* the case so far.

I could just take the easy way out and write one trigger per possibility, but that's ridiculous. There's clearly a better way.

[A recent log from Dun Scaith.](https://www.dropbox.com/s/c70lua9zykz3elu/Dun%20Scaith%2020190816.zip?dl=0)

### PR DESCRIPTION
**Pull request by JLGarber**
(This should resolve #549.)

This probably isn't as self-explanatory as some others, unfortunately. NetRegexes have made most of this possible.

To begin with, anytime there's a trigger that uses a choice of abilities, we now use the `['ID1', 'ID2']` construction, rather than `'ID[12]'`. It's so nice not to use bare regexes anymore!

Certain triggers on Proto-Ultima have been removed. All of them have prominent AoE indicators alongside, so there's really no good reason to keep them. (It was good practice for me at the time though!)

All `addedCombatant` lines have been replaced with `addedCombatantFull` in order to be completely language-agnostic without the need for language-specific regexes. (It's technically not necessary for all of them, but since the Atomos triggers *do* require it, it seemed to make sense to be consistent across the entire trigger file.)

The biggest change is to the Atomos triggers. Two problems existed here. First, anytime there were two tethered Atomos on the field, the user would be notified twice, once for each. This was annoying, and potentially a big deal for users with auditory processing sensitivities. (Particularly so if they happen to be using TTS.)

To understand the second problem, we need to go through the process for how the Atomos mechanic works:

1. Ferdiad summons variously-colored Atomos and 1-2 Aether/Aetherial Chakram.
2. Each Aether/Aetherial Chakram tethers to an Atomos and casts Juggling Sphere.
3. The Aether/Aetherial Chakram warps to the untethered Atomos of the same color and casts its Explosion ability. (`1CA1` for Aether, `1CA2` for Chakram.)

This is fine, but sometimes we get a situation where the packet containing the spawn information is processed out of order and something like this happens:

```
03|2020-08-04T20:47:00.2470000-04:00|40001026|Aether|0|3c|0|
35|2020-08-04T20:47:01.4530000-04:00|40001026|Aether|40001029||C3B3|
20|2020-08-04T20:47:01.5410000-04:00|40001026|Aether|1C9F|Juggling Sphere|40001029||6.70|
03|2020-08-04T20:47:01.4010000-04:00|40001029|Wailing Atomos|0|3c|0|
03|2020-08-04T20:47:01.4010000-04:00|40001028|Wailing Atomos|0|3c|0|
03|2020-08-04T20:47:01.4010000-04:00|40001027|Cursing Atomos|0|3c|0|
```

The old triggers relied on the name of the Atomos being present in the log line, but as we can see, if the Atomos spawn is later in the log, the name field in the Juggling Sphere line will be empty. In order to solve this problem, we have to go to the annoying nonsense of "collect -> delay -> notify -> cleanup". The path through the Atomos triggers should be relatively self-explanatory, but I'll lay it out here so my thought process can be better understood and so that errors might be easier to catch:

1. [SETUP TRIGGER] Atomos spawn in whatever configuration, always 3-4 of them, and always in a 1-2/2-2 split. We add their entity ID to the appropriate `data` property, identifying them by their BNPCname.
2. [COMPILE TRIGGER] When an Aether/Aetherial Chakram starts casting, we delay for half a second to allow the `addedCombatantFull` trigger to populate the `cursing/wailing` properties. We then check what color Atomos the Aether/Chakram is tethered to and increment the `data.atomos` property accordingly. (We do not stringify here because we need this to remain a number through the compilation log lines.)
3. [RESPONSE TRIGGER] At the same time the compile trigger activates, we delay for a full second, and then read the value contained in `data.atomos`. We do some awful manipulation to format the string correctly for the response enumeration object. (The situation where there's exactly one Chakram is coerced to the same value whether or not there's an Aether present, as the handling wouldn't change.) We then return the appropriate response.
4. The cleanup trigger cleans up.

Suggestions on cleaner handling for Atomos information would be welcome. I don't really like the string manipulation necessary to format things for the response object, but it seemed like a relatively simple way of cleanly compiling the information.

Finally, some triggers were doing double duty. Aetherochemical Flare and Supernova are combined on Proto-ultima, while Shadethrust and Nox are combined on Scathach and Diabolos. I considered changing this, since it isn't quite according to modern standards, but I think that it's acceptable even so since the handling in each case is literally identical.

### DIFF
--- a/ui/raidboss/data/03-hw/alliance/dun_scaith.js
+++ b/ui/raidboss/data/03-hw/alliance/dun_scaith.js
@@ -375,12 +375,7 @@
       // Triggering off the Bit appearance
       // The cast time on Aetheromodulator is under 3 seconds
       id: 'Dun Scaith Bit Circles',
-      netRegex: NetRegexes.addedCombatant({ name: 'Proto Bit', capture: false }),
-      netRegexDe: NetRegexes.addedCombatant({ name: 'Proto-Drohne', capture: false }),
-      netRegexFr: NetRegexes.addedCombatant({ name: 'Proto-Foret', capture: false }),
-      netRegexJa: NetRegexes.addedCombatant({ name: 'プロトビット', capture: false }),
-      netRegexCn: NetRegexes.addedCombatant({ name: '原型浮游炮', capture: false }),
-      netRegexKo: NetRegexes.addedCombatant({ name: '프로토 비트', capture: false }),
+      netRegex: NetRegexes.addedCombatantFull({ npcNameId: '3782', capture: false }),
       suppressSeconds: 5,
       infoText: {
         en: 'Avoid Bit AoEs',
@@ -391,12 +386,7 @@
     },
     {
       id: 'Dun Scaith Aether Collectors',
-      netRegex: NetRegexes.addedCombatant({ name: 'Aether Collector', capture: false }),
-      netRegexDe: NetRegexes.addedCombatant({ name: 'Ätherakkumulator', capture: false }),
-      netRegexFr: NetRegexes.addedCombatant({ name: 'Accumulateur D\'Éther', capture: false }),
-      netRegexJa: NetRegexes.addedCombatant({ name: 'エーテル集積器', capture: false }),
-      netRegexCn: NetRegexes.addedCombatant({ name: '以太收集器', capture: false }),
-      netRegexKo: NetRegexes.addedCombatant({ name: '에테르 집적기', capture: false }),
+      netRegex: NetRegexes.addedCombatantFull({ npcNameId: '3781', capture: false }),
       suppressSeconds: 5,
       alertText: {
         en: 'Kill collectors',
--- a/ui/raidboss/data/03-hw/alliance/dun_scaith.js
+++ b/ui/raidboss/data/03-hw/alliance/dun_scaith.js
@@ -283,23 +283,6 @@
     },
     // PROTO-ULTIMA
     {
-      // The trident laser is a series of three separate casts
-      // Each has an incremental ID: 1D96, 1D97, 1D98
-      id: 'Dun Scaith Aetherochemical Laser',
-      netRegex: NetRegexes.startsUsing({ id: '1D96', source: 'Proto Ultima', capture: false }),
-      netRegexDe: NetRegexes.startsUsing({ id: '1D96', source: 'Proto-Ultima', capture: false }),
-      netRegexFr: NetRegexes.startsUsing({ id: '1D96', source: 'Proto-Ultima', capture: false }),
-      netRegexJa: NetRegexes.startsUsing({ id: '1D96', source: 'プロトアルテマ', capture: false }),
-      netRegexCn: NetRegexes.startsUsing({ id: '1D96', source: '究极神兵原型', capture: false }),
-      netRegexKo: NetRegexes.startsUsing({ id: '1D96', source: '프로토 알테마', capture: false }),
-      infoText: {
-        en: 'Dodge trident laser',
-        de: 'Weiche dem Laser aus',
-        fr: 'Evitez le laser',
-        cn: '躲避三向激光',
-      },
-    },
-    {
       // Handles both 1E52 Aetherochemical Flare and 1D9D Supernova
       id: 'Dun Scaith Proto-Ultima Raid Damage',
       netRegex: NetRegexes.startsUsing({ id: ['1E52', '1D9D'], source: 'Proto Ultima', capture: false }),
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: ui/raidboss/data/03-hw/alliance/dun_scaith.js [hunk 1]
```diff
@@ -145,62 +152,94 @@
...
-            de: 'Gehe zu dem nicht verbundenen blauem Atomos',
-            fr: 'Allez vers la Gueule bleue non-liée',
-            cn: '靠近蓝色小怪',
-          };
-        }
+      id: 'Dun Scaith Atomos Compile',
+      netRegex: NetRegexes.startsUsing({ id: ['1C9F', '1CA0'] }),
+      delaySeconds: .5,
+      run: function(data, matches) {
+        data.atomos = data.atomos || 0;
```
quisquous, line 179: I think my inclination when I see a pattern like this that's mapping some more complicated set of data into and out of an integer is to suggest to write it out explicitly.  What do you think about something more along these lines:

```
data.sphere = data.sphere || [];
data.donut = data.donut || [];
const target = data.wailing.includes(matches.targetId) ? 'wailing' : 'cursing';
if (matches.id === '1C9F')
  data.sphere.push(target);
else
  data.donut.push(target);
```

And then later on you could enumerate the possibilities:
```
if (data.donut.length === 2)
  // on any
else if (data.sphere.length === 2)
  // avoid all
else if (data.donut.length === 1)
  if (data.donut[0] === 'wailing')
    // on wailing
  else
    // on cursing  
else // remaining case: data.sphere.length === 1 && data.donut.length === 0
  if (data.sphere[0] === 'wailing')
    // avoid wailing
  else
    // avoid cursing

```

JLGarber: Reasonable. I was more trying to avoid the awful IF tower, and to keep the response localizations all in one place together, but this is definitely just as readable.

(I suppose this is as good a place to ask, are we moving to `===` strict comparison generally now? I would prefer that, but to date the repository seems to use `==` loose.)

JLGarber:
Alright, I think that should do it. Thanks for the cleanup on the Atomos triggers!

> I think this is reasonable. In general, I think it's fine for the same trigger to be used in multiple places if the same response would be used for each. I think for Shadethrust, it's obvious from the regex that it applies to both. The Nox orbs could have a comment that says "these apply in both of these fights" if you think it would help make it clearer.

Fair point, moved the comments for these inside the trigger bodies to be 100% explicit on both.

ui/raidboss/data/03-hw/alliance/dun_scaith.js
quisquous: The if tower isn't great, but it's at least much more readable.  I'm not sure in this case that there's any way to avoid it.

(Re: comparisons.  Yeah, past behavior was mostly just "we were not good JavaScript programmers, sorry." I think my inclination is that we use strict where we can, especially in new code.)

ui/raidboss/data/03-hw/alliance/dun_scaith.js
```diff
@@ -430,8 +430,8 @@
         cn: '躲避AOE后击杀康拉',
       },
     },
-    // These triggers are common to both Scathach and Diabolos
     {
+      // This trigger is common to both Scathach and Diabolos, since handling is 100$ identical.
```
quisquous, line 434: Sorry, did you mean `100%` here and not `100$`?

JLGarber: So I did. Just as well, I caught a couple of other things that needed editing/removing while fixing that.

quisquous:
Thanks for all of this!!

PR #1663, opened 2020-08-11
Raidboss: add ja trigger translation in 4.x alliance

MaikoTan:
And del `missingTranslation` flag in those file.


quisquous:
Thanks!!

PR #1664, opened 2020-08-11
A12N "Tank Limit Break" CN translation.

ShadyWhite:


ui/raidboss/data/03-hw/raid/a12n.js
```diff
@@ -83,7 +84,7 @@
             de: 'Auf DIR sammeln',
             fr: 'Package sur VOUS',
             ja: '自分にスタック',
-            cn: '集合点名',
+            cn: '分摊点名',
```
JLGarber, line 87: To be clear, I took this from https://github.com/quisquous/cactbot/blob/16ae537b79181c04cfb3e562f3086a4fff4785f5/resources/responses.js#L293-L304. Is this just situationally incorrect here, or should `responses` be updated as well? (Because of the necessity to do some conditions within the trigger body, we can't just use a bare `response` object here.)

ShadyWhite: Both are correct, but I think "分摊" is more accurate.But it doesn't matter, because these two words mean the same thing.

ShadyWhite: "集合" means get together,"分摊"means get together and share the damage.So I think “集合” is OK.

JLGarber: Okay. Just wanted to be sure, since I simply lifted it from somewhere else.

PR #1665, opened 2020-08-12
Raidboss: Heroes' Gauntlet initial timeline and triggers

JLGarber:
Untested, but it should work. I'll run this with a trust later tonight depending on how things go for my static and work on timelines/oopsy separately. Most of this I'm confident on, but the zombie triggers could use work. There's got to be a better way to handle that mechanic, or at least to communicate it to players.

### DIFF
--- a/ui/raidboss/data/05-shb/dungeon/heroes_gauntlet.js
+++ b/ui/raidboss/data/05-shb/dungeon/heroes_gauntlet.js
@@ -32,14 +32,6 @@
       },
     },
     {
-      id: 'Heroes Gauntlet Small Zombie Tether',
-      netRegex: NetRegexes.tether({ id: '0011' }),
-      condition: Conditions.targetIsYou(),
-      infoText: {
-        en: 'Zombies away from boss',
-      },
-    },
-    {
       id: 'Heroes Gauntlet Large Zombie Tether',
       netRegex: NetRegexes.tether({ id: '004F' }),
       condition: Conditions.targetIsYou(),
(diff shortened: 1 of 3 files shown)


JLGarber:
Many thanks to quisquous for suggesting running this in a trust to get clean timeline loops, and for double-checking the second boss for me so I didn't have to re-do the whole trust.

Everything seemed to time out nicely against `test_timeline`. I'll need to send in a separate pull request for `encounter_tools` to fix an issue I created when fixing something else though...

ui/raidboss/data/05-shb/dungeon/heroes_gauntlet.js
```diff
@@ -0,0 +1,137 @@
...
+      infoText: {
+        en: 'Away from tether marker',
+      },
+    },
+    {
+      id: 'Heroes Gauntlet Small Zombie Tether',
+      netRegex: NetRegexes.tether({ id: '0011' }),
+      condition: Conditions.targetIsYou(),
+      infoText: {
+        en: 'Zombies away from boss',
```
quisquous, line 39: I still think this is too prescriptive.  If you move early, you have to move twice too.

JLGarber: How did we want to handle this again? Just "small tether zombie", or `Responses.spread()`?

quisquous: I think it'd be fine to leave this silent, especially because you can live through being hit by multiple of these and it's hard to describe.  If you wanted to put anything, I'd say "avoid orbs" or "avoid explosions" when those spawn from the zombies?

JLGarber: MM. Let's do silent then for now, and we can come back to it if people complain.

PR #1677, opened 2020-08-13
raidboss: Add Initial Warrior Of Light Ex Triggers

panicstevenson:
Initial triggers for WoL Ex; mostly tested -- my farm party couldn't get a clear.

Feel free to test: https://panicstevenson.github.io/cactbot/ui/raidboss/raidboss.html

### DIFF
--- a/ui/raidboss/data/05-shb/trial/wol-ex.js
+++ b/ui/raidboss/data/05-shb/trial/wol-ex.js
@@ -328,7 +328,7 @@
       netRegex: NetRegexes.headMarker({ id: '0067' }),
       condition: Conditions.targetIsYou(),
       alertText: {
-        en: 'puddle on you',
+        en: 'Puddle on YOU',
         de: 'Fläche auf DIR',
         fr: 'Zone au sol sur VOUS',
         ja: '自分に水溜り',
--- a/ui/raidboss/data/05-shb/trial/wol-ex.js
+++ b/ui/raidboss/data/05-shb/trial/wol-ex.js
@@ -1,0 +1,368 @@
+'use strict';
+
+[{
+  zoneId: ZoneId.TheSeatOfSacrificeExtreme,
+  timelineFile: 'wol-ex.txt',
+  triggers: [
+    {
+      id: 'WOLEx Terror Unleashed',
+      netRegex: NetRegexes.ability({ source: 'Warrior Of Light', id: '4F09', capture: false }),
+      condition: function(data) {
+        return data.role === 'healer';
+      },
+      suppressSeconds: 5,
+      alertText: {
+        en: 'Full Heal Everyone',
+      },
+    },
+    {
+      id: 'WOLEx To The Limit',
+      netRegex: NetRegexes.startsUsing({ source: 'Warrior Of Light', id: '4F3[456]' }),
+      run: function(data, matches) {
+        data.limitBreaks = {
+          en: {
+            0: 'role positions',
+            1: 'stacks',
+            2: 'meteor',
+          },
+        }[data.displayLang];
+        if (matches.id == '4F34')
+          data.limitBreak = data.limitBreaks[0];
+        if (matches.id == '4F35')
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/05-shb/trial/wol-ex.js
```diff
@@ -0,0 +1,368 @@
...
+        en: 'Protean',
+      },
+    },
+    {
+      id: 'WOLEx Imbued Absolute Fire III',
+      netRegex: NetRegexes.startsUsing({ source: 'Warrior Of Light', id: '4EF3', capture: false }),
+      run: function(data) {
+        data.imbued = data.imbued || [];
+        data.imbuedFire = {
+          en: 'Stop',
```
JLGarber, line 60: Here and elsewhere, is it worth dropping the current translations from `responses` in, given that it's already done? Might be nice for the users who are running between now and when the translators get to this. (Obviously it's going to be imperfect given that non-Latin languages might not map correctly with the `+` joining, but it's something at least, right?)

JLGarber:
Thanks for this! My group will be running tonight, so I'll probably have some actual feedback in a couple hours.

ui/raidboss/data/05-shb/trial/wol-ex.js
panicstevenson: My hope is that we can push Responses.X to the stack and then instead of calling `alertText:` call `response:`? I don't know if that's how that works so I wanted to consult with @quisquous first. But yea, failing that, we're copying/pasting and that can be done either in this PR or following up.

JLGarber: Right, so far as I know that's not currently supported, I've just been doing copypasta for these situations in my own work. But I would be extremely happy to be wrong!

panicstevenson: Sounds like no to my hope of chaining responses, but like you wrote it's a common issue that we can address. But, reading the responses more they're also a little long (ie more than a single word) so they're hard to chain together instead of "Stop + Out + Stack" you get like "Stop everything!! + Get Out + Stack" which isn't as succinct as the callout should be.

General translations pushed.

ui/raidboss/data/05-shb/trial/wol-ex.js
```diff
@@ -0,0 +1,443 @@
...
+        de: 'Tankbuster',
+        fr: 'Tank buster',
+        ja: 'タンクバスター',
+        cn: '坦克死刑',
+        ko: '탱버',
+      },
+    },
+    {
+      id: 'WOLEx Berserk / Deep Darkside',
+      // Either silence, but don't be too noisy
```
quisquous, line 213: I would maybe argue that interrupts that wipe the raid should get an alarm, and it's worth being noisy here?

panicstevenson: Noisy was relative to "spamming this twice," so I am onboard with bumping it to alarm.

ui/raidboss/data/05-shb/trial/wol-ex.js
```diff
@@ -0,0 +1,443 @@
...
+      },
+      run: function(data) {
+        data.deluge = true;
+      },
+    },
+    {
+      id: 'WOLEx Absolute Holy',
+      netRegex: NetRegexes.headMarker({ id: '00A1' }),
+      condition: (data) => !data.deluge,
+      response: Responses.stackOn(),
```
JLGarber, line 359: This is activating during San on the Ninja block, causing multiple callouts. This one needs to be conditioned on "Ninja not active" or something similar. Or San shouldn't have an individual trigger and this should be just a basic "stack", plus a suppress property.

panicstevenson: Should be fixed if you're testing this in real time. Thanks!

PR #1690, opened 2020-08-15
wol ko translate

Una-p:
Before merge, 1 quick question.
https://github.com/quisquous/cactbot/blob/fac948df2e306c665ba2e3d887707290799dbf07/ui/raidboss/data/05-shb/trial/wol.js#L193-L199
What is GTFO means?


Akurosia:
It is the abbreviation for `Get the fuck out`

PR #1691, opened 2020-08-15
raidboss: add wol nm/ex initial translations

MaikoTan:
But what is "GTFO" in wol?

EDIT: Well, I got it from [Akurosia](https://github.com/quisquous/cactbot/pull/1690#issuecomment-674368275), so never mind...


quisquous:
Sorry! Maybe that GTFO one should just say "don't stack"?

MaikoTan:
> Sorry! Maybe that GTFO one should just say "don't stack"?

Never mind, just use any one you prefer. But maybe in the future, a commit message or comment would help when using these abbrivated concept？

PR #1693, opened 2020-08-15
plugin: Fix FateWatcher casting ActorControl143 Category as int instead of ushort

### LINKED ISSUE #1688: Cactbot Eureka not updating
The Eureka plugin for cactbot doesn't seem to want to update NM's as they spawn. It tracks day/night properly and add flags but nothing else. Attached is a picture of a bunny event spawned but it not showing on the cactbot map. Not sure if I have a setting wrong or this was from 5.3 update or what. 
Advanced_Combat_Tracker_kj3sDj9E19

KattTails:
FIxes #1688 

Category should always have been a ushort but it seems with 5.3 there can now be garbage bytes resulting in erroneous values.

### DIFF
--- a/plugin/CactbotEventSource/FateWatcher.cs
+++ b/plugin/CactbotEventSource/FateWatcher.cs
@@ -165,7 +165,7 @@ namespace Cactbot {
     }
 
     public unsafe void ProcessMessage(byte* buffer, byte[] message) {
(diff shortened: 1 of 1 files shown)


quisquous:
Thank you for this super quick fix!!

PR #1694, opened 2020-08-15
raidboss: susano-ex fr

VanSlanzar:
added missing translations fr + removal missingTranslations
`Pointe d'éclair` is an incorrect name, so I corrected it.

I'm putting the link and a capture to you, because I know I bothered you a lot with unnecessary capital letters in `replaceSync` (I was misled due to the different ways written, in-game, logs, fflogs, saintcoinach, garlandtools, as well as the display bug that you fixed in the check timeline) and I apologize and thank you for your patience.
https://xivapi.com/BNpcName/6223?pretty=true
susano

### DIFF
--- a/ui/raidboss/data/04-sb/trial/susano-ex.js
+++ b/ui/raidboss/data/04-sb/trial/susano-ex.js
@@ -13,6 +13,7 @@
       infoText: {
         en: 'look for cloud',
         de: 'Nach Wolke ausschau halten',
+        fr: 'Cherchez le nuage',
         cn: '寻找雷云',
         ko: '구름 확인',
       },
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for the translations.

It doesn't ultimately matter how these syncs are capitalized.  I think when you are adding a new one, do what makes the most sense.  It's really going back and updating old ones that feels the most unnecessary and causes me extra work and that I would like to avoid.  I hope that makes sense.

VanSlanzar:
Of course, but you're right, first in `replaceSync` it doesn't matter and I'll look directly into saint coinach. And very often the capital letters are only for the proper names, thank you to you.

PR #1698, opened 2020-08-16
Raidboss: add ja translation in all ShB dungeons

MaikoTan:


quisquous:
Oh wow, so many translations! Thank you!

PR #1699, opened 2020-08-16
util: support not specifying -rf in test_timeline.py

quisquous:


JLGarber:
That looks about right. Sorry I missed that when specifying the conditions on this!

PR #1700, opened 2020-08-16
Added B rank enabled/disabled via UI config

MyTechnoHunter:
along with all ranks enabled/disabled via user config

### DIFF
--- a/ui/radar/radar.js
+++ b/ui/radar/radar.js
@@ -126,7 +126,7 @@ class Radar {
     if (monster.rank in options.RankOptions)
       options = Object.assign({}, this.options, options.RankOptions[monster.rank]);
     // Check if monster rank enabled or not in config
-    if (monster.rank === 'B' && options.BRankEnabled === false)
+    if (monster.rank === 'B' && !options.BRankEnabled)
       return;
     if (!options.Enabled)
       return;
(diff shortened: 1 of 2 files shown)


ui/radar/radar.js
```diff
@@ -120,7 +125,12 @@ class Radar {
     // option overwrite
     if (monster.rank in options.RankOptions)
       options = Object.assign({}, this.options, options.RankOptions[monster.rank]);
+    // Chceck if monster rank enabled or not in config
+    if(monster.rank === "B" && options.BRankEnabled === false)
+      return;
     if (options.Type === 'mob') {
+      if (options.Enabled === false)
```
quisquous, line 132: This can just be `!options.Enabled`.  Also, I think this check should be above the `if (options.Type === 'mob') {` line so it could apply to non-mobs if for some reason somebody has those defined.

ui/radar/radar_config.js
```diff
@@ -2,6 +2,19 @@
 
 UserConfig.registerOptions('radar', {
   options: [
+    {
+      id: 'BRankEnabled',
+      name: {
+        en: 'B-Rank enabled',
+        de: 'B-Rank enabled',
```
quisquous, line 9: Can you remove translation lines for other languages? Other people will come by and find it, but only if it's missing.

MyTechnoHunter: will do

quisquous:
Thank you!! Hopefully that wasn't too painful.  :)

PR #1706, opened 2020-08-16
raidboss: add some initial puppets' bunker triggers

quisquous:


ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,10 +1,261 @@
...
+        };
+      },
+    },
+    {
+      id: 'Puppet Aegis Aerial Support Surface Laser',
+      netRegex: NetRegexes.headMarker({ id: '0017' }),
+      condition: (data, matches) => data.me === matches.target && data.phase !== 'superior',
+      // TODO: should this be more along the lines of an imperative, like GTFO?
+      alarmText: {
+        en: '10x Sky Laser on YOU',
```
JLGarber, line 41: It occurs to me that this is similar to the meteors on Ozma. Possibly go more prescriptive on this since that has near-universal handling? "Bait lasers left/back" is nearly short enough and captures the imperative maybe a little more?

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,10 +1,261 @@
...
+      // Tell it once, but as this repeats nearly continously forever, only say it once.
+      suppressSeconds: 9999,
+      infoText: {
+        en: 'Dodge into ring gap',
+      },
+    },
+    {
+      id: 'Puppet Light Volt Array',
+      netRegex: NetRegexes.startsUsing({ source: 'Light Artillery Unit', id: '5211' }),
+      condition: (data) => data.CanSilence(),
```
JLGarber, line 81: As a general rule, are we moving to a preference for arrow functions in conditions? I have no strong preference either way, but I can definitely see the appeal of saving two lines.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,10 +1,261 @@
...
+      netRegex: NetRegexes.startsUsing({ id: '4FB4', capture: false }),
+      suppressSeconds: 5,
+      alertText: {
+        en: 'Line Stack (Point Outside)',
+      },
+    },
+    {
+      // TODO: we could theoretically get combatants here when sharp turn starts casting.
+      // Sharp Turn ability ids differentiate left/right.  We could use effect ids to know group.
+      // But how would you communicate where to stand? "behind and slightly/very right"?
```
JLGarber, line 127: I believe it's only three possibilities, right? Viewing the arena as an equilateral triangle with the single point as South, we have "stand left of point", "stand south of point",  and "stand right of point". Since the bosses seem to face counterclockwise, we could call it as "behind boss", "full (south|northwest|northeast)", "boss right flank".

Probably not worth doing for this release, but I think it's doable.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,10 +1,261 @@
...
+          return;
+
+        // The first swipe callout has been cleared to null.
+        // Deliberately skip it so that when the first swipe goes off, we call the second.
+        let swipe = data.swipe.pop();
+        if (!swipe)
+          swipe = data.swipe.pop();
+        if (!swipe)
+          return;
+        return swipe(data, matches);
```
JLGarber, line 202: I was wondering how we could handle this one. Not sure what I would have come up with, but this is definitely better!

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,10 +1,261 @@
...
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Puppet Heavy Active Laser Turret',
+      netRegex: NetRegexes.startsUsing({ source: '905P-Operated Heavy Artillery Unit', id: '4FED', capture: false }),
+      // TODO: how to even describe this.  Could have a second callout when lower disppears?
+      alertText: {
+        en: 'Upper Laser Outside -> Lower',
+      },
+    },
```
JLGarber, line 218: I do think it's better to do two separate calls here, either now or in the next release. This can be a hectic mechanic for those of us who don't play physical ranged. Maybe call the first one as "Outside on high lasers"?

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
quisquous: I think when it's a one-liner, I have been.  I was thinking once 5.3 stuff was done, I'd replace `function` with arrow functions in all triggers, if that makes sense?

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
quisquous: I think the difference from Ozma meteors is that it isn't proximity damage.  It's a chasing aoe you want to not drag through the group.  `Chasing Laser: Get Away`? I've seen two groups go to the same side, and there's plenty of room (although maybe you've seen disaster)?

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
quisquous: What text would you use for these two triggers? I'm failing to come up with anything pithy.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
JLGarber: I have seen a group where two people did the sidewalk dance when they collided. Maybe not such a big deal.

I like `Chasing Laser` though. Might switch to that instead of `Sky Laser` for future work.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
JLGarber: I'm similarly unsure. "Outside: On High Lasers" and "Move To Low Lasers" maybe? It could be we have to just accept that the user has to do the thing whether or not we can tell them exactly what it is. Actually, is there even a log entry for "low laser disappears"?

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,11 +1,514 @@
...
+
+        // Note: do not call out "avoid tank" here because there's a lot of markers going out.
+      },
+      run: (data) => delete data.busterTargets,
+    },
+    {
+      id: 'Puppet Compound 2P Three Parts Disdain',
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      condition: (data) => data.phase === 'compound',
+      response: Responses.stackOn(),
```
panicstevenson, line 454: Should this include a knockback prevention call too?

quisquous: I am iffy on that.  I don't know that I can count on everybody using knockbacks together, and then the second two hits will hit fewer people as some of the party is not knocked back?  Maybe it's worth calling anyway?  What do you think?

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,11 +1,547 @@
...
+          };
+        }
+        return {
+          en: 'Avoid tank buster',
+        };
+      },
+      run: (data) => delete data.busterTargets,
+    },
+    {
+      id: 'Puppet Superior Sliding Swipe First',
```
quisquous, line 242: This is like take 3 on this code.  It still hasn't worked yet.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -1,11 +1,547 @@
...
+    {
+      id: 'Puppet Puppet 2P Prime Blade Puppet Out Corner',
+      netRegex: NetRegexes.startsUsing({ source: 'Puppet 2P', id: '5198', capture: false }),
+      suppressSeconds: 2,
+      // TODO: have only seen this where the 4 clones stay out at 6379.4.
+      // TODO: have seen a report on a guide that the clones can teleport in??
+      alertText: {
+        en: 'Avoid Clones',
+      },
+    },
```
quisquous, line 544: I talked some with Panic about this, but I'm a little bit unsure about what the possibilities are with these puppets.  I think everything is guaranteed up until the 6379.4 block.  On that one, I have seen both 519A (with two clones ending on cardinals you need to get in) and 5198 (with four clones on corners where you need to hide middle).

https://gamerescape.com/2020/08/14/ffxiv-shadowbringers-guide-the-puppets-bunker/ suggests that maybe there's a case where you have to hide outside to dodge a "get out" sword attack?

I'm going to run this through test_timeline and see if we get any mismatches on 6379.4 and if there's another ability here (like the other "out" sword id).  Alternatively, I'd love to know if anybody sees "get in, but corners" for 6379.4.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
JLGarber: This looks like a reasonable way of handling it. I think even as annoying as copying text from `responses` is, it's the right thing to do for now.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
JLGarber: I have a feeling that knockback prevention is going to become the standard just because it's an option, but I think you're right that it's worth waiting to see. (This is sort of... parallel, maybe? to the situation with oopsy in Hades EX, where you had to special-case the meteor circles because the standard strategy is to just eat the center one. This Bunker situation is definitely worth just calling the stack for now and seeing where the community takes it.)

panicstevenson: I had the stack marker and popped knockback prevention; I didn't come close to dying, but I definitely took more damage than usual. I think most people are holding W here so if they didn't get hit on the 2nd stack they'd get hit on the 3rd. I don't have a strong preference here, just that I've seen groups get pushed into the electricity before.

quisquous: Will add this, although I'll have you know that somebody watching my stream hitting knockback replied "I love you just not giving a shit about that knockback stack".  Excuse me, I was in the first hit, and everybody else flew away!!

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
quisquous: I ran test_timeline through 20+ timelines and it's clear that the only two abilities that the clones do are 5198 (out) and 519A (in).  However, I didn't quite do enough testing to figure out how many clones are doing that attack or if there's variations.

I think I'm going to stick with the original thought, that this variation is one of two things, either "get under cardinal clone" or "get middle", and we'll know when somebody complains that there's more possibilities.  I sure haven't seen them in 5-6 iterations of this.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
quisquous: Ok, take 4 seems correct.  (oops pop vs shift)

PR #1709, opened 2020-08-17
raidboss: Add Instance Where WoL Casts Suiton

panicstevenson:
Sometimes this cast originates from the Warrior Of Light, preventing the knockback callout from triggering.

### DIFF
--- a/ui/raidboss/data/05-shb/trial/wol-ex.js
+++ b/ui/raidboss/data/05-shb/trial/wol-ex.js
@@ -443,6 +443,8 @@ const translate = (data, obj) => {
     },
     {
       id: 'WOLEx Suiton: San',
+      // It's possible for this cast to originate from the Warrior of Light instead.
+      // Allow for either so the callout isn't missed.
       netRegex: NetRegexes.startsUsing({ source: ['Warrior Of Light', 'Spectral Ninja'], id: '4EFD', capture: false }),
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/trial/wol-ex.js
```diff
@@ -432,21 +432,21 @@ const translate = (data, obj) => {
     {
       // Katon: San and Absolute Holy share markers
       id: 'WOLEx Spectral Ninja Cleanup',
-      netRegex: NetRegexes.startsUsing({ source: 'Spectral Ninja', id: '4EFD', capture: false }),
-      netRegexDe: NetRegexes.startsUsing({ source: 'Phantom-Ninja', id: '4EFD', capture: false }),
-      netRegexFr: NetRegexes.startsUsing({ source: 'Ninja Spectral', id: '4EFD', capture: false }),
-      netRegexJa: NetRegexes.startsUsing({ source: '幻光の忍者', id: '4EFD', capture: false }),
+      netRegex: NetRegexes.startsUsing({ source: ['Warrior Of Light', 'Spectral Ninja'], id: '4EFD', capture: false }),
```
quisquous, line 435: I had somebody say in the same party as me that they didn't get a callout.  So if you saw this, I wonder if there's some race where an existing combatant named Warrior of Light is turned into Spectral Ninja and immediately casts this and there's a race to update the name.

Can you leave a comment about why this is both?

quisquous:
Thanks for the comment.  Future me is very forgetful and so it'll be nice to remember why it says two things when the log only says one.

PR #1710, opened 2020-08-17
raidboss: Simplify Timeline for WoL Ex Opener

panicstevenson:
The first Imbued Stone/Holy is always Stone, followed by Holy on the next set.


quisquous:
Thanks!

PR #1717, opened 2020-08-18
Minor wolex improvements

quisquous:


ui/raidboss/data/05-shb/trial/wol-ex.js
```diff
@@ -338,25 +358,34 @@ const translate = (data, obj) => {
...
+      netRegexDe: NetRegexes.startsUsing({ source: 'Krieger Des Lichts', id: '5152', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Guerrier De La Lumière Primordial', id: '5152', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'ウォーリア・オブ・ライト', id: '5152', capture: false }),
+      condition: (data) => data.role === 'tank',
+      alarmText: {
+        en: 'TANK LB!!',
+        de: 'TANK LB!!',
+        ja: 'タンクLB!!',
+        fr: 'LB TANK !!',
+        ko: '탱크 LIMITE BREAK!!',
```
Una-p, line 371: `리미트 브레이크!!`

quisquous: This is what it says in Alex Ultimate! I'll fix it here.

ui/raidboss/data/05-shb/trial/wol-ex.js
```diff
@@ -768,8 +798,7 @@ const translate = (data, obj) => {
         'Imbued Ice/Fire': '魔法剣 (氷／火)',
         'Imbued Stone': '魔法剣 (土)',
         'Katon\\: San': '火遁の術：参',
-        '(?<! )Limit 1/3': '幻光召喚フェイス 1/3',
-        '(?<! )Limit 2/3': '幻光召喚フェイス 2/3',
+        '(?<! )Limit(?! Break)': '幻光召喚フェイス',
```
MaikoTan, line 801: I am not very sure about this translation I have added before.
Does it tell the phase change, or just a plain skill (limit break) that boss uses?

quisquous: It is for these lines, which I have changed in this patch: https://github.com/quisquous/cactbot/blob/394faba819e4399e70e058f65bb27a3934cabb61/ui/raidboss/data/05-shb/trial/wol-ex.txt#L107

e.g. "Limit -> BLM/WHM"

It used to say "Limit 2/3 -> BLM/WHM" but I found out it can be any limit break, so I removed the 2/3.  Sorry if I made the translation bad in the process.

MaikoTan: Oh, so it means the fake skill that matches "To the Limit" in the random summons?
If true, then it should be translated as "リミットチャージ", which is the skill "To the Limit" written in Japanese client.

ui/raidboss/data/05-shb/trial/wol-ex.js
Una-p: I'll check out later but it's strange, Korean's English notation is literally same as EN.
And tank translation, we write `탱커`. Old one `탱크` is appear when JP is translated into KO, it make sense but almost unused.

Una-p: I'll check out later but it's strange, Korean's English notation is literally same as EN.
And tank translation, we write `탱커`[tanker]. Old one `탱크`[tank] make sense but almost unused.

purutu: Most of Korean service user called 탱커(tanker) while global service user called 탱크(tank), 
Tank limit break is "탱커 리밋 브레이크" in Korean

Una-p: Hi purutu, I know you make TEA ko translate. But I am one of korean that play global server, few people call it [tank], too.
 I don't mean to blame you. Korean is for Koreans, I think it is right to localize it in Korean where many people used.
Yeah, I got rid of [tanker] word, think it's too long to put in alarmText.

PR #1718, opened 2020-08-18
raidboss: add knockback notice in shiva unreal

MaikoTan:
So players can use anti-knockback skills
(save poor BLM `(*/ω＼*)` )

### DIFF
--- a/ui/raidboss/data/05-shb/trial/shiva-un.js
+++ b/ui/raidboss/data/05-shb/trial/shiva-un.js
@@ -35,7 +35,7 @@
       id: 'ShivaUn Heavenly Strike',
       regex: /Heavenly Strike/,
       beforeSeconds: 5,
-      response: Responses.knockback(),
+      response: Responses.knockback('info'),
     },
   ],
   triggers: [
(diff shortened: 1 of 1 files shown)


quisquous:
Hmm.  I'm not sure about this recommendation, because if you use your knockback here, you won't have it up if you get picked for the avalanche laser.  This knockback is also prevented if there's any sort of shield.  Maybe it'd be better to tell paladins and healers to shield for the knockback?

MaikoTan:
Well, maybe that is the difference that we solve the mechanic?
We always use surecast at `Heavenly Strike`,
then come to `Avalanche`, MT would pull Shiva to the north edge, the chosen one stand at back of shiva and wait for knockback, others should stand in front or at the side of shiva to avoid that knockback laser.

MaikoTan:
Or we can make it as an `info`, not a default `alert` so user can determine how their surecast uses?

quisquous:
Yeah, info sounds good to me.  I can add a separate "shield" trigger for paladins and healers too.

(FWIW, we also pull Shiva to the edge, but then you still got knocked back if you get picked if you don't surecast.)

PR #1722, opened 2020-08-19
raidboss: fix text typo in alex ultimate

MaikoTan:
@r4ha Sorry for disturbing.

https://github.com/quisquous/cactbot/blob/c5a6c24c8690fc7ab1c31d6134dd31d28b3c9806/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js#L2240-L2247

I think the text in here should be "逃亡禁止" instead of "接触禁止". Is it my misunderstanding or just your typo here?
Waiting for your reply.


r4ha:
yes thats a my mistake. thank you!

PR #1729, opened 2020-08-23
raidboss: a12n all language

VanSlanzar:
modifications translations in replaceText + removal useless translations and missingTranslations all language.
added missing translations fr.
All the unnecessary lines probably come from a copy-paste of a12s.


JLGarber:
> All the unnecessary lines probably come from a copy-paste of a12s.

This is correct. I don't have script support for removing unused translations, and I'm too lazy to sit down and actually add it, so I normally just steal the Savage translation blocks and wait for someone else who does have it scripted to take care of it.

quisquous:
Sorry, again, but can you not remove unused lines? It's a bunch of work for me to understand whether or not these lines are actually used or not.

VanSlanzar:
as far as I'm concerned, I deleted all the lines that don't appear in the timeline.

quisquous:
@VanSlanzar I'm addressing you.  Sorry, I feel like we have discussed this several times.  Please do not remove unused translations.  Akurosia has done some of this in the past in bulk with an automated tool, and even that had some errors, and took me a long while to sort through.  But especially in a single patch with other changes where you are adding and modifying things, do not do this.

I will say again that it is not worth your time to remove unused translations.  Even if you feel like it is, I do not want to have to spend the time to understand whether or not those translations are truly unused or not.  Translations are also used for regular expressions and there aren't any tests for those.  I think it would be better to do this with an automatic test so that nobody has to think about whether it's correct or not to remove them.

VanSlanzar:
No problem, everything has been restored.

quisquous:
Thank you! Removing translations is just hard for me to review and I will (eventually) write a test to do it automatically.  I think the highest priority translation work is what `python util/find_missing_translations.py -l fr` returns.  If you need any assistance running it, please let me know.

VanSlanzar:
everything is fine, i use these 3 commands all the time.
```
python util/find_missing_translations.py -l fr
python util/check_translation.py -f 
npm run test
```

PR #1731, opened 2020-08-24
- Fix #1453

### LINKED ISSUE #1453: raidboss: Fix timeline starting when you jump off during a countdown
Currently if you do a `/countdown 15` and then jump to your death, you will wipe, but then the `Engage!` will start the timeline.

Some code needs to figure out that a wipe has occurred mid-countdown and suppress syncs to the timeline until the countdown is over.  Or some other such mechanism to prevent this.

This is pretty frustrating, as there's no real way to stop the timeline other than just wiping again or reloading raidboss.

valarnin:
This PR fixes #1453 by tracking the countdown/wipe state for the timeline and preventing the `Engage!` event from starting the timeline if there was a wipe during countdown.

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -787,6 +787,12 @@ class TimelineUI {
   }
 }
 
+const CountdownState = {
+  none: 0,
+  active: 1,
+  wipe: 2,
+};
+
 class TimelineController {
   constructor(options, ui) {
     this.options = options;
@@ -794,6 +800,12 @@ class TimelineController {
     this.dataFiles = {};
     // data files not sent yet.
     this.timelines = null;
+
+    // Fix for #1453
+    // Countdown state is tracked and if the bug is encountered (countdown, wipe, engage)
+    // then the engage line is never passed to the timeline
+    this.countdownState = CountdownState.none;
+    this.wipeRegex = Regexes.network6d({ command: '40000010' });
   }
 
   SetPopupTextInterface(popupText) {
@@ -801,6 +813,9 @@ class TimelineController {
   }
 
   SetInCombat(inCombat) {
+    // On wipe, set countdown state to Wipe if it was Active
+    if (!inCombat && this.countdownState === CountdownState.active)
+      this.countdownState = CountdownState.wipe;
     if (!inCombat && this.activeTimeline)
       this.activeTimeline.Stop();
   }
@@ -808,8 +823,38 @@ class TimelineController {
   OnLogEvent(e) {
     if (!this.activeTimeline)
       return;
-    for (let i = 0; i < e.detail.logs.length; ++i)
+    for (let i = 0; i < e.detail.logs.length; ++i) {
+      let startMatch =
+        e.detail.logs[i].match(LocaleRegex.countdownStart[this.options.ParserLanguage]);
+      if (startMatch) {
+        // If we're not in a wipe state, reset the state to countdown active
+        if (this.countdownState !== CountdownState.wipe)
+          this.countdownState = CountdownState.active;
+      } else {
+        let engageMatch =
+          e.detail.logs[i].match(LocaleRegex.countdownEngage[this.options.ParserLanguage]);
+        if (engageMatch) {
+          let previousState = this.countdownState;
+          this.countdownState = CountdownState.none;
+          // If we see an engage after a wipe, but before combat has started otherwise
+          // (e.g. countdown > wipe > face pull > engage), don't process this engage line
+          if (previousState === CountdownState.wipe)
+            continue;
+        } else {
+          // On cancel, reset the countdown state
+          let cancelMatch =
+            e.detail.logs[i].match(LocaleRegex.countdownCancel[this.options.ParserLanguage]);
+          if (cancelMatch) {
+            this.countdownState = CountdownState.none;
+          } else {
+            let wipeMatch = e.detail.logs[i].match(this.wipeRegex);
+            if (wipeMatch)
+              this.countdownState = CountdownState.wipe;
+          }
+        }
+      }
       this.activeTimeline.OnLogLine(e.detail.logs[i]);
+    }
   }
 
   SetActiveTimeline(timelineFiles, timelines, replacements, triggers, styles) {
--- a/ui/raidboss/raidboss.html
+++ b/ui/raidboss/raidboss.html
@@ -13,6 +13,7 @@
   <script src="../../resources/responses.js"></script>
   <script src="../../resources/resize_handle.js"></script>
   <script src="../../resources/timerbar.js"></script>
+  <script src="../../resources/translations.js"></script>
   <script src="../../resources/widgetlist.js"></script>
   <script src="../../resources/user_config.js"></script>
 


ui/raidboss/timeline.js
```diff
@@ -794,22 +794,67 @@ class TimelineController {
     this.dataFiles = {};
     // data files not sent yet.
     this.timelines = null;
+
+    // Fix for #1453
+    // Countdown state
```
quisquous, line 799: Instead of magic numbers can I suggest `this.countdownActive` and `this.wipedDuringCountdown`?

ui/raidboss/timeline.js
```diff
@@ -794,22 +794,67 @@ class TimelineController {
...
+    // 0 = No countdown
+    // 1 = Countdown active
+    // 2 = Wiped during countdown
+    this.countdownState = 0;
+    this.wipeRegex = / (?:21:........:400000(?:10|03)|00:0038:cactbot wipe)/;
```
quisquous, line 804: `cactbot wipe` and `4000003` victory both don't reset the timeline (right now, at least).  It also seems unlikely that somebody would beat the fight while the countdown was active as you can't start the countdown mid-combat.  The only thing that stops the timeline is leaving combat or anything that jumps to zero.  (I'm open to these things changing, just telling it like it is.)

Can I suggest `this.wipeRegex = Regexes.network6d({ command: '40000010' });` instead?

ui/raidboss/timeline.js
```diff
@@ -794,22 +794,67 @@ class TimelineController {
...
+            let wipeMatch = e.detail.logs[i].match(this.wipeRegex);
+            if (wipeMatch)
+              this.countdownState = 2;
+          }
+        }
+      }
       this.activeTimeline.OnLogLine(e.detail.logs[i]);
+      // If the previous line caused the timeline to start and we're currently in wiped state,
+      // Reset our state to account for face pull
+      if (this.activeTimeline.timebase !== null && this.countdownState === 2)
```
quisquous, line 855: This is a little awkward to check for every single log line.

It looks like engage / cancel when in state 1 sets you from state 2 to state 0 instead.  Why do you need to handle this separately?

quisquous:
Thanks for looking into this.  This is a hugely irritating bug, and I know a lot of people will be happy to see it fixed.

ui/raidboss/timeline.js
valarnin: This is specifically to handle the case of:
1. Countdown 30s
2. Countdown 25s
3. Wipe from jumping
4. Pull boss with face pull, party decides not to wipe by jumping again

ui/raidboss/timeline.js
valarnin: Actually, that's a good point. Does `cactbot wipe` even exist anymore?

I agree that just 40000010 makes sense.

ui/raidboss/timeline.js
valarnin: Not sure if you're suggesting separating this into two separate flags or if you're suggesting there should be an enum for this?

quisquous: I think either is fine, whichever you think makes this more readable.  Mostly I would just prefer names instead of numbers.

ui/raidboss/timeline.js
quisquous: What about doing this when "in combat" becomes true, if there's a countdown active *and* you've wiped, then clear the wipe flag.  (Maybe it's sufficient to just say "when combat becomes true, always clear the wipe flag?)

I think my main concern is doing this here on every log line vs just on the one log line of the "in combat" transition where this makes sense.

valarnin: How do I detect the line where the "in combat" transition happens given that it starts with Engage or with any timeline-specific sync to 0 e.g. `012s.txt`:
```
0.0 "Start" sync /00:0044:[^:]*:I am the Omega/ window 1,0
2.4 "--sync--" sync /:Omega-M:337D:/ window 3,0.5 # first auto
```

The only change of state to `Timeline` appears to be `timebase !== null`.

quisquous: Sorry, I meant the real "in combat" state vs the "is the timeline active" state.  I was thinking of adding more logic to `SetInCombat`.  My thought is that if combat starts (regardless of the timeline starting), then it's a real pull.

valarnin: `SetInCombat` for `TimelineController` only gets called with `false` (from `TimelineLoader.StopCombat`). 

The only way, as far as I can tell, that combat actually starts from a timeline perspective, is if a sync line matches. It never actually listens to any InCombat events or anything.

quisquous: I am suggesting to change that.  Rather than only calling `StopCombat`, instead pass `SetInCombat` all the way through.

valarnin: Alright, I agree that this probably doesn't make sense so I've dropped it in the latest version. This doesn't address one other really specific edge case though, which would be a countdown -> wipe -> abandon zone, which I believe would never fire the engage line and therefore never clear the state.

ui/raidboss/timeline.js
valarnin: Changed for latest commit.

ui/raidboss/timeline.js
valarnin: This has been adjusted with latest commit, but might not be quite to your liking yet so not marking this resolved yet.

ui/raidboss/timeline.js
```diff
@@ -787,29 +787,76 @@ class TimelineUI {
...
     this.ui = ui;
     this.dataFiles = {};
     // data files not sent yet.
     this.timelines = null;
+
+    // Fix for #1453
+    // Countdown state
+    // 0 = No countdown
+    // 1 = Countdown active
+    // 2 = Wiped during countdown
```
quisquous, line 808: I think the values aren't important to comment anymore.

ui/raidboss/timeline.js
```diff
@@ -787,29 +787,76 @@ class TimelineUI {
...
   }
 
   SetPopupTextInterface(popupText) {
     this.ui.SetPopupTextInterface(popupText);
   }
 
   SetInCombat(inCombat) {
+    // On wipe, set countdown state to Wipe if it was Active
+    if (!inCombat && this.countdownState === CountdownState.active)
+      this.countdownState = CountdownState.wipe;
```
quisquous, line 820: Does the `this.wipeRegex` check below already cover this block?

ui/raidboss/timeline.js
```diff
@@ -787,29 +787,76 @@ class TimelineUI {
...
     if (!this.activeTimeline)
       return;
-    for (let i = 0; i < e.detail.logs.length; ++i)
+    for (let i = 0; i < e.detail.logs.length; ++i) {
+      let startMatch =
+        e.detail.logs[i].match(LocaleRegex.countdownStart[this.options.ParserLanguage]);
+      if (startMatch) {
+        // If we're not in a wipe state, reset the state to countdown active
+        if (this.countdownState !== CountdownState.wipe)
+          this.countdownState = CountdownState.active;
```
quisquous, line 834: Should this section unconditionally set it to `active`? If you wipe and then start another countdown, that should work ok?

ui/raidboss/timeline.js
```diff
@@ -787,29 +787,76 @@ class TimelineUI {
...
+          // If we see an engage after a wipe, but before combat has started otherwise
+          // (e.g. countdown > wipe > face pull > engage), don't process this engage line
+          if (previousState === CountdownState.wipe)
+            continue;
+        } else {
+          // On cancel, reset the countdown state
+          let cancelMatch =
+            e.detail.logs[i].match(LocaleRegex.countdownCancel[this.options.ParserLanguage]);
+          if (cancelMatch) {
+            this.countdownState = CountdownState.none;
```
quisquous, line 850: I would also argue that for simplicity we could skip the cancelling check too?  If starting a countdown unconditionally set the active state, then the next time a countdown started it would work properly.  This would also solve the zoning issue.

ui/raidboss/timeline.js
valarnin: Agreed. Changing that comment block to this instead:
```javascript
    // Fix for #1453
    // Countdown state is tracked and if the bug is encountered (countdown, wipe, engage)
    // then the engage line is never passed to the timeline
```

ui/raidboss/timeline.js
valarnin: This is to specifically handle this call to `StopCombat`, from `popup-text.js` line 310, which then invokes `SetInCombat(false)` on the timeline controller:
```javascript
  SetInCombat(inCombat) {
    if (this.inCombat == inCombat)
      return;

    // Stop timers when stopping combat to stop any active timers that
    // are delayed.  However, also reset when starting combat.
    // This prevents late attacks from affecting |data| which
    // throws off the next run, potentially.
    this.inCombat = inCombat;
    if (!this.inCombat) {
      this.StopTimers();
      this.timelineLoader.StopCombat();
    }
    if (this.inCombat)
      this.Reset();
  }
```

I'm not sure if this is an edge case that is likely to be seen, but I included it in the logic anyways to be thorough.

quisquous: Ah I see, because it stops processing logs when the timeline isn't running.  Got it.  :+1:

ui/raidboss/timeline.js
valarnin: The countdown regex will match on every 5 second interval of the countdown timer. Consider this set of events and the corresponding state:

1. 20s countdown starts, `CountdownState.active`
2. 15s line, `CountdownState.active`
3. wipe, `CountdownState.wipe`
4. 10s line, `CountdownState.active`
5. 5s line, `CountdownState.active`
6. 3/2/1, not sure if these match the regex
7. Engage!, timeline starts despite wipe because state is `CountdownState.active`

Thus, the check for Engage or Cancel is what clears the state back to `CountdownState.none`. It could be rewritten to check `this.countdownState === CountdownState.none`, but some sort of check is required here.

quisquous: Oh, that's my misunderstanding.  I thought the regex included the person's name, which only appears the first time.  (It probably should, but it doesn't.)  Got it, I see this case then.  Ok, then I guess you need to do it conditionally + watch for cancelling + zone changing clearing too?

valarnin: Yeah, need all three. I tested this in-game, it's one of the things I didn't catch on first pass and had to add in.

quisquous:
It looks like changing zones doesn't stop countdowns, so that should be fine.  I'll go ahead and merge this, thanks!

PR #1732, opened 2020-08-25
add ja/cn debug text translation

MaikoTan:
@r4ha @Bluefissure Can you both double-check if I translate them into Japanese/Chinese correctly?


Bluefissure:
lgtm

quisquous:
Sorry, because it looked like you were still waiting on r4ha, I went ahead and merged the korean translations, which caused a conflict here because you were both touching the same files.

quisquous:
If r4ha has comments, we can follow up with another pull request.  Thanks for the translations!

PR #1734, opened 2020-08-25
Raidboss: fix timeline instruction language

Jaehyuk-Lee:
`this.lang` returns `undefined`
There's maybe better object to get language information, but anyway this fixes the problem.

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -615,7 +615,7 @@ class TimelineUI {
   }
 
   AddDebugInstructions() {
-    const lang = this.lang in timelineInstructions ? this.lang : 'en';
+    const lang = this.options.DisplayLanguage in timelineInstructions ? this.options.DisplayLanguage : 'en';
     const instructions = timelineInstructions[lang];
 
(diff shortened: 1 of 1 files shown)


quisquous:
Yeah, sorry.  I think `this.lang` was correct it just got initialized after that use there.  I think it needed to be done like I did in `be51731` to respect the options and use timeline language and not display language.

Thanks for letting me know about the bug and for this fix.  Sorry for the hassle!

Jaehyuk-Lee:
@quisquous 
I saw be51731 I think display language for instruction language is better. 
People usually people set their parser language same as game client language. When I was suggesting timeline's default language be Parser language, timeline shows mostly boss's skills so I thought it'd be better to show their name same as game client's.
But instruction is just an instruction so display language would be better.

quisquous:
I think I disagree a little bit? I think people expect that the options that they've set to apply here.  Most new users will not have set any options, and so will get the parser language, and expert users that have changed the language will get whatever they have set it to.  What do you think?

Jaehyuk-Lee:
ah, ok I got what you're saying. I understood. Good to remain as now.

PR #1741, opened 2020-09-01
Korean v5.2 Update

Jaehyuk-Lee:
Some timeline translation is not yet completed. I don't know them, so I'll leave it to other translators.

* Applied v5.2 Memory Signatures (just copied other server ones)
* Applied Korean text client data


Jaehyuk-Lee:
Checked "jobs" - character HP / Gatherer GP / Crafter CP shown correctly.
So, memory signature is working well I think.

quisquous:
Thank you for all of this!

PR #1743, opened 2020-09-01
plugin: Update Korean FateWatcher to use 5.2 opcodes

KattTails:
Untested as I don't have access to the KO client, but I expect it will match the intl and cn releases.


KattTails:
Whoops. Duplicate of #1741. 😅

PR #1745, opened 2020-09-01
Fix(raidboss): arrows at wrong place in trigger info text

MaikoTan:
https://github.com/quisquous/cactbot/blob/3cdb806d2b0b9cb8325e3e4fc8bdd81eb51a7170/ui/raidboss/popup-text.js#L774-L787

There are some translators translate `Tankbuster on someone` as `死刑 => someone`, but the code above would replace every arrow in the middle as `然后(then)`, so TTS would call out like `死刑然后someone (Tankbuster then someone)`.

(I think the correct TTS info should be like `Tankbuster on someone (死刑 点 someone)`.)

It confused players and should be fixed.

EDIT:
It seems there other languages have the same problem. For now I have only modified Chinese translations, because I don't know whether the other languages should be fixed or not.

### DIFF
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -142,7 +142,7 @@ let Responses = {
         de: 'Tank buster auf ' + data.ShortName(target),
         fr: 'Tank buster sur ' + data.ShortName(target),
         ja: data.ShortName(target) + 'にタンクバスター',
-        cn: '死刑 -> ' + data.ShortName(target),
+        cn: '死刑 点 ' + data.ShortName(target),
         ko: '"' + data.ShortName(target) + '" 탱버',
       };
     };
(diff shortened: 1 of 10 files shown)


quisquous:
> It seems there other languages have the same problem. For now I have only modified Chinese translations, because I don't know whether the other languages should be fixed or not.

Can you point out where this is? I think Chinese is the only language that was using "=>" for "tankbuster on" and other languages were using this for "then".

MaikoTan:
> > It seems there other languages have the same problem. For now I have only modified Chinese translations, because I don't know whether the other languages should be fixed or not.
> 
> Can you point out where this is? I think Chinese is the only language that was using "=>" for "tankbuster on" and other languages were using this for "then".

https://github.com/quisquous/cactbot/blob/760fa49f79d962f29171042bee3eb11d51ea2fbc/resources/responses.js#L714-L758

I think they means `use sleep/stun/interrupt on some boss`, right?
It seems here some arrow may be replaced as `then` either.

PR #1748, opened 2020-09-02
Raidboss: The Burn timelines/triggers

JLGarber:
(Who on earth is Tim? I feel like 80% of the time I end up having to correct `timline` to `timeline` after I write things.)

Nothing special about this one. It was remarkably painless to put together. There are more HP pushes than I expected, but the loops are all extremely straightforward.

Hailfire needs a better callout maybe? The person with the marker has to LoS, but the laser is dangerous to everyone. This was the most concise description I could think of.

On Dissonance, it's a moderately-obvious donut AoE, but I thought it was reasonable to put a trigger on it because the dangerous area is slightly smaller than the obvious donut.

Throttle is a little tricky. I checked through the logs, but unfortunately there isn't any indication which Mining Drone is the one that will explode, so we can't tell the user to stand on line XYZ. A shorter way of putting this would be welcome. (Note that this can be combined with Rock Biters and puddles, so we can't just say "Next to broken drone".)

Fog Plume is also moderately difficult. Maybe it's better to tell non-tanks to stack, so the stars are easier to evade?

Cauterize *seems* to appear only at cardinals or intercardinals, but I haven't figured out all the locations for this release. I may go back and update it later to tell the user which direction is safe, if that's even possible.


quisquous:
> (Who on earth is Tim? I feel like 80% of the time I end up having to correct `timline` to `timeline` after I write things.)

:joy_cat: 

> Hailfire needs a better callout maybe? The person with the marker has to LoS, but the laser is dangerous to everyone. This was the most concise description I could think of.

I think the other option is just to always say "get behind crystal".  The one bad thing about "avoid laser" is that it may make somebody not hide behind the crystal with the laser target when (I think? If I'm remembering correctly?) that's safe.  I vaguely remember pulling Hedetet up next to a crystal so it'd be in melee range when behind it.
 
> On Dissonance, it's a moderately-obvious donut AoE, but I thought it was reasonable to put a trigger on it because the dangerous area is slightly smaller than the obvious donut.

:+1:

Maybe worth leaving a comment here about why something with a ground marker still has a trigger?

> Throttle is a little tricky. I checked through the logs, but unfortunately there isn't any indication which Mining Drone is the one that will explode, so we can't tell the user to stand on line XYZ. A shorter way of putting this would be welcome. (Note that this can be combined with Rock Biters and puddles, so we can't just say "Next to broken drone".)

What you have looks good!

> Fog Plume is also moderately difficult. Maybe it's better to tell non-tanks to stack, so the stars are easier to evade?

I worry about saying "stack" because a lot of people treat that as "stack and plant".  We could call it "bait star explosions"? There's at least "prior art" in "bait confiteor".

> Cauterize _seems_ to appear only at cardinals or intercardinals, but I haven't figured out all the locations for this release. I may go back and update it later to tell the user which direction is safe, if that's even possible.

That would be super great if somehow you could easily figure out where the safe spot was.  I think it's not a huge deal but if it's pretty easy to figure out, that'd be a great follow-up.

Where you planning to do oopsy in a follow-up as well?

JLGarber:
> I think the other option is just to always say "get behind crystal". The one bad thing about "avoid laser" is that it may make somebody not hide behind the crystal with the laser target when (I think? If I'm remembering correctly?) that's safe. I vaguely remember pulling Hedetet up next to a crystal so it'd be in melee range when behind it.

It is safe for all players, that's true. You're right, that's probably the best option while keeping things concise.

> I worry about saying "stack" because a lot of people treat that as "stack and plant". We could call it "bait star explosions"? There's at least "prior art" in "bait confiteor".

Good point, changed to that.

> That would be super great if somehow you could easily figure out where the safe spot was. I think it's not a huge deal but if it's pretty easy to figure out, that'd be a great follow-up.

I genuinely don't know. From a quick look through the logs, there's nothing obvious, but the dragon definitely does have specific coordinates depending on where the Cauterize cast comes from. Fingers crossed etc.

And yes, I have oopsy planned, just wasn't feeling it today. That should be 1-2 days.

quisquous:
> And yes, I have oopsy planned, just wasn't feeling it today. That should be 1-2 days.

Really no hurry at all.  This was more a question of "should I go look at this while it's fresh" or "are you planning on it" and not at all "we're on a tight schedule for all this old content!"

JLGarber:
> Really no hurry at all. This was more a question of "should I go look at this while it's fresh" or "are you planning on it" and not at all "we're on a tight schedule for all this old content!"

Oh, no, no worries at all, I figured that was what was meant but I should have been clearer about it!

Unrelated, I think moving to arrow functions for single-line conditions really makes a lot of sense. So much potential for saving vertical space.

PR #1750, opened 2020-09-02
raidboss: Eden Furor Incineration Trigger fix

Jaehyuk-Lee:
Instant Incineration is physical damage.
See [here](https://fflogs.com/reports/hXYR6ZV8fTb9Djnp#fight=5&type=casts&hostility=1&source=19). (blue is magical, orange is physical)


quisquous:
Thanks!!

PR #1757, opened 2020-09-03
Radar improvements (+ a small netregexes addition)

g0shu:


ui/radar/radar.js
```diff
@@ -264,8 +270,14 @@ class Radar {
       const matches = log.match(NetRegexes.addedCombatantFull());
       if (matches) {
         const monster = this.nameToMonster[matches.groups.name.toLowerCase()];
-        if (monster)
+        if (monster) {
+          if (matches.groups.currentHp == 0) // don't fire if current HP = 0 (hunt is already dead)
```
quisquous, line 274: Can you make this `matches.groups.currentHp === '0'` since it's a string? Or alternatively `parseInt(matches.groups.currentHp) === 0`

ui/radar/radar.js
```diff
@@ -264,8 +270,14 @@ class Radar {
...
+            return;
           this.AddMonster(log, monster, matches.groups);
+          // already has been pulled if current HP < max HP when detected
+          if (matches.groups.currentHp < matches.groups.hp)
```
quisquous, line 278: These are both strings.  If max hp is 10 and currentHp is 2, unfortunately `'10' < '2'`.  I think you could say `matches.groups.currentHp !== matches.groups.hp` or something like that?

Also, I think `this.targetMonsters[name]` hasn't been set yet, if this is the first time the mob has been seen.  I think you could add the `skipPuller` flag directly into `m` here if the max hp and current hp aren't the same: https://github.com/quisquous/cactbot/blob/main/ui/radar/radar.js#L163-L197.

One answer to this would be moving all of this logic (and the hp === 0 logic) down into `AddMonster`.  That would also be nice because that's where a bunch of other checks are made to determine if the monster should really be added or not.  It'll keep the log reading function shorter.

resources/netregexes.js
```diff
@@ -165,6 +165,7 @@ var NetRegexes = {
       8: { field: 'world' },
       9: { field: 'npcNameId' },
       10: { field: 'npcBaseId' },
+      11: { field: 'currentHp' },
```
quisquous, line 168: Yeah, I'm not super happy with hp and currentHp, but not sure what to do.  I will follow up after this is merged to make this maxHp and currentHp.

quisquous:
Thanks for the fixes!! :100:

quisquous:
Thanks!!

Issue #1749, opened 2020-09-02
Remove arrows from trigger text


quisquous:
Arrows in trigger text become "then" (e.g. `stack => out`) when converted to tts.  Therefore, you shouldn't use arrows in trigger text for things like "tankbuster => target", because it will become "tankbuster then target" in tts.

@MaikoTan mentions a few things here for sleep/stun/interrupt and a few extra Korean lines: https://github.com/quisquous/cactbot/pull/1745#issuecomment-685186522

It looks like there needs to be French, Japanese, and Korean fixes.  Can I ping @VanSlanzar , @MaikoTan, and @Jaehyuk-Lee for this? Feel free to comment here about what the lines should be and I can put together a pull request (or send a pull request yourself, if that's easier).

MaikoTan:
Yes. I think in Japanese lines, they should be...
```javascript
// sleep
ja: source + ' にスリプル',
// stun
ja: source + ' にスタン',
// interrupt
ja: source + ' に沈黙',
```

Jaehyuk-Lee:
I think making new PR is better, but I'm busy this week with raid static. So... maybe next week? Or someone else can fix it.

Una-p:
There is one more similar problem that someone suggested to me.
https://github.com/quisquous/cactbot/blob/37ae15b684cb98d17013cf980cd4586653f82dca/ui/raidboss/data/03-hw/raid/a8s.js#L73-L77
https://github.com/quisquous/cactbot/blob/37ae15b684cb98d17013cf980cd4586653f82dca/ui/raidboss/data/03-hw/raid/a8s.js#L89-L93

PR #1598, opened 2020-07-24
update hw,sb KO translation

Una-P:

PR #1759, opened 2020-09-04
raidboss: delete arrow line and unify translations

Una-p:
According https://github.com/quisquous/cactbot/pull/1745#issuecomment-685186522 #1749.
The changes of following .js files are related to the current changes.
This request is included https://github.com/quisquous/cactbot/pull/1598#discussion_r459855008, too.


quisquous:
Thank you for taking care of all of these!

PR #1770, opened 2020-09-07
Oopsyraidsy: Initial Burn package

JLGarber:
Nothing special about this one. I did notice that Falling Rock was displaying for trash that got caught in it, so we'll have to look into a fix for that separately. (Probably really simple to add a check to the `damageWarn` helper for whether the target is a player.)


quisquous:
Thanks! Glad we can check this one off now.

PR #1780, opened 2020-09-10
Jobs: fix bug for Korean server

### LINKED ISSUE #1772: Jobs: UI become invisible while in instance
## Description

The Jobs overlay makes "sound" but not visible while in instance. (sound with /countdown)
This may be Korean server only bug.
I should "Refresh" overlay when I get in to an instance area to make it visible.
This happened in e5s ~ e8s (more than 20 times)
(Also, Copied Factory)
This didn't happen when I unsynced level and go o8s alone.

## Additional information

#1752(jobs: Hide CP bar when not crafting) is working fine though.

Jaehyuk-Lee:
This should fix unintended UI hiding bug.
I wonder if `(?<recipe>.*)` returns string with "x3" thing (multiple craft output)
I added `×[2-9]` in regex, but I'm not sure this is right.

This fixes #1772.

Successfully tested that this code does not make UI hide with other player's craft done.

### DIFF
--- a/resources/translations.js
+++ b/resources/translations.js
@@ -79,7 +79,7 @@ if (typeof module !== 'undefined') {
       fr: '(Vous fabriquez|Votre synthèse d\'essai pour fabriquer) (?<recipe>.*)( a été couronnée de succès!)?',
       ja: '(?<player>\\y{Name})は(?<recipe>.*)(を完成させた|の製作練習に成功した)！',
       cn: '(?<player>\\y{Name})制作(?<recipe>.*)成功！',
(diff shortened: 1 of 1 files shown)


quisquous:
> I wonder if `(?<recipe>.*)` returns string with "x3" thing (multiple craft output)
> I added `×[2-9]` in regex, but I'm not sure this is right.

None of the other regexes do this.  I think it'd be more consistent to not have this and just have `(?<recipe>.*)`.  If somebody wants to include this in the future, they can support it for every language.

Jaehyuk-Lee:
fixed it.

PR #1781, opened 2020-09-10
Various translations adding and fixing

MaikoTan:
But I am not confident with français.
@VanSlanzar sorry for disturbing, do you have time to check it once?
(And there are some brand new configurations lacks fr translations.)


quisquous:
There are only translations for the test timeline and not that important, so I will merge this for now.  Maybe @mooondark can chime in and fix these up if they're not great.

I think I would recommend in the future to separate these things out.  If you have a bunch of Japanese and Chinese translations you feel confident in and can just get merged, then put that in a separate pull request from some French translations you aren't sure about.

Issue #1773, opened 2020-09-09
Remote cactbot: non-English TTS and instance member problem, etc


Jaehyuk-Lee:
## ~~non-English TTS problem~~ (fixed)
I think this is happening to all other languages, non-English is spoken in English pronounce.
I think it's better to watch a video, I've uploaded on Youtube.
In real situation now: [Youtube Link](https://www.youtube.com/watch?v=gy8p2jbKzww)
How Korean should be read: ~~[Youtube Link](https://www.youtube.com/watch?v=jVpBybe_qs4)~~
How English speaker reads: ~~[Youtube Link](https://youtu.be/wVG-eTm5nkU)~~
EDIT: I didn't notice I was recording video while other sounds were playing, sorry.

## Korean server party member problem (guessing it's overlay plugin's bug)

Party member is not loaded even though I refreshed the web page **while in battle**. (I got into the Copied Factory.)

![Screenshot_20200909-212921_Samsung Internet](https://user-images.githubusercontent.com/4695371/92598433-c2ec8580-f2e3-11ea-801b-cd29eae77156.png)

It worked properly when I manually input player parameter. (I played DRG and input DRK's name, remote cactbot worked as I was tank)

## ~~timeline sync is not working well~~ (looks like not-an-issue)
Timeline usually work, but shown late. Cactbot overlay on ACT shows timeline as the battle starts, but remote cactbot is shown late.
Also, sometimes it does not appear at all. The real situation video I posted on the first topic, did not load timeline.

quisquous:
* Non-English TTS problem

Do ACT custom triggers sound correctly? Can you explain a little bit more what you mean by "real situation"? I assume cactbot tts normally sounds correct? If you select `Use Browser for Text to Speech` in the cactbot raidboss options, does it sound correct then? If you select the ko-KR voice on https://embed.plnkr.co/WjmlDnxcXX5dqlQI/ in your browser, does that sound correct? 

Sorry for all the questions, just trying to differentiate which things work and which don't.

* Korean server party member problem

Sorry, I am not sure what's going on here.  When I am in an instance, I can refresh the page and the party will show up.  Even if I start ACT and the remote server while in the instance, the party will still show up.  This is using the `PartyChanged` event from OverlayPlugin.  It's sent whenever an overlay first registers for the `PartyChanged` event or when it changes.

You could try debugging this by adding `console.log(JSON.stringify(e));` to these two places:
[...]

PR #1783, opened 2020-09-11
Remote cactbot: Google TTS language fix

Jaehyuk-Lee:
I don't think we need `encodeURIComponent` to `lang` parameter, so I didn't make it.

This fixes "non-English TTS problem" part of #1773.

### DIFF
--- a/ui/raidboss/browser_tts_engine.js
+++ b/ui/raidboss/browser_tts_engine.js
@@ -45,7 +45,7 @@ class GoogleTTSItem extends TTSItem {
 
 class BrowserTTSEngine {
   constructor(lang) {
-    this.googleTTSLang = lang;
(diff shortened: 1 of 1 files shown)


quisquous:
So, I think there are two different ways to do tts in this file.  I would like to get rid of this undocumented http call to return audio and only use the other SpeechSynthesis api.  I didn't realize this was the one that wasn't working for you.

Jaehyuk-Lee:
Wasn't the original code first check if your device can use SpeechSynthesis api, if not then use Google TTS?
I don't know if SpeechSynthesis api works on every devices, especially on mobile phone.

quisquous:
Newer versions of Chrome should have it.  That example web page I had you try was using the SpeechSynthesis API.

PR #1785, opened 2020-09-11
Fix DE fish reges + static fish data

Akurosia:


quisquous:
Sorry, but when you have two separate changes (one's a bugfix, the other's a content addition), could you put them in two separate pull requests? That way if the bugfix is wrong, I can just revert that whole commit on its own.

PR #1788, opened 2020-09-12
Add monster location to radar log

chunderbolt:
Makes hunt mapping a little easier.

### DIFF
--- a/ui/radar/radar.js
+++ b/ui/radar/radar.js
@@ -196,6 +196,8 @@ class Radar {
       };
       this.targetMonsters[mobKey] = m;
       this.UpdateMonsterDom(m);
+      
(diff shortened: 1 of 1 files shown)


quisquous:
Hmm, this is only going to log when somebody pulls.  I think for recording locations for hunt trains it might make sense to do the logging somewhere in https://github.com/quisquous/cactbot/blob/main/ui/radar/radar.js#L198-L200, which is the first time you see that monster and it appears in the radar.

chunderbolt:
I was using this purely for S rank mapping, where you kill B ranks to narrow down where the S can spawn. Good point though, I suppose it might as well be adapted for both usecases.

quisquous:
Thanks!

PR #1793, opened 2020-09-13
resources: Move WSServer Logic to Separate File

panicstevenson:
Moving the WSServer logic out of common.js and into its own file since
the IIFE is immediately invoked (shocking), making the code within
common unable to be tested given the current import structure of the
JavaScript code.


panicstevenson:
There's a more straightforward fix to simply exit early if running locally.

PR #1795, opened 2020-09-13
Oopsy: add click to copy

### LINKED ISSUE #1782: I wonder if we can make copy&paste works on overlay
## Question

What would you like to ask?

I would happy to add copy and paste support on overlay, especially with oopsy. I sometimes want to copy and paste warnings or death report to ffxiv client, not typing myself.

Jaehyuk-Lee:
This fixes #1782.
I think it's better to be optional since this function overwrites previous copied thing.

### DIFF
--- a/ui/oopsyraidsy/oopsyraidsy.js
+++ b/ui/oopsyraidsy/oopsyraidsy.js
@@ -304,6 +304,16 @@ class OopsyLiveList {
   MakeRow() {
     let div = document.createElement('div');
     div.classList.add('mistake-row');
+    // click-to-copy function
+    div.addEventListener('click', (e) => {
+      const str = e.target.childNodes[0].textContent;
+      const el = document.createElement('textarea');
+      el.value = str;
+      e.target.appendChild(el);
+      el.select();
+      document.execCommand('copy');
+      e.target.removeChild(el);
+    });
     this.items.push(div);
     this.container.appendChild(div);
     return div;
--- a/ui/oopsyraidsy/oopsyraidsy.js
+++ b/ui/oopsyraidsy/oopsyraidsy.js
@@ -123,6 +123,11 @@ const kPartyWipeText = {
   ko: '파티 전멸',
 };
 
+const kCopiedMessage = {
+  en: 'Copied!',
+  ko: '복사 완료!',
+};
+
 // Internal trigger id for early pull
 let kEarlyPullId = 'General Early Pull';
 
@@ -304,6 +309,7 @@ class OopsyLiveList {
   MakeRow() {
     let div = document.createElement('div');
     div.classList.add('mistake-row');
+
     // click-to-copy function
     div.addEventListener('click', (e) => {
       if (this.options.ClickToCopy) {
@@ -314,6 +320,15 @@ class OopsyLiveList {
         el.select();
         document.execCommand('copy');
         e.target.removeChild(el);
+
+        // copied message
+        const msg = document.createElement('div');
+        msg.classList.add('copied-msg');
+        msg.innerText = kCopiedMessage[this.options.DisplayLanguage] || kCopiedMessage['en'];
+        document.body.appendChild(msg);
+        setTimeout(() => {
+          document.body.removeChild(msg);
+        }, 1000);
       }
     });
     this.items.push(div);
--- a/ui/oopsyraidsy/oopsyraidsy.js
+++ b/ui/oopsyraidsy/oopsyraidsy.js
@@ -96,7 +96,7 @@ let Options = {
   },
 };
 
-let kEarlyPullText = {
+const kEarlyPullText = {
   en: 'early pull',
   de: 'zu früh angegriffen',
   fr: 'early pull',
@@ -105,7 +105,7 @@ let kEarlyPullText = {
   ko: '풀링 빠름',
 };
 
-let kLatePullText = {
+const kLatePullText = {
   en: 'late pull',
   de: 'zu spät angegriffen',
   fr: 'late pull',
@@ -125,6 +125,7 @@ const kPartyWipeText = {
 
 const kCopiedMessage = {
   en: 'Copied!',
+  fr: 'Copié !',
   ko: '복사 완료!',
 };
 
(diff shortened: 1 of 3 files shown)


Jaehyuk-Lee:
I'm concerned about which is better:
1. make optional, left-click to copy to clipboard.
2. make always work, right-click to copy to clipboard.

MaikoTan:
I think it'd be better if there is a popup notification that shows "Copied!" when copy succeed?

BTW, what about adding a new button <kbd>Copy All</kbd> at the top of the list?

Jaehyuk-Lee:
Oopsy-copy-msg

Jaehyuk-Lee:
`82:3  ✖  Expected empty line before rule   rule-empty-line-before`
I don't know what this lint wants.

Is there any problem?
https://github.com/Jaehyuk-Lee/cactbot/blob/4510cc66884fc4150574ac802dbe05a45285246d/ui/oopsyraidsy/oopsy_common.css#L78-L85
```
@keyframes fadeOut {
  50% {
    opacity: 1;
  }
  100% {
    opacity: 0;
  }
}
```

MaikoTan:
> `82:3 ✖ Expected empty line before rule rule-empty-line-before`
> I don't know what this stylelint wants.
> 
> Is there any problem?
> https://github.com/Jaehyuk-Lee/cactbot/blob/4510cc66884fc4150574ac802dbe05a45285246d/ui/oopsyraidsy/oopsy_common.css#L78-L85

Just add a empty line here

```diff
@keyframes fadeOut {
  50% {
    opacity: 1;
  }
+
 100% {
    opacity: 0;
  }
}
```

Jaehyuk-Lee:
> > `82:3 ✖ Expected empty line before rule rule-empty-line-before`
> > I don't know what this stylelint wants.
> > Is there any problem?
> > https://github.com/Jaehyuk-Lee/cactbot/blob/4510cc66884fc4150574ac802dbe05a45285246d/ui/oopsyraidsy/oopsy_common.css#L78-L85
> 
> Just add a empty line here
> 
> ```diff
> @keyframes fadeOut {
>   50% {
>     opacity: 1;
>   }
> +
>  100% {
>     opacity: 0;
>   }
> }
> ```

That change results
` 82:2  ✖  Unexpected whitespace at end of line   no-eol-whitespace`
...

Jaehyuk-Lee:
a ha fixed it thanks

Jaehyuk-Lee:
@MaikoTan 
> BTW, what about adding a new button Copy All at the top of the list?

It's easy to copy just clicked line, but it needs much more code with copying everything.
Or should make another variable for storing all texts which adds every line when those lines are created.

MaikoTan:
Well, I was just making a suggestion. It is great that you have done this work. (；´∀｀)

Jaehyuk-Lee:
btw, is there any reason translation variables have "k" on first? I think only oopsyraidy translation variable has "k" letter on first.

ui/oopsyraidsy/oopsyraidsy_config.js
```diff
@@ -222,6 +222,14 @@ UserConfig.registerOptions('oopsyraidsy', {
       type: 'checkbox',
       debugOnly: true,
     },
+    {
+      id: 'ClickToCopy',
+      name: {
+        en: 'Enable click-to-copy',
```
VanSlanzar, line 228: can you added this line, thank you.
`fr: 'Activer le clic pour copier',`

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -123,6 +123,11 @@ const kPartyWipeText = {
   ko: '파티 전멸',
 };
 
+const kCopiedMessage = {
+  en: 'Copied!',
```
VanSlanzar, line 127: can you added this line, thank you.
`fr: 'Copié !',`

ui/oopsyraidsy/oopsyraidsy_config.js
```diff
@@ -222,6 +222,14 @@ UserConfig.registerOptions('oopsyraidsy', {
...
+        ko: '클릭해서 복사하기 활성화',
+      },
+      type: 'checkbox',
```
quisquous, line 231: Can you add a default value here?   What do you think about defaulting this to `true` so that people have a way to turn it off, but it's on by default?

(For what it's worth, I am not sure this needs an option, since it requires a user interaction to do the copy.  It seems unlikely that people will be doing something that requires and copying and pasting while raiding and then click on oopsy in the middle of that operation.  If you think we need an option, then I think this is fine.  I just am a little skeptical.)

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -304,6 +309,28 @@ class OopsyLiveList {
   MakeRow() {
     let div = document.createElement('div');
     div.classList.add('mistake-row');
+
+    // click-to-copy function
+    div.addEventListener('click', (e) => {
+      if (this.options.ClickToCopy) {
```
quisquous, line 315: small style nit, but can you make this `if (!this.options.ClickToCopy) return;` and "early out" so that the rest of the function doesn't have to be indented?

quisquous (commented):
Thanks for this patch! I have two tiny comments, but it looks good.

> btw, is there any reason translation variables have "k" on first? I think only oopsyraidy translation variable has "k" letter on first.

Sorry, I am somewhat inconsistent here, but this is just a way of saying "this is an immutable, constant piece of data, often a lookup value".  The `k` is https://en.wikipedia.org/wiki/Hungarian_notation and is just carried over C++ habits from https://google.github.io/styleguide/cppguide.html#Constant_Names.

ui/oopsyraidsy/oopsyraidsy.js
Jaehyuk-Lee: Is space between é and ! intended?

Jaehyuk-Lee: I guess it's intended looking other translations.

ui/oopsyraidsy/oopsyraidsy.js
Jaehyuk-Lee: I just removed option.

ui/oopsyraidsy/oopsyraidsy_config.js
Jaehyuk-Lee: okay, I was thinking about what is better. I'll go for no option.

ui/oopsyraidsy/oopsyraidsy_config.js
Jaehyuk-Lee: I just removed this option, so didn't apply this.

Jaehyuk-Lee:
@quisquous Thanks for explanation. In my opinion, it should be consistent, use `k` or not. I think most people who don't know C++ practice, would confuse what it is.

quisquous:
> @quisquous Thanks for explanation. In my opinion, it should be consistent, use `k` or not. I think most people who don't know C++ practice, would confuse what it is.

Sorry for the inconsistency across files.  I think there are a number of naming issues I need to clean up (especially around capitalization of functions and variables), but it hasn't been a high enough priority for me to get to them.

JLGarber:
This is really cool, Jaehyuk-Lee! Useful for me, and I know a few people who are just starting their Ultimate progression back up, so this will be very helpful for them.

Jaehyuk-Lee:
@quisquous yep, there's so much patches on cactbot going on, I understand.

ui/oopsyraidsy/oopsyraidsy.js
VanSlanzar: it's normal in french to have this space.

PR #1801, opened 2020-09-15
general: Update IndexOf Usage to Includes Where Applicable

panicstevenson:
Update the general usage of `indexOf()` to instead use `includes()` where the check isn't looking for an actual index but instead checking if value exists within a collection.


quisquous:
Thank you for the cleanup!!

PR #1806, opened 2020-09-16
jobs: WHM UI

Echoring:
whm
- Lily Amount
- Next Lily Countdown
- BloodLily Stack
- DoT Tracker
- Assize Cooldown Tracker
- Lucid Cooldown Tracker

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -2439,7 +2439,7 @@ class Bars {
       }
 
       const l = lilysecondBox.parentNode;
-      if ((lily == 2 && 30 - lilysecond <= 5) | lily == 3)
+      if ((lily == 2 && 30 - lilysecond <= 5) || lily == 3)
         l.classList.add('full');
       else
         l.classList.remove('full');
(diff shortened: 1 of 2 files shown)


ui/jobs/jobs.js
```diff
@@ -2381,12 +2385,102 @@ class Bars {
...
+      id: 'whm-procs-lucid',
+      fgColor: 'whm-color-lucid',
+    });
+
+    // BloodLily Guage
+    const stacksContainer = document.createElement('div');
+    stacksContainer.id = 'whm-stacks';
+    this.addJobBarContainer().appendChild(stacksContainer);
+    const bloodlilyContainer = document.createElement('div');
+    bloodlilyContainer.id = 'whm-stacks-bloodlily';
```
quisquous, line 2413: I think my one nitpick with this UI is that the bloodlily boxes don't line up with the hp and mp bar.  I think this is also true of the smn ruin4 stacks.  It'd be nice to make them all the same width, maybe using flexbox with `justify-content: space-between` to spread them out to the right length.  I think in this case (and the smn case), I am not sure why `whm-stacks-bloodlily` needs to exist.  It seems like all of the stacks could be added directly to `whm-stacks`? It's an improvement to think about for the future.

PR #1807, opened 2020-09-16
fix(jobs): full-size bracket in countdownStartRegex

MaikoTan:
They use full-size (2-bytes) bracket in CN.
So the previous regex would not work.


quisquous:
Sorry about that!

PR #1809, opened 2020-09-17
Refactor(jobs): Hide CP bar when not crafting (#1752)

MaikoTan:
# Description
This PR refactored @Echoring 's previous PR (#1752)

- Add `count` matching
- A few changes make it faster in regex matching

Thank @Jaehyuk-Lee for explaining the Korean language grammar.

## Note

I have tested in JP/EN/FR.
DE *should* work, but I am not very sure if there are exceptions what I have not noticed.
Waiting for @Echoring and @Jaehyuk-Lee for testing in CN/KO.

## How To Test

Download these two files below and replace your local files in the same path (don't forget to backup your files):

[resources/translations.js](https://github.com/MaikoTan/cactbot/raw/refacter_jobs_crafting/resources/translations.js)

[ui/jobs/jobs.js](https://github.com/MaikoTan/cactbot/raw/refacter_jobs_crafting/ui/jobs/jobs.js)


Echoring:
There're quotation marks missing in CN.
It should be `“\ue0bb(?<recipe>.*)”`

Jaehyuk-Lee:
Korean translation looks fine, I didn't try it though.

resources/translations.js
```diff
@@ -63,40 +63,66 @@ if (typeof module !== 'undefined') {
...
-      fr: '(La synthèse échoue\\.{3})|(Votre synthèse d\'essai pour fabriquer (?<recipe>.*) s\'est soldée par un échec\\.{3})',
-      ja: '(?<player>\\y{Name})は(製作|(?<recipe>.*)の製作練習)に失敗した……',
+      en: 'Your synthesis fails!',
+      de: 'Deine Synthese ist fehlgeschlagen!',
+      fr: 'La synthèse échoue\\.{3}',
+      ja: '(?<player>\\y{Name})は製作に失敗した……',
       cn: '(?<player>\\y{Name})制作失败了……',
       ko: '제작에 실패했습니다……\\.',
     },
+    trialCraftingFail: {
```
quisquous, line 106: Thanks.  I think this makes this easier to read to split out like this.

ui/jobs/jobs.js
```diff
@@ -1075,13 +1075,20 @@ class Bars {
...
     this.countdownStartRegex = LocaleRegex.countdownStart[lang] || LocaleRegex.countdownStart['en'];
     this.countdownCancelRegex = LocaleRegex.countdownCancel[lang] || LocaleRegex.countdownCancel['en'];
     this.craftingStartRegex = LocaleRegex.craftingStart[lang] || LocaleRegex.craftingStart['en'];
     this.craftingFinishRegex = LocaleRegex.craftingFinish[lang] || LocaleRegex.craftingFinish['en'];
     this.craftingFailRegex = LocaleRegex.craftingFail[lang] || LocaleRegex.craftingFail['en'];
     this.craftingCancelRegex = LocaleRegex.craftingCancel[lang] || LocaleRegex.craftingCancel['en'];
+    this.trialCraftingStartRegex = LocaleRegex.trialCraftingStart[lang] || LocaleRegex.trialCraftingStart['en'];
+    this.trialCraftingFinishRegex = LocaleRegex.trialCraftingFinish[lang] || LocaleRegex.trialCraftingFinish['en'];
+    this.trialCraftingFailRegex = LocaleRegex.trialCraftingFail[lang] || LocaleRegex.trialCraftingFail['en'];
+    this.trialCraftingCancelRegex = LocaleRegex.trialCraftingCancel[lang] || LocaleRegex.trialCraftingCancel['en'];
```
quisquous, line 1091: There's a lot of crafting regexes now here.  What about combining them so that the logic below is easier? e.g.
```
const trialCraftingStartRegex = ...
const trialCraftingFinishRegex = ...
...

this.craftingStartRegex = Regexes.anyOf(craftingStartRegex, trialCraftingStartRegex);
this.craftingStopRegex = Regexes.anyOf(craftingFailRegex, craftingCancelRegex, trialCraftingFailRegex, trialCraftingCancelRegex);
```

Then you can just check against `this.craftingStartRegex` and `this.craftingStopRegex` below, which should make that logic simpler.

MaikoTan: Yeah. But `Regexes.anyOf` just combine them with `(?:exp1|exp2)`, it will cause *SyntaxError* when there are named capture groups in all the regexes.
image

ui/jobs/jobs.js
```diff
@@ -1493,7 +1499,7 @@ class Bars {
       // Because thresholds are nonmonotonic (when finishing a combo)
       // be careful about setting them in ways that are visually poor.
       if (eyeBox.value >= oldThreshold &&
-          eyeBox.value >= newThreshold)
```
quisquous, line 1496: Why are you changing these? I feel like this and line 1502 are more readable like this.  If I could figure out how to make eslint enforce this indentation, I would turn on a rule.

ui/jobs/jobs.js
quisquous: Oh, whoops.  Sorry about that suggestion, then.  I think the way you've added a helper function makes this more readable.

ui/jobs/jobs.js
```diff
@@ -2689,6 +2695,57 @@ class Bars {
...
+    const container = document.getElementById('jobs-container');
+
+    const anyRegexMatched = (line, array) => {
+      for (const regex of array) {
+        if (regex.test(line))
+          return true;
+      }
+      return false;
+    };
+    const undefinedOrEqual = (value, expected) => {
```
quisquous, line 2709: This is only used once, so maybe doesn't need to be a function?

quisquous (commented):
Looks good in general, with two small comments.  Thanks for making this better!

ui/jobs/jobs.js
MaikoTan: Oops, it might be changed by my auto-formatting tools.
I'll revert it back later.

ui/jobs/jobs.js
MaikoTan: Sure, I agree too.

PR #1812, opened 2020-09-19
4.X & 5.X timeline completely CN translations.

ShadyWhite:
Remove `missingTranslation: true` lines.


ui/raidboss/data/04-sb/raid/o1s.js
```diff
@@ -155,10 +155,9 @@
     },
     {
       'locale': 'cn',
-      'missingTranslations': true,
       'replaceSync': {
         'Alte Roite': '老者',
-        'Ball Of Fire': '火球',
+        'Ball Of Fire': '火焰球',
```
Akurosia, line 160: Are you sure about this change?
all CN translations say, it should be `火球` 
Your translation is the JA one inside the CN block. Can you have a look at it?

ShadyWhite: image
image
`火球` might be 'Ball Of Fire'‘s CN translation in other place.

PR #1813, opened 2020-09-19
2.X & 3.X timeline completely CN translations.

ShadyWhite:
Remove `missingTranslation: true` lines.


quisquous:
Thanks for all the translations!

PR #1814, opened 2020-09-20
Raidboss: full JA translation in HW

MaikoTan:
And a quick question: Is the timeline of A7s not completed? Why are there so many empty lines causing test failed


quisquous:
> And a quick question: Is the timeline of A7s not completed? Why are there so many empty lines causing test failed

Yeah, a7s is not completed.  I need to go back and get more logs because it has different branches, depending on which jail sequence is used.  I think it is better not to translate the timeline at this point until it gets improved. 
 https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/03-hw/raid/a7s.js#L44 should prevent tests from complaining about missing lines.  Which tests are failing for you?

ui/raidboss/data/03-hw/pvp/shatter.js
```diff
@@ -77,12 +85,19 @@
...
       'replaceSync': {
-        'The icebound tomelith A1 activates and begins to emit heat': '冰封的石文A1启动了，冰块变得脆弱了！',
-        'The icebound tomelith A2 activates and begins to emit heat': '冰封的石文A2启动了，冰块变得脆弱了！',
-        'The icebound tomelith A3 activates and begins to emit heat': '冰封的石文A3启动了，冰块变得脆弱了！',
-        'The icebound tomelith A4 activates and begins to emit heat': '冰封的石文A4启动了，冰块变得脆弱了！',
+        'The icebound tomelith A1 activates and begins to emit heat': 'アイスドトームリスA1号基が起動し、氷がもろくなった！',
+        'The icebound tomelith A2 activates and begins to emit heat': 'アイスドトームリスA2号基が起動し、氷がもろくなった！',
+        'The icebound tomelith A3 activates and begins to emit heat': 'アイスドトームリスA3号基が起動し、氷がもろくなった！',
+        'The icebound tomelith A4 activates and begins to emit heat': 'アイスドトームリスA4号基が起動し、氷がもろくなった！',
+        'locale': 'cn',
```
quisquous, line 94: I think `locale: cn` needs to be in its own object.

MaikoTan:
when I remove the `missingTranslation: true,` then running `python test/run_test.py`
It shows
`cactbot/ui/raidboss/data/03-hw/raid/a7s.js:locale ja:no translation for replaceText ''`

I think this is caused by these empty lines:
https://github.com/quisquous/cactbot/blob/4c8611354229b524ce828c185f158a35dd5ed373/ui/raidboss/data/03-hw/raid/a7s.txt#L84

ui/raidboss/data/03-hw/pvp/shatter.js
MaikoTan: Oh sorry, I resolved conflict in browser but forgot to add it back!

quisquous:
> when I remove the `missingTranslation: true,` then running `python test/run_test.py`

Oh, ok, when you remove `missingTranslations`.  Thanks.

PR #1816, opened 2020-09-20
OopsyRaidsy:Fix enable check for early pulls

ErisenIrioth:
If the trigger is enabled for early pulls then they SHOULD be logged.


quisquous:
Whooooops.  Thanks for the fix.

PR #1818, opened 2020-09-20
Raidboss: ARF minor fixes

JLGarber:
(Multiple commits for ease of viewing changes.)

Mostly self-explanatory. The timeline changes on Regula are because it is juuuust possible to have an overlap if the group kills the first turret set super-fast and Regula goes into the second set immediately. The removed lines at the end are because they would never show up in the overlay.

Casing on Height Of Chaos normally wouldn't matter, but it did here. Oops! (I fiexed the casing on End Of Days while I was in there.)

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.txt
+++ b/ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.txt
@@ -101,6 +101,11 @@ hideall "--sync--"
 440.4 "Magitek Slug"
 444.0 "Magitek Slug"
 444.7 "Aetherochemical Grenado"
+449.9 "Aetherochemical Grenado"
+450.4 "Magitek Slug"
+454.1 "Magitek Slug"
+455.1 "Aetherochemical Grenado"
+457.8 "Magitek Slug"
 
 #~~~~~~~~~~~#
 # Harmachis #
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/03-hw/dungeon/aetherochemical_research_facility.txt
```diff
@@ -101,11 +101,6 @@ hideall "--sync--"
 440.4 "Magitek Slug"
 444.0 "Magitek Slug"
 444.7 "Aetherochemical Grenado"
-449.9 "Aetherochemical Grenado"
```
quisquous, line 104: I'm not sure what you mean by "will never show up in the overlay".  The default is 30 seconds, and you can increase that arbitrarily.  So, even at the default, this would show 429 through 459?

Maybe I should file a feature request for this, but I think there'd be some advantage in having a new timeline entry that you could use instead of `jump` called ... `loop`? And if you did that, then the timeline code would automatically include this loop segment lookahead.

JLGarber: That would normally be correct, yes, However, the max default is *also* six lines displayed at a time. My general impression has been that if either condition (30s of timeline, OR six lines) was reached, the rest could be removed. I'm fine not removing it, it was just a thing I came across in passing.

quisquous: You're correct as to when the rest are removed.  However, you can increase both of those limits, and so I've been doing minimum of 30 seconds (ignoring counts) for loops.

JLGarber: Perfectly fair!

PR #1824, opened 2020-09-25
Raidboss: A8N Initial timelines/triggers

JLGarber:
Overall this one wasn't terribly difficult. Most of the triggers were shamelessly ~stolen~ copied from Savage, with appropriate edits.

A few issues remain:

1. Discoid doesn't put head markers out on normal, and for some ridiculous reason the orbs don't do tether log lines. Maybe just a basic trigger to "eat other players' orbs"?

2. The timeline for the robot pairs is a bit fragile depending on where in the rotation each one dies. Unfortunately it's not completely possible to fix this. We have three options. One, we could just unroll the loop for each pair and hardcode three rotations or something, on the assumption that *surely* they will be dead by then. Two, we could just do a single loop for each pair, with maybe a bit of a tail to it, and then just let things drop off, much as Savage does. (This is my preferred option.) Three, we could always just not have the timeline running during the robot pairs.

3. Flarethrower comes up at the same time as Long Needle during parts of Brute Justice's rotation. I think it's probably worth setting a trigger to collect the active tank and then warn them not to stack, since that's a *really* dangerous combination if the healers aren't ready.

I probably won't be getting to oopsy on this for a while. It's tough to sit down and focus these days!


ui/raidboss/data/03-hw/raid/a8n.js
```diff
@@ -0,0 +1,518 @@
...
+  triggers: [
+    {
+      id: 'A8N Megabeam Onslaughter',
+      netRegex: NetRegexes.startsUsing({ source: 'Onslaughter', id: '1732', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Schlachter', id: '1732', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Attaqueur', id: '1732', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'オンスローター', id: '1732', capture: false }),
+      netRegexCn: NetRegexes.startsUsing({ source: '突击者', id: '1732', capture: false }),
+      netRegexKo: NetRegexes.startsUsing({ source: '맹습자', id: '1732', capture: false }),
+      // Insert sound effect from Arthars here.
```
quisquous, line 45: Nice copy and paste :P

quisquous (commented):
> Overall this one wasn't terribly difficult. Most of the triggers were shamelessly ~stolen~ copied from Savage, with appropriate edits.
> 
> A few issues remain:
> 
> 1. Discoid doesn't put head markers out on normal, and for some ridiculous reason the orbs don't do tether log lines. Maybe just a basic trigger to "eat other players' orbs"?

This is probably another case of actor spawning with tether information in the actor spawn data, like t12 birds.  What happens if you eat your own orb in normal mode, will you live? If so, maybe fine to not call anything out if it's complicated?

> 2. The timeline for the robot pairs is a bit fragile depending on where in the rotation each one dies. Unfortunately it's not completely possible to fix this. We have three options. One, we could just unroll the loop for each pair and hardcode three rotations or something, on the assumption that _surely_ they will be dead by then. Two, we could just do a single loop for each pair, with maybe a bit of a tail to it, and then just let things drop off, much as Savage does. (This is my preferred option.) Three, we could always just not have the timeline running during the robot pairs.

Yeah, I think your #2 is the choice I would pick too.

> 3. Flarethrower comes up at the same time as Long Needle during parts of Brute Justice's rotation. I think it's probably worth setting a trigger to collect the active tank and then warn them not to stack, since that's a _really_ dangerous combination if the healers aren't ready.

Seems reasonable! Did you want to do that, or at least leave a TODO?

> I probably won't be getting to oopsy on this for a while. It's tough to sit down and focus these days!

Ain't that the truth.  /o\

ui/raidboss/data/03-hw/raid/a8n.js
JLGarber: It seemed appropriate to leave in even if I'm not Arthars' biggest fan.

JLGarber:
> This is probably another case of actor spawning with tether information in the actor spawn data, like t12 birds. What happens if you eat your own orb in normal mode, will you live? If so, maybe fine to not call anything out if it's complicated?

It's livable. It does damage proportional to the player's HP, approximately 25%. That being the case, I think you're right, we can just ignore it.

. [FLARETHROWER STUFF] Seems reasonable! Did you want to do that, or at least leave a TODO?

Yes, I'll get this all updated a little later today.

ui/raidboss/data/03-hw/raid/a8n.js
```diff
@@ -169,10 +184,19 @@
...
     },
     {
-      id: 'A8N Long Needle',
+      id: 'A8N Long Needle Party',
       netRegex: NetRegexes.headMarker({ id: '003E' }),
+      condition: () => data.me !== data.bruteTank,
       response: Responses.stackOn(),
     },
+    {
+      id: 'A8N Long Needle Active Tank',
```
quisquous, line 193: Hmm, not every long needle comes with a flarethrower.  Also in thinking about this, I guess saying `tank cleave` here conflicts with the flarethrower timeline trigger too?  Do you have any thoughts about how to reconcile all of that?

JLGarber: I don't think Flarethrower is a conflict, really. Looking at this more closely, we could probably just check whether we're on an even or odd count for Long Needle? Something like

```
{
  condition: function(data) {
    data.needleCount = data.needlecount || 0;
    return needleCount % 2 === 1 && [OTHER EXISTING CONDITIONS];
  }
 //do stuff
  run: function(data) {
    data.needlecount += 1;
  }
}
```
And then for the party Long Needle, we can check whether the user is a tank, and if they are, then reverse the condition from the tank Long Needle trigger. It's not trivial, but it's completely doable.

quisquous: Is there a possibility for a phase skip here that would skip one such that a counter wouldn't work? Do we need something like a flarethrower timeline trigger that sets a `flarethrowerSoon = true` and then maybe say "Don't Stack" if that's true? Sorry to make this more complicated.

JLGarber: Well, I can't say for certain that there isn't, but all of my logs seem to indicate that the initial block from Brute Justice appearing until the intermission is 100% fixed. I've never seen an early push into the intermission, and indeed it's highly possible to completely skip it in party finder.

That aside, maybe a better way to handle it would be to enable/disable based on usage of Super Jump and J-Kick? The only dangerous Long Needle happens after J-Kick and before Super Jump, OR at the very start of the rotation. So we can start with the Long Needle "stay out" active, and then once we see Super Jump, disable it until J-Kick?

quisquous: That seems like a reasonable approach if you leave some comments about what's going on.  Thanks!

JLGarber: Alright, thanks for the feedback on this. I think this should better handle those possibilities.

quisquous:
Thanks for all of this!

PR #1827, opened 2020-09-27
Raidboss: Add/Fix translations in SB

MaikoTan:
- add/fix translations in SB alliance/eureka/raid
- unify some translations to be consistent


quisquous:
Thanks for all of these!!

PR #1832, opened 2020-09-28
Raidboss: Add JA missing translations in various files

MaikoTan:


ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -909,16 +909,15 @@
...
     {
       'locale': 'ja',
-      'missingTranslations': true,
       'replaceSync': {
-        '724P-Operated Superior Flight Unit \\(A-Lpha\\)': '７２４Ｐ：強化型飛行ユニット［A-lpha］',
-        '767P-Operated Superior Flight Unit \\(B-Eta\\)': '７６７Ｐ：強化型飛行ユニット［B-eta］',
-        '772P-Operated Superior Flight Unit \\(C-Hi\\)': '７７２Ｐ：強化型飛行ユニット［C-hi］',
+        '724P-Operated Superior Flight Unit \\\\\\(A-Lpha\\\\\\)': '７２４Ｐ：強化型飛行ユニット［A-lpha］',
+        '767P-Operated Superior Flight Unit \\\\\\(B-Eta\\\\\\)': '７６７Ｐ：強化型飛行ユニット［B-eta］',
+        '772P-Operated Superior Flight Unit \\\\\\(C-Hi\\\\\\)': '７７２Ｐ：強化型飛行ユニット［C-hi］',
```
MaikoTan, line 915: I am not very confident about these regex lines.
Because in the txt file it is `724P-Operated Superior Flight Unit \(A-Lpha\)` ( with "\" before "(" )
[Regex101](https://regex101.com/) tell me that regex should be `724P-Operated Superior Flight Unit \\\(A-Lpha\\\)`
image

If mine is right, then those regexes in DE/FR should be fixed.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
```diff
@@ -952,7 +954,7 @@
         'Formation: Sharp Turn': '連携：転回斬撃',
         'Formation: Sliding Swipe': '連携：突進斬撃',
         'Four Parts Resolve': '四連断重撃',
-        '(?<!(Maneuver|Firing Order): )High-Powered Laser': '高出力レーザー',
+        '(?<!: )High-Powered Laser': '高出力レーザー',
```
MaikoTan, line 957: BTW, the previous regex caused python error when running `python run_tests.py`: "look-behind requires fixed-width pattern" error, so I change it like this.

ui/raidboss/data/05-shb/alliance/the_puppets_bunker.js
JLGarber: The `replacesync` regex is intended to match the text string present in the timeline file, and thus you are correct that both the `\` escape and `()` parentheses should be escaped.

MaikoTan: Great!
But I think it is better to separate a single pull request to fix DE/FR things.

quisquous: Yes, I think this is correct as well.  Sorry for how confusing this looks.  Other alternatives to consider in the future are `772P-Operated Superior Flight Unit .C-Hi.`

PR #1838, opened 2020-09-29
jobs: Hide UI except hp/mp bar in pvp zone

### LINKED ISSUE #1672: Jobs: Hide/unload UI when combat in PVP area
Jobs UI will appear when you get in combat no matter PVE or PVP.
PVP is completely different from PVE. Showing UI is not necessary and may cause additional interference.
Can it be hide or unload in PVP specfic zone like the "well fed" icon?

Echoring:
As once discussed in issue #1672 also closes #1672  
Early return from `UpdateJob()` meet some error on zone change, so just hide them.
Seems calling `UpdateJob()` in `OnChangeZone(e)` will cause something wrong.

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -1078,7 +1078,7 @@ class Bars {
     this.abilityFuncMap = {};
 
     this.contentType = 0;
-    this.isPVPZone = false;
+    this.notPVPZone = true;
     this.crafting = false;
 
     const lang = this.options.ParserLanguage;
@@ -1108,13 +1108,16 @@ class Bars {
     ];
   }
 
-  uiDisplay(options) {
+  UpdateUIVisibility(visible) {
     const bars = document.getElementById('bars');
     if (bars) {
       const barList = bars.children;
       for (const bar of barList) {
         if (bar.id === 'hp-bar' || bar.id === 'mp-bar') continue;
-        bar.style.display = options;
+        if (visible === true)
+          bar.style.display = '';
+        if (visible === false)
+          bar.style.display = 'none'
       }
     }
   }
@@ -1328,10 +1331,7 @@ class Bars {
     this.UpdateJobBarGCDs();
 
     // Hide UI except HP and MP bar if in pvp area.
-    if (this.isPVPZone === true)
-      this.uiDisplay('none');
-    else
-      this.uiDisplay('');
+    this.UpdateUIVisibility(this.notPVPZone);
   }
 
   validateKeys() {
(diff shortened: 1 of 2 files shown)


ui/jobs/jobs.js
```diff
@@ -2786,6 +2803,19 @@ class Bars {
 
     for (const func of this.changeZoneFuncs)
       func(e);
+
+    this.isPVPZone = 0;
```
quisquous, line 2807: Can this be `false` and `true` instead of `0` and `1`?

ui/jobs/jobs.js
```diff
@@ -2786,6 +2803,19 @@ class Bars {
...
+    if (zoneInfo) {
+      // 6 => pvp content, 250 => Wolves' Den Pier
+      if (zoneInfo.contentType == 6 || e.zoneID == 250)
```
quisquous, line 2810: `if (zoneInfo.contentType === ContentType.Pvp || e.zoneID === ZoneId.WolvesDenPier)`

ui/jobs/jobs.js
```diff
@@ -1108,6 +1108,17 @@ class Bars {
     ];
   }
 
+  uiDisplay(options) {
```
quisquous, line 1111: What about instead of `uiDisplay(options)` instead this was `SetVisible(visible)` and you could call `SetVisible(true)` or `SetVisible(false)`? I think that would be more readable than passing `'none'` and `''` in two separate places.

Alternatively, you could have `UpdateUIVisibility()` and have it check `this.isPVPZone` internally rather than passing anything.

ui/jobs/jobs.js
```diff
@@ -1108,6 +1108,17 @@ class Bars {
...
+    const bars = document.getElementById('bars');
+    if (bars) {
+      const barList = bars.children;
+      for (const bar of barList) {
+        if (bar.id == 'hp-bar' || bar.id == 'mp-bar') continue;
```
quisquous, line 1116: `===` unless there's a reason not to.  I know code isn't consistent elsewhere and it's not safe to globally change them all, but try to use `===` in new code.

quisquous (commented):
Thanks for taking a look at this.  I think this is good overall, with just some readability comments.

What error is happening if you early return from `UpdateJob` or call `UpdateJob` in `OnChangeZone`? Could you give some more details? I think that approach might be more clear in the code, but I don't understand what's going wrong.

Echoring:
> What error is happening if you early return from UpdateJob or call UpdateJob in OnChangeZone?

If early return from UpdateJob, I need to call UpdateJob in OnChangeZone to hide UI when enter a pvp zone.
But it result in error when change zone like this:
`Uncaught SyntaxError: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty. (Source: file:///C:/Users/DXL/Desktop/ACT/Plugins/ACT.OverlayPlugin/cactbot/ui/jobs/jobs.js, Line: 1155)`
Line 1155 is
`barsLayoutContainer.classList.add(this.job.toLowerCase());`
MP bar can not update and fix to 0 when this error happen.

ui/jobs/jobs.js
Echoring: `Uncaught ReferenceError: ZoneId is not defined`, Maybe we need import it somewhere?

quisquous:
Thanks!!

PR #1840, opened 2020-10-03
fix(Raidboss): directions not fully translated

MaikoTan:
The previous way did not solve the problem itself,
instead of writing complecated regex, it is better
to change the text itself.

### DIFF
--- a/ui/raidboss/data/02-arr/raid/t4.txt
+++ b/ui/raidboss/data/02-arr/raid/t4.txt
@@ -9,7 +9,7 @@
 123.0 "Dreadnaught (NW)" sync / 03:........:Added new combatant Clockwork Dreadnaught\./ window 130,0
 123.0 "4x Bug (S)"
 
-183.0 "2x Rook (E/W center)" sync / 03:........:Added new combatant Spinner-rook\./ window 190,0
+183.0 "2x Rook (E/W)(center)" sync / 03:........:Added new combatant Spinner-rook\./ window 190,0
 183.0 "4x Bug (outside)"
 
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/02-arr/raid/t4.txt
```diff
@@ -9,7 +9,7 @@
 123.0 "Dreadnaught (NW)" sync / 03:........:Added new combatant Clockwork Dreadnaught\./ window 130,0
 123.0 "4x Bug (S)"
 
-183.0 "2x Rook (E/W center)" sync / 03:........:Added new combatant Spinner-rook\./ window 190,0
+183.0 "2x Rook (E/W)(center)" sync / 03:........:Added new combatant Spinner-rook\./ window 190,0
```
quisquous, line 12: I think the original line was intended.  There's two rooks, and they appear east and west of the center, but not very far east or west.  I think it'd be fine to just make this `(E/W)` or `(center)`.  I think `(E/W)(center)` is less understandable, though.

MaikoTan: Sure, maybe just `(center)` is ok.

JLGarber: I can't remember the spawn locations, but I think that if they are close enough to be picked up with a normal tank AoE, it should be labeled as `center`. If not, give the directions so the tank can anticipate their locations.

quisquous: Yeah, I think these are far enough apart that one tank can't pick both of them up.  I'll change this to be E/W and then merge this.  Thanks for the feedback.

PR #1842, opened 2020-10-05
Raidboss: O12S minor timeline fix

JLGarber:
With modern unsynced DPS output, it's possible to enter the duo phase well before the developers intended us to. These windows ensure that the raidboss timeline will always be at the correct place.

(Strictly speaking only the Firewall sync is necessary, but it's nice to have the non-targetable intermission displayed too.)


quisquous:
Thanks for this!

PR #1844, opened 2020-10-05
CI: Some Python related simplification

MaikoTan:
- Add `requirements.txt` for pip installing ( although there are only 4 items for now )
- Add virtual environment-related files into .gitignore
- Add pip cache in python-lint workflow ( theoretically would be a little faster? )

### DIFF
--- a/.github/workflows/python-lint.yml
+++ b/.github/workflows/python-lint.yml
@@ -24,9 +24,25 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: '3.7'
+
+      # Cache pip dependencies
+      # See also: https://github.com/actions/cache/blob/main/examples.md#python---pip
+      - name: Get pip cache dir
+        id: pip-cache
+        run: |
+          echo "::set-output name=dir::$(pip cache dir)"
+
+      - name: pip cache
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.pip-cache.outputs.dir }}
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+
       - name: Install Dependencies
         run: |
-          pip install black==19.10b0
+          pip install -r requirements.txt
       - name: black **/*.py --line-length 100 --check --diff
         run: |
           black **/*.py --line-length 100 --check --diff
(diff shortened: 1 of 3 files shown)


quisquous:
Thanks for this pull request.  I am a little bit short on time to look at pull requests that require a little bit more thought on my part, and so I'll try to get to this in the next few days.

panicstevenson:
If we're installing the same install command prior to both executions, does it make sense to have a separate job for installing dependencies that both the pylint and black jobs `need`?

MaikoTan:
> If we're running the same install command prior to both executions, does it make sense to have a separate job for installing dependencies that both the pylint and black jobs `need`?

I don't very understand what you are referring... ≧ ﹏ ≦

If you are referring to that two jobs using the same install command. 
My thought is, when we need to update the version of pylint and black, we can just change it via `requirements.txt` file.
And since I added pip cache part, it can directly read from cache so even faster than before.

panicstevenson:
>I don't very understand what you are referring... ≧ ﹏ ≦

No worries! I can try to explain a little clearer.

> My thought is, when we need to update the version of pylint and black, we can just change it via requirements.txt file.

Agreed.

> And since I added pip cache part, it can directly read from cache so even faster than before.

Agreed as well.

This code:

```yaml
      # Cache pip dependencies
      # See also: https://github.com/actions/cache/blob/main/examples.md#python---pip
      - name: Get pip cache dir
        id: pip-cache
        run: |
          echo "::set-output name=dir::$(pip cache dir)"
      - name: pip cache
        uses: actions/cache@v2
        with:
          path: ${{ steps.pip-cache.outputs.dir }}
          key: cactbot-pip-${{ hashFiles('**/requirements.txt') }}
          restore-keys: |
            cactbot-pip-
```
is in `.github/workflows/python-lint.yml` twice. Anyone updating this code will update it in two places.

Does it make sense to create a new job within `.github/workflows/python-lint.yml` ?
```yaml
jobs:
  install:
    runs-on: ubuntu-latest
    steps:
      - name: Checkout Code
        uses: actions/checkout@v1
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.7'
      # Cache pip dependencies
      # See also: https://github.com/actions/cache/blob/main/examples.md#python---pip
      - name: Get pip cache dir
        id: pip-cache
        run: |
          echo "::set-output name=dir::$(pip cache dir)"
      - name: pip cache
        uses: actions/cache@v2
        with:
          path: ${{ steps.pip-cache.outputs.dir }}
          key: cactbot-pip-${{ hashFiles('**/requirements.txt') }}
          restore-keys: |
            cactbot-pip-
      - name: Install Dependencies
        run: |
          pip install -r requirements.txt
  black:
    runs-on: ubuntu-latest
    needs: install
    steps:
      - name: Checkout Code
        uses: actions/checkout@v1
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.7'
      - name: black **/*.py --line-length 100 --check --diff
        run: |
          black **/*.py --line-length 100 --check --diff
  pylint:
    runs-on: ubuntu-latest
    needs: install
    steps:
      - name: Checkout Code
        uses: actions/checkout@v1
      - name: Setup Python
        uses: actions/setup-python@v2
        with:
          python-version: '3.7'
      - name: Add Custom Problem Matcher
        run: |
          echo "::add-matcher::.github/matchers/${{ github.job }}.json"
      - name: pylint **/*.py --errors-only
        run: |
          pylint **/*.py --errors-only
```

And we can specify `needs` on `pylint` and `black` to make them run after we've run the install dependencies step?

MaikoTan:
Looks like jobs use seperated environment, tools installed in one job is not usable in other jobs?

panicstevenson:
Hmmm. Looks like you’d have to upload the site-packages from pip install and then download them in each run, which is probably more effort than it’s worth.

MaikoTan:
Well let me revert it first...

quisquous:
Thanks! This all seems reasonable to me now that I've had a chance to look at it.  @panicstevenson, did you have any more comments?

panicstevenson:
I think it looks good to land; if we have anything else, we can update it later

PR #1825, opened 2020-09-26
jobs: refactor combo tracker to be a tree, not a map

quisquous:
Dragoon skill combos share some of the same skills, such as
some having Wheeling Thrust -> Fang and Claw and others having
Fang and Claw -> Wheeling Thrust at the end.

The current combo tracker has a global map of all skills by id,
and so both of these abilities get marked as the "final" ability
in the chain, which is incorrect some of the time.

Refactor the combo tracker to have a tree of nodes instead, so that
there is context about the chain of abilities to get there.
This will let #1804 not have to track specific abilities with timers
and will let it use the combo tracker more naturally.

PR #1848, opened 2020-10-06
jobs: Fix refactored comboTracker(#1825)

### LINKED ISSUE #1845: Jobs: Refactored comboTracker(#1825) cause unexpected work when related skill not use in combo
In `this.comboFuncs.push((skill) => {...}`, `skill` now exist whether in combo or not after PR #1825 merged.
(It only exist when use in combo at earlier patch.)
This makes WAR, PLD and DRG UI work not as expected when related skill not use in combo.
Related lines in jobs.js:
https://github.com/quisquous/cactbot/blob/192dc2e873fd26b0fb3053aeaa4a08c12778b070/ui/jobs/jobs.js#L1504-L1548
https://github.com/quisquous/cactbot/blob/192dc2e873fd26b0fb3053aeaa4a08c12778b070/ui/jobs/jobs.js#L1643-L1654
https://github.com/quisquous/cactbot/blob/192dc2e873fd26b0fb3053aeaa4a08c12778b070/ui/jobs/jobs.js#L2192-L2197
Maybe related: 
https://github.com/quisquous/cactbot/blob/192dc2e873fd26b0fb3053aeaa4a08c12778b070/ui/jobs/jobs.js#L2659-L2662

Echoring:
As issue #1845 notes.
- Solves and closes #1845 
- Add something about finalskill for later use in comboTimer.

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -271,9 +271,8 @@ class ComboTracker {
       this.comboTimer = window.setTimeout(() => {
         this.AbortCombo(null);
       }, kComboDelayMs);
-      if (isFinalSkill) {
+      if (isFinalSkill)
         this.finalSkill = true;
-      }
     }
     if (id)
       this.callback(id);
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -2675,9 +2675,8 @@ class Bars {
   }
 
   OnComboChange(skill) {
-    if (this.combo.IsComboBroken()) {
+    if (this.combo.IsComboBroken())
       skill = null;
-    }
     for (let i = 0; i < this.comboFuncs.length; ++i)
       this.comboFuncs[i](skill);
   }
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -218,6 +218,8 @@ class ComboTracker {
     // A tree of nodes.
     this.startMap = {}; // {} key => { id: str, next: { key => node } }
     this.callback = callback;
+    this.comboBroken = true;
+    this.finalSkill = false;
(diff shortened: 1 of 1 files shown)


quisquous:
What was the bug? It is hard to tell what is going on because you are simultaneously fixing one thing and adding two other things.

Echoring:
They are all for fix this bug, not unrelated feature.
Seems it's hard to explain, I'll show all reason and thought flow.
1. The bug is, cast StormsEye/GoringBlade/Disembowel without combo, and its buff timer will unexpectedly begin to countdown.
2. This bug happen because in `this.comboFuncs.push((skill) => {...}`, `skill` == its HexID whether in combo or not.
(It will be `undefined` in earlier patch if not in combo)
3. `skill` is from `OnComboChange(skill)`.
4. I need to check whether combo is broken or not. So I add `IsComboBroken()` back.
5. If combo is broken, `skill` shouldn't be its HexID, so I make it `null`.
(Nothing use `undefined` to check this earlier, and I think `null` will be better.)
6. Then I find that StormsEye/GoringBlade can not trigger their box even in combo.
7. This is because they are final skill of a combo, and we consider them as combo is broken.
8. So I remove `isFinalSkill` condition in Line 254, and add a new varible to tell is final skill or not.

quisquous:
I took a look at what was broken and came up with what I think is a simpler fix in #1861.  I gave it some testing, but I'd appreciate an extra look to see if that fixes what you're seeing.

Echoring:
Seems nothing wrong and all known issue has been solved. Thank you!

quisquous:
Thanks for taking a look at it! Sorry to write my own patch, but I really wanted to understand what was broken and was wondering if there was a simpler solution here.

PR #1849, opened 2020-10-07
test: fix for prefix check

Jaehyuk-Lee:
example)
`Diadem F` being as common prefix for these
`Diadem Falling Asleep`
`Diadem Found Gather Point`
`Diadem Flag Alert`

fixed to be `Diadem`

But `Eden\'s Gate` becomes `Eden\'s`, so do we have two word ids?

### DIFF
--- a/test/trigger/test_trigger.js
+++ b/test/trigger/test_trigger.js
@@ -275,7 +275,8 @@ let testValidIds = function(file, contents) {
   // as "Prefix " does.  This is a bit rigid, but prevents many typos.
   if (ids.size > 1 && !brokenPrefixes && prefix && prefix.length > 0) {
     // if prefix includes more than one word, just remove latter letters.
-    prefix.substr(0, prefix.indexOf(' '));
+    if(prefix.includes(' '))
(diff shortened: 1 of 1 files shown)


test/trigger/test_trigger.js
```diff
@@ -274,6 +274,8 @@ let testValidIds = function(file, contents) {
   // as the prefix "O4" is not a full word (and have a space after it,
   // as "Prefix " does.  This is a bit rigid, but prevents many typos.
   if (ids.size > 1 && !brokenPrefixes && prefix && prefix.length > 0) {
+    // if prefix includes more than one word, just remove latter letters.
+    prefix.substr(0, prefix.indexOf(' '));
```
quisquous, line 278: I think `lastIndexOf` instead of `indexOf` will solve the multiple word problem.

Also, I think you should only do this `substr` if there is a space in the word, because otherwise `prefix` will become `undefined`.

Jaehyuk-Lee: Ah, that's right

quisquous:
Thanks for fixing this up!

PR #1851, opened 2020-10-07
responses: Korean better translation for "move"

Jaehyuk-Lee:
original one, `움직이기!` is used in two triggers
changed one, `이동하기!` is used in four triggers
These two are slightly different in meaning, but if player knows mechanics in battle, they'll understand with any of which shows.

But, I have some concerns, sometimes other languages don't have exact same meaning words corresponding to English, 
which means it seems fine to use same `Responses` for multiple triggers in English, sometimes it does not in other languages.

How do we deal with it?
I already sacrificed some readability in Korean responses, it's bearable now but there can be very different meaning in Korean in the future.

### DIFF
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -876,7 +876,7 @@ let Responses = {
     return obj;
   },
   moveAround: (sev) => {
-    // to avoid bllizard
+    // should jump or move
     let obj = {};
     obj[defaultInfoText(sev)] = {
       en: 'Move!',
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -856,7 +856,7 @@ let Responses = {
       fr: 'Bougez !',
       ja: '動く！',
       cn: '快动！',
-      ko: '움직이기!',
+      ko: '이동하기!',
     };
     return obj;
   },
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -278,7 +278,7 @@ let Responses = {
     };
     return obj;
   },
-  stack: (sev) => {
+  stackMarker: (sev) => {
     // for stack marker
     let obj = {};
     obj[defaultAlertText(sev)] = {
@@ -304,7 +304,7 @@ let Responses = {
     };
     return obj;
   },
-  stackOn: (sev) => {
+  stackMarkerOn: (sev) => {
     let obj = {};
     obj[defaultAlertText(sev)] = (data, matches) => {
       let target = getTarget(matches);
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -848,7 +848,7 @@ let Responses = {
     };
     return obj;
   },
-  move: (sev) => {
+  moveAway: (sev) => {
     let obj = {};
     obj[defaultInfoText(sev)] = {
       en: 'Move!',
@@ -857,6 +857,18 @@ let Responses = {
       ja: '動く！',
       cn: '快动！',
       ko: '이동하기!',
+    };
+    return obj;
+  },
+  moveAround: (sev) => {
+    let obj = {};
+    obj[defaultInfoText(sev)] = {
+      en: 'Move!',
+      de: 'Bewegen!',
+      fr: 'Bougez !',
+      ja: '動く！',
+      cn: '快动！',
+      ko: '움직이기!',
     };
     return obj;
   },
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -849,6 +849,7 @@ let Responses = {
     return obj;
   },
   moveAway: (sev) => {
+    // to dodge aoes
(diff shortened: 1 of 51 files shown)


MaikoTan:
> But, I have some concerns, sometimes other languages don't have exact same meaning words corresponding to English, 
> which means it seems fine to use same `Responses` for multiple triggers in English, sometimes it does not in other languages.

Can you show some examples about these "words"?
I think if it happens in Korean, it might happen in Japanese or even Chinese too.

Jaehyuk-Lee:
## stack
Korean people call stack marker as `쉐어징` (쉐어 for stack / 징 for marker)
stack marker
But it only mean "that" marker.
sometimes, there's stack, means just stack and bait aoe in the middle, in Korean that would be `모이기`.
Those are slightly different.

쉐어징 problem is Korean community's tacit(?) word, not dictionary thing.

## move
move has two usage, 
1. move to avoid freezing
2. move to dodge aoes

first one, in Korean, it's better to say `움직이기` (okay to move z axis)
second one, in Korean, it's better to say `이동하기` (move x, y axis)

If we only say `움직이기` and without mechanic knowledge, player would understand just to jump.

## etc
There'd been some more examples but those are okay not using exact word, better not separating, for project maintenance.

MaikoTan:
Yeah, it is the same as Japanese and Chinese too.

For "Stack" things, when it refers to "share damage", we always say, "分摊"/"頭割り"; when it refers to "get together", we'll say, "集合"/"集合".
So when I am translating the mechanics I would check it out what is exactly the solution, so I can write the correct word to refer it.

Also, for "Move", we have "移动"/"動く" and "避开"/"避け" and so many words.

I would agree what you said.

quisquous:
> But, I have some concerns, sometimes other languages don't have exact same meaning words corresponding to English,
> which means it seems fine to use same `Responses` for multiple triggers in English, sometimes it does not in other languages.
> How do we deal with it?

If there are phrases that need to be differentiated in other languages, I think we should add multiple responses.  Maybe something like `moveAway` vs `moveAround` (or some other phrasing, with some comments about when to use each)?

> I already sacrificed some readability in Korean responses, it's bearable now but there can be very different meaning in Korean in the future.

Can I ask which responses you have sacrificed readability on? What can be improved?

Jaehyuk-Lee:
`move` part I mentioned on the third comment was it. I thought "sacrifice" was some strong word, but I couldn't think of other words.

Jaehyuk-Lee:
I'll change `responses.js` as quisquous mentioned to split response, then change status to ready for review.

resources/responses.js
```diff
@@ -279,14 +279,28 @@ let Responses = {
...
   },
   stack: (sev) => {
+    // for stack marker
     let obj = {};
     obj[defaultAlertText(sev)] = {
       en: 'Stack',
       de: 'Sammeln',
       fr: 'Packez-vous',
       ja: 'スタック',
       cn: '集合',
```
MaikoTan, line 289:
```suggestion
      ja: '頭割り',
      cn: '分摊',
```

resources/responses.js
```diff
@@ -279,14 +279,28 @@ let Responses = {
...
+    return obj;
+  },
+  getTogether: (sev) => {
+    // for getting together without stack marker
+    let obj = {};
+    obj[defaultAlertText(sev)] = {
+      en: 'Stack',
+      de: 'Sammeln',
+      fr: 'Packez-vous',
+      ja: 'スタック',
```
MaikoTan, line 301:
```suggestion
      ja: '集合',
```

resources/responses.js
```diff
@@ -848,7 +862,21 @@ let Responses = {
...
-  move: (sev) => {
+  moveAway: (sev) => {
+    // to dodge aoes
+    let obj = {};
+    obj[defaultInfoText(sev)] = {
+      en: 'Move!',
+      de: 'Bewegen!',
+      fr: 'Bougez !',
+      ja: '動く！',
+      cn: '快动！',
```
MaikoTan, line 873:
```suggestion
      ja: '避けて！',
      cn: '快躲开！',
```

MaikoTan (commented):
Don't want to open a new PR, just melt into yours.

resources/responses.js
```diff
@@ -279,14 +279,28 @@ let Responses = {
     return obj;
   },
   stack: (sev) => {
+    // for stack marker
```
quisquous, line 282: Do you think it'd be worth renaming this to `stackMarker` and `stackMarkerOn` for clarity?

quisquous:
Thanks!  Please speak up in the future if something like this comes up again.

PR #1852, opened 2020-10-07
style: convert unicode to human-readable character

MaikoTan:


quisquous:
Thank you for these conversions.  Mostly they are there because I was being lazy.  All of the translations have the escaped unicode, and so I have to unescape each one of them manually as I'm adding anything.  Unfortunately I could not find an eslint rule for this, but I think the way you have it is preferred over the escaped format.

Issue #1846, opened 2020-10-06
I have created a vscode extension


MaikoTan:
[cactbot-highlight](https://github.com/MaikoTan/cactbot-highlight)
Here is the repository URL.

As quisquous said in [issue (comment)](https://github.com/quisquous/cactbot/pull/1834#issuecomment-703332443), I try to create an extension for highlighting in VSCode.

This is my first time to create a VSCode extension, it is not very robust.
If you have some suggestion or bug report, feel free to send me an issue.

`cactbot-highlight` now available in [Visual Studio Market](https://marketplace.visualstudio.com/items?itemName=MaikoTan.cactbot-highlight)!

JLGarber:
This looks really nice. I'm actually in the middle of putting a dungeon package together right now, so I'll finish compiling that with this extension enabled and let you know what I come across. The timeline highlighting looks like it could be extremely helpful!

quisquous:
This looks *really* nice.  Maybe I should finally stop using Notepad++ and start using VSCode.  I think my only feedback from the gifs is that translations come later and omitting them lets the tools find when they are missing.  So, I don't think I would ever add every language at once to a trigger.

If you want to send a pull request to add it to the `CONTRIBUTING.md` (or anywhere else you think it would be good to mention), I'd be happy to add a link.

MaikoTan:
> I think my only feedback from the gifs is that translations come later and omitting them lets the tools find when they are missing. So, I don't think I would ever add every language at once to a trigger.

They are just `snippets` and configurable in [snippets/trigger.js.code-snippets](https://github.com/MaikoTan/cactbot-highlight/blob/master/snippets/trigger.js.code-snippets) file. I can't recall which snippets are useful, so I just wrote down some "examples", you can provide some suggestions for which can be added in and which not.

In fact, I think the most thing that we need is `Intellisense`, maybe it is time to migrate cactbot to modern JavaScript with JSDoc and module system (like CommonJS) ?

> If you want to send a pull request to add it to the `CONTRIBUTING.md` (or anywhere else you think it would be good to mention), I'd be happy to add a link.

Yeah, sure of course.
But you know I am Asian, my English is not very well as native. I am afraid that I write something wrong in the document. (´。＿。｀)

quisquous:
> In fact, I think the most thing that we need is `Intellisense`, maybe it is time to migrate cactbot to modern JavaScript with JSDoc and module system (like CommonJS) ?

[...]

PR #1853, opened 2020-10-07
add .vscode ignore

Jaehyuk-Lee:
There's new vscode extension.
I think we need `.vscode/` ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .vs
+.vscode/
 **/*.suo
 bin/
 obj/


panicstevenson:
It's common practice to commit the `.vscode` folder and its contents, since they usually pertain to editor recommendations per repository. What's being added here that is causing issues?

MaikoTan:
Well, I don't support this.

According to https://github.com/flutter/flutter/issues/15555 , it is not recommand to add IDE/OS-specific files in `.gitignore`.

In fact, you can set a global `.gitignore` via `git config --global core.excludeFiles /path/to/your/global/gitignore`.

panicstevenson:
If this is for https://github.com/quisquous/cactbot/issues/1846#issuecomment-704966329 does it make sense to include this (although, I think there are other .txt files in that specific path)? This was something I had mentioned in that same thread: https://github.com/quisquous/cactbot/issues/1846#issuecomment-704655088

We can wait for quisquous for their guidance on VSCode, which they don't use. Although, I do know they have opinions on git.

MaikoTan:
Hmmm...
If we want to include it into `gitignore`, this way is recommanded:

https://github.com/github/gitignore/blob/master/Global/VisualStudioCode.gitignore

```
.vscode/*
!.vscode/settings.json
!.vscode/tasks.json
!.vscode/launch.json
!.vscode/extensions.json
*.code-workspace

# Local History for Visual Studio Code
.history/
```


In my opinion, if this project recommands VSCode as primary developing environment, it is okay to upload those setting files to git.

Jaehyuk-Lee:
Oh I thought this was usual
I did this somewhere else.
okay, closing this PR

quisquous:
Are there user settings that end up in .vscode that shouldn't be committed and should be ignored? Sorry, as panic says above, I don't know much about it.

Jaehyuk-Lee:
I don't know how most people does, but when I was in military service, we added `.vscode` to gitignore. I first met VS code in military and kept that habit until now.

(I mentioned military because I meant my experience is not usual thing since Korean military does not actively use habits of normal internet community or company.)

MaikoTan:
Well, I don't support this because many people in my company using different IDE/editors.

My team mainly develops Java and VueJS applications, but everyone's habits are different. Some people like Idea, some people (such as me) like VSCode, and Vim and Emacs are said to be used. (And the operating system is even more diverse...)
Therefore, our company stipulates not to upload files related to the operating system and editor, unless the project specifies an editor as the main development tool.

Therefore, I think that if it is a personal project, it can naturally be handled at will, but as a community-participated project, these things still need to be considered?

Of course, the final choice lies with quisquous. I'm just talking about my thoughts, not trying to enforce it.

(Sorry I don't know how to describe it so I used Google Translation)

PR #1856, opened 2020-10-08
jobs: Better width for bars and stacks

Echoring:
These changes make bars and stacks match the width of HP bar better.


quisquous:
Thanks for this! It looks much better.

(This is not related to this patch, but I was wondering some if all of the proc boxes should be laid out with `display: flex` rather than being put at very specific absolute positions.  That might make it more natural to have any number of proc boxes or to rearrange them in the future, and it would require fewer css rules.)

Issue #1857, opened 2020-10-08
Jobs: Dealing about difference across patches


Echoring:
Sometimes patch launched with a job adjustment.
CN and KO are late at patch launch, landing latest version may cause some confusion.
For example,
https://github.com/quisquous/cactbot/blob/19fc5c1d1dce83a930475098e972f8df539e4f47/ui/jobs/jobs.js#L1504-L1535
StormsEye is changed up to 60s in patch 5.3, but CN and KO is in patch 5.2.
These player may get confuse that eyeBox can not match their buff time.
Is it possible to make different setup base on game version or language?

Jaehyuk-Lee:
if you see line 927~945, it's already been dealing with it, but not each job's UI, buff timer is.
I think adding each job's override there is good.

quisquous:
I agree that adding overrides by parser language is the right approach here.

I don't know that the Storm's Eye times would fit in `buffOverrides`, because those are just for buffs and get directly applied to `this.buffInfo`.  There's no an equivalent data structure that has all of the effect times for each job.  I guess we could add one, but that also seems like a lot.

An alternative is that it might make sense to just have an `if` statement inside of the `setupWar`, and if this happens again we can think about what it would look like to generalize it more broadly.

quisquous:
Sounds like we have at least an immediate fix here.  I'm happy to discuss in the future if this becomes overwhelming.  It might be nice to figure out how to get the game version from the game itself so we can have these overrides apply automatically (rather than having language dependent selections).

PR #1859, opened 2020-10-10
jobs: WAR quick fix in CN/KO by if condition

Echoring:
As issue #1857 discussed.
Temporarily fix it by simple if condition.
Overrides will be underway since more patch difference is underway.
(Seems that MNK will change a lot in patch 5.4?)


quisquous:
Thanks!

PR #1860, opened 2020-10-10
raidboss: ruby-weapon show stack infoText only once

Jaehyuk-Lee:
stack infoText was shown twice. (there is two stack marker simultaneously)


quisquous:
If that's true, then it would say stack twice for the person who had the markers still? Then, maybe it should just say stack('alert') for everybody with a suppress as soon as any marker comes out, e.g.

```
    {
      id: 'RubyEx High-Powered Homing Lasers',
      netRegex: NetRegexes.startsUsing({ source: 'The Ruby Weapon', id: '4AD8' }),
      netRegexDe: NetRegexes.startsUsing({ source: 'Rubin-Waffe', id: '4AD8' }),
      netRegexFr: NetRegexes.startsUsing({ source: 'Arme Rubis', id: '4AD8' }),
      netRegexJa: NetRegexes.startsUsing({ source: 'ルビーウェポン', id: '4AD8' }),
      netRegexCn: NetRegexes.startsUsing({ source: '红宝石神兵', id: '4AD8' }),
      netRegexKo: NetRegexes.startsUsing({ source: '루비 웨폰', id: '4AD8' }),
      response: Responses.stackOn('alert'),
      suppressSeconds: 1,
    },
```

What do you think? Or is there a better way to solve this?

Jaehyuk-Lee:
So, you mean to remove `RubyEx High-Powered Homing Lasers You`?

quisquous:
Yeah, I think there's potential options here to solve all the double callouts.

(1) don't differentiate info vs alert, and just give everybody one `stackMarker('alert')` callout.  In other words, combine the two triggers into a single alert trigger for everybody.
(2) alert for anybody who has the marker, collect all of the people who got a stack marker, add delaySeconds to the "not you" version, and early out if the person's name is in the array

#2 just seemed more complicated so I suggested #1 as a simplification.

ui/raidboss/data/05-shb/trial/ruby_weapon-ex.js
```diff
@@ -133,7 +122,7 @@
       netRegexKo: NetRegexes.startsUsing({ source: '루비 웨폰', id: '4AD8' }),
       condition: Conditions.targetIsNotYou(),
       suppressSeconds: 1,
-      response: Responses.stack('info'),
+      response: Responses.stackMarker('alert'),
```
quisquous, line 125: I think the condition needs to be removed too, and then this should be good.

PR #1862, opened 2020-10-10
ci: Add GitHub Workflows to Release Artifacts

panicstevenson:
Add GitHub workflows to create a tag off version updates and create a
draft release off of a tag being created.

### DIFF
--- a/.github/workflows/create-tag.yml
+++ b/.github/workflows/create-tag.yml
@@ -1,0 +1,55 @@
+name: Create Tag
+
+on:
+  push:
+    branches:
+      - 'main'
+    paths:
+      - 'plugin/CactbotEventSource/Properties/AssemblyInfo.cs'
+      - 'plugin/CactbotOverlay/Properties/AssemblyInfo.cs'
+
+jobs:
+  create-tag:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


quisquous:
Is there any way for me to test this, or is this a "land and give it a spin" sort of thing?

panicstevenson:
So, this should run in two phases:
1. Updating both AssemblyInfos to a new version should create a tag.
2. Creating a tag should create a draft release.

I think if 1 failed, you could theoretically continue and make a tag yourself. 2 shouldn't do anything without your blessing, but we might need to confirm things look okay before releasing the release.
I think 2 is testable as a standalone, since the code should check if the tag already released; in which case it won't do anything.

We could theoretically break this into two separate PRs if you wanted to test just 2 without 1, or vice-versa? I don't know if the workflows themselves can just be disabled, which would be another option.

quisquous:
Nah, I think that's reasonable.  I think I'll just keep an eye out and if it does anything awry I'll manually clean up whatever happens.  I was going to do a release soon-ish anyway, possibly to give folks a heads up about webpack.  (Although maybe we need some documention / an open issue about this to start a discussion that I can point to.)

PR #1868, opened 2020-10-11
resources: Update Export Property Names

panicstevenson:
Update export property names to match their casing as declared within
the file themselves to work with webpack imports.


test/unittests/responses_test.js
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 let responseModule = require('../../resources/responses.js');
-let Responses = responseModule.responses;
+let Responses = responseModule.Responses;
```
quisquous, line 4: Could I ask you to clean this up as you go, and make this `let { Responses, triggerFunctions, severityMap } = require`, here and below? Sorry for not writing that more cleanly in the first place.

panicstevenson: So, I had another PR planning on cleaning up the imports to use destructuring where things have previously existed. Should I still make that change here or should I just wait until the destructuring PR?

quisquous: Oh, I can wait, then.  :+1:

PR #1871, opened 2020-10-12
Jobs: Show HP number in Healer

Echoring:
Healer's HP is as important as Tank's.
Many times I saw healers are busy at restoring other's HP, 
but ignore their own HP is not enough and got killed in AoE, causing party wipe.
Also healer who makes barrier sometimes want to know how much damage can it nullify.
So I think it's necessary to show HP number in Healer.

### DIFF
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -114,6 +114,9 @@
   height: 10px;
 }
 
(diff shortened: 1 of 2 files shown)


quisquous:
That seems reasonable to me, thanks!

PR #1874, opened 2020-10-12
Jobs: BRD UI

Echoring:
BRD
- Repertoire Box, show how many Repertoire you have. Color means which song you are singing.
- Soul Voice Box, turn orange when >= 95
- **DoT Timer**, show when your DoT tick on your main target.
- 2 DoT Tracker
- Songtime Tracker, change color and threshold for different song.


ui/jobs/jobs.js
```diff
@@ -169,6 +169,14 @@ const kAbility = {
   Dia: '4094',
   Assize: 'DF3',
 };
+// TODO: Some of them are missed in effect_id.js,
+// if effect_id.js is fixed, this can be removed.
+const brdDoTs = {
```
quisquous, line 174: Can you put these in the `known_mapping` section of `util/gen_effect_id.py` and re-run that script rather than including them manually here? You need to use the decimal numbers there, rather than the hex values you have here.

Echoring: Done.

ui/jobs/jobs.js
```diff
@@ -3124,6 +3226,48 @@ class Bars {
...
+      if (m) {
+        if (this.dotTarget.includes(m.groups.targetId))
+          this.lastAttackedDotTarget = m.groups.targetId;
+      }
+    } else if (type === '24') {
+      // line[2] is dotted target id.
+      if ((this.dotTarget.includes(line[2])) &&
+        (line[2] === this.lastAttackedDotTarget) &&
+        (line[4] === 'DoT') &&
+        (line[5] === '0')) // 0 if not fleld setting DoT
```
quisquous, line 3268: This is just a style nit, but do all of these conditions need parentheses? I think you can leave them all off.

ui/jobs/jobs.js
```diff
@@ -2491,6 +2499,105 @@ class Bars {
...
+      id: 'brd-procs-stormbite',
+      fgColor: 'brd-color-stormbite',
+    });
+    [
+      EffectId.Stormbite,
+      EffectId.Windbite,
+    ].forEach((effect) => {
+      this.mobGainEffectFromYouFuncMap[effect] = () => {
+        stormbBiteBox.duration = 0;
+        stormbBiteBox.duration = 30 - 0.5;
```
quisquous, line 2517: Can you leave a comment about what the `0.5` is for? I assume it's something like animation delay?

ui/jobs/jobs.js
```diff
@@ -2491,6 +2499,105 @@ class Bars {
   }
 
   setupBrd() {
+    // DoT
+    const causticBiteBox = this.addProcBox({
+      id: 'brd-procs-causticbite',
+      fgColor: 'brd-color-causticbite',
+    });
+    const stormbBiteBox = this.addProcBox({
```
quisquous, line 2507: spelling nit: `stormbBite` => `stormBite` here and elsewhere

ui/jobs/jobs.js
```diff
@@ -3124,6 +3226,48 @@ class Bars {
...
+          const index = this.dotTarget.indexOf(m.groups.targetId);
+          if (index > -1)
+            this.dotTarget.splice(index, 1);
+        }
+      }
+    } else if (type === '21' || type === '22') {
+      let m = log.match(kYouUseAbilityRegex);
+      if (m) {
+        if (this.dotTarget.includes(m.groups.targetId))
+          this.lastAttackedDotTarget = m.groups.targetId;
```
quisquous, line 3261: I'm not quite sure I follow the reasoning of storing the `lastAttackedDotTarget`.  It it because the repertoire timer then resets constantly and looks weird? If that's the case, can you leave a comment about what's going on here?

Alternatively, a simpler approach might just be `lastDotTarget` where you just keep track of the last thing you've put dots on. if it's already the case that multitarget is tricky.

ui/jobs/jobs.js
```diff
@@ -3124,6 +3226,48 @@ class Bars {
...
+    const brdDoTs = Object.freeze([
+      EffectId.Stormbite,
+      EffectId.Windbite,
+      EffectId.CausticBite,
+      EffectId.VenomousBite,
+    ]);
+    if (type === '26') {
+      let m = log.match(kMobGainsEffectFromYouRegex);
+      if (m) {
+        if (Object.values(brdDoTs).includes(m.groups.effectId.toUpperCase()))
```
quisquous, line 3242: If you're going to uppercase the id twice, could you save it in a local `effectId` variable?

ui/jobs/jobs.js
```diff
@@ -3124,6 +3226,48 @@ class Bars {
           this.buffTracker.onUseAbility(m.groups.id, m.groups);
       }
     }
+    // For extremely complex BRD
+    if (this.job != 'BRD') return;
```
quisquous, line 3230: Can you put return on its own line? This is just my own personal style, but I find it very easy to overlook a `return` at the end of the line.

ui/jobs/jobs.js
```diff
@@ -3124,6 +3226,48 @@ class Bars {
...
+    } else if (type === '21' || type === '22') {
+      let m = log.match(kYouUseAbilityRegex);
+      if (m) {
+        if (this.dotTarget.includes(m.groups.targetId))
+          this.lastAttackedDotTarget = m.groups.targetId;
+      }
+    } else if (type === '24') {
+      // line[2] is dotted target id.
+      if ((this.dotTarget.includes(line[2])) &&
+        (line[2] === this.lastAttackedDotTarget) &&
```
quisquous, line 3266: If this is the `lastAttackedDotTarget` then it's already necessarily in `this.dotTarget`, and so `this.dotTarget.includes` shouldn't need to be checked?

ui/jobs/jobs.js
```diff
@@ -2491,6 +2499,105 @@ class Bars {
...
+      id: 'brd-procs-causticbite',
+      fgColor: 'brd-color-causticbite',
+    });
+    const stormbBiteBox = this.addProcBox({
+      id: 'brd-procs-stormbite',
+      fgColor: 'brd-color-stormbite',
+    });
+    [
+      EffectId.Stormbite,
+      EffectId.Windbite,
```
quisquous, line 2513: Could you leave a comment that iron jaws refreshes these status effects and so doesn't need to be handled separately? I was looking for iron jaws and was a little confused how it worked.

ui/jobs/jobs.js
```diff
@@ -2491,6 +2499,105 @@ class Bars {
   }
 
   setupBrd() {
+    // DoT
+    const causticBiteBox = this.addProcBox({
```
quisquous, line 2503: Do you there should be a straight shot proc box?

ui/jobs/jobs.js
```diff
@@ -2491,6 +2499,105 @@ class Bars {
...
+      classList: ['brd-color-song'],
+    });
+    this.repertoireTimer = this.addTimerBar({
+      id: 'brd-timers-repertoire',
+      fgColor: 'brd-color-song',
+    });
+    // Only with-DoT-target you last attacked will trigger this timer.
+    // So it work not well in mutiple targets fight.
+    this.UpdateDotTimer = () => {
+      this.repertoireTimer.duration = 2.91666;
```
quisquous, line 2552: Ohmygosh, this is GREAT.  I haven't played bard in ages, but having a timer for when your procs might come up just feels really good.

ui/jobs/jobs.css
```diff
@@ -674,6 +686,42 @@
...
+.brd-color-song.ballad {
+  background-color: rgb(143, 90, 143);
+}
+
+.brd-color-song.paeon {
+  background-color: rgb(207, 205, 52);
+}
+
+.brd-color-soulvoice {
+  background-color: rgb(130, 200, 0);
```
quisquous, line 718: I feel like the soulvoice and the minuet colors both are green and kind of similar.  I also think the high color being yellowish looks a lot like the paeon color.  Maybe there can be more differentiation?  I like the songs being different colors, but maybe then soul voice doesn't need as much color normally?

What do you think about soul voice being white-ish / uncolored and then the high color being red-ish or more orange-ish rather than yellow-ish?

quisquous (commented):
Thanks! This is all really good work for a complicated job.  I have a bunch of small questions, but nothing too huge.

ui/jobs/jobs.js
Echoring: For an unknown reason, the effect duration of these two DoT is not actually 30s, but 29.5s.
I think maybe it is because countdown starts before DoT appear on target.

ui/jobs/jobs.js
Echoring: For `lastDotTarget`, there is a condition that when an add appear, you just cast DoT to them and turn back to your main target. And when that add die, it need more code to redirect `lastDotTarget` to your main target. `lastAttackedTarget`, `lastDotTarget` can both not be maintarget, but `lastAttackedDotTarget` must be your main target.

ui/jobs/jobs.css
Echoring: I think so. I mostly pick a color on its icon, but sometimes there is too many similar color.

ui/jobs/jobs.js
Echoring: You means add this annotation into code right?

ui/jobs/jobs.js
Echoring: Seems it will make procs block a little crowded, but not bad to be there.

ui/jobs/jobs.js
quisquous: Ah, I see.  That makes sense!  Could you leave a comment about this scenario?

One other question I had was when do the `dotTargets` get removed? It seems like this list will grow forever.  Maybe it'd be worth capping at N entries and also clearing whenever you're out of combat?

ui/jobs/jobs.js
quisquous: Yeah, I meant add a comment like `// Iron jaws is automatically handled by refreshing these effects and so we don't need to handle this ability ourselves.`

ui/jobs/jobs.js
Echoring: ID in `dotTargets` will be removed when they lost DoT.
Also clear `dotTargets` when passing `UpdateJob`.
I think I can clear it on zone change, or remove it when removing combatant, since ID in `dotTargets` only cannot removed when removing combatant without death, mostly because of zone change. Is this be better way?

quisquous: Oh, sorry.  You're right!  I missed the code where these are removed when the dots are removed.  Thanks for also mentioning the death case.

I think it'd still be worth clearing on zone change too, but then I think there's nothing else to do here.

ui/jobs/jobs.js
JLGarber: Is the internal cooldown on Repertoire not exactly 3 seconds? Or is this value here intended to adjust for some sort of interface delay?

MaikoTan: Yes, the Reperoire tick (or server DoT tick) is in every "Eorzea minute".
We know that 1 "Eorzea day" == 70 "Earth minutes", so 1 "Eorzea minute" equals `(70/(1*24*60))*60 = `2.91667 "Earth seconds".

JLGarber: Ahh, okay. I had always internalized it as 3s and never bothered to time it exactly after it became muscle memory. Thanks for confirming that!

PR #1882, opened 2020-10-15
fisher: Fix Lexical Scoping

### LINKED ISSUE #1880: Latest fisher module not working
After the commit https://github.com/quisquous/cactbot/commit/8432fbcb7bb03604652fd6b6861ba6a5af0ae531, the fisher overlay is no longer working for me.
The errors i got are: 
```
[15.10.2020 18:29:06] Info: fischer: BrowserConsole: Uncaught TypeError: Cannot read property 'IQRHookQuantity' of undefined (Source: file:///E:/FINAL%20FANTASY%20XIV%20-%20A%20Realm%20Reborn/AdditionalExtern/ACT/Plugins/OverlayPlugin_ngld/cactbot/ui/fisher/seabase.js, Line: 274)
[15.10.2020 18:29:06] Info: fischer: BrowserConsole: Uncaught TypeError: Cannot read property 'IQRTugQuantity' of undefined (Source: file:///E:/FINAL%20FANTASY%20XIV%20-%20A%20Realm%20Reborn/AdditionalExtern/ACT/Plugins/OverlayPlugin_ngld/cactbot/ui/fisher/seabase.js, Line: 318)
```
It worked with an older version but not with the latest one (even made sure to update fisher data by using the gen_fisher data script)
@panicstevenson could you have a look at this?

panicstevenson:
Switch functions to instead use arrow functions to preserve `this` without
having to create subsequent variables inside of individual functions.

Fixes #1880

### DIFF
--- a/ui/fisher/seabase.js
+++ b/ui/fisher/seabase.js
@@ -23,10 +23,8 @@ class SeaBase {
   }
 
   getConnection() {
-    let _this = this;
-
-    return new Promise(function(resolve, reject) {
-      let req = window.indexedDB.open(_this._dbName, _this._dbVersion);
+    return new Promise((resolve, reject) => {
+      let req = window.indexedDB.open(this._dbName, this._dbVersion);
 
       req.onsuccess = function(event) {
         resolve(this.result);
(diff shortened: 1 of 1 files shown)


Akurosia:
just tested it out and can confirm this works (tested while fishing for allagan hunter)

ui/fisher/seabase.js
```diff
@@ -23,28 +23,26 @@ class SeaBase {
   }
 
   getConnection() {
-    let _this = this;
+    return new Promise((resolve, reject) => {
+      let req = window.indexedDB.open(this._dbName, this._dbVersion);
```
MaikoTan, line 27: I think if the variable won't change in the future, declare it as `const` would be better than `let` ?
`let` here and some ones below can be changed.

panicstevenson: I don't disagree, but this pattern is used everywhere. I think I'd rather clean this up in a separate pull request.

PR #1883, opened 2020-10-16
general: Convert Cactbot to Use Webpack and ECMAScript Modules

### LINKED ISSUE #1869: Convert cactbot to use Webpack
I'd like to switch cactbot over to using [Webpack](https://webpack.js.org/).  In other words, running a bundling step prior to using any local files.  For end users, there should be no difference, but this will require developers to change their workflow.  Instead of being able to use local files directly out of the repo, instead you'll need to run a step to run a Webpack dev server and load any overlays from there.

Here's my motivations for this:

* One thing we can't do now is to use ESM modules from local `file://` pages.  This means that any js files in node can use `require` but html has to include them with `<script>` in a very fragile way.  Moving to Webpack will let us use CJS or ESM modules in both node and web and will make dependencies a lot cleaner.  This will also make it possible to write a log splitting html front-end to the util/ scripts.  Right now, using any files that depend on other files in both the browser and in node is [very hacky](https://github.com/quisquous/cactbot/blob/main/resources/netregexes.js#L74-L79)

* ngld wants to merge cactbot's plugin code into OverlayPlugin.  One of the things that ngld has mentioned as wanting to solve is the data file loading (e.g. for raidboss triggers and timelines, and oopsy).  This is also filed as #1767.  We can use Webpack to bundle all of these files together locally instead of asking the C# plugin to fetch files from a manifest at runtime.  This will also let raid emulator work without ACT open.  Minor point here also that this means we won't need to list anything in a manifest file anymore.

* I'd also like to move to TypeScript in the future.  We can do this incrementally, and Webpack can help here to handle transforming ts files into js files.  TypeScript will help with catching errors at compile time and providing better Intellisense in VSCode.

I'm mostly filing this issue as a notice for cactbot developers.

panicstevenson:
Convert Cactbot to use Webpack and ECMAScript modules. There are a few
needs for Webpack, or similar; this change doesn't necessarily
implement anything that breaks reliance on `file://` patterns, or
allows for a complete decoupling of the plugin code from the module
code, but this is a logical first step.

It is unfortunately necessary to implement the change to ESM as well
within this patch, in case there are downstream projects directly
importing files from this repository without necessarily forcing them
to address or tackle something similar to Webpack on their end.

Fixes #1869


panicstevenson:
I believe I have tested all modules with this; but, I can't claim to have an intimate familiarity with every module, nor can I claim to have followed every code path (looking at you, TTS). This is a breaking pull request, and as such, I'm hopeful that people are willing to help test it, or critically review it.

Everything should be available at my GitHub pages' URL:
https://panicstevenson.github.io/cactbot/ui/raidboss/raidboss.html as the raidboss example, please substitute out the modules as necessary, and remember that the DPS overlays are under their own subfolders.

Feel free to critique any line, any import, any missing import, the order in which things are imported (I didn't try that hard to format -- I'd love thoughts here).

Current TODOs:

- [ ] Add/Update Documentation
- [ ] Cleanup GitHub Pages GitHub Action (?)

Steps for building this locally if you feel so inclined:

1. Get my repo/fork
1. `npm start`
1. Point to localhost:8080/ui/raidboss/raidboss.html (or similar)

I will try to keep up with the main repository's main branch, but I'm not really willing to give an SLA on that.

Thanks for reading!

package.json
```diff
@@ -142,7 +150,7 @@
         }
       ],
       "strict": [
-        "error",
+        "off",
```
panicstevenson, line 153: Side note: This rule is good, but is very noisy since modules use 'use strict' by default. This will need to be followed up on, since deleting all of the 'use strict' lines at the top of each file:
1. Didn't work very well with `--fix`
1. Makes the diff really difficult to read

resources/netregexes.js
```diff
@@ -1,5 +1,7 @@
 'use strict';
 
+import Regexes from './regexes.js';
```
MaikoTan, line 3: I think don't trailing `.js`, `import Regexes from './regexes';` is ok enough?
Because quisquous said we're going to migrate cactbot to TypeScript?

panicstevenson: So, from my testing, not having the `.js` was okay from Webpack's point of view, but not okay from a strict HTML point of view, if the source type was module. Ultimately, I decided to follow:
https://google.github.io/styleguide/jsguide.html#es-module-imports
Since generally we're using the Google style guide, but I don't know how strictly we're following it.

quisquous: I think for now this is fine.  I'd like to move slowly with figuring out to handle the couple of folks who are building things on top of cactbot outside of the repository.  For now, I think keeping these files as directly importable is good.  I think when we move to typescript, there'll need to be some other changes as well and we can change this at that point.  (Or change all imports at that point, if it makes sense).

ui/raidboss/emulator/data/network_log_converter/ParseLine.js
```diff
@@ -11,13 +37,83 @@ class ParseLine {
...
       return false;
 
-    // This is ugly, but just want to check if we have a specific
-    // line class is defined without having a huge if/else or switch/case
-    let tn = 'LineEvent' + event;
-    if (eval('typeof(' + tn + ')') === 'function')
-      eval('ret = new ' + tn + '(repo, line, parts);');
-    else
+    // This is ugly, but Webpack prefers being explicit
+    switch ('LineEvent' + event) {
```
kuwoyuki, line 41: How about using [dynamic imports](https://webpack.js.org/api/module-methods/) for this?

panicstevenson: Okay, I finally got around to trying this. Keep in mind I'm coming at this with zero expertise, so please bear with me if I did something very not-what-you-expected. I removed all the imports and the switch, and put back the code with the following modifications (no idea why the hex conversion was giving me so much trouble, but that ended up working with the configuration below):
```diff
    // This is ugly, but just want to check if we have a specific
    // line class is defined without having a huge if/else or switch/case
    let tn = 'LineEvent' + event;
+   const hexEvent = parseInt(event, 10).toString(16);
+   const module = import(`./LineEvent0x${hexEvent}.js`);
    if (eval('typeof(' + tn + ')') === 'function') {
      eval('ret = new ' + tn + '(repo, line, parts);');
    } else {
      ret = new LineEvent(repo, line, parts);
    }
```

My problem here now is very similar to my problem with other `eval()`s in this patch: they don't know what they're looking for. I can `import X from 'Y.js';` and the downstream `eval()` won't know what X is. My workaround has been to `import X as _X from 'Y.js'; const X = _X;` If there's a better way to approach this, I'd love to know; I personally feel that the workaround is verging on disgusting.

As for this, I don't know of a way to ``const <insert dynamic name here> = import(`./LineEvent0x${hexEvent}.js`);``. My initial thought was to shove it into a map via `[varName]` syntax, but that doesn't stop the fact that at some point I need to retrieve it explicitly named the name that the `eval()` wants it to be:
```js
    // perfect code above this line
    if (importMap.has(tn)) {
      const LineEvent04 = importMap.get(tn); // How can I know this? How can I get LineEvent05, LineEvent39, etc.?
      eval('ret = new ' + tn + '(repo, line, parts);');
    }
    // perfect code below this line
```

(Some attempts later, I got the above to work via object destructuring and renaming, but the class doesn't know about the class that it extends. D'oh.)

Additionally, I found that this was creating [chunks](https://webpack.js.org/guides/code-splitting/), which seems kind of painful for distributing to people; I think sending 24 additional bundles is kind of a non-starter:
```
0.bundle.js  8.03 KiB             0  [emitted]
1.bundle.js  5.17 KiB             1  [emitted]
. . .
23.bundle.js  4.39 KiB             2  [emitted]
raidemulator.bundle.js   780 KiB  raidemulator  [emitted]  raidemulator
```

I was able to get around this by adding:
```js
  plugins: [
    new webpack.optimize.LimitChunkCountPlugin({
      maxChunks: 1,
    }),
  ],
```
to the Webpack config, but this all seems like a lot to be more clever with the code.

I could be wrong about all of the above; I tried to give it what I felt like was a fair attempt, but I can't find working implementation, and the paths that I've been going down have had more dragons than UCoB. Please correct me if I'm doing something very dumb here!

kuwoyuki: `import()` returns a promise, you could actually do something like
```js
    const hexEvent = parseInt(event, 10)
      .toString(16)
      .toUpperCase()
      .padStart(2, "0");
    let LineEvent;
    try {
      const module = await import(
        // eager so it doesn't generate an extra chunk
        /* webpackMode: "eager" */ `./LineEvent0x${hexEvent}.js`
      );
      if (!Object.prototype.hasOwnProperty.call(module, hexEvent)) {
        throw void 0;
      }
      LineEvent = module[hexEvent];
    } catch (_) {
      ({ default: LineEvent } = await import(
        /* webpackMode: "eager" */ `./LineEvent.js`
      ));
    }
```
Then do a `Promise.all` in the `NetworkLogConverter`
```js
lines = (await Promise.all(lines.map((l) => ParseLine.parse(repo, l)))).filter((l) => l);
```
But I think a much cleaner approach would be to make a dict/map of log lines and just import it?
maybe 
```console
└───network_log_converter
    │   LogRepository.js
    │   ParseLine.js
    │
    └───events
            0x00.js
            ...
            index.js
```
`import` all the events inside index and return an object/dict and simply access them using a property.

kuwoyuki: Another option to abstract it a bit since this is a class anyway.. :unamused: 
```js
export default class LineEventFactory {
  constructor(event, props) {
    switch (event) {
    case "00":
      return new LineEvent00(...props);
    case "01":
      return new LineEvent01(...props);
    // ...
    default:
      return new LineEvent(...props);
    }
  }
}
```
```js
ret = new LineEventFactory(event, [repo, line, parts]);
```

panicstevenson: Okay, so, in fairness, I haven't tried these. I think all of these (I'd have to see about the first one, seems a little magic to me) are better than the current/proposed implementation.

I think I want to hold off on making changes that actually change the code until after this lands, at which point either you or I (it's your code!) can make any one of these follow-up changes. In all reality, I should have left the eval() since I leave it everywhere else and just put that horrid `TODO` banner, but this felt small enough to just nip it and then move on. I think the switch statement is ugly, but there's no cleverness, so I think most people can agree from a passing glance that "this does what it was doing before" even if it's 100 more lines of code or whatever (🤮).

How do you feel? I guess I feel like I could be persuaded that "if I'm fixing the thing, might as well fix it good" but I'm also currently of the opinion that "yea, 100+ files is a fine enough PR for now"

panicstevenson:
There's also a shoutout to @MaikoTan who made a change and submitted a PR on my fork: https://github.com/panicstevenson/cactbot/pull/6
I think we're in agreeance that we'll merge this PR to the main repository and they can follow-up with their patch after.

MaikoTan:
Hey, panic. I have force-updated my changes, it is my first time using `git cherry-pick`, hope it would not break anything.

quisquous:
Ok, I think it's time for this to happen.  Do you mind rebasing this again so I can give it a thorough round of testing?

ui/dps/xephero/xephero.js
```diff
@@ -1,5 +1,11 @@
 'use strict';
 
+import DpsPhaseTracker from './dps_phase_tracker.js';
+import { InitDpsModule, Options } from '../dps_common.js';
+import UserConfig from '../../../resources/user_config.js';
+
+import '../../../resources/zone_id.js';
```
quisquous, line 7: Is this doing anything?

panicstevenson: Don't see how it could be!

ui/fisher/fisher-ui.js
```diff
@@ -16,7 +18,7 @@ if (window.customElements) {
   });
 }
 
-class FisherUI {
+export default class FisherUI {
```
quisquous, line 21: fisher-ui.js uses `Options` but it is defined in the other file.  I think if you pass it to the constructor and use `this.options` in the two places below, fisher will work properly again.

quisquous:
:checkered_flag:

PR #1885, opened 2020-10-17
raidboss: show invuln/provoke/shirk triggers to healers

### LINKED ISSUE #1884: Not getting Invuln/Provoke/Shirk alerts as healer
A few months ago I noticed an issue came up where I'm no longer getting any alerts when a tank uses an invuln/provoke/shirk anymore like I used to. I still get these alerts when I'm on a tank job, but when I heal I get nothing. Is this a setting I need to enable?

quisquous:
Fixes #1884.

### DIFF
--- a/ui/raidboss/data/00-misc/general.js
+++ b/ui/raidboss/data/00-misc/general.js
@@ -1,5 +1,9 @@
 'use strict';
 
+const caresAboutTankStuff = (data) => {
+  return data.role === 'tank' || data.role === 'healer' || data.job == 'BLU';
+};
+
 // Triggers for all occasions and zones.
 [{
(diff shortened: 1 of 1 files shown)


quisquous:
I already have these turned on for all jobs because I care about that sort of thing, but I am not sure if everybody else wants them too.  Anybody have additional opinions one way or the other?

JLGarber:
For myself, I prefer knowing when tanks are swapping or using invulns. It's *usually* going to happen around nebulously Dangerous Mechanics, and being warned helps me as a damage dealer to look out and stay out of the way. (Obviously this is already covered a lot of times by `tankCleave`, but it can still be helpful.)

PR #1887, opened 2020-10-17
eureka: add fate noises in bozja

quisquous:


quisquous:
This was asked for in discord, but having played with it for a while, I think it's not worth it.  There are just SO many fates that it is continually booping.  Putting this here for discussion, but I am inclined to not merge this.

quisquous:
cc @KattTails FYI

KattTails:
I think it could be useful if given the option to choose which fates to notify on. Personally there's a few fates I completely ignore, but chain fates that lead to the engagements are important to me.

FWIW I was looking at making a Bozja overlay but I wanted to include the CE pops, and I got sidetracked trying to track down a hard time tracking down the networking data. It doesn't use the same actorcontrol packets as regular fates, and seems to be internally under a new system called Dynamic Events.

Akurosia:
I would also limit the callouts to the "important" fates that spawn CE with maybe an option to also enable "unimportant" fates

KattTails:
I think a good way of handling might be is having all the fates implemented, defining the important fates (chains for CEs), and allowing the option of showing all fates or only important ones, Preferably there'd be a way to allow the user to choose the fates they want to mark as important.
Also we should probably hide the fates on the map until popped to reduce clutter, as many spawn on the same spots.

KattTails:
I've pushed an WIP of the eureka overlay with bozja support [here](https://github.com/KattTails/cactbot/commit/65d597ef006b9adcf5f2273bb2c81f9d32586341).

It's still noisy as it's missing the important fate filtering, but it's a start.

KattTails:
I've figured out the CE opcode, its 0x299, structure looks like 

| Byte | type | comment |
| ------------- | ------------- | ------------- |
| 0x20 | 4 bytes | pop time_t |
| 0x24 | 2 bytes | CE time remaining (0 until 0x2A is 3 |
| 0x28 | 1 byte | CE key as found DynamicEvents [exd](https://cdn.discordapp.com/attachments/448120801363886090/767223081822978058/DynamicEvent.csv) |
| 0x29 | 1 byte | Number of players in CE |
| 0x2A | 1 byte | Status (1-Registration, 2-Preparing, 3-Commenced) |
| 0x2C | 1 byte | Progress % |

There might be more to it, but this seems like the useful data.
@quisquous I haven't contributed to the plugin in a while, should PR's be targetted to the OverlayPlugin implementation or here?

quisquous:
It looks like eureka is still using cactbot's `onFateEvent`.  OverlayPlugin has fate watching code, but I do not see any events hooked up to it.  So, I would say add it to cactbot and it'll get merged into OverlayPlugin by ngld at a future time.

quisquous:
Abandoning this in favor of #1890.

PR #1888, opened 2020-10-17
resources: fix zone seal translation regexes

### LINKED ISSUE #1877: Emulator: Encounters not splitting correctly across zone seals
It looks like the logic that handles zone seals/unseals for the emulator is somewhat non-functional. Various different dungeons (including Baelsar's Wall, Bardam's Mettle, Fractal Continuum (hard), and Heroes' Gauntlet) either are recorded as one long encounter, or have only two encounters.

https://cdn.discordapp.com/attachments/760609440742899723/766194321061445632/unknown.png
https://cdn.discordapp.com/attachments/760609440742899723/766194571239096350/unknown.png
https://cdn.discordapp.com/attachments/760609440742899723/766194815041404948/unknown.png

As you can see, zone sealing/unsealing doesn't seem to be working correctly. One anomaly that I noticed in my logs is that in the ACT main interface, some boss encounters don't seem to have the zone seal message, even though it's present in the network log.

There's probably more to add for this, but it's late on the east coast and it's probably better to sleep on it and add pertinent information as it's asked for.

quisquous:
\y{Name} was meant to only work for player names.  So, when this changed
to only support 1-2 words of letters, then this broke zone seal messages
for some zones.

Fixes #1877.


quisquous:
@JLGarber does this fix the issue for you?

JLGarber:
From a brief look this seems to, yes. I did see one anomalous 1-second encounter where I think maybe the zone unseal message was offset from the win message by enough to trip the detection, but apart from that everything seems fine. Thanks for the quick fix after Valarnin identified this!

valarnin:
There's other stuff that should be updated as well to reflect this change. Let me get a list.

valarnin:
`docs/AdvancedCactbot.md` says that you can use it for ability names. I thought there was another place that it was used for not-player-names but my grep-skills are failing me.

quisquous:
Ok, I updated the documentation.  I also realized that `\y{Name}` was also supposed to be able to match the empty string.  So, I fixed that too.

I think part of the issue here is that I wanted to switch `\y{Name}` to be `.*?` but that broke some things by being overzealous in matching in some cases.  Sadly, past me did not leave any informative comments in ea1cff1a.  :anger:

PR #1890, opened 2020-10-18
Eureka: add Bozja Southern Front support

KattTails:
I've updated the eureka overlay to include the Bozjan Southern Front, including all FATEs and Critical Engagements. 

With Bozja having many more FATEs and generally being a lot more busy, before this is released there should be some way for the user to configure which FATEs/CEs they are interested in monitoring on the overlay whilst ignoring the rest.
A suggested default would be only to notify on CE pops and their respective chain FATEs.

As for how to implement this, the simplest way would be to have a list of preferred FATE IDs and CE keys in Options, and on FATE pop, if the id is not in this list, ignore it. This isn't very user friendly, and a nicer solution would be the implementation of a user control in the config, whereby users could choose from a list of all FATEs and CEs for which ones to monitor. I'm not confident I can implement that as my js is pretty weak, so help is welcome on that if possible.

Overall, the overlay is functional, but I'd like some general feedback and bug-testing. There's some possibility for additions with the new CE data, as we're exposed to information such as the time the FATE entered each state (registration, preparation and active), the number of sign-ups during the registration phase, and remaining time during the active phase. These could be used to add some visual flare.

As always, suggestions and feedback are welcome.


quisquous:
In terms of options in the config UI, I could imagine "make noise for all fates", "make noise for important fates (leading to CE)", "only make noise for CE" as potential options.  I think that would cover the vast majority of customization needs.

I'll try to take a look at this today or tomorrow.  I'm a little bit distracted trying to sort out some of the pre-Webpack issues.

quisquous:
If folks want to try this, there's a built version of the plugin with these changes attached as an artifact to the msbuild action: https://github.com/quisquous/cactbot/suites/1358588118/artifacts/22123544

ui/eureka/eureka.js
```diff
@@ -2139,6 +2611,48 @@ class EurekaTracker {
...
+    // Upon entering Eureka we usually receive the CE info before
+    // this.zoneInfo is loaded, so lets store the events until we're
+    // able to process them.
+
+    if (!this.zoneInfo) {
+      this.CEQueue.push(e);
+      return;
+    }
+
+    switch (e.detail.eventType) {
```
kuwoyuki, line 2624: This looks quite repetitive, could something like this work?
```js
const handleEvent = (handler) => {
  for (const key of this.nmKeys) {
    const nm = this.nms[key];
    if (e.detail.data.ceKey == nm.ceKey) {
      return handler(nm);
    }
  }
};

switch (e.detail.eventType) {
  case "add":
    handleEvent(this.OnFatePop);
    break;
  case "remove":
    handleEvent(this.OnFateKill);
    break;
  case "update":
    handleEvent(
      (nm) =>
        e.detail.data.status === 3 &&
        this.OnFateUpdate(nm, e.detail.data.progress)
    );
}
```

ui/eureka/eureka.js
```diff
@@ -2240,6 +2754,9 @@ UserConfig.getUserConfigLocation('eureka', Options, function(e) {
   addOverlayListener('onFateEvent', function(e) {
     gTracker.OnFate(e);
   });
+  addOverlayListener('onCEEvent', function(e) {
```
kuwoyuki, line 2757: This doesn't need to be a closure (the above listener as well),
```js
addOverlayListener('onCEEvent', gTracker.OnCE);
```

ui/eureka/eureka.js
KattTails: This looks pretty neat! Thanks for the suggestion, Sorry for taking a while to reply.

ui/eureka/eureka.js
KattTails: Yep they can be condensed, thanks!

KattTails: On reflection I'm going to leave these to keep consistent with the other overlays and pre-existing code.

ui/eureka/eureka.js
```diff
@@ -2139,6 +2611,40 @@ class EurekaTracker {
...
+
+    if (!this.zoneInfo) {
+      this.CEQueue.push(e);
+      return;
+    }
+
+    const handleEvent = (handler) => {
+      for (const key of this.nmKeys) {
+        const nm = this.nms[key];
+        if (e.detail.data.ceKey == nm.ceKey)
```
quisquous, line 2627: `===` instead of `==` please

ui/eureka/eureka.js
```diff
@@ -2139,6 +2611,40 @@ class EurekaTracker {
     }
   }
 
+  OnCE(e) {
+    // Upon entering Eureka we usually receive the CE info before
+    // this.zoneInfo is loaded, so lets store the events until we're
+    // able to process them.
```
quisquous, line 2617: Is there some way to not make pop noises for CEs that have already started?

quisquous (commented):
I sent a few extra changes in https://github.com/KattTails/cactbot/pull/1 which I think should be there when landing.  I can land those separately if you don't want to merge into your pull request.

quisquous:
Also as a bug report, after being in the zone for a while, sometimes I'd spot skirmishes on the game map that were not in the overlay map (active, in progress).  If I reloaded the overlay, it found them, which is surprising.  So, maybe sometimes they can get into a state where they cannot be readded? Or something like that? I have not been able to debug yet, but just mentoining it.  I think it would be fine to land without fixing it too, but I thought I would let you know.

ui/eureka/eureka.js
```diff
@@ -1966,6 +2430,9 @@ class EurekaTracker {
 
     document.getElementById('label-tracker').innerHTML = this.currentTracker;
 
+    if (this.zoneInfo.shortName == 'bozjasouthern')
```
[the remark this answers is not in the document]
KattTails: Yeah I think that'd be a cleaner way of doing this.

> 
> 
> In particular, doesn't this early out mean that castrum won't have a timer because it won't update?
> 
That's a good observation, I'll work on this.

ui/eureka/eureka.js
```diff
@@ -1700,14 +2124,19 @@ class EurekaTracker {
     this.ResetZone();
     this.updateTimesHandle = null;
     this.fateQueue = [];
+    this.CEQueue = [];
```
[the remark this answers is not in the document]
KattTails: done.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -138,68 +183,131 @@ private struct OPCodes {
...
-        RemoveFate(*(int*)&buffer[Param1_Offset]);
-      } else if (a == opcodes[region_].update) {
-        int param1 = *(int*)&buffer[Param1_Offset];
-        int param2 = *(int*)&buffer[Param2_Offset];
-        if (!fates.ContainsKey(param1)) {
-          AddFate(param1);
+    public unsafe void ProcessActorControl143(byte* buffer, byte[] message) {
+      int a = *(ushort*)&buffer[actorControl143.categoryOffset];
+
+      fateSemaphore.WaitAsync();
```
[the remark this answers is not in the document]
KattTails: Each network packet is handled in it's own thread, and from testing I was able to periodically get into a state where stale data was overwriting new data into the dictionaries. It would be possible to queue these up and process them in a single thread, but each transaction is very small (the longest probably being ```!ces[data.ceKey].Equals(data)```). I'm happy to go with your decision here but I think the mutexes are enough for now.

ui/eureka/eureka.js
KattTails: Fixed.

ui/eureka/eureka.js
KattTails: Yeah that should be simple to implement by passing the ceStatus as an optional parameter to onFatePop.

KattTails:
> 
> 
> Also as a bug report, after being in the zone for a while, sometimes I'd spot skirmishes on the game map that were not in the overlay map (active, in progress). If I reloaded the overlay, it found them, which is surprising. So, maybe sometimes they can get into a state where they cannot be readded? Or something like that? I have not been able to debug yet, but just mentoining it. I think it would be fine to land without fixing it too, but I thought I would let you know.

I've been idling in bozja on and off for the past week and catching issues like these every now and then, but mostly for the CEs and not skirmishes as that code was pretty solid from Eureka. I'll keep an eye out for it though.

ui/eureka/eureka.js
```diff
@@ -1700,14 +2125,19 @@ class EurekaTracker {
...
   }
 
-  TransByParserLang(obj) {
-    return obj[this.options.ParserLanguage] || obj['en'];
+  TransByParserLang(obj, key) {
+    if (key == null)
```
kuwoyuki, line 2132: Should probably be strict eq as well
```suggestion
    if (key === null)
```
or falsey check if it's not a guaranteed `null`?
```suggestion
    if (!key)
```

ui/eureka/eureka.js
```diff
@@ -1700,14 +2125,19 @@ class EurekaTracker {
...
+      return obj[this.options.ParserLanguage] || obj['en'];
+    return obj[this.options.ParserLanguage][key] || obj['en'][key];
   }
 
-  TransByDispLang(obj) {
-    return obj[this.options.DisplayLanguage] || obj['en'];
+  TransByDispLang(obj, key) {
+    if (key == null)
```
kuwoyuki, line 2138: this as well

ui/eureka/eureka.js
```diff
@@ -1832,15 +2273,18 @@ class EurekaTracker {
         aspect.classList.remove(aspect.classList.item(0));
       aspect.classList.add('aspect-ratio-' + this.zoneInfo.shortName);
 
-      document.getElementById('map-image').src = this.zoneInfo.mapImage;
+      if (this.zoneInfo.mapImage) {
+        document.getElementById('map-image').src = this.zoneInfo.mapImage;
+        window.clearInterval(this.updateTimesHandle);
+        this.updateTimesHandle = window.setInterval((function() {
+          this.UpdateTimes();
+        }).bind(this), this.options.RefreshRateMs);
```
kuwoyuki, line 2281: perfect use-case for an arrow function?
```suggestion
        this.updateTimesHandle = window.setInterval(
            () => void this.UpdateTimes(),
            this.options.RefreshRateMs
        );
```

ui/eureka/eureka.js
```diff
@@ -1966,6 +2439,9 @@ class EurekaTracker {
 
     document.getElementById('label-tracker').innerHTML = this.currentTracker;
 
+    if (this.zoneInfo.shortName == 'bozjasouthern')
```
kuwoyuki, line 2442:
```suggestion
    if (this.zoneInfo.shortName === 'bozjasouthern')
```

ui/eureka/eureka.js
```diff
@@ -2074,28 +2550,33 @@ class EurekaTracker {
...
         this.AddFlag(match[2], match[3], match[1], match[4]);
-      let gTrackerRegex = gRegex['gTrackerRegex'];
+
+      if (this.fairy) {
+        if (log.includes(' 03:') || log.includes('00:0839:')) {
+          match = log.match(this.fairy.regex);
+          if (match)
+            this.AddFairy(match.groups);
+        }
+      }
```
kuwoyuki, line 2565: this one can be flattened
```suggestion
      if (this.fairy && (log.includes(" 03:") || log.includes("00:0839:"))) {
        match = log.match(this.fairy.regex);
        if (match) this.AddFairy(match.groups);
      }
```

ui/eureka/eureka.js
```diff
@@ -2107,7 +2588,7 @@ class EurekaTracker {
       this.fateQueue.push(e);
       return;
     }
-
+    console.log(e.detail.eventType + ' | ' + e.detail.fateID + ' | ' + e.detail.progress);
```
kuwoyuki, line 2591: :eyes:

ui/eureka/eureka.js
```diff
@@ -2139,6 +2620,40 @@ class EurekaTracker {
...
+
+    if (!this.zoneInfo) {
+      this.CEQueue.push(e);
+      return;
+    }
+
+    const handleEvent = (handler) => {
+      for (const key of this.nmKeys) {
+        const nm = this.nms[key];
+        if (e.detail.data.ceKey == nm.ceKey)
```
kuwoyuki, line 2636: did the `===` get reverted? :eyes:

kuwoyuki (changes requested):
Just a few nits, not sure why there's no `eqeqeq` eslint rule

quisquous:
> Just a few nits, not sure why there's no `eqeqeq` eslint rule

Because there's a lot of old double equals lying around, and so you can't turn it on without fixing all of those.  I'm sure the vast majority of them can be converted to use triple equals, but it'll take some investigation to make sure that the changes are ok.  If you were interested in helping turn this on and wanted to take a closer look at converting them, I'd be happy to assist.

ui/eureka/eureka.js
quisquous: I think the previous code is more readable here, even if it's longer.

quisquous:
@KattTails I've been playing around with this and this all looks good.  I think if you have time to fix up the nits from kuwoyuki, then I think this is ready to be landed whenever you're ready.  I'd probably do a cactbot release to get this into people's hands.

ui/eureka/eureka.js
kuwoyuki: This handleEvent function definition should also go to a higher scope if possible, it doesn't make much sense to create a new function object every CE event.
Perhaps this should be a class method?

KattTails:
> 
> 
> @KattTails I've been playing around with this and this all looks good. I think if you have time to fix up the nits from kuwoyuki, then I think this is ready to be landed whenever you're ready. I'd probably do a cactbot release to get this into people's hands.

Sorry! I've been really busy, I'll try and get to this as soon as possible. I'm glad it's been working though!

PR #1896, opened 2020-10-20
raidboss: allow for timeline overriding

quisquous:
* Change the plugin to also load text files.
* Change user_config.js to allow a custom user file processor.
* Add a custom user file processor to raidboss_config.js.
* In this processor, convert timeline filenames into timeline contents,
  as user-specified timeline files are not part of the data files and so
  can't be loaded by name later.
* Add a new overrideTimelineFile flag that clears previously loaded
  timeline information.
* Add documentation for all of this.

### DIFF
--- a/docs/CactbotCustomization.md
+++ b/docs/CactbotCustomization.md
@@ -294,7 +294,43 @@ and also reading through existing triggers in [ui/raidboss/data](../ui/raidboss/
 
 ## Overriding Raidboss Timelines
 
-Sorry, this is hard to do right now.
+Overriding a raidboss timeline is similar to [overriding a trigger](#overriding-raidboss-triggers).
+
+The steps to override a timeline are:
+
+1) Copy the timeline text file out of cactbot and into your user folder
+
+  For example, you could copy [ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.txt](../ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.txt).
+
+1) Add a section to your user/raidboss.js file to override this.
+
+  Like adding a trigger, you add a section with the `zoneId`,
+  along with `overrideTimelineFile: true`,
+  and a `timelineFile` with the name of the text file.
+
+  ```javascript
+  Options.Triggers.push({
+    zoneId: ZoneId.TheEpicOfAlexanderUltimate,
+    overrideTimelineFile: true,
+    timelineFile: 'the_epic_of_alexander.txt',
+  });
+  ```
+
+  In this case, this assumes that you have followed step 1
+  and there is a `user/the_epic_of_alexander.txt` file.
+
+  By setting `overrideTimelineFile: true`,
+  it tells cactbot to replace the built-in timeline entirely
+  with any new timeline that you add.
+
+1) Edit your new timeline file in your user folder as needed
+
+  Refer to the [timeline guide](TimelineGuide.md) for more documentation on the timeline format.
+
+**Note**: Editing timelines is a bit risky,
+as there may be timeline triggers that refer to specific timeline text.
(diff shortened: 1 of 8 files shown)


quisquous:
I added some logging if timeline triggers no longer applied and hooked it up to cactbot tests.  Unsurprisingly, there were a few cactbot test failures.  So, I'll submit that change separately.  :innocent:

docs/CactbotCustomization.md
```diff
@@ -294,7 +294,43 @@ and also reading through existing triggers in [ui/raidboss/data](../ui/raidboss/
...
 ## Overriding Raidboss Timelines
 
-Sorry, this is hard to do right now.
+Overriding a raidboss timeline is similar to [overriding a trigger](#overriding-raidboss-triggers).
+
+The steps to override a timeline are:
+
+1) Copy the timeline text file out of cactbot and into your user folder
+
+  For example, you could copy [ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.txt](../ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.txt).
```
panicstevenson, line 303:
```suggestion
  For example, you could copy [ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.txt](../ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.txt) to `user/the_epic_of_alexander.txt`.
```

panicstevenson (approved):

docs/CactbotCustomization.md
```diff
@@ -294,7 +294,43 @@ and also reading through existing triggers in [ui/raidboss/data](../ui/raidboss/
...
+
+1) Add a section to your user/raidboss.js file to override this.
```
JLGarber, line 305:
```suggestion
2) Add a section to your user/raidboss.js file to override this.
```

docs/CactbotCustomization.md
```diff
@@ -294,7 +294,43 @@ and also reading through existing triggers in [ui/raidboss/data](../ui/raidboss/
...
+  ```
+
+  In this case, this assumes that you have followed step 1
+  and there is a `user/the_epic_of_alexander.txt` file.
+
+  By setting `overrideTimelineFile: true`,
+  it tells cactbot to replace the built-in timeline entirely
+  with any new timeline that you add.
+
+1) Edit your new timeline file in your user folder as needed
```
JLGarber, line 326:
```suggestion
3) Edit your new timeline file in your user folder as needed
```

JLGarber (commented):
This all looks *very* good. It's a bit of a complex setup for the average user to do this from scratch, but I don't think there's a simpler way to do it. Conversely, this should make it extremely easy to hand a user a timeline file and a code block to paste into `user/raidboss.js`. Explicitly labeling any overrides as being user-generated for logging purposes should help out on troubleshooting.

docs/CactbotCustomization.md
panicstevenson: Actually these are pretty funny (read: not funny at all); so, Markdown in general lets you do the following for numbered lists:

```md
1) Test
1) Test
```
And that will be:
1) Test
1) Test

But, because the numbered items have paragraphs between them that are improperly indented (Markdown says 3 or more is the minimum for indentation), these do actually come out to be:

```md
1.
1.
1.
```
In the preview.

So, either:
```md
1) Point

  Text

2) Point

  Text
```
or:
```md
1) Point

   Text

1) Point

   Text
```
Thanks Markdown 🙄

JLGarber: How very interesting. No, that auto-numbering was something I wasn't aware of. At least it's a "simple" fix, right?

panicstevenson: I wouldn't necessarily say that writing your own markup language to replace Markdown entirely is simple, but... 😏

PR #1897, opened 2020-10-20
raidboss: validate that all timeline triggers match something

quisquous:
This will also print log messages, so that if users edit timeline
files and break timeline triggers, there will be some warning.

This will help with #1896.

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -153,6 +153,14 @@ class Timeline {
     let uniqueid = 1;
     let texts = {};
 
+    // Make all regexes case insensitive, and parse any special \y{} groups.
+    if (triggers) {
+      for (const trigger of triggers) {
+        if (trigger.regex)
+          trigger.regex = Regexes.parse(trigger.regex);
+      }
+    }
+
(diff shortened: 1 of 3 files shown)


ui/raidboss/timeline.js
```diff
@@ -277,11 +285,28 @@ class Timeline {
...
 
-    for (let i = 0; i < this.events.length; ++i) {
-      let e = this.events[i];
+    // Validate that all timeline triggers match something.
+    if (triggers) {
+      for (const trigger of triggers) {
+        let found = false;
+        for (const event of this.events) {
+          let m = event.name.match(trigger.regex);
+          if (m) {
```
panicstevenson, line 294: nit: This feels odd since we don't actually use `m`.

panicstevenson (approved):

ui/raidboss/timeline.js
quisquous: Done, thanks!

kuwoyuki: should've been `regExp.test` too 👀 , it's a much faster boolean check

PR #1900, opened 2020-10-21
haukke_manor.js

Saulchar:
Haukke Manor normal triggers are ready


ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,103 @@
...
+      regex: /The void lamps have begun/,
+      infoText: function(data) {
+      return {
+          en: 'Turn off Lamps',
+        };
+      },
+    },
+    {
+      id: 'Haukke Normal Ladys Candle',
+      regex: /Added new combatant Lady's Candle/,
```
MaikoTan, line 84:
```suggestion
      netRegex: NetRegexes.addedCombatant({ name: 'Lady\'s Candle', capture: false }),
```

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,103 @@
...
+      response: Responses.killAdds('info'),
+    },
+    {
+      id: 'Haukke Normal Ladys Candle Self Destruct',
+      netRegex: NetRegexes.startsUsing({ id: '151' }),
+      response: Responses.getOut('alert'),
+    },
+    {
+      id: 'Haukke Normal Ladys Handmaiden',
+      regex: /Added new combatant Lady's Handmaiden/,
```
MaikoTan, line 94:
```suggestion
      netRegex: NetRegexes.addedCombatant({ name: Lady\'s Handmaiden', capture: false }),
```

MaikoTan (changes requested):
`NetRegexes.addedCombatant` is recommanded

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
Saulchar: Added your changes Maiko, thanks. Didnt see the addedCombatant netregex at first

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+
+[{
+  zoneId: ZoneId.HaukkeManor,
+  timelineFile: 'haukke_manor.txt',
+  triggers: [
+    {
+      id: 'Haukke Normal Dark Mist DPS/Heals',
+      netRegex: NetRegexes.startsUsing({ id: '2C1', source: 'Manor Claviger || Manor Maidservant || Lady Amandine' }),
```
MaikoTan, line 9: if there are multiple names, you can just list them in an array
```suggestion
      netRegex: NetRegexes.startsUsing({ id: '2C1', source: ['Manor Claviger', 'Manor Maidservant', 'Lady Amandine'] }),
```

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,103 @@
...
+      id: 'Haukke Normal Ladys Handmaiden',
+      netRegex: NetRegexes.addedCombatant({ name: Lady\'s Handmaiden', capture: false }),
+      suppressSeconds: 2,
+      infoText: function(data) {
+      return {
+      en: 'Kill Adds, Sentry First',
+       };
+      },
+    },
+  ],
```
MaikoTan, line 102: Sorry I missed a quote.

And you should add indentation in every function.

```suggestion
    {
      id: 'Haukke Normal Ladys Candle',
      netRegex: NetRegexes.addedCombatant({ name: 'Lady\'s Candle', capture: false }),
      response: Responses.killAdds('info'),
    },
    {
      id: 'Haukke Normal Ladys Candle Self Destruct',
      netRegex: NetRegexes.startsUsing({ id: '151', name: 'Lady\'s Candle' }),
      response: Responses.getOut('alert'),
    },
    {
      id: 'Haukke Normal Ladys Handmaiden',
      netRegex: NetRegexes.addedCombatant({ name: 'Lady\'s Handmaiden', capture: false }),
      suppressSeconds: 2,
      infoText: function(data) {
        return {
          en: 'Kill Adds, Sentry First',
         };
      },
    },
  ],
```

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,103 @@
...
+      },
+    },
+    {
+      id: 'Haukke Normal Ladys Candle',
+      netRegex: NetRegexes.addedCombatant({ source: 'Lady\'s Candle', capture: false }),
+      response: Responses.killAdds('info'),
+    },
+    {
+      id: 'Haukke Normal Ladys Candle Self Destruct',
+      netRegex: NetRegexes.startsUsing({ id: '151', source: 'Lady\'s Candle' }),
```
MaikoTan, line 89:
```suggestion
      netRegex: NetRegexes.startsUsing({ id: '151', source: 'Lady\'s Candle' }),
```

According to the test output, `startUsing` accept `source`, not `name`.

Saulchar: Yep, noticed that in the testing, already got it. Thanks though

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,102 @@
...
+      condition: function(data) {
+        return data.role != 'tank';
+      },
+      suppressSeconds: 2,
+      response: Responses.getOut('alert'),
+    },
+    {
+      id: 'Haukke Normal Dark Mist Tank',
+      netRegex: NetRegexes.startsUsing({ id: '2C1', source: ['Manor Maidservant', 'Manor Claviger', 'Lady Amandine'] }),
+      condition: function(data) {
```
JLGarber, line 18: Here and elsewhere, both for situations where you need to stun and for interrupts, you're better off using the `data.canStun()` or `data.canSilence()` construction, rather than just restricting it to tanks.

Also, is Amandine stun-vulnerable? I haven't tested it myself, but I always thought final bosses were immune even in ARR.

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,102 @@
...
+      id: 'Haukke Normal Amandine Void Thunder III',
+      netRegex: NetRegexes.startsUsing({ id: '358', source: 'Manor Claviger' }),
+      condition: function(data) {
+        return data.role == 'tank';
+      },
+      response: Responses.interrupt('info'),
+    },
+    {
+      id: 'Haukke Normal Void Lamps',
+      regex: /The void lamps have begun/,
```
JLGarber, line 74: There is actually a [`netRegex` helper](https://github.com/quisquous/cactbot/blob/0712f53cdde5ae5a8e49fe22cf82c2c8a9047d45/resources/netregexes.js#L286) for this too!
```suggestion
      netRegex: NetRegexes.message({ line: 'The void lamps have begun emitting an eerie glow*?' }),
```

Typically, unless there's a specific reason not to, it's recommended that you use the helpers instead of bare regexes. It makes reviewing and debugging easier.

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,102 @@
...
+      infoText: function(data) {
+        return {
+          en: 'Turn off Lamps',
+        };
+      },
+    },
+    {
+      id: 'Haukke Normal Ladys Candle',
+      netRegex: NetRegexes.addedCombatant({ name: 'Lady\'s Candle', capture: false }),
```
JLGarber, line 83: This one is a personal opinion, so don't worry about it too much, but an alternative you can do here is to use [`addedCombatantFull`](https://github.com/quisquous/cactbot/blob/0712f53cdde5ae5a8e49fe22cf82c2c8a9047d45/resources/netregexes.js#L156), and then use the NPC Name ID instead of the name. This has the advantage that you don't need to translate the combatant name for each language. So this line might look like

```suggestion
      netRegex: NetRegexes.addedCombatantFull({ npcNameId: '425', capture: false }),
```

If you do this, please be sure that the name of the NPC is either in the trigger ID, like you've done here, or at the very least is in a comment line next to the trigger.

JLGarber (commented):
Overall this looks very nice! I'm glad to see other people contributing trigger packages. It looks like you've done a really good job of figuring things out and asking good questions to fill in the gaps that the docs didn't cover.

I saw your remarks in the ACT Discord server about the difficulties you were experiencing with timelines. Could you perhaps expand on those difficulties here? It's entirely possible the game just won't let you do timelines for this, but if it's something where we can help you figure out quirks of the system, we would like to do that.

Finally, as a general rule, we don't typically notify the user to move on visible AoEs. I know you said that you tend to be forgetful about that and need the reminder, but usually we don't do that. (One specific exception that we've made is where there's an in/out mechanic where one of the AoEs is telegraphed and the other isn't.) It's very possible we'll make another exception here, since it's early early game content, but be aware of that for the future.

I look forward to seeing more stuff from you. It will be nice to shrink the backlog of ARR dungeons!

panicstevenson:
Not a comment, but nice to see a new face! 👋

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
Saulchar: After logging 5-10 runs so that I could create timelines, during rigorous testing to ensure matching once they had passed test_timeline satisfactorily, I learned that the bosses in there often delayed their abilities. I would assume this was because they have so few as Haukke Normal level, but most of the time, timelines seemed to be off by 1.5-2 seconds at various intervals throughout the fight, and they seemed random from pull to pull, so I just scrapped the timelines altogether until I get to something higher that's a bit more scripted.

Saulchar: Also, for your comment about Amandine, I will double-check her stun vulnerability right now; actually, you bring up a good point. You are correct about changing to data.canStun(),. I didn't do it at first because I wasn't sure if all the roles had their stuns and interrupts by this level, but it turns out they all do, so I will go ahead and change that.

In terms of not triggering for visible AoEs, I will remove the generic all instance triggers for Dark Mist for anyone without stuns and the trigger for Soul Drain as I would like to keep this consistent with the rest of the repo. I want to make a separate Dark Mist Trigger that applies to everyone that is restricted to Lady Amandine, though. With all the particle effects in her room and all the spell effects from players, sometimes you can't see her Dark Mist AoE at all if she's sitting in the fog at the edges of the room for whatever reason.  This mostly seems to depend on where she is being tanked, whether it be for a reason or simply because the tank is a new player. I want to keep a Dark Mist trigger for Amandine in, though, simply to alleviate issues of possibly getting hit by a mechanic you should be able to see but can't due to room graphics.

Saulchar: Amandine is stun vulnerable by the way.

JLGarber: Okay, yes, thanks for removing those others. I thought about it more today and talked it over, and given that we aren't doing that except for specific situations elsewhere in the repository, it doesn't make sense to start here for the non-Amandine ones. @quisquous will probably have to weigh in on Amandine, but definitely for the future avoid it unless there's a specific reason for it. (An example of where I argued for this is on [Sophia EX](https://github.com/quisquous/cactbot/blob/0712f53cdde5ae5a8e49fe22cf82c2c8a9047d45/ui/raidboss/data/03-hw/trial/sophia-ex.js#L142), as a reference point.)

JLGarber:
It's not really a problem for the repository, since quisquous ends up squashing most commits anyway, but in order to save yourself some time, it's *way* easier to do `npm run test` locally, if you're using NPM. (If you're not and would like to start, I'm sure @quisquous or @panicstevenson have an idea on resources that could help you get started!)

panicstevenson:
RE: JLGarber:
That documentation would exist [here](https://github.com/quisquous/cactbot/blob/main/CONTRIBUTING.md#development-workflow). And, while it's still probably a worthwhile read, I don't think it notes NodeJS/npm/Python/anything else that might be necessary to get started developing. We should fix that!

JLGarber:
Hah. Sorry for throwing you and quisquous under the bus on that one. I'm definitely not the person to give instruction on that since I was barely able to get it set up for myself when I started contributing!

Saulchar:
Added Brayflox Normal Triggers

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
JLGarber: I made a mistake in the casing on the conditions there. It should be `CanStun` or `CanSilence`. See the [raidboss config](https://github.com/quisquous/cactbot/blob/672f8e949bd03643c234ad453875eac562b75a33/ui/raidboss/raidboss_config.js#L375) for the full selection. (Also, it's `CanSilence`, not `CanInterrupt`, because these utility functions were added in Stormblood when that was the terminology.)

I think once you fix those you should see your triggers working. Or, at the very least, when I swapped these out, It Worked On My Machine.

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,85 @@
...
+    },
+    {
+      // Void Lamp Spawn
+      id: 'Haukke Normal Void Lamps',
+      netRegex: NetRegexes.message({ line: 'The void lamps have begun emitting an eerie glow.', capture: false }),
+      infoText: function(data) {
+        return {
+          en: 'Turn off Lamps',
+        };
+      },
```
JLGarber, line 62: Here and for the Handmaiden trigger, it doesn't look like you're using the `infotext` function for anything. Is this a leftover from a previous version?

Saulchar: I thought I had to use infoText to return the statement as the correct type (info, alert, or alarm) to make sure its not bigger then it really needs to be. I will change it to condition though. Thank you so much for catching the CanStun and CanSilence issues, those were driving me bonkers!

JLGarber: Right, no, if you don't need logic in the popup text, you can just return a bare localized object, as
```
infoText: {
  en: 'Some Unchanging Popup Text',
}
```

The only time you need to return a function instead of a bare object is the situation where you're injecting something from the `data` global object, or where you're using the `matches` object from the regex. For instance, here's the logic for [`stackMarkerOn`](https://github.com/quisquous/cactbot/blob/0712f53cdde5ae5a8e49fe22cf82c2c8a9047d45/resources/responses.js#L447):

```
  lookAwayFromTarget: (sev) => {
    let obj = {};
    obj[defaultAlertText(sev)] = (data, matches) => {
      const target = getTarget(matches);
      if (target === data.me)
        return;
      const name = isPlayerId(matches.targetId) ? data.ShortName(target) : target;
      return {
        en: 'Look Away from ' + name,
        de: 'Schau weg von ' + name,
        fr: 'Ne regardez pas ' + name,
        ja: name + 'を見ない',
        cn: '背对' + name,
        ko: name + '에게서 뒤돌기',
      };
    };
    return obj;
  },
```

As you can see, the function takes both `data` and `matches` as arguments, processes `matches` to find the name of the target, and then uses the `ShortName` function from `data` to return a localized object with a customized target name.

Sometimes you'll have a parameter stored on the `data` object but not need the `match` parameter. Let's say that the boss can cast Half Vanish, which makes either the north or the south side of the room unsafe. Which half depends on whether the boss has previously cast Northern Cross or Southern Cross. (This is all made-up for this example. of course.) So you would have one trigger that checks for which Cross is cast, (let's say North,) and that trigger would store it as `data.north = true;`. Then, when the boss casts Half Vanish, you would set the trigger up like this:

```
{
  id: 'Made-up Half Vanish',
  netRegex: NetRegexes.startsUsing({ id: '9876', source: 'Giant Monster', capture: false }),
  alertText: function(data) {
    if (data.north) {
      return {
        /en: 'Go North!',
      };
    }
    return {
      en: 'Go South!',
    }
  }
}
```
Note that regex capturing isn't needed, since we don't care about the particulars of the regex. We just care that there was a match at all.

Sorry for the wall of text. This is all stuff that's easy to use once you know it, but it can be tricky to step through the pieces. There are a lot of details!

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -7,8 +7,8 @@
       // Dark Mist Stun with Suppression for duplicate messages
       id: 'Haukke Normal Dark Mist Stun',
       netRegex: NetRegexes.startsUsing({ id: '2C1', source: ['Manor Maidservant', 'Manor Claviger', 'Lady Amandine'] }),
-      condition: function(data) {
-        return data.canStun();
+      condition: function(data, matches) {
+        return data.CanStun();
       },
       suppressSeconds: 2,
       response: Responses.stun('info'),
```
JLGarber: Note that here and elsewhere, the condition doesn't require `matches`, just `data`. An individual player's ability to stun is not dependent on the results of the regex match.

(That being said, `responses.stun()` *does* require the captured match results, so this particular trigger needs to capture, which you have it correctly set up to do. If you're ever concerned about whether a `response` helper needs the results of the regex match, you can [look it up](https://github.com/quisquous/cactbot/blob/main/resources/responses.js) and check what its parameters are.)

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -27,38 +27,36 @@
...
       // Callout ensures dodge.
       id: 'Haukke Normal Amandine Dark Mist Dodge',
-      netRegex: NetRegexes.startsUsing({ id: '2C1', source: 'Lady Amandine', capture: false }),
-      condition: function(data) {
-        return !data.canStun();
+      netRegex: NetRegexes.startsUsing({ id: '2C1', source: 'Lady Amandine' }),
+      condition: function(data, matches) {
+        return !data.CanStun();
       },
       response: Responses.outOfMelee('alert'),
```
JLGarber: Here, unlike your other  triggers, there shouldn't be a need to capture. `outOfMelee()` takes no arguments, so `matches` won't be needed there, and as noted elsewhere, you don't need `matches` to check whether a job can stun/silence.

(As a general rule, we try to avoid superfluous function parameters where possible. Basically, if a parameter is not used, and it is not "in front of" a parameter that is being used, you can safely not include it.)

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,81 @@
...
+    {
+      // 2 Lady's Handmaiden and 1 Manor Sentry Spawn
+      // The sentry outside the bosses room loads when you enter the zone.
+      // This causes the trigger to go off early, parsing for the Handmaiden fixes the problem.
+      // Suppression included since 2 Handmaiden's spawn at the same time
+      id: 'Haukke Normal Ladys Handmaiden',
+      netRegex: NetRegexes.addedCombatantFull({ npcNameId: '424', capture: false }),
+      suppressSeconds: 2,
+      infoText: {
+        en: 'Kill Adds, Sentry First',
```
quisquous, line 77: Do you think this should just say "Kill Sentry", since that's the thing to focus on first? Maybe also with an alert instead of info?

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,81 @@
...
+      netRegex: NetRegexes.startsUsing({ id: '356', source: 'Lady Amandine' }),
+      condition: function(data, matches) {
+        return data.CanSilence();
+      },
+      response: Responses.interrupt('info'),
+    },
+    {
+      // Void Thunder 3
+      id: 'Haukke Normal Amandine Void Thunder III',
+      netRegex: NetRegexes.startsUsing({ id: '358', source: 'Manor Claviger' }),
```
quisquous, line 48: Should this be Lady Amandine?

Also, should trigger be `Responses.getBehind()`? Am I misremembering, or can you line of sight Thunder III like you can do for Slipstream?

quisquous (commented):
Thanks for this! Here's a few small comments.

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,81 @@
...
+  zoneId: ZoneId.HaukkeManor,
+  triggers: [
+    {
+      // Dark Mist Stun with Suppression for duplicate messages
+      id: 'Haukke Normal Dark Mist Stun',
+      netRegex: NetRegexes.startsUsing({ id: '2C1', source: ['Manor Maidservant', 'Manor Claviger', 'Lady Amandine'] }),
+      condition: function(data, matches) {
+        return data.CanStun();
+      },
+      suppressSeconds: 2,
```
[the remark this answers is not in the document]
Saulchar: Yes, the hallway before the first boss has two Maidservants that like to walk around together, so if one comes you're often getting both. If you don't, the one that is left will often come in with the that's right in front of the first boss door. Im using suppressSeconds to avoid multiple calls and double printing for classes with stuns.

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
Saulchar: Agreed, the trigger Im using is wordy

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
Saulchar: Amandine tracks her target unfortunately with Void Thunder. You are correct it is Amandine

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,81 @@
...
+      },
+      response: Responses.interrupt('info'),
+    },
+    {
+      // Void Thunder 3
+      id: 'Haukke Normal Amandine Void Thunder III',
+      netRegex: NetRegexes.startsUsing({ id: '358', source: 'Lady Amandine' }),
+      condition: function(data, matches) {
+        return data.CanSilence();
+      },
```
quisquous, line 51: Should this be `condition: Conditions.targetIsYou(),`?

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
```diff
@@ -0,0 +1,81 @@
...
+      // Callout ensures dodge.
+      id: 'Haukke Normal Amandine Dark Mist Dodge',
+      netRegex: NetRegexes.startsUsing({ id: '2C1', source: 'Lady Amandine' }),
+      condition: function(data, matches) {
+        return !data.CanStun();
+      },
+      response: Responses.outOfMelee('alert'),
+    },
+    {
+      // Void Fire 3
```
quisquous, line 37: One last tiny nit here is that I don't think there's a need for these comments here when the `id` also explains the ability name.  Sometimes it can be ambiguous and comments are great, but in these cases I think the information is redundant.

quisquous (commented):
Sorry for all the feedback! I think it will only get easier from here.  I have two more small comments, and then this should be good to merge.

ui/raidboss/data/02-arr/dungeon/haukke_manor.js
quisquous: (It looks like this is still the case in your latest patch.)

PR #1903, opened 2020-10-22
refactor(perf): change string.match to regExp.test

kuwoyuki:
Replace all string.match with regExp.test where applicable, also precompile regex in timeline.js `LoadFile()` since they're used in a loop which is a big performance penalty.  
Fixed a4s missing Hydrothermal Missile regex which caused the tests to fail (`!!event.name.match(undefined) === true`)

`string.match` vs `regExp.test` bench http://jsbench.github.io/#edf82fee9c75a58c7f48889ff0449453

### DIFF
--- a/ui/radar/radar.js
+++ b/ui/radar/radar.js
@@ -36,7 +36,7 @@ const kMinDistanceBeforeSound = 100;
 
 let gRadar;
 
-const instanceChangedRegex = {
+const instanceChangedRegexes = {
   en: NetRegexes.gameLog({ code: '0039', line: 'You are now in the instanced area.*?' }),
   de: NetRegexes.gameLog({ code: '0039', line: 'Du bist nun in dem instanziierten Areal.*?' }),
   fr: NetRegexes.gameLog({ code: '0039', line: 'Vous êtes maintenant dans la zone instanciée.*?' }),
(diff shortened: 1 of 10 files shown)


ui/radar/radar.js
```diff
@@ -292,8 +296,7 @@ class Radar {
 
     // change instance
     if (type === '00') {
-      const lang = this.options.ParserLanguage;
-      if (log.match(instanceChangedRegex[lang] || instanceChangedRegex['en'])) {
```
kuwoyuki, line 296: I'm assuming this was a bug since the latter part of the `or` was testing for the English regex existence, so I changed that but if it was intentional I'll revert that :upside_down_face:

ui/radar/radar.js
```diff
@@ -292,8 +296,7 @@ class Radar {
...
+      if (this.instanceChangedRegex.test(log)) {
```
kuwoyuki, line 299: I'm assuming this was a bug since the latter part of the `or` was testing for the English regex existence, so I changed that but if it was intentional I'll revert that :upside_down_face:

ui/raidboss/timeline.js
```diff
@@ -290,7 +304,7 @@ class Timeline {
       for (const trigger of triggers) {
         let found = false;
         for (const event of this.events) {
-          if (event.name.match(trigger.regex)) {
+          if (trigger.regex && trigger.regex.test(event.name)) {
```
quisquous, line 307: Thanks, I think this is better, because it means that anything missing a regex will throw an error.

quisquous:
Thanks for this! I wish there was a way to make eslint complain about these so we don't add more in the future, but these are all good changes.  Microoptimizations like this that don't hurt the readability of the code are great.

PR #1905, opened 2020-10-24
Brayflox normal

Saulchar:
Brayflox Triggers


JLGarber:
(quisquous can probably give you better information than I can, but typically the way I handle branching is always to start from a clean Main, after merging any commits in from upstream. I then *only* do work on any given dungeon or instance after first branching from Main at that point. This way you don't risk cross-contamination, and you won't end up with a giant tower of commits like you have here.)

The issue you're experiencing with those triggers is that neither of those abilities is a `startsUsing()` situation. Both are instant-cast, so those triggers never activate. For both of them, you should be okay to use `netRegexes.ability()` instead.

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,114 @@
...
+      netRegex: NetRegexes.startsUsing({ id: '1FA', source: 'Great Yellow Pelican' }),
+      condition: function(data, matches) {
+        return data.CanStun();
+      },
+      response: Responses.stun('info'),
+    },
+    {
+      // Pelican Adds
+      id: 'Brayflox Normal Pelican Adds',
+      netRegex: NetRegexes.message({ line: 'A flock of ziz emerges from the jungle!', capture: false }),
```
quisquous, line 18: Is it possible to use an addedCombatant trigger here rather than matching against game logs?

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,114 @@
...
+        return data.CanStun();
+      },
+      response: Responses.stun('info'),
+    },
+    {
+      // Hellbender Bubble Others
+      id: 'Brayflox Normal Hellbender Effluvium Others',
+      netRegex: NetRegexes.startsUsing({ id: '3D3', source: 'Hellbender' }),
+      condition: function(data, matches) {
+        return matches.target != data.me;
```
quisquous, line 50: In general, prefer using `===` and `!==` instead of `==` and `!=` (here and elsewhere in this patch).  In this case, you can also say `condition: Conditions.targetIsNotYou(),`

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,114 @@
...
+    {
+      // Hellbender Bubble You
+      id: 'Brayflox Normal Hellbender Effluvium You',
+      netRegex: NetRegexes.startsUsing({ id: '3D3', source: 'Hellbender' }),
+      condition: function(data, matches) {
+        return matches.target == data.me;
+        },
+      alertText: function(data, matches) {
+        return {
+          en: 'Break Bubble on YOU'
```
quisquous, line 67: I realize this is the way a lot of other triggers are written, but probably "Break Your Bubble" reads better.  I think of this kind of thing as "Mechanic on YOU" and I don't think "Break Bubble" is the mechanic, if that makes sense.

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,114 @@
...
+    },
+    {
+      // Healer Esuna Poison Application on Tank
+      id: 'Brayflox Normal Aiatar Salivous Snap Healer',
+      netRegex: NetRegexes.startsUsing({ id: '22E', source: 'Aiatar' }),
+      condition: function(data, matches) {
+        return data.role == 'healer';
+      },
+      infoText: {
+        en: 'Esuna Poison',
```
quisquous, line 88: Can this say who to esuna? (Here, and the other esuna as well?)

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,114 @@
...
+      },
+    },
+    {
+      // Move Aiatar out of Puddles
+      id: 'Brayflox Normal Aiatar Toxic Vomit Tank',
+      netRegex: NetRegexes.message({ line: 'Aiatar is using the poison from the pools to restore its HP!' }),
```
quisquous, line 94: Is there an effect or anything else other than a game log here?

quisquous:
Also, you can run `npm run lintfix` to automatically fix a number of the formatting issues like spacing and missing commas.

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
JLGarber: `26|2020-09-24T19:24:13.1310000-04:00|117|Rehabilitation|9999.00|E0000000||40003AA0|Aiatar|00|33099|||ETC`

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
JLGarber: The NPC Name IDs for the adds are `['1281', '1282', 1283']`. (Blue Back, Violet Back, and Sable Back.) You will want to check for any one of them and include `suppressSeconds`, since there are packs that don't include all three.

Issue #1886, opened 2020-10-17
Support templated strings as trigger output for text customization


quisquous:
One longstanding request is to be able to customize the text output of triggers.  Unfortunately this can be really complicated when triggers have so much JavaScript baked in.

ngld suggests in #1869 a potential way to make trigger output customizable:

> Making trigger text configurable could be fairly easy IMO. If you add a new property to triggers with a list of templates with placeholders (Get Final Nisi (?), Pass {nisi} Nisi, Pass {nisi} Nisi to {player}, ...), the JS trigger could then just pick a string from that list (or map) and fill in the placeholder. The config UI could then offer options to override the text for simple triggers and a more complicated interface for these triggers where it'd show a list of templates and allow the user to modify/override them.

## A half-baked proposal

(I'm making this up just to have something to talk about, so feel free to suggest other/better details)

Triggers now have a `outputStrings` field on them.  `outputStrings` is an object that maps keys to trigger text.  Trigger text must be either a string of a translatable object of strings, i.e. `{en: 'str1', de: 'str2'}` etc.  It must not be a function.

If this new `outputStrings` field exists, then it changes the trigger into an "output string only" trigger.  Any `infoText` / `alertText` / `alarmText` / `ttsText` output must output a key in the `outputStrings` field rather than outputting the literal text.  Then, the trigger would output whatever's in the `outputString`s field that corresponds to the key for that trigger.  Responses themselves can contain `outputStrings`.  We can also add common ones to `resources/translations.js` in some new object.

Once all/many triggers use `outputStrings` then we can add configuration UI to override the output of these.  We should probably deprecate the `Options` trigger overrides for text at this point, at least in the sense of removing it from the documentation, and maybe adding console warnings for anybody using it to change to the newer thing.

### Open questions
* Is there any way to verify at build time whether the keys into outputStrings are valid? (i.e. no typos or missing keys)
* Similarly, it'd be nice to verify that all language have the same set of interpolation fields?
* How do we validate interpolation fields?
* What are the set of fields that can be used for template interpolation? Do we allow data and matches? What about custom computed variables from a JavaScript function? The example below assumes that `data` and `matches` will be available to the interpolation, but what about other variables? It seems like it'd be better to store some `puddleNumber` variable for the interpolation to use rather than `matches[1]` but it also seems not great to have to store them in `data` directly, muddying that up?
[...]

PR #1907, opened 2020-10-25
raidboss: Add outputStrings and cactbot UI for customizing trigger output

quisquous:
This is an implementation of #1886, with a proof-of-concept conversion of a few triggers.

### DIFF
--- a/docs/CactbotCustomization.md
+++ b/docs/CactbotCustomization.md
@@ -50,7 +50,7 @@ ACT -> Plugins -> OverlayPlugin.dll -> Cactbot -> Raidboss,
 there are individual trigger listings.
 You can use these listings to change various exposed configuration settings per trigger.
 
-Settings with a bell (🔔) next to their name are trigger mostly toutputs that you can override.
+Settings with a bell (🔔) next to their name are trigger mostly outputs that you can override.
 For example, maybe there's an 🔔onTarget field whose text is `Tank Buster on ${name}`.
 This is the string that will get played on screen (or via tts) when there is a tank buster on some person.
 `${name}` here is a parameter that will be set dynamically by the trigger.
--- a/docs/CactbotCustomization.md
+++ b/docs/CactbotCustomization.md
@@ -3,6 +3,7 @@
 🌎 [**English**] [[한국어](./ko-KR/CactbotCustomization.md)]
 
 - [Using the cactbot UI](#using-the-cactbot-ui)
+- [Changing Trigger Text with the cactbot UI](#changing-trigger-text-with-the-cactbot-ui)
 - [User Directory Overview](#user-directory-overview)
 - [Setting Your User Directory](#setting-your-user-directory)
 - [Customizing Appearance](#customizing-appearance)
@@ -27,6 +28,7 @@ This has options for things like:
 
 - setting triggers to tts
 - disabling triggers
+- changing the output of triggers
 - changing your cactbot language
 - volume settings
 - getting rid of that cheese icon
@@ -40,6 +42,30 @@ These options are stored in your
 `%APPDATA%\Advanced Combat Tracker\Config\RainbowMage.OverlayPlugin.config.json`
 file.
 You should not need to edit that file directly.
+
+## Changing Trigger Text with the cactbot UI
+
+In the cactbot configuration UI, under
+ACT -> Plugins -> OverlayPlugin.dll -> Cactbot -> Raidboss,
+there are individual trigger listings.
+You can use these listings to change various exposed configuration settings per trigger.
+
+Settings with a bell (🔔) next to their name are trigger mostly toutputs that you can override.
+For example, maybe there's an 🔔onTarget field whose text is `Tank Buster on ${name}`.
+This is the string that will get played on screen (or via tts) when there is a tank buster on some person.
+`${name}` here is a parameter that will be set dynamically by the trigger.
+Anything that looks like `${param}` is such a dynamic parameter.
+
+You could change this to say `${name} is going to die!` instead.
+Or, maybe you don't care who it's on, and you can edit the text to `Buster` to be brief.
+If you want to undo your overriding, just clear the text.
+
+There are some limitations to this overriding.
+You cannot change the logic.
+You cannot make `tts` to say something different than the `alarmText` in most cases.
+You cannot add additional parameters.
+If you want to do any of these more complicated overrides,
+then you will want to look at the [Overriding Raidboss Triggers](#overriding-raidboss-triggers) section.
 
 ## User Directory Overview
 
@@ -180,8 +206,13 @@ If you are not a programmer, be extra careful with what and how you edit.
 ### Example 1: changing the output text
 
 Let's say hypothetically that you are doing UCOB
-and your group decides that they are going to do fire out first
-instead of fire in first like cactbot calls it by default.
+and your group decides that they are going to do "fire out" first
+instead of "fire in" first like cactbot calls it by default.
+Additionally, you *also* want to have the tts say something different for this trigger.
+You keep forgetting to get out, so you want it to repeat a few times.
+
+If you only wanted to change the `infoText`,
+you could do this via [Changing Trigger Text with the cactbot UI](#changing-trigger-text-with-the-cactbot-ui).
 
 One way to adjust this is to edit the trigger output for this trigger.
 You can find the original fireball #1 trigger in
@@ -207,6 +238,9 @@ Options.Triggers.push({
       infoText: {
         en: 'Fire OUT',
       },
+      tts: {
+        en: 'out out out out out',
+      },
       run: function(data) {
         data.naelFireballCount = 1;
       },
@@ -215,7 +249,7 @@ Options.Triggers.push({
 });
 ```
 
-This edit also replaced the `tts` section and removed other languages other than English.
+This edit also removed languages other than English.
 
 ### Example 2: making provoke work for all jobs
 
--- a/docs/RaidbossGuide.md
+++ b/docs/RaidbossGuide.md
@@ -202,9 +202,9 @@ May be a number or a `function(data, matches, output)` that returns a number. If
 Time to wait, in seconds, before showing this trigger again.
 May be a number or a `function(data, matches, output)`.
 The time to wait begins at the time of the initial regex match
- and is unaffected by presence or absence of a delaySeconds value.
- Once a trigger with this element activates,
- it will not activate again until after its timeout period is over.
+and is unaffected by presence or absence of a delaySeconds value.
+Once a trigger with this element activates,
+it will not activate again until after its timeout period is over.
 
 **sound**
 Sound file to play, or one of 'Info', 'Alert', 'Alarm', or 'Long'.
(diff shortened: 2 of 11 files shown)


quisquous:
TODO:
* support outputStrings in responses
* add outputString config UI (probably wait for #1908 to land first to avoid conflicts)
* add validation and testing
* add documentation
* in followup patches, convert all triggers to do this??? :eyes:

ui/raidboss/popup-text.js
```diff
@@ -600,7 +700,9 @@ class PopupText {
       trigger: trigger,
       now: now,
       valueOrFunction: (f) => {
-        let result = (typeof (f) == 'function') ? f(this.data, triggerHelper.matches) : f;
+        let result = f;
+        if (typeof (result) === 'function')
```
panicstevenson, line 704:
```suggestion
        if (typeof result === 'function')
```

ui/raidboss/raidboss_config.js
```diff
@@ -250,6 +251,27 @@ function setOptionsFromOutputValue(options, value) {
...
+class DoNothingFuncProxy {
+  constructor(outputStringsCallback) {
+    return new Proxy(this, {
+      set(target, property, value) {
+        if (property === 'responseOutputStrings') {
+          outputStringsCallback(value);
+          return true;
+        }
+
+        throw new Error('Invalid set of property ${property}');
```
panicstevenson, line 265: In both places, this is explicitly `throw`ing; nothing wrong with it, just that I don't frequently see it in cactbot code and want to make sure it's intentional.

ui/raidboss/raidboss_config.js
```diff
@@ -526,24 +583,33 @@ class RaidbossConfigurator {
...
-    for (let i = 0; i < keys.length; ++i) {
-      let key = keys[i];
-      if (!trig[key])
-        continue;
-      for (let d = 0; d < kFakeData.length; ++d) {
-        if (evalTrigger(trig, key, d))
-          break;
+    // Only evaluate fields if there are not outputStrings.
+    // outputStrings will indicate more clearly what the trigger says.
+    if (Object.keys(trig.outputStrings).length === 0) {
```
panicstevenson, line 605: Kind of an odd request, but can these checks:
```js
if (Object.keys(trig.outputStrings).length === 0)
```
Just be:
```js
if (!trig.outputStrings)
```
and get rid of this line above?
```js
trig.outputStrings = trig.outputStrings || {};
```
It feels odd to set yourself up in this way only to check something self-inflicted, unless there's something I'm missing

panicstevenson (commented):
👍

ui/raidboss/raidboss_config.js
quisquous: That's fair.  We could return '???' or even '?? ERROR ??' or something more obvious, so it appears on screen and also put something in the log.  Does that sound better?

ui/raidboss/raidboss_config.js
quisquous: No, I think I was maybe thinking about wanting it to always be defined for the `Object.assign` above it, but I can just handle it being undefined there instead.  Thanks!

ui/raidboss/popup-text.js
quisquous: Fixed!

quisquous:
I wrote some customization documentation too, but maybe it would be better not to land that until I had converted more triggers?

I was thinking that probably it'd be good to do the last raid tier and all ultimates, and that's probably ok to start.  I worry a good bit about breaking things during conversion.

(cc @JLGarber for any review thoughts here if you have time to look)

ui/raidboss/raidboss_config.js
panicstevenson: What happens if you have a borked trigger? I think it just doesn't play, right? If so, we could arguably be better, but I think logging + maybe something that's easy and doesn't need a translation, like you said with `?? ERROR ??`.

quisquous: Ah, ok ok.  These exceptions are only for setting values.  Getting a bogus value or param already returns `???`.  I still think it's probably better to be kind to user triggers and just print an error.

docs/CactbotCustomization.md
```diff
@@ -41,6 +43,30 @@ These options are stored in your
...
 You should not need to edit that file directly.
 
+## Changing Trigger Text with the cactbot UI
+
+In the cactbot configuration UI, under
+ACT -> Plugins -> OverlayPlugin.dll -> Cactbot -> Raidboss,
+there are individual trigger listings.
+You can use these listings to change various exposed configuration settings per trigger.
+
+Settings with a bell (🔔) next to their name are trigger mostly toutputs that you can override.
```
JLGarber, line 53: Is this a typo? Might be best to rephrase it if not.

docs/RaidbossGuide.md
```diff
@@ -155,28 +172,40 @@ Additionally, as with bare `regex` elements, current practice is to use regex re
...
-If present and a function which returns a promise, will wait for promise to resolve before continuing with trigger. This runs after `preRun` and before the `delaySeconds` delay.
+**promise: function(data, matches, output)**
+If present and a function which returns a promise,
+will wait for promise to resolve before continuing with trigger.
+This runs after `preRun` and before the `delaySeconds` delay.
 
 **delaySeconds**
-An amount of time, in seconds, to wait from the time the regex match is detected until the trigger activates. May be a number or a function(data, matches) that returns a number.
+An amount of time, in seconds, to wait from the time the regex match is detected until the trigger activates.
+May be a number or a `function(data, matches, output)` that returns a number.
```
JLGarber, line 192: Here and elsewhere, maybe it's worth noting for inexperienced users that `output` is largely meaningless for functions that just return a float or integer?

test/trigger/test_trigger.js
```diff
@@ -390,6 +405,136 @@ let testBadZoneId = function(file, contents) {
...
+        }
+      } else {
+        if (trigger.outputStrings && typeof outputStrings !== 'object') {
+          errorFunc(`${file}: '${trigger.id}' outputStrings must be an object.`);
+          continue;
+        }
+        outputStrings = trigger.outputStrings;
+      }
+
+      // TODO: should we prevent `output['phrase with spaces']()` style constructions?
```
JLGarber, line 449: Is there a specific scenario you have in mind where this could be an issue, or wou8ld this just be to keep things consistent with the rest of the repository? I have no strong feelings on it, just curious.

JLGarber:
Nothing much to add to this, it looks great. Definitely going to have to read through it a little more in-depth when I don't have food in the oven, but it does look like it will do a fine job of letting users do more UI config and less writing Javascript.

docs/CactbotCustomization.md
quisquous: Fixed, thanks.

docs/RaidbossGuide.md
quisquous: I added some notes to the `output` section above, thanks.  Let me know if you think it needs more.

test/trigger/test_trigger.js
quisquous: No, I think it was kind of a question.  I think the regex will match anything that's not a closing curly brace, but it was more a question of consistency.

docs/RaidbossGuide.md
```diff
@@ -155,28 +175,40 @@ Additionally, as with bare `regex` elements, current practice is to use regex re
...
+May be a number or a `function(data, matches, output)` that returns a number. If not specified, defaults to 3.
 
 **suppressSeconds**
-Time to wait, in seconds, before showing this trigger again.  May be a number or a function(data, matches).  The time to wait begins at the time of the initial regex match, and is unaffected by presence or absence of a delaySeconds value. Once a trigger with this element activates, it will not activate again until after its timeout period is over.
+Time to wait, in seconds, before showing this trigger again.
+May be a number or a `function(data, matches, output)`.
+The time to wait begins at the time of the initial regex match
+ and is unaffected by presence or absence of a delaySeconds value.
+ Once a trigger with this element activates,
+ it will not activate again until after its timeout period is over.
```
panicstevenson, line 207:
```suggestion
and is unaffected by presence or absence of a delaySeconds value.
Once a trigger with this element activates,
it will not activate again until after its timeout period is over.
```
Micro, overly picky, nit

test/trigger/test_trigger.js
```diff
@@ -390,6 +405,136 @@ let testBadZoneId = function(file, contents) {
...
+            errorFunc(`${file}: missing key '${key}' in '${trigger.id}' outputStrings`);
+            return;
+          }
+          keys.push(key);
+          return fullMatch;
+        });
+
+        for (const key of keys) {
+          for (const param of outputStringsParams[key]) {
+            if (!funcStr.match(`\\b${param}\\s*:`))
```
panicstevenson, line 529: `test`?

test/trigger/test_trigger.js
```diff
@@ -390,6 +405,136 @@ let testBadZoneId = function(file, contents) {
...
+            if (!ok) {
+              errorFunc(`${file}: '${key}' in '${trigger.id}' outputStrings has inconsistent params among languages`);
+              continue;
+            }
+          }
+          outputStringsParams[key] = [...params];
+
+          // Verify that there's no dangling ${
+          if (/\${/.test(template.replace(paramRegex, ''))) {
+            errorFunc(`${file}: '${key}' in '${trigger.id}' outputStrings has a dangling \${`);
```
panicstevenson, line 500: I understand this and I feel it's concise, but I don't know how well it translates both across languages and to people that may be unfamiliar? Something like: "open ${ without matching closing }"?
Feel free to disregard if you think it's obvious.

test/trigger/test_trigger.js
```diff
@@ -390,6 +405,136 @@ let testBadZoneId = function(file, contents) {
...
+
+          // If this is not the first lang, validate it has the same params as previous languages.
+          if (key in outputStringsParams) {
+            // prevParams is an array, params is a set.
+            const prevParams = outputStringsParams[key];
+            let ok = false;
+            if (prevParams.length === params.size) {
+              ok = true;
+              for (const key of prevParams)
+                ok &= params.has(key);
```
panicstevenson, line 489:
```suggestion
              ok = prevParams.every((key) => params.has(key));
```
Feel free to disregard

panicstevenson (approved):
Everything is minor/optional nits; I think this LGTM

docs/RaidbossGuide.md
quisquous: I'm gonna blame VSCode.  :innocent:

test/trigger/test_trigger.js
quisquous: Done.

test/trigger/test_trigger.js
quisquous: Nope, thanks! This is a good suggestion.

test/trigger/test_trigger.js
quisquous: Undone. There's no String.test method.  I could add `new Regexp(`stuff`).test(funcStr)` but why.

PR #1908, opened 2020-10-26
raidboss: add config ui to override duration

### LINKED ISSUE #1889: Add ability to modify trigger text duration from user trigger overrides.
Since https://github.com/quisquous/cactbot/issues/1869 is coming, one thing that got mentioned in discussion is users modifying the `${ENCOUNTER}.js` files themselves with each release and wanting users to move to per-trigger overrides.

One thing I've been doing is extending duration of certain triggers in UCoB for Nael for the quotes on moving in or out, because my brain immediately spaces sometimes and having an additional second or two of it on screen can jolt me into moving correctly.

Currently the documentation for per-trigger overrides doesn't seem to support that based on https://github.com/quisquous/cactbot/blob/main/docs/AdvancedCactbot.md#per-trigger-options.

quisquous:
Fixes #1889.

### DIFF
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -392,9 +392,7 @@ class RaidbossConfigurator {
           input.placeholder = this.base.translate(kMiscTranslations.valueDefault);
           input.value = this.base.getOption('raidboss', 'triggers', trig.id, optionKey, '');
           let setFunc = () => {
-            let val = validDurationOrUndefined(input.value);
-            if (!val)
-              val = '';
+            let val = validDurationOrUndefined(input.value) || '';
             this.base.setOption('raidboss', 'triggers', trig.id, optionKey, val);
           };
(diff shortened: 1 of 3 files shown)


ui/raidboss/raidboss_config.js
```diff
@@ -344,6 +371,37 @@ class RaidbossConfigurator {
...
+          let input = document.createElement('input');
+          div.appendChild(input);
+          input.type = 'text';
+          input.step = 'any';
+          input.placeholder = this.base.translate(kMiscTranslations.valueDefault);
+          input.value = this.base.getOption('raidboss', 'triggers', trig.id, optionKey, '');
+          let setFunc = () => {
+            let val = validDurationOrUndefined(input.value);
+            if (!val)
+              val = '';
```
kuwoyuki, line 397: can be flattened
```suggestion
            let val = validDurationOrUndefined(input.value) || '';
```

PR #1909, opened 2020-10-26
New translations for haukke

Akurosia:


quisquous:
Thank you for speedy translations, as always!

PR #1913, opened 2020-10-26
Brayflox Normal Triggers

Saulchar:
Brayflox Normal Triggers are ready


ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,120 @@
...
+          return {
+            en: 'Esuna Poison on ' + data.ShortName(matches.target),
+          };
+        }
+        if (matches.target == data.me && data.role == 'healer') {
+          return {
+            en: 'Esuna Your Poison',
+          };
+        }
+      },
```
JLGarber, line 30: Maybe rewrite this with a single-line condition like `condition: (data) => data.role === 'healer',`? (Also note the strict non-comparison here, not loose! Unless there's a specific reason for it, we're generally moving to use only strict comparison, and if there *is* a reason, consider rewriting to avoid it.)

```javascript
  condition: (data) => data.role === 'healer',
    if (matches.target !== data.me) {
      return {
        en: 'Esuna Poison on ' + data.ShortName(matches.target),
      };
    }
    return {
      en: 'Esuna Your Poison',
    };
  },

```

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -0,0 +1,120 @@
...
+        if (matches.target != data.me) {
+          return {
+            en: 'Break Bubble on ' + data.ShortName(matches.target),
+          };
+        }
+        if (matches.target == data.me) {
+          return {
+            en: 'Break Your Bubble',
+          };
+        }
```
JLGarber, line 78: As with the pelican poison trigger, consider eliding the final `if` and letting `return` take care of it. (And strict non-comparison for both regardless of whether you remove the `if`.)

PR #1920, opened 2020-10-30
Titan Normal

Saulchar:
Titan Normal Triggers, Timeline and updated manifest to support.


ui/raidboss/data/02-arr/trial/titan-nm.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+
+[{
+  zoneId: ZoneId.TheNavel,
+  timelineFile: 'titan-nm.txt',
+  triggers: [
+    {
+      id: 'TitanNm Tumult',
+      netRegex: NetRegexes.ability({ id: '282', source: 'Titan', capture: false }),
+      condition: (data) => data.role === 'healer' || data.CanAddle(),
```
JLGarber, line 10: As it happens, there's a built-in condition for this, [`caresAboutAOE()`](https://github.com/quisquous/cactbot/blob/a5d5de00566ce83529c52b172112fbfe45051b5b/resources/conditions.js#L13)!

ui/raidboss/data/02-arr/trial/titan-nm.js
```diff
@@ -0,0 +1,31 @@
...
+      netRegex: NetRegexes.ability({ id: '282', source: 'Titan', capture: false }),
+      condition: (data) => data.role === 'healer' || data.CanAddle(),
+      suppressSeconds: 2,
+      response: Responses.aoe(),
+    },
+    {
+      // Gaol callout for both yourself and others
+      id: 'TitanNm Gaols',
+      netRegex: NetRegexes.gainsEffect({ effectId: '124' }),
+      condition: (data) => data.role === 'healer',
```
JLGarber, line 18: Why are only healers being notified for Gaols?

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,107 @@
...
+473.6 "Enrage"
+
+###Phase 4 55->0%
+601.9 "--sync--" sync /:28C:Earthen Fury/ window 199,0 
+603.0 "Earthen Fury" sync /:Titan:28C:/
+
+616.0 "Landslide" sync /:Titan:28A:/
+622.2 "Rock Buster" sync /:Titan:281:/
+626.2 "Weight Of The Land" sync /:Titan:284:/ window 626,23
+632.3 "Tumult x2" duration 2.5 #sync /:Titan:282:/
```
JLGarber, line 77: This is indeed the current way of handling close-in-time abilities, nicely done implementing it correctly first try.

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,107 @@
...
+hideall "--Reset--"
+hideall "--sync--"
+
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+### Phase 1 100->85%
+0.0 "Start"
+0.0 "--sync--" sync /Engage!/ window 0,1
+2.0 "--sync--" sync /:Titan:368:/ window 2,0
+5.0 "Rock Buster" sync /:Titan:281:/
```
JLGarber, line 13: Is there a particular reason you didn't set up a [`tankCleave()`](https://github.com/quisquous/cactbot/blob/a5d5de00566ce83529c52b172112fbfe45051b5b/resources/responses.js#L197) timeline trigger based on these log lines?

JLGarber (commented):
Very nice! That's a really clean timeline for a first attempt. It could honestly go in as it is. One thing you might not be aware of, if you include the `-it` (include targetable) flag when generating the timeline, it will pre-populate `--targetable--/--untargetable--` lines for you without having to manually time them out. (Note that `-it` takes arguments, one per enemy you wish to track. You can't just use the bare flag!)

This is a really good submission overall. Much better than mine were at this point. I'm extremely impressed.

ui/raidboss/data/02-arr/trial/titan-nm.js
Saulchar: I actually didnt notice that, I will replace it. I copied it from another shorthand condition and forget to change it. I'll need to change it to the way I have the Esuna trigger in Brayflox. if then statement.

ui/raidboss/data/02-arr/trial/titan-nm.js
Saulchar: I didnt actually know about that, I will fix that. Thanks!

ui/raidboss/data/02-arr/trial/titan-nm.txt
Saulchar: Well, I was referencing the hardmode timeline for how to sync the phases properly, and saw "Tumult x4" with a duration, so I adapted it for normal

ui/raidboss/data/02-arr/trial/titan-nm.txt
Saulchar: Hmm you are correct, they are far enough apart that would be possible.

ui/raidboss/data/02-arr/trial/titan-nm.js
```diff
@@ -0,0 +1,40 @@
...
+        };
+      },
+    },
+    {
+      // Covers both tanks and non tanks with appropriate text
+      // Needs suppression for occasions in phase 4 where rock busters
+      // get pushed close together by auto attacks to avoid double triggers.
+      id: 'TitanNm Rock Buster',
+      netRegex: NetRegexes.ability({ id: '281', source: 'Titan' }),
+      condition: (data) => data.role !== 'tank' || data.role === 'tank',
```
quisquous, line 35: I think this is a tautology, and maybe you can remove the entire `condition` line?

ui/raidboss/data/manifest.txt
```diff
@@ -1,266 +1,268 @@
...
-05-shb/trial/varis-ex.txt
-05-shb/trial/wol.js
-05-shb/trial/wol.txt
-05-shb/trial/wol-ex.js
-05-shb/trial/wol-ex.txt
-05-shb/ultimate/the_epic_of_alexander.js
-05-shb/ultimate/the_epic_of_alexander.txt
+00-misc/general.js
+00-misc/test.js
+00-misc/test.txt
```
quisquous, line 3: Hmm.  What's going on with this file? Did you change the line endings?

ui/raidboss/data/02-arr/trial/titan-nm.js
```diff
@@ -0,0 +1,40 @@
...
+          en: 'Gaol on YOU',
+        };
+      },
+    },
+    {
+      // Covers both tanks and non tanks with appropriate text
+      // Needs suppression for occasions in phase 4 where rock busters
+      // get pushed close together by auto attacks to avoid double triggers.
+      id: 'TitanNm Rock Buster',
+      netRegex: NetRegexes.ability({ id: '281', source: 'Titan' }),
```
quisquous, line 34: I think this needs to be a timeline trigger.  (Maybe this is what @JLGarber is saying below as well?)  This ability has no castbar and so you can't do a `startsUsing`.  If you use `ability` here, this will go off after the ability has already happened.  So, if you want to do it ahead of time, you need to use a timeline trigger.  (Maybe this is what you were asking about on discord a while back?)

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,105 @@
...
```
quisquous, line 13: Can you also add a `window 5,5` to this line? I usually like to sync the first non-auto as well, since it's not guaranteed that the boss always gets an auto off.

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,105 @@
...
+264.1 "Landslide" sync /:Titan:28A:/
+270.2 "Rock Buster" sync /:Titan:281:/
+274.2 "Tumult" sync /:Titan:282:/ window 17,16 jump 239.3
+
+281.5 "Landslide"
+287.6 "Rock Buster"
+291.6 "Tumult"
+298.9 "Landslide"
+305.1 "Rock Buster"
+309.1 "Tumult"
```
quisquous, line 54: Sorry, would you mind lining up this loop?

```
281.5 "Landslide" #sync /:Titan:28A:/
287.7 "Rock Buster" #sync /:Titan:281:/
291.7 "Tumult" #sync /:Titan:282:/
299.0 "Landslide" #sync /:Titan:28A:/
305.1 "Rock Buster" #sync /:Titan:281:/
309.1 "Tumult" #sync /:Titan:282:/
```

When a timeline loops, it should have the same times as where it jumps to, just adjusted forward in time.  I should really add some sort of `loop` command to do this for you, but for now there's always `util/timeline_adjust.py`.

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,105 @@
...
+435.1 "Rock Throw" sync /:Titan:285:/
+442.9 "Landslide" sync /:Titan:28A:/
+449.1 "Rock Buster" sync /:Titan:281:/
+453.1 "Tumult" sync /:Titan:282:/ window 16,25
+464.6 "Landslide" sync /:Titan:28A:/
+473.6 "Enrage"
+
+###Phase 4 55->0%
+601.9 "--sync--" sync /:28C:Earthen Fury/ window 199,0 
+603.0 "Earthen Fury" sync /:Titan:28C:/
```
quisquous, line 72: This has a 2.1 second cast time?

Is there a reason not to pick a whole number here, like `600.0` instead of `601.9`?

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,105 @@
...
+734.2 "Landslide"
+740.4 "Rock Buster"
+744.4 "Weight Of The Land"
+750.5 "Tumult x2" duration 2.5
+756.7 "Rock Throw"
+763.8 "Landslide"
+770.0 "Rock Buster"
+774.0 "Weight Of The Land"
+780.1 "Tumult x2" duration 2.5
+786.3 "Rock Throw"
```
quisquous, line 105: Similarly to the above loop comment, this should be:
```
734.2 "Landslide" #sync /:Titan:28A:/
740.4 "Rock Buster" #sync /:Titan:281:/
744.4 "Weight Of The Land" #sync /:Titan:284:/ window 26,24
750.5 "Tumult x2" duration 2.5 #sync /:Titan:282:/
756.7 "Rock Throw" #sync /:Titan:285:/
763.8 "Landslide" #sync /Titan:285:/
769.9 "Rock Buster" #sync /:Titan:281:/
773.9 "Weight Of The Land" #sync /:Titan:284:/
780.0 "Tumult x2" duration 2.5 #sync /:Titan:282:/ window 28,23
786.2 "Rock Throw" #sync /:Titan:285:/
```

ui/raidboss/data/02-arr/trial/titan-nm.txt
```diff
@@ -0,0 +1,105 @@
...
+427.6 "Rock Buster" sync /:Titan:281:/ window 20,15
+431.6 "Tumult" sync /:Titan:282:/
+435.1 "Rock Throw" sync /:Titan:285:/
+442.9 "Landslide" sync /:Titan:28A:/
+449.1 "Rock Buster" sync /:Titan:281:/
+453.1 "Tumult" sync /:Titan:282:/ window 16,25
+464.6 "Landslide" sync /:Titan:28A:/
+473.6 "Enrage"
+
+###Phase 4 55->0%
```
quisquous, line 70: This is the tiniest of nits, but can you be consistent in your headings?

Other files also do `### Phase 4: 55% -> 0%` if you want to follow that formatting pattern as well.

quisquous (commented):
Thanks! Timeline looks really good overall, with just a few small nits.

ui/raidboss/data/manifest.txt
Saulchar: I accidentally uploaded my local copy as a new file instead of just editing the existing file

Saulchar: I accidentally uploaded my local copy as a new file instead of just editing the existing file

ui/raidboss/data/02-arr/trial/titan-nm.txt
Saulchar: Titan lands and starts casting Earthen Fury for 1.1s then it goes off. This was the only way I had something to link to that wasn't an RP line.

ui/raidboss/data/02-arr/trial/titan-nm.txt
Saulchar: These were originally that way, I just deleted the syncs because that is my "fake loop" or are we not using fake loops anymore? This was coming out of your timeline guide

ui/raidboss/data/02-arr/trial/titan-nm.js
Saulchar: Will removing the condition allow it to capture the roles still and return the correct tank cleave alert? I wrote it that way just to be sure it captures player roles correctly to assign them the correct alert based on their role.

Saulchar: This way it applies Tank Cleave on YOU to the tank, and Avoid Tank Cleave to everyone else, and it does it in 1 trigger instead of two

ui/raidboss/data/02-arr/trial/titan-nm.js
Saulchar: Would I just move all instances of Rock Buster a couple seconds back in the timeline, so it goes off early enough, or how would I do this properly?

ui/raidboss/data/02-arr/trial/titan-nm.txt
Saulchar: Sure can

ui/raidboss/data/02-arr/trial/titan-nm.js
quisquous: You can use `beforeSeconds` with a timeline trigger.  There's examples in both of the other Titan files.

ui/raidboss/data/02-arr/trial/titan-nm.txt
quisquous: Ah sorry, apparently I can't math.  1.1s just seemed unusually fast to me.  It's fine to sync to that, was mostly just curious if it were truly 1.1s and also why you picked 601.9 or 603 as a time here.

Would you also mind syncing a one-time ability phase like this back to the beginning with a `window 610,0` or something? I usually try to put something a couple of those in a fight to make sure things are on track, just in case.

ui/raidboss/data/02-arr/trial/titan-nm.txt
quisquous: No, fake loops are good.  My comment is 90% about using the right times.  I can go either way on deleting the syncs.  Commenting them out or deleting them is fine.

ui/raidboss/data/02-arr/trial/titan-nm.js
JLGarber: `TankCleave() already handles that logic internally. Because this will be a timeline trigger, we won't have access to the `target`, so the `response` will just look at roles without needing explicit input from the trigger.

ui/raidboss/data/02-arr/trial/titan-nm.js
JLGarber: Quisquous is correct, this is a timeline trigger, sorry I didn't make that clearer! What you want to do is set it up something like this:

```javascript
timelineTriggers: [
    {
      id: 'TitanNm Rock Buster',
      regex: /Rock Buster/,
      beforeSeconds: 7,
      condition: function(data) {
        return data.role == 'healer' || data.role == 'tank';
      },
    },
],
```

(Shamelessly stolen from [Titan EX](https://github.com/quisquous/cactbot/blob/60e3b23d44ff538f8ab248845b502365f38765de/ui/raidboss/data/02-arr/trial/titan-ex.js#L6).)

As you can see, we use a normal `Regex` property, because we're not looking for a log line, we're looking for a timeline match. The `beforeSeconds` property tells the timeline playback that "okay, you have the timeline file loaded, when the timer reaches 7 seconds before any line that contains Rock Buster, activate the trigger".

ui/raidboss/data/02-arr/trial/titan-nm.js
Saulchar: Oh cool, ty!

ui/raidboss/data/02-arr/trial/titan-nm.txt
JLGarber: To expand on that a little, once you have a clean loop established, what you want to do is take the timestamp from the ability at the start of the loop and the timestamp at the next iteration's start. Subtract to find the "length" of the loop. Then run `timeline_adjust.py` with the `-a` flag set to that length value.

The reason for this is that very few encounters have perfect loops, in the sense that there's no timing variance between loop iterations. Bosses moving to do autos will introduce a little drift. This makes the timeline a bit "shaky" when jumping, since the user will see the fake loop abilities suddenly snap to the real loop abilities.

ui/raidboss/data/02-arr/trial/titan-nm.txt
JLGarber: I think what quisquous is asking is "why does this phase start at 601.9 instead of 600?"  It's mostly a stylistic question, but unless there's a very good reason for it, you are probably better off adjusting this loop back by 1.9 seconds to match the rest of the repository. (Or you might adjust it back 3.8 seconds, so that the Earthen Fury cast lines up at 600.)

ui/raidboss/data/02-arr/trial/titan-nm.txt
JLGarber: (Going to personally push for deleting them, since that's what we do in the rest of the repository. It also makes it visually clearer to the user that "hey, this is a fake lookahead section".)

JLGarber:
One other recommendation I would make here, and this is a Github thing not a cactbot thing, once you've finished up with any specific review comment, if you could mark it "resolved", that would make it a lot easier for others when reading the pull request. Not necessary, but it's very helpful for everyone. ^_^

ui/raidboss/data/02-arr/trial/titan-nm.js
```diff
@@ -0,0 +1,39 @@
...
+
+[{
+  zoneId: ZoneId.TheNavel,
+  timelineFile: 'titan-nm.txt',
+  timelineTriggers: [
+    {
+      // Early Callout for Tank Cleave
+      id: 'TitanNm Rock Buster',
+      regex: /Rock Buster/,
+      beforeSeconds: 2,
```
quisquous, line 11:
```suggestion
      beforeSeconds: 5,
```

I think 2 seconds is not much time to move out of the way or cast a heal.  The other titans say 7 seconds.  I think 5 or 7 would be fine here, but mostly arguing for something larger than 2.

ui/raidboss/data/02-arr/trial/titan-nm.txt
quisquous: Sorry, could you only resolve comments that you have pushed changes for?

PR #1926, opened 2020-11-01
Convert all of responses to use output strings

quisquous:


resources/responses.js
```diff
@@ -167,788 +178,686 @@ let Responses = {
...
-
-      };
+      },
+    };
+    return {
+      [defaultAlertText(sev)]: (data, matches, output) => {
+        const target = getTarget(matches);
+        if (target !== data.me)
+          return;
+        return output.earthshaker({ name: source });
```
panicstevenson, line 858: `source` or `target`?

panicstevenson (approved):

resources/responses.js
quisquous: Neither! Removed, thanks.

PR #1928, opened 2020-11-02
made "wake up" alarm into a response

Akurosia:


quisquous:
Thanks for fixing my TODOs.  :100:

PR #1931, opened 2020-11-03
added only-combatant option to make timeline

Akurosia:
Making timelines for Critical engangements can be confusing with action from random mobs.
With this option you can specify the enemy you want to make a timeline around (mostly usefull in eureka and bozja)

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -429,14 +429,14 @@ if __name__ == "__main__":
         "--ignore-combatant",
         nargs="*",
         default=[],
-        help="Combatant names to ignore, e.g. \"Aratama Soul\"",
+        help='Combatant names to ignore, e.g. "Aratama Soul"',
     )
     parser.add_argument(
         "-oc",
         "--only-combatant",
         nargs="*",
         default=[],
-        help="Only the listed combatants will generate timeline data, e.g. \"Aratama Soul\"",
+        help='Only the listed combatants will generate timeline data, e.g. "Aratama Soul"',
     )
     parser.add_argument(
         "-p",
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -436,7 +436,7 @@ if __name__ == "__main__":
         "--only-combatant",
         nargs="*",
         default=[],
-        help="Only the listed combatants will generate timeline data, e.g. \"Aratama Soul\"",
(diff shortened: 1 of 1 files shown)


quisquous:
Is it too much of a pain to use `-ic` here to ignore combatants? Usually there's only a mob or two around each CE, right?

Akurosia:
i tried to do it with some CEs and on some i could bearly see lines from my boss itself which was kinda anoying.
e.g. starting area sometimes has 6-7 different mobs around and its easier to write 1 name instead of like 7

util/make_timeline.py
```diff
@@ -427,6 +431,13 @@ def main(args):
         default=[],
         help="Combatant names to ignore, e.g. Aratama Soul",
     )
+    parser.add_argument(
+        "-oc",
+        "--only-combatant",
+        nargs="*",
+        default=[],
+        help="Combatant names to only pay attantion to, e.g. Aratama Soul",
```
quisquous, line 439:
```suggestion
        help="Combatant names to only pay attention to, e.g. \"Aratama Soul\"",
```

util/make_timeline.py
```diff
@@ -289,6 +289,10 @@ def main(args):
         ):
             continue
 
+        # if only combatants was specified and combatant not in the list
+        if args.only_combatant != [] and entry["combatant"] not in args.only_combatant:
```
quisquous, line 293:
```suggestion
        if args.only_combatant and entry["combatant"] not in args.only_combatant:
```

util/make_timeline.py
JLGarber: Perhaps instead
```suggestion
        help="Only the listed combatants will generate timeline data, e.g. \"Aratama Soul\"",
```

util/make_timeline.py
```diff
@@ -427,6 +431,13 @@ def main(args):
         default=[],
         help="Combatant names to ignore, e.g. Aratama Soul",
     )
+    parser.add_argument(
+        "-oc",
+        "--only-combatant",
+        nargs="*",
+        default=[],
+        help="Only the listed combatants will generate timeline data, e.g. Aratama Soul",
```
quisquous, line 439: Sorry, I think quotes are better here.  (And maybe they're needed elsewhere in the file).  You can't just pass `-oc Aratama Soul` as-is.

PR #1935, opened 2020-11-04
bozja updates

Akurosia:
For translations i tried to stay with a maximum of 20 chars max.
Its hard for some, as many dont have a common abbreviation


quisquous:
Thank you!!

PR #1942, opened 2020-11-05
raidboss: remove unneeded tts entries

quisquous:
tts is automatic.  I think in most cases it's preferable to keep the tts
output the same as the on screen output unless it's dramatically longer
or it's something where it makes sense to have something very short in
tts because the mechanic is urgent.

This leaves unchanged o2 / o3 / o4 which are extremely verbose in their
triggers, which seemed a bit obnoxious to play over tts.


panicstevenson (approved):
LGTM, with maybe the exception of Hyperdrive/Embrace for Kefka. Thoughts?

quisquous:
> LGTM, with maybe the exception of Hyperdrive/Embrace for Kefka. Thoughts?

I'm curious why that is an exception for you.  In both cases for me, I think the important word still gets out first, but then it gets more information too.

JLGarber:
Maybe just normalize Hyperdrive to `tankCleave()`, or `tankBuster()` on Normal? That would sidestep things altogether.

quisquous:
I'm going to be opinionated and say I think the embrace ones are fine as-is, and I'll just leave them.  I converted hyperdrive to be a response.

PR #1946, opened 2020-11-06
Brayflox update

Saulchar:
Update brayflox with suppression for poison on 1st boss


ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.js
```diff
@@ -22,6 +22,7 @@
       id: 'Brayflox Normal Pelican Poison Healer',
       netRegex: NetRegexes.gainsEffect({ effectId: '12' }),
       condition: (data) => data.role === 'healer',
+      suppressSeconds: 2,
```
panicstevenson, line 25: I think this `suppressSeconds` works opposite of how the comment above explains how this trigger should work.
By suppressing here, won't you only see this trigger once for Player A, even if Players A, B, and C all got hit?

For the same effect, you'd need one trigger to collect the people who got hit without a message, and another trigger (with suppressSeconds) to display the message for everyone that got hit.

I would expect something similar to:
```js
{
  id: 'Brayflox Normal Pelican Poison Collect',
  netRegex: NetRegexes.gainsEffect({ effectId: '12' }),
  condition: (data) => data.role === 'healer',
  run: (data, matches) => {
    data.pelicanPoisons = data.pelicanPoisons || [];
    data.pelicanPoisons.push(matches.target);
  }
},
{
  id: 'Brayflox Normal Pelican Poison Healer',
  netRegex: NetRegexes.gainsEffect({ effectId: '12' }),
  condition: (data) => data.role === 'healer',
  delaySeconds: 1,
  suppressSeconds: 2,
  alertText: (data, matches) => {
    if (!data.pelicanPoisons)
      return;

    const names = data.pelicanPoisons.sort();
    data.pelicanPoisons = [];
    if (names.length === 1 && names[0] === data.me) {
      return {
        en: 'Esuna Your Poison',
        de: 'Entferne dein Gift',
        fr: 'Purifiez-vous',
        cn: '康复自己的毒',
      };
    }
    return {
      en: 'Esuna Poison on ' + names.map((x) => data.ShortName(x)).join(', '),
      de: 'Entferne Gift von ' + names.map((x) => data.ShortName(x)).join(', '),
      fr: 'Purifiez le poison sur ' + names.map((x) => data.ShortName(x)).join(', '),
      cn: '康复' + names.map((x) => data.ShortName(x)).join(', '), + '的毒',
    };
  },
},
```

Saulchar: You are correct, I needed to update this because during my daily I had the trigger go off 3 times in quick succession on the tank. Thank you for the input.

JLGarber:
One other thing on the poison trigger (and anywhere you use the Collect --> Analyze --> Call construction) you almost always want to include a Cleanup trigger as well. Something like

```javascript
    {
      id: 'Brayflox Normal Pelican Poison Cleanup',
      netRegex: NetRegexes.gainsEffect({ effectId: '12', capture: false }),
      delaySeconds: 5,
      suppressSeconds: 5,
      run: (data) => {
        delete data.pelicanPoisons;
      },
    },
```

You can see a complete example from [E8n](https://github.com/quisquous/cactbot/blob/33f8aaebf56a47ee9491cfd22300d258bcb7f2c1/ui/raidboss/data/05-shb/raid/e8n.js#L237). (The trigger that's linked there, and then the next two as well.)

panicstevenson:
I had included that in the trigger itself; if you decided to add what @JLGarber said, you'll want to remove it from the alertText function

JLGarber:
Ah, so it was. Sorry for not reading better! That's probably better just on the basis of saving space, even if it is slightly easier to miss on a casual reading.

panicstevenson:
Yea, I am generally for having a separate cleanup function as well; the only reason I personally left it out is that this is a healer only check, not something quite as big as like divebombs or something. That being said, I support either way.

quisquous:
Looks good to me, thanks!

PR #1949, opened 2020-11-07
Raidboss: Hell's Lid timelines and triggers

JLGarber:
It's so straightforward that I don't really have any comments on it. Yay for output strings!


quisquous:
> It's so straightforward that I don't really have any comments on it. Yay for output strings!

It's so straightforward that I don't have any comments either!

PR #1953, opened 2020-11-07
raidboss: Korean translate

Jaehyuk-Lee:
I think adding `sec` to duration is better
https://github.com/quisquous/cactbot/blob/6809fe44899ee01b7bf98e7280b3203b411a7f4a/ui/raidboss/raidboss_config.js#L123-L127
`duration` -> `duration (sec)`

### DIFF
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -125,6 +125,7 @@ let kDetailKeys = {
       fr: 'Durée',
       ja: '存続時間',
       cn: '持续时间',
+      ko: '지속 시간 (초)',
     },
     cls: 'duration-text',
     generatedManually: true,
@@ -216,6 +217,7 @@ const kMiscTranslations = {
     en: '(default)',
     de: '(Standard)',
     fr: '(Défaut)',
+    ko: '(기본값)',
   },
 };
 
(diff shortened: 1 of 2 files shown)


ui/raidboss/raidboss_config.js
```diff
@@ -125,6 +125,7 @@ let kDetailKeys = {
       fr: 'Durée',
       ja: '存続時間',
       cn: '持续时间',
```
MaikoTan, line 127: Great! And you can also add unit to other languages too.

```javascript
      en: 'duration (in seconds)',
      de: 'Dauer (in Sekunden)',
      fr: 'Durée (en secondes)',
      ja: '存続時間 (秒)',
      cn: '持续时间 (秒)',
```

Jaehyuk-Lee: do we need `in seconds` long? I think `sec` is enough and I like short thing.

MaikoTan: Well, I am not very familiar with grammar of English, I just copy them from the "hide cheese icon option".
Do you think it is normal to use `duration (sec)` in English?

Jaehyuk-Lee: it's very normal to use `sec` `min` `hr` in English. I don't know French or German though.

MaikoTan: Maybe just `Sekunden` and `secondes` is fine.
Note: noun in German should be Capitalized always.

quisquous: Thanks! Having the time unit in there makes sense.

PR #1954, opened 2020-11-07
raidboss: wakeup 7min

Jaehyuk-Lee:
I think alarming 7 minute is okay and slightly better.
3 minute until 10 minute AFK is okay


quisquous:
Sounds good!

PR #1956, opened 2020-11-07
raidboss: Korean translate t1-t13

Jaehyuk-Lee:
What does this small `s` means? I couldn't translate it, does it mean it "could" be multiple or single?
https://github.com/quisquous/cactbot/blob/ba2cd3afb7f4276873b9712ead29840d02fbc950/ui/raidboss/data/02-arr/raid/t13.txt#L150


JLGarber:
That is correct, it could be one or it could be more.

PR #1961, opened 2020-11-08
raidboss: convert all remaining arr triggers to output strings

quisquous:


ui/raidboss/data/02-arr/raid/t9.js
```diff
@@ -338,18 +385,12 @@
...
-          ja: 'マーカー: ${naelMarks}',
-          cn: '标记： ${naelMarks}',
-          ko: '카탈징: ${naelMarks}',
-        },
+        return output.marks({
+          dir1: output[data.naelMarks[0]](),
+          dir2: output[data.naelMarks[1]](),
+        });
       },
+      outputStrings: diveOutputStrings,
```
quisquous, line 393: One thing to ask here is what you think of this pattern of reusing an output strings pre-built object across multiple triggers.  In some ways, this makes life simpler, in that there is no way this can be messed up.  However, it also includes strings that may not appear (e.g. marks for the safe zone trigger, and safe zone for the marks trigger), and so might make the config ui more confusing.

I am a little bit torn, but I kind of wonder if "can't get this wrong programming-wise" is the better answer, because I'm not sure I could trust myself to write a test here that would validate this, especially when output string keys in the most complicated cases are dynamic.

panicstevenson: Potentially confusing, but you could do:
```js
const { marks, safeZone, ...directions } = diveOutputStrings;
const marksOutput= Object.assign({}, directions, { marks });
const safeZoneOutput = Object.assign({}, directions, { safeZone });
```
In the sense that you'd be reusing the code, but then extra entries wouldn't appear in the UI?

quisquous: Yeah, I could.  Do you think that's an improvement over the original?

panicstevenson: I guess that depends on how confusing the UI is; I think with a one-liner comment you could explain what you're doing/why and it's not really sacrifice much readability. How do you feel about it?

quisquous: I think mostly I worry about things like wolex which share quite a few things.  I'll give it a go, and you can tell me what you think when we get there.  I'll change this to only have the valid values.

panicstevenson: I guess alternatively, you could put all of the directions into the const at the top, and just reuse those alongside the new object (marks, safeZone) inside the outputStrings within the trigger?
```js
const diveDirectionOutputStrings = {
  unknown: {
    en: '?',
  },
  north: {
    en: 'N',
  },
  ...
  northwest: {
    en: 'NW',
  },
};
```

```js
outputStrings: {
  ...diveDirectionOutputStrings,
  marks: {
    en: 'Marks: ${dir1}, ${dir2}',
    de: 'Markierungen : ${dir1}, ${dir2}',
    fr: 'Marque : ${dir1}, ${dir2}',
    ja: 'マーカー: ${dir1}, ${dir2}',
    cn: '标记： ${dir1}, ${dir2}',
    ko: '카탈징: ${dir1}, ${dir2}',
  },
}
```
Would that work?

quisquous: Oh yeah, that's way better.

panicstevenson (approved):

PR #1984, opened 2020-11-08
oopsyraidsy: Apply ESLint eqeqeq

panicstevenson:


ui/oopsyraidsy/data/03-hw/alliance/weeping_city.js
```diff
@@ -103,7 +103,7 @@
       // This ability name is helpfully called "Attack" so name it something else.
       id: 'Weeping Ozma Tank Laser',
       netRegex: NetRegexes.ability({ id: '1831' }),
-      condition: (e) => e.type != 15,
+      condition: (e) => e.type !== 15,
```
quisquous, line 106: e.type is a string

ui/oopsyraidsy/data/05-shb/dungeon/heroes_gauntlet.js
```diff
@@ -53,7 +53,7 @@
       // This should always be shared.  On all times but the 2nd and 3rd, it's a party share.
       // TODO: on the 2nd and 3rd time this should only be shared with a rock.
       // TODO: alternatively warn on taking one of these with a 472 Magic Vulnerability Up effect
-      condition: (e) => e.type == 15,
+      condition: (e) => e.type === 15,
```
quisquous, line 56: Here too!

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -917,7 +917,7 @@ class DamageTracker {
   OnTrigger(trigger, evt, matches) {
     // If using named groups, treat matches.groups as matches
     // so triggers can do things like matches.target.
-    if ((matches != undefined) && (matches.groups != undefined))
+    if ((matches !== undefined) && (matches.groups !== undefined))
```
quisquous, line 920: I looked through callsites, and I think matches is always defined, so I think this is fine?  (And one callsite already passes matches.groups through instead of just matches, which I guess works out in the end, but is a bit asymmetrical.)

PR #1989, opened 2020-11-09
resources: Apply ESLint eqeqeq

panicstevenson:


resources/widgetlist.js
```diff
@@ -114,12 +114,13 @@ class WidgetList extends HTMLElement {
...
-    if (this.scale != null) this._scale = Math.max(parseFloat(this.scale), 0.01);
-    if (this.elementwidth != null) this._elementwidth = Math.max(parseInt(this.elementwidth), 1);
-    if (this.elementheight != null) this._elementheight = Math.max(parseInt(this.elementheight), 1);
-    if (this.rowcolsize != null) this._rowcolsize = Math.max(parseInt(this.rowcolsize), 1);
-    if (this.toward != null) this.parseToward(this.toward);
-    if (this.maxnumber != null) this._maxnumber = Math.max(parseInt(this.maxnumber), 1);
+    if (this.scale !== null) this._scale = Math.max(parseFloat(this.scale), 0.01);
+    if (this.elementwidth !== null) this._elementwidth = Math.max(parseInt(this.elementwidth), 1);
+    if (this.elementheight !== null)
+      this._elementheight = Math.max(parseInt(this.elementheight), 1);
```
panicstevenson, line 120: This exceeded the maximum line length 😭

MaikoTan: Is it reasonable to convert all `if-null-checking` to bool operators?

for example
```javascript
this._elementheight = Math.max(parseInt(this.elementheight || "1"), 1);
```

panicstevenson: Since there's default values above, we could also do plain ol' ternaries:
```js
this._elementheight = this.elementheight ? Math.max(parseInt(this.elementheight), 1) : 100;
```
Which doesn't add a lot, but it does bring the line length down slightly.

quisquous: Using a boolean or a ternary is not possible, because it conflates falsy values like `0` or `false` with `null`.  In this case, setting to zero is not the same as leaving it null (and using the default value).   A util function is fine, although I think that could be done as a followup.

resources/timericon.js
```diff
@@ -188,18 +188,18 @@ class TimerIcon extends HTMLElement {
...
-    if (this.scale != null) this._scale = Math.max(parseFloat(this.scale), 0.01);
-    if (this.hideafter != null && this.hideafter != '') this._hideAfter = Math.max(parseFloat(this.hideafter), 0);
+    if (this.duration !== null) this._duration = Math.max(parseFloat(this.duration), 0);
+    if (this.width !== null) this._width = Math.max(parseInt(this.width), 1);
+    if (this.height !== null) this._height = Math.max(parseInt(this.height), 1);
+    if (this.bordercolor !== null) this._borderFg = this.bordercolor;
+    if (this.bordersize !== null) this._colorBorderSize = Math.max(parseInt(this.bordersize), 0);
+    if (this.scale !== null) this._scale = Math.max(parseFloat(this.scale), 0.01);
+    if (this.hideafter !== null && this.hideafter !== '') this._hideAfter = Math.max(parseFloat(this.hideafter), 0);
     if (typeof this.onhide !== 'undefined') this._onHide = this.onhide;
```
quisquous, line 198: This isn't your code, but it is a === issue, in that these all should be null, and `!== undefined` is incorrect.

resources/widgetlist.js
kuwoyuki:
```suggestion
    this._elementheight = Math.max(parseInt(this.elementheight ?? 100), 1);
```

quisquous: Unfortunately, OverlayPlugin is not using a recent enough version of Chrome to use nullish coalescing operators.

panicstevenson: Babel after Webpack lands? ;)

PR #1998, opened 2020-11-10
raidboss: Fixup O2N Gravitational Manipulation Float Condition

panicstevenson:


ui/raidboss/data/04-sb/raid/o2n.js
```diff
@@ -59,7 +59,7 @@
       id: 'O2N Gravitational Manipulation Float',
       netRegex: NetRegexes.headMarker({ id: '0071' }),
       condition: function(data, matches) {
-        return !data.levitating && Conditions.targetIsNotYou();
+        return !data.levitating && Conditions.targetIsNotYou()(data, matches);
```
quisquous, line 62: I wish there were a better way to do this, but this is more correct than what was there before.

PR #2005, opened 2020-11-12
docs: Korean Readme

Jaehyuk-Lee:
Overlay module detailed part is yet in English, but I think worth to add this file now.
See markdown [here](https://github.com/Jaehyuk-Lee/cactbot/blob/docs-ko-readme/docs/ko-KR/README.md)


quisquous:
Thanks for this!

PR #2007, opened 2020-11-12
jobs: add option for countdown sound

Jaehyuk-Lee:
I've added countdown alarm sound enable/disable option, or better to add sound volume?

### DIFF
--- a/ui/jobs/jobs_config.js
+++ b/ui/jobs/jobs_config.js
@@ -44,7 +44,7 @@ UserConfig.registerOptions('jobs', {
     {
       id: 'PlayCountdownSound',
       name: {
-        en: 'Enable countdown alarm sound',
+        en: 'Enable countdown notification sound',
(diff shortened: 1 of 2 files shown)


ui/jobs/jobs_config.js
```diff
@@ -41,6 +41,15 @@ UserConfig.registerOptions('jobs', {
       type: 'float',
       default: 0.5,
     },
+    {
+      id: 'PlayCountdownSound',
+      name: {
+        en: 'Enable countdown alarm sound',
```
MaikoTan, line 47: I think use "Alert" instead of "Alarm" would be more consistent ?

JLGarber: That, or "notification". (I say this because `raidboss` already has meanings for "info", "alert", and "alarm", and it might be better to avoid any potential confusion there.)

Issue #1944, opened 2020-11-05
Jobs: Rearrange code by job order


Echoring:
With more and more UI added to Jobs module, seems it's a little... chaos.
I'm considering about rearrange these code by an order, like PLD WAR DRK GNB WHM SCH AST MNK DRG NIN SAM BRD MCH DNC BLM SMN RDM BLU Other(Order on job list).
I know that it's not important, but it makes locating something a little easier(maybe a little obsessive).
This may make the whole file up-side-down, and very hard to review.
May I going to do this?

MaikoTan:
Cactbot is going to migrate to use webpack, I think we can seperate every UI component for individual jobs into related file (for example, `ui/jobs/component/blm.js` for black mage).
And then just dynamic import them to the main class.

quisquous:
I think we could do both.  The order you list above sounds great.  I think we can also break these out into individual files later when it becomes easier.  If you'd like to do it, please do this!

PR #2014, opened 2020-11-14
jobs: Rearrange job.css

Echoring:
Sorry for these wall of changes! As #1944 asked, it's very hard to review.
- Sort blocks and classes order by several parts: Bars, Boxes, Side Icons, Stacks, Procs, and Colors
- Add a lot of comment to show why this width and when you can reuse these class
- Merge some similar class
- Change some resource box width and right side icon position to better match bars
- Change some proc box position to better match bars
- Remove Box 1 defination in Procs, for default has defined that(suggestion needed)

Sorry again for forgot to split to minor changes!


quisquous:
I think this all looks good to me.  Thanks for doing this!  I'm sorry this was such a mess.

If you're interested in more cleanup, I had a bunch of thoughts if any of this sounds interesting or like good ideas to you.  I think these are all ideas for separate cleanups for future patches.  I'd love to hear other thoughts too!

* I feel like there's a lot of repeated `job1-bar, job2-bar, job3-bar {`.  I wonder a little if things like `pld-bar` is a bad idea.  It might be better to just have `.pld .bar` and try to make as many things use `.bar` as possible without having to add a new line for every job.  Especially things like `.war-boxes .text` is sad to repeat for every job.  It'd be nice if we could have one `.boxes .text` rule and if warrior needs something special we could have a `.war.boxes` or `.war .boxes` rule.

* I think there's room to use `calc` a lot more rather than hardcoding sizes and positions.  Things like the size of the hp bar, etc, or positioning buffs next to the resource bar (which could also be done in other ways too like flexbox or relative positioning) rather than per-job offsets for the number of resources.  I think ideally `jobs.js` could just add as many resource boxes as it needed and CSS would just handle it moving buffs without needing to hard code anything per job.

* It's sad to me that all of the proc boxes have to be absolutely positioned.  I think there could be flexbox or grid there.  I think my hope is that you could have  with some margins or gaps to place them exactly where they are now, and we could get rid of all of the proc box positioning by name.

Echoring:
I agree with these thoughts too. I also feel sad that I need to add a lot of similar class when adding a new UI.
But I'm not good at CSS, not sure I can do these all. 
- For the first point, I think I can handle it by add some classes to these elements.
- For the second point, I heard that a CSS class can not get other classes' attributes as variable to calculate itself.
(like `.right-side-icon { left: #hp-bar.width + .boxes > div.width }`) I have no idea for this. 
- For the last point, I think flexbox is a good idea, but I'm not sure how flexbox work. We have 3 type of procs now: less than 3 box, 3 box and 4 box. if using flexbox, UI less than 3 box(mostly 1) may go to too center. I'm not sure how to deal with.

If changes in this PR has nothing wrong, I'll open new branches to do above base on this PR. Let me know if something amiss.

quisquous:
I gave this a brief bit of testing across all the jobs, and everything looked in the right place with the right colors.  Thanks!

PR #2015, opened 2020-11-14
Raidboss: The Ghimlyt Dark timelines and triggers

JLGarber:
The direction the Colossus spins for Magitek Slash is indicated by a head marker, but I couldn't think of a concise way to convey that to the player usefully. Slash itself seems to have a slightly variable timing, up to about 0.5 seconds in either direction, so I just called it good and left it, since everything else seemed to line up okay.

On Prometheus, Needle Gun/Oil Shower are the cast-time telegraphed 90-degree frontal and 270-degree rear cleaves. When running solo, Needle Gun is used twice, while on all my test runs with party members it was invariably Needle Gun followed by Oil Shower. I'm *relatively* sure it's determined by whether there's a target within the affected radius at the time the cast starts, but I figured it was safe to just account for both possibilities.

I didn't see anything easily usable in the logs to indicate where the wall laser was coming from. The glowy spot is visible earlier than the timeline-based notification, but this is the earliest we can tell the user to move without colliding with the Freezing Missile notification. (We could theoretically remove that one though, I just added it because a *lot* of people seem to miss the proximity markers.)

The Soranus Duo is remarkably free of the issues that plagued the Igeyorhm/LahabreaD encounter in Aetherochemical Research Facility. The inactive member's actions don't seem to misalign too much. I did have to remove a couple of syncs here and there where it caused really weird artifacts for `test_timeline`, as for example:

```
2099.987: Matched entry: 2100.0 Artificial Plasma (+0.013s)
2108.146: Matched entry: 2108.4 Innocence (+0.254s)
    Missed sync: Angry Salamander at 2108.2 (last seen at 2331.6209999999996)
2108.400: Matched entry: 2108.2 Angry Salamander (-0.200s)
2108.333: Matched entry: 2108.4 Innocence (+0.067s)
```

I'm not overlay attached to the numerous `--sync--` elements during the Crosshatch intermission, but I thought it might be best to leave them in since we can't sync to the actual Crosshatch casts. (Plus, out of all the phase loops, these syncs are the most exact each run, on the order of milliseconds.)

### DIFF
--- a/ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
+++ b/ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
@@ -1,0 +1,110 @@
+'use strict';
+
+[{
+  zoneId: ZoneId.TheGhimlytDark,
+  timelineFile: 'ghimlyt_dark.txt',
+  timelineTriggers: [
+    {
+      id: 'Ghimlyt Dark Prometheus Laser',
+      regex: /Heat/,
+      beforeSeconds: 5,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Avoid laser from wall',
+        },
+      },
+    },
+  ],
+  triggers: [
+    {
+      id: 'Ghimlyt Dark Jarring Blow',
+      netRegex: NetRegexes.startsUsing({ id: '376E', source: 'Mark III-B Magitek Colossus' }),
+      condition: Conditions.caresAboutPhysical(),
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'Ghimlyt Dark Wild Fire Beam',
+      netRegex: NetRegexes.headMarker({ id: '008B' }),
+      condition: Conditions.targetIsYou(),
+      response: Responses.spread(),
+    },
+    {
+      id: 'Ghimlyt Dark Ceruleum Vent',
+      netRegex: NetRegexes.startsUsing({ id: '3773', source: 'Mark III-B Magitek Colossus', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Ghimlyt Dark Magitek Ray',
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      response: Responses.stackMarkerOn(),
+    },
+    {
+      id: 'Ghimlyt Dark Magitek Slash',
+      netRegex: NetRegexes.startsUsing({ id: '3774', source: 'Mark III-B Magitek Colossus', capture: false }),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Avoid spinning laser',
+        },
+      },
+    },
+    {
+      id: 'Ghimlyt Dark Nitrospin',
+      netRegex: NetRegexes.startsUsing({ id: '3455', source: 'Prometheus', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Ghimlyt Dark Cermet Drill',
+      netRegex: NetRegexes.startsUsing({ id: '3459', source: 'Prometheus' }),
+      condition: Conditions.caresAboutPhysical(),
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'Ghimlyt Dark Freezing Missile',
+      netRegex: NetRegexes.startsUsing({ id: '345C', source: 'Prometheus', capture: false }),
+      suppressSeconds: 5,
+      response: Responses.goMiddle(),
+    },
+    {
+      id: 'Ghimlyt Dark Artifical Plasma',
+      netRegex: NetRegexes.startsUsing({ id: '3727', source: 'Julia Quo Soranus', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Ghimlyt Dark Innocence',
+      netRegex: NetRegexes.startsUsing({ id: '3729', source: 'Julia Quo Soranus' }),
+      condition: Conditions.caresAboutPhysical(),
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'Ghimlyt Dark Delta Trance',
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.txt
```diff
@@ -0,0 +1,225 @@
...
+1000.0 "Start" sync /00:0839:The Impact Crater will be sealed off/ window 1000,1
+1012.2 "Nitrospin" sync /:Prometheus:3455:/
+1023.4 "Needle Gun/Oil Shower" sync /:Prometheus:(345A|3456):/
+1031.1 "--untargetable--"
+1031.2 "Tunnel" sync /:Prometheus:3457:/ window 31.2,5
+1042.8 "Heat" duration 4 # sync /:Prometheus:3458:/
+1048.0 "--targetable--"
+
+1056.2 "Unbreakable Cermet Drill" sync /:Prometheus:3459:/ window 30,30
+1068.3 "Needle Gun/Oil Shower" sync /:Prometheus:(345A|3456):/
```
quisquous, line 59: Is the first one always needle gun? That's what I read your comments as.

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
```diff
@@ -0,0 +1,115 @@
...
+  timelineFile: 'ghimlyt_dark.txt',
+  timelineTriggers: [
+    {
+      id: 'Ghimlyt Dark Prometheus Laser',
+      regex: /Heat/,
+      beforeSeconds: 5,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Avoid laser from wall',
```
quisquous, line 14:
```suggestion
          en: 'Avoid wall laser',
```

For brevity?

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
```diff
@@ -0,0 +1,115 @@
...
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      response: Responses.stackMarkerOn(),
+    },
+    {
+      id: 'Ghimlyt Dark Magitek Slash',
+      netRegex: NetRegexes.startsUsing({ id: '3774', source: 'Mark III-B Magitek Colossus', capture: false }),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Avoid spinning laser',
```
quisquous, line 49: I think you could do "rotate left" or "rotate right" here, as I suspect most people will be facing the boss?

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
```diff
@@ -0,0 +1,115 @@
...
+      // This head marker is used on players and NPCs, so we have to exclude NPCs explicitly.
+      id: 'Ghimlyt Dark Heirsbane',
+      netRegex: NetRegexes.headMarker({ id: '0001' }),
+      condition: (data, matches) => matches.targetId[0] !== '4' && data.role === 'healer',
+      infoText: (data, matches, output) => {
+        return output.text({ player: data.ShortName(matches.target) });
+      },
+      outputStrings: {
+        text: {
+          en: 'Incoming damage on ${player}',
```
quisquous, line 99: Is this "incoming damage" because it can be on non-tanks?

I think it'd be fine to use Responses.tankBuster here, or at least the same strings.  I could argue that the response is fine as-is, as it will inform tanks, who could use tank shields on the target too?

quisquous (commented):
A few small wording comments:

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
JLGarber: It's substantially similar to the Touch of Death mechanic in Temple of the Fist, where a player is marked for HP-to-critical. But you're right, since it's not HP critical, the buster response is a better option here.

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.txt
JLGarber: I've always seen Needle Gun first, but based on the times I saw Needle Gun second or not, I think it's at least *possible* to see a double Oil Shower. (It may not be worth accounting for, since in theory the tank should always be in front.)

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
JLGarber: Sure. I'll just need to double-check which marker is which. We must do research!

ui/raidboss/data/04-sb/dungeon/ghimlyt_dark.js
```diff
@@ -41,12 +41,16 @@
       response: Responses.stackMarkerOn(),
     },
     {
+      // 00A7 is the orange clockwise indicator. 00A8 is the blue counterclockwise one.
       id: 'Ghimlyt Dark Magitek Slash',
-      netRegex: NetRegexes.startsUsing({ id: '3774', source: 'Mark III-B Magitek Colossus', capture: false }),
-      infoText: (data, _, output) => output.text(),
+      netRegex: NetRegexes.headMarker({ id: ['00A7', '00A8'] }),
+      infoText: (data, matches, output) => {
+        const direction = matches.id === '00A7' ? 'Left' : 'Right';
```
Akurosia, line 48: @quisquous how would translation work here?

quisquous: I think I'd consider it an antipattern at this point to build up strings incrementally.  It makes translating hard.  In the past we've done things like `{ en: 'Left' }[data.options.displayLang]` etc, but I think output strings is preferred over code-based maps so that users can edit them too.

In most cases (like this one), I would suggest spelling them out individually, e.g. `rotateLeft` and `rotateRight`.

For other cases where you do want to combine, I would suggest adding `rotate`, `left`, and `right` as output strings and then passing `output.right()` or `output.left()` as a parameter to `output.rotate`, e.g. `output.rotate({ direction: output.right() })`.  However, I think that's unnecessary indirection for this case.

JLGarber: Thanks for pointing that out, I'll be more careful about that in future!

PR #2017, opened 2020-11-15
Raidboss: Xelphatol timeline and triggers

JLGarber:
I found this rattling around in my fork from September. I don't remember why I forgot about it, but I did do a couple of test runs and everything seemed okay. Nothing of real interest here, it was all straightforward apart from some fiddly bits on the first boss. (Dual intermissions seems excessive for a first boss, but oh well, Square will be Square.)

### DIFF
--- a/ui/raidboss/data/03-hw/dungeon/xelphatol.js
+++ b/ui/raidboss/data/03-hw/dungeon/xelphatol.js
@@ -1,0 +1,83 @@
+'use strict';
+
+// Xelphatol
+[{
+  zoneId: ZoneId.Xelphatol,
+  timelineFile: 'xelphatol.txt',
+  timelineTriggers: [
+    {
+      id: 'Xelphatol Short Burst',
+      regex: /Short Burst/,
+      beforeSeconds: 4,
+      condition: Conditions.caresAboutMagical(),
+      suppressSeconds: 5, // Timelines jump sometimes, so let's not be noisy.
+      response: Responses.tankBuster(),
+    },
+    {
+      // This might be better handled by collecting Swiftfeather,
+      // but a timeline trigger is easy and consistent.
+      id: 'Xelphatol On Low',
+      regex: /On Low/,
+      beforeSeconds: 5,
+      response: Responses.tankCleave(),
+    },
+    {
+      id: 'Xelphatol Ixali Aero Buster',
+      regex: /Ixali Aero \(buster\)/,
+      beforeSeconds: 4,
+      condition: Conditions.caresAboutMagical(),
+      response: Responses.tankBuster(),
+    },
+  ],
+  triggers: [
+    {
+      id: 'Xelphatol Long Burst',
+      regex: Regexes.startsUsing({ id: '19C8', source: 'Nuzal Hueloc', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Xelphatol Hot Blast',
+      regex: Regexes.startsUsing({ id: '19CC', source: 'Floating Turret', capture: false }),
+      response: Responses.getUnder(),
+    },
+    {
+      id: 'Xelphatol On High',
+      regex: Regexes.startsUsing({ id: '19CF', source: 'Dotoli Ciloc', capture: false }),
+      response: Responses.knockback(),
+    },
+    {
+      id: 'Xelphatol Dark Wings',
+      netRegex: NetRegexes.headMarker({ id: '0017' }),
+      condition: Conditions.targetIsYou(),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Puddle on YOU',
+          de: 'Fläche auf DIR',
+          fr: 'Zone au sol sur VOUS',
+          ja: '自分に水溜り',
+          cn: '水球点名',
+          ko: '장판 대상자',
+        },
+      },
+    },
+    {
+      id: 'Xelphatol Ixali Aero AOE',
+      regex: Regexes.startsUsing({ id: '19D5', source: 'Tozol Huatotl', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Xelphatol Bill',
+      netRegex: NetRegexes.headMarker({ id: '0046' }),
+      condition: Conditions.targetIsYou(),
+      response: Responses.spread(),
+    },
+    {
+      id: 'Xelphatol Hawk',
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      response: Responses.stackMarkerOn(),
+    },
+  ],
+}];
--- a/ui/raidboss/data/03-hw/dungeon/xelphatol.js
+++ b/ui/raidboss/data/03-hw/dungeon/xelphatol.js
@@ -16,6 +16,8 @@
     {
       // This might be better handled by collecting Swiftfeather,
       // but a timeline trigger is easy and consistent.
+      // This is *technically* avoidable, but if mishandled can kill melee or unwary healers.
+      // (The timing from "cone locked in" to "cone damage snapshots" is 0.70 seconds.)
       id: 'Xelphatol On Low',
       regex: /On Low/,
       beforeSeconds: 5,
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/03-hw/dungeon/xelphatol.js
```diff
@@ -0,0 +1,83 @@
...
+[{
+  zoneId: ZoneId.Xelphatol,
+  timelineFile: 'xelphatol.txt',
+  timelineTriggers: [
+    {
+      id: 'Xelphatol Short Burst',
+      regex: /Short Burst/,
+      beforeSeconds: 4,
+      condition: Conditions.caresAboutMagical(),
+      suppressSeconds: 5, // Timelines jump sometimes, so let's not be noisy.
```
quisquous, line 13: I really need to just make all timeline triggers auto-suppress for 2-5 seconds by default.

ui/raidboss/data/03-hw/dungeon/xelphatol.js
```diff
@@ -0,0 +1,83 @@
...
+      id: 'Xelphatol On Low',
+      regex: /On Low/,
+      beforeSeconds: 5,
+      response: Responses.tankCleave(),
+    },
+    {
+      id: 'Xelphatol Ixali Aero Buster',
+      regex: /Ixali Aero \(buster\)/,
+      beforeSeconds: 4,
+      condition: Conditions.caresAboutMagical(),
```
quisquous, line 28: I think this is probably fine for a dungeon scenario, but one bad part about caresAboutMagical is that it doesn't include when the target is you.  I'm not sure how to more elegantly handle this, though.

This is maybe another case where it'd be nice to have `ConditionFuncs.caresAboutMagical(data) || ConditionFuncs.targetIsYou(data, matches)` or something.

(cc @panicstevenson FYI, just because you've run into this some lately too)

ui/raidboss/data/03-hw/dungeon/xelphatol.js
```diff
@@ -0,0 +1,83 @@
...
+      suppressSeconds: 5, // Timelines jump sometimes, so let's not be noisy.
+      response: Responses.tankBuster(),
+    },
+    {
+      // This might be better handled by collecting Swiftfeather,
+      // but a timeline trigger is easy and consistent.
+      id: 'Xelphatol On Low',
+      regex: /On Low/,
+      beforeSeconds: 5,
+      response: Responses.tankCleave(),
```
quisquous, line 22: This is technically a getBehind, right? I feel like I remember avoiding this on tank.

ui/raidboss/data/03-hw/dungeon/xelphatol.txt
```diff
@@ -0,0 +1,139 @@
...
+
+28.5 "Air Raid" sync /:Nuzal Hueloc:19CA:/ window 5,6
+36.6 "Air Raid" sync /:Nuzal Hueloc:19CA:/
+48.1 "Air Raid" sync /:Nuzal Hueloc:19CA:/
+59.6 "Air Raid" sync /:Nuzal Hueloc:19CA:/ jump 36.6
+71.1 "Air Raid" 
+82.6 "Air Raid"
+94.1 "Air Raid"
+
+100.0 "--sync--" sync / 1E:........:Nuzal Hueloc loses the effect of Hover/ window 100,5
```
quisquous, line 30: This is maybe a weird nitpick, but this would be the first 1E line we've synced against in any timeline.  What do you think about using a 22 line instead? Or is there no 22 line here?

Part of my feeling here is that `test_timeline.py` only supports certain line types, and fflogs only has certain line types, and also in the future I have some hope of using network log lines for timelines (implicitly, not in the timeline file itself) and so having fewer log types to handle is a sort of nice goal.

ui/raidboss/data/03-hw/dungeon/xelphatol.js
JLGarber: "Fortunately" this doesn't matter for a timeline trigger, but yes, it would be nice if `ConditionFuncs` were a thing.

ui/raidboss/data/03-hw/dungeon/xelphatol.js
JLGarber: It's avoidable, but the timing is tight, on the order of less than a second, and improper avoidance can kill melee. (Or healers who happen to be close by. Not that I've ever done that.)

quisquous: Fair enough! Maybe leave this as a comment there for anybody else who sees it?

ui/raidboss/data/03-hw/dungeon/xelphatol.js
quisquous: Oh right, it's a timeline trigger :+1:.

panicstevenson: RE: `ConditionFuncs`, another alternative (potentially a stopgap) would be to add `Conditions.physicalTankbuster()` and `Conditions.magicalTankbuster()` which would be `matches.target || Conditions.caresAboutX()`. It would be better to have composable parts to use as building blocks, but I don't know how many things would benefit from this aside from Tankbusters.

ui/raidboss/data/03-hw/dungeon/xelphatol.txt
JLGarber: There is none. We could in theory wait for an auto-attack.... somewhere? I'm fine with trying to find something else to sync against, but this is the earliest we get to know the floating phase is over.

```
22|2020-05-03T01:12:36.4080000-04:00|400277E4|Floating Turret|19CC|Hot Blast|[PLAYER 1]|[PLAYER 1 ID]|
00|2020-05-03T01:12:36.0000000-04:00|282b||The floating turret uses Hot Blast.|a76b0cd08d41e832b6ac569ec5cfa33a
00|2020-05-03T01:12:36.0000000-04:00|2929||   [PLAYER 1 ID]Balmung takes 633 damage.|ef8c603405355edb168963f92676f5db
37|2020-05-03T01:12:36.9870000-04:00|[PLAYER 1]|[PLAYER 1 ID]|0000C140|18742|30385|10000|10000|0|
00|2020-05-03T01:12:36.0000000-04:00|2ab0||Nuzal Hueloc loses the effect of Invincibility.|8bc410764d917a0e48548b90941fd59f
00|2020-05-03T01:12:36.0000000-04:00|2ab0||Nuzal Hueloc loses the effect of Hover.|
00|2020-05-03T01:12:36.0000000-04:00|2ab0||Nuzal Hueloc loses the effect of Damage Up.|
34|2020-05-03T01:12:37.3880000-04:00|4002779D|Nuzal Hueloc|4002779D|Nuzal Hueloc|01|
38|2020-05-03T01:12:37.3880000-04:00|4002779D|Nuzal Hueloc|003C3C00|167913|320000|8840|8840|0|0|
30|2020-05-03T01:12:37.3880000-04:00|19c|Hover|0.00|4002779D|Nuzal Hueloc|4002779D|Nuzal Hueloc|
30|2020-05-03T01:12:37.3880000-04:00|307|Invincibility|0.00|E0000000||4002779D|Nuzal Hueloc|
30|2020-05-03T01:12:37.3880000-04:00|1bb|Damage Up|0.00|E0000000||4002779D|Nuzal Hueloc|
39|2020-05-03T01:12:38.4140000-04:00|[PLAYER 1]|[PLAYER 1 ID]|
04|2020-05-03T01:12:39.0380000-04:00|400277e4|Floating Turret|0|3c|0|0|
21|2020-05-03T01:12:40.5580000-04:00|4002779D|Nuzal Hueloc|368|Attack|[PLAYER 1 ID]|[PLAYER 1]
```

(Lines cut short for space reasons.)

We could use the "targetable" line instead, if you would prefer, but otherwise we're waiting for something to happen.

quisquous: Yeah, I think the 22 line is preferable to the 1E line.

JLGarber: Thanks for noting that, I'm now not even sure why I used `losesEffect` in the first place, `22` is exactly the same timing.

ui/raidboss/data/03-hw/dungeon/xelphatol.js
JLGarber: I'll think it over a little more and then open an issue about it later tonight. There's definitely something there!

ui/raidboss/data/03-hw/dungeon/xelphatol.js
JLGarber: Likely so. The edge cases where we *would* want the noise haven't come up yet, to my knowledge.

PR #2019, opened 2020-11-15
Varis - fixed KO name from bladesblood (검기) to gunshield (건실드)

Akurosia:
@Jaehyuk-Lee could you please double check if this is now correct (or explain why the name differs in KO) ?


Jaehyuk-Lee:
Ah I made mistake there :sweat_smile:

Akurosia:
ahh ok good to know :)

PR #2020, opened 2020-11-16
eureka: fix tracker functions

Jaehyuk-Lee:


ui/eureka/eureka.js
```diff
@@ -2723,7 +2723,7 @@ class EurekaTracker {
         }
       }
 
-      if (!this.zoneInfo.hasTracker)
+      if (!this.zoneInfo.hasTracker && this.zoneInfo.hasTracker !== undefined)
```
quisquous, line 2726: Oh oops, sorry for not catching this.  Could you leave this conditional the same, and add `hasTracker: true` to the four Eureka zones instead?

quisquous:
Thanks for the update.  I think it's just generally much more understandable when a missing value is the same as false.

PR #2022, opened 2020-11-16
eureka: bozja Korean translated names

Jaehyuk-Lee:
reference: https://docs.google.com/spreadsheets/d/1Nzh1Nb77E4ZMPmdTruqK3X-N9-15nvOpYvf7KLBYYlE/edit#gid=2004262041

* This is not official translation.


quisquous:
These don't need to be official names, for what it's worth.  Ideally you can find something briefer than the full names of the skirmishes and critical engagements.

Una-p:
I'm not sure this is right. It's a translation in advance before they have official one. We didn't translate in advance that have official translation. It will make confuse.

Jaehyuk-Lee:
This translation is based on Japanese client, and mostly Korean translation is based on Japanese ones. So, English client Korean users can be confused with it.

MaikoTan:
I agree what Una-P said.
Those non-official translations are based on japanese version, but you may know that these many terms are different in eu/na and jp.
This really confused someone who playing en/de/fr but using ko as main language in ACT.

Jaehyuk-Lee:
"The Hunt for Red Chotober"'s boss name is "Red Comet", and Japanese name says "赤レッドコメット", maybe the Japanese names are basically named their bosses? I don't really play global server, so I don't exactly know many FATEs there. I have 4 days free login with the campaign.

Una-p:
as you know, almost korean official translation is based on japanese but sometimes based on english or own liberal translation, too.
I'd like to say it is right not to translate because I don't know what the official translation will be like.

Jaehyuk-Lee:
Adding an option for this looks bad, so reverting this seems better.

Una-p:
Thanks for consideration in this regard. :)

Jaehyuk-Lee:
@quisquous please revert this commit (af63e76c962bd35cb39a1f86e1f292d3c1f505f0), revert command would be much easier than making new PR.

quisquous:
Done.  Thanks for the discussion.  It's good to have some agreement about whether it makes sense to land these sorts of translations ahead of time.  For what it's worth, I think it can still make sense to land some trigger translations ahead of the content arriving (since you don't need to know the specific in game name of an ability or mob to say something like "go left" necessarily)?  What do you think?  Or, is there some better way to state a more clear rule here?

Jaehyuk-Lee:
This issue is similar with timeline translating issue. Timeline is not translated before official patch. Trigger is okay, there's few transliterated ones but I write both English and Jspanese ones, so it mostly doesn't matter.

Una-p:
Same as Lee. I think we can proceed as we have done so far. Not to translate the name or the name of the skill that the official name is given. However, other triggers or naming we are already know(unreal shiva etc..) can translate.

PR #2024, opened 2020-11-17
ci: Fix GitHub Actions Using set-env

panicstevenson:
Apparently there is a [security vulnerability](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) in using `set-env`. Updating all `set-env` invocations to use the `$GITHUB_ENV` file, as described in [the updated workflow documentation.](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files)

### DIFF
--- a/.github/workflows/build-artifact.yml
+++ b/.github/workflows/build-artifact.yml
@@ -55,7 +55,7 @@ jobs:
         python .\util\fetch_deps.py
 
     - name: Set up msbuild
-      uses: microsoft/setup-msbuild@v1.0.1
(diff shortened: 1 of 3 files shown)


quisquous:
Thanks! Were you able to test this on your branch?

panicstevenson:
So, this PR itself ran `msbuild`, which had two issues, but the release stuff has not been tested. Although, I can test that after webpack (prior to another release?) once my repo is freed up?

PR #2027, opened 2020-11-17
jobs: always show pull bar

Jaehyuk-Lee:
I wanted pull bar on because I should use Battle Litany on 5 seconds before battle starts (on e8s), but there were no separate option for pull bar. Just Party Buffs option turned off pull bar.

### DIFF
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -34,7 +34,7 @@
   font-size: 12px;
 }
 
-.justbuffs #pull-bar {
+.nopullbar #pull-bar {
   display: none;
 }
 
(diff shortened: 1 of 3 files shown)


quisquous:
Do you think we should always show the pull bar, vs having an option?

Jaehyuk-Lee:
I would like to have always show the pull bar, since it's shown before combat. It doesn't disturb you.

quisquous:
I'm supportive of that.  I agree it doesn't really get in the way, so maybe it's better to just always show it rather than have an option.

quisquous:
Thanks!

quisquous:
If folks really don't want the pull bar, they can add CSS.

PR #2031, opened 2020-11-18
refactor: convert object to static class

MaikoTan:


resources/user_config.js
```diff
@@ -16,17 +16,20 @@ const Responses = _Responses;
...
-  optionTemplates: {},
-  userFileCallbacks: {},
-  savedConfig: null,
-  registerOptions: function(overlayName, optionTemplates, userFileCallback) {
+
+class UserConfig {
+  constructor() {
+    optionTemplates = {};
+    savedConfig = null;
+    userFileCallbacks = {};
```
quisquous, line 24: You're missing a `this.` in front of each of these.

resources/user_config.js
```diff
@@ -256,11 +260,18 @@ let UserConfig = {
...
+      this.instance = new UserConfig();
+    return this.instance;
+  }
+}
+
+const userConfig = UserConfig.getInstance();
 
-// TODO: Convert into static class
-export default UserConfig;
+export default userConfig;
```
quisquous, line 274: Can this be `export default new UserConfig()`, without any need for `getInstance`?

resources/user_config.js
MaikoTan: Oops!

resources/user_config.js
MaikoTan: I am not quite sure how webpack working.
If it exports with `new UserConfig()`, it would be still singleton in the module?

quisquous: To answer your question, the UserConfig module will only be evaluated once in cactbot, and so every thing returned from it is inherently a singleton.  More specifically, ecmascript says that if you have the same top level execution context (e.g. `dist/raidboss.bundle.js`) and have the same specifier (e.g. `../resources/user_config.js`, although the browser will deal with relative paths as you would expect), then javascript will return the exact same module instance when importing it.

As an example:
```
import userConfigA from 'user_config.js';
import userConfigB from 'user_config.js';

console.log(userConfigA === userConfigB); // true
```

PR #2032, opened 2020-11-18
Add supported content list

quisquous:


quisquous:
@JLGarber for thoughts?

JLGarber:
Looks good to me. Where is `content_list.js` generated from, or was that manually sorted from the existing resource files?

quisquous:
> Looks good to me. Where is `content_list.js` generated from, or was that manually sorted from the existing resource files?

It was uh manually typed in from duty finder.  <_<

PR #2036, opened 2020-11-18
ci: set retention of artifacts as 30 days

MaikoTan:
According to https://github.blog/changelog/2020-10-08-github-actions-ability-to-change-retention-days-for-artifacts-and-logs/ ,
we can set retention days as 30,
so don't need to keep massive artifact storage within default 90 days.


panicstevenson (approved):

quisquous:
Sounds good, thanks for finding this!

PR #2039, opened 2020-11-18
jobs: Rearrange jobs.js

### LINKED ISSUE #1944: Jobs: Rearrange code by job order

Echoring:
- Sort list, func block, kability, combo, etc by jobs order
- Add comment to split kability by jobs
- Fix RageOfHalone upper case not matched
- closes #1944

By the way, I notice that there're a lot of unused kability/combo here. May I going to clean them up or just leave it be for probably later use?

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -24,8 +24,8 @@ const kWellFedContentTypes = [
 
 // See user/jobs-example.js for documentation.
 let Options = {
-  ShowHPNumber: ['PLD', 'WAR', 'DRK', 'GNB', 'BLU', 'AST', 'WHM', 'SCH'],
-  ShowMPNumber: ['PLD', 'DRK', 'BLM', 'AST', 'WHM', 'SCH', 'BLU'],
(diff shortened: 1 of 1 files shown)


quisquous:
> By the way, I notice that there're a lot of unused kability/combo here. May I going to clean them up or just leave it be for probably later use?

I think kAbility is fine to have extra.  It does make readability worse, so maybe it would be worth splitting out into a `resources/abilities.js` or `ui/jobs/abilties.js` file.  But from a functionality/performance point of view, it's just a lookup table for name -> ids, so having extra isn't a big deal.

Combos seem more tricky and might make sense to only include the ones that are actually used.  I wonder if it would make sense to refactor jobs so that particular job functions add the combos and combo breakers they care about themselves, rather than adding them all up front.

PR #2042, opened 2020-11-19
Fix: change font when display lang set to ja

MaikoTan:
The default font family makes Japanese shows weird, so add pre-defined font-family css rules.

### DIFF
--- a/resources/user_config.js
+++ b/resources/user_config.js
@@ -83,6 +83,11 @@ let UserConfig = {
       if (!supportedLanguage.includes(options.DisplayLanguage))
         options.DisplayLanguage = options.ParserLanguage || 'en';
 
+      document
+        .getElementsByTagName('body')[0]
+        .classList
+        .add('lang-' + options.DisplayLanguage);
+
       this.addUnlockText(options.DisplayLanguage);
 
       // Handle processOptions after default language selection above,
--- a/resources/user_config.js
+++ b/resources/user_config.js
@@ -83,11 +83,7 @@ let UserConfig = {
       if (!supportedLanguage.includes(options.DisplayLanguage))
         options.DisplayLanguage = options.ParserLanguage || 'en';
 
-      document
-        .getElementsByTagName('body')[0]
-        .classList
-        .add('lang-' + options.DisplayLanguage);
-
+      document.body.classList.add(`lang-${options.DisplayLanguage}`);
       this.addUnlockText(options.DisplayLanguage);
 
       // Handle processOptions after default language selection above,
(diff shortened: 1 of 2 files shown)


panicstevenson:
Forgive my potential ignorance here; what looks weird about it? Would this not apply to Chinese and Korean characters as well?

MaikoTan:
Sure!
But I don't know which font that Korean uses, @Jaehyuk-Lee would you mind tell me some popular Korean font ?

Jaehyuk-Lee:
Korean is okay with Arial, don't know if Arial has Korean font data, it seems very similar to Korean Windows default font "Malgun Gothic". It might be loading Malgun Gothic for Korean, if Arial doesn't have Korean font data.

MaikoTan:
I found it may happened when the sentence including 仮名(kana) and 漢字 and English letters, some characters use serif fonts and some use san serif.
Because Korean rarely use 漢字 in general, so it might not a problem.

resources/user_config.js
```diff
@@ -251,6 +251,7 @@ let UserConfig = {
       textElem = document.createElement('div');
       textElem.id = id;
       textElem.classList.add('text');
+      textElem.classList.add('lang-' + lang);
```
quisquous, line 254: Do you think it would make sense for the `loadUser` function (the caller of this `addUnlockText` function) to add `.lang-ja` to the top of the page, e.g. the body element? It seems like this is something that it might be nice to depend on for every overlay via a `.lang-ja .text` selector.

MaikoTan: Sounds great, I'll try to do this!

Jaehyuk-Lee:
I guess this font problem differs by system language on Windows. I've used English Windows and Korean Windows, both version had different font on English. I had similar problem when I developed DFAssist with Japanese translation, I had no problem reading 漢字 in my PC but Japanese user said they could hardly read or even couldn't read 漢字, so they requested me to add Meiryo font setting for Japanese.
K-004

resources/user_config.js
```diff
@@ -83,6 +83,11 @@ let UserConfig = {
       if (!supportedLanguage.includes(options.DisplayLanguage))
         options.DisplayLanguage = options.ParserLanguage || 'en';
 
+      document
+        .getElementsByTagName('body')[0]
+        .classList
+        .add('lang-' + options.DisplayLanguage);
+
```
quisquous, line 90:
```suggestion
      document.body.classList.add(`lang-${options.DisplayLanguage}`);
```

MaikoTan: Oh, silly me!

MaikoTan:
Okay, thanks again for your explanation !
I think I would open a new PR for Chinese font later, when I have time to test which shows better. This one can be merged now.

PR #2043, opened 2020-11-20
Jobs: Use class selector instead of id

Echoring:
Once discussed in #2014 
Retain id for easier access when special needed.

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -1617,10 +1617,10 @@ class Bars {
     timerBox.hideafter = '';
     timerBox.roundupthreshold = false;
     timerBox.valuescale = options.scale ? options.scale : 1;
-    if (options.id)
+    if (options.id) {
       timerBox.id = options.id;
       timerBox.classList.add('timerbox');
-
+    }
     return timerBox;
   }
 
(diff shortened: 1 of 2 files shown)


ui/jobs/jobs.js
```diff
@@ -1601,6 +1603,7 @@ class Bars {
       container = document.createElement('div');
       container.id = id;
       document.getElementById('bars').appendChild(container);
+      container.classList.add('procbox');
```
MaikoTan, line 1606:
```suggestion
      container.classList.add('proc-box');
```

add a '-' is better

ui/jobs/jobs.js
```diff
@@ -1614,9 +1617,10 @@ class Bars {
     timerBox.hideafter = '';
     timerBox.roundupthreshold = false;
     timerBox.valuescale = options.scale ? options.scale : 1;
-    if (options.id)
+    if (options.id) {
       timerBox.id = options.id;
-
+      timerBox.classList.add('timerbox');
```
MaikoTan, line 1622:
```suggestion
      timerBox.classList.add('timer-box');
```

also here

ui/jobs/jobs.js
```diff
@@ -1628,6 +1632,7 @@ class Bars {
     let timer = document.createElement('timer-bar');
     container.appendChild(timerDiv);
     timerDiv.appendChild(timer);
+    timer.classList.add('timerbar');
```
MaikoTan, line 1635:
```suggestion
    timer.classList.add('timer-bar');
```

and here

quisquous:
Thanks! This is much improved!

PR #2044, opened 2020-11-21
general: bundle data files in webpack

quisquous:
Rather than loading files via a C# cactbotReadDataFiles overlay handler,
instead bundle all of the trigger and timeline files together.  These
use static imports rather than dynamic, which put them in the same
bundle.  Dynamic imports create chunks, which (I think?) require us to
process html files as well in order for this to work (as I think chunks
expect the path of the html files to be in the dist/ folder as well).
Something to consider for the future.

One thing we can do as a followup from here is to change the trigger
files to be ESM modules themselves and import rather than eval them.


webpack/loaders/manifest-loader.cjs
```diff
@@ -0,0 +1,29 @@
...
+  let outputStr = 'export default {';
+
+  for (let name of lines) {
+    if (/^\s*$/.test(name))
+      continue;
+    name = name.replace(/\\/g, '/');
+    name = name.replace(/^\//, '');
+
+    // Use static imports instead of dynamic ones to put files in the bundle.
+    const fileVar = `file${fileIdx++}`;
```
panicstevenson, line 20: This one's minor, but since we're already regexing, we might as well have it cut out potential whitespace instead of checking below.
```suggestion
  const lines = contents.trim().split(/\s*[\r\n]+\s*/);

  let importStr = '';
  let outputStr = 'export default {';

  for (const fileIdx in lines) {
    // normalize filepaths between windows / unix
    const name = lines[fileIdx].replace(/\\/g, '/').replace(/^\//, '');

    // Use static imports instead of dynamic ones to put files in the bundle.
    const fileVar = `file${fileIdx}`;
```

webpack/loaders/manifest-loader.cjs
```diff
@@ -0,0 +1,29 @@
+'use strict';
+
+let fs = require('fs');
```
panicstevenson, line 3:
```suggestion
const fs = require('fs');
```

panicstevenson (approved):
Some nits, but LGTM.
(I thought I turned on `prefer-const` 😭)

ui/raidboss/raidemulator.js
```diff
@@ -208,13 +180,28 @@ let Options = {
...
         encounterTab.refresh();
 
         // If we don't have any encounters stored, show the intro modal
         persistor.listEncounters().then((encounters) => {
-          if (encounters.length === 0)
+          if (encounters.length === 0) {
             showModal('.introModal');
+          } else {
+            let lastEncounter = window.localStorage.getItem('currentEncounter');
+            if (lastEncounter !== undefined && lastEncounter !== undefined) {
```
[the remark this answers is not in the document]
quisquous: `lastEncounter` is a number, so it needs the undefined check here.  Fixed to only have one undefined check, though!

PR #2045, opened 2020-11-21
Raidboss: Fix mirror color not matched

Echoring:
In fact this mirror color is red.
Other languages do not translate the color so not need to add.


quisquous:
Whoops, it is red.  Thanks!

PR #2051, opened 2020-11-22
Jobs: Delete unused combo

Echoring:
- Royal Authority combo is not used(Atonement use effect id to work).
- SAM has at most 2 Iaijutsu to insert, comboTimer is not needed.
- Seems RDM UI no longer care about combo(and combo breaks at start casting, not cast out).


quisquous:
Yeah, rdm used to have 1 2 3 boxes that lit up for where you were in the combo, but I didn't think it was very useful and so I removed it a while back.

Thanks for the cleanup!

PR #2052, opened 2020-11-22
Jobs: Fix bar in stack guage

Echoring:
Due to #2043 , there is a mistake that an unnecessary bar appear in the stack guage.
Other bars define there height there own, but delete this line seems result in a default height, so change it to 0.


quisquous:
Thanks for the fix!

PR #2054, opened 2020-11-23
Convert trigger files into modules

quisquous:
This is a bit rough around the edges, but I wanted to take a pass at it.  Remaining:
- [x] clean up some node test/util initialization (or do that in a followup)
- [x] fix documentation about trigger file format
- [x] convince eslint that dynamic import is acceptable

The non-obvious change here is that previously trigger files were an array of "trigger sets", where a trigger set is an object that contains fields like zoneId, triggers, timeline.  Now, trigger files are modules that export a single trigger set.  All files conformed to this already and many tests assumed there was only one, so it seemed like a good time to make this change as well.  This doesn't affect user triggers.

Also, I did this with a script, so if you want to be picky about the changes to trigger files, I can easily re-do that.  I think looking at this original diff, I probably should add a blank line after imports always.


ui/oopsyraidsy/data/00-misc/buffs.js
```diff
@@ -1,3 +1,6 @@
+import NetRegexes from '../../../../resources/netregexes.js';
```
MaikoTan, line 1: Instead of import with a long relative path, what about setting a [path alias](https://webpack.js.org/configuration/resolve/#resolvealias) to shorten it?

quisquous: Oh neat! Unfortunately, a number of tests and utility scripts load files outside of webpack contexts, and so I don't think it's safe to use webpack features like that everywhere since those tests and scripts won't know about it.

quisquous:
@panicstevenson @JLGarber Do y'all have any opinions about this change to trigger files and trigger file loading?

JLGarber:
Nothing to add from my end beyond agreement with "blank line after imports is good". It's all very clear and understandable.

docs/RaidbossGuide.md
```diff
@@ -4,8 +4,13 @@
 
 ## File Structure
 
+Each trigger file is a module that exports a single trigger set.
+
 ```javascript
-[{
+import ZoneId from '../etc/resources/zone_id.js';
```
panicstevenson, line 10: Microscopic nit, but since there's no `/etc/` in the repository structure, it might be more obvious to say something like:
```suggestion
import ZoneId from '../path/to/resources/zone_id.js';
```

panicstevenson (approved):

package.json
```diff
@@ -165,7 +165,7 @@
       ]
     },
     "parserOptions": {
-      "ecmaVersion": 2018,
+      "ecmaVersion": 2020,
```
panicstevenson, line 168: 🎉

PR #2055, opened 2020-11-23
Util: Automate zone seal messages in make_timeline

JLGarber:
This is another one of those little annoying things that bugged me enough to fix. This will automagically fill in the zone name, title-cased, if the start event is a zone seal.

I'm really not happy with the setup I used to do this, but the approach suffers from some circular problems. `is_zone_seal` depends on being fed pre-split log lines, but during each iteration of the loop, we don't split until after we might already have rejected a zone-seal message. I'd rather not split every log line for performance reasons. I guess maybe we could instead rewrite the "set `started` to true" block to be more like this?

```
            if not started:
                started = True
                last_ability_time = e_tools.parse_event_time(line)
                if e_tools.is_zone_seal(line.split("|")):
                    entry = make_entry(
                        {
                            "line_type": "zone_seal",
                            "time": e_tools.parse_event_time(line),
                            "zone_message": line_fields[4].split(" will be sealed off")[0]
                        }
                    )
                    entries.append(entry)
```

I'm also not terribly happy with selecting the type of start line in `main`, but there's not a lot of help for that without a drastic re-write.

EDIT: Never mind, it's clearly better to make the special entry at the time we know it's the start line. Let's just avoid the considerations altogether.

### DIFF
--- a/util/make_timeline.py
+++ b/util/make_timeline.py
@@ -175,15 +175,32 @@ def parse_file(args):
 
             # We're looking for enemy casts or enemies becoming targetable/untargetable.
             # These lines will start with 21, 22, or 34, and have an NPC ID (400#####)
-            # If this isn't one, skip the line
+            # If there's a zone seal, we want that also.
+            # If none of these apply, skip the line
 
-            if not (line[0:2] in ["21", "22", "34"]) or not line[37:40] == "400":
+            if (not line[0:2] in ["21", "22", "34"] or not line[37:40] == "400") and line[0:2] != "00":
                 continue
             line_fields = line.split("|")
             # We aren't including targetable lines unless the user explicitly says to.
             if line[0:2] == "34" and not line_fields[3] in args.include_targetable:
                 continue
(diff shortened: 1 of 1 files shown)


panicstevenson (approved):

quisquous:
What about:

```
if not line[0:2] in ["00", "21", "22", "34"]:
    continue

line_fields = line.split("|")

# new logic about zone seals here

if not line_fields[0] in ["21", "22", "34"] or not line[37:40] == "400":
    continue
```

Then if somebody adds some other line that we want to care about, then we don't have multiple conditional splits, and there's an obvious place for the new code to go.

I think readability > performance concerns is the right metric here.  If this is slow, we can measure it and try to figure out what would make it better.  There's probably larger issues than small optimizations about split.

JLGarber:
Sure, that all seems reasonable. I didn't benchmark it, but I did run a 200MB file through it and the extra overhead from this change felt negligible.

PR #2066, opened 2020-11-25
ci: add eslint custom rules for locale order

MaikoTan:
## Summary

Last month I planed to add [this new feature](https://github.com/MaikoTan/cactbot-highlight/pull/5) in my VSCode extension cactbot-highlight,
but I found it is hard to implement the fixture from VSCode extension because I don't know how to fetch and parse code blocks around.

So I abandoned that approach, and try to implement this feature via eslint's existing ecosystem.

This also allows contributors using different editors or IDEs, not only VSCode supports it, even Github Actions would report warnings.

So, this PR planning to implement just like the previous PR that I mentioned:

- [x] check locale order
- [x] offer a quick fix interface

## Screenshot

image

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -152,6 +152,7 @@
     "unicode-bom": [
       "error",
       "never"
-    ]
+    ],
+    "cactbot-locale-order": "warn"
   }
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -10,5 +10,10 @@
   "search.exclude": {
     "**/node_modules": true,
     "dist/**": true
-  }
+  },
+  "eslint.options": {
+    "rulePaths": [
+      "./eslint"
+    ]
+  },
 }
--- a/eslint/cactbot-locale-order.js
+++ b/eslint/cactbot-locale-order.js
@@ -7,145 +7,6 @@ const orderMap = {
   ko: 5,
 };
 
-/**
- * Determines whether the given node is a `null` literal.
- * @param {ASTNode} node The node to check
- * @return {boolean} `true` if the node is a `null` literal
- */
-function isNullLiteral(node) {
-  /*
-   * Checking `node.value === null` does not guarantee that a literal is a null literal.
-   * When parsing values that cannot be represented in the current environment (e.g. unicode
-   * regexes in Node 4), `node.value` is set to `null` because it wouldn't be possible to
-   * set `node.value` to a unicode regex. To make sure a literal is actually `null`, check
-   * `node.regex` instead. Also see: https://github.com/eslint/eslint/issues/8020
-   */
-  return node.type === 'Literal' && node.value === null && !node.regex && !node.bigint;
-}
-
-
-/**
- * Returns the result of the string conversion applied to the evaluated value of the given expression node,
- * if it can be determined statically.
- *
- * This function returns a `string` value for all `Literal` nodes and simple `TemplateLiteral` nodes only.
- * In all other cases, this function returns `null`.
- * @param {ASTNode} node Expression node.
- * @return {string|null} String value if it can be determined. Otherwise, `null`.
- */
-function getStaticStringValue(node) {
-  switch (node.type) {
-  case 'Literal':
-    if (node.value === null) {
-      if (isNullLiteral(node))
-        return String(node.value); // "null"
-
-      if (node.regex)
-        return `/${node.regex.pattern}/${node.regex.flags}`;
-
-      if (node.bigint)
-        return node.bigint;
-
-
-      // Otherwise, this is an unknown literal. The function will return null.
-    } else {
-      return String(node.value);
-    }
-    break;
-  case 'TemplateLiteral':
-    if (node.expressions.length === 0 && node.quasis.length === 1)
-      return node.quasis[0].value.cooked;
-
-    break;
-
-    // no default
-  }
-
-  return null;
-}
-
-
-/**
- * Gets the property name of a given node.
- * The node can be a MemberExpression, a Property, or a MethodDefinition.
- *
- * If the name is dynamic, this returns `null`.
- *
- * For examples:
- *
- *     a.b           // => "b"
- *     a["b"]        // => "b"
- *     a['b']        // => "b"
- *     a[`b`]        // => "b"
- *     a[100]        // => "100"
- *     a[b]          // => null
- *     a["a" + "b"]  // => null
- *     a[tag`b`]     // => null
- *     a[`${b}`]     // => null
- *
- *     let a = {b: 1}            // => "b"
- *     let a = {["b"]: 1}        // => "b"
- *     let a = {['b']: 1}        // => "b"
- *     let a = {[`b`]: 1}        // => "b"
- *     let a = {[100]: 1}        // => "100"
- *     let a = {[b]: 1}          // => null
- *     let a = {["a" + "b"]: 1}  // => null
- *     let a = {[tag`b`]: 1}     // => null
- *     let a = {[`${b}`]: 1}     // => null
- * @param {ASTNode} node The node to get.
- * @return {string|null} The property name if static. Otherwise, null.
- */
-function getStaticPropertyName(node) {
-  let prop;
-
-  switch (node && node.type) {
-  case 'ChainExpression':
-    return getStaticPropertyName(node.expression);
-
-  case 'Property':
-  case 'MethodDefinition':
-    prop = node.key;
-    break;
-
-  case 'MemberExpression':
-    prop = node.property;
-    break;
-
-    // no default
-  }
-
-  if (prop) {
-    if (prop.type === 'Identifier' && !node.computed)
-      return prop.name;
-
-
-    return getStaticStringValue(prop);
-  }
-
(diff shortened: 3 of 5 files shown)


panicstevenson:
Very nice! I had written something similar on my branch, [here](https://github.com/panicstevenson/cactbot/tree/eslint-plugin/eslint-plugin-cactbot/lib/rules), with the language ordering specifically [here](https://github.com/panicstevenson/cactbot/blob/eslint-plugin/eslint-plugin-cactbot/lib/rules/language-order.js). A lot of it is a huge work in progress, with notable functions like `doThing()`, but that branch has 4 lints, all with fixes (if that helps any).

The only thing is that I'm a little bit confused why your code looks so vastly different from mine? I haven't tried putting my rule into VSCode, so maybe there's something there that I'm missing? I can try looking into that as well.

Additionally, I had this conversation with @quisquous which lead to some hesitation about merging, but how do you feel about the code living here vs a separate repo? Originally, I had liked the idea of a separate repository to use something like TypeScript a little earlier than this repository for easier null checking, but that was before finding out how to use complex [AST node selectors](https://github.com/panicstevenson/cactbot/blob/eslint-plugin/eslint-plugin-cactbot/lib/rules/timelineTrigger-checks.js#L18-L19), so I think that's a moot point now. However, I do still think there's something to the idea that you have to change directories to the `util/eslint` directory, run `npm install` there, and then go back to be able to run `npm install` or `npm run lint` at the root repository level. I don't know if you were able to avoid this issue, but that felt like enough to be a non-starter for me, since people would try to run their code and not know that they would have to do the local install.

But, I think this all looks great! I think this is the [second time](https://github.com/quisquous/cactbot/pull/1597) you've beaten me to the punch on a similar idea! 😛

quisquous:
This is great! Thank you so much for looking into this.  Agreed very much that having eslint for these sorts of things is the right direction.  It's better than tests and it's better than something specific to vscode.

I think my additional thoughts here are:
* It looks like panic's version is much more concise.  Is there some way to integrate code from that version to simplify looking up properties?
* I definitely think having a --fix option before merging this PR would be great.
* I think things in the cactbot codebase should be apache2 licensed, like cactbot.

MaikoTan:
> Very nice! I had written something similar on my branch, [here](https://github.com/panicstevenson/cactbot/tree/eslint-plugin/eslint-plugin-cactbot/lib/rules), with the language ordering specifically [here](https://github.com/panicstevenson/cactbot/blob/eslint-plugin/eslint-plugin-cactbot/lib/rules/language-order.js). A lot of it is a huge work in progress, with notable functions like `doThing()`, but that branch has 4 lints, all with fixes (if that helps any).

Wow! It looks like your version is more complete.

> The only thing is that I'm a little bit confused why your code looks so vastly different from mine? I haven't tried putting my rule into VSCode, so maybe there's something there that I'm missing? I can try looking into that as well.

Those helper functions are mostly copy and paste from eslint official repo, I was just thinking it would be more robust.

> Additionally, I had this conversation with @quisquous which led to some hesitation about merging, but how do you feel about the code living here vs a separate repo?

I think making a separate repo isn't bad, especially when eslint custom rules are much more overhead.

> However, I do still think there's something to the idea that you have to change directories to the `util/eslint` directory, run `npm install` there, and then go back to be able to run `npm install` or `npm run lint` at the root repository level. I don't know if you were able to avoid this issue, but that felt like enough to be a non-starter for me, since people would try to run their code and not know that they would have to do the local install.

That's why I put `file:util/eslint` in `package.json` in the root of the repo.
When someone runs `npm install`, it should be auto handled by `npm`.


> * It looks like panic's version is much more concise.  Is there some way to integrate code from that version to simplify looking up properties?

Yeaah, I agree panic's version is much simple than mine, maybe we can use panic's before merging?

> * I think things in the cactbot codebase should be apache2 licensed, like cactbot.

Sure!

panicstevenson:
Looking at the vscode repo, it looks like they're using inline eslint rules without a package.json or anything:
https://github.com/microsoft/vscode/tree/master/build/lib/eslint
https://github.com/microsoft/vscode/blob/master/package.json
However, they have theirs as `.ts` files and expect you to have compiled to run eslint. I think if we followed, we'd potentially leave the eslint rules as `.js`, or maybe there's a way to use a typescript parser natively so it wouldn't need to be compiled down to JavaScript.

MaikoTan:
Interesting!

They use the [`rulesdir`](https://eslint.org/docs/user-guide/command-line-interface#-rulesdir) CLI option, which eslint called [runtime rules](https://eslint.org/docs/developer-guide/working-with-rules#runtime-rules).

At the first time I use this approach too, but I cannot enable my extra rules in VSCode so I abandoned it.
But now I found VSCode they use the `rulePaths` option to enable this feature in VSCode.
https://github.com/microsoft/vscode/blob/d9c33ab94146645decc6fbd29e677011c60f91ea/.vscode/settings.json#L45-L47

------

But there is another problem when using runtime rules:

I think that the `type` in `package.json` of cactbot is set to `module`, so JavaScript files of eslint rules would not `require()`-able,
when running `eslint` with `rulesdir` option, it complains `Warning: require() of ES modules is not supported.`

If we can solve this problem, then we can directly use runtime rules feature just like VSCode.

------

And the last thing I don't understand: 

how can VSCode uses `.ts` rules files without `ts-node` or something, could `eslint` recognize `.ts` files and invoke the correct binary to handle them?

panicstevenson:
> I think that the `type` in `package.json` of cactbot is set to `module`, so JavaScript files of eslint rules would not `require()`-able.
> When I running `eslint` with `rulesdir` option, it complains `Warning: require() of ES modules is not supported.`

I think there's the option to either add the babel eslint plugin, which was what I found when I was looking into making these ESM myself, or probably more simply, they can be .cjs files? It definitely requires some exploration, but I think this + tucking them away into a folder in the main repo is probably the easiest / least intrusive to people unfamiliar with the ecosystem.

quisquous:
> > Additionally, I had this conversation with @quisquous which led to some hesitation about merging, but how do you feel about the code living here vs a separate repo?
> 
> I think making a separate repo isn't bad, especially when eslint custom rules are much more overhead.

I think the only concern I had is that I want `npm install` to handle everything without extra steps.  It looks like your patch handles this, so that's great.  I think given that cactbot lint rules don't really apply to other projects and that it's much less overhead to keep it in the same repo, that keeping it here would make a lot more sense.

> > I think that the `type` in `package.json` of cactbot is set to `module`, so JavaScript files of eslint rules would not `require()`-able.
> > When I running `eslint` with `rulesdir` option, it complains `Warning: require() of ES modules is not supported.`
> 
> I think there's the option to either add the babel eslint plugin, which was what I found when I was looking into making these ESM myself, or probably more simply, they can be .cjs files? It definitely requires some exploration, but I think this + tucking them away into a folder in the main repo is probably the easiest / least intrusive to people unfamiliar with the ecosystem.

Yeah.  Naming files `.cjs` is the way to specify that this is a CJS module and not an ESM module.  You can see this in `webpack/loaders/manifest-loader.cjs`.

> > * It looks like panic's version is much more concise.  Is there some way to integrate code from that version to simplify looking up properties?
> 
> Yeaah, I agree panic's version is much simple than mine, maybe we can use panic's before merging?

I think we need to combine all of these ideas somehow.  Your version has a much better `meta` section and installs much more easily, and panic's is  shorter and supports `--fix`.  Would you like to update this pull request to combine them, or would you rather have panic update his pull request instead?


One final comment on the code is that I think the `util/` directory is in my mind a "command line scripts" directory.  I think it'd be better to have a top-level `eslint/` directory for this code.

MaikoTan:
So I can add a `package.json` file in `eslint` directory and specify the type as "commonjs", so that eslint would recognize rules files as commonjs module, which solve the problem I mentioned above.

I would continue to working on this branch later, sorry for waiting so long.

panicstevenson:
> So I can add a `package.json` file in `eslint` directory and specify the type as "commonjs", so that eslint would recognize rules files as commonjs module, which solve the problem I mentioned above.
> 
> I would continue to working on this branch later, sorry for waiting so long.

I think you can specify the file extension as `.cjs` and not need the `package.json`

MaikoTan:
> > So I can add a `package.json` file in `eslint` directory and specify the type as "commonjs", so that eslint would recognize rules files as commonjs module, which solve the problem I mentioned above.
> 
> I think you can specify the file extension as `.cjs` and not need the `package.json`

then eslint could not find the rules, so nope...

quisquous:
Sorry, I haven't forgotten about this.  I wanted to investigate the cjs thing a little bit myself, but have been distracted by 5.4 stuff.  I will definitely come back to this when I have some more time.

quisquous:
Sorry for the delay, but this works for me:
```
git rm eslint/package.json
git mv eslint/cactbot-locale-order.js eslint/cactbot-locale-order.cjs
```

`npm run lint` still finds the rules correctly.

quisquous:
Ok, I think this is good to go.  It's missing a `--rulesdir` on the `lint-staged` section below in `package.json` that husky uses, but I can fix that up.  Thanks for the patience and all the work to get this done.  This will be really great to have for everybody.

PR #2068, opened 2020-11-25
Add VSCode Recommendations and Basic Settings

panicstevenson:
Initial VSCode extension recommendations (basically just ESLint and @MaikoTan's highlighting extension from https://github.com/quisquous/cactbot/issues/1846), as well as some simple settings.

File associations won't be upset if the cactbot-highlight extension isn't installed. I don't know if there's any reason to have the `.git` folder appear in VSCode (potentially maybe node_modules can be here too?). Likewise, I don't think there's much point in search results showing dist or node_modules.

Perhaps these can be extended to include third-party files, such as lodash.min.js, etc. but I figured it might be best to ground something small first.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -3,6 +3,7 @@
   // for the documentation about the extensions.json format
   "recommendations": [
     "dbaeumer.vscode-eslint",
+    "EditorConfig.EditorConfig",
     "maikotan.cactbot-highlight"
   ]
 }
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=827846
+  // for the documentation about the extensions.json format
+  "recommendations": [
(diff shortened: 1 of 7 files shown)


.vscode/extensions.json
```diff
@@ -0,0 +1,8 @@
+{
+  // See https://go.microsoft.com/fwlink/?LinkId=827846
```
quisquous, line 2: github syntax highlighting doesn't like these comments, but I think they are also valid here.  (Trailing commas also appear to be valid?? :eyes:)

Do you think it's worth adding this line to the `.gitattributes` file?
```
.vscode/*.json jsonc
```

panicstevenson: We can certainly try and find out what happens!
I just copied/pasted it from other people; it seems like most everyone has these comments in (despite them being so ugly in GitHub)

panicstevenson: I think GitHub still hates it :(

panicstevenson: GitHub likes this now :)

PR #2073, opened 2020-11-26
eslint: add eslint-plugin-import plugin

quisquous:


quisquous:
I think I wanted this mostly because of the "you're importing a file that doesn't exist" warnings as well as the "you're importing a named export that doesn't exist" ones.  It's nice to catch that when writing code, specially with the vscode eslint integration.

https://github.com/benmosher/eslint-plugin-import has the full list of rules

I considered https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/newline-after-import.md but that makes things look weird like:
```
import chai from 'chai'; // wants a newline after this, but it's nice to group?
const { assert } = chai;
```

I also considered https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-mutable-exports.md, but I need to figure out what's going on with `export let gPopupText;` first.

PR #2074, opened 2020-11-28
radar: Log death of monsters

chunderbolt:
Sometimes a hunt can spawn in the middle of the night when someone's AFKing in the zone - they'll see it go up on the logs, but not down, meaning there's no precise timer. This gives them a time of death.


quisquous:
Thanks!

PR #2075, opened 2020-11-28
eslint: Move Rules to .eslintrc

panicstevenson:


package.json
```diff
@@ -9,8 +9,8 @@
   "scripts": {
     "build": "node_modules/.bin/webpack --config webpack.config.cjs --mode production",
     "start": "node_modules/.bin/webpack-dev-server --config webpack.config.cjs --mode development",
-    "lint": "node_modules/.bin/eslint resources ui user test util",
-    "lintfix": "node_modules/.bin/eslint --fix resources ui user test util",
+    "lint": "node_modules/.bin/eslint -c .eslintrc resources ui user test util",
```
quisquous, line 12: I'm surprised that `./.eslint.rc` is not implicitly found.

One additional thought, re: top-level directories, is that if we're going to soon have eslint rules and you're explicitly passing a path, can we have `eslint/eslintrc.json`? Can vscode understand such a thing?

quisquous: Sounds like punt for now, maybe think about it again in the future.

PR #2098, opened 2020-12-04
util: add coverage report

### LINKED ISSUE #1740: Utilities: Auto-generation of "currently supported content" list
It seems to be a moderately frequent occurrence that someone will ask, here or on Discord, about whether content XYZ is supported by cactbot. Most of us here know that we can check the recent releases/commit history to see what may have been recently added, but for users unaware that Github exists, that might not be intuitive. Additionally, it is almost certainly not intuitive for users to check the Raidboss `manifest.txt`  It Would Be Nice to have a single human-readable list that can be linked in response to this frequent question. I'm envisioning Yet Another Release Script that would build this out in much the same way as, say, `hunts.js`.

The easiest would be to scan through `manifest.txt` and check for whether both a timeline and a trigger file exist. Obviously this wouldn't be a perfect heuristic, but for old content it should work well enough. (This would maybe require that we don't do placeholder timelines though when new content releases? Could be an issue.) A slightly more difficult implementation might parse whether any given trigger file contains triggers or not.

Maybe in addition we could combine this with parsing whether or not a trigger file has a given language translated or not, since that's another common question?

quisquous:
Running `npm run coverage-report` generates the data for this
report.  After that, `util/coverage/coverage.html` will contain a
breakdown of all of the content that cactbot supports.

Arguably, the data collection could be done internal to `coverage.js`
when building the webpage, but there's something nice about having a
multi-step process here separating out the model from the view.

Doing this would allow for different views (markdown?) or alternatively
having a convenient way to generate coverage-over-time graphs for fun.

Fixes #1740.

Follow-ups here include:
* translation support
* oopsy support
* ability to click on / link to a particular row (for discord support)


quisquous:
Looks like this:
image

panicstevenson (approved):

JLGarber (approved):

PR #2099, opened 2020-12-05
COINACH: Add ability to specify file locations in ENV-variable

Akurosia:


MaikoTan (approved):
GREAT!!

(by the way, use double quotes instead of single quotes in python files)

quisquous:
Thanks!

PR #2106, opened 2020-12-07
Add Eden's Promise files

quisquous:


Akurosia:
Is there a reason why the dungeon and trials are not prepared?

quisquous:
> Is there a reason why the dungeon and trials are not prepared?

Mostly because I know that I'm less sure on when those will get done.  I'm a little torn on adding files before they're ready, but I know the raids will get done before anything else.

PR #2112, opened 2020-12-09
raidboss: Initial E9N Triggers

panicstevenson:


JLGarber (approved):

ui/raidboss/data/05-shb/raid/e9n.js
```diff
@@ -1,8 +1,153 @@
...
+      },
+    },
+    {
+      id: 'E9N Rejuvenating Balm',
+      netRegex: NetRegexes.startsUsing({ id: '55E2', source: 'Cloud Of Darkness', capture: false }),
+      durationSeconds: 5,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Away From Tethered Walls',
```
quisquous, line 111: Is it possible to make this "go N/S" or "go E/W"? Or, am I overthinking and this is sufficient?

ui/raidboss/data/05-shb/raid/e9n.js
```diff
@@ -1,8 +1,153 @@
...
+      durationSeconds: 5,
+      response: Responses.aoe(),
+    },
+    {
+      id: 'E9N Mire Of Despair',
+      netRegex: NetRegexes.startsUsing({ id: '5B07', source: 'Cloud Of Darkness', capture: false }),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Out from Under',
```
quisquous, line 53: Sorry, maybe I'm tired and don't remember what this mechanic is, but what does this mean? Is this getOut? or outOfMelee? or?

ui/raidboss/data/05-shb/raid/e9n.js
```diff
@@ -1,8 +1,153 @@
...
+    },
+    {
+      id: 'E9N Stygian Break Tether',
+      netRegex: NetRegexes.headMarker({ id: '000C' }),
+      condition: Conditions.targetIsYou(),
+      delaySeconds: 3,
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Away From Bramble',
```
quisquous, line 88: `Responses.breakChains`?

ui/raidboss/data/05-shb/raid/e9n.js
```diff
@@ -1,8 +1,153 @@
...
+      response: Responses.knockback('info'),
+    },
+    {
+      id: 'E9N Stygian Tether',
+      netRegex: NetRegexes.headMarker({ id: '000C' }),
+      condition: Conditions.targetIsYou(),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Wait Near Bramble',
```
quisquous, line 76: Sorry, can you remind me again what this is too?

JLGarber: This is the "wait for your harmless tether to turn into a thorn tether" bit.

ui/raidboss/data/05-shb/raid/e9n.js
JLGarber: If I'm remembering it correctly, this is during the tile phase when the center tile is gone? I'm not sure either of those responses would be correct enough in this situation.

ui/raidboss/data/05-shb/raid/e9n.js
panicstevenson: This is sufficient, but we can certainly try to add that; the Stygian Arbors are their own mobs with cast bars and positions, so you could certainly figure it out. But, it is very forgiving if you're anywhere aside from the absolute worst position.

ui/raidboss/data/05-shb/raid/e9n.js
panicstevenson: During the tile phase, the center tile becomes an insta-death tile, yea. I tried to reuse something, but I don't know if it makes enough sense?

ui/raidboss/data/05-shb/raid/e9n.js
panicstevenson: Yea, you can hang out near your thorn patch with the tether, you also have a purple floor marker beneath you. Once that floor marker goes away you're safe to run away. Only the distance you run once the floor marker disappears is what counts, so this would be so that you don't run across the map, and then run out of room for running / drag the boss across the map.

quisquous: Ahaha thanks for the mechanic explanation.  I didn't actually realize it's how it worked.

ui/raidboss/data/05-shb/raid/e9n.js
quisquous: Looking at the log and video, I think there's a visible ground aoe here that says don't stand in it, so maybe this trigger isn't needed?

PR #2115, opened 2020-12-09
Resources:sorry for missing CN translation of 'The Diadem'

ShadyWhite:


quisquous:
This should be generated automatically from https://github.com/thewakingsands/ffxiv-datamining-cn/blob/master/TerritoryType.csv#L830 which looks like it hasn't been updated yet for 5.3.  Once it has, I can re-run `util/gen_zone_id_and_info.py` and this will be filled in automatically.

PR #2133, opened 2020-12-10
Raidboss: E12N timeline and basic triggers

JLGarber:
*None of this has had live testing, I haven't had time to do that today and won't until much later tonight. `test_timeline` is mostly happy with it, the only timing variability I've seen is when Stock is cast leading into the second and subsequent primal blocks. Other than that, timing seems consistent across all logs I checked.*

This is a middling-complexity timeline. Not overly difficult, but not a nice straightforward one either. I think the in-line comments mostly cover what's needed. Shiva or Titan can be used first after the intermission, but the coda block that follows Shiva/Titan is fixed. (Odd coda blocks are Maleficium, even coda blocks are Formless Judgment.)

From everything I've seen, the opening block and its primal skills are fixed. It's possible Eden may open with Garuda/Ifrit first, but I've never seen it.

Triggers are extremely basic, since I haven't had time to deal with how to best handle Cast/Stock/Release. I would have preferred to use head markers for Formless Judgment, but because the markers are on both tanks, we would have to collect > call to avoid duplicate callouts, and that's super not worth it for a basic tank cleave.

I don't really like the special callout on Diamond Dust, as opposed to `Responses.stopMoving()`, but given that people might be in bad positions and panic during Diamond Dust, I think it's worth it to pre-call the spread.

TODO: Tether mechanic callouts. From preliminary analysis, each tether type is used for one primal, so I'm thinking we'll just do a collect > analyze > call system based on which tethers are seen.

TODO: Handle the EarthShaker bait --> beam intercept mechanic during the intermission.

TODO: Math out the spawn position of the Titanic Bomb Boulders so we can call the safe direction in much the same way we do on E4s.

TODO: Fix the Rapturous Reach trigger so it doesn't double call during the intermission.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e12n.js
+++ b/ui/raidboss/data/05-shb/raid/e12n.js
@@ -23,7 +23,6 @@ export default {
     {
       id: 'E12N Formless Judgment',
       netRegex: NetRegexes.startsUsing({ id: '5873', source: 'Eden\'s Promise' }),
-      condition: Conditions.caresAboutMagical(),
       response: Responses.tankCleave(),
     },
     {
--- a/ui/raidboss/data/05-shb/raid/e12n.js
+++ b/ui/raidboss/data/05-shb/raid/e12n.js
@@ -1,8 +1,65 @@
+import Conditions from '../../../../../resources/conditions.js';
+import NetRegexes from '../../../../../resources/netregexes.js';
+import { Responses } from '../../../../../resources/responses.js';
 import ZoneId from '../../../../../resources/zone_id.js';
-
 export default {
   zoneId: ZoneId.EdensPromiseEternity,
   timelineFile: 'e12n.txt',
   triggers: [
+    {
+      id: 'E12N Maleficium',
+      netRegex: NetRegexes.startsUsing({ id: '5872', source: 'Eden\'s Promise', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'E12N Formless Judgment',
+      netRegex: NetRegexes.startsUsing({ id: '5873', source: 'Eden\'s Promise' }),
+      condition: Conditions.caresAboutMagical(),
+      response: Responses.tankCleave(),
+    },
+    {
+      id: 'E12N Boulders Impact',
+      netRegex: NetRegexes.ability({ id: '586E', source: 'Titanic Bomb Boulder', capture: false }),
+      suppressSeconds: 5,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Between small boulders',
+        },
+      },
+    },
+    {
+      id: 'E12N Boulders Explosion',
+      netRegex: NetRegexes.ability({ id: '586F', source: 'Titanic Bomb Boulder', capture: false }),
+      suppressSeconds: 5,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Move to last explosions',
+        },
+      },
+    },
+    {
+      id: 'E12N Rapturous Reach',
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      response: Responses.stackMarkerOn(),
+    },
+    {
+      id: 'E12N Diamond Dust',
+      netRegex: NetRegexes.startsUsing({ id: '5864', source: 'Eden\'s Promise', capture: false }),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Spread, don\'t move',
+        },
+      },
+    },
+    {
+      id: 'E12N Frigid Stone',
+      netRegex: NetRegexes.headMarker({ id: '0060' }),
+      condition: Conditions.targetIsYou(),
+      response: Responses.spread(),
+    },
   ],
 };
--- a/ui/raidboss/data/05-shb/raid/e12n.js
+++ b/ui/raidboss/data/05-shb/raid/e12n.js
@@ -53,14 +53,15 @@ export default {
       response: Responses.stackMarkerOn(),
     },
     {
-      id: 'E12N Diamond Dust',
+      id: 'E12N Diamond Dust Spread',
       netRegex: NetRegexes.startsUsing({ id: '5864', source: 'Eden\'s Promise', capture: false }),
-      infoText: (data, _, output) => output.text(),
-      outputStrings: {
-        text: {
-          en: 'Spread, don\'t move',
-        },
-      },
+      response: Responses.spread(),
+    },
+    {
+      id: 'E12N Diamond Dust Stop',
+      netRegex: NetRegexes.startsUsing({ id: '5864', source: 'Eden\'s Promise', capture: false }),
+      delaySeconds: 1, // Avoiding collision with the spread call
+      response: Responses.stopMoving('info'),
     },
     {
       id: 'E12N Frigid Stone',
--- a/ui/raidboss/data/05-shb/raid/e12n.js
+++ b/ui/raidboss/data/05-shb/raid/e12n.js
@@ -2,6 +2,14 @@ import Conditions from '../../../../../resources/conditions.js';
 import NetRegexes from '../../../../../resources/netregexes.js';
 import { Responses } from '../../../../../resources/responses.js';
 import ZoneId from '../../../../../resources/zone_id.js';
+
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/raid/e12n.js
```diff
@@ -1,8 +1,65 @@
...
+      response: Responses.tankCleave(),
+    },
+    {
+      id: 'E12N Boulders Impact',
+      netRegex: NetRegexes.ability({ id: '586E', source: 'Titanic Bomb Boulder', capture: false }),
+      suppressSeconds: 5,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Between small boulders',
```
panicstevenson, line 28: How do you feel about bombs vs boulders?
```suggestion
          en: 'Between small bombs',
```

panicstevenson:
Can the TODOs in the description make their way into the file?

ui/raidboss/data/05-shb/raid/e12n.js
JLGarber: I don't care, and you're right, that's probably more consistent with other triggers.

ui/raidboss/data/05-shb/raid/e12n.js
```diff
@@ -1,8 +1,73 @@
...
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      response: Responses.stackMarkerOn(),
+    },
+    {
+      id: 'E12N Diamond Dust',
+      netRegex: NetRegexes.startsUsing({ id: '5864', source: 'Eden\'s Promise', capture: false }),
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Spread, don\'t move',
```
quisquous, line 62: Is this the pre-spread before ice comes out? If so, should it just use the responses spread?

ui/raidboss/data/05-shb/raid/e12n.txt
```diff
@@ -6,5 +6,154 @@ hideall "--sync--"
...
+
+
+# Formless Judgment Block
+1000.0 "Stock" sync /:Eden's Promise:5860:/
+1008.7 "Formless Judgment" sync /:Eden's Promise:5873:/ window 15,15
+1011.5 "--sync--" sync /:Eden's Promise:5879:/
+1020.9 "Release" sync /:Eden's Promise:5861:/
+
+1034.8 "Stock" sync /:Eden's Promise:5860:/
+1044.4 "Junction Shiva?" sync /:Eden's Promise:5862:/ jump 700.0
```
quisquous, line 152: Is it random here which one happens next, or do they always alternate?

ui/raidboss/data/05-shb/raid/e12n.txt
```diff
@@ -6,5 +6,154 @@ hideall "--sync--"
...
+96.7 "Conflag Strike" sync /:Eden's Promise:585D:/
+107.4 "Ferostorm" sync /:Eden's Promise:585E:/
+118.1 "Conflag Strike" sync /:Eden's Promise:585D:/
+128.7 "Ferostorm" sync /:Eden's Promise:585E:/
+140.1 "Maleficium" sync /:Eden's Promise:5872:/
+145.9 "--sync--" sync /:Eden's Promise:585A:/ window 145.9
+148.4 "--untargetable--"
+
+# Intermission. Can be pushed, unknown threshold, maybe 70% HP?
+196.5 "Eternal Oblivion" sync /:Eden's Promise:587A:/ window 50,10
```
quisquous, line 33:
```suggestion
196.5 "Eternal Oblivion" sync /:Eden's Promise:587A:/ window 200,10
```

ui/raidboss/data/05-shb/raid/e12n.txt
```diff
@@ -6,5 +6,154 @@ hideall "--sync--"
...
+
 0.0 "--sync--" sync /:Engage!/ window 0,1
+1.0 "--sync--" sync /:Eden's Promise:4B1D:/ window 1,1
+14.9 "Maleficium" sync /:Eden's Promise:5872:/ window 14.9,5
+22.1 "--sync--" sync /:Eden's Promise:5879:/
+25.3 "Initialize Recall" sync /:Eden's Promise:5859:/
+34.9 "Judgment Jolt" sync /:Eden's Promise:585F:/
+45.6 "Temporary Current" sync /:Eden's Promise:585C:/
+56.3 "Judgment Jolt" sync /:Eden's Promise:585F:/
+67.0 "Temporary Current" sync /:Eden's Promise:585C:/
```
quisquous, line 21: Does this really happen twice in a row, every time??

ui/raidboss/data/05-shb/raid/e12n.txt
```diff
@@ -6,5 +6,154 @@ hideall "--sync--"
...
+# Paths split here. Titan OR Shiva can be used first,
+# and then this first block, always ending in 5861 Release,
+# is followed by:
+
+# Unknown_5879 > Cast > Cast > Malefecium
+# The time between 5861 Release and Unknown 5879
+# is consistent at 9 seconds for both Shiva and Titan.
+
+# It will then continue to the other of Titan/Shiva,
+# which will in turn be followed by the other closing pattern:
```
quisquous, line 79: Sorry, I'm not quite sure that I follow this loop here.

So one potential option is:
Titan> 5861 Release > 5879 > Cast > Cast > Maleficium > Shiva > Stock > Formless Judgment > 5879 > Release?

Then ... random Shiva or Titan after, or repeat?

Sorry for this suggestion, but I think this would be clearer to just write out the whole Shiva or Titan branch separately, rather than having mini blocks with only a few abilities in them.

JLGarber: Sure. It's like this:
```
1. Post-intermission stuff
2. Shiva OR Titan
3. Release > Cast > Cast >Maleficium
4. Titan OR Shiva
5. Stock > Formless Judgment > 5879 > Release
6. Shiva OR Titan
7. Release > Cast > Cast >Maleficium
etc
```

The issue is that lines 3/5/7 are always encountered in the order seen in this table, but lines 2/4/6 can be switched. Thus, there's no way to "know" whether Shiva or Titan will be followed by Maleficium or Formless Judgment. If you prefer to duplicate the Maleficium/Judgment blocks and append them to both Titan/Shiva, we could do that.

ui/raidboss/data/05-shb/raid/e12n.txt
JLGarber: Every log I've seen, yes.

ui/raidboss/data/05-shb/raid/e12n.txt
JLGarber: Oops! Thanks!

ui/raidboss/data/05-shb/raid/e12n.txt
JLGarber: They always alternate, but as I noted under the other comment, there's no way to know which, Shiva/Titan, follows Maleficium vs Judgment.

ui/raidboss/data/05-shb/raid/e12n.js
JLGarber: It is. I guess that probably shouldn't be too bad to have two separate triggers for that. Probably alert on `spread()`and info on `stopMoving()`?

JLGarber:
Okay, that was a *really* bad mistake on the Formless Judgment trigger. Cleaves should not be conditional!

Separate from that, I did confirm that the timeline works as expected. There's some tweaking that will need to happen for the intermission, but it's not important enough to be worth holding things up over. Everything else is accurate and looks clean.

ui/raidboss/data/05-shb/raid/e12n.js
quisquous: Yeah.  I think `spread()` for the pre-spread and `stopMoving()` for ice seems reasonable? Although arguably you slide a very short distance here, so I could also see `spread()` twice?

ui/raidboss/data/05-shb/raid/e12n.txt
quisquous: Ahhh I see now, so you'd need to duplicate everything and have Titan > Maleficium and Titan > Formless Judgement as well as Shiva > Maleficium and Shiva > Formless Judgment.  This is probably fine, then.

ui/raidboss/data/05-shb/raid/e12n.js
JLGarber: I think twice, yes, once on Diamond Dust and once on head markers.

PR #2166, opened 2020-12-12
Raidboss: E12N full trigger support

JLGarber:
The primary mechanic in this encounter being "remember the last thing Eden stocked", it's a very standard collect > analyze > call setup. Each primal has a unique tether when Eden junctions it, so we just use that to identify upcoming mechanics. (We could technically do this with `startsUsing` lines as well, but tethers are nice because there's guaranteed to be exactly one per primal, while the primals may use multiple lines.)

I need a better name for Leviathan. The combination of Leviathan + Ramuh ends up very awkward. (`In + Out`  ????) "Middle section" maybe? Or we could just call primal names and let the user figure it out, but that seems not-ideal.

After further testing, I dropped the initial Spread call on Diamond Dust. It's too easy to pre-spread and end up too far for healers to catch the user after Diamond Dust. Suggestions on that one welcome, since Spread + Thin Ice is a *nasty* combination in pickup content.

Is there any reason not to move `tetherIds` into the tether collection trigger, given that it's used only once? I had anticipated that we would be using them multiple times, but it didn't end up working out that way.


ui/raidboss/data/05-shb/raid/e12n.js
```diff
@@ -48,26 +117,100 @@ export default {
...
+      },
+      infoText: function(data, _, output) {
+        if (data.stacks.length === 1)
+          return;
+        const names = data.stacks.map((x) => data.ShortName(x)).sort();
+        return output.stacks({ players: names.join(', ') });
+      },
+      outputStrings: {
+        stacks: {
+          en: 'Stack (${players})',
```
panicstevenson, line 139: This can probably steal from E7S, since that has split stacks with a similar callout.

panicstevenson (commented):
Generic nit: can the `function(param) {}` calls become `(param) => {}` arrow functions?

panicstevenson:
> I need a better name for Leviathan. The combination of Leviathan + Ramuh ends up very awkward. (`In + Out` ????) "Middle section" maybe? Or we could just call primal names and let the user figure it out, but that seems not-ideal.

I think I would understand `Out + Middle` or `Out Of Melee + Middle`

ui/raidboss/data/05-shb/raid/e12n.js
JLGarber: Ah, I knew I was missing something. I stole the logic from TEA instead, but that was "optical stacks". Totally forgot that E7s had the same mechanic!

JLGarber:
> Generic nit: can the `function(param) {}` calls become `(param) => {}` arrow functions?

Are we moving 100% to arrow functions even if they contain multiple statements? I have no issue with that, just didn't know it was everywhere as opposed to single-statement bodies.

> I think I would understand `Out + Middle` or `Out Of Melee + Middle`
I guess the Ramuh circle *is* small enough to justify putting it that way. Okay.

quisquous:
Yeah, I think the combinations need special handling for Leviathan which is the most ambiguous, maybe:
* `Out + Middle` = `North / South + Out`
* `Middle + Sides` = `Under + Sides`
* `Middle + Intercards` = `Under + Intercards`

It doesn't have to be done here, but I'm thinking about this for the e12s calls.

JLGarber:
Alright, thanks quisquous, that's definitely way better.

ui/raidboss/data/05-shb/raid/e12n.js
```diff
@@ -50,26 +127,109 @@ export default {
...
+      preRun: (data) => data.tethers = data.tethers.sort(),
+      delaySeconds: 0.5, // Tethers should be first in the log, but let's be SURE
+      alertText: (data, _, output) => {
+        if (data.tethers.length !== 2)
+          return;
+        // Leviathan's mechanics aren't easily described in a single word,
+        // so we special-case them.
+        const comboStr = data.tethers[0] + data.tethers[1];
+        if (comboStr in primalOutputStrings)
+          return { alertText: output[comboStr]() };
```
quisquous, line 215:
```suggestion
          return output[comboStr]();
```

Sorry, the file that I sent you had this as a response and not an alertText.

quisquous:
Sorry, just committing my own suggestion here so I can work through a bunch of pull requests.  :)

JLGarber:
No worries, I should have read it more carefully before doing copypasta.

PR #2168, opened 2020-12-12
raidboss: Fix Randomized Head Markers for E9S

panicstevenson:
E9S has randomized head marker values for The Art Of Darkness
telegraphs; this should solve the variance and call out accordingly.


ui/raidboss/data/05-shb/raid/e9s.js
```diff
@@ -369,27 +370,65 @@ export default {
...
-      preRun: (data) => data.artOfDarkness.push('stackWithPartner'),
-      durationSeconds: (data) => data.finalArtOfDarkness ? 16 : 9,
-      alertText: (data, _, output) => {
-        // Perform the callout now, regardless if it's The Second or Third Art Of Darkness
-        if (data.finalArtOfDarkness)
-          data.artOfDarkness.push(data.finalArtOfDarkness);
-        return data.artOfDarkness.map((key) => output[key]()).join(' -> ');
+      // The Art Of Darkness uses head markers with randomized offsets.  The first
+      // charge is always left or right, and we can solve the rest from there.
+      id: 'E9S The Second / Third Art Of Darkness Charge Solver',
```
panicstevenson, line 388: Question here: When I run this in the emulator, it works perfectly. I'm presuming because this trigger is before the Right / Left and Callout triggers. I guess I was expecting to have to have done something to this trigger to add the output to the data.artOfDarkness array within the run, since the triggers have opposing conditions and it didn't feel like they'd both trigger off the same headmarker in order.

Jaehyuk-Lee:
ah.. this was why I didn't get any call-outs for AoD after calling left or right.

panicstevenson:
> ah.. this was why I didn't get any call-outs for AoD after calling left or right.

😭 Sorry about that.

ui/raidboss/data/05-shb/raid/e9s.js
quisquous: I"m not sure what you mean by "it didn't feel like they'd both trigger off the same headmarker in order".  It is definitely the case that triggers are ordered, and two triggers that match the same line will be processed in the same order they are in the file.  (This is also why order has to be carefully preserved when users override triggers.)  Does that answer your question?

panicstevenson: > Does that answer your question?

Yep! Thanks for the explanation!

PR #2169, opened 2020-12-12
Raidboss: E11N timeline and intial triggers

JLGarber:
(@quisquous next time we add placeholders, we can probably include all imports rather than just `ZoneID`, unless that would cause linting problems.)

Nothing really special on this one. The opening block, introducing the basic mechanics, can be Fire or Lightning in a random order. The block following the Serpent intermission is Holy 100%. After Prismatic Deception, any Burnt Strike can be any element, with the exception that two in succession will not be the same. No matter what element Burnt Strike is, anywhere, it will not affect the timing on subsequent mechanics other than Blastburn/Burnout/Shining Blade.

Triggers are largely stolen from quisquous's work on E11S. I double-checked all IDs and verified that all non-healer/tank-specific callouts are functioning.

I haven't done Bound of Faith yet because I'd like to just copy quisquous's work there wholesale. However, I need to find the tether IDs, since on Normal the tethers are element-specific. (Not hard, but I ran out of time and need at least *some* sleep.)

I also need to figure out how best to handle the serpent intermission with the Lightning --> Fire Burst strikes. Maybe with a fresh night's sleep it will be easier.

That all being said, apart from any reviewer notes, this one is ready for merging immediately.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e11n.js
+++ b/ui/raidboss/data/05-shb/raid/e11n.js
@@ -1,8 +1,133 @@
+import Conditions from '../../../../../resources/conditions.js';
+import NetRegexes from '../../../../../resources/netregexes.js';
+import { Responses } from '../../../../../resources/responses.js';
 import ZoneId from '../../../../../resources/zone_id.js';
+
+// EDEN'S PROMISE: ANAMORPHOSIS
+// E11 NORMAL
+
+// TODO: Handle Bound of Faith
+// TODO: Callouts for the intermission Burnt Strike
+// TODO: See whether it's possible to math out the spawn locations for Blasting Zone
 
 export default {
   zoneId: ZoneId.EdensPromiseAnamorphosis,
   timelineFile: 'e11n.txt',
   triggers: [
+    {
+      id: 'E11N Burnished Glory',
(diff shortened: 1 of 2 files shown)


JLGarber:
It's technically not necessary to do the same handling for Bound of Faith that Savage uses, since Normal has unambiguous tethers *and* has head markers for two of them. However, it makes sense to me for the handling to be common. It works, and it means no additional work is needed to implement.

ui/raidboss/data/05-shb/raid/e11n.js
```diff
@@ -1,8 +1,299 @@
...
+          en: 'Line Cleave + Bait',
+          de: 'Linien AoE -> Ködern',
+          ko: '직선 장판 + 장판 유도',
+        },
+      },
+    },
+    {
+      id: 'E11N Burnt Strike Lightning Clone',
+      netRegex: NetRegexes.startsUsing({ source: 'Fatebreaker\'s Image', id: '5645', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Abbild[p] des fusionierten Ascians', id: '5645', capture: false }),
```
JLGarber, line 204: @Akurosia this one will need your attention.

quisquous:
Thanks for this!!

PR #2203, opened 2020-12-13
Add callouts for e10s giga slash drops

valarnin:
e10s Shadow Servant mechanic has a debuff with separate IDs for each shadow direction. This PR adds north/south/east/west callouts for targeted players.

This is my first time working with the cactbot code since the webpack transition and the outputStrings customization, so please let me know if there's something that should be adjusted there.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e10s.js
+++ b/ui/raidboss/data/05-shb/raid/e10s.js
@@ -185,9 +185,41 @@ export default {
       netRegexDe: NetRegexes.startsUsing({ source: 'Schattenkönig', id: '5B2D', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ source: 'Roi De L\'Ombre', id: '5B2D', capture: false }),
       netRegexJa: NetRegexes.startsUsing({ source: '影の王', id: '5B2D', capture: false }),
-      alertText: (data, _, output) => output.text(),
+      durationSeconds: (data) => data.gigaSlashCleaveDebuffDuration,
+      alertText: (data, _, output) => {
+        let ret = output.noShadow;
+        switch (data.gigaSlashCleaveDebuffId) {
+        case '973':
+          ret = output.west;
+          break;
+        case '974':
+          ret = output.east;
+          break;
+        case '975':
+          ret = output.north;
+          break;
+        case '976':
+          ret = output.south;
+          break;
+        }
+        delete data.gigaSlashCleaveDebuffId;
+        delete data.gigaSlashCleaveDebuffDuration;
+        return ret();
+      },
       outputStrings: {
-        text: {
+        north: {
+          en: 'North',
+        },
+        south: {
+          en: 'South',
+        },
+        east: {
+          en: 'East',
+        },
+        west: {
+          en: 'West',
+        },
+        noShadow: {
           en: 'Left Cleave',
           de: 'Linker Cleave',
           ko: '왼쪽 장판',
@@ -200,13 +232,54 @@ export default {
       netRegexDe: NetRegexes.startsUsing({ source: 'Schattenkönig', id: '5B2C', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ source: 'Roi De L\'Ombre', id: '5B2C', capture: false }),
       netRegexJa: NetRegexes.startsUsing({ source: '影の王', id: '5B2C', capture: false }),
-      alertText: (data, _, output) => output.text(),
+      durationSeconds: (data) => data.gigaSlashCleaveDebuffDuration,
+      alertText: (data, _, output) => {
+        let ret = output.noShadow;
+        switch (data.gigaSlashCleaveDebuffId) {
+        case '973':
+          ret = output.east;
+          break;
+        case '974':
+          ret = output.west;
+          break;
+        case '975':
+          ret = output.south;
+          break;
+        case '976':
+          ret = output.north;
+          break;
+        }
+        delete data.gigaSlashCleaveDebuffId;
+        delete data.gigaSlashCleaveDebuffDuration;
+        return ret();
+      },
       outputStrings: {
-        text: {
+        north: {
+          en: 'North',
+        },
+        south: {
+          en: 'South',
+        },
+        east: {
+          en: 'East',
+        },
+        west: {
+          en: 'West',
+        },
+        noShadow: {
           en: 'Right Cleave',
           de: 'Rechter Cleave',
           ko: '오른쪽 장판',
         },
+      },
+    },
+    {
+      id: 'E10S Shadow Servant Cleave Drop',
+      netRegex: NetRegexes.gainsEffect({ effectId: '97[3456]' }),
+      condition: Conditions.targetIsYou(),
+      run: (data, matches) => {
+        data.gigaSlashCleaveDebuffId = matches.effectId;
+        data.gigaSlashCleaveDebuffDuration = matches.duration;
       },
     },
     {


valarnin:
Since the actual cleave happens 9 seconds after this debuff hits, it may make sense to add a `delaySeconds` to the trigger. In addition, I'm not currently cleaning up the `data.gigaSlashCleaveDirection` property. I wasn't sure if it made sense to clean it up.

JLGarber:
Does the cleave occur as the debuff expires? If so, it might be best to do something like `delaySeconds: (data, matches) => parseFloat(matches.duration) - 6,`, or whatever prior warning number you want.

(Also, what would you think of collapsing the switch into two left/right objects and then returning the correct output string from that, rather than stacking up returns all over? Something like 

```javascript
const leftOutputs = {
  '973': output.goWest,
  '974': output.goEast,
  '975': output.goNorth,
  '976': output.goSouth,
}
const rightOutputs = {
  '973': output.goWest,
  '974': output.goEast,
  '975': output.goSouth,
  '976': output.goNorth,
}
return data.gigaSlashCleaveDirection === 'left' ? leftOutputs[matches.effectId]() : rightOutputs[matches.effectId]()
```
in the interests of vertical space?)

valarnin:
For the reformatting of outputs, I agree, that looks much better. I'll make that adjustment.

For the delay, the exact order of operations per the log I'm looking at is:

183690ms - Servant of Shadow debuff applied, duration 9 seconds
192718ms - Bind debuff applied, duration 2 seconds

So the bind is applied at the end of the debuff's duration, the player has to be completely in place at that time. I'm not sure what timing makes sense for the delay then. I'd think that since it's a repositioning trigger, a 5s warning would make sense?

valarnin:
Actually I'm going to have to rework this anyways, the order of operations is backwards re: slash direction vs debuff. The debuff goes out before the slash direction is decided. The `delaySeconds` question is still relevant, though,

quisquous:
I think the other tricky part here is that later on there's towers + cleaves, and on the first set of towers you go to the tower for your cleave position (because it's very tight).  On the second set, it's towers first and then a delay before cleaves, so it can make more sense to do towers then cleaves (although you could do it the other way too arguably).

Mostly saying that it might make more sense to say "west cleave" rather than "go west", as information rather than imperative.

JLGarber:
Yes, in that case informational rather than imperative is the way to go. And in that case, there might not be reason to delay at all. Maybe instead of delaySeconds, make `effect.duration` the value for a `durationSeconds` property?

valarnin:
I've updated to account for the changed order of network lines and added durationSeconds.
Alert text has been changed to "{cardinal} cleave". I don't have translations for languages other than `en` handy for those strings, so only `en` is included.
The default behavior if you don't get the debuff is the same as current `main`.

Have I missed anything?

ui/raidboss/data/05-shb/raid/e10s.js
```diff
@@ -201,15 +233,56 @@ export default {
...
         },
       },
     },
+    {
+      id: 'E10S Shadow Servant Cleave Drop',
+      netRegex: NetRegexes.gainsEffect({ effectId: '97[3456]' }),
+      condition: Conditions.targetIsYou(),
+      run: (data, _) => {
+        data.gigaSlashCleaveDebuffId = _.effectId;
+        data.gigaSlashCleaveDebuffDuration = _.duration;
```
JLGarber, line 283: This should probably be `(data, matches)`, since we're actually using the parameter and not just skipping over it to `output`.

valarnin: Renamed variable.

panicstevenson:
Just an FYI: I had this call out the wrong mechanic and also sometimes not call out. I am mid-raid, so I can't really look into it right now.

valarnin:
I've only been able to test this against the first giga slash. Were you using the latest commit (e292382)?

panicstevenson:
Just got out of raid, so I will run this through the emulator with the latest. Sorry for the delay.

panicstevenson:
Yea, running the latest in the raid emulator, I'm getting the opposite callout; if it's right cleave, and my shadow is facing east, then it tells me "North Cleave". I assume you're not seeing the same?

valarnin:
It's possible I got the callouts backwards when I adjusted to account for going out on boss action instead of on debuff gain. If you flip the output objects for the two buffs does it appear correct?

panicstevenson:
Yep, the latest commit has both of the `alertText` as the same output.
I believe it should be:
```js
    {
      id: 'E10S Giga Slash Shadow Drop Right',
      alertText: (data, _, output) => {
        let ret = output.default;
        switch (data.gigaSlashCleaveDebuffId) {
        case '973':
          ret = output.westCleave;
          break;
        case '974':
          ret = output.eastCleave;
          break;
        case '975':
          ret = output.northCleave;
          break;
        case '976':
          ret = output.southCleave;
          break;
        }
      },
```
```js
    {
      id: 'E10S Giga Slash Shadow Drop Left',
      alertText: (data, _, output) => {
        let ret = output.default;
        switch (data.gigaSlashCleaveDebuffId) {
        case '973':
          ret = output.eastCleave;
          break;
        case '974':
          ret = output.westCleave;
          break;
        case '975':
          ret = output.southCleave;
          break;
        case '976':
          ret = output.northCleave;
          break;
        }
    },
```

valarnin:
Alright, this should be cleaned up and rebased to latest main to resolve merge conflicts.

ui/raidboss/data/05-shb/raid/e10s.js
```diff
@@ -185,9 +185,41 @@ export default {
...
+          break;
+        }
+        delete data.gigaSlashCleaveDebuffId;
+        delete data.gigaSlashCleaveDebuffDuration;
+        return ret();
+      },
       outputStrings: {
-        text: {
+        northCleave: {
+          en: 'North Cleave',
```
quisquous, line 211: Sorry for the second suggestion here, but I am a little bit worried about the default text being so similar to the proper text.  If I was used to hearing "west cleave" and then I heard "left cleave" I might think those are the same thing.  What about saying "north shadow" and "left cleave" to differentiate?  Or, am I overthinking this?

panicstevenson: Honestly, I would just put the cardinal direction. It’s not like you have to position them very well?

valarnin: So this?
```javascript
      outputStrings: {
        north: {
          en: 'North',
        },
        south: {
          en: 'South',
        },
        east: {
          en: 'East',
        },
        west: {
          en: 'West',
        },
        default: {
          en: 'Left Cleave',
          de: 'Linker Cleave',
          ko: '왼쪽 장판',
        },
      },
```

quisquous: Yeah, that seems good.  Sorry for the small quibble.  I'm just wary of having callouts that might be confusing.

valarnin: The `default` outputString might need to be renamed as well? I'm not sure if that's explanatory enough in the config screen. Maybe it should be `noShadow` instead?

valarnin: This is resolved. I also renamed `default` to `noShadow` for clarity.

panicstevenson: Should this just always call out left cleave/right cleave + your direction, like two separate triggers?
It feels weird if someone is doing callouts based off cactbot if they have "Right Cleave" for one, and "North" for another.

PR #2209, opened 2020-12-14
raidboss: Add E10S Head Markers

panicstevenson:
I wasn't able to find a correlation between the the shadow and the
player, other than the shadow has the same job as the player spawning
it. We can make due with some callouts for this, but it'd be nice to
have something more robust for groups running multiple of the same job.


ui/raidboss/data/05-shb/raid/e10s.js
```diff
@@ -226,9 +223,84 @@ export default {
...
         },
       },
+      run: (data) => data.clones = true,
+    },
+    {
+      // This checks your shadow's job against your job, since your shadow has
+      // the same job as you. If there's multiple of one job, or a shadow has
+      // a job of 0 (player died), then return '?' for the affected players.
+      id: 'E10S Shadow Of A Hero',
+      netRegex: NetRegexes.addedCombatantFull({ name: 'Shadow Of A Hero' }),
```
JLGarber, line 233: If we're already using `addedCombatantFull`, is there a particular reason not to use the `npcNameId` field instead? This would remove the need for translation here, and the name of the NPC is in the trigger ID already.

ui/raidboss/data/05-shb/raid/e10s.js
```diff
@@ -226,9 +223,84 @@ export default {
           cn: '影子放到外圈',
         },
       },
+      run: (data) => data.clones = true,
+    },
+    {
+      // This checks your shadow's job against your job, since your shadow has
+      // the same job as you. If there's multiple of one job, or a shadow has
+      // a job of 0 (player died), then return '?' for the affected players.
```
JLGarber, line 231: Do you have a sample of log lines with player IDs vs NPC IDs? I'm curious whether there's some relationship between player ID sort order and clone shadow spawn order. Or does the HP field possibly match with players? That might be more unique than jobs.

ui/raidboss/data/05-shb/raid/e10s.js
panicstevenson: How can I find the value for this?

ui/raidboss/data/05-shb/raid/e10s.js
panicstevenson: Yes, I will send you a sample. No, there is no relationship as far as I can tell between the player ID order and the shadow spawn order; shadows spawn in a different order pull to pull, even. The HP values also don't match, unfortunately. This would be a case for "Combatant added with a tether already on it."

ui/raidboss/data/05-shb/raid/e10s.js
JLGarber: It's field 9 in [`addCombatant` 03 log lines](https://github.com/quisquous/cactbot/blob/445188579e4d008f22eb6efb75922758d21fda70/resources/netregexes.js#L160). I typically prefer using this for any `addCombatant` triggers because it doesn't require translations.

ui/raidboss/data/05-shb/raid/e10s.js
JLGarber: Mmrf. Nothing useful that I can see. I'll take another look tomorrow just to be sure.

valarnin: That info is passed to the client but I don't think the FFXIV plugin captures it. The shadow's parentId is set to the player spawning it I believe. Here's an example network log line capturing it (player ID asterisk'd out):
```
252|2020-12-12T21:44:01.6250000-05:00|00000298|40013BD8|100*****|00000003|...trim a bunch of fields
03|2020-12-12T21:44:01.3660000-05:00|40013bd8|Shadow Of A Hero|18|50|0|0||9772|12397|14800000|14800000|10000|10000|0|0|118.3551|100.0423|0|-1.616889||b8b00852866a9304af9ddf4be79c63b8
```

valarnin: I don't know of a good way to verify, but it seems like you should be able to use `getCombatants` to determine this if all else fails. Something like this:
```javascript
callOverlayHandler({
    call: 'getCombatants',
    ids: [0x400*****],
    props: ['OwnerID']
}).then(a => console.log(a.combatants[0].OwnerID))
```

valarnin: One more follow up to note that raidemulator can't track the parent/owner ID, because it's not emitted in normal network log lines, so there's no way to test this case in raidemulator.

panicstevenson: Well, I can get combatants by name like we do in E6S, so an array of Shadow Of A Hero combatants, and then line up their owner ids.
I'll try to whip something up locally before raid today.

valarnin: I'm not sure if it's better to do one call by name, or 8 calls by ID.

panicstevenson: My test code doesn't feel too bad:
```js
    promise: async (data) => {
      const shadowOfAHeroLocaleNames = {
        en: 'Shadow Of A Hero',
      };

      const combatantData = await window.callOverlayHandler({
        call: 'getCombatants',
        names: [shadowOfAHeroLocaleNames[data.parserLang]],
        props: ['ownerId', 'id'],
      });

      if (combatantData === null || !combatantData.combatants || !combatantData.combatants.length)
        return;

      const shadowToOwnerMap = {};
      for (const combatant of combatantData.combatants) {
        if (combatant.id)
          shadowToOwnerMap[combatant.id] = data.party.nameFromId(combatant.ownerId);
      }

      data.shadowToOwnerMap = shadowToOwnerMap;
    },
```

Unless anything looks very wrong, I'll run with this in a couple hours.

valarnin: My only concern is how this is going to pan out with suppressSeconds. If there's any sort of weird timing issue vs when the first one's network log hits cactbot vs when all shadow clones are available in the combatants set.

panicstevenson: I didn't have suppressSeconds in the version I had for raid; it was delayed 2 seconds off the Shadow Cleave ability. However, it didn't work and there's not enough logging to tell me why, since everything just says "[Object object]" since I guess I thought console.log would work the same between the browser and ACT.

valarnin:
```javascript
console.log(JSON.stringify(<somevar>));
```

ui/raidboss/data/05-shb/raid/e10s.js
```diff
@@ -244,9 +241,84 @@ export default {
...
+        },
+        'unknown': {
+          en: '?',
+          de: '?',
+          fr: '?',
+          ja: '?',
+          cn: '?',
+          ko: '?',
+        },
+      },
```
MaikoTan, line 319: would you like to add JA translation as well?
which is the same as others, just simply "1", "2" and "3".

ui/raidboss/data/05-shb/raid/e10s.js
panicstevenson: Yep; I guess it will have to wait until tomorrow.

panicstevenson:
Moving this to draft in case any on-lookers are wanting to try it so they know it's basically in-progress.

panicstevenson:
@valarnin I got `[{},{},{},{},{},{},{},{}] ` as my combatant data for:
```js
        const shadowOfAHeroLocaleNames = {
          en: 'Shadow Of A Hero',
        };

        const combatantData = await window.callOverlayHandler({
          call: 'getCombatants',
          names: [shadowOfAHeroLocaleNames[data.parserLang]],
          props: ['ownerId', 'id'],
        });
 ```
 Any idea? I would imagine I'm doing something stupid. Looking at the other combatant data I got, without calling out to `props`, I got "OwnerID", and "ID"; so I'm assuming it's that. It also feels like I can exclude the line entirely. I assume I'm getting empty objects because there's nothing that matches those property names?

**Comment by valarnin** [comment]
Correct. Property names are case-sensitive and they're ProperCase in C# land, so the correct call would be:

```js
        const shadowOfAHeroLocaleNames = {
          en: 'Shadow Of A Hero',
        };

        const combatantData = await window.callOverlayHandler({
          call: 'getCombatants',
          names: [shadowOfAHeroLocaleNames[data.parserLang]],
          props: ['OwnerID', 'ID'],
        });
```

You should be able to verify this with any pet class since pets (e.g. SCH fairy) also have the OwnerID field set with a value.

quisquous:
This has worked well for me in the few times I've tried it.  Is this ready to go, or still a draft and more work to be done?

panicstevenson:
I am still going to try running it today with the OwnerID patch, but if I can't get that to work then maybe it makes sense to just land this as is and then continue working on that over the next weeks.

panicstevenson:
> {"CurrentWorldID":0,"WorldID":0,"WorldName":"","BNpcID":12397,"BNpcNameID":9772,"PartyType":0,"ID":1073758511,"OwnerID":0,"type":2,"Job":21,"Level":80,"Name":"Shadow Of A Hero","CurrentHP":14800000,"MaxHP":14800000,"CurrentMP":10000,"MaxMP":10000,"PosX":100.168457,"PosY":119.0716,"PosZ":0,"Heading":-3.13325167},{"CurrentWorldID":0,"WorldID":0,"WorldName":"","BNpcID":12397,"BNpcNameID":9772,"PartyType":0,"ID":1073758510,"OwnerID":0,"type":2,"Job":25,"Level":80,"Name":"Shadow Of A Hero","CurrentHP":14800000,"MaxHP":14800000,"CurrentMP":10000,"MaxMP":10000,"PosX":118.452751,"PosY":118.710243,"PosZ":0,"Heading":-2.313614},{"CurrentWorldID":0,"WorldID":0,"WorldName":"","BNpcID":12397,"BNpcNameID":9772,"PartyType":0,"ID":1073758509,"OwnerID":0,"type":2,"Job":31,"Level":80,"Name":"Shadow Of A Hero","CurrentHP":14800000,"MaxHP":14800000,"CurrentMP":10000,"MaxMP":10000,"PosX":82.6730652,"PosY":82.6591339,"PosZ":0,"Heading":0.736369848},

This is what I got from my run today. Seems relatively unhelpful?

ui/raidboss/data/05-shb/raid/e10s.js
```diff
@@ -244,9 +241,87 @@ export default {
...
```
quisquous, line 322:
```suggestion
      outputStrings: {
        '1': Outputs.num1,
        '2': Outputs.num2,
        '3': Outputs.num3,
        'unknown': Outputs.unknownTarget,
      },
```

quisquous: Plus the include at the top, sorry.

quisquous (commented):
It sure looks like `OwnerID` is not providing anything.  Maybe using the job is as good as it gets for now without additional network data parsing.

One other tiny comment, but otherwise I think this is ready to merge?

valarnin:
Yeah, looking at the properties on Combatant, it looks like ParentID isn't defined. I had assumed that OwnerID and ParentID are the same since they're at the same offset of the network packet, but I guess that isn't the case.

Legends0:
Worked for me in standard party.

panicstevenson:
Unfortunately due to some real life stuff, I wasn't able to make the changes before most people were raiding so I'll test this out tomorrow in my raid and make sure it's still good to go after committing suggestions.

ui/raidboss/data/05-shb/raid/e10s.js
panicstevenson: So, I was about to come back around and do this; however, since we're going to need the name for the below headMarker regex (do we?), should I leave this as is?

PR #2227, opened 2020-12-15
Raidboss: Matoya's Relict timelines should now function correctly

JLGarber:
Currently the timeline start parsing is not smart enough to figure out that `The` is sometimes elided. This change should restore functionality to the timeline. `Test_timeline` correctly started for every encounter I tried it on.


panicstevenson (approved):

MaikoTan:
😂Sorry I generate prototype timeline with JA client language, and then translate into EN, forgot that there are `the` in placenames

JLGarber:
Makes sense, no problem. Sorry none of us noticed during the initial review!

Issue #2161, opened 2020-12-11
raidemulator needs a lot of polish


quisquous:
There's a number of issues that I think prevent the raidemulator from being more useful than it is.  It's still got a lot of benefits over the old version, but I keep running into things that make it frustrating to use.

I think a first pass of debugging is that I want to see a log and just know the the trigger text that appeared for everybody.  The list of triggers are all triggers that have matched anything (regardless of `condition`), and it is very difficult (especially now with output strings) to click on a trigger and understand what it showed.  It would be nice if instead of saying `00:01 - General Provoke | 00:01` it would say `00:01 - General Provoke | Provoke: Tini Poutini | 00:01` and say the text in the list.  Seeing the values of data are okay but only needed for extreme edge case debugging scenarios.  I can quickly add a `console.log` for debugging if needed.

In priority order, these are improvements that would make my life a lot better:
* very hard to see trigger text from analysis (mentioned above)
* trigger analysis doesn't respect delay seconds
* timeline does not play back in real time (making it really hard to evaluate against video)
* skipping ahead in the timeline does not update the list of timers properly
* timeline triggers aren't handled in analysis
* no feedback when loading files (even in the console)

panicstevenson:
Additionally, `suppressSeconds` doesn't work as intended, and it looks as though all the `addedCombatant` lines are being parsed on log file load?
Adding a `addedCombatant` trigger with `suppressSeconds` defined will only find the first instance of the log and suppress all remaining addedCombatant logs that match.

valarnin:
I can probably take a look at some of these at least this weekend, not sure where you'd prefer I start or if you'd want all of these as one PR or split up.

valarnin:
> timeline does not play back in real time (making it really hard to evaluate against video)

@quisquous Can you elaborate on this? Playback should be realtime after clicking the play button.

panicstevenson:
> @quisquous Can you elaborate on this? Playback should be realtime after clicking the play button.

I think their point is that if you skip in 10 minutes into the fight on the scrollbar at the top, you won't skip ahead in the timeline, and it will keep playing from where it was (or maybe it resets, I can't remember off the top of my head)

quisquous:
Ah no, sorry.  I was saying that the playback was about 0.7x realtime speed.  I had a youtube video up and had to keep pausing it every second or two so that it didn't run ahead.

[...]

PR #2230, opened 2020-12-16
raidemulator: Add modal to show progress when importing encounters

valarnin:
This PR is to address one of the points on #2161.

There's only one real part of this that I'm not happy with, as mentioned on that issue, the exact mechanism for excluding web workers from running the `common.js` hooks doesn't really feel clean to me.

Here's a screenshot of exactly how the new modal looks for reference:

image


resources/common.js
```diff
@@ -104,8 +104,8 @@ export const Util = (() => {
 })();
 
 (function() {
-  // Exit early if running within Node
-  if (typeof location === 'undefined')
+  // Exit early if running within Node or within a web worker
+  if (typeof location === 'undefined' || (typeof WorkerLocation !== 'undefined' && location instanceof WorkerLocation))
```
quisquous, line 108: Hmm! I am not 100% sure offhand what the right way to do this.  I saw somebody else checking `WorkerGlobalScope ` as a way to do this: https://gist.github.com/miguelmota/a0c9aa9398507e606a6e

Also, I do wonder if `typeof location === 'undefined` should be `typeof window === 'undefined'` at the very least to match `EventBus.js` below as a consistent way to detect node vs browser.

ui/raidboss/emulator/data/NetworkLogConverterWorker.js
```diff
@@ -0,0 +1,68 @@
...
+
+onmessage = async (msg) => {
+  const logConverter = new NetworkLogConverter();
+  const localLogHandler = new LogEventHandler();
+  const repo = new LogRepository();
+
+  // Listen for LogEventHandler to dispatch fights and persist them
+  localLogHandler.on('fight', async (day, zoneId, zoneName, lines) => {
+    const enc = new Encounter(day, zoneId, zoneName, lines);
+    if (!(enc.firstPlayerAbility > 0 && enc.firstEnemyAbility > 0))
```
quisquous, line 15: It looks like this is the third time this pattern is repeated, so could this be an `enc.shouldPersistFight()` helper function?

ui/raidboss/emulator/data/NetworkLogConverterWorker.js
```diff
@@ -0,0 +1,68 @@
...
+    });
+  });
+
+  // Convert the message manually due to memory issues with extremely large files
+  const decoder = new TextDecoder('UTF-8');
+  let buf = new Uint8Array(msg.data);
+  let i = 0;
+  let lines = [];
+  let lineCount = 0;
+  for (let a = i; i < buf.length && i !== -1; a = i) {
```
quisquous, line 30: Sorry, too many single letter variables here.  What are `a` and `i`?

ui/raidboss/emulator/data/NetworkLogConverter.js
```diff
@@ -17,9 +11,9 @@ export default class NetworkLogConverter {
     return ret;
   }
 
-  async convertLines(lines) {
+  async convertLines(lines, repo) {
     this.Combatants = {};
-    const repo = new LogRepository();
+    repo = repo || new LogRepository();
```
quisquous, line 16: To make this more readable to me, can the one other caller pass in a new repo too, rather than having this be conditionally implicit?

resources/common.js
valarnin: For the `EventBus.js` case, it may make more sense to change it to use [globalThis](https://caniuse.com/?search=globalThis) instead, as such:

```js
scope = scope !== undefined ? scope : globalThis;
```

It should also be fine to change the `common.js` check to:

```js
  // Exit early if not running within a `Window` scope
  if (typeof Window === 'undefined' || !globalThis instanceof Window)
```

Thoughts?

ui/raidboss/emulator/data/NetworkLogConverterWorker.js
valarnin: Agreed. I'll make this adjustment and update the other references.

ui/raidboss/emulator/data/NetworkLogConverterWorker.js
valarnin: `a` is the advancing index, `i` is the current index. I guess these could be renamed to `currentOffset` and `nextOffset` maybe?

ui/raidboss/emulator/data/NetworkLogConverter.js
valarnin: Yeah, I can adjust the other caller to pass in `repo` as well. This was my aversion to passing around large objects in other languages coming back to bite me 😅.

resources/common.js
quisquous: Seems reasonable to me! I don't think I am confident enough here to know what the "right" way to check this is.

ui/raidboss/emulator/data/NetworkLogConverterWorker.js
```diff
@@ -0,0 +1,70 @@
...
+      ++lineCount;
+      lines.push(line);
+    }
+
+    if (lines.length >= 1000) {
+      lines = await logConverter.convertLines(lines, repo);
+      localLogHandler.parseLogs(lines);
+      postMessage({
+        type: 'progress',
+        lines: lineCount,
```
panicstevenson, line 45: Are lineCount and lines.length the same?

valarnin: No, `lineCount` has the total number of processed lines, whereas the `lines` array only has up to the previous 1000 (for memory reasons when loading large files). The `lines` array gets cleared on line 49 and line 55 (though line 55's clear might be unneeded).

panicstevenson: Ah got it! My misunderstanding.

panicstevenson:
I just pulled this change in locally and it looks great! My only thought would be that the `Done` button is a little weird? I would imagine it's instead a cancel button and upon finishing, the modal would just close automatically. However, that's just a personal nit, and can probably be persuaded otherwise.

valarnin:
> upon finishing, the modal would just close automatically

I wanted to avoid the "did it actually finish importing?" question. I guess maybe it could auto-close after 10 seconds or something.

Canceling the import would be a bit messy code-wise so I didn't put that in. Right now if you click out of the modal it'll close it but keep importing in the background, and then when it's done it'll do the refresh of the encounters tab at the bottom which can be a bit jarring.

quisquous:
It seems fairly obvious to me when imports are done, because they show up in the list.  It's also a modal dialog, so if it's up then it's not done, and if it is then it's not.  I think it'd be fine to just auto-close, personally.

valarnin:
One other thing to consider now that there's a progress modal and such, is that the modal could list all the encounters and allow you to select the ones you want to keep.

quisquous:
> One other thing to consider now that there's a progress modal and such, is that the modal could list all the encounters and allow you to select the ones you want to keep.

Maybe a good thing to do as a follow-up feature?

valarnin:
I decided to cut down the middle for the modal close question, and have it auto-close after 5 seconds. I'll handle the encounter selection as a separate follow up feature when I get the time to.

Is there anything outstanding on this PR that needs action?

ui/raidboss/emulator/data/NetworkLogConverter.js
quisquous: Can this line be removed, then?

quisquous:
When I use this patch to import [this log](https://github.com/quisquous/cactbot/files/5948405/titan_unreal.zip), I get this:

image

It seems to hang forever there.  No error message in the console.  If I reload, it hasn't secretly imported it.  It imports properly in the version of raidemulator on main.  I haven't had time to try other logs; this was the first one I tried, so it could be specific to this.

valarnin:
This is importing fine for me (sorry for the big screenshot, forgot to resize):

image

You may need to restart the webpack server to load it, as I renamed the added entry for the worker to camelCase.

quisquous:
> You may need to restart the webpack server to load it, as I renamed the added entry for the worker to camelCase.

Hmm, even restarting the webpack server doesn't change anything for me.  It still does what I posted for any log I try.  Do you have any debugging suggestions?

valarnin:
Check network tab to see if something got blocked maybe?

ui/raidboss/raidemulator.js
```diff
@@ -52,44 +51,29 @@ const Options = {
...
     emulator = new RaidEmulator(Options);
     progressBar = new ProgressBar(emulator);
     persistor = new Persistor();
-    logEventHandler = new LogEventHandler();
     encounterTab = new EncounterTab(persistor);
     emulatedPartyInfo = new EmulatedPartyInfo(emulator);
     emulatedMap = new EmulatedMap(emulator);
     emulatedWebSocket = new RaidEmulatorOverlayApiHook(emulator);
-    logConverter = new NetworkLogConverter();
+    logConverterWorker = new Worker('../../dist/raidemulator_worker.bundle.js');
```
quisquous, line 69: Here's the issue.  You named it `raidemulatorWorker` in `webpack.config.js` and it's trying to load `raidemulator_worker` instead here.

valarnin: I renamed it in the cjs file to camelCase because the linter didn't like the underscore, and forgot to update it in `raidemulator.js`. Apparently the old file was cached in my browser. Just force pushed the fix.

quisquous: Sorry, it's hard to tell the fix in the force push.  It looks like there's still a mismatch between `raidemulatorWorker` in the config and `raidemulator_worker` here.  I'll fix it and the conflict.

quisquous, line 69:
```suggestion
    logConverterWorker = new Worker('../../dist/raidemulatorWorker.bundle.js');
```

PR #2231, opened 2020-12-17
Util: Test_timeline terminal color-coding

JLGarber:
(This is currently a proof-of-concept only! It needs lots of refinement.)

It occurred to me that it would be nice to color-code certain lines in `test_timeline` that drift by more than a second. The technique I used here is obviously not what will make it into the final version, but at least we have a baseline for what we want. Note that this currently works on my (Windows 10) machine in Powershell, but after some brief testing from @panicstevenson it looks like it does *not* work in Git Bash.

Todo:

1. Make the conditions/string formatting not-ugly. We have plenty of room for improvement there.
2. Make the thresholds configurable at the command line. This should be relatively trivial.
3. 
Test_timeline_colors

### DIFF
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -9,11 +9,13 @@ import re
 import fflogs
 import encounter_tools as e_tools
 
+
 class tcolor:
-    WARN = '\033[33m'
-    FAIL = '\033[91m'
-    SYNCWARN = '\033[93m'
-    END = '\033[0m'
+    WARN = "\033[33m"
+    FAIL = "\033[91m"
+    SYNCWARN = "\033[93m"
+    END = "\033[0m"
+
 
 def load_timeline(timeline):
     """Loads a timeline file into a list of entry dicts"""
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -9,6 +9,11 @@ import re
 import fflogs
 import encounter_tools as e_tools
 
+class tcolor:
+    WARN = '\033[33m'
+    FAIL = '\033[91m'
+    SYNCWARN = '\033[93m'
+    END = '\033[0m'
 
 def load_timeline(timeline):
     """Loads a timeline file into a list of entry dicts"""
@@ -281,11 +286,11 @@ def test_match(event, entry):
 
 
 def drift_fail(entry_text):
-    return "\033[91m" + entry_text + "\033[0m"
+    return tcolor.FAIL + entry_text + tcolor.END
 
 
 def drift_warn(entry_text):
-    return "\033[33m" + entry_text + "\033[0m"
+    return tcolor.WARN + entry_text + tcolor.END
 
 
 def check_event(event, timelist, state):
--- a/util/test_timeline.py
+++ b/util/test_timeline.py
@@ -313,12 +313,12 @@ def check_event(event, timelist, state):
             # Check the timeline drift for anomolous timings
             drift = entry["time"] - timeline_position
             entry_text = "{:.3f}: Matched entry: {} {} ({:+.3f}s)".format(
-                    timeline_position, entry["time"], entry["label"], drift
-                )
+                timeline_position, entry["time"], entry["label"], drift
+            )
             if abs(drift) > 1:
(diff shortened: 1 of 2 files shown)


util/test_timeline.py
```diff
@@ -312,11 +312,18 @@ def check_event(event, timelist, state):
...
             )
+            if abs(drift) > 1:
+                print("\033[91m" + entry_text + "\033[0m")
+            elif 1 > abs(drift) > 0.2:
+                print("\033[33m" + entry_text + "\033[0m")
+            else:
+                print(entry_text)
+
+            if time_progress_seconds > 30:
+                print("    Warning: {:.3f}s since last sync".format(time_progress_seconds))
```
quisquous, line 326:
```suggestion
```

util/test_timeline.py
```diff
@@ -312,11 +312,18 @@ def check_event(event, timelist, state):
...
             drift = entry["time"] - timeline_position
-            print(
-                "{:.3f}: Matched entry: {} {} ({:+.3f}s)".format(
-                    timeline_position, entry["time"], entry["label"], drift
-                )
+            entry_text = "{:.3f}: Matched entry: {} {} ({:+.3f}s)".format(
+                timeline_position, entry["time"], entry["label"], drift
             )
+            if abs(drift) > 1:
+                print("\033[91m" + entry_text + "\033[0m")
```
quisquous, line 319: Can I suggest `print_warning` or `print_error` helper functions for these? Or something along these lines?

JLGarber: That seems reasonable, yes.

JLGarber:
Hmm, as long as we're on this anyway, what would we think of assigning a color to:
```python
            if time_progress_seconds > 30:
                print("    Warning: {:.3f}s since last sync".format(time_progress_seconds))
```

quisquous:
> Hmm, as long as we're on this anyway, what would we think of assigning a color to:
> 
> ```python
>             if time_progress_seconds > 30:
>                 print("    Warning: {:.3f}s since last sync".format(time_progress_seconds))
> ```

I think yes color, but I would also add that this should only print out in cases where there are timeline syncs within those thirty seconds (because I think there are some rare timelines that do that).

MaikoTan:
What about create a class for color output, not just writing color code everywhere.
e.g.
`ColorPrinter.print("warning: xxxxxx", color=ColorPrinter.YELLOW)`

JLGarber:
> What about create a class for color output, not just writing color code everywhere.

That's reasonable too! I'll look at that after raid time.

JLGarber:
(This latest commit isn't intended to be a complete reworking to use class-based colors, just dropping it in for now because I'm frazzled after raid.)

panicstevenson:
> but after some brief testing from @panicstevenson it looks like it does _not_ work in Git Bash.

I don't know if something changed, but the latest commit works for me.

Additionally, when doing something that has multiple phases, the phases will turn red, such as the second half of Emerald Weapon Ex which will have red text for the initial sync.

JLGarber:
I'm not 100% on this fix for jumps, but It Seems To Me to make sense to have it be user-configurable as opposed to hardcoded?

JLGarber:
> I think yes color, but I would also add that this should only print out in cases where there are timeline syncs within those thirty seconds (because I think there are some rare timelines that do that).

I'll have to think through this one a little more on how best to implement. I'm sure it's possible, and I *think* the data is there already? Need to fully grasp what `check_event` is doing.

JLGarber:
Alright, I think this is probably as close to ready as I'll want it to be. I'd like to improve the "more than 30s" text, but it's a little more complex than I had thought, and modifying it is probably beyond the scope I wanted for this. We can have a look at that another time.

quisquous:
> Alright, I think this is probably as close to ready as I'll want it to be. I'd like to improve the "more than 30s" text, but it's a little more complex than I had thought, and modifying it is probably beyond the scope I wanted for this. We can have a look at that another time.

Ah yeah, sorry.  I meant to come back and say "sorry for the feature request that probably should be in a follow-up".  Glad that message was telepathically received.

PR #2235, opened 2020-12-17
raidemulator: Fix issues with getCombatants and issue with NetworkAbility line parsing

valarnin:
- Fix event 0x15 not mapping x/y/z for combatant state
- Fix issues with RaidEmulatorOverlayApiHook's getCombatants override vs C# property casing
- Fix RaidEmulatorOverlayApiHook's getCombatants override not having the correct timestamp offset when doing initial encounter analysis
- Fix issue with suppressSeconds in RaidEmulator
- Fix issue with slow PCs causing delay in playback, causing desync vs videos

### DIFF
--- a/ui/raidboss/emulator/data/AnalyzedEncounter.js
+++ b/ui/raidboss/emulator/data/AnalyzedEncounter.js
@@ -95,6 +95,8 @@ export default class AnalyzedEncounter extends EventBus {
     };
 
     for (const log of this.encounter.logLines) {
+      await this.dispatch('analyzeLine', log);
+
       if (this.encounter.combatantTracker.combatants[ID].hasState(log.timestamp)) {
         popupText.OnPlayerChange({
           detail: {
--- a/ui/raidboss/emulator/data/CombatantState.js
+++ b/ui/raidboss/emulator/data/CombatantState.js
@@ -23,4 +23,17 @@ export default class CombatantState {
         Number(props.MP) || this.MP,
         Number(props.maxMP) || this.maxMP);
   }
+
+  toPluginState() {
+    return {
+      PosX: this.posX,
+      PosY: this.posY,
+      PosZ: this.posZ,
+      Heading: this.heading,
+      CurrentHP: this.HP,
+      MaxHP: this.maxHP,
+      CurrentMP: this.MP,
+      MaxMP: this.maxMP,
+    };
+  }
 }
--- a/ui/raidboss/emulator/data/CombatantTracker.js
+++ b/ui/raidboss/emulator/data/CombatantTracker.js
@@ -20,15 +20,12 @@ export default class CombatantTracker {
   }
 
   initialize(logLines) {
-    const keyedLogLines = {};
     // First pass: Get list of combatants, figure out where they
-    // start at if possible, build our keyed log lines
+    // start at if possible
     for (let i = 0; i < logLines.length; ++i) {
       const line = logLines[i];
       this.firstTimestamp = Math.min(this.firstTimestamp, line.timestamp);
       this.lastTimestamp = Math.max(this.lastTimestamp, line.timestamp);
-
-      keyedLogLines[line.timestamp + '_' + i] = line;
 
       switch (line.hexEvent) {
       // Source/target events
@@ -44,14 +41,15 @@ export default class CombatantTracker {
         this.addCombatantFromLine(line);
         this.addCombatantFromTargetLine(line);
         break;
+      case '00':
+        // For 00 chat lines, don't ever generate combatant data from them
+        break;
 
       default:
         this.addCombatantFromLine(line);
         break;
       }
     }
-
-    const sortedTimestamps = Object.keys(keyedLogLines).sort();
 
     // Between passes: Create our initial combatant states
     for (const id in this.initialStates) {
@@ -71,8 +69,8 @@ export default class CombatantTracker {
 
     // Second pass: Analyze combatant information for tracking
     const eventTracker = {};
-    for (let i = 0; i < sortedTimestamps.length; ++i) {
-      const line = keyedLogLines[sortedTimestamps[i]];
+    for (let i = 0; i < logLines.length; ++i) {
+      const line = logLines[i];
       let state = this.extractStateFromLine(line);
       if (state) {
         eventTracker[line.id] = eventTracker[line.id] || 0;
--- a/ui/raidboss/emulator/data/RaidEmulator.js
+++ b/ui/raidboss/emulator/data/RaidEmulator.js
@@ -11,12 +11,14 @@ export default class RaidEmulator extends EventBus {
     this.currentTimestamp = null;
     this.currentLogLineIndex = null;
     this.lastLogTimestamp = null;
+    this.internalTimestampTracker = null;
   }
   addEncounter(encounter) {
     this.encounters.push(encounter);
   }
   setCurrent(index) {
     this.currentEncounter = new AnalyzedEncounter(this.options, this.encounters[index], this);
+    this.dispatch('preCurrentEncounterChanged', this.currentEncounter);
     this.currentEncounter.analyze(this.popupText).then(() => {
       this.dispatch('currentEncounterChanged', this.currentEncounter);
     });
@@ -44,6 +46,7 @@ export default class RaidEmulator extends EventBus {
     this.currentTimestamp = this.currentTimestamp ||
       this.currentEncounter.encounter.logLines[firstIndex].timestamp;
     this.currentLogLineIndex = this.currentLogLineIndex || firstIndex - 1;
+    this.internalTimestampTracker = Date.now();
     // Use setInterval since it should account for differences in execution time automagically
     this.playing = window.setInterval(this.tick.bind(this), RaidEmulator.playbackSpeed);
     this.dispatch('play');
@@ -52,6 +55,7 @@ export default class RaidEmulator extends EventBus {
 
   pause() {
     window.clearInterval(this.playing);
+    this.internalTimestampTracker = null;
     this.playing = null;
     this.dispatch('pause');
     return true;
@@ -61,6 +65,8 @@ export default class RaidEmulator extends EventBus {
     await this.dispatch('preSeek', time);
     this.currentLogLineIndex = -1;
     let logs = [];
+    const playing = this.playing;
+    this.playing = null;
     for (let i = this.currentLogLineIndex + 1;
       i < this.currentEncounter.encounter.logLines.length;
       ++i) {
@@ -89,6 +95,7 @@ export default class RaidEmulator extends EventBus {
 
     await this.dispatch('postSeek', time);
     await this.dispatch('tick', this.currentTimestamp, this.lastLogTimestamp);
+    this.playing = playing;
   }
 
   async tick() {
@@ -96,8 +103,11 @@ export default class RaidEmulator extends EventBus {
       this.pause();
       return;
     }
+    if (this.playing === null)
+      return;
     const logs = [];
-    const lastTimestamp = this.currentTimestamp + RaidEmulator.playbackSpeed;
+    const timeDiff = Date.now() - this.internalTimestampTracker;
+    const lastTimestamp = this.currentTimestamp + timeDiff;
     for (let i = this.currentLogLineIndex + 1;
       i < this.currentEncounter.encounter.logLines.length;
       ++i) {
@@ -109,7 +119,8 @@ export default class RaidEmulator extends EventBus {
       }
       break;
     }
-    this.currentTimestamp += RaidEmulator.playbackSpeed;
+    this.currentTimestamp += timeDiff;
+    this.internalTimestampTracker += timeDiff;
     if (logs.length)
       await this.dispatch('emitLogs', { logs: logs });
 
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.js
@@ -50,10 +50,11 @@ export class LineEvent0x15 extends LineEvent {
     this.sourceMp = parts[36 + offset];
     this.sourceMaxMp = parts[37 + offset];
 
-    this.sourceX = parts[40 + offset];
-    this.sourceY = parts[41 + offset];
-    this.sourceZ = parts[42 + offset];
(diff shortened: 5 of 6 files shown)


panicstevenson:
So, I just ran this against my E9S Summon branch and my E9S logs:
I still see the clones in the wrong position, and the boss doesn't have a BNpcID field, which I'm assuming is expected.
Do I need to do something special here, like run the fight with Network Debugging enabled?

valarnin:
BNpcID isn't included in normal ACT logs, so it can't be included in raidemulator. I think that field is the actual model ID or something?

The analyzed version of the encounter is stored in the IndexedDB table, so you'll need to re-import to fix the positioning/heading issue. I would like to make the `Reparse` button fix this but when I tried I ran into a weird issue with it dropping some lines, so put that on the backburner for now.

JLGarber:
> BNpcID isn't included in normal ACT logs, so it can't be included in raidemulator. I think that field is the actual model ID or something?

To be clear, this means that *only* the line fields in the parsed ACT log are available via `getCombatants`?

(As for "why BNPCID?", it, and NPCBaseID, are language-agnostic and will *usually* be included in the network data even if the entity name is not, for certain log lines. I think panic was wondering whether that would be possible here.)

valarnin:
Looking at your e9s Summon PR, you're using BNpcID to determine the real one. You could instead check PosX and PosY. Round them and they should both equal 100 for the boss:

```js
        const boss = bossData.combatants.filter((boss) => Math.round(boss.PosX) === 100 && Math.round(boss.PosY) === 100)[0];
```

image

panicstevenson:
> Looking at your e9s Summon PR, you're using BNpcID to determine the real one. You could instead check PosX and PosY. Round them and they should both equal 100 for the boss:

I can settle on using the MaxHP too, since all the clones have 44, and the boss obviously has substantially more than that; BNpcID just felt the most "official"

valarnin:
> I think panic was wondering whether that would be possible here.

It's possible more long-term as long as the logs were generated with debug network capture enabled, but actually pulling that information from the 252 lines is going to take a lot more effort. I'm not even sure if I should keep the 252 lines around in a parsed line scope, given how much that would bloat the encounter data.

valarnin:
@JLGarber Looking at this further, I think I'm a bit confused. Can you or someone else confirm something? There are two IDs included in `03` lines that might correspond to `BNpcID`, if someone can confirm which if either maps to the actual `getCombatants` return data, I can pull that info into raidemulator as well.

JLGarber:
[Sure, it's field 9](https://github.com/quisquous/cactbot/blob/0b47b9f42c613bf53e0aa6a3c0d677244b886a96/resources/netregexes.js#L150). For a line like `03|2020-12-17T01:50:35.4890000-06:00|40006430|Fatebreaker|0|1|0|0||9707|9020|44|44|0|10000|0|0|100|91|0|-4.792213E-05||cd21237b1278fd2f77352afc9aae6e07`, it would be 9707.


(An example of why it matters is something like `03|2020-12-17T01:50:35.3980000-06:00|4000641d||0|1|0|0||108|9020|44|44|0|10000|0|0|100|100|2.384186E-07|3.141593||672cb04d8d25f8a4bcc828153d528ed1`, where there's no name field.)

valarnin:
Alright, thanks.

Looking at the fields defined in `getCombatants`, https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/EventSources/MiniParseEventSource.cs#L40-L62

Comparing against the list of properties currently being tracked in https://github.com/quisquous/cactbot/blob/65e2637b97ba8e86b12e88ff8ac21c3ca60d8438/ui/raidboss/emulator/data/CombatantState.js#L27-L38

Currently the following fields aren't being tracked to emulate this call perfectly:
- BNpcID, field 9 of event 03 as you stated, I'll fix this
- ID, tracked elsewhere but not included in the emulated response, I'll fix this
- Name, tracked elsewhere but not included in the emulated response, I'll fix this
- Job, tracked elsewhere but not included in the emulated response, I'll fix this
- Level, tracked elsewhere but not included in the emulated response, I'll fix this
- OwnerID, maybe event 03 field 6?
- CurrentWorldID, maybe event 03 field 7?
- WorldID, maybe event 03 field 7?
- WorldName, maybe event 03 field 8?
- BNpcNameID, maybe event 03 field 9 or 10?
- PartyType, not in the logs at all AFAIK, derived completely from memory data?
- type, I have no clue what this is supposed to map to?

JLGarber:
Actually, looking at it again, I think field 9 is `BNPCNameID`, while `BNpcID` is field 10. `type` is the line type, the 2/3-digit field  at index 0. Probably safe to ignore it. Your other assumptions look right, I think.

valarnin:
`WorldID` vs `CurrentWorldID` is almost definitely wrong but I'm not sure which is in the 03 event and where the other would be. It's probably safe to map them both over to event 03 field 7, with a comment that one of them isn't right? I can't envision a trigger that would use that property but I want to at least have something in place even if it's just a placeholder.

`type` in this case is some part of the Combatant structure in ACT, not a line event type.

valarnin:
I've added tracking for all of the properties mentioned before, but I'm running into one more issue, related to the way encounters are being split. Right now, raidemulator is following the logic here to determine if an encounter ends and a new one should begin:
https://github.com/quisquous/cactbot/blob/65e2637b97ba8e86b12e88ff8ac21c3ca60d8438/ui/raidboss/emulator/data/LogEventHandler.js#L41-L64

The problem is the `wipeRegex` which checks against `NetRegexes.network6d({ command: '40000010' })` which is the `Fade in` command. This command occurs after all of the non-boss actors are added via `03` events, which means that those log lines aren't included in the encounter (they end up in the previous encounter).

The Log Guide mentions changing to `Fade out` (`40000005`) but that command doesn't line up well either. Instead I'm seeing a `8000000D` command which seems to be at the exact correct position in the logs for where I'd like to split them, but doesn't seem to be documented as to what this command is.

I'd like to switch to the `8000000D` command, since that seems to line up correctly, but I'm pretty sure that's not going to work for older content. So I'd have to, I guess, somehow check if the encounter's first line is a `8000000D` command, and then skip the `40000010` command?

The other piece of relevant code is this line, which is doing some sorting. This is to work around log lines being out-of-order according to their timestamp, so the other option is to remove this line but that's going to cause all sorts of other issues with the getCombatant-related stuff since data might not be accurate to what it would be in-game.
https://github.com/quisquous/cactbot/blob/65e2637b97ba8e86b12e88ff8ac21c3ca60d8438/ui/raidboss/emulator/data/NetworkLogConverter.js#L30

quisquous:
Re: splitting.  What's wrong with the fade out network message? That should be before all of the new combatants get added again.

Re: sorting.  I don't think you should re-sort log lines.  My understanding is that ignoring timestamp, the order of logs in the network log file are the order that cactbot sees them in.  The lines that don't have timestamps or appear out of order are ones that come from the in game log, and always end in `.000`.  Are there other lines that are out of order?  What problems are you addressing by sorting?

valarnin:
The reason for sorting is for the combatant tracker portion. I added the sorting to fix the HP/MP bars bouncing up and down randomly originally, but I think that also got fixed by another change to the way I'm updating the combatant state (basically discarding state updates if they're out of order).

quisquous:
Re: sorting.  I guess I'm just really surprised this is happening.  Can you give some examples?

valarnin:
> Re: sorting. I guess I'm just really surprised this is happening. Can you give some examples?

Well it's a bit late at this point, but the readme for emulator specifically calls this out and I completely forgot about it:

- Lines are sorted based on timestamp and then index, to account for chat lines (`00`) being out of order from network lines

I guess this could be accounted for by not using the timestamp from 00 lines for anything, but that might complicate other things.

valarnin:
The sorting and everything else should be addressed for this PR now.

There's one more outstanding issue I noticed recently that I'd like to address for this PR, which is that the timeline display seems to be desync'd for e12s. I'm looking into that now.

valarnin:
This line specifically is causing the problem:

https://github.com/quisquous/cactbot/blob/1bb55da59b2c0dd5c32920bb5c7a812a7a80e608/ui/raidboss/timeline.js#L407

I can work around it to fix sync when seeking/changing perspectives, which partially fixes the issue, but it's still going to cause problems when seeking on fights with a jump in the timeline.

Realistically, the only way to fully address the issue is to modify it to something like:

```js
const newTimebase = this.getTimeBaseForOffset(fightNow);
```

So that it can be overridden by raidemulator to use the log line's timestamp in replacement of `new Date`.

quisquous:
> * Lines are sorted based on timestamp and then index, to account for chat lines (`00`) being out of order from network lines
> 
> I guess this could be accounted for by not using the timestamp from 00 lines for anything, but that might complicate other things.

Yeah, I think this this is the correct way to handle it.  The lines themselves are not misordered, but the times appear in the past because the `00` lines have `000` milliseconds always.  I treat those lines when analyzing log files as having an invalid time and using the time of any previous log line.

valarnin:
I excluded the `00` lines from combatant state tracking and removed sorting in the previous commit, everything's parsing and running as expected with no jittering or anything, so that's fine.

The outstanding issue for this PR from my side I detailed in this comment (https://github.com/quisquous/cactbot/pull/2235#issuecomment-775399889). Given that this is an issue with the current version in `main`, and given that it requires changing non-emulator code, it should probably be done as a separate PR.

ui/raidboss/emulator/overrides/StubbedPopupText.js
```diff
@@ -7,6 +7,22 @@ export default class StubbedPopupText extends PopupText {
 
   // Stubbed, we don't want overlay hooks
   HookOverlays() { }
+
+  // Use the current emulator timestamp unless we're using an older timestamp
```
quisquous, line 11: Sorry for the belated response, but I'm not sure I understand what's going on here.  Can you explain the different cases, and where 500 comes from?

I think this is really my only question here.

valarnin: In the `PopupText` class, `OnTriggerInternal` creates a `new Date` to pass in to `_onTriggerInternalCheckSuppressed` as well as to `_onTriggerInternalGetHelper`. As the playback in raidemulator uses the original log timestamps, that'll never line up with the `new Date` timestamp values.

https://github.com/quisquous/cactbot/blob/37be1bc0be8c67c4dea88dc449287c759bd6c968/ui/raidboss/popup-text.js#L624-L637

The override for `_onTriggerInternalCheckSuppressed` is for the check to see if a trigger is suppressed specifically, whereas the override for `_onTriggerInternalGetHelper` is for the timestamp to use for suppressing a trigger.

Without these overrides, while realtime playback from the start of the fight works fine, any seeking will cause the entire suppressed logic to fail since the timestamps passed are only a few milliseconds apart.

The ternary operator in `_onTriggerInternalCheckSuppressed` is to account for the difference between seeking (timestamp less than 500ms apart, to allow for slower PCs) vs real-time playback. It might make more sense to expose an `emulator.currentState` property that indicates `paused` vs `playing` vs `seeking` and key off of that instead.

Also, sorry for the 2-day delay in response here, I didn't get an email notification for this one.

quisquous: I think explicitly checking for state vs a heuristic based on time is strictly better, both in terms of readability but also avoiding races.

However, even with that switch, the fact that there needs to be a conditional is very suspicious to me.   Why is it not using the log timestamp at all times?

valarnin: Are you suggesting this?

```js
  _onTriggerInternalCheckSuppressed(trigger, when) {
    when = window.raidEmulator.emulator.currentEncounter.encounter.logLines[
        window.raidEmulator.emulator.currentLogLineIndex
      ].timestamp;
    return super._onTriggerInternalCheckSuppressed(trigger, when);
  }
```

There's still some potential for issues here due to the way seek batches log lines, but that potential exists with the conditional as well (basically if there's a situation where two log lines 100 lines or less apart could exceed the suppress timeout of a trigger).

quisquous: Sorry, I don't know that I have a clear suggestion here, other than saying having an if statement like that makes me both know that I really don't understand how this system is working and wondering if there's some different way that the emulator should be considering time.

Glancing at the code, it seems like part of the issue is that `OnTriggerInternal` is using the current time.  If it used a passed in time that the emulator provided, then the emulator could use a fake time throughout that it controlled.

Maybe trigger suppression needs to be fixed more thoroughly, maybe in it's own PR rather than here? Suppression is also broken during analysis, but maybe that's a totally different problem.

valarnin: > Sorry, I don't know that I have a clear suggestion here, other than saying having an if statement like that makes me both know that I really don't understand how this system is working and wondering if there's some different way that the emulator should be considering time.

The emulator always considers time according to the log line's timestamp. There's never an invocation of `new Date` that doesn't pass in a timestamp aside from this specific one for the suppress workaround.

> Glancing at the code, it seems like part of the issue is that OnTriggerInternal is using the current time. If it used a passed in time that the emulator provided, then the emulator could use a fake time throughout that it controlled.

This is the exact issue that's causing problems with the suppress code here. Realistically I'd like to modify this specific segment of code in PopupText:

https://github.com/quisquous/cactbot/blob/169e11e4a02322e3b59d718ed9e23678b352db26/ui/raidboss/popup-text.js#L624-L628

To instead read something like:

```js
  getTimestamp(trigger, matches) {
    // check matches to see if the timestamp was matched, if so return the timestamp for that
    // otherwise return `+new Date()`
  }

  OnTriggerInternal(trigger, matches) {
    const now = this.getTimestamp(trigger, matches);

    if (this._onTriggerInternalCheckSuppressed(trigger, now))
      return;
```

Then the new function could be overrode in emulator.

> Maybe trigger suppression needs to be fixed more thoroughly, maybe in it's own PR rather than here? Suppression is also broken during analysis, but maybe that's a totally different problem.

I agree that if this requires changes to non-emulator code, it should be fixed in its own PR. If that's the case, I can revert the change for suppress for now.

valarnin:
I have reverted the Suppress fixes as implied in the previous conversation. Once this PR is merged, I'll start a new PR with the fix for that (with accompanying PopupText changes).

@quisquous I believe that was the only thing outstanding for this PR?

PR #2236, opened 2020-12-17
Raidboss: E11S timeline fixes and QoL

JLGarber:
Some buddies were running E11S last night and noted that they had some small issues with the Cycle portion of the timeline. The general feeling was that there was just too much information in the timeline, and that not all of it was useful.

### Fixes
1. (Temporary only, to be properly fixed when there's time.) I commented out the Solemn Charge syncs at 471. Unfortunately, the possibilities for each overlap here, causing bad timeline jumpiness. This should be resolved by figuring out the actual IDs for each ability and then doing a full match instead of the compact format used here. (That is, `sync /:Fatebreaker:(5659|5960):/` as opposed to `sync /:Fatebreaker:56(59|60):/`, and similarly for other possible collision points.)
2.  Holy Cycle closing jumps. This is a *bad* mistake and I'm not sure how it didn't turn up when I tested this before submitting. Sorry about that!

### Quality of Life
I removed the "Is the enrage happening here?" Burnished Glory from all the lookahead blocks on the cycles. Players who are at this point should *know* whether they are in cycle 3 or not. I also removed Floating Fetters, since it's not terribly relevant compared to everything else that's happening.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e11s.txt
+++ b/ui/raidboss/data/05-shb/raid/e11s.txt
@@ -199,8 +199,8 @@ hideall "--sync--"
 
 828.2 "--center--" sync /:Fatebreaker:5908:/ window 30,30
 835.0 "Burnished Glory?"
-837.2 "Cycle Of Faith" sync /:Fatebreaker:5692:/
-837.2 "--sync--" sync /:Fatebreaker:568A:/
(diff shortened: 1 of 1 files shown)


Jaehyuk-Lee:
Ah, yes I had timeline broken on last phase too.

PR #2245, opened 2020-12-19
Raidboss: E10n timeline and triggers

JLGarber:
Nothing much to say on the encounter itself. It's very straightforward for both timeline and triggers. Much appreciation to @quisquous for finishing up E10s so I could borrow!

I noted it in-line in the timeline file, but the two phase change abilities are not translated: `英雄の影が受けた攻撃：影の遠吠え`, "Attack on the Shadow of the Hero: Howling Shadow", and `英雄の影が受けた攻撃：ヴォイドパルセーション`, "Attack on the Shadow of the Hero: Void Pulsation". I translated or `"--sync--"`d both. (This is also 100% necessary because both the plugin and `test_timeline` complain about having "unknown characters" if the untranslated names are even in timeline comments. It's easily solvable in `test_timeline` by ensuring the timeline is opened as UTF-8, but I'm not 100% sure of the solution for the timeline controller.) Thanks @panicstevenson for talking me through the encoding issue!

Finally, `"Front/Back Shadow Implosion"` is not correct. It should be `"Forward/Backward Shadow Implosion"`. I shortened it this way in the interests of space. Improvement suggestions are welcome.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e10n.txt
+++ b/ui/raidboss/data/05-shb/raid/e10n.txt
@@ -4,7 +4,77 @@
 hideall "--Reset--"
 hideall "--sync--"
 
+# Two phase block abilities are not translated:
+# 5B0D, 英雄の影が受けた攻撃：影の遠吠え, which is
+# Attack on the Shadow of the Hero: Howling Shadow (shortened here as Howling Shadow)
+
+# 56CC, 英雄の影が受けた攻撃：ヴォイドパルセーション, which is
+# Attack on the Shadow of the Hero: Void Pulsation (--sync--, since it lines up with Void Pulse)
+
+# -ii 56B0 56B6 56B9 56BA 56BB 56BD 56BE 56C0 56C1 56D4 56D5 56D8 56D9 56E1 56E7 5B09 5B26
+
 0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
 
-0 "Start"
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/raid/e10n.txt
```diff
@@ -4,7 +4,80 @@
...
+# Two phase block abilities are not translated:
+# 5B0D, https://xivapi.com/action/23309, which is
+# Attack on the Shadow of the Hero: Howling Shadow (shortened here as Howling Shadow)
+
+# 56CC, https://xivapi.com/action/22220, which is
+# Attack on the Shadow of the Hero: Void Pulsation (--sync--, since it lines up with Void Pulse)
+
+# Neither of these attacks' names can safely be in the actual file,
+# since multiple utilities throw exceptions based on characters they contain.
+
```
quisquous, line 16:
```suggestion
```

This is just a suggestion, but I think it is fine to skip and `-ii` these.  These are all ignored in savage, because so many abilities hit the shadows, and I don't think it's useful to the player.

quisquous (commented):
Looks good overall with one extremely small piece of feedback:

ui/raidboss/data/05-shb/raid/e10n.txt
JLGarber: Looking at it again, that does seem reasonable, yes.

PR #2252, opened 2020-12-20
Raidboss: Fix diametrically wrong E12N N/S callouts

JLGarber:
Thanks to @valarnin for chasing down instances where this happened. It looks like the coordinate system Square uses is just awful all round. (If it weren't completely impossible given what we know of Square, I would have speculated that the coordinates were randomized on each entry.)


quisquous:
+y down is a perfectly normal coordinate system.  :P

PR #2253, opened 2020-12-20
util: handle edge cases of broken lines in encounter tools

quisquous:


quisquous:
I uh apparently need a new hard drive, maybe.  But I have some corruption in my log files, like:
`00|2027-02-01T22:13:52.0000000-08:00|38207365|38 damage.;z�_�`

And so encounter tools just breaks down and can't process anything with `-lf`:
```
  File "util/make_timeline.py", line 499, in <module>
    print("\n".join(main(args)))
  File "util/make_timeline.py", line 273, in main
    return parse_file(args)
  File "util/make_timeline.py", line 154, in parse_file
    encounter_sets = e_tools.find_fights_in_file(file)
  File "util\encounter_tools.py", line 123, in find_fights_in_file
    if is_instance_begun(line_fields):
  File "util\encounter_tools.py", line 84, in is_instance_begun
    return line_fields[4].endswith("has begun.")
IndexError: list index out of range
```

I'm not sure if there's a better way to handle this?  Should we try/catch?  Should we have an `is_valid_log_line` to check up front?

JLGarber:
Mmrf, I agree that duplicating the logic everywhere is horrible, but it would be two lines in each function whether or not we extract it to a helper. I don't think there's a better way to handle this that would be all at the same time more correct, more concise, and more understandable.

quisquous:
> Mmrf, I agree that duplicating the logic everywhere is horrible, but it would be two lines in each function whether or not we extract it to a helper. I don't think there's a better way to handle this that would be all at the same time more correct, more concise, and more understandable.

Another alternative is a single `is_valid_line` helper that we check before calling any of these functions, rather than putting it in each function?

JLGarber:
Maybe? I guess that probably would be better, even if it is One More Thing In The Giant Function Tower.

quisquous:
Haha, it sounds like neither of us have strong feelings.  I'll just merge this as-is.

PR #2255, opened 2020-12-22
oopsy: add e9n warnings

quisquous:


JLGarber (approved):
Thanks for handling all of these! I was probably going to do this Eventually(TM) but having them now is great.

PR #2256, opened 2020-12-22
oopsy: add e10n warnings

quisquous:


JLGarber (approved):
Looks very nice! Do you suppose it might be worth handling "knocked back into death wall" as well, or is that too much trigger work for one thing that most people just invuln anyway?

quisquous:
> Looks very nice! Do you suppose it might be worth handling "knocked back into death wall" as well, or is that too much trigger work for one thing that most people just invuln anyway?

Does the wall outright kill you? Or is it a high damage tick? This is one of those cases where I really wish we had log lines for hitting death walls.  :C

PR #2260, opened 2020-12-22
raidboss: handle e12s door boss headmarkers

quisquous:


ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -146,10 +138,183 @@ const primalOutputStrings = {
   },
 };
 
+// Due to changes introduced in patch 5.2, overhead markers now have a random offset
+// added to their ID. This offset currently appears to be set per instance, so
+// we can determine what it is from the first overhead marker we see.
+// The first 1B marker in the encounter is the formless tankbuster, ID 004F.
+const firstHeadmarker = parseInt('00DA', 16);
+const getHeadmarkerId = (data, matches) => {
+  // If we naively just check !data.decOffset and leave it, it breaks if the first marker is 004F.
```
JLGarber, line 147:
```suggestion
  // If we naively just check !data.decOffset and leave it, it breaks if the first marker is 00DA.
```

For now, until we move this function into a utility file somewhere?

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -146,10 +138,183 @@ const primalOutputStrings = {
...
+          tankBusters: {
+            en: 'Tank Busters',
+          },
+        };
+
+        // Already called out in the headmarker trigger.
+        if (data.formlessTargets && data.formlessTargets.includes(data.me))
+          return;
+
+        // TODO: should this call out who to cover if you are a paladin?
```
JLGarber, line 306: By default, no. Maaaaaybe set it up as configurable like we've done for the TEA Wormhole bit or the Shiva whatever-it-was there?

JLGarber (approved):
Only tiny nits here, looks good overall.

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: What's wrong for doing this with paladins? It's not a strat change option, it's more of a "how can I save the person who is second in enmity because the tank died and forgot to provoke".

JLGarber: Oh, right, sorry! I was thinking of some "cover +invuln" nonsense like the Earth Shaker bit from O4s, but of course that doesn't apply here.

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -146,10 +138,185 @@ const primalOutputStrings = {
...
+const firstHeadmarker = parseInt('00DA', 16);
+const getHeadmarkerId = (data, matches) => {
+  // If we naively just check !data.decOffset and leave it, it breaks if the first marker is 004F.
+  // (This makes the offset 0, and !0 is true.)
+  if (typeof data.decOffset === 'undefined')
+    data.decOffset = parseInt(matches.id, 16) - firstHeadmarker;
+  // The leading zeroes are stripped when converting back to string, so we re-add them here.
+  // Fortunately, we don't have to worry about whether or not this is robust,
+  // since we know all the IDs that will be present in the encounter.
+  return '00' + (parseInt(matches.id, 16) - data.decOffset).toString(16).toUpperCase();
```
panicstevenson, line 154:
```suggestion
  return (parseInt(matches.id, 16) - data.decOffset).toString(16).toUpperCase().padStart(4, '0');
```

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -146,10 +138,185 @@ const primalOutputStrings = {
...
+        // Track tankbuster targets, regardless if this is on you or not.
+        // Use this to make more intelligent calls when the cast starts.
+        if (id === '00DA') {
+          data.formlessTargets = data.formlessTargets || [];
+          data.formlessTargets.push(matches.target);
+        } else if (id === '00BB') {
+          data.weightTargets = data.weightTargets || [];
+          data.weightTargets.push(matches.target);
+
+          // Handle double blue titan on 2nd and 3rd iterations.
```
panicstevenson, line 226: Third iteration I don't think your partner will ever matter, since it's a T/H and a DPS, so I imagine these spots aren't being called on the fly?

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -146,10 +138,185 @@ const primalOutputStrings = {
...
+      id: 'E12S Promise Weight Cleanup',
+      netRegex: NetRegexes.startsUsing({ source: 'Eden\'s Promise', id: '58A5', capture: false }),
+      run: (data) => {
+        delete data.weightTargets;
+        data.seenFirstBombs = true;
+      },
+    },
+    {
+      id: 'E12S Promise Formless Judgment',
+      netRegex: NetRegexes.startsUsing({ source: 'Eden\'s Promise', id: '5858A9A5', capture: false }),
```
panicstevenson, line 291: Interesting abilityId!

ui/raidboss/data/05-shb/raid/e12s.js
quisquous:
```
          // The second and two rounds of bombs have a partner.
          // The third is technically fixed by role with a standard party (one dps, one !dps),
          // but call out your partner anyway in case you've got 8 blus or something.
```

It's kind of a "might as well call it out" scenario, since we're already doing the work.

quisquous:
This patch mostly didn't work, probably because of what panic suggested (as headmarkers don't all start with 00).  A few things got called out (correctly) but most things didn't get called at all.  There were also some errors I fixed in fa67b4f.  I also need to make the headmarkers not apply to the oracle fight.  At any rate, this needs another round of testing before I can even believe it half works.

quisquous:
Ok, I ran this through the emulator.  The emulator doesn't seem to be analyzing everybody (?!) but for the people that it's analyzing, I think it should be ok.  I'd like to give it another manual test before landing, though.

panicstevenson (approved):
Works well in my raid.

Legends0:
This work well. I think adding position of statues would be neat thing, most groups do north/south.

PR #2262, opened 2020-12-22
raidboss: add some oracle triggers

quisquous:


JLGarber (approved):

JLGarber:
Can't test these myself yet, but it all looks fine!

quisquous:
Ok, fixed some bugs, but want to give this a manual test as well.

Legends0:
They don't seem to work in my experience. No errors reported.

quisquous:
> They don't seem to work in my experience. No errors reported.

How did you apply the patch? Are you running the webpack dev server?

Legends0:
> 
> 
> > They don't seem to work in my experience. No errors reported.
> 
> How did you apply the patch? Are you running the webpack dev server?

I overwrite the last commit here of the file changed in this PR then open ACT. I may be missing other commits?

quisquous:
> > > They don't seem to work in my experience. No errors reported.
> > 
> > How did you apply the patch? Are you running the webpack dev server?
> 
> I overwrite the last commit here of the file changed in this PR then open ACT. I may be missing other commits?

Yeah, it's less about other commits, and more a question of if you have run `npm run build` or `npm start`.  #2316 might be what you're looking for.

panicstevenson:
One alternative too, is you can [take this commit and push it up to your fork](https://github.com/quisquous/cactbot/blob/60a89da24abcbc0c653c2e4a80ffc6803717c6c9/CONTRIBUTING.md#validating-changes-via-remote-urls) and point to your GitHub Pages URL if you have that setup.

Legends0:
Look Away call during Basic Relativity is happening after the rewind debuff has already expired.

It seems to work well for intermediate. The three callouts could perhaps use a bit of a pause between each debuff, but maybe that would be too long of a text.

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -351,6 +389,165 @@ export default {
...
+        const key = data.intermediateDebuffs && data.intermediateDebuffs.shift();
+        if (!key)
+          return { infoText: output.moveAway() };
+        return { alertText: output[key]() };
+      },
+    },
+    {
+      // For all relativities, players should look outside during the final return effect.
+      id: 'E12S Relativity Look Outside',
+      netRegex: NetRegexes.gainsEffect({ effectId: '994' }),
```
Legends0, line 541: This is incorrectly triggered during Basic Relativity. Instead, for Basic Relativity it should be a few seconds before the Spell-in-Waiting: Return (9A0) ends.

quisquous:
> It seems to work well for intermediate. The three debuff callouts could perhaps use a bit of a pause between each debuff, but maybe that would be too long of a text. 

Are you using tts? By "three debuff callouts", do you mean the initial callout?

The usual pattern for stuff like this (and quintuplecast) for me is to put one long `infoText` with all the mechanics, and then call out each one individually.  I think this makes a lot of sense when you can keep that text on screen during the mechanic but much less sense for tts users.

> EDIT: I think the look away may be incorrect for basic relativity, but I am not sure if it was due to other issues in the group I was with. I had to look away the moment before Spell-in-Waiting: Return (9A0) went off, but I also tended to be looking away as Return went off.

My understanding of the mechanic is that for basic relativity, you get `Spell-in-Waiting: Return`.  When that debuff expires, you drop a small circle and you get `Return`.  When `Return` goes off, your character (and its facing direction) are teleported back to where `Spell-in-Waiting: Return` expired.  The facing direction is only at the very very end of the mechanic when `Return` finally expires, and that holds true for all relativities.

Legends0:
> 
> Are you using tts? By "three debuff callouts", do you mean the initial callout?
> 

Yea the tts during the initial callout of relativities. I never had much issue with quintuple cast. I think it will just take some getting use to as it did for the e11s cycles where its another similar situation of lots of info in a short time.

> 
> My understanding of the mechanic is that for basic relativity, you get `Spell-in-Waiting: Return`. When that debuff expires, you drop a small circle and you get `Return`. When `Return` goes off, your character (and its facing direction) are teleported back to where `Spell-in-Waiting: Return` expired. The facing direction is only at the very very end of the mechanic when `Return` finally expires, and that holds true for all relativities.

In my situation may just have been due to varied party strat or someone standing in a wrong position in the pug I was testing it with. I wasn't very aware of who/what was causing each persons doom. Some strategies for Basic have positioning that do not need the look outside for Basic, but that may have been role specific position in another party I was with. I suppose it's best to just keep it in and let people enable/disable it.

Some additional triggers for Apocalypses (may vary due to strats) that may be helpful:
Singular Apocalypse: could be helpful to add an indicator for when middle is safe, such as with the cw/ccw callout add something like out of middle or wait for middle. The timing for the out of middle would be a bit after Black Halo is cast.
Dual Apocalypse: could benefit from similar, after Cataclysm cast. For this mechanic, it seems most parties will be waiting for the center AoE to go off, then going center. The callout hear is just for non-tanks and the stack callout should work for the tanks for that part.
Triple Apocalypes: could benefit from an out of middle or wait for middle when Darkest Dance is started casting (or maybe a bit before?).

Legends0:
An update, I figured out what is going on in Basic. There are two times you have to look away. Once before the rewind and then at the end as is normal in all relativities. It's probably in the others as well but positioning is just different.

Legends0:
I wrote some code for the shadow eye in Basic Relativity that is basically copied from e2s. https://github.com/Legends0/cactbot/commit/1a3eba998e552f34bf0004da38f29f9bafd28ed7 I'll be able to test it out this weekend.

quisquous:
> An update, I figured out what is going on in Basic. There are two times you have to look away. Once before the rewind and then at the end as is normal in all relativities. It's probably in the others as well but positioning is just different.

Oh, you're right.  Shadoweye does go off with Spell-In-Waiting: Return.  We have the tanks and healers in a line roughly so if everybody is facing the boss, everybody naturally is facing away.  I think there's no second look away for basic, as that's the only shadoweye.

PR #2264, opened 2020-12-23
resources: add new OutputStrings dictionary of translations

### LINKED ISSUE #1307: Add repository of common translations
Mentioned in several discussions at this point, and so collecting into a single issue here.

`responses.js` has done a good job of eliminating large translations, but it got a little bit complicated to handle a lot of things like tankbusters.  Each property on `Responses` is a function that takes a severity and returns an object for a response field on a trigger, e.g. `{ alertText: { en: 'Buster on YOU', de: etc etc } }`.  This is complicated to use in any real way.

I think we need some separate global object that folks can get translations from? It could also live in `responses.js` because it's probably going to be used for responses, but it should be a separate object from` Responses` itself.  Putting it in the `responses.js` file also means it'd be straightforward to have all of the things in `Responses` use these new translations too, which sounds nice to avoid even more duplication.

Opinions welcome for what we should call this new thing? `Translations` feels long.  Names are hard.

quisquous:
Fixes #1307.


resources/output_strings.js
```diff
@@ -0,0 +1,536 @@
...
+    });
+  }
+}
+
+// General guidelines:
+// * property names should closely match English text
+// * use OnPlayer suffix for things with `${player}`
+// * use OnTarget suffix for things with `${name}`
+// * any other parameters (of which there are none, currently) should use consistent suffixes.
+// * the value of each property should be a single object with localized keys
```
quisquous, line 35: Feel free to disagree or propose other guidelines.  This is what I thought made sense to me for a set of translations that would get used elsewhere outside of responses.

I do think the indirection causes a little bit of cognitive friction, which is why I proposed the rules of closely match text, and use common suffixes for the two included parameter types.

resources/output_strings.js
```diff
@@ -0,0 +1,536 @@
...
+const OutputStrings = new ThrowOnInvalidProxy({
```
quisquous, line 36: Also not sure that `OutputStrings` is the best name here, but I think it kind of fits with the usage?

resources/output_strings.js
panicstevenson: The only one that I feel like bothers me here is `lookAwayFromTarget`, since I feel that one is usually player, but has the corner case of sometimes being look away from boss (which would just be `lookAway`). Is there anything else you have to distinctly look away from?

For context, I'm fine with it staying `lookAwayFromTarget`, but my issue is that it's mostly player and maybe we should just add a line here or something that it OnPlayer is distinctly always player mechanics, or something like that.

I get it largely doesn't matter at the end of the day, but it did make me do a double take in the review if `lookAwayFromTarget` didn't get updated properly.

resources/output_strings.js
panicstevenson: I like `Outputs`, following the pattern of `Conditions`, `Responses`, `Outputs`, but I don't feel strongly.

panicstevenson (approved):
LGTM, if you want a free nit, maybe we can remove the comments from responses since it feels weird to duplicate them.

quisquous:
> LGTM, if you want a free nit, maybe we can remove the comments from responses since it feels weird to duplicate them.

I am torn a little bit.  I think because there's so many responses and outputs, even if they were in the same file, it'd be easy to have some confusion on what things are.  I think people who are building triggers might look in either place to get the strings or responses they need.  I don't feel strongly, but that was my reasoning.

resources/output_strings.js
quisquous: We could arguably have both?

JLGarber:
> I am torn a little bit. I think because there's so many responses and outputs, even if they were in the same file, it'd be easy to have some confusion on what things are. I think people who are building triggers might look in either place to get the strings or responses they need. I don't feel strongly, but that was my reasoning.

I think this is probably the better option. Strictly speaking Panic is right, we don't need to duplicate stuff like that, but realistically, this can serve users better, and I think that should be the overriding concern here.

resources/output_strings.js
quisquous: Added it.  Do you think `lookAwayFrom` should use both of them? Or fine to just have both in `Outputs`?

panicstevenson (approved):

Issue #2063, opened 2020-11-24
Add mocha for unit tests


quisquous:
Currently cactbot unit tests are a bit freeform and it'd be nice to use something like mocha to give them some structure, be able to run specific tests.  Just filing this because I keep thinking about it, and wanted to write it down somewhere.

panicstevenson:
I think conceptually the work here is very straightforward, if someone was looking to pick something up with minimal oh-no-you-broke-people's-triggers-mid-raid worries. If anyone's interested, feel free to chime in. If not, I can pick this up in a couple days.

Trim21:
Will we choose jest or mocha as test tool? Both tools can handle typescript painless.

MaikoTan:
Personally I prefer mocha, though.

panicstevenson:
From what @quisquous and I tested a while ago, we had to use Jest's [transformers](https://jestjs.io/docs/en/configuration#transform-objectstring-pathtotransformer--pathtotransformer-object) to use the tests how we use them, since we aren't testing code but instead are running text files through tests.

quisquous:
Fixed in #2483, #2486, #2487, #2488.

I liked Jest's transformers better because you could ask Jest to go test `ui/raidboss/data/00-misc/test.js` and it would run a trigger test on it.  However, I couldn't figure out how to get Jest tests to deal with modules (or typescript) properly and it was a huge pain to set up.  Mocha is a lot more direct, although there's a weird indirect in #2483 to deal with testing data files.

PR #2265, opened 2020-12-23
ci: make test faster with asyncio

MaikoTan:
Require Python 3.7 and over, so I don't know if this is a good idea.

Would make test faster than before for about 20 sec.

This is just a demo, would continue to optimize it soon.

### DIFF
--- a/test/run_tests.py
+++ b/test/run_tests.py
@@ -1,5 +1,6 @@
 """Aggregate test runner for Cactbot Python tests."""
 
+import concurrent.futures
 import sys
 import test_manifests
 import test_timelines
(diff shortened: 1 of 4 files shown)


quisquous:
I definitely agree that tests are really slow, but I think unfortunately that this approach won't work in a future where the test harness switches to jest.  (Sorry!  It's a really good idea, but I'm just not sure it will apply any more in the future.)

I think the test harness that exists right now really has no features (you can't specify test names, or any number of things you want in a test runner), and so switching to something better will help.  Switching to something like jest or mocha will also make it more possible to handle typescript tests.

See also: #2063
See also: https://jestjs.io/docs/en/cli#--maxworkersnumstri

MaikoTan:
Ok, didn't notice that.

I have not used jest, but mocha is definitly a better test framework, maybe I can help when you start to work on typescript or test things.

PR #2266, opened 2020-12-23
raidboss: Remove Unnecessary Headmarker Regex

panicstevenson:
There was a message on Discord about this potentially not working in some instances; I don't know if this is _the_ fix, but this seems unnecessary to be included.


quisquous:
Seems reasonable!!

PR #2267, opened 2020-12-23
raidboss: update e12n rapturous reach ids

quisquous:


JLGarber (approved):
That looks right. Sorry I didn't catch it the first pass!

PR #2270, opened 2020-12-23
github: run lint workflow on lint rule changes

quisquous:


.github/workflows/javascript-lint.yml
```diff
@@ -4,12 +4,18 @@ on:
   push:
     paths:
       - '**/*.js'
+      - '**/*.cjs'
+      - '**/*.mjs'
       - '**/*.json'
+      - 'eslint/'
```
panicstevenson, line 10: Feels redundant? Only .json and .js files in there?

quisquous: Thanks, good point!

PR #2271, opened 2020-12-23
eslint: Replace --rulesdir with rulesdir plugin

quisquous:
This is a followup to #2066 to fix the programmatic use of eslint in `util/gen_coverage_report.js` that is breaking here: https://github.com/quisquous/cactbot/runs/1602596519


quisquous:
Other options including using `npm` to install a local directory into `node_modules`, but that also means that you have to run `npm install` any time anything in that local directory changes.  There's also `npm link`, but can't symlink on Windows.  So, this seemed like the least worst approach.

panicstevenson (approved):
It feels like this all might need a mild revisit when we do TypeScript stuff, so it feels like this is all very reasonable for now

quisquous:
> It feels like this all might need a mild revisit when we do TypeScript stuff, so it feels like this is all very reasonable for now

Oh, interesting.  I'd be curious what you're thinking of here.

PR #2272, opened 2020-12-23
eslint: lint all the things

quisquous:


webpack/webpack.config.cjs
```diff
@@ -1,5 +1,3 @@
-'use strict';
```
panicstevenson, line 1: Are these fine to just go away? I thought they wanted to be here, but I admittedly have no idea what strict mode does.

quisquous: No, I think you're correct.  I think cjs files all need use strict in them, but I cannot seem to convince eslint that these are commonjs modules.

quisquous: As per https://eslint.org/docs/rules/strict#rule-details, if `'sourceType': 'module'` is set, then the strict mode rule is ignored.

quisquous: OK found some override options!!

PR #2288, opened 2020-12-24
Raidboss: add missing ja translation in e10s

MaikoTan:


quisquous:
Ohmygosh, thanks for all of these translations!!!

MaikoTan:
I wrote a new eslint rule to find out where lacks specified translations, and call it from vscode tasks.


I think this is personized that everyone needs different options, so I didn't upload it.

PR #2289, opened 2020-12-24
ci: add problem matcher for eslint

MaikoTan:
image

### DIFF
--- a/.github/matchers/eslint.json
+++ b/.github/matchers/eslint.json
@@ -1,0 +1,22 @@
+{
+  "problemMatcher": [
+    {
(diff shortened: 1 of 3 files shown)


quisquous:
I think this is built in to github's problem matchers by default, and doesn't need to be specified separately.

test here: https://github.com/quisquous/cactbot/commit/4e6fa61bf6cce71761bcdaaec0b2d28e9d7a70e5

MaikoTan:
Oh I got it, it is automatically added by setup-node step, sorry for redundant (TдT)

PR #2292, opened 2020-12-24
Add method to register overlays with external URL. Add ZeffUI

Zeffuro:
Had to add new method to allow overlays with external URLs.
Hope that's okay.

### DIFF
--- a/plugin/CactbotEventSource/CactbotEventSource.cs
+++ b/plugin/CactbotEventSource/CactbotEventSource.cs
@@ -628,7 +628,7 @@ namespace Cactbot {
     {
         Registry.RegisterOverlayPreset(new OverlayPreset
         {
-            Name = $"Cactbot External {name}",
+            Name = $"{name}",
             Url = url,
             Size = new int[] { width, height },
             Locked = locked,
--- a/plugin/CactbotEventSource/CactbotEventSource.cs
+++ b/plugin/CactbotEventSource/CactbotEventSource.cs
@@ -624,14 +624,14 @@ namespace Cactbot {
       });
     }
 
-    private void RegisterExternalPreset(string name, string url, int width, int height, bool locked)
+    private void RegisterExternalPreset(string name, string url, int width, int height)
     {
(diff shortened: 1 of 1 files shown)


plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -624,6 +624,17 @@ struct OverlayPreset : IOverlayPreset {
       });
     }
 
+    private void RegisterExternalPreset(string name, string url, int width, int height, bool locked)
+    {
+        Registry.RegisterOverlayPreset(new OverlayPreset
+        {
+            Name = $"Cactbot External {name}",
```
quisquous, line 631:
```suggestion
            Name = $"{name}",
```

I think it'd be fine to just call this `ZeffUI`.  I think calling it a cactbot anything will likely be more confusing to users than just calling it what it is.  I think I'm just looking towards the future where there's not a cactbot plugin, and so this would be `ZeffUI` in a world where you could just add this to OverlayPlugin directly.

plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -624,6 +624,17 @@ struct OverlayPreset : IOverlayPreset {
...
+            Url = url,
+            Size = new int[] { width, height },
+            Locked = locked,
```
quisquous, line 634: I think this should probably just be `Locked = false` (with no function param) like the other ones until there's an example that needs it.

Zeffuro: That's fair, I'm kinda used to leaving things open as an option but I get it.

Zeffuro:
Made the suggested changes!

plugin/CactbotEventSource/CactbotEventSource.cs
Zeffuro: Yeah that's fine. Make sure to let me know when OverlayPlugin is far enough along for me to switch over :)

quisquous:
Thanks!

PR #2293, opened 2020-12-24
webpack: share raidboss/oopsy data across entries

quisquous:
Requires upgrading to webpack 5.0.0+ to support the `dependsOn` field / more complex `entry` field types.  Upgraded to latest because why not.


panicstevenson (approved):

quisquous:
Sorry, I thought I tested this better, but whoops.  I kept the webpack upgrade and reverted the sharing in 232715fd31bb41bb22923abbced7928461433412.

I wonder if this is a case where we need to have everything in webpack or if I am just holding it wrong.

PR #2296, opened 2020-12-25
other: add python-related settings of vscode

MaikoTan:
For consistency.
Use `Black` as formator, `pylint` as linter.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,10 @@
 {
+  "eslint.options": {
+    "rulePaths": [
+      "./eslint"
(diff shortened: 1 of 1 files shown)


.vscode/settings.json
```diff
@@ -1,4 +1,10 @@
 {
+  "eslint.options": {
+    "rulePaths": [
+      "./eslint"
+    ]
+  },
```
quisquous, line 6: I think this shouldn't be needed.  I changed this to be done through a plugin instead so that it would work without options.

PR #2299, opened 2020-12-26
jobs: make space for BLM's Xeno counter resource box

Jessidhia:
The #right-side-icons are being pulled back behind the resource box as BLM is being treated as having 0 boxes on the right side.

Before: 
image

After: 
image


quisquous:
Thank you for the bug report plus fix! Super appreciated.

PR #2301, opened 2020-12-26
docs: Korean translation with some old contents update

Jaehyuk-Lee:
Translating to Korean is on progress, I have some concerns with old contents with readme.

## obvious changes
1. `utils` directory is changed to `util`
1. "curse" is changed to "curseforge", changed link for BigWigs Bossmods.
1. timeline and trigger uses same directory, so I've put them together.

## about [readme.md#building from source](https://github.com/quisquous/cactbot/blob/main/README.md#building-from-source) paragraph
After [v0.20.0](https://github.com/quisquous/cactbot/releases/tag/v0.20.0), building step is added.
Need to run `npm run build` for building `dist` folder. Prior to this, we need `node.js` install and `npm install`.
So, we should separate building `dll` file and building `dist/*.js` file.
I can write about building `dist/*.js` files but I think it'll be look like translationese, I think it would be better to be written by native English speaker.

Also, I've written new sentences about timeline and trigger file locations, please review them. readme.md line 271

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ To install dependencies there are 2 methods: **per script** and **manually**
 ### Dependencies: Script Method
 
 1. `curl` MUST be installed (this is used to download dependencies)
-1. Execute the `./utils/fetch_deps.py` script
+1. Execute the `./util/fetch_deps.py` script
 1. Continue with **Steps to build**
 
 ### Dependencies: Manual Method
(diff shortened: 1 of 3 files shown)


quisquous:
The changes to the English readme look good to me.  Thanks for looking into this.

I can follow up with `npm` instructions after merging this, if that makes sense? Is there anything else preventing this pr from being merged?

Jaehyuk-Lee:
okay, then I'll make new PR after new npm build thing added. I planned to open PR after finishing all translation, but I think yeah, later is fine.

PR #2303, opened 2020-12-26
Raidboss: E9N timeline

JLGarber:
This wasn't too bad to put together, but it does look like there's more variance on this than I'm used to. Wide-Angle Phaser and Rejuvenating Balm just seem to drift slightly from pull to pull. None of it actually matters since the exact timing on those two mechanics isn't make-or-break for anyone, but it's annoying. My suspicion is that when she tethers the wall, there's a difference in animation time between "tethers front" and "tethers back". Everything else seems to be mostly fine.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e9n.txt
+++ b/ui/raidboss/data/05-shb/raid/e9n.txt
@@ -52,7 +52,7 @@ hideall "--sync--"
 340.1 "Waste Away" sync /:Cloud Of Darkness:55DE:/
 342.4 "Evil Seed" sync /:Cloud Of Darkness:55E7:/
 361.7 "Wide-Angle Phaser" sync /:Cloud Of Darkness:55E0:/
-376.5 "Rejuvenating Balm" sync /:Cloud Of Darkness:55E2:/
+376.8 "Rejuvenating Balm" sync /:Cloud Of Darkness:55E2:/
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/raid/e9n.txt
```diff
@@ -4,7 +4,64 @@
...
 
+# -ii 5626 5B18 55E1 55E3 5159 532F 55DB 55E5
+
 0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
 
-0 "Start"
-0.0 "--sync--" sync /:Engage!/ window 0,1
+0 "Start" sync /Engage!/ window 0,1
+# Unfortunately we can't sync to a first attack here,
+# as Cloud of Darkness uses an auto at random before starting her 3-second rotation.
```
quisquous, line 13: Could sync to the starts casting line of Ground-Razing Particle Beam to start the timeline sooner, maybe?

quisquous (commented):
That seems reasonable overall.  It looks like there's plenty of windows, so small drift seems fine.  How much drift are you talking about, mostly for my own curiosity?

JLGarber:
Not truly awful, it's usually around 0.6-0.8 seconds, but I've seen it max at something like 1.2.

quisquous:
Thanks!

PR #2304, opened 2020-12-27
Add support for user folders

### LINKED ISSUE #860: Add user folder support for subfolders
It's cumbersome to stick all the logic into a single file in the user folder.  It'd be easier if folks could share whole files that could be dropped into

One possibility here is to not just load the javascript file named the same as the overlay, but also the subfolder (and anything underneath it).  In other words, `user/raidboss.js` is loaded, but also any javascript/css files under `user/raidboss/*`.  Maybe we'd want to ignore `.git` directories (or all dot directories?) in case folks put these under version control.

quisquous:
Fixes #860.

### DIFF
--- a/docs/CactbotCustomization.md
+++ b/docs/CactbotCustomization.md
@@ -86,6 +86,23 @@ The `oopsyraidsy` module loads `user/oopsyraidsy.js` and `user/oopsyraidsy.css`.
 And so on, for each module.
 These files are included after cactbot's files and can override its settings.
 
+Additionally, cactbot will also load any files in a subdirectory of the same module name.
+For example, `raidboss` will also load any `.js` and `.css` files
+underneath `user/raidboss/` even if in multiple subfolders.
+This is intended to make it easier to share triggers and customizations with others.
+Timeline `.txt` files must be directly in the same folder as the `.js` that refers to them.
+
+cactbot loads files in subdirectories (alphabetically) before loading files in outer directories.
+This is so that `user/raidboss.js` will always be loaded last
+and can override anything that is set inside a file inside of `user/raidboss/`.
+For example, `user/alphascape/some_file.js` will load before `user/mystatic/file1.js`,
+which will both load before `user/raidboss.js`.
+The same applies to `.css` files.
+
+You can get more information about the loading order
+by looking at the [debug messages](#check-if-your-file-is-loaded)
+when developer mode is turned on.
+
 The `user/` directory already includes some example configuration files,
 which you can rename and use.
 For example the [user/raidboss-example.js](../user/raidboss-example.js) file
@@ -111,13 +128,13 @@ This is often in `%APPDATA%\Advanced Combat Tracker\Plugins\cactbot-version\cact
 
 ## Customizing Appearance
 
-The `user/<name>.css` file can change positions, sizes, colors, etc. for components of
+A user css file can change positions, sizes, colors, etc. for components of
 the UI module. See the `ui/<name>/<name>.css` to find the selectors you can modify.
 
 For example in [ui/raidboss/raidboss.css](../ui/raidboss/raidboss.css),
 you see the `#popup-text-container` and `#timeline-container`
 which can be changed via `user/raidboss.css` to different positions as desired.
-You can use `user/raidboss.css` to add additional styling.
+You can use `user/raidboss.css` or a `.css` file in `user/raidboss/` to add additional styling.
 
 The size and color of info text alerts can also be changed
 by making a CSS rule for the `.info-text` class such as below:
@@ -423,6 +440,8 @@ It will list lines for each local user file it loads:
 `[10/19/2020 6:18:27 PM] Info: raidbossy: BrowserConsole: local user file: C:\Users\tinipoutini\cactbot\user\raidboss.js`
 
 Verify that your user file is loaded at all.
+
+The order that the filenames are printed are the order in which they are loaded.
 
 ### Check if your user file has errors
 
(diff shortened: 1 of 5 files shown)


quisquous:
Re: docs.  Should I change all references to `user/raidboss.js` to say something like "a user raidboss js file", or something along those lines?

docs/CactbotCustomization.md
```diff
@@ -86,6 +86,23 @@ The `oopsyraidsy` module loads `user/oopsyraidsy.js` and `user/oopsyraidsy.css`.
...
+Additionally, cactbot will also load any files in a subdirectory of the same module name.
+For example, `raidboss` will also load any `.js` and `.css` files
+underneath `user/raidboss/` even if in multiple subfolders.
+This is intended to make it easier to share triggers and customizations with others.
+Timeline `.txt` files must be directly in the same folder as the `.js` that refers to them.
+
+cactbot loads files in subdirectories (alphabetically) before loading files in outer directories.
+This is so that `user/raidboss.js` will always be loaded last
+and can override anything that is set inside a file inside of `user/raidboss/`.
+For example, `user/alphascape/some_file.js` will load before `user/mystatic/file1.js`,
```
panicstevenson, line 98:
```suggestion
For example, `user/alphascape/some_file.js` will load before `user/mystatic/some_files.js`,
```

docs/CactbotCustomization.md
```diff
@@ -111,13 +128,13 @@ This is often in `%APPDATA%\Advanced Combat Tracker\Plugins\cactbot-version\cact
 
 ## Customizing Appearance
 
-The `user/<name>.css` file can change positions, sizes, colors, etc. for components of
+A user css file can change positions, sizes, colors, etc. for components of
```
panicstevenson, line 131:
```suggestion
A user-defined css file can change positions, sizes, colors, etc. for components of
```

docs/CactbotCustomization.md
```diff
@@ -424,6 +441,8 @@ It will list lines for each local user file it loads:
 
 Verify that your user file is loaded at all.
 
+The order that the filenames are printed are the order in which they are loaded.
```
panicstevenson, line 444:
```suggestion
The order that the filenames are printed is the order in which they are loaded.
```

plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -476,48 +476,73 @@ protected override void Update()
       this.Log(LogLevel.Info, format, args);
     }
 
-    private Dictionary<string, string> GetLocalUserFiles(string config_dir) {
+    private static string GetRelativePath(string top_dir, string filename) {
+      // TODO: .net 5.0 / .net core 2.0 has Path.GetRelativePath.
+      // There's also a win api function we could call, but that's a bit gross.
+      // However, this is an easy case where filename is known to be rooted in top_dir,
+      // so use this hacky solution for now.  Hi, ngld.
```
panicstevenson, line 483: 👋

panicstevenson:
Do we need a test to include a folder name with spaces, since Windows for some reason doesn't prevent you from doing something insane like that?

docs/CactbotCustomization.md
quisquous: I picked this specifically because `file1` sorts before `some_file`.  Is that confusing?  Is there a better way to capture both?

docs/CactbotCustomization.md
quisquous: I think I was think of "user css" file as "user/ css" file here.  Should I change all instances of "user" to "user-defined"?

plugin/CactbotEventSource/CactbotEventSource.cs
quisquous: :innocent:

test/unittests/config_test.js
```diff
@@ -0,0 +1,98 @@
...
+      'B/subDir2/second.js',
+      'b/some_file.js',
+      'b/subdir1/file.js',
+      'root_file2.js',
+      'a/some file with spaces oof why.js',
+      'b/subdir2/first.js',
+    ];
+    const sortedKeys = [
+      'a/some file with spaces oof why.js',
+      'a/some.js',
```
quisquous, line 23: I'm not sure what I think about this.  This happens because `some ` sorts before `some.`, but for folders, `b` sorts before `b but with some spaces`.

Should we sort filenames ignoring extensions? Or is sorting the full filename ok? What do you think about this inconsistency (or maybe it isn't one).

panicstevenson: Hmmm, interesting. I guess my thought as a human with context is that a filename is different from an extension and that on Windows, I might not even have visible file extensions enabled. I would say that the most correct thing to do would be to sort without file extensions and then put them back, but that sounds like a lot of effort for something that I think everyone universally agrees is bad. I would argue that having the test is enough such that it can be pointed to if someone _really_ has an issue, but I think we're at like a corner of edge cases from both Windows and cactbot points of view, and there's a debug output that can be used as a frontline defense.

That being said, if you want to tackle it, go for it; I just don't think it's necessary.

quisquous: Done!

docs/CactbotCustomization.md
quisquous: Added more doc updates, lmk what you think.

quisquous:
@Jaehyuk-Lee @MaikoTan @Echoring FYI for more doc changes here.  I think this is the last big change that I am planning to make to customization, so hopefully this is the last major change for a while.

PR #2312, opened 2020-12-27
raidboss: convert Eden's Promise to use Outputs

quisquous:


quisquous:
I don't plan to go back and convert everything, but I thought I would at least do the latest tier to see what it looked like.

I think if other folks want to send patches that convert past content to use `Outputs`, I would happily merge them.  I think newer content in the future should always use `Outputs` where possible.

ui/raidboss/data/05-shb/raid/e10n.js
```diff
@@ -41,12 +42,7 @@ export default {
       netRegexJa: NetRegexes.startsUsing({ source: '影の王', id: '56B7', capture: false }),
       alertText: (data, _, output) => output.text(),
       outputStrings: {
-        text: {
-          en: 'Go Front',
-          de: 'Geh nach vorn',
-          ja: '前へ',
-          cn: '去前面',
-        },
+        text: Outputs.goFront,
```
JLGarber, line 45: Thanks for this! I was a little surprised this hadn't come up in any content recent enough to be useful, and I didn't have the time to do it myself when submitting this.

JLGarber (approved):
Very happy to see all this, I think it's a wonderful step forward.

PR #2315, opened 2020-12-27
E10N fixes and E10S issues with directions in FR

mooondark:
Directions strings are inconsistant in FR :
- _à l'est_ is good,
- but _à le nord_ is wrong (we should say _au nord_).

So I changed it to avoid this.


panicstevenson (approved):
This is just translating to `the north` and `the south`, etc. right? So, this would be okay for the other trigger callouts that exist; ie `Déposez le météore à l\'ouest` in Amarot, right? Since that says (roughly) "drop the meteor to the west"?

quisquous:
It looks like a lot of this is also in #2314, for what it's worth.

Sorry, this also needs to have conflicts resolved as I moved a bunch of these into `resources/outputs.js.

MaikoTan:
Sorry! This direction phrases was written by me.
I learnt a little french and was able to write simple sentences, but still made mistakes! 😂

mooondark:
I think the conflict is resolved now.
Thanks everybody.

PR #2316, opened 2020-12-27
docs: add references to nodejs and npm install

quisquous:
This is a follow-up to #2301.

### DIFF
--- a/README.md
+++ b/README.md
@@ -240,7 +240,7 @@ ThirdParty
 
 1. Install [nodejs and npm](https://nodejs.org/en/download/)
 1. Run `npm install` in the root of the cactbot directory.
-1. Run `npm build` or `npm start`.
+1. Run `npm run build` or `npm start`.
 
 See the [contributing](CONTRIBUTING.md#validating-changes-via-webpack) documentation
 for more details about using Webpack.
--- a/README.md
+++ b/README.md
@@ -238,16 +238,19 @@ ThirdParty
 
 ### npm and webpack
 
+If you are not a cactbot developer
+and are trying to modify cactbot for your own personal triggers,
+you should instead refer to the [customization documentation](docs/CactbotCustomization.md)
+instead of changing your local cactbot files.
+
+To install npm and start Webpack, follow these steps:
+
 1. Install [nodejs and npm](https://nodejs.org/en/download/)
 1. Run `npm install` in the root of the cactbot directory.
 1. Run `npm run build` or `npm start`.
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -229,13 +229,26 @@ ThirdParty
...
 1. Open the solution in Visual Studio (tested with Visual Studio 2017).
 1. Build for "Release" and "x64".
 1. The plugin will be built as **bin/x64/Release/CactbotOverlay.dll**.
 1. Add the built plugin directly as an ACT plugin.  In the ACT -> Plugins -> Plugin Listing tab, click the `Browse` button and find the **bin/x64/Release/CactbotOverlay.dll** where this file was built.  Then click `Add/Enable Plugin`.
 
+### npm and webpack
+
+1. Install [nodejs and npm](https://nodejs.org/en/download/)
+1. Run `npm install` in the root of the cactbot directory.
+1. Run `npm build` or `npm start`.
```
panicstevenson, line 243:
```suggestion
1. Run `npm run build` or `npm start`.
```

README.md
```diff
@@ -229,13 +229,26 @@ ThirdParty
...
+
+See the [contributing](CONTRIBUTING.md#validating-changes-via-webpack) documentation
+for more details about using Webpack.
+
+If you are a user and not a cactbot developer,
```
panicstevenson, line 248: This should be at the top of this block imo; kind of like a DRAGONS BELOW PLS DON'T GO HERE UNLESS YOU KNOW WHAT A GIT IS

JLGarber (approved):

README.md
MaikoTan: What is "DRAGONS BELOW" exactly means by the way? Although I can guess lmao. 😂

panicstevenson: It's a reference to [here be dragons](https://en.wikipedia.org/wiki/Here_be_dragons), basically like "beyond this point, your safety is not guaranteed"

quisquous: Ok, I moved it up.

I think I'm trying to strike a balance here between "hi if you're a developer and want to contribute to cactbot, we're happy to help you through getting sorted with anything" and also "hi if you're just trying to modify cactbot for your own use and you aren't a programmer, probably you should use this other thing because we don't have time to support teaching every cactbot user how to use git and npm".  I feel like there's also an implied "if you know what you're doing, do whatever's easiest" which I don't feel like I need to say.  It's hard to emit both the friendliness and the warning at the same time.  

Hopefully this strikes the right balance.

PR #2323, opened 2020-12-28
fix: adding & corrections in outputs

MaikoTan:
@ShadyWhite @r4ha Please review this.


resources/outputs.js
```diff
@@ -88,7 +88,7 @@ const Outputs = new ThrowOnInvalidProxy({
     de: 'Tank buster',
     fr: 'Tank busters',
     ja: 'タンクバスター',
-    cn: '死刑',
+    cn: '坦克死刑',
```
MaikoTan, line 91: These are multiple tankbusters, but there are no plural form in JA/CN, how can we call them?

resources/outputs.js
```diff
@@ -145,7 +145,7 @@ const Outputs = new ThrowOnInvalidProxy({
     en: 'Stack',
     de: 'Sammeln',
     fr: 'Packez-vous',
-    ja: '頭割り',
+    ja: 'スタック',
     cn: '分摊',
```
MaikoTan, line 149: This is stack for other things instead of damage share, how can we call them?

RE: Should we be consistent in JA to call all the stack gimmick as スタック or something?

resources/outputs.js
```diff
@@ -225,15 +225,15 @@ const Outputs = new ThrowOnInvalidProxy({
     en: 'Look Towards Boss',
     de: 'Anschauen Boss',
     fr: 'Regardez le boss',
-    ja: '見る',
-    cn: '背对',
+    ja: 'ボスを見る',
+    cn: '正对',
```
MaikoTan, line 229: Maybe there are a better translation in CN?

MaikoTan:
By the way, @quisquous there are directions like north/south/east/west in outputs, but no northeast/southwest, could I add them in?

resources/outputs.js
r4ha: 頭割り is popular in Japanese.

MaikoTan: You might misunderstand me.
https://github.com/quisquous/cactbot/blob/126e5cf537d2d8c97ad26afaa9ee3bd84419a12e/resources/outputs.js#L134-L142
Here is the damage share, we call it 頭割り.

https://github.com/quisquous/cactbot/blob/126e5cf537d2d8c97ad26afaa9ee3bd84419a12e/resources/outputs.js#L143-L151
And here is get together, not damage share I think. What can we call it?

r4ha: Umm, maybe it should follow en so we call スタック ...?

MaikoTan: That's the point that I am not sure.
In my opinion, スタック is usually used in programing field as the data structure `stack`, not for refering "getting together".
What about following cn, that they always call it as 集合? or 集まる if prefer 和語?

r4ha: 集合 is good idea!

MaikoTan: Ok, I'll change it tomorrow.
good night!

resources/outputs.js
ShadyWhite, line 229:
```suggestion
    cn: '面向',
```

ShadyWhite (changes requested):

quisquous:
> By the way, @quisquous there are directions like north/south/east/west in outputs, but no northeast/southwest, could I add them in?

Go for it!

I think it would be useful to have the full names, e.g. `northeast` but then also have something for the abbreviated `NE` version.  Maybe call those `dirNE` to avoid having single letter keys? Or, I can add them in too if you just want to do the intercardinal longer versions.

quisquous:
Thanks!!

PR #2326, opened 2020-12-29
util: convert unicode ability names into --sync--

quisquous:


quisquous:
I am a bit skeptical of this, but my Emerald Weapon normal has untranslated autos that make `make_timeline` complain when trying to print them.  I could change print to use `.encode("utf-8")` but then I can't print to the command line without the newlines being escaped.

Maybe there's some better way to do this? It does seem a bit risky to just assume no unicode chars.

JLGarber:
Huh. I didn't even notice this, it looks like it's terminal-specific? At least on my Windows 10 machine, Powershell seems to display it correctly, `cmd` can't display the characters but doesn't throw errors, and Git Bash just stops and complains.

I'm *slightly* concerned about silently swallowing anything non-ASCII, but given that to date it hasn't been an issue, it's *probably* okay?

quisquous:
I think this seems like the wrong approach.  I'll try a different shell if I see this again.

PR #2171, opened 2020-12-12
raidboss: Add E9S Summon

panicstevenson:

PR #2329, opened 2020-12-30
Raidboss(E11S): Add Prismatic Invincible Safe Zone Alert

MaikoTan:
Learnt from #2171 
Untested for now, I'll continue working on this soon.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e11s.js
+++ b/ui/raidboss/data/05-shb/raid/e11s.js
@@ -654,7 +654,7 @@ export default {
     },
     {
       id: 'E11S Prismatic Deception',
-      netRegex: NetRegexes.gainsEffect({ effectId: '655' }),
+      netRegex: NetRegexes.gainsEffect({ effectId: '655', capture: false }),
       promise: async (data) => {
         const fateBreakersImageLocaleNames = {
           en: 'Fatebreaker\'s image',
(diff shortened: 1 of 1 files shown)


quisquous:
So, I talked with @JLGarber about this a while back, and I am actually a little bit hesitant about adding this trigger.  I feel like there is a lot of anger over cactbot knowing things that players can't know, like which square exactly e4s titan fists hit (even though it doesn't matter, imo) and calling out the TEA alpha and beta phases super early before anybody would know what was on them.

I am trying to walk a line where cactbot tells you stuff that you could ostensibly know by looking at the screen or memorizing a fight, and not tell you stuff that you couldn't have known at all.  I feel like something like this is maybe a bit over the line community-wise and is maybe better for a personal trigger?

MaikoTan:
Ok, I'll close it.

quisquous:
I'm sorry to be opinionated on this one.  I think it's a great idea and I would definitely be interested in seeing the final version once you get it tested, I just worry a lot about the general community reaction of having something called out that is so obviously not something any one player can know.  Sorry!

mooondark:
@MaikoTan,
Is there a way to test it as a user trigger ?

MaikoTan:
> @MaikoTan,
> Is there a way to test it as a user trigger ?

You can checkout [this doc](https://github.com/quisquous/cactbot/blob/main/docs/CactbotCustomization.md)

mooondark:
I tried but where to put the calculatePrismaticSafeZone const ? If I place it with the other const it throws an error.
(sorry I'm just a translator not a dev 😔)

PR #2331, opened 2020-12-30
build: split raidboss/oopsy data out

MaikoTan:


quisquous:
Does this work for you locally? This has the same issue that #2293 had for me, in that if I use this branch, restart webpack dev server via `npm start`, then raidboss and oopsy don't work at all.  There's no errors, it's more that the files are not loaded at all.  If I look in the inspector, there are no files loaded from the dev server.

main:
image

patch:
image

MaikoTan:
Looks like that we should have a explicit `import()` or `require()` so this approach would work.
Another way is to add a `script` label in html file to load these js?

quisquous:
> Looks like that we should have a explicit `import()` or `require()` so this approach would work.
> Another way is to add a `script` label in html file to load these js?

At the moment, I don't think it's possible to add an `import` or `require` from the main html page.  This is so that cactbot can continue being loaded via `file://` urls, which doesn't support ESM.

Is there another place that the `import` can be placed?

MaikoTan:
Sorry I found it did even not loaded the `raidboss.js`, besides, webpack complains that `raidboss.js` didn't export anything, so cause the problem.

ui/config/config.html
```diff
@@ -4,6 +4,7 @@
   <meta charset="utf-8" />
   <link rel="stylesheet" href="../../resources/defaults.css">
   <link rel="stylesheet" type="text/css" href="config.css">
+  <script src="../../dist/raidboss_data.bundle.js"></script>
```
quisquous, line 7:
```suggestion
  <script src="../../dist/raidboss_data.bundle.js"></script>
  <script src="../../dist/oopsyraidsy_data.bundle.js"></script>
```

quisquous (commented):
Re: HtmlWebpackPlugin.  I think that seems reasonable, if you want to send a PR for it.  I think the only thing I'd want to make sure is that we keep all the `raidboss_alerts_only.html` redirect htmls working.

I think having something that generates the includes automatically will be much better than manual, as if you forget any shared script then nothing loads at all, and we have no testing for that.

ui/config/config.html
quisquous: I think with this change, this is ready to merge, in my opinion.

MaikoTan:
> Re: HtmlWebpackPlugin.  I think that seems reasonable, if you want to send a PR for it.  I think the only thing I'd want to make sure is that we keep all the `raidboss_alerts_only.html` redirect htmls working.

Then we need an extra source directory to put the "template html", if we want to make sure all the html url would not change.
i.e: `src/pages/raidboss.html -> ui/raidboss/raidboss.html`

PR #2344, opened 2021-01-02
config: show user triggers in the ui

### LINKED ISSUE #2083: config: add user triggers to raidboss section
Mentioned [in discord](https://discordapp.com/channels/551474815727304704/551476873717088279/782651950079672331).

I think it might be nice to be able to add some configuration for user triggers to the ui.  I think there's an argument that if you're writing user triggers you can just...edit the trigger yourself with no need for the ui, but:

* in the future with #860 maybe people will start distributing triggers more
* some people are worried about whether they've written their triggers correctly, and so having a "check the UI to see if they're there for the right zones" would at least be a quick smoke check

quisquous:
Fixes #2083.

User triggers now show up in the ui.  Each user-defined js file is shown
in its own section as a peer to each expansion section.  Within that
section, each trigger set is given its own subsection even if multiple
trigger sets apply to the same zone.

Like cactbot triggers, user-defined triggers must use output strings in
order to have configurable output text, but other options like duration
and tts vs text can still be configured as long as the trigger has an
id.

This pr handles deprecated-but-still-supported-in-the-wild user
features like zoneRegex and missing trigger ids.  It's nice that the
code to evaluate trigger text is still there, as it applies to
user triggers without output strings.

Triggers with missing ids can't be configured, because all config
options are handled via globally unique ids (whoops?).  This probably
also means that such triggers don't handle suppressSeconds either.
Maybe we should think about using ids for overriding but some other key
for suppressSeconds/config uniqueness.

### DIFF
--- a/resources/user_config.js
+++ b/resources/user_config.js
@@ -127,6 +127,10 @@ class UserConfig {
       reloadOnce();
     });
 
+    this.loadUserFiles(overlayName, options, callback);
+  }
+
+  loadUserFiles(overlayName, options, callback) {
     const readOptions = callOverlayHandler({
       call: 'cactbotLoadData',
       overlay: 'options',
--- a/ui/config/config.js
+++ b/ui/config/config.js
@@ -1,3 +1,4 @@
+import Regexes from '../../resources/regexes.js';
 import UserConfig from '../../resources/user_config.js';
 import ZoneInfo from '../../resources/zone_info.js';
 import contentList from '../../resources/content_list.js';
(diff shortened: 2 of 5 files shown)


quisquous:
Here's a brief example from my mess of user trigger testing: 

image

panicstevenson (approved):
LGTM; only thing (as a followup, probably) might be to add like a tooltip or something, but I don't really think this happens by accident

quisquous:
> LGTM; only thing (as a followup, probably) might be to add like a tooltip or something, but I don't really think this happens by accident

Sorry, a tooltip for what?

PR #2356, opened 2021-01-07
Update NARI DirectorUpdate docs URL to new repo location

ayyaruq:
What it says on the tin, NARI repo was moved to an org account. The repo itself will auto redirect but GitHub Pages doesn't so here we are.


quisquous:
Thanks for the update!

PR #2357, opened 2021-01-07
raiboss: e12s oracle prevent adv yellow tethers check trigger during basic

Legends0:
After reviewing some OverlayPlugin debug output, I noticed E12S Adv Relativity Hourglass Collect Yellow Tethers errors like TypeError: Cannot read property '...' of undefined (Source: webpack://cactbot/./ui/raidboss/popup-text.js?, Line: 84). This happened to be caused by the tethers appearing in Basic Relativity that I overlooked which activate the trigger.


quisquous:
Thanks!

PR #2362, opened 2021-01-08
build: webpack timeline loader

MaikoTan:
A too-young-too-simple timeline loader who remove comments and blank lines only.

### DIFF
--- a/webpack/loaders/timeline-loader.cjs
+++ b/webpack/loaders/timeline-loader.cjs
@@ -1,6 +1,6 @@
 'use strict';
 
-const commentRegex = /(?:^(?:[^"]*)#.*$|(?<=^[^"]*"[^"]*".*)#.*$)/i;
+const commentRegex = /(?<=^(?:[^"/]*(?:|"[^"]*"))[^"/]*(?:|sync\s*\/[^/]*\/[^"/]*))#.*$/i;
 
 module.exports = function(content, map, meta) {
   let ret = '';
--- a/webpack/loaders/timeline-loader.cjs
+++ b/webpack/loaders/timeline-loader.cjs
@@ -1,0 +1,15 @@
+'use strict';
+
+module.exports = function(content, map, meta) {
+  let ret = '';
+
+  content.split(/\r?\n/).forEach((_line) => {
+    let line = _line.trim();
+    if (!line || line.startsWith('#'))
+      return;
+    line = line.replace(/#.*$/, '').trim();
+    ret += line + '\r\n';
+  });
+
+  return ret;
+};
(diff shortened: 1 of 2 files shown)


webpack/loaders/timeline-loader.cjs
```diff
@@ -0,0 +1,15 @@
...
+
+module.exports = function(content, map, meta) {
+  let ret = '';
+
+  content.split(/\r?\n/).forEach((_line) => {
+    let line = _line.trim();
+    if (!line || line.startsWith('#'))
+      return;
+    line = line.replace(/#.*$/, '').trim();
+    ret += line + '\r\n';
```
quisquous, line 11: What about removing lines that are only comments but `#` doesn't start at the beginning of the line too?

```suggestion
    let line = _line.replace(/#.*$/, '').trim();
    if (!line)
      return;
    ret += line + '\r\n';
```

But more importantly, either of these will break `100.0 "Some ability #1" sync /stuff/`.  Can you fix the regex to handle # inside of quotes as well?

MaikoTan: Oh, yes.
I gonna fix it soon.

MaikoTan: How about this regex?
https://regex101.com/r/CkKpW6/1

The only issue is when comment contain one quote (") would causing regex unmatched.

quisquous: https://regex101.com/r/FwTpMe/2 maybe?

MaikoTan: lookbehind sounds good, but is it possible that '#' in `sync /regex/` things?

quisquous: Oh, that's quite right!

MaikoTan: Looks like javascript regex doesn't offer something like [`(?(exp)yes|no)`](https://docs.microsoft.com/en-gb/dotnet/standard/base-types/regular-expression-language-quick-reference?redirectedfrom=MSDN#alternation-constructs), so don't know how to capture the `sync /regex/` part as well.

quisquous: Hmm, how about: https://regex101.com/r/FwTpMe/3

MaikoTan: I tried several times and it works well, thanks!

quisquous: Wait sorry: https://regex101.com/r/FwTpMe/4

One character change to support slashes in the ability name!

quisquous: Haha, I see that you literally just updated this as I had this thought.

PR #2366, opened 2021-01-08
refactor: simplify crafting detection

MaikoTan:
via array util functions


quisquous:
Thanks for the cleanup!

Issue #751, opened 2019-11-13
oopsy: implement suppressSeconds


quisquous:
Right now this doesn't exist, but it'd be super useful if it worked like it did in raidboss.

PR #2369, opened 2021-01-10
oopsy: add e12s chiseled sculpture lasers

### LINKED ISSUE #2343: oopsy: e12s door boss lasers should tell you whose laser you got hit by
We could track headmarkers and who they were tethered to, and then that would let us know which number went awry.

Just wanted to write this idea down before I forgot.

quisquous:
Fixes #2343.


quisquous:
This is a first cut, untested.  Unfortunately, I think I need to implement #751 so that I can use `suppressSeconds` to count the lasers.

PR #2373, opened 2021-01-12
raidboss: v5.3 Korean

Jaehyuk-Lee:
Checking if everything is fine. I'll open PR before Parser update comes out.


Jaehyuk-Lee:
okay, I'll make a new one.

PR #2376, opened 2021-01-12
add effect id of Sheltron of PLD

Trim21:


quisquous:
Hi! If you'd like to add this, can you also rerun that script and commit the changes it generates so that the `resources/` file is also updated?

Trim21:
done

PR #2380, opened 2021-01-12
add timeline trigger for UWU

Trim21:


Trim21:
Copy from titan-hm, maybe it's better to response tankCleave for all condition?

quisquous:
> Copy from titan-hm, maybe it's better to response tankCleave for all condition?

Yeah, I think `tankCleave` in both cases seems good to me.

Trim21:
there is another question I asked in last UWU trigger PR

> @quisquous I have a question about this. There is another groud effect ability(the laser in Ultimate Suppression, I don't know it's name in English) in UWU, and it's random, chose 1 from 3 choies(which can be distinguished by different ability id). It's not that visible like Land Weight and happened in the same time with another groud effect. Could I issue a pr for this?


> image

quisquous:
> there is another question I asked in last UWU trigger PR
> 
> > @quisquous I have a question about this. There is another groud effect ability(the laser in Ultimate Suppression, I don't know it's name in English) in UWU, and it's random, chose 1 from 3 choies(which can be distinguished by different ability id). It's not that visible like Land Weight and happened in the same time with another groud effect. Could I issue a pr for this?
> 
> > image
> 
> And I can add it in this PR.

This is the "Aetherochemical Laser" trident coming from Ultima? Although I don't think that has three abilities? Can you give me the ability ids?

Trim21:
@quisquous yes, I'll provide the trigger and raw log later.

PR #2381, opened 2021-01-12
raidboss: update Korean v5.3

Jaehyuk-Lee:
I was doing v5.3 translation slowly, but Akurosia never let me to.. 😅
just joking 😃

maybe I should translate those Korean-knowledge people only translatable things first, then other obvious things next time.


quisquous:
Thanks for merging your changes in on top of that!

PR #2384, opened 2021-01-13
plugin: re-add shield values for 5.41

### LINKED ISSUE #2107: plugin: find new shield value offset for 5.4
When updating the plugin in aeb55b2945e I couldn't find the shield percentage any more.  If anybody with cheat engine expertise has time to look into this, it'd be much appreciated.

KattTails:
Re-enables the shield percentage since it seems to be back in patch 5.41. 
It feels a bit messy to copy such a large amount of memory for just one byte so far away from the start but I guess it's fine?

Thanks to @AmenneHolelane in https://github.com/quisquous/cactbot/issues/2107#issuecomment-758983952 (and Pohky on the goat place discord!) for finding the value.

Fixes #2107


quisquous:
Thank you!!! This is much appreciated.  <3

PR #2385, opened 2021-01-13
plugin, jobs: v5.3 Korean server update

Jaehyuk-Lee:
I guess this will work, didn't really tried it though since the parser is not out yet.

### DIFF
--- a/resources/translations.js
+++ b/resources/translations.js
@@ -49,8 +49,6 @@ export const LocaleNetRegex = {};
     },
     // Recipe name always start with \ue0bb
     // HQ icon is \ue03c
-    // TODO:KO have not launched patch 5.3 yet, so they do not have trial synthesis.
-    // Translation will be added once launch.
     craftingStart: {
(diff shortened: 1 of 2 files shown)


resources/translations.js
```diff
@@ -65,6 +63,7 @@ export const LocaleNetRegex = {};
       fr: 'Vous commencez une synthèse d\'essai pour une? \ue0bb(?<recipe>.*)\\.',
       ja: '(?<player>\\y{Name})は\ue0bb(?<recipe>.*)の製作練習を開始した。',
       cn: '(?<player>\\y{Name})开始练习制作\ue0bb(?<recipe>.*)。',
+      ko: '\ue0bb(?<recipe>.*) 제작 연습을 시작합니다.',
```
MaikoTan, line 66: the last char, `\\.` instead of `.` ?

Jaehyuk-Lee:
updated v5.3 memory signature things here, the latest release is not working with Korean server.

PR #2390, opened 2021-01-13
update latest data (generated-automatically)

Jaehyuk-Lee:
I found out how to solve saintcoinach problem, thought the game version number should be same as latest patch date, but it was actually needed to be just same as saintcoinach showed on manual execute.
Was written in `/util/readme.md` though.


quisquous:
Thank you! I'm glad you got SaintCoinach fixed.  (I think you wrote that doc too?)

Jaehyuk-Lee:
I remember I write it, but forgot the information. :(

PR #2392, opened 2021-01-15
Raidboss: Make Japanese translations more natural (STORMBLOOD)

nekomimi-mode:
Fixing Japanese translations of Raidboss timeline to make more natural. At this moment, most of "stormblood" contents only.


ui/raidboss/data/04-sb/raid/o3n.js
```diff
@@ -331,7 +331,7 @@ export default {
         text: {
           en: 'Mindjack: Left',
           de: 'Geistlenkung: Links',
-          ja: 'マインドジャック: 左折',
+          ja: 'マインドジャック: 左方向',
```
MaikoTan, line 334: "左折" and "右折" is the original word in-game, do you think it is better to translate it?

ui/raidboss/data/04-sb/raid/o8s.js
```diff
@@ -188,7 +188,7 @@ export default {
           en: 'Future: Stack and Through',
           de: 'Zukunft: Sammeln und Durchlaufen',
           fr: 'Futur : Stack et traversez',
-          ja: '未来: シェア後ボス通り抜ける',
+          ja: '未来: シェア後ボスを通り抜ける',
```
MaikoTan, line 191:
```suggestion
          ja: '未来: 頭割り後ボスを通り抜ける',
```
I think to describe "damage share" gimmick, we don't use "シェア" anymore, all use "頭割り" instead.

MaikoTan (approved):

ui/raidboss/data/04-sb/raid/o3n.js
nekomimi-mode: I've reverted them.
But I think both are fine. Furthermore, only "左" or "右" is enough for this gimmick.

MaikoTan: Yep, I agree that.
For example they wrote only "左" and "右" in cn things.

I am just thinking to use the same word in-game would be more accurate.

PR #2402, opened 2021-01-17
raidboss: support English timelineReplace

### LINKED ISSUE #2146: Wrong common translation works
## Description

Some timelines are not translated correctly.
Happens in e9s

## Additional information

`S` does not mean `South` but it translates to south in each languages.
스크린샷(3)

quisquous:
Fixes #2146.

In the future, rather than abbreviating abilities in the timeline
itself, we should keep all abilities as their full names, and
then abbreviate in English via timelineReplace.

### DIFF
--- a/test/timeline/test_timeline.js
+++ b/test/timeline/test_timeline.js
@@ -170,8 +170,13 @@ const tests = {
       if (!locale)
         continue;
       // English is a partial translation.
-      if (locale === 'en')
+      if (locale === 'en') {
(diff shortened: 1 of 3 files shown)


quisquous:
cc @Akurosia in case this breaks your scripts, and cc @JLGarber who has asked for this feature

JLGarber (approved):

Akurosia:
thanks for letting me know. should not break anything on my side

PR #2407, opened 2021-01-20
raidboss: e12s oracle basic relativity shadoweye target callouts

Legends0:
These will call out "Eye on you" similar to e2s, and then later call out look away from player 1 and player 2, or the other player (if there).

Not sure if we want to keep the older one as an option, but this one would work well for when there is a non-standard comp.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e12s.js
+++ b/ui/raidboss/data/05-shb/raid/e12s.js
@@ -860,22 +860,72 @@ export default {
       },
     },
     {
-      id: 'E12S Basic Relativity Shadoweye',
+      id: 'E12S Oracle Basic Relativity Shadow Eye Me',
+      netRegex: NetRegexes.gainsEffect({ effectId: '998' }),
+      condition: (data, matches) => data.phase === 'basic',
+      preRun: (data, matches) => {
+        data.eyes = data.eyes || [];
+        data.eyes.push(matches.target);
+      },
+      // Delay so that we don't overwrite Hourglass callout
+      delaySeconds: 4,
+      infoText: (data, matches, output) => {
+        // The trigger repeats for collection; only call out first match
+        if (data.eyes.includes(data.me) && !data.eyecalled) {
+          data.eyecalled = true;
+          return output.text();
+        }
+        return;
+      },
+      outputStrings: {
+        text: {
+          en: 'Eye on YOU',
+          de: 'Auge auf DIR',
+          fr: 'Œil sur VOUS',
+          ja: '自分に目',
+          cn: '石化眼点名',
+          ko: '시선징 대상자',
+        },
+      },
+    },
+    {
+      id: 'E12S Oracle Basic Relativity Shadow Eye Other',
       netRegex: NetRegexes.gainsEffect({ effectId: '998' }),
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -860,22 +860,72 @@ export default {
...
+      },
+      outputStrings: {
+        lookAwayFromPlayers: {
+          en: 'Look Away from ${player1} and ${player2}',
+          // TODO: Verify these localizations:
+          // de: 'Schau weg von ${player1} und ${player2}',
+          // fr: 'Ne regardez pas ${player1} et ${player2}',
+          // ja: '${player1}と${player2}を見ない',
+          // cn: '背对${player1}和${player2}',
+          // ko: '${player1}와 ${player2}에게서 뒤돌기',
```
MaikoTan, line 926: ja and cn is good.

Legends0: Thanks

quisquous:
Is there a strategy that people are using where this is useful? My group uses a strategy where healers and tanks do literally the same thing regardless of what debuff they have, and everybody is naturally looking away.  Even "look away" is a bit overkill.

Mostly just trying to understand what other people are using where this might help vs if this is just extra noise.

Legends0:
> 
> 
> Is there a strategy that people are using where this is useful? My group uses a strategy where healers and tanks do the same thing regardless of what debuff they have, and everybody is naturally looking away. Even "look away" is a bit overkill.
> 
> Mostly just trying to understand what other people are using where this might help vs if this is just extra noise.

The "look away" trigger was helpful for learning the mechanic. I have seen variations in which roles soak which hourglass, but "look away" trigger is a bit overkill for the leading pug strategy (TPS).

I am not certain on the perspective of the tanks, but as a healer, once the strategy is perfected, the "look away" serves as a reminder to adjust looking slightly outward if not already. I could see the change being better for tanks if it just said look away from tanks/healers or as is here in the PR as it may give more of a indicator as to which direction to look if their positioning happens to be off.

Outside of solo healer strategies, this isn't particularly more useful than the original: it adds one additional alert if targeted, and extends the original alert with more details. I'm not sure what strategy the current solo healer kills have used. Likely when a healer is targeted with gaze, they have the other person play the role of the healer and would thus change their positioning, in which case the Eye on You callout has benefits.

Would the PR here benefit more with an option for if Eye on You should be enabled and/or if Look Away should use names?

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -860,22 +860,72 @@ export default {
...
+          return output.lookAwayFromPlayer({ player: data.ShortName(player1) });
+        }
+
+        // Return empty when only you have eye
+        return;
+      },
+      outputStrings: {
+        lookAwayFromPlayers: {
+          en: 'Look Away from ${player1} and ${player2}',
+          // TODO: Verify these localizations:
```
quisquous, line 921: In general you can leave translations out with no comment, and the `util/find_missing_translations.py` script will point out that these are missing to anybody looking.

PR #2408, opened 2021-01-21
oopsy: add damage down to e10s oopsy

quisquous:


quisquous:
@JLGarber is this too noisy? I think there's some parts of damage down (e.g. messing up tethers or not having enough people in the orb stack) which otherwise won't show up.  Also, feel free to wordsmith "partial stack", as I'm still iffy on it.

JLGarber:
I don't think we have a better option on wording for orb failures, no. 

I don't have time to look through my logs right now, but is there literally nothing for Shackle failure, or just "this is very new stuff we haven't tried before"? I'm thinking here of something like flags on zero-damage attacks or something similar.

On the tower failures, what about tracking whether players have/don't have the Bleed effect at the moment towers resolve? That would allow for exclusion of towers as a failure source, at least.

quisquous:
Yeah, I did not see anything for shackle failure.  Not a tether change, not anything else, just a damage down after the debuff expired.

For towers, I thought about bleed, but that doesn't solve shadow clones in towers.  I guess we could still blame people not in towers, but...

JLGarber:
Mm, okay, that being the case, it's probably okay as it is. It might be noisy on reports, but for real-time "wait, what happened?" situations it should be fine.

PR #2413, opened 2021-01-23
feat(coverage): keep table header on top

MaikoTan:
via adding a tricky sticky css rule.

### DIFF
--- a/util/coverage/coverage.css
+++ b/util/coverage/coverage.css
@@ -1,5 +1,6 @@
 body {
   margin: 20px;
+  padding-top: 2em;
 }
 
 div {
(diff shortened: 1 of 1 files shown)


quisquous:
I like this idea a lot, but now the header is on top of the text:
image

Is there some way in CSS to make this not happen and to create some space for the sticky table header?

MaikoTan:
Maybe we can add a background color to the headers?

quisquous:
> Maybe we can add a background color to the headers?

Sounds reasonable to me! It could just be the white background color of the page, and then it would at least be opaque.

MaikoTan:
image
Maybe better?

quisquous:
Looks good to me!

PR #2416, opened 2021-01-27
oopsy: modify triggers to handle shielded mechanics

quisquous:


quisquous:
cc @JLGarber 

This does make me wonder if we should treat "does >0 damage" as something we do for all triggers? We could separate out into `abilityWarn` vs `damageWarn` if that was an important distinction, but I am trying to think if there are issues that this could cause.

JLGarber:
I think you're right on "did it do damage". If nothing else, we can move to separating them as you said, and now that we have `matches` elements easily available in oopsy functions, it should be a lot easier to work with flags if that's the easiest way to tell things.

quisquous:
So, I had a contrary thought here, which is ... ignoring these mechanics where people shield to take zero damage, what if somebody just had a lot of shielding but still stood in something (that then stripped their shield before the real mechanic).  That still seems like a mistake?  I don't know.  I think I'll keep thinking on it.

JLGarber:
Hrrm, that's a good point. I'll think on it as well.

PR #2429, opened 2021-02-02
Raidboss: O4S timeline refresh

JLGarber:
In light of BLU updates, quisquous and I thought it would be best to modernize the outstanding Deltascape timelines. Many thanks to Shasta for the initial Very Definitely Good Enough timeline that's only now being replaced.

This timeline is unfortunately not as accurate as I should like. There are a lot of places where timing is off by 0.05-0.15 seconds. This currently is rather difficult to fix because I don't have any logs more recent than mid-2019. (I'd rather not hunt-and-peck replace individual times without having modern logs to compare against.) I think we can let this ride for a week or so and then make adjustments as necessary.

Neither part of the encounter uses an `Engage!` sync, since in two-part encounters that's ambiguous. Fortunately, even though Neo doesn't have autos, the `startsCasting` line for Almagest is just about at the same place as an auto *would* be, so we lose nothing here.

We'll very likely have to make some adjustments to triggers to accommodate BLU, but it's probably best to hold on that until we see patch notes.

### DIFF
--- a/ui/raidboss/data/04-sb/raid/o4s.txt
+++ b/ui/raidboss/data/04-sb/raid/o4s.txt
@@ -1,92 +1,79 @@
 # Omega - Deltascape V4.0 (Savage) - O4S
-# Created by Shasta Kota of Death & Taxes (DnT) on Gilgamesh
-# Shasta's reddit: /u/shastaxc
-# Last Updated: 08/01/2017
 
-#############################################################
-########CUSTOMIZE your timeline. Remove the hashtag from
-########the beginning of the following lines to prevent
-########them from appearing on your timeline.
+hideall "--Reset--"
+hideall "--sync--"
 
-#hideall "--Reset--"
-#########DPS CAN HIDE THESE
-
-
-#########HEALERS CAN HIDE THESE
-
-
-#########TANKS CAN HIDE THESE
-
-##############################################################
-################## Windows 8 & 10 Voices ##################
-########REMOVE THE HASTAG to select a voice
-#define speaker "voice" "Microsoft Zira Desktop" 0 100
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/04-sb/raid/o4s.txt
```diff
@@ -1,211 +1,212 @@
...
+181.5 "--untargetable--"
+
+185.3 "--sync--" sync /:Exdeath:240A:/
+186.4 "Collision" sync /:Exdeath:2409:/
+188.7 "Holy" sync /:Exdeath:2403:/
+192.4 "Zombie Breath" sync /:Exdeath:240B:/
+197.1 "Flare" sync /:Exdeath:2401:/
+
+200.0 "--targetable--"
+203.9 "Blizzard III/Fire III/Thunder III" sync /:Exdeath:23F[BCD]:/
```
quisquous, line 60: Can you add an English translation to make this `Blizzard/Fire/Thunder III` since this is pretty long?

quisquous (commented):
Overall looks good, thanks!

Is there any `--adds targetable--` line that can be added for the neo meteors while you're here?

quisquous:
> Are there -p and -ii for neo?

Oh, I also meant to follow up here and ask if there is a different `-p` and `-ii` for neo, or is it combined in exfaust?

JLGarber:
Oops, yes, I tend to be bad about remembering the phase flag. I don't think it would be terribly meaningful for Exdeath since it's under 5 minutes, but I've added in phases corresponding to the first Grand Cross Delta, Grand Cross Omega, and Neverwhere. As for the ignored abilities, I separated those out, good catch.

PR #2430, opened 2021-02-02
plugin: update Bozja CE Opcode for global 5.45

KattTails:
Untested, will check when patch goes live as I'll be doing the new content!


quisquous:
Thank you for keeping this up to date! It's a little bit unfortunate how often this changes, vs the Eureka ones.

This looks like it works to me, so merging.

KattTails:
The Eureka ones change too, the opcode is just pulled from the ACT plugin. The CE opcode has no use for the ACT plugin so Ravahn doesn't track it.

PR #2443, opened 2021-02-05
e12s - Add Chiseled Sculpture triggers

valarnin:
This PR adds Chiseled Sculpture triggers for when to bait which statue.

You'll need a bunch of the raidemulator fixes I've been working on to test this in raidemulator. I've compiled them into a patch which can be applied with `git apply patch.patch`, rather than muddying this PR with those fixes:

[patch.zip](https://github.com/quisquous/cactbot/files/5929390/patch.zip)

This patch includes the two raidemulator PRs that are already open, as well as some other minor fixes to raidemulator.

I haven't cleaned up the comments related to the statues yet, wanted to get the code/logic up for review.

An additional patch for the bug with suppressSeconds which I forgot to include in the original patches:

```patch
diff --git a/ui/raidboss/emulator/data/PopupTextAnalysis.js b/ui/raidboss/emulator/data/PopupTextAnalysis.js
index 0f0e6db6..1b0e6182 100644
--- a/ui/raidboss/emulator/data/PopupTextAnalysis.js
+++ b/ui/raidboss/emulator/data/PopupTextAnalysis.js
@@ -88,7 +88,7 @@ export default class PopupTextAnalysis extends StubbedPopupText {

   _onTriggerInternalCheckSuppressed(trigger, when) {
     this.currentFunction = '_onTriggerInternalCheckSuppressed';
-    const ret = super._onTriggerInternalCheckSuppressed(trigger, when);
+    const ret = super._onTriggerInternalCheckSuppressed(trigger, this.realNow);
     this.currentTriggerStatus.suppressed = ret;
     if (ret) {
       this.delayResolver();
diff --git a/ui/raidboss/emulator/overrides/StubbedPopupText.js b/ui/raidboss/emulator/overrides/StubbedPopupText.js
index 8640a802..471b9f2e 100644
--- a/ui/raidboss/emulator/overrides/StubbedPopupText.js
+++ b/ui/raidboss/emulator/overrides/StubbedPopupText.js
@@ -7,6 +7,22 @@ export default class StubbedPopupText extends PopupText {

   // Stubbed, we don't want overlay hooks
   HookOverlays() { }
+
+  // Use the current emulator timestamp unless we're using an older timestamp
+  _onTriggerInternalCheckSuppressed(trigger, when) {
+    when = when > ((+new Date()) - 500) ?
+      window.raidEmulator.emulator.currentEncounter.encounter.logLines[
+        window.raidEmulator.emulator.currentLogLineIndex
+      ].timestamp : when;
+    return super._onTriggerInternalCheckSuppressed(trigger, when);
+  }
+
+  // Override the `.now` property to use our internal tracked timestamp
+  _onTriggerInternalGetHelper(trigger, matches, now) {
+    const helper = super._onTriggerInternalGetHelper(trigger, matches, now);
+    helper.now = window.raidEmulator.emulator.lastLogTimestamp;
+    return helper;
+  }
 }

 StubbedPopupText.globalTriggerSets = null;
```

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e12s.js
+++ b/ui/raidboss/data/05-shb/raid/e12s.js
@@ -136,6 +136,37 @@ const primalOutputStrings = {
   },
 };
 
+const statueBaitResponse = (data, _, output) => {
+  // cactbot-builtin-response
+  output.responseOutputStrings = {
+    inner: {
+      en: 'Bait Inner Statue',
+    },
+    outer: {
+      en: 'Bait Outer Statue',
+    },
+    unknown: {
+      en: '???',
+    },
+    sides: Outputs.sides,
+  };
+  data.statueLaserCount = data.statueLaserCount + 1 || 1;
+  let response;
+  if (data.statueLaserCount === data.statueTetherNumber)
+    response = { alertText: output[data.statueDir]() };
+  else
+    response = { infoText: output.sides() };
+
+  if (data.statueLaserCount >= 4) {
+    delete data.statueTetherNumber;
+    delete data.statueLaserCount;
+    delete data.statueLastId;
+    delete data.statues;
+  }
+
+  return response;
+};
+
 // Due to changes introduced in patch 5.2, overhead markers now have a random offset
 // added to their ID. This offset currently appears to be set per instance, so
 // we can determine what it is from the first overhead marker we see.
@@ -400,6 +431,8 @@ export default {
       id: 'E12S Promise Chiseled Sculpture',
       netRegex: NetRegexes.headMarker({}),
       condition: (data, matches) => data.isDoorBoss && matches.target === data.me,
+      // Show headmarker number for whole phase
+      durationSeconds: 20,
       response: (data, matches, output) => {
         // cactbot-builtin-response
         output.responseOutputStrings = {
@@ -421,10 +454,9 @@ export default {
         const firstLaserMarker = '0091';
         const lastLaserMarker = '0098';
         if (id >= firstLaserMarker && id <= lastLaserMarker) {
-          // We could arguably tell you which giant you're tethered to by finding their position?
-          // And then saying something like "North" but that's probably more confusing than helpful.
           // ids are sequential: #1 square, #2 square, #3 square, #4 square, #1 triangle etc
           const decOffset = parseInt(id, 16) - parseInt(firstLaserMarker, 16);
+          data.statueTetherNumber = (decOffset % 4) + 1;
           return {
             alertText: [
               output.square1(),
@@ -439,6 +471,108 @@ export default {
           };
         }
       },
+    },
+    {
+      id: 'E12S Promise Chiseled Sculpture Tether',
+      netRegex: NetRegexes.tether({ target: 'Chiseled Sculpture', id: '0011' }),
+      durationSeconds: 20,
+      promise: async (data, matches) => {
+        // Calculate distance to center to determine inner vs outer
+        data.statues = data.statues || [];
+        if (data.statues.filter((s) => s.id === matches.targetId).length === 0) {
+          // Push statue to array immediately or else async causes the array
+          // to get duplicates of the statues
+          const statue = {
+            id: matches.targetId,
+          };
+          data.statues.push(statue);
+          const statueId = parseInt(matches.targetId, 16);
+          const statueData = await window.callOverlayHandler({
+            call: 'getCombatants',
+            ids: [statueId],
+          });
+
+          if (statueData === null || !statueData.combatants || !statueData.combatants.length)
+            return;
+
+          // Center = 0, -75
+          const x = statueData.combatants[0].PosX;
+          const y = statueData.combatants[0].PosY - -75;
+          statue.dist = Math.hypot(x, y);
+          if (data.statues.length > 1) {
+            // Sort so that closest statue (inner) is first
+            data.statues.sort((a, b) => a.dist - b.dist);
+          }
+        }
+      },
+      infoText: (data, matches, output) => {
+        // Player is the source of the tether for this one, oddly enough
+        if (data.me !== matches.source)
+          return;
+
+        if (data.statues.length > 0 && data.statues[0].id === matches.targetId)
+          data.statueDir = 'inner';
+        else if (data.statues.length > 1 && data.statues[1].id === matches.targetId)
+          data.statueDir = 'outer';
+        else
+          data.statueDir = 'unknown';
+
+        return output[data.statueDir]();
+      },
+      outputStrings: {
+        inner: {
+          en: 'Inner Statue',
+        },
+        outer: {
+          en: 'Outer Statue',
+        },
+        unknown: {
+          en: '???',
+        },
+      },
+    },
+    {
+      id: 'E12S Promise Palm Of Temperance SE',
+      netRegex: NetRegexes.startsUsing({ source: 'Guardian Of Eden', id: '58B4', capture: false }),
+      durationSeconds: 10,
+      infoText: (_1, _2, output) => output.knockback(),
+      outputStrings: {
+        knockback: {
+          en: 'Knockback SE',
+        },
+      },
+    },
+    {
+      id: 'E12S Promise Palm Of Temperance SW',
+      netRegex: NetRegexes.startsUsing({ source: 'Guardian Of Eden', id: '58B5', capture: false }),
+      durationSeconds: 10,
+      infoText: (_1, _2, output) => output.knockback(),
+      outputStrings: {
+        knockback: {
+          en: 'Knockback SW',
+        },
+      },
+    },
+    {
+      // Time the first dodge for lasers from the Rapturous Reach left/right split
+      id: 'E12S Promise Statue Laser First',
+      netRegex: NetRegexes.startsUsing({ source: 'Eden\'s Promise', id: '58A[DE]', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Edens Verheißung', id: '58A[DE]', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Promesse D\'Éden', id: '58A[DE]', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'プロミス・オブ・エデン', id: '58A[DE]', capture: false }),
+      // Check against `data.statueDir` to limit this trigger to just the Chiseled Sculpture phase
+      condition: (data) => data.statueDir !== undefined,
+      durationSeconds: 5,
+      response: statueBaitResponse,
+    },
+    {
+      // Time the other dodges from the previous lasers going off
+      id: 'E12S Promise Statue 2nd/3rd/4th Laser',
+      netRegex: NetRegexes.ability({ source: 'Chiseled Sculpture', id: '58B3', capture: false }),
+      condition: (data) => (console.log(data) || true) && data.statueLaserCount < 4,
+      durationSeconds: 3,
+      suppressSeconds: 1,
+      response: statueBaitResponse,
     },
     {
       id: 'E12S Promise Weight Cleanup',


valarnin:
Known issues that I'll address with the next commit (when I'm not about to go into raid):

- `Tempeance` -> `Temperance`
- Linting
- `Temperance` and `Laser First` triggers shouldn't be capturing
- `Temperance` triggers missing translations

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -440,6 +468,98 @@ export default {
         }
       },
     },
+    {
+      id: 'E12S Promise Chiseled Sculpture Tether',
+      netRegex: NetRegexes.tether({ target: 'Chiseled Sculpture', id: '0011' }),
+      promise: async (data, matches) => {
```
quisquous, line 474: Hmm, would you mind switching this to not use `getCombatants`?  You can get the positions of the statues from the `Classical Sculpture` ability: https://github.com/quisquous/cactbot/blob/main/ui/oopsyraidsy/data/05-shb/raid/e12s.js#L89-L100

I think it's more reliable long-term to do this from just log lines directly.

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -440,6 +468,98 @@ export default {
...
+        if (data.statues.length > 0 && data.statues[0].id === matches.targetId) {
+          data.statueDir = output.inner();
+        } else if (data.statues.length > 1 && data.statues[1].id === matches.targetId) {
+          data.statueDir = output.outer();
+        } else {
+          data.statueDir = '???';
+        }
+        return output.statue({ statueDir: data.statueDir });
+      },
+      outputStrings: {
```
quisquous, line 517: I think in this case where there are only two outputs, it's maybe easier for users / translators if it's just spelled out `Inner Statue` and `Outer Statue`.

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -440,6 +468,98 @@ export default {
...
+      condition: (data) => data.statueDir !== undefined && (!data.statueLaserCount || data.statueLaserCount <= 2),
+      durationSeconds: 3,
+      response: statueBaitResponse,
+      outputStrings: statueBaitOutputStrings,
+    },
+    {
+      // Time the other dodges from the previous lasers going off
+      id: 'E12S Promise Statue 2nd/3rd/4th Laser',
+      netRegex: NetRegexes.ability({ source: 'Chiseled Sculpture', id: '58B3' }),
+      // Extra condition here to handle if multiple people get clipped by the lasers
```
quisquous, line 557: I've used suppressSeconds successfully for oopsy to time lasers if that's more straightforward?

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: The `Classical Sculpture` ability always shows source entity position as center of arena (0, 75), I'm guessing because damage is triggered before the entities are moved to healer positions.

```
22|2021-01-23T23:28:38.1450000-05:00|400078F5|Chiseled Sculpture|58B2|Classical Sculpture|********|***** *****|F1730006|1F0D4001|F40E|82A0000|1B|58B28000|0|0|0|0|0|0|0|0|0|0|127297|127297|3839|10000|0|1000|0.1265127|-72.03751|75|-0.04555082|148000|148000|0|10000|0|1000|0|-75|75|-3.141593|00004D63|9e34592d7df884fbb12951a55482ff5e
```

I guess it could go based on healer position, maybe? (e.g. trigger on 58B2, early return if target isn't a healer, otherwise store it in statues array as before)

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: I tried `suppressSeconds` but it wasn't cooperating. I'm not sure if that's due to some issue on my end, some issue with raidemulator, or if it just doesn't work for this setup, and I gave up on debugging it after 20 minutes or so, just wanted to get the trigger to work.

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: `data.statueDir` is stored off for later use in the `bait` triggers. I figured making it a separate string here would allow it to be modified for both locations easily.

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: I have seen this in my logs:
```
[21:27:42.609] 16:4000DA5F:Chiseled Sculpture:58B2:Classical Sculpture:id:name:730003:6D214001:F40E:82A0000:1B:58B28000:0:0:0:0:0:0:0:0:0:0:140477:148513:10000:10000:0:1000:-0.2594604:-59.06775:75:3.117912:148000:148000:0:10000:0:1000:-0.6748441:-58.87721:75:3.099695:00004663
```

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: This is one of the things that outputstrings makes more complicated.  If you depend on the output string from one trigger in another, then it will be non-obvious to a user that they need to go change `E12S Promise Chiseled Sculpture Tether` in order to affect `E12S Promise Statue Laser First` and `E12S Promise Statue 2nd/3rd/4th Laser` text.

The usual pattern here is to store a key into outputStrings, e.g. store `'inner'` and `'outer'` in `data.statueDir` and then do `output[data.statueDir]()` .

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: I think `suppressSeconds` is a bug in raidemulator, as per https://github.com/quisquous/cactbot/issues/2161#issuecomment-743408063.

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: Might be latency-dependant or something. Looking at about 15 different pulls, 7 of them had center coordinates for me, including two pulls where one statue had center coords and the other had correct coords. These are from the log files themselves.

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: This should have been addressed in the latest updates, please confirm that I understood your intention correctly.

valarnin:
I'm not sure how to address these two test errors, given that I'd like the trigger to have translatable strings and it needs to return either info or alert text:

```
/home/runner/work/cactbot/cactbot/ui/raidboss/data/05-shb/raid/e12s.js: found both 'response' and 'outputStrings in 'E12S Promise Statue Laser First'.
/home/runner/work/cactbot/cactbot/ui/raidboss/data/05-shb/raid/e12s.js: found both 'response' and 'outputStrings in 'E12S Promise Statue 2nd/3rd/4th Laser'.
```

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: I actually had a fix for suppressSeconds but forgot to include it in the patch file, whoops. Updated the first PR comment to include that patch as well. This has been addressed in the latest commit.

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: Here's a gist of the raw numbers I was looking at:
https://gist.github.com/valarnin/5be37cbf53c153236005206fcd7fa1a6

quisquous: Ugh, that's unfortunate.  Thanks for the data.  This is another for the pile of "positional data in abilities is flaky" causing headaches.  I guess I need to fix my oopsy triggers as well to use getCombatants, but there's not support for promises there.

quisquous:
> I'm not sure how to address these two test errors, given that I'd like the trigger to have translatable strings and it needs to return either info or alert text:
> 
> ```
> /home/runner/work/cactbot/cactbot/ui/raidboss/data/05-shb/raid/e12s.js: found both 'response' and 'outputStrings in 'E12S Promise Statue Laser First'.
> /home/runner/work/cactbot/cactbot/ui/raidboss/data/05-shb/raid/e12s.js: found both 'response' and 'outputStrings in 'E12S Promise Statue 2nd/3rd/4th Laser'.
> ```

Sorry about this.  It's a bit of a weirdness around:
* built-in responses want to be able to set output strings
* I didn't want to support multiple levels of output string overriding
* therefore, if you have a response, then the response function itself must set the output strings
* (it also must have a special cactbot-builtin-response comment to verify it's safe to eval and will provide output strings so that the config UI can eval it)

Example: https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/raid/e12s.js#L278-L297

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -440,6 +474,109 @@ export default {
...
+      },
+    },
+    {
+      // Time the first dodge for lasers from the left/right split
+      id: 'E12S Promise Statue Laser First',
+      netRegex: NetRegexes.startsUsing({ source: 'Eden\'s Promise', id: '58A[DE]', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Edens Verheißung', id: '58A[DE]', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Promesse D\'Éden', id: '58A[DE]', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'プロミス・オブ・エデン', id: '58A[DE]', capture: false }),
+      condition: (data) => data.statueDir !== undefined && !data.statueLaserCount,
```
quisquous, line 565: If statueDir is undefined, could treat it as "unknown"?

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: :+1:

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: This trigger is based on the Rapturous Reach left/right that happens at the beginning. The check against `data.statueDir` is to ensure that the trigger only happens during the correct phase.

Technically the second half of the condition (`!data.statueLaserCount`) is superfluous in this case and can probably be removed, since the 4th laser should clean up all the statue-related data properties.

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: As I said before, this could probably be addressed by filtering to just the two healer positions, but given that position data is inconsistent I'm leery about trusting that as well.

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: Oh, I see!  Thanks for the explanation.  Can you leave a comment to that effect?

valarnin:
I think I've addressed everything outstanding for this PR. Please review again and let me know if there's anything else outstanding.

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -440,6 +472,108 @@ export default {
...
+      // Check against `data.statueDir` to limit this trigger to just the Chiseled Sculpture phase
+      condition: (data) => data.statueDir !== undefined,
+      durationSeconds: 5,
+      response: statueBaitResponse,
+    },
+    {
+      // Time the other dodges from the previous lasers going off
+      id: 'E12S Promise Statue 2nd/3rd/4th Laser',
+      netRegex: NetRegexes.ability({ source: 'Chiseled Sculpture', id: '58B3', capture: false }),
+      condition: (data) => (console.log(data) || true) && data.statueLaserCount < 4,
```
quisquous, line 572: Can you remove this debugging line?

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -136,6 +136,37 @@ const primalOutputStrings = {
...
+      en: '???',
+    },
+    sides: Outputs.sides,
+  };
+  data.statueLaserCount = data.statueLaserCount + 1 || 1;
+  let response;
+  if (data.statueLaserCount === data.statueTetherNumber)
+    response = { alertText: output[data.statueDir]() };
+  else
+    response = { infoText: output.sides() };
```
quisquous, line 158: I think this is the part that I'm least sure about in this patch.  Am I right in reading it will just call `Sides, Sides, Sides, Bait Inner Statue!`

I think it might be nicer to count them out and say `1, 2, 3, Bait Inner Statue`

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -400,6 +431,8 @@ export default {
       id: 'E12S Promise Chiseled Sculpture',
       netRegex: NetRegexes.headMarker({}),
       condition: (data, matches) => data.isDoorBoss && matches.target === data.me,
+      // Show headmarker number for whole phase
```
quisquous, line 434: Can this persisting number call say something like `1 (Inner)` or `4 (Outer)` instead?

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -440,6 +472,108 @@ export default {
...
+      durationSeconds: 10,
+      infoText: (_1, _2, output) => output.knockback(),
+      outputStrings: {
+        knockback: {
+          en: 'Knockback SW',
+        },
+      },
+    },
+    {
+      // Time the first dodge for lasers from the Rapturous Reach left/right split
```
quisquous, line 557: Do you think it's worth combining the haircut call with the #1 bait call? Mostly I'm trying to imagine what this would be like for tts users.

valarnin: That would require somehow suppressing the default haircut trigger for this mechanic specifically, which I'm not sure really makes sense?

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: I'm not sure exactly how to handle this given that the tether and the headmarker are two separate triggers.

Can we rely on the order of the network lines for this? Here's an example from logs from last night:

```
27|2021-02-07T19:05:12.8460000-05:00|********|Player One|0000|0000|018D|0000|0000|0000||82513ce6107aeb3d25f6a8d10425e083
35|2021-02-07T19:05:12.8460000-05:00|********|Player One|40018B3B|Chiseled Sculpture|9598|0000|0011|40018B3B|000F|0000||be10e5a16e567a8c8273a7c224d44e74
27|2021-02-07T19:05:12.8460000-05:00|********|Player Two|0000|0000|0191|0000|0000|0000||fa22088288495908ac9be5d2af60c15f
35|2021-02-07T19:05:12.8460000-05:00|********|Player Two|40018B3A|Chiseled Sculpture|9598|0000|0011|40018B3A|000F|0000||049e70d56ab785dcbf7465666766d31e
27|2021-02-07T19:05:12.8460000-05:00|********|Player Three|0000|0000|018E|0000|0000|0000||1f86664445717844cc485cfa2a85c7b4
35|2021-02-07T19:05:12.8460000-05:00|********|Player Three|40018B3B|Chiseled Sculpture|9598|0000|0011|40018B3B|000F|0000||ec7f664a7167ac7fda13756541ec29c3
27|2021-02-07T19:05:12.8460000-05:00|********|Player Four|0000|0000|0192|0000|0000|0000||1206109c2010dce5a3c1f1fbb9da38e6
35|2021-02-07T19:05:12.8460000-05:00|********|Player Four|40018B3A|Chiseled Sculpture|9598|0000|0011|40018B3A|000F|0000||34c06f84a7da705714c452d678848847
27|2021-02-07T19:05:12.8460000-05:00|********|Player Five|0000|0000|018F|0000|0000|0000||056bab79dc151cac7891e05d02d52be2
35|2021-02-07T19:05:12.8460000-05:00|********|Player Five|40018B3B|Chiseled Sculpture|9598|0000|0011|40018B3B|000F|0000||acec9ff58f91b8b3452cc2f1fe65d310
27|2021-02-07T19:05:12.8460000-05:00|********|Player Six|0000|0000|0193|0000|0000|0000||52e93b2c86d3a83bf17d5341ff0b24da
35|2021-02-07T19:05:12.8460000-05:00|********|Player Six|40018B3A|Chiseled Sculpture|9598|0000|0011|40018B3A|000F|0000||a87da7109863367c4f7ffbe2749a7499
27|2021-02-07T19:05:12.8460000-05:00|********|Player Seven|0000|0000|0190|0000|0000|0000||5cb09d0ad1e4880c00cd02e55bb288d7
35|2021-02-07T19:05:12.8460000-05:00|********|Player Seven|40018B3B|Chiseled Sculpture|9598|0000|0011|40018B3B|000F|0000||d859bb36e9631dec384d97b77cf5436e
27|2021-02-07T19:05:12.8460000-05:00|********|Player Eight|0000|0000|0194|0000|0000|0000||372050caf1b8b86521e080785dce26bc
35|2021-02-07T19:05:12.8460000-05:00|********|Player Eight|40018B3A|Chiseled Sculpture|9598|0000|0011|40018B3A|000F|0000||2f430ac45f2df6311b6f6bdf3719614a
```

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: Does that make sense when the order ends up being, for example, `1, Bait Inner Statue, 3, 4`?

Additionally, in the same sense, should the knockback direction influence the laser bait direction?

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: Removed for next commit. Apparently when I ran a search for it before, I typo'd console 🤷

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: I was trying to suggest modifying the default haircut trigger.  To have it go "is this person the 1st bait, if so then call out "Left + Bait Inner" rather than having a separate trigger for the 1st bait, if that makes sense?

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: I think you should be able to rely on the ordering.  The other option is to add a slight 0.25-0.5s delay from the number and then you can have text like `#1` if the tether wasn't there vs more information if it was.  That might be more robust for the future.

ui/raidboss/data/05-shb/raid/e12s.js
quisquous: Re: order.  It could say "2 (Bait Inner)" with alert/alarm vs info if you think it's asymmetrical to not have the number there.  I could also see having this same trigger call out the number with info and the "Bait Inner Statue" with alert.  For tts users, that would say "bait inner statue", but visual text users would see both.  Do any of those options sound better to you?

Re: knockback direction.  I think this is one of those strat-specific things.  I think it'd be easier to just say "bait" here and trust the player to go to the right spot.  My group does haircut-specific directions usually.

ui/raidboss/data/05-shb/raid/e12s.js
valarnin: The default haircut uses two separate triggers for left and right, which would both have to be modified for this or combined in some manner.

quisquous: I don't think modifying both sounds that bad, but if you think it'd be better separate, that's fine too.

valarnin:
Closing in favor of #2587

PR #2444, opened 2021-02-05
fix minor CN translation in TEA raidboss

miaonima:
冷却 is ambiguous here saying the jump is "cold". I think the trigger intended to let tank bait jump with some mitigation.


MaikoTan:
Well, "冷却" usually means "cooldown" in most rpg games, so i think this is good.
See if @ShadyWhite has comments.

miaonima:
> Well, "冷却" usually means "cooldown" in most rpg games, so i think this is good.
> See if @ShadyWhite has comments.

I understand players in international server call that "use/pop cooldown". But in CN server you call “T开减伤” while nobody (at least I have not seen one) would say "T用冷却".

ShadyWhite (approved):

PR #2455, opened 2021-02-07
Raidboss: O3S timeline refresh

JLGarber:
Nothing much on this one. It was time-consuming but relatively straightforward. There seems to be some randomness on the Queen's Waltz cast timings in phase 4. Nothing major, but it drifts by about 0.3 seconds sometimes.

I feel like we can do better with the `Spellblade Blizzard/Fire/Thunder` entries by using the English `replaceText` functionality, but I'm not sure how to put it more succinctly without incorrectly including `Spellblade Holy`. ("Random Spellblade" doesn't indicate that.)

### DIFF
--- a/ui/raidboss/data/04-sb/raid/o3s.txt
+++ b/ui/raidboss/data/04-sb/raid/o3s.txt
@@ -100,8 +100,8 @@ hideall "--sync--"
 456.0 "The Queen's Waltz (Random)" sync /:Halicarnassus:230[68AE]:/
 469.0 "Dimensional Wave" sync /:Halicarnassus:22F6:/
 477.1 "Haste" sync /:Halicarnassus:22F4:/ window 30,30
-480.8 "Spellblade Blizzard/Fire/Thunder" sync /:Halicarnassus:22E[CDE]:/
-485.4 "Spellblade Fire/Thunder/Blizzard" sync /:Halicarnassus:22E[CDE]:/
+480.8 "Spellblade Blizzard/Fire/Thunder 1" sync /:Halicarnassus:22E[CDE]:/
+485.4 "Spellblade Blizzard/Fire/Thunder 2" sync /:Halicarnassus:22E[CDE]:/
 491.5 "Dimensional Wave" sync /:Halicarnassus:22F6:/
 501.7 "Place Token (Ninjas/Giant)" sync /:Halicarnassus:22FB:/
 512.0 "Haste III" sync /:Halicarnassus:22F5:/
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/04-sb/raid/o3s.txt
```diff
@@ -1,176 +1,136 @@
...
+10.7 "Critical Hit" sync /:Halicarnassus:22EB:/ window 10.7,5
+19.7 "Spellblade Holy" sync /:Halicarnassus:22EF:/
+25.4 "Pole Shift" # sync /:Halicarnassus:22F[23]:/
+28.4 "Holy Blur/Holy Edge" sync /:Halicarnassus:22F[01]:/
+37.4 "the Queen's Waltz (Clock)" sync /:Halicarnassus:2306:/ window 37.4,5
+37.9 "Sword Dance" sync /:Halicarnassus:2307:/
+41.4 "Haste" sync /:Halicarnassus:22F4:/
+44.8 "Spellblade Thunder III" sync /:Halicarnassus:22EE:/
+50.3 "Dimensional Wave" sync /:Halicarnassus:22F6:/
+59.3 "the Playing Field" sync /:Halicarnassus:2300:/
```
quisquous, line 22: the => The, here and elsewhere, if you don't mind?

ui/raidboss/data/04-sb/raid/o3s.txt
```diff
@@ -1,176 +1,136 @@
...
-197	"Ribbit"
-203	"Dimensional Wave"	#Dimensional Wave
-203	"Dragon Conal AoE" #Frostbite
-217	"Critical Hit"	#Tank Buster
-228	"Place Dark Token"	#spawns 5 untargetable adds
-240	"Queen's Waltz (Crystals)"	#AoE + Stand on blue tiles
-#242	"Cross Reaper/Gusting Gouge"	#Middle AoE + knockback on edge tiles
-248	"Random Elemental" #Fire, Blizz, or Thunder
-253	"Dimensional Wave"	#AoE
+171.1 "Panel Swap (Cave)" sync /:Halicarnassus:2304:/ window 85,10
```
quisquous, line 47: If this is a push, can you leave a comment (and maybe sync all the way back to the beginning?)

Is it worth putting larger syncs on other panel swap phases? (I could go either way, honest question.)

quisquous (changes requested):
Can you remove the `timelineNeedsFixing: true,` line from the js file?

> I feel like we can do better with the `Spellblade Blizzard/Fire/Thunder` entries by using the English `replaceText` functionality, but I'm not sure how to put it more succinctly without incorrectly including `Spellblade Holy`. ("Random Spellblade" doesn't indicate that.)

`Elemental Spellblade`?

ui/raidboss/data/04-sb/raid/o3s.txt
JLGarber: Blast, there's always one set that gets overlooked.

ui/raidboss/data/04-sb/raid/o3s.txt
JLGarber: I don't know for sure. I'm mostly trying to ensure that we have the widest possible safe sync windows for those situations where we forget to turn ACT on until halfway through. (Also unfortunately we can't sync either The Game or Panel Swap to the beginning since it's the same ability each time.) If that's not worthwhile I can cut them down to the normal 30,30 windows.

JLGarber:
I thought about Elemental Spellblade, but it falls into the same problem I mentioned, that Holy is technically an element. Or do you think people will recognize it as "only the three specific ones"?

quisquous:
> I thought about Elemental Spellblade, but it falls into the same problem I mentioned, that Holy is technically an element. Or do you think people will recognize it as "only the three specific ones"?

I mean sure, holy is technically an element, but maybe I wouldn't consider it if you asked me what that meant.  I think your original `Spellblade Blizzard/Fire/Thunder` makes sense too.

ui/raidboss/data/04-sb/raid/o3s.txt
quisquous: Oh, ok.  So, not a push, just an extra sync.  Sounds good!  I think this is fine.  :)

JLGarber:
Thinking it over, you're right. Also, it occurs to me that I've directly argued in the past that "people should sort of get what things mean in Savage as long as what we call it is unambiguous". On that basis, `Elemental Spellblade` should be safe enough.

PR #1616, opened 2020-07-26
jobs: use effect ids instead of effect names

quisquous:

PR #2458, opened 2021-02-10
Refactor(jobs): melt brd specific code into others

MaikoTan:
I think this change would help when refactoring jobs later.

cc @Echoring 

Also fix a bug might come from #1616 while invoking functions with an undefined "name".

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -3897,7 +3897,7 @@ class Bars {
         const effectId = m.groups.effectId.toUpperCase();
         const f = this.gainEffectFuncMap[effectId];
         if (f)
-          f(name, m.groups);
+          f(effectId, m.groups);
         this.buffTracker.onYouGainEffect(effectId, m.groups);
       }
       m = log.match(kMobGainsEffectRegex);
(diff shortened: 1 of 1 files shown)


ui/jobs/jobs.js
```diff
@@ -3972,8 +3962,13 @@ class Bars {
       // but lastAttackedDotTarget must be your main target.
       if (line[2] === this.lastAttackedDotTarget &&
         line[4] === 'DoT' &&
-        line[5] === '0') // 0 if not fleld setting DoT
-        this.UpdateDotTimer();
+        line[5] === '0') {
+        // 0 if not fleld setting DoT
```
quisquous, line 3966:
```suggestion
        // 0 if not field setting DoT
```

ui/jobs/jobs.js
```diff
@@ -3886,36 +3889,56 @@ class Bars {
     const log = e.rawLine;
 
     const type = line[0];
+    const trackedDoTs = Object.keys(this.mobGainEffectFromYouFuncMap);
```
quisquous, line 3892: This is maybe premature optimization, but rather than calling this once per log line, can you set `this.trackedDots` once after the job setup?

quisquous (commented):
Thanks for the refactoring and bugfixes!

ui/jobs/jobs.js
MaikoTan: Oops, didn't catch this issue!

PR #2460, opened 2021-02-11
build: generate source map when in development

MaikoTan:
This helps when we are debugging with modules,
In chrome or CEF (in OverlayPlugin), we can directly set breakpoints in source code files provides by `webpack-dev-server`,
and then make changes in-game, to catch variables changes in code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build": "node_modules/.bin/webpack --config webpack/webpack.config.cjs --mode production",
-    "start": "node_modules/.bin/webpack-dev-server --config webpack/webpack.config.cjs --mode development",
+    "start": "node_modules/.bin/webpack-dev-server --config webpack/webpack.config.cjs --mode development --devtool source-map",
     "lint": "node_modules/.bin/eslint **/*.{js,cjs,mjs}",
     "lintfix": "node_modules/.bin/eslint --fix **/*.{js,cjs,mjs}",
     "stylelint": "node_modules/.bin/stylelint resources/**/*.css ui/**/*.css user/**/*.css test/**/*.css util/**/*.css",
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
   },
   "scripts": {
     "build": "node_modules/.bin/webpack --config webpack/webpack.config.cjs --mode production",
-    "start": "node_modules/.bin/webpack-dev-server --config webpack/webpack.config.cjs --mode development --devtool source-map",
+    "start": "node_modules/.bin/webpack-dev-server --config webpack/webpack.config.cjs --mode development",
     "lint": "node_modules/.bin/eslint **/*.{js,cjs,mjs}",
     "lintfix": "node_modules/.bin/eslint --fix **/*.{js,cjs,mjs}",
     "stylelint": "node_modules/.bin/stylelint resources/**/*.css ui/**/*.css user/**/*.css test/**/*.css util/**/*.css",
--- a/webpack/webpack.config.cjs
+++ b/webpack/webpack.config.cjs
@@ -2,68 +2,71 @@
 
 const path = require('path');
 
-module.exports = {
-  entry: {
-    config: './ui/config/config.js',
-    coverage: './util/coverage/coverage.js',
-    rdmty: './ui/dps/rdmty/dps.js',
-    xephero: './ui/dps/xephero/xephero.js',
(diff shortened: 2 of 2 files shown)


quisquous:
Can this be done from within the webpack config, rather than in the shortcut?

MaikoTan:
It can, but we are using the same config file in development/production mode, I think the source map should not be included in production, so only add it in CLI.

But maybe we can assign this configuration value in runtime?

quisquous:
What's the reason not to have a source map in production? (Sorry if that's a weird question! I don't know!)

MaikoTan:
Source map files are like debugging symbol files (*.pdb) in C/C++/C# developing when you are using Visual Studio.
Including it in production would only increase the release size, because they are usually bigger than the corresoponding bundled JS file.

image

MaikoTan:
And if we want to assign `devtool` option in runtime, we can modify `webpack.config.cjs` like this:
(webpack also accept function as configuration)

```javascript
'use strict';

const path = require('path');

module.exports = function(env, argv) {
  return {
    entry: {
      config: './ui/config/config.js',
      coverage: './util/coverage/coverage.js',
      rdmty: './ui/dps/rdmty/dps.js',
      xephero: './ui/dps/xephero/xephero.js',
      eureka: './ui/eureka/eureka.js',
      fisher: './ui/fisher/fisher.js',
      jobs: './ui/jobs/jobs.js',
      oopsyraidsy: './ui/oopsyraidsy/oopsyraidsy.js',
      pullcounter: './ui/pullcounter/pullcounter.js',
      radar: './ui/radar/radar.js',
      raidboss: './ui/raidboss/raidboss.js',
      raidemulator: './ui/raidboss/raidemulator.js',
      test: './ui/test/test.js',
    },
    optimization: {
      splitChunks: {
        cacheGroups: {
          'raidboss_data': {
            test: /[\\/]ui[\\/]raidboss[\\/]data[\\/]/,
            name: 'raidboss_data',
            chunks: 'all',
          },
          'oopsyraidsy_data': {
            test: /[\\/]ui[\\/]oopsyraidsy[\\/]data[\\/]/,
            name: 'oopsyraidsy_data',
            chunks: 'all',
          },
        },
      },
    },
    devtool: argv.mode === 'development' ? 'source-map' : undefined,
    output: {
      filename: '[name].bundle.js',
      path: path.resolve(__dirname, '../dist'),
    },
    devServer: { writeToDisk: true },
    module: {
      rules: [
        {
          test: /\.css$/,
          use: ['style-loader', 'css-loader'],
        },
        {
          test: /data[\\\/]manifest\.txt$/,
          use: [
            {
              loader: './webpack/loaders/manifest-loader.cjs',
            },
          ],
        },
        {
          test: /data[\\\/](?!manifest\.txt).*\.txt$/,
          use: [
            {
              loader: 'raw-loader',
            },
            {
              loader: './webpack/loaders/timeline-loader.cjs',
            },
          ],
        },
      ],
    },
  };
};
```

quisquous:
Having the source map on github pages would allow better debugging in that sense though, right? I guess then we'd need a different release vs gh-pages action.  Maybe something to add in once cactbot and overlayplugin merge and there aren't releases in the same way.  Ok, so development only makes sense for now.  I like the second approach that you posted in a comment better, as I personally prefer to have all the options in the config file and not specified separately if possible.

MaikoTan:
> Having the source map on github pages would allow better debugging in that sense though, right?

Right though I cannot get any sense. When we publish to github pages, it should be available at least. Almost no  condition should we debug the online version. If debug needed, why not debug on local server, using webpack-dev-server?

quisquous:
> > Having the source map on github pages would allow better debugging in that sense though, right?
> 
> Right though I cannot get any sense. When we publish to github pages, it should be available at least. Almost no condition should we debug the online version. If debug needed, why not debug on local server, using webpack-dev-server?

If folks have a local server, then it makes sense to just use the local server.  But, I think my thought is that there are probably some cactbot users out there who might be able to debug a problem and are using the online version and file a good bug report but wouldn't bother to set up their computer for local development.

MaikoTan:
Sounds reasonable but nowadays we can start dev server in just 2 commands (`npm i` and `npm run start`), if someone knows how to debug, he probably knows how to deal with a dev server. So I cannot see an usage about this.
However, if you want to do this, you can just add `--devtool source-map` option after build script, just like what I did at tge first time.
Also, according to webpack's documents, `source-map` option is the slowest selection, would cause the build workflow running a long time.

PR #2462, opened 2021-02-11
build: add basic TypeScript config

Trim21:
this PR add basic  typescript config(tsc, webpack and eslint) <del>and refactor `resources/widgetlist.ts` as example.</del>


tsconfig.json
```diff
@@ -0,0 +1,79 @@
+{
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5",
+    /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "module": "commonjs",
+    /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
```
MaikoTan, line 10: Can you change `target` and `module` as `esnext`, and remove these redundant comments?

tsconfig.json
```diff
@@ -0,0 +1,79 @@
...
+    /* Strict Type-Checking Options */
+    "strict": true,
+    /* Enable all strict type-checking options. */
+    // "noImplicitAny": true,                 /* Raise error on expressions and declarations with an implied 'any' type. */
+    // "strictNullChecks": true,              /* Enable strict null checks. */
+    // "strictFunctionTypes": true,           /* Enable strict checking of function types. */
+    // "strictBindCallApply": true,           /* Enable strict 'bind', 'call', and 'apply' methods on functions. */
+    // "strictPropertyInitialization": true,  /* Enable strict checking of property initialization in classes. */
+    // "noImplicitThis": true,                /* Raise error on 'this' expressions with an implied 'any' type. */
+    // "alwaysStrict": true,                  /* Parse in strict mode and emit "use strict" for each source file. */
```
MaikoTan, line 38: These options should be always enabled.

tsconfig.json
```diff
@@ -0,0 +1,79 @@
...
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5",
+    /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "module": "commonjs",
+    /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
+    // "allowJs": true,                       /* Allow javascript files to be compiled. */
+    // "checkJs": true,                       /* Report errors in .js files. */
+    // "jsx": "preserve",                     /* Specify JSX code generation: 'preserve', 'react-native', or 'react'. */
+    // "declaration": true,                   /* Generates corresponding '.d.ts' file. */
```
MaikoTan, line 15: Can you enable `declaration` and output typing filed to `./typings`? Also, remember to add `typing` entry in `package.json`.

tsconfig.json
```diff
@@ -0,0 +1,79 @@
...
+  "compilerOptions": {
+    /* Visit https://aka.ms/tsconfig.json to read more about this file */
+
+    /* Basic Options */
+    // "incremental": true,                   /* Enable incremental compilation */
+    "target": "es5",
+    /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019', 'ES2020', or 'ESNEXT'. */
+    "module": "commonjs",
+    /* Specify module code generation: 'none', 'commonjs', 'amd', 'system', 'umd', 'es2015', 'es2020', or 'ESNext'. */
+    // "lib": [],                             /* Specify library files to be included in the compilation. */
```
MaikoTan, line 11: This should not be empty, right?
At least `dom` and `esnext` should be included in?

tsconfig.json
```diff
@@ -0,0 +1,79 @@
...
+    /* Experimental Options */
+    // "experimentalDecorators": true,        /* Enables experimental support for ES7 decorators. */
+    // "emitDecoratorMetadata": true,         /* Enables experimental support for emitting type metadata for decorators. */
+
+    /* Advanced Options */
+    "skipLibCheck": true,
+    /* Skip type checking of declaration files. */
+    "forceConsistentCasingInFileNames": true
+    /* Disallow inconsistently-cased references to the same file. */
+  }
```
MaikoTan, line 78: Also, can you explicitly add `include` and `exclude` in this file?

resources/widgetlist.ts
```diff
@@ -238,34 +274,33 @@ class WidgetList extends HTMLElement {
...
     element.style.position = 'relative';
-    element.style.left = element.style.top = 0;
+    element.style.left = element.style.top = '0';
 
     const container = document.createElement('div');
     container.appendChild(element);
     container.id = 'child' + id;
 
-    this.rootElement.appendChild(container);
+    this.rootElement!.appendChild(container);
```
MaikoTan, line 288: Why are here and everywhere using `!.` oprand instead of a explicit null-check?

resources/widgetlist.ts
```diff
@@ -285,8 +320,8 @@ class WidgetList extends HTMLElement {
     if (!this._connected)
       return;
 
-    this.rootElement.style.width = this._rowcolsize * this._elementwidth;
-    this.rootElement.style.height = this._rowcolsize * this._elementheight;
+    this.rootElement.style.width = String(this._rowcolsize * this._elementwidth);
+    this.rootElement.style.height = String(this._rowcolsize * this._elementheight);
```
MaikoTan, line 324: I think not to use `String` constructor, directly invpke `toString()` would be better? also, a template string with number literal is not bad though. here and everywhere.

resources/widgetlist.ts
```diff
@@ -1,17 +1,38 @@
 // NOTE
 // This class should be considered deprecated, and any users of this class should
 // just switch over to using CSS grid.
+type Sorter = () => number;
 
 class WidgetList extends HTMLElement {
+  private _nextId!: number;
+  private _nameToId!: { [key: string]: number };
```
MaikoTan, line 8: Can `[key:string]` be more precise?
And why is here declared using `!.` oprator.

MaikoTan (changes requested):

resources/widgetlist.ts
Trim21: I just translation what old js do into ts. And in this case, root element won't be null.

Trim21:
a gist without deprecated api https://gist.github.com/Trim21/3f193e45bca9653aa109efbfd8d119fb

tsconfig.json
Trim21: required lib will be the default value for different target.

quisquous:
Thanks for looking into this!  I think really my concern is that I wanted to wait until the testing system got switched over to jest before beginning conversion to typescript, as jest can then handle converting typescript when running tests.  (cc @panicstevenson)

I think this is probably one of the only files you could convert at this point, given that most other things end up getting imported into tests, which would then break, if I understand this correctly.

Trim21:
We can compile ts to js then test them with old test system before we switch to new test system.

MaikoTan:
I think we can directly use `ts-node` for typescript files testing later, even jest test files itself can be written in typescript.

package.json
```diff
@@ -8,16 +8,19 @@
   },
   "scripts": {
     "build": "node_modules/.bin/webpack --config webpack/webpack.config.cjs --mode production",
+    "compile": "tsc",
```
MaikoTan, line 11: ye don't need to `tsc`, `webpack` and `ts-loader`would handle this.

resources/util.ts
```diff
@@ -0,0 +1,243 @@
+const kTankJobs = ['GLA', 'PLD', 'MRD', 'WAR', 'DRK', 'GNB'];
```
MaikoTan, line 1: Why not putting them in a `namespace` instead of seperate out?

resources/util.ts
```diff
@@ -0,0 +1,243 @@
...
+  | 29
+  | 30
+  | 31
+  | 32
+  | 33
+  | 34
+  | 35
+  | 36
+  | 37
+  | 38
```
MaikoTan, line 58: To use `keyof` would save your life.

resources/util.ts
```diff
@@ -0,0 +1,243 @@
...
+  'NIN': 30,
+  'MCH': 31,
+  'DRK': 32,
+  'AST': 33,
+  'SAM': 34,
+  'RDM': 35,
+  'BLU': 36,
+  'GNB': 37,
+  'DNC': 38,
+};
```
MaikoTan, line 142: Redundant declaration.

tsconfig.json
```diff
@@ -0,0 +1,43 @@
+{
+  "compilerOptions": {
+    "target": "ES6",
+    "module": "commonjs",
```
MaikoTan, line 4:
```suggestion
    "target": "esnext",
    "module": "esnext",
```

MaikoTan (changes requested):

resources/util.ts
Trim21: will cause cycle definition

MaikoTan:
By the way, I think `"type": "module",` in package.json should be removed since we bundle it up with webpack and pretend to using jest to make unit test?

Trim21:
CI could pass if we refactor a js file to ts now. (which should be import from file name without extension)

tsconfig.json
```diff
@@ -0,0 +1,43 @@
...
+    "strictFunctionTypes": true,
+    "strictBindCallApply": true,
+    "strictPropertyInitialization": true,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUncheckedIndexedAccess": true,
```
quisquous, line 24: It might be nice to add `"noPropertyAccessFromIndexSignature": false, // used by trigger output strings` just as a comment about why that isn't turned on.  Assuming this is one of those json files that allows comments.  <_<

tsconfig.json
```diff
@@ -0,0 +1,43 @@
...
+    "moduleResolution": "node",
+    "baseUrl": "./",
+    "esModuleInterop": true,
+    "experimentalDecorators": false,
```
quisquous, line 28: This is off by default, so maybe doesn't need to be listed here?

tsconfig.json
```diff
@@ -0,0 +1,43 @@
+{
+  "compilerOptions": {
```
quisquous, line 2: Can you also add `"preserveConstEnums": true,`

tsconfig.json
```diff
@@ -0,0 +1,43 @@
...
+    "baseUrl": "./",
+    "esModuleInterop": true,
+    "experimentalDecorators": false,
+    "skipLibCheck": true,
+    "forceConsistentCasingInFileNames": true
+  },
+  "include": [
+    "./ui",
+    "./resources",
+    "./test"
```
quisquous, line 35: "./util" as well?

.eslintrc.cjs
```diff
@@ -20,6 +20,17 @@ module.exports = {
...
       },
     },
+    {
+      'files': ['**/*.ts'],
+      'parser': '@typescript-eslint/parser',
+      'extends': [
+        'plugin:@typescript-eslint/recommended',
+      ],
+      'rules': {
+        '@typescript-eslint/no-non-null-assertion': 2,
```
quisquous, line 30: How does eslint with typescript work here? Will the eslint rules also apply to the .ts files or the converted ones? Or do all the equivalent rules turned on for eslint need to be turned on for typescript as well?

tsconfig.json
```diff
@@ -0,0 +1,43 @@
+{
+  "compilerOptions": {
+    "target": "ESNext",
+    "module": "ESNext",
+    "allowJs": true,
+    "checkJs": false,
```
quisquous, line 6: This seems like something that would be good to turn on and fix errors for, even now (or in a followup PR), unless I'm misunderstanding?

typing/trigger.d.ts
```diff
@@ -0,0 +1,54 @@
...
+  tts?: LanguageText,
+  run?(data: Data, matches: Match): void,
+  outputStrings?: Record<string, unknown>,
+}
+
+export interface TimeLineTrigger extends BaseTrigger {
+  beforeSeconds?: number,
+}
+
+export interface LogTrigger extends BaseTrigger {
```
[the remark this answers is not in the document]
Trim21: `beforeSeconds` is useless in LogTrigger, it only works in TimelineTrigger, and netRefeg doesn't work in TimelineTrigger, am I misunderstanding?

test/definitions.py
```diff
@@ -11,6 +11,7 @@
 TEST_DIRECTORY = "test"
 UI_DIRECTORY = "ui"
 MANIFEST_FILENAME = "manifest.txt"
+NODE_COMMAND = "node --experimental-specifier-resolution=node --loader ts-node/esm".split()
```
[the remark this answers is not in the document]
Trim21: No, this flag has nothing to do with syntax `import` or `require`, it only tell node to auto-add `.js` extension when importing or requiring a new module. So we can `import { something } from './file'`. And ts-node will try to add `.ts` extension, too. In this way we can import a js file or ts file and it will work as expected. We can't import a file with `.ts` extension, node doesn't support that.

.eslintrc.cjs
Trim21: > How does eslint with typescript work here? Will the eslint rules also apply to the .ts files or the converted ones? Or do all the equivalent rules turned on for eslint need to be turned on for typescript as well?

It will merge rules from `@typescript-eslint/recommended` in to old js rules. We don't need to do it twice

typing/trigger.d.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+interface Match {
+  target?: string
+  [key: string]: string
+}
+
+
+type LanguageText = Record<lang, string>
+type Response = (data: Data, matches: Match, output: Output) => string | LanguageText
```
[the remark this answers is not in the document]
Trim21: > I think overall it would be better to put this file in a separate pull request as there's a bunch of details here that need to be ironed out, and this file isn't used yet.

OK, I'll leave only `typings/data.d.ts` to see if ci running as expected, and move other type definition to a seprated pr

tsconfig.json
```diff
@@ -0,0 +1,43 @@
...
+    "strict": true,
+    "noImplicitAny": true,
+    "strictNullChecks": true,
+    "strictFunctionTypes": true,
+    "strictBindCallApply": true,
+    "strictPropertyInitialization": true,
+    "noImplicitThis": true,
+    "alwaysStrict": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
```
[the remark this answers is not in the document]
MaikoTan: > I think this is going to be tricky for the number of trigger fields that do `(data, matches, output) => output.text()` style things.

Write something like `(_, _, output) => output.text()`?
But I think this check still worth to be an eslint rule?

tsconfig.json
Trim21: It looks that this will affect performance...

tsconfig.json
Trim21: It seems that this option is added in ts 4.2.0 and current latest stable version is 4.1.5

.eslintrc.cjs
MaikoTan: ESLint doesn't know JavaScript or TypeScript, it only knows AST(Abstract Syntax Tree). The `@typescript-eslint/parser` would parse `.ts` file to AST that compatiable with ESLint's. So ESLint can capture common syntax error (TypeScript claims it is a superset of JavaScript), but if that is a TypeScript-specific syntax, only `@typescript-eslint` could do the such thing.

tsconfig.json
quisquous: I think the "we won't read" part is not true.  For developers, probably.  For users writing their own triggers, they will probably look at the generated javascript code.

tsconfig.json
quisquous: Ah ok!  Thanks.  Sorry, I was just reading through options from the typescript page.

typing/trigger.d.ts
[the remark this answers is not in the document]
quisquous: Oh sorry, I see what you're doing.  My misunderstanding, that makes sense.

tsconfig.json
Trim21: In fact I think user won't read nether `raidboss_data.bundle.js` or the trigger file compiled by tsc(both unreadable even if we set this option to true), They should read the original TS file instead.

This is what generated code would look like, and there it's not worth reading.
image

tsconfig.json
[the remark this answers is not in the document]
quisquous: I think you can't use `_` twice there, but you could say `(_data, _matches, output)`.  It seems reasonable to add.  Mostly just saying that it'll be a lot of churn to turn on.

.eslintrc.cjs
quisquous: Thanks for the explanation.

tsconfig.json
quisquous: They certainly won't read the bundle, but they will likely read typescript -> javascript converted trigger files.  A lot of users copy and paste triggers and so I think there needs to be some solution for them.  I think when publishing to gh-pages (or in a release??) we will need to convert the typescript trigger files to javascript so users have something to copy and paste from in their user trigger files, e.g. [this workflow](https://github.com/quisquous/cactbot/blob/main/docs/CactbotCustomization.md#example-1-changing-the-output-text)

Trim21: I think we can do that with cli flags when generating a preview(maybe, I'm not sure how to call it), and disable it when building the final bundle file.

quisquous: Ok, that works for me, thanks!

tsconfig.json
```diff
@@ -0,0 +1,42 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    "module": "ES2020",
+    "allowJs": true,
+    "checkJs": true,
```
quisquous: Oops, I think this was maybe a bad suggestion as now all the imports have errors because they're implictly any.

tsconfig.json
Trim21: There are too much error in js file, I think it would be better to migrate js to ts and fix them in the same time. It will be very hard to fix all js errors in this or a single PR.

tsconfig.json
Trim21: yes, I just disabled it again....

Trim21:
stranger things, tests passed locally...

Trim21:
I think we need to upgrade node version to 14 to make sure that ts-node works as expected.

tsconfig.json
quisquous: Sounds good! Sorry for the churn.

types/global.d.ts
```diff
@@ -0,0 +1 @@
+export type Lang = 'cn' | 'de' | 'en' | 'fr' | 'ja' | 'ko';
```
MaikoTan, line 1: I think to follow an order like `'en' | 'de' | 'fr' | 'ja' | 'cn' | 'ko'` is consistency with our [common locale order](https://github.com/quisquous/cactbot/blob/main/docs/RaidbossGuide.md#trigger-properties).

package.json
```diff
@@ -31,12 +34,15 @@
...
+    "ts-loader": "^8.0.17",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.1.5",
     "webpack": "^5.11.0",
     "webpack-cli": "^3.3.12",
     "webpack-dev-server": "^3.11.0"
   },
   "engines": {
-    "node": ">=13.2.0"
+    "node": ">=14.15.5"
```
quisquous, line 45: Is there something specific in this version? Or do you just need `14.x.0` for some x? Mostly asking, because I'm at `14.15.1` and I don't see any reason to make me/everyone upgrade if not needed.

tsconfig.json
quisquous: ?

package.json
Trim21: Not specific reason, just current LTS version. I think I should downgrade it to 14.15.0.

quisquous:
Do you also need to add `**/*.ts` to `.github/workflows/javascript-lint.yml`?

tsconfig.json
Trim21: Sorry, add it to exclude by mistake

types/data.d.ts
```diff
@@ -0,0 +1,32 @@
...
+  party: PartyTracker;
+  lang: string;
+  currentHP: number;
+  options: Option;
+  ShortName(x: string): string;
+  StopCombat(): void;
+  /**
+   * @deprecated Use parseFloat instead
+   */
+  ParseLocaleFloat(string: string): number;
```
MaikoTan, line 25: one line can be 
```suggestion
  /** @deprecated Use parseFloat instead */
  ParseLocaleFloat(string: string): number;
```

tsconfig.json
```diff
@@ -0,0 +1,42 @@
...
+    "./resources",
+    "./util",
+    "./test"
+  ],
+  "exclude": [
+    "./eslint",
+    "./.eslintrc.cjs",
+    "./webpack",
+    "./dist",
+    "**/*.spec.ts"
```
MaikoTan, line 40: What is `*.spec.ts` use for by the way?

Trim21: tests(which haven't been written)

MaikoTan: image

PR #2463, opened 2021-02-12
Raidboss + oopsyraidsy: St Mocianne's Arboretum (hard)

JLGarber:
Nothing special really. The only thing of note is that we have no way of syncing the timeline sooner on boss 2 for the path splits. The user *can* tell which it will be, since the golems will be in a line if it's the Earthquake pattern and will be in a box if it's the Earth Shaker pattern. However, the timeline has no good way of determining this.

### DIFF
--- a/ui/raidboss/data/04-sb/dungeon/st_mocianne_hard.js
+++ b/ui/raidboss/data/04-sb/dungeon/st_mocianne_hard.js
@@ -29,7 +29,7 @@ export default {
     },
     {
       id: 'St Mocianne Hard Vine Whip',
-      netRegex: NetRegexes.startsUsing({ id: '2E48' }),
+      netRegex: NetRegexes.startsUsing({ id: '2E48', source: 'Nullchu' }),
       condition: Conditions.caresAboutPhysical(),
       response: Responses.tankBuster(),
     },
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/04-sb/dungeon/st_mocianne_hard.js
```diff
@@ -0,0 +1,141 @@
...
+      outputStrings: {
+        text: {
+          en: 'Sludge puddle on YOU',
+        },
+      },
+    },
+    {
+      id: 'St Mocianne Hard Fault Warren',
+      netRegex: NetRegexes.headMarker({ id: '003E' }),
+      suppressSeconds: 5,
```
quisquous, line 50: Can you leave a comment explaining this one? I think it's something I could see accidentally removing later or not understanding why it's here.

quisquous (commented):
Looks good, thanks!! Who knew boss 2 had a branching timeline??? (I sure didn't.)

ui/raidboss/data/04-sb/dungeon/st_mocianne_hard.js
JLGarber: Good point!

PR #2466, opened 2021-02-13
Refactor: Some common utils TypeScript-ization

MaikoTan:


resources/netregexes.ts
```diff
@@ -56,10 +68,10 @@ const parseHelper = (params, funcName, fields) => {
       throw new Error(`${funcName}: invalid value: ${JSON.stringify(value)}`);
 
     const fieldName = fields[key].field;
-    const fieldValue = 'value' in fields[key] ? fields[key].value.toString() : matchDefault;
+    const fieldValue = fields[key].value?.toString() ?? matchDefault;
```
quisquous, line 71: I am looking forward to both `?` and `??`.  That will be so useful to have.

resources/netregexes.ts
```diff
@@ -102,9 +120,12 @@ export default class NetRegexes {
     });
   }
 
-  // matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#15-networkability
-  // matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#16-networkaoeability
-  static abilityFull(params) {
+  const abilityFullParams = ['sourceId', 'source', 'id', 'ability', 'targetId', 'target', 'flags', 'damage', 'targetCurrentHp', 'targetMaxHp', 'x', 'y', 'z', 'heading'] as const;
+  /**
+   * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#15-networkability
+   * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#16-networkaoeability
```
quisquous, line 126: What is the comment style change here for?

MaikoTan: JSDoc style, so most editor with JSDoc extension would display these text when hover it or type it. (VSCode support this as built-in feature though.)

MaikoTan: Just like:
image

MaikoTan:
As I said in https://github.com/quisquous/cactbot/pull/2478#issuecomment-782088605 , I think we can only to remove the `"type": "module",` option in `package.json`?

quisquous:
Ok, I am really stuck on how to fix the import of regexes into netregexes.  Let me write down everything I've learned and understand and maybe we can muddle through this together:

I see at least three options:

### 1) `import from './regexes.ts'`

:heavy_check_mark: WebPack
:x: mocha
:x: lint (ts(2691))

This results in a TS compile error: `resources/netregexes.ts(1,33): error TS2691: An import path cannot end with a '.ts' extension. Consider importing './regexes' instead.`  One option is that we could technically ... just ignore it with a `// @ts-ignore` or some similar suppression, although this causes an eslint@typescript-eslint/ban-ts-comment error.  However, mocha does run correctly with @ts-ignore, for what it's worth.  So clearly webpack and mocha know what to do here.

This comment: https://github.com/Microsoft/TypeScript/issues/ 27481#issuecomment-426026215 (link deliberately broken to avoid creating more noise on that old issue) points out that this could cause issues and that "you should omit the file extension from your imports, or use the .js extension".  I'm not sure I fully understand the ramifications here, but my intuition is that this is probably not the correct answer.

So, probably we shouldn't do this, even though maybe we could get away with it?


### 2) `import from './regexes.js'`

:x: WebPack
:heavy_check_mark: mocha
:x: lint (eslintimport/no-unresolved)

This approach is what https://github.com/TypeStrong/ts-node/issues/ 1007#issue-598417180 calls "idiomatic typescript".

I think the lint issues here are the least important.  We could find some better/smarter linter for typescript imports.

However, I cannot convince WebPack to do the right thing here.  `Module not found: Error: Can't resolve './regexes.js' in 'C:\Users\enne\cactbot\resources'`.  I've tried `ts-loader` and `awesome-typescript-loader`.  I've tried running .js files through the `ts-loader` as well.

One thing I haven't tried is writing my own loader that does something like "if file doesn't exist, try a .ts file of the same name".  I did a LOT of searching, but I was unable to find anybody else having this same issue.

### 3) `import from './regexes'`

  :heavy_check_mark: WebPack
  :x: mocha
  :heavy_check_mark: lint

WebPack understands how to do this because of the `resolve:{ extensions: [.ts, .js] }` block.  If an import has no extension, it will consider these extensions (in order, I think).

Unfortunately, Mocha doesn't know what to do here:
```
Error: ERR_MODULE_NOT_FOUND C:\Users\enne\cactbot\resources\regexes C:\Users\enne\cactbot\resources\netregexes.ts module
    at finalizeResolution (C:\Users\enne\cactbot\node_modules\ts-node\dist-raw\node-esm-resolve-implementation.js:374:11)
```

If you look at [that file](https://github.com/TypeStrong/ts-node/blob/master/dist-raw/node-esm-resolve-implementation.js#L350-L362), there is a block of code there that's looking for `--experimental-specifier-resolution=node` being passed in.  If you hack that branch of the code to true (as if that option were there), then mocha runs properly and it can find all of its files.

Unfortunately, if you actually pass that option to mocha, e.g. `node --experimental-specifier-resolution=node --loader=ts-node/esm node_modules/mocha/bin/_mocha` then you get this error:
```
TypeError [ERR_INVALID_RETURN_PROPERTY_VALUE]: Expected string to be returned for the
 "format" from the "loader getFormat" function but got type object.
    at Loader.getFormat (internal/modules/esm/loader.js:110:13)
```
This error is possibly due to https://github.com/nodejs/node/issues/ 33259#issuecomment-624306938.  However, this is from May of last year.  If you look at the [esm documentation](https://github.com/nodejs/node/blob/master/doc/api/esm.md#customizing-esm-specifier-resolution-algorithm) it sure looks like extensionless esm is supported.  Additionally, I dug up https://github.com/nodejs/node/pull/ 35098 which is fixing an issue that sure looks like this issue, and so maybe this is fixed in node 15.0.0 and later.  (I tried install node 15.9.0 but it made a huge mess of package-lock.json and I couldn't get anything to work anymore.)

### Summary

`¯\_(ツ)_/¯`

quisquous:
I tried to use https://webpack.js.org/plugins/normal-module-replacement-plugin/ to implement #2 above, but it appears that this resolving hook does not get called for regexes.js (maybe because it doesn't exist? or some other reason?).

MaikoTan:
So I did not understand that why not consider `commonjs` soluion instead? I can see everywhere that ESM is still an experimental feature, not fully supported.
As for CommonJS, we have all the mature method on resolving, linting, testing, and bundling. Additionally, when we bundle `cactbot` up and ship it to user, it is no matter that it is ESM or CommonJS (webpack compile them with its own loading function via eval()).

quisquous:
I think part of the reason I wanted to use ESM was so that downstream users could import the .js files in a browser if they needed.  Typescript certainly makes this harder, but not impossible.  But https://github.com/quisquous/cactbot-chat/pull/2/files#diff-3488b72567c312043498c71ef6123cde991d4b0b672fe5a4520221071b3991d4 sort of thing.  Maybe this is not something we should support, but that was part of the initial reason.

As for solution #2, I managed to fix the WebPack part of this here: https://github.com/MaikoTan/cactbot/pull/77, although the eslint import issues still need to be fixed before this can land.

MaikoTan:
> I think part of the reason I wanted to use ESM was so that downstream users could import the .js files in a browser if they needed. 

What do you mean that "users could import the .js files in a browser"?


> As for solution #2, I managed to fix the WebPack and eslint part of this here: https://github.com/MaikoTan/cactbot/pull/77.  VSCode is still confused about this, and I'm not sure how to get it to work, but tests and lint should pass.

The issue of `ts-loader` that you posted before said that there is an option to force VSCode recognize .js as .ts, I didn't try that though, it might be a solution if you insist on this method.

MaikoTan:
> I think part of the reason I wanted to use ESM was so that downstream users could import the .js files in a browser if they needed. 

What do you mean that "users could import the .js files in a browser"?


> As for solution #2, I managed to fix the WebPack and eslint part of this here: https://github.com/MaikoTan/cactbot/pull/77.  VSCode is still confused about this, and I'm not sure how to get it to work, but tests and lint should pass.

The issue of `ts-loader` that you posted before said that there is an option to force VSCode recognize .js as .ts, I didn't try that though, it might be a solution if you insist on this method.

quisquous:
> > I think part of the reason I wanted to use ESM was so that downstream users could import the .js files in a browser if they needed.
> 
> What do you mean that "users could import the .js files in a browser"?

By "user" here, I mean "developer making an overlay that depends on cactbot files".  You can see the PR I linked to and the repo that it's attached to as an example.  I think I was trying to imagine what somebody would have to do to import some cactbot files and make an overlay that was outside of the repo without necessarily having to set up webpack or a whole bunch of infra.

I think, re: esm vs cjs.  CJS feels to me like "the deprecated way" and ESM is "the one that's not quite ready yet".  I think I believed that it was "ready enough", but maybe that's not quite true.  I think import overall has better semantics.  Dynamic import is async vs the sync require.  The lexical import is known at parsing time vs require being dynamic.  Import allows for better tree-shaking in webpack, although I doubt this will matter for cactbot.  So, it just seemed like "a better way forward", if that makes sense.

> > As for solution #2, I managed to fix the WebPack and eslint part of this here: [MaikoTan#77](https://github.com/MaikoTan/cactbot/pull/77).  VSCode is still confused about this, and I'm not sure how to get it to work, but tests and lint should pass.
> 
> The issue of `ts-loader` that you posted before said that there is an option to force VSCode recognize .js as .ts, I didn't try that though, it might be a solution if you insist on this method.

VSCode figured it out eventually.  I think it just took a bit to refresh the change to the eslint config.  So, I think this solution is probably good to go?

If you'd prefer that I land those changes separately rather than merging them into your PR, I'd be happy to do that too.

MaikoTan:
I think merging in this PR is well, but I got a stomachache today, I think I can take a look tonight or tomorrow...

MaikoTan:
Sorry that so long I return back to this PR, 'cause I am so busy these days.

> I think, re: esm vs cjs. CJS feels to me like "the deprecated way" and ESM is "the one that's not quite ready yet". I think I believed that it was "ready enough", but maybe that's not quite true. I think import overall has better semantics. Dynamic import is async vs the sync require. The lexical import is known at parsing time vs require being dynamic. Import allows for better tree-shaking in webpack, although I doubt this will matter for cactbot. So, it just seemed like "a better way forward", if that makes sense.

I think I was trying to refer to just remove the "type: module" property in `package.json`, not to say something like "we should use the deprecated way, to `require()` and `module.export = xxx;`"?
In that way, not only `webpack` runs well with extensionless import statements, `mocha` can also run tests via `esm` and `ts-node/register`, and eslint would not be affected too.

We can definitely keep all our files in repo being ESM style even remove that property. For example, [underscore.js](https://github.com/jashkenas/underscore/blob/master/package.json) didn't declare the "type: module" property although it provides ESM exports.
(Webpack can do tree-shaking on these packages due to the "module" field in `package.json`, see: https://webpack.js.org/guides/author-libraries/#final-steps )

After all, removing "type: module" property would only affect how nodejs loads files in the repo, but didn't change the way in bundling, or being used by other packages.

quisquous:
Oh, I misunderstood what you merged and merged my partial fix suggestion above.  I think you need that, but you're also missing https://github.com/quisquous/cactbot/commit/6f9b8bf1e6462d3f832c682e82b059d5f2ba14da and https://github.com/quisquous/cactbot/commit/2dc7c97ba1097b4d28cae50bec6e8fa2ab29eee1.  Sorry!

PR #2469, opened 2021-02-13
Raidboss + oopsy: Kugane Castle

JLGarber:
I'm not sure why I never bothered with this one before, it was remarkably painless. Nothing special here.


quisquous:
Thanks for taking the time to send this in!  Glad this one wasn't too tricky for once.  :)

PR #2478, opened 2021-02-15
build: disable type checking temporary

MaikoTan:
Disable type checking in TypeScript transpiling temporary, this option should be removed when we successfully refactor all the code into TypeScript.

### DIFF
--- a/webpack/webpack.config.cjs
+++ b/webpack/webpack.config.cjs
@@ -46,7 +46,14 @@ module.exports = function(env, argv) {
       rules: [
         {
           test: /\.ts$/,
-          loader: 'ts-loader',
+          use: {
+            loader: 'ts-loader',
+            options: {
+              // TODO: disable type checking for now
+              // enable this when we refactor all TypeScript
+              transpileOnly: true,
+            },
+          },
         },
         {
           test: /\.css$/,
@@ -72,6 +79,9 @@ module.exports = function(env, argv) {
           ],
         },
       ],
+    },
+    resolve: {
+      extensions: ['.ts'],
     },
   };
 };


quisquous:
I'm not sure I understand the reasoning behind this.  Why disable type checking, even temporarily?

MaikoTan:
Because we are now mixed JavaScript and TypeScript in building, `tsc` would report a bunch of type errors, so temporary turn it off in webpack configuration.
Type checking for only `*.ts` is still live when lint with `tsc --noEmit` that I believe @Trim21 had written that.

Trim21:
If we refactor TypeScript from basic library file, I think we don't need to disable type check.

quisquous:
Does `"checkJs": false,` in the tsconfig.json not prevent such errors?

MaikoTan:
@quisquous The `"type": "module",` option in `package.json` might need to be removed cause webpack complains:

```
ERROR in ./ui/raidboss/popup-text.js 5:0-46
Module not found: Error: Can't resolve '../../resources/regexes' in '/mnt/e/repo/ACT_FFXIV/cactbot-workspace/cactbot/ui/raidboss'
Did you mean 'regexes.ts'?
BREAKING CHANGE: The request '../../resources/regexes' failed to resolve only because it was resolved as fully specified
(probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
The extension in the request is mandatory for it to be fully specified.
Add the extension to the request.
 @ ./ui/raidboss/raidemulator.js 10:0-53 166:53-71
```

quisquous:
> > Does `"checkJs": false,` in the tsconfig.json not prevent such errors?
> 
> @quisquous I found this would not happen when refactored common dependencies first, so never mind.
> 
> (But
> 
> ```
>     resolve: {
>       extensions: ['.ts'],
>     },
> ```
> 
> part should be added in, otherwise webpack cannot recognize `.ts` files.

That's interesting, because I felt like it works for me when I changed `resources/util.ts`, e.g.

```
> webpack-dev-server --config webpack/webpack.config.cjs --mode development

i ∩╜ówds∩╜ú: Project is running at http://localhost:8080/
i ∩╜ówds∩╜ú: webpack output is served from /
i ∩╜ówdm∩╜ú: assets by path *.js 15.7 MiB 16 assets
assets by path resources/ 2.02 KiB
  asset resources\util.d.ts 1.31 KiB [emitted]
  asset resources\util.d.ts.map 731 bytes [emitted]
```

> But the `"type": "module"` option in `package.json` might need to be removed cause webpack complains:
> 
> ```
> ERROR in ./ui/raidboss/popup-text.js 5:0-46
> Module not found: Error: Can't resolve '../../resources/regexes' in '/mnt/e/repo/ACT_FFXIV/cactbot-workspace/cactbot/ui/raidboss'
> Did you mean 'regexes.ts'?
> BREAKING CHANGE: The request '../../resources/regexes' failed to resolve only because it was resolved as fully specified
> (probably because the origin is a '*.mjs' file or a '*.js' file where the package.json contains '"type": "module"').
> The extension in the request is mandatory for it to be fully specified.
> Add the extension to the request.
>  @ ./ui/raidboss/raidemulator.js 10:0-53 166:53-71
> ```

Where are these errors coming from?

I think in general, I would prefer to keep "type: module" on, because that's what we're using.  I think I don't understand why that would be an issue, as module type (CJS vs ESM) is not related to typescript vs javascript, other than just extension?

> Although something like `import Regexes from './regexes.ts';` also work in webpack while `"type": "module"` option enabled, but then `tsc` complains:
> 
> ```
> An import path cannot end with a '.ts' extension. Consider importing './regexes' instead. ts(2691) [1, 33]
> ```

Where are we using tsc directly? How do I reproduce these errors.

> I think adding this option at the first is to invoke the ESM test js file via node without any extensions, right?
> But since we have successfully refactored our tests into Mocha, I think this option is not needed anymore?

I think the `module` setting is really to say that .js files are ESM and not CJS.  I think setting "type: module" is required to use ESM with ts-node, as per https://github.com/TypeStrong/ts-node/issues/1007

MaikoTan:
But it really reported these errors in my PR https://github.com/quisquous/cactbot/actions/runs/585495205

I haven't change any settings with eslint right?

quisquous:
Oh! Maybe that's the issue then.  Maybe this is the first instance of a .ts file importing another .ts file.

PR #2483, opened 2021-02-17
test: test timeline files with mocha

quisquous:
This is an adaptation of @panicstevenson's https://github.com/panicstevenson/cactbot/tree/convert-timeline-test branch.
I've left a lot of TODOs.

In the short term, files that are excluded in the mocharc.js file should be converted to use describe/it style tests, and once they've been converted, then remove them from the mocha exclusion and then delete the python scaffolding that calls them.


MaikoTan:
Would it be helpful to add `--bail` and `--parallel` options in moch cli?

quisquous:
> Would it be helpful to add `--bail` and `--parallel` options in mocha cli?

Hmm! I'm a bit iffy on including `--bail`.  I find it pretty useful sometimes to know if I have 100 errors or just one, and it's kinda frustrating when you fix one and have to run it again to find the next one.  I think for the test gh-action, it should not have `--bail` but if you feel strongly that it'd be nice to have locally and we can figure out a way to only have that happen locally (or maybe we only include that for lint-staged?), then maybe that's a way forward?

`--parallel` does not seem to make any performance difference here.  Maybe it's just me, but I find the mochajs documentation very hard to follow.  I think (and please correct me if I'm wrong), but it seems like the parallelization is at the per file level.  The vast majority of the time in cactbot testing is spent doing the timeline translation collision detection (~250ms per timeline x 150 timelines = ~30 seconds, although maybe we should improve this?), which is all in a single file, and so there ends up being no advantage.  Arguably there's a disadvantage in that all the output is buffered so you can't see progress until it's done thirty seconds later.

MaikoTan:
> I think for the test gh-action, it should not have `--bail` but if you feel strongly that it'd be nice to have locally and we can figure out a way to only have that happen locally (or maybe we only include that for lint-staged?), then maybe that's a way forward?

What about to have two test npm scripts, for example, `npm run test` for local with `--bail` enabled, and `npm run test:ci` for github actions (we can also add more detailed test methods like code coverage to this version).


> I think (and please correct me if I'm wrong), but it seems like the parallelization is at the per file level. 

Yes it is.

> The vast majority of the time in cactbot testing is spent doing the timeline translation collision detection (~250ms per timeline x 150 timelines = ~30 seconds, although maybe we should improve this?), which is all in a single file, and so there ends up being no advantage.  Arguably there's a disadvantage in that all the output is buffered so you can't see progress until it's done thirty seconds later.

I think this is the most part that we need parallel.
It is no matter to follow the order when testing various timelines, so we can modify this part later, for example, via creating background workers (worker works in single thread if I am correct?) to do the such testing.

quisquous:
> > I think for the test gh-action, it should not have `--bail` but if you feel strongly that it'd be nice to have locally and we can figure out a way to only have that happen locally (or maybe we only include that for lint-staged?), then maybe that's a way forward?
> 
> What about to have two test npm scripts, for example, `npm run test` for local with `--bail` enabled, and `npm run test:ci` for github actions (we can also add more detailed test methods like code coverage to this version).

This seems reasonable.  Let me follow up with this after landing the initial mocha changes.  I think it's better to just get this conversion done so typescript work can proceed in parallel, and then we can make testing better from there.

> > I think (and please correct me if I'm wrong), but it seems like the parallelization is at the per file level.
> 
> Yes it is.

Thanks!

> > The vast majority of the time in cactbot testing is spent doing the timeline translation collision detection (~250ms per timeline x 150 timelines = ~30 seconds, although maybe we should improve this?), which is all in a single file, and so there ends up being no advantage.  Arguably there's a disadvantage in that all the output is buffered so you can't see progress until it's done thirty seconds later.
> 
> I think this is the most part that we need parallel.
> It is no matter to follow the order when testing various timelines, so we can modify this part later, for example, via creating background workers (worker works in single thread if I am correct?) to do the such testing.

Yeah, I agree that there's no ordering constraint and also that it is the part that needs it the most.

I'm taking a look at it now and I think if I drop all of the Regex.parses and store regexes directly in a Map, that will speed in up by 2x.  I see another opportunity to speed it up as well by matching everything once and then doing a second pass with anything that matched the first time.  Let me try a few things here first, and then we can see what is necessary here or not.

panicstevenson (approved):

quisquous:
Ok, I think I'm going to land this now, and we can follow up with --bail / --parallel changes as needed.

PR #2484, opened 2021-02-17
Raidboss: Targetable timings O8N and E6N

JLGarber:
I have been annoyed with my bad timings on the E6N timeline for ages but never annoyed enough to remember to fix it. (It was from before we had a helpful flag to auto-generate these timings, but I still should have done better at the time.) As long as I was doing this, I figured I would also drop in timings for Kefka normal, since I ran that after. I left Kefka as separate commits since that's an older timeline. I don't plan to regenerate it right now, but I might look into it at some point in the future. If it's better to just do all that at once later, I can revert the Kefka changes.


quisquous:
I didn't realize the o8n timeline doesn't have syncs for all abilities.  I guess that needs a timelineNeedsFixing too.  :C

PR #2485, opened 2021-02-17
Eureka: Update Bozja opcodes for v5.45hotfix

### LINKED ISSUE #2482: in Bozja,the CE can not be trigger.other is ok
## Question

What would you like to ask?

KattTails:
I've also removed the old opcodes from the active code. I had hoped they would be useful for the other regions but it doesn't look like that is going to be the case. I've left a comment with the previous opcodes for reference.

This also fixes #2482


quisquous:
Thank you, as always! I am sorry this feature is so much ongoing hassle.

PR #2491, opened 2021-02-18
test: speed up timeline by reducing n in O(n^2) op

quisquous:
The pre-replacement collision tests that two regexes that apply to the
same text can be applied in either order without causing issues.  In
these cases, we only need to consider the regexes that apply to
a particular item, and not every single regex.

This is about a 20% speedup.  Unfortunately, there's no way to make
the post-collision test any better.

Mentioned in #2483.

### DIFF
--- a/test/helper/test_timeline.js
+++ b/test/helper/test_timeline.js
@@ -132,14 +132,21 @@ const testTimelineFiles = (timelineFiles) => {
 
             // For both texts and syncs...
             for (const testCase of testCases) {
-            // For every unique replaceable text or sync the timeline knows about...
+              // For every unique replaceable text or sync the timeline knows about...
               for (const orig of testCase.items) {
-              // For every translation for that timeline...
+                // For every translation for that timeline...
+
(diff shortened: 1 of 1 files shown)


MaikoTan:
I can recall there is a algorithm (data sturcture?) called trie that helps finding coilision in just one pass.
Maybe we can search in npmjs for some reliable implementations? 
https://www.npmjs.com/search?q=trie

quisquous:
> I can recall there is a algorithm (data sturcture?) called trie that helps finding coilision in just one pass (or O(n)).
> Maybe we can search in npmjs for some reliable implementations?
> https://www.npmjs.com/search?q=trie

I remember what a trie is, but I'm not quite sure I understand how it applies here.  I think the issue is really that these are generalized regular expressions, and so there's no way to know that replacing `/regex/` with `str` will create a conflict with some arbitrary `/otherRegex/` without checking that regex against the replaced string (and thus doing that work for all regexes).  If you see a way to speed this up with a trie or some other data structure, I'd be interested in hearing what the algorithm looks like because I am not able to see it.

PR #2493, opened 2021-02-19
Jobs: New DoT support for BLU

Echoring:
tormentBox will also trace DoT from AetherialSpark and Nightbloom now.
These new DoT cannot be stacked with SongOfTorment.

### DIFF
--- a/ui/jobs/components/blu.js
+++ b/ui/jobs/components/blu.js
@@ -35,6 +35,7 @@ export function setup(bars) {
     tormentBox.duration = 0;
     tormentBox.duration = 30;
   });
+  // +0.5&0.8 for animation delay
   bars.onUseAbility(kAbility.AetherialSpark, () => {
     tormentBox.duration = 0;
     tormentBox.duration = 15 + 0.5;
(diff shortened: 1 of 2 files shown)


ui/jobs/components/blu.js
```diff
@@ -35,6 +35,14 @@ export function setup(bars) {
     tormentBox.duration = 0;
     tormentBox.duration = 30;
   });
+  bars.onUseAbility(kAbility.AetherialSpark, () => {
+    tormentBox.duration = 0;
+    tormentBox.duration = 15 + 0.5;
+  });
+  bars.onUseAbility(kAbility.Nightbloom, () => {
+    tormentBox.duration = 0;
+    tormentBox.duration = 60 + 0.8;
```
quisquous, line 44: Where does the +0.8 and +0.5 come from?

Echoring: Animation delay.
In fact I'm thinking about use gain effect instead If this can make countdown more accurate...

MaikoTan: What about `onMobGainEffectFromYou` just like what you did in BRD module?

Echoring: hmmm... gaineffect is also not precise, just let it be. I'll make a comment.

quisquous:
Thanks for taking a look, and for adding a comment!

PR #2495, opened 2021-02-20
typescript: update resources/util.ts

quisquous:
This also updates the mocha options to support running with typescript.
On the downside, this also makes tests go as slow as they used to
(~30s) before the recent speedup.


quisquous:
I don't really know what I'm doing with typescript yet.  I assume I should move Job and Role to live in a definition file?

quisquous:
cc @MaikoTan if you have any time to look at this, since I think this will fix some of the test issues from your patch

resources/util.ts
```diff
@@ -0,0 +1,152 @@
...
+    canStun: (job: Job) => stunJobs.includes(job),
+    canSilence: (job: Job) => silenceJobs.includes(job),
+    canSleep: (job: Job) => sleepJobs.includes(job),
+    canCleanse: (job: Job) => cleanseJobs.includes(job),
+    canFeint: (job: Job) => feintJobs.includes(job),
+    canAddle: (job: Job) => addleJobs.includes(job),
+  };
+})();
+
+export default Util;
```
Trim21, line 152: Why IIFE?

quisquous: Mmmmm I can't remember why it was that way originally.  I'll remove the IIFE to make it cleaner.

PR #2496, opened 2021-02-21
refactor: migrate outputs and conditions to TS

Trim21:


quisquous:
I think some of this is also in #2466?

Trim21:
> I think some of this is also in #2466?

resources/conditions.ts is duplicated...

MaikoTan:
Yep, I removed it 'cause Trim does it better.

resources/outputs.ts
```diff
@@ -39,15 +12,15 @@ const numberToOutputString = (n) => {
...
-};
+}
 
 // General guidelines:
 // * property names should closely match English text
 // * use OnPlayer suffix for things with `${player}`
 // * use OnTarget suffix for things with `${name}`
 // * any other parameters (of which there are none, currently) should use consistent suffixes.
 // * the value of each property should be a single object with localized keys
-const Outputs = new ThrowOnInvalidProxy({
```
quisquous, line 50: I wonder if this is worth keeping around until all triggers are converted.

Trim21: It looks tests passed, so `.toJSON()` isn't coveraged yet?

How can I test this 🤔

quisquous: `.toJSON` is just there to support somebody doing `JSON.stringify` on `Outputs`.

This proxy is there to do what typescript will do in the future, in that it validates that any key you call on Outputs is valid.  You can test this by changing any access on Outputs and running npm run test, e.g. change `e12s.js` line 546 `junctionWithCast: Outputs.spread` to `junctionWithCast: Outputs.thisDoesntExist` and then run tests.  You'll get an error.

Once all the trigger files are converted to typescript such that typescript can provide this error, I think this can be safely removed.

Trim21: image

image

In current tests, I think we could remove this proxy?

quisquous: Oh!  If tests already catch undefined output strings, then that seems good to remove after all.  Sorry!

quisquous:
:tada: Thank you!!

PR #2499, opened 2021-02-21
style: add eslint rules

### LINKED ISSUE #2497: Proposal: limit function definition style
There are too many function definition style in code. add eslint rule [func-style](https://eslint.org/docs/rules/func-style) to allow only some of them for code readability.

Trim21:
set func-style to `'expression', { 'allowArrowFunctions': true }` in TS.
Doesn't affect old js tyle, only ts files.

close #2497


.eslintrc.cjs
```diff
@@ -35,6 +35,20 @@ module.exports = {
...
+          'multiline': {
+            'delimiter': 'semi',
+            'requireLast': true,
+          },
+          'singleline': {
+            'delimiter': 'semi',
+            'requireLast': true,
+          },
+        }],
+        'object-shorthand': 'error',
```
quisquous, line 48: I'm kind of indifferent to this rule, but it looks like it's `--fix`able so, :+1:

.eslintrc.cjs
```diff
@@ -35,6 +35,20 @@ module.exports = {
...
+        '@typescript-eslint/method-signature-style': ['error', 'method'],
```
quisquous, line 49: Why `method` over `property`? It seems like you get better strict checking with `property` as per https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/method-signature-style.md#enforces-using-a-particular-method-signature-syntax-method-signature-style

.eslintrc.cjs
```diff
@@ -35,6 +35,20 @@ module.exports = {
       'rules': {
         '@typescript-eslint/no-non-null-assertion': 2,
         '@typescript-eslint/no-explicit-any': 2,
+        '@typescript-eslint/member-delimiter-style': ['error', {
```
quisquous, line 38: :+1:

.eslintrc.cjs
```diff
@@ -35,6 +35,20 @@ module.exports = {
...
+            'requireLast': true,
+          },
+          'singleline': {
+            'delimiter': 'semi',
+            'requireLast': true,
+          },
+        }],
+        'object-shorthand': 'error',
+        '@typescript-eslint/method-signature-style': ['error', 'method'],
+        '@typescript-eslint/explicit-module-boundary-types': ['error'],
```
quisquous, line 50: This seems good, especially with `allowDirectConstAssertionInArrowFunctions: true` as a default.

Do you think `allowHigherOrderFunctions: false` is worth setting?

quisquous:
> set func-style to `'expression', { 'allowArrowFunctions': true }` in TS.
> Doesn't affect old js tyle, only ts files.
> 
> close #2497

...and also set like six other things :stuck_out_tongue:

.eslintrc.cjs
Trim21: "method" uniform the style with object-shorthand

.eslintrc.cjs
Trim21: > Do you think `allowHigherOrderFunctions: false` is worth setting?

Good idea.

.eslintrc.cjs
quisquous: I think `object-shorthand` makes sense for JavaScript, but using `method-signature-style: property` fixes type issues that could lead to bugs.  So, even if `object-shorthand` looks better, I think the stronger typing is better overall.  Maybe the compromise is to say `object-shorthand: ['error', 'consistent']` so that it could be used for objects that do not have methods? Alternatively, we could leave it unset?

Trim21: > I think the stronger typing is better overall.

In my understanding, Both style contains same type information.

```typescript
// method-signature-style: method
export interface Data {
  CanAddle(s: string): boolean;
}

// method-signature-style: property
export interface Data {
  CanAddle: (s: string) => boolean;
}
```

quisquous: As per https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/method-signature-style.md#enforces-using-a-particular-method-signature-syntax-method-signature-style:
> A good practice is to use the TypeScript's strict option (which implies strictFunctionTypes) which enables correct typechecking for function properties only (method signatures get old behavior).
> 
> TypeScript FAQ:
> 
> A method and a function property of the same type behave differently. Methods are always bivariant in their argument, while function properties are contravariant in their argument under strictFunctionTypes.

I had to go read https://www.stephanboyer.com/post/132/what-are-covariance-and-contravariance to understand what covariance/contravariance/bivariance means.  I think the root of it is that it adds strictness when checking if two functions are type equivalent when the arguments to those functions are subtypes of each other.

PR #2507, opened 2021-02-23
Better CN callouts for TEA pre-observation

miaonima:
Make buff names easier to recognize and consistent with TEA Beta Instructions Callout, and provide instructions for ORANGE BAIT and PURPLE BAIT (as in other translations). Confirmed this is the popular strategy (if not the only viable strategy) for this set of mechanics on CN servers.

I saw many raiders in the CN TEA discussion group would like to disable these 4 triggers because of the confusing official names. The buffs are more easily recognized by the color rather than their official names.


Trim21:
Not a good idea... Maybe `大光` `大暗` is even better

miaonima:
> Not a good idea... Maybe `大光` `大暗` is even better

I agree. Will close this and revise.

ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
```diff
@@ -1,6 +1,6 @@
-import Conditions from '../../../../../resources/conditions.ts';
+import Conditions from '../../../../../resources/conditions.js';
 import NetRegexes from '../../../../../resources/netregexes.js';
-import Outputs from '../../../../../resources/outputs.ts';
+import Outputs from '../../../../../resources/outputs.js';
```
Trim21, line 3: why making this change？

miaonima:
Sorry for the mess! I'm quite new to git. Please let me know if this is correct now.

quisquous:
Seems good to me to match what other languages are doing, especially if other people don't like it.  Thanks for the feedback!

@Trim21 does this look good to you?

(Also, sorry that git is confusing, especially when you're new to it.  Thanks for making it work!  I am happy to help in the future if git behaves badly and you have questions or need something fixed.)

Trim21:
@quisquous LGTM

PR #2514, opened 2021-02-27
Lint: add eslint rule for linting outputStrings

MaikoTan:


eslint/cactbot-output-strings.js
```diff
@@ -0,0 +1,111 @@
...
+            if (!outputStringProps) return;
+            const textFunctions = node
+              .properties
+              .filter((prop) => textProps.includes(prop.key.name))
+              .map((prop) => prop.value);
+            textFunctions.forEach((func) => {
+              if (t.isFunctionExpression(func) || t.isArrowFunctionExpression(func)) {
+                const param = func.params[2] && func.params[2].name;
+                const source = sourceCode.getText(func);
+                const m = source.match(new RegExp(`(?<=${param}\\.)(\\w+)(?=\\()`, 'g'));
```
quisquous, line 87: Can you `const param` => `const outputParam` for readability?

Do you need to check `if (!param) return;` here?

I think `(?<=${param}\\.)` should be `(?<=\b${param}\\.)` as well?

eslint/cactbot-output-strings.js
```diff
@@ -0,0 +1,111 @@
...
+      description: 'suggest outputStrings in cactbot',
+      category: 'Stylistic Issues',
+      recommended: true,
+      url: 'https://github.com/quisquous/cactbot/blob/main/docs/RaidbossGuide.md#trigger-elements',
+    },
+    fixable: 'code',
+    schema: [],
+    messages: {
+      noOutputStrings: 'no outputStrings in trigger',
+      notFoundProperty: 'not found \'{{prop}}\' in \'{{outputParam}}\'',
```
quisquous, line 45:
```suggestion
      notFoundProperty: 'no \'{{prop}}\' in \'{{outputParam}}\'',
```

eslint/cactbot-output-strings.js
```diff
@@ -0,0 +1,111 @@
...
+              .properties
+              .filter((prop) => textProps.includes(prop.key.name))
+              .map((prop) => prop.value);
+            textFunctions.forEach((func) => {
+              if (t.isFunctionExpression(func) || t.isArrowFunctionExpression(func)) {
+                const param = func.params[2] && func.params[2].name;
+                const source = sourceCode.getText(func);
+                const m = source.match(new RegExp(`(?<=${param}\\.)(\\w+)(?=\\()`, 'g'));
+                if (m) {
+                  m.forEach((output) => {
```
quisquous, line 89: `output` => `outputProperty`?

quisquous (commented):
I pulled this change locally and tried to get it to run with `npm run lint` or in VSCode but could not get it to generate any errors.  Am I missing someting?

MaikoTan:
Sorry, I wrote the wrong glob in eslint configuration file, now it shoild be working.

eslint/cactbot-output-strings.js
```diff
@@ -0,0 +1,111 @@
...
+            const textFunctions = node
+              .properties
+              .filter((prop) => textProps.includes(prop.key.name))
+              .map((prop) => prop.value);
+            textFunctions.forEach((func) => {
+              if (t.isFunctionExpression(func) || t.isArrowFunctionExpression(func)) {
+                const param = func.params[2] && func.params[2].name;
+                const source = sourceCode.getText(func);
+                const m = source.match(new RegExp(`(?<=${param}\\.)(\\w+)(?=\\()`, 'g'));
+                if (m) {
```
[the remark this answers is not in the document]
MaikoTan: Okay, i can do the refactoring soon.

eslint/cactbot-output-strings.js
MaikoTan: Yes, yes.

And in fact, there is an issue here. In somewhere, there are something like `output[func(xxx)]()` to dynamically invoke outputs, and cannot be catching by this regex, which can only catch `output.prop(`.

eslint/cactbot-output-strings.js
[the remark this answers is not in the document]
MaikoTan: Or should I extract them to be a new function?

eslint/cactbot-output-strings.js
quisquous: Yeah, I don't think there's anything to be done about using a variable.  Maybe typescript could help a little here, but it feels a bit unsolvable from a linting/testing perspective.

quisquous:
image

Is there a way to only highlight the particular output property rather than the entire function?

image

I understand that this is an error on the "whole trigger", but this is a bit jarring to have a huge error, especially when writing new triggers.  Is there some way to make this less loud? (Maybe only highlight the alertText?) Or maybe "missing output strings entirely" something better left for tests than syntax highlighting?

eslint/cactbot-output-strings.js
```diff
@@ -0,0 +1,112 @@
...
+    const sourceCode = context.getSourceCode();
+    const spreadNameMaps = new Map();
+    return {
+      VariableDeclarator(node) {
+        if (t.isIdentifier(node.id) && t.isObjectExpression(node.init))
+          spreadNameMaps.set(node.id.name, node.init.properties);
+      },
+      ObjectExpression(node) {
+        const propNames = getAllKeys(node.properties, spreadNameMaps);
+        if (propNames.includes('id') && propNames.some((name) => textProps.includes(name))) {
```
quisquous, line 58: I think this could be an early out as well.

MaikoTan:
> Is there a way to only highlight the particular output property rather than the entire function?

There is. It is just a bit complecated so I only check the source of whole function at the first time. I can change like that if you prefer.

> I understand that this is an error on the "whole trigger", but this is a bit jarring to have a huge error, especially when writing new triggers.  Is there some way to make this less loud? (Maybe only highlight the alertText?) Or maybe "missing output strings entirely" something better left for tests than syntax highlighting?

Sounds good to only highlight the `xxxText` property.

quisquous:
> > Is there a way to only highlight the particular output property rather than the entire function?
> 
> There is. It is just a bit complecated so I only check the source of whole function at the first time. I can change like that if you prefer.

If it's not too unreasonable, I think that'd be better?  If it's extremely complicated, then even just the alertText vs the entire text of the function would be better.

MaikoTan:
> > > Is there a way to only highlight the particular output property rather than the entire function?
> > 
> > There is. It is just a bit complecated so I only check the source of whole function at the first time. I can change like that if you prefer.
> 
> If it's not too unreasonable, I think that'd be better?  If it's extremely complicated, then even just the alertText vs the entire text of the function would be better.

AFAIK there are 2 ways to implement this, one is to calculate the relative position of the `output.prop` part, and generate `loc` for eslint dynamically. Another is complecated but robust, to search for `CallExpression` or `MemberExpression` in the current function.

quisquous:
> > > > Is there a way to only highlight the particular output property rather than the entire function?
> > > 
> > > 
> > > There is. It is just a bit complecated so I only check the source of whole function at the first time. I can change like that if you prefer.
> > 
> > 
> > If it's not too unreasonable, I think that'd be better?  If it's extremely complicated, then even just the alertText vs the entire text of the function would be better.
> 
> AFAIK there are 2 ways to implement this, one is to calculate the relative position of the `output.prop` part, and generate `loc` for eslint dynamically. Another is complecated but robust, to search for `CallExpression` or `MemberExpression` in the current function.

I think given that you're already searching for it via a regular expression, searching again to find its position doesn't seem too awkward.  I think if you were going to do the robust search, then I'd probably say finding the errors in the first place should be done that way.  I think either is fine.

MaikoTan:
I found that ESLint has a `getScope()` function which helps fetching variables declared in code, I think at least I can change the current code to use this function rather than mentain variables by the rule itself, and increase a little bit of robustness.
Also the "report output properties positions" issue can be done with it, too. I will work on it this evening when I have time.

MaikoTan:
image

image

Now the highlight scope is minified.

Use ESLint's selector syntax, which I am not very familiar with it.
Hope it would apply to all kinds of outputs (except computed indexing)

quisquous:
Thanks! This is great!

PR #2515, opened 2021-02-27
rename all ts imports to js imports

quisquous:


quisquous:
I think all imports should be `.js` with the extension at this point.  Maybe the `import/extensions` rule could help here to prevent importing files with no extensions.  I think once everything is converted to typescript then tsc will naturally prevent importing `.ts`, so maybe we don't need to bother with that rule.

cc @MaikoTan @Trim21

Trim21:
Why... What will happen if we import `.ts` files...

MaikoTan:
> Why... What will happen if we import `.ts` files...

See here: https://github.com/quisquous/cactbot/pull/2466#issuecomment-782918099
and this commit https://github.com/quisquous/cactbot/pull/2466/commits/5013e32e91249d4458abcacd3654f360433c1068

Trim21:
I think this is a really bad workaround...

Trim21:
I think we should remove `"type": "module",` from package.json and rename all `.js` files to '.mjs`, instead of importing `.ts` as `.js` extension.

quisquous:
The webpack loader is a workaround, but I think that is a webpack bug.  esbuild, node, VSCode, and eslint all handle this case properly.  I think there are more general reasons than just "make this work in cactbot" listed in the linked comment from Maiko.

Importing typescript files with an import than has an explicit ts extension is not an answer, as per the TS2691 error listed there.

From everything that I have read, importing typescript via a bare import specifier (no extension) or with a javascript extension are both supported.  Visual Studio even has an option to auto-complete typescript imports with .js extensions.  I think this is not some weird workaround, this is normal.

I think as a short-term solution, I would prefer to leave extensions as .js while we are converting to typescript, because it makes each typescript conversion have less churn.  You don't have to change two hundred files when converting one file to typescript because everything continues to work.

Trim21:
I think I can fix this these days. (with bare import) mocha is not a big problem here.

quisquous:
> I think I can fix this these days. (with bare import) mocha is not a big problem here.

I don't think there's anything wrong.  Everything is working at the moment.

Trim21:
I just fixed mocha problem....

quisquous:
> I think this is a really bad workaround...(also the webpack loader)

Can you explain where you're coming from?  I feel like you haven't explained your reasoning here and just keep saying it's bad.  It really seems to me that bare imports and js extensions are both valid ways to import a typescript file from another typescript file and I don't understand what you're objecting to.

Trim21:
@quisquous You are right, It works fine, and many tools support it.

But my objecting comes from 2 point.

1. The .js file being referenced doesn't exist; and won't exist until after it's been transpiled to js, and when that happens in many cases it'll be stored somewhere else. Source code shouldn't be referring to it's own build artifacts.
2. It can be fixed very easily in this way. https://github.com/quisquous/cactbot/pull/2526/files#diff-b55cdbef4907b7045f32cc5360d48d262cca5f94062e353089f189f4460039e0

Trim21:
So will we remove all .js extensions in a PR?

quisquous:
I'll abandon this change.

> So will we remove all .js extensions in a PR, or step by step

One PR seems fine.  Do you want to do it?

PR #2518, opened 2021-02-28
refactor: typescript-ize translation.js and party.js

MaikoTan:
@Trim21 帮我review一下


resources/party.ts
```diff
@@ -21,9 +24,17 @@ export default class PartyTracker {
...
 
     // role -> [names] but only for party
-    this.roleToPartyNames_ = {};
+    this.roleToPartyNames_ = {} as Record<Role, string[]>;
     for (const role of Util.getAllRoles())
       this.roleToPartyNames_[role] = [];
+  }
+
+  // Bind this to PartyChanged events.
+  onPartyChanged(e: { party: Party[]}): void {
```
Trim21, line 33: will it be better to add a types/`events.d.ts` for all events?

resources/translations.ts
```diff
@@ -0,0 +1,149 @@
...
+    en: 'You abandoned trial synthesis\\.',
+    de: 'Testsynthese abgebrochen\\.',
+    fr: 'Vous avez interrompu la synthèse d\'essai\\.',
+    ja: '(?<player>\\y{Name})は製作練習を中止した。',
+    cn: '(?<player>\\y{Name})停止了练习。',
+    ko: '제작 연습을 중지했습니다\\.',
+  },
+};
+
+type LocaleRegexesObj = Record<keyof typeof localeLines, Record<'en' | 'de' | 'fr' | 'ja' | 'cn' | 'ko', RegExp>>;
```
Trim21, line 114:
```TS
import { Lang } from '../types/global'

type LocaleRegexesObj = Record<keyof typeof localeLines, Record<Lang, RegExp>>;
```

resources/party.ts
MaikoTan: Those are typed for OverlayPlugin/Cactbot backend, I think to put them in the `global` file is well, not neccesary to create many files? They are not very long, even fill up all the types later.

resources/translations.ts
```diff
@@ -0,0 +1,150 @@
+import { Lang } from '../types/global';
+import Regexes from './regexes';
+import NetRegexes from './netregexes';
+
+// Fill in LocaleRegex so that things like LocaleRegex.countdownStart.de is a valid regex.
+const localeLines = {
```
quisquous, line 6:
```suggestion
const localeLines: { [key: string]: Record<Lang, string> } = {
```
I think it's expected that all languages are filled out in this file.  I think especially if we are going to treat these as records, we should enforce that here.  What do you think?

resources/translations.ts
```diff
@@ -0,0 +1,150 @@
...
+
+type LocaleRegexesObj = Record<keyof typeof localeLines, Record<Lang, RegExp>>;
+
+// Here overrides the build-in Proxy type definition
+// because it assumes that the source object type should
+// be the same as the target object, but here we use
+// proxy to dynamically return RegExp instead of strings
+// in the source object. So create a hacky Type to override
+// the default definition to make this work.
+interface OverrideProxyConstructor {
```
quisquous, line 123: Is it possible to not use a proxy here?  I recently sped up unittests a bunch by remove `Regexes.parse` lines (which these helpers both do).  One thing that was nice about the previous code is that it generates regexes for all languages once, and then never again.  I would prefer not to have the proxy dynamically compute it every time.  It's also a little bit suspicious to my eyes to have a Proxy for a Record (which should never return undefined) have a bunch of "return null" cases.

Is it possible to make a more generic object and then use `as LocaleRegexesObj` or something to tell typescript that it's correct? (I tried doing that, but then it thought that LocaleRegex.countdownCancel might be undefined, even with the `as`? Maybe you can figure out something else?) Alternatively, is there some other way of constructing this?

resources/party.ts
```diff
@@ -41,99 +52,99 @@ export default class PartyTracker {
...
+  jobName(name: string): Job | null {
     const partyIndex = this.partyNames.indexOf(name);
     if (partyIndex >= 0)
-      return Util.jobEnumToJob(this.details[partyIndex].job);
+      return Util.jobEnumToJob(this.details[partyIndex]?.job as number);
     return null;
   }
 
-  nameFromId(id) {
+  nameFromId(id: string): string | undefined {
```
quisquous, line 147: Maybe this is for a separate PR, but it is definitely odd that this one returns undefined but the others return null.

On the other hand, I think having the rest return undefined instead of null would clean up a bunch of the types.  You could say stuff like this, without needing all of the `as string` lines:

```
  otherHealer(name: string): string | undefined {
    const names = this.roleToPartyNames_['healer'];
    if (names.length !== 2)
      return;
    if (names[0] === name)
      return names[1];
    if (names[1] === name)
      return names[0];
  }
```

What do you think?

resources/translations.ts
MaikoTan: True and working for now. But do you remember the time that intl server updated patch 5.3 but CN/KR didn't, we had `en/de/fr/ja` completed but `cn/kr` lacking.
If this happens again, then the type as this would not be correct.

Or, something like this, ensure `en` required only?
```typescript
type LocaleLines = {
  [key: string]: {
    en: string;
  } & Partial<Record<Exclude<Lang, 'en'>, string>>;
};
```

resources/translations.ts
MaikoTan: Correct, I tried before but I found it is hard to simply add types on the original code.
I think if not to use proxy, the better way is to create a dummy class which caches regexes at the runtime and export the entries to access to `NetRegex` and `Regex`?

resources/party.ts
MaikoTan: So you means to modify other methods to return `undefined` too?

resources/translations.ts
quisquous: Could the proxy precalculate them all ahead of time and access the precalculated object that used to be returned?

resources/party.ts
quisquous: Yeah, I think I was suggesting that they all could return `string | undefined` and wondering if that would be more consistent / clean things up.  Feel free to disagree here.  It was more just me thinking out loud and wondering about the `as string` and the inconsistency.

resources/translations.ts
quisquous: Ah, that's true, that maybe they're missing.  But if they're missing then you can't have `Record<Lang, RegExp>` below, because it's not a Record anymore if there are missing keys.

I think there are two alternatives.  Either we make it not a record (like you suggest with Partial above) or we add in something (a fake duplicate regex of what `en` says, so at least it exists even if it's wrong?) for other languages just so something exists.  I don't think we need to list it in localeLines, but you could have the proxy still return something.

I just am worried about having a Record with missing keys.

resources/party.ts
```diff
@@ -41,99 +52,99 @@ export default class PartyTracker {
...
     const names = this.tankNames;
     if (names.length !== 2)
-      return null;
+      return;
     if (names[0] === name)
       return names[1];
     if (names[1] === name)
       return names[0];
-    return null;
+    return;
```
quisquous, line 124:
```suggestion
```

resources/party.ts
```diff
@@ -41,99 +52,99 @@ export default class PartyTracker {
...
+    const names = this.roleToPartyNames_['healer'];
     if (names.length !== 2)
-      return null;
+      return;
     if (names[0] === name)
       return names[1];
     if (names[1] === name)
       return names[0];
-    return null;
+    return;
```
quisquous, line 136:
```suggestion
```

resources/party.ts
```diff
@@ -41,99 +52,99 @@ export default class PartyTracker {
...
 
   // returns the job name of the specified party member
-  jobName(name) {
+  jobName(name: string): Job | undefined {
     const partyIndex = this.partyNames.indexOf(name);
     if (partyIndex >= 0)
-      return Util.jobEnumToJob(this.details[partyIndex].job);
-    return null;
+      return Util.jobEnumToJob(this.details[partyIndex]?.job as number);
+    return;
```
quisquous, line 144:
```suggestion
```

PR #2519, opened 2021-02-28
fix(type): regexes param can be undefined

MaikoTan:


resources/netregexes.ts
```diff
@@ -35,7 +35,7 @@ const changeZoneParams = ['id', 'name'] as const;
 const network6dParams = ['instance', 'command', 'data0', 'data1', 'data2', 'data3'] as const;
 
 const parseHelper = (
-    params: { timestamp?: string; capture?: boolean },
+    params: { timestamp?: string; capture?: boolean } | undefined,
```
Trim21, line 38: What's the difference between the style here and below.

MaikoTan: There are somewhere called regex helper functions without a parameter, for example:
https://github.com/quisquous/cactbot/blob/33bb7a866ac0b9ffc1d5912f9edac0169589f3e3/ui/jobs/utils.js#L11
Since we have `typeof f === 'undefined` type guard here, so would not break anything.

PR #2522, opened 2021-03-01
raidboss: DRS Trinity Seeker add chains

Legends0:
Chains (23: 0080) go out 3.111 seconds before Scorching hit, headmarker (1B: 01AC) appears 7.111 earlier than the hits. This is timed to when you should start moving then duration to hit(s) if you are late.

### DIFF
--- a/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.txt
+++ b/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.txt
@@ -56,7 +56,7 @@ hideall "--sync--"
 4105.5 "Baleful Onslaught" sync /:Trinity Seeker:5AD6:/
 4115.6 "Phantom Edge?" sync /:Trinity Seeker:5ABD:/
 4117.8 "--middle--" sync /:Trinity Seeker:5A9A:/
-4118.4 "--chains--" sync /23:0000:0000:0080:/ duration 3.1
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.txt
```diff
@@ -59,6 +56,7 @@ hideall "--sync--"
 4105.5 "Baleful Onslaught" sync /:Trinity Seeker:5AD6:/
 4115.6 "Phantom Edge?" sync /:Trinity Seeker:5ABD:/
 4117.8 "--middle--" sync /:Trinity Seeker:5A9A:/
+4118.4 "--chains--" sync /23:0000:0000:0080:/ duration 3.1
```
quisquous, line 59: Sorry, this is just a tiny nitpick, but in general I try to sync to abilities and starts using.  I'll use `00` lines and `03` lines if I have to and maybe `22` lines if I'm desperate, but even then I usually add an ability to sync to.  But, I am trying to keep the number of different line types down in timelines, esp as fflogs doesn't have this info and so test_timeline won't report this properly.

I think there's plenty of ability syncs around this point, so I think these would be fine without a sync (or put the sync in a comment?).

Legends0: I have commented the syncs. Thanks for the input on syncing these lines, I wasn't too sure if the syncs to 23 were ok to use.

Issue #2521, opened 2021-02-28
DRS Trigger for Trinity Seeker Incorrect


jekyll87:
The trigger currently listed for Trinity Seeker as "DelubrumSav Seeker Baleful Onslaught Stillness" is not actually a stillness check. The MT can continue to push actions and attack throughout. What the attack actually does is do a knock-back IF other tanks come to the front to try to share the TB. More accurately, this is a solo tank check cleave that does 0 damage and no knock-back if the MT is standing by themselves. 

I realize people can change the text from "Stillness Tank Cleave" to w/e they want but I feel something like "Solo Tank Check" is more accurate.

quisquous:
Ah, thanks! I feel like I saw streams and guides that said it was stillness.  I see that lyra's guide says this, and so maybe I misread it before or maybe it's been updated.  I'll update it to "Solo Tank Cleave".

jekyll87:
> 
> 
> Ah, thanks! I feel like I saw streams and guides that said it was stillness. I see that lyra's guide says this, and so maybe I misread it before or maybe it's been updated. I'll update it to "Solo Tank Cleave".

No problem. I think it's a common misconception from the first few weeks but it definitely only checks that the tank is alone then does nothing.

Legends0:
Oops, didn't realize @jekyll87 wrote this issue when he contacted me about it. I submitted a PR for it.

quisquous:
Should be fixed!

PR #2523, opened 2021-03-01
raidboss: DRS Trinity Seeker change Stillness Tank Cleave => Solo Tank Cleave

Legends0:
Knockback here occurs if the tank cleave is shared. The tank is able to move during the cleave.

### DIFF
--- a/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
+++ b/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
@@ -79,7 +79,7 @@ export default {
       },
     },
     {
-      id: 'DelubrumSav Seeker Baleful Onslaught Stillness',
+      id: 'DelubrumSav Seeker Baleful Onslaught Solo',
       netRegex: NetRegexes.startsUsing({ source: 'Trinity Seeker', id: '5AD6', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ source: 'Trinität Der Sucher', id: '5AD6', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ source: 'Trinité Soudée', id: '5AD6', capture: false }),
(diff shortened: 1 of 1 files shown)


Legends0:
Fixes #2521

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -79,16 +79,16 @@ export default {
...
       netRegexDe: NetRegexes.startsUsing({ source: 'Trinität Der Sucher', id: '5AD6', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ source: 'Trinité Soudée', id: '5AD6', capture: false }),
       netRegexJa: NetRegexes.startsUsing({ source: 'トリニティ・シーカー', id: '5AD6', capture: false }),
       alertText: (data, _, output) => output.text(),
       outputStrings: {
         text: {
-          en: 'Stillness Tank Cleave',
-          de: 'Stille Tank Cleave',
+          en: 'Solo Tank Cleave',
+          de: 'Solo Tank Cleave',
```
quisquous, line 91:
```suggestion
          de: 'Solo Tank Cleave',
```

In general, when changing translations, I usually remove everything else so that translators can come back and add a new translation.  (Unless this is proper German?)

Akurosia: You could change it to `Solo Tank Cleave` and would understood by 99%.
But if this is just a normal tank cleave, couldn't we just use the response for that (as for other solo cleaves we dont specify the solo part)

Legends0: Use of 'solo' over the same tank cleave puts more emphasis on its distinction from the other tank cleave in the same encounter, plus its near proximity with its first appearance which I think is why the extra word may be preferred. As the translation appears good enough I'll leave it in the PR and next time I'll keep in mind to remove the prior translations.

Up for further discussion if others think the generic response should be used instead.

PR #2525, opened 2021-03-01
Dualcast fix o4s.js

Aho-Senpai:
Changed comments from `Decisive Battle` to `Dualcast`

### DIFF
--- a/ui/raidboss/data/04-sb/raid/o4s.js
+++ b/ui/raidboss/data/04-sb/raid/o4s.js
@@ -41,7 +41,7 @@ export default {
       },
     },
     {
-      // Fire III not after Dualcast.
+      // Fire III + Dualcast.
       id: 'O4S Exdeath Fire III Counter',
       netRegex: NetRegexes.startsUsing({ id: '23F5', source: 'Exdeath', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ id: '23F5', source: 'Exdeath', capture: false }),
@@ -62,7 +62,7 @@ export default {
       },
     },
     {
-      // Blizzard III not after Dualcast.
+      // Blizzard III + Dualcast.
       id: 'O4S Exdeath Blizzard III',
       netRegex: NetRegexes.startsUsing({ id: '23F7', source: 'Exdeath', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ id: '23F7', source: 'Exdeath', capture: false }),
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/04-sb/raid/o4s.js
```diff
@@ -62,7 +62,7 @@ export default {
       },
     },
     {
-      // Blizzard III not after Decisive Battle.
+      // Blizzard III not after Dualcast.
```
quisquous, line 65: I do not understand this change.  Dualcast is the very first ability cast, and so everything is after dualcast.

If you look in the timeline, Decisive Battle happens ~1 minute into the fight.  Blizzard III before that point is 23F7.  Blizzard III after that point is 23FC.  I think the comment is correct.

Aho-Senpai: That doesn't sound quite right ?

https://github.com/Aho-Senpai/cactbot/blob/main/ui/raidboss/data/04-sb/raid/o4s.txt#L66

The main reason for the change is because the spells change if he cast `Dualcast` before or not, if he does, it's the double ground AoEs, if he doesn't (and thus tether his hand to the void) it's raidwide freeze (need to keep moving)

quisquous: I think there is some misunderstanding here.

There's phase 1 (before decisive battle) and phase 2 (after decisive battle) which use different ability ids.  I agree with you that decisive battle is not the critical piece here, and that dualcast is the reason.  However, looking at the timeline, it's still correct to say "before decisive battle" and "after decisive battle" to differentiate these two cases.

You are changing this to say "Blizzard III not after Dualcast" which is incorrect.  It is after Dualcast.  (Arguably everything is after Dualcast, as it is the first ability used, but this one is also immediately after Dualcast.)

I think if you want to change these comments to be in relation to Dualcast rather than in relation to which phase they are used in, it would be more useful to say something like "Blizzard III + Dualcast" for this one and "Blizzard III not dualcast" for the other.

Aho-Senpai: Yes, actually. Lemme change that. The main issue i had was more that a dualcast version of the cast can still happen after the first Decisive Battle, especially since there's 2 Decisive battle

quisquous:
Thanks for the clarification! I think it is much clearer now.

PR #2526, opened 2021-03-02
Fix/import ts in ts

Trim21:


quisquous:
Ah, I see!  This fixes the issue of typescript not emitting the correct extension when you use bare imports.  Thanks, this looks really clean.

Just so I understand, the idea is that we'd land this, and then change all the imports to not use extensions?

Trim21:
yes, the config added in webpack allow bare import in js files now.

quisquous:
Thanks!! :+1:

PR #2527, opened 2021-03-02
Jobs: SAM UI

Echoring:
sam
- Kenki amount
- Meditation stacks
- Shifu&Jinpu&Higanbana duration
- Tsubame-gaeshi cooldown
- Combo timer bar


ui/jobs/components/sam.js
```diff
@@ -1,6 +1,97 @@
...
+    tsubameGaeshi.duration = 60;
+  });
+
+  const higanbana = bars.addProcBox({
+    id: 'sam-procs-higanbana',
+    fgColor: 'sam-color-higanbana',
+  });
+  bars.onMobGainsEffectFromYou([
+    EffectId.Higanbana,
+  ], () => {
```
MaikoTan, line 82:
```suggestion
  bars.onMobGainsEffectFromYou(EffectId.Higanbana, () => {
```

ui/jobs/components/sam.js
```diff
@@ -1,6 +1,97 @@
...
+      meditationGauge.parentNode.classList.add('high');
+    else
+      meditationGauge.parentNode.classList.remove('high');
+  });
+
+  const shifu = bars.addProcBox({
+    id: 'sam-procs-shifu',
+    fgColor: 'sam-color-shifu',
+  });
+  bars.onYouGainEffect(EffectId.Shifu, (name, matches) => {
```
MaikoTan, line 43: There is a `onYouGainEffect` with shifu above too, it would be overrided if you call this function twice. So you either write down the shifu asignment here, otherwise merge these function into the function above.

In fact, the parameters of the function are **ID** and Matches, although it is not being used here.

MaikoTan (changes requested):

MaikoTan (approved):

quisquous:
Code looks fine to me.  Mostly I am wondering if it's worth displaying getsu/setsu/ka somewhere.  (Possibly underneath the combo bar in the same getsu/setsu/ka order that the in-game gauge uses as boxes like blm / smn has?  Or, maybe in some other way?)

MaikoTan:
It seems pointless to me to make it repeat on the bar since there are already enough prominent Sen(Setsu/Getsu/Ka) guage in the game?

quisquous:
> It seems pointless to me to make it repeat on the bar since there are already enough prominent Sen(Setsu/Getsu/Ka) guage in the game?

I mean, I understand that feeling, but the rest of the cactbot jobs ui for all other jobs entirely replaces the in-game gauge.  You can hide it off to the side and only use the cactbot stuff.  It's odd to me to have one gauge that you need to keep using.

Echoring:
Done. By the way, in-game order is setsu-getsu-ka in simple mode..
sen

quisquous (commented):
Looks great, thanks for adding the extra ui.  Just one small comment about some of the code, and then this is good to go.

ui/jobs/components/sam.js
```diff
@@ -1,6 +1,145 @@
...
+
+  const setsuContainer = document.createElement('div');
+  setsuContainer.id = 'sam-stacks-setsu';
+  stacksContainer.appendChild(setsuContainer);
+  const setsu = [];
+  for (let i = 0; i < 1; ++i) {
+    const d = document.createElement('div');
+    setsuContainer.appendChild(d);
+    setsu.push(d);
+  }
```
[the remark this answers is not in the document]
MaikoTan: Maybe the `setsuContainer`, `getsuContainer` and `kaContainer` are not needed, either.
I think just put the div elements in the `stackContainer` (or to rename it as `senContainer` would be more accurate?) is ok?

ui/jobs/components/sam.js
```diff
@@ -1,6 +1,121 @@
...
+    else
+      sen[0].classList.remove('active');
+    if (jobDetail.getsu > 0)
+      sen[1].classList.add('active');
+    else
+      sen[1].classList.remove('active');
+    if (jobDetail.ka > 0)
+      sen[2].classList.add('active');
+    else
+      sen[2].classList.remove('active');
```
MaikoTan, line 60:
```suggestion
    if (jobDetail.setsu)
      sen[0].classList.add('active');
    else
      sen[0].classList.remove('active');
    if (jobDetail.getsu)
      sen[1].classList.add('active');
    else
      sen[1].classList.remove('active');
    if (jobDetail.ka)
      sen[2].classList.add('active');
    else
      sen[2].classList.remove('active');
```

setsu, getsu and ka are booleans.

PR #2532, opened 2021-03-04
refactor(resources): migrate WidgetList to TypeScript

Trim21:


resources/user_config.ts
```diff
@@ -1,46 +1,77 @@
...
+  options: {
+    id: string;
+    name: {
+      en: string;
+      de: string;
+      fr: string;
+      ja: string;
+      cn: string;
+      ko: string;
+    };
```
MaikoTan, line 35:
```suggestion
    name: Record<Lang, string>;
```

resources/user_config.ts
```diff
@@ -190,20 +219,24 @@ class UserConfig {
...
+        this.savedConfig[overlayName] || {},
+        this.optionTemplates[overlayName],
       );
 
       // If the overlay has a "Debug" setting, set to true via the config tool,
       // then also print out user files that have been loaded.
-      const printUserFile = options.Debug ? (x) => console.log(x) : (x) => {};
+      const printUserFile = options.Debug ? console.log : () => {
+        return undefined;
+      };
```
MaikoTan, line 231:
```suggestion
      const printUserFile = options.Debug ? console.log : () => undefined;
```

resources/user_config.ts
```diff
@@ -231,7 +264,19 @@ class UserConfig {
...
+              NetRegexes,
+              Regexes,
+              Responses,
+              Outputs,
+              Util,
+              ZoneId,
+              ZoneInfo,
+              Options,
+              ...window,
+            }, localFiles[jsFile]);
```
MaikoTan, line 279: Wise thought, but then we should use `this.ZoneId` to reference to those import names?

resources/user_config.ts
```diff
@@ -288,7 +333,7 @@ class UserConfig {
     });
   }
 
-  handleSkin(skinName) {
+  handleSkin(skinName: 'default' | 'lippe') {
```
MaikoTan, line 336: type should be `string`, although there are only 2 skins for now.

types/events.d.ts
```diff
@@ -0,0 +1,19 @@
+import { Lang } from './global';
+
+export interface LoadUserEvent {
```
MaikoTan, line 3: I'll prefer to use a event map just like what I did in #2518

types/global.d.ts
```diff
@@ -1 +1,13 @@
 export type Lang = 'en' | 'de' | 'fr' | 'ja' | 'cn' | 'ko';
+
+export type OverlayName = 'raidboss' | 'config'
+
+type callOverlayHandler = (param: { call: string; overlay: OverlayName }) => {};
```
MaikoTan, line 5: not neccesary to declare a type right? just merge them in Window declaration.

ui/dps/rdmty/dps.js
```diff
@@ -42,44 +42,44 @@ var ____Class0=React.Component;for(var ____Class0____Key in ____Class0){if(____C
         return (
             this.props.perSecond === '---' ? null :
             React.createElement("li", {
-                className: 'row ' + this.props.job.toLowerCase() + (this.props.isSelf ? ' self' : ''), 
-                onClick: this.props.onClick}, 
+                className: 'row ' + this.props.job.toLowerCase() + (this.props.isSelf ? ' self' : ''),
+                onClick: this.props.onClick},
```
MaikoTan, line 46: ？

MaikoTan (changes requested):

resources/user_config.ts
Trim21: I haven't tested this, I'm not sure it works.

package.json
```diff
@@ -53,14 +53,13 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "npm run lint-staged"
+      "pre-commit": "tsc --noEmit && npm run lint-staged"
```
Trim21, line 56: `tsc --noEmit $filename` ignore `tsconfig.json` and other files, So we have to run tsc on the whole project.

Or maybe we should add this in ci? @quisquous

MaikoTan: From TypeScript official:
```bash
# Transpile just the index.ts with the compiler defaults
tsc index.ts
```

resources/widgetlist.ts
```diff
@@ -1,17 +1,36 @@
...
   }
 
   // All visual dimensions are scaled by this.
-  set scale(s) {
-    this.setAttribute('scale', s);
+  set scale(s: string | null) {
+    if (s === null)
+      this.removeAttribute('scale');
+    else
+      this.setAttribute('scale', s);
```
MaikoTan, line 31: Maybe you can use conditional ternary operator to simplify these checking.

resources/widgetlist.ts
```diff
@@ -238,42 +250,33 @@ class WidgetList extends HTMLElement {
...
+    const container = this.shadowRoot?.getElementById(`child${id}`);
+    if (container !== null && container !== undefined) {
+      const element = container.childNodes[0];
+      this.rootElement.removeChild(container);
+      return element;
     }
-
-    this.layout();
-    return element;
+    return undefined;
```
MaikoTan, line 279:
```suggestion
    return;
```
?

resources/widgetlist.ts
```diff
@@ -285,32 +288,33 @@ class WidgetList extends HTMLElement {
...
     let count = 0;
 
-    for (const i in this._sorted) {
-      const id = this._sorted[i];
-      console.assert(id !== 0, 'An id in _sorted isn\'t in _elements?');
-      const container = this.shadowRoot.getElementById('child' + id);
-      console.assert(container !== null, 'Element with id child' + id + ' is missing?');
+    this._sorted.forEach((id: number) => {
+      if (id === 0)
+        throw new Error('An id in _sorted isn\'t in _elements?');
```
MaikoTan, line 301: Throwing an error would interupt the workflow, may be not intented?

types/data.d.ts
```diff
@@ -6,7 +6,12 @@ export interface Option {
   ParserLanguage: Lang;
   LangShortLocale: Lang;
   DisplayLanguage: Lang;
-  [key: string]: unknown;
+  Language: Lang;
+  SystemLocale: string;
+  ShortLocale: string;
+  Debug: boolean;
+  Skin: 'default' | 'lippe';
```
MaikoTan, line 13:
```suggestion
  Skin: string;
```

types/shadow-dom.d.ts
```diff
@@ -0,0 +1,106 @@
...
+  interface ShadowRoot extends DocumentFragment, DocumentOrShadowRoot {
+    host: HTMLElement;
+    mode: ShadowRootMode;
+  }
+
+  interface Element {
+    attachShadow: (shadowRootInitDict: ShadowRootInit) => ShadowRoot;
+
+    shadowRoot: ShadowRoot | null;
+
```
MaikoTan, line 22: This is already declared in `libdom.d.ts`, you dont need to duplicate it.

MaikoTan (changes requested):

resources/widgetlist.ts
Trim21: `console.assert()` will do the same thing.

types/shadow-dom.d.ts
Trim21: No, some properties are not. (I forget which ones, I'll reply after I finish `ui/test/test.ts`

resources/widgetlist.ts
MaikoTan: https://developer.mozilla.org/zh-CN/docs/Web/API/Console/assert
MDN tells me it wouldn't interrupt.

Trim21: The code should interrupt here, some value is null, and can't operate it.

resources/widgetlist.ts
```diff
@@ -21,41 +40,56 @@ class WidgetList extends HTMLElement {
   // and the second being the direction is wraps for the next
   // row/column. eg. "left down" will grow a list toward the left,
   // and subsequent rows will be below the first.
-  set toward(s) {
-    this.setAttribute('toward', s);
+  set toward(s: string | null) {
+    if (s === null)
+      this.removeAttribute('toward');
+    else
+      this.setAttribute('toward', s);
```
quisquous, line 47: I'm not sure I understand these changes to the setters.  Is it possible to have `set toward(s: string)` with no null possibility here, and elsewhere? The `removeAttribute` does not seem like how it behaved previously?

resources/widgetlist.ts
```diff
@@ -322,35 +330,26 @@ class WidgetList extends HTMLElement {
...
   test() {
     for (let i = 0; i < 8; ++i) {
       const div = document.createElement('div');
-      div.style.width = this._elementwidth * 3 / 4;
-      div.style.height = this._elementheight * 3 / 4;
+      div.style.width = String(this._elementwidth * 3 / 4);
+      div.style.height = String(this._elementheight * 3 / 4);
       div.style.overflow = 'hidden';
-      div.style.backgroundColor = '#' + parseInt(Math.random() * 10) + '' + parseInt(Math.random() * 10) + '' + parseInt(Math.random() * 10);
+      div.style.backgroundColor = `#${Math.random() * 10}${Math.random() * 10}${Math.random() * 10}`;
```
quisquous, line 342: I think this is not the same without the `parseInt`.  Probably this was intended to be `Math.round()` or `Math.floor()`?

ui/raidboss/autoplay_helper.ts
```diff
@@ -1,20 +1,20 @@
 export default class AutoplayHelper {
-  static Check() {
+  static Check(): boolean {
```
quisquous, line 2: For future conversions, can you do one file at a time (unless it's so interrelated that it's hard to do them separately)?  This is fine for now.

resources/widgetlist.ts
```diff
@@ -1,17 +1,36 @@
...
 // This class should be considered deprecated, and any users of this class should
 // just switch over to using CSS grid.
+type Sorter = () => number;
 
 class WidgetList extends HTMLElement {
+  private _nextId = 1;
+  private _nameToId: { [key: string]: number } = {};
+  private _elements: { [key: number]: Sorter } = {};
+  private _sorted: number[] = [];
+  private _elementwidth = 100;
```
quisquous, line 11: Also, I think it's redundant to set all of these values here and then re-set them in attached.  Can these just be types and then the default values only be set in one place?

resources/widgetlist.ts
Trim21: originally, it's `this.setAttribute('toward', w);`, which according to MDN, `setAttribute('prop', null)` means `removeAttribute`

quisquous: Can you link me to this documentation?

I guess my confusion here is that we are providing this `toward` API here, and so we can assert "you must pass a non-null string".  I don't think any caller of any of these components in resources would want to remove an attribute, only set the value.

Trim21: https://developer.mozilla.org/en-US/docs/Web/API/Element/setAttribute#parameters

MaikoTan: Seems like I should change like this in my PR too?

quisquous: I guess my question is more about "what is the api for `toward`".  I don't think anybody is or should be calling `toward = null`.    I also don't think removing attributes makes sense for this (or other) components?  I don't think any of the underlying code is expect attributes to be removed.  So, I was wondering more if this should be `set toward(s: string) {`, where we only accept a non-null string here.

(Actually, it looks like `toward` must be an array of two elements, each of which is up/down/left/right, if I'm reading the `parseToward` code below?)

Trim21: What the getter should return when `getAttribute` return `null`? Types in getter and setter should match.

quisquous: So, just trying this out, calling `removeAttribute` does not invoke the `attributeChangedCallback` and just sets the attribute to its default(?) value.  This seems like a browser bug, as the documentation says that it should call the `attributeChangedCallback`.  It seems like this would probably will cause issues for things that need to parse the attributes to set other attributes indirectly.

I guess the options are:
```
// This is what Maiko did in resourcebar.ts:

set toward(s: string | null) {
  this.setAttribute('toward', s ?? 'right down');
}
get toward(): string | null {
  return this.getAttribute('toward');
}

// ... OR ... ?

set toward(s: string) {
  this.setAttribute('toward', s);
}
get toward(): string{
  // maybe print an error here if this happens?
  return this.getAttribute('toward') ?? '';
}
```

Or something else?

Trim21: I think the below one works

quisquous:
Looks like this introduces a bug: 
image

Using two buffs (e.g. a potion and something else) should not stack on top of each other like that.

resources/widget_list.ts
```diff
@@ -21,115 +41,70 @@ class WidgetList extends HTMLElement {
...
-
+  attachedCallback() {
     this._connected = true;
     this.layout();
   }
 
-  disconnectedCallback() {
+  detachedCallback() {
     this._connected = false;
   }
```
MaikoTan, line 105: connectedCallback/disconnectedCallback is the modern one...
(草你这是喝了多少假酒)

MaikoTan (changes requested):

resources/widget_list.ts
```diff
@@ -1,18 +1,38 @@
 // NOTE
 // This class should be considered deprecated, and any users of this class should
 // just switch over to using CSS grid.
+type Sorter = () => number;
+
+const getRandomInt = function(max: number) {
+  return Math.floor(Math.random() * Math.floor(max));
+};
 
 class WidgetList extends HTMLElement {
```
MaikoTan, line 10:
```suggestion
export default class WidgetList extends HTMLElement {
```

MaikoTan (changes requested):

resources/widgetlist.ts
quisquous: Ok, sorry for the runaround on this.  I was looking at the revert of @MaikoTan's patch and @panicstevenson's fix in #2633.  I think I misunderstood how these getters are working, and didn't think they were getting used anywhere that would return null, but they are getting used in the attaching function when the component is being set up.  So, having them return null when unset is important.  I also think null is better for unset values than empty string, since we have some string attributes.  And, I agree with the previous idea that getters and setters should have the same type.

Now that I understand this better, I think @TRim21's original proposal of:
```typescript
  set toward(s: string | null) {
    if (s === null)
      this.removeAttribute('toward');
    else
      this.setAttribute('toward', s);
```

...is more consistent and better overall.  Sorry about my confusion.

Trim21: ok

Trim21:
> Looks like this introduces a bug:
> image
> 
> Using two buffs (e.g. a potion and something else) should not stack on top of each other like that.

image
Can't reproduce, mine looks good.

quisquous:
> Can't reproduce, mine looks good.

I cannot reproduce now either, so looks good to me.

> (any change can I test this without using a position?

There used to be an `/echo :test:jobs:` but I think that never got updated with the switch to NetRegexes.

quisquous:
Thank you!  Sorry for the confusion on my part about how best to do this.  I think I just fundamentally misunderstood how these web components worked and when the getters were being used.  You had the right idea to begin with, and I just didn't understand it enough.

PR #2534, opened 2021-03-05
format ui/dps/rdmty/dps.js

Trim21:
format this file with `eslint --fix`
at least we should remove trailing whitespaces


quisquous:
Why closed? This seems good?

Trim21:
> Why closed? This seems good?

I found that this file is generated...

quisquous:
It may have been generated originally, but it's not generated inside of cactbot.  It got included and then modified over time, so it's probably worth formatting?

PR #2545, opened 2021-03-06
raidboss: add slime/golem timelines

quisquous:


Legends0:
I can test this out today.

Legends0:
At least for golems, the timeline starts considerably late, probably around 50 seconds into the encounter. The golem spawn triggers also are off, so maybe I was wrong in my previous timeline about how often they spawn? For instance the rock golems were spawning about 8 seconds before the timeline.

Legends0:
I took a closer look at a Golem enrage log for the 03: addCombatants for Ruins Golem.

It happens every 40 seconds, but then once the last Bicolor Golem dies on the correct colored pad, the Ruins Golem spawn changes. I'm not sure how it adjusts for the first spawn, but it appears that it cuts off remaining time on the timer such that it is within 30 seconds of the previous spawn, then the next spawn becomes 20 seconds, and then the next are always 10 seconds.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.txt
```diff
@@ -11,9 +11,116 @@ hideall "--sync--"
...
+1344.7 "Death Scythe" #sync /:Grim Reaper:5747:/
+1347.7 "Death Scythe" #sync /:Grim Reaper:5747:/
+1350.8 "Death Scythe" #sync /:Grim Reaper:5747:/
+1353.8 "Death Scythe" #sync /:Grim Reaper:5747:/
+1356.8 "Death Scythe" #sync /:Grim Reaper:5747:/
+1359.8 "Death Scythe" #sync /:Grim Reaper:5747:/
+
+
+### Golems (The Granary, right side)
+2000.0 "--sync--" sync / 22:........:Bicolor Golem:........:01/ window 2000,0
```
quisquous, line 79: Oh! It's starting late because I messed this line up.  I think it needs to be:
`2000.0 "--sync--" sync / 22:........:Bicolor Golem:........:Bicolor Golem:01/ window 2000,0`

quisquous:
Do you think maybe it's worth just dropping the ruin golem spawns from the timeline, then? I'm not sure that that it'd be possible to handle that sort of dynamic cooldown shift.

Legends0:
> 
> 
> Do you think maybe it's worth just dropping the ruin golem spawns from the timeline, then? I'm not sure that that it'd be possible to handle that sort of dynamic cooldown shift. Or maybe including a few, and then dropping them as soon as they might get off?
I see what you mean. Seems simple to do a shift for the Ruins Golems, but to have the timeline still be accurate for the Grim Reaper becomes difficult. You'd basically need a trigger to track for change in Ruins Golems spawn as there isn't an ideal way to track that the Bicolor Golems have been successfully resolved outside of the rate of Ruins Golems changing. Then on top of that trigger, you need some other programming for handling where to jump to in a timeline that just contains 10 seconds interval Ruin Golems with Grim Reaper at the end.

Probably too complex for what it's worth tracking the Ruins Golem spawn changes. If we kept the Ruins Golem spawns every 40 seconds, it would be correct 25% of the time at worse case.

Legends0:
Speaking of Ruins Golem rate changes, do we know if there similar changes that occur for Slimes? I think in most situations the Golems will be the ones to run into the rate change issues and so if a similar issue also exists with the Slime timeline it is less likely to be encountered.

quisquous:
> Speaking of Ruins Golem rate changes, do we know if there similar changes that occur for Slimes? I think in most situations the Golems will be the ones to run into the rate change issues as they appear to more often resolve the mechanic before Slimes alliance and so if a similar issue also exists with the Slime timeline it is less likely to be encountered.

I am not 100% sure.  The slimes timeline has seemed spot on to me so far.  I think it's also the case that slimes run away as soon as they resolve the mechanic, so even if it does change, it doesn't ultimately matter as the mechanic is done at that point.

quisquous:
Do you think this is "good enough" to land with the start sync fixed? Or, would it be better to drop the ruins golems entirely off the timeline after the time when golems are likely to be done?

Legends0:
> 
> 
> Do you think this is "good enough" to land with the start sync fixed? Or, would it be better to drop the ruins golems entirely off the timeline after the time when golems are likely to be done?

Looks good enough to land to me.

PR #2548, opened 2021-03-07
Refactor: Typescript-ize overlay_plugin_api.js

MaikoTan:


MaikoTan:
https://github.com/quisquous/cactbot/blob/20289eb784eef4e2d4a1f2f5ae9b34f694ae64ba/ui/raidboss/emulator/overrides/RaidEmulatorOverlayApiHook.js#L1

Just found that the emulator hooks the apis from overlayplugin by re-assigning, so cause the test error.
How can I solve this?

quisquous (commented):
> https://github.com/quisquous/cactbot/blob/20289eb784eef4e2d4a1f2f5ae9b34f694ae64ba/ui/raidboss/emulator/overrides/RaidEmulatorOverlayApiHook.js#L1-L13
> 
> Just found that the emulator hooks the apis from overlayplugin by re-assigning, so cause the test error.
> How can I solve this?

One possibility is to consolidate all of the apis into a static class (e.g. `OverlayPluginApi`), and then export that static class.  Then, in any raidboss code that was calling `window.callOverlayHandler` (or any other function), store `OverlayPluginApi` as a member variable in that code, and so code would then do `this.overlayPlugin.callOverlayHandler`.  Then, the emulator can implement its own version of the `OverlayPluginApi` interface and override the `this.overlayPlugin` member varialbe.

MaikoTan:
I don't quite understand what you mean, are you saying that the file overlay_plugin_api exports a static object containing APIs like callOverlayHandler and then in the raidboss data file do something like:

```js
import OverlayPluginApi from 'path/to/overlay_plugin_api';

export default {
  zoneId: xxx,
  overlayPlugin: OverlayPluginApi,
  triggers: [
    {
      promise: async (data) => {
         await this.overlayPlugin.callOverlayHandler(); // blahblah
      },
    },
  ],
};
```

And then we just replace that property in the default exported object when we hook later?

quisquous:
Hmm! I think I was imagining a couple of options.

1) `data.overlayPluginApi` for triggers, and then the emulator could override the data member as needed.
2) Have some way to say OverlayPluginApi.setOverride(someEmulatorFakeApi). and then check in every function if it's been overridden.  Then code can still have a global `OverlayPluginApi` that is used.

Probably 1 is a little bit cleaner, but it's just "one more thing" on data, as well.

ui/raidboss/emulator/overrides/RaidEmulatorOverlayApiHook.js
```diff
@@ -1,16 +1,20 @@
...
+    this.originalAdd = addOverlayListener;
+    this.originalRemove = removeOverlayListener;
+    this.originalCall = callOverlayHandler;
     this.timestampOffset = 0;
 
     window.dispatchOverlayEvent = this.dispatch.bind(this);
-    window.addOverlayListener = this.add.bind(this);
-    window.removeOverlayListener = this.remove.bind(this);
-    window.callOverlayHandler = this.call.bind(this);
+    setOverride({
```
quisquous, line 13: Can `dispatchOverlayEvent` also be part of the overrides for consistency?

resources/overlay_plugin_api.ts
```diff
@@ -0,0 +1,221 @@
...
+  return p;
+};
+
+interface Overrides {
+  addOverlayListenerOverride?: <T extends EventType>(event: T, cb: EventMap[T]) => void;
+  removeOverlayListenerOverride?: <T extends EventType>(event: T, cb: EventMap[T]) => void;
+  callOverlayHandlerOverride?: IOverlayHandler;
+}
+
+const kOverrides: Overrides = {};
```
quisquous, line 204: Sorry about the `k` prefix in some places.  I really meant anything with a `k` to be typescript `readonly`.  Usually it's just for pre-declared constant data that is never changed.  I think this should just be `overrides`.

resources/overlay_plugin_api.ts
```diff
@@ -0,0 +1,221 @@
...
+        resolve(data === null ? null : JSON.parse(data));
+      });
+    });
+  }
+
+  return p;
+};
+
+interface Overrides {
+  addOverlayListenerOverride?: <T extends EventType>(event: T, cb: EventMap[T]) => void;
```
quisquous, line 199: Is it possible to shorten this with `addOverlayListenerOverride: typeof addOverlayListener` sort of thing, so that the written out type is not duplicated?

resources/overlay_plugin_api.ts
```diff
@@ -0,0 +1,221 @@
...
+
+const sendMessage = (
+    msg: { [s: string]: unknown },
+    cb?: (value: string | null) => unknown,
+): void => {
+  if (ws) {
+    if (queue)
+      queue.push(msg);
+    else
+      ws?.send(JSON.stringify(msg));
```
quisquous, line 41:
```suggestion
      ws.send(JSON.stringify(msg));
```

In this case, ws is always not-null, due to the `if` above?

types/event.d.ts
```diff
@@ -0,0 +1,311 @@
...
+    phoenixReady: number;
+  };
+  'RDM': {
+    whiteMana: number;
+    blackMana: number;
+  };
+}
+
+export interface EventMap {
+  // #region OverlayPlugin build-in Event
```
quisquous, line 122:
```suggestion
  // #region OverlayPlugin built-in Event
```

types/trigger.d.ts
```diff
@@ -1,21 +1,18 @@
...
 export type Matches = {
   target?: string;
   [s: string]: string;
 }
 
+export type TranslatedText = Record<Lang, string>;
 
-type TranslatableText = {
-  [s in Lang]: string;
+export type TranslatableText = {
```
quisquous, line 10: It is hard for me to tell the difference between what `TranslatedText` and `TranslatableText` are supposed to mean as type names.

I think my idea with `TranslatableText` was meant to be "this is an object that can be translated" (i.e. is translatable), with "translating" being going from object -> string by selecting one of the properties.  (There's a bunch of functions called `translate` that do this sort of thing.)  `TranslatedText` with past tense "translated", sounds like "this has been translated already", and makes me expect that it's a string.

I really don't have good ideas on what to call these.  `TranslatableText` is already an awkward mouthful.  I'd be hesitant to suggest `FullyTranslatableText` as something to call the `Record<Lang, string>` as that seems even worse.

cc @JLGarber for naming bikesheds if you have any opinions

quisquous (commented):
The other missing piece here is that `resources/user_config.js` also needs to import this, as these new functions should be available for user eval'd triggers.  See the code there where things like regexes is imported.

I do think that this will break backwards compatibility for anybody who was previously saying `window.callOverlayHandler` inside of a user trigger.  Is it possible to make this still continue to work for user triggers, or is this a breaking change, and I need to announce it?

MaikoTan:
> The other missing piece here is that `resources/user_config.js` also needs to import this, as these new functions should be available for user eval'd triggers.  See the code there where things like regexes is imported.

Done!

> I do think that this will break backwards compatibility for anybody who was previously saying `window.callOverlayHandler` inside of a user trigger.  Is it possible to make this still continue to work for user triggers, or is this a breaking change, and I need to announce it?

If user are using `window.callOverlayHandler`, then would be not working. But if they are using `callOverlayHandler` (without prefix `window`), then would not be affected.

quisquous:
> > I do think that this will break backwards compatibility for anybody who was previously saying `window.callOverlayHandler` inside of a user trigger.  Is it possible to make this still continue to work for user triggers, or is this a breaking change, and I need to announce it?
> 
> If user are using `window.callOverlayHandler`, then would be not working. But if they are using `callOverlayHandler` (without prefix `window`), then would not be affected.

Is there a way to make it work with window, but just for user triggers?  It's not like they are (or can be) typescript, so they don't benefit for any typing here.

Or, is this too complicated?

MaikoTan:
> Is there a way to make it work with window, but just for user triggers?  It's not like they are (or can be) typescript, so they don't benefit for any typing here.
> 
> Or, is this too complicated?

We can just override the global Window and do a `window.callOverlayHandler = callOverlayHandler` to ensure the backward compatiability?
I can write the JSDoc with `deprecated` flag to notify user to switch to the new way.

MaikoTan:
Oh sorry didn't see the review comments 'cause I am using github Android App, I'll fix them tonight.

types/trigger.d.ts
MaikoTan: What about `LocaleText`?

JLGarber: No, unfortunately I don't think there are any really good options here. I definitely like `LocaleText` in general, or maybe `LocalizableText`? In this case, given how specific we're being with type names, I think we can afford to go a little longer than we normally would for the sake of precision.

quisquous: `LocaleText` is good for me.  What about `FullLocaleText` for the `Record` version?

MaikoTan: Done

resources/overlay_plugin_api.ts
```diff
@@ -0,0 +1,265 @@
...
+    };
+
+    waitForApi();
+  }
+
+  // Assign in `window` object to ensure backward compatiability
+  window.addOverlayListener = addOverlayListener;
+  window.removeOverlayListener = removeOverlayListener;
+  window.callOverlayHandler = callOverlayHandler;
+  window.dispatchOverlayEvent = processEvent;
```
quisquous, line 264: Is it possible to only do this for user trigger files?

MaikoTan: Ok, I'll move them in user_config.js

MaikoTan: But the `dispatchOverlayEvent` should be still in this file, right ?

quisquous: Sorry for belated reply, as I'm a little bit busy/exhausted these days.  I am not actually sure what `dispatchOverlayEvent` is.  It doesn't look like it is called in cactbot, from what I can tell.  I also can't see that it's called via OverlayPlugin either.  So, maybe it can also be handled in the same way (i.e. exported, and only on window)?

Also, `window` is a global, so even if this is only set inside of user_config, that will still be visible to other files.  Is it possible to set and un-set it off of window around the `eval` blocks?

MaikoTan: > Sorry for belated reply, as I'm a little bit busy/exhausted these days. I am not actually sure what `dispatchOverlayEvent` is. It doesn't look like it is called in cactbot, from what I can tell. I also can't see that it's called via OverlayPlugin either. So, maybe it can also be handled in the same way (i.e. exported, and only on window)?
> 
> Also, `window` is a global, so even if this is only set inside of user_config, that will still be visible to other files. Is it possible to set and un-set it off of window around the `eval` blocks?

Oh, I thought it would be called by OverlayPlugin or something in Cactbot's C# backend, 'cause it looks like an event dispatcher callback.
Okay, I'll change like what you said.

quisquous:
Thank you!

PR #2552, opened 2021-03-07
raidboss: DRS additional triggers

Legends0:
Adds callouts for most AoEs and tankcleaves/busters.
Add Stygimoloch Lord triggers for Fateful Word and Devastating Bolt
Add Icy/Fiery Portent to Queen's Guard and The Queen
Add Bozjan Phantom triggers
Add The Queen Ball Lightning callout


ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -31,6 +31,15 @@ export default {
     },
   ],
   triggers: [
+    {
+      id: 'DelubrumSav Seeker Verdant Tempest',
```
quisquous, line 35: Usually I like to put aoes on castbars becuase it's simpler, but I think it might be worth putting this on a 10 second beforeSeconds timeline trigger instead, so that folks have time to heal up and throw mitigation?  What do you think?

Similarly the other aoes below?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -249,6 +258,103 @@ export default {
...
       },
     },
+    {
+      id: 'DelubrumSav Guard Fiery Portent',
+      netRegex: NetRegexes.startsUsing({ source: 'Queen\'s Soldier', id: '583F', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Soldat Der Königin', id: '583F', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Soldat De La Reine', id: '583F', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'クイーンズ・ソルジャー', id: '583F', capture: false }),
+      delaySeconds: 4,
+      response: Responses.stopEverything('alert'),
```
quisquous, line 268:
```suggestion
      response: Responses.stopEverything('alarm'),
```

What about making this one alarm and keeping icy as an alert, so that the sounds are differentiated?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -249,6 +258,103 @@ export default {
...
+      // Buff expires about 16 seconds on first cast, ~8 seconds later casts)
+      netRegex: NetRegexes.startsUsing({ source: 'Bozjan Phantom', id: '57BC', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Bozja-Phantom', id: '57BC', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Fantôme Bozjien', id: '57BC', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'ボズヤ・ファントム', id: '57BC', capture: false }),
+      delaySeconds: 3,
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Stop Attacking, Wait for Dispel',
```
quisquous, line 328: Maybe `Stop Attacking, Dispel Ice Spikes`?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -249,6 +258,103 @@ export default {
...
+      condition: tankBusterOnParty,
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'DelubrumSav Avowed Wrath Of Bozja',
+      netRegex: NetRegexes.startsUsing({ source: 'Trinity Avowed', id: '594E' }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Trinität Der Eingeschworenen', id: '594E' }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Trinité Féale', id: '594E' }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'トリニティ・アヴァウド', id: '594E' }),
+      response: Responses.tankCleave('alert'),
```
quisquous, line 347: This one needs more special handling, I think.  Maybe if you are a tank say `triple shared tank cleave` or just `shared tank cleave`, healers can get `tank cleave` and everybody else `avoid tank cleave`?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -317,6 +423,36 @@ export default {
...
+    {
+      id: 'DelubrumSav Queen Cleansing Slash',
+      netRegex: NetRegexes.startsUsing({ source: 'The Queen', id: '59F5' }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Kriegsgöttin', id: '59F5' }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Garde-La-Reine', id: '59F5' }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'セイブ・ザ・クイーン', id: '59F5' }),
+      condition: tankBusterOnParty,
+      // Probably this is where you swap, but maybe that's not something you can
+      // count on in an alliance raid, where there might not even be another tank.
+      response: Responses.tankBuster(),
```
quisquous, line 445: Maybe this needs "Double tankbuster on X" for tanks/healers?

quisquous (commented):
Thank you for more triggers!

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -249,6 +258,103 @@ export default {
         },
       },
     },
+    {
+      id: 'DelubrumSav Guard Fiery Portent',
+      netRegex: NetRegexes.startsUsing({ source: 'Queen\'s Soldier', id: '583F', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Soldat Der Königin', id: '583F', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Soldat De La Reine', id: '583F', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'クイーンズ・ソルジャー', id: '583F', capture: false }),
+      delaySeconds: 4,
```
quisquous, line 267: Why the delaySeconds on this? Maybe better to have the early callout + durationSeconds if it needs to stay up for longer? Same question on the icy portent below.

Legends0: The cast is 6 seconds long, and so I figure a cast on startsUsing is too early. To me, 3 second delay still felt too early and 4 seconds feels too short.

This does need a durationSeconds though. The Pyretic (3C0) debuff lasts 3 seconds.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Pyretic ticks hit for 100k+, even with some mitigations. One tick is survivable if you are at full hp. However, I will make this adjustment as in my experience people tend to die from the Pyretic tick(s) if they get hit by it.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Trinity Avowed (TA) and Queen fight may need these. The other fights the AoE does considerably less damage, more so when everyone is using Pure Essences. In TA and Queen there is an AoE that also applies a bleed and the damage needs considerably more mitigation.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: Sounds reasonable! I was just wondering if it was better to just call this out as early as possible.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I opted for shared tank cleave to keep it short. Unless we want to also clarify the shared tank cleave in Trinity Seeker to be double, I think shared tank cleave is fine here. One could also interpret shared tank cleave to default as '2 tanks' or 'all tanks', but I don't particularly have a preference as I do not play tank role...

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I have changed the AoE + Bleeds spell triggers to match e11s.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: It's not possible to survive without immunity as there is a vulnerability up that occurs. However, In the logs I have where a tank gets hit and survives it because they "Dodge" the first hit (during immunity) and then get hit by the second.
It's odd, but all tanks with immunity will Dodge the first hit, and then if they are WAR or DRK, they get hit by the second which applies the Doom.

Legends0: I think most groups will be having tank invuln the ability and then swap after it ends. I have chosen to alter the comment about the trigger to clarify the spell behavior. Maybe a tank swap callout would be good so tanks swap after the second hit as WAR/DRKs would have a Physical Vulnerability Up and for the next Cleansing Slash a new tank is needed as the original tank's invulnerability is not up for it.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: The updated timings (delaySeconds of 3.5) lined up with the my group's manual callouts, so felt good now.

JLGarber: What would you think of modifying `delaySeconds` to take a function of the cast duration instead? As `delaySeconds: (data, matches) => parseFloat(matches.duration) - 2,`. It's easier to see exactly how much advance notification we're giving the user when it's written this way.

Legends0:
For Maelstrom's Bolt, could have a trigger on the 25: NetworkActionSync where the Ball Lightning reaches 0 hp as that signifies it dieing and spawning a shield into the Environment or just assume a shield exists and say get into shield?

Also possible to check for 20B, Lightning Resistance Up to tell if someone is in the shield or not...

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Good idea @JLGarber I have added this change.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: That's fair, but I think even if they say bigAoe, the TA and Queen ones should be timeline triggers as I mentioned above to give people more time (although arguably all aoes in the instance could be timeline triggers for consistency, imo.)

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -309,6 +472,61 @@ export default {
...
+        output.responseOutputStrings = {
+          getOut: Outputs.out,
+          getIn: Outputs.in,
+        };
+
+        if (matches.effectId === '97F')
+          return { alertText: output.getOut() };
+        else if (matches.effectId === '97E')
+          return { alertText: output.getIn() };
+        return;
```
quisquous, line 503:
```suggestion
```

Unnecessary return?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -309,6 +472,61 @@ export default {
...
+      // Remaining set of debuffs go out 24.3 seconds before Fateful Word is cast
+      netRegex: NetRegexes.gainsEffect({ effectId: '97[EF]' }),
+      condition: (data, matches) => {
+        return (matches.target === data.me);
+      },
+      delaySeconds: (data, matches) => {
+        if (data.secondFates)
+          return 21;
+        data.secondFates = 1;
+        return 4;
```
quisquous, line 490: I think this might be cleaner as a timeline trigger before Fateful Words rather than having multiple delays?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: The shortest AoE cast is Malediction Of Agony at 4.7s. The others are at least 5 seconds, with the majority being 5.5 seconds. Gods Save The Queen being the longest at 6 seconds. Use of a timelineTrigger seems unnecessary given that some examples of timelineTriggers in cactbot I found in ultimate are using 5 seconds before.

The biggest issue I think will be The Queen's ending AoEs. Could look to how other raid end tier bosses are handled, but then again these are in quick succession and we have more mitigation from lost actions available to us here. Calling out all the AoEs would lead to a callout every 3 seconds, which I think is a bit excessive compared to previous encounters. The timeline I think could use some cleaning up here as well. The AoEs themselves can be grouped into three sets:
Set 1: Double AoE + Double AoE
-- 5 second "calm" --, there is no real break the next cast starts immediately at end of set
Set 2: Double Aoe + Double AoE + Big AoE (AoE + Bleed)
-- 6.3 second "calm" --, technically 1.3 seconds of no casts
Set 3: Four single AoEs during bleed

So at the minimum I think 3 callouts would be alright for the ending AoEs, which would need to be done by a timelineTrigger.

The Guard's AoEs appear to do the same damage as in Queen's Guard encounter.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: Yeah, I'm not sure what the text here should be.  Is there a target for this attack? If so, and the player is a paladin / gnb maybe say "invuln tankbuster".  If not, maybe say "tankbuster + swap".  For other tanks or healers, I'm not sure what to say.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -317,6 +535,65 @@ export default {
...
+    },
+    {
+      id: 'DelubrumSav Queen Ball Lightning',
+      // Players with Reflect should destroy one for party to stand in the shield left behind
+      netRegex: NetRegexes.addedCombatantFull({ npcNameId: '7974', capture: false }),
+      suppressSeconds: 1,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Orbs',
```
quisquous, line 593:
```suggestion
          en: 'Reflect Orbs',
```

(Sorry, I think verbs are helpful here to tell people what to do when there's only one thing to do.)

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: It seemed unlikely for the branching to execute that return, but in case it did somehow I didn't know if an implied return would work. Could a player possibly have a debuff, then die and respawn before the 21 second delaySeconds go off and cause an error here without return?

Legends0: Nevermind, I guess in that case the effectId is stored from what they had before they died.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Would we need to store the result of the debuff in this trigger using the netregex and then use the timelineTrigger to attempt to collect that debuff?

Legends0: I see timelineTrigger would also be better because this event happens in a loop and given the 21 second delay could lead to the callout happening long after the phase is completed.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: > The shortest AoE cast is Malediction Of Agony at 4.7s. The others are at least 5 seconds, with the majority being 5.5 seconds. Gods Save The Queen being the longest at 6 seconds. Use of a timelineTrigger seems unnecessary given that some examples of timelineTriggers in cactbot I found in ultimate are using 5 seconds before.

I think the difference here is more that raises are really hard to come by here and the instance is really long, so penalties for messing up are just really high.  I've had a bunch of people ask me to give more warning before raidwides (since if folks are really low and you need two emergency gcd heals, 5s might not be enough).  So, I thought maybe it'd be worth doing in this instance especially, more than any other place.

Yeah, it's different, but I kinda think it's a little bit more warranted just because keeping people alive is a lot more important here.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -317,6 +536,86 @@ export default {
...
+        if (data.me === matches.target) {
+          if (data.role === 'tank') {
+            if (data.job === 'PLD' || data.job === 'GNB')
+              return { alertText: output.tankInvuln() };
+            return { alertText: output.tankBusterAndSwap() };
+          }
+          return { alarmText: output.tankBusterOnYou() };
+        }
+        if (data.role === 'healer')
+          return { alertText: output.tankBusters() };
```
quisquous, line 584:
```suggestion
        if (data.role === 'healer' || data.role === 'tank')
          return { alertText: output.tankBusterOnPlayer({ player: matches.target });
```

I think it's worth calling out the target, and informing tanks about it, as well.

Legends0: The condition would need to be modified as well to inform the healers/tanks outside of the party.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: FYI, not moving during Icy Portent causes Deep Freeze '4E6'. Also the behavior and debuffs are the same in The Queen.

Legends0:
It may be possible to have a chess solver for Queen, based on a strategy such as the "C" pattern strategy, but that may be a bit much.

For Trinity Avowed Hot & Cold cleaves, the adds have a spawn location that can be retreived, but because they turn I think we aren't able to write a solver with current netregexes. However, if someone ever wants to try something, here are what I verified each cleave's spell ID represents:
` ``  
      // Trinity Avowed or Avowed Avatar cast these pairs
      // +1
      // 5942 = right cleave, heat+1(1) paired with 5944
      // 5943 = right cleave, cold+1(1) paired with 5945
      // 5944 = right cleave, heat+1(1) paired with 5942
      // 5945 = right cleave, cold+1(1) paired with 5943
      
      // 5946 = left cleave, cold+1(1) paired with 5948
      // 5947 = left cleave, heat+1(1) paired with 5949
      // 5948 = left cleave, cold+1(1) paired with 5946
      // 5949 = left cleave, heat+1(1) paired with 5947
      
      // +2
      // 5956 = right cleave, heat+1(2) paired with 5958
      // 5957 = right cleave, cold+1(2) paired with 5959
      // 5958 = right cleave, heat+1(2) paired with 5956
      // 5959 = right cleave, cold+1(2) paired with 5957
      
      // 595A = left cleave heat+1(2) paired with 595C
      // 595B = left cleave cold+1(2) paired with 595D
      // 595C = left cleave heat+1(2) paired with 595A
      // 595D = left cleave cold+1(2) paired with 595B
```

**Comment by quisquous** [comment]
> It may be possible to have a chess solver for Queen, based on a strategy such as the "C" pattern strategy, but that may be a bit much.

The real issue is that I haven't seen the limit cut headmarkers in log lines, so I'm not sure how to figure out which of the states you're in.

**Comment by quisquous** [comment]
I also think it'd be nice to get this landed at some point.  Would it make sense to stop adding things here so we can wrap these up?

File: ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js (hunk 10 above)
Comment (Legends0) [reply ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js]: Sorry, I removed the party condition after the comment in 04c9ea76b46e3bd73f8a85a9e5bf90b64c4bd9fa

**Comment by Legends0** [comment]
> 
> 
> I also think it'd be nice to get this landed at some point. Would it make sense to stop adding things here so we can wrap these up?

If we find that we need a timelineTrigger for Queen's Guard's AoEs, I would recommend that have a condition that it only calls it out during The Queen's Guard encounter so that it doesn't cause a lot of noise during The Queen.

Similar to Thunderous Discharge, An erroneous Verdant Tempest callout could happen during Trinity Seeker if the phase pushes sooner. However I think in the case of Thunderous Discharge, most groups should be killing it before the second callout.

Queen's Guard may want to have a Dispel Boost callout or I think that boosted AoE causes a wipe. The Automation Turrets have a similar mechanic.

I can work on additional triggers in another PR.

File: ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js [hunk 11]
```diff
@@ -249,6 +335,130 @@ export default {
...
+        output.responseOutputStrings = {
+          goSouth: {
+            en: 'Go South, Unavoidable Knocbkack',
+          },
+          goNorth: {
+            en: 'Go North, Unavoidable Knocbkack',
+          },
+        };
+
+        if (matches.y === -403.5)
```
quisquous, line 403: This will not work, as `matches.y` is a string.  (Someday, typescript will warn about this!)

Also, I think I don't like to trust that values are exact.  What's the other value? Can you list both of them in a comment and then do something like `parseFloat(matches.y) < 410` instead of `matches.y === -403.5`?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -309,6 +519,32 @@ export default {
         return { infoText: output.avoidCleave() };
       },
     },
+    {
+      id: 'DelubrumSav Lord Labyrinthine Fate Collect',
+      // Result used by timelineTrigger id 'DelubrumSav Lord Fateful Words'
+      netRegex: NetRegexes.gainsEffect({ effectId: '97[EF]' }),
+      condition: (data, matches) => {
+        return (matches.target === data.me);
+      },
```
quisquous, line 528:
```suggestion
      condition: Conditions.targetIsYou(),
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -317,6 +553,121 @@ export default {
...
+      netRegexDe: NetRegexes.startsUsing({ source: ['Kriegerin Der Königin', 'Ritter Der Königin', 'Schütze Der Königin', 'Soldat Der Königin'], id: ['5A16', '5A08', '5A35', '5A23'], capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: ['Guerrière De La Reine', 'Chevalier De La Reine', 'Fusilier De La Reine', 'Soldat De La Reine'], id: ['5A16', '5A08', '5A35', '5A23'], capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: ['クイーンズ・ウォリアー', 'クイーンズ・ナイト', 'クイーンズ・ガンナー', 'クイーンズ・ソルジャー'], id: ['5A16', '5A08', '5A35', '5A23'], capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      // Only call out the beginning of a set of two casts
+      suppressSeconds: 5,
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Multiple Guards\' AOEs',
```
quisquous, line 667:
```suggestion
          en: 'Multiple Big AOEs',
```

This is just a tiny nitpick, but I think the word "Guard" isn't doing much here.  Maybe "Big" as I wrote or just "Multiple AOEs"?

quisquous (commented):
This all looks great.  Thanks for your patience with a bunch of review and thanks for adding so many triggers.  I have a few tiny comments, but I am ready to merge this.

quisquous:
> For Trinity Avowed Hot & Cold cleaves, the adds have a spawn location that can be retreived, but because they turn I think we aren't able to write a solver with current netregexes.

I think this would be possible with getCombatants and asking each of them for position and heading.  You can see something like this in e6s and also the Iron Splitter trigger in DRS.  Once they spawn, delay x seconds until they have turned, and then call getCombatants to get their current heading from OverlayPlugin.  It's a bit cumbersome, but do-able.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: The comment is under the id.
It always spawns on the y at -403.5 (North) or -344.5 (South).

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: To avoid confusion between the other Big AoE callouts, I have shortened it to Multiple AoEs. I was also thinking it could be shortened to Guards' AoEs, but Multiple requires less thought about it.

Legends0:
> 
> 
> > For Trinity Avowed Hot & Cold cleaves, the adds have a spawn location that can be retreived, but because they turn I think we aren't able to write a solver with current netregexes.
> 
> I think this would be possible with getCombatants and asking each of them for position and heading. You can see something like this in e6s and also the Iron Splitter trigger in DRS. Once they spawn, delay x seconds until they have turned, and then call getCombatants to get their current heading from OverlayPlugin. It's a bit cumbersome, but do-able.

Oh that's what that call function is for, interesting. Another note on that then is that may be also how to do Cataclysm direction callouts for e12s.

With chess after you pointed out the headmarker problem, I looked at the logs and I too couldn't see the markers for the Queens' Guard, so although we can get their facing and positions, I guess we don't know how many steps they are going to take. For the players, players get debuffs so we can at least know how many steps they need to take. The Queens' Guard get a single debuff (it's not named so comes up as Unkown_808 or something), but I am guessing there is some other way that the number of steps they will take is conveyed. I haven't looked at enough comparisons, but the ones I saw only had them more 1 or 3 spaces and it was always in pairs, the Sprint buffed ones (ones that visually get tethered) were different numbers. No idea if it's possible to output a safe path, if the numbers are always the same, then I think it's possible we don't need to know what each one has and in certain situations can have an answer (based on where they spawn knowing that it'd be only 1 or 3 movements). As soon as it gets a 2 (if it does), then that wouldn't work.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I put in the midway point between to spawns and chose < for North. It looks to me like the y values are inverted.

quisquous: Yeah +x is east and +y is south on all maps.  Thanks!

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I will add a note to this that due to the nature of the current Trinity Avowed timeline with how it jumps to different phases as different times, Glory of Bozja can be double called out or even missed.

quisquous: I'll fix this up.  Double callouts are an unfixed bug with timeline triggers that I forgot.  The issue is that if the timeline syncs you backwards 0.1s in time or something, it doesn't realize it's already called the timeline callout for that time already.  It's not because the Trinity Avowed timeline jumps around, it's just a timeline trigger issue.

Glory of Bozja should never be missed.  I could imagine if it was at the beginning of the block and you jumped so that it was happening sooner than 7 seconds, then it could get missed.  However, it always comes at the end of a block, and so 7 seconds before it will never be missed.  If this is happening, I think I need more information on why.

Would you mind filing new issues for these when you find bugs.  I can respond here, but it's easy for me to lose track of replies on old merged issues.

PR #2553, opened 2021-03-08
raidboss: more delubrum savage triggers

quisquous:


quisquous:
I tried to write some baleful comet triggers to tell you where to go and rotate to, but something is awry, so I removed them.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -125,6 +142,70 @@ export default {
...
+        // white middle = 14
+        // blue outer = 18?
+        // white outer = 22?
+
+        const isWhite = Math.floor(data.splitterDist / 4) % 2;
+        return isWhite ? output.goBlue() : output.goWhite();
+      },
+      outputStrings: {
+        goBlue: {
+          en: 'Go Blue',
```
quisquous, line 202: @Legends0 Do you have any opinions about these triggers? Do other folks who aren't me say blue/white or does it need to say like "Go Blue Stone" or "Go White Sand" or some other variation?

Legends0: My group has stated "Stand on Sand" or "Stand on Brick".

Legends0: I think 'Be on' will be a better phrase to use as it gets you to look down at your feet rather than looking around for someone to go to. The noun used may vary, sand seems pretty universal to use, but brick caught me off the first time it was used but looking at the ground more closely it made sense after hearing it.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -125,6 +142,70 @@ export default {
...
       },
     },
+    {
+      id: 'DelubrumSav Seeker Dead Iron',
+      // Headmarkers are randomized, so use the tether instead.
+      netRegex: NetRegexes.tether({ target: 'Trinity Seeker', id: '01DB' }),
+      condition: (data, matches) => matches.source === data.me,
+      alarmText: (data, _, output) => output.earthshaker(),
+      outputStrings: {
+        earthshaker: Outputs.earthshakerOnYou,
```
Legends0, line 152: I recommend a strategy to have these players get away from the boss and spread around the avatar as the earthshaker tethers come from the avatar that spawns and alliance is at the boss. However, I think it's nicer to keep with the familiar cactbot earthshaker callout and let the player decide how to handle it.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -58,6 +61,20 @@ export default {
...
+      id: 'DelubrumSav Seeker Iron Impact',
+      // This is an early warning on the Verdant Path cast.
+      netRegex: NetRegexes.startsUsing({ source: 'Trinity Seeker', id: '5A99', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ source: 'Trinität Der Sucher', id: '5A99', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ source: 'Trinité Soudée', id: '5A99', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ source: 'トリニティ・シーカー', id: '5A99', capture: false }),
+      alertText: (data, _, outputs) => outputs.text(),
+      outputStrings: {
+        text: {
+          en: 'Get Behind For Line Stack',
```
Legends0, line 74: May want to just say "Line Stack" or "Line AoE" as it is more in line with how cactbot has previously handled similar mechanics. The drawback here is there isn't a marker for it like some of the usage has. I believe Iron Impact targets a random player out of all 48 players.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: That's fair.  I suspect delubrum savage will have more fixed strategies, probably?  I'd be fine being a little bit more perscriptive here, and if people complain, we can adjust.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: I think "line stack" on its own usually means "look for the line stack, and go to it".  I suspect that most groups stack behind.  I think "stack for line stack" just sounded too awkward, so I went with "get behind".

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: My worry here is that the thing you have to do is several words in, and the timing is pretty tight.

I'd almost say "Blue stone" or "white sand" on its own, unless there's some way to say like "blue stone (stand on it!)" that sounds less awkward.  I think people in savage should know what this mechanic is, and I think there's some advantage to getting out the important word as quickly as possible.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Looks good.

PR #2556, opened 2021-03-08
update TEA translate for CN

Trim21:


ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
```diff
@@ -113,31 +113,31 @@ const radiantOutputStrings = {
...
     ko: '성례: 남',
   },
   west: {
     en: 'Sacrament West',
     de: 'Sacrement Westen',
     fr: 'Sacrement Ouest',
     ja: '拝火は西',
-    cn: '拜火 西',
+    cn: '西侧月环',
     ko: '성례: 서',
```
MaikoTan, line 141: To just call them 拜火 is ok, or you can name direction as 上(北)

Trim21: I haven't see anyone call it 拜火 in cn bbs or videos, I think it's better to call it 月环, as most people are doing.

PR #2557, opened 2021-03-08
add a trigger for TEA alpha phase

Trim21:
@MaikoTan help me to translate it.

I haven't tested it in-game, I haven't passed phase 3 yet. I'm not sure if it works fine.

Hoping someone could test it.

### DIFF
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
@@ -747,7 +747,7 @@ export default {
     {
       // Applies to both limit cuts.
       id: 'TEA Limit Cut Numbers',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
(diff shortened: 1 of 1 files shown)


quisquous:
Does `TEA Alpha Safe Spot` do you what you're trying to do already?

MaikoTan:
I got no idea about what you're trying to do, can you explain it first?
And why does it return blank data by XIVAPI with the abilily id you've written?
https://xivapi.com/Action/18591?columns=Name,Name_de,Name_fr,Name_ja

miaonima:
What mechanic is this? On which phase? Thanks.

Trim21:
@miaonima alpha phase, safe location. Avoiding 3 Sacrament

miaonima:
I think cactbot already has callouts for p4 alpha phase. Besides your enumeration, it says LEFT/RIGHT for safe spot, (and opposite direction if you have Defamation huge circle AOE). I am wondering if your proposal is the same as 'TEA Alpha Safe Spot' trigger. Please let me know your thoughts.

PR #2558, opened 2021-03-09
Refactor: generic regex/netregex helper function improve

MaikoTan:
This change makes it easier for us to refactor the Trigger data file later on, we can write
```typescript
{
  id: 'General Provoke',
  netRegex: NetRegexes.ability({ id: '1D6D' }),
  condition: function(data: Data, matches: Matches<typeof NetRegexes.ability>): boolean {
    if (matches.source !== data.me && !data.party.inAlliance(matches.source))
      return false;
    return caresAboutTankStuff(data);
  },
},
```
and then TypeScript will automatically derive it as:
```typescript
(parameter) matches: {
    sourceId: string;
    source: string;
    id: string;
    ability: string;
    targetId: string;
    target: string;
    timestamp: string;
}
```

As in VSCode, it would be like this:
image

### DIFF
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -32,10 +32,10 @@ type TargetableMatches = Matches<Regex<
 
 export default {
   targetIsYou(): (data: Data, matches: TargetableMatches) => boolean {
-    return (data: Data, matches: TargetableMatches) => data.me === matches?.target;
+    return (data: Data, matches: TargetableMatches) => data.me === matches.target;
   },
   targetIsNotYou(): (data: Data, matches: TargetableMatches) => boolean {
-    return (data: Data, matches: TargetableMatches) => data.me !== matches?.target;
+    return (data: Data, matches: TargetableMatches) => data.me !== matches.target;
   },
   caresAboutAOE(): (data: Data) => boolean {
     return (data: Data) => data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -5,12 +5,37 @@
 import { Data } from '../types/data';
 import { Matches } from '../types/trigger';
 
+import {
+  StartsUsingParams,
+  AbilityParams,
+  AbilityFullParams,
+  HeadMarkerParams,
+  GainsEffectParams,
+  StatusEffectExplicitParams,
+  LosesEffectParams,
+  TetherParams,
+  WasDefeatedParams,
+  Return,
+} from '../resources/regexes';
+
+type TargetableMatches = Matches<Return<
+  StartsUsingParams |
+  AbilityParams |
+  AbilityFullParams |
+  HeadMarkerParams |
+  GainsEffectParams |
+  StatusEffectExplicitParams |
+  LosesEffectParams |
+  TetherParams |
+  WasDefeatedParams
+>>;
+
 export default {
-  targetIsYou(): (data: Data, matches: Matches<unknown>) => boolean {
-    return (data: Data, matches: Matches<unknown>) => data.me === matches.target;
+  targetIsYou(): (data: Data, matches: TargetableMatches) => boolean {
+    return (data: Data, matches: TargetableMatches) => data.me === matches?.target;
   },
-  targetIsNotYou(): (data: Data, matches: Matches<unknown>) => boolean {
-    return (data: Data, matches: Matches<unknown>) => data.me !== matches.target;
+  targetIsNotYou(): (data: Data, matches: TargetableMatches) => boolean {
+    return (data: Data, matches: TargetableMatches) => data.me !== matches?.target;
   },
   caresAboutAOE(): (data: Data) => boolean {
     return (data: Data) => data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -6,11 +6,11 @@ import { Data } from '../types/data';
 import { Matches } from '../types/trigger';
 
 export default {
-  targetIsYou(): (data: Data, matches: Matches) => boolean {
-    return (data: Data, matches: Matches) => data.me === matches.target;
+  targetIsYou(): (data: Data, matches: Matches<unknown>) => boolean {
+    return (data: Data, matches: Matches<unknown>) => data.me === matches.target;
   },
-  targetIsNotYou(): (data: Data, matches: Matches) => boolean {
-    return (data: Data, matches: Matches) => data.me !== matches.target;
+  targetIsNotYou(): (data: Data, matches: Matches<unknown>) => boolean {
+    return (data: Data, matches: Matches<unknown>) => data.me !== matches.target;
   },
   caresAboutAOE(): (data: Data) => boolean {
     return (data: Data) => data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -15,10 +15,10 @@ import {
   LosesEffectParams,
   TetherParams,
   WasDefeatedParams,
-  Return,
+  Regex,
 } from '../resources/regexes';
 
-type TargetableMatches = Matches<Return<
+type TargetableMatches = Matches<Regex<
   StartsUsingParams |
   AbilityParams |
   AbilityFullParams |
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -32,10 +32,10 @@ type TargetableMatches = Matches<Regex<
 
 export default {
   targetIsYou(): (data: Data, matches: TargetableMatches) => boolean {
-    return (data: Data, matches: TargetableMatches) => data.me === matches.target;
+    return (data: Data, matches: TargetableMatches) => data.me === matches?.target;
   },
   targetIsNotYou(): (data: Data, matches: TargetableMatches) => boolean {
-    return (data: Data, matches: TargetableMatches) => data.me !== matches.target;
+    return (data: Data, matches: TargetableMatches) => data.me !== matches?.target;
   },
   caresAboutAOE(): (data: Data) => boolean {
     return (data: Data) => data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -1,9 +1,12 @@
-import Regexes, { Params, Return } from './regexes';
+import { BaseRegExp } from 'types/trigger';
+import Regexes, { Params } from './regexes';
 
 interface Fields {
   field: string;
   value?: string;
 }
+
+export type NetRegex<T extends string> = BaseRegExp<Exclude<T, 'capture'>>;
 
 // Differences from Regexes:
 // * may have more fields
@@ -129,7 +132,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#14-networkstartscasting
    */
-  static startsUsing(params?: Params<StartsUsingParams>): Return<StartsUsingParams> {
+  static startsUsing(params?: Params<StartsUsingParams>): NetRegex<StartsUsingParams> {
     return parseHelper(params, 'startsUsing', {
       0: { field: 'type', value: '20' },
       1: { field: 'timestamp' },
@@ -148,7 +151,7 @@ export default class NetRegexes {
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#15-networkability
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#16-networkaoeability
    */
-  static ability(params?: Params<AbilityParams>): Return<AbilityParams> {
+  static ability(params?: Params<AbilityParams>): NetRegex<AbilityParams> {
     return parseHelper(params, 'ability', {
       0: { field: 'type', value: '2[12]' },
       1: { field: 'timestamp' },
@@ -166,7 +169,7 @@ export default class NetRegexes {
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#15-networkability
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#16-networkaoeability
    */
-  static abilityFull(params?: Params<AbilityFullParams>): Return<AbilityFullParams> {
+  static abilityFull(params?: Params<AbilityFullParams>): NetRegex<AbilityFullParams> {
     return parseHelper(params, 'abilityFull', {
       0: { field: 'type', value: '2[12]' },
       1: { field: 'timestamp' },
@@ -191,7 +194,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#1b-networktargeticon-head-markers
    */
-  static headMarker(params?: Params<HeadMarkerParams>): Return<HeadMarkerParams> {
+  static headMarker(params?: Params<HeadMarkerParams>): NetRegex<HeadMarkerParams> {
     return parseHelper(params, 'headMarker', {
       0: { field: 'type', value: '27' },
       1: { field: 'timestamp' },
@@ -205,7 +208,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#03-addcombatant
    */
-  static addedCombatant(params?: Params<AddedCombatantParams>): Return<AddedCombatantParams> {
+  static addedCombatant(params?: Params<AddedCombatantParams>): NetRegex<AddedCombatantParams> {
     return parseHelper(params, 'addedCombatant', {
       0: { field: 'type', value: '03' },
       1: { field: 'timestamp' },
@@ -220,7 +223,7 @@ export default class NetRegexes {
    */
   static addedCombatantFull(
       params?: Params<AddedCombatantFullParams>,
-  ): Return<AddedCombatantFullParams> {
+  ): NetRegex<AddedCombatantFullParams> {
     return parseHelper(params, 'addedCombatantFull', {
       0: { field: 'type', value: '03' },
       1: { field: 'timestamp' },
@@ -247,7 +250,7 @@ export default class NetRegexes {
    */
   static removingCombatant(
       params?: Params<RemovingCombatantParams>,
-  ): Return<RemovingCombatantParams> {
+  ): NetRegex<RemovingCombatantParams> {
     return parseHelper(params, 'removingCombatant', {
       0: { field: 'type', value: '04' },
       1: { field: 'timestamp' },
@@ -261,7 +264,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#1a-networkbuff
    */
-  static gainsEffect(params?: Params<GainsEffectParams>): Return<GainsEffectParams> {
+  static gainsEffect(params?: Params<GainsEffectParams>): NetRegex<GainsEffectParams> {
     return parseHelper(params, 'gainsEffect', {
       0: { field: 'type', value: '26' },
       1: { field: 'timestamp' },
@@ -283,7 +286,7 @@ export default class NetRegexes {
    */
   static statusEffectExplicit(
       params?: Params<StatusEffectExplicitParams>,
-  ): Return<StatusEffectExplicitParams> {
+  ): NetRegex<StatusEffectExplicitParams> {
     return parseHelper(params, 'statusEffectExplicit', {
       0: { field: 'type', value: '38' },
       1: { field: 'timestamp' },
@@ -307,7 +310,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#1e-networkbuffremove
    */
-  static losesEffect(params?: Params<LosesEffectParams>): Return<LosesEffectParams> {
+  static losesEffect(params?: Params<LosesEffectParams>): NetRegex<LosesEffectParams> {
     return parseHelper(params, 'losesEffect', {
       0: { field: 'type', value: '30' },
       1: { field: 'timestamp' },
@@ -325,7 +328,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#23-networktether
    */
-  static tether(params?: Params<TetherParams>): Return<TetherParams> {
+  static tether(params?: Params<TetherParams>): NetRegex<TetherParams> {
     return parseHelper(params, 'tether', {
       0: { field: 'type', value: '35' },
       1: { field: 'timestamp' },
@@ -342,7 +345,7 @@ export default class NetRegexes {
    * 'target' was defeated by 'source'
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#19-networkdeath
    */
-  static wasDefeated(params?: Params<WasDefeatedParams>): Return<WasDefeatedParams> {
+  static wasDefeated(params?: Params<WasDefeatedParams>): NetRegex<WasDefeatedParams> {
     return parseHelper(params, 'wasDefeated', {
       0: { field: 'type', value: '25' },
       1: { field: 'timestamp' },
@@ -357,7 +360,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#00-logline
    */
-  static echo(params?: Params<EchoParams>): Return<EchoParams> {
+  static echo(params?: Params<EchoParams>): NetRegex<EchoParams> {
     if (typeof params === 'undefined')
       params = {};
     Regexes.validateParams(params, 'echo', ['type', 'timestamp', 'code', 'name', 'line', 'capture']);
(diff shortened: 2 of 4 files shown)


panicstevenson:
I don't know if this makes sense, but would it be possible to infer the type based off of the `netRegex` attribute type? From there, you would know that anything that has a `netRegex` of `ability` can only have `condition` with no parameters, just `data`, or `data` and `Matches<ability>` (or similar).
That way the auto-complete would then only give you the options of the type you already specified in `netRegex`. I don't know if that makes sense or is better? It might be more work, but (I imagine) it would be less typing in the triggers themselves.

MaikoTan:
I tried before but no.
TypeScript cannot infer types from siblings generic types, this is the simpliest approach I can find. Or if you have better idea?

quisquous:
> I tried before but no.
> TypeScript cannot infer types from siblings generic types (unless we specify the type onto the trigger object), this is the simpliest approach I can find. Or if you have better idea?

Along these lines, I was wondering if we could specify the Trigger types as a union (of all regex + netregex + timeline types?).  And then we could do something like:

```
type NetRegexA = 'a';
type NetRegexB = 'b';

type MatchesA = 'a';
type MatchesB = 'b';

type TriggerA = {
  id: number;
  netRegex: NetRegexA;
  condition?: MatchesA;
};
type TriggerB = {
  id: number;
  netRegex: NetRegexB;
  condition: MatchesB;
};

type Trigger = TriggerA | TriggerB;

// Ok!
const x: Trigger = { id: 1, netRegex: 'a', condition: 'a' };
// Ok!
const y: Trigger = { id: 2, netRegex: 'a' };
// Error!
const z: Trigger = { id: 3, netRegex: 'a', condition: 'b' };

```

MaikoTan:
Ah, you're right!
How can I forget that way! I'll change it later.

quisquous:
> Ah, you're right!
> How can I forget that way! I'll change it later.

I think this patch will still be really nice to have a concise way to specify the matches for each regex type, but I think my hope is that individual triggers can be implicitly typed (for brevity) and not have to type that for every function.

MaikoTan:
It looks like the triggers in an array not only the type hint didn't work, even doesn't guarantee the types of parameters of `infoText`.

[Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLHQjRMLAZBzAQooKHreCnxFKU-RFWchSICSJPCChI3CPNZ0kodyQoFlyQNP4HQUqT50XBkVzXbddwNf5xMkszBwgMBhBALtzLsiSxBkmNNO7ezFS0ogYRcySvIIKUGCAA)
I have created a playground with basic types, see if you have better thought.

quisquous:
> It looks like the triggers in an array not only the type hint didn't work, even doesn't guarantee the types of parameters of `infoText`.
> 
> [Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLHQjRMLAZBzAQooKHreCnxFKU-RFWchSICSJPCChI3CPNZ0kodyQoFlyQNP4HQUqT50XBkVzXbddwNf5xMkszBwgMBhBALtzLsiSxBkmNNO7ezFS0ogYRcySvIIKUGCAA)
> I have created a playground with basic types, see if you have better thought.

Hmmm.  This is the best I could come up with: [Playground Link 2](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLBLKIoklNRgBpMINANEAyOQGitAAa2QMhPAIDCsPMBCigoet4KfEUpT9EVZyFIgtK08IKEjcI81nbSh3JCgWXJA0-gdIydPnRcGRXNdt13A1-k07SPMHCAwGEEAu08gKtLEPSY2s7tAsVGyiBhMLtJiggpXQjRMLAZBZKfSgAHUwLUYF9wfJToBUtSNKins9IM2KApMkdzOMSyLCqzzQAXJcfkMVctxIFzKDcsrjO83z-MCgLgsTUL+q0yLwqmprkBhZAMUrGDa0UuTSPI7yILCxLb2k1L0wgLoygweg1qfCg3w-c7EMKCjO3c3TE0q5warMnELMoKznBa+zl1+Tq4Gco89z67tRB8qA-Pc7SxujAS5um6KwsVIA)

It lets you specify the trigger type once, and then all functions get the right types, vs on every parameter of every function.

MaikoTan:
> > It looks like the triggers in an array not only the type hint didn't work, even doesn't guarantee the types of parameters of `infoText`.
> > [Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLHQjRMLAZBzAQooKHreCnxFKU-RFWchSICSJPCChI3CPNZ0kodyQoFlyQNP4HQUqT50XBkVzXbddwNf5xMkszBwgMBhBALtzLsiSxBkmNNO7ezFS0ogYRcySvIIKUGCAA)
> > I have created a playground with basic types, see if you have better thought.
> 
> Hmmm. This is the best I could come up with: [Playground Link 2](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLBLKIoklNRgBpMINANEAyOQGitAAa2QMhPAIDCsPMBCigoet4KfEUpT9EVZyFIgtK08IKEjcI81nbSh3JCgWXJA0-gdIydPnRcGRXNdt13A1-k07SPMHCAwGEEAu08gKtLEPSY2s7tAsVGyiBhMLtJiggpXQjRMLAZBZKfSgAHUwLUYF9wfJToBUtSNKins9IM2KApMkdzOMSyLCqzzQAXJcfkMVctxIFzKDcsrjO83z-MCgLgsTUL+q0yLwqmprkBhZAMUrGDa0UuTSPI7yILCxLb2k1L0wgLoygweg1qfCg3w-c7EMKCjO3c3TE0q5warMnELMoKznBa+zl1+Tq4Gco89z67tRB8qA-Pc7SxujAS5um6KwsVIA)
> 
> It lets you specify the trigger type once, and then all functions get the right types, vs on every parameter of every function.

I think to use an `as` operator in array literal is odd, and then would be tired to specify types over and over again.
But if this is the only way, I would go with that.

quisquous:
> > > It looks like the triggers in an array not only the type hint didn't work, even doesn't guarantee the types of parameters of `infoText`.
> > > [Playground Link](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLHQjRMLAZBzAQooKHreCnxFKU-RFWchSICSJPCChI3CPNZ0kodyQoFlyQNP4HQUqT50XBkVzXbddwNf5xMkszBwgMBhBALtzLsiSxBkmNNO7ezFS0ogYRcySvIIKUGCAA)
> > > I have created a playground with basic types, see if you have better thought.
> > 
> > 
> > Hmmm. This is the best I could come up with: [Playground Link 2](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgCJzHZBvAsAKGWQFsIAuZAZzClAHMBuAogGzhDouto6cOQQBXKFAjgAEgAUKIQcQBG0PkQDKACwD2UMADk4pCgAoAHlxr0AlMgC8APirnezZCrAaADgGENCjEat2yABuGsAAJsrIAPQAVDHIAAJhEO6iCBgQYcgAqpQo7nBQeQBiLBoYyKDUEHBZMVHOkoV5ADIa6SwQpeVgRtz0Zjx0AfayCkrOnuyugiD+NvbyGhqd7JFTICrAnSBI84FLKzUg69OdKfuLy6snk+yeN3mXyIc3pyDFEKC9yIYjL9dju8AIJhMKdZ6vIEEAC+BAI32g8CQyAAQnA8gAlCB0ACixncAB4ACrICDGSAgMKUBxDezkynU5DYvEEnDOOhQDSCdyUAD8FDw-CIAG0aaBkMSALoC2n0SIwvhw-AEMAAT3cKAAshgEGoIJQSWSKWImf0OPZrMyIAgtGESQAaOUWviqjUoNodCDEhk2ZBNbTAOAsQnY21Qe0AcjEkeQAB9kJHkrGE5GYFAU4mAFZwSNO810Wy2V34dWa5AAcTE0GACGJtDodGgRoZpppBct7OFlTCg3lziIoibpjRGIgLPxROJxYHlRAMA0PopRjCGDgFHQmCdxF1+soFB1YD1BpJtn+nuD3oZSvhpfdkobTagNlnVZANbrj+bkco3KgSEzSM4HkbZgHVSN7DjV9q1oT9gEbb9MCgJswAgktbRAahrWHA0-T+Ch0SxHFJ0JSMkJQiCFmQd8AHdsMnQwyINVCLBLKIoklNRgBpMINANEAyOQGitAAa2QMhPAIDCsPMBCigoet4KfEUpT9EVZyFIgtK08IKEjcI81nbSh3JCgWXJA0-gdIydPnRcGRXNdt13A1-k07SPMHCAwGEEAu08gKtLEPSY2s7tAsVGyiBhMLtJiggpXQjRMLAZBZKfSgAHUwLUYF9wfJToBUtSNKins9IM2KApMkdzOMSyLCqzzQAXJcfkMVctxIFzKDcsrjO83z-MCgLgsTUL+q0yLwqmprkBhZAMUrGDa0UuTSPI7yILCxLb2k1L0wgLoygweg1qfCg3w-c7EMKCjO3c3TE0q5warMnELMoKznBa+zl1+Tq4Gco89z67tRB8qA-Pc7SxujAS5um6KwsVIA)
> > It lets you specify the trigger type once, and then all functions get the right types, vs on every parameter of every function.
> 
> In my thought, using an `as` operator in array literal is odd, for we would be tired to specify types over and over again.
> 
> But if this is the only way, I would go with that.

Yeah, I'm not very happy about it.  Using `as` is usually a last-ditch effort, because it's kind of lying to typescript.  Actually, if I play with it more, and I say `as Trigger<'ability'>` it won't throw any errors, which seems strictly worse than labelling each of the function args with types.

Ok, so that's not really a solution, sorry!

MaikoTan:
@quisquous I pushed a commit that guarantees the return value of the regexes helper function, maybe a combination with `as` would work?

quisquous:
> @quisquous I pushed a commit that guarantees the return value of the regexes helper function, maybe a combination with `as` would work?

Hmm! I am not sure I understand enough to know what that would fix.  I think the more that I think about the `as` solution, I think it is worrisome as it will coerce types.  If you take my original example and say `as Trigger<'ability'>` it doesn't throw any issues, for example.  I am not sure it is a good idea.  Maybe guaranteed return values will fix something, but I'd need to see a playground example to understand.

I did find that if I add a separate variable called `kind`, then that's enough to get type inference without specifying parameters.  See: [Playground Link 3](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEJwM4QEoQOYCiAHgA4A8AKshEZCACYbIZhSh4B81tEDTuhUsgDeAWABQyZHigB7AK4kMAfgBcIiVKkBtJqGQUAumuat2Abk3IAvpfHWJEhLJAtkAWTgBrCBThQ8CDABGmQAXmQACgBKdXQsAWJyAHIwf0CwZK4wrhAIAHdkRNJI1IgWZOi7CTAATxIUTzAEAAtyym46RlM2EE5woohnKHpKABoe9g5q8TqG5ABxXmhgBAo2PECoDpouphZe-oixSS0ofBo4zBx8JMppqykvUHp1CjHH5AQ4ABsfgCNEF51JEALZwZptDDqJqtdoUDjRcJcABusmA9DsDlm9RQ62Am2gAyWeTYaw2WzIyTg-2AP2AdSyyAAPotlmT8YTtql0kEsjMnC43GYuRgAJKgkj0hAMt4U6DaQwDbRWE5aLTPBjqHkBPkfU7q86BIgw7y+XnBC5EGL69VSb5-QEIYFRcGQ8pInIiay2mz6wzmIA)

However, I can't decide if using something like `kind` is an improvement.  It's kind of awkward, in some ways.  It might be better to just specify all the types on all the parameters after all.

After looking at this a bit, I think I am preferring just specifying types explicitly on all the parameters (but maybe you or others feel differently?).  Maybe typescript will get better in the future and learn how to infer these types.

panicstevenson:
I think another aspect that makes the specifying-types-explicitly option more appealing is that if we ever come up with a solution at a future date, it will be easier to reason about making that change than it would be to undo whatever we would implement now while making also that change.

types/trigger.d.ts
```diff
@@ -1,10 +1,16 @@
...
+export interface BaseRegExp<T> extends RegExp {
+  groups?: {
+    [s in T]?: string;
+  };
 }
 
+export type Matches<T> =
+  T extends BaseRegExp ? T['groups'] :
+  T extends RegExp ? { [s: string]: string } :
+  unknown;
```
quisquous, line 12: When is T not a BaseRegExp or a RegExp?

resources/netregexes.ts
```diff
@@ -108,7 +129,7 @@ export default class NetRegexes {
   /**
    * matches: https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#14-networkstartscasting
    */
-  static startsUsing(params?: Params<typeof startsUsingParams[number]>): RegExp {
+  static startsUsing(params?: Params<StartsUsingParams>): Return<StartsUsingParams> {
```
quisquous, line 132: I meant to say this earlier, but this is really great.  Thanks for making this all so generic.  (I learned a bunch about typescript that I didn't know!)

resources/regexes.ts
```diff
@@ -1,7 +1,11 @@
+import { BaseRegExp } from '../types/trigger';
+
 export type Params<T extends string> =
-  Partial<Record<T, string | string[]> &
+  Partial<Record<Exclude<T, 'timestamp' | 'capture'>, string | string[]> &
   { 'timestamp': string; 'capture': boolean }>;
 
+export type Return<T extends string> = BaseRegExp<Exclude<T, 'capture'>>;
```
quisquous, line 7: I do have one naming question here, in that I understand that this is `Return` because it's what that the regex/netregex functions return.  However, I think it *is* a regex.  Maybe it could be called `Regex` for clarity? (Maybe it'd be worth having `Regex` and `NetRegex` differently typed versions too so that you can't mix and match them?)

quisquous (commented):
Sorry again for the distraction around trying to make parameter types implicit in triggers.  I'm glad to have looked into it, even if it didn't turn up anything.

(I will add also that I think we should make some gh-pages / ci step that transpiles into JavaScript so that users can have a JavaScript example to copy from trigger files before we start converting those trigger files.)

types/trigger.d.ts
MaikoTan: Yep, when T is not either a BaseRegExp nor RegExp, then return an `unknown`.
Or we can return a `never` so that `tsc` would report an error that you have written a wrong generic type.

resources/regexes.ts
MaikoTan: OK, that is not difficult.

MaikoTan:
@quisquous I think I can merge this PR as a regex precision first and reintroduce a new PR when I have time to reconsider how to more precisely infer the type of matches. sorry to let it linger so long.

resources/conditions.ts
```diff
@@ -5,12 +5,37 @@
...
+  LosesEffectParams |
+  TetherParams |
+  WasDefeatedParams
+>>;
+
 export default {
-  targetIsYou(): (data: Data, matches: Matches) => boolean {
-    return (data: Data, matches: Matches) => data.me === matches.target;
+  targetIsYou(): (data: Data, matches: TargetableMatches) => boolean {
+    return (data: Data, matches: TargetableMatches) => data.me === matches?.target;
```
quisquous, line 35: Why does `matches?.target` need the null conditional operator? Shouldn't matches always be non-null?

quisquous (commented):
Thanks for finishing this off.  I think this is all really good.  I have one quick question:

resources/conditions.ts
MaikoTan: Just forgot to delete it(._.)

MaikoTan: Seems like it can be undefined sometimes?

quisquous: Sorry, could you explain to me how/when that could happen?

MaikoTan: for example, `Regexes.ability({ capture: false })`？

quisquous: Oh, right!

PR #2562, opened 2021-03-11
Raidboss: E3N timeline touch-up

JLGarber:
I happened to get E3N in Mentor Roulette tonight, and I noticed that the Rip Current at 77.4 can be skipped. I made a change to account for that.

While I was at it, I added more and wider sync windows to better match modern practice. This should guarantee that the user won't be de-synced for too long. (Previously they would have had to wait until the second Undersea Quake for the timeline to catch up.)

Finally, I added numbers to Spinning Dive and Temporary Current. This better fits with modern expectations of "if it's different and close together, differentiate it a little in the text."


quisquous:
Thank you! Much appreciated!!

PR #2563, opened 2021-03-11
update translate for CN server

Trim21:
Sorry for my poor English, I'll describe this PR in Chinese.

PGBO3UST()P076D0LH5}4)4

“中”的上面是场地正中间

地火L跑法，二穿一跑法现在基本已经绝迹了

不知道这地方应该如何措辞（

### DIFF
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
@@ -747,7 +747,7 @@ export default {
     {
       // Applies to both limit cuts.
       id: 'TEA Limit Cut Numbers',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         // Here and elsewhere, it's probably best to check for whether the user is the target first,
         // as that should short-circuit more often.
@@ -811,7 +811,7 @@ export default {
     {
       // Applies to both limit cuts.
       id: 'TEA Limit Cut Knockback',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         return data.me === matches.target && (/00(?:4F|5[0-6])/).test(getHeadmarkerId(data, matches));
       },
@@ -951,7 +951,7 @@ export default {
     },
     {
       id: 'TEA Ice Marker',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         return data.me === matches.target && getHeadmarkerId(data, matches) === '0043';
       },
@@ -993,7 +993,7 @@ export default {
     },
     {
       id: 'TEA Enumeration YOU',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         return data.me === matches.target && getHeadmarkerId(data, matches) === '0041';
       },
@@ -1011,7 +1011,7 @@ export default {
     },
     {
       id: 'TEA Enumeration Everyone',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         return getHeadmarkerId(data, matches) === '0041';
       },
@@ -1501,7 +1501,7 @@ export default {
     },
     {
       id: 'TEA Judgment Crystal',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         return data.me === matches.target && getHeadmarkerId(data, matches) === '0060';
       },
@@ -1680,7 +1680,7 @@ export default {
     },
     {
       id: 'TEA Cactbot Wormhole Strat',
-      netRegex: NetRegexes.headMarker({ }),
+      netRegex: NetRegexes.headMarker({}),
       condition: function(data, matches) {
         if (!data.options.cactbotWormholeStrat)
           return false;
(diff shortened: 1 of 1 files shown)


miaonima:
Thanks for the PR. Will test with video tonight.

ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.js
```diff
@@ -2873,47 +2873,47 @@ export default {
           de: 'Warte in der Mitte, ausweichen nach Norden',
           fr: 'Attendez au milieu, esquivez au Nord',
           ja: '中央から北へ',
-          cn: '中间 -> 北',
+          cn: '上 -> 右',
```
miaonima, line 2876: I think 右上 -> 右 would be better.

miaonima: Expecting "右 -> 中", then "中".

Trim21: which condition should be this?

Trim21:
I have tested `上 -> 右`(`waitMiddleDodgeNorth`) is correct

quisquous:
Sorry, I think somebody is going to have to explain to me what's going on here.  It looks like you're implementing some other strategy here, though?

My general feeling is that if you want to change the strategy that cactbot is using, then I think customizing the text via the config UI or via some `tea-cn.js` file that people can drop into their `user/raidboss/` folder might be a better approach.  I think it's generally better to have all languages be the same content, just translated.

Trim21:
@quisquous Yes, we are replacing CN translation with the most common (and currently the only one way in CN server) strategy in CN server.

I think it's not fair for CN users, as there already has been someone who would like to maintain it.

User config file is hard to maintain and hard to use for both authors and users, and built-in info text will misleading users without a specific user config file.

quisquous:
> @quisquous Yes, we are replacing CN translation with the most common strategy (and currently the only one strategy) in the CN server.
> 
> I think it's not fair for CN users, as there already has been someone who would like to maintain it.
> 
> User config file is hard to maintain and hard to use for both authors and users, and built-in info text will misleading users without a specific user config file.

The issue for me is that if there are people on non-Chinese servers who happen to have their cactbot language set to output in Chinese, then they will get a totally different strategy than the rest of their party who is not playing in Chinese.  This is why it's desirable to me to have all languages use the same strategy.

I think the user config file is much less hard to maintain these days with user subfolders.  You can create a separate file for this one override and have people put it in their `user/raidboss/` subfolder.

MaikoTan:
I would agree because I am the one, sometimes I'll using Japanese timeline with Chinese callout, and play in JP server.
Since you can now modify the output of any trigger via the Cactbot configuration UI, I don't think such inconsistencies are necessary.

quisquous:
I will also mention that there wasn't really a common strategy that I saw for this on any servers here.  I picked one just so that cactbot had *something*.  I think in general I try not to prescribe a strategy when there's multiple, but in this case I thought it was better to have something.  I would guess that the majority of people using cactbot on the international servers either turned this off or did something else entirely.

Trim21:
@quisquous OK, I'll close this pr.

I think we need an better way to share config file between users than copy and paste.

quisquous:
> I think we need an better way to share config file between users than copy and paste.

Sure! I'm happy to talk about what that could look like.  I think that would probably help everybody.

When you say "copy and paste" do you mean "copy and paste files into user subdirectories"?

Trim21:
Yes, I think it's a better way to share user-config than copy and paste code into `user/raidboss.js`.

I'm planning to write an ACT plugin to fetch remote user config files and update local user config files.

MaikoTan:
> Yes, I think it's a better way to share user-config than copy and paste code into `user/raidboss.js`.
> 
> I'm planning to write an ACT plugin to fetch remote user config files and update local user config files.

Why not implement this feature in cactbot?

Trim21:
> Why not implement this feature in cactbot?

I know nothing about C#

MaikoTan:
> > Why not implement this feature in cactbot?
> 
> I know nothing about C#
> 

I mean, in JS.
`fetch()`, `localStorage.setItem()` and `eval()`.

Trim21:
> > > Why not implement this feature in cactbot?
> > 
> > 
> > I know nothing about C#
> 
> I mean, in JS.
> `fetch()`, `localStorage.setItem()` and `eval()`.

I think it's better to update local files than cache them in localStorage.

Maybe I should learn some c# and add them in cactbot

quisquous:
> > When you say "copy and paste" do you mean "copy and paste files into user subdirectories"?
> >
> Yes, I think it's a better way to share user-config than copy and paste code into `user/raidboss.js`.

I think there continues to be a misunderstanding here.  `user/raidboss.js` is not the only way to append user code.  There is also support for user**sub**directories.  In other words, you can create a file called `user/raidboss/tea-cn-trines.js` or something like that, and that will also be loaded.  The only thing others have to do is to put that file in their `user/raidboss/` folder.  There is no need for copying and pasting of code, only of files.

I hope that helps clear things up.

> I'm planning to write an ACT plugin to fetch remote user config files and update local user config files.

I'd be interested in having some discussion about what this would look like (on some new issue) before you get too far into implementing such a feature.

Trim21:
> I think there continues to be a misunderstanding here. `user/raidboss.js` is not the only way to add user code. There is also support for user **sub**directories. In other words, you can create a file called `user/raidboss/tea-cn-trines.js` or something like that, and that will also be loaded. The only thing others have to do is to put that file in their `user/raidboss/` folder. There is no need for copying and pasting of code, only of files.

I understand. I mean the file content is JS code.

Trim21:
> I'd be interested in having some discussion about what this would look like (on some new issue) before you get too far into implementing such a feature.

I think I won't write it in C# or JS.

I'd like to write a simple act plugin that could fetch remote files and update files on disk, it will do nothing with cactbot itself.

PR #2566, opened 2021-03-12
Refactor: return steps(DNC), arcanums(AST) as list

### LINKED ISSUE #2549: Feature Request: Job details return proper data?
Most of the properties in the `JobDetail` data currently accepted by the JS front-end are simple and easy to understand, such as PLD's `oath` which is a number, and MCH's `overheatActive` which is obviously a boolean value.
But there are also some less clear properties, such as `arcanums` for AST and `steps` for DNC, which in theory should be arrays of strings, but currently, the C# backend returns data as strings split by ", ", which makes it difficult for the JS frontend to traverse and calculate the data.
Therefore, could this data be output to the front end in the correct way (i.e. as an array of strings)?

Translated with www.DeepL.com/Translator (free version)

MaikoTan:
**BREAKING CHANGES**

This commit should be merged at a patch update,
in order to reduce unavailability.

Fixes #2549

### DIFF
--- a/plugin/CactbotEventSource/FFXIVProcessCn.cs
+++ b/plugin/CactbotEventSource/FFXIVProcessCn.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
 using Newtonsoft.Json.Linq;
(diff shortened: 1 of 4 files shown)


quisquous:
Thanks! Sounds great!

PR #2567, opened 2021-03-12
add rule to check output method call

Trim21:
will throw errors on output strings like this

```JS
{
  id: 'TEA Trigger 2',
  regex: /Fluid Swing/,
  infoText: function(data, _, output) {
    if (data.tt)
      return output.tankBusterOn({ play: data.me }); <- error missing `play` in template
    return output.tankBusterOn(); <- error missing params in function call
    return output.tankBusters({ 'a': '1', 'b': '2' });
  },
  outputStrings: {
    tankBusters: {
      en: '${a} ${b}',
      de: '${a} ${b}',
      fr: '${a} ${b}',
      ja: '${a} ${b}',
      cn: '${a} ${b}',
      ko: '${a} ${b}',
    },
    tankBusterOn: {
      en: 'Tank Buster on ${player}',
      de: 'Tank buster auf ${player}',
      fr: 'Tank buster sur ${player}',
      ja: '${player}にタンクバスター',
      cn: '死刑 点 ${player}',
      ko: '"${player}" 탱버',
    },
  }
};

```


eslint/cactbot-output-strings.js
```diff
@@ -54,12 +57,43 @@ const ruleModule = {
...
+          }
+          const triggerID = node.parent.parent.properties.find((prop) => prop.key && prop.key.name === 'id').value.value;
+          outputTemplates.set(triggerID, outputTemplateKey);
+        }
+      },
+      'Program > VariableDeclaration > VariableDeclarator > ObjectExpression': function(node) {
         globalVars.set(node.parent.id.name, getAllKeys(node.properties));
       },
-      [`Property[key.name=/${textProps.join('|')}/] > :function`](node) {
+      [`Property[key.name=/${textProps.join('|')}/] > :function`]: function(node) {
```
MaikoTan, line 96: Why do you change the function declare type?

eslint/cactbot-output-strings.js
```diff
@@ -54,12 +57,43 @@ const ruleModule = {
...
+      /**
+       *
+       * @param node {t.ObjectExpression}
+       */
+      'Property[key.name=/outputStrings/] > ObjectExpression': function(node) {
+        const outputTemplateKey = {};
+        for (const outputString of node.properties) {
+          // each outputString
+          const values = [];
+          outputString.value.properties.forEach((x) => {
```
MaikoTan, line 74: The `outputString` here can be `SpreadElement`, so you have to filter(and add its properties with the `getAllKeys` function) it.
For example using the `@babel/types` helper functions.

eslint/cactbot-output-strings.js
```diff
@@ -19,10 +19,13 @@ const ruleModule = {
     messages: {
       noOutputStrings: 'no outputStrings in trigger',
       notFoundProperty: 'no \'{{prop}}\' in \'{{outputParam}}\'',
+      notFoundTemplate: 'no \'{{prop}}\' in \'{{outputParam}}\'',
```
MaikoTan, line 22: Redundant duplicated?

eslint/cactbot-output-strings.js
```diff
@@ -54,12 +57,43 @@ const ruleModule = {
...
+       */
+      'Property[key.name=/outputStrings/] > ObjectExpression': function(node) {
+        const outputTemplateKey = {};
+        for (const outputString of node.properties) {
+          // each outputString
+          const values = [];
+          outputString.value.properties.forEach((x) => {
+            values.push(x.value.value);
+          });
+          const v = values.map((x) => Array.from(x.matchAll(/\${\s*([^}\s]+)\s*}/g))).map((x) => x.length ? x.map((v) => v[1]) : null);
```
MaikoTan, line 77: `v` can be more clear?

Trim21:
example: https://gist.github.com/Trim21/aca4917b1abb077576e444889fa403df

eslint/cactbot-output-strings.js
```diff
@@ -19,10 +19,13 @@ const ruleModule = {
     messages: {
       noOutputStrings: 'no outputStrings in trigger',
       notFoundProperty: 'no \'{{prop}}\' in \'{{outputParam}}\'',
+      notFoundTemplate: '`output.{{outputParam}}(data)` doesn\'t take property \'{{prop}}\' on `data`',
```
quisquous, line 22:
```suggestion
      notFoundTemplate: '`output.{{prop}}(...)` doesn\'t have template \'{{template}}\'.',
```

This was confusing to read compared to `notFoundProperty`, because `outputParam` on `notFoundProperty` refers to the `output` variable, but `outputParam` on `notFoundTemplate` refers to the property on `output`.  I think this naming makes things more clear.  What do you think?

eslint/cactbot-output-strings.js
```diff
@@ -19,10 +19,13 @@ const ruleModule = {
...
+      tooMuchParams: 'function `output.{{call}}()` take only {{num}} params',
```
quisquous, line 23: Is it possible to write a "missingTemplate", where it says which template parameters are missing?

eslint/cactbot-output-strings.js
```diff
@@ -54,8 +57,36 @@ const ruleModule = {
...
+      const outputTemplateKey = {};
+      for (const outputString of node.properties.filter((s) => s.type !== 'SpreadElement' && s.value.type !== 'MemberExpression')) {
+        // each outputString
+        const values = [];
+
+        outputString.value.properties.forEach((x) => {
+          if (x.value.value !== undefined)
+            values.push(x.value.value);
+        });
+        const templateIDs = values.map((x) => Array.from(x.matchAll(/\${\s*([^}\s]+)\s*}/g))).map((x) => x.length ? x.map((v) => v[1]) : null);
```
quisquous, line 71: This is just a subjective opinion, but can you name anything with an `id` suffix as `Id` and not `ID`, e.g. `templateId` to be consistent with other uses in cactbot and even in this file (e.g. `messageId`)?

eslint/cactbot-output-strings.js
Trim21: `tooMuchParams` means `output.call()` with template or `output.call({...})` without template. 

I'll add a error for template in outputString is not fulfilled

eslint/cactbot-output-strings.js
```diff
@@ -19,10 +19,14 @@ const ruleModule = {
     messages: {
       noOutputStrings: 'no outputStrings in trigger',
       notFoundProperty: 'no \'{{prop}}\' in \'{{outputParam}}\'',
+      notFoundTemplate: '`output.{{prop}}(...)` doesn\'t have template \'{{template}}\'.',
+      missingTemplateValue: 'template \'{{prop}}\' is missing in function call',
+      tooManyParams: 'function `output.{{call}}()` take only {{num}} params',
```
quisquous, line 24: Does `notFoundTemplate` and `missingTemplateValue` already cover `tooManyParams`, and `tooManyParams` can be removed now? It seems like if you have 0 template parameters and you try to provide one, then you would get `notFoundTemplate`, and if you try to provide too many then you would get `missingTemplateValue`?

Trim21: I don't think so . If method is called `output.text({}, shouldNotExist)`

quisquous: Oh!  Thank you for the explanation.  I see what it's trying to do; it's trying to check the number of parameters to output.  I am unable to get the "zero" case to trigger.  Can you give me an example of incorrect text for zero?  I also think that it's ok to have 0 or 1 output properties for 0 template parameters (`output.text({})` is valid, I think?).

Can you try these triggers? You'll also need to fix code to handle missing ids, which currently throws an error in two places.
```
    {
      // This is valid.
      alarmText: (data, _, output) => output.text({}),

      // This is correctly an error.
      alertText: (data, _, output) => output.text(true, false),

      // These should be an error, but are not.
      infoText: (data, _, output) => output.text({ foo: 3 }),
      tts: (data, _, output) => output.text(true),

      outputStrings: {
        text: {
          en: 'text',
        },
      },
    },
    {
      // This is valid.
      alarmText: (data, _, output) => output.textWithParam({ param: 3 }),

      // This is correctly an error.
      alertText: (data, _, output) => output.textWithParam({}),

      // This should be an error but is not.
      infoText: (data, _, output) => output.textWithParam(),

      // This is an error, but has incorrect error text:
      // error  template 'param' is missing in function call               rulesdir/cactbot-output-strings
      // error  `output.param2(...)` doesn't have template '{{template}}'  rulesdir/cactbot-output-strings
      tts: (data, _, output) => output.textWithParam({ param2: 3 }),

      outputStrings: {
        textWithParam: {
          en: '${param}',
        },
      },
    },
```

I could also argue that typescript will handle the number and type of parameters eventually? Typescript will never know about template parameters so that's great to check, but it should eventually handle the number of parameters on its own when we convert trigger files and give `output` a type.

Trim21: yes, typescript should be able to check the number of args. But we haven't started migrating triggers to ts so I think it's necessary now.

eslint/cactbot-output-strings.js
```diff
@@ -97,6 +139,73 @@ const ruleModule = {
             },
           });
         }
+        if (t.isIdentifier(node.property) && stack.outputProperties.includes(node.property.name)) {
+          const args = node.parent.callee.parent.arguments;
+          const outputOfTriggerID = stack.outputTemplates ?? {};
+          const outputTemplate = outputOfTriggerID?.[node.property.name];
+
+          if (args.length === 0) {
+            if ((node.property.name in outputOfTriggerID) && outputTemplate === undefined) {
```
quisquous, line 148: Can you give me an example that would trigger this?

eslint/cactbot-output-strings.js
```diff
@@ -97,6 +139,73 @@ const ruleModule = {
             },
           });
         }
+        if (t.isIdentifier(node.property) && stack.outputProperties.includes(node.property.name)) {
+          const args = node.parent.callee.parent.arguments;
+          const outputOfTriggerID = stack.outputTemplates ?? {};
```
quisquous, line 144: `outputOfTriggerID` => `outputOfTriggerId`

eslint/cactbot-output-strings.js
quisquous: These two from my previous example still do not create errors.  Should they?
```
      infoText: (data, _, output) => output.text({ foo: 3 }),
      tts: (data, _, output) => output.text(true),
```

Also, I think grammar-wise, saying "take only 1 params" could be better.  I am still not sure why there is a zero case or how to trigger that, but maybe this could say: `'function \'output.{{call}}()\' takes only 1 parameter',` for the one case?

package.json
```diff
@@ -21,6 +21,7 @@
     "find-translations": "node --loader ts-node/esm util/find_missing_translations.js"
   },
   "devDependencies": {
+    "@babel/types": "7.12.7",
```
Trim21, line 24: this is required because 7.13.x break current code.

package.json
```diff
@@ -21,6 +21,7 @@
     "find-translations": "node --loader ts-node/esm util/find_missing_translations.js"
   },
   "devDependencies": {
+    "@babel/types": "7.12.*",
```
Trim21, line 24: 7.13.x breaks current code.

quisquous: Wait, the newer version breaks current code? What is the issue?

Trim21: It breaks the function `getAllKeys` wroten by @MaikoTan , I didn't dig into details

eslint/cactbot-output-strings.js
```diff
@@ -19,6 +18,10 @@ const ruleModule = {
     messages: {
       noOutputStrings: 'no outputStrings in trigger',
       notFoundProperty: 'no \'{{prop}}\' in \'{{outputParam}}\'',
+      notFoundTemplate: '`output.{{prop}}(...)` doesn\'t have template \'{{template}}\'.',
+      missingTemplateValue: 'template \'{{prop}}\' is missing in function call',
+      tooManyParams: 'function `output.{{call}}()` take only 1 params',
```
quisquous, line 23:
```suggestion
      tooManyParams: 'function `output.{{call}}()` takes only 1 parameter',
```

eslint/cactbot-output-strings.js
```diff
@@ -19,6 +18,10 @@ const ruleModule = {
...
+      typeError: 'function `output.{{call}}(...) take only object as params',
```
quisquous, line 24:
```suggestion
      typeError: 'function `output.{{call}}(...) only takes an object as a parameter',
```

package.json
MaikoTan: Okay, i'll try to figure it out when I have time.

PR #2570, opened 2021-03-13
Refactor: TypeScript-ize resources/resourcesbar

MaikoTan:


resources/resourcebar.ts
```diff
@@ -1,96 +1,123 @@
...
   // progress bar starts full and goes to empty.
-  set style(s) {
-    this.setAttribute('style', s);
+  set style(s: 'empty' | 'full' | null) {
+    this.setAttribute('style', s ?? 'empty');
   }
-  get style() {
-    return this.getAttribute('style');
+  get style(): 'empty' | 'full' | null {
+    return this.getAttribute('style') as 'empty' | 'full' | null;
```
MaikoTan, line 120: The `style` property here conflict the original implementation of `HTMLElement`, I think it should be changed to another name, like `styleFill` or something?

quisquous: Sounds good!

MaikoTan:
Done.
And I didn't found there is code using `resources-bar`'s style property?
Maybe this can be merge now.

quisquous:
> And I didn't found there is code using `resources-bar`'s style property?
> Maybe this can be merge now.

Yes, it looks like this is only used by the rdm resouce bars.

PR #2571, opened 2021-03-13
CI: add type checking to lint

MaikoTan:


quisquous:
I think I'm a little confused why this is in both test and lint? I think it would make more sense to only put it in lint?

Also, is adding `npm run` entries with colons something other folks do? Mostly curious because previous entries have all used hyphens (e.g. `tsc-no-emit`).

MaikoTan:
> I think I'm a little confused why this is in both test and lint? I think it would make more sense to only put it in lint?

Because they are triggered by different files?
But looks like putting in lint only would make sense.

> Also, is adding `npm run` entries with colons something other folks do? Mostly curious because previous entries have all used hyphens (e.g. `tsc-no-emit`).

Sure!

quisquous:
> > I think I'm a little confused why this is in both test and lint? I think it would make more sense to only put it in lint?
> 
> Because they are triggered by different files?
> But looks like putting in lint only would make sense.

Sorry, is that true? Are there files in test that should be covered by this tsc check that are not covered by lint?

MaikoTan:
Sorry, my bad, I have misunderstood about that the `ui/raidboss/**` pattern in `test.yml` matches trigger files which should be type-checked as well, and just realize that these files were already covered by `*.ts` and `*.js` pattern. So never mind...`(´。＿。｀)`

quisquous:
> Sorry, my bad, I have misunderstood about that the `ui/raidboss/**` pattern in `test.yml` matches trigger files which should be type-checked as well, and just realize that these files were already covered by `*.ts` and `*.js` patterns in `lint.yml`. So never mind...`(´。＿。｀)`

No need for apologies, it was an honest question! I was wondering if I was misunderstanding here.  Thanks for thinking about it too so we can do the more efficient thing here.  :)

PR #2573, opened 2021-03-13
i18n: localize 'Active:' text in timeline

MaikoTan:
@Akurosia @Jaehyuk-Lee May you have time to check if this localization correct?

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -59,7 +59,7 @@ const activeText = {
   fr: 'Active :',
   ja: '(進行):',
   cn: '(进行中):',
-  ko: 'Active:',
+  ko: '시전중:',
 };
 
 function computeBackgroundColorFrom(element, classList) {
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -55,7 +55,7 @@ const timelineInstructions = {
 
 const activeText = {
   en: 'Active:',
-  de: 'aktiv:',
+  de: 'Aktiv:',
   fr: 'Active :',
   ja: '(進行):',
   cn: '(进行中):',
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -57,9 +57,9 @@ const activeText = {
   en: 'Active:',
   de: 'aktiv:',
   fr: 'Active :',
-  ja: '進行:',
-  cn: '进行中:',
-  ko: '진행중:',
+  ja: '(進行):',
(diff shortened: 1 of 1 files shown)


Jaehyuk-Lee:
Actually, I think Korean text to be remain as before, "Active: "
Some skill names are similar with "□□□: □□□□" but there's never "(English): (한국어)", always "(한국어): (한국어)".
English letters and Korean letters are very different so people's eye catches "Active: 스킬" only "스킬" part.
I don't think people be confused if "진행중 :" as a part of skill name though, people need to read and recognize it as it's not a skill name.

MaikoTan:
Hmmmm...
Maybe we can add brackets on it, like `「진행중」: 긴 시전바`?
If I add `「进行中」` and `「進行」` on Chinese and Japanese timeline, folks would never confused because there are no `「」` in any skills.

Jaehyuk-Lee:
Korean never use 「」 brackets except proper noun, for example, book name, thesis or paper. That's awkward in Korean, the `:` is correct with the "Active".
I'm asking for opinion on Korean community, I got new translation, `시전중`.

Jaehyuk-Lee:
I don't play the game lately, I would myself run some raid with those changes if I have my account active.
There's no strong opinion on one side but slightly on the "let's translate" side.
So, let's first apply `시전중: (skill name)` this time, this is not a critical thing, let's see.

MaikoTan:
I am now on the train, without a laptop. maybe you can comment a suggestion so that I can apply your request in one-click.

ui/raidboss/timeline.js
```diff
@@ -53,6 +53,15 @@ const timelineInstructions = {
   ],
 };
 
+const activeText = {
+  en: 'Active:',
+  de: 'aktiv:',
+  fr: 'Active :',
+  ja: '(進行):',
+  cn: '(进行中):',
+  ko: 'Active:',
```
Jaehyuk-Lee, line 62:
```suggestion
  ko: '시전중:',
```

Jaehyuk-Lee: I didn't know that suggesting a commit was this easy, I should use this 😃

ui/raidboss/timeline.js
```diff
@@ -53,6 +53,15 @@ const timelineInstructions = {
   ],
 };
 
+const activeText = {
+  en: 'Active:',
+  de: 'aktiv:',
```
Akurosia, line 58: as it stand alone and just show a state, i would capitalize it to `Aktiv:`

PR #2574, opened 2021-03-13
raidboss: DRS TA Unwavering Apparations Callout

Legends0:
I wrote this up overnight... could be better optimized. To be tested...

Goal is to callout the location of the box  (safe spot) that is safe to stand in to avoid all cleaves and also call where to stand relative to the safe spot if needed to change current temperature.


ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -20,6 +20,21 @@ const tankBusterOnParty = (data, matches) => {
   return data.party.inParty(data.target);
 };
 
+const getFacing = (combatant) => {
+  // Snap heading to closest card/intercard (aka PI/4).  N = PI, E = PI/2.
+  const facing = Math.round(combatant.Heading * 4 / Math.PI) / 4 * Math.PI;
```
quisquous, line 25: Ah, I see this came from e9s, probably.  I think it's a little weird floating point-wise to multiply by Math.PI here and then use Math.PI below.  This is especially true when north is both `Math.PI` and `-Math.PI`.

I think this might be more robust to do something like:
```
// 0 = N, 1 = E, 2 = S, 3 = W
const facing = (2 - Math.round(combatant.Heading * 4 / Math.PI) / 2) % 4;
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +611,454 @@ export default {
...
+          combatantDataBoss.combatants &&
+          combatantDataBoss.combatants.length) ||
+          !(combatantDataAvatars !== null &&
+          combatantDataAvatars.combatants &&
+          combatantDataAvatars.combatants.length)) {
+          data.safeZone = null;
+          return;
+        }
+
+        // we need to filter for the Trinity Avowed with the lowest ID
```
quisquous, line 717: I think arguably you could figure this out from positional data too, if you're calling getCombatants? Are the ids here a known way this works?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +611,454 @@ export default {
...
+        // that one is always cleaving on one of the cardinals
+        // Trinity Avowed is always East (-267, -87)
+        const eastCombatant =
+          combatantDataBoss.combatants.sort((a, b) => a.ID - b.ID).pop();
+
+        // we need to filter for the three Avowed Avatars with the lowest IDs
+        // as they cast cleave at the different cardinals
+        // First Avowed Avatar is always North (-277, -97)
+        const northCombatant =
+          combatantDataAvatars.combatants.sort((a, b) => a.ID - b.ID).pop();
```
quisquous, line 727: I don't think you need to sort this repeatedly.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +611,454 @@ export default {
...
+      // 595B = left cleave cold (2) paired with 595D
+      // 595C = left cleave heat (2) paired with 595A
+      // 595D = left cleave cold (2) paired with 595B
+      netRegex: NetRegexes.startsUsing({ source: ['Trinity Avowed', 'Avowed Avatar'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      netRegexDe: NetRegexes.startsUsing({ source: ['Trinität Der Eingeschworenen', 'Spaltteil der Eingeschworenen'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      netRegexFr: NetRegexes.startsUsing({ source: ['Trinité Féale', 'Clone De La Trinité Féale'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      netRegexJa: NetRegexes.startsUsing({ source: ['トリニティ・アヴァウド', 'アヴァウドの分体'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      run: (data, matches) => {
+        data.blades = data.blades || {};
+        data.blades[matches.targetId.toUpperCase()] = matches.id;
```
quisquous, line 1059: sourceId?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: It is possible to get their position from X and Y, but from the few logs I viewed it's always been the same four combatant order so I just assumed they would be in the right spot. It's a bit of extra code to handle if they could be in different locations.

quisquous: Sounds good! Mostly just curious.  :)

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Ah, I never worked with the heading before. Logs were showing old data and I never tested out which headings map to which cardinal. I guess sometimes it's negative sometimes its positive pi radians. This works out when I test for the four cardinals. My original values were wrong then.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Woops, overlooked that one. Thanks.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: You can use the cactbot test overlay to see your character heading, if you need an easy example.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I am looking at more logs other than my own as mine always produced the same positions, I think I may have to add in code for to verify the location as I see logs on fflogs where they are not always in the same position.

https://www.fflogs.com/reports/zZgdKMLvjDB6aFwT#fight=7&hostility=1&view=replay
Trinity Avowed East, but the Avoweds appear to be in different locations assuming that FFlogs numbers are based on appearance order.

Legends0: I am going to continue to assume the boss always goes East. It auto moves to that position there before starting the mechanic.

Legends0:
I wonder if I am overlooking that two adds cleaving a single zone could create a zone with 0 temperature but not be safe...

Legends0:
So the change in b2d617a now has two maps, and since the zones adjacent to the zone that no add is cleaving are only cleaved by one add we shouldn't have to adjust for multiple cleaves in a zone as it should just ignore those keys in the map.

Legends0:
Looking over the logs some more as I want to test out that I actually have the correct sort so that overlay gives the correct combatants... there is more going on with the mechanic such that 4 other entities are involved. These other entities are perhaps what explain the additional spell IDs.

I may need to ensure that the npc ids that I am aiming for match up with the spells that I regex and that I am sorting the right direction (I may need to reverse the sort).

I based the current code off of the entities that received the Cold/Hot Blade buffs, which they do cast the Blade of Entropy, but I need to ensure that the Blades of Entropy I am checking on are coming from the same buffed target. I may modify the comments about Blade of Entropy spell ids with the clarification as to why there are two of each spell as it may turn out that these other 4 entities are casting the paired spell and I need to determine which of the two belong to which entities. I also may need to check that the the entities don't share from the entire pool of spell ids.

Legends0:
I must be misunderstanding the value that combatant.ID is. I had assumed it was a string such as '400008f4, but if that were the case, then the sort methods that I see used wouldn't give the correct results, they would just reverse the array. a.localeCompare(b) would do the correct sort of these values.

quisquous:
> I must be misunderstanding the value that combatant.ID is. I had assumed it was a string such as '400008F4', but if that were the case, then the sort methods that I see used wouldn't give the correct results, they would just reverse the array. a.localeCompare(b) would do the correct sort of these values.

`combatant.ID` is the integer value of the id.  I am not sure I understand why `sort()` would not work here regardless.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +610,470 @@ export default {
...
+      // 595B = left cleave cold (2) paired with 595D
+      // 595C = left cleave heat (2) paired with 595A
+      // 595D = left cleave cold (2) paired with 595B
+      netRegex: NetRegexes.startsUsing({ source: ['Trinity Avowed', 'Avowed Avatar'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      netRegexDe: NetRegexes.startsUsing({ source: ['Trinität Der Eingeschworenen', 'Spaltteil der Eingeschworenen'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      netRegexFr: NetRegexes.startsUsing({ source: ['Trinité Féale', 'Clone De La Trinité Féale'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      netRegexJa: NetRegexes.startsUsing({ source: ['トリニティ・アヴァウド', 'アヴァウドの分体'], id: ['5942', '5943', '5946', '5947', '5956', '5957', '595A', '595B'] }),
+      run: (data, matches) => {
+        data.blades = data.blades || {};
+        data.blades[matches.sourceId.toUpperCase()] = matches.id.toUpperCase();
```
Legends0, line 1074: > 
> `combatant.ID` is the integer value of the id. I am not sure I understand why `sort()` would not work here regardless.

Ah, thanks for that. I think I should convert matches.id to integer value here then. 

Later on, the map is used to match combatant.ID to the spell id of the Blade of Entropy they cast, but the key there is a string of the hex value of the ID.

Legends0:
I think the only lingering issue to verify is that the heading data is accurate when delayed 1 second. So long as we get correct heading data, it should now work. I likely won't be able to test until next week.

Also the callout is hard to get.

For some debug here are some lines to add:
```
        console.log(`Boss (${eastCombatant.ID}) heading ${eastCombatant.Heading}, Position: ${eastCombatant.PosX}, ${eastCombatant.PosY}`);
        console.log(`First (${avatarOne.ID}) avatar heading ${avatarOne.Heading}, Position: ${avatarOne.PosX}, ${avatarOne.PosY}`);
        console.log(`Second (${avatarTwo.ID}) avatar heading ${avatarTwo.Heading}, Position: ${avatarTwo.PosX}, ${avatarTwo.PosY}`);
        console.log(`Third (${avatarThree.ID}) avatar heading ${avatarThree.Heading}, Position: ${avatarThree.PosX}, ${avatarThree.PosY}`);
```

```
        console.log(`East combatant facing ${eastCombatantFacing} cleaving ${bladeSides[eastCombatantBlade]}`);
        console.log(`North combatant facing ${northCombatantFacing} cleaving ${bladeSides[eastCombatantBlade]}`);
        console.log(`West combatant facing ${westCombatantFacing} cleaving ${bladeSides[eastCombatantBlade]}`);
        console.log(`South combatant facing ${southCombatantFacing} cleaving ${bladeSides[eastCombatantBlade]}`);
```

```
        if (data.currentBrand) {
          currentBrand = data.currentBrand;
          console.log(`Brand is currently ${currentBrand}`);
        }
        if (data.currentTemperature) {
          currentTemperature = data.currentTemperature;
          console.log(`Temperature is currently ${currentTemperature}`);
        }
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -20,6 +20,20 @@ const tankBusterOnParty = (data, matches) => {
   return data.party.inParty(data.target);
 };
 
+const getFacing = (combatant) => {
+  // Snap heading to closest card.
+  // N = 0, E = 1, S = 2, W = 3
+  return (2 - Math.round(combatant.Heading * 4 / Math.PI) / 2) % 4;
```
Legends0, line 26: I think truncating like so the problem?
Math.trunc(2 - Math.round(heading * 4 / Math.PI) / 2) % 4

Legends0:
Thinking on this a bit more, I may have over complicated the mechanic. I am pretty sure it's simplified to this:
```
const adjacentZones = {
    0: bladeValues[northCombatantBlade], 
    1: bladeValues[eastCombatantBlade],
    2: bladeValues[southCombatantBlade],
    3: bladeValues[westCombatantBlade],
 };

if ((eastCombatantFacing === 0 || eastCombatantFacing === 2) && (southCombatantFacing === 1 || southCombatantFacing === 3)) {
  // if east cleaving east and south cleaving south, safe spot is se
  safeZone = output.southeast();
} else if ((eastCombatantFacing === 0 || eastCombatantFacing === 2) && (northCombatantFacing === 1 || northCombatantFacing === 3)) {
  // if east cleaving east and north cleaving north, safe spot is ne
  safeZone = output.northeast();
} else if ((westCombatantFacing === 0 || westCombatantFacing === 2) && (southCombatantFacing === 1 || southCombatantFacing === 3)) {
  // if west cleaving west and south cleaving south, safe spot is sw
  safeZone = output.southwest();
} else if ((westCombatantFacing === 0 || westCombatantFacing === 2) && (northCombatantFacing === 1 || northCombatantFacing === 3)) {
  // if west cleaving west and north cleaving north, safe spot is nw
  safeZone = output.northwest();
} else {
  // facing did not evaluate properly
  safeZone = null;
}
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I am going to leave this issue in for now and hope the delaySeconds 1 resolves it. The mechanic has plenty of time to resolve. It should at least always call out the adjacent spots correctly now, but as those are relative you still need someone to give you a safe spot callout else you are dealing with a 1 in 4 odds.

quisquous:
> Thinking on this a bit more, I may have over complicated the mechanic. I am pretty sure it's simplified to this:
> 
> ```
> const adjacentZones = {
>     0: bladeValues[northCombatantBlade], 
>     1: bladeValues[eastCombatantBlade],
>     2: bladeValues[southCombatantBlade],
>     3: bladeValues[westCombatantBlade],
>  };
> 
> if ((eastCombatantFacing === 0 || eastCombatantFacing === 2) && (southCombatantFacing === 1 || southCombatantFacing === 3)) {
>   // if east cleaving east and south cleaving south, safe spot is se
>   safeZone = output.southeast();
> } else if ((eastCombatantFacing === 0 || eastCombatantFacing === 2) && (northCombatantFacing === 1 || northCombatantFacing === 3)) {
>   // if east cleaving east and north cleaving north, safe spot is ne
>   safeZone = output.northeast();
> } else if ((westCombatantFacing === 0 || westCombatantFacing === 2) && (southCombatantFacing === 1 || southCombatantFacing === 3)) {
>   // if west cleaving west and south cleaving south, safe spot is sw
>   safeZone = output.southwest();
> } else if ((westCombatantFacing === 0 || westCombatantFacing === 2) && (northCombatantFacing === 1 || northCombatantFacing === 3)) {
>   // if west cleaving west and north cleaving north, safe spot is nw
>   safeZone = output.northwest();
> } else {
>   // facing did not evaluate properly
>   safeZone = null;
> }
> ```

I have seen this "guide": 
image

...which implies that you just need to find a cardinal that is looking towards that cardinal or opposite (e.g. north looking north or south) and then its cleave direction will tell you the safe spot.  That might make the logic easier.  I think also using some `const dirNum = { north: 0, east: 1, south: 2, west: 3 }` might make it easier to read `westCombatantFacing === dirNum.north`

Legends0:
> 
> I have seen this "guide":
> 
> ...which implies that you just need to find a cardinal that is looking towards that cardinal or opposite (e.g. north looking north or south) and then its cleave direction will tell you the safe spot. That might make the logic easier. I think also using some `const dirNum = { north: 0, east: 1, south: 2, west: 3 }` might make it easier to read `westCombatantFacing === dirNum.north`

I used this guide as a reference to fix the logic. I think I had this before I simplified, just overtly complex. This _should_ be correct now.

quisquous:
This seemed to work just fine for me too!

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+      },
+      infoText: (data, _, output) => {
+        return !data.safeZone ? output.unknown() : data.safeZone;
+      },
+      outputStrings: {
+        getCleaved: {
+          en: 'Safe Spot inside ${dir1}, Get cleaved 1 square ${dir2}',
+        },
+        safeSpot: {
+          en: 'Safe Spot: inside ${dir}',
```
quisquous, line 903:
```suggestion
          en: '${dir} Safe Spot',
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          data.safeZone = output.safeSpot({ dir: safeZone });
+        else
+          data.safeZone = null;
+      },
+      infoText: (data, _, output) => {
+        return !data.safeZone ? output.unknown() : data.safeZone;
+      },
+      outputStrings: {
+        getCleaved: {
+          en: 'Safe Spot inside ${dir1}, Get cleaved 1 square ${dir2}',
```
quisquous, line 900:
```suggestion
          en: '${dir1} Safe Spot => ${dir2} for cleave',
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+
+        // Only need to check cleaves from two parallel clones to determine safe spots
+        // because if the clone is cleaving inside, then we know where other clones
+        // are cleaving in order to make a '+' where the ends are each cleaved by one
+        // clone and the middle square is safe
+        let safeZone = null;
+        let adjacentZones = {};
+        if ((northCombatantFacing === dirNum.north && bladeSides[northCombatantBlade]) ||
+          (northCombatantFacing === dirNum.south && !bladeSides[northCombatantBlade])) {
+          // North clone cleaving inside east
```
quisquous, line 812:
```suggestion
          // North clone cleaving inside east (and therefore east clone cleaving north).
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          safeZone = output.southwest();
+          adjacentZones = {
+            0: bladeValues[eastCombatantBlade],
+            1: bladeValues[northCombatantBlade],
+            2: bladeValues[southCombatantBlade],
+            3: bladeValues[westCombatantBlade],
+          };
+        } else if ((northCombatantFacing === dirNum.north && !bladeSides[northCombatantBlade]) ||
+          (northCombatantFacing === dirNum.south && bladeSides[northCombatantBlade])) {
+          // North clone cleaving inside west
```
quisquous, line 822:
```suggestion
          // North clone cleaving inside west (and therefore west clone cleaving north).
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+        // are cleaving in order to make a '+' where the ends are each cleaved by one
+        // clone and the middle square is safe
+        let safeZone = null;
+        let adjacentZones = {};
+        if ((northCombatantFacing === dirNum.north && bladeSides[northCombatantBlade]) ||
+          (northCombatantFacing === dirNum.south && !bladeSides[northCombatantBlade])) {
+          // North clone cleaving inside east
+          safeZone = output.southwest();
+          adjacentZones = {
+            0: bladeValues[eastCombatantBlade],
```
quisquous, line 815: `0` here could also be `[dirNum.north]` for readability, here and elsewhere.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          safeZone = output.southeast();
+          adjacentZones = {
+            0: bladeValues[westCombatantBlade],
+            1: bladeValues[eastCombatantBlade],
+            2: bladeValues[southCombatantBlade],
+            3: bladeValues[northCombatantBlade],
+          };
+        } else if ((southCombatantFacing === dirNum.south && bladeSides[southCombatantBlade]) ||
+          (southCombatantFacing === dirNum.north && !bladeSides[southCombatantBlade])) {
+          // South clone cleaving inside west
```
quisquous, line 832:
```suggestion
          // South clone cleaving inside west (and therefore west clone cleaving south).
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          safeZone = output.northeast();
+          adjacentZones = {
+            0: bladeValues[northCombatantBlade],
+            1: bladeValues[eastCombatantBlade],
+            2: bladeValues[westCombatantBlade],
+            3: bladeValues[southCombatantBlade],
+          };
+        } else if ((southCombatantFacing === dirNum.north && bladeSides[southCombatantBlade]) ||
+          (southCombatantFacing === dirNum.south && !bladeSides[southCombatantBlade])) {
+          // South clone cleaving inside east
```
quisquous, line 842:
```suggestion
          // South clone cleaving inside east (and therefore east clone cleaving south).
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+        east: Outputs.east,
+        southeast: Outputs.southeast,
+        south: Outputs.south,
+        southwest: Outputs.southwest,
+        west: Outputs.west,
+        northwest: Outputs.northwest,
+      },
+    },
+    {
+      id: 'DelubrumSav Avowed Blade of Entropy Collect',
```
quisquous, line 924: Can you put the collect trigger before the normal trigger?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          adjacentZones = null;
+        }
+
+        let currentBrand = 0;
+        let currentTemperature = 0;
+
+        if (data.currentBrand)
+          currentBrand = data.currentBrand;
+        if (data.currentTemperature)
+          currentTemperature = data.currentTemperature;
```
quisquous, line 862:
```suggestion
        const currentBrand = data.currentBrand ? data.currentBrand : 0;
        const currentTemperature = data.currentTemperature ? data.currentTemperature : 0;
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          const dirs = {
+            0: output.north(),
+            1: output.east(),
+            2: output.south(),
+            3: output.west(),
+          };
+          adjacentZone = dirs[Object.values(calculatedZones).indexOf(calculatedZones.sort((a, b) =>
+            b - a).pop())];
+        } else {
+          adjacentZone = null;
```
quisquous, line 883:
```suggestion
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+        // Callout safe spot and get cleaved spot if both are known
+        // Callout safe spot only if no need to be cleaved
+        if (adjacentZone)
+          data.safeZone = output.getCleaved({ dir1: safeZone, dir2: adjacentZone });
+        else if (safeZone)
+          data.safeZone = output.safeSpot({ dir: safeZone });
+        else
+          data.safeZone = null;
+      },
+      infoText: (data, _, output) => {
```
quisquous, line 895:
```suggestion
      alertText: (data, _, output) => {
```

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+
+        // we need to filter for the Trinity Avowed with the lowest ID
+        // that one is always cleaving on one of the cardinals
+        // Trinity Avowed is always East (-267, -87)
+        const eastCombatant =
+          combatantDataBoss.combatants.sort((a, b) => b.ID - a.ID).pop();
+
+        // we need to filter for the three Avowed Avatars with the lowest IDs
+        // as they cast cleave at the different cardinals
+        const avatarOne = combatantDataAvatars.combatants.sort((a, b) => b.ID - a.ID).pop();
```
quisquous, line 745: Do you need to sort these? How many avatars get these buffs?  Is it just the three?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
```diff
@@ -596,6 +596,364 @@ export default {
...
+          console.error(`Avowed Avatar: null data`);
+          data.safeZone = null;
+          return;
+        }
+        if (!combatantDataAvatars.combatants) {
+          console.error(`Avowed Avatar: null combatants`);
+          data.safeZone = null;
+          return;
+        }
+        if (combatantDataAvatars.combatants.legnth < 3) {
```
quisquous, line 717:
```suggestion
        if (combatantDataAvatars.combatants.length < 3) {
```

Do you get more here?

quisquous (commented):
Sorry for a bunch of small comments.  If you're not running any more DRS and want me to apply these fixups myself, I can do that.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: The comma was there to intentionally place a pause as it sounded better pausing between the two clauses. Does => add a pause?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Just three avatars and the boss get the buff.

Unless the method to get the avatars is filtered differently, I think the sort is still required as logs indicate many more Avatars with different IDs in the encounter.

Legends0: May be possible to collect the info via a parseInt(matches.sourceId,16) during the buffs going out, adding enough delay there... but that gets to a more complicated race condition.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Currently, it should get more than 3, but if we change the trigger used to do the promise, we can get it so that only three (or 4 if boss included) are checked.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: It's a lot more complicated to code the collection off of the debuffs for matching to the adds than it is to sort, but I don't know the performance difference.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I defaulted to order in which the spells occurred in the encounter, but it does make it easier to read when the collect is before the trigger issuing the alert.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: oh that's how you do that as a key

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: Something like letting the trigger go off 4 times, each with its own promise and then a third trigger (or the last trigger to go off does a preRun to check if it is the one to execute the alert.

Legends0:
> 
> 
> Sorry for a bunch of small comments. If you're not running any more DRS and want me to apply these fixups myself, I can do that.

I don't mind if you can take over from here. I resolved most of the outstanding issues.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
Legends0: I forgot to mention, but regarding what I said about using sourceId, you need to be careful when handling this if regex triggers off the boss, the boss (randomly?) gets its buff refreshed, so I had switched to having the trigger just be based on the Avatars getting it.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
MaikoTan: `->` or `=>` inside sentence would be replaced as "then" in various languages, e.g. replaced to "puis"  for fr players.
So make it clear when display on screen as well as in TTS.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js
quisquous: Ah ok ok, thanks for all the details.  I think I am a little bit iffy on depending on actor ids when you don't have to, but I agree that it's probably more complicated to do other ways.  This is also kind of painful to test, so I think I'm inclined to just leave it as-is and get this landed.

PR #2576, opened 2021-03-13
Revert "Refactor: Typescript-ize overlay_plugin_api.js"

quisquous:
Reverts quisquous/cactbot#2548


quisquous:
@MaikoTan Sorry for reverting, but this seems to break everything.  There's a couple of errors (user_config.ts needs to refer to OverlayPluginApi.callOverlayHandler/addOverlayListener in a few places), but even with those fixed it seems like callbacks from callOverlayHandler never get delivered and so overlays never start.

MaikoTan:
Wh-what's wrong here...Σ(ﾟДﾟ)

Sorry I have not tested when apply the last several commits...(T_T)

quisquous:
No worries!  I am not sure what the error was.  I fixed the obvious thing (user_config.ts syntax errors), but couldn't figure out why overlays weren't working after that.  If you can't reproduce it after applying your patch and testing, let me know.

MaikoTan:
https://github.com/quisquous/cactbot/blob/2736eac0402428771db569bc5e89cdd742f9a447/resources/user_config.js#L275

Weird. I found that the callback function in `then` was not even been executed, but when I check out `callOverlayHandler` in `overlay_plugin_api.ts`, it returned a `Promise` as intended, so I have no idea why.

PR #2577, opened 2021-03-14
raidboss: DRS Fix Fateful words not calling out; Change duration => castTime

Legends0:
Some issues that I found during testing that I could resolve and test live.


Legends0:
I think the Coat of Arms is still off.

quisquous:
> I think the Coat of Arms is still off.

Ah sure, there's a couple more.  I'll fix them up.

PR #2580, opened 2021-03-14
raidboss: add suppressSeconds to DRS timeline triggers

quisquous:
This is an unfixed bug with timelines where sometimes
timeline triggers get double callouts if not suppressed.


quisquous:
@Legends0 This should fix the double callout issue you were seeing.

PR #2581, opened 2021-03-14
raidboss: fix DRS broken avowed timeline jump

quisquous:


quisquous:
@Legends0 I think this is the cause of the missing Glory of Bozja timeline trigger calls.

PR #2585, opened 2021-03-15
Jobs: WAR improvement: Better way to handle StormsEye duration

Echoring:
This makes StormsEye more accurate, even if some log lines are missing (like in Bozjan).


quisquous:
Oh, thank you! I had meant to file a bug about this being inconsistent.  Thank you!

FWIW, I was having inconsistent buff tracking in bozja for BLM, but the more recent changes to the ffxiv parser plugin to find effects in delubrum savage have fixed this for me.

PR #2586, opened 2021-03-15
Fix issue with party tracker's members never clearing

valarnin:
As reported by Elestriel on Discord, since https://github.com/quisquous/cactbot/commit/707a44d5124812fc051c70179b43b5a1f65ca77e the party members aren't being cleared properly in PartyTracker when receiving an event.

This commit moves the defaults to the property definitions, removes the spurious call to onPartyChanged from the constructor, and moves the reset logic to its own specific function in case an external piece of code needs to invoke it.

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -3,37 +3,21 @@ import { Job, Role } from '../types/job.js';
 import Util from './util';
 
 export default class PartyTracker {
-  details: Party[];
-  partyNames_: string[];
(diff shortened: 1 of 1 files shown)


valarnin:
Apparently my eslint plugin in VScode isn't working all of a sudden, sorry if there are a few amends as I fix the eslint issues.

quisquous:
Thank you!!!

PR #2587, opened 2021-03-16
raidboss: e12s chiseled sculpture triggers (alternative)

quisquous:
We were running a bunch of e12s and so I had time to test this out
and iron it out.

I personally found that having a trigger on:
* initial number
* left right
* laser bait
* other three lasers
* knockback location

...was just incredibly noisy.

I have cut this back from #2443 to say:
* initial number (inner outer)
* left/right OR left/right + bait #1
* bait #n laser (no other lasers called out)
* knockback location

Inner/Outer seems to work with the changes that I made here to collect
the sculptures separately from the tether.

(@valarnin, sorry to implement a patch that you were in the middle
of, but I wanted to get this in before I didn't have a chance to
test it anymore.)

Closes #2443.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e12s.js
+++ b/ui/raidboss/data/05-shb/raid/e12s.js
@@ -400,43 +400,189 @@ export default {
       id: 'E12S Promise Chiseled Sculpture',
       netRegex: NetRegexes.headMarker({}),
       condition: (data, matches) => data.isDoorBoss && matches.target === data.me,
-      response: (data, matches, output) => {
-        // cactbot-builtin-response
-        output.responseOutputStrings = {
-          // This is sort of redundant, but if folks want to put "square" or something in the text,
-          // having these be separate would allow them to configure them separately.
-          square1: numberOutputStrings[1],
-          square2: numberOutputStrings[2],
-          square3: numberOutputStrings[3],
-          square4: numberOutputStrings[4],
-          triangle1: numberOutputStrings[1],
-          triangle2: numberOutputStrings[2],
-          triangle3: numberOutputStrings[3],
-          triangle4: numberOutputStrings[4],
-        };
-
+      run: (data, matches) => {
         const id = getHeadmarkerId(data, matches);
 
         // Statue laser mechanic.
         const firstLaserMarker = '0091';
         const lastLaserMarker = '0098';
         if (id >= firstLaserMarker && id <= lastLaserMarker) {
-          // We could arguably tell you which giant you're tethered to by finding their position?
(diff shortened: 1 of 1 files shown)


quisquous:
https://clips.twitch.tv/HonestDistinctGooseMingLee-NZRMW8o0n-jzQkHn

valarnin:
Ultimately my goal was to get the triggers into cactbot in general, I'm glad that my original pass at it was sufficient to get there, even if in a modified form.

I tend to err on the side of more callouts with the understanding that it's easier to pare them down than to add additional later. I agree that it was potentially a bit excessive in this case.

In the end this is one less trigger I'll have to maintain for my static in the future, and one more trigger that everyone can benefit from.

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -400,43 +400,189 @@ export default {
...
+    {
+      id: 'E12S Promise Chiseled Sculpture Collector',
+      netRegex: NetRegexes.addedCombatantFull({ npcNameId: '9818' }),
+      run: (data, matches) => {
+        // Collect both sculptures up front, so when we find the tether on the
+        // current player we can look up both of them immediately.
+        data.statueIds = data.statueIds || [];
+        data.statueIds.push(parseInt(matches.id, 16));
+      },
+    },
```
valarnin, line 425: This is the only thing that jumps out at me as iffy. I'm not sure if the log lines for this happen at fade in or during the fight, if it's the latter then the trigger will fail if it's started after zone-in but before pull.

quisquous: These happen mid-fight:
```
[20:07:59.272] 03:4000900B:Added new combatant Chiseled Sculpture.  Job: N/A Level: 80 Max HP: 148000 Max MP: 10000 Pos: (0,-75,75) (98180000012445).
[20:07:59.272] 03:4000900A:Added new combatant Chiseled Sculpture.  Job: N/A Level: 80 Max HP: 148000 Max MP: 10000 Pos: (0,-75,75) (98180000012445).
[20:08:04.397] 16:4000900B:Chiseled Sculpture:58B2:Classical Sculpture:etc
```

valarnin: Alright. I wasn't at my computer to verify this at the time.

quisquous: Thanks for the question.  There's been a lot of things added early and then moved later this tier, so it's good to check.  :)

PR #2592, opened 2021-03-16
Fix web worker for raidemulator loading over file:// URIs

### LINKED ISSUE #2564: raidemulator should display an error when loaded over file://
This no longer works because of the web workers, and so it it'd be nice to display some error for users rather than silently not working.

valarnin:
This PR fixes #2564 though the fix isn't that great. I added the `worker-loader` loader to npm/webpack and used that to load the worker, which resolves loading over file URI by automatically switching to a blob, but it makes the actual worker import code a bit ugly. Copy/pasting the comment for reference:

```js
// @TODO: For some reason this absolutely has to be imported as `Worker`, no matter what settings
// are specified in webpack.config.cjs
// Using `{ Worker as NetworkLogConverterWorker }` also fails for some reason.
// Using `{}` works for some reason too.
// eslint-disable-next-line import/default
import Worker from './emulator/data/NetworkLogConverterWorker';
```

The `Worker` portion of the import line also shows as an error in vscode and eslint without the ignore rule.

@quisquous I'm pretty iffy on this one, it feels like there should be a better way than this, but I can't see it. Maybe this'll be better once raidemulator is moved over to typescript (if that's in the plans in general).

### DIFF
--- a/ui/raidboss/raidemulator.js
+++ b/ui/raidboss/raidemulator.js
@@ -18,6 +18,9 @@ import { TimelineLoader } from './timeline';
 import Tooltip from './emulator/ui/Tooltip';
 import UserConfig from '../../resources/user_config';
 import raidbossFileData from './data/manifest.txt';
+// eslint can't detect the custom loader for the worker
+// eslint-disable-next-line import/default
+import NetworkLogConverterWorker from './emulator/data/NetworkLogConverterWorker';
 
 // @TODO: Some way to not have this be a global?
 
@@ -65,7 +68,7 @@ const Options = {
     emulatedPartyInfo = new EmulatedPartyInfo(emulator);
     emulatedMap = new EmulatedMap(emulator);
     emulatedWebSocket = new RaidEmulatorOverlayApiHook(emulator);
-    logConverterWorker = new Worker('../../dist/raidemulatorWorker.bundle.js');
+    logConverterWorker = new NetworkLogConverterWorker();
 
     // Listen for the user to click a player in the party list on the right
     // and persist that over to the emulator
--- a/package-lock.json
+++ b/package-lock.json
@@ -9248,6 +9248,52 @@
       "integrity": "sha512-Hz/mrNwitNRh/HUAtM/VT/5VH+ygD6DV7mYKZAtHOrbs8U7lvPS6xf7EJKMF0uW1KJCl0H701g3ZGus+muE5vQ==",
       "dev": true
     },
+    "worker-loader": {
+      "version": "3.0.8",
+      "resolved": "https://registry.npmjs.org/worker-loader/-/worker-loader-3.0.8.tgz",
+      "integrity": "sha512-XQyQkIFeRVC7f7uRhFdNMe/iJOdO6zxAaR3EWbDp45v3mDhrTi+++oswKNxShUNjPC/1xUp5DB29YKLhFo129g==",
+      "dev": true,
+      "requires": {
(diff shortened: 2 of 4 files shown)


ui/raidboss/raidemulator.js
```diff
@@ -18,6 +18,12 @@ import { TimelineLoader } from './timeline';
 import Tooltip from './emulator/ui/Tooltip';
 import UserConfig from '../../resources/user_config';
 import raidbossFileData from './data/manifest.txt';
+// @TODO: For some reason this absolutely has to be imported as `Worker`, no matter what settings
```
quisquous, line 21: Maybe this, if you want to name the constructor?
```
        {
          // Worker has to go before normal js
          test: /NetworkLogConverterWorker\.(?:c|m)?js$/,
          loader: 'worker-loader',
          options: {
            esModule: true,
            inline: 'fallback',
            worker: {
              type: 'SharedWorker',
              options: {
                type: 'classic',
                name: 'NetworkLogConverterWorker',
              },
            },
          },
          resolve: {
            fullySpecified: false,
          },
        },
```

ui/raidboss/raidemulator.js
```diff
@@ -18,6 +18,12 @@ import { TimelineLoader } from './timeline';
...
+// are specified in webpack.config.cjs
+// Using `{ Worker as NetworkLogConverterWorker }` also fails for some reason.
+// Using `{}` works for some reason too.
+// eslint-disable-next-line import/default
```
quisquous, line 25: Would https://www.npmjs.com/package/eslint-import-resolver-webpack fix this?

quisquous (commented):
Thanks for looking into this!

ui/raidboss/raidemulator.js
valarnin: A slightly modified version of this (`Worker` instead of `SharedWorker`) works. Fix pushed.

ui/raidboss/raidemulator.js
valarnin: I tried various options for this and wasn't able to get at least vscode's eslint plugin to recognize the loader. I didn't try running `npm lint` to see if that would pick it up properly since it's about time for me to go to sleep here.

quisquous: If `npm lint` works, then it should work.  I find that vscode lags a little bit in picking up eslint changes, but will eventually get it.

valarnin: Installed package with:
```
npm install eslint-import-resolver-webpack --save-dev
```

Added this config to `.eslintrc.cjs`:
```diff
  'settings': {
    'import/resolver': {
      'node': {
        'extensions': ['.d.ts', '.ts', '.js'],
      },
      'typescript': {
        'alwaysTryTypes': true,
        'project': './tsconfig.json',
      },
+      'webpack': {
+        'config': './webpack/webpack.config.cjs',
+      },
    },
  },
```

```
npm run lint

> cactbot@ lint C:\Users\valar\Source\cactbot2
> eslint **/*.{js,cjs,mjs,ts}


C:\Users\valar\Source\cactbot2\ui\raidboss\raidemulator.js
  21:8  error  No default export found in imported module "./emulator/data/NetworkLogConverterWorker"  import/default
```

quisquous: Yeah, I can't get this to work either.  Maybe it doesn't understand `webpack-loader`.  I think we could write a custom eslint import resolver here, but that's probably overkill for a single line.

PR #2593, opened 2021-03-17
raidboss: e12s oracle cataclysm callout

Legends0:
Call out opposite direction of cataclysm, a bit more advanced than "getBehind".

I had tested getBehind early on in the tier, but never made a PR with it as ideally a cardinal would have been nicer to have. I will test this out later this week, I am unsure if delaySeconds 0.5 is enough.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e12s.js
+++ b/ui/raidboss/data/05-shb/raid/e12s.js
@@ -989,8 +989,8 @@ export default {
         // select the Oracle Of Darkness with same source id
         let oracleData = null;
         oracleData = await window.callOverlayHandler({
-            call: 'getCombatants',
-            ids: [parseInt(matches.sourceId, 16)],
+          call: 'getCombatants',
+          ids: [parseInt(matches.sourceId, 16)],
         });
 
         // if we could not retrieve combatant data, the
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -978,6 +978,81 @@ export default {
...
+          data.safeZone = null;
+          return;
+        }
+        if (!combatantData.combatants) {
+          console.error(`Oracle Of Darkness: null combatants`);
+          data.safeZone = null;
+          return;
+        }
+
+        // we need to filter for the Oracle Of Darkness with the highest ID
```
quisquous, line 1021: Is it the same id as the one who is doing this cast such that you could use `matches.sourceId`?

valarnin: Looks like that's the case to me at least, though I'd need to test it to be sure:
```
20|2021-02-13T22:37:29.2990000-06:00|4001ED5E|Oracle Of Darkness|58C2|Cataclysm|E0000000||5.70||e1c6cf0d8bbd16afba599ca69c645583
```

As such the code above could probably be:
```js
        const combatantData = await window.callOverlayHandler({
          call: 'getCombatants',
          ids: [matches.sourceId],
        });
```

Legends0: I didn't realize it, but yea I can use ids like so to get the correct combatant. Thanks.

quisquous:
Also, the original patch worked great for me tonight when I tested it.

PR #2594, opened 2021-03-17
eureka: Korean regex capture group fix

Jaehyuk-Lee:


Jaehyuk-Lee:
plus, additional translation

PR #2595, opened 2021-03-17
raidboss: delubrum_reginae fr

VanSlanzar:
- adds missing translations fr
- deletion of `Knockback` de for added in `common` and ja already exists in `common.`


ui/raidboss/common_replacement.js
```diff
@@ -119,6 +119,7 @@ export const commonReplacement = {
       ko: '--점프--',
     },
     '--knockback--': {
+      de: '--Rückstoß--',
```
MaikoTan, line 122: Maybe an odd question, but here to capitalize first-letter is right?

VanSlanzar: The last time, I wanted to remove a capital letter from a translation, I was told to leave it as it was written. So I just did a copy and paste.

MaikoTan: Ah, I see. Thanks!

Issue #2541, opened 2021-03-06
TypeError when executing find_missing_translations.py


Akurosia:
node version: v14.15.5
npm version: 6.14.11
python version: Python 3.9.1 (happens also with 3.7.x and 3.8.x)

Steps to reproduce:
1. remove existing node_modules
2. reinstall node_modules by running `npm i`
```
PS F:\Github\cactbot_aku2> npm i

> husky@4.2.5 install F:\Github\cactbot_aku2\node_modules\husky
> node husky install

husky > Setting up git hooks
husky > Done

> husky@4.2.5 postinstall F:\Github\cactbot_aku2\node_modules\husky
> opencollective-postinstall || exit 0

Thank you for using husky!
If you rely on this package, please consider supporting our open collective:
> https://opencollective.com/husky/donate

npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.13 (node_modules\webpack-dev-server\node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.13: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@2.3.2 (node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@2.3.2: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})

added 1051 packages from 508 contributors and audited 1056 packages in 44.279s

126 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
```
3. reinstall python requirements by using `python -m pip install -r requirements.txt`

```
PS F:\Github\cactbot_aku2> python -m pip install -r requirements.txt
Requirement already satisfied: black==19.10b0 in c:\users\akurosia\appdata\roaming\python\python39\site-packages (from -r requirements.txt (line 2)) (19.10b0)
Requirement already satisfied: pylint==2.5.3 in c:\users\akurosia\appdata\roaming\python\python39\site-packages (from -r requirements.txt (line 3)) (2.5.3)
Requirement already satisfied: PyYAML in c:\users\akurosia\appdata\roaming\python\python39\site-packages (from -r requirements.txt (line 6)) (5.3.1)
[...]

PR #2603, opened 2021-03-19
util: Convert find_missing_translations to JavaScript

panicstevenson:
Convert find_missing_translations from Python to JavaScript.

Example usage:
```sh
node --loader ts-node/esm ./util/find_missing_translations.js --locale fr
```
or
```sh
npm run find-translations -- --locale=fr
```

Fixes https://github.com/quisquous/cactbot/issues/2541

### DIFF
--- a/util/find_missing_timeline_translations.js
+++ b/util/find_missing_timeline_translations.js
@@ -4,16 +4,19 @@ import Regexes from '../resources/regexes';
 import { Timeline } from '../ui/raidboss/timeline';
 import { commonReplacement, partialCommonReplacementKeys } from '../ui/raidboss/common_replacement';
 
-const triggersFile = process.argv[2];
-const locale = process.argv[3];
-const localeReg = 'regex' + locale[0].toUpperCase() + locale[1];
-findMissing(triggersFile, locale, localeReg);
+let kTriggersFile;
+let kLocale;
+let kLocaleReg;
 
-async function findMissing(triggersFile, locale, localeReg) {
+export async function findMissing(triggersFile, locale, localeReg) {
+  kTriggersFile = triggersFile;
+  kLocale = locale;
+  kLocaleReg = localeReg;
+
   // Hackily assume that any file with a txt file of the same name is a trigger/timeline.
   const timelineFile = triggersFile.replace(/\.js$/, '.txt');
   if (!fs.existsSync(timelineFile))
-    process.exit(-1);
+    return;
 
   const timelineText = String(fs.readFileSync(timelineFile));
   const timeline = new Timeline(timelineText);
@@ -40,7 +43,7 @@ async function findMissing(triggersFile, locale, localeReg) {
     break;
   }
 
-  findMissingRegex(triggerSet.triggers, triggerLines, timeline, trans, localeReg);
+  findMissingRegex(triggerSet.triggers, triggerLines, timeline, trans);
   findMissingTimeline(timelineFile, triggerSet, timeline, trans);
 }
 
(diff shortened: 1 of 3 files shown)


util/find_missing_timeline_translations.js
```diff
@@ -4,16 +4,19 @@ import Regexes from '../resources/regexes';
 import { Timeline } from '../ui/raidboss/timeline';
 import { commonReplacement, partialCommonReplacementKeys } from '../ui/raidboss/common_replacement';
 
-const triggersFile = process.argv[2];
-const locale = process.argv[3];
-const localeReg = 'regex' + locale[0].toUpperCase() + locale[1];
-findMissing(triggersFile, locale, localeReg);
+let kTriggersFile;
```
quisquous, line 7: Sorry for the bikeshedding, but I'm going to say maybe don't introduce more k-prefixed constants.  I originally intended `k` prefix things as constants, in a Typescript `readonly` sense, like for arrays of data that would never change.  I think global variables set with `let` are not good candidates for this, and probably we should remove them all because it's confusing to everyone but me.

Also, can you make these not global? This is a race condition here because `findMissing` is an async function and sets these values.  I could see:

findMissing1: call sets values to 1, waits
findMissing2: call sets values to 2, waits
findMissing1: import finishes, global values wrong now??

I think it would be clearer if they were explicit too, but that's my fault for writing it like this.

util/find_missing_translations.js
```diff
@@ -0,0 +1,147 @@
...
+  help: 'Limits the results to only match specific files/path',
+});
+
+// Directory names to ignore when looking for JavaScript files.
+const ignoreDirs = [
+  '.git',
+  'publish',
+  'ThirdParty',
+  'node_modules',
+];
```
quisquous, line 29:
```suggestion
const ignoreDirs = [
  '.git',
  'dist',
  'node_modules',
  'publish',
  'ThirdParty',
  'user',
];
```

quisquous (commented):
`node --loader ts-node/esm ./util/find_missing_translations.js` is a bit of a mouthful.  What do you think about an `npm run find-translations -- --locale=fr` style thing instead?

util/find_missing_timeline_translations.js
panicstevenson: Only reason I went with this approach was because if I pass these via the functions they will exceed 100 characters. I have no problem changing it, but I was just trying to (in my mind) best follow what was there.

quisquous: Could put them in some sort of shorter named object too, if that helps? There sure are a lot of parameters.

panicstevenson: I opted for doing less touching and just moving the localeReg closer to where it's being used, which puts us just under the 100 character limit.

MaikoTan:
> `node --loader ts-node/esm ./util/find_missing_translations.js` is a bit of a mouthful.  What do you think about an `npm run find-translations -- --locale=fr` style thing instead?

I think it is better to create an entry js file, and then link all the scripts into this entry file.
for example, 
```bash
node util/index.js find-translations --locale=fr
node util/index.js generate effect
node util/index.js generate hunt
# ...etc

# or as a npm script?
npm run util -- find-translations --locale=fr
```

With some modern CLI tools framework (e.g. commander/vorpal), we can even create an interactive CLI as well.

So to avoid a long npm script list.

quisquous:
> I think it is better to create an entry js file, and then link all the scripts into this entry file as its subcommand.

I think a script like this would be a great thing to follow up and add, especially for all of the resource generation commands.

I think I'd like to keep the find-translations as a top level option for now, mostly because I think that will be used by many translators.  If package.json starts getting too crowded, maybe we can reconsider?  I think there's a bunch now, but it's not too bad yet?

PR #2606, opened 2021-03-19
raidemulator - Implement timelineTriggers support

### LINKED ISSUE #2504: bug： timeline Trigger can't modify data in raid emulator
## Description

I'm testing a set of triggers, and I found that timeline trigger can't modify data in any way. And `run`, `preRun` method are omitted. If I modify data in funxtion like `infoText` the modifition will be omitted.

## Additional information

log: 
[pure.log](https://github.com/quisquous/cactbot/files/6023557/pure.log)

trigger:
```JS
Options.Triggers.push({
  zoneId: ZoneId.TheEpicOfAlexanderUltimate,
  timelineTriggers: [
    {
      id: 'TEA 子言 时停阶段',
      regex: /Propeller Wind/,
      // delaySeconds: 20,
      infoText: (data, _, output) => {
        data.userPhase = 'Temporal Stasis';
// modify data here and it doesn't works,
// weather in `run(data)` or `preRun(data)` in later trigger infomation
// There's no such property or any other modification I made here
        console.log(data.userPhase); // output 'Temporal Stasis'
        return '提前进入时停阶段，反正中间也没啥东西'; // but this works fine and output correctly
      },
    },
  ],
});

```

valarnin:
This PR fixes #2504 by supporting `timelineTriggers` during encounter analysis.

This is a first pass. It's functional but causes a lot of DOM updates (due to the timeline events actually firing during analysis) which makes the initial encounter load slightly slower.

I'm not sure if it's worth the effort of fixing this, I can barely tell a difference on my hardware so I'd prefer if someone with slower hardware can test and let me know how much the performance impact is vs the current `main` code.

### DIFF
--- a/ui/raidboss/emulator/data/AnalyzedEncounter.js
+++ b/ui/raidboss/emulator/data/AnalyzedEncounter.js
@@ -52,6 +52,7 @@ export default class AnalyzedEncounter extends EventBus {
   }
 
   async analyzeFor(ID) {
+    let currentLogIndex = 0;
     const partyMember = this.encounter.combatantTracker.combatants[ID];
 
     if (!partyMember.job) {
@@ -65,8 +66,10 @@ export default class AnalyzedEncounter extends EventBus {
     const timelineUI = new RaidEmulatorTimelineUI(this.options);
     const timelineController =
         new RaidEmulatorTimelineController(this.options, timelineUI, raidbossFileData);
+    timelineController.bindTo(this.emulator);
     const popupText = new PopupTextAnalysis(
         this.popupText.options, new TimelineLoader(timelineController), raidbossFileData);
+    timelineUI.popupText = popupText;
 
     timelineController.SetPopupTextInterface(new PopupTextGenerator(popupText));
 
@@ -91,6 +94,32 @@ export default class AnalyzedEncounter extends EventBus {
       zoneID: parseInt(this.encounter.encounterZoneId, 16),
     });
 
+    timelineController.activeTimeline.SetTrigger(async (trigger, matches) => {
+      // Some async magic here, force waiting for the entirety of
+      // the trigger execution before continuing
+      const delayPromise = new Promise((res) => {
+        popupText.delayResolver = res;
+      });
+      const promisePromise = new Promise((res) => {
+        popupText.promiseResolver = res;
+      });
+      const runPromise = new Promise((res) => {
+        popupText.runResolver = res;
+      });
+
+      popupText.OnTrigger(trigger, matches);
+
+      await delayPromise;
+      await promisePromise;
+      const triggerHelper = await runPromise;
+
+      popupText.currentTriggerStatus.finalData = EmulatorCommon.cloneData(popupText.data);
+
+      popupText.callback(
+          this.encounter.logLines[currentLogIndex],
+          triggerHelper, popupText.currentTriggerStatus);
+    });
+
     popupText.callback = (log, triggerHelper, currentTriggerStatus, finalData) => {
       this.perspectives[ID].triggers.push({
         triggerHelper: triggerHelper,
@@ -107,7 +136,8 @@ export default class AnalyzedEncounter extends EventBus {
       finalData: popupText.data,
     };
 
-    for (const log of this.encounter.logLines) {
+    for (; currentLogIndex < this.encounter.logLines.length; ++currentLogIndex) {
+      const log = this.encounter.logLines[currentLogIndex];
       await this.dispatch('analyzeLine', log);
 
       if (this.encounter.combatantTracker.combatants[ID].hasState(log.timestamp)) {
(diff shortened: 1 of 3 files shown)


quisquous:
Sorry, I haven't forgotten about this, but just haven't had enough time to take a closer look at it.

valarnin:
> Sorry, I haven't forgotten about this, but just haven't had enough time to take a closer look at it.

No problem, this isn't holding up the rest of the items in the "needs improvement" issue, I injured myself last Saturday and that's put a damper on my ability to work on this for the next week or two.

ui/raidboss/emulator/overrides/RaidEmulatorTimelineController.js
```diff
@@ -43,7 +43,10 @@ export default class RaidEmulatorTimelineController extends TimelineController {
     e.detail.logs.forEach((line) => {
       this.activeTimeline.emulatedTimeOffset = line.offset;
       this.ui.emulatedTimeOffset = line.offset;
+      this.activeTimeline.timebase =
+        new Date() - (line.offset - this.emulator.currentEncounter.encounter.initialOffset);
```
quisquous, line 47: Maybe this is a continuation of the suppressSeconds conversation from the other patch, but it is weird to me to ever see `new Date()` in any emulator code.  My expectation is that time should be some "virtual time" entirely controlled by the emulator.

It looks like `new Date()` is used in `SyncTo` and `_OnUpdateTimer()`.

What if:
* `activetimeline.OnLogLine` also took a current time parameter (the `SyncTo` case)
* `_OnUpdateTimer` also took a current time parameter
* update `timeline.js` to pass `new Date()` to the callers of these two functions
* Pass the log line time to `OnLogLine`, which will set the timebase for you
* Call `_OnUpdateTimer` with the current log line time as well

quisquous (commented):
Performance-wise this seems fine to me.  Thanks for implementing it.

I had a couple of code questions (sorry):

ui/raidboss/emulator/data/AnalyzedEncounter.js
```diff
@@ -78,6 +81,32 @@ export default class AnalyzedEncounter extends EventBus {
...
+      popupText.OnTrigger(trigger, matches);
+
+      await delayPromise;
+      await promisePromise;
+      const triggerHelper = await runPromise;
+
+      popupText.currentTriggerStatus.finalData = EmulatorCommon.cloneData(popupText.data);
+
+      popupText.callback(
+          this.encounter.logLines[currentLogIndex],
```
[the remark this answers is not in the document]
valarnin: The callback definition immediately below it uses `log.timestamp` to calculate when the trigger was resolved in the pull.

```js
    popupText.callback = (log, triggerHelper, currentTriggerStatus, finalData) => {
      this.perspectives[ID].triggers.push({
        triggerHelper: triggerHelper,
        status: currentTriggerStatus,
        logLine: log,
        resolvedOffset: (log.timestamp - this.encounter.startTimestamp) +
          (currentTriggerStatus.delay * 1000),
      });
    };
```

In addition, `EmulatedPartyInfo.js`'s uses the timestamp of the log line for calculations.

ui/raidboss/emulator/overrides/RaidEmulatorTimelineController.js
valarnin: Any time there's a `new Date` call that's not isolated, I have to account for it with extra logic like this in emulator. That's why I was suggesting that calls to `new Date` in raidboss be replaced with calls to a class function `getDate` or something that could be overridden in emulator as a global "return the current log line's timestamp" logic.

quisquous: I agree that `new Date()` needs to be abstracted.  In general, I prefer passing in the time explicitly to functions vs the implicit overriding of a functions.  It means that internal functions can't call `getDate` repeatedly (which is usually an anti-pattern for a processing function) and also makes it more clear that the date is being controlled explicitly externally and may not behave monotically.

valarnin: I think this would belong in a separate PR, so the question becomes should this PR wait until that change, or should this be merged as-is and the date stuff be a follow up PR?

quisquous: Oops, somehow I lost track of responding to this comment.  Thanks for submitting the other PR, you read my mind.  :)

valarnin:
@quisquous I have updated this PR to take into account the changes to the way date is passed. Please re-review. I'd like to get started on converting emulator to typescript as soon as I get the remaining two outstanding issues resolved, which I'll do in my next PR after this one is merged.

quisquous:
Thanks for fixing up the time issues.  This makes this patch feel much better to me.

PR #2613, opened 2021-03-21
Refactor: TypeScript-ize timerbar.js

MaikoTan:


quisquous:
styleFill does not appear to be working properly.  With this patch, the raidboss timers do not fill from left to right, and instead go right to left.

Also, I think given that this is .ts, there is no way to run the timerbar_test.  Do we need to have a webpack entry for it?

resources/timerbar.ts
```diff
@@ -1,91 +1,135 @@
...
+  private _leftText: string;
+  private _centerText: string;
+  private _rightText: string;
+  private _hideAfter: number;
+  private _loop: boolean;
+  private _connected: boolean;
+  private _hideTimer: number | null;
+  private _animationFrame: number | null;
+
+  static get observedAttributes(): string[] {
```
quisquous, line 25: I think at the very least, `styleFill` needs to be listed here.  That's not enough to fix the issue though, so something else needs to change as well.

MaikoTan:
@quisquous Sorry I have to take care of my grandfather in hospital these days, if you can commit a suggestion, I will be glad to merge it.

PR #2614, opened 2021-03-21
Refactor: TypeScript-ize timerbox.js

MaikoTan:


quisquous:
(timerbox?)

resources/timerbox.ts
```diff
@@ -1,111 +1,141 @@
...
+  _duration: number;
+  rootElement: HTMLDivElement;
+  largeBoxElement: HTMLDivElement;
+  largeBoxBackgroundElement: HTMLDivElement;
+  largeBoxForegroundElement: HTMLDivElement;
+  smallBoxElement: HTMLDivElement;
+  smallBoxBackgroundElement: HTMLDivElement;
+  smallBoxForegroundElement: HTMLDivElement;
+  timerElement: HTMLDivElement;
+  kLargeSize: number;
```
quisquous, line 12: Can these constants all be set here and be readonly (maybe static too?)

resources/timerbox.ts
```diff
@@ -1,111 +1,141 @@
 class TimerBox extends HTMLElement {
-  static get observedAttributes() {
+  _start: number;
```
quisquous, line 2: Can these all be private?

resources/timerbox.ts
```diff
@@ -114,23 +144,67 @@ class TimerBox extends HTMLElement {
...
+    if (this.bg !== null) this._bg = this.bg;
+    if (this.fg !== null) this._fg = this.fg;
+    if (this.scale !== null) this._scale = Math.max(parseFloat(this.scale), 0.01);
+    if (this.toward !== null) this._towardTop = this.toward !== 'bottom';
+    if (this.style !== null) this._styleFill = this.styleFill !== 'empty';
+    if (this.hideafter !== null && this.hideafter !== '') this._hideAfter = Math.max(parseFloat(this.hideafter), 0);
   }
 
-  init(root) {
+  // // These would be used by document.registerElement, which is deprecated but
```
quisquous, line 193: Yeah, I think these comments are out of date.  CEF is new enough that we don't need `registerElement` anymore.

resources/timerbox.ts
```diff
@@ -1,111 +1,141 @@
...
   }
 
   // The displayed value is scaled by this.
-  set valuescale(s) {
-    this.setAttribute('valuescale', s);
+  set valuescale(s: string | null) {
+    this.setAttribute('valuescale', s ?? '');
   }
-  get valuescale() {
+  get valuescale(): string | null {
```
quisquous, line 66: Is it possible for any of these to return null?

MaikoTan: `this.getAttribute` returns `string | null`, or we can make a null-check here to ensure it returns `string` only?

quisquous: Sorry, to match #2649 (as per the discussion in https://github.com/quisquous/cactbot/pull/2532#discussion_r602517869), I think these should all be `string | null` after all, here and in the other resource PRs.

MaikoTan: Done.

resources/timerbox.ts
```diff
@@ -1,136 +1,226 @@
...
+  private _styleFill: boolean;
+  private _hideAfter: number;
+  private _bigAtZero: boolean;
+  private _roundUpThreshold: boolean;
+  private _connected: boolean;
+  private _hideTimer: number | null;
+  private _timer: number | null;
+  private _animationFrame: number | null;
+
+  static get observedAttributes(): string[] {
```
quisquous, line 32: I think this needs styleFill as well, although it seems like setting this to empty is working correctly, whereas it is not for timer bars.

PR #2617, opened 2021-03-21
refactor(TS): complete type Option

Trim21:


types/global.d.ts
```diff
@@ -56,3 +56,14 @@ export interface HandlerMap {
 }
 
 export type HandlerType = keyof HandlerMap;
+
+export interface Option {
```
quisquous, line 60: I'm a little surprised at taking out `[key: string]: unknown;` from this type.  See the extra fields in `ui/raidboss/raidboss_options.js`.  `raidboss_config.js` also sets a lot of options as well.

Is this intended to be `BaseOptions`, e.g. the interface that something like `RaidbossOptions` would derive from?

PR #2622, opened 2021-03-23
Fix: LocaleNetRegex should use `this.netRegexes`

MaikoTan:
Sorry that I've not noticed at the last time.


quisquous:
Oops, good catch!

PR #2624, opened 2021-03-23
refactor: define class in es6 style

Trim21:


quisquous:
`ui/dps/xephero/manual_dps_test.html` does not appear to be working at the moment.  It probably broke a while ago and then never got fixed.  I think we can just remove `manual_dps_test.html` and `fake_dps.js` and if I ever have time, I can fix it up.

PR #2632, opened 2021-03-26
Revert "Fix(resourcebar): getter and setter accept string"

quisquous:
Reverts quisquous/cactbot#2623

Sorry, this causes the mp bar to say `NaN` in the jobs overlay.  I haven't looked into why this is happening, but just wanted to roll back to a working version.

### DIFF
--- a/resources/resourcebar.ts
+++ b/resources/resourcebar.ts
@@ -31,93 +31,93 @@ export default class ResourceBar extends HTMLElement {
   }
 
   // All visual dimensions are scaled by this.
-  set scale(s: string) {
-    this.setAttribute('scale', s);
+  set scale(s: string | null) {
+    this.setAttribute('scale', s ?? '');
   }
-  get scale(): string {
-    return this.getAttribute('scale') ?? '';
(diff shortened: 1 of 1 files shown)


MaikoTan:
What Job were you using when this bug appeared?
I think `NaN` is usually due to the divisor being 0. I looked at the resourcebar code and there are only two lines dividing by `this._maxValue`, I think it might have been assigned to 0 in some places.

I'll try to look for it ehen I have more time.

quisquous:
I think the issue was that the null check / empty string that Panic put up a PR for in #2633.  However, after some discussion (see https://github.com/quisquous/cactbot/pull/2532#discussion_r602517869), I think Trim21's original idea is the better one here.  I didn't realize the getters were actually being used, and if they are, then it's better to treat null as "unset" rather than use empty strings.

So, I think that issue is fixed, but now the other typescript patches should match what that one does, to support null and removeAttribute.  Sorry for all the change back and forth here.  This is just due to my own confusion and lack of understanding.

PR #2633, opened 2021-03-26
Update ResourceBar Getter and Setter Functions

panicstevenson:
Applying commit 6288af5 with fixes to prevent NaN error.

### DIFF
--- a/resources/resourcebar.ts
+++ b/resources/resourcebar.ts
@@ -31,93 +31,93 @@ export default class ResourceBar extends HTMLElement {
   }
 
   // All visual dimensions are scaled by this.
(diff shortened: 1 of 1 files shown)


panicstevenson:
After an offline discussion with @quisquous, I think the intention of this code is to preserve the existing functionality of the getter/setter functions (`setAttribute` maybe a little less, but definitely keep the `getAttribute` functionality, see: [this comment](https://github.com/quisquous/cactbot/pull/2532#discussion_r599167793)). With that, I think it makes the most sense to keep the code as-is, and potentially address the `setAttribute` section separately if necessary.

PR #2634, opened 2021-03-26
raidboss: more dr normal triggers

quisquous:


quisquous:
Hmm, I'm still having a lot of bad luck getting the savage version to work, and that makes me extra skeptical that this normal mode is working.

quisquous:
> Hmm, I'm still having a lot of bad luck getting the savage version to work, and that makes me extra skeptical that this normal mode is working.

AHA, looks like I had two of the sword directions reversed.

Issue #2626, opened 2021-03-23
No timeline check with command `util/check_translation.py -f e12s.js -t de` and other lang. Just for e12s.


VanSlanzar:
**no problem, except for e12s.**
```
Bauer@DESKTOP-SO533H8 MINGW64 ~/Documents/cactbot (e12s-fr)
$ util/check_translation.py -f e12s.js -t de
### Eden 12 (Savage)
# Eden's Promise: Eternity (Savage)

hideall "--Reset--"
hideall "--sync--"

Traceback (most recent call last):
  File "C:\Users\Bauer\Documents\cactbot\util\check_translation.py", line 316, in <module>
    main(args)
  File "C:\Users\Bauer\Documents\cactbot\util\check_translation.py", line 273, in main
    print_timeline(args.timeline, timeline_file, trans, args.grep_missing)
  File "C:\Users\Bauer\Documents\cactbot\util\check_translation.py", line 231, in print_timeline
    filter_print_timeline(translate_timeline(line.strip(), trans[locale]), missing_filter)
  File "C:\Users\Bauer\Documents\cactbot\util\check_translation.py", line 191, in translate_timeline
    did_work = did_work or re.search(old, text, re.IGNORECASE)
  File "C:\Users\Bauer\AppData\Local\Programs\Python\Python39\lib\re.py", line 201, in search
    return _compile(pattern, flags).search(string)
  File "C:\Users\Bauer\AppData\Local\Programs\Python\Python39\lib\re.py", line 304, in _compile
    p = sre_compile.compile(pattern, flags)
  File "C:\Users\Bauer\AppData\Local\Programs\Python\Python39\lib\sre_compile.py", line 768, in compile
    code = _code(p, flags)
  File "C:\Users\Bauer\AppData\Local\Programs\Python\Python39\lib\sre_compile.py", line 607, in _code
    _compile(code, p.data, flags)
  File "C:\Users\Bauer\AppData\Local\Programs\Python\Python39\lib\sre_compile.py", line 182, in _compile
    raise error("look-behind requires fixed-width pattern")
re.error: look-behind requires fixed-width pattern
```

PR #2635, opened 2021-03-26
Fix Negative Lookback Groups

panicstevenson:
Python's regex only allows fixed-length strings for lookbacks; change
from variable-length lookback strings to instead be looking for spaces,
implying a multi-word string.

Fixes https://github.com/quisquous/cactbot/issues/2626


quisquous:
Thank you!

Issue #2608, opened 2021-03-20
`npm run find-translations` lists entries in zone_info


quisquous:
I haven't looked into it, but this seems wrong:

```
$ npm run find-translations -- --locale=de

C:\Users\enne\cactbot/resources/zone_info.js:1855 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:1871 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:1887 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:2057 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:2486 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:3494 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:3510 "'name': {"
C:\Users\enne\cactbot/resources/zone_info.js:3526 "'name': {"
C:\Users\enne\cactbot/ui/raidboss/data/05-shb/alliance/the_copied_factory.js:831 "'replaceText': {"

etc etc etc more real stuff follows from here
```

None of these seem like real errors.  I could imagine skipping files that say that they're autogenerated, but I wonder if this is indicative of some logic error that needs to be fixed too.

Akurosia:
i found another incorrect thing with the new find-translation script:

While doing translations for th elatest delubrum, it only showed me following missing translations for the savage file:
```
F:\Github\cactbot_aku2/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js:459 "clockwise: {"
F:\Github\cactbot_aku2/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js:462 "counterclockwise: {"
F:\Github\cactbot_aku2/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js:636 "knockbackNoFlare: {"
F:\Github\cactbot_aku2/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js:639 "knockbackWithFlare: {"
F:\Github\cactbot_aku2/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.js:1021 "text: {"
```
During the translation, i saw that one of the `text` was not listed in the script (see attached screenshot)
<img width="760" alt="sublime_text_Esjj9IWilb" src="https://user-images.githubusercontent.com/7465469/111955556-f2c5e280-8ae9-11eb-9522-f935de0047d8.png">

[...]

PR #2637, opened 2021-03-26
Fixup find_missing_translations.js

panicstevenson:
Remove auto-generated files from consideration, and fix a bug where
text blocks including variables were not considered.

Fixes https://github.com/quisquous/cactbot/issues/2608

### DIFF
--- a/util/find_missing_translations.js
+++ b/util/find_missing_translations.js
@@ -87,10 +87,16 @@ const parseJavascriptFile = (file, locales) => {
   let keys = [];
   let openMatch = null;
 
-  const openObjRe = new RegExp('(\\s*)(.*{)\\s*');
+  const openObjRe = new RegExp('(\\s*)(.*{)\\s*$');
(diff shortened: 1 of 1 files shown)


util/find_missing_translations.js
```diff
@@ -87,10 +87,16 @@ const parseJavascriptFile = (file, locales) => {
   let keys = [];
   let openMatch = null;
 
-  const openObjRe = new RegExp('(\\s*)(.*{)\\s*');
+  const openObjRe = new RegExp('(\\s*)(.*{)\\s*$');
```
quisquous, line 90: Sorry to be slow, but how does this fix the variable issue? I was expecting a change to `keyRe`?

panicstevenson: This was matching lines like `en: 'Hello ${name}'` based off the `{`, since `\\s*` can match 0 results as well.

quisquous: Oho! Thanks for the explanation.

PR #2642, opened 2021-03-28
Revert "refactor(resources): migrate WidgetList to TypeScript"

quisquous:
Reverts quisquous/cactbot#2532

@Trim21 I'm sorry to revert this, but this patch was causing all of my overlays to run with extreme lag.  Normally I wouldn't revert and would just fix, but I think this is a pretty bad bug.  My jobs overlay would print `BrowserConsole: Element with id child20 is missing? (Source: http://localhost:8080/dist/jobs.bundle.js, Line: 10214)` many times a second.  When I reverted this patch locally it stopped happening, so I feel pretty confident that this is the issue.

### DIFF
--- a/resources/widgetlist.js
+++ b/resources/widgetlist.js
@@ -1,40 +1,17 @@
 // NOTE
 // This class should be considered deprecated, and any users of this class should
 // just switch over to using CSS grid.
-type Sorter = () => number;
 
-const getRandomInt = function(max: number) {
-  return Math.floor(Math.random() * Math.floor(max));
-};
-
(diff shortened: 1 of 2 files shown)


Trim21:
Looks like there is some bug in `removeElement`， I'll try to fix it

Trim21:
look like I'll get a notification every this commit is pushed to a fork...

MaikoTan:
Click unsubscribe button?

Trim21:
> Click unsubscribe button?

Then I just unsubscribed from this commit in this forked repo.

PR #2653, opened 2021-04-01
raidemulator - Rework the way analyzed triggers are displayed

valarnin:
This PR reworks the way analyzed triggers are displayed in raidemulator to make it more obvious at a glance what various triggers do.

This PR also adds an additional checkbox to the trigger display to hide "collector" triggers, e.g. triggers that ran but produced no user-visible output.

Here's an example screenshot:

image

### DIFF
--- a/ui/raidboss/emulator/data/PopupTextAnalysis.js
+++ b/ui/raidboss/emulator/data/PopupTextAnalysis.js
@@ -167,7 +167,6 @@ export default class PopupTextAnalysis extends StubbedPopupText {
   _onTriggerInternalResponse(triggerHelper) {
     this.currentFunction = '_onTriggerInternalResponse';
     super._onTriggerInternalResponse(triggerHelper);
-    this.currentTriggerStatus.response = triggerHelper.response;
   }
 
   _onTriggerInternalAlarmText(triggerHelper) {
@@ -188,6 +187,11 @@ export default class PopupTextAnalysis extends StubbedPopupText {
   _onTriggerInternalTTS(triggerHelper) {
     this.currentFunction = '_onTriggerInternalTTS';
     super._onTriggerInternalTTS(triggerHelper);
+    if (triggerHelper.ttsText !== undefined &&
+      this.currentTriggerStatus.responseType === undefined) {
+      this.currentTriggerStatus.responseType = 'tts';
+      this.currentTriggerStatus.responseLabel = triggerHelper.ttsText;
+    }
   }
 
   _onTriggerInternalPlayAudio(triggerHelper) {
@@ -208,11 +212,28 @@ export default class PopupTextAnalysis extends StubbedPopupText {
   }
 
   _createTextFor(text, textType, lowerTextKey, duration) {
-    // No-op
+    // No-op for functionality, but store off this info for feedback
+    this.currentTriggerStatus.responseType = textType;
+    this.currentTriggerStatus.responseLabel = text;
   }
 
   _playAudioFile(url, volume) {
-    // No-op
+    // No-op for functionality, but store off this info for feedback
+
+    // If we already have text and this is a default alert sound, don't override that info
+    if (this.currentTriggerStatus.responseType) {
+      if (this.currentTriggerStatus.responseType === 'info' &&
+          url === this.options.InfoSound)
+        return;
+      if (this.currentTriggerStatus.responseType === 'alert' &&
+          url === this.options.AlertSound)
+        return;
+      if (this.currentTriggerStatus.responseType === 'alarm' &&
+          url === this.options.AlarmSound)
+        return;
+    }
+    this.currentTriggerStatus.responseType = 'audiofile';
+    this.currentTriggerStatus.responseLabel = url;
   }
 
   _scheduleRemoveText(container, e, delay) {
--- a/ui/raidboss/emulator/ui/EmulatedPartyInfo.js
+++ b/ui/raidboss/emulator/ui/EmulatedPartyInfo.js
@@ -10,7 +10,8 @@ export default class EmulatedPartyInfo extends EventBus {
     this.emulator = emulator;
     this.$partyInfo = document.querySelector('.partyInfoColumn .party');
     this.$triggerInfo = document.querySelector('.triggerInfoColumn');
-    this.$triggerHideCheckbox = document.querySelector('.triggerHideSkipped');
+    this.$triggerHideSkippedCheckbox = document.querySelector('.triggerHideSkipped');
+    this.$triggerHideCollectCheckbox = document.querySelector('.triggerHideCollector');
     this.$triggerBar = document.querySelector('.playerTriggers');
     this.triggerBars = [];
     this.latestDisplayedState = 0;
@@ -37,12 +38,17 @@ export default class EmulatedPartyInfo extends EventBus {
       this.latestDisplayedState = Math.max(this.latestDisplayedState, time);
     });
     this.updateTriggerState = () => {
-      if (this.$triggerHideCheckbox.checked)
+      if (this.$triggerHideSkippedCheckbox.checked)
         this.hideNonExecutedTriggers();
       else
         this.showNonExecutedTriggers();
+      if (this.$triggerHideCollectCheckbox.checked)
+        this.hideCollectorTriggers();
+      else
+        this.showCollectorTriggers();
     };
-    this.$triggerHideCheckbox.addEventListener('change', this.updateTriggerState);
+    this.$triggerHideSkippedCheckbox.addEventListener('change', this.updateTriggerState);
+    this.$triggerHideCollectCheckbox.addEventListener('change', this.updateTriggerState);
 
     this.$triggerItemTemplate = document.querySelector('template.triggerItem').content.firstElementChild;
     this.$playerInfoRowTemplate = document.querySelector('template.playerInfoRow').content.firstElementChild;
@@ -63,6 +69,19 @@ export default class EmulatedPartyInfo extends EventBus {
       n.classList.remove('d-none');
     });
   }
+
+  hideCollectorTriggers() {
+    this.$triggerInfo.querySelectorAll('.trigger-no-output').forEach((n) => {
+      n.classList.add('d-none');
+    });
+  }
+
+  showCollectorTriggers() {
+    this.$triggerInfo.querySelectorAll('.trigger-no-output').forEach((n) => {
+      n.classList.remove('d-none');
+    });
+  }
+
 
   /**
    * @param {RaidEmulator} emulator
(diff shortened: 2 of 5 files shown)


ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -208,11 +212,15 @@ export default class PopupTextAnalysis extends StubbedPopupText {
...
-    // No-op
+    // No-op for functionality, but store off this info for feedback
+    this.currentTriggerStatus.responseType = textType;
+    this.currentTriggerStatus.responseLabel = text;
   }
 
   _playAudioFile(url, volume) {
-    // No-op
+    // No-op for functionality, but store off this info for feedback
+    this.currentTriggerStatus.responseType = 'audiofile';
```
quisquous, line 222: From code inspection, I'm sort of surprised why this doesn't override the `_createTextFor` when there is the normal audio alert + text on screen.

quisquous (commented):
One CSS comment is that the mouseover highlight is a little tough to read:
image

Overall this looks SO good, though.  Thank you!  This will make the raid emulator a million times more useful for me.

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
```diff
@@ -286,10 +344,37 @@ export default class EmulatedPartyInfo extends EventBus {
         false);
   }
 
-  _wrapCollapse(label, $obj, onclick) {
+  _wrapCollapse(time, name, icon, text, classes, $obj, onclick) {
```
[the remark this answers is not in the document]
valarnin: I'm not sure how that would improve readability where the function is called, since the object would have to be constructed there anyways? I guess I'm just not seeing the benefit here since it's only called from three locations.

Maybe it would make more sense to make this a helper class that could chain function calls instead, such that calls became:
```js
let wrapped = 
    new WrapCollapse($initDataViewer)
    .setTime('00:00')
    .setName('Initial Data')
    .setClasses('data')
    .setCallback(() => {
      $initDataViewer.textContent = JSON.stringify(per.initialData, null, 2);
    }))
    .get();
```

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
```diff
@@ -219,31 +238,40 @@ export default class EmulatedPartyInfo extends EventBus {
...
+          [per.triggers[i].status.responseType],
+          $triggerDataViewer, () => {
+            $triggerDataViewer.textContent = JSON.stringify(per.triggers[i], null, 2);
+          });
       if (per.triggers[i].status.executed)
         $trigger.classList.add('trigger-executed');
       else
         $trigger.classList.add('trigger-not-executed');
 
+      if (triggerText === undefined || triggerText === null)
```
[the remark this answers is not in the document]
valarnin: This is a case of defensive programming due to raidemulator being in JS still. I can drop the check against undefined in this case but I think there are some edge cases where `PopupText`'s `valueOrFunction` can return `undefined`, or at least I recall running into a few cases where that happened with my own custom triggers.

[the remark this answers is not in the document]
quisquous: Oh yes, `valueOrFunction` can return undefined.  I think it will return any type but object or function.  But it all goes through `getTriggerLabelText`, so my thought was that function can fix anything up to be string or null, rather than making everything else deal with that.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: I'm not sure I understand this comment?

`_createTextFor` is overridden right above this to handle info/alert/alarm but checking for if there's an audio file queued to play requires access to the `triggerHelper` which isn't available in that function.

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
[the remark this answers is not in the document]
quisquous: Oh, the readability is just from sheer number of parameters, especially with multiple duplicate nulls.  I meant more:

```javascript
    $container.append(this._wrapCollapse({
        time: '00:00',
        name: 'Initial Data',
        icon: null, // optional??
        text: null,
        classes: ['data'],
        callback: $initDataViewer, () => {
          $initDataViewer.textContent = JSON.stringify(per.initialData, null, 2);
        },
    }));
```

It sounds like you disagree about readability though?

ui/raidboss/emulator/data/PopupTextAnalysis.js
quisquous: Sorry, this was just a drive-by "trying to understand the code without doing much debugging".

I guess my confusion was that even for an `alertText`, a sound file gets played via `_playAudioFile`, and so what prevents `this.currentTriggerStatus.responseType = 'audiofile';` from clobbering the value set in `this.currentTriggerStatus.responseType = textType;` from `_createTextFor`, as `_playAudioFile` runs after `_createTextFor`.

valarnin: Currently, nothing. There doesn't seem to be a good way to display both audio and text info within the confines of the button lines. I guess the icon could be the audio file icon, the border color could go to blue to indicate, but it could keep the text displayed?

Functionally this would be just removing the `this.currentTriggerStatus.responseLabel = url;` assignment.

quisquous: Sorry, maybe I'm only creating more confusion about my confusion.  What I'm trying to ask is, why does an `infoText` end up with `responseType === 'info'` and not `responseType === 'audiofile'` since `infoText` also plays a sound.  Your code appears to do what it should (the former) but I don't see from reading this why it's not doing the latter.

Sorry for the weird question, I was hoping you could clean up whatever misconception I was having so I didn't have to go do a bunch of debugging to understand why things work.

valarnin: Looks like this was a side-effect of https://github.com/quisquous/cactbot/commit/e0c67d03e98361ebb75400a3c128b987bbf351ae, `audioAllowed` was changed to `AudioAllowed` but not updated in `raidemulator.js` to reflect that change.

The `Options` global in `raidemulator.js` should probably be updated to just import from `raidboss_options.js`.

That will result in the expected behavior here (e.g. everything plays an audio alert so it all gets overwritten by the audio alert display), see screenshot below.

I think that in this case it doesn't really make sense to show the audio alert for default audio alerts? Functionally, in `_playAudioFile` check that the url matches the global `Options` entries and if so then don't overwrite the text display?

In addition, it probably doesn't make sense to overwrite the text for audio alerts, maybe it should just overwrite icon/color but leave the text the same (if there's text) since the audio file played can be pulled from details if needed.

image

quisquous: Aha!  Thank you for figuring it out.  Definitely agreed that text and icon should not take precedence if it is the default.  I think the previous behavior was good.

I'm not sure what the best way to do this functionally is.  I think somebody could write a trigger that plays the alert sound manually for a trigger but nothing else, so I'm a little bit skeptical of url checking.

Maybe `triggerHelper` needs to change the field it sets from `soundUrl` to `implicitSoundUrl` (or default? or auto? or some word?) and then you can figure out at the time of playing a sound whether or not the sound url is explicit or implicit.  You could have some `_playImplicitAudioFile` that the emulator could override to nothing?

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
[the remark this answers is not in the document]
valarnin: Switched to object for latest commit.

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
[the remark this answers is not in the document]
valarnin: Switched over to defaulting to undefined since it makes handling the object-as-parameter of wrapCollapse a bit more clean. Cleaned this up otherwise.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: Review the latest commit's logic in `PopupTextAnalysis.js`'s `_playAudioFile` function. Also swapped the `raidemulator.js` `Options` property to pull from `raidboss_options.js`.

valarnin:
Hover opacity updated so it's more legible:

image

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
```diff
@@ -286,19 +347,54 @@ export default class EmulatedPartyInfo extends EventBus {
...
   }
 
-  _wrapCollapse(label, $obj, onclick) {
+  _wrapCollapse(params = {
+    time: undefined,
+    name: undefined,
+    icon: undefined,
+    text: undefined,
+    classes: undefined,
+    $obj,
```
quisquous, line 356:
```suggestion
    $obj: undefined,
```

I am a little surprised that eslint `'object-shorthand': ['error', 'consistent'],` did not complain about this, but I think it should be consistent.

valarnin: That parameter is required which is why I didn't default it to undefined. I'm not sure how I would otherwise indicate as such without being able to use typescript.

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
```diff
@@ -286,19 +347,54 @@ export default class EmulatedPartyInfo extends EventBus {
...
+    onclick: undefined,
+  }) {
```
quisquous, line 358:
```suggestion
  _wrapCollapse(params = {
    // all parameters are optional except for $obj
    time: undefined,
    name: undefined,
    icon: undefined,
    text: undefined,
    classes: undefined,
    $obj: undefined, // required
    onclick: undefined,
  }) {
```

I think this is just documentation and the values and keys here are not used in practice.  So, maybe the answer is just more documentation?  I'm not sure I understood the nuance that $obj was required by initializing it with shorthand until you explained it to me.

valarnin: Hmm, given that it's the only 100% required parameter, maybe it should be a separate function argument, e.g.:
```js
  _wrapCollapse($obj, params = {
    // All params properties are optional
    time: undefined,
    name: undefined,
    icon: undefined,
    text: undefined,
    classes: undefined,
    onclick: undefined,
  }) {
```

valarnin: Whichever of these two options makes more sense to you, I'll switch to.

quisquous: I think I'd say just put it all in params with a comment, but I only weakly prefer that.  As you say, with typescript this is all kind of moot, and so mostly I just want something that's a little bit more readable and will remind me (or you) in the future of what these parameters mean when it becomes typescript.

valarnin: How's this?
image

quisquous: Even better, but I still would prefer it to say `${obj}: undefined` for shorthand consistency.

quisquous:
Thanks!!

PR #2657, opened 2021-04-03
typescript: remove tsconfig noImplicitReturns

quisquous:
When functions have return types, TypeScript already handles the case
where a function cannot have an undefined return type, implicit or
explicit.

I think it noisy to have to add a `return;` at the end of a function
that can return undefined as a type.  This will especially be the
case for triggers, which use implicit return undefined as a way
to not output things in many trigger functions.

Turning this off for now will help make that conversion easier.
If this creates an issue, then once everything has turned into
TypeScript we can revisit and turn it back on.

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -121,7 +121,6 @@ export default class PartyTracker {
       return names[1];
     if (names[1] === name)
       return names[0];
-    return;
   }
 
   // see: otherTank, but for healers.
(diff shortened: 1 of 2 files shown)


panicstevenson (approved):

quisquous:
cc @MaikoTan @Trim21 FYI if you didn't see this

PR #2661, opened 2021-04-03
fisher: fix place names with articles

### LINKED ISSUE #2651: Fisher : french translation - fishing spot misspelled
Nothing much to say : with the french version of the game, i found a misspelled spot in Azys Lla.

"Arbre à pappus" should be spelled "L'arbre à pappus". Works perfectly when changed in local files.

Thanks in advance ! :)

quisquous:
`Arbre à pappus` is `L'arbre à pappus` in game.  This can be fixed
by using the Name and not NameNoArticle field from ffxiv.

Doing this adds the article `The` to many English place names.
This requires removing the `The` handling from several places in
code.

However, using the article field also changes the capitalization of
some English places (e.g. `Limsa Lominsa Lower Decks` from
`Limsa Lominsa lower decks`). The latter is correct in game, and
so we also need to make all string comparisons case insensitive.

This will unfortunately break any local saved data for these
locations, but there is not much to be done about that.

foof

Fixes #2651.


ui/fisher/seabase.js
```diff
@@ -11,9 +11,20 @@ export default class SeaBase {
...
+    // lowercase it in the data unless we made another copy.
+    const lcVal = val.toLowerCase();
+    return Object.keys(obj).find((key) => {
+      if (Array.isArray(obj[key])) {
+        for (const subVal of obj[key]) {
+          if (subVal.toLowerCase() === lcVal)
+            return true;
+        }
+        return false;
+      }
```
panicstevenson, line 25:
```suggestion
      if (Array.isArray(obj[key])) {
        return obj[key].some((subVal) => subVal.toLowerCase() === lcVal);
      }
```

panicstevenson (approved):

PR #2611, opened 2021-03-21
Refactor: TypeScript-ize overlay_plugin_api.ts (again)

MaikoTan:
Following #2548 and #2576
I have tested this in ~~`raidboss`~~, `jobs`, `oopsy`, `radar`, `eureka` and `fisher` work properly.

PR #2666, opened 2021-04-04
Fix recursion error from overlay_plugin_api.ts

quisquous:
After #2611, the raid emulator enters infinite recursion.
```
overlay_plugin_api.ts:97 Uncaught (in promise) RangeError: Maximum call stack size exceeded
    at Object.addOverlayListener [as addOverlayListenerOverride] (overlay_plugin_api.ts:97)
    at Object.addOverlayListener [as addOverlayListenerOverride] (overlay_plugin_api.ts:98)
    at Object.addOverlayListener [as addOverlayListenerOverride] (overlay_plugin_api.ts:98)
    at Object.addOverlayListener [as addOverlayListenerOverride] (overlay_plugin_api.ts:98)
```

This is because of this call sequence:
* original addOverlayListener -> checks for override
* calls RaidEmulatorOverlayApiHook version -> calls original function
* original function checks for override again

The root of the issue is that the "original" function needs to not
include the override.  Fix this by returning an internal function
that does not include the override from the setOverride function.

Also, to avoid having to write this for the add/remove/dispatch,
remove those overrides as they are not used.

### DIFF
--- a/resources/overlay_plugin_api.ts
+++ b/resources/overlay_plugin_api.ts
@@ -41,13 +41,6 @@ declare global {
 type IAddOverlayListener = <T extends EventType>(event: T, cb: EventMap[T]) => void;
 type IRemoveOverlayListener = <T extends EventType>(event: T, cb: EventMap[T]) => void;
 
-interface Overrides {
-  addOverlayListenerOverride?: IAddOverlayListener;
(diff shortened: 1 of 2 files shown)


panicstevenson (approved):

quisquous:
cc @MaikoTan @valarnin

PR #2674, opened 2021-04-04
util: process and generate triggers branch

quisquous:
As trigger files are diverging more and more from things that users
can copy and paste into javascript file, create a triggers branch
that can be viewed online that has the most recent version of all
of the triggers.

This will allow us to start converting trigger files and still
have a place that users can look to see a JavaScript transpiled
version that they can more easily copy and paste as needed.

### DIFF
--- a/.github/workflows/update-triggers-branch.yml
+++ b/.github/workflows/update-triggers-branch.yml
@@ -1,0 +1,41 @@
+name: Update triggers
+on:
+  push:
+    branches: [ main ]
+
+jobs:
+  check-triggers:
+    runs-on: ubuntu-latest
+    outputs:
+      found: ${{ steps.check-triggers.outputs.found }}
+    steps:
+      - name: Check triggers Exists
+        id: check-triggers
+        shell: bash
+        run: |
+          GITHUB_URL="https://github.com/${{ github.repository }}"
+          if git ls-remote --exit-code --heads $GITHUB_URL triggers; then
+            echo "::set-output name=found::true"
+            echo "triggers branch found. Updating..."
+          else
+            echo "::set-output name=found::false"
+            echo "triggers branch not found. Skipping..."
+          fi
+  update-triggers:
+    runs-on: ubuntu-latest
+    needs: check-triggers
+    if: needs.check-triggers.outputs.found == 'true'
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build
+        run: |
+          npm install
+          npm run process-triggers
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./dist/triggers/ui/raidboss/data
+          user_name: "github-actions"
+          user_email: "github-actions@github.com"
+          commit_message: "gh-pages build: ${{ github.event.head_commit.message }}"
--- a/.github/workflows/update-triggers-branch.yml
+++ b/.github/workflows/update-triggers-branch.yml
@@ -36,6 +36,8 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./dist/triggers/ui/raidboss/data
+          publish_branch: triggers
           user_name: "github-actions"
           user_email: "github-actions@github.com"
-          commit_message: "gh-pages build: ${{ github.event.head_commit.message }}"
+          commit_message: "triggers build: ${{ github.event.head_commit.message }}"
+          enable_jekyll: true # Don't commit the .nojekyll file
--- a/package.json
+++ b/package.json
@@ -19,7 +19,8 @@
     "lint-staged": "node_modules/.bin/lint-staged",
     "coverage-report": "node --loader ts-node/esm util/gen_coverage_report.js",
     "find-translations": "node --loader ts-node/esm util/find_missing_translations.js",
-    "translate-timeline": "node --loader ts-node/esm util/translate_timeline.js"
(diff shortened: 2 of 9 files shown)


tsconfig_triggers.json
```diff
@@ -0,0 +1,28 @@
+{
```
panicstevenson, line 1: Does it make sense to do this programmatically instead (feel free to say no)?
https://github.com/Microsoft/TypeScript/wiki/Using-the-Compiler-API#a-minimal-compiler

package.json
```diff
@@ -18,7 +18,8 @@
     "test": "node_modules/.bin/mocha",
     "lint-staged": "node_modules/.bin/lint-staged",
     "coverage-report": "node --loader ts-node/esm util/gen_coverage_report.js",
-    "find-translations": "node --loader ts-node/esm util/find_missing_translations.js"
+    "find-translations": "node --loader ts-node/esm util/find_missing_translations.js",
+    "process-triggers": "node --loader ts-node/esm util/process_triggers_folder.js"
```
panicstevenson, line 22: Should this be .ts?

panicstevenson (commented):
I think overall LGTM? Want me to put this all on my repo and see how it works out?

tsconfig_triggers.json
quisquous: Oh yeah, that'd be a much better way to do it.

.github/workflows/update-triggers-branch.yml
```diff
@@ -0,0 +1,41 @@
...
+          npm install
+          npm run process-triggers
+      - name: Deploy
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          publish_dir: ./dist/triggers/ui/raidboss/data
+          user_name: "github-actions"
+          user_email: "github-actions@github.com"
+          commit_message: "gh-pages build: ${{ github.event.head_commit.message }}"
```
panicstevenson, line 41:
```suggestion
        with:
          github_token: ${{ secrets.GITHUB_TOKEN }}
          publish_dir: ./dist/triggers/ui/raidboss/data
          publish_branch: triggers
          user_name: "github-actions"
          user_email: "github-actions@github.com"
          commit_message: "triggers build: ${{ github.event.head_commit.message }}"
          enable_jekyll: true # Don't commit the .nojekyll file
```

util/process_triggers_folder.ts
```diff
@@ -0,0 +1,105 @@
+import { execSync } from 'child_process';
+import eslint from 'eslint';
+import fs from 'fs';
+import path from 'path';
+import process from 'process';
+
+const root = '../dist/triggers/ui/raidboss/data/';
+const tscCmd = `..\\node_modules\\.bin\\tsc --build ..\\tsconfig_triggers.json`;
```
panicstevenson, line 8:
```suggestion
const tscCmd = `${['..', 'node_modules', '.bin', 'tsc'].join(path.sep)} --build ${['..', 'tsconfig_triggers.json'].join(path.sep)}`;
```

panicstevenson (commented):
Here are the changes I made to get this working on my fork.

tsconfig_triggers.json
quisquous: Also a much more complicated way.  I'll uh leave a comment for future me.

quisquous:
@panicstevenson I assume you needed to create the triggers branch first for this to work? Anything else before I merge?

quisquous:
Should I also updated `RaidbossGuide.md` to differentiate user files and repository files?

panicstevenson:
> I assume you needed to create the triggers branch first for this to work?

Yea; otherwise the action will just be a no-op.

> Should I also update RaidbossGuide.md to differentiate user files and repository files? 

I don't understand what's being differentiated or why?

> Mostly I'm just not sure what the best way to broadcast this information to people is.

Do people need this information now? I thought this was planning to exist as a place to get !TypeScript triggers, since users won't be able to copy/paste/edit TypeScript triggers and put them into cactbot that way. I would expect the documentation update/broadcast at the point of trigger conversion. What are your thoughts here if I'm missing something?

quisquous:
> > Should I also update RaidbossGuide.md to differentiate user files and repository files?
> 
> I don't understand what's being differentiated or why?

Mmmm mostly the difference between user js files and repository ts files?  Mostly I worry that to non-programmers that the difference is much less obvious.

> > Mostly I'm just not sure what the best way to broadcast this information to people is.
> 
> Do people need this information now? I thought this was planning to exist as a place to get !TypeScript triggers, since users won't be able to copy/paste/edit TypeScript triggers and put them into cactbot that way. I would expect the documentation update/broadcast at the point of trigger conversion. What are your thoughts here if I'm missing something?

I think it's more that I expect trigger conversion to take a while and so this will be a slow transition, and people would probably want to know this at the point of TypeScript triggers existing.  I was thinking about writing new trigger files in typescript or converting at least one file soon, so maybe "any TypeScript triggers" might be in the near future even if "all TypeScript triggers" is far away.  (However, I suspect that there are probably some things in `util/` or `test/` that might not know what to do about `.ts` files yet, so there's more dependencies still?)

panicstevenson:
> I think it's more that I expect trigger conversion to take a while and so this will be a slow transition, and people would probably want to know this at the point of TypeScript triggers existing

Does it have to be a lengthy conversion process? Is it possible that we could loosen the TypeScript compilation rules for triggers and essentially move everything to `.ts`, and then come back and work on making them stricter in the future? I don't know if there's necessarily any issue with moving one trigger file at a time, but I think it would be a faster break-fix cycle to do it all in one go, even if the benefits are essentially negligible at that point.

PR #2675, opened 2021-04-05
util: fix find_missing_translations to report timelineReplace mismatches

quisquous:
This used to work, but when we switched over to netregexes from regexes
this broke and nobody ever noticed.  Unfortunately, there's a lot of
regexes that don't need to be translated, so this adds an
*EXTREMELY* hacky method to detect them.

NetRegexes know what fields that they have on construction, and only
particular fields need translation.  Therefore, if a global flag is
flipped and a NetRegex needs that field, insert a noop character
sequence at the beginning of the NetRegex that we can then detect
in order to know if this NetRegex had such fields when it was
constructed.

### DIFF
--- a/util/find_missing_timeline_translations.js
+++ b/util/find_missing_timeline_translations.js
@@ -1,8 +1,11 @@
 import fs from 'fs';
 import path from 'path';
 import Regexes from '../resources/regexes';
+import NetRegexes from '../resources/netregexes';
 import { Timeline } from '../ui/raidboss/timeline';
 import { commonReplacement, partialCommonReplacementKeys } from '../ui/raidboss/common_replacement';
+
+NetRegexes.setFlagTranslationsNeeded(true);
 
 export async function findMissing(triggersFile, locale) {
   // Hackily assume that any file with a txt file of the same name is a trigger/timeline.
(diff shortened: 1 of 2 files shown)


util/find_missing_timeline_translations.js
```diff
@@ -1,9 +1,12 @@
 import fs from 'fs';
 import path from 'path';
 import Regexes from '../resources/regexes';
+import NetRegexes from '../resources/netregexes';
 import { Timeline } from '../ui/raidboss/timeline';
 import { commonReplacement, partialCommonReplacementKeys } from '../ui/raidboss/common_replacement';
 
+NetRegexes.setFlagTranslationsNeeded(true);
```
panicstevenson, line 8: Sorry to be a pain, but should this have an additional comment here about why we're turning this on in the test file? It's kind of obvious when looking at files in the reverse manner (code -> test), but it might not be immediately obvious when looking at the test file alone.

panicstevenson (approved):

util/find_missing_timeline_translations.js
quisquous: No no, thanks.  This really does need a comment.

PR #2679, opened 2021-04-05
Add Regex.zoneSealed and Regex.zoneUnsealed functions #2677

Akurosia:
(accedentily removed branch before)
basic concept was added.
@quisquous  might you be able to help, why it does not recognize other languages other than en (are commen replacements not applied here?)?

Sealed example: `[10:35:26.000] 00:0839:Noch 15 Sekunden, bis sich die Rambade schließt.`
Unsealed example: `[10:35:44.000] 00:0839:Die Rambade öffnet sich wieder.`

### DIFF
--- a/resources/regexes.ts
+++ b/resources/regexes.ts
@@ -129,6 +129,8 @@ const gameLogParams = ['timestamp', 'code', 'line', 'capture'] as const;
 const gameNameLogParams = ['timestamp', 'code', 'name', 'line', 'capture'] as const;
 const changeZoneParams = ['timestamp', 'name', 'capture'] as const;
 const network6dParams = ['timestamp', 'instance', 'command', 'data0', 'data1', 'data2', 'data3', 'capture'] as const;
+const zoneSealedParams = ['timestamp', 'name', 'time', 'capture'] as const;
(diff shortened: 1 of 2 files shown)


Akurosia:
i tried to make it work with netregex as well, but i cant get it to work at all (e.g. you would need the same information as you have now with th enormal message line)

valarnin:
Are these not covered by `resources/translations.ts`?

quisquous:
Common replacements are only applied to timelines right now.  There's also no way to know the current language when constructing a regex (it's too low level for that), so you wouldn't be able to pick the correct one.

I am not sure that there's really a great way to do it.  You'd have to pass the language in as a parameter to it, e.g. `NetRegexes.areaSeal({ locale: 'en', name: 'Some Zone Name }),` and then construct the regex from there, somehow.

> Are these not covered by `resources/translations.ts`?

`translations.ts` has translations that match any zone seal/unseal by language, but not a specific one.  You could say `netRegex: LocaleNetRegex.areaSeal.en,` but that would match any zone, and you'd then have to name all the areas specifically in a condition to make such a trigger work.

PR #2682, opened 2021-04-06
Plugin: update CN 5.4 Offset

DieMoe233:


tssailzz8:
the cedirector_cn is 0x129

PR #2690, opened 2021-04-09
raidboss: Change new Date calls to emit from log events

valarnin:
This PR changes the calls to `new Date` in PopupText and the Timeline classes to instead be emitted from log line events, per various discussions.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -602,6 +602,9 @@ export class PopupText {
   }
 
   OnLog(e) {
+    // This could conceivably be determined based on the line's contents as well, but
+    // not sure if that's worth the effort
+    const currentTime = +new Date();
     for (const log of e.detail.logs) {
       if (log.includes('00:0038:cactbot wipe'))
         this.SetInCombat(false);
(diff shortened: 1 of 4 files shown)


ui/raidboss/popup-text.js
```diff
@@ -602,39 +602,43 @@ export class PopupText {
   }
 
   OnLog(e) {
+    // This could concievably be determined based on the line's contents as well, but
```
quisquous, line 605:
```suggestion
    // This could conceivably be determined based on the line's contents as well, but
```

I think line timings get weird because of the dropped milliseconds on `00` lines.  So, it's a little bit more reliable to have an independent timer, I think.

quisquous (commented):
Thanks for this refactoring! One comment on a comment, but otherwise looks great.

ui/raidboss/popup-text.js
valarnin: Yeah, plus this has the added benefit of exactly following the old behavior whereas if it pulled from the log line it could cause weird behavior if there was lag in the line parsing or whatever.

valarnin: The suggested change didn't fix both misspellings so I `--amend`-ed the commit to fix them both instead.

quisquous: Maybe I'm not seeing it, but did you change it?

valarnin: Yes, it's changed and force pushed.

valarnin: Or not. Now it should be. I really need to fix my vscode git integration.

PR #2700, opened 2021-04-11
lint: Add timelineTrigger Syntax Checks

panicstevenson:
Add rules for checking the timelineTrigger syntax.

Ensure that the `regex` property uses a regular expression literal:

Ensure properties such as `netRegex` or `regexFr` are not present:


quisquous:
Thanks!!

PR #2713, opened 2021-04-12
Refactor: TypeScript-ize files in utils/ to TypeScript

MaikoTan:
.


quisquous:
I agree with the idea that it's weird to have a mix of Python / JavaScript / TypeScript and that centralizing on TypeScript makes the most sense to me over time.  I do think it's a lower priority to convert Python scripts to TypeScript than it is to convert existing JavaScript code to TypeScript, especially code that is more user-facing and outside of `utils/`, but that's just my own bias towards what I would like to see first.

> It is hard to split every file as they should be refactored at the same time.

Hmm.  My concern here is that in a general programming sense, any change comes with risk of introducing bugs, especially when there's not any testing (oops).  Looking at all of the JavaScript -> TypeScript conversions, almost all of them have introduced bugs.  I would argue that a Python ->TypeScript rewrite is even more change than just massaging the JavaScript code to be type friendly.

To be clear, saying that all the changes have introduced bugs is not to assign any blame! Bugs happen; it's just how it goes.  I didn't see any of them myself either when looking at the patches.  Writing perfect code is not possible because we don't have a perfect understanding of the system.  However, the larger a change is, the more likely it is that something is going to be missed in code review, and more bugs will be introduced.

I think it would be impossible for me to review a patch that changed the entirety of `utils/` and so I would really recommend trying to break up this idea into smaller patches.

Here's a couple of ideas:
* convert the existing JavaScript files to TypeScript in individual PRs
* where there are utility files, add a separate parallel TypeScript version (e.g. add `csv_util.ts` but keep `csv_util.py` until there are no users of `csv_util.py`)
* convert the `gen_etc.py` files in individual PRs, which mostly use `csv_util.ts`
  * and then follow up to update the `gen_etc.ts` files to emit TypeScript files

What do you think about such an approach?

MaikoTan:
Sure! Certainly you're right!
It is better to keep the original files until we succeed converting all of them.

PR #2715, opened 2021-04-12
Fix small missed translation

Akurosia:


quisquous:
I missed it too! No need to apologize, thanks for the patch!!

PR #2716, opened 2021-04-13
refactor: add TypeScript version of coinach.py

MaikoTan:
Also add types for node.


util/coinach.ts
```diff
@@ -0,0 +1,176 @@
...
+  write(filename: string, scriptname: string, variable: string | null, d: unknown[]): void {
+    const fullPath = path.join(this.cactbotPath, filename);
+
+    let f = `// Auto-generated from ${scriptname}
+    // DO NOT EDIT THIS FILE DIRECTLY
+
+
+    export default `;
+
+    let str = JSON.stringify(d.sort());
```
quisquous, line 154: `JSON.stringify` is not the same as `json.dumps(d, sort_keys=True, indent=2, ensure_ascii=False)`.  In particular, it's missing the indent=2 and `sort` isn't a function on objects.

Re: indenting, one possibility here is to run eslint, see: https://github.com/quisquous/cactbot/blob/main/util/gen_coverage_report.js#L230-L248

Re: sorting, I think `JSON.stringify(d, Object.keys(d).sort())` will sort the top level keys.  If that's not enough and we need to sort the inner keys too, maybe we need https://www.npmjs.com/package/json-stable-stringify?

util/coinach.ts
```diff
@@ -0,0 +1,176 @@
...
+
+// Helper for automating SaintCoinach -- https://github.com/ufx/SaintCoinach
+
+import { promisify } from 'util';
+import fs from 'fs';
+import path from 'path';
+import { exec } from 'child_process';
+
+
+const _COINACH_EXE = 'SaintCoinach.Cmd.exe';
```
quisquous, line 11: Do you think it's be worth converting these to camelCase names? The uppercase constants is a Python PEP8 convention.

quisquous (commented):
Thanks for this!

util/coinach.ts
panicstevenson: I would expect to follow the convention set by [typescript-eslint/naming-convention](https://github.com/typescript-eslint/typescript-eslint/blob/v4.9.1/packages/eslint-plugin/docs/rules/naming-convention.md), which would accept `coinachExe` (like you suggested), or `COINACH_EXE` (what exists without the leading underscore). Throwing my personal opinion in, I like either format but I would consider `coinachExe` to be more correct given the usage is constrained solely to this class.

util/coinach.ts
panicstevenson: [JSON.stringify()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify) can take a third argument for spacing; I don't know if that covers the need here; `JSON.stringify(d, Object.keys(d).sort(), 2)` in this instance.

util/coinach.ts
```diff
@@ -0,0 +1,176 @@
+#!/usr/bin/env node
```
panicstevenson, line 1: I think the shebang here would be updated as well:
```suggestion
#!/usr/bin/env ts-node
```
I think this also invites a conversation if this file should be executable to match the previous Python script usage, or if we should instead drop the shebang at the top of each TypeScript util file (drop their executable nature) and instead expect them all to run from `npm run`/`yarn` commands.

panicstevenson (commented):
Just echo'ing what quisquous wrote: thanks for this!

util/coinach.ts
quisquous: I'm not sure I have strong opinions here.  I think mostly `npm run` commands are there to be easier for commonly run commands (e.g. find-translations) or things that are run by build scripts (e.g. the coverage report, just to have a layer of abstraction so that it's easy to know "ok I run X command to do the same thing as the build step").  Other things in util are probably used once a patch by 1-2 people tops, and so it doesn't feel like it needs the same level of accessibility in some ways.

Is there a reason not to have the shebang?

util/coinach.ts
quisquous: Ah yeah, thanks! I think we'd still want to eslint as well, though.

util/coinach.ts
panicstevenson: > Is there a reason not to have the shebang?

Strictly, I think no. I think the shebang, regardless if the file is committed as an executable or not (I don't think it is here, but I can't firmly tell?), is okay to "just have". I would want to confirm that running this file as an executable (`./coinach.ts`) works as expected; I guess at that point the expectation there would be someone installing TypeScript globally, since I doubt `ts-node` is on your path if it's only in your `node_modules/` directory (something that should be documented accordingly, in an ideal world).

I guess this is more a question of "Should these files have two entrypoints?" which I think the answer is going to always want to be no, so we should firm up our single entrypoint for these, which is diverging from the place where entrypoints live. Sorry, this may be a lot of thinking for something kind of pointless, but I guess this is mainly "Should we close the loop on this pattern and expect people to only run from the commands in `package.json`"?

quisquous: Maiko also made this suggestion on another patch: https://github.com/quisquous/cactbot/pull/2603#issuecomment-802670848 in terms of having a single npm entrypoint for all the scripts (with a chooser), which I think would prevent proliferation of a million gen_x_data entry points in `package.json`

Regardless, I'm fine with removing the shebangs for now in favor of moving towards a more consistent approach in the future.  I don't think I call any scripts directly like that.

panicstevenson: I think I'm of the same opinion, mostly because I think that an implicit goal of converting things to TypeScript should be that you can do more without having to do external installs. For instance, a pain point around Python is that you have to install it separately from the repository. Ideally, we could limit the dev world down to only needing to install Node and then everything that the repo expects would live inside node_modules and can be called from there. Allowing these files to be called directly would mean that the end user is installing TypeScript globally, with a potentially separate version.

MaikoTan: Okay, I'll remove the shebang later.

util/coinach.ts
```diff
@@ -0,0 +1,176 @@
...
+      console.log('ffxiv_path: ' + this.ffxivPath);
+      console.log('cmd: ' + cmd);
+    }
+    // # This will throw an exception if stuff is VERY wrong
+    // # however, return code is still 0 even if all exports fail.
+    // # Also, it seems to need to be run from the SaintCoinach directory.
+    const output = String(promisify(exec)(cmd));
+    // # Manually check output for errors.
+    let m = /^([0-9])* files exported, ([0-9])* failed/m.exec(output);
+    if (!m) throw new CoinachError('Unknown output', cmd, output);
```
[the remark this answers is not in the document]
MaikoTan: Sure

util/coinach.ts
```diff
@@ -0,0 +1,176 @@
...
+    fs.writeFileSync(fullPath, f);
+
+    if (this.verbose) console.log(`wrote: ${filename}`);
+  }
+}
+
+const reader = new CoinachReader(null, null, true);
+reader.exd(process.argv[2] ?? '', 'en').then((lines) => {
+  new CoinachWriter(null, true).write('status.js', 'this', null, lines);
+}).catch((e) => console.error(e));
```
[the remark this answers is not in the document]
MaikoTan: Thanks, I just know that there is an equivalent of `if __name__ == '__main__'` in nodejs!

package-lock.json
```diff
@@ -1,7 +1,11595 @@
 {
   "name": "cactbot",
+  "lockfileVersion": 2,
```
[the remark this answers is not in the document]
MaikoTan: But when I read the doc of npm, it said version 2 should be backwards compatible with version 1.

https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion

Which version of npm are you using, maybe you need an upgrade?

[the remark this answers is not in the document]
quisquous: It is ... sort of backwards compatible? npm says "I'll try my best" and prints out a warning.  It seems to work, but that also seems fragile.  My npm is 6.14.8.

Regardless, I think if you want to upgrade the lockfile, it'd be better to do that in a separate PR where nothing else changes but updating the lockfile version.  It makes the diff impossible to read if you are both bumping the version and adding new packages.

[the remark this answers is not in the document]
MaikoTan: Well, I just don't know how to prevent `npm` to commit the newer lockfile version, and I have not found solutions on Internet (maybe my searching keyword is wrong?)
Maybe I should manually revert this.

And sorry that I'm busy these days, didn't respond so long.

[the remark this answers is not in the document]
quisquous:
```diff
diff --git a/package-lock.json b/package-lock.json
index 5af70bb36..d0abd338c 100644
--- a/package-lock.json
+++ b/package-lock.json
@@ -492,9 +492,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "13.1.1",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-13.1.1.tgz",
-      "integrity": "sha512-hx6zWtudh3Arsbl3cXay+JnkvVgCKzCWKv42C9J01N2T2np4h8w5X8u6T
pz5mj38kE3M9FM0Pazx8vKFFMnjLQ==",
+      "version": "14.14.43",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-14.14.43.tgz",
+      "integrity": "sha512-3pwDJjp1PWacPTpH0LcfhgjvurQvrZFBrC6xxjaUEZ7ifUtT32jtjPxEM
Mblpqd2Mvx+k8haqQJLQxolyGN/cQ==",
       "dev": true
     },
     "@types/normalize-package-data": {
```

Is the diff in lockfile version 1, if that helps? Maybe a separate PR to talk about bumping to lockfile version 2 and what that means.

quisquous:
Also, sorry for being slow on getting back to this review too.  I have added a new function to `coinach.py` to write typescript files that you might want to pick up (either here or in a followup pr, if you'd rather just land this first).

util/coinach.ts
```diff
@@ -0,0 +1,243 @@
...
+      for (const p of defaultFfxivPaths) {
+        if (fs.existsSync(path.join(p, ffxivExe))) {
+          ffxivPath = p;
+          break;
+        }
+      }
+    }
+
+    if (!coinachPath) throw new CoinachError('not found coinach');
+    if (!ffxivPath) throw new CoinachError('not found ffxiv');
```
quisquous, line 62:
```suggestion
    if (!coinachPath) throw new CoinachError('coinach path not found');
    if (!ffxivPath) throw new CoinachError('ffxiv path not found');
```

package-lock.json
```diff
@@ -492,9 +492,9 @@
       "dev": true
     },
     "@types/node": {
-      "version": "13.1.1",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-13.1.1.tgz",
-      "integrity": "sha512-hx6zWtudh3Arsbl3cXay+JnkvVgCKzCWKv42C9J01N2T2np4h8w5X8u6Tpz5mj38kE3M9FM0Pazx8vKFFMnjLQ==",
+      "version": "14.14.37",
```
quisquous, line 495: I think this is causing the process triggers tsc error.

quisquous (commented):
Looks good in general, two tiny comments.

package-lock.json
MaikoTan: Forgot why I add this, but let me try to revert it. ಠ_ಠ

quisquous: Looks like you didn't need it?

PR #2723, opened 2021-04-14
raidboss/oopsy: Leviathan Unreal

quisquous:
Same as extreme, but search/replaced with new ids.

Changes:
* Sahagin appear to do more abilities (or maybe I just saw more?) so
  ignored more of those.
* Changed up some of the tidal roar syncs.  I think the timeline itself
  is kind of a mess in the middle part of the fight and there's really
  just not much to do about it.
* Changed oopsy to ignore Waterspout and Dread Tide damage as it appears
  to always be 0x16 even in unreal, so looks like it's shared when it's
  not.
* Added more spinning dive ability ids.  It seems like unreal has more
  unless I just missed some in extreme.


panicstevenson (approved):
Mantle and Veil of the Whorls are backwards (assuming the same for Levi Ex), sorry for not catching that sooner.
Can we add a delay to when the add spawns to when you should start stunning it? Maybe a 5 second delay?

PR #2724, opened 2021-04-14
Raidboss: Paglth’an timeline and triggers

JLGarber:
Still a work in progress, but triggers are basically done. No testing yet.


ui/raidboss/data/05-shb/dungeon/paglthan.js
```diff
@@ -0,0 +1,102 @@
...
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Paglthan Lightning Rod Gain',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'A0E' }),
+      condition: Conditions.targetIsYou(),
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Drop lightning outside',
```
panicstevenson, line 40: I don't have a particularly brief way to say this, but I think the mechanic here is to put the lightning circle on the rod so you can go back to hitting the boss. Someone can correct me if I'm wrong.

ui/raidboss/data/05-shb/dungeon/paglthan.js
```diff
@@ -0,0 +1,102 @@
...
+      response: Responses.stackMarkerOn(),
+    },
+    {
+      id: 'Paglthan Mega Flare',
+      netRegex: NetRegexes.headMarker({ id: 'E000' }),
+      condition: Conditions.targetIsYou(),
+      response: Responses.spread(),
+    },
+    {
+      id: 'Paglthan Kan Rhai',
```
panicstevenson, line 79: This might need a tiny bit more complexity around it, since the move will continue hitting in the place that it's dropped.
In other words, this mechanic is dissimilar from the plus mechanics in Lakshmi (a one-time hit from the player). I don't know if that's obvious or if there should be a MOVE! alert to follow this if it picked you.

Feel free to disagree!

panicstevenson (commented):
Looks good! What did you feel like is missing to publish this?

ui/raidboss/data/05-shb/dungeon/paglthan.js
JLGarber: Honestly, I didn't even see the rods outside, I was mostly focused on "where are these trust NPCs going, why are they stacked on me, PLEASE STOP, GET AWAY". I haven't had time yet to run this again, but I'll be doing that relatively soon and I'll double-check.

ui/raidboss/data/05-shb/dungeon/paglthan.js
JLGarber: I kind of think that's covered by "Drop AoE Outside"? It seems to me that's the typical phrasing we used for shedding persistent environmental hazards, as opposed to targeted Chariot-type AoEs that would be covered by `spread` or `getOut`. I suppose we could do "Drop Cross Puddle Outside"?

JLGarber:
As far as what's missing, I need to find the loop point on boss 3 and set that up on the timeline. I *think* that's the loop right there at the end, but I'm not 100% sure, and even with trust NPCs it still went significantly faster than I expected.

Boss 2 should probably have the knockback set up with a warning as well, if only because it's easy to be caught out by it.

(Also I have no idea why Git dredged up year-old commits on this branch.)

ui/raidboss/data/05-shb/dungeon/paglthan.js
panicstevenson: Oh, I was thinking more along the lines of like a delaySeconds: 3 that shouts `Move!` on the person with the headmarker but that might be excessive.

ui/raidboss/data/05-shb/dungeon/paglthan.js
```diff
@@ -0,0 +1,107 @@
...
+      netRegex: NetRegexes.headMarker({ id: '0104' }),
+      condition: Conditions.targetIsYou(),
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Drop AoE outside',
```
panicstevenson, line 90: It's not informative or useful, but should this just say `Kan Rhai on YOU`? The difference just being that one is just a "wake up, you have the mechanic" and the other is potentially telling someone to disconnect.

Again, feel free to disagree here as well.

JLGarber: Actually, that's likely the best solution. It's an extremely visible AoE as opposed to something like the chasing puddles on E8N.

JLGarber:
I think this is probably good enough for an initial release. We can improve on this as bug reports and complaints come in, but as far as "let's get this into peoples' hands", I think it's fine.

ui/raidboss/data/05-shb/dungeon/paglthan.js
```diff
@@ -0,0 +1,107 @@
...
+import NetRegexes from '../../../../../resources/netregexes';
+import { Responses } from '../../../../../resources/responses';
+import ZoneId from '../../../../../resources/zone_id';
+
+export default {
+  zoneId: ZoneId.Paglthan,
+  timelineFile: 'paglthan.txt',
+  timelineTriggers: [
+    {
+      // This is a rear cone attack that always follows Wide Blaster.
```
quisquous, line 11: Would it make sense to trigger this off of wide blaster, then? (Not that timeline triggers should be avoided, but they feel less safe than triggering off of a log directly.)

ui/raidboss/data/05-shb/dungeon/paglthan.js
```diff
@@ -0,0 +1,107 @@
...
+      },
+    },
+    {
+      id: 'Paglthan Lightning Rod Lose',
+      netRegex: NetRegexes.losesEffect({ effectId: 'A0E' }),
+      condition: Conditions.targetIsYou(),
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Away from lightning circles',
```
quisquous, line 51: Is it simpler to say "get middle"?

ui/raidboss/data/05-shb/dungeon/paglthan.js
quisquous: Yeah, I was having the same thought as Panic, in that the surprising part of this mechanic is just the "move away" part at the end, and the "don't hit people with the giant cross on you" is more obvious.

quisquous (commented):
I merged this already, but here's a few bikeshed asides for any future followup.  Feel free to disagree/ignore all of these thoughts, as I thought the original patch was good.  Just thought I'd throw some post-merge drive-by opinions.  O:)

JLGarber:
Nono, happy to have more bikesheds here, this was solely a "get it implemented so the support list has the checkmark" submission.

I suppose the center is always safe during lightning rods, so yes, that would make sense. I'll update that for the next release.

Okay, I see where you both were going with the Kan Rhai stuff. There's the end of the castbar, which I *think* is the snapshot? It ends about 1s before the first strike on Kan Rhai. If that doesn't work, there's an Unknown ability that hits right before the first strike. (I'd prefer not to use arbitrary delays for mechanics with semi-strict timing if we can avoid it. We don't want the user to move in before the marker drops.)

PR #2730, opened 2021-04-15
raidboss: Initial Tower at Paradigm's Breach Triggers

panicstevenson:
Rough first pass at the 5.5 alliance raid, Tower at Paradigm's Breach triggers.


ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -0,0 +1,490 @@
...
+    {
+      id: 'Paradigm Knave Magic Artillery Beta Collect',
+      netRegex: NetRegexes.headMarker({ id: '00DA' }),
+      run: function(data, matches) {
+        data.busterTargets = data.busterTargets || [];
+        data.busterTargets.push(matches.target);
+      },
+    },
+    {
+      id: 'Paradigm Knave Magic Artillery Beta Not You',
```
quisquous, line 90:
```suggestion
      id: 'Paradigm Knave Magic Artillery Beta',
```

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -0,0 +1,490 @@
...
+      outputStrings: {
+        text: {
+          en: 'Knockback (from clone)',
+          de: 'Rückstoß (vom Klon)',
+        },
+      },
+      run: (data) => data.cloneLunge = true,
+    },
+    {
+      id: 'Paradigm Copied Knave Lunge Out of Middle',
```
quisquous, line 166: Is there a Lunge Get Middle?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -0,0 +1,490 @@
...
+      response: Responses.stackMarker(),
+    },
+    {
+      id: 'Paradigm Red Girl Cruelty',
+      netRegex: NetRegexes.startsUsing({ id: '6012', source: 'Red Girl', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ id: '6012', source: 'Rot(?:e|er|es|en) Mädchen', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ id: '6012', source: 'Fille En Rouge', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ id: '6012', source: '赤い少女', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      durationSeconds: 4,
```
quisquous, line 251: There's a lot of durationSeconds: 4 here.  Is there a reason to override the default of 3 in so many places?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -0,0 +1,490 @@
...
+    },
+    {
+      id: 'Paradigm False Idol Darker Note You',
+      netRegex: NetRegexes.headMarker({ id: '008B' }),
+      condition: Conditions.targetIsYou(),
+      durationSeconds: 4,
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Tank Buster and Spread',
```
quisquous, line 331: This is kinda wordy.  I think spread should come first, at least.  But I'd suggest `Spread + Tank Buster` or even just `Tank Cleave`?

This is also a tiny thing (which you should feel very free to disagree about) but I wonder about putting this output string in the `Paradigm False Idol Darker Note` trigger, so then you only have one collect trigger (no output) and one output trigger (that handles all players).

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -0,0 +1,490 @@
...
+      id: 'Paradigm Copied Knave Lunge Out of Middle',
+      netRegex: NetRegexes.startsUsing({ id: '60C8', source: 'Knave of Hearts', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ id: '60C8', source: 'Herzbube', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ id: '60C8', source: 'Jack', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ id: '60C8', source: 'ジャック', capture: false }),
+      durationSeconds: 5,
+      alertText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Knockback (from clone) + Out of Middle',
```
quisquous, line 175: `Get Knocked Out Of Middle` or `Knockback -> Out Of Middle`?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
panicstevenson: Don't know! There's `Assuming one Lunge + Colossal Impact pattern; needs confirmation` in the TODO

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
panicstevenson: Doesn't have to, just felt more appropriate since the cast times were pretty long (in my opinion)

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
panicstevenson: Sure! Took this from the other alliance raid

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
panicstevenson: I mean, the root of the callout here is: You have a spread circle that does tank buster damage, and if the circles overlap you will kill both of you instantly.
It might be just fine to say "Tank Buster" if the spread + no overlap is obvious

quisquous: Yeah, I guess I'm asking how it's different than the previous tankbusters in the alliance raid where all three tanks get marked with a (presumed) tank buster cleave and also need to spread there.  This one has a marker and a cute circle, but it feels somewhat like the mechanic is the same in both cases, is my line of thinking.

Mostly my worry about the "and spread" part is that there's no avoiding the buster, it's a "spread so you can take this tank buster" not "take this tank buster...and then book it somewhere else", which is what the text might make somebody think.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
panicstevenson: Do these knockbacks need/want to mention their origin (boss/clone)?

quisquous: I'm a little indifferent here, mostly because I expect followup triggers to say Knockback South or Knockback East etc.

panicstevenson:
Moving this out of draft; I don't think these triggers are perfect, and they are objectively incomplete, but I think it's something and has been sufficiently looked over. If anyone has any remaining thoughts/opinions, feel free to still post them.

JLGarber (approved):

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -68,14 +76,12 @@ export default {
       delaySeconds: 2.1,
       durationSeconds: 5,
       alertText: (data, _, output) => output.text(),
+      tts: {
+        en: 'Go North/South Sides',
```
quisquous, line 80: tts nit, but you should also use output strings here, so that the tts string can be edited in the config.  Sorry that tests are not smart enough to tell you this.

PR #2734, opened 2021-04-15
raidboss: Initial Diamond Weapon Triggers

panicstevenson:


ui/raidboss/data/05-shb/trial/diamond_weapon.js
```diff
@@ -0,0 +1,147 @@
...
+      id: 'Diamond Diamond Rain',
+      netRegex: NetRegexes.startsUsing({ source: 'The Diamond Weapon', id: '5FA7', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      durationSeconds: 4,
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Diamond Claw Swipe East',
+      netRegex: NetRegexes.startsUsing({ source: 'The Diamond Weapon', id: '5F9A', capture: false }),
+      delaySeconds: 2,
```
JLGarber, line 29: Could we write this as `(data, matches) => parseFloat(matches.duration) - DELAY`? This way it's easier to see exactly how long before the end of the cast we want the user to be notified, as opposed to approximating it with an arbitrary number.

JLGarber (approved):

ui/raidboss/data/05-shb/trial/diamond_weapon.js
panicstevenson: I can; in this case it would be `(data, matches) => parseFloat(matches.duration) - 2,` though (if that makes a difference).
Alternatively, these don't really need a delay, or also don't really need to be called this early. 
I decided to go the middle route between the two and call it halfway through the castbar and extend the duration telling you what to do.

kshman:
right claw
* 5F9A nothing
* 5FA2 spread
* 5FA4 share

left claw
* 5F9B nothing
* 5FA3 spread
* 5FA5 share

panicstevenson:
> right claw
> 
> * 5F9A nothing
> * 5FA2 spread
> * 5FA4 share
> 
> left claw
> 
> * 5F9B nothing
> * 5FA3 spread
> * 5FA5 share

Are there other variants of the claw attacks in normal mode that I am unaware of? I only have 5F9A and 5F9B in my logs.

ui/raidboss/data/05-shb/trial/diamond_weapon.js
JLGarber: That's fair. I don't really care about *whether* there's a delay, honestly, just that if there *is* we display it as a function so we know how much time we're giving before the cast bar completes.

ui/raidboss/data/05-shb/trial/diamond_weapon.js
```diff
@@ -0,0 +1,147 @@
...
+      },
+    },
+    {
+      id: 'Diamond Articulated Bits',
+      netRegex: NetRegexes.ability({ source: 'The Diamond Weapon', id: '5FA9', capture: false }),
+      durationSeconds: 4,
+      infoText: (data, _, output) => output.text(),
+      outputStrings: {
+        text: {
+          en: 'Avoid hands',
```
quisquous, line 117: This is a microbikeshed, but maybe `Avoid Bits` (since folks know what bits are, and this is summoning bits?) or even `Avoid lasers` (if you want to say what they're doing rather than what they are)?

ui/raidboss/data/05-shb/trial/diamond_weapon.js
quisquous: My general rule of thumb is to call things as early as possible.  The only exceptions are really knockbacks (call 5 seconds beforehand) or maybe something like "move around"/"stop" where you want to call it closer to the point where somebody has to take (or not take) action vs having them stand around and do nothing for a bit.  Something like "move to the other side" should happen as soon as possible, imo.

ui/raidboss/data/05-shb/trial/diamond_weapon.js
```diff
@@ -0,0 +1,140 @@
...
+      id: 'Diamond Diamond Rain',
+      netRegex: NetRegexes.startsUsing({ source: 'The Diamond Weapon', id: '5FA7', capture: false }),
+      condition: Conditions.caresAboutAOE(),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Diamond Claw Swipe East',
+      netRegex: NetRegexes.startsUsing({ source: 'The Diamond Weapon', id: '5F9A', capture: false }),
+      durationSeconds: 10,
+      outputStrings: sharedOutputStrings,
```
quisquous, line 29: Sorry for the tiniest of nits, but outputStrings goes last.  I'm honestly surprised the test didn't yell about this ordering.

quisquous:
> Are there other variants of the claw attacks in normal mode that I am unaware of? I only have 5F9A and 5F9B in my logs.

I also only see 5F9A and 5F9B in my logs and don't see any of these others.

PR #2736, opened 2021-04-16
raidboss: Diamond Weapon Extreme timeline

valarnin:
This PR has an initial timeline for Diamond Weapon Extreme.

I've only done limited testing for this but the fight is fairly static.

As mentioned in Discord, there's a bug in raidemulator which prevents the timeline from syncing. I'll try to track that down and get a PR in for it soon.

### DIFF
--- a/ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
+++ b/ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
@@ -1,0 +1,123 @@
+### Diamond Weapon Ex
+
+hideall "--Reset--"
+hideall "--sync--"
+
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+# -it "The Diamond Weapon" -ii 5FEB 5FBE 5FAA 5FEC 5FBA 5FFD 5FBB 6055 6159 5FF9 5FCD 5FFA 5FF9 612E 5FB3 5FCE 5FD2 5FFE 5FD3 5FB4 5FB6 5FB1 5FB0 5FBF 5FAE 5FA6 5FC3 5FC9 5FCA 5FC4
+
+### Phase 1
+0 "--sync--" sync /:5FA7:The Diamond Weapon/
+0 "Start" sync /Engage!/ window 1,1
+21.3 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/ window 25,10
+
+# 5FA2 = Spread (Homing Laser, 5FA0)
+# 5F9B = "nothing" (Diamond Rain, 5FA7)
+# 5FA5 = Stack (Diamond Flash, 5FA1)
+36.5 "Adamant Purge" sync /:The Diamond Weapon:5FA[2345]:/ window 20,5
+46.7 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+49.9 "Homing Laser/Diamond Flash" sync /:The Diamond Weapon:5FA[01]:/
+67.1 "Photon Burst" sync /:The Diamond Weapon:5FA8:/
+83.4 "Adamant Purge" sync /:The Diamond Weapon:5F9[8B]:/
+93.6 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+104.8 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/
+120.0 "Adamant Purge" sync /:The Diamond Weapon:5FA[2345]:/
+130.2 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+133.7 "Homing Laser/Diamond Flash" sync /:The Diamond Weapon:5FA[01]:/
+151.4 "Photon Burst" sync /:The Diamond Weapon:5FA8:/
+
+# Phase transition
+167.7 "Code Chi-Xi-Stigma" sync /:The Diamond Weapon:5FAD:/
+168.0 "--untargetable--"
+183.3 "Auri Cyclone" sync /:The Diamond Weapon:5FB9:/
+188.0 "Diamond Shot" sync /:Articulated Bit:5FC0:/
+
+### Phase 2
+197.8 "--targetable--"
+210.0 "Outrage" sync /:The Diamond Weapon:5FBC:/
+218.2 "Outrage" sync /:The Diamond Weapon:5FBC:/
+229.6 "Auri Arts (Z)" sync /:The Diamond Weapon:5FAF:/
+229.6 "--untargetable--"
+236.4 "Auri Arts (Jump)" sync /:The Diamond Weapon:5FB2:/
+241.0 "--targetable--"
+250.4 "Outrage" sync /:The Diamond Weapon:5FBC:/
+257.6 "Auri Doomstead" sync /:The Diamond Weapon:5FBD:/
+265.5 "--untargetable--"
+271.8 "Auri Arts (Cleave)?" sync /:The Diamond Weapon:5FB5:/
+272.8 "Auri Arts (Jump)" sync /:The Diamond Weapon:5FCF:/
+272.8 "--targetable--"
+278.7 "Vertical Cleave" sync /:The Diamond Weapon:5FD0:/
+291.1 "Outrage" sync /:The Diamond Weapon:5FBC:/
+301.3 "Articulated Bits" sync /:The Diamond Weapon:5FC1:/
+312.4 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+312.9 "Auri Arts (Z)" sync /:The Diamond Weapon:5FAF:/
+312.9 "--untargetable--"
+319.7 "Auri Arts (Jump)" sync /:The Diamond Weapon:5FB2:/
+320.4 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+324.2 "--targetable--"
+333.7 "Outrage" sync /:The Diamond Weapon:5FBC:/
+340.9 "Auri Doomstead" sync /:The Diamond Weapon:5FBD:/
+353.1 "Articulated Bits" sync /:The Diamond Weapon:5FC1:/
+355.7 "--untargetable--"
+360.7 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+362.2 "Auri Arts (Cleave)?" sync /:The Diamond Weapon:5FB5:/
+364.0 "--targetable--"
+364.0 "Auri Arts (Jump)" sync /:The Diamond Weapon:5FCF:/
+369.9 "Vertical Cleave" sync /:The Diamond Weapon:5FD0:/
+370.7 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+382.4 "Outrage" sync /:The Diamond Weapon:5FBC:/
+388.0 "--untargetable--"
+
+### Phase 3
+397.9 "--targetable--"
+422.1 "--sync--" sync /:The Diamond Weapon:5FA6:/
+422.7 "Flood Ray 1" #sync /:The Diamond Weapon:5FC7:/
+424.2 "Flood Ray 2" #sync /:The Diamond Weapon:5FC7:/
+425.7 "Flood Ray 3" #sync /:The Diamond Weapon:5FC7:/
+427.2 "Flood Ray 4" #sync /:The Diamond Weapon:5FC7:/
+428.7 "Flood Ray 5" #sync /:The Diamond Weapon:5FC7:/
+430.2 "Flood Ray 6" #sync /:The Diamond Weapon:5FC7:/
+431.7 "Flood Ray 7" #sync /:The Diamond Weapon:5FC7:/
+433.2 "Flood Ray 8" #sync /:The Diamond Weapon:5FC7:/
+450.7 "Photon Burst" sync /:The Diamond Weapon:5FA8:/
+470.0 "Articulated Bits" sync /:The Diamond Weapon:5FA9:/
+477.1 "Adamant Purge" sync /:The Diamond Weapon:5FA[2345]:/
+481.6 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+486.2 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+489.7 "Homing Laser/Diamond Flash" sync /:The Diamond Weapon:5FA[01]:/
+490.6 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+501.3 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/
+516.5 "Articulated Bits" sync /:The Diamond Weapon:5FA9:/
+523.6 "Diamond Shrapnel" sync /:The Diamond Weapon:5FAC:/
+528.1 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+537.1 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+541.5 "Burst" sync /:The Diamond Weapon:5FC[56]:/
+546.8 "Photon Burst" sync /:The Diamond Weapon:5FA8:/
+566.1 "Articulated Bits" sync /:The Diamond Weapon:5FA9:/
+# This Purge seems to be able to be at least 5FA3, 5FA4, 5FA5, so including every possiblility
+573.2 "Adamant Purge" sync /:The Diamond Weapon:5F(9[ABCD]|A[2345]):/
+577.7 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+583.4 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+586.8 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+597.8 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/
+621.0 "--sync--" sync /:The Diamond Weapon:5FA6:/
+621.1 "Flood Ray 1" #sync /:The Diamond Weapon:5FC7:/
+622.6 "Flood Ray 2" #sync /:The Diamond Weapon:5FC7:/
+624.1 "Flood Ray 3" #sync /:The Diamond Weapon:5FC7:/
+625.6 "Flood Ray 4" #sync /:The Diamond Weapon:5FC7:/
+627.1 "Flood Ray 5" #sync /:The Diamond Weapon:5FC7:/
+628.6 "Flood Ray 6" #sync /:The Diamond Weapon:5FC7:/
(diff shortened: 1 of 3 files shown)


valarnin:
Slight update to account for differing variations of `Adamant Purge`.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+# There are no autoattacks from the boss until after the initial raidwide, so nothing to sync against for autos
+0 "Start" sync /Engage!/ window 1,1
+21.3 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/ window 25,10
+
+# Adamant Purge is a mini-phase that will be either stack, spread, or "nothing"
+# 5FA2 = Spread (Homing Laser, 5FA0)
+# 5F9B = "nothing" (Diamond Rain, 5FA7)
+# 5FA5 = Stack (Diamond Flash, 5FA1)
+# This mechanic repeats in phase 3 with additional bits and an extra tracking If circle aoe
+# @TODO: Someone better at timelines than me, sync these to mini-phases instead
```
quisquous, line 20: I am not sure what you're proposing.  I think this P1 is good as it is.  It's a linear timeline with a minor A/B -> B/A branch.  Usually we don't break those out into separate blocks unless it's more like o7s level branching.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+36.5 "Adamant Purge" sync /:The Diamond Weapon:5F(9[ABCD]|A[2345]):/ window 20,5
+46.7 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+49.9 "Spread/Stack?" sync /:The Diamond Weapon:5FA[01]:/
+67.1 "Photon Burst" sync /:The Diamond Weapon:5FA8:/
+83.4 "Adamant Purge" sync /:The Diamond Weapon:5F(9[ABCD]|A[2345]):/
+93.6 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+96.8 "Spread/Stack?" sync /:The Diamond Weapon:5FA[01]:/
```
quisquous, line 27: Is this true? I think it's always claw 1, spread/stack, claw 2, nothing, claw 3, stack/spread.  I do not see such a 5FA[01] in my logs.  e.g. I see a 5FA3/5FA0, a 5F9A on its own, then a 5FA5/5FA1.  Do you have a counterexample log?

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+hideall "--sync--"
+
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+# -ii 5FEB 5FBE 5FAA 5FEC 5FBA 5FFD 5FBB 6055 6159 5FF9 5FCD 5FFA 5FF9 612E 5FB3 5FCE 5FD2 5FFE 5FD3 5FB4 5FB6 5FB1 5FB0 5FBF 5FAE 5FA6 5FC3 5FC9 5FCA 5FC4
+
+### Phase 1
+# There are no autoattacks from the boss until after the initial raidwide, so nothing to sync against for autos
+0 "Start" sync /Engage!/ window 1,1
+21.3 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/ window 25,10
```
quisquous, line 13: You could put a `sync /:5FA7:The Diamond Weapon/` before this to start syncing earlier too in case somebody pulls without a countdown (or there's an egregious early pull).

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+# Phase transition
+167.7 "Code Chi-Xi-Stigma" sync /:The Diamond Weapon:5FAD:/
+183.3 "Auri Cyclone" sync /:The Diamond Weapon:5FB9:/
+184.3 "Auri Cyclone" sync /:The Diamond Weapon:5FD1:/
+188.0 "Diamond Shot" sync /:Articulated Bit:5FC0:/
+
+### Phase 2
+210.0 "Outrage" sync /:The Diamond Weapon:5FBC:/
+218.2 "Outrage" sync /:The Diamond Weapon:5FBC:/
+229.6 "Auri Arts" sync /:The Diamond Weapon:5FAF:/
```
quisquous, line 43: Do you think it'd be worth labeling these somehow, e.g. `Auri Arts (Z)` and `Auri Arts (knockback)`.  Or, if names are ambiguous, at least numbers to differentiate in case you forget where you are in the rotation?

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+
+# Adamant Purge is a mini-phase that will be either stack, spread, or "nothing"
+# 5FA2 = Spread (Homing Laser, 5FA0)
+# 5F9B = "nothing" (Diamond Rain, 5FA7)
+# 5FA5 = Stack (Diamond Flash, 5FA1)
+# This mechanic repeats in phase 3 with additional bits and an extra tracking If circle aoe
+# @TODO: Someone better at timelines than me, sync these to mini-phases instead
+36.5 "Adamant Purge" sync /:The Diamond Weapon:5F(9[ABCD]|A[2345]):/ window 20,5
+46.7 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+49.9 "Spread/Stack?" sync /:The Diamond Weapon:5FA[01]:/
```
quisquous, line 23: I think in general ability names are preferred, and folks can replace with spread/stack in their timelineReplace if they want.  So, `"Homing Laser/Diamond Flash"` instead here.  Also, as a convention, `?` is only used when something may or may not happen.  If it's one or the other, then it's just `A/B` with no `?`.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+104.8 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/
+120.0 "Adamant Purge" sync /:The Diamond Weapon:5F(9[ABCD]|A[2345]):/
+130.2 "Claw Swipe" sync /:The Diamond Weapon:5F9[EF]:/
+133.7 "Spread/Stack?" sync /:The Diamond Weapon:5FA[01]:/
+151.4 "Photon Burst" sync /:The Diamond Weapon:5FA8:/
+
+# Phase transition
+167.7 "Code Chi-Xi-Stigma" sync /:The Diamond Weapon:5FAD:/
+183.3 "Auri Cyclone" sync /:The Diamond Weapon:5FB9:/
+184.3 "Auri Cyclone" sync /:The Diamond Weapon:5FD1:/
```
quisquous, line 37: You have 5FB9 from the boss on itself, and 5FD1 and 5FD2 are the two jumps.  I think I'd write this as just the 5FB9 line and ignore 5FD1 everywhere.  Or, if you felt like the two separate hits were important, ignore 5FB9 and include 5FD1 and 5FD2

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
+### Diamond Weapon Ex
+
+hideall "--Reset--"
+hideall "--sync--"
+
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+# -ii 5FEB 5FBE 5FAA 5FEC 5FBA 5FFD 5FBB 6055 6159 5FF9 5FCD 5FFA 5FF9 612E 5FB3 5FCE 5FD2 5FFE 5FD3 5FB4 5FB6 5FB1 5FB0 5FBF 5FAE 5FA6 5FC3 5FC9 5FCA 5FC4
```
quisquous, line 8: Can you also run this with `-it "Diamond Weapon"` so it includes the targetable/untargetable lines?

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+0 "Start" sync /Engage!/ window 1,1
+21.3 "Diamond Rain" sync /:The Diamond Weapon:5FA7:/ window 25,10
+
+# Adamant Purge is a mini-phase that will be either stack, spread, or "nothing"
+# 5FA2 = Spread (Homing Laser, 5FA0)
+# 5F9B = "nothing" (Diamond Rain, 5FA7)
+# 5FA5 = Stack (Diamond Flash, 5FA1)
+# This mechanic repeats in phase 3 with additional bits and an extra tracking If circle aoe
+# @TODO: Someone better at timelines than me, sync these to mini-phases instead
+36.5 "Adamant Purge" sync /:The Diamond Weapon:5F(9[ABCD]|A[2345]):/ window 20,5
```
quisquous, line 21: This doesn't correspond to the comment about ability ids above.  Is it really all of these possibilities?

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+210.0 "Outrage" sync /:The Diamond Weapon:5FBC:/
+218.2 "Outrage" sync /:The Diamond Weapon:5FBC:/
+229.6 "Auri Arts" sync /:The Diamond Weapon:5FAF:/
+236.4 "Auri Arts" sync /:The Diamond Weapon:5FB2:/
+250.4 "Outrage" sync /:The Diamond Weapon:5FBC:/
+257.6 "Auri Doomstead" sync /:The Diamond Weapon:5FBD:/
+271.8 "Auri Arts" sync /:The Diamond Weapon:5FB5:/
+272.8 "Auri Arts" sync /:The Diamond Weapon:5FCF:/
+277.0 "Vertical Cleave" sync /:The Diamond Weapon:5FB7:/
+278.7 "Vertical Cleave" sync /:The Diamond Weapon:5FD0:/
```
quisquous, line 50: I think it's a little confusing to have two ability names in a row that do different things.  I would suggest either (1) dropping one of them (5FB7 is the cast, 5FD0 is the knockback) or (2) renaming 5FD0 to `--knockback--`.  I know I suggested using ability names above, but I think a double ability name is one place to ignore that rule.

Similarly, the 5FB5/5FCF Auri Arts above, which are both part of the same mechanic from the player's perspective.  (Does that always happen, since there's an optional extra jump sometimes on that one?)

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+
+### Phase 3
+422.7 "Flood Ray 1" sync /:The Diamond Weapon:5FC7:/
+424.2 "Flood Ray 2" sync /:The Diamond Weapon:5FC7:/
+425.7 "Flood Ray 3" sync /:The Diamond Weapon:5FC7:/
+427.2 "Flood Ray 4" sync /:The Diamond Weapon:5FC7:/
+428.7 "Flood Ray 5" sync /:The Diamond Weapon:5FC7:/
+430.2 "Flood Ray 6" sync /:The Diamond Weapon:5FC7:/
+431.7 "Flood Ray 7" sync /:The Diamond Weapon:5FC7:/
+433.2 "Flood Ray 8" sync /:The Diamond Weapon:5FC7:/
```
quisquous, line 76: I keep meaning to write a timeline test for this, but syncs have a -2.5,2.5 window by default, and so these all overlap with each other.  My rule is that anything with the same sync that's not more than 3-4 seconds apart should just have the sync commented out, e.g. `422.7 "Flood Ray 1" #2737 sync /:The Diamond Weapon:5FC7:/`

If you don't want to include the 5FA6 Flood Ray ability from the boss on the boss (which I think makes sense), it might be good to include the 5FA6 starts casting line to make sure these sync up if they can't have syncs themselves.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,116 @@
...
+360.7 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+362.2 "Auri Arts" sync /:The Diamond Weapon:5FB5:/
+364.0 "Auri Arts" sync /:The Diamond Weapon:5FCF:/
+368.2 "Vertical Cleave" sync /:The Diamond Weapon:5FB7:/
+369.9 "Vertical Cleave" sync /:The Diamond Weapon:5FD0:/
+370.7 "Aetherial Bullet" sync /:Articulated Bit:5FAB:/
+382.4 "Outrage" sync /:The Diamond Weapon:5FBC:/
+
+### Phase 3
+422.7 "Flood Ray 1" sync /:The Diamond Weapon:5FC7:/
```
quisquous, line 69: It looks like you're skipping the tank lasers here and elsewhere.  Including every hit of them is maybe too noisy, but maybe there could be one line with a duration where they happen?  That would probably help tanks plan their mitigation around them better when looking at the timeline.

quisquous (commented):
Thanks for this!!  This is really good.  I have a bunch of tiny comments.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
Akurosia: can confirm i expirienced always the same that 2. claw doe snot havespread/stack but it has the orbs (where the grous need to be split up on the sides)

quisquous:
Also, this is a lot of comments, so feel free to turn any of these suggestions into TODOs and we can followup later.

valarnin:
Thanks for the feedback. I'll clean up the easier to clean up stuff, and turn the rest into TODOs.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
valarnin: The tank lasers are autos, and I've seen some inconsistencies in timing elsewhere in the fight, that's why I excluded them. Inconsistencies might just be jitter due to my east-coast ping though.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
valarnin: The Adamant Purge for Diamond Rain can be either 5F9A or 5F9B depending on cleave side (see the triggers PR), but otherwise now that I can more easily check the actual mechs without manually matching up events with movements, I can confirm that across about 50 pulls it's always stack/spread, then aoe, then stack/spread in both phases. I have updated the timeline to reflect this.

valarnin:
I think I've actually covered everything here, if you could re-review to make sure.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
valarnin: I've added what I think you meant, but not 100% sure. Can't get in-game to test it and haven't tracked down the raidemulator sync issue for the timeline yet.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,123 @@
...
+hideall "--Reset--"
+hideall "--sync--"
+
+0.0 "--Reset--" sync / 21:........:40000010:/ window 10000 jump 0
+
+# -it "The Diamond Weapon" -ii 5FEB 5FBE 5FAA 5FEC 5FBA 5FFD 5FBB 6055 6159 5FF9 5FCD 5FFA 5FF9 612E 5FB3 5FCE 5FD2 5FFE 5FD3 5FB4 5FB6 5FB1 5FB0 5FBF 5FAE 5FA6 5FC3 5FC9 5FCA 5FC4
+
+### Phase 1
+0 "--sync--" sync /:5FA7:The Diamond Weapon/
+0 "Start" sync /Engage!/ window 1,1
```
quisquous, line 12:
```suggestion
0 "Start" sync /Engage!/ window 1,1
16.3 "--sync--" sync /:5FA7:The Diamond Weapon/ window 17,10
```

It's a 5 second cast, so I was suggesting to sync starting here.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
```diff
@@ -0,0 +1,123 @@
...
+210.0 "Outrage" sync /:The Diamond Weapon:5FBC:/
+218.2 "Outrage" sync /:The Diamond Weapon:5FBC:/
+229.6 "Auri Arts (Z)" sync /:The Diamond Weapon:5FAF:/
+229.6 "--untargetable--"
+236.4 "Auri Arts (Jump)" sync /:The Diamond Weapon:5FB2:/
+241.0 "--targetable--"
+250.4 "Outrage" sync /:The Diamond Weapon:5FBC:/
+257.6 "Auri Doomstead" sync /:The Diamond Weapon:5FBD:/
+265.5 "--untargetable--"
+271.8 "Auri Arts (Cleave)?" sync /:The Diamond Weapon:5FB5:/
```
quisquous, line 47:
```suggestion
271.8 "Auri Arts (Cleave)" sync /:The Diamond Weapon:5FB5:/
```

I'm not sure I understand why these have a question mark.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
quisquous: No worries.  I think it's something that can be added with a TODO.  Even if it's not precise and there aren't syncs, it'd be good to add those lines in eventually.

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
quisquous: `test_timeline.py` is usually the tool I go to to make sure timelines look good across a variety of logs.  raidemulator is good for making sure triggers all work correctly, but `test_timeline.py` helps with finding a lot of timeline-specific issues.

JLGarber (approved):

ui/raidboss/data/05-shb/trial/diamond_weapon-ex.txt
valarnin: If it's a two-orb pattern, the boss does a vertical cleave through the platform before the circle aoe. If it's one-orb, the vertical cleave doesn't happen.

quisquous: Makes sense, thanks!

PR #2738, opened 2021-04-16
raidboss: palgthan fr

VanSlanzar:
added missing translations fr


quisquous:
Could you put the fr keys after the de keys?

VanSlanzar:
That good.

PR #2740, opened 2021-04-17
test: make outputStrings the last field

quisquous:


quisquous:
Feel free to disagree with this.  I forgot to add this to the ordering test when I added outputStrings, so I am enforcing it now.

I think it's not clear where outputStrings should go.  The fields are ordered in their order of evaluation to try to make logical sense, but outputStrings are evaluated in multiple places (or arguably are plain old data, so it doesn't matter).  I think in general they're pretty wordy, so I thought it made sense to put them last so you could more easily see all the trigger functions in one place.

(While you're here being opinionated, I also was curious if it made sense to move the `supressSeconds` evaluation up before condition to make triggers more readable.  For triggers that are not suppressed, `suppressSeconds` gets evaluated along with `delaySeconds`, but for triggers that are suppressed, `suppressSeconds` takes effect before `condition`.  Moving it there puts the evaluation / takes effect into the same place.  This is maybe in line with `delaySeconds` where the delay happens after `delaySeconds` in terms of trigger ordering.)

panicstevenson:
I think that I would argue that `outputStrings` should be above `run`, since `run` can't do anything that interacts with `outputStrings` AND it's the last thing that happens in the trigger itself. While I kind of agree that `outputStrings` are lengthier, I don't think that _has_ to be true, and there are plenty of long `run`s (Unending Coil dragons say hi). But, I do see where you're coming from, and I could be convinced that average lines is the deciding factor, and `outputStrings` definitely wins.

Point 2 I'm relatively indifferent on; I think maybe because I'm just used to seeing all the `xSeconds` in a block together for so long. I see the logic behind moving it, and I would be in favor of it. I think I am also fine with it staying where it is. That being said, it might "stand out" more if it's moved up, making it more obvious why the trigger didn't fire when debugging (I don't know how often this happens? I assume... not).

JLGarber:
I think I come down on the side of "output last", all things considered. While I agree that `run` has no interactions with `output`, I think putting the function elements together makes more sense. Furthermore, precisely *because* of the potential length of `output` in general, it's easier to look at the trigger as a whole if it's at the end. (Obviously we have syntax highlighting, but it's easier to miss a `run: (data) => delete data.foo` hanging on at the end than it would be to miss an entire `outputStrings` object.) Finally, `output` basically doesn't get evaluated in the same sense other properties do, so it should be at the end out of the way.

As for `suppressSeconds`, let's move it to its evaluation point. I agree with Panic that it's nice to have all the time-related properties close together, but I think having the evaluation order match is more important.

quisquous:
To be clear on `suppressSeconds`, it *is* at its evaluation point already, technically.  I guess I'm arguing to move its evaluation point to where we also check for suppressed triggers.  (The point where we check if a trigger is suppressed does not re-evaluate the `suppressSeconds` field, is what I'm trying to say.)

PR #2742, opened 2021-04-17
raidboss: diamond weapon fr

VanSlanzar:
added missing translations fr

`'missingTranslations': true,` Are there any lines that will be added or is it no longer useful?


quisquous:
Re: missing translations.  The timeline will be added, so it probably makes sense to mark things as missing for now, because as soon as the timeline is added it'll need to be marked as missing translations.

PR #2743, opened 2021-04-17
raidboss: Correct Arrays That Should Be Maps

panicstevenson:
Fix incorrect instances of arrays that should be maps, given their usage
throughout the rest of the trigger file. This was not done exhaustively,
searching for instances of arrays within raidboss and ensuring that they
are not accessed in unexpected ways.


quisquous:
Typescript will catch the rest!

Issue #2543, opened 2021-03-06
lint-staged tsc doesn't respect tsconfig.json


quisquous:
Repro steps:
1) make a change to resources/party.ts
2) git commit -a -m errors

```
× tsc --noEmit:
resources/party.ts(55,7): error TS1056: Accessors are only available when targeting E
CMAScript 5 and higher.
resources/party.ts(59,7): error TS1056: Accessors are only available when targeting E
CMAScript 5 and higher.
resources/party.ts(64,7): error TS1056: Accessors are only available when targeting E
CMAScript 5 and higher.
resources/party.ts(69,7): error TS1056: Accessors are only available when targeting E
CMAScript 5 and higher.
resources/party.ts(74,7): error TS1056: Accessors are only available when targeting E
CMAScript 5 and higher.
resources/party.ts(79,7): error TS1056: Accessors are only available when targeting E
CMAScript 5 and higher.
```

This is because in package.json:
```
  "lint-staged": {
    "*.ts": "tsc --noEmit",
  },
```

`tsc` when provided files does not respect the tsconfig.json file.  If you provide one with `--project` you cannot provide input files.  There is no way to provide a tsconfig.json file here.  See: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html

We have a lot of options set, and I don't want to duplicate those onto the command line.

Some possible workarounds in these issues (broken to avoid needless linking, sorry):
1) https://github.com/microsoft/TypeScript/issues/ 27379
2) https://github.com/okonet/lint-staged/issues/ 468

quisquous:
cc @Trim21 @MaikoTan

Trim21:
As we mentioned at https://github.com/quisquous/cactbot/pull/2532/files/b97400d51d6f4802dfde7b28945970e864ae49fb#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

quisquous:
Ah sorry, I missed that hiding in the resolved comment.

I am not sure about running tsc on everything on every commit.  lint-staged is already a little bit slow and I don't want it to get worse.  On my computer, editing a comment in one trigger file and then doing this:

```
$ time git commit -a -m temp

[...output elided...]

real    0m9.665s
user    0m0.000s
sys     0m0.015s
```

```
$ time node_modules/.bin/tsc --noEmit

real    0m5.284s
user    0m0.030s
sys     0m0.105s
```

[...]

Issue #2739, opened 2021-04-17
eslint key ordering warning shows false positive warnings


quisquous:
See: https://github.com/quisquous/cactbot/pull/2736/files#diff-6e4e9ad12c6f2792e12247c8aa2ab1613ac2fed50a678684beced2721211c552

cc @MaikoTan

quisquous:
Ah, this is a real warning, it's just not present in that tree that the PR is on top of.

quisquous:
I wonder if this should be an error instead?

MaikoTan:
If it were set as an error, then users should always order the keys before the PR merged.
I think both okay, though.

PR #2744, opened 2021-04-17
lint: Make lint-staged Respect TypeScript Config

panicstevenson:
Make lint-staged respect the TypeScript configuration for the
repository. Additionally, change ESLint to fail on any warnings, since
warnings are useful distinctions from errors within an IDE, but no
warnings should make it through the commit at this time.

Fixes https://github.com/quisquous/cactbot/issues/2543


package.json
```diff
@@ -64,7 +65,8 @@
   },
   "lint-staged": {
     "*.css": "stylelint",
-    "*.{js,cjs,mjs,ts}": "eslint",
+    "*.{js,cjs,mjs}": "eslint --max-warnings=0",
```
quisquous, line 68: Is max warnings something that can (or should?) be in the eslint config itself?

panicstevenson:
Also linking related discussion: https://github.com/quisquous/cactbot/issues/2739 for rationale regarding why `--max-warnings=0` was added.

package.json
quisquous: It looks the answer is that this only CLI :C

PR #2747, opened 2021-04-17
various: Fix typo of 'itt' vs 'it'

valarnin:
Noticed when running `npm test` that one of the final tests had a typo. Ran a search on the code and corrected all instances of `itt` vs `it`, even though the rest of them were in comments.

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -80,32 +80,32 @@ export default class PartyTracker {
     return this.roleToPartyNames_['dps'];
   }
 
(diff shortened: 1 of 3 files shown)


quisquous:
`iff` is intentional, sorry: https://en.wikipedia.org/wiki/If_and_only_if

valarnin:
I've never seen the term before now. Thanks for the link.

Also the irony that I typo'd the commit and comment about the typo...

quisquous:
😂

PR #2748, opened 2021-04-17
timerbar: fix #2707

### LINKED ISSUE #2707: timerbar `toward` attribute does not work
You can see this in the `resources/timerbar_test.html` test file.  Regardless of the `toward` direction, the bar animates the same way.  I suspect that this was broken by cc7c6762cde67cc312690b5a79eb7da3c83198ff and nobody ever noticed because it's not being used.

valarnin:
Fix #2707


resources/timerbar.ts
```diff
@@ -419,10 +416,13 @@ export default class TimerBar extends HTMLElement {
 
     // To start full and animate to empty, we animate backwards and flip
     // the direction.
-    if (this._towardRight !== this._fill)
-      this.foregroundElement.style.transformOrigin = '100% 0%';
-    else
-      this.foregroundElement.style.transformOrigin = '0% 0%';
+    if (this._towardRight) {
```
quisquous, line 419:
```suggestion
    if (this._towardRight !== this._fill) {
```

I think if it says "fill towards left" that it should be filling from right to left.

If you checkout the revision before cc7c6762, that's what it used to do:

image

It makes sense to me that the empty and fill towards left would be aligned but have their colors reversed.

quisquous:
Also, thanks for sending this bugfix! I played with the CSS some and couldn't quite figure out how to make it align right properly, so this is much appreciated.

valarnin:
Playing around with some options, give me a few minutes.

valarnin:
Alright, I give up. There's a visual "jitter" when it's right-aligned but I tried 4 or 5 things to fix it and nothing really worked. Included your fix for the fill in latest commit. Should be good now.

PR #2756, opened 2021-04-18
Add TypeScript Support for Utils and Tests

panicstevenson:


test/helper/test_manifest.js
```diff
@@ -51,7 +51,7 @@ const testManifestFile = (file) => {
   it('correct file types in manifest', () => {
     for (const file of manifestLines) {
       // All files must be .js and .txt.
-      if (/\.txt$/.test(file) || /\.js$/.test(file))
+      if (/\.txt$/.test(file) || /\.[jt]s$/.test(file))
```
quisquous, line 54: Can you fix the comment above this line (or remove it)?

test/helper/test_timeline.js
```diff
@@ -23,16 +23,18 @@ const setup = (timelineFiles) => {
   timelineFiles.forEach((timelineFile) => {
     // For each timeline file, ensure that its corresponding trigger file is pointing to it.
     const filename = timelineFile.split('/').slice(-1)[0];
-    const triggerFilename = timelineFile.replace('.txt', '.js');
-    if (!fs.statSync(triggerFilename))
+    const triggerFilenameJS = timelineFile.replace('.txt', '.js');
+    const triggerFilenameTS = timelineFile.replace('.txt', '.ts');
+    if (!fs.existsSync(triggerFilenameJS) && !fs.existsSync(triggerFilenameTS))
```
quisquous, line 28: Can you store the results of existsSync in a variable? I know it's premature optimization, but checking if a file exists on disk is probably the most expensive operation here.

test/helper/test_timeline.js
```diff
@@ -99,7 +101,8 @@ const testTimelineFiles = (timelineFiles) => {
         let timeline;
 
         before(async () => {
-          const importPath = '../../' + path.relative(process.cwd(), triggersFile).replace(/\\/g, '/');
+          // Normalize path
+          const importPath = '../../' + path.relative(process.cwd(), triggersFile).replace(/\\/g, '/').replace('.ts', '.js');
```
quisquous, line 105: Also as I'm looking at this, is this a case for `path.posix.relative` and then not needing the replacing backslashes with slashes?

PR #2760, opened 2021-04-18
raidboss: Add Trigger Source to Config

panicstevenson:
Add the ability to view the `triggers` branch from within the config
UI.

Ideally, the link would open in Chrome to take advantage of text
highlighting, but that functionality doesn't exist yet.

### DIFF
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -579,6 +579,27 @@ class RaidbossConfigurator {
 
           triggerDetails.appendChild(div);
         }
+
+        const label = document.createElement('div');
+        triggerDetails.appendChild(label);
+
+        const div = document.createElement('div');
+        div.classList.add('option-input-container', 'trigger-source');
+        const baseUrl = 'https://github.com/quisquous/cactbot/blob/triggers';
+        const path = key.split('-');
+        let urlFilepath;
+        if (path.length === 3) {
+          // 00-misc/general.js
+          urlFilepath = `${path[0]}-${path[1]}/${[...path].slice(2).join('-')}`;
+        } else {
+          // 02-arr/raids/t1.js
+          urlFilepath = `${path[0]}-${path[1]}/${path[2]}/${[...path].slice(3).join('-')}`;
+        }
+        const uriComponent = encodeURIComponent(trig.id).replace(/\'/g, '%5C%27');
+        const urlString = `${baseUrl}/${urlFilepath}.js#:~:text=${uriComponent}`;
+        div.innerHTML = `<a href="${urlString}" target="_blank">(View Trigger Source)</a>`;
+
+        triggerDetails.appendChild(div);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


ui/raidboss/raidboss_config.js
```diff
@@ -579,6 +579,27 @@ class RaidbossConfigurator {
...
+        const path = key.split('-');
+        let urlFilepath;
+        if (path.length === 3) {
+          // 00-misc/general.js
+          urlFilepath = `${path[0]}-${path[1]}/${[...path].slice(2).join('-')}`;
+        } else {
+          // 02-arr/raids/t1.js
+          urlFilepath = `${path[0]}-${path[1]}/${path[2]}/${[...path].slice(3).join('-')}`;
+        }
+        const uriComponent = encodeURIComponent(trig.id).replace(/\'/g, '%5C%27');
```
quisquous, line 598: I do think trigger ids are referenced by name sometimes in a file, so probably this needs to be more like `id: ${trig.id}`?

quisquous (commented):
Thank you for this! It is sad that the there's no text fragment support here.

Do you think it's worth the effort to run something like https://github.com/quisquous/cactbot/blob/main/util/find_missing_timeline_translations.js#L59-L79 and jump to a line number (which will probably be ~close modulo differences between last release and current main branch)? Or is the file enough for now?

ui/raidboss/raidboss_config.js
panicstevenson: > Do you think it's worth the effort to run something like https://github.com/quisquous/cactbot/blob/main/util/find_missing_timeline_translations.js#L59-L79 and jump to a line number (which will probably be ~close modulo differences between last release and current main branch)? Or is the file enough for now?

I personally think it's fine since I think the adoption of this would be rather low to start, and hopefully by the time adoption ramps up it would either open in Chrome directly or CEF would be updated, but I would agree that jumping to a line number would be a better user experience in the interim, so I'd be happy to make the change if you thought it was worthwhile.

panicstevenson:
> Do you think it's worth the effort to run something like https://github.com/quisquous/cactbot/blob/main/util/find_missing_timeline_translations.js#L59-L79 and jump to a line number (which will probably be ~close modulo differences between last release and current main branch)? Or is the file enough for now?

I personally think it's fine since I think the adoption of this would be rather low to start, and hopefully by the time adoption ramps up it would either open in Chrome directly or CEF would be updated, but I would agree that jumping to a line number would be a better user experience in the interim, so I'd be happy to make the change if you thought it was worthwhile.

quisquous:
> > Do you think it's worth the effort to run something like https://github.com/quisquous/cactbot/blob/main/util/find_missing_timeline_translations.js#L59-L79 and jump to a line number (which will probably be ~close modulo differences between last release and current main branch)? Or is the file enough for now?
> 
> I personally think it's fine since I think the adoption of this would be rather low to start, and hopefully by the time adoption ramps up it would either open in Chrome directly or CEF would be updated, but I would agree that jumping to a line number would be a better user experience in the interim, so I'd be happy to make the change if you thought it was worthwhile.

Ok, file is enough for now then.  ✅

quisquous:
cc @ngld for another thing that an updated CEF version would help with

PR #2761, opened 2021-04-18
refactor: merge tsconfig

Trim21:


quisquous:
Thanks! This is a tiny nit, but can we have the filenames match, e.g. `tsconfig.base.json` and `tsconfig.triggers.json` OR `tsconfig_base.json` and `tsconfig_triggers.json`, but not a mix of underscore and period and jsonc and json?

Trim21:
ok

quisquous:
Thanks!

PR #2762, opened 2021-04-18
typo in type definition

Trim21:
#2611 @MaikoTan 

```js
// global.d.ts
export type NonEnLang = Exclude<Lang, 'en'>;
// trigger.d.ts
import { Lang, NotEnLang } from './global';
```


MaikoTan:
Oops!

PR #2764, opened 2021-04-18
Revert "refactor: merge tsconfig"

quisquous:
Reverts quisquous/cactbot#2761


quisquous:
Sorry @Trim21, this breaks `npm run process-triggers` (which was why there was a separate tsconfig in the first place).  I didn't look at the details, but reverting this patch makes that run again.

PR #2765, opened 2021-04-18
raidboss: Diamond Ex headmarkers

valarnin:
Adding headmarkers for LC/Flood Ray mechanic.

Haven't had a chance to test in-zone but raidemulator testing shows that they're correct over three days of pulls/clears.


valarnin:
I've confirmed that the headmarker trigger works properly in-zone now.

quisquous:
Thanks!!

PR #2768, opened 2021-04-18
test: Remove Posix Relative Path

panicstevenson:
Partially reverts https://github.com/quisquous/cactbot/pull/2760

### DIFF
--- a/test/helper/test_timeline.js
+++ b/test/helper/test_timeline.js
@@ -108,7 +108,7 @@ const testTimelineFiles = (timelineFiles) => {
 
         before(async () => {
           // Normalize path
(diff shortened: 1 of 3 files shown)


quisquous:
I can't remember why the backslash replacing was there in the first place.  Did you end up testing this in `cmd`?

panicstevenson:
> I can't remember why the backslash replacing was there in the first place. Did you end up testing this in `cmd`?

Tested in both Git Bash and Command Prompt on Windows 10, as well as whatever the CI is using (bash on Linux?)

quisquous:
Thanks for the fix.  Sorry for the awful suggestion that caused this.  <_<

PR #2773, opened 2021-04-19
raidboss: the_puppets_bunker fr

VanSlanzar:
- added missing translations fr 
- repair 519B
`6166.6 "Three Parts Disdain" sync /:Compound 2P:519B:/ duration 3.6`
`6204.6 "Four Parts Resolve" sync /:Compound 2P:519E:/ duration 9.4`

I did not find `Three Parts Resolve` and **519B** is `Three Parts Disdain`.

Tell me if I made a mistake.


Akurosia:
I think this just means that Three parts disdained has ended, as it has multiple hits

VanSlanzar:
I will wait to see @quisquous answer.

@Akurosia while you're there here you should be looking if you don't have the same poster issue on the timeline as I did.
`6344.7 "Klingensequenz (Out/Behind)" sync /:2P: Fusion:519[89]:/`

I had this problem in fr too. I had to create all the lines so as not to have it.

Akurosia:
yeah i have to go through all timelines again with my new version of the script to check all old ones^^ but thanks for letting me know

quisquous:
Yeah, I think this is just a typo on my part, mixing up Four Parts Resolves with Three Parts Disdain.  Thanks for catching this!

PR #2781, opened 2021-04-23
raidboss: Convert Browser TTS Engine to TypeScript

panicstevenson:


ui/raidboss/browser_tts_engine.ts
```diff
@@ -21,28 +25,44 @@ class SpeechTTSItem extends TTSItem {
...
 }
 
-class GoogleTTSItem extends TTSItem {
-  constructor(text, lang) {
-    super();
-    this.lang = lang;
+class GoogleTTSItem implements TTSItem {
+  readonly text: string;
+  readonly lang: string;
+  private item?: Element | null;
```
quisquous, line 31: (just for my own knowledge, does the `?` already imply can be null, or do you also have to say null?)

ui/raidboss/browser_tts_engine.ts
```diff
@@ -96,13 +111,17 @@ export default class BrowserTTSEngine {
     }
   }
 
-  playSpeechTTS(text) {
-    this.ttsItems[text] = new SpeechTTSItem(text, this.speechLang, this.speechVoice);
-    this.ttsItems[text].play();
+  playSpeechTTS(text: string): void {
+    if (this.speechLang && this.speechVoice) {
+      const ttsItem = new SpeechTTSItem(text, this.speechLang, this.speechVoice);
```
quisquous, line 116: Up above, speechLang and speechVoice are optional, but here they're required.  What happens if these are undefined?

ui/raidboss/browser_tts_engine.ts
quisquous: Or, looking at this again, when can it be null? Should it just be `private item?: Element`?

ui/raidboss/browser_tts_engine.ts
panicstevenson: That's my fault for not updating the constructor; the constructor can be made to take lang and voice as optional parameters, and this check isn't necessary.

ui/raidboss/browser_tts_engine.ts
panicstevenson: `?` implies `undefined`; in this case, `Document.createElement` returns `HTMLIFrameElement | null`. Removing the `| null` will return:
```
Type 'Element | null' is not assignable to type 'HTMLIFrameElement | undefined'.
  Type 'null' is not assignable to type 'HTMLIFrameElement | undefined'
  ```

PR #2784, opened 2021-04-24
util: Convert scripts to emit typescript resources

quisquous:
The scripts themselves haven't been otherwise updated.


quisquous:
Clearly these will need to have *actual* types when anything wants to use them, but I think this is a good first step.

panicstevenson (approved):

PR #2785, opened 2021-04-24
resources: convert weather.js to typescript

quisquous:


resources/weather.ts
```diff
@@ -1,24 +1,26 @@
...
 import ZoneInfo from './zone_info';
 
-export function getWeather(timeMs, zoneId) {
+export const getWeather = (timeMs: number, zoneId: number): string | undefined => {
   const chance = getWeatherChanceValue(timeMs);
 
   // See weather_rate.js and territory_type.js for details.
-  const rateIdx = ZoneInfo[zoneId].weatherRate;
-  const entry = gWeatherRates[rateIdx];
+  const rateIdx = ZoneInfo[zoneId]?.weatherRate;
```
panicstevenson, line 8: Should the comment above this be updated?

panicstevenson (approved):

resources/weather.ts
quisquous: Good idea, thanks!

PR #2787, opened 2021-04-24
resources: convert content_list.js to typescript

quisquous:


panicstevenson (approved):

resources/content_list.ts
```diff
@@ -6,7 +6,7 @@ import ZoneId from './zone_id';
 // These are not things that cactbot necessarily supports, but things that it
 // theoretically could be supporting in the future with raidboss and oopsy.
 
-export default [
+const contentList: (number | null)[] = [
```
panicstevenson, line 9: Should ZoneId.MatchAll return like -1 instead so we don't have to check null?

quisquous: It could, but I'd want to change that in a different PR.  I can leave a TODO if you think that'd be better.

panicstevenson: Up to you; I don't think it's necessary. We can get to it whenever, if ever

PR #2788, opened 2021-04-24
util: add types to generated resource files

quisquous:


panicstevenson (approved):

Trim21:
Why don't we just type it `as const`?

quisquous:
No, you're right.  I think things that are just simple maps or lists should just be `as const` so that there's no "maybe this property is undefined".  But, I think for things that are more complicated structured data like hunts, those need a real type.  I'll fix this up.

PR #2792, opened 2021-04-24
responses: stop handling unnamed matches

quisquous:


quisquous:
Once upon a time, we didn't have fancy named capture groups, and so the target/source could have been matches[0] and matches[1].  The only worry here is that somebody has some user trigger with a bare regex that is also using responses that is somehow relying on this behavior.  That seems...extremely unlikely however.

JLGarber:
I think the "bare regex but using responses with this behavior" group is a small enough portion of the userbase that it's fine. Once it's release time, maybe we could have a line in the release notes that we can point to if users actually experience breakages.

PR #2797, opened 2021-04-25
eslint: add typescript consistent-type-assertions rule

quisquous:


resources/party.ts
```diff
@@ -10,7 +19,7 @@ export default class PartyTracker {
   allianceIds_: string[] = [];
   nameToRole_: Record<string, Role> = {};
   idToName_: Record<string, string> = {};
-  roleToPartyNames_: Record<Role, string[]> = {} as Record<Role, string[]>;
+  roleToPartyNames_: Record<Role, string[]> = emptyRoleToPartyNames;
```
Trim21, line 22: Won't this object be shared cross `PartyTracker` instance?

quisquous: Oh whoops, thanks!

Trim21: I'm not sure if it works after your changes...

Copy and original object share array, too.

quisquous: Thanks for catching my (repeated) mistakes.  *<_<*

panicstevenson (approved):

Trim21 (approved):

PR #2798, opened 2021-04-26
Responses: combineFuncs will now combine identical severity levels

### LINKED ISSUE #2793: Resources: Selecting info severity level for the tankBuster() response fails
I came across this one while doing some work on Lakshmi. `tankBuster()` works, and `tankBuster('alarm') works, but `tankBuster('info') fails. ([Sohm Al normal](https://github.com/quisquous/cactbot/blob/56e201238fccddb67f6005f3e3f6610c205c94b2/ui/raidboss/data/03-hw/dungeon/sohm_al.js) has one, although it's probably easier to reproduce by editing something else.)

There's nothing immediately obvious (to me, anyway) on why it's failing. I poked through `responses` a little and there doesn't *seem* to be any reason `info` fails while `alarm` doesn't.

JLGarber:
(This resolves #2793 )

Briefly, for those not immediately familiar with the internals of `Responses`, some response functions take two parameters, the "primary" severity, and the "secondary". (Usually these functions are assigned different default severity levels.) The user can then choose specific severity levels for the response in the trigger body. If the severity levels are different, the two different output functions are assigned to different severity levels in the trigger response object and it's then returned.

If the severity levels are the same, the response object is supposed to be a function that returns the correct response. However, a missing `return` meant that this function just evaluated `true` and exited. Since it's *incredibly* uncommon to assign non-default severity to `Responses`, we never noticed before.

To be clear, quisquous tracked down this problem, all I did was throw data at them.

### DIFF
--- a/resources/responses.js
+++ b/resources/responses.js
@@ -109,7 +109,7 @@ const combineFuncs = function(text1, func1, text2, func2) {
     obj[text2] = func2;
   } else {
     obj[text1] = (data, matches, output) => {
-      func1(data, matches, output) || func2(data, matches, output);
+      return func1(data, matches, output) || func2(data, matches, output);
     };
   }
   return obj;


quisquous:
Thanks for reporting and debugging and fixing!!

PR #2799, opened 2021-04-26
Raidboss: Emanation (Lakshmi normal mode) timeline and triggers

JLGarber:
I modified the trigger names in EX so as to avoid any possibility of collisions. I didn't *think* this was a problem since it's a necessary disambiguation, but I can revert those changes and submit them separately if necessary.

One of the patterns involving stack markers combined with other markers does technically give people enough time to stack, but it's probably best to keep them away so puddle people don't clip the stack. This is particularly important during Chanchala phases, since the puddles expand significantly.

There are a lot of abilities that have different IDs at different points in the timeline. This is because all her skills are empowered under Chanchala, and Square implements this by just using different skills.

This doesn't attempt to check for whether a non-tank has the cleave (`the Path of Light`,) so that we can warn them specifically to use Vril. It can be done, but the weekend is over and I won't have a lot of time to look at this until next weekend at the earliest.

### DIFF
--- a/ui/raidboss/data/04-sb/trial/lakshmi-ex.js
+++ b/ui/raidboss/data/04-sb/trial/lakshmi-ex.js
@@ -347,7 +347,6 @@ export default {
         'Lakshmi': 'Lakshmi',
       },
       'replaceText': {
-        '--Chanchala over--': '--Chanchala terminé--',
         '/dance': '/danse',
         '\\(mid\\)': '(milieu)',
         '\\(out\\)': '(extérieur)',
@@ -443,7 +442,6 @@ export default {
         'The Pall Of Light': '빛의 폭포',
         'The Path Of Light': '빛의 파동',
         'The Pull Of Light': '빛의 급류',
-        'over': '끝',
       },
     },
   ],
--- a/ui/raidboss/data/04-sb/trial/lakshmi-ex.js
+++ b/ui/raidboss/data/04-sb/trial/lakshmi-ex.js
@@ -9,7 +9,7 @@ export default {
   timelineFile: 'lakshmi-ex.txt',
   timelineTriggers: [
     {
-      id: 'Lakshmi EX Path of Light',
+      id: 'LakshmiEx Path of Light',
       regex: /Path of Light/,
       beforeSeconds: 5,
       condition: function(data) {
@@ -20,7 +20,7 @@ export default {
   ],
   triggers: [
     {
-      id: 'Lakshmi EX Chanchala Gain',
+      id: 'LakshmiEx Chanchala Gain',
       netRegex: NetRegexes.startsUsing({ id: '2148', source: 'Lakshmi', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ id: '2148', source: 'Lakshmi', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ id: '2148', source: 'Lakshmi', capture: false }),
@@ -32,7 +32,7 @@ export default {
       },
     },
     {
-      id: 'Lakshmi EX Chanchala Lose',
+      id: 'LakshmiEx Chanchala Lose',
       netRegex: NetRegexes.losesEffect({ target: 'Lakshmi', effectId: '582', capture: false }),
       netRegexDe: NetRegexes.losesEffect({ target: 'Lakshmi', effectId: '582', capture: false }),
       netRegexFr: NetRegexes.losesEffect({ target: 'Lakshmi', effectId: '582', capture: false }),
@@ -44,7 +44,7 @@ export default {
       },
     },
     {
-      id: 'Lakshmi EX Pull of Light Tank',
+      id: 'LakshmiEx Pull of Light Tank',
       netRegex: NetRegexes.startsUsing({ id: '215E', source: 'Lakshmi' }),
       netRegexDe: NetRegexes.startsUsing({ id: '215E', source: 'Lakshmi' }),
       netRegexFr: NetRegexes.startsUsing({ id: '215E', source: 'Lakshmi' }),
@@ -57,7 +57,7 @@ export default {
       response: Responses.tankBuster('info'),
     },
     {
-      id: 'Lakshmi EX Pull of Light Unexpected',
+      id: 'LakshmiEx Pull of Light Unexpected',
       netRegex: NetRegexes.startsUsing({ id: '215E', source: 'Lakshmi' }),
       netRegexDe: NetRegexes.startsUsing({ id: '215E', source: 'Lakshmi' }),
       netRegexFr: NetRegexes.startsUsing({ id: '215E', source: 'Lakshmi' }),
@@ -70,7 +70,7 @@ export default {
       response: Responses.tankBuster('alarm'),
     },
     {
-      id: 'Lakshmi EX Divine Denial',
+      id: 'LakshmiEx Divine Denial',
       netRegex: NetRegexes.startsUsing({ id: '2149', source: 'Lakshmi', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ id: '2149', source: 'Lakshmi', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ id: '2149', source: 'Lakshmi', capture: false }),
@@ -90,7 +90,7 @@ export default {
       },
     },
     {
-      id: 'Lakshmi EX Divine Desire',
+      id: 'LakshmiEx Divine Desire',
       netRegex: NetRegexes.startsUsing({ id: '214B', source: 'Lakshmi', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ id: '214B', source: 'Lakshmi', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ id: '214B', source: 'Lakshmi', capture: false }),
@@ -110,7 +110,7 @@ export default {
       },
     },
     {
-      id: 'Lakshmi EX Divine Doubt',
+      id: 'LakshmiEx Divine Doubt',
       netRegex: NetRegexes.startsUsing({ id: '214A', source: 'Lakshmi', capture: false }),
       netRegexDe: NetRegexes.startsUsing({ id: '214A', source: 'Lakshmi', capture: false }),
       netRegexFr: NetRegexes.startsUsing({ id: '214A', source: 'Lakshmi', capture: false }),
@@ -130,7 +130,7 @@ export default {
       },
     },
     { // Stack marker
-      id: 'Lakshmi EX Pall of Light',
+      id: 'LakshmiEx Pall of Light',
       netRegex: NetRegexes.headMarker({ id: '003E' }),
       alertText: function(data, matches, output) {
         if (!data.chanchala)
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/04-sb/trial/lakshmi.js
```diff
@@ -0,0 +1,338 @@
...
+      },
+      infoText: function(data, matches, output) {
+        if (data.avoidStack.includes(data.me))
+          return;
+        if (data.me === matches.target)
+          return output.stackOnYou();
+        return output.stack({ target: matches.target });
+      },
+      outputStrings: {
+        dontStack: {
```
panicstevenson, line 135: I don't know if I fully remember the fight, but I'm assuming based on the logic here that you will have two triggers at the same time, one telling you Cross Marker and the other saying not to stack? Does that feel like too much?

ui/raidboss/data/04-sb/trial/lakshmi.js
```diff
@@ -0,0 +1,338 @@
...
+          en: 'Stack on YOU',
+          de: 'Stack auf DIR',
+          fr: 'Package sur VOUS',
+          ja: '自分に頭割り',
+          cn: '分摊点名',
+          ko: '쉐어징 대상자',
+        },
+        stack: {
+          en: 'Stack on ${target}',
+        },
```
panicstevenson, line 148: I think this is the general pattern, just in case any future find and replace operations take place. (Would require changing `target: matches.target` above to `player: matches.target`)
```suggestion
        stackOn: {
          en: 'Stack on ${player}',
          de: 'Sammeln auf ${player}',
          fr: 'Packez-vous sur ${player}',
          ja: '${player}と頭割り',
          cn: '靠近${player}分摊',
          ko: '"${player}" 쉐어징',
        },
```

ui/raidboss/data/04-sb/trial/lakshmi.txt
```diff
@@ -0,0 +1,98 @@
...
+270.7 "The Path Of Light" sync /:Lakshmi:24A1:/ window 270.7,30
+284.9 "Hand Of Grace" sync /:Lakshmi:2486:/
+292.1 "The Pall Of Light" sync /:Lakshmi:2491:/
+299.1 "Blissful Arrow" sync /:Lakshmi:248A:/
+299.7 "Blissful Spear" sync /:Lakshmi:248C:/
+305.3 "Stotram" sync /:Lakshmi:249E:/
+307.6 "--vril spawn--"
+320.2 "Divine Denial" sync /:Lakshmi:2485:/ window 30,30
+320.2 "--vril despawn--"
+321.6 "--chanchala lose--"
```
panicstevenson, line 60: The extreme version uses `--Chanchala over--` I don't know if it makes sense to bring that over here as well

ui/raidboss/data/04-sb/trial/lakshmi.js
```diff
@@ -0,0 +1,338 @@
...
+          data.avoidStack.push(matches.target);
+      },
+    },
+    {
+      // Activating on any use of Hand of Beauty/Grace or Alluring Arm.
+      // Head markers don't appear until the end of the castbar,
+      // and every head marker section begins with one of these abilities,
+      // so this should be perfectly safe.
+      id: 'Lakshmi Headmarker Cleanup',
+      netRegex: NetRegexes.startsUsing({ id: ['2486', '2487', '2488'], source: 'Lakshmi', capture: false }),
```
panicstevenson, line 110:
```suggestion
      netRegex: NetRegexes.startsUsing({ id: '248[678]', source: 'Lakshmi', capture: false }),
```
Feels more readable in my opinion, but feel free to disagree here.

ui/raidboss/data/04-sb/trial/lakshmi-ex.js
```diff
@@ -9,7 +9,7 @@ export default {
   timelineFile: 'lakshmi-ex.txt',
   timelineTriggers: [
     {
-      id: 'Lakshmi Path of Light',
+      id: 'Lakshmi EX Path of Light',
```
panicstevenson, line 12: I think the pattern here is usually to include Ex as a suffix to the word, so LakshmiEx (or LakEx I see sometimes?)
I don't care here, commenting to find out if there's a rationale for it or if it doesn't matter at all.

panicstevenson (approved):

ui/raidboss/data/04-sb/trial/lakshmi-ex.js
quisquous: I think the only use of `EX` so far is `SophiaEX` (no space).  All of the other extremes are some variant of `SomethingEx`.  I don't think there's a rationale, but maybe it'd make sense to be consistent? I'm not sure why Susano, Byakko, Rathalos are all shortened other than I just didn't want long ids, but there's not a more consistent rationale.

ui/raidboss/data/04-sb/trial/lakshmi.js
quisquous: Not to muddy the waters, but I could go either way on this. The shorter version is more readable but the longer version is more searchable.

ui/raidboss/data/04-sb/trial/lakshmi.txt
```diff
@@ -0,0 +1,98 @@
...
+0 "Start" sync /Engage!/ window 0,1
+3.9 "Vril" sync /:Dreaming Kshatriya:2482:/ window 3.9,5
+13.0 "Stotram" sync /:Lakshmi:2483:/ window 13,5
+16.1 "Inner Demons" # sync /:Dreaming Kshatriya:258D:/
+19.0 "Aether Drain" sync /:Vril:248D:/
+29.3 "Tail Slap" # sync /:Dreaming Kshatriya:258C:/
+
+# We can't sync to abilities here to make the loop visible,
+# since two adds are present, and the timeline would jerk around.
+60.0 "--sync--" sync /.*/ window 0,30 jump 30
```
quisquous, line 21: Hmm.  This is a lot of timeline resyncing just to keep a loop going.  Is there a single ability that can loop back here, or would it make sense to just move `100.0` farther into the future to be `1000.0`?

One future feature that I think would be nice would be to have a pause/break sort of keyword where you could insert like `60.0 pause` or something and the timeline would pause there until it reached another sync and no abilities past that time would be shown past there.  We could use that for phase breaks everywhere.

ui/raidboss/data/04-sb/trial/lakshmi-ex.js
JLGarber: Fair point on the casing. I don't think we need to worry about the shortened forms of the names, but I'll definitely change it over to `LakshmiEx`.

ui/raidboss/data/04-sb/trial/lakshmi.js
JLGarber: Right, my general feeling here (weakly) is that we *have* regex helpers now, so it's a little self-defeating to then just go back to basic regex syntax. `grep/findstr` will find the array format easily enough with a regex-formatted command, but it's a little tougher to tell your search "oh, and what you're looking for *might* be string-coded regex", assuming you even know that any occurrences *are* such.

In the case of Diamond Weapon, for instance, where there are ridiculous numbers of abilities that all have to be synced on one line, the collapsed syntax makes sense, but here I weakly lean toward the expanded form.

ui/raidboss/data/04-sb/trial/lakshmi.txt
JLGarber: No, there's nothing usable. (The `/.*/` construction was borrowed from EX, which is admittedly *fairly* old.) The adds may or may not be in sync, depending on which tank they target and how far they have to travel to start auto-attacking. In this case it's likely best to just start the syncs from 1000.

ui/raidboss/data/04-sb/trial/lakshmi.txt
JLGarber: Maybe we could change both to the more precise `--Chanchala ends--`?

ui/raidboss/data/04-sb/trial/lakshmi.js
JLGarber: It sort of is, but the issue is that at least the circle markers have environmental effects when they resolve, and it's *really* easy for people to autopilot based on triggers. (I dislike it when people run from stack markers, but I get *really* frustrated when people with spread markers attempt to stack.)

The way I see it, each of the non-stack markers has a specific thing they need to *do*, and they should do that and not even attempt to stack. Adding "don't stack" doesn't always make sense since the stack marker isn't always up when the circle/cross/cone marker are, so we would then need to special-case each one based on whether or not there's a stack marker on the field. This then *further* means we need to do a collect/call setup for each marker, rather than just the stack marker.

Conversely, every time there's a stack marker, it's accompanied by *something* else, so it makes sense to warn those other people off from within the stack trigger, at least to me.

ui/raidboss/data/04-sb/trial/lakshmi.txt
quisquous: I think any of these options are fine, but having both files be consistent would be good.

JLGarber: Alright, then I'll make some changes to both to make it maybe a little more sensible. I do think `chanchala over` doesn't really convey what we want, and looking at it more, `chanchala gain` isn't good either.

(At some point I'll need to go through and redo the EX timeline as well, since it's from the days of manual timing. It looks great, but there aren't a lot of syncs.)

ui/raidboss/data/04-sb/trial/lakshmi.txt
```diff
@@ -0,0 +1,96 @@
...
+1393.2 "Blissful Spear" sync /:Lakshmi:2494:/
+1402.3 "The Pull Of Light" sync /:Lakshmi:2492:/
+1407.4 "--vril despawn--"
+1421.5 "Alluring Arm" sync /:Lakshmi:2488:/
+1430.9 "The Path Of Light" sync /:Lakshmi:248F:/ window 30,30
+1435.7 "Blissful Spear" sync /:Lakshmi:2494:/
+1440.0 "The Pall Of Light" sync /:Lakshmi:2490:/
+1452.2 "The Pull Of Light" sync /:Lakshmi:2492:/
+
+1465.5 "Chanchala" sync /:Lakshmi:2484:/ window 30,30 jump 365.0
```
quisquous, line 93:
```suggestion
1465.5 "Chanchala" sync /:Lakshmi:2484:/ window 30,30 jump 1365.0
```

ui/raidboss/data/04-sb/trial/lakshmi-ex.txt
```diff
@@ -40,7 +40,7 @@ hideall "Alluring Arm"
 225.5 "The Pull Of Light" # end of castbar
 228.5 "The Path Of Light" duration 5.5 # hat -> cleave
 246.0 "Divine Desire" # end of castbar
-249.5 "--Chanchala over--"
+249.5 "--chanchala end--"
```
quisquous, line 43: If you are going to change this, can you edit (remove?) the timeline replacements as well that are using this?

ui/raidboss/data/04-sb/trial/lakshmi.txt
JLGarber: Oops!

JLGarber:
Thanks for catching those. I think we're probably ready for this one to merge now. Oopsy will be forthcoming in [6 to 8 weeks](https://meta.stackexchange.com/a/19514).

PR #2801, opened 2021-04-27
resources: convert user_config.js to TypeScript

quisquous:
This implements a few things from config that will probably need
to be moved there.

### DIFF
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -95,10 +95,11 @@ class UserConfig {
     return keys.sort((keyA, keyB) => {
       const listA = splitKeyMap[keyA];
       const listB = splitKeyMap[keyB];
+      // Convince TypeScript these two exist.
       if (listA === undefined || listB === undefined)
-        return 0; // Convince TypeScript these exist.
-      const maxLen = Math.max(listA.length, listB.length);
+        return 0;
 
+      const maxLen = Math.max(listA.length, listB.length);
       for (let idx = 0; idx < maxLen; ++idx) {
         const entryA = listA[idx];
         const entryB = listB[idx];
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -1,3 +1,9 @@
+import { Lang } from '../types/global';
+import { BaseOptions } from '../types/data';
+import { CactbotLoadUserRet, SavedConfig, SavedConfigEntry } from '../types/event';
+import { LocaleText } from '../types/trigger';
+import { addOverlayListener, callOverlayHandler } from './overlay_plugin_api';
+
 // TODO:
 // The convention of "import X as _X; const X = _X;" is currently
 // being used as a method to workaround for downstream code
@@ -5,7 +11,6 @@
 // create a variable of the same name, the eval()'d code does not know
 // about the import, and thus throws ReferenceErrors.
 // Used by downstream eval
-import { addOverlayListener, callOverlayHandler } from './overlay_plugin_api';
 import _Conditions from './conditions';
 const Conditions = _Conditions;
 import _ContentType from './content_type';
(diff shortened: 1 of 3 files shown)


resources/user_config.ts
```diff
@@ -242,12 +298,11 @@ class UserConfig {
...
             }
 
-            if (this.userFileCallbacks[overlayName])
-              this.userFileCallbacks[overlayName](jsFile, localFiles, options, basePath);
+            this.userFileCallbacks[overlayName]?.(jsFile, localFiles, options, basePath);
           } catch (e) {
             // Be very visible for users.
             console.log('*** ERROR IN USER FILE ***');
-            console.log(e.stack);
+            console.log(e);
```
quisquous, line 305: This is the only (intended) logic change.  TypeScript doesn't really support error types very well and `console.log(e)` covers this just as well.

types/data.d.ts
```diff
@@ -2,10 +2,11 @@ import PartyTracker from '../resources/party.js';
 import { Lang } from './global';
 import { Job, Role } from './job';
 
-export interface Option {
+export interface BaseOptions {
```
quisquous, line 5: I also changed this to `BaseOptions` with the idea that we'd have `RaidbossOptions extends BaseOptions` later.

resources/user_config.ts
MaikoTan: What about `(e as Error).stack` ?

resources/user_config.ts
```diff
@@ -25,21 +30,55 @@ const ZoneId = _ZoneId;
 import _ZoneInfo from './zone_info';
 const ZoneInfo = _ZoneInfo;
 
+// Convince TypeScript and eslint that these are used.  TypeScript doesn't have a great way
```
panicstevenson, line 33: This feels odd to me; none of `disable-next-line`, `disable-line`, or `disable` are preferable here?

resources/user_config.ts
```diff
@@ -53,29 +92,41 @@ class UserConfig {
     //  * b/some_file.js
     //  * root_file1.js
     //  * root_file2.js
-    return keys.sort((a, b) => {
-      const maxLen = Math.max(splitKeyMap[a].length, splitKeyMap[b].length);
+    return keys.sort((keyA, keyB) => {
+      const listA = splitKeyMap[keyA];
+      const listB = splitKeyMap[keyB];
+      if (listA === undefined || listB === undefined)
+        return 0; // Convince TypeScript these exist.
```
panicstevenson, line 99: Which are we convincing here?

Trim21: ts think listA and listB maybe undefined

resources/user_config.ts
quisquous: Yeah, I think that would work, but ultimately there's no interesting stack here (as compared to a runtime error).  The only stack is just "you're in the middle of loading a user file".

resources/user_config.ts
quisquous: I can do `/* eslint-disable @typescript-eslint/no-unused-vars */` but then there is still a `ts(6133)` warning in VSCode (or an error when running normally).  I can use a `// @ts-ignore` to ignore the next line, but that ignores everything.  There's (deliberately??) no way to disable a single warning for a single line.

So, I thought "let me just use these" and then we at least get all the other errors still, but suppress the "unused" one.

resources/user_config.ts
quisquous: > Which are we convincing here?

Sorry, I'm not sure what your question is.  Are you saying that these might actually be undefined?  Can you clarify what you're trying to get at?

panicstevenson: I don't know what "these" represents in the comment. It's a line comment, so I feel like "these" is the zero. But, that makes no sense, so I'm assuming it's listA and listB

quisquous: Moved the comment, thanks!

panicstevenson (approved):

PR #2802, opened 2021-04-27
plugin: Update Global CE Opcode for 5.5hotfix

KattTails:
Untested, can verify later once servers are up.


KattTails:
Just got around to testing it, works as expected.

PR #2808, opened 2021-04-28
raidboss: Remove instanceof Check for Browser TTS

panicstevenson:
Fixes https://github.com/quisquous/cactbot/issues/2809

### DIFF
--- a/ui/raidboss/browser_tts_engine.ts
+++ b/ui/raidboss/browser_tts_engine.ts
@@ -48,8 +48,13 @@ class GoogleTTSItem implements TTSItem {
   }
 
   play() {
-    if (this.item && this.item instanceof HTMLAudioElement)
-      void this.item.play();
+    if (this.item) {
+      // Unfortunately, it doesn't seem like there's a great way to avoid casting here since
+      // HTMLIFrameElement Window's version of HTMLAudioElement uses a different constructor
+      // than the global Window's version.
+      // https://github.com/microsoft/TypeScript/issues/21568
+      void (this.item as HTMLAudioElement).play();
+    }
   }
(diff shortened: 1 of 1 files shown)


ngld:
You can probably keep the check if you compare against `iframe.contentWindow.HTMLAudioElement` instead. That said, why does this code use iframes in the first place and why are they never cleaned up? Considering that each iframe needs its own v8 and blink context and a new iframe is created for each TTS line, this seems like a wonderful way to waste RAM.

quisquous:
I had suggested `iframe.contentWindow.HTMLAudioElement` offline but we couldn't get it to work.

I think enough people are complaining (somehow? why is this so popular??) on discord that I'd rather land this now and figure out what followups make sense.

@ngld I was planning to rip out this particular path anyway in favor of `window.speechSynthesis` but it's waiting on the CEF version upgrade so it can be used in overlays as well.  Do you have any thought of when/if that CEF version bump might happen?

SiliconExarch:
> somehow? why is this so popular??

Speaking for myself, I use it because I get really unpleasant audio artifacts (sounds like clipping) using the built in ACT TTS system unless I turn the volume down to a level I can barely hear over the game audio.

ngld:
> Do you have any thought of when/if that CEF version bump might happen?

Just updating CEF isn't enough to make TTS work and even if we can make it work, it'd just do the same as the `cactbotSay` / `say` handler that we already have. I'll have to dig into the CefSharp and CEF code and figure out why it's not working in OverlayPlugin since it works fine in cefclient.

The good news is that the issues with transparent backgrounds are fixed which prevented the CEF update before.

PR #2811, opened 2021-04-29
raidboss: Fix E12S Basic Relativity Debuffs

nebularg:
Missed a few `matches.effectId.toUpperCase()`

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e12s.js
+++ b/ui/raidboss/data/05-shb/raid/e12s.js
@@ -1263,20 +1263,22 @@ export default {
           },
         };
 
-        if (matches.effectId === '998')
+        const id = matches.effectId.toUpperCase();
+
+        if (id === '998')
           return { infoText: output.shadoweye() };
-        if (matches.effectId === '99D')
+        if (id === '99D')
           return { infoText: output.water() };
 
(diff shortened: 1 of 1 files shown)


nebularg:
I'm not sure how I can fix those errors. It's the same as https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/raid/e12s.js#L1633-L1646

ui/raidboss/data/05-shb/raid/e12s.js
```diff
@@ -1263,20 +1263,22 @@ export default {
           },
         };
 
-        if (matches.effectId === '998')
+        const id = matches.effectId.toUpperCase();
```
quisquous, line 1266: (Thanks for the fix!) I see what the issue is.  In order to understand output strings inside of a response, cactbot makes the assumption that anything that has the `// cactbot-builtin-response` string in it can be called with `matches = {}` and not have an error.  In practice, `matches.effectId` has to exist if the regex matched, but in tests and the config ui that's not always true.
```suggestion
        if (!matches.effectId)
          return;
        const id = matches.effectId.toUpperCase();
```

nebularg:
I wondered why it wasn't just mocked for testing, but it does make sense the config ui just wants to get the outputStrings. Thanks!

PR #2815, opened 2021-04-30
raidboss: Convert Timeline Matches to Named Variables

panicstevenson:
Destructure RegExpExecArray results from RegExp.exec() to named
variables. Not a necessary step for the TypeScript conversion, but it is
a more readable alternative to `(match && match[1] && match[2])`.

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -178,7 +178,7 @@ export class Timeline {
       speaker: /define speaker "[^"]*"(\s+"[^"]*")?\s+(-?[0-9]+(?:\.[0-9]+)?)\s+(-?[0-9]+(?:\.[0-9]+)?)/,
       syncCommand: /(?:[^#]*?\s)?(sync\s*\/(.*)\/)(\s.*)?$/,
       tts: /^alertall\s+"([^"]*)"\s+before\s+(-?[0-9]+(?:\.[0-9]+)?)\s+(sound|speak\s+"[^"]*")\s+"([^"]*)"$/,
-      windowCommand: /(?:[^#]*?\s)?(window\s+(?:([0-9]+(?:\.[0-9]+)?),)?([0-9]+(?:\.[0-9]+)?))(?:\s.*)?$/,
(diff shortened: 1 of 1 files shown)


ui/raidboss/timeline.js
```diff
@@ -199,45 +199,44 @@ export class Timeline {
...
 
-      match = regexes.popupText.exec(line);
-      if (match) {
-        texts[match[2]] = texts[match[2]] || [];
-        texts[match[2]].push({
-          type: match[1],
-          secondsBefore: parseFloat(match[3]),
-          text: match[4] ? match[4] : match[2],
+      const [, popupType, popupId, popupSeconds, popupText] = regexes.popupText.exec(line) || [];
+      if (popupType && popupId && popupSeconds) {
```
quisquous, line 229: (I had this brief thought of "what if popupSeconds is 0"? However, apparently even though `"0" == false` probably because it converts to a number first, `"0"` converted to a boolean is true.  [Thanks, JavaScript?!?!].)

That said, this style of conditional here and elsewhere doesn't allow for empty string, even though the regex allows for empty strings in a number of places.  I'm not saying it necessarily should just saying that this is a change in behavior in a number of places.  Do you think it'd be worth checking only defined here instead of truthy to convince TypeScript that they are valid?  Alternatively, I could go through and check them more carefully.

ui/raidboss/timeline.js
```diff
@@ -199,45 +199,44 @@ export class Timeline {
...
-        this.ignores[match[1]] = true;
+      const [, ignoreId] = regexes.ignore.exec(line) || [];
+      if (ignoreId) {
+        this.ignores[ignoreId] = true;
         continue;
       }
 
-      match = regexes.tts.exec(line);
-      if (match) {
+      const [, ttsId, ttsSeconds, ttsCommand, ttsText] = regexes.tts.exec(line) || [];
```
quisquous, line 208: Do you think it's worth rewriting the regexes as well to have named capture groups rather than using ids by position? e.g. changing the above to `tts: /^alertall\s+"(?<ttsId>[^"]*)"\s+before\s+(?<ttsSeconds>-?[0-9]+(?:\.[0-9]+)?)\s+(?<ttsCommand>sound|speak\s+"[^"]*")\s+"(?<ttsText>[^"]*)"$/,`, and then destructring into an object instead of an array?

PR #2816, opened 2021-04-30
resources: convert player_override.js to TypeScript

quisquous:


quisquous:
This needs to be tested, but thought I would at least push the changes so that other folks didn't also do this work.

resources/player_override.ts
```diff
@@ -12,14 +15,18 @@ import Util from './util';
...
+
+export const addPlayerChangedOverrideListener = (playerName: string,
+    func: PlayerChangedFunc): void => {
   if (!func)
     return;
 
-  let lastPlayerChangedEvent = null;
-  let lastPlayerJob = null;
+  let lastPlayerChangedEvent: PlayerChangedDetail | null = null;
+  let lastPlayerJob: (Job | null) = null;
```
panicstevenson, line 27: Why does this one have parentheses? Any reason?

quisquous: No idea, fixed.

PR #2821, opened 2021-05-01
raidboss: Add Strict Undefined Checks for Timeline

panicstevenson:
Part of the TypeScript conversion for timelines; trying to separate
breaking logic changes.

### DIFF
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -201,10 +201,9 @@ export class Timeline {
       let match = regexes.ignore.exec(line);
       if (match && match['groups']) {
         const ignore = match['groups'];
-        if (ignore.id) {
+        if (ignore.id)
           this.ignores[ignore.id] = true;
-          continue;
-        }
+        continue;
       }
 
       match = regexes.tts.exec(line);
@@ -858,14 +857,12 @@ export class TimelineUI {
     const activeBar = this.activeBars[e.id];
     if (activeBar) {
       const div = activeBar.parentNode;
-      if (div && div.parentNode)
-        div.parentNode.removeChild(div);
+      div?.parentNode?.removeChild(div);
     }
 
     div.style.order = e.sortKey;
     div.id = e.id;
-    if (this.timerlist)
-      this.timerlist.appendChild(div);
+    this.timerlist?.appendChild(div);
     this.activeBars[e.id] = bar;
     if (e.id in this.expireTimers) {
       window.clearTimeout(this.expireTimers[e.id]);
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -209,19 +209,19 @@ export class Timeline {
       match = regexes.tts.exec(line);
       if (match && match['groups']) {
         const tts = match['groups'];
-        if (tts.id && tts.beforeSeconds && tts.command) {
-          // TODO: Support alert sounds?
-          if (tts.command === 'sound')
-            continue;
-          const ttsItems = texts[tts.id] || [];
-          texts[tts.id] = ttsItems;
-          ttsItems.push({
-            type: 'tts',
-            secondsBefore: parseFloat(tts.beforeSeconds),
-            text: tts.text ? tts.text : tts.id,
-          });
+        if (!tts.id || !tts.beforeSeconds || !tts.command)
           continue;
-        }
+        // TODO: Support alert sounds?
+        if (tts.command === 'sound')
+          continue;
+        const ttsItems = texts[tts.id] || [];
+        texts[tts.id] = ttsItems;
+        ttsItems.push({
+          type: 'tts',
+          secondsBefore: parseFloat(tts.beforeSeconds),
+          text: tts.text ? tts.text : tts.id,
+        });
+        continue;
       }
       match = regexes.soundAlert.exec(line);
       if (match)
@@ -233,16 +233,16 @@ export class Timeline {
       match = regexes.popupText.exec(line);
       if (match && match['groups']) {
         const popupText = match['groups'];
-        if (popupText.type && popupText.id && popupText.beforeSeconds) {
-          const popupTextItems = texts[popupText.id] || [];
-          texts[popupText.id] = popupTextItems;
-          popupTextItems.push({
-            type: popupText.type,
-            secondsBefore: parseFloat(popupText.beforeSeconds),
-            text: popupText.text ? popupText.text : popupText.id,
-          });
+        if (!popupText.type || !popupText.id || !popupText.beforeSeconds)
           continue;
-        }
+        const popupTextItems = texts[popupText.id] || [];
+        texts[popupText.id] = popupTextItems;
+        popupTextItems.push({
+          type: popupText.type,
+          secondsBefore: parseFloat(popupText.beforeSeconds),
+          text: popupText.text ? popupText.text : popupText.id,
+        });
+        continue;
       }
       match = regexes.line.exec(line);
       if (!(match && match['groups'])) {
@@ -890,8 +890,7 @@ export class TimelineUI {
     const bar = this.activeBars[e.id];
     if (bar) {
       const div = bar.parentNode;
-      if (div && div.parentNode)
-        div.parentNode.removeChild(div);
+      div?.parentNode?.removeChild(div);
       delete this.activeBars[e.id];
     }
   }
--- a/ui/raidboss/timeline.js
+++ b/ui/raidboss/timeline.js
@@ -116,16 +116,17 @@ export class Timeline {
     for (const r of this.replacements) {
       if (r.locale && r.locale !== replaceLang)
         continue;
-      if (!r[replaceKey])
+      const replacement = r[replaceKey];
+      if (!replacement)
         continue;
-      for (const [key, value] of Object.entries(r[replaceKey]))
+      for (const [key, value] of Object.entries(replacement))
         text = text.replace(Regexes.parse(key), value);
     }
     // Common Replacements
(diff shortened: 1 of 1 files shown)


ui/raidboss/timeline.js
```diff
@@ -572,21 +593,23 @@ export class Timeline {
...
       } else if (t.type === 'alarm') {
-        if (this.showAlarmTextCallback)
+        if (this.showAlarmTextCallback && t.text)
           this.showAlarmTextCallback(t.text);
       } else if (t.type === 'tts') {
-        if (this.speakTTSCallback)
+        if (this.speakTTSCallback && t.text)
           this.speakTTSCallback(t.text);
-      } else if (t.type === 'trigger') {
+      } else if (t.type === 'trigger' && t.trigger && t.matches) {
```
panicstevenson, line 612: I'm not a huge fan of this section, so I'm open to any input here.

ui/raidboss/timeline.js
```diff
@@ -705,6 +743,9 @@ export class TimelineUI {
     this.init = true;
 
     this.root = document.getElementById('timeline-container');
+    if (!this.root)
+      return;
```
panicstevenson, line 747: Does this make sense to just return here?

ui/raidboss/timeline.js
```diff
@@ -871,13 +925,13 @@ export class TimelineUI {
...
   OnSyncTime(fightNow, running) {
-    if (!this.options.Debug)
+    if (!this.options.Debug || !this.debugElement)
       return;
 
     if (!running) {
       if (this.debugFightTimer)
         this.debugElement.removeChild(this.debugFightTimer);
-      delete this.debugFightTimer;
+      this.debugFightTimer = null;
```
panicstevenson, line 934: `delete` means that this can be `{type} | undefined`, and the type is already `TimerBar | null`, so I thought this made more sense; additionally, this is set to null explicitly in `Init()`

ui/raidboss/timeline.js
```diff
@@ -65,8 +65,8 @@ const activeText = {
 function computeBackgroundColorFrom(element, classList) {
   const div = document.createElement('div');
   const classes = classList.split('.');
-  for (let i = 0; i < classes.length; ++i)
-    div.classList.add(classes[i]);
+  for (const c of classes)
```
panicstevenson, line 68: `class` is protected, willing to take suggestions here.

ui/raidboss/timeline.js
quisquous: Are all of these necessary with TypeScript? If you define texts to always have `text` if it's a `tts` type or `trigger` and `matches` if it's a `trigger` type, do you need these checks?

At the very least, `t.trigger && t.matches` should go inside the `t.type === 'trigger'` conditional for consistency since this is really just a switch statement.

ui/raidboss/timeline.js
```diff
@@ -202,23 +201,28 @@ export class Timeline {
...
+          continue;
+        }
       }
 
       match = regexes.tts.exec(line);
       if (match && match['groups']) {
         const tts = match['groups'];
-        // TODO: Support alert sounds?
-        if (tts.command === 'sound')
+        if (tts.id && tts.beforeSeconds && tts.command) {
```
quisquous, line 213:
```suggestion
        // Convince TypeScript these exist.
        if (tts.id && tts.beforeSeconds && tts.command)
          continue;
```

(+ early return instead of more nesting, here and elsewhere)

ui/raidboss/timeline.js
```diff
@@ -249,76 +256,86 @@ export class Timeline {
...
-      if (line) {
-        let commandMatch = regexes.durationCommand.exec(line);
-        if (commandMatch && commandMatch['groups']) {
-          const durationCommand = commandMatch['groups'];
-          line = line.replace(durationCommand.text, '').trim();
-          e.duration = parseFloat(durationCommand.seconds);
-        }
+      if (parsedLine.text && parsedLine.time && parsedLine.name) {
+        line = line.replace(parsedLine.text, '').trim();
+        // There can be # in the ability name, but probably not in the regex.
```
quisquous, line 261: not your code but wow this is optimistic

ui/raidboss/timeline.js
```diff
@@ -249,76 +256,86 @@ export class Timeline {
...
-        lineNumber: lineNumber,
-      };
-      if (line) {
-        let commandMatch = regexes.durationCommand.exec(line);
-        if (commandMatch && commandMatch['groups']) {
-          const durationCommand = commandMatch['groups'];
-          line = line.replace(durationCommand.text, '').trim();
-          e.duration = parseFloat(durationCommand.seconds);
-        }
+      if (parsedLine.text && parsedLine.time && parsedLine.name) {
```
quisquous, line 259: early out instead of indent?

ui/raidboss/timeline.js
```diff
@@ -249,76 +256,86 @@ export class Timeline {
...
+          id: uniqueid++,
+          time: seconds,
+          // The original ability name in the timeline.  Used for hideall, infotext, etc.
+          name: parsedLine.name,
+          // The text to display.  Not used for any logic.
+          text: this.GetReplacedText(parsedLine.name),
+          activeTime: 0,
+          lineNumber: lineNumber,
+        };
+        if (line) {
```
quisquous, line 275: more early outs?

ui/raidboss/timeline.js
```diff
@@ -808,14 +855,17 @@ export class TimelineUI {
...
 
     // Adding a timer with the same id immediately removes the previous.
-    if (this.activeBars[e.id]) {
-      const div = this.activeBars[e.id].parentNode;
-      div.parentNode.removeChild(div);
+    const activeBar = this.activeBars[e.id];
+    if (activeBar) {
+      const div = activeBar.parentNode;
+      if (div && div.parentNode)
+        div.parentNode.removeChild(div);
```
quisquous, line 862:
```suggestion
      div?.parentNode?.removeChild(div);
```

ui/raidboss/timeline.js
quisquous: Makes sense.  I think picking one of `null` or `undefined` to represent "unset" seems reasonable to me.  (I think either would be ok here.)

ui/raidboss/timeline.js
```diff
@@ -808,14 +855,17 @@ export class TimelineUI {
...
     }
 
     div.style.order = e.sortKey;
     div.id = e.id;
-    this.timerlist.appendChild(div);
+    if (this.timerlist)
+      this.timerlist.appendChild(div);
```
quisquous, line 868:
```suggestion
    this.timerlist?.appendChild(div);
```

ui/raidboss/timeline.js
quisquous: (or feel free not to leave a comment? Maybe it's weird to drop these comments everywhere, but I think there's something useful to me in knowing "me as the programmer really expects this path to not be possible given what I know" vs "this is a check I expect to possibly fire in practice"  I don't know how to make this more readable?  Do we need a `convinceTypeScript() { throw new Error('This shouldn\'t happen OOPS'); }`?  Other ideas???)

ui/raidboss/timeline.js
```diff
@@ -724,11 +766,15 @@ export class TimelineUI {
...
     // Helper for positioning/resizing when locked.
     const helper = document.getElementById('timeline-resize-helper');
+    if (!helper)
+      return;
     const rows = Math.max(6, this.options.MaxNumberOfTimerBars);
     helper.style.gridTemplateRows = 'repeat(' + rows + ', 1fr)';
 
     for (let i = 0; i < this.options.MaxNumberOfTimerBars; ++i) {
       const helperBar = document.createElement('div');
+      if (!helperBar)
```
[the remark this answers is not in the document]
panicstevenson: Unfortunately it will return `HTMLDivElement | null`

ui/raidboss/timeline.js
panicstevenson: I removed these for now; when the TypeScript change happens, I will come back and see what I can do here.

ui/raidboss/timeline.js
panicstevenson: I don't feel like "Convince TypeScript these exist" is the right call here, though. Right? There's no explicit contract that says these values have to exist. I think it would be more appropriate to log an error and continue. How do you feel about that?

ui/raidboss/timeline.js
quisquous: WAIT.  Sorry for this suggestion, but I didn't think you could use `?.` in JavaScript yet until OverlayPlugin updates its CEF version.  How is this not complaining?

ui/raidboss/timeline.js
quisquous: There is no explicit contract, but I think logically this regex has no optional capture groups.  If it matches, then all of these fields are defined.  Arguably, somebody could change the regex and not this code, and maybe that's what you're getting at?

What about rephrasing it as `notReached() { throw new Error('This code shouldn\'t be reached'); }`? Or do you think an exception is too heavy?

Maybe I'm overthinking this?

ui/raidboss/timeline.js
panicstevenson: It might be complaining; I haven't loaded the overlay since I initially made the PR since I'm just addressing comments right now

panicstevenson: Committed and reverted.

ui/raidboss/timeline.js
panicstevenson: Added!

Issue #2822, opened 2021-05-02
Hook to post-process loaded raidboss triggers


lyze:
I'd like to add a hook to post-process all loaded triggers for the raidboss plugin. I'm more than happy to contribute the code to do this.

If I understand correctly, the current mechanism for overriding a single trigger is all-or-nothing, and involves replacing the trigger in its entirety. My specific use case is to keep the existing triggers, but change what the TTS module outputs. For example, I could make each TTS callout more moogly with a "kupo~":

```js
Options.PostProcessTriggers = triggerSets => {
  triggerSets.triggers.forEach(trigger => { trigger.tts = trigger.infoText + ' クポ~' });
};
```

PR #2825, opened 2021-05-02
Make alertText idempotent in e9s triggers.

lyze:


lyze:
Found this as part of #2822.

quisquous:
I think "don't modify code in alertText" is a consideration if it makes code more readable, but I also am not sure it should be mandatory or worth changing triggers to support it at this point.

Personally, I think if this was something that we wanted to change, it'd make sense to wait until all triggers were typescript, and then we could pass a readonly data to text functions and verify that this was true.

lyze:
> Personally, I think if this was something that we wanted to change, it'd make sense to wait until all triggers were typescript, and then we could pass a readonly data to text functions and verify that this was true.

Right, absolutely. This was the only obvious instance that I found. I felt this was a reasonable one-off cleanup to just send out regardless of whether we do anything in #2822.

panicstevenson:
I feel like it's reasonable; however, there are (from my quick count) at least ~30 other cases of this, and my simple regular expression lookup didn't even find this case.

quisquous:
Overall I'm not opposed to the code changes.  However, I don't think `RaidbossGuide` should be updated, mostly as stated before I don't think it's worth going through and changing old triggers to behave like this without some way to enforce it.  This is done and so it's here, but I don't think it's worth more PRs in this vein.

lyze:
Got rid of the `RaidbossGuide` change.

PR #2826, opened 2021-05-02
Add option to transform TTS output.

lyze:
Issue #2822.

### DIFF
--- a/ui/raidboss/popup-text.js
+++ b/ui/raidboss/popup-text.js
@@ -241,11 +241,14 @@ export class PopupText {
     if (this.options.IsRemoteRaidboss) {
       this.ttsEngine = new BrowserTTSEngine(this.displayLang);
       this.ttsSay = (text) => {
-        this.ttsEngine.play(text);
+        this.ttsEngine.play(this.options.TransformTts(text));
       };
     } else {
-      this.ttsSay = function(text) {
-        const cmd = { 'call': 'cactbotSay', 'text': text };
+      this.ttsSay = (text) => {
+        const cmd = {
+          'call': 'cactbotSay',
+          'text': this.options.TransformTts(text),
+        };
         callOverlayHandler(cmd);
       };
     }
--- a/ui/raidboss/raidboss_options.ts
+++ b/ui/raidboss/raidboss_options.ts
@@ -35,6 +35,8 @@ type RaidbossNonConfigOptions = {
   Triggers: unknown; // one giant TODO
   PlayerNameOverride: string | null;
   IsRemoteRaidboss: boolean;
+  // Transforms text before passing it to TTS.
+  TransformTts: (text: string) => string;
 };
 
 // These options are ones that are not auto-defined by raidboss_config.js.
@@ -58,6 +60,8 @@ const defaultRaidbossNonConfigOptions: RaidbossNonConfigOptions = {
 
   PlayerNameOverride: null,
   IsRemoteRaidboss: false,
+
+  TransformTts: (t) => t,
 };
 
 // TODO: figure out how to get this type from raidboss_config??


lyze:
I tested this in the raid emulator with a user file like this: https://gist.github.com/lyze/32a8bf44f513466b8c14289bf2c2fef1

Trim21:
But `OrderedTriggerList` is not documented yet.

quisquous:
I agree that this function wrapping gist would work with this patch, but I'm a little bit unsure that this is the best way to implement this.  My worry is that this would make everybody who wanted to modify trigger output programatically to have to implement all the code that's in your gist and have to understand the details of everything triggers can return (now and into the future).

Are there other use cases to modify triggers programatically? If no, I might counterpropose a simpler `ProcessTriggerText(text: string) => string` option, where it passes in strings and returns modified strings.  (Possibly `ProcessTriggerText(text: string, triggerId: string, triggerField: string) => string`, in case you wanted to know which trigger this came from and whether this was alertText / tts / etc?)

> `t.response?.['infoText']`

A couple of tiny other details:

There's no ES2020 support in OverlayPlugin's version of CEF yet, so you can't do optional chaining inside of an overlay yet, even if it works in the browser in raid emulator.

Also, `response` can also return a function that returns `{ infoText, etc }` and not just an object that contains the same.

quisquous:
> But `orderedTriggers: OrderedTriggerList` is not documented yet.

Sorry, I'm not sure I understand what your concern is here.

Trim21:
> > But `orderedTriggers: OrderedTriggerList` is not documented yet.
> 
> Sorry, I'm not sure I understand what your concern is here.

```JS
this.options.PostProcessTriggers?.(orderedTriggers);
```

Users will have to dig into the source code to see what this value is. But `OrderedTriggerList` is neither documented nor typed.

lyze:
Thanks for the quick response and helpful comments.

> I agree that this function wrapping gist would work with this patch, but I'm a little bit unsure that this is the best way to implement this. My worry is that this would make everybody who wanted to modify trigger output programatically to have to implement all the code that's in your gist and have to understand the details of everything triggers can return (now and into the future).

I agree with this. The trigger internals are fairly complex. But I think most of the heavy lifting can be covered by a single function along the lines of something like this:

```js
// Returns a trigger field created from the specified trigger field (such as `alertText`, `infoText`, `tts`).
// The returned value is suitable for use as a field in a trigger.
// `triggerValueFn` is a callback that is called with the resolved value of the trigger. Some triggers
// (such as with locales, or with `data, matches, output` trigger functions) will only have this
// triggerValueFn called when the trigger field is evaluated.
// `fieldName` is `infoText`, `alertText`, `tts`, etc.
//
// Example usage:
//  --snip--
//
function ResolveTriggerField(trigger, fieldName, triggerValueFn) {
  let triggerField = trigger[fieldName];
  if (!triggerField) {
    if (typeof trigger.response === 'function') {
      triggerField = (data, matches, output) => trigger.response(data, matches, output)[fieldName];
    } else {
      triggerField = trigger.response[fieldName];
    }
  }
  switch (typeof triggerField) {
    case 'string':
      return triggerValueFn(triggerField);
    case 'function':
      return (data, matches, output) => {
        triggerValueFn(triggerField(data, matches, output), data, matches, output);
      };
    case 'object':
      for (const [k, tf] of triggerField) {
        triggerField[k] = ResolveTriggerField(tf, k, triggerValueFn);
      }
      return triggerField;
  }
}
```

The downside here is that it doesn't distinguish easily between "immediate" triggers (like returning string) and "deferred" triggers (requiring `data, matches, output` or locale). But I see this distinction more as a question about optimization.

> Are there other use cases to modify triggers programatically? If no, I might counterpropose a simpler ProcessTriggerText(text: string) => string option, where it passes in strings and returns modified strings. (Possibly ProcessTriggerText(text: string, triggerId: string, triggerField: string) => string, in case you wanted to know which trigger this came from and whether this was alertText / tts / etc?)

I was originally considering this approach. If there was a hook that took the text destined for TTS and transformed it before actually passing it to the TTS module, that would work for my purposes. I felt that this interface was too limited, but I am happy to go this route.

Is it in the spirit of the project to expose an internal, use-at-your-own-risk API? One could imagine that `ProcessTriggerText(text, id, field)` could be implemented with the footgun form of `PostProcessTriggers`.

quisquous:
> Thanks for the quick response and helpful comments.
> 
> > I agree that this function wrapping gist would work with this patch, but I'm a little bit unsure that this is the best way to implement this. My worry is that this would make everybody who wanted to modify trigger output programatically to have to implement all the code that's in your gist and have to understand the details of everything triggers can return (now and into the future).
> 
> I agree with this. The trigger internals are fairly complex. But I think most of the heavy lifting can be covered by a single function along the lines of something like this:

The question is less *can* it be written and more does it make sense for all users to have to write it (or copy it).

> The downside here is that it doesn't distinguish easily between "immediate" triggers (like returning string) and "deferred" triggers (requiring `data, matches, output` or locale). But I see this distinction more as a question about optimization.

Are there any triggers at this point that return a string directly? Everything is using output strings these days, and so goes through a function.

> > Are there other use cases to modify triggers programatically? If no, I might counterpropose a simpler ProcessTriggerText(text: string) => string option, where it passes in strings and returns modified strings. (Possibly ProcessTriggerText(text: string, triggerId: string, triggerField: string) => string, in case you wanted to know which trigger this came from and whether this was alertText / tts / etc?)
> 
> I was originally considering this approach. If there was a hook that took the text destined for TTS and transformed it before actually passing it to the TTS module, that would work for my purposes. I felt that this interface was too limited, but I am happy to go this route.

Are there other use cases here?

> Is it in the spirit of the project to expose an internal, use-at-your-own-risk API? One could imagine that `ProcessTriggerText(text, id, field)` could be implemented with the footgun form of `PostProcessTriggers`.

The issue for me is what I stated before, in that everybody who wants to have `ProcessTriggerText` needs to now write `PostProcessTriggers` in their raidboss user file.  It's hard for me to imagine what other use cases would motivate this.

valarnin:
> Are there other use cases here?

I actually have a fair number of e12s triggers that end up being "copy the entirety of the existing trigger, but do X before/after the default logic", or "change the callout of X trigger from boss-relative to absolute", which this would be nice for.

The most basic user config implementation based on this PR would appear to be:

```js
Options.PostProcessTriggers = ts => {
  console.log('Post processing trigger set', ts);
};
```

The other code in the gist is just for what lyze was trying to change with triggers with their specific use case.

I'm not sure it makes sense to implement this at the point in PopupText where it is, however. There are a number of checks and optimizations to the trigger set that would cause it to no longer match the trigger format from the file (e.g. lang-specific regexes are selected). To me it would make more sense for this to be a pre-process hook, or maybe a post-trigger-execution hook as Quisquous suggested with maybe an additional field to clarify the type of text, e.g.:
```js
ProcessTriggerText(text: string, triggerId: string, triggerField: string, textType: 'info' | 'alert' | 'alarm' | 'tts') => string
```

However, back to the Post Process logic, there are at least two notable issues in general with this process:

1. This would be fired every time the user zones or changes jobs, not just once (ReloadTimelines calls).
2. This appears to allow modifying the original trigger data which means that it would cause all modifications recursively any time ReloadTimeline is called, meaning you could quickly end up with something like `trigger.response = (data, matches, output) => {/*do something*/; oldTriggerResponse(data, matches, output);}` where `oldTriggerResponse ends up nested 10 deep or something.
   - This is noted in the comments for the `ProcessTrigger` function

Realistically, PopupText should probably be doing something like this:

```js
        for (const originalTrigger of set.triggers) {
            const trigger = Object.assign({}, originalTrigger);
```

Which would then avoid modifying the original trigger and negate the issues above. I'm not sure if that's really a discussion for this PR though.

quisquous:
> I actually have a fair number of e12s triggers that end up being "copy the entirety of the existing trigger, but do X before/after the default logic", or "change the callout of X trigger from boss-relative to absolute", which this would be nice for.

Depending on your ordering constraints, user-defined triggers execute after built-in triggers.  So, in some cases the "after the default logic" could be done via appending new triggers vs modifying existing ones.

These feel like different use cases.  One is "I want to modify O(3000) triggers to have different output", where it *is* possible to do such a thing via the config UI, but it would be a gigantic pain.  Another is "I want to not copy and paste O(10) triggers because I'd rather modify them partially than copy and paste completely", where there's already an existing method to do such a thing.

Personally, I feel like "copy the entirety of the existing triggers" is a looser coupling and so is less likely to break if the built-in triggers are modified to behave differently.  I understand that copypasta feels bad, but I think "resilience to change" >>> "don't repeat yourself" in this case.  In my experience, it is far more likely for built-in triggers to change than for ffxiv to change.  It's one reason why I moved away from the "per trigger options" where (for example) people could override just the `Condition` function vs overriding the entire trigger on its own.

lyze:
> Personally, I feel like "copy the entirety of the existing triggers" is a looser coupling and so is less likely to break if the built-in triggers are modified to behave differently.

My initial thinking is the hook would handle cross-cutting concerns. For my use, changing TTS output doesn't need to be trigger-specific. 

Is your concern that the hook I'm proposing makes it too easy for people to rely on trigger internals? How much of this would be addressed by clearer documentation ("here be dragons")? As a new user to cactbot triggers, I found it surprising that triggers had to be overridden in their entirety.  Playing devil's advocate, I could say that if I copied a trigger and overwrote it, then I wouldn't know about improvements to a trigger's implementation because I have to maintain my own copy of it. I personally found this to be fairly clumsy, and I wasn't excited to maintain all aspects of the triggers I would have to copy.

quisquous:
> Playing devil's advocate, I could say that if I copied a trigger and overwrote it, then I wouldn't know about improvements to a trigger's implementation because I have to maintain my own copy of it.  I personally found this to be fairly clumsy, and I wasn't excited to maintain all aspects of the triggers I would have to copy.

You keep saying "maintain" here, and this is where I disagree.

Historically, FFXIV content changes very little. I can think of only two examples across all of time where FFXIV content has changed and triggers needed to be updated (a11s optical sight ability ids, and ucob nael rp text in English).  However, triggers themselves in cactbot are more likely to be updated for various reasons (cleanups, refactorings, adding new triggers that combine with old triggers).

I am arguing that it is *vastly* more likely that somebody refactors the existing built-in trigger in a way that breaks your logic than FFXIV content changes and there is some maintenance savings to be had.  For example, if you add a hypothetical post-process hook to override `alertText` for some trigger.  Possible things that could do this are moving code from `condition` into `alertText` which your override won't handle.  Other things could be moving logic from `alertText` into `run` (e.g. your own #2825), and suddenly `shift` happens incorrectly twice.  This tight coupling makes it much more likely that small logic changes will break your user triggers.  Copying and pasting, even if it *feels* more clunky, is saying "I will handle this entire mechanic in this user file" and so even if the built-in trigger(s) for that mechanic are modified, your user triggers will continue working because they are handling the mechanic entirely.

The argument I'm trying to make is that there is ~zero maintenance of triggers as the ffxiv content itself ~doesn't change.  A looser coupling between user triggers and built-in triggers prevents much more breakage and headache.

lyze:
This is ready for review now. I went with the simplest option with a `string => string` callback that transforms the TTS output. We can probably reevaluate fancier options if there's a need.

Sorry for the delay. I was a bit busy recently, and couldn't find the time to do this until now.

PR #2832, opened 2021-05-03
Refactor: TypeScript-ize jobs/buff_tracker.ts

MaikoTan:
add common types in jobs module

### DIFF
--- a/ui/jobs/buff_tracker.ts
+++ b/ui/jobs/buff_tracker.ts
@@ -1,5 +1,5 @@
 import { BaseRegExp, Matches } from '../../types/trigger';
-import { IJobsOptions } from './types';
+import { JobsOptions } from './types';
 
 import WidgetList from '../../resources/widget_list';
 import EffectId from '../../resources/effect_id';
@@ -38,7 +38,7 @@ export interface Aura {
 export class Buff {
   name: string;
   info: BuffInfo;
-  options: IJobsOptions;
+  options: JobsOptions;
   activeList: WidgetList;
   cooldownList: WidgetList;
   readyList: WidgetList;
@@ -48,7 +48,7 @@ export class Buff {
   readySortKeyBase: number;
   cooldownSortKeyBase: number;
 
-  constructor(name: string, info: BuffInfo, list: WidgetList, options: IJobsOptions) {
+  constructor(name: string, info: BuffInfo, list: WidgetList, options: JobsOptions) {
     this.name = name;
     this.info = info;
     this.options = options;
@@ -200,7 +200,7 @@ export class Buff {
 
 export class BuffTracker {
   buffInfo: Record<string, BuffInfo>;
-  options: IJobsOptions;
+  options: JobsOptions;
   playerName: string;
   leftBuffDiv: WidgetList;
   rightBuffDiv: WidgetList;
@@ -212,7 +212,7 @@ export class BuffTracker {
   mobLosesEffectMap: Record<string, BuffInfo[]>;
 
   constructor(
-      options: IJobsOptions,
+      options: JobsOptions,
       playerName: string,
       leftBuffDiv: WidgetList,
       rightBuffDiv: WidgetList,
--- a/ui/jobs/buff_tracker.ts
+++ b/ui/jobs/buff_tracker.ts
@@ -9,7 +9,7 @@ import { kAbility } from './constants';
 import { makeAuraTimerIcon } from './utils';
 
 export interface BuffInfo {
-  name?: string;
+  name: string;
   gainAbility?: string;
   gainEffect?: string;
   loseEffect?: string;
@@ -203,7 +203,7 @@ export class Buff {
 }
 
 export class BuffTracker {
-  buffInfo: { [s: string]: BuffInfo };
+  buffInfo: { [s: string]: Omit<BuffInfo, 'name'> };
   options: JobsOptions;
   playerName: string;
   leftBuffDiv: WidgetList;
@@ -514,8 +514,11 @@ export class BuffTracker {
       mobLosesEffect: this.mobLosesEffectMap,
     } as const;
 
-    for (const [key, buff] of Object.entries(this.buffInfo)) {
-      buff.name = key;
+    for (const [key, buffOmitName] of Object.entries(this.buffInfo)) {
+      const buff = {
+        ...buffOmitName,
+        name: key,
+      };
 
       const overrides = this.options.PerBuffOptions[buff.name] ?? null;
       buff.borderColor = overrides?.borderColor ?? buff.borderColor;
@@ -613,7 +616,7 @@ export class BuffTracker {
     this.onLoseEffect(this.mobLosesEffectMap[name], matches);
   }
 
-  onBigBuff(name = '', seconds = 0, info: BuffInfo, source = ''): void {
+  onBigBuff(name: string, seconds = 0, info: BuffInfo, source = ''): void {
     if (seconds <= 0)
       return;
 
@@ -637,7 +640,7 @@ export class BuffTracker {
     buff?.onGain(seconds, source);
   }
 
-  onLoseBigBuff(name = ''): void {
+  onLoseBigBuff(name: string): void {
     this.buffs[name]?.onLose();
   }
 
--- a/ui/jobs/buff_tracker.ts
+++ b/ui/jobs/buff_tracker.ts
@@ -43,8 +43,8 @@ export class Buff {
   cooldownList: WidgetList;
   readyList: WidgetList;
   active: Aura | null;
-  cooldown: Record<string, Aura>;
-  ready: Record<string, Aura>;
+  cooldown: { [s: string]: Aura };
+  ready: { [s: string]: Aura };
   readySortKeyBase: number;
   cooldownSortKeyBase: number;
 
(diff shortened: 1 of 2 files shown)


ui/jobs/types.d.ts
```diff
@@ -0,0 +1,48 @@
...
+  PldLowMPThreshold: number;
+  DrkMediumMPThreshold: number;
+  DrkLowMPThreshold: number;
+  /**  One more fire IV and then despair. */
+  BlmMediumMPThreshold: number;
+  /** Should cast despair. */
+  BlmLowMPThreshold: number;
+}
+
+export type IJobsOptions = IDefaultOptions & IInitOptions;
```
quisquous, line 46: I'm not sure I understand the `I` prefix here.  Is this Hungarian/Microsoft naming? I'd maybe suggest:

* `IJobsOptions` => `JobsOptions`
* `IDefaultOptions` => `DefaultOptions` or `NonConfigOptions`
* `IInitOptions` => `InitOptions` or `ConfigOptions`

ui/jobs/buff_tracker.ts
```diff
@@ -1,10 +1,54 @@
+import { BaseRegExp, Matches } from '../../types/trigger';
+import { IJobsOptions } from './types';
+
+import WidgetList from '../../resources/widget_list';
 import EffectId from '../../resources/effect_id';
+import { AbilityParams, GainsEffectParams, LosesEffectParams } from '../../resources/netregexes';
```
quisquous, line 6: This is just a question for my own curiosity because it came up in `ui/radar/radar.ts` that I'm working on, but what happens when you need to import both default and non-default parameters? Is that possible?

ui/jobs/buff_tracker.ts
```diff
@@ -25,12 +69,12 @@ export class Buff {
...
   }
 
-  addCooldown(source, effectSeconds) {
+  addCooldown(source: string, effectSeconds: number): void {
     if (!this.info.cooldown)
       return;
     if (this.cooldown[source]) {
       // Unexpected use of the same cooldown by the same name.
-      this.cooldown[source].removeCallback();
+      this.cooldown[source]?.removeCallback();
```
quisquous, line 77: I think you could remove `if (this.cooldown[source]) {` check before this if you're using optional chaining.  Maybe the comment could be "// Remove any preexisting cooldowns with the same name in case they unexpectedly exist" instead.

Same thing for the ready check below.

ui/jobs/buff_tracker.ts
```diff
@@ -67,42 +111,56 @@ export class Buff {
...
-          if (expireCallback)
-            expireCallback();
-        }, seconds * 1000);
-      }
+
+        if (seconds > 0) {
+          aura.removeTimeout = window.setTimeout(() => {
+            aura.removeCallback();
+            if (expireCallback)
+              expireCallback();
```
quisquous, line 155: `
```suggestion
            expireCallback?.();
```

ui/jobs/buff_tracker.ts
```diff
@@ -1,10 +1,54 @@
...
 export class Buff {
-  constructor(name, info, list, options) {
+  name: string;
+  info: BuffInfo;
+  options: IJobsOptions;
+  activeList: WidgetList;
+  cooldownList: WidgetList;
+  readyList: WidgetList;
+  active: Aura | null;
+  cooldown: Record<string, Aura>;
```
quisquous, line 46: This is just subjective opinion, but I think `Record` should only be used in cases where all of the keys are known and finite and have a mapping to a value.  So, for things like jobs -> names or roles -> players, where the keys are a union of values.  I think it makes more sense to use an index signature on "infinite" types like strings, e.g. `{ [buffName: string]: Aura }`.  I think this also makes the key semantics self-documenting because you have to label them.  For a `Record` using a union type as a key field, it is also self-documenting.

Not that Stack Overflow answers are authoritative, but https://stackoverflow.com/questions/51936369/what-is-the-record-type-in-typescript says "If you want to accept all possible string-valued keys, you could do something like Record<string, T>, but the idiomatic way of doing that is to use an index signature like { [k: string]: T }." which I agree with.

I know there are `Record<string, something>` examples elsewhere, but I understand `Record` a lot better now than when I first saw those.

ui/jobs/buff_tracker.ts
```diff
@@ -445,99 +519,106 @@ export class BuffTracker {
       mobLosesEffect: this.mobLosesEffectMap,
     };
 
-    for (let i = 0; i < keys.length; ++i) {
-      const buff = this.buffInfo[keys[i]];
-      buff.name = keys[i];
+    for (const key in this.buffInfo) {
+      const buff = this.buffInfo[key];
+      if (!buff) continue;
```
quisquous, line 524:
```suggestion
    for (const [key, buff] in Object.entries(this.buffInfo)) {
```

ui/jobs/buff_tracker.ts
```diff
@@ -46,19 +90,19 @@ export class Buff {
...
 
     // TODO: could consider looking at the party list to make initials unique?
     let txt = '';
     const initials = source.split(' ');
     if (initials.length === 2)
-      txt = initials[0][0] + initials[1][0];
+      txt = `${initials[0]?.charAt(0) ?? '?'}${initials[1]?.charAt(0) ?? '?'}`;
     else
-      txt = initials[0];
+      txt = initials[0] ?? '?';
```
quisquous, line 105: Is there something that can make this more readable?

Maybe `txt = initials.map((str) => str?.charAt(0) ?? '?').join('');`

ui/jobs/buff_tracker.ts
```diff
@@ -113,19 +171,19 @@ export class Buff {
...
 
-  clear() {
+  clear(): void {
     this.onLose();
 
-    const cooldownKeys = Object.keys(this.cooldown);
-    for (let i = 0; i < cooldownKeys.length; ++i)
-      this.cooldown[cooldownKeys[i]].removeCallback();
+    Object.values(this.cooldown).forEach((aura) => {
+      aura?.removeCallback();
```
quisquous, line 178: Does this need the `?`? `this.cooldown` and `this.ready` have the same type.

ui/jobs/buff_tracker.ts
```diff
@@ -445,99 +519,106 @@ export class BuffTracker {
...
 
-  onUseAbility(id, matches) {
+  onUseAbility(id: string, matches: Matches<BaseRegExp<AbilityParams>>): void {
     const buffs = this.gainAbilityMap[id];
     if (!buffs)
       return;
 
     for (const b of buffs)
-      this.onBigBuff(b.name, b.durationSeconds, b, matches.source);
+      this.onBigBuff(b.name ?? '', b.durationSeconds ?? 0, b, matches?.source ?? '');
```
quisquous, line 575: I think this would be cleaner if these were optional parameters so that the function itself can handle it rather than the caller.  (Also, are all these fields really optional on buffs?)

ui/jobs/buff_tracker.ts
```diff
@@ -445,99 +519,106 @@ export class BuffTracker {
...
+      buff.icon = overrides?.icon ?? buff.icon;
+      buff.side = overrides?.side ?? buff.side ?? 'right';
+      buff.sortKey = overrides?.sortKey || buff.sortKey;
+      buff.hide = overrides?.hide ?? buff.hide ?? false;
 
       for (const prop in propToMapMap) {
         if (!(prop in buff))
           continue;
-        const key = buff[prop];
+        const key = buff[prop as keyof typeof propToMapMap];
```
quisquous, line 537: Yeah, this construction is definitely a little bit weird, where it's using keys to index fields in another type.  It's not a very TypeScript-y way to do things, that's for sure.  I can't come up with a way to not use `as` here, but in general I'm pretty suspicious of using it.

Could you at least move the `as` up to the top, e.g.
```typescript
      for (const propStr in propToMapMap) {
        const prop = propStr as keyof typeof propToMapMap;
```

...that way there's only one `as` and it's a lot less suspicious looking, which makes the code more readable to my eyes.

Can you make `propToMapMap` be marked `as const` as well?

quisquous (commented):
Thank you for splitting this up!  This is a lot of change to look through even for just this one file.

ui/jobs/types.d.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export type JobFunc = (id: string, matches: Matches) => void;
```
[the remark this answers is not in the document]
MaikoTan: Oh, this is used for typing `Bars.onUseAbility`/`Bars.onYouGainEffect`/etc in `ui/jobs/jobs.ts`, the callback parameter of those functions.

ui/jobs/types.d.ts
MaikoTan: Yep, the `I` prefix means `Interface`, which is widely used in Microsoft's code. But I think it is good to change as you suggested.

ui/jobs/buff_tracker.ts
MaikoTan: It is just my guess, and I have not verified it: I think that you can write something like,
```typescript
import { AbilityParams } from 'path/to/netregexes';
type NonDefaultParameters = 'para1' | 'para2' | 'para3';
type CustomizedMatches = Matches<BaseRegExp<AbilityParams | NonDefaultParameters>>;
```

In my opinion, the typing of `Matches` and `BaseRegExp` is not good for these usages. Perhaps I and @Trim21 would rewrite the typing later.

ui/jobs/buff_tracker.ts
MaikoTan: You're right, thanks for catching this.

ui/jobs/buff_tracker.ts
MaikoTan: Changed as default value.

ui/jobs/buff_tracker.ts
quisquous: Oh oh, sorry.  I think I didn't explain myself well.  I didn't realize until now that you could do `import NetRegexes, { AbilityParams } from './netregexes';`.  Sorry I meant default/non-default *exports* and not *parameters*.  I misspoke.

MaikoTan: Ah, yes! You could do that! ^_~

ui/jobs/buff_tracker.ts
```diff
@@ -558,19 +630,14 @@ export class BuffTracker {
...
-    buff.onGain(seconds, source);
+    buff?.onGain(seconds, source);
   }
 
-  onLoseBigBuff(name) {
-    const buff = this.buffs[name];
-    if (!buff)
-      return;
-    buff.onLose();
+  onLoseBigBuff(name = ''): void {
```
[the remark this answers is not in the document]
MaikoTan: Found a better way, that we can omit the `key` property when define buff items, and then add it back when process them.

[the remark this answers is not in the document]
quisquous: Nice!!

PR #2833, opened 2021-05-03
Refactor: TypeScript-ize jobs/combo_tracker

MaikoTan:


ui/jobs/combo_tracker.ts
```diff
@@ -1,7 +1,19 @@
 import { kAbility, kComboBreakers } from './constants';
 
+type StartMap = Record<string, {
```
quisquous, line 3: Same comment as the other patch, but please use an index signature here like `{ [skill: string]: etc }` instead of `Record<string, etc>`

ui/jobs/combo_tracker.ts
```diff
@@ -1,7 +1,19 @@
...
+  id?: string;
+  next?: StartMap;
+}>;
+
 export default class ComboTracker {
-  constructor(comboBreakers, callback) {
+  comboTimer?: number | undefined;
```
quisquous, line 9: I think `comboTimer: number | undefined` or `comboTimer?: number`, but no need for both.

ui/jobs/combo_tracker.ts
```diff
@@ -1,7 +1,19 @@
 import { kAbility, kComboBreakers } from './constants';
 
+type StartMap = Record<string, {
+  id?: string;
+  next?: StartMap;
```
quisquous, line 5: I think these are both required.  Am I misunderstanding? `next` might be `{}` but that's a valid `StartMap`.

ui/jobs/combo_tracker.ts
```diff
@@ -1,7 +1,19 @@
...
 export default class ComboTracker {
-  constructor(comboBreakers, callback) {
+  comboTimer?: number | undefined;
+  comboBreakers: readonly string[];
+  startMap: StartMap;
+  callback: (id: string | null) => void;
+  considerNext: StartMap;
+  isFinalSkill: boolean;
+
+  constructor(comboBreakers: readonly string[], callback: (id: string | null) => void) {
```
quisquous, line 16: Would it be worth having a `type ComboSkill = string | null` to encapsulate this here and elsewhere? Maybe also a `type ComboTrackerCallback = (id: ComboSkill) => void;`?

ui/jobs/combo_tracker.ts
```diff
@@ -41,15 +52,15 @@ export default class ComboTracker {
       this.AbortCombo(id);
   }
 
-  StateTransition(id, nextState) {
+  StateTransition(id: string | null, nextState?: { id?: string; next?: StartMap } | null): void {
```
quisquous, line 55:
```suggestion
  StateTransition(id: string | null, nextState: StartMap[string] | null): void {
```

I think `nextState` is always defined, and maybe it's possible to not repeat the type of `StartMap` as well.  That also removes a bunch of `?.` and `??` from down below.

ui/jobs/combo_tracker.ts
```diff
@@ -27,11 +38,11 @@ export default class ComboTracker {
       if (!nextMap[id])
         nextMap[id] = node;
 
-      nextMap = nextMap[id].next;
-    }
+      nextMap = nextMap[id]?.next ?? {};
```
quisquous, line 41: I think this could be more readable without the unnecessary `?`.  `nextMap[id]` and `nextMap[id].next` are both always defined.  What about:

```typescript
      let nextEntry = nextMap[id];
      if (!nextEntry)
        nextEntry = nextMap[id] = node;
      nextMap = nextEntry.next;
```

ui/jobs/combo_tracker.ts
MaikoTan: Yes, you're right =_=

ui/jobs/combo_tracker.ts
MaikoTan: Sure!

ui/jobs/combo_tracker.ts
```diff
@@ -41,33 +56,33 @@ export default class ComboTracker {
...
   }
 
-  StateTransition(id, nextState) {
+  StateTransition(id?: string, nextState?: StartMap[string]): void {
     window.clearTimeout(this.comboTimer);
     this.comboTimer = undefined;
 
-    this.isFinalSkill = nextState && Object.keys(nextState.next).length === 0;
-    if (nextState === null || this.isFinalSkill) {
+    this.isFinalSkill = (nextState && Object.keys(nextState?.next ?? {}).length === 0) ?? false;
```
quisquous, line 63:
```suggestion
    this.isFinalSkill = (nextState && Object.keys(nextState.next).length === 0) ?? false;
```

I think (?) you don't need all of the optional chaining here.

ui/jobs/combo_tracker.ts
```diff
@@ -1,7 +1,23 @@
 import { kAbility, kComboBreakers } from './constants';
 
+type StartMap = {
+  [s: string]: {
+    id: string;
+    next: StartMap;
+  };
+};
+
+type ComboCallback = (id?: string) => void;
```
quisquous, line 10: Ah nice, I checked too and nothing is using `null`, just checking for `!id`.  So, this seems safe to change here.

PR #2834, opened 2021-05-03
Refactor: TypeScript-ize jobs/utils & jobs/jobs

MaikoTan:
Hard to split these two files, 'cause they reference each other.
Should fix type errors after other PRs merged. (* ￣︿￣)


ui/jobs/jobs.ts
```diff
@@ -53,39 +58,118 @@ const Options = {
...
-    this.comboTimer = null;
-    this.regexes = new RegexesHolder(this.options.ParserLanguage);
-
     this.skillSpeed = 0;
     this.spellSpeed = 0;
 
+    this.distance = -1;
+    this.inCombat = false;
+    this.combo = undefined;
+    this.regexes = new RegexesHolder(this.options.ParserLanguage);
```
quisquous, line 169: I wonder if it would be better to refactor this (ahead of time? in it's own PR?) to only construct `this.regexes` if there's a language and a player name), so that `this.regexes` is the optional piece (and you could early out in log testing if it doesn't exist) rather than having every field on `this.regexes` be independently optional.

ui/jobs/jobs.ts
```diff
@@ -396,41 +482,49 @@ class Bars {
...
     if (!container) {
       container = document.createElement('div');
       container.id = id;
-      document.getElementById('bars').appendChild(container);
+      document.getElementById('bars')?.appendChild(container);
       container.classList.add('proc-box');
     }
 
-    const timerBox = document.createElement('timer-box');
+    const timerBox = document.createElement('timer-box') as unknown as TimerBox;
```
quisquous, line 501: `as unknown`?

ui/jobs/jobs.ts
```diff
@@ -396,41 +482,49 @@ class Bars {
...
+  addTimerBar(options: {
+    id: string;
+    fgColor: string;
+  }): TimerBar {
     const container = this.addJobBarContainer();
 
     const timerDiv = document.createElement('div');
     timerDiv.id = options.id;
-    const timer = document.createElement('timer-bar');
+    const timer = document.createElement('timer-bar') as TimerBar;
```
quisquous, line 527: I was talking to @panicstevenson about this, but I wonder if we need a static `create` function so we could hide this `as` and `timer-bar` implicit name, e.g. like a `create(): TimerBar` function on `TimerBar`.  I don't think that should be done in this patch (maybe before? or after?), but it's something on my mind.

ui/jobs/jobs.ts
```diff
@@ -1002,31 +1089,32 @@ class Bars {
...
+  _test(): void {
     const logs = [];
     const t = '[10:10:10.000] ';
-    logs.push(t + '1A:10000000:' + this.me + ' gains the effect of Medicated from ' + this.me + ' for 30.2 Seconds.');
-    logs.push(t + '15:10000000:Tako Yaki:1D60:Embolden:10000000:' + this.me + ':500020F:4D70000:0:0:0:0:0:0:0:0:0:0:0:0:0:0:42194:42194:10000:10000:0:1000:-655.3301:-838.5481:29.80905:0.523459:42194:42194:10000:10000:0:1000:-655.3301:-838.5481:29.80905:0.523459:00001DE7');
-    logs.push(t + '1A:10000000:' + this.me + ' gains the effect of Battle Litany from  for 25 Seconds.');
-    logs.push(t + '1A:10000000:' + this.me + ' gains the effect of The Balance from  for 12 Seconds.');
+    const me = this.me ?? '';
+    logs.push(`${t}1A:10000000:${me} gains the effect of Medicated from ${me} for 30.2 Seconds.`);
+    logs.push(`${t}15:10000000:Tako Yaki:1D60:Embolden:10000000:${me}:500020F:4D70000:0:0:0:0:0:0:0:0:0:0:0:0:0:0:42194:42194:10000:10000:0:1000:-655.3301:-838.5481:29.80905:0.523459:42194:42194:10000:10000:0:1000:-655.3301:-838.5481:29.80905:0.523459:00001DE7'`);
```
quisquous, line 1100: Ironically, none of these lines work anymore because I never updated them to be network lines.

ui/jobs/jobs.ts
```diff
@@ -53,39 +58,118 @@ const Options = {
...
+    paeonStacks: number;
+    museStacks: number;
+    circleOfPower: number;
+  };
+
+  dotTarget: string[];
+  trackedDoTs: string[];
+  comboFuncs: ((skill: string | null) => void)[];
+  // hard to type this when we don't know what job is now
+  jobFuncs: ((jobDetail: never) => void)[];
```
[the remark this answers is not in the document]
MaikoTan: Well, in fact, you can simply do `JobDetail[keyof JobDetal]` if you want to get a union of `values` of an interface. 
So I think the types that you mentioned in `event.d.ts` should be updated too.ヾ(•ω•`)o

ui/jobs/jobs.ts
MaikoTan: (*￣3￣)╭ Sounds good, and I think it is better to put in a new PR.

ui/jobs/jobs.ts
```diff
@@ -114,29 +198,30 @@ class Bars {
...
     const bars = document.getElementById('bars');
     if (bars) {
       const barList = bars.children;
       for (const bar of barList) {
-        if (bar.id === 'hp-bar' || bar.id === 'mp-bar') continue;
+        if (bar.id === 'hp-bar' || bar.id === 'mp-bar')
+          continue;
         if (this.isPVPZone)
-          bar.style.display = 'none';
+          (bar as HTMLDivElement).style.display = 'none';
```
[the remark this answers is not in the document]
MaikoTan: Sure!

ui/jobs/jobs.ts
MaikoTan: ＞︿＜ Oh, this was written before the TypeScript-ization of `resources/timerbox.js`, and I forgot to delete the `unknown`-cast

ui/jobs/jobs.ts
MaikoTan: Maybe a "Factory class" is better?
i.e. something like
```typescript
class CustomElementFactory {
  static create(s: "timer-bar"): TimerBar;
  static create(s: "timer-icon"): TimerIcon;
  static create(s: string): HTMLElement {
    // return custom element based on the parameter.
  }
}
```

ui/jobs/jobs.ts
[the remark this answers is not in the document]
quisquous: Mmm, @panicstevenson and I tried that but couldn't get it to work when we had to use it in `ui/test/test.ts`.  If you can figure it out, please share!

quisquous:
Just checking in, but are you planning to get back to this? I think the changes to `jobs.ts` are still not captured by any other PRs yet.

MaikoTan:
Yeah, but I think I should open a new PR based on the changes recently, rather than merge up branches and deal with the completed conflicts.

PR #2840, opened 2021-05-05
resources: move notReached to common UnreachableCode error

quisquous:
Unfortunately, even a function that returns `never` is not enough
for TypeScript to trust this in all cases, so convert this to an
Error type instead.

This should remove the need for "convince TypeScript" sort of comments
as well.

### DIFF
--- a/resources/not_reached.ts
+++ b/resources/not_reached.ts
@@ -1,0 +1,14 @@
+// Helper Error for TypeScript situations where the programmer thinks they
+// know better than TypeScript that some situation will never occur.
+
+// The intention here is that the programmer does not expect a particular
+// bit of code to happen, and so has not written careful error handling.
+// If it does occur, at least there will be an error and we can figure out why.
+
+// One common example is a regex, where if the regex matches then all of the
+// (non-optional) regex groups will also be valid, but TypeScript doesn't know.
+export class UnreachableCode extends Error {
+  constructor() {
+    super('This code shouldn\'t be reached');
+  }
+}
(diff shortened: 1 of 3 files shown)


panicstevenson:
Sorry to force you to expound upon an already lengthy comment at the top, but should this include something about how we're trying to avoid casting or other things that usually say that we know more than TypeScript? For instance, we're not trying to know more than TypeScript, but we're _pretty_ confident that we know more.

MaikoTan:
> Unfortunately, even a function that returns `never` is not enough
> for TypeScript to trust this in all cases, so convert this to an
> Error type instead.


I think a function that returns `never` is okay for these usage though ?
What issue happened?

quisquous:
> > Unfortunately, even a function that returns `never` is not enough
> > for TypeScript to trust this in all cases, so convert this to an
> > Error type instead.
> 
> I think a function that returns `never` is okay for these usage though ?
> What issue happened?

Yes, I thought so as well, but https://github.com/quisquous/cactbot/commit/2a1f1f4ed802e46866433bdcd065f1a31fd7c9e1#diff-84bfd0469f7c02b315057a1d4bc390ba9c040c65ef708cde30fd593a4b1d1cb5R100

quisquous:
> Sorry to force you to expound upon an already lengthy comment at the top, but should this include something about how we're trying to avoid casting or other things that usually say that we know more than TypeScript? For instance, we're not trying to know more than TypeScript, but we're _pretty_ confident that we know more.

How would you reword this? Or, what would you add if you had your druthers?

panicstevenson:
> How would you reword this? Or, what would you add if you had your druthers?

I think you can just add a small sentence at the end of the intention paragraph:

```js
// The intention here is that the programmer does not expect a particular
// bit of code to happen, and so has not written careful error handling.
// If it does occur, at least there will be an error and we can figure out why.
// This is preferable to casting or disabling TypeScript altogether in order to
// avoid syntax errors.
```

panicstevenson (approved):

PR #2841, opened 2021-05-05
radar: Convert radar to TypeScript

quisquous:


quisquous:
```diff
diff --git a/ui/radar/radar.js b/ui/radar/radar.ts
index aac33a66e..19a77989b 100644
--- a/ui/radar/radar.js
+++ b/ui/radar/radar.ts
@@ -1,12 +1,66 @@
-import { callOverlayHandler, addOverlayListener } from '../../resources/overlay_plugin_api';
+import { BaseOptions } from '../../types/data';
+import { EventMap } from '../../types/event';
+import { Lang } from '../../types/global';
+import { Matches, NetRegex } from '../../types/trigger';
 
-import gMonster from '../../resources/hunt';
-import NetRegexes from '../../resources/netregexes';
+import { callOverlayHandler, addOverlayListener } from '../../resources/overlay_plugin_api';
+import HuntData, { HuntEntry, HuntMap, Rank } from '../../resources/hunt';
+import NetRegexes, { AddedCombatantFullParams } from '../../resources/netregexes';
 import UserConfig from '../../resources/user_config';
 
 import './radar_config';
 
-const Options = {
+type RadarType = 'mob' | 'any';
+
+// TODO: how to avoid copying these from ./radar_config?
+// One option could be for X_config to export its default values.
+// We could make a helper function to extract them from the template option.
+
+// Options that come radar_config and can be configured in the UI.
+const defaultRadarConfigOptions = {
+  BRankEnabled: true,
+  DetectionRange: 0,
+  TTS: false,
+  PopSoundAlert: true,
+  PopVolume: 0,
+  Puller: true,
+  Position: true,
+} as const;
+type RadarConfigOptions = typeof defaultRadarConfigOptions;
+
+// Additional options that can be configured via the user file.
+interface RadarOptions extends BaseOptions, RadarConfigOptions {
+  PopSound: string;
+  RankOptions: {
+    [rank in Rank]?: {
+      Type?: RadarType;
+      Enabled?: boolean;
+      PopSoundAlert?: boolean;
+    };
+  };
+  CustomMonsters: HuntMap;
+}
+
+// "Monster" is the live version of the static "HuntEntry" data.
+type Monster = {
+  id: string;
+  name: string;
+  rank?: Rank;
+  hp: number;
+  currentHp: number;
+  battleTime: number;
+  pos: Point2D;
+  posZ: number;
+  addTime: number;
+  dom: HTMLElement;
+  puller?: string;
+  // already pulled before being detected
+  skipPuller: boolean;
+}
+
+const Options: RadarOptions = {
+  ...UserConfig.getDefaultBaseOptions(),
+  ...defaultRadarConfigOptions,
   PopSound: '../../resources/sounds/freesound/sonar.ogg',
   RankOptions: {
     'S': {
@@ -40,8 +94,6 @@ const Options = {
 // TODO: this would also let us handle mobs with the same name better.
 const kMinDistanceBeforeSound = 100;
 
-let gRadar;
-
 const instanceChangedRegexes = {
   en: NetRegexes.gameLog({ code: '0039', line: 'You are now in the instanced area.*?' }),
   de: NetRegexes.gameLog({ code: '0039', line: 'Du bist nun in dem instanziierten Areal.*?' }),
@@ -52,7 +104,10 @@ const instanceChangedRegexes = {
 };
 
 class Point2D {
-  constructor(x, y) {
+  public x: number;
+  public y: number;
+
+  constructor(x: number, y: number) {
     this.x = x;
     this.y = y;
   }
@@ -63,83 +118,124 @@ class Point2D {
   }
 
   // Calculate delta vector
-  delta(target) {
+  delta(target: Point2D) {
     return new Point2D(target.x - this.x, target.y - this.y);
   }
 
   // Calculate distance between 2 points
-  distance(target) {
+  distance(target: Point2D) {
     return this.delta(target).length();
   }
 }
 
-function posToMap(h) {
+class Point2DWithZ extends Point2D {
+  public z: number;
+
+  constructor(x: number, y: number, z: number) {
+    super(x, y);
+    this.z = z;
+  }
+}
+
+const posToMap = (h: number) => {
   const offset = 21.5;
   const pitch = 0.02;
   return h * pitch + offset;
-}
+};
 
 
-function PlaySound(monster, options) {
+const PlaySound = (monster: Monster, options: RadarOptions) => {
   if (options.TTS) {
-    callOverlayHandler({
+    void callOverlayHandler({
       call: 'cactbotSay',
-      text: monster.rank + ' ' + monster.name,
+      text: `${monster.rank ?? ''} ${monster.name}`,
     });
   } else if (options.PopSoundAlert && options.PopSound && options.PopVolume) {
     const audio = new Audio(options.PopSound);
     audio.volume = options.PopVolume;
-    audio.play();
+    void audio.play();
   }
-}
+};
 
 class Radar {
-  constructor(element) {
+  private targetMonsters: { [mobKey: string]: Monster };
+  private playerPos: Point2DWithZ | null;
+  private playerRotation: number;
+  private table: HTMLElement;
+  private options: RadarOptions;
+  private hunts: HuntMap;
+  private lang: Lang;
+  private nameToHuntEntry: HuntMap;
+  private regexes: {
+    abilityFull: RegExp;
+    addedCombatantFull: RegExp;
+    instanceChanged: RegExp;
+    wasDefeated: RegExp;
+  };
+
+  constructor(element: HTMLElement) {
     this.targetMonsters = {};
-    this.playerPos = {};
+    this.playerPos = null;
     this.playerRotation = 0;
     this.table = element;
     this.options = Options;
-    this.monsters = Object.assign({}, gMonster, Options.CustomMonsters);
-    this.lang = this.options.ParserLanguage || 'en';
-    this.nameToMonster = {};
-    this.regex = {
+    this.hunts = Object.assign({}, HuntData, Options.CustomMonsters);
+    this.lang = this.options.ParserLanguage ?? 'en';
+    this.nameToHuntEntry = {};
+    this.regexes = {
       abilityFull: NetRegexes.abilityFull(),
       addedCombatantFull: NetRegexes.addedCombatantFull(),
       instanceChanged: instanceChangedRegexes[this.options.ParserLanguage] || instanceChangedRegexes['en'],
       wasDefeated: NetRegexes.wasDefeated(),
     };
 
-    for (const i in this.monsters) {
-      const monster = this.monsters[i];
-      monster.name = monster.name[this.lang] || monster.name['en'];
+    for (const hunt of Object.values(this.hunts)) {
+      let name = hunt.name;
+      if (typeof name === 'object' && !Array.isArray(name))
+        name = name[this.lang] ?? name['en'];
 
-      // Names are either strings or arrays of strings.
-      if (typeof monster.name === 'string') {
-        this.nameToMonster[monster.name.toLowerCase()] = monster;
+      if (typeof name === 'string') {
+        this.nameToHuntEntry[name.toLowerCase()] = hunt;
       } else {
-        for (let i = 0; i < monster.name.length; ++i)
-          this.nameToMonster[monster.name[i].toLowerCase()] = monster;
+        for (const nameItem of name)
+          this.nameToHuntEntry[nameItem.toLowerCase()] = hunt;
       }
     }
   }
 
-  AddMonster(log, monster, matches) {
-    if (monster.id && matches.npcNameId !== monster.id)
+  AddMonster(log: string, hunt: HuntEntry,
+      matches: Matches<NetRegex<AddedCombatantFullParams>>) {
+    if (!this.playerPos)
+      return;
+    if (!matches)
+      return;
+    if (matches.id === undefined ||
+        matches.name === undefined ||
+        matches.npcNameId === undefined ||
+        matches.hp === undefined ||
+        matches.currentHp === undefined ||
+        matches.x === undefined ||
+        matches.y === undefined ||
+        matches.z === undefined) {
+      // not reached
+      return;
+    }
+
+    if (hunt.id && matches.npcNameId !== hunt.id)
       return;
-    if (monster.regex && !monster.regex.test(log))
+    if (hunt.regex && !hunt.regex.test(log))
       return;
-    if (monster.hp && parseFloat(matches.hp) < monster.hp)
+    if (hunt.hp && parseFloat(matches.hp) < hunt.hp)
       return;
     if (matches.currentHp === '0') // hunt is already dead
       return;
 
     let options = this.options;
     // option overwrite
-    if (monster.rank in options.RankOptions)
-      options = Object.assign({}, this.options, options.RankOptions[monster.rank]);
+    if (hunt.rank && hunt.rank in options.RankOptions)
+      options = Object.assign({}, this.options, options.RankOptions[hunt.rank]);
     // Check if monster rank enabled or not in config
-    if (monster.rank === 'B' && !options.BRankEnabled)
+    if (hunt.rank === 'B' && !options.BRankEnabled)
       return;
     if (options.Enabled === false)
       return;
@@ -151,10 +247,13 @@ class Radar {
     }
 
     const mobKey = matches.name.toLowerCase();
-    if (mobKey in this.targetMonsters) {
+    if (!mobKey)
+      return;
+    const targetMob = this.targetMonsters[mobKey];
+    if (targetMob) {
       // Get positions
       const playerPos = new Point2D(this.playerPos.x, this.playerPos.y);
-      const oldPos = this.targetMonsters[mobKey].pos;
+      const oldPos = targetMob.pos;
       const newPos =
         new Point2D(parseFloat(matches.x), parseFloat(matches.y));
 
@@ -164,19 +263,19 @@ class Radar {
 
       // Update position only if its closer than the current one
       if (newDistance < oldDistance) {
-        this.targetMonsters[mobKey].pos = newPos;
-        this.targetMonsters[mobKey].posZ = matches.z;
+        targetMob.pos = newPos;
+        targetMob.posZ = parseFloat(matches?.z ?? '0');
 
         // Update DOM
-        this.UpdateMonsterDom(this.targetMonsters[mobKey]);
+        this.UpdateMonsterDom(targetMob);
 
         // Play sound only if its far enough
         if (oldPos.distance(newPos) >= kMinDistanceBeforeSound)
-          PlaySound(this.targetMonsters[mobKey], options);
+          PlaySound(targetMob, options);
       }
     } else {
       // Add DOM
-      const arrowId = 'arrow-' + matches.id;
+      const arrowId = `arrow-${matches.id}`;
       const tr = document.createElement('tr');
       let th = document.createElement('th');
       const img = document.createElement('img');
@@ -190,20 +289,23 @@ class Radar {
       th.setAttribute('align', 'left');
       th.appendChild(document.createElement('div'));
       tr.appendChild(th);
-      this.table.insertBefore(tr, this.table.childNodes[0]);
+      const node = this.table.childNodes[0];
+      // TODO: not reached
+      if (node instanceof HTMLElement)
+        this.table.insertBefore(tr, node);
 
       const m = {
         'id': matches.id,
         'name': matches.name,
-        'rank': monster.rank || '',
+        'rank': hunt.rank,
         'hp': parseFloat(matches.hp),
         'currentHp': parseFloat(matches.hp),
         'battleTime': 0,
         'pos': new Point2D(parseFloat(matches.x), parseFloat(matches.y)),
-        'posZ': matches.z,
-        'addTime': Date.now(),
+        'posZ': parseFloat(matches.z),
+        'addTime': +Date.now(),
         'dom': tr,
-        'puller': null,
+        'puller': undefined,
         'skipPuller': matches.hp !== matches.currentHp, // already pulled before being detected
       };
       this.targetMonsters[mobKey] = m;
@@ -211,25 +313,29 @@ class Radar {
 
       const mapX = posToMap(m.pos.x).toFixed(1);
       const mapY = posToMap(m.pos.y).toFixed(1);
-      console.log(`Found: ${monster.name} (${mapX}, ${mapY})`);
+      console.log(`Found: ${m.name} (${mapX}, ${mapY})`);
 
-      PlaySound(this.targetMonsters[mobKey], options);
+      PlaySound(m, options);
     }
   }
 
-  UpdateMonsterPuller(monster, puller) {
+  UpdateMonsterPuller(monster: Monster, puller: string) {
     if (!this.options.Puller)
       return;
-    if (monster.puller !== null || monster.skipPuller)
+    if (monster.puller || monster.skipPuller)
       return;
     monster.puller = puller;
     this.UpdateMonsterDom(monster);
     console.log(`Pulled: ${puller} => ${monster.name}`);
   }
 
-  UpdateMonsterDom(monster) {
+  UpdateMonsterDom(monster: Monster) {
+    // Wait for OnPlayerChange.
+    if (!this.playerPos)
+      return;
+
     let options = this.options;
-    if (monster.rank in options.RankOptions)
+    if (monster.rank && monster.rank in options.RankOptions)
       options = Object.assign({}, this.options, options.RankOptions[monster.rank]);
     const tr = monster.dom;
     // calculate rotation based on facing
@@ -240,21 +346,24 @@ class Radar {
         targetVector.y - playerVector.y,
     );
     if (tr) {
-      tr.childNodes[1].innerHTML = monster.rank + '&nbsp;&nbsp;&nbsp;&nbsp;' + monster.name;
-      if (Math.abs(this.playerPos.z - monster.posZ) > 5)
-        tr.childNodes[1].innerHTML += '&nbsp;&nbsp;' + (this.playerPos.z < monster.posZ ? '↑' : '↓');
-      tr.childNodes[1].innerHTML += '<br>' + deltaVector.length().toFixed(2) + 'm';
-      if (Date.now() / 1000 <= monster.battleTime + 60) {
-        tr.childNodes[1].innerHTML += ' ' + (monster.currentHp * 100 /
-          monster.hp).toFixed(2) + '%';
-      }
-      if (monster.puller)
-        tr.childNodes[1].innerHTML += '&nbsp;&nbsp;' + monster.puller;
-      // Z position is relative to the map so it's omitted.
-      if (options.Position) {
-        tr.childNodes[1].innerHTML += '<br>X: ' +
-          posToMap(monster.pos.x).toFixed(1) + '&nbsp;&nbsp;Y:' +
-          posToMap(monster.pos.y).toFixed(1);
+      const node = tr.childNodes[1];
+      if (node && node instanceof HTMLElement) {
+        node.innerHTML = `${monster.rank ?? ''}&nbsp;&nbsp;&nbsp;&nbsp;${monster.name}`;
+        if (Math.abs(this.playerPos.z - monster.posZ) > 5)
+          node.innerHTML += '&nbsp;&nbsp;' + (this.playerPos.z < monster.posZ ? '↑' : '↓');
+        node.innerHTML += '<br>' + deltaVector.length().toFixed(2) + 'm';
+        if (Date.now() / 1000 <= monster.battleTime + 60) {
+          node.innerHTML += ' ' + (monster.currentHp * 100 /
+            monster.hp).toFixed(2) + '%';
+        }
+        if (monster.puller)
+          node.innerHTML += '&nbsp;&nbsp;' + monster.puller;
+        // Z position is relative to the map so it's omitted.
+        if (options.Position) {
+          node.innerHTML += '<br>X: ' +
+            posToMap(monster.pos.x).toFixed(1) + '&nbsp;&nbsp;Y:' +
+            posToMap(monster.pos.y).toFixed(1);
+        }
       }
     }
     if (options.DetectionRange > 0 && deltaVector.length() > options.DetectionRange)
@@ -267,50 +376,63 @@ class Radar {
     const angle = deltaTheta * 180 / Math.PI;
     const arrowId = 'arrow-' + monster.id;
     const arrow = document.getElementById(arrowId);
-    arrow.style.transform = 'rotate(' + angle + 'deg)';
+    if (arrow)
+      arrow.style.transform = `rotate(${angle}deg)`;
   }
 
-  RemoveMonster(mobKey) {
-    if (mobKey in this.targetMonsters) {
-      console.log(`Killed: ${this.targetMonsters[mobKey].name}`);
-      this.targetMonsters[mobKey].dom.remove();
-      delete this.targetMonsters[mobKey];
-    }
+  RemoveMonster(mobKey: string) {
+    const monster = this.targetMonsters[mobKey];
+    if (!monster)
+      return;
+    console.log(`Killed: ${monster.name}`);
+    monster.dom.remove();
+    delete this.targetMonsters[mobKey];
   }
 
-  OnNetLog(e) {
+  OnNetLog(e: { line: string[]; rawLine: string }) {
     const type = e.line[0];
     const log = e.rawLine;
 
     // added new combatant
     if (type === '03') {
-      const matches = this.regex.addedCombatantFull.exec(log);
-      if (matches && matches.groups) {
-        const monster = this.nameToMonster[matches.groups.name.toLowerCase()];
-        if (monster)
-          this.AddMonster(log, monster, matches.groups);
-      }
-      return;
+      const m = this.regexes.addedCombatantFull.exec(log);
+      const matches = m?.groups;
+      if (!matches)
+        return;
+      const name = matches.name?.toLowerCase();
+      if (!name)
+        return;
+      const hunt = this.nameToHuntEntry[name];
+      if (!hunt)
+        return;
+      this.AddMonster(log, hunt, matches);
     }
 
     // network ability
     if (type === '21' || type === '22') {
-      const matches = this.regex.abilityFull.exec(log);
-      if (matches && matches.groups) {
-        const monster = this.targetMonsters[matches.groups.target.toLowerCase()];
-        if (monster) {
-          // provoke doesn't work on hunt mobs
-          const isProvoke = matches.groups.id === '1D6D';
-          if (!isProvoke)
-            this.UpdateMonsterPuller(monster, matches.groups.source);
-        }
+      const m = this.regexes.abilityFull.exec(log);
+      const matches = m?.groups;
+      if (!matches)
+        return;
+      const name = matches.target?.toLowerCase();
+      if (!name)
+        return;
+      const monster = this.targetMonsters[name];
+      if (!monster)
+        return;
+
+      // provoke doesn't work on hunt mobs
+      const isProvoke = matches?.id === '1D6D';
+      if (!isProvoke) {
+        const source = matches?.source;
+        if (source)
+          this.UpdateMonsterPuller(monster, source);
       }
-      return;
     }
 
     // change instance
     if (type === '00') {
-      if (this.regex.instanceChanged.test(log)) {
+      if (this.regexes.instanceChanged.test(log)) {
         // don't remove mobs lasting less than 10 seconds
         this.ClearTargetMonsters(10);
       }
@@ -319,49 +441,60 @@ class Radar {
 
     // removing combatant
     if (type === '25') {
-      const matches = this.regex.wasDefeated.exec(log);
-      if (matches && matches.groups)
-        this.RemoveMonster(matches.groups.target.toLowerCase());
-      return;
+      const m = this.regexes.wasDefeated.exec(log);
+      const matches = m?.groups;
+      if (!matches)
+        return;
+      const name = matches.target?.toLowerCase();
+      if (name)
+        this.RemoveMonster(name);
     }
   }
 
-  OnPlayerChange(e) {
-    this.playerPos.x = e.detail.pos.x;
-    this.playerPos.y = e.detail.pos.y;
-    this.playerPos.z = e.detail.pos.z;
+  OnPlayerChange(e: Parameters<EventMap['onPlayerChangedEvent']>[0]) {
+    if (!this.playerPos) {
+      this.playerPos = new Point2DWithZ(e.detail.pos.x, e.detail.pos.y, e.detail.pos.z);
+    } else {
+      this.playerPos.x = e.detail.pos.x;
+      this.playerPos.y = e.detail.pos.y;
+      this.playerPos.z = e.detail.pos.z;
+    }
+
     this.playerRotation = e.detail.rotation;
-    for (const i in this.targetMonsters)
-      this.UpdateMonsterDom(this.targetMonsters[i]);
+    for (const monster of Object.values(this.targetMonsters))
+      this.UpdateMonsterDom(monster);
   }
 
-  OnChangeZone(e) {
+  OnChangeZone() {
     this.ClearTargetMonsters();
   }
 
-  ClearTargetMonsters(deltaTime) {
-    const deltaTimeThreshold = deltaTime || 0;
-    for (const i in this.targetMonsters) {
-      if ((Date.now() - this.targetMonsters[i].addTime) / 1000 > deltaTimeThreshold) {
-        this.targetMonsters[i].dom.remove();
-        delete this.targetMonsters[i];
+  ClearTargetMonsters(deltaTime?: number) {
+    const deltaTimeThreshold = deltaTime ?? 0;
+    for (const [mobKey, monster] of Object.entries(this.targetMonsters)) {
+      if ((Date.now() - monster.addTime) / 1000 > deltaTimeThreshold) {
+        monster.dom.remove();
+        delete this.targetMonsters[mobKey];
       }
     }
   }
 }
 
 UserConfig.getUserConfigLocation('radar', Options, () => {
+  const elem = document.getElementById('radar-table');
+  if (!elem)
+    throw new Error('missing radar element');
+  const radar = new Radar(elem);
+
   addOverlayListener('LogLine', (e) => {
-    gRadar.OnNetLog(e);
+    radar.OnNetLog(e);
   });
 
   addOverlayListener('onPlayerChangedEvent', (e) => {
-    gRadar.OnPlayerChange(e);
+    radar.OnPlayerChange(e);
   });
 
-  addOverlayListener('ChangeZone', (e) => {
-    gRadar.OnChangeZone(e);
+  addOverlayListener('ChangeZone', () => {
+    radar.OnChangeZone();
   });
-
-  gRadar = new Radar(document.getElementById('radar-table'));
 });
```

quisquous:
I separated out `HuntEntry` from `Monster` here.  Everything was called "Monster" previously, which ended up being pretty confusing from a type perspective.  Now `HuntEntry` is the static hunt data (either from resources or user entries) and `Monster` is a hunt that has been found and is currently live.

I think this is the first example where we've had to construct options.  I think this is a little bit of a janky setup, but I added the common `UserConfig.getDefaultBaseOptions()` function that could get reused and left some comments about how to handle default options coming from `radar_config.ts`.  `BaseOptions` really wants a few language options to be valid (and I think it's better to treat them as always existing), so they need to be initialized up front.  Maybe @MaikoTan has run into this same sort of issue in converting ui/jobs?

MaikoTan:
Yep, that is better than before. Thanks for notify me.

ui/radar/radar.ts
```diff
@@ -0,0 +1,499 @@
...
+        new Point2D(parseFloat(matches.x), parseFloat(matches.y));
+
+      // Calculate distances
+      const oldDistance = playerPos.distance(oldPos);
+      const newDistance = playerPos.distance(newPos);
+
+      // Update position only if its closer than the current one
+      if (newDistance < oldDistance) {
+        targetMob.pos = newPos;
+        targetMob.posZ = parseFloat(matches?.z ?? '0');
```
panicstevenson, line 267:
```suggestion
        targetMob.posZ = parseFloat(matches.z ?? '0');
```
I don't think matches can be undefined?

ui/radar/radar.ts
```diff
@@ -0,0 +1,499 @@
...
+      const m = {
+        'id': matches.id,
+        'name': matches.name,
+        'rank': hunt.rank,
+        'hp': parseFloat(matches.hp),
+        'currentHp': parseFloat(matches.hp),
+        'battleTime': 0,
+        'pos': new Point2D(parseFloat(matches.x), parseFloat(matches.y)),
+        'posZ': parseFloat(matches.z),
+        'addTime': +Date.now(),
```
panicstevenson, line 305:
```suggestion
        'addTime': Date.now().valueOf(),
```

ui/radar/radar.ts
```diff
@@ -0,0 +1,499 @@
...
+        targetVector.y - playerVector.y,
+    );
+    if (tr) {
+      const node = tr.childNodes[1];
+      if (node && node instanceof HTMLElement) {
+        node.innerHTML = `${monster.rank ?? ''}&nbsp;&nbsp;&nbsp;&nbsp;${monster.name}`;
+        if (Math.abs(this.playerPos.z - monster.posZ) > 5)
+          node.innerHTML += '&nbsp;&nbsp;' + (this.playerPos.z < monster.posZ ? '↑' : '↓');
+        node.innerHTML += '<br>' + deltaVector.length().toFixed(2) + 'm';
+        if (Date.now() / 1000 <= monster.battleTime + 60) {
```
panicstevenson, line 354:
```suggestion
        if (Date.now().valueOf() / 1000 <= monster.battleTime + 60) {
```

ui/radar/radar.ts
```diff
@@ -0,0 +1,499 @@
...
+    }
+    if (options.DetectionRange > 0 && deltaVector.length() > options.DetectionRange)
+      monster.dom.classList.add('hide');
+    else
+      monster.dom.classList.remove('hide');
+
+    let deltaTheta = Math.atan2(deltaVector.y, deltaVector.x);
+    deltaTheta -= Math.PI - this.playerRotation;
+    const angle = deltaTheta * 180 / Math.PI;
+    const arrowId = 'arrow-' + monster.id;
```
panicstevenson, line 376:
```suggestion
    const arrowId = `arrow-${monster.id}`;
```
Surprised this didn't complain?

ui/radar/radar.ts
```diff
@@ -0,0 +1,499 @@
...
+  }
+
+  OnNetLog(e: { line: string[]; rawLine: string }) {
+    const type = e.line[0];
+    const log = e.rawLine;
+
+    // added new combatant
+    if (type === '03') {
+      const m = this.regexes.addedCombatantFull.exec(log);
+      const matches = m?.groups;
```
panicstevenson, line 398: Should this be `if (!(matches && matches.groups))` similar to timeline or should timeline adopt this style?

panicstevenson (approved):

ui/radar/radar.ts
quisquous: I think part of it is that trigger functions in all trigger files and the `AddMonster` function here take `matches` (but is really `groups`).  And so, there's some code that says `matches` is what gets returned from a regex match, and there's other code that say `matches` is the `groups` parameter.  I was trying to make this just match the parameter here.

Do you think everything should adopt this style? (I'm not sure what to call `m` though.)

quisquous:
> https://github.com/quisquous/cactbot/blob/ced74878580cc90e9969ac577385f4aa093ce83b/resources/user_config.ts#L171
> 
> What about to change this line as `getUserConfigLocation<TOptions>(overlayName: string, options: TOptions, callback: (option: BaseOption & TOptions) => void)`? So that we can get the actual typing of `options`.  Additionally, it is not recommended to modify an argument of a function theoretically, I think to let it returned as an argument of the callback would be more "JavaScriptism"?

Hmm.  I think this would make types more readable to humans, but it doesn't seem technically necessary to me in a TypeScript sense.  As far as I can tell, it's not going to be able to catch any differences between the types that we add to `RadarOptions` and the fields that get set via `registerOptions`.  Those are only runtime issues.

Do you think it would make sense to make this change separately first? (Probably `TOptions extends BaseOptions` too?)

> Also, we can declare `registerOptions` function as a generic function with some types like `RadarConfigOptionKeys`, then TS would know how many keys would be add in `Options` variable. (But this would break almost all modules, I think this would be more like a long-term plan?)

I am having a hard time understanding how this would work.  `registerOptions` feels like more of a runtime thing and I'm not sure how TS would know what types would be there for later.  If you have an example PR or more explanation of what this would look like for one module, I would be interested to see it.

ui/radar/radar.ts
panicstevenson: I don't have strong opinions here; I saw it as a means for one pattern to absorb the other and be an emergent pattern going forward for anyone that might run into a similar pattern of code. For Timeline, I think the names were already somewhat useful? Or, if not, I tried to make them useful:
```
commandMatch = regexes.syncCommand.exec(line);
```
And I would expect that from there if you wanted matches to be that, or you wanted that to be groups, you could reassign from there.

Additionally, we can ignore this comment and see what shakes out in further conversions.

quisquous: Yeah, I think this is a larger conversation to have to clean up bits in multiple places.

PR #2842, opened 2021-05-06
O12S Improvements

quisquous:


JLGarber (approved):

ui/raidboss/data/04-sb/raid/o12s.js
```diff
@@ -111,58 +359,80 @@ export default {
...
       netRegex: NetRegexes.headMarker({ id: '0057' }),
       condition: Conditions.targetIsYou(),
-      response: Responses.meteorOnYou('info'),
+      response: Responses.meteorOnYou('alarm'),
+    },
+    {
+      id: 'O12S Optimized Sagittarius Arrow',
+      netRegex: NetRegexes.headMarker({ id: '0017' }),
+      condition: Conditions.targetIsYou(),
+      alertText: (data, _, output) => output.text(),
```
panicstevenson, line 430:
```suggestion
      alarmText: (data, _, output) => output.text(),
```

ui/raidboss/data/04-sb/raid/o12s.js
```diff
@@ -57,10 +151,10 @@ export default {
       netRegexJa: NetRegexes.gainsEffect({ target: 'オメガ', effectId: '67E', capture: false }),
       netRegexCn: NetRegexes.gainsEffect({ target: '欧米茄', effectId: '67E', capture: false }),
       netRegexKo: NetRegexes.gainsEffect({ target: '오메가', effectId: '67E', capture: false }),
-      infoText: (data, _, output) => output.text(),
+      alertText: (data, _, output) => output.text(),
       outputStrings: {
         text: {
-          en: 'Keep Bosses Apart',
+          en: 'Move Bosses Apart',
```
panicstevenson, line 157: Should we clear the below translations?

panicstevenson (approved):

ui/raidboss/data/04-sb/raid/o12s.js
quisquous: Counterpoint: it's good to differentiate meteor and arrow? At least in non-BLU land you do different things.

ui/raidboss/data/04-sb/raid/o12s.js
quisquous: I think this is probably fine as-is.  Some of the below translations already look like they say this.

I wonder if this is a case where we need a `// FIXME` sort of support in the tools, where you can say "this translation is still good enough to keep, but somebody needs to look at it again."

ui/raidboss/data/04-sb/raid/o12s.js
panicstevenson: That's fair; I forgot that they're not both just "run away" normally

ui/raidboss/data/04-sb/raid/o12s.js
MaikoTan: ja and cn are okay to keep, the verbs in the sentences has already the meaning of "pull the boss" or "move the boss to somewhere".

PR #2848, opened 2021-05-08
raidboss: convert top level raidboss file to TypeScript

quisquous:


quisquous:
I am really not sure what to do about the manifest files coming from webpack.  I need to solve that for `ui/config` and oopsy as well.

Also, the repeated types of the default (with some values) is not great.  I am not sure how to statically get that from the various x_config.js files.  I think we could potentially move all of the default values out of `options` and into `defaultOptions` separately and then export that (and a type??).  I think I am happy to TODO our way through this and then try to wrangle it afterwards.

Trim21:
> I am really not sure what to do about the manifest files coming from webpack. I need to solve that for `ui/config` and oopsy as well.

```ts

declare module "*.txt" {

}
```

quisquous:
> > I am really not sure what to do about the manifest files coming from webpack. I need to solve that for `ui/config` and oopsy as well.
> 
> ```ts
> // types/custome.d.ts
> declare module '*/manifest.txt' {
>   import { ManiFest } from './manifest';
>   export default ManiFest;
> }

Hmm! Thanks!! I gave this a try in d563ccaba187ec1c348da3a7c515cb8cdf5df0c4, but I'm not sure that I understand the `declare module` syntax.  I'm unable to say "only make this apply to ui/raidboss/data/manifest.txt and not other manifest.txt files".  I think I don't understand why I need the `*` as well, but I can't figure out how to make it work without it.

MaikoTan:
> Hmm! Thanks!! I gave this a try in d563ccaba187ec1c348da3a7c515cb8cdf5df0c4, but I'm not sure that I understand the `declare module` syntax.  I'm unable to say "only make this apply to ui/raidboss/data/manifest.txt and not other manifest.txt files".  I think I don't understand why I need the `*` as well, but I can't figure out how to make it work without it.

AFAIK, it is because `declare module 'xxx'` is just a declaration, the 'xxx' part is a pattern to match something that you import into other modules. For example, you can write `declare module '*.png';` to support image importing in TypeScript while getting type hints, instead of an error like "Type 'typeof import("*.png")' is not assignable to type 'string'.".

When you're writing `import raidbossManifest from './data/manifest';`, it is not matched by the pattern like `*ui/raidboss/data/manifest.txt`. Therefore, you would lost all the typing.

I think you can add path aliases via webpack's aliases configuration and tsconfig's paths config, but then you might get another pain from ESLint. ¯\_(ツ)_/¯

Trim21:
> I think you can add path aliases via webpack's aliases configuration and tsconfig's paths config, but then you might get another pain from ESLint. `¯\_(ツ)_/¯`

path alias will break current tests.

quisquous:
I'm thinking that means that I need to rename `manifest.txt` to `raidboss_manifest.txt` and `oopsy_manifest.txt` so that they can have different types.  Otherwise, there's no way to differentiate them.

Trim21:
> I'm thinking that means that I need to rename `manifest.txt` to `raidboss_manifest.txt` and `oopsy_manifest.txt` so that they can have different types. Otherwise, there's no way to differentiate them.

import with path alias like `@/ui/raidboss/data/manifest.txt` and `declare module '*raidboss/data/manifest.txt'`

quisquous:
Hmm, but a webpack path alias is just an option, right?  People could still attempt to import it from the path directly and then it would be untyped.  If that's true, that seems a little too magical for me and maybe renaming is the clearer option.

PR #2854, opened 2021-05-10
ci: add code coverage report

Trim21:
install this github app https://github.com/apps/codecov


MaikoTan:
I saw there is codecov bot who would comment in a PR with the coverage report in many repositories, is that implementable in cactbot?

MaikoTan:
Hmm, looks like we should install the [Coodecov App](https://github.com/apps/codecov) to enable this feature?

Trim21:
> Hmm, looks like we should install the [Codecov App](https://github.com/apps/codecov) to enable this feature?

> I make PR comments on public repositories using Codecov! However, I'd prefer you use the GitHub App, instead! (see: https://github.com/apps/codecov)

MaikoTan:
> > Hmm, looks like we should install the [Codecov App](https://github.com/apps/codecov) to enable this feature?
> 
> > codecov-commenter: I make PR comments on public repositories using Codecov! However, I'd prefer you use the GitHub App, instead! (see: https://github.com/apps/codecov)

Yep, that's what I refer to.

Trim21:
> > > Hmm, looks like we should install the [Codecov App](https://github.com/apps/codecov) to enable this feature?
> > 
> > 
> > > codecov-commenter: I make PR comments on public repositories using Codecov! However, I'd prefer you use the GitHub App, instead! (see: https://github.com/apps/codecov)
> 
> Yep, that's what I refer to.

that's what I said in the beginning...

MaikoTan:
> that's what I said in the beginning...

Oh, sorry. I don't know why but I can't see the comment of the bot in Github Android App...
When I switch to PC then I saw that report...(/ω＼*)

quisquous:
I appreciate the enthusiasm here (and I apologize for my pessimism here) but I have to admit that I am a little hesitant to add this.  From my perspective, it's another CI thing to be maintained and extra noise on PRs.  Sorry for being longwinded here, but I don't want to just say no without explaining myself.

Personally speaking, I haven't found a lot of value in coverage reports on different projects I've worked on in the past.  I think "does this code have coverage" is not a good mapping onto "which systems need testing" and "what set of interactions needs to be tested" and "which systems are at risk and thus are the priority for testing".  Even worse, coverage doesn't consider data, like large parts of raidboss and oopsy trigger files.

This project is really not at a good point where we have testing infrastructure for many things.  I think there are a few things (e.g. more regex types) that make sense to add unit tests for where we can, and missing those can be caught in review right now.  For everything else we're really missing the infrastructure where somebody can see missing coverage and "add a quick unit test".

I guess I'm trying to get to this: what actions do you foresee people taking in PRs based on coverage reports? If there aren't useful actions to take from additional information, then the cost (noise on PRs, maintaining CI code) is probably not worth it.

In general, I would not be supportive of somebody looking at a coverage report and adding testing just to make coverage better.  It is hard for me to believe that would generate quality PRs.  For example, if somebody went "oh look, parts of `uwu.js` aren't tested, I'll write a unit test", I would come back and say that we probably needed to come up with a system to test trigger files more thoroughly rather than do a one-off test.  In other words, the solutions that we need are at higher levels and not at the line level, and so information about coverage will not drive us towards those solutions.


Looking at our cactbot testing situation, here are my priorities (maybe you disagree?):

(1) Oopsy has no tests at all (other than the coverage report 👀), it probably needs basic smoketests like raidboss has.  

(2) I think raidboss itself could use more extensive tests on a per-trigger basis; I'm picturing something like Chromium/WebKit's web tests with expectation files.  The inputs are log line inputs (trimmed to only what's necessary) -> JSON output that says alert/alarm/info/tts and the text that was output.  You check in the expected output and then run and compare the actual output to the checked in output.  The reason for this testing is to make sure that refactorings of raidboss systems or the trigger files themselves don't regress what the trigger files used to do.  In particular, I'm thinking of how I broke all the ultimates when I added output strings.  I think this would also be like "raid emulator light" (and probably should reuse the analysis code from raid emulator) and so could also help when writing triggers to make sure that everything worked as expected.

(3) Tests for things like radar, fishing, jobs.  I'm not sure what this would look like or how to do it.

(4) Testing what ends up on screen ultimately.  It'd be good to have a few end-to-end smoke tests to make sure that timeline bars appear properly, skins work, etc etc.

quisquous:
Sorry if the wall of text came off as rude.  If you feel like this would be useful, I'm really happy to chat about it more.

Trim21:
> Sorry if the wall of text came off as rude. If you feel like this would be useful, I'm really happy to chat about it more.

I didn't take it that way. (I'm not a native English speaker, hope I didn't make any misunderstanding)

I think it would be useful for basic utils in `resources/`.

quisquous:
Ok, I'm glad it didn't seem rude because it wasn't intended to be.  I think I just didn't know what to read into the "no comment -> close", so I wanted to check.

Have you found things in `resources/` from running this that need coverage?

Is this something that can be run once or twice as needed vs on every PR?

PR #2856, opened 2021-05-12
Fix(jobs): MPTicker/repertoreTimer bar goes left

MaikoTan:
image

It seems like the last PR changed the default value of `toward` property in `resources/timerbar.ts`, so the MPTicker doesn't go right as before.

### DIFF
--- a/ui/jobs/components/brd.js
+++ b/ui/jobs/components/brd.js
@@ -60,6 +60,7 @@ export function setup(bars) {
     id: 'brd-timers-repertoire',
     fgColor: 'brd-color-song',
   });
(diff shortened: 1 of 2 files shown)


Echoring:
BRD's repertoireTimer shares the same issue as well.

MaikoTan:
> BRD's repertoireTimer shares the same issue as well.

Fixed! (/≧▽≦)/~快夸我~

quisquous:
I think it was more the case that `toward` was broken before and so these had the wrong default values, so when that was fixed these started """working""" again.  (I think this is also true for durations in timelines, but I think it looks better to have that one go the way it does now?)

PR #2857, opened 2021-05-13
user_config: return options rather than using global options

quisquous:
The lint is changed for rdmty because changing a file that is ignored
is a warning and we have a maximum number of warnings of zero.
So, instead eslint-disable is added to that file.

### DIFF
--- a/ui/raidboss/emulator/data/RaidEmulator.js
+++ b/ui/raidboss/emulator/data/RaidEmulator.js
@@ -2,9 +2,10 @@ import EventBus from '../EventBus';
 import AnalyzedEncounter from './AnalyzedEncounter';
 
 export default class RaidEmulator extends EventBus {
-  constructor(options) {
+  constructor() {
     super();
-    this.options = options;
+    // Set later via setOptions.
+    this.options = undefined;
     this.encounters = [];
     this.currentEncounter = null;
     this.playing = false;
@@ -13,10 +14,14 @@ export default class RaidEmulator extends EventBus {
     this.lastLogTimestamp = null;
     this.internalTimestampTracker = null;
   }
+  setOptions(options) {
+    this.options = options;
+  }
   addEncounter(encounter) {
     this.encounters.push(encounter);
   }
   setCurrent(index) {
+    console.assert(this.options);
     this.currentEncounter = new AnalyzedEncounter(this.options, this.encounters[index], this);
     this.dispatch('preCurrentEncounterChanged', this.currentEncounter);
     this.currentEncounter.analyze(this.popupText).then(() => {
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -73,7 +73,6 @@ module.exports = {
     'plugin/',
     'publish/',
     'resources/lib/',
-    'ui/dps/rdmty/',
   ],
   'env': {
     'browser': true,
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -168,7 +168,8 @@ class UserConfig {
     });
   }
 
-  getUserConfigLocation(overlayName: string, options: BaseOptions, callback: () => void) {
+  getUserConfigLocation<OverlayOptions extends BaseOptions>(overlayName: string,
+      initialOptions: OverlayOptions, callback: (options: OverlayOptions) => void) {
     let currentlyReloading = false;
     const reloadOnce = () => {
       if (currentlyReloading)
@@ -184,10 +185,11 @@ class UserConfig {
       reloadOnce();
     });
 
-    this.loadUserFiles(overlayName, options, callback);
+    this.loadUserFiles(overlayName, initialOptions, callback);
   }
 
-  loadUserFiles(overlayName: string, options: BaseOptions, callback: () => void) {
+  loadUserFiles<OverlayOptions extends BaseOptions>(overlayName: string,
+      initialOptions: OverlayOptions, callback: (options: OverlayOptions) => void) {
     const readOptions = callOverlayHandler({
       call: 'cactbotLoadData',
       overlay: 'options',
@@ -199,6 +201,8 @@ class UserConfig {
       // Linux (?!?), support any style of slashes elsewhere.
       const basePath = e.detail.userLocation.replace(/[/\\]*$/, '') + '\\';
       const localFiles = e.detail.localUserFiles;
+
+      const options: OverlayOptions = { ...initialOptions };
 
       // The plugin auto-detects the language, so set this first.
       // If options files want to override it, they can for testing.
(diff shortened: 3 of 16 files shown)


quisquous:
@MaikoTan is this what you were thinking of in https://github.com/quisquous/cactbot/pull/2841#issuecomment-838404910?

MaikoTan:
Hmmm, It looks well, but I think there may be a better solution to achieve this. Sorry I am very busy these days, might I can take a look tomorrow or weekend.

ui/raidboss/emulator/data/RaidEmulator.js
```diff
@@ -13,10 +14,14 @@ export default class RaidEmulator extends EventBus {
...
     this.internalTimestampTracker = null;
   }
+  setOptions(options) {
+    this.options = options;
+  }
   addEncounter(encounter) {
     this.encounters.push(encounter);
   }
   setCurrent(index) {
+    console.assert(this.options);
```
valarnin, line 24: Maybe this `console.assert` should have a message of some sort? Otherwise raidemulator-related changes look good.

quisquous:
> Hmmm, It looks well, but I think there may be a better solution to achieve this. Sorry I am very busy these days, might I can take a look tomorrow or weekend.

Hmm! I would be interested to know what you're thinking of here.

Do you think it makes sense to land #2841 now and wait on this? Or, wait on both?

quisquous:
I think I am going to go ahead and land the radar and top level raidboss changes and hold off on this until @MaikoTan has a chance to give more feedback.  I don't think those changes will make it harder to do something around options, and having some existing TypeScript `Options` in the codebase will help prove a good solution here.

MaikoTan:
> I think I am going to go ahead and land the radar and top level raidboss changes and hold off on this until @MaikoTan has a chance to give more feedback. I don't think those changes will make it harder to do something around options, and having some existing TypeScript `Options` in the codebase will help prove a good solution here.

Sorry for hanging so long, I was really busy these days.

What I thought the first time is that we can make class `UserConfig` as a `generic` class and receive `ConfigOptions` as one of its generic types, so we don't need to add those configuration types to the `Whatever-DefaultOptions` type.

But I just found that `user_config.ts` file exports a `UserConfig` instance instead of the class type, so my idea, in this case, would be a bit complicated overhead. So the cost-performance ratio here might be lower, might not be worth applying it.

Sorry again for letting you wait so long.

quisquous:
Oh! Ok, I think I see where you're coming from with that.

One issue is that UserConfig is really behaving like a static class / a singleton here.  And so we could export the class instead of an instance of the class, but we'd need the options registering and the top level init function both use the same instance of that class.  Options registering could just be a static function, I guess?

The other issue is that I don't see how to create the RaidbossOptions type from inside of UserConfig.  Logically, we know that a set of options has defaults and will set values, but those are all done dynamically in a way that TypeScript can't know about inside of `processOptions`.  I think *something* has to create the defaults statically outside of UserConfig.

One possibility for refactoring might be:
* raidboss_config.ts now defines `defaultRaidbossConfigOptions` and `RaidbossConfigOptions` instead of `raidboss.ts`
  * defaults get pulled out of `templateOptions` in that file into `defaultRaidbossConfigOptions`
  * we remove the code that sets defaults in `processOptions` and assume that default values in options are set prior to calling `UserConfig.getUserConfigLocation`
  * maybe we add some asserts in `UserConfig.processOptions` (`or registerOptions`) that all `templateOptions` have a default pre-set
* raidboss_options.ts then imports `defaultRaidbossConfigOptions` and `RaidbossConfigOptions` rather than duplicating it
  * or maybe we just move this file into `raidboss_config.ts`?
* `UserConfig.registerOptions` is now a static function and returns a `UserConfig<T>` instance
  * this is used in two places: (1) for an individual overlay to use later and (2) for `ui/config/config.js` to build ui from
  * calling `registerOptions` ALSO returns a UserConfig<RaidbossConfigOptions> for `raidboss.ts` to use for case 1 above
  * `raidboss_config.ts` exports this UserConfig<RaidbossConfigOptions> that it gets from `registerOptions` and `raidboss.ts` uses it instead of importing from `user_config.ts`?
* `UserConfig<RaidbossConfigOptions>.getUserConfigLocation` no longer needs to have `defaultOptions` as a parameter, because it already got the default options as a part of `registerOptions` and can then have a callback of `(options: RaidbossConfigOptions) => void` because it is parameterized?

How does that sound?

MaikoTan:
Sorry I almost forgot this PR /_ \

> * raidboss_config.ts now defines `defaultRaidbossConfigOptions` and `RaidbossConfigOptions` instead of `raidboss.ts`
>   
>   * defaults get pulled out of `templateOptions` in that file into `defaultRaidbossConfigOptions`
>   * we remove the code that sets defaults in `processOptions` and assume that default values in options are set prior to calling `UserConfig.getUserConfigLocation`
>   * maybe we add some asserts in `UserConfig.processOptions` (`or registerOptions`) that all `templateOptions` have a default pre-set

Yeah, sounds good. And for the typing, I think we can have a util function that adds types of `defaultOptions` to `raidbossOption` in the callback of `UserConfig.getUserConfigLocation`, then it can be `callback: (options: OverlayOptions & BaseOptions) => void` or something?

> * raidboss_options.ts then imports `defaultRaidbossConfigOptions` and `RaidbossConfigOptions` rather than duplicating it
>   * or maybe we just move this file into `raidboss_config.ts`?

Maybe we should change the construction of the whole object that passed into `UserConfig.registerOptions`, to make it easier to be determined by TypeScript?

For example, an object like `options: Record<OptionsKey, { name: LocaleObject, type: string, }>` would be clearer than `options: { id: OptionsKey, name: LocaleObject, type: string }[]` for TypeScript typing?

> * `UserConfig.registerOptions` is now a static function and returns a `UserConfig<T>` instance
>   
>   * this is used in two places: (1) for an individual overlay to use later and (2) for `ui/config/config.js` to build ui from
>   * calling `registerOptions` ALSO returns a UserConfig for `raidboss.ts` to use for case 1 above
>   * `raidboss_config.ts` exports this UserConfig that it gets from `registerOptions` and `raidboss.ts` uses it instead of importing from `user_config.ts`?

I am a bit doubting that how is the (2) `ui/config/config.js` building UI from `UserConfig.registerOptions` when it becomes a static function, as then you should handle a bunch of `UserConfig<Raidboss>`, `UserConfig<Radar>`, `UserConfig<Jobs>` and so on?

But I am still thinking that this would be painful when we are doing this refactoring, is it even worth?
Perhaps we can do it step by step...╮(╯-╰)╭

quisquous:
I think I'll abandon this PR for now and come back with some followups later based on what we've discussed here.  :)

PR #2859, opened 2021-05-14
raidboss: Convert Timeline to Typescript

panicstevenson:


panicstevenson:
Opened as a draft since there's no rush to merge this and it touches a lot with some opinions.

ui/raidboss/timeline.ts
```diff
@@ -443,7 +542,7 @@ export class Timeline {
       this.syncTimeCallback(fightNow, false);
   }
 
-  SyncTo(fightNow, currentTime) {
+  protected SyncTo(fightNow: number, currentTime: number): void {
```
quisquous, line 545: I don't feel super strongly about protected vs public.  Private at least has some immediate meaning that says "you can see all of the uses in this file (probably)".  Protected is like "you probably shouldn't use this I guess, but who knows how it's actually being used".

ui/raidboss/timeline.ts
```diff
@@ -424,8 +523,8 @@ export class Timeline {
     });
   }
 
-  Stop() {
-    this.timebase = null;
+  public Stop(): void {
+    this.timebase = 0;
```
quisquous, line 527: I think this seems reasonable.  Stopped is still `0`, and there's nothing special about `null` here.

panicstevenson: I had some concerns about it still working with the raidboss emulator, but I just tested it and it seems to be working as expected.

types/trigger.d.ts
```diff
@@ -55,3 +55,16 @@ export type ResponseOutput<Matches> = {
 // The type of a response trigger field.
 export type ResponseFunc<Matches> =
     (data: Data, matches: Matches, output: Output) => ResponseOutput<Matches>;
+
+export type Trigger = {
+  id: string;
+  regex: RegExp;
+  beforeSeconds: number;
+}
```
quisquous, line 63: Can you leave a TODO here so it's more obvious to other people that this isn't used?

types/trigger.d.ts
```diff
@@ -55,3 +55,16 @@ export type ResponseOutput<Matches> = {
 // The type of a response trigger field.
 export type ResponseFunc<Matches> =
     (data: Data, matches: Matches, output: Output) => ResponseOutput<Matches>;
+
+export type Trigger = {
```
quisquous, line 59: I think `beforeSeconds` is only for timeline triggers.  Maybe call this `TimelineTrigger` instead?

ui/raidboss/timeline.ts
```diff
@@ -63,8 +68,72 @@ const activeText = {
...
+type Text = ParsedText & { time: number };
+
+interface RaidbossOptions extends BaseOptions {
+  TimelineLanguage: Lang;
+  MaxNumberOfTimerBars: number;
+  BarExpiresSoonSeconds: number;
+  ShowTimerBarsAtSeconds: number;
+  KeepExpiredTimerBarsForSeconds: number;
+  PerTriggerAutoConfig: { [triggerId: string]: TriggerAutoConfig };
+}
```
quisquous, line 133: Sorry for the conflict but I just checked in the top level raidboss.ts conversion, which adds a RaidbossOptions type to `raidboss_options.ts`.  Can this get merged into there?

quisquous (commented):
Is this still a draft, or ready to go?

PR #2861, opened 2021-05-14
add option to colorize the output of timeline translate script

Akurosia:
This will make it easier to spot missing translations in the terminal

### DIFF
--- a/util/translate_timeline.js
+++ b/util/translate_timeline.js
@@ -26,6 +26,26 @@ parser.addArgument(['-c', '--colorize'], {
 
 const rootDir = 'ui/raidboss/data';
 
+const colorize = (lineText, line, lineSync) => {
+  if (lineText)
+    line = line.replace(` "${lineText.text}"`, ` \x1b[93m"${lineText.text}"\x1b[0m`);
+
+  if (lineSync)
+    line = line.replace(`sync /${lineSync.regex.source}/`, `\x1b[31msync\x1b[0m \x1b[32m/${lineSync.regex.source}/\x1b[0m`);
+
+  // if a # is in the line, dont make uneccassary colorizatiosn
+  if (line.includes('#')) {
+    line = line.replace('#', `\x1b[90m#`) + '\x1b[0m';
+  } else {
+    line = line.replace(/ window (\d+(,\d+(\.\d+)?)?)/, `\x1b[31m window\x1b[0m \x1b[35m$1\x1b[0m`);
+    line = line.replace(/ duration (\d+(\.\d+(\.\d+)?)?)/, `\x1b[31m duration\x1b[0m \x1b[35m$1\x1b[0m`);
+    line = line.replace(/ jump (\d+(\.\d+(\.\d+)?)?)/, `\x1b[31m jump\x1b[0m \x1b[35m$1\x1b[0m`);
+  }
+  // colorize the lineText.time numbers
+  line = line.replace(/^(\d+\.\d+)/, `\x1b[35m$1\x1b[0m`);
+  return line;
+};
+
 const findTriggersFile = (shortName) => {
   // strip extensions if provided.
   shortName = shortName.replace(/\.(?:[jt]s|txt)$/, '');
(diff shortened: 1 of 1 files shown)


Akurosia:
found a case where the "logic" is not working correctly 
`2074.9 "Perpetueller Strahl x5" duration 8.4 #sync /:The Manipulator:13B6:/` will not colorize duration 8,4 correctly

quisquous:
I'm not trying to be rude, but what about just using `grep` to find lines with MISSINGTEXT/MISSINGSYNC?

Akurosia:
its not about the fully missing lines.
E.g. some translations are only partily there and looking at a 200 line long white list where times, descriptions, synx etc is all white in white makes it hard to confirm that e.g. all translations are correct
As an example:
WindowsTerminal_6PaHDKukNV
WindowsTerminal_MvFlls0qC7

quisquous:
Mostly my concern here is that you're reimplementing timeline parsing here and so it has bugs (e.g. `1000.0 "Some thing #1"` will treat # like a comment).

Are you mostly looking at texts (and not syncs) for translations? I think I would be more supportive of something that trimmed the output down to just the times and texts themselves rather than trying to color each individual part of the line.

Akurosia:
but the sync stuff needs to be verified too (best example was the issues with Opressor 0.5 in the past)
i am also fine with just closing it, i just thought it would be quite handy to have a better visualisation

util/translate_timeline.js
```diff
@@ -87,16 +92,40 @@ const run = async (args) => {
...
+      // if a # is in the line, dont make uneccassary colorizatiosn
+      if (line.includes('#')) {
+        line = line.replace('#', `\x1b[90m#`) + '\x1b[0m';
+      } else {
+        line = line.replace(/ window (\d+(,\d+)?)/, `\x1b[31m window\x1b[0m \x1b[35m$1\x1b[0m`);
+        line = line.replace(/ duration (\d+(\.\d+)?)/, `\x1b[31m duration\x1b[0m \x1b[35m$1\x1b[0m`);
+        line = line.replace(/ jump (\d+(\.\d+)?)/, `\x1b[31m jump\x1b[0m \x1b[35m$1\x1b[0m`);
+      }
+      // colorize the lineText.time numbers
+      line = line.replace(/^(\d+\.\d+)/, `\x1b[35m$1\x1b[0m`);
```
quisquous, line 112: Honestly it's this section that feels like the most re-parsing.  Would coloring the text, sync, and MISSINGTEXT/MISSINGSYNC be enough (and this block could be removed)?

I can see in this section that:
* if a line has a comment in a text, it will colorize incorrectly
* if a line has a comment, then window/duration/jump don't get colorized
* if a line has a number inside of a text or a comment it will colorize that too

util/translate_timeline.js
```diff
@@ -87,16 +92,40 @@ const run = async (args) => {
...
-      line += ' #MISSINGSYNC';
-    if (textErrors[lineNumber])
-      line += ' #MISSINGTEXT';
+
+    if (args?.colorize === 'true') {
+      if (lineText)
+        line = line.replace(` "${lineText.name}"`, ` \x1b[93m"${lineText.text}"\x1b[0m`);
+      const lineSync = lineToSync[lineNumber];
+      if (lineSync)
+        line = line.replace(`sync /${lineSync.origRegexStr}/`, `\x1b[31msync\x1b[0m \x1b[32m/${lineSync.regex.source}/\x1b[0m`);
```
quisquous, line 101: This whole `(args?.colorize)` block feels like a repeat of the below.

What about adding a `colorize()` function? e.g.
```
const colorize = (str, colorStr) => {
  if (args?.colorize !== 'true')
    return str;
  // do some lookup of color names here,
  const colorCode = /* something with colorStr */
  return `${colorCode}${str}${endColorCode}`;
};

// ... later ...

      if (lineText)
        line = line.replace(` "${lineText.name}"`, ` "${colorize(lineText.text, 'yellow')}"`);
```

util/translate_timeline.js
Akurosia: problem with comments are, that commented sync lines are not translated and could lead to confusion e.g. `#sync /:Lamebrix Strikebocks:1AB[89AB]:/` from a10s

the number inside text should not happen as the last line-.replace matches on line start as it includes `^`.
But the other issues are there, thats right

quisquous: > problem with comments are, that commented sync lines are not translated and could lead to confusion e.g. `#sync /:Lamebrix Strikebocks:1AB[89AB]:/` from a10s

Yeah, that's part of the reason why I was wondering if there was some better way to handle this whole problem, e.g. printing only the texts or only the syncs and not everything else.

util/translate_timeline.js
```diff
@@ -26,6 +26,26 @@ parser.addArgument(['-c', '--colorize'], {
 
 const rootDir = 'ui/raidboss/data';
 
+const colorize = (lineText, line, lineSync) => {
```
quisquous, line 29: Hmm sorry, this is not what I meant at all.  I was looking for a helper function that takes in a string and a color and optionally turns it into a colorized string, e.g.

```typescript
const colorize = (str, colorStr) => {
  if (args?.colorize !== 'true')
    return str;
  // do some lookup of color names here,
  const colorCode = /* something with colorStr */
  return `${colorCode}${str}${endColorCode}`;
};
```

And then there's only one block for printing out everything that always calls colorize for each replacement, rather than a `if (colorize) { print everything colorized } else { print everything not colorized`.  I want a `print everything colorized (where the helper function to colorize sometimes does it or not)`.

Akurosia: ok than tbh i am lost ^^'
far to inexpirienced to do this kind of magic as i would not even know where to start with it than.

PR #2862, opened 2021-05-15
raidemulator: Fix delaySeconds/suppressSeconds implementation

valarnin:
This PR fixes the delaySeconds and suppressSeconds implementations for raidemulator, resolving the last two outstanding tasks in #2161.

Also fixes two other issues that have cropped up over time while cleaning up the other tasks in #2161.

Fixing delaySeconds required rewriting a pretty large chunk of code related to how triggers are tracked and processed. I've run this set of changes through the following scenarios:

1. e12s clear
2. e12s partial clear, no delaySeconds got clipped short
3. Several encounters with no triggers
4. Several encounters with only delaySeconds triggers
5. Several encounters with only suppressSeconds triggers

The only scenario I couldn't test is where a trigger with `delaySeconds` never has its delay resolve before the end of the encounter. I don't anticipate there being any issues with this but if someone has a log sitting around where this happens, I'd appreciate a test run.

### DIFF
--- a/ui/raidboss/emulator/data/AnalyzedEncounter.js
+++ b/ui/raidboss/emulator/data/AnalyzedEncounter.js
@@ -107,17 +107,19 @@ export default class AnalyzedEncounter extends EventBus {
         popupText.runResolver = res;
       });
 
-      popupText.OnTrigger(trigger, matches);
+      const currentLine = this.encounter.logLines[currentLogIndex];
+
+      popupText.OnTrigger(trigger, matches, currentLine.timestamp);
 
       await delayPromise;
       await promisePromise;
       const triggerHelper = await runPromise;
 
-      popupText.currentTriggerStatus.finalData = EmulatorCommon.cloneData(popupText.data);
+      triggerHelper.resolver.status.finalData = EmulatorCommon.cloneData(popupText.data);
 
       popupText.callback(
-          this.encounter.logLines[currentLogIndex],
-          triggerHelper, popupText.currentTriggerStatus);
+          currentLine,
+          triggerHelper, triggerHelper.resolver.status);
     });
 
     popupText.callback = (log, triggerHelper, currentTriggerStatus, finalData) => {
@@ -129,6 +131,7 @@ export default class AnalyzedEncounter extends EventBus {
           (currentTriggerStatus.delay * 1000),
       });
     };
+    popupText.triggerResolvers = [];
 
     this.perspectives[ID] = {
       initialData: EmulatorCommon.cloneData(popupText.data, []),
--- a/ui/raidboss/emulator/data/PopupTextAnalysis.js
+++ b/ui/raidboss/emulator/data/PopupTextAnalysis.js
@@ -1,20 +1,64 @@
 import EmulatorCommon from '../EmulatorCommon';
 import StubbedPopupText from '../overrides/StubbedPopupText';
 
+class Resolver {
+  constructor() {
+    this.promise = null;
+    this.run = null;
+    this.delayUntil = null;
+    this.final = null;
+  }
+  async isResolved(log) {
+    if (this.delayUntil) {
+      if (this.delayUntil < log.timestamp) {
+        this.delayUntil = null;
+        this.delayResolver();
+        await this.delayPromise;
+      } else {
+        return false;
+      }
+    }
+    if (this.promise)
+      await this.promise;
+    if (this.run)
+      this.run();
+    if (this.final)
+      this.final();
+    return true;
+  }
+  setDelay(delayUntil) {
+    this.delayUntil = delayUntil;
+    return this.delayPromise = new Promise((res) => {
+      this.delayResolver = res;
+    });
+  }
+  setPromise(promise) {
+    this.promise = promise;
+  }
+  setRun(run) {
+    this.run = run;
+  }
+  setFinal(final) {
+    this.final = final;
+  }
+  setHelper(triggerHelper) {
+    this.triggerHelper = triggerHelper;
+  }
+}
+
 export default class PopupTextAnalysis extends StubbedPopupText {
-  OnTriggerInternal(trigger, matches, currentTime) {
-    this.currentTriggerStatus = {
-      initialData: EmulatorCommon.cloneData(this.data),
-      condition: undefined,
-      response: false,
-      result: false,
-      delay: false,
-      suppressed: false,
-      executed: false,
-      promise: undefined,
-    };
-    this.currentFunction = 'initial';
-    super.OnTriggerInternal(trigger, matches, currentTime);
+  constructor(options, timelineLoader, raidbossFileData) {
+    super(options, timelineLoader, raidbossFileData);
+    this.triggerResolvers = [];
+  }
+
+  // Override `OnTrigger` so we can use our own exception handler
+  OnTrigger(trigger, matches, currentTime) {
+    try {
+      this.OnTriggerInternal(trigger, matches, currentTime);
+    } catch (e) {
+      console.log(trigger, e);
+    }
   }
 
   async OnLog(e) {
@@ -27,29 +71,30 @@ export default class PopupTextAnalysis extends StubbedPopupText {
       for (const trigger of this.triggers) {
         const r = log.match(trigger.localRegex);
         if (r) {
-          // Some async magic here, force waiting for the entirety of
-          // the trigger execution before continuing
-          const delayPromise = new Promise((res) => {
-            this.delayResolver = res;
+          const resolver = this.currentResolver = new Resolver();
+          resolver.status = {
+            initialData: EmulatorCommon.cloneData(this.data),
+            condition: undefined,
+            response: false,
+            result: false,
+            delay: false,
+            suppressed: false,
+            executed: false,
+            promise: undefined,
+          };
+          this.triggerResolvers.push(resolver);
+
+          this.OnTrigger(trigger, r, logObj.timestamp);
+
+          resolver.setFinal(() => {
+            resolver.status.finalData = EmulatorCommon.cloneData(this.data);
+            delete resolver.triggerHelper.resolver;
+            this.callback(logObj, resolver.triggerHelper, resolver.status, this.data);
           });
-          const promisePromise = new Promise((res) => {
-            this.promiseResolver = res;
-          });
-          const runPromise = new Promise((res) => {
-            this.runResolver = res;
-          });
-
-          this.OnTrigger(trigger, r);
-
-          await delayPromise;
-          await promisePromise;
-          const triggerHelper = await runPromise;
-
-          this.currentTriggerStatus.finalData = EmulatorCommon.cloneData(this.data);
-
-          this.callback(logObj, triggerHelper, this.currentTriggerStatus);
         }
       }
+
+      await this.checkResolved(logObj);
     }
   }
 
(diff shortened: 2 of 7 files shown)


ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -27,29 +75,37 @@ export default class PopupTextAnalysis extends StubbedPopupText {
       for (const trigger of this.triggers) {
         const r = log.match(trigger.localRegex);
         if (r) {
-          // Some async magic here, force waiting for the entirety of
-          // the trigger execution before continuing
-          const delayPromise = new Promise((res) => {
-            this.delayResolver = res;
+          const resolver = this.currentResolver = new Resolver();
+          resolver.status = {
```
quisquous, line 79: Confirming: this status lives on `Resolver` because some triggers are delayed and so you need to maintain the status of multiple triggers now? I'm mostly confused because the status seems unrelated to resolving (more like trigger helper??), so it's a bit of a mismatch at first glance.

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -60,132 +116,121 @@ export default class PopupTextAnalysis extends StubbedPopupText {
...
 
   _onTriggerInternalSuppressSeconds(triggerHelper) {
-    this.currentFunction = '_onTriggerInternalSuppressSeconds';
+    this.currentTriggerStatus = triggerHelper.resolver.status;
     super._onTriggerInternalSuppressSeconds(triggerHelper);
   }
 
   _onTriggerInternalPromise(triggerHelper) {
-    this.currentFunction = '_onTriggerInternalPromise';
+    this.currentTriggerStatus = triggerHelper.resolver.status;
```
quisquous, line 194: I'm a little confused what `this.currentTriggerStatus` is doing here, and elsewhere.  You set `this.currentTriggerStatus` in a lot of places, but in reading through the code it seems like most of those could be dropped and instead `triggerHelper.resolver.status.promise = ret` could be used instead.  What's the side effect of setting this that I don't understand?

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -1,20 +1,68 @@
...
+  constructor() {
+    this.promise = null;
+    this.run = null;
+    this.delayTo = null;
+  }
+  async IsResolved(log) {
+    if (this.delayTo) {
+      if (this.delayTo < log.timestamp) {
+        this.delayTo = null;
+        this.delay();
```
quisquous, line 14: Can this be called `this.delayResolver()` like it was before?

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -1,20 +1,68 @@
...
+        await this.delayPromise;
+      } else {
+        return false;
+      }
+    }
+    if (!this.delayTo) {
```
quisquous, line 20: this.delayTo is always false here, right? I'm not sure I understand this conditional or the return false at the bottom of this function.

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -1,20 +1,68 @@
...
+  }
+  setDelay(delayTo) {
+    this.delayTo = delayTo;
+    return this.delayPromise = new Promise((res) => {
+      this.delay = res;
+    });
+  }
+  setPromise(promise) {
+    this.promise = async () => {
+      await promise;
```
quisquous, line 39: Can this just be this.promise = promise?

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -1,20 +1,68 @@
...
+        await this.delayPromise;
+      } else {
+        return false;
+      }
+    }
+    if (!this.delayTo) {
+      if (this.promise)
+        await this.promise();
+      if (this.run)
+        await this.run();
```
quisquous, line 24: Is this asynchronous?

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -27,29 +75,37 @@ export default class PopupTextAnalysis extends StubbedPopupText {
...
         }
       }
+
+      for (let i = 0; i < this.triggerResolvers.length; ++i) {
+        const ret = await this.triggerResolvers[i].IsResolved(logObj);
+        if (ret)
+          this.triggerResolvers[i] = false;
+      }
+
+      this.triggerResolvers = this.triggerResolvers.filter((resolver) => resolver);
```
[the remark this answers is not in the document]
valarnin: Can't use filter directly because we can't use `await` in the filter function.

The for loop could alter the array in-line instead but I felt that was a worse solution?

ui/raidboss/emulator/overrides/RaidEmulatorTimelineController.js
```diff
@@ -43,11 +43,10 @@ export default class RaidEmulatorTimelineController extends TimelineController {
     e.detail.logs.forEach((line) => {
       this.activeTimeline.emulatedTimeOffset = line.offset;
       this.ui.emulatedTimeOffset = line.offset;
-      this.activeTimeline.timebase = this.activeTimeline.timebase || line.timestamp;
       this.activeTimeline.OnLogLine(
           line.properCaseConvertedLine || line.convertedLine,
           line.timestamp);
-      this.activeTimeline._OnUpdateTimer(line.timestamp);
+      this.activeTimeline.timebase && this.activeTimeline._OnUpdateTimer(line.timestamp);
```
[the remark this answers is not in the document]
valarnin: I'm avoiding calling it because if `_OnUpdateTimer` is called without a `timebase` value set, it throws a console log, and 4000-5000 console logs per encounter analysis is a bit excessive.

ui/raidboss/emulator/ui/EmulatedPartyInfo.js
```diff
@@ -243,15 +243,13 @@ export default class EmulatedPartyInfo extends EventBus {
...
-      callback: () => {
-        $initDataViewer.textContent = JSON.stringify(per.initialData, null, 2);
-      },
     }));
 
     const $triggerContainer = $container.querySelector('.d-flex.flex-column');
 
     for (const i in per.triggers.sort((l, r) => l.resolvedOffset - r.resolvedOffset)) {
       const $triggerDataViewer = this.$jsonViewerTemplate.cloneNode(true);
+      $triggerDataViewer.textContent = JSON.stringify(per.triggers[i], null, 2);
```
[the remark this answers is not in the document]
valarnin: From the PR comments:
> Also fixes two other issues that have cropped up over time while cleaning up the other tasks in #2161.

More specifically, this part of the commit message:
> raidemulator: Fix JSON for triggers not showing properly

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: This is an awkward setup, because the way the analysis was written originally it was really cornered into handling only one trigger at a time. This is difficult because we need to be able to reference both the resolver and the status at various points and the scoping for the current trigger needs to change depending on which async function is currently executing.

As such, the `Resolver` and `status` properties need to be accessible async, and the only `PopupText` property we can piggyback off of to handle this is `triggerHelper`. So basically they both get references to each other and get stored on `triggerHelper` as well, so that we can dynamically change scope in the async executors.

The only other real way I can see to handle this without some other significant change to the way the base raidboss `PopupText` class works would be to have an instance of `PopupTextAnalysis` per trigger per player which is... excessive and slow.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: See my comment on the review above this, https://github.com/quisquous/cactbot/pull/2862#discussion_r632994855

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: Yes, `this.delayTo` is always false-y here. This is a leftover from a previous iteration, I'll clean it up.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: Just to clarify, the three properties should be:
```js
this.delayTo;
this.delayPromise;
this.delayResolver;
```

Correct?

ui/raidboss/emulator/data/PopupTextAnalysis.js
```diff
@@ -1,20 +1,68 @@
...
-      suppressed: false,
-      executed: false,
-      promise: undefined,
+class Resolver {
+  constructor() {
+    this.promise = null;
+    this.run = null;
+    this.delayTo = null;
+  }
+  async IsResolved(log) {
```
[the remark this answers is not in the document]
valarnin: Will fix for next commit.

ui/raidboss/emulator/overrides/RaidEmulatorTimelineController.js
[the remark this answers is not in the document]
valarnin: Changed to this for the next commit:

```js
      // Only call _OnUpdateTimer if we have a timebase from the previous call to OnLogLine
      // This avoids spamming the console with a ton of messages
      if (this.activeTimeline.timebase)
        this.activeTimeline._OnUpdateTimer(line.timestamp);
```

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: Adjusted for next commit. This used to do more in the inner async function.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: I assume you're referencing `this.run` which shouldn't be async, no. Fixing for next commit.

ui/raidboss/emulator/data/PopupTextAnalysis.js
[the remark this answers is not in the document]
valarnin: For reference, the error that results if trying to use filter directly:
image

And can't pass an async function to filter because those return a Promise which means nothing gets filtered out.

ui/raidboss/emulator/data/PopupTextAnalysis.js
quisquous: I meannnnn if you're going to ask, I'd bikeshed and say `delayTo` => `delayUntil` for grammar reasons, but the other two are good.  Mostly I just think `this.delay()` operates differently than the other members and so it made it more readable to name it differently.

ui/raidboss/emulator/data/AnalyzedEncounter.js
```diff
@@ -107,7 +107,9 @@ export default class AnalyzedEncounter extends EventBus {
         popupText.runResolver = res;
       });
 
-      popupText.OnTrigger(trigger, matches);
+      const currentLine = this.encounter.logLines[currentLogIndex];
```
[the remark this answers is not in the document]
quisquous: Ok, just checking my understanding, thanks.  It's a big patch so trying to cut down on what I needed to understand.

ui/raidboss/emulator/data/PopupTextAnalysis.js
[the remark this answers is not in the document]
quisquous: https://advancedweb.hu/how-to-use-async-functions-with-array-filter-in-javascript/ has a one-liner suggestion using `Promise.all` to map everything to a promise and then await them all.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: Yeah, that makes sense. Changes made for next commit.

ui/raidboss/emulator/data/PopupTextAnalysis.js
quisquous: This doesn't really answer my question here about storing a handle to `triggerHelper.resolver.status` in `this.currentTriggerStatus`.  Where is `this.currentTriggerStatus` being used in a place that you don't have the `triggerHelper` or the `resolver`?

ui/raidboss/emulator/data/PopupTextAnalysis.js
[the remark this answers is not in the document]
valarnin: Swapped to this setup, which works:
```js
  async checkResolved(logObj) {
    await Promise.all(
        this.triggerResolvers.map(async (resolver) => await resolver.isResolved(logObj)))
      .then((results) => {
        this.triggerResolvers = this.triggerResolvers.filter((_, index) => !results[index]);
      });
  }
```

Any additional adjustments to this, or is it good for the next commit?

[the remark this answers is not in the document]
quisquous: Looks good to me, thanks! It just seems better from a TypeScript/functional programming pov to not reuse the triggerResolvers entries.

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: These functions in `PopupTextAnalysis`:

- _onTriggerInternalCheckSuppressed
- _makeTextElement
- _createTextFor
- _playAudioFile

quisquous: What about: `_makeTextElement`, `_createTextFor`, and `_playAudioFile` get a `triggerHelper` parameter on their functions.  It looks pretty straightforward to pass that through.

`_onTriggerInternalCheckSuppressed` is weirder, but I'm not sure why status needs to store suppressed, as it doesn't seem to get used anywhere? Could it just not store that at all and suppressed triggers can just be dropped entirely?

Maybe if the right information can get passed in here, then the currentTriggerStatus setting can go away and it can all be done through triggerHelper.

valarnin: > What about: `_makeTextElement`, `_createTextFor`, and `_playAudioFile` get a `triggerHelper` parameter on their functions. It looks pretty straightforward to pass that through.
This would resolve the issue in general, yes. Should I include that as part of this PR, given that it's going to require changes in both raidboss and raidemulator even if it's just that change?

> `_onTriggerInternalCheckSuppressed` is weirder, but I'm not sure why status needs to store suppressed, as it doesn't seem to get used anywhere? Could it just not store that at all and suppressed triggers can just be dropped entirely?
I decided to store the suppressed state in case we want to indicate it later. As it is right now, there's no visual indicator of suppressed triggers, however it's still visible that a trigger is suppressed in the JSON output due to being stored here.

> Maybe if the right information can get passed in here, then the currentTriggerStatus setting can go away and it can all be done through triggerHelper.
These changes would allow getting rid of the `this.currentTriggerStatus` property, however the circular reference of `triggerHelper.resolver.triggerHelper` would still be needed because the callback needs `triggerHelper` which isn't available otherwise, and `triggerHelper.resolver` is needed elsewhere as well.

quisquous: > > What about: `_makeTextElement`, `_createTextFor`, and `_playAudioFile` get a `triggerHelper` parameter on their functions. It looks pretty straightforward to pass that through.
> 
> This would resolve the issue in general, yes. Should I include that as part of this PR, given that it's going to require changes in both raidboss and raidemulator even if it's just that change?

Yeah, I think it would be fine to include it here.  You could also submit it as a separate PR to land first if you wanted to separate it out.  Those changes seem like they should be pretty minor plumbing.

> > `_onTriggerInternalCheckSuppressed` is weirder, but I'm not sure why status needs to store suppressed, as it doesn't seem to get used anywhere? Could it just not store that at all and suppressed triggers can just be dropped entirely?
> 
> I decided to store the suppressed state in case we want to indicate it later. As it is right now, there's no visual indicator of suppressed triggers, however it's still visible that a trigger is suppressed in the JSON output due to being stored here.

Personally I think it would be better to remove the noise of suppressed triggers.  I have really never had a situation in which it was useful to look at them.  Triggers also get suppressed prior to the point of constructing a triggerHelper too.

> > Maybe if the right information can get passed in here, then the currentTriggerStatus setting can go away and it can all be done through triggerHelper.
> 
> These changes would allow getting rid of the `this.currentTriggerStatus` property, however the circular reference of `triggerHelper.resolver.triggerHelper` would still be needed because the callback needs `triggerHelper` which isn't available otherwise, and `triggerHelper.resolver` is needed elsewhere as well.

I think it might be possible to remove the resolver -> triggerHelper lookup (maybe by setting final somewhere else?) but I think it's not a big deal and this could land with that circular reference.  Those two objects are tightly coupled together.

It's really `this.currentTriggerStatus` as global state being set everywhere that really feels like the blocking issue to me.

quisquous:
I was trying to test #2866 with and without this patch, and there were definitely some issues with the Hello World debuffs (`O12S Defamation`, `O12S Latent Defect`, `O12S Rot`, `O12S Hello World Stack`, `O12S Hello World No Marker`).  "Run correctly" here means showed up in the analysis window.

* Without #2866 or this patch, everything appeared to run correctly.
* With just #2866, everyone gets the "No Marker" callout but the short stack target, but everything else working ok.  (This makes some sense, as #2866 add delay/suppress, so not surprised it's broken.)
* With just this patch, the two "No Marker" calls are correct and the "short stack (on somebody else)" is correct, but no normal Hello World debuff triggers show up.  If I unclick "show skipped" they seem to be there.
* With both #2866 and this patch, everybody but the short stack target gets no marker still and the debuffs don't show up.

Moreover, if I add console.log messages, it looks like only the first debuff is collected even though there's a delay, which is probably why it thinks 7 people have no debuff.

[AlphascapeV40Savage_20210513_045020_7m59s_wipe.log](https://github.com/quisquous/cactbot/files/6488452/AlphascapeV40Savage_20210513_045020_7m59s_wipe.log)

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: Do you have a preference on if triggerHelper should go first or last in the arguments?

quisquous: I don't feel strongly.  It looks like there's only one function (`_addTextFor`) that takes both a `triggerHelper` and something else so far.  I can see both "always go first" or "always goes last" making sense.

valarnin:
> I was trying to test #2866 with and without this patch, and there were definitely some issues with the Hello World debuffs (`O12S Defamation`, `O12S Latent Defect`, `O12S Rot`, `O12S Hello World Stack`, `O12S Hello World No Marker`). "Run correctly" here means showed up in the analysis window.

I think this was being caused by the (`run` != async) issue. Can you re-test with the latest pushed version and confirm that you're still seeing the issues with both the old and new versions of the triggers?

ui/raidboss/emulator/data/PopupTextAnalysis.js
valarnin: I opted to prepend the parameter since it's the more common parameter in the class scope. Changes are pushed, please let me know if there are any other adjustments for this.

quisquous:
> > I was trying to test #2866 with and without this patch, and there were definitely some issues with the Hello World debuffs (`O12S Defamation`, `O12S Latent Defect`, `O12S Rot`, `O12S Hello World Stack`, `O12S Hello World No Marker`). "Run correctly" here means showed up in the analysis window.
> 
> I think this was being caused by the (`run` != async) issue. Can you re-test with the latest pushed version and confirm that you're still seeing the issues with both the old and new versions of the triggers?

It looks like my #2866 was busted.  I forgot that "suppressSeconds" starts suppressing prior to the delay happening, even though its field comes afterwards.  I needed to separate into two triggers.  (Although arguably I think maybe it's more useful to have suppress happen after delay so that single trigger collect in preRun, delay, suppress can work? Some other PR...)

This looks fine though from my local testing.  I'll take another look through review.

quisquous:
Thanks for all the changes.  Having the triggerHelper passed everywhere made this a lot more understandable for me.

PR #2864, opened 2021-05-15
Korean v5.4 update

Jaehyuk-Lee:
Will open pull request when v5.4 update is released. (May 18th, KST)
CE opcodes won't be fixed nearly, other contributor might fix this though.


Jaehyuk-Lee:
`python gen_zone_id_and_info.py` returns this.
```
output csv: C:\SaintCoinach\2021.04.07.0000.0000\rawexd\ContentFinderCondition.csv
csv lines: 786
  File "gen_zone_id_and_info.py", line 347, in <module>
    territory_info = generate_zone_info(
  File "gen_zone_id_and_info.py", line 277, in generate_zone_info
    output["contentType"] = int(cfc_map_by_lang["en"][cfc_id]["content_type_id"])
ValueError: invalid literal for int() with base 10: ''
```
I've checked the generated csv file is identical to https://github.com/xivapi/ffxiv-datamining 's one.

Jaehyuk-Lee:
@quisquous Just accept this PR please. I think fixing that python error would take time. Of course if there's another problem, please feel free to tell me.
Korean users got 0.22.5 release after releasing my own 0.22.4.kr1 release, so my updates for Korean users got overwritten by the latest official release. (My release checks update from the quisquous version.) I think merging this, in spite of the resource problem, is better.

quisquous:
I am really sorry about not merging this before doing another release.  I did not mean to cause you a hassle.

I am not having any errors when running `gen_zone_id_and_info.py`.  If you don't mind debugging to figure out what's causing the error, I would appreciate it.  I will run this for you and submit it.

Jaehyuk-Lee:
That was only for me.. huh...
maybe me having both python 3.x and 2.7 installed is the problem.

PR #2869, opened 2021-05-17
raidboss: Remove Redundant Response Severities

panicstevenson:


quisquous:
I think this isn't unreasonable to do, but at the same time I am also unsure.  Here are various thoughts:

1) it's nice to have default severities because that creates consistency 
2) it's also nice to have explicit severities because that's more readable in each trigger (alternatively: this isn't important information and is just noise)
3) it could be nice to have default severities because we could change them if needed (alternatively: this seems unlikely?)
4) changing default severities is impossible because each set of triggers usually has some subjective balance of alarm/alert/info and so this changing this arbitrarily would probably make some content "feel bad"
5) if this change is desirable, should we enforce that you don't set the default? (alternatively: if we don't want this, should we land this PR at all?)

JLGarber:
For me, I think I come down (moderately weakly) on the side of "if defaults exist, use them". As far as "explicit severity information in-line is nice", I think this really isn't that important. To me, as a "dev user", a large part of using canned responses is that they *are* set/forget, and are configurable *if* that's needed. When viewing triggers in a text editor, "what level of severity is this" doesn't really come up except when multiple triggers interact on the same or interrelated events. In these cases, if there is a reason we need to override the defaults, I sort of feel like comments in-line explaining why would be more useful than enforced explicit severity.

That being said, "Explicit is better than implicit" counts for a lot, and none of my opinionation accounts for pull request reviews, automated testing, or the continuing TypeScript conversions. If those end up being more important from a repository maintenance perspective, maybe that should govern.

panicstevenson:
I don't think I have a strong opinion either way, but I would like it to be consistent (not urgently, but this was confusing to me when I was debugging another issue, and that's how I got here).

I think it basically comes down to the preference between:
```ts
{
  ...
  // This needs to be alarm because your character gets deleted if you mess up.
  response: Responses.spread('alarm'),
},
{
  ...
  response: Responses.spread(),
},
```
Or
```ts
{
  ...
  response: Responses.spread('alarm'),
},
{
  ...
  response: Responses.spread('info'),
},
```
I feel like I like the latter for a couple reasons, one being that it's easier to enforce (we'd leave defaults, but people may already be using overrides that don't use the defaults), and another being that the intention of the trigger is fairly obvious. I dislike it since I feel like it "self-documents" in a poor way. That being said, nothing forces anyone to add comments about the why for any severity override; and, even if we did, they'd just get copy/pasted with everything else.

That being said, I think explicitly calling each of them out is annoying for some things; I think calling all `aoe('info')` and all `tankBusterSwap('alarm', 'alert')` would get very old very quick and might even cause some fatigue when it comes to pattern-matching triggers from older fights to bring into newer fights. I think explicitly mentioning the severity is easy to parse to know exactly the SEVERITY of a trigger, but I don't know that we (as developers) really care about that too often? I think that the argument here is that there's a cognitive overload for remembering that getUnder is info and that getIn is alert; however, I think looking it at a trigger file-level view it's very helpful to know if this is a run-of-the-mill callout, like a normal tankbuster, or if it's special.

The counterpoint to that I can see would be that a VERY special tankbuster that camouflages in the explicit-severity world probably doesn't look like a standard 3 line tankbuster trigger at all; and, if it does, it _should_ have some form of a comment, just by nature of preventing any future gotchas that might blanketly update all triggers of similar types, etc. I think having the defaults and non-defaults mixed and match lets you know the "kind of special, but not really THAT special" middle-ground that I think is actually fairly common.

quisquous:
At any rate, sounds like lets merge this and we can think on whether it makes sense to enforce this.  (I'd probably want an eslint fixer if we did, so it was automatic.)  Unfortunately I caused you some merge conflicts, but can merge after that.

panicstevenson:
> At any rate, sounds like lets merge this and we can think on whether it makes sense to enforce this

How do you feel we should follow up and come to a decision on this? I'm comfortable with leaving this PR open until we decide which direction to enforce, or just closing it; I don't think I had envisioned this change having a discussion, but I think it's good. I'm a little hesitant to merge this and see nothing come of it if we want to see something come out of the enforcing aspect.

quisquous:
I think I'm less hesitant to merge this if you feel like this led to some confusion for you and this is more readable.  I think landing this and (doing nothing | enforcing this | removing defaults be explicit everywhere) are all ok options here with different tradeoffs.

I am ok deferring a decision until we find a stronger reason to pick one of the followup options, or if somebody feels motivated to write an eslint fixer so these are automatically removed, I think that'd be ok too.

JLGarber:
Thinking it over more, we probably should wait to merge this until after such a lint rule is put in place. As much as I feel like this PR is clean and fine, it's work that will potentially have to be done again if we just do the cleanup without the rule. (Granted, I say this while explicitly not being in a position to contribute to *writing* the rule.)

Maybe one thing we could add with this PR is some guidance in the style guide noting that "we should use implicit rather than explicit default severity, unless doing so would collide with another trigger"? Or I can write that up separately this weekend.

quisquous:
I guess my thought here is that I think if we have such a rule, it should be an eslint rule that supports --fix.  That would prevent there being any extra work to do (we'd run it once to clean everything up) and also prevent it from being a burden on programmers.  I think in that view, that's why I was more supportive of landing this now.

panicstevenson:
> Mmmmmmm what prevents this from getting out of sync with `responses.ts`? Is it possible to share this defaultSeverityResponseMap between the two, and have responses use that as the canonical source of default severity?

I don't think that we can source from the Responses unless we had Responses publish something? (Granted, all these things live in the same area currently, we could just reach across and look at it for now and if we ever split anything, deal with that later). My thoughts here were a little less strict and more in line of thinking of having the opposite map, which could would map the function name to the response severity, and if it didn't have an entry it would throw.

Granted, the way you've listed has a lot less overall friction for minor things, and my only complaint there is a design issue that will most likely never exist

quisquous:
Yeah, I guess I was imagining some separate file that Responses imports or some data structure inside of Responses that it exports. Then, rather than just saying `defaultInfoText` we'd say `defaultText('earthshaker')` or something.  A followup also sounds fine.

PR #2871, opened 2021-05-19
Jobs: Fix BRD, MNK(cn/ko5.4)

Echoring:
- Fix bug about BRD repertoireTimer goes empty
- Delete MNK lightningTimer for cn/ko5.4
- <del>Job-specific UI will be reset upon party wipe now. </del>

### DIFF
--- a/ui/jobs/components/mnk.js
+++ b/ui/jobs/components/mnk.js
@@ -5,9 +5,9 @@ import { computeBackgroundColorFrom } from '../utils';
 const lightningFgColors = [];
 
 export function setup(bars) {
-  // TODO: Remove bars timer when cn/ko update 5.4
+  // TODO: Remove bars timer when ko update 5.4
   let lightningTimer = null;
-  if (['cn', 'ko'].includes(bars.options.ParserLanguage)) {
+  if (['ko'].includes(bars.options.ParserLanguage)) {
     lightningTimer = bars.addTimerBar({
       id: 'mnk-timers-lightning',
       fgColor: 'mnk-color-lightning-0',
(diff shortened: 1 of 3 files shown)


ui/jobs/components/mnk.js
```diff
@@ -48,34 +36,9 @@ export function setup(bars) {
...
-          lightningTimer.value = lightningSeconds;
-        }
-      }
-    } else {
-      // For now, we just assign bars.speedBuffs.lightningStacks
-      // as corresponding stacks via current level
-      bars.speedBuffs.lightningStacks = getLightningStacksViaLevel(bars.level);
-    }
+    // For now, we just assign bars.speedBuffs.lightningStacks
+    // as corresponding stacks via current level
```
panicstevenson, line 40:
```suggestion
    // After the 5.4 changes, we just assign bars.speedBuffs.lightningStacks
    // as corresponding stacks via current level
```
I know it's not your comment, but I think it might age a little better this way.

panicstevenson (approved):

ui/jobs/jobs.js
```diff
@@ -681,9 +681,12 @@ class Bars {
   }
 
   _onPartyWipe(e) {
-    // TODO: add reset for job-specific ui
     if (this.buffTracker)
       this.buffTracker.clear();
+    // Reset job-specific ui
```
quisquous, line 686: This is too many things in one PR.  Can you remove this bit and submit another PR? The other changes are fine, but I think resetting timers is not the same as resetting all job ui.  It doesn't clear resources or any other things.  I think it'd be better if each job had a `reset()` function to clean up anything it added.  However, I think that's too many changes for this PR and so it'd be better to do separately.

Echoring: Reverted. I'll try to write a `reset()` function later.

PR #2872, opened 2021-05-20
raidboss: Refactor Timeline Types

panicstevenson:
Refactor Timeline types for brevity.


panicstevenson:
I'm not tied to any of the names, feel free to suggest something if there's strong opinions to be had!

PR #2874, opened 2021-05-20
raidboss: KR fix

Bing-su:
- e11s
only that line was translated 'lightning(번개)' into 'fire(화염)'

- bozjan
typo


Jaehyuk-Lee:
That was my fault :rofl:

PR #2875, opened 2021-05-21
raidemulator: Convert EmulatorCommon, LineEvent, LogRepository to typescript

valarnin:
This is part one of many PRs to convert raidemulator to typescript.

This PR also includes a slight addition to the webpack config to show errors for child compilations, which is needed to catch errors related to the web worker used in emulator.


quisquous:
```diff
diff --git a/ui/raidboss/emulator/EmulatorCommon.js b/ui/raidboss/emulator/EmulatorCommon.ts
index f108cee73..66c9e3dea 100644
--- a/ui/raidboss/emulator/EmulatorCommon.js
+++ b/ui/raidboss/emulator/EmulatorCommon.ts
@@ -1,9 +1,14 @@
 import { LocaleNetRegex } from '../../../resources/translations';
 import NetRegexes from '../../../resources/netregexes';
 
+// Disable ban-types for cloneData as it needs to work on raw objects for performance reasons.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DataType = { [key: string]: any } | null;
+
 export default class EmulatorCommon {
-  static cloneData(data, exclude = ['options', 'party']) {
-    let ret;
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  static cloneData(data: DataType, exclude = ['options', 'party']): DataType {
+    let ret: DataType;
     if (Array.isArray(data))
       ret = [];
     else
@@ -18,12 +23,14 @@ export default class EmulatorCommon {
       if (typeof data[i] === 'object')
         ret[i] = EmulatorCommon._cloneData(data[i]);
       else
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
         ret[i] = data[i];
     }
     return ret;
   }
 
-  static _cloneData(data) {
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  static _cloneData(data: DataType): DataType {
     if (typeof data === 'object') {
       if (Array.isArray(data)) {
         const ret = [];
@@ -39,7 +46,7 @@ export default class EmulatorCommon {
       if (data instanceof RegExp)
         return new RegExp(data);
 
-      const ret = {};
+      const ret: DataType = {};
       for (const i in data)
         ret[i] = EmulatorCommon._cloneData(data[i]);
 
@@ -48,72 +55,73 @@ export default class EmulatorCommon {
     return data;
   }
 
-  static timeToString(time, includeMillis = true) {
+  static timeToString(time: number, includeMillis = true): string {
     const negative = time < 0 ? '-' : '';
     time = Math.abs(time);
+    const millisNum = time % 1000;
+    const secsNum = ((time % (60 * 1000)) - millisNum) / 1000;
     // Milliseconds
-    const millis = ('00' + (time % 1000)).substr(-3);
-    const secs = ('0' + ((time % (60 * 1000)) - millis) / 1000).substr(-2);
-    const mins = ('0' + ((((time % (60 * 60 * 1000)) - millis) / 1000) - secs) / 60).substr(-2);
+    const millis = `00${millisNum}`.substr(-3);
+    const secs = `0${secsNum}`.substr(-2);
+    const mins = `0${((((time % (60 * 60 * 1000)) - millisNum) / 1000) - secsNum) / 60}`.substr(-2);
     return negative + mins + ':' + secs + (includeMillis ? '.' + millis : '');
   }
 
-  static timeToDateString(time) {
-    const date = new Date(time);
-    return date.getFullYear() + '-' + EmulatorCommon.zeroPad(date.getMonth() + 1) + '-' + EmulatorCommon.zeroPad(date.getDate());
+  static timeToDateString(time: number): string {
+    return this.dateObjectToDateString(new Date(time));
   }
 
-  static timeToTimeString(time, includeMillis = false) {
-    const date = new Date(time);
-    let ret = EmulatorCommon.zeroPad(date.getHours()) + ':' + EmulatorCommon.zeroPad(date.getMinutes()) + ':' + EmulatorCommon.zeroPad(date.getSeconds());
+  static dateObjectToDateString(date: Date): string {
+    const year = date.getFullYear();
+    const month = EmulatorCommon.zeroPad((date.getMonth() + 1).toString());
+    const day = EmulatorCommon.zeroPad(date.getDate().toString());
+    return `${year}-${month}-${day}`;
+  }
+
+  static timeToTimeString(time: number, includeMillis = false): string {
+    return this.dateObjectToTimeString(new Date(time), includeMillis);
+  }
+
+  static dateObjectToTimeString(date: Date, includeMillis = false): string {
+    const hour = EmulatorCommon.zeroPad(date.getHours().toString());
+    const minute = EmulatorCommon.zeroPad(date.getMinutes().toString());
+    const second = EmulatorCommon.zeroPad(date.getSeconds().toString());
+    let ret = `${hour}:${minute}:${second}`;
     if (includeMillis)
-      ret = ret + '.' + EmulatorCommon.zeroPad(date.getMilliseconds(), 3);
+      ret = ret + `.${date.getMilliseconds()}`;
 
     return ret;
   }
 
-  static msToDuration(ms) {
+  static msToDuration(ms: number): string {
     const tmp = EmulatorCommon.timeToString(ms, false);
     return tmp.replace(':', 'm') + 's';
   }
 
-  static dateTimeToString(time, includeMillis = false) {
+  static dateTimeToString(time: number, includeMillis = false): string {
     const date = new Date(time);
-    let ret = date.getFullYear() + '-' + EmulatorCommon.zeroPad(date.getMonth() + 1) + '-' + EmulatorCommon.zeroPad(date.getDate());
-    ret = ret + ' ' + EmulatorCommon.zeroPad(date.getHours()) + ':' + EmulatorCommon.zeroPad(date.getMinutes()) + ':' + EmulatorCommon.zeroPad(date.getSeconds());
-    if (includeMillis)
-      ret = ret + '.' + date.getMilliseconds();
-
-    return ret;
+    return `${this.dateObjectToDateString(date)} ${this.dateObjectToTimeString(date, includeMillis)}`;
   }
 
-  static zeroPad(str, len = 2) {
+  static zeroPad(str: string, len = 2): string {
     return ('' + str).padStart(len, '0');
   }
 
-  static properCase(str) {
+  static properCase(str: string): string {
     return str.replace(/([^\W_]+[^\s-]*) */g, (txt) => {
       return txt.charAt(0).toUpperCase() + txt.substr(1).toLowerCase();
     });
   }
 
-  static spacePadLeft(str, len) {
+  static spacePadLeft(str: string, len: number): string {
     return str.padStart(len, ' ');
   }
-}
-
-EmulatorCommon.sealRegexes = {};
-EmulatorCommon.engageRegexes = {};
-EmulatorCommon.countdownRegexes = {};
-EmulatorCommon.unsealRegexes = {};
 
-for (const lang in LocaleNetRegex.areaSeal) {
-  EmulatorCommon.sealRegexes[lang] = LocaleNetRegex.areaSeal[lang];
-  EmulatorCommon.engageRegexes[lang] = LocaleNetRegex.countdownEngage[lang];
-  EmulatorCommon.countdownRegexes[lang] = LocaleNetRegex.countdownStart[lang];
-  EmulatorCommon.unsealRegexes[lang] = LocaleNetRegex.areaUnseal[lang];
+  static sealRegexes: Record<string, RegExp> = LocaleNetRegex.areaSeal;
+  static engageRegexes: Record<string, RegExp> = LocaleNetRegex.countdownEngage;
+  static countdownRegexes: Record<string, RegExp> = LocaleNetRegex.countdownStart;
+  static unsealRegexes: Record<string, RegExp> = LocaleNetRegex.areaUnseal;
+  static wipeRegex = NetRegexes.network6d({ command: '40000010' });
+  static winRegex = NetRegexes.network6d({ command: '40000003' });
+  static cactbotWipeRegex = NetRegexes.echo({ line: 'cactbot wipe.*?' });
 }
-
-EmulatorCommon.wipeRegex = NetRegexes.network6d({ command: '40000010' });
-EmulatorCommon.winRegex = NetRegexes.network6d({ command: '40000003' });
-EmulatorCommon.cactbotWipeRegex = NetRegexes.echo({ line: 'cactbot wipe.*?' });
```

valarnin:
For future reference, do you want diffs of only the files that github doesn't pick up as a rename? Or of all files changed in the PR?

quisquous:
> For future reference, do you want diffs of only the files that github doesn't pick up as a rename? Or of all files changed in the PR?

I think in general I'd say "only the file that github doesn't pick up as a rename".  If you have multiple files with so many changes that they don't show up as renames, that's probably too many changes for one PR.  If you're talking about the network logs specifically, I'm not sure what would make it easier to review if it's more than 50% remove/delete instead of rename.  I may just download and diff locally.

valarnin:
> If you have multiple files with so many changes that they don't show up as renames, that's probably too many changes for one PR.

That should be true for almost everything, except that I think there are multiple LineEvent0x** files that didn't get picked up as renames in the previous PR.

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -0,0 +1,127 @@
...
+      if (typeof data[i] === 'object')
+        ret[i] = EmulatorCommon._cloneData(data[i]);
+      else
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+        ret[i] = data[i];
+    }
+    return ret;
+  }
+
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
quisquous, line 32:
```suggestion
```

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -0,0 +1,127 @@
+import { LocaleNetRegex } from '../../../resources/translations';
+import NetRegexes from '../../../resources/netregexes';
+
+// Disable ban-types for cloneData as it needs to work on raw objects for performance reasons.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DataType = { [key: string]: any } | null;
+
+export default class EmulatorCommon {
+  // eslint-disable-next-line @typescript-eslint/ban-types
```
quisquous, line 9:
```suggestion
```

I think encapsulating the `any` inside of `DataType` prevents errors here already without an explicit disable.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -1,45 +1,50 @@
...
-    this.decEvent = parseInt(this.parts[0]);
+  public offset = 0;
+  decEvent: number;
+  hexEvent: string;
+  timestamp: number;
+  networkLine: string;
+  convertedLine = '';
+  invalid = false;
+  index = 0;
+  constructor(repo: LogRepository, line: string, public parts: string[]) {
```
quisquous, line 16:
```suggestion

  constructor(repo: LogRepository, line: string, public parts: string[]) {
```

I can't find an eslint rule for this, but I'd appreciate a linebreak between members and functions.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -1,45 +1,50 @@
...
-
-    this.decEvent = parseInt(this.parts[0]);
+  public offset = 0;
+  decEvent: number;
+  hexEvent: string;
+  timestamp: number;
+  networkLine: string;
+  convertedLine = '';
+  invalid = false;
+  index = 0;
```
quisquous, line 15: Are these all public? Some private?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -1,45 +1,50 @@
...
 
-  prefix() {
+  prefix(): string {
     return '[' + EmulatorCommon.timeToTimeString(this.timestamp, true) + '] ' + this.hexEvent + ':';
   }
 
-  static isDamageHallowed(damage) {
-    return parseInt(damage, 16) & parseInt('1000', 16);
+  static isDamageHallowed(damage: string): boolean {
+    return ((parseInt(damage, 16) & parseInt('1000', 16)) > 0);
```
quisquous, line 35:
```suggestion
    return (parseInt(damage, 16) & parseInt('1000', 16)) > 0;
```

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -1,45 +1,50 @@
...
   }
 
-  static isDamageBig(damage) {
-    return parseInt(damage, 16) & parseInt('4000', 16);
+  static isDamageBig(damage: string): boolean {
+    return ((parseInt(damage, 16) & parseInt('4000', 16)) > 0);
```
quisquous, line 39:
```suggestion
    return (parseInt(damage, 16) & parseInt('4000', 16)) > 0;
```

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -0,0 +1,127 @@
+import { LocaleNetRegex } from '../../../resources/translations';
+import NetRegexes from '../../../resources/netregexes';
+
+// Disable ban-types for cloneData as it needs to work on raw objects for performance reasons.
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type DataType = { [key: string]: any } | null;
```
quisquous, line 6: I can't come up with any better way to do this, and I think it's probably cleaner to use `any` here than to use `Data` from `data.d.ts` with a `[key: string]: any` field on it.  So, this all seems good to me, thanks!

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -0,0 +1,127 @@
...
+    // Use extra logic for top-level extend for property exclusion
+    // This cut the execution time of this code from 41,000ms to 50ms when parsing a 12 minute pull
+    for (const i in data) {
+      if (exclude.includes(i))
+        continue;
+
+      if (typeof data[i] === 'object')
+        ret[i] = EmulatorCommon._cloneData(data[i]);
+      else
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
quisquous, line 26:
```suggestion
        // Assignment of any to any.  See DataType definition above for reasoning.
        // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```

I think any eslint-disable should at least get some explanation as to why, even if it's just pointing to something else.

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -0,0 +1,127 @@
...
+type DataType = { [key: string]: any } | null;
+
+export default class EmulatorCommon {
+  // eslint-disable-next-line @typescript-eslint/ban-types
+  static cloneData(data: DataType, exclude = ['options', 'party']): DataType {
+    let ret: DataType;
+    if (Array.isArray(data))
+      ret = [];
+    else
+      ret = {};
```
quisquous, line 15: This is a drive-by review of old code, but is this actually possible? I don't think `DataType` or `Data` can be an array.  Feel free to leave a TODO for later if you don't want to minimize behavior changes in this patch.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -1,45 +1,50 @@
...
 
-    this.timestamp = +new Date(this.parts[1]);
+    this.timestamp = +new Date(this.parts[1] || '');
 
     this.networkLine = line;
     repo.updateTimestamp(this.timestamp);
   }
 
-  convert() {
+  convert(_: LogRepository | undefined = undefined): void {
```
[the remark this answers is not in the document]
valarnin: It's optional. Only two subclasses use it (only the log lines that need to look up a combatant name).
0x1F - Job Gauge
0x18 - HoT/DoT tick

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -1,45 +1,50 @@
...
   }
 
-  static calculateDamage(damage) {
+  static calculateDamage(damage: string): number {
     if (LineEvent.isDamageHallowed(damage))
       return 0;
 
     damage = EmulatorCommon.zeroPad(damage, 8);
-    const parts = [
+    const parts: [string, string, string, string] = [
```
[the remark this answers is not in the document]
valarnin: Adjusted for next commit.

ui/raidboss/emulator/data/network_log_converter/LogRepository.ts
```diff
@@ -1,19 +1,19 @@
+export type Combatant = {
+  name: string | undefined;
+  job: string | undefined;
```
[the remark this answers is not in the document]
valarnin: Adjusted for next commit.

ui/raidboss/emulator/data/network_log_converter/LogRepository.ts
```diff
@@ -1,19 +1,19 @@
...
+  spawn: number;
+  despawn: number;
+}
+
 export default class LogRepository {
-  constructor() {
-    this.Combatants = {};
-    this.firstTimestamp = Number.MAX_SAFE_INTEGER;
-  }
+  Combatants: Record<string, Combatant> = {};
```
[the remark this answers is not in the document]
valarnin: Adjusted for next commit.

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  static spacePadLeft(str: string, len: number): string {
+    return str.padStart(len, ' ');
+  }
+
+  static sealRegexes: Record<string, RegExp> = LocaleNetRegex.areaSeal;
+  static engageRegexes: Record<string, RegExp> = LocaleNetRegex.countdownEngage;
+  static countdownRegexes: Record<string, RegExp> = LocaleNetRegex.countdownStart;
+  static unsealRegexes: Record<string, RegExp> = LocaleNetRegex.areaUnseal;
```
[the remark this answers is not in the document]
valarnin: The narrower type causes issues with `LogEventHandler`'s usage of a helper method to check against the records.

```js
  static doesLineMatch(line, regexes) {
    for (const i in regexes) {
      const res = regexes[i].exec(line);
      if (res) {
        if (LocaleNetRegex.areaSeal[i])
          res.groups.language = i;
        return res;
      }
    }
    return false;
  }
```

This is used for both the LocaleNetRegex exports as well as the other static RegExp values defined on EmulatorCommon, compare these two calls:
```js
    res = LogEventHandler.doesLineMatch(line, EmulatorCommon.engageRegexes);
// vs
    res = LogEventHandler.doesLineMatch(line, [EmulatorCommon.winRegex]);
```

ui/raidboss/emulator/EmulatorCommon.ts
valarnin: Yeah, this and the one above were leftover from trying to get this to pass eslint, missed cleaning them up after defining the `DataType` type. Removed for next commit.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
valarnin: Added line break for next commit. As for an eslint rule, maybe this one could be configured? I'm not very well versed in eslint configuration.

https://eslint.org/docs/rules/padding-line-between-statements

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
valarnin: They're all public. I used vscode's auto-generate-members option to add them, I'm not sure why it added a public for offset but not for any of the others. Should they all have the public modifier since they're all intended to be public?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
valarnin: Adjusted these two for next commit.

ui/raidboss/emulator/EmulatorCommon.ts
valarnin: Yeah, I struggled with this one for a while, originally I wanted to just use `object` and disable the warning for that because that's what it's actually working on, but then it complained on every line that tried to access the object's properties with `data[<some_variable>]` so ultimately I went with this.

ui/raidboss/emulator/EmulatorCommon.ts
valarnin: Agreed. The multiple comments above an un-braced `else` strikes me as a bit painful, though. Added comment for next commit.

ui/raidboss/emulator/EmulatorCommon.ts
valarnin: Checking usages for this, I don't think it can be an array anymore. This helper used to be used in a few spots in the UI portion where it could be an array, if I remember correctly, and that's why it's there? Either way, I'm comfortable removing it. Removed for next commit.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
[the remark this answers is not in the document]
quisquous: Yes, but `repo` is unconditionally passed in `convertLines`, right?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
quisquous: I think consistency is the most important thing at the moment.  I might consider adding https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-member-accessibility.md at some point, and if I do, then everything can get fixed up then.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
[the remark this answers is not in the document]
valarnin: Yeah, sorry, I see where you're coming from now. Yes, it's always passed but not always used, so the question then becomes which is the better implementation practice here? Should it be on the base class but be excluded from subclass parameters? Should it be defined as an `_` argument?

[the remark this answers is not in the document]
valarnin: I think regardless of the answer here, the change to this file will stay the same, so I'll adjust this for the next commit and just need clarification on how to handle the `LineEvent0x**` files for the next PR at some point. I'll mark this as resolved but feel free to respond with clarification.

ui/raidboss/emulator/EmulatorCommon.ts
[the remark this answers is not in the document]
valarnin: Probably this can be resolved upstream in `LogEventHandler` by adjusting the `regexes` parameter to be a union type between the two possible values. Either way, I don't think this should hold up this PR, adjusting for next commit.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
valarnin: Adjusted to all `public`. I think in this case it makes sense to be more explicit.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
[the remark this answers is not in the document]
quisquous: So, from a TypeScript perspective, it is an error to say `convert()` in the base class and `convert(repo: LogRepository)` in the derived class, as the second function cannot be assigned to the first.  It is *technically* possible to say `convert(repo?: LogRepository)` in the base class and `convert(repo: LogRepository)` in the derived class, see [this playground](https://www.typescriptlang.org/play?strict=true#code/MYGwhgzhAEBCkFNoG8CwAoa0BmBXAdsABQAeA-AFzT64C2ARggE4CUKGWWwA9vhNyAQA6ENwDmRAAb1EVACTISAX0ksA3B2hKM29BlCQYAEWYBLAG4IAJtAQkALgnxWY8CEjSYcBYgE8qNAzMbJ6c0Dx8AsKiEpJWZpZW8si+Kuqaurr6vBD20CTQALzUCADu0CZMFtZE6egkQniERADMdQD07Q1NxHXZfHn+cIhF+Rrovo0+rXWTPbVq0J3Q9gAWpjAb0BAA1gj2wKu+QA).

However, that's a little bit sketchy.  You can assign a derived object to a variable with the type of the base class and call it with the base class prototype in an unexpected way.

What makes sense to me (both for readability, but also for correctness) is to have every function in the prototype chain have the same signature.  (Or, less strictly, to say that the base class function should list everything and the derived functions could list everything or could only list what they or anything derived from them needed?)  At any rate, that's why I was trying to argue that `repo` was required because the derived classes required it.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
quisquous: Thanks, I'll take a look.

PR #2878, opened 2021-05-21
Jobs: Reset job-specific UI

Echoring:
As discussed in #2871 ,a `reset` function is built to solve this issue.


quisquous:
Thanks for this!

PR #2879, opened 2021-05-21
raidemulator: Convert LineEvent0x** to typescript

valarnin:
This PR converts LineEvent0x** files to typescript. Per request, diffs of the typescript files that git didn't detect a rename for will be in comments below.


valarnin:
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x04.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x04.ts
index c7f9c51d..d3bf16cc 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x04.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x04.ts
@@ -1,19 +1,23 @@
-import { LineEvent0x03 } from './LineEvent0x03';
+import { LineEvent0x03, LineEvent0x03Parts } from './LineEvent0x03';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x04Parts = LineEvent0x03Parts;

 // Removed combatant event
 // Extend the add combatant event to reduce duplicate code since they're
 // the same from a data perspective
 export class LineEvent0x04 extends LineEvent0x03 {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x04Parts) {
     super(repo, line, parts);
   }

-  convert() {
-    this.convertedLine = this.prefix() +
-      this.id.toUpperCase() + ':' +
-      'Removing combatant ' + this.name +
-      '.  Max HP: ' + this.maxHp +
-      '. Pos: (' + this.parts[17] + ',' + this.parts[18] + ',' + this.parts[19] + ').';
+  convert(): void {
+    this.convertedLine = `\
+${this.prefix()}\
+${this.id.toUpperCase()}:\
+Removing combatant ${this.name}\
+.  Max HP: ${this.maxHp}\
+. Pos: (${this.parts[17]},${this.parts[18]},${this.parts[19]}).`;
   }
 }

```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x0C.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x0C.ts
index 9e3741d1..5bc17062 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x0C.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x0C.ts
@@ -1,12 +1,37 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x0CParts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Class
+  string, // 3: Strength
+  string, // 4: Dexterity
+  string, // 5: Vitality
+  string, // 6: Intelligence
+  string, // 7: Mind
+  string, // 8: Piety
+  string, // 9: Attack Power
+  string, // 10: Direct Hit
+  string, // 11: Critical Hit
+  string, // 12: Attack Magic Potency
+  string, // 13: Heal Magic Potency
+  string, // 14: Determination
+  string, // 15: Skill Speed
+  string, // 16: Spell Speed
+  string, // 17: 0?
+  string, // 18: Tenacity
+  string, // 19: Unknown?
+  string, // 20: Checksum
+];

 // Player stats event
 export class LineEvent0x0C extends LineEvent {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x0CParts) {
     super(repo, line, parts);
   }

-  convert() {
+  convert(): void {
     this.convertedLine = this.prefix() +
       'Player Stats: ' + this.parts.slice(2, this.parts.length - 1).join(':').replace(/\|/g, ':');
   }
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x14.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x14.ts
index 0ab1a538..ada5dd38 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x14.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x14.ts
@@ -1,14 +1,39 @@
 import LineEvent from './LineEvent';
 import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x14Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Ability ID
+  string, // 5: Ability Name
+  string, // 6: Target ID
+  string, // 7: Target Name
+  string, // 8: Duration
+  string, // 9: Unknown/Blank?
+  string, // 10: Checksum
+];

 // Ability use event
 export class LineEvent0x14 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public abilityId: string;
+  public abilityName: string;
+  public targetId: string;
+  public targetName: string;
+  public castTime: string;
+  public properCaseConvertedLine = '';
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x14Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
     this.name = parts[3];
     repo.updateCombatant(this.id, {
+      job: undefined,
       name: this.name,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -20,6 +45,7 @@ export class LineEvent0x14 extends LineEvent {
     this.targetId = parts[6].toUpperCase();
     this.targetName = parts[7];
     repo.updateCombatant(this.targetId, {
+      job: undefined,
       name: this.targetName,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -28,17 +54,19 @@ export class LineEvent0x14 extends LineEvent {
     this.castTime = parts[8];
   }

-  convert() {
+  convert(): void {
     const target = this.targetName.length === 0 ? 'Unknown' : this.targetName;

-    this.convertedLine = this.prefix() +
-      this.abilityId + ':' +
-      this.name + ' starts using ' + this.abilityName +
-      ' on ' + target + '.';
-    this.properCaseConvertedLine = this.prefix() +
-      this.abilityId + ':' +
-      EmulatorCommon.properCase(this.name) + ' starts using ' + this.abilityName +
-      ' on ' + EmulatorCommon.properCase(target) + '.';
+    this.convertedLine = `\
+${this.prefix()}${this.abilityId}:\
+${this.name} \
+starts using ${this.abilityName} \
+on ${target}.`;
+    this.properCaseConvertedLine = `\
+${this.prefix()}${this.abilityId}:\
+${EmulatorCommon.properCase(this.name)} \
+starts using ${this.abilityName} \
+on ${EmulatorCommon.properCase(target)}.`;
   }
 }

```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.ts
index 56da990e..751c7810 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.ts
@@ -1,13 +1,94 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x15Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Ability ID
+  string, // 5: Ability Name
+  string, // 6: Target ID
+  string, // 7: Target Name
+  string, // 8: Flags
+  string, // 9: Everything here and lower is complicated
+  string, // 10:
+  string, // 11:
+  string, // 12:
+  string, // 13:
+  string, // 14:
+  string, // 15:
+  string, // 16:
+  string, // 17:
+  string, // 18:
+  string, // 19:
+  string, // 20:
+  string, // 21:
+  string, // 22:
+  string, // 23:
+  string, // 24:
+  string, // 25:
+  string, // 26:
+  string, // 27:
+  string, // 28:
+  string, // 29:
+  string, // 30:
+  string, // 31:
+  string, // 32:
+  string, // 33:
+  string, // 34:
+  string, // 35:
+  string, // 36:
+  string, // 37:
+  string, // 38:
+  string, // 39:
+  string, // 40:
+  string, // 41:
+  string, // 42:
+  string, // 43:
+  string, // 44:
+  string, // 45:
+  string, // 46: Checksum
+];

 // Ability hit single target event
 export class LineEvent0x15 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public abilityId: string;
+  public abilityName: string;
+  public targetId: string;
+  public targetName: string;
+  public flags: string;
+  public damage: number;
+  public targetHp: string;
+  public targetMaxHp: string;
+  public targetMp: string;
+  public targetMaxMp: string;
+  public targetX: string;
+  public targetY: string;
+  public targetZ: string;
+  public targetHeading: string;
+  public sourceHp: string;
+  public sourceMaxHp: string;
+  public sourceMp: string;
+  public sourceMaxMp: string;
+  public x: string;
+  public sourceX: string;
+  public y: string;
+  public sourceY: string;
+  public z: string;
+  public sourceZ: string;
+  public heading: string;
+  public sourceHeading: string;
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x15Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
     this.name = parts[3];
     repo.updateCombatant(this.id, {
+      job: undefined,
       name: this.name,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -19,6 +100,7 @@ export class LineEvent0x15 extends LineEvent {
     this.targetId = parts[6].toUpperCase();
     this.targetName = parts[7];
     repo.updateCombatant(this.targetId, {
+      job: undefined,
       name: this.targetName,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -31,30 +113,30 @@ export class LineEvent0x15 extends LineEvent {
     if (this.flags === '3F')
       offset = 2;

-    this.damage = LineEvent.calculateDamage(parts[9 + offset]);
+    this.damage = LineEvent.calculateDamage(parts[9 + offset] as string);

-    this.targetHp = parts[24 + offset];
-    this.targetMaxHp = parts[25 + offset];
+    this.targetHp = parts[24 + offset] as string;
+    this.targetMaxHp = parts[25 + offset] as string;

-    this.targetMp = parts[26 + offset];
-    this.targetMaxMp = parts[27 + offset];
+    this.targetMp = parts[26 + offset] as string;
+    this.targetMaxMp = parts[27 + offset] as string;

-    this.targetX = parts[30 + offset];
-    this.targetY = parts[31 + offset];
-    this.targetZ = parts[32 + offset];
-    this.targetHeading = parts[33 + offset];
+    this.targetX = parts[30 + offset] as string;
+    this.targetY = parts[31 + offset] as string;
+    this.targetZ = parts[32 + offset] as string;
+    this.targetHeading = parts[33 + offset] as string;

-    this.sourceHp = parts[34 + offset];
-    this.sourceMaxHp = parts[35 + offset];
+    this.sourceHp = parts[34 + offset] as string;
+    this.sourceMaxHp = parts[35 + offset] as string;

-    this.sourceMp = parts[36 + offset];
-    this.sourceMaxMp = parts[37 + offset];
+    this.sourceMp = parts[36 + offset] as string;
+    this.sourceMaxMp = parts[37 + offset] as string;

     // Also map these to x/y/z/heading for combatant state updates
-    this.x = this.sourceX = parts[40 + offset];
-    this.y = this.sourceY = parts[41 + offset];
-    this.z = this.sourceZ = parts[42 + offset];
-    this.heading = this.sourceHeading = parts[43 + offset];
+    this.x = this.sourceX = parts[40 + offset] as string;
+    this.y = this.sourceY = parts[41 + offset] as string;
+    this.z = this.sourceZ = parts[42 + offset] as string;
+    this.heading = this.sourceHeading = parts[43 + offset] as string;
   }
 }

```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x16.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x16.ts
index 1b27a28c..869fdfc4 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x16.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x16.ts
@@ -1,10 +1,12 @@
-import LineEvent from './LineEvent';
-import { LineEvent0x15 } from './LineEvent0x15';
+import { LineEvent0x15, LineEvent0x15Parts } from './LineEvent0x15';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x16Parts = LineEvent0x15Parts;

 // Ability hit multiple/no target event
 // Duplicate of 0x15 as far as data
 export class LineEvent0x16 extends LineEvent0x15 {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x16Parts) {
     super(repo, line, parts);
   }
 }
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x17.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x17.ts
index 8ec50201..b49c22bb 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x17.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x17.ts
@@ -1,8 +1,26 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x17Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Ability ID
+  string, // 5: Ability Name
+  string, // 6: Reason
+  string, // 7: Unknown/Blank?
+  string, // 8: Checksum
+];

 // Cancel ability event
 export class LineEvent0x17 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public abilityId: string;
+  public abilityName: string;
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x17Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x18.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x18.ts
index df798f37..81543601 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x18.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x18.ts
@@ -1,14 +1,55 @@
 import LineEvent from './LineEvent';
 import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x18Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Target ID
+  string, // 3: Target Name
+  string, // 4: Type
+  string, // 5: Effect ID
+  string, // 6: Damage
+  string, // 7: Current HP
+  string, // 8: Max HP
+  string, // 9: Current MP
+  string, // 10: Max MP
+  string, // 11: Current TP
+  string, // 12: Max TP
+  string, // 13: X
+  string, // 14: Y
+  string, // 15: Z
+  string, // 16: Heading
+  string, // 17: Unknown/Blank?
+  string, // 18: Checksum
+];

 // DoT/HoT event
 export class LineEvent0x18 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public type: string;
+  public effectId: string;
+  public damage: number;
+  public hp: number;
+  public maxHp: number;
+  public mp: number;
+  public maxMp: number;
+  public x: number;
+  public y: number;
+  public z: number;
+  public heading: number;
+  public resolvedName: string;
+  public effectName: string | undefined = '';
+  public properCaseConvertedLine = '';
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x18Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
     this.name = parts[3];
     repo.updateCombatant(this.id, {
+      job: undefined,
       name: this.name,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -32,35 +73,40 @@ export class LineEvent0x18 extends LineEvent {
     this.resolvedName = repo.resolveName(this.id, this.name);
   }

-  convert(repo) {
-    this.effectName = LineEvent0x18.showEffectNamesFor[this.effectId.toUpperCase()];
+  convert(): void {
+    if (this.effectId.toUpperCase() in LineEvent0x18.showEffectNamesFor)
+      this.effectName = LineEvent0x18.showEffectNamesFor[this.effectId.toUpperCase()];
     let effectPart = '';
     if (this.effectName)
       effectPart = this.effectName + ' ';

-    this.convertedLine = this.prefix() +
-      effectPart +
-      this.type + ' Tick on ' + this.resolvedName +
-      ' for ' + this.damage + ' damage.';
+    this.convertedLine = `\
+${this.prefix()}\
+${effectPart}\
+${this.type} \
+Tick on ${this.resolvedName} \
+for ${this.damage} damage.`;

-    this.properCaseConvertedLine = this.prefix() +
-      effectPart +
-      this.type + ' Tick on ' + EmulatorCommon.properCase(this.resolvedName) +
-      ' for ' + this.damage + ' damage.';
+    this.properCaseConvertedLine = `\
+${this.prefix()}\
+${effectPart}\
+${this.type} \
+Tick on ${EmulatorCommon.properCase(this.resolvedName)} \
+for ${this.damage} damage.`;
   }
+
+  static showEffectNamesFor: Record<string, string> = {
+    '4C4': 'Excognition',
+    '35D': 'Wildfire',
+    '1F5': 'Doton',
+    '2ED': 'Salted Earth',
+    '4B5': 'Flamethrower',
+    '2E3': 'Asylum',
+    '777': 'Asylum',
+    '798': 'Sacred Soil',
+    '4C7': 'Fey Union',
+    '742': 'Nascent Glint',
+  };
 }

 export class LineEvent24 extends LineEvent0x18 {}
-
-LineEvent0x18.showEffectNamesFor = {
-  '4C4': 'Excognition',
-  '35D': 'Wildfire',
-  '1F5': 'Doton',
-  '2ED': 'Salted Earth',
-  '4B5': 'Flamethrower',
-  '2E3': 'Asylum',
-  '777': 'Asylum',
-  '798': 'Sacred Soil',
-  '4C7': 'Fey Union',
-  '742': 'Nascent Glint',
-};
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x19.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x19.ts
index 32623caa..8ff738ba 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x19.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x19.ts
@@ -1,14 +1,35 @@
 import LineEvent from './LineEvent';
 import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x19Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Target ID
+  string, // 3: Target Name
+  string, // 4: Source ID
+  string, // 5: Source Name
+  string, // 6: Unknown/Blank?
+  string, // 7: Checksum
+];

 // Combatant defeated event
 export class LineEvent0x19 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public targetId: string;
+  public targetName: string;
+  public resolvedName: string | false;
+  public resolvedTargetName: string | false;
+  public properCaseConvertedLine = '';
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x19Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
     this.name = parts[3];
     repo.updateCombatant(this.id, {
+      job: undefined,
       name: this.name,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -17,6 +38,7 @@ export class LineEvent0x19 extends LineEvent {
     this.targetId = parts[4].toUpperCase();
     this.targetName = parts[5];
     repo.updateCombatant(this.targetId, {
+      job: undefined,
       name: this.targetName,
       spawn: this.timestamp,
       despawn: this.timestamp,
@@ -31,14 +53,16 @@ export class LineEvent0x19 extends LineEvent {
       this.resolvedTargetName = repo.resolveName(this.targetId, this.targetName);
   }

-  convert() {
-    this.convertedLine = this.prefix() +
-      this.resolvedName || this.name +
-      ' was defeated by ' + this.resolvedTargetName || this.targetName + '.';
+  convert(): void {
+    this.convertedLine = `\
+${this.prefix()}\
+${this.resolvedName || this.name} \
+was defeated by ${this.resolvedTargetName || this.targetName}.`;

-    this.properCaseConvertedLine = this.prefix() +
-      EmulatorCommon.properCase(this.resolvedName || this.name) +
-      ' was defeated by ' + EmulatorCommon.properCase(this.resolvedTargetName || this.targetName) + '.';
+    this.properCaseConvertedLine = `\
+${this.prefix()}\
+${EmulatorCommon.properCase(this.resolvedName || this.name)} \
+was defeated by ${EmulatorCommon.properCase(this.resolvedTargetName || this.targetName)}.`;
   }
 }

```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1A.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1A.ts
index 7b99ffac..2e1b2923 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1A.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1A.ts
@@ -1,9 +1,40 @@
 import LineEvent from './LineEvent';
 import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x1AParts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Ability ID
+  string, // 3: Ability Name
+  string, // 4: Duration
+  string, // 5: Source ID
+  string, // 6: Source Name
+  string, // 7: Target ID
+  string, // 8: Target Name
+  string, // 9: Stacks
+  string, // 10: Target HP???
+  string, // 11: Source HP???
+  string, // 12: Unknown?
+  string, // 13: Checksum
+];

 // Gain status effect event
 export class LineEvent0x1A extends LineEvent {
-  constructor(repo, line, parts) {
+  public abilityId: string;
+  public abilityName: string;
+  public duration: number;
+  public stacks: number;
+  public id: string;
+  public name: string;
+  public targetId: string;
+  public targetName: string;
+  public resolvedName: string;
+  public resolvedTargetName: string;
+  public fallbackResolvedTargetName: string;
+  public properCaseConvertedLine = '';
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x1AParts) {
     super(repo, line, parts);

     this.abilityId = parts[2].toUpperCase();
@@ -18,6 +49,7 @@ export class LineEvent0x1A extends LineEvent {
       name: this.name,
       spawn: this.timestamp,
       despawn: this.timestamp,
+      job: undefined,
     });

     this.targetId = parts[7].toUpperCase();
@@ -26,6 +58,7 @@ export class LineEvent0x1A extends LineEvent {
       name: this.targetName,
       spawn: this.timestamp,
       despawn: this.timestamp,
+      job: undefined,
     });

     this.resolvedName = repo.resolveName(this.id, this.name);
@@ -35,34 +68,33 @@ export class LineEvent0x1A extends LineEvent {
       repo.resolveName(this.id, this.name, this.targetId, this.targetName);
   }

-  convert() {
+  convert(): void {
     let stackCountText = '';
     if (this.stacks > 0 && this.stacks < 20 &&
       LineEvent0x1A.showStackCountFor.includes(this.abilityId))
-      stackCountText = ' (' + this.stacks + ')';
+      stackCountText = ` (${this.stacks})`;

-    this.convertedLine = this.prefix() +
-      this.targetId + ':' + this.targetName +
-      ' gains the effect of ' + this.abilityName +
-      ' from ' + this.fallbackResolvedTargetName +
-      ' for ' + this.parts[4] + ' Seconds.' + stackCountText;
+    this.convertedLine = `\
+${this.prefix()}${this.targetId}:${this.targetName} \
+gains the effect of ${this.abilityName} \
+from ${this.fallbackResolvedTargetName} \
+for ${this.parts[4]} Seconds.${stackCountText}`;

-    this.properCaseConvertedLine = this.prefix() +
-      this.targetId + ':' + EmulatorCommon.properCase(this.targetName) +
-      ' gains the effect of ' + this.abilityName +
-      ' from ' + EmulatorCommon.properCase(this.fallbackResolvedTargetName) +
-      ' for ' + this.parts[4] + ' Seconds.' + stackCountText;
+    this.properCaseConvertedLine = `\
+${this.prefix()}${this.targetId}:${EmulatorCommon.properCase(this.targetName)} \
+gains the effect of ${this.abilityName} \
+from ${EmulatorCommon.properCase(this.fallbackResolvedTargetName)} \
+for ${this.parts[4]} Seconds.${stackCountText}`;
   }
+  static showStackCountFor: readonly string[] = [
+    '130', // Aetherflow
+    '196', // Vulnerability Down
+    '15e', // Vulnerability Down
+    '2ca', // Vulnerability Up
+    '1f9', // Damage Up
+    '4d7', // Embolden
+    '511', // Embolden
+  ] as const;
 }

 export class LineEvent26 extends LineEvent0x1A {}
-
-LineEvent0x1A.showStackCountFor = [
-  '130', // Aetherflow
-  '196', // Vulnerability Down
-  '15e', // Vulnerability Down
-  '2ca', // Vulnerability Up
-  '1f9', // Damage Up
-  '4d7', // Embolden
-  '511', // Embolden
-];
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1B.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1B.ts
index 9fd6d54b..25081d79 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1B.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1B.ts
@@ -1,8 +1,27 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x1BParts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Target ID
+  string, // 3: Target Name
+  string, // 4: Unknown?
+  string, // 5: Unknown?
+  string, // 6: Headmarker ID
+  string, // 7: 0000
+  string, // 8: 0000
+  string, // 9: 0000
+  string, // 10: Unknown/Blank?
+  string, // 11: Checksum
+];

 // Head marker event
 export class LineEvent0x1B extends LineEvent {
-  constructor(repo, line, parts) {
+  public targetId: string;
+  public targetName: string;
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x1BParts) {
     super(repo, line, parts);

     this.targetId = parts[2].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1C.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1C.ts
index 919b6cf4..ade81abf 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1C.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1C.ts
@@ -1,8 +1,25 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x1CParts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Operation
+  string, // 3: Waymark
+  string, // 4: Source ID
+  string, // 5: Source Name
+  string, // 6: X
+  string, // 7: Y
+  string, // 8: Z?
+  string, // 9: Checksum
+];

 // Floor waymarker event
 export class LineEvent0x1C extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x1CParts) {
     super(repo, line, parts);

     this.id = parts[4].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1D.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1D.ts
index 02df5ef1..c09e28e5 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1D.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1D.ts
@@ -1,8 +1,22 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x1DParts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Operation
+  string, // 3: Waymark
+  string, // 4: Source ID
+  string, // 5: Source Name
+  string, // 6: Target ID
+  string, // 7: Target Name
+  string, // 8: Unknown/Blank?
+  string, // 9: Checksum
+];

 // Waymarker
 export class LineEvent0x1D extends LineEvent {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x1DParts) {
     super(repo, line, parts);
   }
 }
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1E.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1E.ts
index cb8b3ecd..5582b495 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1E.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x1E.ts
@@ -1,31 +1,33 @@
-import LineEvent from './LineEvent';
-import { LineEvent0x1A } from './LineEvent0x1A';
+import { LineEvent0x1A, LineEvent0x1AParts } from './LineEvent0x1A';
 import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x1EParts = LineEvent0x1AParts;

 // Lose status effect event
 // Extend the gain status event to reduce duplicate code since they're
 // the same from a data perspective
 export class LineEvent0x1E extends LineEvent0x1A {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x1EParts) {
     super(repo, line, parts);
   }
-  convert() {
+  convert(): void {
     let stackCountText = '';
     if (this.stacks > 0 && this.stacks < 20 &&
       LineEvent0x1A.showStackCountFor.includes(this.abilityId))
-      stackCountText = ' (' + this.stacks + ')';
+      stackCountText = ` (${this.stacks})`;

-    this.convertedLine = this.prefix() +
-      this.targetId + ':' + this.targetName +
-      ' loses the effect of ' + this.abilityName +
-      ' from ' + this.fallbackResolvedTargetName +
-      '.' + stackCountText;
+    this.convertedLine = `\
+${this.prefix()}${this.targetId}:${this.targetName} \
+loses the effect of ${this.abilityName} \
+from ${this.fallbackResolvedTargetName} \
+for ${this.parts[4]} Seconds.${stackCountText}`;

-    this.properCaseConvertedLine = this.prefix() +
-      this.targetId + ':' + EmulatorCommon.properCase(this.targetName) +
-      ' loses the effect of ' + this.abilityName +
-      ' from ' + EmulatorCommon.properCase(this.fallbackResolvedTargetName) +
-      '.' + stackCountText;
+    this.properCaseConvertedLine = `\
+${this.prefix()}${this.targetId}:${EmulatorCommon.properCase(this.targetName)} \
+loses the effect of ${this.abilityName} \
+from ${EmulatorCommon.properCase(this.fallbackResolvedTargetName)} \
+for ${this.parts[4]} Seconds.${stackCountText}`;
   }
 }

```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x22.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x22.ts
index 3a1e2b90..df21aa2c 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x22.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x22.ts
@@ -1,8 +1,26 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x22Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Target ID
+  string, // 3: Target Name
+  string, // 4: Source ID
+  string, // 5: Source Name
+  string, // 6: Enable Flag
+  string, // 7: Checksum
+];

 // Nameplate toggle
 export class LineEvent0x22 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public targetId: string;
+  public targetName: string;
+  public targetable: boolean;
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x22Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x23.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x23.ts
index 9dffd461..271d6fda 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x23.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x23.ts
@@ -1,8 +1,31 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x23Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Target ID
+  string, // 5: Target Name
+  string, // 6: Unknown?
+  string, // 7: Unknown?
+  string, // 8: Tether ID
+  string, // 9: Unknown?
+  string, // 10: Unknown?
+  string, // 11: Unknown?
+  string, // 12: Unknown/Blank?
+  string, // 13: Checksum
+];

 // Tether event
 export class LineEvent0x23 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public targetId: string;
+  public targetName: string;
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x23Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x24.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x24.ts
index bd6cb7ec..906fd9d7 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x24.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x24.ts
@@ -1,16 +1,29 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x24Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Value
+  string, // 3: Bars
+  string, // 4: Checksum
+];

 // Limit gauge event
 export class LineEvent0x24 extends LineEvent {
-  constructor(repo, line, parts) {
+  public valueDec: number;
+  public valueHex: string;
+  public bars: string;
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x24Parts) {
     super(repo, line, parts);
-    this.value = parseInt(parts[2], 16);
+    this.valueHex = parts[2];
+    this.valueDec = parseInt(this.valueHex, 16);
     this.bars = parts[3];
   }

-  convert() {
-    this.convertedLine = this.prefix() +
-      'Limit Break: ' + this.parts[2];
+  convert(): void {
+    this.convertedLine = `${this.prefix()}Limit Break: ${this.valueHex}`;
   }
 }

```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x25.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x25.ts
index a09779be..9c355702 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x25.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x25.ts
@@ -1,9 +1,44 @@
 import LineEvent from './LineEvent';
-import { LineEvent0x1E } from './LineEvent0x1E';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x25Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Sequence ID
+  string, // 5: Current HP
+  string, // 6: Max HP
+  string, // 7: Current MP
+  string, // 8: Max MP
+  string, // 9: Current TP
+  string, // 10: Max TP
+  string, // 11: X
+  string, // 12: Y
+  string, // 13: Z
+  string, // 14: Heading
+  string, // 15: Unknown?
+  string, // 16: Unknown?
+  string, // 17: Unknown?
+  string, // 18: Unknown?
+  string, // 19: Unknown/Blank?
+  string, // 20: Checksum
+];

 // Action sync event
 export class LineEvent0x25 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public hp: string;
+  public maxHp: string;
+  public mp: string;
+  public maxMp: string;
+  public x: string;
+  public y: string;
+  public z: string;
+  public heading: string;
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x25Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x26.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x26.ts
index 5dab9d0e..d56cdd66 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x26.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x26.ts
@@ -1,10 +1,49 @@
 import LineEvent from './LineEvent';
 import EmulatorCommon from '../../EmulatorCommon';
 import Util from '../../../../../resources/util';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x26Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Job/Level Data
+  string, // 5: Current HP
+  string, // 6: Max HP
+  string, // 7: Current MP
+  string, // 8: Max MP
+  string, // 9: Current TP
+  string, // 10: Max TP
+  string, // 11: X
+  string, // 12: Y
+  string, // 13: Z
+  string, // 14: Heading
+  string, // 15: Unknown?
+  string, // 16: Unknown?
+  string, // 17: Unknown?
+  string, // 18: Unknown/Blank?
+  string, // 19: Checksum
+];

 // Network status effect event
 export class LineEvent0x26 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public jobIdHex: string;
+  public jobIdDec: number;
+  public jobName: string;
+  public level: number;
+  public hp: string;
+  public maxHp: string;
+  public mp: string;
+  public maxMp: string;
+  public x: string;
+  public y: string;
+  public z: string;
+  public heading: string;
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x26Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
```
```diff
diff --git a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x27.js b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x27.ts
index 1aefdb98..aff24bc2 100644
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent0x27.js
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent0x27.ts
@@ -1,8 +1,39 @@
 import LineEvent from './LineEvent';
+import LogRepository from './LogRepository';
+
+export type LineEvent0x27Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Source ID
+  string, // 3: Source Name
+  string, // 4: Current HP
+  string, // 5: Max HP
+  string, // 6: Current MP
+  string, // 7: Max MP
+  string, // 8: Current TP
+  string, // 9: Max TP
+  string, // 10: X
+  string, // 11: Y
+  string, // 12: Z
+  string, // 13: Heading
+  string, // 14: Unknown/Blank?
+  string, // 15: Checksum
+];

 // Network update hp event
 export class LineEvent0x27 extends LineEvent {
-  constructor(repo, line, parts) {
+  public id: string;
+  public name: string;
+  public hp: string;
+  public maxHp: string;
+  public mp: string;
+  public maxMp: string;
+  public x: string;
+  public y: string;
+  public z: string;
+  public heading: string;
+
+  constructor(repo: LogRepository, line: string, parts: LineEvent0x27Parts) {
     super(repo, line, parts);

     this.id = parts[2].toUpperCase();
```

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
```diff
@@ -1,14 +1,24 @@
...
+  string, // 2: Type
+  string, // 3: Speaker
+  string, // 4: Message
+  string, // 5: Checksum
+];
 
 // Chat event
 export class LineEvent0x00 extends LineEvent {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x00Parts) {
```
quisquous, line 15: `LineEvent0x00Parts` seems a bit suspicious.  I think `parts` comes from `split('|')` and from data off disk (possibly constructed, corrupted, incorrect).  So, it really seems like we don't know anything more about it other than `string[]`.  I don't think it makes sense to assert via `as` that the incoming parameter has 6 string fields, as that assertion doesn't add any type safety.   It seems safer to say `parts: string[]` and just handle the potentially undefined parts as needed.  If it

Additionally, I think it should be `parts: string[]` and not `public parts: string[]`, as `LineEvent` already stores parts and `LineEvent0x00` doesn't need to also?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
```diff
@@ -1,14 +1,24 @@
...
     super(repo, line, parts);
     if (this.parts.slice(4).join('|').split('\u001f\u001f').length > 1)
```
quisquous, line 17: Old code review, but can you leave a comment for whatever this line is doing?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.ts
```diff
@@ -0,0 +1,143 @@
...
+    this.flags = parts[8];
+
+    let offset = 0;
+
+    if (this.flags === '3F')
+      offset = 2;
+
+    this.damage = LineEvent.calculateDamage(parts[9 + offset] as string);
+
+    this.targetHp = parts[24 + offset] as string;
```
quisquous, line 118: What are all the `as string` doing here? Can they be removed?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x03.ts
```diff
@@ -36,19 +79,19 @@ export class LineEvent0x03 extends LineEvent {
...
-      ' Max MP: ' + this.maxMp +
-      ' Pos: (' + this.parts[17] + ',' + this.parts[18] + ',' + this.parts[19] + ')';
+    this.convertedLine = `\
+${this.prefix()}${this.id.toUpperCase()}:\
+Added new combatant ${CombatantName}.\
+  Job: ${this.jobName}\
+ Level: ${this.level}\
+ Max HP: ${this.maxHp}\
+ Max MP: ${this.maxMp}\
+ Pos: (${this.parts[17]},${this.parts[18]},${this.parts[19]})`;
```
quisquous, line 94:
```suggestion
    this.convertedLine = this.prefix() +
      this.id.toUpperCase() + ':' +
      'Added new combatant ' + CombatantName +
      '.  Job: ' + this.jobName +
      ' Level: ' + this.level.toString() +
      ' Max HP: ' + this.maxHp.toString() +
      ' Max MP: ' + this.maxMp.toString() +
      ' Pos: (' + this.parts[17] + ',' + this.parts[18] + ',' + this.parts[19] + ')';
```

I find this multiline template literal form to not be very readable as compared to the original code.  (Mostly it's the indentation that's very jarring to my eyes?) What about using the previous code with `toString()` attached to anything that's not a string?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
```diff
@@ -1,14 +1,24 @@
...
+import LogRepository from './LogRepository';
+
+export type LineEvent0x00Parts = [
+  string, // 0: ID
+  string, // 1: Timestamp
+  string, // 2: Type
+  string, // 3: Speaker
+  string, // 4: Message
+  string, // 5: Checksum
+];
```
quisquous, line 11: I think if you want to document what the various parts are here, I think there are stronger ways to do it where the names become identifiers rather than comments.

One option here is to make an "enum" (e.g. `const field = { id: 0, timestamp: 1, etc } as const;` and then some other code can say `parts[field.type]` instead of `parts[2]`).

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
valarnin: This is a typecasting adjustment. Without the custom type, the array is actually read as `(string | undefined)[]` for some reason, so it results in this:

```typescript
  constructor(repo: LogRepository, line: string, parts: string[]) {
    super(repo, line, parts);

    this.id = parts[2].toUpperCase();
    this.name = parts[3];

    this.abilityId = parts[4].toUpperCase();
    this.abilityName = parts[5];
  }
```

```
[tsl] ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts(26,15)
      TS2532: Object is possibly 'undefined'.

ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts
27:4-13
[tsl] ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts(27,5)
      TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts
29:21-29
[tsl] ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts(29,22)
      TS2532: Object is possibly 'undefined'.

ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts
30:4-20
[tsl] ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x17.ts(30,5)
      TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

The `public` assignment is to avoid this issue in the `convert` function, as well.

quisquous: I'm trying to argue that TypeScript is correct, and that it is potentially undefined.  Asserting that it is defined does not make it so.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
valarnin: I'm trying to remember back to the reason this exists. I think it has something to do with this issue which we were unable to reproduce, so it can probably be removed? I don't have any logs at hand which reproduce this behavior right now so I'm not sure.

https://github.com/ravahn/FFXIV_ACT_Plugin/issues/250

ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.ts
valarnin: More weird behavior with `(string | undefined)`. Removing the `as string` for these results in the same errors as outlined in the 0x00 file, despite the `parts` parameter being properly typed, because the indexing of `parts` is being adjusted by the offset.

```
ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x15.ts
116:44-61
[tsl] ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x15.ts(116,45)
      TS2345: Argument of type 'string | undefined' is not assignable to parameter of type 'string'.
  Type 'undefined' is not assignable to type 'string'.

ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x15.ts
118:4-17
[tsl] ERROR in C:\Users\valar\Source\cactbot2\ui\raidboss\emulator\data\network_log_converter\LineEvent0x15.ts(118,5)
      TS2322: Type 'string | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.
```

ui/raidboss/emulator/data/network_log_converter/LineEvent0x03.ts
valarnin: I have no strong opinions either way on this one, I can swap back to string appending instead. I assume you want this adjustment for all the line events for consistency?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
valarnin: I don't know how that'll pan out with the issues outlined in the first review re: `as string` etc.

valarnin: Additionally, that's really not going to work for 0x15 and 0x16, with the offset stuff.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x03.ts
quisquous: Yeah, that would be great, thanks!

ui/raidboss/emulator/data/network_log_converter/LineEvent0x15.ts
quisquous: I think if `parts` is just a `string[]` and you treat everything as potentially undefined (either by `this.hp = parts[7] ?? '0') or `hp?: number` in its declaration) then you don't need `as string` here.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
valarnin: Throwing an `assert` to check that it's actually the correct length and all values are string doesn't fix tsc complaining about it potentially being undefined. I'm not sure what would fix this satisfactorily. Does the constructor become 200+ lines to handle the potential undefined values (in the case of 0x15 lines for example) to work around the issue?

Adding a check `if parts length is less than expected or anything in it isn't a string` fixes the logic issue but tsc still complains that it could be undefined even though we're checking that it's not already.

quisquous: Yeah, agreed.  A length check is not sufficient for TypeScript.

What about nullish coalescing?  I don't think that would add a ton of extra lines.

```typescript
export class LineEvent0x00 extends LineEvent {
  constructor(repo: LogRepository, line: string, parts: string[]) {
    super(repo, line, parts);
    if (this.parts.slice(4).join('|').split('\u001f\u001f').length > 1)
      this.invalid = true;
  }

  convert(_: LogRepository): void {
    const type = this.parts[2] ?? '';
    // If speaker is blank, it's excluded from the converted line
    const speaker = this.parts[3] && this.parts[3] !== '' ? `${this.parts[3]}:` : '';
    this.convertedLine =
      this.prefix() +
      type + ':' +
      speaker +
      this.parts.slice(4, this.parts.length - 1).join('|').trim();
    this.convertedLine = LineEvent00.replaceChatSymbols(this.convertedLine);
  }
```

valarnin: That seems to resolve the issues properly, but it's still going to be ugly for some of the classes. For example, these cases:
```typescript
    this.abilityId = parts[2].toUpperCase();
...
    this.stacks = parseInt(parts[9], 16);
...
    this.convertedLine = this.prefix() + this.targetId + ':' + this.targetName +
    ' gains the effect of ' + this.abilityName +
    ' from ' + this.fallbackResolvedTargetName +
    ' for ' + this.parts[4] +
    ' Seconds.' + stackCountText;
```

Compare to:

```typescript
    this.abilityId = (parts[2] ?? '').toUpperCase();
...
    this.stacks = parseInt(parts[9] ?? '0', 16);
...
    this.convertedLine = this.prefix() + this.targetId + ':' + this.targetName +
    ' gains the effect of ' + this.abilityName +
    ' from ' + this.fallbackResolvedTargetName +
    ' for ' + (this.parts[4] ?? '') +
    ' Seconds.' + stackCountText;
```

quisquous: Personally, I think nullish coalescing is better than using `as`.  There's probably cases right now where the raid emulator will throw an exception if you have some log with a corrupted line in it (such that the line is truncated and leaves some field as undefined), so I think this is a robustness improvement even if it makes the code slightly more awkward.

* `this.abilityId = parts[2]?.toUpperCase() ?? '';` is slightly more readable, as an alternative
* `this.stacks = parseInt(parts[9] ?? '0', 16);` seems fine (or alternatively make some `EmulatorCommon.parseHex(str?: string)` helper?)
* `(this.parts[4] ?? '')` can get a named temporary, like the previous comment, which I think is a readability improvement over the previous code

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
```diff
@@ -1,14 +1,24 @@
...
 export class LineEvent0x00 extends LineEvent {
-  constructor(repo, line, parts) {
+  constructor(repo: LogRepository, line: string, public parts: LineEvent0x00Parts) {
     super(repo, line, parts);
     if (this.parts.slice(4).join('|').split('\u001f\u001f').length > 1)
       this.invalid = true;
   }
 
-  convert() {
+  convert(_: LogRepository): void {
```
panicstevenson, line 21: Why is this signature different from the other `convert`s? Also, I would expect LineEvent to have this as an optional parameter?

valarnin: We had a whole conversation on this in the PR for the parent `LineEvent` class, here's a (hopefully working) link:
https://github.com/quisquous/cactbot/pull/2875#discussion_r636566724

All of the `convert` signatures should include a `LogRepository` argument, unless I missed one somewhere.

panicstevenson: I guess what I am trying to ask is why does 0x00 have this signature, which is different from all the other converts, aside from 0x1F (which is understandable)

quisquous: > All of the convert signatures should include a LogRepository argument, unless I missed one somewhere.

It looks like you missed a number of them.

valarnin: They should have all said they had a `LogRepository` param, apparently my regex wasn't correct for checking these. I'll fix for the next commit, thanks for spotting it.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
valarnin: I'm working on cleaning this up now, but I have a follow up question. In the case that a log line is invalid such that it can't realistically be used in the emulator, should we mark it as invalid so that it's ignored? That would be a change to current behavior so it might be better looked into as a future PR?

quisquous: Yeah, I think that'd be a nice follow-up.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
MaikoTan: What about writing like this:
```typescript
export type LineEvent0x00Parts = [
  id: string,
  timestamp: string,
  type: string,
  speaker: string,
  message: string,
  checksum: string,
];
```

So you could get a type hint as well.
image

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
MaikoTan: There is a trick in typescript to check if an array has a constant length, which is to add a `is-check` function, for expample:
```typescript
function isLineEvent0x00Parts(array: string[]): array is LineEvent0x00Parts {
  return array.length === 6;
}
```

I think this way would be better?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts
valarnin: Given that many of these properties are needed regardless, maybe it makes sense to just flat out make these getters on the class, such that:

```typescript

// Chat event
export class LineEvent0x00 extends LineEvent {
  constructor(repo: LogRepository, line: string, public parts: LineEvent0x00Parts) {
    super(repo, line, parts);
    if (this.parts.slice(4).join('|').split('\u001f\u001f').length > 1)
      this.invalid = true;
  }

  public get id(): string {
    return this.parts[0];
  }

  public get timestamp(): string {
    return this.parts[1];
  }
...
```

Thoughts?

quisquous: I assume with nullish coalescing or something to handle missing parts as well?

```typescript
  public get id(): string {
    return this.parts[0] ?? '';
  }

  public get timestamp(): string {
    return this.parts[1] ?? '';
  }
  ```

That seems reasonable to me!

Comment (valarnin) [reply ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts:11]: Yeah, with nullish coalescing, and `parseInt`/`parseFloat` where appropriate.

Comment (valarnin) [reply ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts:11]: Given this adjustment, I'm going to pull `LineEvent.ts` into this PR to put `id` and `timestamp` on the base class, otherwise it'll cause issues and those fields will have to be adjusted again later.

File: ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts (hunk 2 above)
Comment (valarnin) [reply ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts]: I think this is the only issue not addressed in the latest commit. I'm not sure what the best course here is, removing the check or adding a comment with what I think was the reason?

**Comment by valarnin** [comment]
Per various conversations on reviews, I've redone the conversion to typescript for these files. Please re-review.

File: ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts (hunk 2 above)
Comment (quisquous) [reply ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts]: I think a comment would be good.  If you think it can be removed, feel free to send another PR to remove it in the future?

Comment (valarnin) [reply ui/raidboss/emulator/data/network_log_converter/LineEvent0x00.ts]: I guess a comment like this might work?

```typescript
// The exact reason for this check isn't clear anymore but may be related to
// https://github.com/ravahn/FFXIV_ACT_Plugin/issues/250
```

ui/raidboss/emulator/data/network_log_converter/LineEvent0x01.ts
```diff
@@ -1,16 +1,24 @@
...
 // Zone change event
 export class LineEvent0x01 extends LineEvent {
-  constructor(repo, line, parts) {
-    super(repo, line, parts);
-    this.zoneId = parts[2].toUpperCase();
-    this.zoneName = parts[3];
-    this.zoneNameProperCase = EmulatorCommon.properCase(this.zoneName);
+  public properCaseConvertedLine = '';
+
+  public get zoneID(): string {
```
quisquous, line 9:
```suggestion
  public get zoneId(): string {
```

To match other uses of `zoneId` and `thingId` elsewhere.

ui/raidboss/emulator/data/network_log_converter/LineEvent0x03.ts
```diff
@@ -0,0 +1,151 @@
...
+      ' Max HP: ' + this.maxHpString +
+      ' Max MP: ' + this.maxMpString +
+      ' Pos: (' + this.xString + ',' + this.yString + ',' + this.zString + ')';
+
+    // This last part is guesswork for the area between 9 and 10.
+    const UnknownValue = this.npcNameId +
+      EmulatorCommon.zeroPad(this.npcBaseId, 8 + Math.max(0, 6 - this.npcNameId.length));
+
+    if (UnknownValue !== '00000000000000')
+      this.convertedLine += ' (' + UnknownValue + ')';
```
quisquous, line 145:
```suggestion
    const unknownValue = this.npcNameId +
      EmulatorCommon.zeroPad(this.npcBaseId, 8 + Math.max(0, 6 - this.npcNameId.length));

    if (unknownValue !== '00000000000000')
      this.convertedLine += ' (' + unknownValue + ')';
```

ui/raidboss/emulator/data/network_log_converter/LineEvent0x18.ts
```diff
@@ -0,0 +1,114 @@
+import LineEvent from './LineEvent';
+import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+// DoT/HoT event
+export class LineEvent0x18 extends LineEvent {
+  public resolvedName: string;
+  public effectName: string | undefined = '';
```
quisquous, line 8:
```suggestion
  public effectName?: string;
```

ui/raidboss/emulator/data/network_log_converter/LineEvent0x19.ts
```diff
@@ -0,0 +1,63 @@
+import LineEvent from './LineEvent';
+import EmulatorCommon from '../../EmulatorCommon';
+import LogRepository from './LogRepository';
+
+// Combatant defeated event
+export class LineEvent0x19 extends LineEvent {
+  public resolvedName: string | false;
+  public resolvedTargetName: string | false;
```
quisquous, line 8:
```suggestion
  public resolvedName?: string;
  public resolvedTargetName?: string;
```

What about undefined instead of false as an unset type?

ui/raidboss/emulator/data/network_log_converter/LineEvent0x19.ts
```diff
@@ -0,0 +1,63 @@
...
+    return this.parts[4]?.toUpperCase() ?? '';
+  }
+
+  public get targetName(): string {
+    return this.parts[5] ?? '';
+  }
+
+  convert(_: LogRepository): void {
+    const defeatedName = (this.resolvedName || this.name);
+    const killerName = (this.resolvedTargetName || this.targetName);
```
quisquous, line 55:
```suggestion
    const defeatedName = (this.resolvedName ?? this.name);
    const killerName = (this.resolvedTargetName ?? this.targetName);
```

quisquous:
Oops, I didn't mean to merge this, but I also only had some small comments left.  Would you mind finishing those in another PR rather than me reverting?  Sorry for the clumsiness.

valarnin:
> Oops, I didn't mean to merge this, but I also only had some small comments left. Would you mind finishing those in another PR rather than me reverting? Sorry for the clumsiness.

Yeah. I don't have any more free time tonight to work on this but I'll get a new PR in tomorrow for the outstanding comments.

PR #2790, opened 2021-04-24
raidboss: Convert Test to TypeScript

panicstevenson:

PR #2880, opened 2021-05-22
raidboss: Add TriggerSet types; convert some initial triggers

quisquous:
This is mostly just trying to put together some thoughts of what this might look like.  Feedback super welcome.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -66,6 +66,8 @@ module.exports = {
         'rulesdir/cactbot-output-strings': 'error',
         'rulesdir/cactbot-timeline-triggers': 'error',
         'rulesdir/cactbot-response-default-severities': 'error',
+        // Raidboss data files always export a trigger set, and explicit types are noisy.
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
       },
     },
   ],
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -64,6 +64,8 @@ module.exports = {
         'no-unused-vars': ['error', { 'args': 'all', 'argsIgnorePattern': '^_\\w+' }],
         'prefer-arrow/prefer-arrow-functions': 'warn',
         'rulesdir/cactbot-output-strings': 'error',
+        // Only meant to be used for `output` parameters!
+        '@typescript-eslint/no-non-null-assertion': 'off',
         'rulesdir/cactbot-timeline-triggers': 'error',
         'rulesdir/cactbot-response-default-severities': 'error',
         // Raidboss data files always export a trigger set, and explicit types are noisy.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -72,21 +72,11 @@ type OptionalUnlessVoid<T> = T extends void ? void : T | undefined;
 export type TriggerField<Data, Return> =
     TriggerFunc<Data, MatchesAny, OptionalUnlessVoid<Return>> | OptionalUnlessVoid<Return>;
 
-// TODO: I am not sure that it is possible to type triggers such that we have some gigantic union
-// of Trigger<TetherParams> | Trigger<AbilityParams> | Trigger<AddedCombatantParams> | etc etc
-// and so the idea is that individual files would use a narrower type for Matches if they wanted to
-// and that's something we (maybe?) could enforce via tests rather than via TypeScript.  This
-// approach would not preclude some TypeScript gigabrain solution that can figure out how to do
-// the same thing at the TypeScript level.
-//
-// TODO: I think there are two types of TriggerSets.  There is the "cactbot + user defined triggers"
-// and also the "cactbot built-in only" triggers, which can be more strict.  For example,
-// * is `id` required, and `regex` for timeline triggers? (yes for built-in, no for user triggers)
-// * should we handle nonsense like `condition` returning non-booleans?
-// We could have one type that built-in triggers define themselves as more strictly and then a
-// second type that popup-text.js uses and needs to do more narrowing for.
+// This trigger type is what we expect cactbot triggers to be written as,
+// in other words `id` is not technically required for triggers but for
+// built-in triggers it is.
 export type BaseTrigger<Data> = {
-  id?: string;
+  id: string;
   disabled?: boolean;
   condition?: TriggerField<Data, boolean>;
   preRun?: TriggerField<Data, void>;
@@ -126,25 +116,21 @@ export type RegexTrigger<Data> = BaseTrigger<Data> & {
 }
 
 export type TimelineTrigger<Data> = BaseTrigger<Data> & {
-  regex?: RegExp;
-  // TODO: can this also be a function?
-  beforeSeconds?: number;
+  regex: RegExp;
+  beforeSeconds: number;
 };
-
-export type Trigger<Data> = RegexTrigger<Data> | NetRegexTrigger<Data>;
 
 // Because timeline functions run during loading, they only support the base RaidbossData.
 export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
 
 export type TriggerSet<Data> = {
   // ZoneId.MatchAll (aka null) is not supported in array form.
-  zoneId?: ZoneId | number[];
-  zoneRegex?: RegExp | { [lang in Lang]?: RegExp };
+  zoneId: ZoneId | number[];
   resetWhenOutOfCombat?: boolean;
   overrideTimelineFile?: boolean;
   timelineFile?: string;
   timeline?: TimelineFunc;
-  triggers?: Trigger<Data>[];
+  triggers?: NetRegexTrigger<Data>[];
   timelineTriggers?: TimelineTrigger<Data>[];
   timelineReplace?: {
     locale: Lang;
@@ -152,4 +138,17 @@ export type TriggerSet<Data> = {
     replaceText?: { [regex: string]: string };
     replaceSync?: { [regex: string]: string };
   }[];
+}
+
+// Less strict type for user triggers + built-in triggers, including deprecated fields.
+export type LooseTimelineTrigger = Partial<TimelineTrigger<RaidbossData>>;
+
+export type LooseTrigger =
+    Partial<RegexTrigger<RaidbossData>> | Partial<NetRegexTrigger<RaidbossData>>;
+
+export type LooseTriggerSet = Exclude<Partial<TriggerSet<RaidbossData>>, 'triggers' | 'timelineTriggers'> & {
+    /** @deprecated Use zoneId instead */
+    zoneRegex?: RegExp | { [lang in Lang]?: RegExp };
+    triggers?: LooseTrigger[];
+    timelineTriggers?: LooseTimelineTrigger[];
 }
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -131,7 +131,8 @@ export type Trigger = RegexTrigger | NetRegexTrigger;
 export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
 
 export type TriggerSet = {
-  zoneId?: ZoneId | ZoneId[];
+  // ZoneId.MatchAll (aka null) is not supported in array form.
+  zoneId?: ZoneId | number[];
   zoneRegex?: RegExp | { [lang in Lang]?: RegExp };
   resetWhenOutOfCombat?: boolean;
   overrideTimelineFile?: boolean;
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -1,4 +1,5 @@
 import { Lang, NonEnLang } from './global';
+import { RaidbossData } from './data';
 
 export interface BaseRegExp<T> extends RegExp {
   groups?: {
@@ -37,34 +38,111 @@ export type Output = {
 
 // The output of any non-response raidboss trigger function.
 export type TriggerOutput =
-    void | undefined | null | LocaleText | string | number | (() => TriggerOutput);
+    undefined | null | LocaleText | string | number | boolean | (() => TriggerOutput);
 
 // The type of a non-response trigger field.
-export type TriggerFunc<Matches> = (data: Data, matches: Matches, output: Output) => TriggerOutput;
+export type TriggerFunc<Matches, Return> =
+    (data: RaidbossData, matches: Matches, output: Output) => Return;
 
 // Valid fields to return from a ResponseFunc.
 type ResponseFields = 'infoText' | 'alertText' | 'alarmText' | 'tts';
 
 // The output from a response function (different from other TriggerOutput functions).
 export type ResponseOutput<Matches> = {
-  [text in ResponseFields]?: TriggerFunc<Matches>;
+  [text in ResponseFields]?: TriggerFunc<Matches, TriggerOutput>;
 };
 // The type of a response trigger field.
 export type ResponseFunc<Matches> =
-    (data: Data, matches: Matches, output: Output) => ResponseOutput<Matches>;
+    (data: RaidbossData, matches: Matches, output: Output) => ResponseOutput<Matches>;
 
-export type Trigger = {
-  id: string;
-  regex: RegExp;
-  // TODO: complete this type
-  [key: string]: unknown;
-}
-
-export type TimelineTrigger = Trigger & { beforeSeconds: number };
+export type ResponseField = ResponseFunc<MatchesAny> | ResponseOutput<MatchesAny>;
 
 export type TriggerAutoConfig = {
   Output?: Output;
   Duration?: number;
   BeforeSeconds?: number;
   OutputStrings?: { [outputKey: string]: Lang };
+}
+
+export type MatchesAny = { [s in T]?: string } | undefined;
+export type TriggerField<Return> = TriggerFunc<MatchesAny, Return> | Return;
+
+// TODO: I am not sure that it is possible to type triggers such that we have some gigantic union
+// of Trigger<TetherParams> | Trigger<AbilityParams> | Trigger<AddedCombatantParams> | etc etc
+// and so the idea is that individual files would use a narrower type for Matches if they wanted to
+// and that's something we (maybe?) could enforce via tests rather than via TypeScript.  This
+// approach would not preclude some TypeScript gigabrain solution that can figure out how to do
+// the same thing at the TypeScript level.
+//
+// TODO: I think there are two types of TriggerSets.  There is the "cactbot + user defined triggers"
+// and also the "cactbot built-in only" triggers, which can be more strict.  For example,
+// * is `id` required, and `regex` for timeline triggers? (yes for built-in, no for user triggers)
+// * should we handle nonsense like `condition` returning non-booleans?
+// We could have one type that built-in triggers define themselves as more strictly and then a
+// second type that popup-text.js uses and needs to do more narrowing for.
+export type BaseTrigger = {
+  id?: string;
+  disabled?: boolean;
+  condition?: TriggerField<boolean | undefined>;
+  preRun?: TriggerField<void>;
+  delaySeconds?: TriggerField<number | undefined>;
+  durationSeconds?: TriggerField<number | undefined>;
+  suppressSeconds?: TriggerField<number | undefined>;
+  promise?: TriggerField<Promise<void> | undefined>;
+  sound?: TriggerField<string | undefined>;
+  soundVolume?: TriggerField<number | undefined>;
+  response?: ResponseField;
+  alarmText?: TriggerField<TriggerOutput>;
+  alertText?: TriggerField<TriggerOutput>;
+  infoText?: TriggerField<TriggerOutput>;
+  tts?: TriggerField<TriggerOutput>;
+  run?: TriggerField<void>;
+  outputStrings?: {
+    [key: string]: LocaleText;
+  };
+}
+
+export type NetRegexTrigger = BaseTrigger & {
+  netRegex: RegExp;
+  netRegexDe?: RegExp;
+  netRegexFr?: RegExp;
+  netRegexJa?: RegExp;
+  netRegexCn?: RegExp;
+  netRegexKo?: RegExp;
+}
+
+export type RegexTrigger = BaseTrigger & {
+  regex: RegExp;
+  regexDe?: RegExp;
+  regexFr?: RegExp;
+  regexJa?: RegExp;
+  regexCn?: RegExp;
+  regexKo?: RegExp;
+}
+
+export type TimelineTrigger = BaseTrigger & {
+  regex?: RegExp;
+  // TODO: can this also be a function?
+  beforeSeconds?: number;
+};
+
+export type Trigger = RegexTrigger | NetRegexTrigger;
+
+export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
+
+export type TriggerSet = {
+  zoneId?: number;
+  zoneRegex?: RegExp | { [lang in Lang]?: RegExp };
+  resetWhenOutOfCombat?: boolean;
+  overrideTimelineFile?: boolean;
+  timelineFile?: string;
+  timeline?: TimelineFunc;
+  triggers?: Trigger[];
+  timelineTriggers?: TimelineTrigger[];
+  timelineReplace?: {
+    locale: Lang;
+    missingTranslations?: boolean;
+    replaceText?: { [regex: string]: string };
+    replaceSync?: { [regex: string]: string };
+  }[];
 }
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -131,7 +131,7 @@ export type Trigger = RegexTrigger | NetRegexTrigger;
 export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
 
 export type TriggerSet = {
-  zoneId?: number;
+  zoneId?: ZoneId | ZoneId[];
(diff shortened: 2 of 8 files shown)


ui/raidboss/data/02-arr/trial/titan-ex.ts
```diff
@@ -67,13 +70,17 @@ export default {
...
+      // or possibly even just pre-define them all into MatchesTether or something.
+      alertText: (data: Data, matches: Matches<NetRegex<TetherParams>>,
+          output: Output): string | undefined => {
+        if (matches?.source === data.me || matches?.target === data.me)
+          return output.jailOnYou?.();
       },
-      infoText: (data, matches, output) => {
-        if (matches.source !== data.me && matches.target !== data.me)
-          return output.jails();
+      infoText: (data: Data, matches: Matches<NetRegex<TetherParams>>,
```
quisquous, line 80: One interesting thing is that this function (only) made me specify the output type of the function and also the types of Data/Matches/Output.  However, other functions (above) did not.

types/trigger.d.ts
```diff
@@ -43,34 +46,110 @@ export type Output = {
...
+  // TODO: can this also be a function?
+  beforeSeconds?: number;
+};
+
+export type Trigger = RegexTrigger | NetRegexTrigger;
+
+export type TimelineFunc = string | string[] | ((data: Data) => TimelineFunc);
+
+export type TriggerSet = {
+  zoneId?: number;
```
Trim21, line 141: also `number[]`

types/trigger.d.ts
```diff
@@ -43,34 +46,110 @@ export type Output = {
...
+  regex?: RegExp;
+  // TODO: can this also be a function?
+  beforeSeconds?: number;
+};
+
+export type Trigger = RegexTrigger | NetRegexTrigger;
+
+export type TimelineFunc = string | string[] | ((data: Data) => TimelineFunc);
+
+export type TriggerSet = {
```
Trim21, line 140: This should be a genetic type I think. We need to extend `Data` for a specific trigger set.

Trim21:
I wrote something like this:

```TS
type TriggerFunction<T, D = Data, R = T | string | undefined> = (data: D, matches?: Matches, output?: Output) => R;

interface TimeLineTrigger<D> extends BaseTrigger<D> {
  beforeSeconds?: number | TriggerFunction<number, D>;
}

interface _TriggerSet<D extends Data> {
  zoneId: number | null | number[];
  timelineFile?: string;
  timeline?: Array<string | ((d: D) => string)>;
  timelineTriggers?: TimeLineTrigger<D>[];
  locale?: Lang;
  timelineReplace?: TimelineReplace[];
  triggers?: LogTrigger<D>[];
  overrideTimelineFile?: boolean;
  resetWhenOutOfCombat?: boolean;
}

export type TriggerSet<D = Data> = _TriggerSet<D>

```

in trigger file
```TS
import { TEAData } from "./data";
import { TriggerSet } from "../../cactbot";

const trigger: TriggerSet<TEAData> = {}
export default trigger;
```

types/trigger.d.ts
quisquous: I am not sure that's true.  Because `data` is reset to the base `data`, then every field in `TEAData` must be optional.  And, because all fields are optional, you can do what I did in `t1.ts`, where I have `ThisData extends Data` and use it locally inside of that function in place of `Data`.  In other words, I can imagine each trigger file having its own type that extends the base data type and using that as a parameter for Data locally.

Trim21: code outside the trigger file itself didn't need to know the extend type, they just need to know it's a `TriggerSet<Data>`.

Trim21: Oh sorry, I import `TEAData` because I have multiple trigger files for TEA, not because I'm using this type outside the trigger files.

quisquous: Oh oh oh, I see!  I get it.  Thanks, that's a good idea!

types/trigger.d.ts
```diff
@@ -43,34 +46,110 @@ export type Output = {
...
+  zoneId?: ZoneId | ZoneId[];
```
Trim21, line 141: this becomes `number | null | Array<number | null>`

quisquous: I think that's technically correct? `zoneId` is either a single `ZoneId` or an array of `ZoneIds`.

I am not sure what you are suggesting.  Do you think it should be `number | null | number[]`?

Trim21: Yes, I think it should be `number | null | number[]`. Array with null as an element is unnecessary.

quisquous: Done, thanks!

types/trigger.d.ts
quisquous: Done in the latest patch, what a great idea!

ui/raidboss/data/02-arr/trial/titan-ex.ts
quisquous: @Trim21 Do you understand why I have to specify types here but not on other similar functions?

ui/raidboss/timeline.ts
```diff
@@ -456,7 +460,7 @@ export class Timeline {
         }
       }
       if (!found) {
-        const text = `No match for timeline trigger ${trigger.regex.source} in ${trigger.id}`;
+        const text = `No match for timeline trigger ${trigger.regex?.source ?? ''} in ${trigger?.id ?? ''}`;
```
panicstevenson, line 463: I can understand why trigger.regex is optional here, but why trigger itself?

ui/raidboss/data/02-arr/raid/t1.ts
```diff
@@ -37,7 +44,7 @@ export default {
       netRegexCn: NetRegexes.ability({ source: '神杖巨蛇', id: '4BA' }),
       netRegexKo: NetRegexes.ability({ source: '카두케우스', id: '4BA' }),
       condition: Conditions.targetIsYou(),
-      alertText: (data, _, output) => output.text(),
+      alertText: (data: Data, _: unknown, output: Output) => output.text?.(),
```
panicstevenson: I wonder if, before going forward, it makes sense to make the change these `data, _, output` definitions to be `_data, _matches, output`, etc. now that we've added the rule in https://github.com/quisquous/cactbot/pull/2790

quisquous: Yeah, that seems like a reasonable change!

ui/raidboss/data/02-arr/trial/titan-ex.ts
quisquous: Ok, fixed this by removing the `@typescript-eslint/explicit-module-boundary-types` rule for raidboss triggers, which I think makes sense.

ui/raidboss/data/02-arr/raid/t1.ts
quisquous: Thanks for fixing it!

.eslintrc.cjs
```diff
@@ -66,6 +66,8 @@ module.exports = {
         'rulesdir/cactbot-output-strings': 'error',
         'rulesdir/cactbot-timeline-triggers': 'error',
         'rulesdir/cactbot-response-default-severities': 'error',
+        // Raidboss data files always export a trigger set, and explicit types are noisy.
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
```
Trim21, line 70: there is a override option to disable this for trigger files only.

quisquous: I think that's what this is, unless I'm misunderstanding you.

Trim21: I mean we can add a new override rules to disable this for trigger files only instead of disable it at all.

quisquous: Sorry, that's what I'm trying to say.  Isn't this in an override rule for trigger files only already?

Trim21: Oh, sorry 🤣 there isn't.

types/trigger.d.ts
```diff
@@ -43,34 +46,118 @@ export type Output = {
...
+  regex: RegExp;
+  regexDe?: RegExp;
+  regexFr?: RegExp;
+  regexJa?: RegExp;
+  regexCn?: RegExp;
+  regexKo?: RegExp;
+}
+
+export type TimelineTrigger<Data> = BaseTrigger<Data> & {
+  regex?: RegExp;
```
Trim21, line 137: how will happen if `TimelineTrigger` doesn't have a `regex` property? 
will it match all timeline record?

ui/raidboss/data/02-arr/raid/t1.ts
```diff
@@ -37,7 +44,7 @@ export default {
       netRegexCn: NetRegexes.ability({ source: '神杖巨蛇', id: '4BA' }),
       netRegexKo: NetRegexes.ability({ source: '카두케우스', id: '4BA' }),
       condition: Conditions.targetIsYou(),
-      alertText: (_data, _matches, output) => output.text(),
+      alertText: (_data, _matches, output) => output.text?.(),
```
Trim21, line 47: No need to do this, there is an eslint option

```js
"@typescript-eslint/no-unused-vars": ["error", {
          "args": "after-used",
        }]
```

types/trigger.d.ts
MaikoTan: Then the current trigger would be ignored.
I think this should be a required field?

ui/raidboss/data/02-arr/raid/t1.ts
```diff
@@ -101,7 +108,7 @@ export default {
       netRegexKo: NetRegexes.message({ line: '알라그 유적 will be sealed off.*?', capture: false }),
       delaySeconds: 35,
       suppressSeconds: 5,
-      infoText: (_data, _matches, output) => output.text(),
+      infoText: (_data, _matches, output) => output.text?.(),
```
Trim21, line 111: I think we should disable this null check for trigger files.

ui/raidboss/data/02-arr/raid/t1.ts
```diff
@@ -1,9 +1,16 @@
+import { RaidbossData } from '../../../../../types/data';
+import { TriggerSet } from '../../../../../types/trigger';
+
 import Conditions from '../../../../../resources/conditions';
 import NetRegexes from '../../../../../resources/netregexes';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 
-export default {
+interface Data extends RaidbossData {
```
Trim21, line 9: It would be better to expect built-in raid data type so user using webpack to build their raidboss.js could import the

quisquous: I suspect that you are the only person writing personal triggers and using webpack.  :P

Just to clarify, when you said "expect built-in data type", did you mean "export"?  Are you suggesting to `export interface Data`?  If so, I don't see any reason not to.

ui/raidboss/data/02-arr/raid/t1.ts
quisquous: If we could disable the undefined check just for this one object, I think that would be good, but I don't want to do it in general.  I think the benefit of the undefined check is better than the irritating at having to write this.  The only other alternative is `output.text!()`, which isn't much better.

I tried a lot of different things, but I don't think there's any way to say that `output.text` is always defined.  Unfortunately, `Proxy` does not seem like a first-class TypeScript object, and so it doesn't know that it can look at the `get` function to understand what types can be returned.

ui/raidboss/data/02-arr/raid/t1.ts
Trim21: yes, that's a typo, I mean `interface Data extends RaidbossData`

In fact, I copy most of type definition from @MaikoTan

ui/raidboss/data/02-arr/raid/t1.ts
Trim21: Looks like currently in TypeScript `Proxy` just return a type as same as it's target

types/trigger.d.ts
quisquous: Ok, I separated out the types into "loose" types (aka things that might be from users, where we don't want to assume that fields are set, meant to be used in like timeline/popup-text for processing triggers) and into "strict" types (where more things are required, deprecated fields are removed, and meant to be used in built-in triggers).

ui/raidboss/data/02-arr/raid/t1.ts
quisquous: Unfortunately, if you make `Output` be `any`, then any call on that `any` value like `text()` is also an "unsafe call on an any value" error.  So, I'm not sure that's the solution.

Trim21:
find that it's hard to write trigger in the old way with ` "strictNullChecks": true,`

Trim21:
And I think we should remove this type annotation  because it makes all method on `Responses` optional.

https://github.com/quisquous/cactbot/blob/cc1bda93735d2f80ff3f4de83a05a4e832fe37d5/resources/responses.ts#L148-L155

quisquous:
> And I think we should remove this type annotation because it makes all methods on `Responses` optional like `Output`.
> 
> https://github.com/quisquous/cactbot/blob/cc1bda93735d2f80ff3f4de83a05a4e832fe37d5/resources/responses.ts#L148-L155

I am not sure which commit you are commenting on, but that is fixed on the main branch, and rebased here.  Take a look and let me know if you think there is more to do.

Trim21:
> I am not sure which commit you are commenting on, but that is fixed on the main branch, and rebased here. Take a look and let me know if you think there is more to do.

sorry, my HEAD is at the latest tag, didn't contain that commit.

quisquous:
> find that it's hard to write trigger in the old way with ` "strictNullChecks": true,`

Do you have an example of something that gets harder?  Is it more than just `output.text?.()`?

Trim21:
> > find that it's hard to write trigger in the old way with ` "strictNullChecks": true,`
> 
> Do you have an example of something that gets harder? Is it more than just `output.text?.()`?

got many problems with null checking in properties of `matches`

try this file: https://gist.github.com/Trim21/8f1fac30be1928d7b2b33046bfd0975e

Trim21:
could we add a method on output named render or something else with type `(infi: string, params: Record): string` to avoid this type problem?

quisquous:
> could we add a method on output named render or something else with type `(info: string, params: Record): string` to avoid this type problem?

Personally, I think `output.render(info: 'text', params: { player: target })` is harder to type and read than `output.text?.({ player: target })`.  Is it really so bad to type `output.text?.()` instead of `output.text()`?

quisquous:
```diff
diff --git a/ui/raidboss/data/05-shb/ultimate/tea.ts b/ui/raidboss/data/05-shb/ultimate/tea.ts
index 2a1de597..f842c56b 100644
--- a/ui/raidboss/data/05-shb/ultimate/tea.ts
+++ b/ui/raidboss/data/05-shb/ultimate/tea.ts
@@ -29,16 +29,20 @@ export interface TEAData extends RaidbossData {
 
 
 const getHeadmarkerId = (data: TEAData, matches: MatchesHeadMarker): string => {
+  const id = matches?.id;
+  if (!id)
+    return '0000';
+
   // If we naively just check !data.decOffset and leave it, it breaks if the first marker is 004F.
   // (This makes the offset 0, and !0 is true.)
   if (typeof data.decOffset === 'undefined') {
     // The first 1B marker in the encounter is Limit Cut 1, ID 004F.
-    data.decOffset = parseInt(matches.id, 16) - 79;
+    data.decOffset = parseInt(id, 16) - 79;
   }
   // The leading zeroes are stripped when converting back to string, so we re-add them here.
   // Fortunately, we don't have to worry about whether or not this is robust,
   // since we know all the IDs that will be present in the encounter.
-  return '00' + (parseInt(matches.id, 16) - data.decOffset).toString(16).toUpperCase();
+  return '00' + (parseInt(id, 16) - data.decOffset).toString(16).toUpperCase();
 };
 
 const limitCut = {
@@ -61,10 +65,8 @@ export const trigger: TriggerSet<TEAData> = ({
       beforeSeconds: 3,
       durationSeconds: 4,
       alertText(data) {
-        if ([5, 6].includes(data.limitCutNumber))
+        if ([5, 6].includes(data.limitCutNumber ?? 0))
           return '踩灵泉';
-
-        return '';
       },
     },
     {
@@ -73,10 +75,8 @@ export const trigger: TriggerSet<TEAData> = ({
       beforeSeconds: 3,
       durationSeconds: 4,
       alertText(data) {
-        if ([7, 8].includes(data.limitCutNumber))
+        if ([7, 8].includes(data.limitCutNumber ?? 0))
           return '踩灵泉';
-
-        return '';
       },
     },
     {
@@ -85,10 +85,8 @@ export const trigger: TriggerSet<TEAData> = ({
       beforeSeconds: 3,
       durationSeconds: 4,
       alertText(data) {
-        if ([1, 2].includes(data.limitCutNumber))
+        if ([1, 2].includes(data.limitCutNumber ?? 0))
           return '踩灵泉';
-
-        return '';
       },
     },
   ],
@@ -116,7 +114,7 @@ export const trigger: TriggerSet<TEAData> = ({
         // Here and elsewhere, it's probably best to check for whether the user is the target first,
         // as that should short-circuit more often.
         return (
-          data.me === matches.target &&
+          data.me === matches?.target &&
           /00(?:4F|5[0-6])/.test(getHeadmarkerId(data, matches)) &&
           data.phase !== 'wormhole'
         );
@@ -158,7 +156,7 @@ export const trigger: TriggerSet<TEAData> = ({
         // Here and elsewhere, it's probably best to check for whether the user is the target first,
         // as that should short-circuit more often.
         return (
-          data.me === matches.target &&
+          data.me === matches?.target &&
           /00(?:4F|5[0-6])/.test(getHeadmarkerId(data, matches)) &&
           data.phase === 'wormhole'
         );
@@ -194,7 +192,7 @@ export const trigger: TriggerSet<TEAData> = ({
       // return data.limitCutDelay
       // },
       infoText: function(data, _, output) {
-        return output[`num${data.limitCutNumber}`]();
+        return output[`num${data.limitCutNumber ?? ''}`]();
       },
       outputStrings: limitCut,
     },
```

I don't think this is that bad, to be honest.  I think the `[5, 6].includes(etc)` constructs is rough for TypeScript because TypeScript strongly types the includes function based on the type of the Array.

I am very reluctant to drop the strict null check, because one of the things I really wanted out of typescript was to make sure that we were initializing `data` properties and not using them before they were assigned.  I think needing to add a `?` to `matches` or `output` usage is a small price to pay for that.

quisquous:
I think this is almost ready to go.  I think the one thing that feels like a blocking issue for me is that using conditional chaining causes really obnoxious looking transpiled javascript.  If users are looking at the [triggers branch](https://github.com/quisquous/cactbot/tree/triggers) to copy and paste JavaScript for user files, then this will look really bad.  It seems unlikely that OverlayPlugin will update CEF any time soon (I filed https://github.com/ngld/OverlayPlugin/issues/209.  Maybe I should just look into doing it myself??).

So, I think for me that means we need to pick one of:
1) disable the eslint error against `!` in triggers only and do something like `matches!.text` and `output.text!()`
2) disable null checks in user files (but I feel that strict null check on `data` is important, as we've had a bunch of bugs there over time? maybe everybody else disagrees with this?)
3) accept that transpiled triggers with optional chaining will be pretty unreadable and just live with that

My feeling here is kind of to use option 1 and then if OverlayPlugin CEF version bumps we can just search and replace `!` to `?.` as needed.


Other ideas:

I see Maiko has a PR up for making the Matches have required fields in #2994, but I don't understand how the callers will be typed in order to make that work.  Maybe I'm missing something.

We *could* write an output type that has every outputStrings field name in it.  We could generate it via some automated script that then spits out some `Output` type with every valid field name in it, and then we would never need to write `output.text?.()` or `output.text!()` and we could just write `output.text()`.  This is kind of weird and maybe obnoxious, but it's a thought that I had.

Trim21:
> > could we add a method on output named render or something else with type `(info: string, params: Record): string` to avoid this type problem?
> 
> Personally, I think `output.render(info: 'text', params: { player: target })` is harder to type and read than `output.text?.({ player: target })`. Is it really so bad to type `output.text?.()` instead of `output.text()`?

It's not about typing, but reanspiled null checker.

quisquous:
> > Personally, I think `output.render(info: 'text', params: { player: target })` is harder to type and read than `output.text?.({ player: target })`. Is it really so bad to type `output.text?.()` instead of `output.text()`?
> 
> It's not about typing, but reanspiled null checker. It's a litt bit annoying.
> 
> I'd rather choose `!`

I added a `!` for output in the last patch.  I also took Maiko's idea from another patch and made `Matches` always non-null.  Its fields may still be undefined, but `matches` itself is never undefined.  I think that cleans things up a lot.

quisquous:
@Trim21 @MaikoTan @panicstevenson @valarnin I think this is ready to go at this point after a lot of discussion with all of you (thank you!!).  Does anybody have any other thoughts or concerns here?  I wanted to just make sure we were in a good place before more triggers started being converted.

valarnin:
I hadn't provided any feedback on this I thought, but I'll gladly weigh in with my quick thoughts.

> I suspect that you are the only person writing personal triggers and using webpack. :P

I also write personal triggers and use webpack ;)

My only real question here is the move of `export default` in trigger files to the end, as:

```typescript
export default triggerSet;
```

Was there no way to make this work as it did originally? For example:

```typescript
export default {
// ...
} as TriggerSet<Data>;
```

quisquous:
> I hadn't provided any feedback on this I thought, but I'll gladly weigh in with my quick thoughts.
> 
> > I suspect that you are the only person writing personal triggers and using webpack. :P
> 
> I also write personal triggers and use webpack ;)

Noted!

> Was there no way to make this work as it did originally? For example:
> 
> ```ts
> export default {
> // ...
> } as TriggerSet<Data>;
> ```

Unfortunately, using `as` here makes this mean something very different.  It just asserts that the file is exporting a `TriggerSet<Data>` without checking that's true.  It's similar to saying `export default {} as string;`

If there's some way to declare and export a variable with an explicit type in a single line, I'd be happy to change it.  It feels awkward to me too.

Trim21:
There isn't as far as I know.... We have to do it in this way.

or `export default helper<Data>({})` which make it more awkward.

valarnin:
Hmm, unfortunate then.

But yeah, otherwise everything here looks good.

panicstevenson:
I converted a few different trigger files: Ifrit NM; Binding Coil Turn 8; and (started) Unending Coil.

Things I noticed:
```ts
export { RaidbossData as Data } from '../../../../../types/data';
```
In Titan Extreme originally confused me; it made more sense using Turn 1 as an example here.

`Regex.hasHP` in UCoB returns `'regex' does not exist in type 'NetRegexTrigger<Data>'`

`output.xyz({ num: num })` will throw that num isn't a string; this can be solved with `String(num)` but that feels maybe obnoxious to put everywhere.

It would be great if there were a way to guarantee things existing within `data`, otherwise you can get into a situation where condition, infoText, alarmText, and run all check if `data.something !== undefined`

I don't think any of these are blocking to me; otherwise it felt good. I don't have any personal triggers outside of my user file to compare that aspect, but the normal trigger conversion seemed straightforward for the most part.

quisquous:
> I converted a few different trigger files: Ifrit NM; Binding Coil Turn 8; and (started) Unending Coil.
> 
> Things I noticed:
> 
> ```ts
> export { RaidbossData as Data } from '../../../../../types/data';
> ```
> 
> In Titan Extreme originally confused me; it made more sense using Turn 1 as an example here.

It made some sense to me to have every trigger also export its `Data` type in case we needed it? We can not do that too.

> `Regex.hasHP` in UCoB returns `'regex' does not exist in type 'NetRegexTrigger<Data>'`

Oof, welllll, I'll defer this for a later PR.

I've wondered for a while if we should find some other way to handle `hasHP` anyway.  A lot of people turn off this setting too (especially because it's very noisy in logs).

> `output.xyz({ num: num })` will throw that num isn't a string; this can be solved with `String(num)` but that feels maybe obnoxious to put everywhere.

Alternatively ``output.xyz({ num: `${num}`)`` Alternatively alternatively we could allow params to be numbers as well and turn them into strings internally?

> It would be great if there were a way to guarantee things existing within `data`, otherwise you can get into a situation where condition, infoText, alarmText, and run all check if `data.something !== undefined`

Yeah, this is definitely unfortunate.  One thing that can help some is that probably things with info+alarm could be responses, which cuts down a little bit on this.  But, that doesn't really solve it either.

One future feature that we could consider doing is having some function on triggerSet to initialize data to some known values when you start the zone or on reset.  Then the `Data` object could have required fields on it from the perspective of the trigger set.

panicstevenson:
> Alternatively `` output.xyz({ num: `${num}`) ``

Ah, yea. I originally started with that and it didn't work, but I think that's due to a lack of converting the rest of the file? Now that a lot more of the types are defined this works as expected. Otherwise, it will complain that you can't use `number | undefined` in a template string, but that seems like a non-issue now.

valarnin:
> I've wondered for a while if we should find some other way to handle `hasHP` anyway. A lot of people turn off this setting too (especially because it's very noisy in logs).

I think this option isn't on by default, IIRC? Either way, this could key off of something like autoattack and just do the math on current vs max hp, since those should be included in those lines? (specifically I'm talking about `21|` lines).

> One future feature that we could consider doing is having some function on triggerSet to initialize data to some known values when you start the zone or on reset. Then the `Data` object could have required fields on it from the perspective of the trigger set.

This is actually something that could be pretty useful, e.g. on e12s for `phase`.

quisquous:
🏁

PR #2882, opened 2021-05-22
Raidboss: notify as tankbuster from Nixie for tank

MaikoTan:
The target of the skill `Crash-Smash` is Nixie itself, so change to `HeadMarker` version.

I just write down my opinion on `de`, `fr` and `ko`, might not the correct ones.

### DIFF
--- a/ui/raidboss/data/05-shb/dungeon/matoyas_relict.js
+++ b/ui/raidboss/data/05-shb/dungeon/matoyas_relict.js
@@ -1,5 +1,6 @@
 import Conditions from '../../../../../resources/conditions';
 import NetRegexes from '../../../../../resources/netregexes';
+import Outputs from '../../../../../resources/outputs';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/05-shb/dungeon/matoyas_relict.js
```diff
@@ -62,22 +63,21 @@ export default {
...
-          fr: 'Évitez le tank et les liens',
-          ja: 'タンクと線から離れる',
-          cn: '远离坦克及其连线',
-          ko: '탱커와 선 피하기',
+          en: 'Avoid ${player} and tethers',
+          de: 'Weiche den ${player}-Verbindungen aus', // FIXME
+          fr: 'Évitez ${player} et les liens', // FIXME
+          ja: '${player}と線から離れる',
+          cn: '远离${player}及其连线',
+          ko: '${player}와 선 피하기', // FIXME
```
quisquous, line 80:
```suggestion
          ja: '${player}と線から離れる',
          cn: '远离${player}及其连线',
```

I think in cases where you're not sure, it's maybe better to just remove it and have somebody re-translate it.

MaikoTan: Okay!

PR #2883, opened 2021-05-22
resources: export Responses as const

quisquous:


resources/responses.ts
```diff
@@ -462,3 +462,10 @@ export const Responses: ResponsesMap = {
   },
   wakeUp: (sev?: Severity) => staticResponse(defaultAlarmText(sev), Outputs.wakeUp),
 } as const;
+
+// Don't give `Responses` a type in its declaration so that it can be treated as more strict
+// than `ResponsesMap`, but do assert that its type is correct.  This allows callers to know
+// which properties are defined in Responses without having to conditionally check for undefined.
+const responseMapTypeAssertion: ResponsesMap = Responses;
+// Suppress unused variable warning.
+console.assert(responseMapTypeAssertion);
```
panicstevenson, line 471: Can we instead just suppress the warning?

panicstevenson (approved):

resources/responses.ts
quisquous: If it were just an eslint error, then yes.  I think I am averse to disabling ts errors because you can only disable all of them or none of them.  So, if there were some other ts error, we wouldn't see it.  It's the same reason I used assert it `user_config.ts`.

panicstevenson: Seems fair; I didn't realize it was so gross to disable these. For reference:
```ts
// Don't give `Responses` a type in its declaration so that it can be treated as more strict
// than `ResponsesMap`, but do assert that its type is correct.  This allows callers to know
// which properties are defined in Responses without having to conditionally check for undefined.
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
// eslint-disable-next-line @typescript-eslint/no-unused-vars
const responseMapTypeAssertion: ResponsesMap = Responses;
```

quisquous: Yeah, I think there are other options here, but maybe I'll leave this as an assert to match `user_config` and if we come up with something better a followup can fix both of them.

PR #2897, opened 2021-05-22
lint: Enable no-unused-vars for Raidboss Data

panicstevenson:


Trim21:
Why don't we use `"args": "after-used",` to aviod `_` prefix?

panicstevenson:
We're using the _ prefix to know if the capture groups are needed for the trigger tests: https://github.com/quisquous/cactbot/pull/2891 and then using eslint to actually confirm

Trim21:
thanks

PR #2899, opened 2021-05-22
raidboss: Apply ESLint prefer-arrow/prefer-arrow-functions to 00-misc

quisquous:
Converted via:
find ui/raidboss/data/00-misc/ -type f | xargs -i sed -i 's/function(\(.*\)) {/(\1) => {/' {}


quisquous:
I will also add that on this (and others), if there is a `this` being used in the function body of an arrow function, eslint will yell about it.  That's what led to #2896.  So, I think this is a relatively safe conversion.

panicstevenson (approved):

PR #2902, opened 2021-05-22
raidboss: Fix Non-Outputs stack Outputs

panicstevenson:


panicstevenson:
There's more of these, but they all vary slightly across various languages.

MaikoTan:
Yeah, maybe we should differentiate "Stack for damage share", "Stack for bait AoE" and "Stack for other things", 'cause these are different sentences in `ja` and `cn` (also `ko` I think?).

panicstevenson:
Is that the same issue here then: https://github.com/quisquous/cactbot/pull/2903/files ? Should we revert that change?

MaikoTan:
Well, looks like those text in #2903 are all stack for damage sharing?
I think it is OK though.

panicstevenson:
Okay, `自分にスタック` and `自分に頭割り` are similar enough for that context? Sorry to be annoying, I just want to make sure I didn't make issues for anyone.

MaikoTan:
Well, technically nope.
`スタック` just means "Stack", and `頭割り` means "Sharing damage".

PR #2910, opened 2021-05-23
raidboss: Use Implicit Returns for One-Line Arrow Functions

panicstevenson:


ui/raidboss/data/04-sb/raid/o8s.js
```diff
@@ -513,9 +507,7 @@ export default {
       netRegexJa: NetRegexes.startsUsing({ id: '28CE', source: 'ケフカ', capture: false }),
       netRegexCn: NetRegexes.startsUsing({ id: '28CE', source: '凯夫卡', capture: false }),
       netRegexKo: NetRegexes.startsUsing({ id: '28CE', source: '케프카', capture: false }),
-      condition: (data) => {
-        return (data.antics && data.fireMarker === 'spread') || (data.truths && data.fireMarker === 'stack');
-      },
+      condition: (data) => (data.antics && data.fireMarker === 'spread') || (data.truths && data.fireMarker === 'stack'),
```
quisquous, line 510: Agreed that this fits on one line, but this is one heck of a line.  Do you think this would be worth breaking out into:
```javascript
condition: (data) => {
  const isFakeSpread = data.antics && data.fireMarker === 'spread';
  const isTrueStack = data.truths && data.fireMarker === 'stack';
  return isFakeSpread || isTrueStack;
},
```

panicstevenson: Yea, I can appreciate that. I did all of this with a regex find + replace then ran eslint to find anything over the max length. Do you think I should run a search for anything that has more than X logical operators?

panicstevenson: `=> (.*(\|\||&&)){3}` only returns these from O8S, setting it to {2} will catch things like "tank || healer || blu"

quisquous: I think some of it is the mixed operators that makes this harder to read? If it's just x || y || z, that's a bit more straightforward.

PR #2917, opened 2021-05-23
raidemulator: Follow up fixes for #2879

valarnin:
Follow up PR for fixes mentioned in #2879.

Opted to move `LineEvent0x18`'s `effectName` property initialization to the constructor instead of the `convert` method, since it doesn't need to be delayed for repo population like some other properties.


valarnin:
Added the missing comment as well. I think that's everything outstanding.

quisquous:
Thanks!

PR #2919, opened 2021-05-23
raidemulator: Convert LogEventHandler and NetworkLogConverter to typescript

valarnin:
`NetworkLogConverterWorker` will need to wait until the `CombatantTracker` class and dependencies are converted.

This PR should be considered downstream from #2917 due to `zoneID` => `zoneId` property rename. As such, this PR's compile will fail until that PR is merged. When that one's merged I'll rebase this to upstream/main.


valarnin:
```diff
diff --git a/ui/raidboss/emulator/data/LogEventHandler.js b/ui/raidboss/emulator/data/LogEventHandler.ts
index 7868f946..79c1d2c3 100644
--- a/ui/raidboss/emulator/data/LogEventHandler.js
+++ b/ui/raidboss/emulator/data/LogEventHandler.ts
@@ -1,13 +1,16 @@
 import EmulatorCommon from '../EmulatorCommon';
 import EventBus from '../EventBus';
 import { LocaleNetRegex } from '../../../../resources/translations';
+import { Lang } from 'types/global';
+import LineEvent from './network_log_converter/LineEvent';
+import { LineEvent0x01 } from './network_log_converter/LineEvent0x01';
 
 export default class LogEventHandler extends EventBus {
-  static doesLineMatch(line, regexes) {
+  static doesLineMatch(line: string, regexes: Record<string, RegExp>): RegExpExecArray | false {
     for (const i in regexes) {
-      const res = regexes[i].exec(line);
+      const res = regexes[i]?.exec(line);
       if (res) {
-        if (LocaleNetRegex.areaSeal[i])
+        if (LocaleNetRegex.areaSeal[i as Lang] && res.groups)
           res.groups.language = i;
         return res;
       }
@@ -15,96 +18,101 @@ export default class LogEventHandler extends EventBus {
     return false;
   }
 
-  static isMatchStart(line) {
+  static isMatchStart(line: string): RegExpMatchArray | false {
     let res;
     res = LogEventHandler.doesLineMatch(line, EmulatorCommon.countdownRegexes);
     if (res) {
-      res.groups.StartIn = res.groups.time * 1000;
+      res.groups = res.groups ?? {};
+      res.groups.StartIn = (parseInt(res.groups.time ?? '0') * 1000).toString();
       res.groups.StartType = 'Countdown';
       return res;
     }
     res = LogEventHandler.doesLineMatch(line, EmulatorCommon.sealRegexes);
     if (res) {
-      res.groups.StartIn = 0;
+      res.groups = res.groups ?? {};
+      res.groups.StartIn = '0';
       res.groups.StartType = 'Seal';
       return res;
     }
     res = LogEventHandler.doesLineMatch(line, EmulatorCommon.engageRegexes);
     if (res) {
-      res.groups.StartIn = 0;
+      res.groups = res.groups ?? {};
+      res.groups.StartIn = '0';
       res.groups.StartType = 'Engage';
       return res;
     }
     return false;
   }
 
-  static isMatchEnd(line) {
+  static isMatchEnd(line: string): RegExpMatchArray | false {
     let res;
-    res = LogEventHandler.doesLineMatch(line, [EmulatorCommon.winRegex]);
+    res = LogEventHandler.doesLineMatch(line, { '': EmulatorCommon.winRegex });
     if (res) {
+      res.groups = res.groups ?? {};
       res.groups.EndType = 'Win';
       return res;
     }
-    res = LogEventHandler.doesLineMatch(line, [EmulatorCommon.wipeRegex]);
+    res = LogEventHandler.doesLineMatch(line, { '': EmulatorCommon.wipeRegex });
     if (res) {
+      res.groups = res.groups ?? {};
       res.groups.EndType = 'Wipe';
       return res;
     }
-    res = LogEventHandler.doesLineMatch(line, [EmulatorCommon.cactbotWipeRegex]);
+    res = LogEventHandler.doesLineMatch(line, { '': EmulatorCommon.cactbotWipeRegex });
     if (res) {
+      res.groups = res.groups ?? {};
       res.groups.EndType = 'Cactbot Wipe';
       return res;
     }
     res = LogEventHandler.doesLineMatch(line, EmulatorCommon.unsealRegexes);
     if (res) {
+      res.groups = res.groups ?? {};
       res.groups.EndType = 'Unseal';
       return res;
     }
     return false;
   }
 
-  constructor() {
-    super();
+  public currentFight: LineEvent[] = [];
+  public currentZoneName?: string;
+  public currentZoneId = -1;
 
-    this.currentZoneName = null;
-    this.currentFight = [];
-    this.currentZoneId = -1;
-  }
-
-  parseLogs(logs) {
+  parseLogs(logs: LineEvent[]): void {
     for (let i = 0; i < logs.length; ++i) {
-      const lineObj = logs[i];
+      const lineObj = logs[i] as LineEvent;
 
       this.currentFight.push(lineObj);
 
-      lineObj.offset = lineObj.timestamp - this.currentFight[0].timestamp;
+      lineObj.offset = lineObj.timestamp - (this.currentFight[0]?.timestamp ?? 0);
 
       const res = LogEventHandler.isMatchEnd(lineObj.networkLine);
       if (res) {
         this.endFight();
-      } else if (lineObj.zoneName) {
-        this.currentZoneId = lineObj.zoneId;
-        this.currentZoneName = lineObj.zoneName;
+      } else if (lineObj instanceof LineEvent0x01) {
+        const castLineObj = lineObj;
+        this.currentZoneId = parseInt(castLineObj.zoneId ?? '0');
+        this.currentZoneName = castLineObj.zoneName;
         this.endFight();
       }
     }
   }
 
-  endFight() {
+  endFight(): void {
     if (this.currentFight.length < 2)
       return;
 
-    const start = new Date(this.currentFight[0].timestamp);
+    const start = new Date(this.currentFight[0]?.timestamp ?? 0).toISOString();
+    const end = new Date(this.currentFight.slice(-1)[0]?.timestamp ?? 0).toISOString();
     this.currentZoneName = this.currentZoneName || 'Unknown';
     this.currentZoneId = this.currentZoneId || -1;
 
     console.debug(`Dispatching new fight
 Start: ${start}
-End: ${new Date(this.currentFight[this.currentFight.length - 1].timestamp)}
+End: ${end}
 Zone: ${this.currentZoneName}
 Line Count: ${this.currentFight.length}
 `);
-    this.dispatch('fight', start.toISOString().substr(0, 10), this.currentZoneId, this.currentZoneName, this.currentFight);
+    void this.dispatch('fight', start.substr(0, 10), this.currentZoneId, this.currentZoneName, this.currentFight);
 
     this.currentFight = [];
   }
 
```
```diff
diff --git a/ui/raidboss/emulator/data/NetworkLogConverter.js b/ui/raidboss/emulator/data/NetworkLogConverter.ts
index 90b983f3..e4ef38ee 100644
--- a/ui/raidboss/emulator/data/NetworkLogConverter.js
+++ b/ui/raidboss/emulator/data/NetworkLogConverter.ts
@@ -1,30 +1,41 @@
 import EventBus from '../EventBus';
+import LineEvent from './network_log_converter/LineEvent';
 import LogRepository from './network_log_converter/LogRepository';
 import ParseLine from './network_log_converter/ParseLine';
 
 export default class NetworkLogConverter extends EventBus {
-  async convertFile(data) {
+  convertFile(data: string): LineEvent[] {
     const repo = new LogRepository();
-    const ret = await this.convertLines(
+    return this.convertLines(
         // Split data into an array of separate lines, removing any blank lines.
         data.split(NetworkLogConverter.lineSplitRegex).filter((l) => l !== ''),
         repo,
     );
-    return ret;
   }
 
-  async convertLines(lines, repo) {
-    this.Combatants = {};
+  convertLines(lines: (string | LineEvent)[], repo: LogRepository): LineEvent[] {
     repo = repo || new LogRepository();
-    lines = lines.map((l) => ParseLine.parse(repo, l)).filter((l) => l);
+    // This is a bit ugly but we can't keep the original lines array values around
+    // due to memory issues
+    // Convert from string to LineEvent
+    lines =
+      (
+        (lines as string[])
+          .map((l) => ParseLine.parse(repo, l)) as (LineEvent | boolean)[])
+        .filter((l) => l) as LineEvent[];
+
+    // Call `convert` to convert the network line to non-network format and update indexing values
+    lines = (lines as LineEvent[])
+      .map((l, i) => {
+        l.convert(repo);
+        l.index = i;
+        return l;
+      });
 
-    for (let i = 0; i < lines.length; ++i) {
-      lines[i].convert(repo);
-      lines[i].index = i;
-    }
     // Sort the lines based on `${timestamp}_${index}` to handle out-of-order lines properly
-    return lines.sort((l, r) => (l.timestamp + '_' + l.index).localeCompare(r.timestamp + '_' + r.index));
+    return (lines as LineEvent[])
+      .sort((l, r) => (`${l.timestamp}_${l.index}`).localeCompare(`${r.timestamp}_${r.index}`));
   }
-}
 
-NetworkLogConverter.lineSplitRegex = /\r?\n/gm;
+  static lineSplitRegex = /\r?\n/gm;
+}
```

valarnin:
As a follow up question, it seems like a lot of these typescript conversions aren't going to be detected as a rename. Should these PRs instead be two commits (renames first, then code adjustments)? I believe that will fix the rename detection issue.

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
+import EmulatorCommon from '../EmulatorCommon';
+import EventBus from '../EventBus';
+import { LocaleNetRegex } from '../../../../resources/translations';
+import { Lang } from 'types/global';
+import LineEvent from './network_log_converter/LineEvent';
+import { LineEvent0x01 } from './network_log_converter/LineEvent0x01';
+
+export default class LogEventHandler extends EventBus {
+  static doesLineMatch(line: string, regexes: Record<string, RegExp>): RegExpExecArray | false {
```
quisquous, line 9: Can this be `RegExpExecArray | undefined` here?

Also, can it be `regexes: Record<Lang, RegExp> | RegExp`? It seems like it'd be better for the callers to do something like `res = LogEventHandler.doesLineMatch(line, EmulatorCommon.cactbotWipeRegex);` rather than the `{ '': regex }` construction.  All of this would also let you remove the `as Lang` and the `regexes[i]?` below.

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+    return false;
+  }
+
+  public currentFight: LineEvent[] = [];
+  public currentZoneName?: string;
+  public currentZoneId = -1;
+
+  parseLogs(logs: LineEvent[]): void {
+    for (let i = 0; i < logs.length; ++i) {
+      const lineObj = logs[i] as LineEvent;
```
quisquous, line 82:
```suggestion
    for (const lineObj of logs) {
```

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      this.currentFight.push(lineObj);
+
+      lineObj.offset = lineObj.timestamp - (this.currentFight[0]?.timestamp ?? 0);
```
quisquous, line 86: It might be worth a helper getter for startTime / endTime?

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+
+      const res = LogEventHandler.isMatchEnd(lineObj.networkLine);
+      if (res) {
+        this.endFight();
+      } else if (lineObj instanceof LineEvent0x01) {
+        const castLineObj = lineObj;
+        this.currentZoneId = parseInt(castLineObj.zoneId ?? '0');
+        this.currentZoneName = castLineObj.zoneName;
```
quisquous, line 94:
```suggestion
        this.currentZoneId = parseInt(lineObj.zoneId ?? '0');
        this.currentZoneName = lineObj.zoneName;
```

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+      res.groups = res.groups ?? {};
+      res.groups.EndType = 'Unseal';
+      return res;
+    }
+    return false;
+  }
+
+  public currentFight: LineEvent[] = [];
+  public currentZoneName?: string;
+  public currentZoneId = -1;
```
quisquous, line 78: Unrelated to this patch, but do you think we need a `ZoneId.InvalidZone`?

ui/raidboss/emulator/data/NetworkLogConverter.ts
```diff
@@ -0,0 +1,41 @@
...
+    const repo = new LogRepository();
+    return this.convertLines(
+        // Split data into an array of separate lines, removing any blank lines.
+        data.split(NetworkLogConverter.lineSplitRegex).filter((l) => l !== ''),
+        repo,
+    );
+  }
+
+  convertLines(lines: (string | LineEvent)[], repo: LogRepository): LineEvent[] {
+    repo = repo || new LogRepository();
```
quisquous, line 17: Is this necessary? It looks like from both the type and the callers that repo always exists.

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  endFight(): void {
+    if (this.currentFight.length < 2)
+      return;
+
+    const start = new Date(this.currentFight[0]?.timestamp ?? 0).toISOString();
+    const end = new Date(this.currentFight.slice(-1)[0]?.timestamp ?? 0).toISOString();
+    this.currentZoneName = this.currentZoneName || 'Unknown';
+    this.currentZoneId = this.currentZoneId || -1;
```
quisquous, line 107:
```suggestion
    this.currentZoneName = this.currentZoneName ?? 'Unknown';
    this.currentZoneId = this.currentZoneId ?? -1;
```

Alternatively, maybe a `reset()` function for setting all of the `current` variables back to default values initially/after endFight?

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,119 @@
...
+    res = LogEventHandler.doesLineMatch(line, EmulatorCommon.countdownRegexes);
+    if (res) {
+      res.groups = res.groups ?? {};
+      res.groups.StartIn = (parseInt(res.groups.time ?? '0') * 1000).toString();
+      res.groups.StartType = 'Countdown';
+      return res;
+    }
+    res = LogEventHandler.doesLineMatch(line, EmulatorCommon.sealRegexes);
+    if (res) {
+      res.groups = res.groups ?? {};
```
quisquous, line 32:
```suggestion
    if (res && res.groups) {
```

ui/raidboss/emulator/data/NetworkLogConverter.ts
```diff
@@ -0,0 +1,41 @@
...
+  convertFile(data: string): LineEvent[] {
+    const repo = new LogRepository();
+    return this.convertLines(
+        // Split data into an array of separate lines, removing any blank lines.
+        data.split(NetworkLogConverter.lineSplitRegex).filter((l) => l !== ''),
+        repo,
+    );
+  }
+
+  convertLines(lines: (string | LineEvent)[], repo: LogRepository): LineEvent[] {
```
quisquous, line 16: Who is calling `convertLines` with `LineEvent[]`?

ui/raidboss/emulator/data/NetworkLogConverter.ts
```diff
@@ -0,0 +1,41 @@
...
+    lines = (lines as LineEvent[])
+      .map((l, i) => {
+        l.convert(repo);
+        l.index = i;
+        return l;
+      });
+
+    // Sort the lines based on `${timestamp}_${index}` to handle out-of-order lines properly
+    return (lines as LineEvent[])
+      .sort((l, r) => (`${l.timestamp}_${l.index}`).localeCompare(`${r.timestamp}_${r.index}`));
```
quisquous, line 37: There is...a lot going on in this function.  Can I suggest something that doesn't require `as` everywhere?

```suggestion
   const lineEvents = lines.map((l) => ParseLine.parse(repo, l)).filter(isLineEvent);

    // Call `convert` to convert the network line to non-network format and update indexing values
    lineEvents.map((l, i) => {
      l.convert(repo);
      l.index = i;
      return l;
    });

    // Sort the lines based on `${timestamp}_${index}` to handle out-of-order lines properly
    return lineEvents.sort((l, r) => (`${l.timestamp}_${l.index}`).localeCompare(`${r.timestamp}_${r.index}`));
```

Plus a type-guard function up above to filter undefined out without an `as`:

```typescript
const isLineEvent = (line?: LineEvent): line is LineEvent => {
  return !!line;
};
```

I'm not sure that I understand what the memory issues are here.  The caller is holding onto `lines` as well, so clobbering `lines` here doesn't remove the final ref.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: Switching to `undefined` makes sense here, yeah. I don't know how to make `regexes: Record<Lang, RegExp> | RegExp` work in the scope of the function though.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: `i` was used at one point for this loop but isn't anymore. Adjusted for next commit.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: Adjusted for next commit.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: I could have sworn I tried this at one point and tsc complained, but it seems fine now? 🤷‍♂️

Adjusted for next commit.

ui/raidboss/emulator/data/NetworkLogConverter.ts
valarnin: This is leftover code from the original bootstrapping raidemulator test framework I was using to compare network lines to converted non-network lines pulled from ACT. Removed for next commit.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: We explicitly don't want to reset the state to default for LogEventHandler since the ZoneChange line only happens once whereas we might load a bunch of encounters after it for that zone.

I think instead these should just be defaulted in the property declaration? `currentZoneId` is already being defaulted but `currentZoneName` isn't.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: This is an annoying case where `groups` should always be present, but on the off chance that it isn't (due to changes in `LocaleNetRegex` maybe), we'd want to throw an error instead of continuing blindly because it's going to cause issues with the logic in `Encounter` if there's ever a case where there's a start/end line that doesn't have the property set?

Basically, I'm saying that `if (res && !res.groups)` is an error state that should cause the log import process to either halt entirely or skip to the next valid case, and there's not really a good mechanism right now to do the latter.

ui/raidboss/emulator/data/NetworkLogConverter.ts
valarnin: This has to do with the conversation below related to memory. Depending on how we solve that, this will likely change.

ui/raidboss/emulator/data/LogEventHandler.ts
quisquous:
```typescript
  static doesLineMatch(line: string,
      regexes: Record<Lang, RegExp> | RegExp): RegExpExecArray | null {
    if (regexes instanceof RegExp)
      return regexes.exec(line);

    for (const langStr in regexes) {
      const lang = langStr as keyof typeof regexes;
      const res = regexes[lang].exec(line);
      if (res) {
        if (LocaleNetRegex.areaSeal[lang] && res.groups)
          res.groups.language = lang;
        return res;
      }
    }
    return null;
  }
```

quisquous: Actually `LocaleNetRegex.areaSeal[lang]` is always true, so that can be elided too.

ui/raidboss/emulator/data/LogEventHandler.ts
quisquous: Ah yeah, that makes sense!

ui/raidboss/emulator/data/LogEventHandler.ts
quisquous: Not depending on the internals of `LocaleNetRegex` is a good argument, thanks.

My followup suggestion is that these should all be `res.groups ??= {}` instead, then.

ui/raidboss/emulator/data/NetworkLogConverter.ts
valarnin: The issue isn't the strings in the array, the issue is the array itself.

Before the change to use the worker, the `lines` array passed into the function could have had an exceedingly large amount of elements, for instance the biggest single log file I have sitting around right now is 3,833,046 lines long.

That array alone is a little under 30mb just to store and track the references to the strings, not including the string data itself, and that's all assuming the javascript implementation is storing the array internally as an array of pointers and not as some more complex data type.

It's probably not nearly as important anymore and the adjustments you've suggested above would probably be fine, but it's potentially not and I'd prefer that someone double check the logic in worker to make sure I'm not missing some edge case where it might pass through a ton of lines, because that was the cause of a lot of browser crashes before.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: I don't think that having a `ZoneId.InvalidZone` makes sense because every zone ID should always be valid. It's more of a case of `ZoneId.UnknownZone` as a catch-all which I'm not sure is really possible?

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: Thanks for this. I don't know why, but I was absolutely failing to string this one together. Adjusted for next commit, including the `areaSeal` removal.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: Adjusted for next commit.

ui/raidboss/emulator/data/LogEventHandler.ts
valarnin: I keep forgetting that `??=` is an operator in typescript. Adjusted for next commit.

valarnin: As a follow up, does it make sense to note this as a comment somewhere to prevent confusion in the future?

quisquous: Yeah, that makes sense.  Some sort of `// Currently all of these regexes have groups if they match at all, but be robust to that changing in the future.` sort of thing?

ui/raidboss/emulator/data/NetworkLogConverter.ts
quisquous: I think I just don't understand how changes inside of `convertLines` saves memory, as all callers of `convertLines` are still holding onto a ref to `lines`.  (Maybe the `convertFile` one doesn't because it's just a param, vs a named variable?  But that's a v8 detail I don't know enough about to say either way.)  But for the worker, the worker still has a ref on the array, so it doesn't matter if you clobber it inside of `convertLines` because there's still a ref on it and it can't be gc'd until the function is done and you set `lines = []` in the worker.  That's really the root of my confusion, as I don't understand how the previous code made any difference to the gc situation.

(Also, maybe this is not the PR to change it, but I still do not think that you should sort the lines.  The lines have out of order times in them but the order in the file is the order in which they are presented to the plugin.  So, I think it's incorrect to change that order.)

valarnin: In current state I don't think it's going to be a memory issue anymore. I'll make these adjustments.

As for sorting, the only reason that's in place is to address issues with CombatantTracker updating states out of order. I think that's a fix for after CombatantTracker and dependents have been converted to Typescript.

However, during testing of this I've noticed that there's an issue with the getter for `LineEvent.timestamp` that needs fixed. I'm going to get a PR in for that real quick before continuing on this PR.

valarnin:
Had to include a slight adjustment to NetworkLogConverterWorker because `convertLines` is no longer async.

quisquous (commented):
Also, just as feedback, it is very hard to review your patches when you force push on every change.  I can't see what changed and have to re-read the entire thing and try to remember what was there before.  This is especially true the larger the patch is.

If you want to make the review process easier for reviewers, I would recommend:
* add separate commits for any changes due to review, so folks can look at a single commit to see what changed
* only rebase on main and force push when you need to include other changes from main
  * but don't squash commits when you do this
  * alternatively, merge main in to preserve commit history on the branch
* ideally push a commit first and then publish comments second that say "done".  If you comment "will be done in the future", then now I have to remember at some indefinite point to try to compare what was done.

ui/raidboss/emulator/data/LogEventHandler.ts
```diff
@@ -0,0 +1,130 @@
...
+      if (res) {
+        if (res.groups)
+          res.groups.language = lang;
+        return res;
+      }
+    }
+    return null;
+  }
+
+  static isMatchStart(line: string): RegExpMatchArray | false {
```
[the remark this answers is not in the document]
valarnin: It's used in `Encounter`. It lives in `LogEventHandler` because `isMatchEnd` also lives there and is used there. Given that all of these are using regexes from `EmulatorCommon`, maybe it makes sense to move them both there?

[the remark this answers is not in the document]
quisquous: That seems reasonable!

valarnin:
> 
> 
> Also, just as feedback, it is very hard to review your patches when you force push on every change. I can't see what changed and have to re-read the entire thing and try to remember what was there before. This is especially true the larger the patch is.

I started squashing the commits down (or using `git commit --amend`, in actuality) because that's what you were doing when merging them in regardless. If you'd rather a bunch of commits and squash them on merge instead, that's fine too. That does circle back to a question I asked previously on this PR:
> As a follow up question, it seems like a lot of these typescript conversions aren't going to be detected as a rename. Should these PRs instead be two commits (renames first, then code adjustments)? I believe that will fix the rename detection issue.

Should I start doing this as well?

> If you want to make the review process easier for reviewers, I would recommend:
> 
>     * add separate commits for any changes due to review, so folks can look at a single commit to see what changed
>     * only rebase on main and force push when you need to include other changes from main
>       * but don't squash commits when you do this
>       * alternatively, merge main in to preserve commit history on the branch

I think a rebase to main shouldn't require a force push, though I might be wrong? As such, my preferred methodology is to rebase to main any time I need to make further commits, just to avoid future merge issues and to make sure that something hasn't changed on main that would impact the changes in the PR.

>     * ideally push a commit first and then publish comments second that say "done".  If you comment "will be done in the future", then now I have to remember at some indefinite point later to try to compare what was done.
>       * My workflow is usually to leave a draft review comment that says "done" and then publish all the comments at once when I push changes.

I don't think I can draft a comment on someone else's review? I can't seem to find that option, either way. Would you prefer that I comment that the changes have been made after pushing the commit, or just mark the conversation as resolved without commenting? I'm not sure which states github notifies for.

quisquous:
> > Also, just as feedback, it is very hard to review your patches when you force push on every change. I can't see what changed and have to re-read the entire thing and try to remember what was there before. This is especially true the larger the patch is.
> 
> I started squashing the commits down (or using `git commit --amend`, in actuality) because that's what you were doing when merging them in regardless. If you'd rather a bunch of commits and squash them on merge instead, that's fine too. That does circle back to a question I asked previously on this PR:

Yeah, I think the only time it makes sense to have a clean history in the PR itself is if you think the change should be rebase+merge instead of squash+merge onto main so that the commits are preserved.  But I think that's probably a really rare case and probably those should just be different squashed PRs in the end.

> > As a follow up question, it seems like a lot of these typescript conversions aren't going to be detected as a rename. Should these PRs instead be two commits (renames first, then code adjustments)? I believe that will fix the rename detection issue.
> 
> Should I start doing this as well?

I think if it's in the same PR, then yeah, maybe rename first then changes as separate commits might help make it more readable.

> > If you want to make the review process easier for reviewers, I would recommend:
> > ```
> > * add separate commits for any changes due to review, so folks can look at a single commit to see what changed
> > * only rebase on main and force push when you need to include other changes from main
> >   * but don't squash commits when you do this
> >   * alternatively, merge main in to preserve commit history on the branch
> > ```
> 
> I think a rebase to main shouldn't require a force push, though I might be wrong? As such, my preferred methodology is to rebase to main any time I need to make further commits, just to avoid future merge issues and to make sure that something hasn't changed on main that would impact the changes in the PR.

Yeah, a rebase requires a force push when you've changed the base.  The only time a rebase wouldn't require a force push is if there were no commits since the last time you did it.

> > ```
> > * ideally push a commit first and then publish comments second that say "done".  If you comment "will be done in the future", then now I have to remember at some indefinite point later to try to compare what was done.
> >   * My workflow is usually to leave a draft review comment that says "done" and then publish all the comments at once when I push changes.
> > ```
> 
> I don't think I can draft a comment on someone else's review? I can't seem to find that option, either way. Would you prefer that I comment that the changes have been made after pushing the commit, or just mark the conversation as resolved without commenting? I'm not sure which states github notifies for.

This is github interface weirdness, but this is the difference between the main PR page (where you can cancel/comment/resolve) and the files changed page (where you can cancel/add single comment/start a review).  It's the same difference between merging in somebody's suggestions one at a time vs batching them in the files changed page.  At any rate, "starting a review" should let you write draft comments and then publish at once.  "Add single comment" or "comment" from the main PR page both are instant with no draft.

ui/raidboss/emulator/data/LogEventHandler.ts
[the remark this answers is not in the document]
valarnin: This change is done.

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -112,6 +113,82 @@ export default class EmulatorCommon {
...
+      if (res) {
+        if (res.groups)
+          res.groups.language = lang;
+        return res;
+      }
+    }
+    return null;
+  }
+
+  static matchStart(line: string): RegExpMatchArray | false {
```
quisquous, line 133:
```suggestion
  static matchStart(line: string): RegExpMatchArray | undefined {
```

ui/raidboss/emulator/EmulatorCommon.ts
```diff
@@ -112,6 +113,82 @@ export default class EmulatorCommon {
...
+    if (res) {
+      res.groups ??= {};
+      res.groups.StartIn = '0';
+      res.groups.StartType = 'Engage';
+      return res;
+    }
+    return false;
+  }
+
+  static matchEnd(line: string): RegExpMatchArray | false {
```
quisquous, line 161:
```suggestion
  static matchEnd(line: string): RegExpMatchArray | undefined {
```

valarnin:
Was hoping I wouldn't have to change my local back to the branch to apply those but looks like that's a no-go. Let me fix those up, give me a few minutes...

PR #2921, opened 2021-05-23
util: add eval check to process-triggers

quisquous:
This will make sure that trigger files actually are usable as user
files on the triggers branch.

Also, make line length a warning now that we have a limit of
zero warnings.  This will make it less noisy and let us detect
other errors in trigger files.

### DIFF
--- a/util/process_triggers_folder.ts
+++ b/util/process_triggers_folder.ts
@@ -80,7 +80,7 @@ const processFile = async (filename: string) => {
   }
 
   const contents = lintResult.output;
-  if (!contents) {
+  if (contents === undefined) {
     console.error(`${filename}: Lint returned no contents`);
     process.exit(4);
   }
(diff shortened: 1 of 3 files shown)


util/process_triggers_folder.ts
```diff
@@ -44,35 +46,57 @@ const changeExportToPush = (lines: string[]) => {
...
+    // Warnings aren't great, but tsc can create unfixable line length warnings.
+    // Print them, but don't stop.
+    console.error(`${filename}: Lint ran with warnings: ${JSON.stringify(lintResult.messages)}`);
+  }
+
+  const contents = lintResult.output;
+  if (!contents) {
+    console.error(`${filename}: Lint returned no contents`);
+    process.exit(4);
+  }
```
panicstevenson, line 86: Does output return an all-clear normally?

panicstevenson (approved):

util/process_triggers_folder.ts
quisquous: What do you mean by that?

panicstevenson: If everything is good, we still expect an output?

quisquous: Yeah, the output is the linted file.  I guess technically this could be empty (and valid) vs undefined, so I should probably be more precise here.

PR #2922, opened 2021-05-24
raidemulator: Fix issue with LineEvent getter data passing postMessage boundary

valarnin:
After the conversion of LineEvent to use getters, that data was no longer capable of passing the postMessage boundary from the worker to the main window thread. This fixes the issue by iterating the getters on the lines and populating them on the passed generic object.

Note that, even if we were to get rid of the Worker threading, this would still be required as IndexedDB only stores plain objects.

This may make more sense to put directly in the `EmulatorCommon.cloneData` function as base functionality but the performance impact on that is noticeable.

### DIFF
--- a/ui/raidboss/emulator/data/NetworkLogConverter.ts
+++ b/ui/raidboss/emulator/data/NetworkLogConverter.ts
@@ -21,7 +21,6 @@ export default class NetworkLogConverter extends EventBus {
     let lineEvents = lines.map((l) => ParseLine.parse(repo, l)).filter(isLineEvent);
     // Call `convert` to convert the network line to non-network format and update indexing values
     lineEvents = lineEvents.map((l, i) => {
-      l.convert(repo);
       l.index = i;
       return l;
     });
--- a/ui/raidboss/emulator/data/NetworkLogConverterWorker.js
+++ b/ui/raidboss/emulator/data/NetworkLogConverterWorker.js
@@ -4,6 +4,17 @@ import LogEventHandler from './LogEventHandler';
 import NetworkLogConverter from './NetworkLogConverter';
 import LogRepository from './network_log_converter/LogRepository';
 
+const getClassPropertyDescriptors = (obj) => {
+  if (obj && obj !== Object.prototype) {
+    const proto = Object.getPrototypeOf(obj);
+    return {
+      ...getClassPropertyDescriptors(proto),
+      ...Object.getOwnPropertyDescriptors(obj),
+    };
+  }
+  return null;
+};
+
 onmessage = async (msg) => {
   const logConverter = new NetworkLogConverter();
   const localLogHandler = new LogEventHandler();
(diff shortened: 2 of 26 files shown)


quisquous:
Hmmmmmmmmmmmm.

If I had realized this was going to be an issue, I think I would have said no to the getters when you asked.  I think maybe it would make more sense to have `public readonly` member variables instead of trying to work around the getters.

valarnin:
Yeah, I didn't realize this issue until doing a full test after the changes to NetworkLogConverter in the other PR, they only show when trying to load an imported encounter, not on import. I can instead close this PR and change the LineEvent classes, but I think there are some properties that can't be readonly, basically anything being set from the `convert` method, since they require a fully populated `repo` object to do their lookup properly.

quisquous:
The convert method is called almost immediately after parsing though, right?  Could/should convert just be called directly in the constructor?

valarnin:
There's an important logic distinction here in that convert is called second in each batch. So the objects are constructed in batches of 1000, then converted in batches of 1000. In fact, it might even lead to an edge case bug that I didn't think of when moving things over to the worker setup, where if a line that needs its repo info is processed at the end of one batch, the info it needs won't be available.

valarnin:
Now that I've had time to investigate further, looking through the code, the only case where a name lookup absolutely has to happen is job gauge (`31`/`0x1F`) where that is, for some reason, sent absolutely first upon logging in.

I guess that we could just ignore that specific case (by marking the line as invalid if there's no entry in `LogRepository` when the line is constructed), at which point all member variables could be readonly?

quisquous:
> Now that I've had time to investigate further, looking through the code, the only case where a name lookup absolutely has to happen is job gauge (`31`/`0x1F`) where that is, for some reason, sent absolutely first upon logging in.
> 
> I guess that we could just ignore that specific case (by marking the line as invalid if there's no entry in `LogRepository` when the line is constructed), at which point all member variables could be readonly?

You will always get the `0x02` line before `0x1F`.  I think in some extreme error case, it'd be better to just mark the line as having a blank name than marking the whole line as invalid?

valarnin:
Every log file I have locally right now starts with:

```
249|2021-05-13T16:08:59.1714902-04:00|Language ID: 1, Disable Damage Shield: False, Disable Combine Pets: False, Parse Filter: None, DoTCrits: False, RealDoTs: False, Parse Potency: False|989345ddb19420d2c44ee731a516e11d
249|2021-05-13T16:08:59.1814905-04:00|Selected Process ID: 0, Disable Network Data: False, Enable HP Scan: False, Dump All Network Data: True, Disable Combat Log: False, Selected IP: , WinPcap: False, Socket 
Filter: True|820d88da183f73944fddf5f933334931
31|2021-05-14T20:47:58.4930000-04:00|100ABCDE|18|00|00|00|bc4725cadebb6d423accb6b864e3f013
39|2021-05-14T20:47:59.9630000-04:00|100EDCBA||122634||10000||0|0||||||316eba1911bd687c9f11df40ebb86b0c
01|2021-05-14T20:48:01.2960000-04:00|155|The Goblet|1f43bd37b371c28656b49618c9f7396e
02|2021-05-14T20:48:01.2960000-04:00|100abcde|Some One|dbc852241900a67fd5faef44df03ff1f
```

The order is always 31, potentially misc other lines, 01, then 02, then 03 lines. This holds true for the 8 log files I have locally right now as well as the couple I have backed up elsewhere, across three different versions of the parser plugin.

quisquous:
Hmm! That's not my experience with log files.  I see a bunch of 253/249/251 and then 01/251/02/03 always.  All of my Network_20900_etc log files look like that.  Maybe I'm logging out on a crafter?

At any rate, I think it would be fine to mark any such lines as having a blank name.

valarnin:
Alright. I'll make the adjustments to readonly members instead. Is it alright to use this PR for the changes, given that all the conversation about the change is here?

quisquous:
Yeah, sounds good! This PR is about fixing the postMessage issue anyway, so seems like a great place.

valarnin:
Converted everything over to use readonly properties with the const field lookup as discussed in https://github.com/quisquous/cactbot/pull/2879#discussion_r637092942

valarnin:
Also did some full end-to-end testing after these changes to make sure they fully resolved the issue.

quisquous:
Thanks for all the changes back and forth to make this work!

PR #2924, opened 2021-05-25
ci: Use Consistent Action Dependencies

panicstevenson:
Use consistent action dependencies across workflows using the checkout
action as well as setup-node in an attempt to fix the most recent build
error: https://github.com/quisquous/cactbot/runs/2661413115


quisquous:
I'll keep my eyes out for this error again after this PR.  Thanks!

PR #2925, opened 2021-05-25
resources: add Matches type aliases

quisquous:
This changes Matches from being `Matches<RegexType<Params>>` to just
`Matches<Params>` for simplicity, as there's no concrete differences
between these types.

It also seems useful to have an indirection between our Match types
and the actual templated type so that we don't have to change
a million trigger files if we ever change how they are implemented.

### DIFF
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -1,10 +1,12 @@
-import { NetRegex } from 'types/trigger';
+import { BaseRegExp } from '../types/trigger';
 import Regexes, { Params } from './regexes';
 
 interface Fields {
   field: string;
   value?: string;
 }
+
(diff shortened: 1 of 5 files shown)


resources/netregexes.ts
```diff
@@ -1,11 +1,13 @@
-import { NetRegex } from 'types/trigger';
+import { BaseRegExp } from '../types/trigger';
```
panicstevenson, line 1: Should this still be 'types/trigger' and not the relative path?

panicstevenson (approved):

resources/netregexes.ts
quisquous: I am not sure how that was working, so I'm going to stick with the relative path import.

PR #2927, opened 2021-05-26
plugin: Update intl CE Opcode for 5.55

KattTails:
I'll likely start on adding Zadnor support to the eureka overlay soon. I took a quick look at the network data already.

I also still plan on reworking opcodes so it can be read from a file, removing the need for a re-release every time it changes, it's on my todo list, sorry.


quisquous:
> I'll likely start on adding Zadnor support to the eureka overlay soon. I took a quick look at the network data already.
> 
> I also still plan on reworking opcodes so it can be read from a file, removing the need for a re-release every time it changes, it's on my todo list, sorry.

No need for apologies whatsoever!  Thank you for always finding this opcode and all your work!

PR #2931, opened 2021-05-26
raidemulator: Clean up some issues with typescript conversion

valarnin:
Found two outstanding issues with typescript conversion for raidemulator while testing the parsing plugin update yesterday.

I'll probably revisit the `zoneId` issue once the rest of raidemulator is converted, since it makes more sense to do the `parseInt` conversion when loading and parsing the log data, rather than later when persisting the encounter to DB.

### DIFF
--- a/ui/raidboss/emulator/data/Encounter.js
+++ b/ui/raidboss/emulator/data/Encounter.js
@@ -32,8 +32,9 @@ export default class Encounter {
       if (res) {
         this.firstLineIndex = i;
         this.startStatus.add(res.groups.StartType);
-        if (parseInt(res.groups.StartIn) >= 0)
-          this.engageAt = Math.min(line.timestamp + parseInt(res.groups.StartIn), this.engageAt);
+        const startIn = parseInt(res.groups.StartIn);
+        if (startIn >= 0)
+          this.engageAt = Math.min(line.timestamp + startIn, this.engageAt);
       } else {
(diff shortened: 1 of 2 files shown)


ui/raidboss/emulator/data/Encounter.js
```diff
@@ -32,8 +32,8 @@ export default class Encounter {
       if (res) {
         this.firstLineIndex = i;
         this.startStatus.add(res.groups.StartType);
-        if (res.groups.StartIn >= 0)
-          this.engageAt = Math.min(line.timestamp + res.groups.StartIn, this.engageAt);
+        if (parseInt(res.groups.StartIn) >= 0)
+          this.engageAt = Math.min(line.timestamp + parseInt(res.groups.StartIn), this.engageAt);
```
quisquous, line 36:
```suggestion
        const startIn = parseInt(res.groups.StartIn);
        if (startIn >= 0)
          this.engageAt = Math.min(line.timestamp + startIn, this.engageAt);
```

quisquous (commented):
Yeah, agreed with your feeling that the `parseInt` conversion should happen during loading.

ui/raidboss/emulator/data/Encounter.js
valarnin: Makes sense to me.

PR #2940, opened 2021-05-27
remove some implicitly any

Trim21:


quisquous:
I am not sure what you are fixing here in particular, as these appear to be `never[]` and not `any[]`?

quisquous:
Also, I think if you wanted to change this, I think something like this (that doesn't use `as` type assertions) might be better:

```typescript
const emptyRoleToPartyNames = (): Record<Role, string[]> => {
  return {
    tank: [],
    healer: [],
    dps: [],
    crafter: [],
    gatherer: [],
    none: [],
  };
};
```

PR #2941, opened 2021-05-27
added german translations for Zadnor

Akurosia:


KattTails:
Just a warning, some of these translations are really long and there might be some major readability issues if multiple pop at the same time.

Akurosia:
some of them cant made shorter as they either 1) duplicate themself or 2) would not make sense to put in as a text
From my testing sofar even if they are close by you still can make out which fate is which

PR #2947, opened 2021-05-28
raidemulator: Fix various issues related to using relative offsets instead of log line timestamps

valarnin:
This PR fixes a number of minor issues related to using offsets instead of log line timestamps which were left over from #2862, which I incorrectly attributed to the changes in #2942.

Also included in this PR is a fix for a long outstanding issue with seeking during an encounter with a jump (e.g. e12s p2) which these fixes exposed the root cause of.

### DIFF
--- a/ui/raidboss/emulator/data/RaidEmulator.js
+++ b/ui/raidboss/emulator/data/RaidEmulator.js
@@ -10,7 +10,7 @@ export default class RaidEmulator extends EventBus {
     this.playingInterval = null;
     this.currentLogLineIndex = null;
     this.lastLogTimestamp = null;
-    this.emulatedTimestamp = null;
+    this.lastTickTime = null;
   }
   addEncounter(encounter) {
     this.encounters.push(encounter);
@@ -33,7 +33,7 @@ export default class RaidEmulator extends EventBus {
 
   selectPerspective(ID) {
     this.currentEncounter.selectPerspective(ID);
-    this.seekTo(this.currentTimestamp);
+    this.seekTo(this.currentLogTime);
   }
 
   play() {
@@ -42,11 +42,10 @@ export default class RaidEmulator extends EventBus {
 
     const firstIndex = this.currentEncounter.encounter.firstLineIndex;
 
-    this.currentTimestamp = this.currentTimestamp ||
+    this.currentLogTime = this.currentLogTime ||
       this.currentEncounter.encounter.logLines[firstIndex].timestamp;
     this.currentLogLineIndex = this.currentLogLineIndex || firstIndex - 1;
-    this.emulatedTimestamp = Date.now();
-    // Use setInterval since it should account for differences in execution time automagically
+    this.lastTickTime = Date.now();
     this.playingInterval = window.setInterval(this.tick.bind(this), RaidEmulator.playbackSpeed);
     this.dispatch('play');
     return true;
@@ -54,7 +53,7 @@ export default class RaidEmulator extends EventBus {
 
   pause() {
     window.clearInterval(this.playingInterval);
-    this.emulatedTimestamp = null;
+    this.lastTickTime = null;
     this.playingInterval = null;
     this.dispatch('pause');
     return true;
@@ -83,7 +82,7 @@ export default class RaidEmulator extends EventBus {
           await this.dispatch('emitLogs', { logs: logs });
           logs = [];
         }
-        this.currentTimestamp = this.lastLogTimestamp = line.timestamp;
+        this.currentLogTime = this.lastLogTimestamp = line.timestamp;
         ++this.currentLogLineIndex;
         await this.dispatch('midSeek', line);
         continue;
@@ -98,7 +97,7 @@ export default class RaidEmulator extends EventBus {
     }
 
     await this.dispatch('postSeek', seekTimestamp);
-    await this.dispatch('tick', this.currentTimestamp, this.lastLogTimestamp);
+    await this.dispatch('tick', this.currentLogTime, this.lastLogTimestamp);
     if (playing)
       this.play();
   }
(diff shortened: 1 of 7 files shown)


ui/raidboss/emulator/data/RaidEmulator.js
```diff
@@ -7,11 +7,10 @@ export default class RaidEmulator extends EventBus {
     this.options = options;
     this.encounters = [];
     this.currentEncounter = null;
-    this.playing = false;
-    this.currentTimestamp = null;
+    this.playingInterval = null;
     this.currentLogLineIndex = null;
     this.lastLogTimestamp = null;
-    this.internalTimestampTracker = null;
+    this.emulatedTimestamp = null;
```
quisquous, line 13: Is it possible to use different words to differentiate what time spaces these variables are operating in?  `emulatedTimestamp` here seems to be tracking "real time", but `currentTimestamp` and `lastLogTimestamp` seem to be "log time".

Proposing:
* `emulatedTimestamp` => `lastTickTime`
* `currentTimestamp` => `currentLogTime`
* `lastLogTimestamp` => `lastLogTime`

ui/raidboss/emulator/overrides/RaidEmulatorTimeline.js
```diff
@@ -21,24 +16,21 @@ export default class RaidEmulatorTimeline extends Timeline {
...
 
-    this.SyncTo(this.emulatedFightSync +
-      ((timestampOffset - this.emulatedFightSyncLastOffset) / 1000), timestampOffset);
-    this._OnUpdateTimer(timestampOffset);
-  }
+    // This is a bit complicated due to jumps in timelines. If we've already got a timebase,
+    // fightNow needs to be calculated based off of that instead of engageAt
+    // timebase = 0 when not set
+    const baseTimestamp = this.timebase || this.emulator.currentEncounter.encounter.engageAt;
+    const fightNow = (currentTimestamp - baseTimestamp) / 1000;
```
quisquous, line 27: This does make me wonder if `this.timebase` should be `null` instead of `0` when stopped.  Not a thing to fix in this patch, but this took a little bit of staring at to remember that `this.timebase` is a "real" time (and not an offset) that is set to zero when unset.

ui/raidboss/emulator/data/RaidEmulator.js
valarnin: `currentTimestamp` is tracking elapsed time since the start of the fight according to playback and will almost never line up with an actual log timestamp. It might be better described as `currentFightTime`.

The other two look fine.

ui/raidboss/emulator/overrides/RaidEmulatorTimeline.js
valarnin: Yeah, I scratched my head a bit at that. I can look into a PR for it after this one is merged?

quisquous: Sounds good!

ui/raidboss/emulator/data/RaidEmulator.js
quisquous: What about `currentTimestamp` => `elapsedTime`

valarnin: To be clear, `currentTimestamp`:

- Is a unix timestamp
- Always occurs between the first log line of the encounter and the last log line of the encounter
- Does not necessarily correspond to a given log line
- Increases every tick of the emulator
- May decrease if seeking to a previous point in the encounter

quisquous: > currentTimestamp is tracking elapsed time

:thinking:  I guess I think of "elapsed" as meaning a time delta, subtracting "current time - start time".

I agree that it doesn't correspond to a given log line, but it is still "log time".  I think of there being different time spaces that we are converting between.  There's real time (Date.now()), there's timeline time (corresponding to `fightNow` in timeline), there's log time (in the emulator, corresponding to the timestamps on logs) and there's synthetic time for lack of a better word (which is `this.timebase` and can be either real time or log time, depending on whether it's in the emulator or not).

The space in which `currentTimestamp` moves around is in "log time", in that if X < `currentTimestamp` < Y, then log X happens before log Y.

That's why I'm trying to differentiate "log time" vs "timeline time" vs "real time".

valarnin: My worry is that `currentLogTime` implies that this might correspond to a given log line due to `lastLogTime` corresponding to a given log line, but that's not the case.

Maybe instead:

- `emulatedTimestamp` => `lastTickTime`
- `currentTimestamp` => `currentLogTime`
- `lastLogTimestamp` => `lastLogLineTime`

Just to differentiate the other timestamp?

quisquous: Sounds good! :+1:

ui/raidboss/emulator/data/RaidEmulator.js
```diff
@@ -46,72 +45,78 @@ export default class RaidEmulator extends EventBus {
     this.currentTimestamp = this.currentTimestamp ||
       this.currentEncounter.encounter.logLines[firstIndex].timestamp;
     this.currentLogLineIndex = this.currentLogLineIndex || firstIndex - 1;
-    this.internalTimestampTracker = Date.now();
+    this.emulatedTimestamp = Date.now();
     // Use setInterval since it should account for differences in execution time automagically
```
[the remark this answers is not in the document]
valarnin: Removed.

ui/raidboss/emulator/data/RaidEmulator.js
quisquous: (What happened to `lastLogLineTime`?)

valarnin: (It turns out, if you forget to stage your final commit changes, they don't actually get added to the commit. Whoops.)

PR #2950, opened 2021-05-29
raidemulator: Convert Tooltip to typescript

valarnin:
Ran into a new (for me) issue with the custom `<template>` tag in the emulator's HTML not translating well directly to typescript. Worked around it as best I was able to figure out but there's probably a better way.


ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -0,0 +1,167 @@
...
+
+  static getTemplate(dir: string): HTMLTemplateElement {
+    const elemName = `${dir}TooltipTemplate`;
+    const ret = document.getElementById(elemName);
+    if (ret === null) {
+      const msg = `Missing element ${elemName}`;
+      console.error(msg);
+      throw new Error(msg);
+    }
+    return ret as HTMLTemplateElement;
```
quisquous, line 159:
```suggestion
    if (ret && ret instanceof HTMLTemplateElement)
      return ret;

    const msg = `Missing element ${elemName}`;
    console.error(msg);
    throw new Error(msg);
```

ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  static cloneTemplate(direction: ValidDirection): HTMLElement {
+    const template = Tooltip.templates[direction];
+    return template.content.firstElementChild?.cloneNode(true) as HTMLElement;
```
quisquous, line 164:
```suggestion
    if (node instanceof HTMLElement)
      return node;

    const msg = `Incorrect element type for ${direction}`;
    console.error(msg);
    throw new Error(msg);
```

ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -1,44 +1,103 @@
...
-    this.offset = {
-      x: 0,
-      y: 0,
-    };
     this.target = target;
     this.direction = direction;
-    // Technically, templates should use querySelector, but this gets called so often
-    // on trigger-heavy encounters that we should hack its performance a bit
-    this.tooltip = document.getElementById(`${direction}TooltipTemplate`)
-      .content.firstElementChild.cloneNode(true);
```
quisquous: Alternatively, could this just check if `firstElementChild` is an `HTMLTemplateElement` and `cloneNode` returns an `HTMLElement`, or is it worth the code to pre-cache them as you added here?

ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -1,44 +1,103 @@
...
+      autoShow = true,
+      autoHide = true) {
+    Tooltip.initializeTemplates();
+
+    let target: HTMLElement | null;
+
+    if (typeof targetRef === 'string')
+      target = document.querySelector<HTMLElement>(targetRef);
+    else
+      target = targetRef;
```
quisquous, line 53: FWIW, I think it'd be better for the caller to pass in the HTMLElement rather than having the constructor find it and fail.

ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -1,44 +1,103 @@
...
+      const msg = `Tooltip template direction ${direction} missing .tooltip-inner`;
+      console.error(msg);
+      throw new Error(msg);
+    }
+    this.inner = innerElem;
+    const arrowElem = this.tooltip.querySelector<HTMLElement>('.arrow');
+    if (!arrowElem) {
+      const msg = `Tooltip template direction ${direction} missing .arrow`;
+      console.error(msg);
+      throw new Error(msg);
```
quisquous, line 75: There is a lot of this pattern in this file.  One thing I've been doing elsewhere where something can't fail or will only fail if somebody breaks some major assumptions about the code like this is using `throw new UnreachableCode()`.  You could add a message parameter there if you wanted.  But, I've been using that for "this should ~never happen".

(I think this is different than the first `!target` check.  Having a caller pass in some name that may or may not exist seems like a failure case that should be handled [which is why I suggested not making the constructor handle it] but having a missing element in a template seems like something you should be able to just assume internally is true, if that make sense?)

ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -0,0 +1,167 @@
...
+      const msg = 'Invalid selector or element passed to Tooltip';
+      console.error(msg);
+      throw new Error(msg);
+    }
+
+    this.target = target;
+    this.direction = direction;
+    this.tooltip = Tooltip.cloneTemplate(direction);
+    const innerElem = this.tooltip.querySelector<HTMLElement>('.tooltip-inner');
+    if (!innerElem) {
```
quisquous, line 65: This is just a type assertion, right? `querySelector` isn't doing anything fancy with the parameter other than casting? If so, then maybe:

```suggestion
    const innerElem = this.tooltip.querySelector('.tooltip-inner');
    if (!innerElem || !(innerElem instanceof HTMLElement)) {
```

ui/raidboss/emulator/ui/Tooltip.ts
valarnin: I don't think `instanceof` will work here, since the abstract declaration is just a way of tricking tsc into letting this work?

Or it could be checking `instanceof HTMLElement` instead?

image

ui/raidboss/emulator/ui/Tooltip.ts
valarnin: I was planning on cleaning that up after the typescript conversion for all of the upstream dependencies for this class. Maybe I should add a `// @TODO:` for it as a reminder?

ui/raidboss/emulator/ui/Tooltip.ts
valarnin: So to make sure I'm understanding correctly, `throw new UnreachableCode()` everywhere but the constructor's `!target` check?

ui/raidboss/emulator/ui/Tooltip.ts
valarnin: The 'performance hack' here is using `getElementById` to lower the scope that `querySelector` would need to check against. Technically this could instead be:

```typescript
  static cloneTemplate(direction: ValidDirection): HTMLElement {
    const template = Tooltip.templates[direction];
    return template.content.querySelector('.tooltip') as HTMLElement;
  }
```

They're functionally the same.

For reference, `firstElementChild` is actually an `HTMLDivElement`.

ui/raidboss/emulator/ui/Tooltip.ts
quisquous: TODO sounds good, thanks!

ui/raidboss/emulator/ui/Tooltip.ts
quisquous: Yeah, that makes sense to me!  Feel free to disagree, but that's how I read this code.  Missing `target` feels like a plausible runtime error that should be handled but the others feel like "this should never happen".

ui/raidboss/emulator/ui/Tooltip.ts
quisquous: Oh fair.  `instance of HTMLElement` seems better then, if that's what it really is.

ui/raidboss/emulator/ui/Tooltip.ts
quisquous: Yeah yeah, agreed than caching the `getElementById` lookup.  My original comment was mostly saying "you added a bunch of code to find and cache tooltip elements by name" and wondering if that was necessary vs the original code.

ui/raidboss/emulator/ui/Tooltip.ts
valarnin: It's `instanceof HTMLTemplateElement` but there's no way to get tsc to see the browser-generated-at-runtime type, as far as I'm aware.

valarnin:
This revision feels much better overall to me. Apparently I hadn't run `npm install` before I started working on this originally, because `HTMLTemplateElement` does actually exist in Typescript proper.

ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -0,0 +1,161 @@
...
+    if (ret instanceof HTMLElement)
+      return ret as HTMLTemplateElement;
+    throw new UnreachableCode();
+  }
+
+  static cloneTemplate(direction: ValidDirection): HTMLElement {
+    const template = Tooltip.templates[direction];
+    const node = template.content.querySelector('.tooltip');
+    if (node instanceof HTMLElement)
+      return node.cloneNode(true) as HTMLElement;
```
quisquous, line 158:
```suggestion
    const node = template.content.querySelector('.tooltip')?.cloneNode(true);
    if (node instanceof HTMLElement)
      return node;
```

My general reasoning here is that you should use `as` as little as possible.  And, if you have to use it, then it probably needs a comment.

valarnin: I somehow keep forgetting that if you check `instanceof` on a variable, tsc will automatically determine its type going forward.

PR #2952, opened 2021-05-29
add a github workflow to set labels

### LINKED ISSUE #2945: add Labels to github PR
It's a little bit nosying to watch all PR.

Add labels to pr would help to filter PR like "translation: CN"

Trim21:
close #2945

only `translation: en/...` and `module: utils/raidboss/...` are implemented ( what else do we need? )

no extra step that contributors need to do. all from PR diff.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -409,6 +409,12 @@
       "integrity": "sha512-/+CRPXpBDpo2RK9C68N3b2cOvO0Cf5B9aPijHsoDQTHivnGSObdOF2BRQOYjojWTDy6nQvMjmqRXIxH55VjxxA==",
       "dev": true
     },
+    "@types/argparse": {
+      "version": "1.0.38",
+      "resolved": "https://registry.npmjs.org/@types/argparse/-/argparse-1.0.38.tgz",
+      "integrity": "sha512-ebDJ9b0e702Yr7pWgB0jzm+CX4Srzz8RcXtLJDJB+BSccqMa36uyH/zUsSYao5+BD1ytv3k3rPYCq4mAE1hsXA==",
+      "dev": true
+    },
     "@types/color-name": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/@types/color-name/-/color-name-1.1.1.tgz",
--- a/package.json
+++ b/package.json
@@ -19,11 +19,12 @@
     "lint-staged": "node_modules/.bin/lint-staged",
     "coverage-report": "node --loader ts-node/esm util/gen_coverage_report.js",
     "find-translations": "node --loader ts-node/esm util/find_missing_translations.js",
-    "translate-timeline": "node --loader ts-node/esm util/translate_timeline.js",
+    "translate-timeline": "node --loader ts-node/esm util/translate_timeline.ts",
     "process-triggers": "node --loader ts-node/esm util/process_triggers_folder.js"
   },
   "devDependencies": {
     "@babel/types": "7.12.*",
+    "@types/argparse": "^1.0.10",
     "@typescript-eslint/eslint-plugin": "^4.15.0",
     "@typescript-eslint/parser": "^4.15.1",
     "@zoltu/typescript-transformer-append-js-extension": "^1.0.1",
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -81,7 +81,7 @@ type Style = {
   regex: RegExp;
 }
 
-type Event = {
+export type Event = {
   id: number;
   time: number;
   name: string;
@@ -100,7 +100,7 @@ type Error = {
   error: string;
 }
 
-type Sync = {
+export type Sync = {
   id: number;
   origRegexStr: string;
   regex: RegExp;
--- a/util/translate_timeline.ts
+++ b/util/translate_timeline.ts
@@ -2,8 +2,10 @@ import fs from 'fs';
 import path from 'path';
 import { ArgumentParser } from 'argparse';
 import { findMissing } from './find_missing_timeline_translations';
-import { Timeline } from '../ui/raidboss/timeline';
+import { Event, Sync, Timeline } from '../ui/raidboss/timeline';
 import { walkDirSync } from './file_utils';
+import { Lang } from 'types/global';
+import Options from '../ui/raidboss/raidboss_options';
 
 const parser = new ArgumentParser({
   addHelp: true,
@@ -11,6 +13,7 @@ const parser = new ArgumentParser({
 });
 parser.addArgument(['-l', '--locale'], {
   required: true,
+  type: 'string',
   help: 'The locale to translate the timeline for, e.g. de',
 });
 parser.addArgument(['-t', '--timeline'], {
@@ -21,7 +24,7 @@ parser.addArgument(['-t', '--timeline'], {
 
 const rootDir = 'ui/raidboss/data';
 
-const findTriggersFile = (shortName) => {
+const findTriggersFile = (shortName: string): string | undefined => {
   // strip extensions if provided.
   shortName = shortName.replace(/\.(?:[jt]s|txt)$/, '');
 
@@ -33,29 +36,31 @@ const findTriggersFile = (shortName) => {
   return found;
 };
 
-const run = async (args) => {
+const run = async (args: { locale: Lang; timeline: string }) => {
+  if (!process.argv[1]) {
+    console.error('Unable to determine current process filepath, aborting.');
+    process.exit(-2);
+  }
   process.chdir(path.join(path.dirname(process.argv[1]), '..'));
 
   const triggersFile = findTriggersFile(args.timeline);
   if (!triggersFile) {
     console.error(`Couldn\'t find '${args.timeline}', aborting.`);
     process.exit(-2);
-    return;
   }
 
   const timelineFile = triggersFile.replace(/\.[jt]s$/, '.txt');
   if (!fs.existsSync(timelineFile)) {
     console.error(`Couldn\'t find '${timelineFile}', aborting.`);
     process.exit(-2);
-    return;
   }
 
   const locale = args.locale;
 
   // Use findMissing to figure out which lines have errors on them.
-  const syncErrors = {};
-  const textErrors = {};
-  await findMissing(triggersFile, locale, (filename, lineNumber, label, message) => {
+  const syncErrors: { [lineNumber: number]: boolean } = {};
+  const textErrors: { [lineNumber: number]: boolean } = {};
+  await findMissing(triggersFile, locale, (filename: string, lineNumber: number, label: string) => {
     if (!filename.endsWith('.txt') || !lineNumber)
       return;
     if (label === 'text')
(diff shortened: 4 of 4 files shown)


.github/workflows/labels.yaml
```diff
@@ -0,0 +1,77 @@
...
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/github-script@v4
+        id: my-script
+        with:
+          github-token: ${{secrets.GITHUB_TOKEN}}
+          script: |
```
quisquous, line 15: Can this script live in a separate file?

quisquous:
Can I propose shorter names? For modules, just the module on its own, e.g `resources`, `raidboss`, `util`, `eureka` etc.  For translations, what about ":globe_with_meridians:cn" or ":speech_balloon:cn"

Trim21:
> Can I propose shorter names? For modules, just the module on its own, e.g `resources`, `raidboss`, `util`, `eureka` etc. For translations, what about "🌐cn" or "💬cn"

ok

Trim21:
I may need to import some lib that has nothing to do with cactbot for supporting translation detection of `timelineReplace`.

It seems that it's better to write a github action instead of a javascript file here, but which will make code outside the cactbot repo, and being executed in this repo.

quisquous:
Is it possible to have a script in a separate file but keep the action in this repo?

Trim21:
> Is it possible to have a script in a separate file but keep the action in this repo?

I find it's way too complex to detect translation change of timeline replace... So maybe just keep this PR like this.

Trim21:
oh I get an idea. I could eval timeline replace part and do an object diff instead of source code diff.

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,63 @@
...
+      return module;
+    }
+
+    if (f.filename.startsWith('resources/'))
+      return `resources`;
+
+    if (f.filename.startsWith('util/'))
+      return `util`;
+
+    return undefined;
```
quisquous, line 17:
```suggestion
```

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,63 @@
...
+    pull_number: pullNumber,
+    // this is used by github api, nothing I can do with it.
+  });
+  const listFilesResponse = await github.paginate(listFilesOptions);
+  return listFilesResponse;
+};
+
+const parseChangedLang = (patch) => {
+  const set = new Set();
+  for (const lang of ['en', 'cn', 'de', 'ja', 'fr']) {
```
quisquous, line 51:
```suggestion
  for (const lang of ['cn', 'de', 'ja', 'fr']) {
```

I think we shouldn't include English as a translation here.

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,63 @@
...
+  }
+};
+
+const getChangedFiles = async (github, owner, repo, pullNumber) => {
+  const listFilesOptions = github.pulls.listFiles.endpoint.merge({
+    owner,
+    repo,
+    // eslint-disable-next-line camelcase
+    pull_number: pullNumber,
+    // this is used by github api, nothing I can do with it.
```
quisquous, line 43:
```suggestion
    'owner': owner,
    'repo': repo,
    'pull_number': pullNumber,
```

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,63 @@
...
+  });
+  const listFilesResponse = await github.paginate(listFilesOptions);
+  return listFilesResponse;
+};
+
+const parseChangedLang = (patch) => {
+  const set = new Set();
+  for (const lang of ['en', 'cn', 'de', 'ja', 'fr']) {
+    const pattern = new RegExp(`[-+].*${lang}: `);
```
quisquous, line 52:
```suggestion
    const pattern = new RegExp(`^\s*[+].*(?:${lang}|'${lang}'): `);
```

* consider added lines here as adding translations
* be a little bit more strict about matching only whitespace before the language
* optional quotes

.github/workflows/labels.yaml
```diff
@@ -0,0 +1,24 @@
+name: Auto Label
+
+on:
+  pull_request:
+    types: [ opened, edited, reopened, ready_for_review, synchronize ]
```
quisquous, line 5:
```suggestion
    types: [ opened, synchronize ]
```

Maybe I misunderstand actions, but I think `ready_for_review` and `edited` do not imply a change in files, so this shouldn't need to be run? And if it runs on `opened` and `synchronize`, then I think `reopened` is taken care of?

.github/scripts/auto-label.cjs
Trim21: No, `  39:5  error  Identifier 'pull_number' is not in camel case  camelcase`

Trim21:
detection of `timelineReplace` is implemented

.github/scripts/auto-label.cjs
Trim21: and `pull_number` is object key defined by github sdk.

Trim21:
ci failured becuase PR it didn't have permission before merged.

.github/scripts/auto-label.cjs
quisquous: I think this is a cleaner fix than an eslint-disable and a comment.  Do you disagree?

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,165 @@
...
+const prefixLabelMap = {
+  'resources/': 'resources',
+  'util/': 'util',
+  'test/': 'test',
+  '.mocharc.cjs': 'test',
+  'eslint/': 'style',
+  '.eslintrc.js': 'style',
+  '.github/workflows/': 'ci',
+  '.github/scripts/': 'ci',
+
```
quisquous, line 20: Hmm, one thing I was thinking is that we already have a `fisher` label which is `fishing 🎣`.  I think also `oopsyraidsy` might be better off just being `oopsy` for brevity.  What do you think about listing everything out here explicitly, e.g.

```suggestion
  'ui/config/': 'config',
  'ui/dps/': 'dps',
  'ui/eureka/': 'eureka',
  'ui/fisher/': 'fishing🎣',
  'ui/jobs/': 'jobs',
  'ui/oopsyraidsy/': 'oopsy',
  'ui/pullcounter/': 'pull',
  'ui/radar/': 'radar',
  'ui/raidboss/': 'raidboss',
  'ui/test/': 'test',
```

.github/scripts/auto-label.cjs
Trim21: Oh, I mis-understand. Didn't expect that quote would fix this.

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,165 @@
...
+};
+
+const maybeTimelineReplaceChange = (changedFiles) => {
+  const s = new Set();
+
+  changedFiles.forEach((file) => {
+    if (!file.filepath.startsWith('ui/raidboss/data/'))
+      return;
+
+    if (path.extname(file.filepath) === '.js') {
```
quisquous, line 109:
```suggestion
    if (['.js', '.ts'].includes(path.extname(file.filepath))) {
```

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,165 @@
...
+
+const nonNullUnique = (s) => {
+  return lodash.uniq(s).filter(Boolean);
+};
+
+const rawUrl = (owner, repo, sha, path) => {
+  return `https://github.com/${owner}/${repo}/raw/${sha}/${path}`;
+};
+
+const maybeTimelineReplaceChange = (changedFiles) => {
```
quisquous, line 102: I would maybe call this `getTimelineReplaceChanges` so it starts with a verb?

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,165 @@
+'use strict';
+const path = require('path');
+
+const babelParser = require('recast/parsers/babel');
+const json5 = require('json5');
+const lodash = require('lodash');
```
quisquous, line 6: Do we really need a `lodash` dependency just to use `uniq` and `flatten`? Is it possible to not add this dependency?  I think these are just one line of code each, right?

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,165 @@
...
+  const { data: pullRequest } = await github.rest.pulls.get({
+    owner,
+    repo,
+    // eslint-disable-next-line camelcase
+    pull_number: pullNumber,
+  });
+  if (!pullRequest.mergeable) {
+    console.log('not mergeable PR, skip');
+    return;
+  }
```
quisquous, line 33: Only `maybeTimelineReplaceChange` doesn't work if the patch is not mergeable, correct? Could we still do the other checks and apply labels even for a non-mergeable patch?

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,165 @@
...
+  return Array.from(s.values());
+};
+
+const getTimelineReplace = (fileContent) => {
+  const ast = recast.parse(fileContent, {
+    parser: babelParser,
+  });
+  const exportDefault = ast.program.body.filter((body) => body.type === 'ExportDefaultDeclaration')[0];
+
+  if (exportDefault) {
```
quisquous, line 127: Just as a style nit, if a conditional lasts for the rest of the function, can you do a `if (!exportDefault) \n return;` and `if (!timeReplace) \n return;` sort of thing and prefer "early outs"? I (personally) think code is more readable when there are fewer indented contexts to have to keep track of.

quisquous (commented):
Ahhhhh this is super exciting!  Thank you for writing this!  I think this will be really helpful in having a better PR overview, and will also let people filter by various changes.

.github/scripts/auto-label.cjs
Trim21: So which committs to compare for non-mergeable PR? 

It's impossible to compare with base branch and head branch, Maybe just all PR commits?

.github/scripts/auto-label.cjs
Trim21: I didn't write a parser for ts triggers...

.github/scripts/auto-label.cjs
Trim21: `lodash` is required by `@babel/core` already.

.github/scripts/auto-label.cjs
Trim21: I think we can compare all commits in this PR if it's not mergeable.

.github/scripts/auto-label.cjs
Trim21: And there isn't a TS trigger yet, I can't test it.

quisquous: Ok, we can address this later then.

.github/scripts/auto-label.cjs
```diff
@@ -0,0 +1,169 @@
...
+  '.eslintrc.js': 'style',
+  '.github/workflows/': 'ci',
+  '.github/scripts/': 'ci',
+  'ui/config/': 'config',
+  'ui/dps/': 'dps',
+  'ui/eureka/': 'eureka',
+  'ui/fisher/': 'fishing🎣',
+  'ui/jobs/': 'jobs',
+  'ui/oopsyraidsy/': 'oopsy',
+  'ui/pullcounter/': 'pull',
```
panicstevenson, line 26:
```suggestion
  'ui/pullcounter/': 'pullcounter',
```
@quisquous thoughts?
Also, I think we should have a `raidemulator` distinction within ui/raidboss/emulator if possible?

quisquous: `pullcounter` is fine.  Special-casing `raidemulator` sounds useful too.  Do you think a raid emulator change should be both `raidboss` and `raidemulator`?

Trim21: maybe `raidemulator` only

panicstevenson: It definitely could, but I think that all people interested in raidemulator changes are interested in raidboss changes, but the opposite is certainly not true, so maybe not for the purpose of this?
(Another piece might be that all people interested in raidemulator title their PRs as such, but that feels lazy to rely on people doing that)

panicstevenson: To summarize my thoughts above since I didn't:
I think raidemulator only for raidemulator changes is fine, or both labels is also fine.

Trim21:
oh, I need to disable pull bot...

Trim21:
I'll create a private from another branch later

PR #2953, opened 2021-05-29
eureka: shorten English zadnor CE names

quisquous:


panicstevenson (approved):

panicstevenson:
I think I would say "Machines" instead of "Magitek" and "Familiar" instead of "Face", but that's my unsolicited two cents.

quisquous:
> I think I would say "Machines" instead of "Magitek" and "Familiar" instead of "Face", but that's my unsolicited two cents.

The reason I said "Magitek" was to differentiate "Attack of the Machines" from "Lean, Mean, Magitek Machines".  Maybe "Attack of the Machines" should be "Attack" and "Lean, Mean, Magitek Machines" should be "Magitek"? Internal to raidboss I think there's no ambiguity, but when you add skirmishes into the mix for the eureka code there is.

quisquous:
> > I think I would say "Machines" instead of "Magitek" and "Familiar" instead of "Face", but that's my unsolicited two cents.
> 
> The reason I said "Magitek" was to differentiate "Attack of the Machines" from "Lean, Mean, Magitek Machines". Maybe "Attack of the Machines" should be "Attack" and "Lean, Mean, Magitek Machines" should be "Magitek"? Internal to raidboss I think there's no ambiguity, but when you add skirmishes into the mix for the eureka code there is.

Oh, but there's also Attack of the Supersoldiers.  So, maybe machines, magitek, supersoldiers.  /o\

PR #2955, opened 2021-05-30
eureka: improve duel support

quisquous:
* config option for duels now plays only on CEs before duels
* different coloring for duel CEs
* cooldown timer for duels (and Castrum/Dalraida) shown on map

This should help some for folks chasing duels.

### DIFF
--- a/ui/eureka/eureka.css
+++ b/ui/eureka/eureka.css
@@ -119,6 +119,10 @@
   color: #d25454;
 }
 
+.duel .nm-time {
+  color: #e7baba;
+}
+
 @keyframes pulse {
   0% {
     transform: scale(1);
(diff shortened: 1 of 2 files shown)


panicstevenson (approved):

oukenteikan:
My ACT Eureka encountered an abnormal display problem after this update. After reading the commit history, I suspect that a change was introduced in this commit. 

- Before the update, if the weather for a NM does not arrive, it will show red color in the map. 
- However after the update, if the weather for a NM does not arrive and it didn't die during my exploring, it shows white color and the time is how long it is before we can trigger the NM.

I find it inconvenient because I need a second sight on the weather panel to distinguish whether the number means "Oh I need to wait XXX if I want to trigger NM" or "Oh we still have XXX time for triggering, hurry up".

PR #2958, opened 2021-05-30
raidboss: Convert Short One-Line Functions to Single Line

panicstevenson:
Convert any short (45 characters or less) single line function to be
an implicit returns instead.

### DIFF
--- a/ui/raidboss/data/03-hw/raid/a8n.js
+++ b/ui/raidboss/data/03-hw/raid/a8n.js
@@ -201,9 +201,7 @@ export default {
       netRegexJa: NetRegexes.ability({ source: 'ブルートジャスティス', id: '174C' }),
       netRegexCn: NetRegexes.ability({ source: '残暴正义号', id: '174C' }),
       netRegexKo: NetRegexes.ability({ source: '포악한 심판자', id: '174C' }),
-      run: (data, matches) => {
-        data.bruteTank = matches.target;
-      },
+      run: (data, matches) => data.bruteTank = matches.target,
     },
     {
       // The only dangerous Flarethrower is the first one in any rotation.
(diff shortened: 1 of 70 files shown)


ui/raidboss/data/03-hw/raid/a8n.js
```diff
@@ -221,9 +219,7 @@ export default {
       netRegexCn: NetRegexes.ability({ source: '残暴正义号', id: ['1750', '1756'] }),
       netRegexKo: NetRegexes.ability({ source: '포악한 심판자', id: ['1750', '1756'] }),
       suppressSeconds: 5,
-      run: (data, matches) => {
-        data.bruteTankOut = matches.id === '1756';
-      },
+      run: (data, matches) => data.bruteTankOut = matches.id === '1756',
```
panicstevenson, line 222: This is the only `==` I found within the diff (not that there isn't other implicit ways to return)

ui/raidboss/data/04-sb/alliance/orbonne_monastery.js
```diff
@@ -194,9 +188,7 @@ export default {
       netRegexJa: NetRegexes.ability({ id: '3851', source: '模造聖域' }),
       netRegexCn: NetRegexes.ability({ id: '3851', source: '仿制圣域' }),
       netRegexKo: NetRegexes.ability({ id: '3851', source: '모조 성역' }),
-      run: (data, matches) => {
-        data.halidom.push(matches.target);
-      },
+      run: (data, matches) => data.halidom.push(matches.target),
```
panicstevenson, line 191: Technically this will return the length of `data.halidom`?

ui/raidboss/data/03-hw/raid/a8n.js
quisquous: Do you mean `===` and not `==`? I'm not sure what you're commenting on, sorry?

panicstevenson: Either, `===` will be caught by searching for `==`; I'm mainly looking to see if there's any form of logic changes (although I think we don't care about the return value from preRun/run ?)

ui/raidboss/data/04-sb/alliance/orbonne_monastery.js
quisquous: `run` is a function with a `void` return type, so I think this is fine.  Similarly, `T1 Initiated` is `run: (data) => data.started = true,` which returns `true` in my triggerset patch.

PR #2961, opened 2021-05-30
oopsy: Add Oopsyraidsy Data to ESLint Overrides

panicstevenson:


ui/oopsyraidsy/data/00-misc/buffs.js
```diff
@@ -114,6 +114,7 @@ const missedMitigationBuff = (args) => {
   });
 };
 
+// eslint-disable-next-line no-unused-vars
```
panicstevenson, line 117: 🚨

quisquous:
Mmmm, is this all arrow functions + shorten things into one-liner arrows + unused var renaming??

panicstevenson:
> Mmmm, is this all arrow functions + shorten things into one-liner arrows + unused var renaming??

Yes, but only shortening things that are short one-liner returns that don't return data objects, not every short line like in raidboss.

The process:

1. Convert all functions to arrow functions with regex find and replace.
1. Update all unused parameter names via VSCode.
1. Update the ESLint rule override. (NOTE: Bardam's Mettle abilityWarn and buffs' missedDamageBuff)
1. Remove any tail-end `_matches)`, `_data)`, etc with find and replace for additional cleanup.
1. Update any short one-liner returns (45 characters or less) to use the arrow function implicit return.

Additionally, convert all files to TypeScript and run tsc to check my work then convert back.

quisquous:
Hmmmmm ok.  I think as a future suggestion, it's easy for me to review mechanical changes when the mechanical change applies to everything identically.  But when there's multiple mechanical changes, then each one is some slightly different combination of those changes and it becomes a lot harder to review.  I think in the future, I'd rather review 70 files x 1 change x 5 prs rather than 70 files x 5 changes x 1 pr.  I can do this because it's here and there's a ~manual step in the middle that would be obnoxious to redo, but a thought for the future.

ui/oopsyraidsy/data/05-shb/trial/ruby_weapon-ex.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
```
quisquous, line 1: ?

panicstevenson: Genuinely unsure..

PR #2966, opened 2021-05-31
Added timeline and trigger translation

Akurosia:
hope that these are all currently missing translations for DE done, only missing are the common replacements (seperate PR)


quisquous:
Thanks for all the translations, as always!!

PR #2969, opened 2021-05-31
ci: Update NPM Cache and Install Commands

panicstevenson:
Update `npm install` to instead use `npm ci`.

After looking into an error in the [update-gh-pages](https://github.com/quisquous/cactbot/runs/2707837316?check_suite_focus=true) action, I noticed a
number of workflows are still using `npm install` that should be using
`npm ci`. While looking for best practices, I found [this blog post](https://www.voorhoede.nl/en/blog/super-fast-npm-install-on-github-actions/), which
I've implemented the "even faster alternative" here.

### DIFF
--- a/.github/workflows/build-artifact.yml
+++ b/.github/workflows/build-artifact.yml
@@ -66,21 +66,18 @@ jobs:
       run: msbuild /p:Configuration=Release /p:Platform=x64 plugin/Cactbot.sln
       shell: cmd
 
-    - name: Get npm cache directory
-      id: npm-cache-dir
-      run: |
-        echo "::set-output name=dir::$(npm config get cache)"
-
-    - uses: actions/cache@v2
+    - name: Cache Dependencies
+      id: cache
+      uses: actions/cache@v2
       with:
-        path: ${{ steps.npm-cache-dir.outputs.dir }}
-        key: ${{ runner.os }}-node-${{ hashFiles('**/package-lock.json') }}
-        restore-keys: |
-          ${{ runner.os }}-node-
+        path: ./node_modules
+        key: modules-${{ hashFiles('package-lock.json') }}
 
     - name: Install Node Dependencies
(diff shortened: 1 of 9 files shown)


Trim21:
could be better with 

```yaml
      - uses: actions/cache@v2
        id: restore-cache
        with:
          path: ./node_modules
          key: modules-${{ hashFiles('package-lock.json') }}

      - run: npm ci $args
        if: steps.restore-cache.outputs.cache-hit != 'true'
```

panicstevenson:
> could be better with

I'm not sure which thing(s) is the improvement, if you don't mind elaborating?

- Removing action names
- `npm ci --ignore-scripts` -> `npm ci $args`
- `cache` -> `restore-cache`

```diff
-     - name: Cache Dependencies
+     - id: restore-cache
        uses: actions/cache@v2
        with:
          path: ./node_modules
          key: modules-${{ hashFiles('package-lock.json') }}

-     - name: Install Dependencies
+     - if: steps.restore-cache.outputs.cache-hit != 'true'
        run: |
-         npm ci --ignore-scripts
+         npm ci $args
```

Trim21:
> * args

sorry, I just didn't want to type `--ignore-scripts` again, sorry for the misleading.

Trim21:
just add  `id: restore-cache` and - if: steps.restore-cache.outputs.cache-hit != 'true'`

panicstevenson:
> just add `id: restore-cache` and `- if: steps.restore-cache.outputs.cache-hit != 'true'`

Sorry if the diff is misleading, those lines are already there; they're just `cache` and not `restore-cache`, unless I missed it somewhere?

Trim21:
emmmm emmmm emmmm

sorry for the trouble, didn't know why I didn't see it at first time.

panicstevenson:
No worries! I just thought I was missing something entirely 😄

PR #2970, opened 2021-05-31
raidemulator: Fix loading encounters with a different base language, issue with getCombatants

valarnin:
As discussed in #2942, set Options.ParserLanguage to the auto-detected language when analyzing and loading encounters for playback.

This can be done via a single set in `RaidEmulator.setCurrent`.

This PR also fixes an issue with the getCombatants call which was caused by writing the additional non-tracked field code against the old version of LineEvent whereas the new version doesn't store the `parts` array because all relevant parts are extracted to proper fields now.

### DIFF
--- a/ui/raidboss/emulator/data/RaidEmulator.js
+++ b/ui/raidboss/emulator/data/RaidEmulator.js
@@ -16,7 +16,14 @@ export default class RaidEmulator extends EventBus {
     this.encounters.push(encounter);
   }
   setCurrent(index) {
-    this.currentEncounter = new AnalyzedEncounter(this.options, this.encounters[index], this);
+    const enc = this.encounters[index];
+
+    // If language was autodetected from the encounter, set the current ParserLanguage
+    // appropriately
+    if (enc.language)
+      this.options.ParserLanguage = enc.language;
+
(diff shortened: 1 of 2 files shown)


quisquous:
Do you think it's worth parsing the `249` line as well?

Reading through Encounter.js, re: language, I also have some questions:
* petNames appears to only use English?
* `if (res && res.groups && res.groups.language)` doesn't seem like it would ever be true?
* where does Encounter.language get set?

valarnin:
Sorry, on mobile so hopefully this makes sense.

petNames only in English - this is a minor issue, but I can resolve this
evening if needed.
The `res` check - this can definitely be true, but it's hard to line up due
to the massive indenting of the blocks above. Basically it'll be true if
the current line is a seal/unseal/countdown/engage line.
Encounter.language gets set on parsing in the NetworkLogConverter code
tree, where it figures out where encounters start/end.

quisquous:
Ah, I see.  `language` here is one of the extra fields that gets set outside of the regex somewhere else.  Ok, thanks!

So if somebody is in a raid with no seal and they don't use a countdown, then there's no language detected?

I'll go ahead and merge this so we can have this fix in, and the pet issue can get fixed after.

PR #2971, opened 2021-05-31
new common translations + some adjustments

Akurosia:


quisquous:
Thanks! These all look great to add.

PR #2973, opened 2021-05-31
build: Use HtmlWebpackPlugin to Generate HTML

panicstevenson:
Use [HtmlWebpackPlugin](https://webpack.js.org/plugins/html-webpack-plugin/) to partially auto-generate the HTML files for
individual modules with their respective JavaScript bundles.

Heavily leverages the great work put forward by @MaikoTan in https://github.com/quisquous/cactbot/pull/2420.
The main difference is that this PR is only attempting to use the
HtmlWebpackPlugin and update the GitHub pages build, and does not
impact CSS files for the time being.


quisquous:
Have you done manual testing with this?  The one issue I brought up on MaikoTan's patch is that config didn't work for me when I was using a plugin built out of the build folder, and I'm suspicious if that still applies if you're only handling html here.  I know this is probably n=1 developers impacted, but <_<

panicstevenson:
> Have you done manual testing with this? The one issue I brought up on MaikoTan's patch is that config didn't work for me when I was using a plugin built out of the build folder, and I'm suspicious if that still applies if you're only handling html here. I know this is probably n=1 developers impacted, but <_<

I tested all the other things from that comment, and it looks as though MaikoTan had already solved them. I have done manual testing both pointing to my local with the webpack-dev-server and pointing to my GitHub's gh-pages branch.

Trim21:
So in ngld, we will need to set html path in dist folder after this patch?

panicstevenson:
> So in ngld, we will need to set html path in dist folder after this patch?

Unfortunately, yes.

https://panicstevenson.github.io/cactbot/ui/jobs/jobs.html
http://localhost:8080/dist/ui/jobs/jobs.html

Are what I have, both currently working. Is there a way around keeping the HTML as is, but also having it update on auto-generate without messing up git?

Trim21:
Created a dummy hi folder and copy them from dist dir, then upload to gh-pages.

I think this at least didn't break gh-pages users.

panicstevenson:
> Created a dummy ui folder and copy html from dist dir, then upload to gh-pages.
> 
> I think this at least didn't break gh-pages users.

The current PR doesn't break gh-pages users, only webpack-dev-server users.

Trim21:
We can set proxy or redirect rules in dev server

And sorry, I mean, it breaks users using it with file protocol.

Trim21:
I'm not sure you are doing it correctly...

dev server port will be automatically 8081 if 8080 is taken, (which I was using it as nginx port)

Trim21:
Would it possible just do redirecting，like redirecting from ui/raidboss/raidboss.html to dist/hi/raidboss/raisboss.html?

 I think it would be easier to understand.

And this proxy rewrite looks wrong.

If http://localhost:8080/dist/ui/jobs/jobs.html is working, it will be proxied to /ui/jobs/jobs.html

MaikoTan:
What I remembered is that there has been an issue that plugin (i.e. C# code) reads config html file from `ui/config/config.html` only as it is [hard coded](https://github.com/quisquous/cactbot/blob/56e201238fccddb67f6005f3e3f6610c205c94b2/plugin/CactbotEventSource/CactbotEventSource.cs#L179-L189), so when in development, this file path is not working inconsistently. I didn't find a solution unless change the whole file skeleton, or change the config html path as configurable. Do you have some idea?

quisquous:
> > Have you done manual testing with this? The one issue I brought up on MaikoTan's patch is that config didn't work for me when I was using a plugin built out of the build folder, and I'm suspicious if that still applies if you're only handling html here. I know this is probably n=1 developers impacted, but <_<
> 
> I tested all the other things from that comment, and it looks as though MaikoTan had already solved them. I have done manual testing both pointing to my local with the webpack-dev-server and pointing to my GitHub's gh-pages branch.

Sorry, but I am skeptical.  My use case has been that I build the plugin inside the cactbot directory and use that plugin.  [This code](https://github.com/quisquous/cactbot/blob/main/plugin/CactbotEventSource/CactbotEventSource.cs#L179-L181) is probably not pointing to the correct config html anymore?

image

`FileName: C:\Users\enne\cactbot\bin\x64\Release\CactbotOverlay.dll` is the path to my DLL.

I'm not saying we have to keep my use case working if I'm the only person, but probably there needs to be some plugin changes.

panicstevenson:
> Sorry, but I am skeptical.

Skeptical of which exactly? I did not test this use case. I only tested the other items from that comment.

quisquous:
> > Sorry, but I am skeptical.
> 
> Skeptical of which exactly? I did not test this use case. I only tested the other items from that comment.

Oh!  Sorry, I misread! You tested all the *other* things.

panicstevenson:
> Oh! Sorry, I misread! You tested all the _other_ things.

Sorry, yea; I was unfamiliar with the use case. I think I understand now. Does it make sense to leave config out of this change?

quisquous:
> > Oh! Sorry, I misread! You tested all the _other_ things.
> 
> Sorry, yea; I was unfamiliar with the use case. I think I understand now. Does it make sense to leave config out of this change?

Yeah, I think it might make sense to leave this as a TODO and solve it separately.

panicstevenson:
> Yeah, I think it might make sense to leave this as a TODO and solve it separately.

Alternatively, I can add:
```cs
      var configFile = "ui/config/config.html";
      var distFolder = "dist/";
      var dir = new VersionChecker(this).GetCactbotDirectory();
      var url = Path.GetFullPath(Path.Combine(dir, distFolder, configFile));
      // Attempt to use the local webpack override, otherwise fallback to default path
      if (!File.Exists(url))
        url = Path.GetFullPath(Path.Combine(dir, configFile));
```

quisquous:
Yeah, that plugin change seems great! Go for it.

quisquous:
I did a little testing and things seem good to me!  Thanks!

quisquous:
> And it may breaks users like me requiring HTML via file protocol, didn't know how many users are doing this.

Sorry that I didn't wait for this to be resolved.  I think I read through the comments and thought it had been addressed, but now I am not sure anymore.

I am not sure quite what your use case is, but if you are not using the dev server and are running `npm run build` yourself, then I think you should be able to change your `file://` links to point to `cactbot/dist/ui/raidboss/raidboss.html` instead of `cactbot/ui/raidboss/raidboss.html` and things should continue to work.  Let me know if that's not the case and I've created a hassle for you.

PR #2974, opened 2021-05-31
fix auto label workflow permission error

Trim21:


quisquous:
haha I was just about to suggest what you added in 77f80be :+1:

PR #2977, opened 2021-06-01
ci: add plugin and docs labels

quisquous:


.github/scripts/auto-label.cjs
```diff
@@ -23,6 +23,10 @@ const prefixLabelMap = {
   '.eslintrc.js': 'style',
   '.github/workflows/': 'ci',
   '.github/scripts/': 'ci',
+  'plugin/': 'plugin',
+  'docs/': 'docs',
+  'screenshots/': 'docs',
+  'README.md': 'docs',
```
quisquous, line 29: This is a little bit suspicious, I know.  I could change all of these prefixes to be regexes instead, and then I could match .md files anywhere?  Maybe that would be better?

panicstevenson: I think `*.md` for just the top level might be a fair compromise here? Netting you all of README, CONTRIBUTING, and CODE_OF_CONDUCT? I think most other Markdown files I would want to fit this would be in `docs/`, and would also get labelled docs.
This wouldn't catch something like [.github/workflows/README.md](https://github.com/quisquous/cactbot/blob/main/.github/workflows/README.md); I can go either way on that, personally.

.github/scripts/auto-label.cjs
```diff
@@ -23,6 +23,10 @@ const prefixLabelMap = {
   '.eslintrc.js': 'style',
   '.github/workflows/': 'ci',
   '.github/scripts/': 'ci',
+  'plugin/': 'plugin',
+  'docs/': 'docs',
```
panicstevenson, line 27: Should the various ko-KR/, zh-CN/, zh-TW/ subdirectories get something for translations?

panicstevenson (approved):

.github/scripts/auto-label.cjs
Trim21: Why not just add a file extension checking?

.github/scripts/auto-label.cjs
quisquous: Done, although having the "only the first match applies" and then "this can apply multiple labels" made things slightly more complicated, but I think it's ok.

.github/scripts/auto-label.cjs
quisquous: There is a lot of regex in cactbot.  I think it'd be better to continue to use that then to add some other pattern matching mechanism.

Trim21: we have been using minimatch in lint-staged config

Trim21:
labels config file like this `ui/raidboss/raidboss_config.js` with config, too?

.github/scripts/auto-label.cjs
quisquous: Hmm, that's fair.

.github/scripts/auto-label.cjs
```diff
@@ -14,32 +14,38 @@ const github = require('@actions/github');
...
-  'ui/raidboss/raidemulator': 'raidemulator',
-  'ui/raidboss/emulator': 'raidemulator',
+const langToLabel = (lang) => `💬${lang.toUpperCase()}`;
+
+// Only the first match applies.
+const regexLabelMap = {
+  '^docs/ko-KR/': ['docs', langToLabel('ko')],
+  '^docs/zh-CN/': ['docs', langToLabel('cn')],
+  '^docs/zh-TW/': ['docs', langToLabel('cn')],
+  '.*\\.md': ['docs'],
```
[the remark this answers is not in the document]
quisquous: Done, thanks!

quisquous:
> label config file like this `ui/raidboss/raidboss_config.js` with `config`, too? #2984

Also done!

.github/scripts/auto-label.cjs
```diff
@@ -14,28 +14,43 @@ const github = require('@actions/github');
...
+
+// Only the first match applies.
+const regexLabelMap = {
+  '^docs/ko-KR/': ['docs', langToLabel('ko')],
+  '^docs/zh-CN/': ['docs', langToLabel('cn')],
+  '^docs/zh-TW/': ['docs', langToLabel('cn')],
+  '.*\\.md$': ['docs'],
+  '^docs/': ['docs'],
+  '^screenshots/': ['docs'],
+  '^resources/': ['resources'],
```
panicstevenson, line 27: Triple legend nit: timerbar_test.html is in resources.

panicstevenson (approved):

.github/scripts/auto-label.cjs
quisquous: I think we should move timerbar_test to some test/manual/ directory then?

panicstevenson: `test/danakj_timecapsule/`

PR #2936, opened 2021-05-26
raidemulator: Convert CombatantTracker and subclasses to typescript

valarnin:
Converting CombatantTracker and related dependencies to typescript.

I had to touch a lot of the LineEvent classes to let pulling combatant info work after converting to concrete types, which caused this PR to be bigger than I would have liked.

There's a fair amount of code here that'll probably need updated once someone with some more deep knowledge of typescript reviews, there's a lot of `... as ...` casting in Combatant.ts for example due to bounds checks that tsc doesn't seem to understand.

PR #2981, opened 2021-06-01
raidemulator: Convert CombatantJobSearch to typescript

valarnin:
As discussed in #2936


ui/raidboss/emulator/data/CombatantJobSearch.ts
```diff
@@ -0,0 +1,133 @@
+export default class CombatantJobSearch {
+  static getJob(abilityID: number): string | undefined {
+    for (const Job in CombatantJobSearch.abilities) {
+      if (CombatantJobSearch.abilities[Job]?.includes(abilityID))
```
quisquous, line 4: As a future (not in this PR) performance / simplicity suggestion you could turn the map below into an ability -> job map and then just return a single lookup rather than a loop and an implicit loop in includes.  Could also skip the parseInt too and store the hex string.  Maybe that's just premature optimization.  :)

ui/raidboss/emulator/data/CombatantJobSearch.ts
```diff
@@ -0,0 +1,133 @@
...
+        return Job;
+    }
+  }
+
+  static readonly abilityMatchRegex = /[a-fA-F0-9]{1,4}/i;
+
+  static readonly abilities: Record<string, number[]> = {
```
quisquous, line 11:
```suggestion
  static readonly abilities: { [job: string]: number[] } = {
```

I don't think `Record<string, etc>` should get used.  It's really just an index type, and not a `Record`, even if it technically works.

ui/raidboss/emulator/data/CombatantJobSearch.ts
quisquous, line 4:
```suggestion
  static getJob(abilityId: number): string | undefined {
    for (const job in CombatantJobSearch.abilities) {
      if (CombatantJobSearch.abilities[job ]?.includes(abilityId))
```

valarnin: Both of these are CPU vs Memory tradeoffs. I decided to opt for memory-friendly more than CPU-friendly because this isn't very often-executed code (only executed against an ability line if the combatant is a player and their job hasn't been detected yet).

The parseInt will happen upstream in LineEvent once those refactors are done.

ui/raidboss/emulator/data/CombatantJobSearch.ts
valarnin: I thought in a previous PR you had indicated that `Record` should be used for maps with static pre-defined keys? Or maybe I misunderstood and you meant keys that map to types like `Lang`?

Technically the type could also instead be auto-inferred I think, like this (though I'm not at my desktop to verify right now):
```typescript
static readonly abilities = {
...
};
```

MaikoTan: I think quis meant if the key type is a built-in basic type, we prefer the indexing approach; if the key type is a user-defined union type, we prefer to use `Record`?

quisquous: Sorry, here's my long-winded reasoning.

For context (sorry if this is repeating what you already know), my understanding is that `Record<Job, number[]>` means something very different to TypeScript than `Record<string, number[]>`.  More than just the type of the key, a `Record` on a union type (or some other type with finite values) also requires that every key is present in the `Record` with a valid value and that there are no keys that are not part of that union type.  I think there's a reason that the TypeScript page itself only lists a `Record` with a union type key as an example.

(The nice part here is that it gives you an error if anybody ever adds an additional `Job` and doesn't update a `Record<Job, etc>`. Arguably, your code here could be `Record<Job, number[]>` but you'd need to `as keyof typeof abilities` like you did in the other patch and also add crafting jobs unless you partitioned `Job` out into `BattleJob` or somesuch.  I'm not sure this is better?  Maybe it is?)

At any rate, my understanding is that `Record<Job, number[]>` is a shorter way to write `{ [key in Job]: number[] }`, My understanding also is that `Record<string, number[]>` is equivalent to all of the following:
1) `{ [key: string]: number[] }`
2) `{ [key: string]?: number[] }`
3) `{ [key in string]: number[] }`
4) `{ [key in string]?: number[] }`

I am not a TypeScript expert, but personally it feels like if you want to say "this is an object with string keys" then (1) is the "idiomatic" way to write that, and unlike `Record` it also gives you a chance to name the key.

To sum up, my feelings here are mostly:
* naming the key is nice
* `Record` behaves specially with union types, and subjectively feels like it should only be used for union types
* if there's 5 ways to do something, let's pick one idiomatic one and do that everywhere

Issue #664, opened 2019-10-15
jobs: buff tracker should track cooldowns of jobs that don't land on you


quisquous:
The buff cooldown system was built on top of showing active buffs.  Currently, if you don't get hit by a buff because you're standing too far away or you're dead, you also won't see the cooldown.

This needs to get fixed so that cooldowns happen if anybody gets an effect.

panicstevenson:
Does this depend on https://github.com/quisquous/cactbot/issues/694 / wherever the idea of party information being brought to light lives? In my line of thinking, as this would be implemented now you would be tracking buffs for people outside of your party as well (ie "missing" litany from the DRG in the other alliance).

quisquous:
Yeah, I think you need the party info to do this properly.

Echoring:
I have an idea. We can add a function `gainAbilityFromPartyMember` which triggers a buff tracker with duration 0 with a shared cooldown with the original one. Buff log is a little later than ability log, so if you get the buff, this missing tracker will be overwritten.
Such like this:
```
      emboldenmissingyou: {
        gainAbilityFromPartyMember: kAbility.Embolden,
        durationSeconds: 0,
        icon: '../../resources/ffxiv/status/embolden.png',
        // Lime.
        borderColor: '#57FC4A',
        sortKey: 3,
        cooldown: 120,
        sharesCooldownWith: ['embolden'],
      },
```

PR #2983, opened 2021-06-02
Jobs: buff tracker improvement

Echoring:
- **Use  `mobGainsEffect`/`gainEffect` instead of `gainAbility`.**
If buff misses you or debuff misses target, `gainAbility` will still errorly announce it.
This also causes that Embolden from RDM out of your party can trigger buff tracker. 

Co-Authored-By: 迷子 (Maiko Tan) <19927330+MaikoTan@users.noreply.github.com>


ui/jobs/buff_tracker.ts
```diff
@@ -314,13 +317,27 @@ export class BuffTracker {
...
+        icon: '../../resources/ffxiv/status/embolden.png',
+        // Lime.
+        borderColor: '#57FC4A',
+        sortKey: 3,
+        cooldown: 120,
+      },
+      emboldenself: {
+        // RDM himself gains a different buff with a same name.
+        // EffectId forbids a name to have two ids,
+        // so hard code here.
```
quisquous, line 336: What do you think about calling this `EmboldenSelf` in the effect_id list with a comment about it?

ui/jobs/buff_tracker.ts
```diff
@@ -277,26 +277,29 @@ export class BuffTracker {
         sortKey: 0,
       },
       offguard: {
-        gainAbility: kAbility.OffGuard,
```
quisquous, line 280: Should `gainAbility` be removed in general or are there other uses still?

quisquous (commented):
(If you are interested in other buff tracker improvements, I think fixing #664 so abilities that don't hit you from your party still go on cooldown would be a really great thing to fix in a future PR.)

ui/jobs/buff_tracker.ts
MaikoTan: Then needs a refactoring in `gen_effect_id.py`, to add a `ManualAbilityMap` or something.

ui/jobs/buff_tracker.ts
MaikoTan: I think it is better to keep it remained, 'cause we don't know if there are skills that can only be handled by `gainAbility` added in the new patch.

ui/jobs/buff_tracker.ts
quisquous: Yeah, I think adding the effect name there is better than just having an explicit value here.

ui/jobs/buff_tracker.ts
Echoring: I think we will put it to use again in the next PR _(:3rz)_

util/gen_effect_id.py
```diff
@@ -51,6 +51,14 @@
     "Flourishing Fan Dance": "1820",
     "Higanbana": "1228",
     "Wildfire": "861",
+    "Chain Stratagem": "1221",
+    "Vulnerability Up": "638",
+}
+
+# These costom name of effect will not be checked.
```
quisquous, line 58: costom => custom here and elsewhere (sorry)

util/gen_effect_id.py
```diff
@@ -51,6 +51,14 @@
...
+# Use this only when you need to handle different effects with a same name.
+costom_mapping = {
+    "Embolden (Self)": "1239",
```
quisquous, line 61: I think it might be clearer if the CustomMapping just used the name directly rather than going through the `clean_name` function.  In other words, make this `EmboldenSelf: "1239",`

Echoring: Done. But I think a clean_name function here is necessary to prevent someone using this later add a not clean name.

quisquous: What about an assert/warning/error if a custom_mapping doesn't have a clean name?

Echoring: Fine, though I think this makes code longer?

util/gen_effect_id.py
```diff
@@ -51,6 +51,15 @@
...
     "Wildfire": "861",
+    "Chain Stratagem": "1221",
+    "Vulnerability Up": "638",
+}
+
+# These custom name of effect will not be checked.
+# Use this only when you need to handle different effects with a same name.
+# The name should be clean.
+custom_mapping = {
+    "Emboldenself": "1239",
```
quisquous, line 62:
```suggestion
    "EmboldenSelf": "1239",
```

Sorry, one last comment.  Can this be `EmboldenSelf` like it was before? I think it's still two words.  <_<

Echoring: Sorry, but a clean name should only have the first character uppercase in one word. Other letter should be lower case.
If you want a finally EmboldenSelf in effect_id, you should "Embolden Self" here.

quisquous: Oh no.  I see.  Sorry for poor suggestions, then.  I think if it's a choice between `EmboldenSelf` or `Embolden (Self)` + clean or `Emboldenself` + error check, I'd prefer the first one with `EmboldenSelf` on its own.  Sorry for not remembering how the clean function works.  :C

Echoring: Well, you means directly input a clean name and not check it after?

quisquous: Yeah.  I think it's clearer to just have the name specified directly when we are making it up.

Echoring: Done.

quisquous: (I'm really sorry!! I didn't mean to go around in such a circle.  :C)

PR #2984, opened 2021-06-02
fix(config): view source fragment

Trim21:
generating url in this way: https://github.com/quisquous/cactbot/blob/triggers/04-sb/dungeon/bardams_mettle.js#:~:text=id%3A%20%27Bardam%5C%27s%20Mettle%20Rush%27

it's `id: 'Bardam\'s Mettle Rush'`

previous:

https://github.com/quisquous/cactbot/blob/triggers/04-sb/dungeon/bardams_mettle.js#:~:text=id%3A%20Bardam%5C%27s%20Mettle%20Rush

it's `id: Bardam\'s Mettle Rush`


Trim21:
this should be labeled as `config`, too?

PR #2987, opened 2021-06-02
plugin: update CE opcode for 5.55 hotfix

KattTails:
Tested working.

I still plan on changing how these are loaded so you don't need to re-release every time. I apologise for any hassle regarding it.


quisquous:
You keeping this up to date is the opposite of a hassle.  <3

PR #2988, opened 2021-06-03
ci: make language pr labels lowercase

quisquous:


Trim21:
It's already in low case

quisquous:
> It's already in low case

#2985 has uppercase, and the code says uppercase.  If you agree they should be lowercase, then I'm glad because that's what this PR is trying to do.

Trim21:
Sorry, I mean `lang.toLowerCase()`

quisquous:
> Sorry, I mean `lang.toLowerCase()`
> 
> lang is in low case already

OH! oops I understand now

PR #2990, opened 2021-06-03
Revert "fix(config): view source fragment"

quisquous:
Reverts quisquous/cactbot#2986.

`replaceAll` is a Chrome M85 feature, which OverlayPlugin does not support.

This patch caused "trig.id.replaceAll is not a function" errors which prevented the config from loading.


quisquous:
Sorry, @Trim21.

MaikoTan:
Maybe we need a babel-polyfill?

Trim21:
emmmm

Trim21:
Didn't think that

Trim21:
we can add a babel in webpack

PR #2992, opened 2021-06-03
doc: update name/link to *.ts

MaikoTan:
Since these files are already TypeScript


quisquous:
Thank you!  I'd been trying to search for files when I remember, but I forgot these.

PR #2994, opened 2021-06-03
style(jobs): make matches always not null

MaikoTan:


quisquous:
https://www.typescriptlang.org/play?#code/C4TwDgpgBAsghsAxgCwgZwIICMCWAbHUKAXigG8BYAKCijlwNAH4AuKNYAJxwDsBzANzUAvlAA+UAK48AJhABmvCDKFVqoSLAQp0GHiBLlqtANoBrCCDYdu-ALrWuvQSPFTZCpSurVEAex4Odj8AWwgAMWlEQwAKEO1UNDZ4JETsfEIQAEoSAD5yYVVfAKD41N19ZISKg1IyQuo0UIiouOrMfSyiqn9A4CgynUwGTKry4YyiOoaqJrDInkQ28fTGbO7ivuD5qIAlCABHSRxOZVjBxLYAOQCryTw8ejwIAB59o5PlF5Sh1czc3I5Yj5erdAD0YKgAFFOJw-JwmFAABJ+ADudFOUBAfkkUDAjx4PGcUGAfigiDgDxJiWg8iiwBwJSYjWaC0Q72Opxky1+nVUcxaiw5n25F10I1AXSAA

I'm not sure I understand how these functions are going to be called.

MaikoTan:
> I'm not sure I understand how these functions are going to be called.

Because these functions are called with fully matched regex, so the `matches` should not be `undefined` and should have all the properties.

On the other hands, I am thinking to add function override on helper functions in `regexes.ts` and `netregexes.ts`, let them return a `matches` with full properties (or `undefined` if not matched?) when `capture` is true or undefined, and return an empty `matches` (or `undefined`) when `capture` is false, which should be suitable with our usage?

quisquous:
> > I'm not sure I understand how these functions are going to be called.
> 
> Because these functions are called with fully matched regex, so the `matches` should not be `undefined` and should have all the properties.

Yeah, I agree that's what happens in practice, but I think the required properties part is much harder to convince TypeScript of.  The non-nullable part is good though, and I think we should just change all of the `MatchesThing` to be not `undefined` right now.  Thanks for that suggestion.

> On the other hands, I am thinking to add function override on helper functions in `regexes.ts` and `netregexes.ts`, let them return a `matches` with full properties (or `undefined` if not matched?) when `capture` is true or undefined, and return an empty `matches` (or `undefined`) when `capture` is false, which should be suitable with our usage?

Even if we had such an override, my concern is that `popup-text.js` or other places have loops that look like this:

```typescript
for (const log of log) {
  for (const trig of triggers) {
    const m = trig.regex.specialMatchingFunction(log);
    if (m)
      trig.alertText(data, m, output);
  }
}
```

Even if we had some special function on `NetRegexes` that returned a particular type of matches with required fields, what type is `m` such that it can apply for all types of these matches with required fields?

quisquous:
Ok, after some discussion with @panicstevenson, I came up with this hack.  I'm not sure what I think, but you can tell me what you think.  We can treat everything as `MatchesAny` inside `popup-text.ts` and then inside trigger files, we can treat them as `MatchesStartsUsing`.  It's all a little bit of a lie, but I think with the lint rule that @panicstevenson to verify that the right type of `MatchesFoo` is called with `netregexFoo` then I think it's probably ok?

Without the `| MatchesAny` there's a lot of type errors because `MatchesAny` and `MatchesStartsUsing` are unrelated.  (We could also make them interfaces and use `extends` but I think it's the same thing.)

```diff
diff --git a/types/trigger.d.ts b/types/trigger.d.ts
index 70ff9d79..c13f7722 100644
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -7,7 +7,7 @@ export interface BaseRegExp<T> extends RegExp {
   };
 }

-export type Matches<Params> = { [s in Params]?: string };
+export type Matches<Params> = { [s in Params]: string } | MatchesAny;

 // TargetedMatches can be used for generic functions in responses or conditions
 // that use matches from any number of Regex or NetRegex functions.

```

What do you think?

PR #2997, opened 2021-06-03
raidemulator: Refactor LineEvent properties to be normalized/consistent

valarnin:
Second PR requested from #2936

### DIFF
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
@@ -65,7 +65,7 @@ export default class LineEvent {
 // Type guards for these interfaces require their own descriptor property
 // because we don't want every line event with an id/name
 // to update combatant state, for example
-export interface LineEventSource {
+export interface LineEventSource extends LineEvent {
   readonly isSource: true;
   readonly id: string;
   readonly name: string;
@@ -80,12 +80,11 @@ export interface LineEventSource {
   readonly maxMp?: number;
 }
 
-export const isLineEventSource =
-  (line: LineEventSource | LineEvent): line is LineEventSource => {
-    return 'isSource' in line;
-  };
+export const isLineEventSource = (line: LineEvent): line is LineEventSource => {
+  return 'isSource' in line;
+};
 
-export interface LineEventTarget {
+export interface LineEventTarget extends LineEvent {
   readonly isTarget: true;
   readonly targetId: string;
   readonly targetName: string;
@@ -99,32 +98,27 @@ export interface LineEventTarget {
   readonly targetMaxMp?: number;
 }
 
-export const isLineEventTarget =
-  (line: LineEventTarget | LineEvent): line is LineEventTarget => {
-    return 'isTarget' in line;
-  };
+export const isLineEventTarget = (line: LineEvent): line is LineEventTarget => {
+  return 'isTarget' in line;
+};
 
-export interface LineEventJobLevel {
+export interface LineEventJobLevel extends LineEvent {
   readonly isJobLevel: true;
-  readonly job?: string;
-  readonly jobId?: number;
-  readonly level?: number;
+  readonly job: string;
+  readonly jobId: number;
+  readonly level: number;
 }
 
-export const isLineEventJobLevel =
-  (line: LineEvent | LineEventSource | LineEventTarget | LineEventJobLevel):
-  line is LineEventJobLevel => {
-    return 'isJobLevel' in line;
-  };
+export const isLineEventJobLevel = (line: LineEvent): line is LineEventJobLevel => {
+  return 'isJobLevel' in line;
+};
 
-export interface LineEventAbility {
+export interface LineEventAbility extends LineEvent {
   readonly isAbility: true;
-  readonly abilityId?: number;
-  readonly abilityName?: string;
+  readonly abilityId: number;
+  readonly abilityName: string;
 }
 
-export const isLineEventAbility =
-  (line: LineEvent | LineEventSource | LineEventTarget | LineEventAbility):
-  line is LineEventAbility => {
-    return 'isAbility' in line;
-  };
+export const isLineEventAbility = (line: LineEvent): line is LineEventAbility => {
+  return 'isAbility' in line;
+};
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
@@ -61,3 +61,70 @@ export default class LineEvent {
       ).toString(16), 16);
   }
 }
+
+// Type guards for these interfaces require their own descriptor property
+// because we don't want every line event with an id/name
+// to update combatant state, for example
+export interface LineEventSource {
+  readonly isSource: true;
+  readonly id: string;
+  readonly name: string;
+  readonly x?: number;
+  readonly y?: number;
+  readonly z?: number;
+  readonly heading?: number;
+  readonly targetable?: boolean;
+  readonly hp?: number;
+  readonly maxHp?: number;
+  readonly mp?: number;
+  readonly maxMp?: number;
+}
+
+export const isLineEventSource =
+  (line: LineEventSource | LineEvent): line is LineEventSource => {
+    return 'isSource' in line;
+  };
+
+export interface LineEventTarget {
+  readonly isTarget: true;
+  readonly targetId: string;
+  readonly targetName: string;
+  readonly targetX?: number;
+  readonly targetY?: number;
+  readonly targetZ?: number;
+  readonly targetHeading?: number;
+  readonly targetHp?: number;
+  readonly targetMaxHp?: number;
+  readonly targetMp?: number;
+  readonly targetMaxMp?: number;
+}
+
+export const isLineEventTarget =
+  (line: LineEventTarget | LineEvent): line is LineEventTarget => {
+    return 'isTarget' in line;
+  };
+
+export interface LineEventJobLevel {
+  readonly isJobLevel: true;
+  readonly job?: string;
+  readonly jobId?: number;
+  readonly level?: number;
+}
+
+export const isLineEventJobLevel =
+  (line: LineEvent | LineEventSource | LineEventTarget | LineEventJobLevel):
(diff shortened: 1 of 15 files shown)


valarnin:
We may want to flag the encounters DB as upgraded and force a re-parse of encounter lines with this one, since it'll cause issues due to some property types being changed?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -61,3 +61,70 @@ export default class LineEvent {
...
+  readonly heading?: number;
+  readonly targetable?: boolean;
+  readonly hp?: number;
+  readonly maxHp?: number;
+  readonly mp?: number;
+  readonly maxMp?: number;
+}
+
+export const isLineEventSource =
+  (line: LineEventSource | LineEvent): line is LineEventSource => {
```
quisquous, line 84:
```suggestion
  (line: LineEvent): line is LineEventSource => {
```

What about this change along with making `LineEventSource` / `LineEventTarget` / etc all extend `LineEvent`?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -61,3 +61,70 @@ export default class LineEvent {
...
+  };
+
+export interface LineEventAbility {
+  readonly isAbility: true;
+  readonly abilityId?: number;
+  readonly abilityName?: string;
+}
+
+export const isLineEventAbility =
+  (line: LineEvent | LineEventSource | LineEventTarget | LineEventAbility):
```
quisquous, line 127: This was a little confusing for me to read.  I understand why it needs to be there, but what do you think about just having `(line: LineEvent)` here instead of this confusing union and solving the Typescript error by having `LineEventAbility extends LineEvent`.  I think that would solve this, and might make `LineEventAbility` a more useful type at the same time.

quisquous (commented):
> We may want to flag the encounters DB as upgraded and force a re-parse of encounter lines with this one, since it'll cause issues due to some property types being changed?

That makes sense.  Did you want to do that here as well?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -61,3 +61,70 @@ export default class LineEvent {
       ).toString(16), 16);
   }
 }
+
+// Type guards for these interfaces require their own descriptor property
+// because we don't want every line event with an id/name
+// to update combatant state, for example
+export interface LineEventSource {
+  readonly isSource: true;
```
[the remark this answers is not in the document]
valarnin: "Source" in this context means the LineEvent object has a source or primary combatant.

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
valarnin: Resolved the one above and we can talk about it on this one.

If each of these extends LineEvent, then how can a LineEvent be both LineEventSource and LineEventTarget? Typescript doesn't support multiple inheritance for classes, only interfaces, and LineEvent is a class.

quisquous: I was trying to suggest that `LineEventSource extends LineEvent` and `LineEventTarget extends LineEvent`.  `LineEvent` itself just stays as a class.

valarnin: Except that this isn't valid syntax:

```typescript
class LineEvent {}

interface LineEventSource extends LineEvent {}
```

An interface can't extend a class.

If it were instead:

```typescript
class LineEvent {}

class LineEventSource extends LineEvent {}
```

Then this is no longer possible because a class can't extend multiple classes:

```typescript
class LineEvent0x1A extends LineEvent, LineEventSource, LineEventJobLevel {}
```

quisquous: https://www.typescriptlang.org/play?strict=true#code/MYGwhgzhAEAyCWA7ApgUQG7MQF2gbwF8BYAKFKW2QCcAzMYZOJNTHAZQHsBXKh6ZAB6VEAExgIUGLLkKlSQA seems to say it's ok? `tsc` also doesn't complain to me when I make that change in your patch.  Can you point me at the documentation for this for my own understanding?

valarnin: I found this page that states that it's possible, but it's marked as deprecated and the newer page doesn't state directly that it's okay?

https://www.typescriptlang.org/docs/handbook/interfaces.html#interfaces-extending-classes

quisquous: "it" here being deprecated is just the page as a whole, but not that feature specifically.  Agreed that the newer page doesn't state it, but I also can't find anything else about it either.  ("Type guards" aren't in the newer documentation either, but maybe I'm just missing them.)

I guess I don't feel that strongly about it.  It seems pretty reasonable to me, but if it seems sketchy to you, then this is fine as-is.

valarnin: Looks like type guards are in the documentation but under a different name?
https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

I'm fine with `interface extends class` but I'm always a bit leery of using something that might be deprecated.

I don't see anything in the typescript 2.0 release notes that would indicate that the syntax/feature would be deprecated, so maybe they just didn't bring it forward in their documentation?

ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -61,3 +61,70 @@ export default class LineEvent {
...
+export const isLineEventJobLevel =
+  (line: LineEvent | LineEventSource | LineEventTarget | LineEventJobLevel):
+  line is LineEventJobLevel => {
+    return 'isJobLevel' in line;
+  };
+
+export interface LineEventAbility {
+  readonly isAbility: true;
+  readonly abilityId?: number;
+  readonly abilityName?: string;
```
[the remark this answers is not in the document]
valarnin: Sorry, the Source and Target interfaces needed the optionals and I just carried the pattern forward to these, but it does make more sense for them to be required. Fixed.

valarnin:
> 
> 
> > We may want to flag the encounters DB as upgraded and force a re-parse of encounter lines with this one, since it'll cause issues due to some property types being changed?
> 
> That makes sense. Did you want to do that here as well?

I think this is the only outstanding question at this point. I'm a bit torn on if this should be handled in this PR, as well as exactly how to handle it.

It's going to be a bit user-unfriendly to go through the entire DB when opening the page and update all the encounters, probably going to be a long browser freeze, but at the same time there's not a good way to flag that the encounter needs updated, except maybe by checking for the relevant properties on the lines when loading?

quisquous:
> > > We may want to flag the encounters DB as upgraded and force a re-parse of encounter lines with this one, since it'll cause issues due to some property types being changed?
> > 
> > 
> > That makes sense. Did you want to do that here as well?
> 
> I think this is the only outstanding question at this point. I'm a bit torn on if this should be handled in this PR, as well as exactly how to handle it.
> 
> It's going to be a bit user-unfriendly to go through the entire DB when opening the page and update all the encounters, probably going to be a long browser freeze, but at the same time there's not a good way to flag that the encounter needs updated, except maybe by checking for the relevant properties on the lines when loading?

Just to clarify (because my memory is bad), the db stores both the original encounter lines and the parsed LineEvent versions?

If so, do encounters have a version? (If no, we could add one.)  We could have a global version number, and if you ever load an encounter that has an old version, we could reparse the LineEvent objects.  That way you wouldn't have to go through the entire DB when the version bumped, and instead you could just pay an extra tax when loading old files.  What about something along those lines?

valarnin:
The DB stores the `Encounter` object, minus `combatantTracker`, plus an auto-incremented `id`, which is a bit of a mess and something I'll likely revisit shortly after moving `Persistor` to typescript.

image

The `logLines` property is `logLines: LineEvent[]`.

LineEvent stores the original log line as `networkLine: string`.

The encounter's logLines can be rebuilt via a fairly simple snippet, and then reparsed, based on existing code for the `Reparse Encounter` button:

```javascript
// Note that the static property comes first here intentionally to deal with `undefined`
if (Encounter.encounterVersion < enc.version) {
  const converter = new NetworkLogConverter();
  enc.logLines = converter.convertLines(
    enc.logLines.map((l) => l.networkLine),
    repo
  );
  enc.initialize();
  await persistor.persistEncounter(enc);
  encounterTab.refresh();
}
```

We could then modify the `Encounter` class something like this:

```javascript
export default class Encounter {
  static readonly encounterVersion = 1;
// ...
  initialize() {
    this.version = Encounter.encounterVersion;
```

And then `Persistor.loadEncounter` like this:

```javascript
          const enc = req.result;
          const ret = new Encounter(enc.encounterDay,
              enc.encounterZoneId,
              enc.encounterZoneName,
              enc.logLines);
          ret.id = enc.id;
          ret.version = enc.version;
          res(ret);
```

quisquous:
Yeah, that's exactly the kind of thing I was thinking too, thanks!

valarnin:
Hmm, I feel like the exact mechanism for upgrading the encounter could be handled better? Maybe an `upgrade` method on `Encounter` that accepts a version, instead?

```javascript
  upgrade(version) {
    if (Encounter.encounterVersion <= version)
      return false;

    const converter = new NetworkLogConverter();
    this.logLines = converter.convertLines(
      this.logLines.map((l) => l.networkLine),
      repo
    );
    this.version = Encounter.encounterVersion;
    this.initialize();

    return true;
  }
```

Then `Persistor` can be updated as such:

```javascript
          ret.id = enc.id;
          if (ret.upgrade(enc.version)) {
            this.persistEncounter(ret).then(() => {
              res(ret);
            });
          } else {
            res(ret);
          }
```

That also means that the version can be set in the constructor (or defaulted when it's moved to typescript) instead of being set in initialize:

```javascript
export default class Encounter {
  constructor(encounterDay, encounterZoneId, encounterZoneName, logLines) {
    this.version = Encounter.encounterVersion;
```

PR #2998, opened 2021-06-03
raidboss: Fix e9s callout not being declared properly

### LINKED ISSUE #2995: e9s some triggers are not working after recent update
## Description

Those triggers alerted after these triggers below are not working.
E9S The Second Art Of Darkness Right
E9S The Second Art Of Darkness Left
E9S The Third Art Of Darkness Right
E9S The Third Art Of Darkness Left

I guess these trigger `E9S The Second / Third Art Of Darkness Charge Solver` and `E9S The Second / Third Art Of Darkness Callout` are not working.

Don't know if this happens on KR server only. I heard other person also had this problem.

I've checked tuesday working well. This happened after the recent update.

valarnin:
Noticed this issue while testing the language fix for raidemulator.

Edit:

Fixes #2995.

### DIFF
--- a/ui/raidboss/data/05-shb/raid/e9s.js
+++ b/ui/raidboss/data/05-shb/raid/e9s.js
@@ -538,7 +538,7 @@ export default {
       durationSeconds: (data) => data.finalArtOfDarkness ? 16 : 9,
       alertText: (data, _matches, output) => {
         // Perform the callout now, regardless if it's The Second or Third Art Of Darkness
-        callout = data.artOfDarkness.slice();
+        const callout = data.artOfDarkness.slice();
         if (data.finalArtOfDarkness)
           callout.push(data.finalArtOfDarkness);
         return callout.map((key) => output[key]()).join(' -> ');


panicstevenson (approved):

ui/raidboss/data/05-shb/raid/e9s.js
```diff
@@ -538,7 +538,7 @@ export default {
       durationSeconds: (data) => data.finalArtOfDarkness ? 16 : 9,
       alertText: (data, _matches, output) => {
         // Perform the callout now, regardless if it's The Second or Third Art Of Darkness
-        callout = data.artOfDarkness.slice();
+        const callout = data.artOfDarkness.slice();
```
quisquous, line 541: Can you explain how this fixes #2995?  It seems like even if `callout` is a global here, that it gets clobbered in this function and so I'm not sure I follow how the behavior changes by adding `const`.

valarnin: Except that it's not a global, it's a local, and it's being used without being declared.

image

Trim21:
our lint should check this.

ui/raidboss/data/05-shb/raid/e9s.js
quisquous: Oh, it's `use strict` that makes this an error.  Got it.

PR #2999, opened 2021-06-03
fix(config): view source fragment

Trim21:


Trim21:
`g` flag in regex just mean replace All

PR #3003, opened 2021-06-04
lint: Add Rule for Mismatched Matches TypeReferences

panicstevenson:


panicstevenson:
For:
```js
{
  id: 'TitanEx Rock Throw',
  netRegex: NetRegexes.startsUsing({ ... }),
  alertText: (data, matches: MatchesTether, output) => { ... },
  outputStrings: { ... },
},
```
```
cactbot\ui\raidboss\data\02-arr\trial\titan-ex.ts
  76:34  error  Function matches type does not match trigger matches type: expected MatchesStartsUsing, found MatchesTether  rulesdir/cactbot-matches-type-references
  80:33  error  Function matches type does not match trigger matches type: expected MatchesStartsUsing, found MatchesTether  rulesdir/cactbot-matches-type-references
```

Result after applying `--fix`:
```js
{
  id: 'TitanEx Rock Throw',
  netRegex: NetRegexes.startsUsing({ ... }),
  alertText: (data, matches: MatchesStartsUsing, output) => { ... },
  outputStrings: { ... },
},
```

eslint/cactbot-matches-type-references.js
```diff
@@ -0,0 +1,72 @@
...
+      recommended: true,
+      url: 'https://github.com/quisquous/cactbot/blob/main/docs/RaidbossGuide.md#trigger-elements',
+    },
+    fixable: 'code',
+    schema: [],
+  },
+
+  create: (context) => {
+    const netRegexTypeByTriggerId = {};
+    const matchesTypeToNetRegexType = {
```
quisquous, line 20: Should this just be implicit? e.g. prepend `Matches` and capitalize the first letter of the NetRegex type?

eslint/cactbot-matches-type-references.js
```diff
@@ -0,0 +1,72 @@
+const findTriggerIdFromSiblingProperty = (node) => {
+  return node.parent.parent.properties.find((property) => property.key.name === 'id').value.value;
```
quisquous, line 2: This is...quite a statement.  I guess there's no way to make this simpler or more readable?

panicstevenson: I could spend the time to say what it's doing, but not really? We're inside of a CallExpression, so you have to go up to get to the Property, and then up once more to get to the whole trigger ObjectLiteral. From there, we find the id and then you have to get the value's value, which is to say the actual content inside the quoted trigger id.

I can make that more explicit here if we feel like that's necessary or useful, sure

Issue #2980, opened 2021-06-01
Potential Logic Error in UCOB - UCU Nael Fireball 4


Datachange:
Hello beautiful people,

Please refer to this trigger:
https://github.com/quisquous/cactbot/blob/triggers/04-sb/ultimate/unending_coil_ultimate.js#L876

On Nael's fourth fireball, if I took three fireballs, I would expect to get an alert text telling me to avoid the 4th fireball. However, in the video below, I did not get such an alert text. (Note: I am overriding that trigger, however, that aspect of it should not be affected.)

However, I believe I understand why this trigger is not working... I think `data.tookThreeFireballs` (the variable which drives the alert text) will never be populated because `preRun()` runs before `delaySeconds()`.

If I understand the fireball triggers correctly, because the target of the tether cannot (currently) be calculated, four triggers listen to an event near the beginning of the phase, and set a delay from there, one for each tether at a predetermined interval. Since, at the time of this initial event, zero fireballs have gone out, and since `preRun()` runs before `delaySeconds()`, I would therefore think that it is impossible for `data.tookThreeFireballs` to be populated at this time.

Video: https://drive.google.com/file/d/1KqBSlkMp_Rw0Jjct9VufqQg3wuBOrXlc/view?usp=sharing
Log: https://drive.google.com/file/d/1v_IBjWQ2p_JixkgNBJGIq83WuB0vgMLN/view?usp=sharing
User file: https://drive.google.com/file/d/1gJxqLv5lL8St-Vl1ZP0io9caw0H14tVW/view?usp=sharing

That's all I got! Thank you very much for your time,
Data

panicstevenson:
Your assessment makes sense, and seems to be right. I used the log you provided (thank you!) in the emulator and it appears to be fixed with the suggested change.

PR #3004, opened 2021-06-04
raidboss: Fix Unending Coil Nael Fireball 4 Callout

panicstevenson:


panicstevenson:
Thanks to @Datachange for finding both the bug and the fix!

Closes https://github.com/quisquous/cactbot/issues/2980.

PR #3007, opened 2021-06-04
ci: label script will match key with lang only

Trim21:
```js
mocha.setup('bdd');

const assert = chai.assert;

/**
 * @param {string} patch
 */
const parseChangedLang = (patch) => {
  const set = new Set();
  for (const lang of ['cn', 'de', 'ja', 'fr', 'ko']) {
    const pattern = new RegExp(String.raw`^\+\s*(?:${lang}|'${lang}'):`);
    for (const line of patch.split('\n')) {
      if (pattern.test(line)) {
        // TODO: it'd be nice to add the file/line number here.
        if (!set.has(lang))
          console.log(`label: ${lang} [translation]`);
        set.add(lang);
      }
    }
  }
  return Array.from(set.values());
};

describe('parseChangedLang', () => {
  it('should match only single word', () => {
    assert.deepEqual(parseChangedLang(`+  de: 'hello',`), ['de']);
  });

  it('should match quoted', () => {
    assert.deepEqual(parseChangedLang(`+  'de': 'hello',`), ['de']);
  });

  it('should match only single word', () => {
    assert.deepEqual(parseChangedLang(`+  // parentNode: { classList: () };`), []);
  });

  it('should match only single word', () => {
    assert.deepEqual(parseChangedLang(`+  parentNode: { classList: () };`), []);
  });
});


mocha.run();
```
https://codepen.io/Trim21/pen/gOmeyGo

### DIFF
--- a/.github/scripts/auto-label.cjs
+++ b/.github/scripts/auto-label.cjs
@@ -156,7 +156,7 @@ const getChangedFiles = async (github, identifier) => {
 const parseChangedLang = (patch) => {
   const set = new Set();
   for (const lang of ['cn', 'de', 'ja', 'fr', 'ko']) {
-    const pattern = new RegExp(`^\s*[+] *(?:${lang}|'${lang}'): `);
+    const pattern = new RegExp(`^\s*[+]\s*(?:${lang}|'${lang}'): `);
     for (const line of patch.split('\n')) {
       if (pattern.test(line)) {
         // TODO: it'd be nice to add the file/line number here.
--- a/.github/scripts/auto-label.cjs
+++ b/.github/scripts/auto-label.cjs
@@ -156,7 +156,7 @@ const getChangedFiles = async (github, identifier) => {
 const parseChangedLang = (patch) => {
   const set = new Set();
   for (const lang of ['cn', 'de', 'ja', 'fr', 'ko']) {
-    const pattern = new RegExp(`^\s*[+].*(?:${lang}|'${lang}'): `);
+    const pattern = new RegExp(`^\s*[+] *(?:${lang}|'${lang}'): `);
     for (const line of patch.split('\n')) {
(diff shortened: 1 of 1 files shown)


Trim21:
should fix the label `de` of https://github.com/quisquous/cactbot/pull/3001

valarnin:
Why use ` ` instead of `\s`? I realize that ` ` should always work due to styling enforcement but syntactically it could be space or tab?

Trim21:
> Why use ` ` instead of `\s`? I realize that ` ` should always work due to styling enforcement but syntactically it could be space or tab?

ok

.github/scripts/auto-label.cjs
```diff
@@ -156,7 +156,7 @@ const getChangedFiles = async (github, identifier) => {
 const parseChangedLang = (patch) => {
   const set = new Set();
   for (const lang of ['cn', 'de', 'ja', 'fr', 'ko']) {
-    const pattern = new RegExp(`^\s*[+].*(?:${lang}|'${lang}'): `);
+    const pattern = new RegExp(String.raw`^\s*[+]\s*(?:${lang}|'${lang}'): `);
```
quisquous, line 159:
```suggestion
    const pattern = /^[+]\s*(?:${lang}|'${lang}'): /;
```

What about just using `/regex/` as a simplification? Also I think I suggested this, but Id on't think there's any whitespace before the initial `+`.

Trim21: not working

valarnin: Maybe it makes sense to add https://eslint.org/docs/rules/prefer-regex-literals ?

quisquous: Oh right sorry, template interpolation.  Sorry, I should have more coffee before making suggestions.  <_<

PR #3009, opened 2021-06-04
extend type of `document.createElement`

Trim21:


resources/timerbar.ts
```diff
@@ -485,7 +485,7 @@ export default class TimerBar extends HTMLElement {
   }
 
   // Apply all styles from an object where keys are CSS properties
-  applyStyles(styles: { [s: string]: string}): void {
+  applyStyles(styles: { [s: string]: string }): void {
```
quisquous, line 488: Thanks! I wonder if there's eslint for this.

MaikoTan: I think it is [@typescript-eslint/object-curly-spacing](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/object-curly-spacing.md)

PR #3010, opened 2021-06-04
Jobs: Trace the cooldown of raidbuffs which not lands on you or misses target

### LINKED ISSUE #664: jobs: buff tracker should track cooldowns of jobs that don't land on you

Echoring:
- **Now buffTracker can trace the cooldown of raidbuffs which not lands on you or misses target (except `Devotion`).**
-- Import `partyTracker` and add a new field of `partyonly`.
-- Split an `onCooldown()` from `onGain()`.
-- Change `gainAbility` to `cooldownAbility`.
-- Refactor `onBigBuff()`, `onUseAbility()` and `onGainEffect()`
-- `Devotion` is too complex. It cooldown at owner use `Aetherpact`, but take effect at pet use `Devotion`. Delay between them are both random and maneuverable, which makes cooldown extremely unreliable. And its source is pet, which is not considered as party member. This make it hard to trace missing `Devotion`. A FIXME is left here.
-- Fixes #664


ui/jobs/buff_tracker.ts
```diff
@@ -294,7 +307,18 @@ export class BuffTracker {
...
         sortKey: 1,
         cooldown: 60,
-        sharesCooldownWith: ['offguard'],
+        sharesCooldownWith: ['offguard', 'peculiarmiss'],
+      },
+      peculiarmiss: {
+        gainAbility: kAbility.PeculiarLight,
+        // durationSeconds: 0 can not trigger the cooldown timer, so use 0.001
+        // you may see an icon flashes by if you gaze at bufftracker.
+        durationSeconds: 0.001,
```
quisquous, line 316: This is a neat idea.  I have a couple of small concerns that maybe can be fixed up.

* I don't like the duplication of `peculiar` and `peculiarmiss` in terms of `borderColor` / `sortKey` / `cooldown` / etc etc.
* I wonder if there's a better way to handle `durationSeconds: 0.001`

Mostly I'm wondering if there's some way to combine `peculiar` and `peculiarmiss` and maybe specially mark the ability in a way that doesn't need `durationSeconds: 0.001`.

Can I propose something like making a new `cooldownAbility` field? So, you'd add `cooldownAbility: kAbility.PeculiarLight` to the `peculiar: {` object above.  That field will never be an active use (so no need for `durationSeconds: 0.001`) and will only start the cooldown.  Maybe something else would have to be fixed up to make this work, but that's one possible solution that I can picture.

Echoring:
I take some fight in Alliance Raid, seems nothing wrong. Is this what you pictured?

quisquous:
> I take some fight in Alliance Raid, seems nothing wrong. Is this what you pictured?

Yeah, this is exactly it.  I wanted to give it a try myself real quick and ran out of time last night.  I'll take a look today, thanks!

ui/jobs/buff_tracker.ts
```diff
@@ -655,7 +700,10 @@ export class BuffTracker {
         existingBuff.clearCooldown(source);
     }
 
-    buff?.onGain(seconds, source);
+    if (option === 'active' && seconds > 0)
+      buff?.onGain(seconds);
+    else if (option === 'cooldown')
+      buff?.onCooldown(seconds, source);
```
quisquous, line 706: If you make the above code:
```typescript
if (!buff)
  buff = this.buffs[name] = new Buff(name, info, list, this.options);
```

Then you don't need `buff?.` here and can just say `buff.`

ui/jobs/buff_tracker.ts
```diff
@@ -199,18 +200,24 @@ export class Buff {
     this.active.removeCallback();
     this.active = null;
   }
+
+  onCooldown(seconds: number, source: string): void {
+    this.clearCooldown(source);
+    this.addCooldown(source, seconds);
```
quisquous, line 206: FWIW, when using Battle Voice on myself, because the ability occurs before the effect, it briefly flashes 180 before 20.  Do you think it's worth changing the `addCooldown` to add like 0.5 or 1 second minimum before showing the cooldown just to avoid this race?  I also have the setting where my minimum time is set to 9999.  (Or is there some other way to avoid this?)

ui/jobs/buff_tracker.ts
```diff
@@ -307,9 +320,11 @@ export class BuffTracker {
         cooldown: 60,
       },
       litany: {
+        cooldownAbility: kAbility.BattleLitany,
         gainEffect: EffectId.BattleLitany,
         loseEffect: EffectId.BattleLitany,
         useEffectDuration: true,
+        partyonly: true,
```
quisquous, line 327: Sorry for the naming nit, but can you make this `partyOnly` since it's two words?

ui/jobs/buff_tracker.ts
```diff
@@ -320,9 +335,12 @@ export class BuffTracker {
         // On each embolden stack changes,
         // there will be a gain effect log with a wrong duration (always 20).
         // So using stack to identify the first log.
+        cooldownAbility: kAbility.Embolden,
         gainEffect: EffectId.Embolden,
         loseEffect: EffectId.Embolden,
-        durationSeconds: 20,
+        useEffectDuration: true,
+        durationSeconds: 20 + 1, // for self tracing use, not actual time
```
quisquous, line 342: Do you mean "tracking" instead of "tracing"?  Even so, I'm not sure I understand the `+1` here, would you mind explaining more?

ui/jobs/buff_tracker.ts
```diff
@@ -485,29 +510,39 @@ export class BuffTracker {
...
         useEffectDuration: true,
+        partyonly: true,
         icon: '../../resources/ffxiv/status/brotherhood.png',
         // Dark Orange.
         borderColor: '#994200',
         sortKey: 11,
         cooldown: 90,
       },
       devotion: {
+        // FIXME: pet is not be considered as inParty, so cannot trace it if missing you.
```
quisquous, line 525:
```suggestion
        // FIXME: pet is not considered inParty, so this cannot track it if it misses you.
```

ui/jobs/buff_tracker.ts
```diff
@@ -485,29 +510,39 @@ export class BuffTracker {
...
+        // By the way, pet can delay using devotion after been ordered
+        // and if you order it to continue moving, it can greatly delay upto 30s,
```
quisquous, line 527:
```suggestion
        // and if you order it to continue moving, it can greatly delay up to 30s,
```

Thanks for the comment.  I think that seems fine to have this limitation on devotion.

ui/jobs/buff_tracker.ts
```diff
@@ -428,33 +445,41 @@ export class BuffTracker {
...
+        useEffectDuration: true,
         icon: '../../resources/ffxiv/status/standard-finish.png',
         // Green.
         borderColor: '#32CD32',
         sortKey: 6,
       },
       technicalFinish: {
+        // Sorry, but 0-3 step Technical Finish will not trigger this if it misses you.
+        // and I think it meaningless to trace a poor DNC
+        // who can perform 0-3 step Technical Finish and miss you at the same time.
```
quisquous, line 457: Haha, that's fair.  I think we could support an array of abilities for `cooldownAbility` (or an array of effects for `gainEffect` and combine embolden/emboldenSelf), and we already have all the ability ids for the other technical finishes.  Just saying we *couuuuld* do it, but agree that "I don't think we should bother" is a fine answer.  :)

ui/jobs/buff_tracker.ts
```diff
@@ -332,16 +350,15 @@ export class BuffTracker {
       },
       emboldenself: {
         // RDM himself gains a different buff.
-        // This makes RDM himself can also track embolden cooldown.
+        // This makes RDM himself can also trace embolden.
```
quisquous, line 353:
```suggestion
        // This makes RDM himself can also see the embolden buff.
```

ui/jobs/buff_tracker.ts
```diff
@@ -604,7 +650,9 @@ export class BuffTracker {
       if ('stack' in b && b.stack !== parseInt(matches?.count ?? '0'))
         return;
 
-      this.onBigBuff(b.name, seconds, b, matches?.source);
+      this.onBigBuff(b.name, seconds, b, matches?.source, 'active');
+      // some cooldown cannot be handled by action, so also cooldown them here.
+      this.onBigBuff(b.name, seconds, b, matches?.source, 'cooldown');
```
quisquous, line 655:
```suggestion
      // Some cooldowns (like potions) have no `cooldownAbility`, so also track them here.
      if (!b.cooldownAbility)
        this.onBigBuff(b.name, seconds, b, matches?.source, 'cooldown');
```

I think it's good to give an example here of which cooldowns aren't being handled.  I think also it's better for cooldowns that have abilities to not call `onBigBuff(... 'cooldown')` twice for them.  I think it technically works in that the second cooldown call will cancel the first, but I think that's a little confusing and could cause a bug in the future, so it'd be better to just call it once.  (Unless I'm misunderstanding and we need to call this even for things that have `cooldownAbility`??)

quisquous (commented):
Thanks!  This worked great! I have a few small comments about naming / comment grammar / tiny logic changes, but overall this is so good.  I've been wanting this feature for a long time!

ui/jobs/buff_tracker.ts
Echoring: Oh, sorry for typo!
This +1 is for RDM himself. Ability log line is earlier than effect log line.
Embolden cooldown first, than Embolden(self) active. So if without +1, user who set minimum time above 100 will see two Embolden icon (one actived for self and one cooldowning for party) at the last second.

ui/jobs/buff_tracker.ts
Echoring: Sounds great.
By the way, `Technical Step` cooldown at start dancing, but take effect on finish, the time between this is variable so it may also not accurate.

ui/jobs/buff_tracker.ts
quisquous: Yeah, I think an array might be better then because it would treat them both as the same buff.

ui/jobs/buff_tracker.ts
quisquous: Ah, that's fair! Maybe it's worth shortening the cooldown by the "usual" time it takes to complete?

ui/jobs/buff_tracker.ts
Echoring: In fact there is an issue.
Cooldown triggered by ability knows no duration. If user set a high minimum time, the will see active buff and cooldowning buff at the same time. To deal with this we should add `durationsecond` for every raidbuffs, and cannot use `useEffectduration`.

ui/jobs/buff_tracker.ts
Echoring: I think it not necessary? If DNC always dance on time and make every step right, cooldown of buff is exactly 120, for cooldown is triggered on finish now.

quisquous: Oh ok, then it's just if they add extra delay then it might not be accurate.

ui/jobs/buff_tracker.ts
quisquous: Oh, I didn't think about that when combining them.  That makes sense!

ui/jobs/buff_tracker.ts
Echoring: Well, I'll add something to ensure `secondsUntilShow` at least 1

ui/jobs/buff_tracker.ts
Echoring: Changed to array. Embolden(self) now will be considered as the same raidbuff.

ui/jobs/buff_tracker.ts
Echoring: Changed to array to handle all kinds of finish.
Though cooldown may still not be accurate if dancing some extra step or less.

ui/jobs/buff_tracker.ts
```diff
@@ -561,8 +589,12 @@ export class BuffTracker {
         }
 
         const map = propToMapMap[prop];
-        map[key] = map[key] || [];
-        map[key]?.push(buff);
+        if (Array.isArray(key)) {
+          key.forEach((k) => map[k] = [buff, ...map[k] ?? []]);
+        } else {
+          map[key] = map[key] || [];
```
quisquous, line 595:
```suggestion
          map[key] ??= [];
```

ui/jobs/buff_tracker.ts
quisquous: What happened to the conditional here? (Did I suggest something incorrect?)

Echoring: Oh seems I misunderstand you? I I thought you means that it makes sense so leave it be...

ui/jobs/buff_tracker.ts
MaikoTan: Dang!
I didn't even know there is such an operator make it much clear!

ui/jobs/buff_tracker.ts
Echoring: Well, conditional added.
`durationSeconds` has been added back to prevent user with high minimum time seeing duplicated icon of active and cooldown.
But if you miss the buff, cooldown will appear after normal duration end, not at once.
(Hope that not too many user set 9999)

quisquous: Oops, sorry, I guess I'm saying that it might cause a problem in the future if buffs that have both an effect and a cooldown ability call onBigBuff('cooldown') twice.  I was trying to say that it made more sense (to me?) to only call 'cooldown' once, either on the ability or on the effect if there is no cooldown ability.  That's why I added that conditional in the suggested code.  Sorry if I didn't explain myself well enough.

PR #3014, opened 2021-06-04
refactor fetch_deps to TypeScript

Trim21:


Trim21:
I'd like to refactor code related to coinach but I can't run its cli tool on my computer.

util/fetch_deps.ts
```diff
@@ -0,0 +1,244 @@
+/* eslint-disable @typescript-eslint/no-non-null-assertion */
```
quisquous, line 1: Is it possible to not disable this? Or, were you planning to follow up with a different PR to change this?

Trim21: Oh, I forget to push commit that this has been removed

Trim21:
@quisquous ping

Trim21:
anything block this？

quisquous:
> anything block this？

Sorry nope, just didn't quite have enough brain to look at this yet.  I was just about to look right now.  Sorry for the delay.

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+  const depsPath = path.join(projectRoot, 'util/DEPS.json5');
+
+  const deps: Record<string, Meta> = json5.parse(fs.readFileSync(depsPath).toString());
+
+  let cache: Record<string, Meta> = {};
+  const cachePath = path.join(projectRoot, 'util/DEPS.cache');
+  const dlPath = path.join(projectRoot, 'util/.deps_dl');
+
+  if (isFile(cachePath))
+    cache = JSON.parse(fs.readFileSync(cachePath).toString()) as Record<string, Meta>;
```
quisquous, line 80: Can this be `json5.parse` as well (with no `as`)?

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+        cache[key] = meta;
+      }
+    }
+    if (obsolete.size) {
+      console.log('Removing old dependencies...');
+      const count = obsolete.size;
+
+      Array.from(obsolete.values()).forEach((key, i) => {
+        console.log(`[${pad(i + 1)}/${pad(count)}]: ${key}`);
+        const meta = cache[key] as Meta;
```
quisquous, line 162:
```suggestion
        const meta = cache[key];
        if (!meta)
          return;
```

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  for (const [key, meta] of Object.entries(deps)) {
+    if (!isDir(path.join(projectRoot, meta['dest'])))
+      missing.add(key);
+    else if (
+      ('hash' in meta) &&
+      (key in oldCache) &&
+      _.isEqual(_.get(cache, [key, 'hash'], [null, null]), meta['hash'][1])
+    )
+      outdated.add(key);
```
quisquous, line 97: I think `_.isEqual(_.get(cache, [key, 'hash'], [null, null]), meta['hash'][1])` is incorrect.  `_.get(cache, [key, 'hash'], [null, null])` is a `[algorithm, hash]` array and `meta['hash'][1]` is just the `hash`.  Additionally, I think you're looking for not equal?

Can I suggest for clarity:

```suggestion
    if (!isDir(path.join(projectRoot, meta['dest'])))
      missing.add(key);
    else if (oldCache.has(key) && cache[key]?.['hash'][1] !== meta['hash'][1])
      outdated.add(key);
```

I think this would also let you drop lodash as a dependency in this file, as `_has` and `_unset` seem pretty straightforward to write without lodash.

I think I'm not super keen on adding additional (logical) dependencies.  I agree that it is already an implicit npm dependency, but that doesn't mean that it is used in cactbot either other than one you just added.  It's still an additional knowledge burden.

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+  } finally {
+    console.log('Saving dependency cache...');
+    fs.writeFileSync(cachePath, JSON.stringify(cache, null, 2));
+
+    if (repMap) {
+      console.log('Updating hashes...');
+      console.log(repMap);
+      const data = fs.readFileSync(depsPath).toString();
+
+      for (const [old, pyNew] of Object.entries(repMap))
```
quisquous, line 181: `pyNew`?

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+        const i = Array.from(tmp.values()).indexOf(key);
+        // for (const [i, key] of enumerate(missing | outdated)) {
+        console.log(`[${pad(i + 1)}/${pad(count)}]: ${key}`);
+        const meta = deps[key];
+        if (_.isEmpty(meta) || !meta)
+          continue;
+        const baseFileName = path.basename(meta['url']).split('.', 1)[0];
+        if (!baseFileName)
+          // this value won't be null, not TypeScript require this
+          return;
```
quisquous, line 126:
```suggestion
        const baseFileName = path.basename(meta['url']).split('.', 1)[0] ?? '';
```

You could always do this?

Alternatively:

```typescript
if (!baseFileName)
  throw new UnreachableCode();
```

...is what we've done elsewhere in cases where it's like "I know this is correct, but I need to convince TypeScript".

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+        const dlname = path.join(dlPath, baseFileName);
+        const dest = path.join(projectRoot, meta['dest']);
+        // TODO: replace with http client in nodejs
```
quisquous, line 129: :+1:

util/fetch_deps.ts
```diff
@@ -0,0 +1,229 @@
...
+import tar from 'tar-fs';
+import JSZip from 'jszip';
+import json5 from 'json5';
+
+
+type Meta = {
+  'url': string;
+  'dest': string;
+  'strip': number;
+  'hash': [string, string];
```
quisquous, line 18:
```suggestion
  'hash': [algorithm: string, hash: string];
```

quisquous:
> I'd like to refactor code related to coinach but I can't run its cli tool on my computer.

What's wrong with it?

util/fetch_deps.ts
```diff
@@ -0,0 +1,218 @@
...
+  const repMap: { [key: string]: string } = {};
+
+  try {
+    const tmp = new Set([...missing, ...outdated]);
+
+    const count = tmp.size;
+    if (count) {
+      console.log('Fetching missing or outdated dependencies...');
+
+      for (const [i, key] of Array.from(tmp.values()).entries()) {
```
Trim21, line 103: todo: this could be parallelized, but not that important.

quisquous, line 103: Could this be simplified to this?

```suggestion
      Array.from(tmp).forEach((key, i) => {
```

(plus the closing paren)

Trim21: No, there are some await async in loops.

quisquous: Ok.  I guess `Array.from(tmp).forEach(async (key, i) => {` is about as long/complicated as what's there.  I think I'm just used to seeing `forEach` when you need the index and not `entries()` destructuring.

PR #3015, opened 2021-06-05
Jobs: Auto remove readied raidbuffs that not be used in 3min

Echoring:
- **Auto remove readied raidbuffs that not be used in 3min.** 
If a readied raidbuff not be used in 3min, we can assume that this player has left the battlefield, or that is a DNC who no longer dance with you, or a DRG who no longer consider you as his left eye, or at least his raidbuff is unexpectable and not worth tracing. 
-- This usually happen in Bozjan/Eureka that a lot of NIN cast `Trick Attack` and leave the battlefield, leaving a lot of `Trick Attack ready` on your `buffTracker`, preventing you to read the coming cooldown of other raidbuffs.

### DIFF
--- a/ui/jobs/buff_tracker.ts
+++ b/ui/jobs/buff_tracker.ts
@@ -104,6 +104,12 @@ export class Buff {
     const readyKey = 'r:' + this.name + ':' + source;
     this.ready[source] = this.makeAura(readyKey, this.readyList, -1, 0,
         this.readySortKeyBase, color, txt, 0.6);
+
+    // if a readied raidbuff not be used in 3min, we can assume that
+    // this player has left the battlefield, or at least his raidbuff is unexpectable.
+    setTimeout(() => {
+      this.ready[source]?.removeCallback();
+    }, 3 * 60 * 1000);
   }
 
   makeAura(


ui/jobs/buff_tracker.ts
```diff
@@ -104,6 +104,12 @@ export class Buff {
     const readyKey = 'r:' + this.name + ':' + source;
     this.ready[source] = this.makeAura(readyKey, this.readyList, -1, 0,
         this.readySortKeyBase, color, txt, 0.6);
+
+    // if a readied raidbuff not be used in 3min, we can assume that
+    // this player has left the battlefield, or at least his raidbuff is unexpectable.
+    setTimeout(() => {
+      this.ready[source]?.removeCallback();
+    }, 3 * 60 * 1000);
```
quisquous, line 112: I think some raid buffs have 3 minute cooldowns (and sometimes you hold that three minute cooldown for a bit to line up with other buffs).  So, I think either the time needs to be extended to like 5-6 minutes or possibly some other way to approach this could be better?

One option might be to only clear cooldowns for zones that are not raids and trials? (There's a ZoneInfo/ContentType to tell you this).  Another option is to not show cooldowns at all in bozja.  Do either of these options sound better?

Echoring: If you hold a **readied** 3 minutes raidbuff for 3 minutes, you will absolutely lose at least one time of possible use, and I think one should never do it. I think 3 minutes is enough. I've not remembered a battle can have downtime more than 3 minutes, even so you will know that everything is ready like opener.
There's also situation that one member abondon duty in raids and another join party in progress, so I think zoneinfo is not a good choice.

quisquous: Oh, "readied".  Hmm!  That's true.  Only really potions at that point start being something you could hold for 3 minutes and still be worth it.

I guess I was trying to differentiate if you felt like this was an issue in raiding content or only in bozja/eureka/open world.  It sounds like you think this would be helpful everywhere though.  That sounds good!

PR #3016, opened 2021-06-05
Jobs: Fix missing BLU attenuation debuff and DRG RightEye effect ID out of date

Echoring:
- **Fix missing BLU attenuation debuff**
`astralAttenuationFire` and `umbralAttenuationIce` is missing.
- **Fix DRG RightEye effect ID out of date**
This effect has been changed to a new ID at 5.0


ui/jobs/buff_tracker.ts
```diff
@@ -252,6 +252,14 @@ export class BuffTracker {
         borderColor: '#e0cb5c',
         sortKey: 0,
       },
+      astralAttenuationFire: {
```
quisquous, line 255: The condensed libra only has three states.  There's physical, there's fire/wind/lightning, and there's water/earth/ice.  The game has a single icon for each of these.  I couldn't come up with a good single icon to represent them here in the buff tracker, and so instead picked two out of the three.  I played with three for a while and I felt like all three buffs took up a lot of space, and so I picked two instead as a compromise.

I think there's two decisions blu mages make when playing around the libra buff:
(1) is it time to final sting, keep libra-ing until physical debuff appears
(2) pick which 220 potency filler spell to use to match the debuff

And #2 doesn't need to know all three attenuations, it just needs to know which of three libra states it's in.  Even if they're using Blaze, they can see "oh it's the lightning/wind one so I'll use Blaze".  (And arguably nobody has room for 220 filler spells anymore at level 70 anyway these days.)

Another option is coming up with a better single icon here that represents all three states?

Echoring: I tried to compress three icon into one. How do you think of this?
astralumbral

PR #3017, opened 2021-06-05
raidboss: Add Transition on TimerBar Removed

panicstevenson:


panicstevenson:
Adds a transition to fade timerbar out on removal.

https://user-images.githubusercontent.com/8849608/120900872-3f4e9380-c5ec-11eb-862d-53c13a16e158.mp4

Can be extended to be turned off or changed as the user sees fit.

ui/raidboss/timeline.ts
```diff
@@ -883,7 +883,7 @@ export class TimelineUI {
 
     this.timerlist = document.getElementById('timeline');
     if (this.timerlist)
-      this.timerlist.style.gridTemplateRows = `repeat(${this.options.MaxNumberOfTimerBars}, 1fr)`;
+      this.timerlist.style.gridTemplateRows = `repeat(${this.options.MaxNumberOfTimerBars}, min-content)`;
```
quisquous, line 886: Was this the cause of the truncating you were looking at?

ui/raidboss/timeline.ts
```diff
@@ -1019,10 +1019,23 @@ export class TimelineUI {
...
+    if (!bar)
+      return;
+
+    const div = bar.parentNode;
+    const removeBar = () => {
       div?.parentNode?.removeChild(div);
       delete this.activeBars[e.id];
+    };
+    const element = document.getElementById(e.id.toString());
+    if (element) {
```
quisquous, line 1031: (This could be `if (!element) return` too.)

ui/raidboss/timeline.ts
panicstevenson: Since grid rows and grid columns can't be used as transition fields, you have to use width, min/max-width, height, or min/max-height as the transition field for the elements in the grid. With this as 1fr, changing the height on the top element (as the animation for collapsing the bar) won't affect the other bars, and it will simply do the animation and then "blink" the 2nd bar into the top bar slot (as is currently happening). By setting this to min-content, the grid rows are more flexible and will take on changes to height. By changing the removed bar's height, the other bars will "slide" into position.

PR #3020, opened 2021-06-06
lint: apply sort-keys to .eslintrc.cjs

quisquous:
I am not sure what I think about this.  I feel like the eslint file is getting really big and things are added all over the place.  Forcing the keys to be sorted keeps it in a more reasonable order, but also makes it slightly less readable?  I am not sure.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -7,7 +7,7 @@ const path = require('path');
 const rulesDirPlugin = require('eslint-plugin-rulesdir');
 rulesDirPlugin.RULES_DIR = path.join(__dirname, 'eslint');
 
-module.exports = {
+const settings = {
   'env': {
     'browser': true,
     'es6': true,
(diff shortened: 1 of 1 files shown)


valarnin:
It might make more sense to have top-level keys be unsorted, but subkeys be sorted?

quisquous:
Is there a way to convince the eslint rule to do that?

valarnin:
Hmm, doesn't look like it with the default implementation, we'd need to extend it or something.

quisquous:
I don't think the top-level rule ordering is *that* bad? It was kind of nice to have the larger rules/overrides entries at the bottom, but I don't know that there's really an expected order there either.

MaikoTan:
Since it is a `JS` file, what about splitting some large sub-objects into some separated variables?

valarnin:
> 
> 
> Since it is a `JS` file, what about splitting some large sub-objects into some separated variables?

Exactly what I was about to suggest, I think maybe the overrides section would benefit from, e.g., having `const tsOverrides = ...;` as its own object then pulled into the export?

quisquous:
Separated variables make this a million times better, thanks!

PR #3027, opened 2021-06-06
raidemulator: Convert Combatant, CombatantTracker to typescript

valarnin:
Last PR broken out from #2936

My only real thought here is that maybe `Partial<CombatantState>` should have been an exported type from `CombatantState.ts` instead of repeating it so much?


quisquous:
Suggestions to remove type assertions got out of hand, so here's a patch (sorry):
[0001-remove-type-assertions.zip](https://github.com/quisquous/cactbot/files/6604722/0001-remove-type-assertions.zip)

It's untested so you probably want to give it a review yourself and not trust me.

Mostly I am a little bit averse to `as number` or `as Combatant` in place of `!== undefined` because it's a really big hammer.  I think in this case it's fine (right now), because you're turning `number | undefined` into `number`.  However, I think it in the future if some type changes, then code using type assertions will silently be broken and not generate the errors you would hope would be generated.

I agree that TypeScript is a bit obnoxious around understanding if fields in objects or entries in arrays are defined or not even if you've checked them.  Usually temporary named variables goes a long way towards convincing it.

quisquous:
> Last PR broken out from #2936
> 
> My only real thought here is that maybe `Partial<CombatantState>` should have been an exported type from `CombatantState.ts` instead of repeating it so much?

What else would you call it? I'd be half tempted to name it PartialCombatantState which doesn't really make it that much easier to type.  Would you rename CombatantState to FullCombatantState (or something else)?

valarnin:
The patch looks good in general. About to do some testing with it.

I'll add some comments to the `nextSignificantState` and maybe some other spots that might be confusing to look at in 6 months when it's not fresh in my mind.

PR #3029, opened 2021-06-06
raidemulator: Convert Encounter to typescript 

valarnin:
This PR also fixes some potential issues with downstream classes referencing `engageAt` instead of `initialOffset`, thus causing the 'fight begins' bar of the top progress bar to not be set properly if a countdown wasn't used.

Had to use `x as Lang` in two spots on this one, which I'm not really happy with, but I don't see how else to handle those two cases?

I plan on doing a pass over raidemulator after all the typescript conversion is done, and any time there are private properties on classes, adding proper modifiers for the properties.


valarnin:
Also, apparently vscode isn't running eslint again for some reason?

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -1,56 +1,78 @@
...
 import PetNamesByLang from '../../../../resources/pet_names';
 import EmulatorCommon from '../EmulatorCommon';
 import LogRepository from './network_log_converter/LogRepository';
 import NetworkLogConverter from './NetworkLogConverter';
+import { Lang } from 'types/global';
+import LineEvent, { isLineEventSource, isLineEventTarget } from './network_log_converter/LineEvent';
+import { UnreachableCode } from '../../../../resources/not_reached';
 
-const isPetName = (name, language = undefined) => {
+const isPetName = (name: string, language: Lang | undefined = undefined) => {
```
quisquous, line 10:
```suggestion
const isPetName = (name: string, language?: Lang) => {
```

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -1,56 +1,78 @@
...
   if (language)
     return PetNamesByLang[language].includes(name);
 
   for (const lang in PetNamesByLang) {
-    if (PetNamesByLang[lang].includes(name))
+    if (PetNamesByLang[lang as Lang].includes(name))
```
quisquous, line 15: We could also add a helper function somewhere that converts `string` to `Lang` (and throws otherwise) for cases where you know a string is a lang?

Alternatively, something like this seems better, as it doesn't assume the key type, just converts from string to whatever the key type is:

```typescript
for (const langStr in PetNamesByLang)
  const lang = langStr as keyof typeof PetNamesByLang;
```

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -1,56 +1,78 @@
...
       return true;
   }
 
   return false;
 };
 
+const validTimestamp = (timestamp: number) => {
```
quisquous, line 22: tiny naming nit, but maybe add a verb here like `isValidTimestamp`?

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -1,56 +1,78 @@
...
-    this.firstEnemyAbility = Number.MAX_SAFE_INTEGER;
-
-    this.firstLineIndex = 0;
+  initialize(): void {
+    const startStatuses = new Set<string>();
 
     for (let i = 0; i < this.logLines.length; ++i) {
       const line = this.logLines[i];
+      if (!line)
+        throw new UnreachableCode();
```
quisquous, line 58: For this style of for loop where you also need the index, you could also consider something like this to appease TypeScript:

```typescript
this.logLines.forEach((line, i) => {
```

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -68,46 +90,38 @@ export default class Encounter {
         }
       }
       if (res && res.groups && res.groups.language)
-        this.language = res.groups.language || this.language;
+        this.language = res.groups.language as Lang || this.language;
```
quisquous, line 93:
```suggestion
        this.language = res.groups.language as Lang ?? this.language;
```

Maybe this is a case for a helper type guard function to turn string into Lang?

valarnin: Resolving this one in favor of the other comment thread.

ui/raidboss/emulator/data/Encounter.ts
valarnin: I think a typeguard would work fine here, but it should probably live in `global.d.ts`?

quisquous: Sounds good!

valarnin: You want that in this PR, or you want a separate PR and then a rebase for this one?

quisquous: I think this PR is fine.  It's really only a couple of lines.

valarnin:
Hmm, it didn't seem to like adding the `isLang` to global, and I'm not sure why.

```
ERROR in ./types/global.d.ts
Module build failed (from ./node_modules/ts-loader/index.js):
Error: TypeScript emitted no output for C:\Users\valar\Source\cactbot2\types\global.d.ts.
    at makeSourceMapAndFinish (C:\Users\valar\Source\cactbot2\node_modules\ts-loader\dist\index.js:53:18)
    at successLoader (C:\Users\valar\Source\cactbot2\node_modules\ts-loader\dist\index.js:40:5)
    at Object.loader (C:\Users\valar\Source\cactbot2\node_modules\ts-loader\dist\index.js:23:5)
```

quisquous:
I think it's something about definition files being special and only having types in them, so exporting a concrete function is causing an issue.  I'm not quite sure where this would go.  `resources/util.ts` wouldn't be unreasonable?

valarnin:
I can move it there but then `global.d.ts` would need to import from `resources/util.ts` to pick up the common `langs` array, or actual langs would have to be listed in two locations?

quisquous:
Maybe something like this in `util.ts`?  This is the least worst thing I could come up with.

```typescript
// This is a repeat of `Lang`'s definition from global.d.ts, but
// uses a Record to verify that it is identical.  This lets
// global.d.ts export the type and lets us use Lang from other
// definition files, but also use it as an array here.
const langRecord: Record<Lang, null> = {
  en: null,
  de: null,
  fr: null,
  ja: null,
  cn: null,
  ko: null,
};
const langs = Object.keys(langRecord);

const Util = {
  isLang: (lang: string): lang is Lang => {
    return langs.includes(lang);
  },
... etc
```

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -67,47 +90,39 @@ export default class Encounter {
           }
         }
       }
-      if (res && res.groups && res.groups.language)
+      if (res && res.groups && res.groups.language && isLang(res.groups.language))
         this.language = res.groups.language || this.language;
```
quisquous, line 94: And maybe something more like this?

```suggestion
      const matchedLang = res?.groups?.language;
      if (matchedLang && Util.isLang(matchedLang))
        this.language = matchedLang;
```

valarnin:
Is there a reason not to just move the entire `Lang` typing out to its own `.ts` file? I'd rather do this the most correct way, whatever that is, and I'm fine with opening a separate PR for it if needed.

Also, `isLang` doesn't really seem to be a good fit for `Util` in general?

quisquous:
> Is there a reason not to just move the entire `Lang` typing out to its own `.ts` file? I'd rather do this the most correct way, whatever that is, and I'm fine with opening a separate PR for it if needed.

I think the other definition files can't import a non-definition file, so it's useful to for LocaleText etc to have Lang defined inside a definition file.  That's why I proposed what I did.  If you can make something else work that you think is better, feel free to put a patch up?

> Also, `isLang` doesn't really seem to be a good fit for `Util` in general?

Also fair, but `Util` is still called `Util` and not `JobUtil` so it's not that out of place either?  The other option is to put this one function in its own file?

valarnin:
> I think the other definition files can't import a non-definition file

`./types/data.d.ts` pulls `import PartyTracker from '../resources/party';` which is not a .d.ts file, so this seems to be fine?

valarnin:
Alright, I think this should be sorted out now, hopefully the rebase didn't break anything.

valarnin:
Testing looks good, no issues from the rebase and changes.

quisquous:
Thanks for sorting out the `isLang` stuff! That'll be useful elsewhere for sure.

PR #3031, opened 2021-06-07
build: Use Webpack v5 Native TerserPlugin

panicstevenson:
Webpack v5 has a native TerserPlugin. Creating a new one led to an
increase in build times by nearly 50%.

### DIFF
--- a/webpack/webpack.config.cjs
+++ b/webpack/webpack.config.cjs
@@ -154,6 +154,7 @@ module.exports = function(env, argv) {
     optimization: {
       minimize: true,
       minimizer: [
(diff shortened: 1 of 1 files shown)


webpack/webpack.config.cjs
```diff
@@ -155,7 +154,7 @@ module.exports = function(env, argv) {
     optimization: {
       minimize: true,
       minimizer: [
-        new TerserWebpackPlugin({
+        () => ({
```
quisquous, line 157: Do you think it's worth a comment here about why this is not what the previous code was?

PR #3034, opened 2021-06-07
util: Move `Lang` to its own file, add `isLang` typeguard

valarnin:
As discussed in #3029, I think this is better served as its own PR.

I think the `lang as Lang` typecast here is fine, since it's a typeguard check against an `Array.includes`?


util/translate_timeline.ts
```diff
@@ -4,7 +4,7 @@ import { ArgumentParser } from 'argparse';
 import { findMissing } from './find_missing_timeline_translations';
 import { Event, Sync, Timeline } from '../ui/raidboss/timeline';
 import { walkDirSync } from './file_utils';
-import { Lang } from 'types/global';
+import { Lang } from 'resources/languages';
```
MaikoTan, line 7: I think should this be a relative path like 

```suggestion
import { Lang } from '../resources/languages';
```

valarnin: I tried to keep the relative vs absolute paths the same as they were before. I can adjust them if needed?

resources/languages.ts
```diff
@@ -0,0 +1,7 @@
+export const Languages = ['en', 'de', 'fr', 'ja', 'cn', 'ko'] as const;
+
+export type Lang = typeof Languages[number];
+
+export type NonEnLang = Exclude<Lang, 'en'>;
+
+export const isLang = (lang?: string): lang is Lang => Languages.includes(lang as Lang);
```
quisquous, line 7:
```suggestion
export const isLang = (lang?: string): lang is Lang => Languages.includes(lang);
```

I think this is unnecessary.  `Languages` is a `string[]` and `lang` is a `string`.

util/translate_timeline.ts
quisquous: I think this is fine to keep for now, but I think it'd be good to go back and be consistent (and eslint) in a followup PR?

resources/languages.ts
quisquous: Also I thought I typed this in but maybe I edited it away accidentally.  Can `Languages` be `languages` instead? I know cactbot is not consistent with capitalization, but ideally variables should be `camelCase`, and I'd like to get there eventually.

valarnin: I wasn't sure if exported variables should be capitalized like classes, or if they should be camelCase?

util/translate_timeline.ts
valarnin: There are spots all over the place that would need to be updated for an eslint to pass, I think. I can work on a PR for it, looks like 61 cases in 26 files according to a quick regex search. Sounds like something for me to do during work meetings tomorrow.

Trim21: Is there an eslint rule that could prevent this?

MaikoTan: I think `no-restricted-imports` with a pattern that only allows path starts with "./" or "../" would help?

resources/languages.ts
valarnin: Renamed, made the adjustment as well, and tested to make sure it works since I had some weird issues with a slightly different approach that was along the same vein.

util/translate_timeline.ts
valarnin: This should work, though I'll have to test it at some point.

```js
"no-restricted-imports": ["error", {
    "patterns": ["!./*"]
}]
```

Trim21: no, not working

valarnin: I've got a working version but it needs things excluded manually if they're globals or whatever. I've excluded the core ts imports, for example, and it's still giving:

```
C:\Users\valar\Source\cactbot2\util\process_triggers_folder.ts
  2:1  error  'eslint' import is restricted from being used by a pattern                           no-restricted-imports
  5:1  error  'process' import is restricted from being used by a pattern                          no-restricted-imports
  6:1  error  './file_utils' import is restricted from being used by a pattern                     no-restricted-imports
  7:1  error  '../resources/user_config' import is restricted from being used by a pattern         no-restricted-imports
  8:1  error  '../ui/raidboss/raidboss_options' import is restricted from being used by a pattern  no-restricted-imports
```

The rule at this point is already getting fairly big.

```js
  'no-restricted-imports': [
    'error',
    {
      'patterns': [
        // Include all imports
        '*',
        // Exclude relative imports
        '!./*', '!../*',
        // Exclude core imports
        '!assert',
        '!buffer',
        '!child_process',
        '!cluster',
        '!crypto',
        '!dgram',
        '!dns',
        '!domain',
        '!events',
        '!freelist',
        '!fs',
        '!http',
        '!https',
        '!module',
        '!net',
        '!os',
        '!path',
        '!punycode',
        '!querystring',
        '!readline',
        '!repl',
        '!smalloc',
        '!stream',
        '!string_decoder',
        '!sys',
        '!timers',
        '!tls',
        '!tracing',
        '!tty',
        '!url',
        '!util',
        '!vm',
        '!zlib'
      ],
    }
  ],
```

valarnin: (I realize that it's not 100% working yet, I'm still tweaking, this is just an example.)

valarnin: @quisquous I don't think this should hold up this PR, the discussion can keep happening here or be moved to an issue?

PR #3036, opened 2021-06-07
raidboss: Fix issue with Dun Scaith Ferdiad Hollow trigger

### LINKED ISSUE #3033: Dun Scaith: Ferdiad Hollow Wormhole callouts' Atomos colors are wrong about half of the time
## Description

I've been running Dun Scaith a few times in the past few days, and I've noticed that the callouts for Ferdiad Hollow's Wormhole (where the blue and yellow Atomos warp either round or donut AoE markers around) seems to be wrong in terms of the Atomos colors about half of the time, give or take a little. I haven't figured out any specific pattern as to when it'll be right or wrong, and seemingly it can happen on all possible variations of the mechanic. I think it does always get the donut vs circle right but the color seems more or less randomly successful.

## Additional information

For example:

Single donut calling the color right:
right
And the same thing but wrong (should be "go to untethered blue"):
wrong

[Log of run in question](https://gist.github.com/Kryomancer/8453a2d00656765fab75a1925a99820f#file-splitnetwork_20907_20210607-2021-06-07t17-24-44-171z-txt). I'm sorry I can't point out which specific instances of the mech corresponds to the images in question so I'm not sure how helpful this is.

ACT version 3.5.0, release 273
FFXIV act plugin version : 2.0.9.7
Overlay plugin NGLD version: 0.16.6.0
Cactbot version: 0.22.10.0

(I'm fully aware this isn't particularly relevant content nor a difficult mechanic, so the priority of this is not that high. It's more of a "would be nice if" -thing.)

valarnin:
Fixes #3033

### DIFF
--- a/ui/raidboss/data/03-hw/alliance/dun_scaith.js
+++ b/ui/raidboss/data/03-hw/alliance/dun_scaith.js
@@ -192,7 +192,8 @@ export default {
       run: (data, matches) => {
         data.cursing = data.cursing || [];
         data.wailing = data.wailing || [];
-        matches.npcNameId === '5510' ? data.wailing.push(matches.id) : data.cursing.push(matches.id);
+        const id = matches.id.toUpperCase();
+        matches.npcNameId === '5510' ? data.wailing.push(id) : data.cursing.push(id);
       },
     },
     {


valarnin:
A fairly exhaustive search through raidboss triggers (`matches\.id`) doesn't turn up any other spots where it looks like this might be an issue, as far as I was able to see.

Most other hits for that search were for headmarkers or for IDs that were already being uppercase'd or were being parseInt'd, with the exception of `ruby_weapon-ex.js`'s `RubyEx Raven's Image` and `RubyEx Change of Heart` triggers, which don't seem to do anything with those IDs.

quisquous:
> A fairly exhaustive search through raidboss triggers (`matches\.id`) doesn't turn up any other spots where it looks like this might be an issue, as far as I was able to see.

Thanks for looking!

> Most other hits for that search were for headmarkers or for IDs that were already being uppercase'd or were being parseInt'd, with the exception of `ruby_weapon-ex.js`'s `RubyEx Raven's Image` and `RubyEx Change of Heart` triggers, which don't seem to do anything with those IDs.

Ironically I have an Ruby Weapon trigger improvement patch that is almost ready to submit.  <_<

PR #3047, opened 2021-06-08
raidboss: Fix some triggers using `matches.duration` instead of `matches.castTime`

valarnin:
Noticed when testing performance in alliance raids, there were some triggers using `matches.duration` instead of `matches.castTime`. Checked with this regex and found another case:

```
NetRegexes.startsUsing.*?\n(?:      [^ ].*\n)*      delaySeconds.*duration
```

I checked any other cases of `\.duration[^a-zA-Z]` in triggers and it looks like these are the only cases?

### DIFF
--- a/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
+++ b/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
@@ -163,7 +163,7 @@ export default {
       netRegexDe: NetRegexes.startsUsing({ id: '5EB1', source: 'Herzbube' }),
       netRegexFr: NetRegexes.startsUsing({ id: '5EB1', source: 'Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '5EB1', source: 'ジャック' }),
-      delaySeconds: (_data, matches) => matches.duration - 6,
+      delaySeconds: (_data, matches) => matches.castTime - 6,
       durationSeconds: 5,
       infoText: (_data, _matches, output) => output.text(),
       outputStrings: {
@@ -184,7 +184,7 @@ export default {
       netRegexFr: NetRegexes.startsUsing({ id: '5EB1', source: 'Réplique De Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '5EB1', source: '複製サレタジャック' }),
       condition: (data) => !data.cloneLunge,
-      delaySeconds: (_data, matches) => matches.duration - 6,
+      delaySeconds: (_data, matches) => matches.castTime - 6,
       durationSeconds: 5,
       infoText: (_data, _matches, output) => output.text(),
       run: (data) => data.cloneLunge = true,
@@ -206,7 +206,7 @@ export default {
       netRegexFr: NetRegexes.startsUsing({ id: '60C7', source: 'Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '60C7', source: 'ジャック' }),
       // Half a second longer cast time than the Lunge itself
-      delaySeconds: (_data, matches) => matches.duration - 6.5,
+      delaySeconds: (_data, matches) => matches.castTime - 6.5,
       durationSeconds: 5,
       alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -163,7 +163,7 @@ export default {
       netRegexDe: NetRegexes.startsUsing({ id: '5EB1', source: 'Herzbube' }),
       netRegexFr: NetRegexes.startsUsing({ id: '5EB1', source: 'Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '5EB1', source: 'ジャック' }),
-      delaySeconds: (_data, matches) => matches.duration - 6,
+      delaySeconds: (_data, matches) => matches.castTime - 6,
```
quisquous, line 166: Knockbacks I usually only give 5 seconds for, since the knockback ability usually is a different (uncasted) ability than the cast that causes it.

```suggestion
      delaySeconds: (_data, matches) => parseFloat(matches.castTime) - 5,
```

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -184,7 +184,7 @@ export default {
       netRegexFr: NetRegexes.startsUsing({ id: '5EB1', source: 'Réplique De Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '5EB1', source: '複製サレタジャック' }),
       condition: (data) => !data.cloneLunge,
-      delaySeconds: (_data, matches) => matches.duration - 6,
+      delaySeconds: (_data, matches) => matches.castTime - 6,
```
quisquous, line 187:
```suggestion
      delaySeconds: (_data, matches) => parseFloat(matches.castTime) - 5,
```

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -206,7 +206,7 @@ export default {
       netRegexFr: NetRegexes.startsUsing({ id: '60C7', source: 'Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '60C7', source: 'ジャック' }),
       // Half a second longer cast time than the Lunge itself
-      delaySeconds: (_data, matches) => matches.duration - 6.5,
+      delaySeconds: (_data, matches) => matches.castTime - 6.5,
```
quisquous, line 209:
```suggestion
      delaySeconds: (_data, matches) => parseFloat(matches.castTime - 5),
```

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
```diff
@@ -226,7 +226,7 @@ export default {
       netRegexDe: NetRegexes.startsUsing({ id: '60C8', source: 'Herzbube' }),
       netRegexFr: NetRegexes.startsUsing({ id: '60C8', source: 'Jack' }),
       netRegexJa: NetRegexes.startsUsing({ id: '60C8', source: 'ジャック' }),
-      delaySeconds: (_data, matches) => matches.duration - 6.5,
+      delaySeconds: (_data, matches) => matches.castTime - 6.5,
```
quisquous, line 229:
```suggestion
      delaySeconds: (_data, matches) => parseFloat(matches.castTime) - 5,
```

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.js
valarnin: This wasn't meant to be a logic change, just a bug fix?

quisquous: Ok, that's fair, then leave the 6, but I think this should have a `parseFloat`, imo.  Even if JavaScript allows "numbery string - number = number", I think it's a bit weird.

valarnin:
There are some other places where triggers are using `matches.duration` without passing them through `parseFloat`, but I'm guessing those can be cleaned up as those triggers are converted to typescript.

image

Issue #3039, opened 2021-06-08
lint: Add rule to disallow non-relative imports


valarnin:
Continuing the discussion from https://github.com/quisquous/cactbot/pull/3034#discussion_r647023921

CC @MaikoTan @Trim21 @quisquous since you all commented on the original thread.

I played around with the `no-restricted-imports` rule for a while tonight and wasn't able to get it to allow `./*` or `../*` paths. There doesn't seem to be another rule that would work here, either. All other related rules are to specifically disallow relative imports.

That seems to lead to the following two questions:

1. Is enforcing only relative imports the correct action, given that all rules seem to go towards the opposite?
2. Should we roll our own eslint rule to handle this if so?

Trim21:
this maybe out of topic, “types/trigger” looks like a package that should be looked into `node_modules`, why it's working?

quisquous:
I would also be ok with absolute imports everywhere instead of relative.  I think consistency is most important to me.

We're also at the point where https://eslint.org/docs/rules/sort-imports might be good too.

valarnin:
I didn't mean to imply "hey, we should do it this way", just to point out what I was able to see with 30 minutes or so of googling and poking at the issue. It's entirely possible that there's no rule to enforce it fully the other way either, or that stylistically it makes more sense to use only relative imports.

Ultimately either way is fine with me and I agree that it's better for it to be consistent regardless of which way the import rule goes. I'm perfectly fine writing the rule myself if it needs to be done.

MaikoTan:
Oops, sorry for that I suggested the wrong way.
I have searched for a while and found nothing about enforcing the relative imports.
(and I think what Trim said is right, given a path like `types/trigger` is not only like a package path but also possible cause conflict?)

panicstevenson:
This is because the [base URL](https://github.com/quisquous/cactbot/blob/main/tsconfig.base.json#L17) is set to `./`, allowing you to import relative to the base when using the non-relative import syntax. It looks to be [purely cosmetic](https://www.typescriptlang.org/tsconfig#baseUrl) if we wanted to disable it (and update affected import statements).

Trim21:
So we didn't need a lint rule just remove `baseUrl` from tsconfig looks enough

valarnin:
Closing this now that the PR is merged, I don't think we need any further discussion around this.

PR #3049, opened 2021-06-08
Remove baseUrl definition from tsconfig

valarnin:
In regards to #3039.

Probably someone should review to make sure this fully forces relative paths, and should test it in an overlay to make sure it still works as expected.

### DIFF
--- a/resources/player_override.ts
+++ b/resources/player_override.ts
@@ -1,6 +1,6 @@
-import { Party, PlayerChangedRet } from 'types/event';
-import { Lang } from 'resources/languages';
-import { Job } from 'types/job';
+import { Party, PlayerChangedRet } from '../types/event';
+import { Lang } from '../resources/languages';
+import { Job } from '../types/job';
 
 import { addOverlayListener } from './overlay_plugin_api';
 import Util from './util';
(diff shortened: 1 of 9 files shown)


Trim21:
I have tested that importing like `'resources/languages'; ` didn't compile anymore without this configuration.

Trim21:
Just find that `npm run build` is not called in pr ci...

quisquous:
> Just find that `npm run build` is not called in pr ci...

I think `npm run tsc-no-emit` is run as a part of linting.  That's not 100% the same as `npm run build`, but it would catch errors like this.  Is that what you're concerned about?

> Considering add a [vercel](https://vercel.com/) to build front-end so we don't need to build pr locally and manually?

Sorry, I don't know what that is.  Maybe worth opening an issue and we can chat about it there?

Personally, given that you have to adjust an overlay url (in ACT or browser), pulling PRs locally and running out of my already set up cactbot folder is the most straightforward workflow for me.  Do you have a different workflow?

Trim21:
> > Just find that `npm run build` is not called in pr ci...
> 
> I think `npm run tsc-no-emit` is run as a part of linting. That's not 100% the same as `npm run build`, but it would catch errors like this. Is that what you're concerned about?

 Sorry, didn't catch that. Look good.

quisquous:
This seems fine to me, and a brief smoke test didn't turn anything up.

PR #3056, opened 2021-06-09
util: Adjust typings in trigger.d.ts

valarnin:
While working on converting popup-text.js to typescript, I noticed a typing issue with trigger.d.ts.

Basically, `./ui/raidboss/data/00-misc/test.js` has the following trigger which doesn't use language keys. Either the `Output` type in trigger.d.ts needs adjusted, or this trigger needs adjusted. Either way, to avoid having to redefine types, the typing of OutputStrings was moved to its own type definition and exported.

I opted to allow plain strings in this PR because I don't know if this is the only trigger that's using plain strings here and I don't have a good way to search all the triggers to check.

```js
    {
      id: 'Test Response',
      netRegex: NetRegexes.echo({ line: 'cactbot test response.*?', capture: false }),
      netRegexDe: NetRegexes.echo({ line: 'cactbot test antwort.*?', capture: false }),
      response: (_data, _matches, output) => {
        // cactbot-builtin-response
          alarmOne: '1',
          alertTwo: '2',
          infoThree: '3',
          ttsFour: '4',
        };
        return {
          alarmText: output.alarmOne(),
          alertText: output.alertTwo(),
          infoText: output.infoThree(),
          tts: output.ttsFour(),
        };
      },
    },
```

### DIFF
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -31,7 +31,7 @@ export type OutputStrings = { [outputKey: string]: LocaleText };
 // TODO: is it awkward to have Outputs the static class and Output the unrelated type?
 // This type corresponds to TriggerOutputProxy.
 export type Output = {
-  responseOutputStrings: OutputStrings | string;
+  responseOutputStrings: OutputStrings;
 } & {
   [key: string]: (params?: { [param: string]: string | undefined }) => string;
 };
@@ -92,9 +92,7 @@ export type BaseTrigger<Data> = {
   infoText?: TriggerField<Data, TriggerOutput>;
   tts?: TriggerField<Data, TriggerOutput>;
   run?: TriggerField<Data, void>;
-  outputStrings?: {
-    [key: string]: LocaleText;
-  };
+  outputStrings?: OutputStrings;
 }
 
 export type NetRegexTrigger<Data> = BaseTrigger<Data> & {
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -26,12 +26,12 @@ export type LocaleText = LocaleObject<string>;
 
 export type ZoneId = number | null;
 
+export type OutputStrings = { [outputKey: string]: LocaleText };
+
 // TODO: is it awkward to have Outputs the static class and Output the unrelated type?
 // This type corresponds to TriggerOutputProxy.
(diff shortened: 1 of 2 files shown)


valarnin:
It might also make sense to change the typing for `BaseTrigger<Data>`'s `outputStrings` to use the new `OutputStrings` type as well?

types/trigger.d.ts
```diff
@@ -26,12 +26,12 @@ export type LocaleText = LocaleObject<string>;
 
 export type ZoneId = number | null;
 
+export type OutputStrings = { [outputKey: string]: LocaleText };
```
quisquous, line 29:
```suggestion
export type OutputStrings = { [outputKey: string]: LocaleText | string };
```

I think this is what you're describe that the trigger does.

types/trigger.d.ts
```diff
@@ -26,12 +26,12 @@ export type LocaleText = LocaleObject<string>;
...
+
 // TODO: is it awkward to have Outputs the static class and Output the unrelated type?
 // This type corresponds to TriggerOutputProxy.
 export type Output = {
-  responseOutputStrings: {
-    [outputKey: string]: LocaleText;
-  };
+  responseOutputStrings: OutputStrings | string;
```
quisquous, line 34:
```suggestion
  responseOutputStrings: OutputStrings;
```

If the above is true, then this doesn't need `| string`.

types/trigger.d.ts
valarnin: I wasn't sure if that was applicable to both cases (the `outputStrings` trigger property vs the `responseOutputStrings` in-trigger-response override).

Frankly, I think it might make more sense to enforce `LocaleText` on the value here, but as I said before I'm not sure if there's a good way to find other cases that aren't using the correct structure.

quisquous:
> It might also make sense to change the typing for `BaseTrigger<Data>`'s `outputStrings` to use the new `OutputStrings` type as well?

Yes, definitely!

valarnin:
I decided that I'd make the change to the more strict format instead. I did an exhaustive check for responseOutputStrings that don't follow the lang pattern and it was just the one test trigger, so I think this way is better.

PR #3059, opened 2021-06-10
Jobs: Flash proc box when expired and in combat

rcj:
In case a proc box expires (reaches a duration of 0) and we are still
in combat, continuously flash the box to draw attention to it.

This is intended for proc boxes like dots or self-bufs which should be
always active/running while in combat.

The feature can be turned on in the Jobs settings via the variable
FlashExpiredProcsInCombat.

Proc boxes that should flash should pass
  flashWhenExpired: true
to the addProcBox call that creates them.

### DIFF
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -1141,14 +1141,3 @@
 .blu-color-lucid {
   background-color: rgb(227, 148, 210);
 }
-
-.proc-box[incombat] .flash-when-expired.expired {
-  display: block;
-  animation: flash 1s infinite;
-}
-
-@keyframes flash {
-  50% {
-    opacity: 0%;
-  }
-}
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -1141,3 +1141,14 @@
 .blu-color-lucid {
   background-color: rgb(227, 148, 210);
 }
+
+.proc-box[incombat] .flash-when-expired.expired {
+  display: block;
+  animation: flash 1s infinite;
+}
+
+@keyframes flash {
+  50% {
+    opacity: 0%;
+  }
+}
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -1,7 +1,3 @@
-:root {
-  --proc-box-flash-repeat: infinite;
-}
-
 #jobs {
   position: relative;
 }
@@ -1147,9 +1143,9 @@
 }
 
 /* Proc box animation */
-.proc-box.in-combat .flash-when-expired.expired {
+.proc-box.in-combat .notify-when-expired.expired {
   display: block;
-  animation: var(--proc-box-flash-animation, proc-box-flash) var(--proc-box-flash-duration, 1s) var(--proc-box-flash-repeat);
+  animation: var(--proc-box-notfy-animation, proc-box-flash) var(--proc-box-notify-duration, 1s) var(--proc-box-notify-repeat, infinite);
 }
 
 @keyframes proc-box-flash {
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -1,3 +1,7 @@
+:root {
+  --proc-box-flash-repeat: infinite;
+}
+
 #jobs {
   position: relative;
 }
@@ -1140,4 +1144,16 @@
 
 .blu-color-lucid {
   background-color: rgb(227, 148, 210);
+}
+
+/* Proc box animation */
+.proc-box.in-combat .flash-when-expired.expired {
+  display: block;
+  animation: var(--proc-box-flash-animation, proc-box-flash) var(--proc-box-flash-duration, 1s) var(--proc-box-flash-repeat);
+}
+
+@keyframes proc-box-flash {
+  50% {
+    opacity: 0;
+  }
 }
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -125,26 +125,14 @@ class Bars {
     this.isPVPZone = false;
     this.crafting = false;
 
-    this.addProcFlashCSS();
+    this.updateProcBoxFlashRepeat();
   }
 
-  addProcFlashCSS() {
-    console.log(`Add css, repeat: ${this.options.FlashExpiredProcsInCombat}`);
+  updateProcBoxFlashRepeat() {
     if (this.options.FlashExpiredProcsInCombat >= 0) {
       const repeats = this.options.FlashExpiredProcsInCombat === 0 ? 'infinite' : this.options.FlashExpiredProcsInCombat;
-      const animationRule = `.proc-box.in-combat .flash-when-expired.expired {
-  display: block;
-  animation: flash 1s ${repeats};
-}
 
-@keyframes flash {
-  50% {
-    opacity: 0;
-  }
-}`;
-      const animationStyle = document.createElement('style');
-      animationStyle.innerHTML = animationRule;
-      document.head.appendChild(animationStyle);
+      document.documentElement.style.setProperty('--proc-box-flash-repeat', repeats);
     }
   }
 
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -125,14 +125,14 @@ class Bars {
     this.isPVPZone = false;
     this.crafting = false;
 
-    this.updateProcBoxFlashRepeat();
+    this.updateProcBoxNotifyRepeat();
   }
 
-  updateProcBoxFlashRepeat() {
-    if (this.options.FlashExpiredProcsInCombat >= 0) {
-      const repeats = this.options.FlashExpiredProcsInCombat === 0 ? 'infinite' : this.options.FlashExpiredProcsInCombat;
+  updateProcBoxNotifyRepeat() {
+    if (this.options.NotifyExpiredProcsInCombat >= 0) {
+      const repeats = this.options.NotifyExpiredProcsInCombat === 0 ? 'infinite' : this.options.NotifyExpiredProcsInCombat;
 
-      document.documentElement.style.setProperty('--proc-box-flash-repeat', repeats);
+      document.documentElement.style.setProperty('--proc-box-notify-repeat', repeats);
     }
   }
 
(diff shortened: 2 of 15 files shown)


ui/jobs/jobs.css
```diff
@@ -1141,3 +1141,14 @@
 .blu-color-lucid {
   background-color: rgb(227, 148, 210);
 }
+
+.proc-box[incombat] .flash-when-expired.expired {
+  display: block;
+  animation: flash 1s infinite;
```
MaikoTan, line 1147: I don't think `infinite` is a good idea here, even it is only limited in combat.
To think when we have a boss who got a long long cutscene, for example, e1n, an infinite loop flash might be painful.
I suggest that `10` times or `8` times here is ok.

ui/jobs/jobs_config.ts
```diff
@@ -210,5 +210,18 @@ UserConfig.registerOptions('jobs', {
...
     },
+    {
+      id: 'FlashExpiredProcsInCombat',
+      name: {
+        en: 'Flash procs boxes of inactive dots/etc. while in combat.',
+        de: 'Dot/etc. boxen blinken when im Kampf und dot ist nicht aktiv.',
+        fr: 'Flash procs boxes of inactive dots/etc. while in combat.',
+        ja: 'Flash procs boxes of inactive dots/etc. while in combat.',
+        cn: 'Flash procs boxes of inactive dots/etc. while in combat.',
+        ko: 'Flash procs boxes of inactive dots/etc. while in combat.',
```
MaikoTan, line 221: If you don't know what is it in various languages, I think you should only add those you know.
i.e. delete these lines.

ui/jobs/jobs.js
```diff
@@ -565,6 +569,18 @@ class Bars {
...
   }
 
+  _updateProcBoxFlashState() {
+    if (this.options.FlashExpiredProcsInCombat) {
+      const boxes = document.getElementsByClassName('proc-box');
+      for (const box of boxes) {
+        if (this.inCombat)
+          box.setAttribute('incombat', '');
+        else
+          box.removeAttribute('incombat');
```
MaikoTan, line 579: Maybe a CSS `class` is more suitable for this usage?

Echoring:
There's an issue that if your level is synced or some skills have not learnt, they will flash disappointingly.
I think we'd better have a map about level and skill to prevent this happen?

ui/jobs/jobs.css
rcj: I didn't found this to be a problem, however I see your point, what about make this configurable with options:
6/8/10/infinite?

ui/jobs/jobs.js
rcj: I don't mind either, will change

rcj:
> 
> 
> There's an issue that if your level is synced or some skills have not learnt, they will flash disappointingly.
> I think we'd better have a map about level and skill to prevent this happen?
> or we can let it flash only once actived?

Hm good point. I would prefer the map option as why showing it in the first place if it isn't available anyway?

ui/jobs/jobs.css
MaikoTan: what about a number textbox? and 0 means infinite.

rcj:
> 
> 
> There's an issue that if your level is synced or some skills have not learnt, they will flash disappointingly.
> I think we'd better have a map about level and skill to prevent this happen?
> Or we can let it flash only when these boxes have been actived once?

I changed it to only flash if the box was active before.

I didn't implement the level check as it isn't clear to me yet how this should behave.
Should a box that isn't available due to level restriction not be rendered at all or should it simply be inactive?
The former has the downside that each time you level sync the order/available boxes might be different, which could be rather annoying.
And implementing it just to prevent the initial flash seems needles complexity. In that case only flashing after it was used once seems the better alternative, as everything else would behave the same.

ui/jobs/jobs.js
```diff
@@ -124,6 +124,28 @@ class Bars {
...
     this.crafting = false;
+
+    this.addProcFlashCSS();
+  }
+
+  addProcFlashCSS() {
+    console.log(`Add css, repeat: ${this.options.FlashExpiredProcsInCombat}`);
+    if (this.options.FlashExpiredProcsInCombat >= 0) {
+      const repeats = this.options.FlashExpiredProcsInCombat === 0 ? 'infinite' : this.options.FlashExpiredProcsInCombat;
+      const animationRule = `.proc-box.in-combat .flash-when-expired.expired {
```
panicstevenson, line 135: By no means am I a CSS expert, so feel free to push back here, but should end users be able to change this animation?

Speaking a bit more generically, should this be "add animation for expired timers" and then by default we would use this animation, but users (and skins?) would be welcome to override the CSS file and make it grow/shrink in size (or whatever they want?).

Maybe similar to https://github.com/quisquous/cactbot/pull/3017 (which should probably also get a small update to be a little more verbose as to what the CSS variable is doing)?

I don't think CSS overriding and bring-your-own animations is anything we've really considered before that other patch, but maybe that's the more appropriate pattern here? Let me know your thoughts!

(Also, welcome! 👋)

ui/jobs/jobs.css
```diff
@@ -1,3 +1,7 @@
+:root {
+  --proc-box-flash-repeat: infinite;
```
rcj, line 2: I'm wondering if there is any benefit defining this here instead of just providing a default in case the variable isn't defined like done for the other parameters. As this isn't reused anywhere else, I don't really see any benefit. Any opinions?

ui/jobs/jobs.js
rcj: > By no means am I a CSS expert, so feel free to push back here, but should end users be able to change this animation?

Neither am I. I learned it some 23 years ago when it was new and haven't really used it much in the last 15 years and a lot has changed/evolved since then. So I'm happy about any tips/suggestions.
I don't imagine a lot of people will do this, but also see no real downside in making this possible and certainly nice for those few who might want to do so.

> Speaking a bit more generically, should this be "add animation for expired timers" and then by default we would use this animation, but users (and skins?) would be welcome to override the CSS file and make it grow/shrink in size (or whatever they want?).

What other timers are there that can expire where such a thing might be useful? Also, is there any way to measure if a specific feature is being used? I haven't found anything in that regard.

> Maybe similar to #3017 (which should probably also get a small update to be a little more verbose as to what the CSS variable is doing)?
> 
> I don't think CSS overriding and bring-your-own animations is anything we've really considered before that other patch, but maybe that's the more appropriate pattern here? Let me know your thoughts!

I think if this should be a feature that is actually used by more than a handful of people the configuration menu would have to provide some selectable options and a preview how it looks like (assuming that the majority of users isn't technically).

Anyway, I think this is probably a bigger change and would require some more discussion as to how/what/etc.

> (Also, welcome! 👋)
Thanks!

ui/jobs/jobs.css
panicstevenson: I don't have the ability to test this right now. Does this affect in any way the ability to override these values? If not, then I'm inclined to say that they can all use the default syntax like you have in other areas

ui/jobs/jobs.js
panicstevenson: > Neither am I. I learned it some 23 years ago when it was new and haven't really used it much in the last 15 years and a lot has changed/evolved since then. So I'm happy about any tips/suggestions.
I don't imagine a lot of people will do this, but also see no real downside in making this possible and certainly nice for those few who might want to do so.

I think what's here is good! Thanks.

> What other timers are there that can expire where such a thing might be useful?

I don't know if people would want to add this to other timers (or, at least, I don't think of anything that uses proc boxes and would want them to nag you), but I do think that there might be people that want to override what we're doing here. One example might be [ZeffUI](https://github.com/Zeffuro/ZeffUI).

> Also, is there any way to measure if a specific feature is being used? I haven't found anything in that regard.

Not currently; although this is something we're interested in. It might be hard to know the granularity of what features are being used, but it would be nice to know overall adoption that aren't measured in GitHub downloads, and from there maybe a breakdown of a few settings: modules in use, languages in use, etc.

> I think if this should be a feature that is actually used by more than a handful of people the configuration menu would have to provide some select-able options and a preview how it looks like (assuming that the majority of users isn't technically).
> 
> Anyway, I think this is probably a bigger change and would require some more discussion as to how/what/etc.

Definitely; I think cactbot personalization could use a bit of an overhaul overall, but I think that has historically not been a major concern. Hopefully we can get a few things in before Endwalker around this.

panicstevenson:
One last thing here, do you think "Flash" should be something else, like maybe "Alert" (potentially an overloaded term) in case people want to override it? I don't think I have strong feelings here, since the default is to flash and anyone overriding that (I presume) would know what the default was, just curious what your thoughts are.

ui/jobs/jobs.css
rcj: The way I read the documentation, no. Default is simply used if there is no definition for the variable.

rcj:
Yeah, now that it can be changed a more generic term like "notify" or "alert" is a better description.

ui/jobs/jobs.js
```diff
@@ -565,6 +578,18 @@ class Bars {
...
   }
 
+  _updateProcBoxNotifyState() {
+    if (this.options.NotifyExpiredProcsInCombat >= 0) {
+      const boxes = document.getElementsByClassName('proc-box');
+      for (const box of boxes) {
+        if (this.inCombat)
+          box.classList.add('in-combat');
+        else
+          box.classList.remove('in-combat');
```
quisquous, line 588: Thanks for this whole patch! This is a great idea, and I agree with the feedback to not flash things until they've been used once.  My only feedback is this one suggestion to also reset the notify state when you start combat again.  This just extends the "until they've been used once" to "until they've been used once in the current combat."

This would prevent a situation where a proc expires, then going out of combat, and then zoning into a lower level dungeon without ever changing jobs and having it flash at you the whole time.

What do you think?

```suggestion
        if (this.inCombat) {
          box.classList.add('in-combat');
          for (const child of box.children)
            child.classList.remove('expired');
        } else {
          box.classList.remove('in-combat');
        }
```

quisquous:
Thanks for this!! You are also the lucky 100th contributor to cactbot! 
image
:tada:

PR #3060, opened 2021-06-10
build: only generate TypeScript declaration in dev

MaikoTan:
Well in fact, I don't like this solution. But it is not very well to bundle `*.d.ts` and `*.map` in release and ship it to end users. So disable it, and enable in development.

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -5,8 +5,8 @@
   },
   "compilerOptions": {
     "checkJs": false,
-    "declaration": true,
-    "declarationMap": true,
+    "declaration": false,
+    "declarationMap": false,
     "sourceMap": true,
     "allowSyntheticDefaultImports": true,
     "outDir": "./dist",
--- a/webpack/webpack.config.cjs
+++ b/webpack/webpack.config.cjs
@@ -218,6 +218,9 @@ module.exports = function(env, argv) {
         {
           test: /\.ts$/,
           loader: 'ts-loader',
+          options: {
+            compilerOptions: dev ? { declaration: true, declarationMap: true } : {},
+          },
         },
         {
           test: /\.css$/,


Trim21:
disable `MiniCssExtractPlugin` and `minimize` in dev, too?

MaikoTan:
It is more likely to separate them such as `webpack.config.base.cjs`, `webpack.config.prod.cjs` and `webpack.config.dev.cjs`

panicstevenson:
I agree on both of the above comments; I think we should disable the minimizer in dev, and that should be a change to diverge the webpack configs.

Semi-related: Do we want to be publishing source maps in build or only dev?

Trim21:
I think source map should be generated in both production and development mode

MaikoTan:
Or add one more "release" mode without source map?

Trim21:
production(or releases) build is transpiled and minimized, didn't see a reason why we didn't publish source map with them...

MaikoTan:
They are just not used by end users, if you need these files, you probably would clone the whole project of cactbot, instrad of digging the source map or declaration files. So these files in release are just increasing the file size.

quisquous:
I think we should ship source maps if it's not a large burden.  I think it can help in situations where you have a savvy end user who debugs something and files a better bug report.  I could understand the argument to not ship the sourcemaps in the zip file (since we have a lot of issues with updating and lots of files?) but it seems like shipping them to gh-pages is free.

Trim21:
it's just 3MB to 6MB difference.

MaikoTan:
In fact, these are declaration files and [declarationMap](https://www.typescriptlang.org/tsconfig#declarationMap), which is only being used in development letting editor/IDE knows where is the TypeScript source code from.
The source map effected in Chrome DevTool is generated by `devtool` option in webpack, which is extremely big and not recommended to be shipped to end users.

quisquous:
I'm a little confused what the state of this PR is.  Are you planning to split the config into separate files as you mentioned [above](https://github.com/quisquous/cactbot/pull/3060#issuecomment-858700467), or is that a later PR?

> In fact, these are declaration files and [declarationMap](https://www.typescriptlang.org/tsconfig#declarationMap), which is only being used in development letting editor/IDE knows where is the TypeScript source code from.
> The source map effected in Chrome DevTool is generated by `devtool` option in webpack, which is extremely big and also not recommended to be shipped to end users.

OH.  <_<  Whoops, sorry.  In that case, not shipping declaration files for TypeScript sounds good.  I do think we should still ship the source map to gh-pages.

> If you want to include source maps in gh-pages, what about to tweak the `devtool` option in webpack to be `inline-source-map` or something?

Sure!

MaikoTan:
> I'm a little confused what the state of this PR is. Are you planning to split the config into separate files as you mentioned [above](https://github.com/quisquous/cactbot/pull/3060#issuecomment-858700467), or is that a later PR?

Yes, I am going to separate the files later, please wait for me a while.

PR #3061, opened 2021-06-10
various: Fixes in preparation for popup-text typescript conversion

valarnin:
A lot of miscellaneous adjustments I hit while converting popup-text to typescript.

These are fine to push to main before the actual popup-text conversion (which is also done), so I thought I should PR these separately so that the popup-text PR is smaller.

Tested these changes with raidemulator and in-game within overlayplugin overlay via raidboss. Didn't test jobs or the other overlays, since I'm not familiar with them, but they shouldn't have any problems?

### DIFF
--- a/resources/player_override.ts
+++ b/resources/player_override.ts
@@ -15,10 +15,10 @@ import Util from './util';
 // jobs remotely due to gauge data being local and many bits of information
 // loaded from memory.
 
-type PlayerChangedDetail = { detail: PlayerChangedRet };
+export type PlayerChangedDetail = { detail: PlayerChangedRet };
 type PlayerChangedFunc = (e: PlayerChangedDetail) => void;
 
-export const addPlayerChangedOverrideListener = (playerName: string,
+export const addPlayerChangedOverrideListener = (playerName: string | null,
     func: PlayerChangedFunc): void => {
   if (!func)
     return;
--- a/resources/player_override.ts
+++ b/resources/player_override.ts
@@ -18,6 +18,7 @@ import Util from './util';
 export type PlayerChangedDetail = { detail: PlayerChangedRet };
 type PlayerChangedFunc = (e: PlayerChangedDetail) => void;
 
+// @TODO: Swap the order of these arguments, make playerName optional instead
 export const addPlayerChangedOverrideListener = (playerName: string | null,
     func: PlayerChangedFunc): void => {
   if (!func)
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -1,5 +1,5 @@
 import { Lang, NonEnLang } from '../resources/languages';
-import { Replacement } from '../ui/raidboss/timeline';
+import { TimelineReplacement, TimelineStyle } from '../ui/raidboss/timeline';
 import { RaidbossData } from './data';
 
 export interface BaseRegExp<T> extends RegExp {
@@ -135,8 +135,8 @@ export type TriggerSet<Data> = {
   timeline?: TimelineFunc;
   triggers?: NetRegexTrigger<Data>[];
   timelineTriggers?: TimelineTrigger<Data>[];
-  timelineReplace?: Replacement[];
-  timelineStyles?: Style[];
+  timelineReplace?: TimelineReplacement[];
+  timelineStyles?: TimelineStyle[];
 }
 
 // Less strict type for user triggers + built-in triggers, including deprecated fields.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -1,4 +1,5 @@
 import { Lang, NonEnLang } from '../resources/languages';
+import { Replacement } from '../ui/raidboss/timeline';
 import { RaidbossData } from './data';
 
 export interface BaseRegExp<T> extends RegExp {
@@ -37,8 +38,9 @@ export type Output = {
 };
 
 // The output of any non-response raidboss trigger function.
-export type TriggerOutput =
-    undefined | null | LocaleText | string | number | boolean | (() => TriggerOutput);
+export type TriggerOutput<Data, Matches> =
+    undefined | null | LocaleText | string | number | boolean |
+    ((d: Data, m: Matches, o: Output) => TriggerOutput<Data, Matches>);
 
 // The type of a non-response trigger field.
 export type TriggerFunc<Data, Matches, Return> =
@@ -49,7 +51,7 @@ type ResponseFields = 'infoText' | 'alertText' | 'alarmText' | 'tts';
 
 // The output from a response function (different from other TriggerOutput functions).
 export type ResponseOutput<Data, Matches> = {
-  [text in ResponseFields]?: TriggerFunc<Data, Matches, TriggerOutput>;
+  [text in ResponseFields]?: TriggerFunc<Data, Matches, TriggerOutput<Data, Matches>>;
 };
 // The type of a response trigger field.
 export type ResponseFunc<Data, Matches> =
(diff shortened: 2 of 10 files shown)


ui/raidboss/raidboss_options.ts
```diff
@@ -1,26 +1,30 @@
...
   SpeechAlert: boolean;
   GroupSpeechAlert: boolean; // TODO: we should remove this
   SoundOverride: string;
   VolumeOverride: number;
-  Condition: unknown; // TODO: trigger condition function
-  InfoText: unknown; // TODO: trigger function
-  AlertText: unknown; // TODO: trigger function
-  AlarmText: unknown; // TODO: trigger function
-  TTSText: unknown; // TODO: trigger function
+  Condition?: TriggerField<RaidbossData, boolean>; // TODO: trigger condition function
```
panicstevenson, line 17: If this is already Partial does it also need this question mark?

resources/player_override.ts
```diff
@@ -15,10 +15,10 @@ import Util from './util';
 // jobs remotely due to gauge data being local and many bits of information
 // loaded from memory.
 
-type PlayerChangedDetail = { detail: PlayerChangedRet };
+export type PlayerChangedDetail = { detail: PlayerChangedRet };
 type PlayerChangedFunc = (e: PlayerChangedDetail) => void;
 
-export const addPlayerChangedOverrideListener = (playerName: string,
+export const addPlayerChangedOverrideListener = (playerName: string | null,
```
panicstevenson, line 21: This isn't an actionable thing, just a thought I've had a few times, but I wonder if all the places we're doing `| null` could use a re-examination.

webpack/webpack.config.cjs
```diff
@@ -286,6 +286,7 @@ module.exports = function(env, argv) {
     ],
     stats: {
       children: true,
+      errorDetails: true,
```
panicstevenson, line 289: Does this likewise need to be in the [dev-server configuration](https://webpack.js.org/configuration/dev-server/#devserverstats-)? Or is this intended to help in debugging the gh-pages build?

resources/player_override.ts
valarnin: I thought about adjusting this to be optional instead, but it can't be specified as `playerName?: string` because the second parameter is required still?

panicstevenson: Definitely not in this PR, but we could swap the order of the parameters? Also, this isn't targeted at this line of code, I'm just curious how much `null` should come up (outside of WebComponents, which is beyond our control). In my day to day work, I don't feel like I see `null` used often.

ui/raidboss/raidboss_options.ts
valarnin: It doesn't, apparently. I was having other issues with accessing this in popup-text while converting that adding the optional to the field resolved for some reason, but it seems fine removing it now on that branch, so I'll remove it with the next set of changes I guess?

valarnin: Also all these TODOs might be okay to remove now, unless they're referencing something else?

webpack/webpack.config.cjs
valarnin: This entire `stats` key would go in the dev server config, but that change isn't in main yet.

ui/raidboss/raidboss_options.ts
panicstevenson: There's also `// one giant TODO` below which is probably also okay to drop?

resources/player_override.ts
valarnin: I can add a TODO for that change and someone can follow up in a later PR, after the popup-text stuff is merged?

webpack/webpack.config.cjs
panicstevenson: I guess another way to ask this: Why is this being added? I'm definitely not opposed, just curious what it gets us/what problem you were solving

valarnin: When child components (in this case the worker for raidemulator) get built, the `children` flag allows general error information, and the `errorDetails` flag allows a full stack if there's a compile error when building.

In this case it helped me track down a specific issue with the manifest change that wasn't visible without the flag enabled.

valarnin: Sorry, I see now what you're asking. This is something else, not what you linked, I believe.

https://webpack.js.org/configuration/stats/

panicstevenson: That sounds reasonable. Thanks!

resources/player_override.ts
panicstevenson: I don't think it's necessary to add a TODO, more so I saw this and it popped back in my head that I'm curious is `| null` is an anti-pattern or if it's fine overall. Again, not really an actionable item for this PR, just thoughts.
@quisquous if they have thoughts.

valarnin: Unresolving this so that quis can weigh in and we can continue the discussion. I think that it makes more sense to make playerName an optional second parameter, myself. I just don't think it fits for this PR, since it would have to be updated downstream as well.

quisquous: Agreed with the general idea here.  I think my feelings on null and undefined are:
* `null | undefined` is a code smell, because you shouldn't need two "unset" values
* prefer `undefined` over `null` where possible, because it feels more "ergonomic" in a TypeScript sense (e.g. `?` for optionally undefined things, but no similar thing for `null`)
* if you have to use `null` (e.g. HTML stuff), then use `null` instead of `undefined`

I think a TODO is not useful in this case because there's a lot of places where we do this, and so this one TODO on this one spot feels a bit out of place.  If you'd like to make sure we don't lose this, you could file a "take a look at null vs undefined and clean things up" issue.

valarnin: I think the TODO is okay here because this is only used for popup-text?
I can just roll that fix into the popup-text conversion as well?

quisquous: Ah sure, sorry I misunderstood what you meant.  Sounds good.

types/trigger.d.ts
```diff
@@ -130,12 +135,8 @@ export type TriggerSet<Data> = {
...
   triggers?: NetRegexTrigger<Data>[];
   timelineTriggers?: TimelineTrigger<Data>[];
-  timelineReplace?: {
-    locale: Lang;
-    missingTranslations?: boolean;
-    replaceText?: { [regex: string]: string };
-    replaceSync?: { [regex: string]: string };
-  }[];
+  timelineReplace?: Replacement[];
+  timelineStyles?: Style[];
```
quisquous, line 139: I know it's a bit long, but could you call these `TimelineReplacement` and `TimelineStyle`? I think the bare words on their own are ambiguous and could collide with other things.

ui/raidboss/timeline.ts
```diff
@@ -70,10 +71,11 @@ const activeText = {
...
 };
 
-type Replacement = {
+export type Replacement = {
   locale: string;
-  replaceSync: { [key: string]: string };
-  replaceText: { [key: string]: string };
+  missingTranslations?: boolean;
+  replaceSync?: { [key: string]: string };
+  replaceText?: { [key: string]: string };
```
quisquous, line 78:
```suggestion
  replaceSync?: { [regexString: string]: string };
  replaceText?: { [timelineText: string]: string };
```

Just as slightly more informative names.

quisquous (commented):
Overall looks great, thanks!

types/trigger.d.ts
valarnin: These were pulled in from timeline.ts, they would need to be renamed there, so as long as it's fine to pull that change into this PR, sure.

quisquous: Sure!

PR #3062, opened 2021-06-10
eslint: ban `Record`

Trim21:
mentioned in https://github.com/quisquous/cactbot/pull/3014#discussion_r647110732

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -2,7 +2,7 @@ import { Party } from '../types/event';
 import { Job, Role } from '../types/job';
 import Util from './util';
 
-const emptyRoleToPartyNames = () => {
+const emptyRoleToPartyNames = (): { [key in Role]: string[] } => {
   return {
     tank: [],
     healer: [],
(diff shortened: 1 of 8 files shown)


resources/party.ts
```diff
@@ -21,7 +21,7 @@ export default class PartyTracker {
   allianceIds_: string[] = [];
   nameToRole_: { [name: string]: Role } = {};
   idToName_: { [id: string]: string } = {};
-  roleToPartyNames_: Record<Role, string[]> = emptyRoleToPartyNames();
```
quisquous, line 24: I think `Record` is fine for cases where the key value is a union type.  It's not that I want to remove all uses of `Record`.  It's more that I think `Record<string, etc>` and `Record<number, etc>` are clearer as index types?

Trim21:
I think there isn't such a rule...

quisquous:
Sorry.  I'm happy to revisit if other people feel differently about using `Record`.  Maybe everybody else disagrees with me thinking it's weird?

Trim21:
looks like this rule could support Record<string, etc

PR #3065, opened 2021-06-11
raidboss: Disable Animations for Skips

panicstevenson:


panicstevenson:
Update animations to not run when doing timeline syncs:

Old:

https://user-images.githubusercontent.com/8849608/121622809-75907680-ca23-11eb-9766-67e1bb61d689.mp4

New:

https://user-images.githubusercontent.com/8849608/121622800-6f9a9580-ca23-11eb-921c-036aaed28bb3.mp4

panicstevenson:
There's a bug in the latest iteration where there is no animation playing for the Heavenly Strike. I think given all the other animations, users would expect that the ability just casted would still have an animation. Unfortunately, there is a bigger issue here, which can be seen in the older version: when the jump happens it puts the currently expiring bar at the bottom (logical, it's the timerbar with the highest event time), but this is fairly jarring given that timer bars have never done this in the past, so we're going to sweep it under the rug for now.

I think this change is sufficient in that there's animations for anything else that doesn't fall into this category, and also mainly because I think a proper fix here would include a much larger change in timeline's overall behavior.

PR #3066, opened 2021-06-11
raidboss: Shiva Ex Timeline Trigger Suppressions

panicstevenson:


panicstevenson:
Occasionally in testing https://github.com/quisquous/cactbot/pull/3065, these would fire multiple times.

quisquous:
Seems reasonable.  I do wonder if we should just auto-suppress for (default sync window) seconds (or possibly more?)

panicstevenson:
> Seems reasonable. I do wonder if we should just auto-suppress for (default sync window) seconds (or possibly more?)

I think it would be reasonable to at the least do `trigger.suppressSeconds ?? 0.5` or similar just to alleviate any low-hanging noise.

PR #3076, opened 2021-06-13
French translation for Zadnor NMs

MKhayle:
tested and ©works on my machine

### DIFF
--- a/ui/eureka/eureka.js
+++ b/ui/eureka/eureka.js
@@ -2403,6 +2403,7 @@ const Options = {
           label: {
             en: 'Beasts',
             de: 'Bestienbändigerin',
+            fr: 'Un jour, je serai la meilleure dresseuse',
           },
           x: 24.1,
           y: 37.4,
@@ -2412,6 +2413,7 @@ const Options = {
           label: {
             en: 'Parcel',
             de: 'Oboro',
+            fr: 'Les astres du jour et de la nuit',
           },
           x: 22.8,
           y: 34.2,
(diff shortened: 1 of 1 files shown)


quisquous:
Hi, thanks for the PR!  This looks good, except I think some of these names are really long.  I took some time to try to shorten the English names so that the text didn't overlap on the screen.  Is it possible to do that in French?

For example, something like these?
* `Dabog, soldat augmenté` => `Dabog`
* `La revanche de Lyon` => `Lyon`
* `Un cavalier qui surgit hors de la nuit` => `Un Cavalier` (or just `Cavalier`?)
* `Clarricie, sans défense ou presqu` => `Clarricie`

(although forgive me if they seem weird in French, I was just suggesting them as ideas for you to consider)

MKhayle:
Those names are the exact names from Zadnor, same as what was already done for Bozja's Southern Front so I wanted to respect the way it was done in the other half of this content. 

Visually, that doesn't seem to be annoying even with FATEs close one to another ? 
image

I can consider modifying them if needed tho if you do think it would be better, but if I do, I'd like to also rework the BSF translations if you allow me to.

quisquous:
I think my map is definitely sized much smaller than yours and so things (especially in Zadnor?) sometimes will overlap confusingly.

I think having French translations at all (especially consistent with Bozja) seems like a really good improvement, so let's merge this., In the future, if you feel like shorter names would make the map clearer, feel free to send another PR to clean them both up.

PR #3080, opened 2021-06-13
raidboss: Allow partial TTS entries

### LINKED ISSUE #3063: Support partial tts entries in triggers
Mentioned in #3055.  It would be useful if tts could be specified only in particular languages, e.g. `tts: { ja: 'some text' }` and for other languages fall back to the default infoText/alertText/alarmText as tts.

valarnin:
Closes #3063 

Makes `trigger.tts` completely Partial, adjusts logic for tts to only set `triggerHelper.ttsText` if the value resolves.

If `triggerHelper.ttsText` is undefined, `_onTriggerInternalPlayAudio` will by default fall back to the audio alert.

There might need to be a new `Default alert output` option for `TTS w/ fallback to Sound`, since otherwise this state can't be encountered in normal operation anyways?

### DIFF
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -42,6 +42,11 @@ export type TriggerOutput<Data, Matches> =
     undefined | null | LocaleText | string | number | boolean |
     ((d: Data, m: Matches, o: Output) => TriggerOutput<Data, Matches>);
 
+// Used if the function doesn't need to return an en key
+export type PartialTriggerOutput<Data, Matches> =
+    undefined | null | Partial<LocaleText> | string | number | boolean |
+    ((d: Data, m: Matches, o: Output) => PartialTriggerOutput<Data, Matches>);
+
 // The type of a non-response trigger field.
 export type TriggerFunc<Data, Matches, Return> =
     (data: Data, matches: Matches, output: Output) => Return;
@@ -95,7 +100,7 @@ export type BaseTrigger<Data> = {
   alarmText?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
   alertText?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
   infoText?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
-  tts?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
+  tts?: TriggerField<Data, PartialTriggerOutput<Data, MatchesAny>>;
   run?: TriggerField<Data, void>;
   outputStrings?: OutputStrings;
 }
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -51,12 +51,12 @@ export type PartialTriggerOutput<Data, Matches> =
 export type TriggerFunc<Data, Matches, Return> =
     (data: Data, matches: Matches, output: Output) => Return;
 
-// Valid fields to return from a ResponseFunc.
-type ResponseFields = 'infoText' | 'alertText' | 'alarmText' | 'tts';
-
 // The output from a response function (different from other TriggerOutput functions).
 export type ResponseOutput<Data, Matches> = {
-  [text in ResponseFields]?: TriggerFunc<Data, Matches, TriggerOutput<Data, Matches>>;
+  infoText?: TriggerFunc<Data, Matches, TriggerOutput<Data, Matches>>;
+  alertText?: TriggerFunc<Data, Matches, TriggerOutput<Data, Matches>>;
+  alarmText?: TriggerFunc<Data, Matches, TriggerOutput<Data, Matches>>;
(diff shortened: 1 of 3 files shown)


valarnin:
... That commit was meant to be on a different branch. Let me see if I can fix this...

types/trigger.d.ts
```diff
@@ -95,7 +100,7 @@ export type BaseTrigger<Data> = {
   alarmText?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
   alertText?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
   infoText?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
-  tts?: TriggerField<Data, TriggerOutput<Data, MatchesAny>>;
+  tts?: TriggerField<Data, PartialTriggerOutput<Data, MatchesAny>>;
```
quisquous, line 103: Should this be the type in `ResponseOutput` above as well? It seems like it'd be surprising if somebody updates responses to have custom tts for some language for some response and it behaved differently.

ui/raidboss/popup-text.ts
```diff
@@ -1152,9 +1154,8 @@ export class PopupText {
           result = triggerHelper.valueOrFunction(resp.tts);
       }
 
-      result ??= triggerHelper.defaultTTSText;
-
-      triggerHelper.ttsText = result?.toString();
+      if (result)
+        triggerHelper.ttsText = result?.toString();
```
quisquous, line 1158: Hmm.  What about https://github.com/quisquous/cactbot/search?q=%22tts%3A+null%22?

Or, in other words, how should we say "please don't play tts for this trigger (it is too long / complicated / collides with other tts that is more important)"?

quisquous:
Could you also add some documentation to the RaidbossGuide about how tts behaves differently?

> If `triggerHelper.ttsText` is undefined, `_onTriggerInternalPlayAudio` will by default fall back to the audio alert.
>
> There might need to be a new `Default alert output` option for `TTS w/ fallback to Sound`, since otherwise this state can't be encountered in normal operation anyways?

I am not sure what you are getting at here.  I think the fallback here is that if you don't have a partial tts entry, it falls back to the tts from the alarm/alert/info text.  If you don't have that, then you don't have an alarm/alert/info sound.  The only other sound is if you explicitly specified a `sound:` in your trigger, which should still play in these cases.  (I didn't test this, but that's my memory of how all this works).

ui/raidboss/popup-text.ts
valarnin: I think those triggers are invalid according to the defined data model anyways? Maybe this should be ` | false` to disable tts entirely?

valarnin:
> I think the fallback here is that if you don't have a partial tts entry, it falls back to the tts from the alarm/alert/info text.

I misunderstood the original request, I thought the request was to fall back to the info/alert/alarm sound, not the tts. I'll adjust to this instead, but it'll be later tonight.

quisquous:
> > I think the fallback here is that if you don't have a partial tts entry, it falls back to the tts from the alarm/alert/info text.
> 
> I misunderstood the original request, I thought the request was to fall back to the info/alert/alarm sound, not the tts. I'll adjust to this instead, but it'll be later tonight.

Ah, no.  The thought is that if you don't specify explicit `tts:` then you get the "default tts" from info/alert/alarm.  If you specify `tts:` with a localetext but not an entry for the language, it should behave as if `tts:` was not specified at all.

valarnin:
> Ah, no. The thought is that if you don't specify explicit `tts:` then you get the "default tts" from info/alert/alarm. If you specify `tts:` with a localetext but not an entry for the language, it should behave as if `tts:` was not specified at all.

This adjustment has been made.

ui/raidboss/popup-text.ts
quisquous: I think the data model is still being refined, and even so I think they're valid because they're in `TriggerOutput`.

What about changing `if (result)` to `if (result && result !== undefined)`? That way `undefined` continues to behave as if `tts` wasn't specified at all and `null` and `false` will both prevent tts as they currently do?

valarnin: That would just default it back to the defaultTTSText. I think the block would need to be this instead:

```typescript
      if (result !== undefined) {
        if (result)
          triggerHelper.ttsText = result?.toString();
      } else {
        triggerHelper.ttsText = triggerHelper.defaultTTSText;
      }
```

quisquous: Oh right, thanks for the correction.  Do you think it's worth a "// allow null and false to disable tts" comment to make this more obvious?

valarnin: Adjusted the logic, I think the comment makes sense.

PR #3081, opened 2021-06-13
raidboss: Add option for trigger sets to initialize data object

### LINKED ISSUE #3071: Add a Way for TriggerSets to Initialize Data
## Summary

As mentioned at the end of [this comment](https://github.com/quisquous/cactbot/pull/2880#issuecomment-855293396), it would be helpful to have a way to let triggersets define what is guaranteed inside data on initialization. I don't think this is necessary before converting other triggers to TypeScript, but it might create more work than necessary if it's not solved.

The problem is that an individual triggerset's data looks like this:
```ts
  fieldOne?: { [key: string]: string };
}
```
where fieldOne isn't guaranteed to exist, leading various triggers to inspect if `fieldOne` is undefined before doing any sort of logic on it. TypeScript won't actually guard here; for instance:
```ts
  fieldOne: { [key: string]: string };
}
```
won't throw an error, but it would be more correct to add a way to initialize variables we want to be initialized, for example:
```ts
  init: ((data: RaidbossData) => {
    return {
      ...data,
      fieldOne: {},
    };
  }),
  . . .
}
  ```

### PR DESCRIPTION
**Pull request by valarnin**
Closes #3071 as long as this implementation is what you're looking for.

I can track down a trigger set that would benefit from this and convert as part of this PR as well, if that's something you'd want?

### DIFF
--- a/docs/RaidbossGuide.md
+++ b/docs/RaidbossGuide.md
@@ -46,6 +46,7 @@ A trigger set must have one of zoneId or zoneRegex to specify the zone
 
 **initData**
 A function that can be used to initialize the data this trigger set uses.
+It should return an object that sets values for any fields in `data` that need to be initialized.
 This function is called any time the fight is reset, mainly on zone change or wipe.
 See [t1.ts](../ui/raidboss/data/02-arr/raid/t1.ts) for an example implementation.
 
--- a/docs/RaidbossGuide.md
+++ b/docs/RaidbossGuide.md
@@ -44,6 +44,11 @@ Prefer using this over zoneRegex.
 A trigger set must have one of zoneId or zoneRegex to specify the zone
 (but not both).
 
+**initData**
+A function that can be used to initialize the data this trigger set uses.
+This function is called any time the fight is reset, mainly on zone change or wipe.
+See [t1.ts](../ui/raidboss/data/02-arr/raid/t1.ts) for an example implementation.
+
 **zoneRegex**
 A regular expression that matches against the zone name (coming from ACT).
 If the regular expression matches, then the triggers will apply to that zone.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -131,6 +131,8 @@ export type TimelineTrigger<Data> = BaseTrigger<Data> & {
 // Because timeline functions run during loading, they only support the base RaidbossData.
 export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
 
+export type DataInitializeFunc<Data extends RaidbossData> = (data: RaidbossData) => Data;
+
 export type TriggerSet<Data> = {
   // ZoneId.MatchAll (aka null) is not supported in array form.
   zoneId: ZoneId | number[];
@@ -142,6 +144,7 @@ export type TriggerSet<Data> = {
   timelineTriggers?: TimelineTrigger<Data>[];
   timelineReplace?: TimelineReplacement[];
   timelineStyles?: TimelineStyle[];
+  initData?: DataInitializeFunc;
 }
 
 // Less strict type for user triggers + built-in triggers, including deprecated fields.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -131,7 +131,7 @@ export type TimelineTrigger<Data> = BaseTrigger<Data> & {
 // Because timeline functions run during loading, they only support the base RaidbossData.
 export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
 
-export type DataInitializeFunc<Data extends RaidbossData> = (data: RaidbossData) => Data;
+export type DataInitializeFunc<Data extends RaidbossData> = () => Omit<Data, keyof RaidbossData>;
 
 export type TriggerSet<Data extends RaidbossData> = {
   // ZoneId.MatchAll (aka null) is not supported in array form.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -133,7 +133,7 @@ export type TimelineFunc = string | string[] | ((data: RaidbossData) => Timeline
 
 export type DataInitializeFunc<Data extends RaidbossData> = (data: RaidbossData) => Data;
 
-export type TriggerSet<Data> = {
+export type TriggerSet<Data extends RaidbossData> = {
   // ZoneId.MatchAll (aka null) is not supported in array form.
   zoneId: ZoneId | number[];
   resetWhenOutOfCombat?: boolean;
@@ -144,7 +144,7 @@ export type TriggerSet<Data> = {
   timelineTriggers?: TimelineTrigger<Data>[];
   timelineReplace?: TimelineReplacement[];
   timelineStyles?: TimelineStyle[];
-  initData?: DataInitializeFunc;
+  initData?: DataInitializeFunc<Data>;
 }
 
 // Less strict type for user triggers + built-in triggers, including deprecated fields.
--- a/ui/raidboss/data/02-arr/raid/t1.ts
+++ b/ui/raidboss/data/02-arr/raid/t1.ts
@@ -12,10 +12,9 @@ export interface Data extends RaidbossData {
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.TheBindingCoilOfBahamutTurn1,
-  initData: (data: RaidbossData): Data => {
+  initData: () => {
     return {
       started: false,
-      ...data,
     };
   },
   triggers: [
--- a/ui/raidboss/data/02-arr/raid/t1.ts
+++ b/ui/raidboss/data/02-arr/raid/t1.ts
@@ -7,7 +7,7 @@ import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 
 export interface Data extends RaidbossData {
-  started?: boolean;
+  started: boolean;
 }
 
 const triggerSet: TriggerSet<Data> = {
--- a/ui/raidboss/data/02-arr/raid/t1.ts
+++ b/ui/raidboss/data/02-arr/raid/t1.ts
@@ -12,6 +12,12 @@ export interface Data extends RaidbossData {
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.TheBindingCoilOfBahamutTurn1,
+  initData: (data: RaidbossData): Data => {
+    return {
+      started: false,
+      ...data,
+    };
+  },
   triggers: [
     {
       id: 'T1 High Voltage',
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -425,7 +425,10 @@ export class PopupText {
   protected triggerSets: ProcessedTriggerSet[] = [];
   protected zoneName = '';
   protected zoneId = -1;
-  protected dataInitializers: DataInitializeFunc<RaidbossData>[] = [];
+  protected dataInitializers: {
+    file: string;
+    func: DataInitializeFunc<RaidbossData>;
+  }[] = [];
 
   constructor(
       protected options: RaidbossOptions,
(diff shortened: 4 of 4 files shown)

### CODE REVIEW

File: types/trigger.d.ts [hunk 1]
```diff
@@ -137,6 +139,7 @@ export type TriggerSet<Data> = {
   timelineTriggers?: TimelineTrigger<Data>[];
   timelineReplace?: TimelineReplacement[];
   timelineStyles?: TimelineStyle[];
+  initData?: DataInitializeFunc;
```
quisquous, line 142:
```suggestion
  initData?: DataInitializeFunc<Data>;
```

Could you also convert `t1.ts` just to have a working example?

ui/raidboss/popup-text.ts
```diff
@@ -786,6 +791,9 @@ export class PopupText {
     this.data = this.getDataObject();
     this.StopTimers();
     this.triggerSuppress = {};
+
+    for (const init of this.dataInitializers)
+      this.data = init(this.data);
```
quisquous, line 796: I realize that TypeScript will prevent this for built-in triggers, but I think we also need to handle the case of "bad" user triggers clobbering data here.  We should definitely print errors (and reset data again?) if somebody forgets to return anything from `initData` at all.

(Unsure if we should also check if all the fields in what is returned from `this.getDataObject()` are also the same values in this final `this.data`? This only gets called infrequently on wipe, so it might be worth being loud here if somebody holds it wrong.)

quisquous (commented):
Could you also add some documentation to the RaidbossGuide for this new field?

ui/raidboss/popup-text.ts
valarnin: So, my thought here is that I'd rather store the return in a temp var and check that it has the `me` property or something, rather than clobbering, because that gets a bit odd with having to Object.apply or something?

valarnin: Updated to reflect this. Let me know if this still doesn't make sense.

quisquous: Not that I'm saying we have to clobber, but I don't think that requires `Object.apply`.  You could do something like:

```typescript
    this.data = {
      ...this.data,
      ...this.getDataObject(),
    };
```

valarnin: Alright, how's this latest version look? Sort of a both approach?

quisquous: Yeah, looks good.  Then one init function that forgets to return something doesn't break all the other init functions.

valarnin: Maybe this should log to console as an `else`?

quisquous: Sounds good!

quisquous:
Looks good to go other than adding a bit about this in the RaidbossGuide.

valarnin:
> Looks good to go other than adding a bit about this in the RaidbossGuide.

Documentation added. It might make sense to adjust the test trigger set to have this as well and point to that for an example implementation?

quisquous:
> > Looks good to go other than adding a bit about this in the RaidbossGuide.
> 
> Documentation added. It might make sense to adjust the test trigger set to have this as well and point to that for an example implementation?

Yeah, having this in the test triggerset would be good and linking to that instead of `t1.ts` would be good.  It could even set a variable used in a trigger or something so that we could verify initData was working manually.

valarnin:
I'd have to convert the test triggerset as part of this PR then, or else it wouldn't really make sense?

Trim21:
I'm a little confused, it looks like `DataInitializeFunc` didn't need to take a data argument, just need to return 
a object with extra properties.

quisquous:
> I'd have to convert the test triggerset as part of this PR then, or else it wouldn't really make sense?

Oh, right.  Maybe it'd be better to just use `t1.ts` for now then.

quisquous:
> I'm a little confused, it looks like `DataInitializeFunc` didn't need to take a data argument, just need to return
> a object with extra properties.

I think the idea was that it would force trigger sets to fully initialize their `Data` object by making them return it.


I don't know if this is what you are suggesting, but what if we changed the type to: `export type DataInitializeFunc<Data extends RaidbossData> = () => Omit<Data, keyof RaidbossData>;`  In other words, the init function returns an object that has all of the fields that `Data` has that `RaidbossData` doesn't, something like:

```typescript
  initData: () => {
    return {
      started: true,
    };
  },
```

I think this is cleaner and removes a lot of the awkwardness, especially in the trigger files.

Trim21:
I'm saying how I think the code is dealing with this new function... In fact I'm expecting something like this
```js
this.data = { ...this.data }; 
```

but your suggestions looks good.

valarnin:
> I don't know if this is what you are suggesting, but what if we changed the type to: `export type DataInitializeFunc<Data extends RaidbossData> = () => Omit<Data, keyof RaidbossData>;` In other words, the init function returns an object that has all of the fields that `Data` has that `RaidbossData` doesn't, something like:

I think I like this solution better than what's currently in place. I'll see if I can pull in the changes from main to fix the merge conflict, then make this adjustment.

valarnin:
Merge conflicts fixed and logic rewritten.

panicstevenson:
nit: t1 should not have started as optional within data

valarnin:
> 
> 
> nit: t1 should not have started as optional within data

Fixed, thanks.

ui/raidboss/popup-text.ts
```diff
@@ -788,6 +793,18 @@ export class PopupText {
     this.data = this.getDataObject();
     this.StopTimers();
     this.triggerSuppress = {};
+
+    for (const init of this.dataInitializers) {
+      const data = init();
+      if (data) {
```
panicstevenson, line 799: I either don't understand what this is guarding or why we're guarding it. I understand this is for a scenario where we somehow have a malformed function, but it feels like this shouldn't be possible aside from a user file override, in which case this should already bubble up an issue? Maybe there's something else here that I'm missing?

valarnin: This is specifically for user file overrides. Because those are javascript-only, we can't rely on the typing at all, so if they return undefined or something we don't want to cause that to error out. Maybe it should be checking `if(isObject(data))` instead, but I think this check is required regardless.

panicstevenson: So, I guess my mental model is that a user override wouldn't really need/want this? Or care about it? And, we should get a bigger error (in general) that something bad happened? In either sense, I guess I'm proposing dropping this or making it very explicit that something failed within `initData`.

Is there a use-case where a user might want an extremely complex initData function that actually has data initialized? For most cases, I was just assuming this was a "clear having to check undefined 100 times in a TypeScript file" which doesn't apply for JavaScript user files

valarnin: My thought process here is as follows:

1. There's no good way to distinguish between a user trigger set and a built-in trigger set.
2. `initData` is available on both
3. Therefore, a user may use `initData` in a custom trigger set
4. If a user uses `initData` and doesn't return a proper object, we don't want to completely break raidboss for them

panicstevenson: Right, but then the user isn't getting what they expect, and there's a log about data initializers which might not explicitly tell them that the thing they did doesn't work? I'm thinking in the sense of a user gets another user's user files and has no idea what they're doing, then they drop those user files into their user directory and it "doesn't work" it's a lot greyer of an area to support than an explicit log that says, "ERROR IN USER FILE OVERRIDE INIT DATA", and doesn't have nearly the same impact as crashing entirely from "I need to back these out immediately" or "I added a file and now cactbot doesn't work. HELP"

valarnin: I guess I don't understand the question/concern here. If the `if(data)` check isn't present, then a bad user-supplied initData function will clobber the raidboss data object and cause every other trigger to break.

So either we:

- Don't allow user-supplied initData functions at all, which would require some sort of additional checking to not push them to the array, and goes against the data model
- Allow user-supplied initData functions, but keep users from breaking their raidboss entirely, and instead warn them and ignore the bad initData function
- Allow user-supplied initData functions, and let users completely break their raidboss

Right now, the second above is what is in this PR.

docs/RaidbossGuide.md
```diff
@@ -44,6 +44,11 @@ Prefer using this over zoneRegex.
 A trigger set must have one of zoneId or zoneRegex to specify the zone
 (but not both).
 
+**initData**
+A function that can be used to initialize the data this trigger set uses.
+This function is called any time the fight is reset, mainly on zone change or wipe.
```
quisquous, line 49:
```suggestion
It should return an object that sets values for any fields in `data` that need to be initialized.
This function is called any time the fight is reset, mainly on zone change or wipe.
```

ui/raidboss/popup-text.ts
quisquous: I do think checking `if (typeof data === 'object')` instead of `if (data)` would be both clearer to the reader here and would catch more errors.

docs/RaidbossGuide.md
MaikoTan: I think to add a type hint makes sense?
i.e. something like:

> **initData**
> `(data: RaidbossData) => Data`
> blahblah

quisquous: In general I think I would agree with you, but in this specific case I worry that `() => Omit<Data, keyof RaidbossData>` might be more confusing than helpful.

ui/raidboss/popup-text.ts
quisquous: After some more discussion/thought, I think one missing piece from the error handling here is the filename (which I guess would need to be tracked per init function, sorry).  Having something closer to ``console.log(`Error in file: ${filename}: these triggers may not work; initData function returned invalid object: ${init.toString()}`);``

But, I think it would be helpful to less technical users who aren't trigger authors (possibly copying and pasting triggers from friends) to know which file is busted and what the result of that might be.

panicstevenson: Sorry, to follow up on this (work got busy): I am not suggesting the first point at all; I don't think that's viable. I think latter two points are both okay, and both have their benefits. I think that if you're going to go with the second approach (as is currently in the PR), you need to be very explicit with the fact that there is an error. I would be surprised if more than 1% of users are reading the cactbot logs, so you need an approach that makes it either fairly obvious to the user that they did something wrong (the third point, hopefully), or something fairly obvious to whatever technical support (Discord, GitHub, etc.) that there were issues. I think what @quisquous has above works well; I think as long as the words `Error` and `initData` come across (filename super helpful too!) then it shouldn't be impossible to navigate from there.

ui/raidboss/data/02-arr/raid/t1.ts
```diff
@@ -7,11 +7,16 @@ import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 
 export interface Data extends RaidbossData {
-  started?: boolean;
+  started: boolean;
 }
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.TheBindingCoilOfBahamutTurn1,
+  initData: () => {
```
panicstevenson, line 15: Hardly a nit, since I like the current pattern, but before this rolls out to everywhere does anyone prefer:
```js
initData: () => ({
  started: false,
}),
```

panicstevenson:
Used this in 3099 and it felt great! Thanks for this!

ui/raidboss/data/02-arr/raid/t1.ts
valarnin: Both work, so it's more a matter of style preference for the reference trigger?

ui/raidboss/popup-text.ts
valarnin: See latest commits, please. Sorry for the rebase/force-push, I wanted to pick up the webpack config changes so I could actually run this in dev server and test properly.

Specifically this commit should address the issues?

https://github.com/quisquous/cactbot/pull/3081/commits/42ae20988ad0450d0c4e56faf369406d5f77bd1e

PR #3083, opened 2021-06-14
cache webpack build

Trim21:


webpack/webpack.config.cjs
```diff
@@ -175,6 +175,10 @@ module.exports = function(env, argv) {
         },
       },
     },
+    cache: {
+      type: 'filesystem',
+      allowCollectingMemory: true,
```
quisquous, line 180:
```suggestion
```

From https://webpack.js.org/configuration/other-options/#cacheallowcollectingmemory it looks like this already defaults to true in development mode.  (Unless I'm misunderstanding?)

PR #3086, opened 2021-06-14
refactor: import image instead of path

MaikoTan:


ui/jobs/buff_tracker.ts
```diff
@@ -5,6 +5,33 @@ import EffectId from '../../resources/effect_id';
 import { MatchesAbility, MatchesGainsEffect, MatchesLosesEffect } from '../../resources/matches';
 import PartyTracker from '../../resources/party';
 
+import potionImage from '../../resources/ffxiv/status/potion.png';
```
panicstevenson, line 8: I don't have thoughts here, but this might be for @quisquous' benefit if they care; I think we can dynamically import the images using [require.context()](https://webpack.js.org/guides/dependency-management/#requirecontext) and you could specify them individually. It's not the best since we'll import everything in this folder, but maybe it helps think about asset folder structure. Actually, it might [not be possible](https://stackoverflow.com/questions/65655682/require-context-alternative-when-using-es-modules) with ES modules? Curious if you found anything in trying to clean up these import statements?

webpack/webpack.config.cjs
```diff
@@ -241,6 +242,10 @@ module.exports = function(env, argv) {
             },
           ],
         },
+        {
+          test: /\.(png|jpe?g)/,
+          type: 'asset/resource',
```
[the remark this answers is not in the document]
MaikoTan: Yup, basically I followed the guide from webpack official document, and didn't tried other options.
I'll take a look later.

webpack/webpack.config.cjs
```diff
@@ -241,6 +242,10 @@ module.exports = function(env, argv) {
             },
           ],
         },
+        {
+          test: /\.(png|jpe?g)/,
```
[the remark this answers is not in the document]
MaikoTan: Yeeees, I am just thinking to add more image fotmat that usually being used.

webpack/webpack.config.cjs
```diff
@@ -269,19 +274,14 @@ module.exports = function(env, argv) {
       ...htmlPluginRules,
       new CopyPlugin({
         patterns: [
-          {
-            // copy sounds and images
-            from: 'resources/@(ffxiv|sounds)/**/*',
-          },
+          // {
+          //   // copy sounds and images
+          //   from: 'resources/@(ffxiv|sounds)/**/*',
```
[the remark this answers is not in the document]
MaikoTan: Oops! This is what I used for test if webpack copied files as intended. I just forgot to uncomment this before push haha. ^_^

ui/jobs/buff_tracker.ts
MaikoTan: Directly `icon: await import('/path/to/image'),` is also an approach I think?

panicstevenson: Does that let you dynamically import? I guess I'm somewhat of the opinion that it'd be nice to be like `import statusImages from '../../resources/ffxiv/status/*.png'` and then you can call `statusImages['battle-litany'];` I don't know if that's an option though. Also, it doesn't guarantee a lot of safety, since nothing stops you from writing `statusImages['battle-littany'];` and now jobs is broken

MaikoTan: Since the `asset` configuration landed, whatever the import method is, these images would be translated as `data-url`. So I think performance/memory is not the issue?

panicstevenson: Yea, the concern here is verbosity; "concern" to be used lightly. It was more a question of should we/could we. I can try to preface some of these PR thoughts as musings rather than actionable, addressable concerns so hopefully people don't feel tied to doing things I'm thinking of sparingly.

> Since the asset configuration landed, these imported images would be translated as `data-url`. So I think performance/memory is not the issue?

Bundle-size is the question here, right? I think most of our images have a one-to-one mapping with modules, so I don't think there's any duplication concerns (which, I _think_ is what people generally key off of to know whether to inline assets); however, since everyone will get these files regardless (via download) I think the demographic to determine if data inlining is helpful would be to look at players using the GitHub pages URL and think about what images are being used and how often. I think here, since a given player will probably never use all of the images available in one session (log in, play BLM in raid with `jobs` module, log out) I wonder if we should re-evaluate? (I know that it was my suggestion, sorry!)

In other words, is there a benefit for a BLM to load all potential images from `jobs`? I think they should only be concerned with what they're concerned about, pulled down as needed.

I'm also not a webpack/front-end person at all, so if someone has guidance here, I'd take that over anything I posted above 😛

Trim21: @panicstevenson I think webpack would handle this.

import same image in different js files won't make it bundled twice

panicstevenson: Maybe that's fair, since the images aren't being inlined into the CSS, so nothing should be blocking from a CSS call perspective (with the exception of raidemulator). The issue then is that data urls have additional overhead, but I doubt we care since we're talking about ~1 kb additional per image? I think the argument then would come back to "for the cactbot updater, is it better to have a bunch of small files or a handful of big files," since sometimes we do see people's cactbot plugin cache get corrupted due to failed updates. However, we don't have any data for that, so it's hard to make a judgment call.

Trim21: I think the benefit here is more about managing assets with webpack.

ui/jobs/buff_tracker.ts
```diff
@@ -5,6 +5,33 @@ import EffectId from '../../resources/effect_id';
...
+import boleImage from '../../resources/ffxiv/status/bole.png';
+import ewerImage from '../../resources/ffxiv/status/ewer.png';
+import spearImage from '../../resources/ffxiv/status/spear.png';
+import spireImage from '../../resources/ffxiv/status/spire.png';
+import ladyOfCrownsImage from '../../resources/ffxiv/status/lady-of-crowns.png';
+import lordOfCrownsImage from '../../resources/ffxiv/status/lord-of-crowns.png';
+import devilmentImage from '../../resources/ffxiv/status/devilment.png';
+import standardFinishImage from '../../resources/ffxiv/status/standard-finish.png';
+import technicalFinishImage from '../../resources/ffxiv/status/technical-finish.png';
+import battlevoiceImage from '../../resources/ffxiv/status/battlevoice.png';
```
quisquous, line 28: Sorry, this is the tiniest nit, but can this be `battleVoiceImage`? Battle Voice is two words.  <_<

ui/jobs/buff_tracker.ts
quisquous: To reiterate my feelings here, I think:
* I think this is a good improvement as-is
* probably in another PR we should move all of these images to another file just for convenience?
* if we want this to be dynamic, I think that can be a good follow-up PR but I think I'd want to see some performance problems before making things more complicated

quisquous (commented):
Is this still a Draft? It seems ready to land.

MaikoTan:
> Is this still a Draft? It seems ready to land.

Should the "dynamic import" part that panic mentioned be landed or not, which I have still no perfect solution to deal with though.

PR #3089, opened 2021-06-14
refactor(util): add util/csv_util.py TypeScript version

MaikoTan:


Trim21:
adding some tests？

MaikoTan:
@Trim21 Nay, help me, why is it not expected?

quisquous:
I haven't forgotten about this.  I'll try to take a look tomorrow.  Thanks for looking at utils!

util/csv_util.ts
```diff
@@ -0,0 +1,159 @@
...
+export const getCnTable: GetTableFunc = (
+    table: string, inputs: (string | number)[], outputs?: string[],
+) => {
+  const url = `${_BASE_GITHUB}${_CN_GITHUB}${table}.csv`;
+  return __getRemoteTable(url, inputs, outputs);
+};
+
+
+export const getLocaleTable: {
+  <T extends (string | number), K extends string>(
```
quisquous, line 130: I'm not sure why I understand why this has `extends` in it anywhere.  Are you planning for these tables to have different input and output types?

I'm also not sure why `getLocaleTable` has a different return type than `makeMap` since all implementations of `getLocaleTable` end up returning a value from `makeMap`.  I think `makeMap`'s definition is a little bit more understandable to me, for what it's worth.

MaikoTan: I am trying to make the results from these functions inferable and also type-checkable.
image

util/csv_util.ts
```diff
@@ -0,0 +1,159 @@
+#!/usr/bin/env ts-node-script
+
+import { parseString } from '@fast-csv/parse';
+import fetch from 'node-fetch';
+import { Lang } from '../resources/languages';
+
+const _BASE_GITHUB = 'https://raw.githubusercontent.com/';
```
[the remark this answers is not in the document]
MaikoTan: Okay.

util/csv_util.ts
quisquous: !!!!!!!!!!!!

Should `makeMap` also have this call signature so that it can be typed as well? I think there's at least one call outside of this file.

util/csv_util.ts
```diff
@@ -0,0 +1,159 @@
...
+  <T extends (string | number), K extends string>(
+    table: string, inputs: [key: K, ...indices: T[]],
+  ): Promise<Table<K, T>>;
+  <T extends (string | number), K extends string>(
+    table: string, inputs: (string | number)[], outputs?: [key: K, ...indices: T[]],
+  ): Promise<Table<K, T>>;
+};
+
+// Turn names from tables into JavaScript-safe ascii string keys.
+export const cleanName = (str?: string | null): string | undefined | null => {
```
quisquous, line 26: What calls this with null?

MaikoTan:
I have fixed a bunch of error but there are still issues.
I test this with `gen_effect_id` and I found that [this status](https://xivapi.com/Status/1379) still got wrong capitalization with current regex, the original `capitalize` function of python version results `_Rsv_1379_1_1_C0_0Status`, but the regex method of my TypeScript version generate `_rsv_1379_1_1_c0_0status`.

quisquous (commented):
> I have fixed a bunch of error but there are still issues.
> I test this with `gen_effect_id` and I found that [this status](https://xivapi.com/Status/1379) still got wrong capitalization with current regex, the original `capitalize` function of python version results `_Rsv_1379_1_1_C0_0Status`, but the regex method of my TypeScript version generate `_rsv_1379_1_1_c0_0status`.

If you wanted, we could also remove leading underscores too and then that would "solve" the problem.  I don't think anybody is going to be using this particular effect by name (and we could consider replacing it with the real name).

util/csv_util.ts
MaikoTan: There might be some null value in CSVs I think?

MaikoTan:
> > I have fixed a bunch of error but there are still issues.
> > I test this with `gen_effect_id` and I found that [this status](https://xivapi.com/Status/1379) still got wrong capitalization with current regex, the original `capitalize` function of python version results `_Rsv_1379_1_1_C0_0Status`, but the regex method of my TypeScript version generate `_rsv_1379_1_1_c0_0status`.
> 
> If you wanted, we could also remove leading underscores too and then that would "solve" the problem. I don't think anybody is going to be using this particular effect by name (and we could consider replacing it with the real name).

I fixed it, I realized that is because the regex `[A-z]` contains special characters like `[\]_` so I convert it to lowercase and then just do `[a-z]` would fix it.

util/csv_util.ts
quisquous: I think everything is a string (possibly the empty string, but still a string).

MaikoTan: Sure, but then you should always make sure the parameter is not null before passed into the function.

PR #3091, opened 2021-06-14
fix webpack cache

Trim21:


webpack/webpack.config.cjs
```diff
@@ -175,8 +175,8 @@ module.exports = function(env, argv) {
         },
       },
     },
-    cache: {
-      type: dev ? 'memory' : 'filesystem',
+    cache: dev ? true : {
```
panicstevenson, line 178: What is happening before / after with this change? We want type 'filesystem' it seems like for production, but do we only need to supply `true` to cache for dev?

MaikoTan: `true` is equal to `cache: { type: 'memory' }`

The main problem on this pr is that when `cache` receive an object and type is set to `memory`, the `name` property should not be set, otherwise webpack would start to complain.

valarnin: I can confirm that without this change, `npm run start` won't run and complains about this.

quisquous: Thanks for the explanation.  It's hard to understand what "fix webpack cache" means with no text and no linked bug.

Trim21: sorry for that!

PR #3093, opened 2021-06-14
build: add babel/preset-env

MaikoTan:
to ensure js code would be transpiled,
and recognizable by current version of CEF in OverlayPlugin.
Also allow us to use new ES syntax to simplify code

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -33,49 +33,76 @@
         "@babel/highlight": "^7.0.0"
       }
     },
+    "@babel/compat-data": {
+      "version": "7.14.5",
+      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.14.5.tgz",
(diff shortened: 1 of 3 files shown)


quisquous:
This is a great idea, thanks!

Trim21:
TypeScript files too?

MaikoTan:
> TypeScript files too?

tsc would handle them

unfortunately babel knows typescript but don't do the type-checking.

panicstevenson:
Should we run this also for cjs files?

Trim21:
> > TypeScript files too?
> 
> tsc would handle them
> 
> unfortunately babel knows typescript but don't do the type-checking.

can we pipeline `babel-loader` after `ts-loader`?

MaikoTan:
> Should we run this also for cjs files?

It is possible but currently cjs files are used out of webpack....

MaikoTan:
> can we pipe TS files to `babel-loader` after `ts-loader`?

Sounds good.

PR #3097, opened 2021-06-14
Jobs: Add notification sound to proc boxes

rcj:
In case a visual animation when a proc box expired isn't enough or
desired, playing a short alert sound can be helpful.

By default sounds are disabled.

If sounds are desired, they can be configured to be played either when
the threshold of an ability is reached or if the ability expired.


resources/timerbox.ts
```diff
@@ -416,6 +431,12 @@ export default class TimerBox extends HTMLElement {
   advance(): void {
     const elapsedSec = (new Date().getTime() - this._start) / 1000;
     if (elapsedSec >= this._duration) {
+      // We need to check for this._duration > 0 here, as for undocumented reason the
+      // duration of a timerbox is always set to zero before it is set to the
+      // actual duration. As a result this would otherwise trigger a sound each time
+      // the ability is activated.
```
rcj, line 437: I couldn't find any documentation or reason in a commit message why the duration is always set to zero before it is immediately after set to the correct value. Removing the assignment to zero didn't show any issues in my short tests. Does anyone remember what issues this solves?

resources/timerbox.ts
```diff
@@ -416,6 +431,12 @@ export default class TimerBox extends HTMLElement {
...
+      if (this._duration > 0 && this._playSoundNotification && this._soundNotification === 'expired')
+        this.playNotification();
```
rcj, line 439: I don't really like adding this to the timerbox directly and was wondering if adding a callback/event mechanism that fires when the threshold is reached or the timer expired would be a better alternative. This would allow attaching different actions to different timerboxes and would also remove the requirement to pass all kinds of configuration options into the timerbox.

resources/timerbox.ts
```diff
@@ -431,9 +452,20 @@ export default class TimerBox extends HTMLElement {
...
+    if (remainingTime <= this._threshold && this._soundNotification === 'threshold' && this._playSoundNotification && this._duration > 0)
+      this.playNotification();
+
     this.draw();
   }
 
+  playNotification(): void {
+    this._playSoundNotification = false;
+    const audio = new Audio('../../resources/sounds/BigWigs/Alert.ogg');
+    audio.volume = 0.3;
```
rcj, line 465: For now just hard-coded, should probably be configurable similar to the eureka overlay.

ui/jobs/jobs_config.ts
```diff
@@ -219,5 +219,26 @@ UserConfig.registerOptions('jobs', {
...
+      name: {
+        en: 'Play a sound notification if a proc box for dots/etc. expires while in combat.',
+        de: 'Spiele einen Alarm Sound wenn eine dot/etc. box im Kampf inaktiv wird.',
+      },
+      type: 'select',
+      options: {
+        en: {
+          'Disabled': 'disabled',
+          'When counter reaches 0': 'expired',
+          'When counter reaches threshold': 'threshold',
```
rcj, line 233: I don't like this description as I'm unsure how many users actually know what is meant by threshold. Would appreciate any suggestions.

resources/timerbox.ts
quisquous: I agree.  If possible, I think a callback would be nicer here rather than baking sounds inside of the web component.  Is it possible to do that in this PR?

resources/timerbox.ts
quisquous: I think it would be nice to have some new sound for this, to avoid making it sound like mechanics are happening.  Do you want to find an acceptably licensed sound off of freesound.org or something?)

ui/jobs/jobs.js
```diff
@@ -453,8 +453,10 @@ class Bars {
       timerBox.id = id;
       timerBox.classList.add('timer-box');
     }
-    if (notifyWhenExpired)
+    if (notifyWhenExpired) {
       timerBox.classList.add('notify-when-expired');
+      timerBox.soundnotification = this.options.NotifyExpiredProcsInCombatSound;
```
quisquous, line 458: I think these options could be separate.  I think `notifyWhenExpired` could be just the animation and then `this.options.NotifyExpiredProcsInCombatSound;` could be just the sound, and then somebody could turn one on but not the other.

ui/jobs/jobs_config.ts
quisquous: "When counter gets close to zero" or "When the proc box gets big when nearly expired" or something along those lines?

resources/timerbox.ts
quisquous: Ah sorry, these web components have existed foreverrrrrrr and are a bit dusty.

I think this would be worth tracking down and fixing separately.  My memory here is that some(?) of the web components in `resources/` needed to have some(?) attributes set to zero to properly reset them, but I don't remember why anymore.  It would be nice to not have to do this in general.

resources/timerbox.ts
rcj: Yes, I'll change it to support callbacks for the different relevant states.

resources/timerbox.ts
rcj: I'll look into it.

ui/jobs/jobs.js
quisquous: Oh right, sorry.  `notifyWhenExpired` is whether the proc box itself is on an proc where this should happen at all.  Got it, sorry for the confusion!

resources/timerbox.ts
```diff
@@ -408,14 +423,27 @@ export default class TimerBox extends HTMLElement {
     clearTimeout(this._timer ?? 0);
     this._timer = null;
     this.classList.remove('expired');
+    this._notifyThresholdCallbacks = true;
 
     this._start = new Date().getTime();
+
+    for (const f of this._onResetCallbacks)
+      setTimeout(f, 0);
```
quisquous, line 431: Can I ask what `setTimeout` is doing here vs just calling these callbacks directly? Is there some ordering constraint here or are you just making sure they run outside of the advance/reset callstacks?

ui/jobs/jobs.js
```diff
@@ -504,6 +509,13 @@ class Bars {
     return bar;
   }
 
+  playNotification() {
+    console.log('playing sound');
```
quisquous, line 513:
```suggestion
```

ui/jobs/jobs.js
```diff
@@ -504,6 +509,13 @@ class Bars {
...
+    const audio = new Audio('../../resources/sounds/freesound/alarm.ogg');
```
quisquous, line 514: Thanks for finding a sound!

ui/jobs/jobs.js
rcj: Thanks, I need to add a pre-commit hook to look for those...

resources/timerbox.ts
rcj: Pretty much. It is probably irrelevant, but as a callback could in theory do anything, including changing properties of the current timerbox, I wanted to ensure the current operation finishes without the need to potentially think about side effects from any of the callbacks.

Issue #3037, opened 2021-06-08
Question: Get Enemies' Location Realtime


MaikoTan:
As the title says, is there an exposed API from cactbot or OverlayPlugin to get the realtime location of enemies?

I know that currently we are using type `03: AddCombatant` LogLine that includes position, but only the position when it SHOWS/APPEARS. It is okay for mob hunting as those mobs would not run away from the appearing location so far, but when I want to track a enemy in dungeon, the logline methods doesn't work anymore.

valarnin:
You can call `getCombatants` on-demand to get info. See various triggers such as `E9S Clone The Art Of Darkness` or `LeviUn Slam Location`. Is there something else you're looking for beyond that?

MaikoTan:
> You can call `getCombatants` on-demand to get info. See various triggers such as `E9S Clone The Art Of Darkness` or `LeviUn Slam Location`. Is there something else you're looking for beyond that?

It is based on WebSocket message, right?
If I call it frequently like every .5 seconds or every 100 milliseconds, will it cause performance issues?

valarnin:
.5s should be fine, but .1s might be a bit too fast for some slower PCs. What's the scope of this? raidboss trigger, or some other component?

MaikoTan:
> .5s should be fine, but .1s might be a bit too fast for some slower PCs. What's the scope of this? raidboss trigger, or some other component?

A personal tracker used in solo Deep Dungeon, there are some scary mobs for a single player so I want to monitor their pace and decide whether I should hide myself.
Yeah, maybe .5 is okay for this use.

valarnin:
If it's personal you can run adjust timing as needed, maybe make sure you run it in a setTimeout loop and not a setInterval so that if it's delayed it doesn't immediately run again causing cascading lag?

valarnin:
This might be useful at some point for actual raidboss/etc functionality, if that ends up being the case I can handle setting up some sort of util that accepts essentially two callbacks, the first to get a request object, and the second to process the response and determine if it should keep watching, but that's a large chunk of code to write if it's not going to be used for anything yet.

MaikoTan:
Oh you're right!
I should change setInterval to setTimeout.

And another question, I saw that `getCombatant` receive a array of names, so that means we can track multiple enemies in a single request? Also If I sent a empty array will it return all the combatants that it recorded?

valarnin:
Yes to both of those questions.

Functionally:

[...]

PR #3099, opened 2021-06-15
raidboss: Convert Various Coil Triggers to TypeScript

panicstevenson:


panicstevenson:
I meant to open this as a draft, but obviously this can't be merged until https://github.com/quisquous/cactbot/pull/3081 goes in, since this relies on `initData`. This was mainly meant to be an exercise in seeing what else needs to happen between now an mass trigger conversion. Some take-aways for me, personally:

1. We can't convert anything that uses Regex currently.
1. It would be nice to have the ESLint rule that checks if MatchesX is the same as the NetRegex being used also ensure that matches has a type at all.
1. It would be really nice to have something linting the imports.
1. It would be nice to enforce the order of fields in a TriggerSet. Where should `initData` go? I thought above `timelineTriggers` (which is above `triggers`) made sense to me.
1. I find `export type Data = RaidbossData;` to be nicer than `export { RaidbossData as Data } from '../../../../../types/data';`
1. Goodbye `data.x = data.x || y` pattern?
1. The logic in `T13 Gigaflare Phase Change` felt a little weird with the `initData` addition?

quisquous:
Overall, I think this looks good to me.  I think the changes seem pretty minimal (which is good).

> I meant to open this as a draft

...I think you can change it after the fact?
 
> 1. We can't convert anything that uses Regex currently.

It is just the hp regex, right? Or are there others?

> 2. It would be nice to have the ESLint rule that checks if MatchesX is the same as the NetRegex being used also ensure that matches has a type at all.

:+1:

> 3. It would be really nice to have something linting the imports.

https://eslint.org/docs/rules/sort-imports?

Or were you imagining some other kind of linting.

> 4. It would be nice to enforce the order of fields in a TriggerSet. Where should `initData` go? I thought above `timelineTriggers` (which is above `triggers`) made sense to me.

Yes, I think after `timeline` but before `timelineTriggers` sounds great.  If somebody wanted to add a test (or maybe extend the trigger lint?) for this, that'd be good.

> Other, much less important stuff:
> 
> 1. I find `export type Data = RaidbossData;` to be nicer than `export { RaidbossData as Data } from '../../../../../types/data';` for things that don't use `data`.

:+1:

> 2. Goodbye `data.x = data.x || y` pattern?

HECK YES

> 3. The logic in `T13 Gigaflare Phase Change` felt a little weird with the `initData` addition so I updated it.

:+1:

valarnin:
For the `hasHP` issue, has anyone reached out to Ravahn to see if that's a line that they're willing to add as a network log event?

If not, would it make sense to make some sort of `getCombatants` helper that can be used as a trigger promise to monitor HP instead?

Something like:

```
      id: 'T7 Phase 2',
      netRegex: NetRegexes.startsUsing({ name: 'Melusine', capture: false }),
      suppressSeconds: 9999,
      promise: (data, matches) => watchCombatant(
          { ids: [matches.id], every: 100 },
          (combatant) => (combatant.CurrentHP / combatant.MaxHP) <= 79,
      ),
    },
```

For example. `watchCombatant` would return a promise which would fetch combatant data every `every` ms, run the `(combatant)` function, and if that function returns true, resolve the promise.

It's similar to what I outlined in #3037

panicstevenson:
> ...I think you can change it after the fact?

Neat!

> https://eslint.org/docs/rules/sort-imports?
>
> Or were you imagining some other kind of linting.

This should be sufficient, I'd think.

> It is just the hp regex, right? Or are there others?

I think hasHP in general is the only thing in Regexes that isn't in NetRegexes (with nameToggle being the only one in the opposite camp), so yea.

> For the hasHP issue, has anyone reached out to Ravahn to see if that's a line that they're willing to add as a network log event?

I certainly haven't reached out, but these logs can be pretty noisy so I think I would be surprised.

> If not, would it make sense to make some sort of getCombatants helper that can be used as a trigger promise to monitor HP instead?

I think this could work; I also don't know that the wheel needs to be reinvented if we're willing to accept `regex` into `TriggerSet`. I can imagine that's not something we'd be too keen on doing though. I also imagine we're not too excited about the prospect of dropping all regexes aside from `hasHP` to shave down the surface area between regexes and triggersets. If both of those are non-starters, then we can look into dropping the triggers (possible? HP pushes aren't really much of a thing anymore in the sense that there's usually an ability, dialogue line, or _something_ to clue you in that the boss has phased) and you could argue that people doing old content seriously could learn the phase percentages now that HP percent is easily visible, and that anyone else doing old content is probably breezing passed it to a point that triggers and timelines are approaching less-than-useful-noise territory. Or, we can come up with an alternative, in which case I think what you've listed is really nice. I think I'd lower the polling to be more in the "human perceivable" range of ms, since I don't think anyone is going to be upset that their phase change came in 500ms late, but that's more of a nit on a proof of concept idea than anything.

valarnin:
> I think I'd lower the polling to be more in the "human perceivable" range of ms, since I don't think anyone is going to be upset that their phase change came in 500ms late, but that's more of a nit on a proof of concept idea than anything.

Yeah, not really obvious from the proof of concept but delay would be an optional parameter with a sane default of 1000ms or 5000ms or something.

Also the helper is deliberately designed to support checking against any combatant property so it's not just for HP, though that's the most obvious use. It could also track and check against position for example.

Realistically the first param should probably be defined as something like:

```typescript
type WatchCombatantFunc = ({
  ids?: number[];
  names?: string[];
  props?: string[];
  delay?: number;
}, (combatant: PluginState) => boolean);
```

Where `PluginState` is from raidemulator and should be moved up to event.d.ts?

quisquous:
Yeah, I think a helper here could be nice.  I think there is a huge amount of logs when a bunch of adds detonate from 100->1 and I feel like I've subjectively experienced a frame hiccup at that point.  Being able to tell users that they can turn this field off (or being able to eliminate one other option that cactbot users have to care about) would be a positive thing.

valarnin:
I can open a separate PR with a helper like I described above, would it make sense to put it in `util.ts`?

panicstevenson:
> I can open a separate PR with a helper like I described above, would it make sense to put it in `util.ts`?

It's a raidboss only thing, so I don't know if there's a home for raidboss utils, but if there should be then now would be a good time to break one out; otherwise, I think utils is _probably_ fine? quisquous is probably more opinionated here.

quisquous:
> > I can open a separate PR with a helper like I described above, would it make sense to put it in `util.ts`?
> 
> It's a raidboss only thing, so I don't know if there's a home for raidboss utils, but if there should be then now would be a good time to break one out; otherwise, I think utils is _probably_ fine? quisquous is probably more opinionated here.

I don't really have a better place to put it, and `Utils` is already included for user raidboss files.

valarnin:
Hit an annoying issue with implementing this, basically "how do we cancel this on reset". Probably, this would need to be tracked in an array of rejectors with a util method to clear them as well.

PR #3102, opened 2021-06-15
add ts indent rule

Trim21:
There are some More complicated config can tell union to indent
https://stackoverflow.com/questions/59851672/eslint-indent-and-ignorenodes-trouble-getting-ast-selectors-to-work-correctl/59852368#59852368

But I think it's a waste of time to write a complicated rule for code indent……

### DIFF
--- a/types/event.d.ts
+++ b/types/event.d.ts
@@ -288,7 +288,7 @@ interface CactbotLoadUserRet {
 
 // Structured JSON data saved in OverlayPlugin config files.
 export type SavedConfigEntry = string | number | boolean | [ SavedConfigEntry] |
-   { [nestedName: string]: SavedConfigEntry };
+{ [nestedName: string]: SavedConfigEntry };
 export type SavedConfig = {
   [overlayName: string]: SavedConfigEntry;
 };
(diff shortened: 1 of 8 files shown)


Trim21:
any tools to format code except eslint?

same eslint lint rules can get may get many different indent.

MaikoTan:
prettier
with some configurations.

Trim21:
There are some rule of our eslint can't be disabled in prettier.

Trim21:
after prettier 
image

panicstevenson:
Yea, prettier is kind of a non-starter since the quote thing is non-customizable and we parse JavaScript. Also, I personally just don't like that rule.
I looked into https://www.npmjs.com/package/prettier-eslint a little bit, but it still has prettier issues, namely around the printWidth wanting to auto format things to various lengths.

types/event.d.ts
```diff
@@ -311,9 +311,9 @@ interface PlayerChangedRet {
   // Potentially add the job to the jobDetail passed back from the C# plugin, and use
   // that information to decide the type
   jobDetail: JobDetail['PLD'] & JobDetail['WAR'] & JobDetail['DRK'] & JobDetail['GNB'] & JobDetail['WHM'] &
-    JobDetail['SCH'] & JobDetail['AST'] & JobDetail['PGL'] & JobDetail['MNK'] & JobDetail['DRG'] &
-    JobDetail['NIN'] & JobDetail['SAM'] & JobDetail['BRD'] & JobDetail['MCH'] & JobDetail['DNC'] &
-    JobDetail['THM'] & JobDetail['BLM'] & JobDetail['ACN'] & JobDetail['SMN'] & JobDetail['RDM'];
+  JobDetail['SCH'] & JobDetail['AST'] & JobDetail['PGL'] & JobDetail['MNK'] & JobDetail['DRG'] &
```
quisquous, line 314: Similarly, why is this not indented either?

All the other indents seem fine except for ones that end in `|` or `&`

Trim21:
Yes, I find that this is a broken case of `typescript-eslint`.

Maybe we shouldn't use this rule.

quisquous:
> Yes, I find that this is a broken case of `typescript-eslint`. and maintainer suggest not using this rule.

I liked all the other changes though.  Didn't the stack overflow link you posted up above have an example of how to fix this?

Trim21:
No, it just ignores certain AST nodes like union or something else.

quisquous:
Sure but that seems ok-ish? At least the other things are indented properly.

Trim21:
Which means it can be indented with any indent I think...



------------------&nbsp;Original&nbsp;------------------
From: ***@***.***&gt;; 
Date: 2021年6月16日(星期三) 下午2:01
To: ***@***.***&gt;; 
Cc: ***@***.***&gt;; "State ***@***.***&gt;; 
Subject: Re: [quisquous/cactbot] add ts indent rule (#3102)





 
Sure but that seems ok-ish? At least the other things are indented properly.

quisquous:
Yeah, I'm saying the same thing.  I guess I'm saying if adding this rule fixes some things but not everything, that's ok.  It's still an improvement?  Unless there's some other eslint plugin that fixes all of this and would be better to use instead?

PR #3107, opened 2021-06-15
raidemulator: Update job fields to use Job type

valarnin:
I feel like there's a better way to handle CombatantJobSearch but I can't see it?

This is in prep for converting PopupTextAnalysis to typescript. I'll probably visit the JobDetail type next since that's also causing some annoyances with conversion.

### DIFF
--- a/ui/raidboss/emulator/data/CombatantJobSearch.ts
+++ b/ui/raidboss/emulator/data/CombatantJobSearch.ts
@@ -3,8 +3,10 @@ import { Job } from '../../../../types/job';
 export default class CombatantJobSearch {
   static getJob(abilityId: number): Job | undefined {
     for (const [key, value] of Object.entries(CombatantJobSearch.abilities)) {
-      if (value.includes(abilityId))
-        return key as Job;
+      if (value) {
+        if (value.includes(abilityId))
+          return key as Job;
+      }
     }
   }
 
--- a/ui/raidboss/emulator/data/CombatantJobSearch.ts
+++ b/ui/raidboss/emulator/data/CombatantJobSearch.ts
@@ -1,14 +1,20 @@
+import { Job } from '../../../../types/job';
+
 export default class CombatantJobSearch {
-  static getJob(abilityId: number): string | undefined {
-    for (const job in CombatantJobSearch.abilities) {
-      if (CombatantJobSearch.abilities[job]?.includes(abilityId))
+  static getJob(abilityId: number): Job | undefined {
+    for (const key in CombatantJobSearch.abilities) {
+      const job = key as Job;
+      const list = CombatantJobSearch.abilities[job];
+      if (!list)
+        continue;
+      if (list.includes(abilityId))
         return job;
     }
   }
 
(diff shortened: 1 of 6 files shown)


panicstevenson:
> I feel like there's a better way to handle CombatantJobSearch but I can't see it?

You could spend the upfront time to create a map of <abilityId, Job>; seems kind of unreasonable, but it might not be given that you could have the map registered before loading any fights. Otherwise, you could have the abilityIds be a Set so you could `.has()`, or you could make all the abilityIds strings and use a Trie 👀. Alternatively, you could store a local variable, run through everything, and `return foundJob ?? 'NONE';`, which might help any !job checks.

valarnin:
I like this suggestion a lot more than what I have. The only difference is that it shouldn't return 'NONE', because that breaks the downstream logic.

Performance for this function isn't that important, it's more important that the map of jobs to ability IDs is maintainable. This function gets called at most 40 times per encounter but most likely it won't get called at all. This function is primarily used to determine the job of a player if they don't fire a line event that includes job ID during the encounter, and it'll only be called if a job hasn't been determined yet (using this function or a line that includes a job ID).

panicstevenson:
> I like this suggestion a lot more than what I have. The only difference is that it shouldn't return 'NONE', because that breaks the downstream logic.

That's fair; I wasn't sure if there was a reason to have it be undefined or if it made more sense for it to always return a Job, even if that job is none.

(Also sorry for the bug!)

ui/raidboss/emulator/data/CombatantJobSearch.ts
```diff
@@ -1,14 +1,18 @@
...
-  static getJob(abilityId: number): string | undefined {
-    for (const job in CombatantJobSearch.abilities) {
-      if (CombatantJobSearch.abilities[job]?.includes(abilityId))
-        return job;
+  static getJob(abilityId: number): Job | undefined {
+    for (const [key, value] of Object.entries(CombatantJobSearch.abilities)) {
+      if (value) {
+        if (value.includes(abilityId))
+          return key as Job;
+      }
```
panicstevenson, line 9: Can this just be shortened to:
```suggestion
      if (value?.includes(abilityId))
        return key as Job;
```
?

ui/raidboss/emulator/data/CombatantJobSearch.ts
```diff
@@ -1,14 +1,16 @@
...
+
 export default class CombatantJobSearch {
-  static getJob(abilityId: number): string | undefined {
-    for (const job in CombatantJobSearch.abilities) {
-      if (CombatantJobSearch.abilities[job]?.includes(abilityId))
-        return job;
+  static getJob(abilityId: number): Job | undefined {
+    for (const [key, value] of Object.entries(CombatantJobSearch.abilities)) {
+      if (value?.includes(abilityId))
+        return key as Job;
```
quisquous, line 7:
```suggestion
        return key as keyof typeof CombatantJobSearch.abilities;
```

I realize it's like two lines below, but at least this validates that if you changed the type of `abilities` you would also have to fix this up to be correct as well.

PR #3108, opened 2021-06-15
other: Rework PlayerChangedRet to auto-infer jobDetail type

valarnin:
This PR reworks the way `PlayerChangedRet` types `job` and `jobDetail` to automatically infer the type of `jobDetail` based on the value of `job`.

This is using some circular infer syntax that I'm not 100% solid on so please give this one a thorough review, and test some of the Jobs overlays.

I tested this against the Cactbot Test overlay and that was picking up changes fine for WHM/WAR/DNC, which are the only three classes I can properly test with.

If `job` is checked against a job that doesn't have a defined JobDetail, it will type `jobDetail` to unknown, as such:

image

### DIFF
--- a/types/event.d.ts
+++ b/types/event.d.ts
@@ -306,10 +306,6 @@ type PlayerChangedRet = Job extends infer T ? T extends Job ? {
   currentGP: number;
   maxGP: number;
   currentShield: number;
-  // TODO: Is there a cleaner way to do this? It would be better if there were a way to
-  // determine which job was passed in with the event and explicitly use that JobDetail
-  // Potentially add the job to the jobDetail passed back from the C# plugin, and use
-  // that information to decide the type
   jobDetail: JobDetail[T];
   pos: {
     x: number;
--- a/types/event.d.ts
+++ b/types/event.d.ts
@@ -293,9 +293,9 @@ export type SavedConfig = {
   [overlayName: string]: SavedConfigEntry;
 };
 
-interface PlayerChangedRet {
+type PlayerChangedRet = Job extends infer T ? T extends Job ? {
   name: string;
-  job: Job;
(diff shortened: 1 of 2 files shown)


types/event.d.ts
```diff
@@ -310,10 +310,7 @@ interface PlayerChangedRet {
   // determine which job was passed in with the event and explicitly use that JobDetail
   // Potentially add the job to the jobDetail passed back from the C# plugin, and use
   // that information to decide the type
-  jobDetail: JobDetail['PLD'] & JobDetail['WAR'] & JobDetail['DRK'] & JobDetail['GNB'] & JobDetail['WHM'] &
-    JobDetail['SCH'] & JobDetail['AST'] & JobDetail['PGL'] & JobDetail['MNK'] & JobDetail['DRG'] &
-    JobDetail['NIN'] & JobDetail['SAM'] & JobDetail['BRD'] & JobDetail['MCH'] & JobDetail['DNC'] &
-    JobDetail['THM'] & JobDetail['BLM'] & JobDetail['ACN'] & JobDetail['SMN'] & JobDetail['RDM'];
+  jobDetail: JobDetail[T];
```
panicstevenson, line 313: You can delete my long-winded comment above that says "I'll defer to a wizard," as the wizard has arrived.

valarnin: I'm hardly a wizard, just an acolyte who found the right spellbook in the library 😉

MaikoTan: If T of `T extends Job` is not the key in `JobDetail`, would it be `never` or something?

MaikoTan (approved):

quisquous:
Holy moly you fixed it.  :tada:

PR #3110, opened 2021-06-16
lint: Add Eslint Import Plugin for Auto-Sort

panicstevenson:


panicstevenson:
As mentioned in https://github.com/quisquous/cactbot/pull/3099

The `sort-imports` rule doesn't have an auto-fix, so that feels like a non-starter. There were other plugins that solve something similar, but this one seems to have a couple of useful knobs. The only oddity in my opinion is that things are sorted by `from` instead of the actual import, but it should lead to less overall churn, since adding an import early in the alphabet won't cause everything to need to be rearranged.

Open to opinions/criticism.

ui/raidboss/raidboss.ts
```diff
@@ -1,16 +1,13 @@
...
-
-import UserConfig from '../../resources/user_config';
 import { addRemotePlayerSelectUI } from '../../resources/player_override';
+import UserConfig from '../../resources/user_config';
 import raidbossFileData from './data/raidboss_manifest.txt';
-
+import { PopupText, PopupTextGenerator } from './popup-text';
 import Options from './raidboss_options';
+import { TimelineController, TimelineLoader, TimelineUI } from './timeline';
+
```
quisquous, line 8: Do these two below blocks of import not count for newline-between: never?

panicstevenson: Ah, so no; this plugin never touches "imports with side effects"; although they do need to be clumped together. I went with putting them below, following the import order above and putting *script before css/assets.

Open to thoughts

Trim21:
maybe add rule to our scripts like auto-label.cjs?

ui/jobs/bar.d.ts
```diff
@@ -1,8 +1,7 @@
-// TODO: remove this when ui/jobs/jobs.js is converted to typescript
```
Trim21, line 1: Why this comment is removed

MaikoTan:
I'd prefer the groups and/or orders like
```
node libraries
third parties' libraries (in node_modules)
modules from other packages/modules (like common utils)
modules in itself
```

ui/jobs/bar.d.ts
panicstevenson: Hmmm, weird. I didn't delete it, so it must have, but it dealt with other comments just fine? I'll add it back, but good catch.

panicstevenson:
This now has a newline between each group; which distinguishes between internal, external, parents, and siblings. I kind of think the distinction between parents and siblings isn't too useful? But, I'll let someone else say if they have opinions. Also, this is the default ordering.

ui/raidboss/raidboss.ts
quisquous: Ah no, mostly was just curious how that worked.  It makes sense not to resort those, I guess.

quisquous:
The newlines definitely help break it up a little bit.  I think the ordering seems fine as it is now, but I am not that opinionated on what the best ordering is here.

Do @MaikoTan or @Trim21 (or other folks?) have more opinions on what they'd prefer, or if this is looking good?

MaikoTan:
LGTM!

ui/jobs/buff_tracker.ts
```diff
@@ -1,38 +1,36 @@
...
-import astralImage from '../../resources/ffxiv/status/astral.png';
-import umbralImage from '../../resources/ffxiv/status/umbral.png';
-import physicalImage from '../../resources/ffxiv/status/physical.png';
-import offguardImage from '../../resources/ffxiv/status/offguard.png';
-import peculiarLightImage from '../../resources/ffxiv/status/peculiar-light.png';
-import trickAttackImage from '../../resources/ffxiv/status/trick-attack.png';
-import battleLitanyImage from '../../resources/ffxiv/status/battle-litany.png';
-import emboldenImage from '../../resources/ffxiv/status/embolden.png';
 import arrowImage from '../../resources/ffxiv/status/arrow.png';
+import astralImage from '../../resources/ffxiv/status/astral.png';
```
panicstevenson, line 3: Explicit callout for this file, if people care.

MaikoTan: What's wrong?

panicstevenson: Nothing is wrong, just pointing out this one since it has the img imports

panicstevenson:
Apologies for force push; I thought it'd be easier to reset against main and run the linter rather than wrangle any changes that have happened in the last week.

types/net_props.d.ts
```diff
@@ -1,6 +1,5 @@
 import { NetFields } from './net_fields';
 
-
 type Params<T extends string> =
```
panicstevenson, line 3: Technically this is my change; I came through looking for `^import.*\n\n` since there were a couple places where the auto fix didn't do a great job at cleaning the mess it made.

ui/raidboss/raidboss.ts
```diff
@@ -1,16 +1,14 @@
...
 import { PopupText, PopupTextGenerator } from './popup-text';
+import defaultOptions from './raidboss_options';
 import { TimelineController, TimelineLoader, TimelineUI } from './timeline';
 
-import UserConfig from '../../resources/user_config';
-import { addRemotePlayerSelectUI } from '../../resources/player_override';
-import raidbossFileData from './data/raidboss_manifest.txt';
-
-import defaultOptions from './raidboss_options';
+import '../../resources/timerbar';
```
panicstevenson, line 10: Just to reiterate, this is my change; side effect changes come last, and I split them into JS (following the same order enforced elsewhere by the plugin), and then new line followed by CSS (also following the same order).

PR #3111, opened 2021-06-16
raidboss: Add helper to watch combatants

valarnin:
This PR adds a `Util.watchCombatants` helper method to monitor combatants, as well as a test trigger showing the usage.

As discussed in #3099

### DIFF
--- a/resources/util.ts
+++ b/resources/util.ts
@@ -84,7 +84,6 @@ type WatchCombatantParams = {
   names?: string[];
   props?: string[];
   delay?: number;
-  maxRetries?: number;
   maxDuration?: number;
 };
 
@@ -93,7 +92,6 @@ type WatchCombatantFunc = (params: WatchCombatantParams,
 
 type WatchCombatantMapEntry = {
   cancel: boolean;
-  tries: number;
   start: number;
 };
 
@@ -102,8 +100,6 @@ const watchCombatantMap: WatchCombatantMapEntry[] = [];
 const shouldCancelWatch =
   (params: WatchCombatantParams, entry: WatchCombatantMapEntry): boolean => {
     if (entry.cancel)
-      return true;
-    if (params.maxRetries !== undefined && entry.tries > params.maxRetries)
       return true;
     if (params.maxDuration !== undefined && Date.now() - entry.start > params.maxDuration)
       return true;
@@ -129,14 +125,12 @@ const watchCombatant: WatchCombatantFunc = (params, func) => {
 
     const entry: WatchCombatantMapEntry = {
       cancel: false,
-      tries: 0,
       start: Date.now(),
     };
 
     watchCombatantMap.push(entry);
 
     const checkFunc = () => {
-      ++entry.tries;
       if (shouldCancelWatch(params, entry)) {
         rej();
         return;
--- a/resources/util.ts
+++ b/resources/util.ts
@@ -89,54 +89,47 @@ type WatchCombatantFunc = (params: {
 const watchCombatantMap: { cancel: boolean }[] = [];
 
 const watchCombatant: WatchCombatantFunc = (params, func) => {
-  let resolver: (value: boolean) => void;
-  let rejecter: () => void;
-  const promise = new Promise<boolean>((res, rej) => {
-    resolver = res;
-    rejecter = rej;
-  });
+  return new Promise<boolean>((res, rej) => {
+    const delay = params.delay ?? 1000;
 
-  const delay = params.delay ?? 1000;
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/00-misc/test.js
```diff
@@ -242,6 +243,33 @@ export default {
         };
       },
     },
+    {
+      id: 'Test Watch',
+      netRegex: NetRegexes.echo({ line: 'cactbot test watch.*?', capture: false }),
+      promise: (data) => Util.watchCombatant({ names: [data.me, 'Striking Dummy'] }, (ret) => {
+        const me = ret.combatants.filter((c) => c.Name === data.me)[0];
```
panicstevenson, line 250:
```suggestion
        const me = ret.combatants.find((c) => c.Name === data.me);
```

panicstevenson:
Neat!

ui/raidboss/data/00-misc/test.js
```diff
@@ -242,6 +243,33 @@ export default {
         };
       },
     },
+    {
+      id: 'Test Watch',
+      netRegex: NetRegexes.echo({ line: 'cactbot test watch.*?', capture: false }),
+      promise: (data) => Util.watchCombatant({ names: [data.me, 'Striking Dummy'] }, (ret) => {
```
quisquous, line 249: Can you put a `LocaleText` of striking dummy names at the top of the file so this can be translated?

resources/util.ts
```diff
@@ -77,6 +79,66 @@ const jobToRoleMap: Map<Job, Role> = (() => {
...
+  names?: string[];
+  props?: string[];
+  delay?: number;
+}, func: (ret: GetCombatantsRet) => boolean) => Promise<boolean>;
+
+const watchCombatantMap: { cancel: boolean }[] = [];
+
+const watchCombatant: WatchCombatantFunc = (params, func) => {
+  let resolver: (value: boolean) => void;
+  let rejecter: () => void;
```
Trim21, line 93: Why not just wrap all things in promise initialization?

valarnin: Originally I was storing off the rejecter in the watchCombatantMap, but that ended up not really working out. Now that it's not needed outside the scope of the promise, I've changed it as suggested.

Trim21:
A max retries options?

valarnin:
> 
> 
> A max retries options?

Should be pretty easy to add, but I'm not sure it makes sense more than, for example, a max duration option? Both of these options are also implementable on the trigger side.

Also keep in mind that these get cancelled on fight reset automatically.

valarnin:
I decided that to make supporting this easier in raidemulator in the log run, I'll implement max retries and max duration in the helper itself.

MaikoTan:
I think it is worth to add a function that calculate Euclidean distance in Utils?

valarnin:
> 
> 
> I think it is worth to add a function that calculate Euclidean distance in Utils?

Something like this?

```typescript
  getDistance: (c1: PluginCombatantState, c2: PluginCombatantState): number => {
    const distX = Math.abs(c1.PosX - c2.PosX);
    const distY = Math.abs(c1.PosY - c2.PosY);
    return Math.hypot(distX, distY);
  },
```

Probably useful but I don't know if it should go in this PR?

MaikoTan:
Maybe to put it in a new PR is better, and I think that not only calculations with 2D coordinates (x/y) but also 3D (x/y/z) coordinates, (i.e `getDistance2D(c1, c2)` and `getDistance3D(c1, c2)`)?

resources/util.ts
```diff
@@ -77,6 +79,84 @@ const jobToRoleMap: Map<Job, Role> = (() => {
   return map;
 })();
 
+type WatchCombatantParams = {
+  ids?: number[];
+  names?: string[];
+  props?: string[];
+  delay?: number;
+  maxRetries?: number;
+  maxDuration?: number;
```
quisquous, line 88: What's the use case for both of these? It seems like they are saying the same thing in different units.

valarnin: maxRetries doesn't account for execution time of the overlay call, so if the call takes 50ms, after 20 retries that's a full second's worth of skew. I'd think just `maxDuration` should be sufficient but I'm not sure what use case @Trim21 was thinking of when they suggested adding retries, so I opted to add both options.

quisquous: Ah, I see how they're not exactly equivalent.  It still seems like solving roughly the same problem.  Maybe `maxDuration` is the more conservative estimate and that on its own is good enough?  I can't quite imagine why somebody would need `maxRetries` precisely.

PR #3112, opened 2021-06-16
lint: Create ESLint Utils for Ordering Functions

panicstevenson:


panicstevenson:
Extract helpful functions out for other ordering rules that we want to auto-fix.

PR #3113, opened 2021-06-16
raidemulator: Convert AnalyzedEncounter to typescript

valarnin:
This PR also adds nullable to the jobDetail property since that's a possible outcome, see https://github.com/ngld/OverlayPlugin/blob/master/OverlayPlugin.Core/MemoryProcessors/FFXIVProcessIntl.cs#L250

Not very happy with how `updateState` came out but I couldn't see a different way to handle it other than making jobDetail completely optional, which wouldn't reflect OverlayPlugin. Maybe someone else can see how to handle it better?


ui/raidboss/emulator/data/AnalyzedEncounter.ts
```diff
@@ -0,0 +1,225 @@
...
+          x: state.posX,
+          y: state.posY,
+          z: state.posZ,
+        },
+        rotation: state.heading,
+        bait: 0,
+        debugJob: '',
+      },
+    };
+    // For emulator purposes, jobDetail is always null, but typescript really doesn't like that
```
quisquous, line 101: I FIGURED IT OUT.  \*\o/\*

The issue is not `jobDetail` as you might suspect, but instead the issue is `job`.  Because `Job` is a union type, `T extends Job` distributes the `Job` union across all possibilities leading to an incompatible / impossible type to initialize.

This is the error:
```
      TS2322: Type '{ name: string; job: Job; level: number; currentHP: number;
maxHP: number; currentMP: number; maxMP: number; currentCP: number; maxCP: numbe
r; currentGP: number; maxGP: number; currentShield: number; ... 4 more ...; debu
gJob: string; }' is not assignable to type '{ name: string; job: "NONE"; level:
number; currentHP: number; maxHP: number; currentMP: number; maxMP: number; curr
entCP: number; maxCP: number; currentGP: number; maxGP: number; currentShield: n
umber; ... 4 more ...; debugJob: string; } | ... 37 more ... | { ...; }'.
  Type '{ name: string; job: Job; level: number; currentHP: number; maxHP: numbe
r; currentMP: number; maxMP: number; currentCP: number; maxCP: number; currentGP
: number; maxGP: number; currentShield: number; ... 4 more ...; debugJob: string
; }' is not assignable to type '{ name: string; job: "DNC"; level: number; curre
ntHP: number; maxHP: number; currentMP: number; maxMP: number; currentCP: number
; maxCP: number; currentGP: number; maxGP: number; currentShield: number; ... 4
more ...; debugJob: string; }'.
    Types of property 'job' are incompatible.
      Type 'Job' is not assignable to type '"DNC"'.
        Type '"NONE"' is not assignable to type '"DNC"'.
```

You can verify that job is the issue because you can say some nonsense like `job: job as 'DNC'` and it will then become the correct type.

You can fix the underlying issue with `type PlayerChangedRet = Job extends infer T ? [T] extends [Job] ? {`.

I ran into this same thing earlier this week and https://www.reddit.com/r/typescript/comments/gmjfxe/weirdness_with_booleans/ solved it then as well.

valarnin: That doesn't actually fix the problem. `jobDetail` just ends up always typed as `unknown` even when testing `job`.

image
image

quisquous: Whoooooooops.  That's what I get for posting something before bed.  I think I get it now.

(Explaining this to myself out loud for future me who is reading this, since you likely already know it.)  You need the union distributed because that's what causes it to be a discriminated union that's able to tell the types apart.  Using the brackets makes it no longer a discriminated union, so `jobDetail` becomes unknown again.  Unfortunately you can't easily construct discriminated unions from variables, because they need to be statically typed in some way (e.g. [this problem](https://stackoverflow.com/questions/51064527/defining-a-variable-as-one-variant-of-a-discriminated-union-in-typescript).

The only option I could come up with after some (minor) sleep is to add another case where we use `null` as another case in the discriminated union.  In other words, if you set `jobDetail: null` then `job` can be anything.  This allows you to at least construct your one case (but doesn't solve synthetically generating anything with a real `jobDetail`), e.g.

```typescript
type PlayerChangedJobDetails<T> = {
  job: T;
  jobDetail: JobDetail[T];
} | {
  job: Job;
  jobDetail: null;
}

type PlayerChangedBase = {
  name: string;
  level: number;
  currentHP: number;
  maxHP: number;
  currentMP: number;
  maxMP: number;
  currentCP: number;
  maxCP: number;
  currentGP: number;
  maxGP: number;
  currentShield: number;
  pos: {
    x: number;
    y: number;
    z: number;
  };
  rotation: number;
  bait: number;
  debugJob: string;
};

type PlayerChangedRet = Job extends infer T ? T extends Job ?
  PlayerChangedJobDetails<T> & PlayerChangedBase : never : never;
```

valarnin: This actually seems to work for constructing one wholesale as well. e.g. the following work.

Made this adjustment and removed the casting.

```typescript
    const detail: PlayerChangedDetail = {
      detail: {
        name: combatant.name,
        job: 'WHM',
        level: combatant.level ?? 0,
        currentHP: state.hp,
        maxHP: state.maxHp,
        currentMP: state.mp,
        maxMP: state.maxMp,
        currentCP: 0,
        maxCP: 0,
        currentGP: 0,
        maxGP: 0,
        currentShield: 0,
        jobDetail: {
          lilyStacks: 0,
          lilyMilliseconds: 0,
          bloodlilyStacks: 0,
        },
        pos: {
          x: state.posX,
          y: state.posY,
          z: state.posZ,
        },
        rotation: state.heading,
        bait: 0,
        debugJob: '',
      },
    };
```

```typescript
    const job: string = 'WHM';
    const detail = {
      name: combatant.name,
      level: combatant.level ?? 0,
      currentHP: state.hp,
      maxHP: state.maxHp,
      currentMP: state.mp,
      maxMP: state.maxMp,
      currentCP: 0,
      maxCP: 0,
      currentGP: 0,
      maxGP: 0,
      currentShield: 0,
      pos: {
        x: state.posX,
        y: state.posY,
        z: state.posZ,
      },
      rotation: state.heading,
      bait: 0,
      debugJob: '',
    };
    if (job === 'WHM') {
      const fullDetail: PlayerChangedDetail = {
        detail: {
          ...detail,
          job: 'WHM',
          jobDetail: {
            lilyStacks: 0,
            lilyMilliseconds: 0,
            bloodlilyStacks: 0,
          },
        },
      };
    }
```

quisquous: > This actually seems to work for constructing one wholesale as well. e.g. the following work.

I could be wrong, but I think those work (and would have worked before my suggestion) because in both cases TypeScript knows statically that job is `WHM`.  I think the heart of the issue is that you can't assign a job whose type is just `Job`.  If the `Job` is literally the string `WHM` or you do a conditional where you verify that it is `WHM` then that's enough information to know statically.

ui/raidboss/emulator/data/AnalyzedEncounter.ts
```diff
@@ -1,66 +1,132 @@
...
-      party: this.encounter.combatantTracker.partyMembers.map((ID) => {
-        return {
-          name: this.encounter.combatantTracker.combatants[ID].name,
-          job: Util.jobToJobEnum(this.encounter.combatantTracker.combatants[ID].job),
-          inParty: true,
-        };
-      }),
-    });
-    this.popupText.OnPlayerChange({
+  selectPerspective(ID: string): void {
```
quisquous, line 41: Do you want to change this ID while you're here (or another PR)?

ui/raidboss/emulator/data/AnalyzedEncounter.ts
```diff
@@ -0,0 +1,220 @@
...
+      const event = {
+        type: 'onLogEvent',
+        detail: {
+          logs: [log],
+        },
+      };
+
+      await popupText.OnLog(event as EmulatorLogEvent);
+      await popupText.OnNetLog(event as EmulatorNetworkLogEvent);
+      timelineController.OnLogEvent(event as EmulatorLogEvent);
```
quisquous, line 216: I would like to avoid these as well.  I tried looking at the definitions of LogEvent and EmulatorLogEvent and I'm a little bit confused how to disentangle it.  Redefining logs as two totally different types is a little bit suspicious.  I'll take a look tomorrow when I have a little bit more brain.

valarnin: I could do some refactoring to avoid this but I'd have to modify all of the following files to compensate:

- raidemulator.js
- AnalyzedEncounter.ts
- PopupTextAnalysis.ts
- RaidEmulatorPopupText.js
- RaidEmulatorTimelineController.ts
- StubbedPopupText.ts

Basically I'd stub out `OnLog` and `OnNetLog` and equivalents in the parent classes, and replace existing emulator overrides for `OnLog` and `OnNetLog` with new `OnEmulatorLog` calls. Do you want me to do that as part of this PR?

valarnin:
I'll have to rebase this to main to pick up #3127 after it lands.

valarnin:
Rebased. The only new changes since last review were to update `ID` to `id`

ui/raidboss/emulator/data/AnalyzedEncounter.ts
valarnin: I'm going to mark this as resolved, since it's fixed from a code perspective for this PR. Might be something to revisit later on though.

ui/raidboss/emulator/data/AnalyzedEncounter.ts
valarnin: I've done this refactor with the latest commit. Please re-review and let me know if this is a better solution.

PR #3115, opened 2021-06-16
ui: Stop using global Options variables everywhere

quisquous:
This is a partial version of #2857.  Not dealing with globals just
seems like a good refactoring to have.

The lint is changed for rdmty because changing a file that is ignored
is a warning and we have a maximum number of warnings of zero.
So, instead eslint-disable is added to that file.


ui/raidboss/raidemulator.js
```diff
@@ -41,7 +41,7 @@ import './raidemulator.css';
   let logConverterWorker;
 
   document.addEventListener('DOMContentLoaded', () => {
-    emulator = new RaidEmulator(Options);
+    emulator = new RaidEmulator();
```
valarnin, line 44: This change (and associated `RaidEmulator` class changes I feel aren't the best way to address this? It seems to me that it would make more sense to move the block of `new <some class>` constructor calls and associated `...on(` event listener binds down to be within the event handler for loading config instead, minus the `Persistor` constructor?

Otherwise this makes `RaidEmulator` not follow the general pattern of the rest of the code, where it accepts the options in the constructor.

quisquous: Yeah, agreed.  I think `RaidEmulator` already isn't following the pattern of the rest of the code to defer its constructions, so that seems like a good suggestion all around.

ui/fisher/fisher.js
```diff
@@ -147,8 +146,8 @@ class Fisher {
       },
     };
 
-    this.ui = new FisherUI(element, Options);
-    this.seaBase = new SeaBase(Options);
+    this.ui = new FisherUI(element, this.options);
```
MaikoTan, line 149: I think the `options` parameter should be the first one? Just like others do.

ui/oopsyraidsy/oopsyraidsy.js
```diff
@@ -228,12 +228,12 @@ Examples:
 
 /* eslint-enable */
 
-function ShortNamify(name) {
+function ShortNamify(name, options) {
```
MaikoTan, line 231: Maybe just to pass in `PlayerNicks` is better in this usage.

ui/fisher/fisher.js
quisquous: Done, thanks for the suggestion!

ui/raidboss/raidemulator.js
quisquous: Done.

ui/oopsyraidsy/oopsyraidsy.js
quisquous: Done, thanks!

ui/raidboss/raidemulator.js
```diff
@@ -29,6 +29,7 @@ import defaultOptions from './raidboss_options';
 import '../../resources/defaults.css';
 import './raidemulator.css';
 
+// TODO: why is this IIFE, and why are these lets not just consts in the addEventListener?
```
valarnin, line 32: If I remember correctly, before the typescript conversion process and bundling, the IIFE was required to ensure that this executed after all the JS loaded, to allow the various classes to be available. Probably not required anymore?

quisquous: Yeah, I don't think it is.  I'll follow up after this and remove it.

PR #3116, opened 2021-06-16
lint: Add ESLint for Trigger Property Order

panicstevenson:


ui/oopsyraidsy/data/03-hw/raid/a12n.js
```diff
@@ -44,8 +44,8 @@ export default {
     {
       id: 'A12N Assault Cleanup',
       netRegex: NetRegexes.gainsEffect({ effectId: '461' }),
-      suppressSeconds: 5,
       delaySeconds: 20,
+      suppressSeconds: 5,
```
panicstevenson, line 48: This is a bit concerning, but I'm not sure of a great way (other than the obvious splitting ESLint rules across oopsyraidsy and raidboss) to prevent this on oopsy triggers that have all fields that are otherwise valid for raidboss. Also, it only happened twice.

quisquous: Why is this concerning?

Also, I could see some argument in splitting the rules (or having an option for the field ordering that gets moved to the eslint config?), since oopsy and raidboss do have different fields.  Otherwise, is the plan to put `mistake` and `deathReason` in that list?

ui/raidboss/data/05-shb/trial/hades-ex.js
```diff
@@ -779,13 +779,13 @@ export default {
         },
       },
     },
-    { // After tanks take tower damage
```
quisquous, line 782: Could you maybe keep this comment and move it down by the delaySeconds?

panicstevenson: Hmmm. This seems like a bug then if the comments got dropped. Good catch!

ui/oopsyraidsy/data/03-hw/raid/a12n.js
panicstevenson: I mean, theoretically the list could contain the superset of both, but that (I feel) would become pretty difficult to parse. I can add an optional override so they can be done per raidboss/oopsy, but then that would be cloning the overrides block in eslintrc (not really an issue, but it’s a thing)

I don’t know if there’s an easy way to tell if we’re in an oopsy or raidboss trigger (other than fields or filepath) that we could key off of

quisquous: Yeah, that's why I was suggesting doing it in the eslint config.  You could pass the `orderList` in as an eslint option to the rule, and then apply a rule + different option to each directory.

I think we could have a "applies to both" and then a "applies to raidboss" and a "applies to oopsy" section so there's not a duplication of the entire overrides block.

.eslintrc.cjs
```diff
@@ -268,6 +268,12 @@ const overrides = [
       'rulesdir/cactbot-trigger-property-order': 'error',
```
quisquous, line 268: It is a little weird that the raidboss one applies to both. Can this be pulled out into a separate raidboss-only section, module: raidboss, as a peer to oopsyraidsy directly below.

ui/raidboss/data/05-shb/trial/hades-ex.js
quisquous: Did anything change to fix this happening in the future?

panicstevenson: No, and I'm not entirely sure on the best way to do it either; I looked at other plugins for their approach, and it seems like they generally bail if they would mess up comments.

I feel like it might be okay for the future since these should be alerting you as you're typing, not an afterthought? I guess the argument would be for people not using VSCode that get blocked by Husky on commit and accidentally remove all of their comments.

quisquous: Yeah, I think the worry is that somebody runs an `npm run lintfix` and it's destructive.  What about comments on their own line, comments on their own line after the last property, or comments on their own line before the first property?

For this case, is it possible to include the textrange between the start of node and the first property?

PR #3109, opened 2021-06-16
lint: Ensure Matches is Always Typed

panicstevenson:

PR #3120, opened 2021-06-16
More auto-type-inferrence for triggers

### LINKED ISSUE #3119: Have TriggerSet Matches Determine the Correct Parameters
## Summary

Currently, trigger functions use `MatchesAny` for their matches parameter, which means that you can run into places like [this](https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/02-arr/raid/t8.ts#L57):

```ts
infoText: (data, matches: MatchesAbility, output) => {
  if (matches.target && matches.target in data.landmines)
    return;
  const num = Object.keys(data.landmines).length + 1;
  return output.landmine!({ num: `${num}` });
},
```

Where, although the code is specifying  that it is a `MatchesAbility`, it states that `matches.target` could potentially be undefined, leading to an additional truthy check that wasn't present in the JavaScript version.

After playing around in the trigger type definition file, I removed `MatchesAny` from `type Matches<Param>`:
```ts
export type Matches<Params> = { [s in Params]: string };
```
And updated the `TriggerOutput`, `PartialTriggerOutput`, and `TriggerFunc` types to be:
```ts
// The output of any non-response raidboss trigger function.
export type TriggerOutput<Data, T> =
    T extends MatchesAbility ? undefined | null | LocaleText | string | number | boolean |
    ((d: Data, m: MatchesAbility, o: Output) => TriggerOutput<Data, MatchesAbility>) :
    undefined | null | LocaleText | string | number | boolean |
    ((d: Data, m: Matches, o: Output) => TriggerOutput<Data, Matches>);

// Used if the function doesn't need to return an en key
export type PartialTriggerOutput<Data, T> =
    T extends MatchesAbility ? undefined | null | Partial<LocaleText> | string | number | boolean |
    ((d: Data, m: MatchesAbility, o: Output) => PartialTriggerOutput<Data, MatchesAbility>) :
    undefined | null | Partial<LocaleText> | string | number | boolean |
    ((d: Data, m: Matches, o: Output) => PartialTriggerOutput<Data, Matches>);

// The type of a non-response trigger field.
export type TriggerFunc<Data, T, Return> =
    T extends MatchesAbility ? (data: Data, matches: MatchesAbility, output: Output) => Return :
    (data: Data, matches: Matches, output: Output) => Return;
```

Which, with the conditional typing, I can now remove the `matches.target` from the trigger's infoText function conditional, but I don't think copying and pasting the type conditional for each Matches type is the best way about this. I'm curious if anyone has any other thoughts or ideas to make this logic cleaner, either through type conditionals or something else entirely.

valarnin:
This wizard's library sure has a lot of nice books.

Closes #3119 

You still need to specify the type for callbacks, but this fixes type inference:

image
image

I couldn't see a better way to handle the return value for `Regexes.parse` and `NetRegexes.parse`?

### DIFF
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -2,14 +2,14 @@
 // are also functions.
 
 import { RaidbossData as Data } from '../types/data';
-import { TargetedMatches } from '../types/trigger';
+import { MatchesAny, TargetedMatches } from '../types/trigger';
 
 export default {
   targetIsYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: TargetedMatches) => data.me === matches?.target;
+    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me === matches?.target;
   },
   targetIsNotYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: TargetedMatches) => data.me !== matches?.target;
+    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me !== matches?.target;
   },
   caresAboutAOE(): (data: Data) => boolean {
     return (data: Data) => data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -6,10 +6,10 @@ import { MatchesAny, TargetedMatches } from '../types/trigger';
 
 export default {
   targetIsYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me === matches?.target;
+    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me === matches.target;
   },
   targetIsNotYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me !== matches?.target;
+    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me !== matches.target;
   },
   caresAboutAOE(): (data: Data) => boolean {
     return (data: Data) => data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -60,6 +60,32 @@ export type ChangeZoneParams = typeof changeZoneParams[number];
 export type Network6dParams = typeof network6dParams[number];
 export type NameToggleParams = typeof nameToggleParams[number];
 
+export interface ParamsAll {
+  'StartsUsingParams': StartsUsingParams;
+  'AbilityParams': AbilityParams;
+  'AbilityFullParams': AbilityFullParams;
+  'HeadMarkerParams': HeadMarkerParams;
+  'AddedCombatantParams': AddedCombatantParams;
+  'AddedCombatantFullParams': AddedCombatantFullParams;
+  'RemovingCombatantParams': RemovingCombatantParams;
+  'GainsEffectParams': GainsEffectParams;
+  'StatusEffectExplicitParams': StatusEffectExplicitParams;
+  'LosesEffectParams': LosesEffectParams;
+  'TetherParams': TetherParams;
+  'WasDefeatedParams': WasDefeatedParams;
+  'EchoParams': EchoParams;
+  'DialogParams': DialogParams;
+  'MessageParams': MessageParams;
+  'GameLogParams': GameLogParams;
+  'GameNameLogParams': GameNameLogParams;
+  'StatChangeParams': StatChangeParams;
+  'ChangeZoneParams': ChangeZoneParams;
+  'Network6dParams': Network6dParams;
+  'NameToggleParams': NameToggleParams;
+}
+
+export type ParamsAllTypes = keyof ParamsAll;
+
 // If NetRegexes.setFlagTranslationsNeeded is set to true, then any
 // regex created that requires a translation will begin with this string
 // and match the magicStringRegex.  This is maybe a bit goofy, but is
@@ -75,11 +101,11 @@ const keysThatRequireTranslation = [
   'line',
 ];
 
-const parseHelper = (
-    params: { timestamp?: string; capture?: boolean } | undefined,
-    funcName: string,
-    fields: { [s: string]: Fields },
-): RegExp => {
+const parseHelper = <T extends ParamsAllTypes>(
+  params: Params<T> | undefined,
+  funcName: string,
+  fields: { [s: string]: Fields },
+): NetRegex<ParamsAll[T]> => {
   params = params ?? {};
   const validFields: string[] = [];
   for (const value of Object.values(fields)) {
@@ -151,7 +177,7 @@ const parseHelper = (
     if (key >= (maxKey ?? 0 as number))
       break;
   }
-  return Regexes.parse(str);
+  return NetRegexes.parse(str);
 };
 
 export default class NetRegexes {
@@ -521,5 +547,36 @@ export default class NetRegexes {
       3: { field: 'name' },
       6: { field: 'toggle' },
     });
+  }
+
+  static parse<T extends ParamsAllTypes>(regexpString: RegExp | string): NetRegex<ParamsAll[T]> {
+    const kCactbotCategories = {
+      Timestamp: '^.{14}',
+      NetTimestamp: '.{33}',
+      NetField: '(?:[^|]*\\|)',
+      LogType: '[0-9A-Fa-f]{2}',
+      AbilityCode: '[0-9A-Fa-f]{1,8}',
+      ObjectId: '[0-9A-F]{8}',
+      // Matches any character name (including empty strings which the FFXIV
+      // ACT plugin can generate when unknown).
+      Name: '(?:[^\\s:|]+(?: [^\\s:|]+)?|)',
+      // Floats can have comma as separator in FFXIV plugin output: https://github.com/ravahn/FFXIV_ACT_Plugin/issues/137
+      Float: '-?[0-9]+(?:[.,][0-9]+)?(?:E-?[0-9]+)?',
+    };
+
+    // All regexes in cactbot are case insensitive.
+    // This avoids headaches as things like `Vice and Vanity` turns into
+    // `Vice And Vanity`, especially for French and German.  It appears to
+    // have a ~20% regex parsing overhead, but at least they work.
+    let modifiers = 'i';
+    if (regexpString instanceof RegExp) {
+      modifiers += (regexpString.global ? 'g' : '') +
+                    (regexpString.multiline ? 'm' : '');
+      regexpString = regexpString.source;
+    }
+    regexpString = regexpString.replace(/\\y\{(.*?)\}/g, (match, group) => {
+      return kCactbotCategories[group as keyof typeof kCactbotCategories] || match;
+    });
+    return new RegExp(regexpString, modifiers) as NetRegex<ParamsAll[T]>;
   }
 }
--- a/resources/regexes.ts
+++ b/resources/regexes.ts
@@ -152,6 +152,31 @@ export type GameNameLogParams = typeof gameNameLogParams[number];
 export type ChangeZoneParams = typeof changeZoneParams[number];
 export type Network6dParams = typeof network6dParams[number];
 
+export interface ParamsAll {
+  'StartsUsingParams': StartsUsingParams;
+  'AbilityParams': AbilityParams;
+  'AbilityFullParams': AbilityFullParams;
+  'HeadMarkerParams': HeadMarkerParams;
+  'AddedCombatantParams': AddedCombatantParams;
+  'AddedCombatantFullParams': AddedCombatantFullParams;
+  'RemovingCombatantParams': RemovingCombatantParams;
+  'GainsEffectParams': GainsEffectParams;
+  'StatusEffectExplicitParams': StatusEffectExplicitParams;
+  'LosesEffectParams': LosesEffectParams;
+  'TetherParams': TetherParams;
+  'WasDefeatedParams': WasDefeatedParams;
+  'EchoParams': EchoParams;
+  'DialogParams': DialogParams;
+  'MessageParams': MessageParams;
+  'GameLogParams': GameLogParams;
+  'GameNameLogParams': GameNameLogParams;
+  'StatChangeParams': StatChangeParams;
+  'ChangeZoneParams': ChangeZoneParams;
+  'Network6dParams': Network6dParams;
+}
+
+export type ParamsAllTypes = keyof ParamsAll;
+
 export default class Regexes {
   /**
    * fields: source, id, ability, target, capture
@@ -695,7 +720,7 @@ export default class Regexes {
     return anyOfArray(array);
   }
 
-  static parse(regexpString: RegExp | string): RegExp {
+  static parse<T extends ParamsAllTypes>(regexpString: RegExp | string): Regex<ParamsAll[T]> {
     const kCactbotCategories = {
       Timestamp: '^.{14}',
       NetTimestamp: '.{33}',
@@ -723,7 +748,7 @@ export default class Regexes {
     regexpString = regexpString.replace(/\\y\{(.*?)\}/g, (match, group) => {
       return kCactbotCategories[group as keyof typeof kCactbotCategories] || match;
     });
-    return new RegExp(regexpString, modifiers);
+    return new RegExp(regexpString, modifiers) as Regex<ParamsAll[T]>;
   }
 
   // Like Regex.Regexes.parse, but force global flag.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -1,19 +1,30 @@
 import { Lang, NonEnLang } from '../resources/languages';
 import { TimelineReplacement, TimelineStyle } from '../ui/raidboss/timeline';
 import { RaidbossData } from './data';
+import { MatchesAllTypes, MatchesAbilityFull, MatchesAbility, MatchesGainsEffect, MatchesLosesEffect, MatchesStartsUsing, MatchesTether, MatchesWasDefeated } from '../resources/matches';
+import { Regex } from '../resources/regexes';
+import { NetRegex } from '../resources/netregexes';
 
-export interface BaseRegExp<T> extends RegExp {
+export interface BaseRegExpExecArray<T> extends RegExpExecArray {
   groups?: {
-    [s in T]?: string;
+    [s in T]: string;
   };
 }
 
-export type Matches<Params> = { [s in Params]: string } | MatchesAny;
+export interface BaseRegExp<T> extends RegExp {
+  exec: (string: string) => BaseRegExpExecArray<T> | null;
+}
+
+export type Matches<Params> = { [s in Params]: string };
 
 // TargetedMatches can be used for generic functions in responses or conditions
 // that use matches from any number of Regex or NetRegex functions.
-export type TargetedParams = 'sourceId' | 'source' | 'targetId' | 'target';
-export type TargetedMatches = Matches<TargetedParams>;
+export type TargetedMatches =
+  MatchesStartsUsing | MatchesAbility | MatchesAbilityFull | MatchesGainsEffect |
+  MatchesLosesEffect | MatchesTether | MatchesWasDefeated;
+
+export type MatchesAny = (MatchesAllTypes extends infer T ? T extends MatchesAllTypes ?
+  Matches<T> : never : never) | { [k: string]: string };
 
 export type FullLocaleText = Record<Lang, string>;
 
@@ -38,18 +49,18 @@ export type Output = {
 };
 
 // The output of any non-response raidboss trigger function.
-export type TriggerOutput<Data, Matches> =
+export type TriggerOutput<Data, Matches extends MatchesAny> =
     undefined | null | LocaleText | string | number | boolean |
-    ((d: Data, m: Matches, o: Output) => TriggerOutput<Data, Matches>);
+    ((d: Data, m: MatchesAll<Matches>, o: Output) => TriggerOutput<Data, Matches>);
 
 // Used if the function doesn't need to return an en key
-export type PartialTriggerOutput<Data, Matches> =
+export type PartialTriggerOutput<Data, Matches extends MatchesAny> =
     undefined | null | Partial<LocaleText> | string | number | boolean |
-    ((d: Data, m: Matches, o: Output) => PartialTriggerOutput<Data, Matches>);
+    ((d: Data, m: MatchesAll<Matches>, o: Output) => PartialTriggerOutput<Data, Matches>);
 
 // The type of a non-response trigger field.
 export type TriggerFunc<Data, Matches, Return> =
-    (data: Data, matches: Matches, output: Output) => Return;
+    (data: Data, matches: MatchesAll<Matches>, output: Output) => Return;
 
 // The output from a response function (different from other TriggerOutput functions).
 export type ResponseOutput<Data, Matches> = {
@@ -59,10 +70,11 @@ export type ResponseOutput<Data, Matches> = {
   tts?: TriggerFunc<Data, Matches, PartialTriggerOutput<Data, Matches>>;
 };
 // The type of a response trigger field.
-export type ResponseFunc<Data, Matches> =
-    (data: Data, matches: Matches, output: Output) => ResponseOutput<Data, Matches>;
+export type ResponseFunc<Data, Matches extends MatchesAny> =
+    (data: Data, matches: MatchesAll<Matches>, output: Output) => ResponseOutput<Data, Matches>;
 
-export type ResponseField<Data> = ResponseFunc<Data, MatchesAny> | ResponseOutput<Data, MatchesAny>;
+export type ResponseField<Data, Matches> =
+    ResponseFunc<Data, Matches> | ResponseOutput<Data, Matches>;
 
 export type TriggerAutoConfig = {
   Output?: Output;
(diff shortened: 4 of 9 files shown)


panicstevenson:
> You still need to specify the type for callbacks

There should be a lint rule that enforces this either way, right? Do you have the ESLint extension installed? Curious that `matches` isn't underlined in that screenshot.

valarnin:
> There should be a lint rule that enforces this either way, right? Do you have the ESLint extension installed? Curious that `matches` isn't underlined in that screenshot.

The lint rule seems to enforce the correct type only if the type is specified.

resources/conditions.ts
```diff
@@ -2,14 +2,14 @@
 // are also functions.
 
 import { RaidbossData as Data } from '../types/data';
-import { TargetedMatches } from '../types/trigger';
+import { MatchesAny, TargetedMatches } from '../types/trigger';
 
 export default {
   targetIsYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: TargetedMatches) => data.me === matches?.target;
+    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me === matches?.target;
```
panicstevenson, line 9:
```suggestion
    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me === matches.target;
```
And below

valarnin:
Not goona lie, this one really fried my brain. I just about gave up when tsc stopped working entirely. For future reference if that ever happens for someone else, the cause for me was `TriggerFunc` being defined like this:

```typescript
export type TriggerFunc<Data, Matches, Return> =
    (data: Data, matches: Matches, output: Output) => Return;
```

That just completely broke tsc for some reason I'm still not sure about.

panicstevenson:
> The lint rule seems to enforce the correct type only if the type is specified.

That's not good? It should be there since https://github.com/quisquous/cactbot/pull/3109; did you have VSCode opened before you pulled latest? I've found that the local rules only register once every time VSCode's ESLint server is restarted, either via closing and opening VSCode or doing Ctrl+Shift+P => Restart ESLint Server

valarnin:
It wasn't a problem with the code previously defined, it was a problem with that type being passed a string for `Matches` generic type from the changes upstream in `RegexTrigger` and `NetRegexTrigger` instead of an actual Matches type.

quisquous:
🤯 🤯 🤯

resources/netregexes.ts
```diff
@@ -522,4 +548,35 @@ export default class NetRegexes {
       6: { field: 'toggle' },
     });
   }
+
+  static parse<T extends ParamsAllTypes>(regexpString: RegExp | string): NetRegex<ParamsAll[T]> {
+    const kCactbotCategories = {
```
quisquous, line 553: Is there a way to not duplicate this `(regexpString: RegExp | string) => RegExp` common code?

resources/regexes.ts
```diff
@@ -695,7 +720,7 @@ export default class Regexes {
     return anyOfArray(array);
   }
 
-  static parse(regexpString: RegExp | string): RegExp {
+  static parse<T extends ParamsAllTypes>(regexpString: RegExp | string): Regex<ParamsAll[T]> {
```
quisquous, line 723: So, `Regexes.parse` has been the base function that many things have called for a long time.  See [this code](https://github.com/quisquous/cactbot/blob/0109118d9bbfbfd9abd5a4cca29a2abda28b2136/ui/oopsyraidsy/oopsyraidsy.js#L1217) for example.

One option could be to change these to parse functions to `private parseInternal` or something and have a separate `Regxes.parse` that returns a `RegExp` and then fix up the other callsites later as needed.

(This function returning a typed regex is a great idea!)

types/trigger.d.ts
```diff
@@ -74,56 +86,58 @@ export type TriggerAutoConfig = {
...
 }
 
-export type MatchesAny = { [s in T]?: string };
-
 // Note: functions like run or preRun need to be defined as void-only as (confusingly)
 // it is not possible to assign `(d: Data) => boolean` to a void | undefined, only to void.
-export type TriggerField<Data, Return> =
-  [Return] extends [void] ? TriggerFunc<Data, MatchesAny, void> :
-  TriggerFunc<Data, MatchesAny, Return | undefined> | Return | undefined;
+export type TriggerField<Data, Return, Matches> =
```
quisquous, line 91: I am sorry for the most obnoxious nit here, but can these be ordered left to right like they'd be in a trigger? e.g. ?<data, matches, (output), return>?

resources/conditions.ts
```diff
@@ -2,14 +2,14 @@
 // are also functions.
 
 import { RaidbossData as Data } from '../types/data';
-import { TargetedMatches } from '../types/trigger';
+import { MatchesAny, TargetedMatches } from '../types/trigger';
 
 export default {
   targetIsYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: TargetedMatches) => data.me === matches?.target;
+    return (data: Data, matches: MatchesAny) => 'target' in matches && data.me === matches.target;
```
quisquous, line 9: Is `MatchesAny` just to be safe if the wrong type gets used here?

For what it's worth, I'm not sure either the above `TargetedMatches` adds any type safety here.  I wonder a little if this should just be `MatchesAny`.  It's not clear to me if it is possible to get popuptext to care about matches types here.

types/trigger.d.ts
```diff
@@ -74,56 +86,58 @@ export type TriggerAutoConfig = {
...
+  netRegex: NetRegex<Matches<T>>;
+  netRegexDe?: NetRegex<Matches<T>>;
+  netRegexFr?: NetRegex<Matches<T>>;
+  netRegexJa?: NetRegex<Matches<T>>;
+  netRegexCn?: NetRegex<Matches<T>>;
+  netRegexKo?: NetRegex<Matches<T>>;
+}) : never : never;
+
+export type RegexTrigger<Data> =
+  MatchesAllTypes extends infer T ? [T] extends [MatchesAllTypes] ?
```
quisquous, line 130: aha the old infer/extends at the heart of it all

resources/regexes.ts
valarnin: I assume this is in relation to the other comment about argument types (`regexpString: RegExp | string`)?

I could make a pass through the code and clean them all up but it really seems like it doesn't fit with this PR. Maybe open an issue so it doesn't get lost?

quisquous: Ah no, this is separate from the above comment.  (And the above comment about `(regexpString: RegExp | string) => RegExp` was not about argument types but about the body of the function being duplicated between `NetRegexes.parse` and `Regexes.parse`.)

This comment is about the fact that you are taking a common function called by both Regexes and NetRegexes (i.e. `Regexes.parse`) and making it typed to only return a `Regex<>` but without changing any of the callers that are now using it incorrectly.

valarnin:
I think I might take a second pass at this and rework how the type defs for the base Matches and Params types are set up to make this work 100% across the board. Basically, right now I've poked around further at this and I've gotten it to the point where it can infer the typing down part-way but it won't go all the way due to overlap of the field defs for a given log line:

image

If I rework how the base types are so that, for instance, `NetRegexes.startsUsing` returns a `NetRegex` that resolves to a type with `type: '20'` as a property, that won't be assignable directly to another `NetRegex` with `type: '21'` as a property, which would allow type inference to detect all the way through to things like `condition` params.

valarnin:
Alright, I've banged my head against this for about 8 hours now today, including a back and forth with someone significantly better than me with typescript, and this doesn't really look feasible. The reason the PR as it exists "works" isn't because of type inference but because of removing the optional from the base matches type and instead making MatchesAny more thorough, which could still cause problems at runtime because if you're not capturing matches, then matches would end up an empty object, and there's no type checking on that.

If we want full type inference and hinting/checking for this, it would require actually properly typing the triggers as they're created, which would mean a different or second syntax for creating triggers. The old syntax would still be fine in all cases, but any triggers in typescript would need to use a different syntax if they want proper type hinting.

I'm open to suggestions and/or just saying this is impossible and closing the PR, but I'd like to hear thoughts on something like this for syntax:

```typescript
  zoneId: ZoneId.TheBindingCoilOfBahamutTurn1,
  initData: () => {
    return {
      started: false,
    };
  },
  triggers: [
    // new format
    new NetRegexTrigger<Data, 'StartsUsingParams'>(
        'T1 High Voltage',
        'StartsUsingParams',
        {
          id: '5A7',
        },
        {
          en: { source: 'Ads' },
          de: { source: 'Abwehrsystem' },
          fr: { source: 'Sphère De Contrôle' },
          ja: { source: '制御システム' },
          cn: { source: '自卫系统' },
          ko: { source: '제어 시스템' },
        },
        {
          condition: (data) => data.CanSilence(),
          response: Responses.interrupt(),
        },
    ),
    // old format
    {
      id: 'T1 High Voltage',
      netRegex: NetRegexes.startsUsing({ source: 'Ads', id: '5A7' }),
      netRegexDe: NetRegexes.startsUsing({ source: 'Abwehrsystem', id: '5A7' }),
      netRegexFr: NetRegexes.startsUsing({ source: 'Sphère De Contrôle', id: '5A7' }),
      netRegexJa: NetRegexes.startsUsing({ source: '制御システム', id: '5A7' }),
      netRegexCn: NetRegexes.startsUsing({ source: '自卫系统', id: '5A7' }),
      netRegexKo: NetRegexes.startsUsing({ source: '제어 시스템', id: '5A7' }),
      condition: (data) => data.CanSilence(),
      response: Responses.interrupt(),
    },
  ],
};
```

- Pros:
  - Fully typed, everything auto-detects typing and enforces types
  - Less repeated fields so less chance of copy/paste issues, e.g. `id: '5A7'` is spread to all languages automatically
  - Might be easier to localize properly, things like `name` could be pulled to a const at the top instead of being repeated per-trigger
- Cons:
  - A bit less nice to look at maybe?
  - Less horizontal space, more vertical space used
  - Have to repeat 'StartsUsingParams' twice, once for the actual typing and once for figuring out which `NetRegexes` function to call, I couldn't figure out a way past that. It does enforce that they're equal, though.
  - Have to repeat `Data` per trigger, again, I couldn't find a way past that.

valarnin:
This is what a fully converted `t1.ts` file would look like, for reference:

```typescript
import { RaidbossData } from '../../../../../types/data';
import { TriggerSet } from '../../../../../types/trigger';

import Conditions from '../../../../../resources/conditions';
import { Responses } from '../../../../../resources/responses';

  started: boolean;
}

const adsNames = {
  en: { source: 'Ads' },
  de: { source: 'Abwehrsystem' },
  fr: { source: 'Sphère De Contrôle' },
  ja: { source: '制御システム' },
  cn: { source: '自卫系统' },
  ko: { source: '제어 시스템' },
};

const caduceusNamesRegex = {
  en: { name: 'Caduceus.*?' },
  de: { name: 'Caduceus.*?' },
  fr: { name: 'Caducée.*?' },
  ja: { name: 'カドゥケウス.*?' },
  cn: { name: '神杖巨蛇.*?' },
  ko: { name: '카두케우스.*?' },
};

const caduceusNames = {
  en: { source: 'Caduceus' },
  de: { source: 'Caduceus' },
  fr: { source: 'Caducée' },
  ja: { source: 'カドゥケウス' },
  cn: { source: '神杖巨蛇' },
  ko: { source: '카두케우스' },
};

  zoneId: ZoneId.TheBindingCoilOfBahamutTurn1,
  initData: () => {
    return {
      started: false,
    };
  },
  triggers: [
    new NetRegexTrigger<Data, 'StartsUsingParams'>(
        'T1 High Voltage',
        'StartsUsingParams',
        { id: '5A7' },
        adsNames,
        {
          condition: (data) => data.CanSilence(),
          response: Responses.interrupt(),
        },
    ),
    // Indiscriminate Hood Swing
    new NetRegexTrigger<Data, 'AbilityParams'>(
        'T1 Initiated',
        'AbilityParams',
        { id: '4B8.*?', capture: false },
        adsNames,
        {
          condition: (data) => data.CanSilence(),
          response: Responses.interrupt(),
        },
    ),
    new NetRegexTrigger<Data, 'AbilityParams'>(
        'T1 Regorge',
        'AbilityParams',
        { id: '4BA' },
        adsNames,
        {
          alertText: (_data, _matches, output) => output.text!(),
          outputStrings: {
            text: {
              en: 'Spit on YOU',
              de: 'Spucke auf DIR',
              fr: 'Crachat sur VOUS',
              ja: '自分にリゴージ',
              cn: '吐痰点名',
            },
          },
        },
    ),
    new NetRegexTrigger<Data, 'AddedCombatantParams'>(
        'T1 Split',
        'AddedCombatantParams',
        {},
        caduceusNamesRegex,
        {
          condition: (data) => data.started,
          suppressSeconds: 5,
          alertText: (_data, _matches, output) => output.text!(),
          outputStrings: {
            text: {
              en: 'Split',
              de: 'Teilen',
              fr: 'Division',
              ja: '分裂',
              cn: '分裂',
            },
          },
        },
    ),
    new NetRegexTrigger<Data, 'AbilityParams'>(
        'T1 Hood Swing',
        'AbilityParams',
        { id: '4B8' },
        caduceusNames,
        {
          delaySeconds: 8,
          suppressSeconds: 5,
          infoText: (_data, _matches, output) => output.text!(),
          outputStrings: {
            text: {
              en: 'Hood Swing in 10',
              de: 'Kapuzenschwung in 10',
              fr: 'Coup de capot dans 10s',
              ja: '十秒以内タンクバスター',
              cn: '10秒内死刑',
            },
          },
        },
    ),
    new NetRegexTrigger<Data, 'MessageParams'>(
        'T1 Slime Timer First',
        'MessageParams',
        { capture: false },
        {
          en: { line: 'The Allagan megastructure will be sealed off.*?' },
          de: { line: 'Allagische Superstruktur will be sealed off.*?' },
          fr: { line: 'Mégastructure allagoise will be sealed off.*?' },
          ja: { line: 'アラグの遺構 will be sealed off.*?' },
          cn: { line: '亚拉戈遗构 will be sealed off.*?' },
          ko: { line: '알라그 유적 will be sealed off.*?' },
        },
        {
          delaySeconds: 35,
          suppressSeconds: 5,
          infoText: (_data, _matches, output) => output.text!(),
          outputStrings: {
            text: {
              en: 'Slime Soon',
              de: 'Schleim bald',
              fr: 'Gluant bientôt',
              ja: 'まもなくスライム',
              cn: '软泥即将出现',
            },
          },
        },
    ),
    new NetRegexTrigger<Data, 'AddedCombatantParams'>(
        'T1 Split',
        'AddedCombatantParams',
        { capture: false },
        {
          en: { name: 'Dark Matter Slime.*?' },
          de: { name: 'Dunkelmaterien-Schleim.*?' },
          fr: { name: 'Gluant De Matière Sombre.*?' },
          ja: { name: 'ダークマター・スライム.*?' },
          cn: { name: '暗物质粘液怪.*?' },
          ko: { name: '암흑물질 슬라임.*?' },
        },
        {
          delaySeconds: 35,
          suppressSeconds: 5,
          infoText: (_data, _matches, output) => output.text!(),
          outputStrings: {
            text: {
              en: 'Slime Soon',
              de: 'Schleim bald',
              fr: 'Gluant bientôt',
              ja: 'まもなくスライム',
              cn: '软泥即将出现',
            },
          },
        },
    ),
  ],
  timelineReplace: [
    {
      'locale': 'de',
      'replaceSync': {
        'Ads': 'Abwehrsystem',
        'Caduceus': 'Caduceus',
        'Dark Matter Slime': 'Dunkelmaterien-Schleim',
        'The Allagan megastructure': 'Allagische Superstruktur',
      },
    },
    {
      'locale': 'fr',
      'replaceSync': {
        'Ads': 'Sphère De Contrôle',
        'Caduceus': 'Caducée',
        'Dark Matter Slime': 'Gluant De Matière Sombre',
        'The Allagan megastructure': 'Mégastructure allagoise',
      },
    },
    {
      'locale': 'ja',
      'replaceSync': {
        'Ads': '制御システム',
        'Caduceus': '蛇杖撃',
        'Dark Matter Slime': 'ダークマター・スライム',
        'The Allagan megastructure': 'アラグの遺構',
      },
    },
    {
      'locale': 'cn',
      'replaceSync': {
        'Ads': '自卫系统',
        'Caduceus': '蛇杖击',
        'Dark Matter Slime': '暗物质粘液怪',
        'The Allagan megastructure': '亚拉戈遗构',
      },
    },
    {
      'locale': 'ko',
      'replaceSync': {
        'Ads': '제어 시스템',
        'Caduceus': '뱀 지팡이',
        'Dark Matter Slime': '암흑물질 슬라임',
        'The Allagan megastructure': '알라그 유적',
      },
    },
  ],
};

export default triggerSet;

```

quisquous:
> Alright, I've banged my head against this for about 8 hours now today, including a back and forth with someone significantly better than me with typescript, and this doesn't really look feasible. The reason the PR as it exists "works" isn't because of type inference but because of removing the optional from the base matches type and instead making MatchesAny more thorough, which could still cause problems at runtime because if you're not capturing matches, then matches would end up an empty object, and there's no type checking on that.

Sorry, I'll take a closer look tomorrow but you having spent 8 hours of your time does not sound promising for me to sort anything out better.  Thanks for so much for all of your time on this.  The previous patch felt so close though; can you explain a little bit more what the issue(s) you're trying to solve?

I think not capturing returning nothing is fine if that's a blocker.  It's not the best, but we have enough testing there that you only use `matches` when capturing that it is not a worry for me.

valarnin:
> Sorry, I'll take a closer look tomorrow but you having spent 8 hours of your time does not sound promising for me to sort anything out better. Thanks for so much for all of your time on this. The previous patch felt so close though; can you explain a little bit more what the issue(s) you're trying to solve?

The original goal of the PR was to:
1. Have it automatically infer types based on the `NetRegexes/Regexes` function called to build a regex.
  That never really worked, it was just typing weirdness that made it look like it worked with the two triggers I was testing with. Every single bit of `X extends infer Y ? Y extends X` code in this PR does absolutely nothing
2. Remove the check for if a `matches` parameter is set
  This works, but can be accomplished with far less code than what's in this PR

> I think not capturing returning nothing is fine if that's a blocker. It's not the best, but we have enough testing there that you only use `matches` when capturing that it is not a worry for me.

That's not really a blocker so much as something that I thought the original PR would fix unintentionally but doesn't.

> I was hoping to eventually solve these by using the `timelineReplace` block and just dropping the `netRegexDe` etc lines and translating them automatically. See: #1306.

Wouldn't it be better to do either of these instead?
- Keying triggers with enemy names off of bNpcId
- Pulling translation info from XIVAPI and looking up from that

It seems a bit odd to me that a lot of triggers keep on keying off of name instead of ID in general, and something like one of the following would be better all around:

```typescript
    {
      id: 'T1 High Voltage',
      // `npcName` in this case is { [englishNpcName: string]: LocaleText }
      netRegex: NetRegexes.startsUsing({ source: npcName['ADS'][data.options.ParserLanguage], id: '5A7' }),
      condition: (data) => data.CanSilence(),
      response: Responses.interrupt(),
    },
    {
      id: 'T1 High Voltage',
      // `npcName` in this case is { [npcId: number]: LocaleText }
      netRegex: NetRegexes.startsUsing({ source: npcName[1459][data.options.ParserLanguage], id: '5A7' }),
      condition: (data) => data.CanSilence(),
      response: Responses.interrupt(),
    },
    {
      id: 'T1 High Voltage',
      // If the fight only has one actor using '5A7', why key off of name at all?
      netRegex: NetRegexes.startsUsing({ id: '5A7' }),
      condition: (data) => data.CanSilence(),
      response: Responses.interrupt(),
    },
```

`timelineReplace` could look up the names in the same way, or even do an auto map after the fact, looping over all the relevant IDs or whatever and adding them to `timelineReplace` without having to copy/paste as much, something like:

```typescript
const npcIds = ['ADS', 'Caduceus', 'dark matter slime'];

for (const id of npcIds) {
  for (const lang of languages) {
    let langMap = triggerSet.timelineReplace?.find((m) => m.locale === lang);
    if (!langMap) {
      langMap = {
        'locale': lang,
      };
      triggerSet.timelineReplace?.push(langMap);
    }
    if (!langMap.replaceSync)
      langMap.replaceSync = {};
    if (!langMap.replaceText)
      langMap.replaceText = {};
    if (!(id in langMap.replaceSync))
      langMap.replaceSync[id] = npcName[id][lang];
    if (!(id in langMap.replaceText))
      langMap.replaceText[id] = npcName[id][lang];
  }
}
```

That could even be a Util helper, pass it the trigger set and the array of IDs and it'll map everything over, and automatically pick up new translations whenever the xivapi export or whatever is updated.

quisquous:
> Wouldn't it be better to do either of these instead?
> 
> * Keying triggers with enemy names off of bNpcId
> * Pulling translation info from XIVAPI and looking up from that

Would you mind copying that discussion to #1306 so it's not sequestered off on this PR (which will probably close first)?

quisquous:
### Priorities

For context on my priorities, one thing I was hoping to be able to solve prior to converting more triggers was being able to make properties on matches (e.g. `matches.target`) a non-optional property.  This to me is a "nice to have" and isn't blocking to converting triggers.  Full inference of matches feels even more like a "nice to have" if we could do it, but also not blocking.

One reason I feel like full inference is less of an issue than before we started TypeScript is that @panicstevenson added some eslint rules to make sure the `MatchesFoo` type existed and corresponded to the `netRegex`.  The fact that it's auto-fixable makes this feel like less overhead for a trigger author and reviewer.  Non-optional `matches.target` properties will make it easier to convert triggers (and probably introduce fewer bugs because there's fewer changes).

That said, full inference would be great.  I think depending on tsc instead of eslint seems like a better option when we can do it.

### This patch as-is (or at least the matches part)

By changing the type of matches, this patch solves the non-optional matches property issue.  I do think it is not a perfect solution (not knocking your code, just thinking about all the implications) For one, if you have `netRegex.startsUsing` you can use `Matches.Ability` and tsc won't complain.  Similarly, `matches` is an `any` parameter unless explicitly defined.  However, I think the eslint rules cover this in the same way we would have wanted tsc to:

image

Or, at least in trigger sets, it does.  I think any loose calls to `netRegex.startsUsing(...)` might be able to incorrectly use the wrong matches type.  If that's the case, then we could consider some follow-up where we ...override NetRegex.match/exec/etc to return the right kind of matches (if capture is true?).  Or something along those lines?

I would sum up this approach as being "(automated) required explicit matches parameters in order to get non-optional properties on matches".  Personally, I think this is a net win.  Triggers are really the bulk of cactbot code, so anything we can do to reduce burden there pays off a lot.  Subjectively, I think other folks (like Trim21 and MaikoTan) seem even more annoyed about having to null check things than I do.

### Proposal: new NetRegexTrigger<Data, 'StartsUsingParams'>

Yeah, this is along the same lines as what ngld suggested in [that discord conversation with you and me](https://discord.com/channels/551474815727304704/594899820976668673/821089539640262677).

Please feel free to disagree (and I could just be change averse here) but this still feels to me like a pretty heavy solution to get free type inference on `matches`.  It gets the job done, but foof it feels like a lot.

### Alternative: explicit type parameter

I would propose another option, which is to make `Trigger` explicitly a discriminated union based on some new `type` field.

```typescript
    {
      id: 'T8 Landmine Explosion',
      type: 'ability',
      netRegex: NetRegexes.ability({ id: '7D1', source: 'Allagan Mine' }),
      infoText: (data, matches, output) => {
        if (matches.target && matches.target in data.landmines)
          return;
        const num = Object.keys(data.landmines).length + 1;
        return output.landmine!({ num: `${num}` });
      },
```

I didn't prototype this out all the way, but [this playground link](https://www.typescriptlang.org/play?#code/C4TwDgpgBAsghsAxgCwgZwDwBUoQB7AQB2AJmlGsAE4CWRA5gHxQC8UA3lANoDWEIUOlCwBdAFwVqdelAC+AbgCwAKBWhIsBCnQBBAEY0ANjVCsOUOAeOgJlWgwA0FAPYBbCMGTSAoobQRbKQY5JWV1aHgkVDQsOCp6DzNOYDiE4ED7eicU+I8ASRIM6RC1cAitaKwPVCokwULJTOzUjyLghRVSjW88MGNEEyxaegSqfSMTATZ2FSg5qHCJAHJLCdAlh1n5qggEvAkAJV2esE3leahEOENDPThEHgkAClcK9AlI7TRx6xAASlYzAAbs4aCRQh0wmUoCd+oNhqNYrlgEktnNFlAljk0hs0VAdntDsdemcLlcbncHs9XlF3ppaTEWsAASxgaDwSpIeEYb04cAhjQRhAqFVPMLUed5hisdVhbjJXMCfgifQTqT5uTbvdHlAXm80B99aKaiy2WCIaEutABULamxYTQBvyEcKfpMoAAfHl9R3wwWIpme718m2I2VUS3KRDOIiUBYuqhoPKuH1OiSh4VcERmLhbGYKqVlZarX7yi6K3bKqAAekoAFcAGYN6vqsnXLVU3U0r6my4xtDOQwQAB0hmc9D1DOHJcmf3VsnV+fLCyLmMIYqoZfLSv2NfrTYATC28XNNZSdZOe4C+7HByOxxPu9Fh9iPHOtguVCJ5EA) is from the discord conversation above and this seems enough to convince TS to do inference.  (I am like 80% expecting you to say "oh yeah I tried this and this idea doesn't work for X, Y, Z reasons".)

I think the thought here would be to adjust the lint rule to verify `type` and `netRegex` matched instead of `matches` and `netRegex` like it does now.

valarnin:
I've written out discriminated union code for this a couple of times in that 8 hour session, it works fine but I wasn't sure if you wanted to have an additional field on the trigger to discriminate like that. I can do a full PR with that setup instead. If it's done correctly, you don't even need to adjust the eslint rule because the discriminated type can enforce a subtype of regex (like in this PR) e.g.:

```ts
type ExplicitTriggerAbility = {
    type: 'ability',
    regex: CactbotRegex<MatchesAbility>,
    callback: (matches: MatchesAbility) => void;
};
```

And calls to `NetRegexes`/`Regexes` can return the appropriate type.

quisquous:
Haha, I figured you'd considered that approach at some point.  (Were there other workable approaches you thought of and discarded as well?)

Maybe you disagree, but I think an additional field is much less invasive than wrapping every trigger in a function.  I think it also makes it more plausible to have a lint rule auto-add the type field properly based on the `netRegex` type to help automate converting triggers (and much harder to do the wrapping in a function).

I think my preferences for approaches are along the lines of "explicit field" >= "patch as-is" >>> "do nothing" >>>>> "builder function".  I think "explicit field" and "patch as-is" both have some trade-offs either way that make me consider them, but "explicit field" is a lot cleaner from a typescript perspective.

valarnin:
There were a few other approaches I tried but none of them worked out due to various reasons. If we're on the same page for `explicit field + discriminated union`, I can close this PR out and start working on that instead.

Normally when I do a PR, I'll do all the code changes in one commit, but for this one I think I'll do the type/logic changes in one and the trigger changes in the second, so they're a bit separated.

quisquous:
`explicit field + discriminated union` sounds good to me, thanks!

valarnin:
Closing. I'll try to get the new PR in either tonight or tomorrow.

PR #3121, opened 2021-06-16
ci: Fix Triggers Branch Deployment

panicstevenson:


panicstevenson:
This broke with the creation of the manifest declarations (and the dependency on them inside popup-text). Surprised to see that nothing reported this being broken.

quisquous:
Yeah, I think it would be good if something caused a non-zero error code in this path.

panicstevenson:
> Yeah, I think it would be good if something caused a non-zero error code in this path.

I checked the exit status from running the same code as in the workflow while this was broken and it was non-zero; I was confused.

PR #3122, opened 2021-06-17
ci: add workflow lint pr title

MaikoTan:
Import [amannn/action-semantic-pull-request](https://github.com/amannn/action-semantic-pull-request) to make sure that PR title matches [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/),
i.e. something like:
```
<type>[optional scope]: <description>
```

in particular, this PR defines types such as 
- build: build-related change only
- ci: CI(Github Actions currently) changes only
- fix: fixed a bug
- feat: added a new feature
- refactor: neither fix a bug nor add a feature
- chore: others
- docs: documents changes only
- style: code style changes only
- perf: improved performance
- test: test changes only
- i18n: translation only

Also defines scopes such as:
- resources
- ui
- pullcounter
- raidboss
- oopsy
- radar
- eureka
- jobs
- test
- util
- config

### DIFF
--- a/.github/workflows/sematic-pr-title.yml
+++ b/.github/workflows/sematic-pr-title.yml
@@ -1,0 +1,54 @@
+name: "Sematic PR Title"
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - edited
+      - synchronize
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: amannn/action-semantic-pull-request@v3.4.0
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          types: |
+            build
+            ci
+            fix
+            feat
+            refactor
+            chore
+            docs
+            style
+            perf
+            test
+          scopes: |
+            resources
+            ui
+            pullcounter
+            raidboss
+            oopsy
+            radar
+            eureka
+            jobs
+            test
+            util
+            config
+
+          requireScope: false
+          # For work-in-progress PRs you can typically use draft pull requests 
+          # from Github. However, private repositories on the free plan don't have 
+          # this option and therefore this action allows you to opt-in to using the 
+          # special "[WIP]" prefix to indicate this state. This will avoid the 
+          # validation of the PR title and the pull request checks remain pending.
+          # Note that a second check will be reported if this is enabled.
+          wip: true
+          # When using "Squash and merge" on a PR with only one commit, GitHub
+          # will suggest using that commit message instead of the PR title for the
+          # merge commit, and it's easy to commit this by mistake. Enable this option
+          # to also validate the commit message for one commit PRs.
+          validateSingleCommit: true
--- a/.github/workflows/sematic-pr-title.yml
+++ b/.github/workflows/sematic-pr-title.yml
@@ -26,6 +26,7 @@ jobs:
             style
             perf
             test
+            i18n
           scopes: |
             resources
             ui


.github/workflows/sematic-pr-title.yml
```diff
@@ -0,0 +1,55 @@
...
+            ci
+            fix
+            feat
+            refactor
+            chore
+            docs
+            style
+            perf
+            test
+            i18n
```
quisquous, line 29: I think this is the part I'm least sure about.  If other folks want this, I'm happy to go for it with some minor tweaks, but if it was just up to me I would be not sure.

My feelings and thoughts here are:
* The nicest part of this is having a bot being able to enforce commit message structure.
* This list as written has lots of overlap in `types` and `scopes`.  It seems like build, ci, docs, style, test, i18n are all `scopes` and not `types`?
* I don't find a lot of value in knowing `fix` vs `feat`.  I think it's often pretty ambiguous (features can exist to "fix" old/missing code?)
* This feels slightly redundant with https://chris.beams.io/posts/git-commit/#imperative, where I find that a good PR title has an imperative verb to start (e.g. add/fix/refactor/remove) and so `fix(raidboss): fix thing` or `feat(oopsy): add thing` don't seem to add much more than `raidboss: fix thing` or `oopsy: add thing` and it takes up space in an already space-limited PR title
* I know we have PR labels, but I do think `scopes` is still useful for browsing commit messages, as the labels don't appear in `git log`.
* This is maybe a larger issue, but I am definitely not following semantic versioning for cactbot, mostly because I have never figured out what to really consider a breaking change (happy to discuss if people want to define this better?), and so semantic PR titles have less value because we're not doing semantic versioning

quisquous:
Sorry for ping, but I wanted to solicit extra opinions from everybody to understand what folks find useful and what they'd like in PR titles.  Do folks want something like this and find this valuable?  Are folks indifferent?  Any other thoughts or feelings?

CC @Trim21 @panicstevenson @valarnin @JLGarber @Jaehyuk-Lee @Akurosia @Echoring @MKhayle @rcj (these are all folks who have committed stuff in the last ~month, but please leave opinions if you have them and are not listed, just trying not to @ everyone)

valarnin:
From my viewpoint, the two most helpful things are the type or part of the change prefixing the PR title, e.g. `raidboss: ...`, and the labels which works fantastic already.

The PR title prefix acts as an indicator as to the target component of the change, and the labels act to show impact areas, so I know that if I see for example a PR with `fisher:` even though I haven't really worked on fisher directly, if it's impacting something I'm familiar with according to labels, I can pop in and check the PR.

To me, as long as the PR title matches something like this, I think it's sufficient. How exactly you'd configure to enforce this, I don't know.

```ts
/(?:resources|ui|oopsy|<etc>)(?:[ ]?\([^)]+\))?: <description>/
```

rcj:
I don't see a lot of additional value of having this in the commit title compared to e.g. some line like
    Scope: <whatever> at the end of the commit message.

Just makes it harder to keep the commit message within 50 chars.
But I don't really mind it either in case others see this as beneficial.

Trim21:
I think this world help if we also want to generate change log from commits, otherwise maybe not really helpful

.github/workflows/sematic-pr-title.yml
MaikoTan: > * This list as written has lots of overlap in `types` and `scopes`.  It seems like build, ci, docs, style, test, i18n are all `scopes` and not `types`?

This is what I am basically following the [Angular Commit Message Guildeline](https://gist.github.com/brianclements/841ea7bffdb01346392c), and I think most of their thoughts are suitable to ours.

> * I don't find a lot of value in knowing `fix` vs `feat`.  I think it's often pretty ambiguous (features can exist to "fix" old/missing code?)

I think if there are new functions/options/dungeon supports, I would treat it as a `feature`. If there are old code updating, such as "update actions id to match the new patch in-game" or "delete unusable monk lightning gauge because the new patch removed it", I prefer to treat it as a `fix` more than a `feature`, this code are actually "broken"/"bugged" in the new patch, isn't it?

If there is a PR contains both types of changes, I think `feat` is more important than `fix`, the latter one can be written as the description text of the PR.

Also, for `docs`/`style`/`i18n`/etc, they should be used if the PR contains the current type of changes ONLY. If there are some other changes, then the PR should be noted as `feat`/`fix` or something.

> * This feels slightly redundant with https://chris.beams.io/posts/git-commit/#imperative, where I find that a good PR title has an imperative verb to start (e.g. add/fix/refactor/remove) and so `fix(raidboss): fix thing` or `feat(oopsy): add thing` don't seem to add much more than `raidboss: fix thing` or `oopsy: add thing` and it takes up space in an already space-limited PR title

As my knowing, in this situation, you can just write `feat(raidboss): things` such as `feat(raidboss): Leviathan Unreal timeline` or `feat(raidboss): UCU triggers improvement` if there are a bunch of changes.

> * This is maybe a larger issue, but I am definitely not following semantic versioning for cactbot, mostly because I have never figured out what to really consider a breaking change (happy to discuss if people want to define this better?), and so semantic PR titles have less value because we're not doing semantic versioning

Mostly I was trying to make PR easy to be reviewed, so a semantic title would helpful in this situation. On the other hand, as Trim said, we can generate changelogs from semantic commit messages other than writing it manually.

Jaehyuk-Lee:
I don't mind, but I see `plugin` not listed.

panicstevenson:
I don't have strong opinions here; but, this was one of the things I had brought up with quisquous back when originally bringing up the migration from Travis CI to GitHub Actions would be that the build, and even a changelog, could be automatically generated. Some time has passed, but overall we never really settled on a strict enforcement of PR titles, although it still kind of coalesced (somewhat) as a norm. I figured it was somewhere in our documentation, but a quick search left me unable to find it.

My thoughts really come down to what changelog generation would look like, and what does type vs scope really mean. I don't perceive a large value from the types vs scopes distinction, and I don't see a changelog as having a large benefit to users with that distinction; or, for example, I don't see a large benefit in:
```markdown
# Changelog
## Raidboss
### Features
- Add Sastasha triggers and timeline (#...)
- Add Odin triggers and timeline (#...)
### Fixes
- Fix issue where bongos played during fates (#...)
## Oopsyraidsy
### Features
- Add Sastasha oopsy (#...)
## Fisher
### Fixes
- Fix fishing bars randomly disappearing (#...)
``` 
vs
```markdown
# Changelog
## Raidboss
- Add Sastasha triggers and timeline (#...)
- Add Odin triggers and timeline (#...)
- Fix issue where bongos played during fates (#...)
## Oopsyraidsy
- Add Sastasha oopsy (#...)
## Fisher
- Fix fishing bars randomly disappearing (#...)
``` 
I could be wrong, but I feel as though the verbiage and language used in the title wouldn't (can't?) change much as well. If everything after `type(scope):` is taken as the description in changelog automation, then, while `fix(raidboss): Fate Bongos` may mean something to a developer, `Fate Bongos (#...)` is relatively worthless to a user, even if it is coming under the Raidboss and Fixes subheaders (or so I personally feel). So, with that, by adding `fix(...)` I've taken away 5 characters from an already very limited space (I count the absolute max GitHub is willing to show differs from the conventional git ettiquite that I learned of 52 character titles. GitHub will show 70 characters, with 8 of those being taken up with ` (#____)` for the PR link if the PR has 4 digits). Mostly, I don't know that a distinction between `refactor(build): Refactor Webpack Configs` says a lot more than `build: Refactor Webpack Configs`, since I feel the information is going to be present either way, but again that might be entirely personal to me.

Within the [conventional commit spec](https://www.conventionalcommits.org/en/v1.0.0/#summary), scope is optional but type is not; I would think that to retain what we currently have, we would swap the scope and the types to be such that the types are the module name and that the scope is left undefined, opting for `raidboss:` over `raidboss(fix):` since the latter seems a bit silly. That being said, if there's generation tools that let us say the type is optional and to key off the scope, that's fine too. Or, if the changelog can only be ordered by scope.

Overall, I feel like I've written a lot here, but I generally don't care very strongly either way; I think the current system works fine, and could be enforced and elevated into changelog automation, but I'm not unwilling to add 6 characters to a PR title. I think the thing for me here is that I would imagine that we end up removing the types (as they are in this iteration, fix, feat, refactor, etc.) from the changelog anyway, which would make them kind of fruitless to enforce adding to commits.

However, all of above being said, I am strongly in favor of settling on _something_, and I would love for robots to enforce it, so I am very much in favor of this pull request, basically in any form that it ends up landing in.

Trim21:
I think the first one is better than the second one...

quisquous:
> I think the first one is better than the second one...

Sorry to disagree, but I find the second one much clearer.

I think right now I am writing the changelog for two audiences:
* users who want to know why they should bother upgrading
* people doing a lot of discord support (Techno) who want to know what's changed (and thus might be broken)

I deliberately ignore translations, dev stuff, ci changes, lint changes, etc, mostly because it feels too noisy and not what most users care about.  (I can see some argument that a changelog should also support a "devs who want to know what has changed without reading every commit" and maybe we need a "for users" and "for devs" changelog section, with the "discord support" category caring about both, potentially).

However, I suspect most users or devs don't read the changelog at all already and so any extra friction in their way like more sections that "bury the lede" so to speak will probably make fewer people read the changelog.  (Example: despite saying `jobs: Add notification sound to proc boxes (#3097)` there's been already two questions on discord about where the noise is coming from.)  Mostly saying that making the changelog easy to read seems important.

What repurposing the types to be what we're currently doing but with no scope, e.g.
```
          types: |
            resources
            ui
            pullcounter
            raidboss
            oopsy
            radar
            eureka
            jobs
            test
            util
            config
            lint
            ci
```

In terms of automating the changelog, being able to separate out both "dev-only" and "translation" changes from other kinds of changes would be the biggest help.

MaikoTan:
Close that I think it might not suitable for our use, especially that we have multiple "modules" and focus on "where it changed" instead of "what type of the changing".

quisquous:
Hmm.  I am not so opposed to this that I am vetoing.  I know that my voice carries a lot of weight, but I wasn't meaning to express an opinion so heavily, and more trying to have a discussion.  It seems like we're not going to have a unanimous agreement here, so maybe we can talk about what parts of this are important to you and see if we can meet somewhere in the middle?

Is the "semantic" part of this important to you, in terms of having semantic versioning as a next step? (If so, what constitutes a "breaking" change for you in cactbot? Or said differently, what do you consider the supported API of cactbot, when we also have trigger and CSS overriding?)

Honestly I think the biggest issue for me is just the length.  `feat()` is an extra 6 characters, or >10% of a 50 character PR title.  Would you be supportive of emojis instead of types? e.g. bug = 🐛 / 🦟, feat = 💡 /, breaking change = 💥 / 💣 / ❗ , etc? Then titles would be `🐛raidboss: fix broken trigger` sort of thing?

MaikoTan:
I am just thinking that the action I added is not so customisable to the usage that you suggested. I'd like to create an action manually just like @Trim21 working of label tagging rather than depending on the action. I did understand your comment and didn't want to just give up this thing, however I have no many time so my pace would be slower. Sorry that I didn't describe the reason. Once I prepared the new version, I can open a new PR?

quisquous:
That sounds good!  Sorry if I misunderstood the situation, but I didn't want you to close this thinking that there wasn't support for this idea.  I think having some sort of action to enforce PR titles and length would be great, and I think I am still not sure what the best title format is, but am happy to talk more another PR in the future.  :)

PR #3129, opened 2021-06-18
resources: Improve Timerbar Performance

panicstevenson:


panicstevenson:
Open to other ideas or suggestions for this; originally my thought was to move from recalculating width to something that doesn't need to run through layout, and it turns out `scaleX` is basically exactly the same as `width` here, but only fires a repaint. Additionally, I removed `will-change: transform` from the foreground element's CSS, which saw performance gains as well for me, but it may be doing something under the hood that would make sense to leave it in.

Data collected from timerbar_test at 6x CPU slowdown:

image

Additionally, I saw my max frames per second in timerbar_test at 6x CPU slowdown go from 33.6 FPS with the implementation on `main`, to 49.2 FPS on this branch.

quisquous:
Could you test the lippe skin with this if you haven't already?

panicstevenson:
> Could you test the lippe skin with this if you haven't already?

Yep, it looks good!

PR #3130, opened 2021-06-18
lint: Have Ordering Fix Preserve Comments

panicstevenson:


panicstevenson:
As mentioned in https://github.com/quisquous/cactbot/pull/3116

panicstevenson:
This has an unintended side effect that it will insert newlines to things that don't have them, such as:
```js
  collectSeconds: 0.5,
  damageRegex: '482A', // Line comment after
  suppressSeconds: 5,
```
Will become:
```js
  damageRegex: '482A',
  collectSeconds: 0.5,
  // Line comment after
  suppressSeconds: 5,
```
But that feels fine overall.

quisquous:
> This has an unintended side effect that it will insert newlines to things that don't have them, such as:

That also seems fine to me.

Does the code as-is catch comments after the last property too? Or is that not a thing?

PR #3131, opened 2021-06-18
raidboss: Remove Timeline Callback Indirection

panicstevenson:


ui/raidboss/timeline.ts
```diff
@@ -947,14 +923,14 @@ export class TimelineUI {
...
     if (this.timeline) {
-      this.timeline.SetAddTimer(this.OnAddTimer.bind(this));
-      this.timeline.SetRemoveTimer(this.OnRemoveTimer.bind(this));
-      this.timeline.SetShowInfoText(this.OnShowInfoText.bind(this));
-      this.timeline.SetShowAlertText(this.OnShowAlertText.bind(this));
-      this.timeline.SetShowAlarmText(this.OnShowAlarmText.bind(this));
-      this.timeline.SetSpeakTTS(this.OnSpeakTTS.bind(this));
-      this.timeline.SetTrigger(this.OnTrigger.bind(this));
-      this.timeline.SetSyncTime(this.OnSyncTime.bind(this));
+      this.timeline.addTimerCallback = this.OnAddTimer.bind(this);
```
quisquous, line 926: What do you think about just setting `TimelineUI` on `Timeline` and having it make all of these `OnFoo` calls directly rather than calling a collection of callback functions?

panicstevenson: Doesn't TimelineUI have a Timeline, which would then have a TimelineUI?

quisquous: Yeah, `TimelineUI` has a reference to the active `Timeline` and `Timeline` itself calls back into `TimelineUI` when it needs to do things.  I think this is fine for `Timeline` to have a reference to `TimelineUI` while it is active.

An alternative here is that you could arguably have `TimelineController` own the active `Timeline` and set the `TimelineUI` on the `Timeline` when it becomes active/inactive and `TimelineUI` doesn't need a ref to it at all.

PR #3133, opened 2021-06-19
fix: remove `isMain` section in coinach.ts

MaikoTan:
`require.main` is not valid in es modules.
There is an [opening issue](https://github.com/nodejs/modules/issues/274)
which is trying to implement `import.meta.main` but currently there isn't.

In this file, just remove the section completely.

also fix some new lines to be consistent as python version


util/coinach.ts
```diff
@@ -236,8 +237,7 @@ export default data;`;
       console.log(`wrote: ${filename}`);
   }
 }
-
-if (require.main === module) {
+if (isMain(import.meta)) {
```
quisquous, line 240: We could arguably just drop this section.  I don't think I've ever run this on its own.

MaikoTan: You mean remove all the `if __name__ == "__main__":`-like section in these files completely?
I never run this code, either.

By the way, I am currently writing the `util entry` that I said [before](https://github.com/quisquous/cactbot/pull/2603#issuecomment-802670848) based on [inquirer.js](https://www.npmjs.com/package/inquirer)
when I complete that work, we don't need this `isMain` code completely.

quisquous: I mean in this file in particular.  I'd have to check other files to know for sure.

MaikoTan: Done.

PR #3135, opened 2021-06-20
raidemulator: refactor raidemulator.js initialization

quisquous:
This removes a bunch of `let`s, an IIFE, and turns a `.then` into an `await`.

### DIFF
--- a/ui/raidboss/raidemulator.js
+++ b/ui/raidboss/raidemulator.js
@@ -29,6 +29,31 @@ import defaultOptions from './raidboss_options';
 import '../../resources/defaults.css';
 import './raidemulator.css';
 
+
+function showModal(selector) {
+  const modal = document.querySelector(selector);
+  const body = document.body;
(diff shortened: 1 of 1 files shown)


ui/raidboss/raidemulator.js
```diff
@@ -29,398 +29,382 @@ import defaultOptions from './raidboss_options';
...
+    persistor: persistor,
+    encounterTab: encounterTab,
+    emulatedPartyInfo: emulatedPartyInfo,
+    emulatedMap: emulatedMap,
+    emulatedWebSocket: emulatedWebSocket,
+    timelineUI: timelineUI,
+  };
+});
 
 function showModal(selector) {
```
valarnin, line 409: Might make sense to move the modal functions to the top of the file now from a flow standpoint?

quisquous: Thanks, done!

PR #3137, opened 2021-06-20
util: Rework trigger structure to have union discriminator

valarnin:
Following up for #3120

This PR:

1. Adjusts the trigger structure for network triggers to have a discriminator type. I had to additionally discriminate between NetRegexTrigger and RegexTrigger, so the former has `netType` and the latter has `type`, to match the naming schema. `type` never needs to be specified, it just has to exist for typescript to properly discriminate between `NetRegex` and `Regex`.
2. Uses the new discriminated unions to enforce that all properties of a network trigger are the correct typing, e.g. a trigger must use the same `NetRegexes.***` helper for each `netRegex*` property.
3. Uses the new discriminated unions to automatically provide type matching for all trigger parameters. Someone who knows what they're doing with the eslint rules should update `cactbot-matches-type-references.js` to maybe make specifying type optional if the trigger is discriminated, as well as update the references to the indexed `NetAllMatches` type instead?
4. Manages to make the discriminator optional. If all trigger properties use canned conditions/responses/etc, the discriminator isn't required. Even if you don't specify the discriminator, it will still error if you try to use two different types of NetRegexes calls, so that type checking is still present. For example, the `T13 Earthshaker` trigger doesn't need the `netType` discriminator, so I didn't adjust it in this PR.

All of the logic changes are in https://github.com/quisquous/cactbot/commit/86254d0e2495cbe202ad671bbe405148c32eced5
All of the trigger updates are in the two commits after that.


valarnin:
I wasn't sure where `netType` should go in the trigger ordering, so I didn't add it to any of the documentation, eslint rules, or tests.

MaikoTan:
Yeah, when I wrote the `Matches` generic type I never thought that it would be used such widely, I would be glad to see it gets refactoring and cleaning.

ui/raidboss/data/02-arr/raid/t8.ts
```diff
@@ -112,18 +116,19 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
+      netType: 'StartsUsing',
```
quisquous, line 119: > I had to additionally discriminate between NetRegexTrigger and RegexTrigger, so the former has netType and the latter has type, to match the naming schema. type never needs to be specified, it just has to exist for typescript to properly discriminate between NetRegex and Regex.

Forgive me for being skeptical, but how does an optional field contribute to discrimination?

Things seem to work if I change `PartialRegexTrigger` to be defined like this (and `s/PartialRegexTrigger<'None'>/PartialRegexTrigger/`):
```typescript
type PartialRegexTrigger = {
  regex: RegExp;
  regexDe?: RegExp;
  regexFr?: RegExp;
  regexJa?: RegExp;
  regexCn?: RegExp;
  regexKo?: RegExp;
};
```

Mostly I say this, because if I had my druthers, I would much rather just call this `type` everywhere and not bother with type inference on `RegexTrigger`.

My other feedback is that I also would order `type` after id but before `netRegex`.  I think `id` being first is helpful for scanning trigger files.

Could you also update the following to mention `type` and its ordering?
* [this lint](https://github.com/quisquous/cactbot/blob/main/eslint/cactbot-trigger-property-order.js#L29)
* [this test](https://github.com/quisquous/cactbot/blob/main/test/helper/test_trigger.js#L329)
* [this doc](https://github.com/quisquous/cactbot/blob/main/docs/RaidbossGuide.md#trigger-structure)
* [responses](https://github.com/quisquous/cactbot/blob/main/resources/responses.ts#L37)

types/net_matches.d.ts
```diff
@@ -0,0 +1,16 @@
+import { NetAllFields } from './net_fields';
+import { TriggerTypes } from './net_trigger';
+
+export type Matches = { [s in T]?: string };
+export type ToMatches<F> = Record<keyof F, string>;
+
+
+export type NetMatches = Matches & {
+};
```
quisquous, line 9:
```suggestion
export type NetMatches = Matches;
```

Am I misunderstanding what this is doing? (Alternatively maybe there's no need for a separate `NetMatches` if it's the same as `Matches`?)

types/trigger.d.ts
```diff
@@ -1,19 +1,17 @@
...
 import { TimelineReplacement, TimelineStyle } from '../ui/raidboss/timeline';
 import { RaidbossData } from './data';
-
-export interface BaseRegExp<T> extends RegExp {
-  groups?: {
-    [s in T]?: string;
-  };
-}
+import { CactbotBaseRegExp, TriggerTypes } from './net_trigger';
+import { NetAllMatches, Matches as MatchesAny } from '../types/net_matches';
```
quisquous, line 5: I think there is only one use of `MatchesAny` so no need to import it with a separate name.

Can you leave a TODO to remove `Matches<Params>` below? (I assume that's the followup here?, to switch everything to `NetAllMatches`?)

resources/regexes.ts
```diff
@@ -695,7 +724,7 @@ export default class Regexes {
     return anyOfArray(array);
   }
 
-  static parse(regexpString: RegExp | string): RegExp {
+  static parse<T extends ParamsAllTypes>(regexpString: RegExp | string): Regex<ParamsAll[T]> {
```
quisquous, line 727: There's the same issue here as with the previous patch.  There are parts of code like [this line](https://github.com/quisquous/cactbot/blob/main/ui/oopsyraidsy/oopsyraidsy.js#L1217) that are calling `Regexes.parse` with a `NetRegexes`-built regex.

The compiler would have yelled if anything in TypeScript was doing this (and so this is more just an issue for the future and a conceptual issue right now).  We can defer changing something in this PR (because there's enough going on), but this feels like something that could get cleaned up afterwards (either by having two separate parse functions, or changing callers appropriately, or both, or something else).

types/net_fields.d.ts
```diff
@@ -0,0 +1,321 @@
...
+  skillSpeed: 15;
+  spellSpeed: 16;
+  tenacity: 18;
+};
+
+// 0x0D, non-net-log only
+export type NetHasHPFields = NetFields & {
+  name: 2;
+  hp: 3;
+};
```
quisquous, line 78: Would it make sense to just drop this? See comment in `trigger.d.ts` about having triggers only be netregex.

types/net_matches.d.ts
```diff
@@ -0,0 +1,16 @@
+import { NetAllFields } from './net_fields';
+import { TriggerTypes } from './net_trigger';
+
+export type Matches = { [s in T]?: string };
+export type ToMatches<F> = Record<keyof F, string>;
```
quisquous, line 5:
```suggestion
export type ToMatches<Fields> = Record<keyof Fields, string>;
```

Too many single letters in this file for my brain.  :)

types/trigger.d.ts
```diff
@@ -140,7 +149,7 @@ export type TriggerSet<Data extends RaidbossData> = {
   overrideTimelineFile?: boolean;
   timelineFile?: string;
   timeline?: TimelineFunc;
-  triggers?: NetRegexTrigger<Data>[];
+  triggers?: (NetRegexTrigger<Data> | RegexTrigger<Data>)[];
```
quisquous, line 152: I think this should stay `NetRegexTrigger<Data>` honestly.  If somebody is itching to convert a file that has a regex hp trigger in it, they can convert it to the new utility first.  I think it's fine to be stricter about the kinds of triggers that exist in built-in cactbot.

quisquous:
Thank you for making this work.  I am *so* excited to see that this is possible.  I had given up hope.

types/net_fields.d.ts
```diff
@@ -0,0 +1,321 @@
...
+
+export type NetFields = Fields & {
+  event: 0;
+  timestamp: 1;
+};
+
+// 0x00
+export type NetGameLogFields = NetFields & {
+  code: 2;
```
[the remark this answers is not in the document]
valarnin: With this pass at the typing, I included the field offsets as well, in the hopes that that'll allow cleaning things up properly later. That was the one big thing missing from the `MatchesAbility` etc types that were already set up.

types/net_matches.d.ts
valarnin: This was leftover from when I was originally trying to have RegexTrigger have typing as well, before I hit the 50 union entries limit.

types/trigger.d.ts
valarnin: That would conflict with the `export type Matches<Params>` below, that's why the rename in the import. Something to fix with cleanup in followup.

types/trigger.d.ts
valarnin: This causes issues downstream in PopupText that would need some coercion to fix, IIRC, though I'm not at my desktop now to confirm.

types/net_trigger.d.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class CactbotRegExpExecArray<T extends TriggerTypes> extends RegExpExecArray {
+  groups?: NetAllMatches[T];
+}
+
+export class CactbotBaseRegExp<T extends TriggerTypes> extends RegExp {
+  exec(string: string): CactbotRegExpExecArray<T> | null;
+}
+
+export type TriggerTypes = 'GameLog' |
```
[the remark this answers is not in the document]
valarnin: Yeah, this should be fine. I wasn't doing it that way initially due to some weirdness with the `'None'` type but that was fixed eventually.

ui/raidboss/data/02-arr/raid/t8.ts
```diff
@@ -112,18 +116,19 @@ const triggerSet: TriggerSet<Data> = {
...
       id: 'T8 Brainjack',
       netRegex: NetRegexes.startsUsing({ id: '7C3', source: 'The Avatar' }),
       netRegexDe: NetRegexes.startsUsing({ id: '7C3', source: 'Avatar' }),
       netRegexFr: NetRegexes.startsUsing({ id: '7C3', source: 'Bio-Tréant' }),
       netRegexJa: NetRegexes.startsUsing({ id: '7C3', source: 'アバター' }),
       netRegexCn: NetRegexes.startsUsing({ id: '7C3', source: '降世化身' }),
       netRegexKo: NetRegexes.startsUsing({ id: '7C3', source: '아바타' }),
-      alertText: (data, matches: MatchesStartsUsing, output) => {
+      alertText: (data, matches: NetAllMatches['StartsUsing'], output) => {
```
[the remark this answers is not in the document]
valarnin: Yeah, I talked about this in the initial PR comment, this is 100% auto-inferred now if `netType` is specified, so the logic should probably be `netType is set || <previous logic>`, I'm just not familiar at all with the eslint scripts.

ui/raidboss/data/02-arr/raid/t8.ts
valarnin: I tested this out a bit more this morning, and it does seem to work fine with `type` and no second discriminator. I'm really not sure why it wasn't working last night, maybe just leftover issues from when eslint/tsc broke on the more-than-50-union-members issue.

ui/raidboss/data/02-arr/raid/t8.ts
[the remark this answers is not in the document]
valarnin: Actually, upon further review, the old eslint rule isn't required at all now I think? Since in all cases it's either auto-inferred or irrelevant due to using only canned conditions/responses.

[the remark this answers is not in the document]
valarnin: I removed the eslint rule with the latest commit, since that's what made sense to me. Let me know if you'd rather it be reverted and modified instead.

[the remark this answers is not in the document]
MaikoTan: I am not quite understand what is wrong here. Is it just the lint rule complains but `tsc` says nothing?

types/net_fields.d.ts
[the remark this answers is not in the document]
valarnin: I went through and did most of the cleanup here anyways 😃.

There's still some more cleanup that could be done related to Regexes but there's Params differences between that and NetRegexes (e.g. `sourceId` vs `id`) so I don't know if that can be fully cleaned up.

ui/raidboss/data/02-arr/raid/t8.ts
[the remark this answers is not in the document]
valarnin: The lint rule doesn't serve any purpose after this PR, since the `matches` parameter is either always properly typed (if `type: ` property is specified) or irrelevant (if only canned conditions/responses are used) if the trigger is a netRegex, and since all typescript conversions will have to be netRegex per quisquous in another comment on this PR, it'll always be unneeded.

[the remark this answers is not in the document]
valarnin: To clarify further, consider the following cases:

1. `type` is specified, `matches` is used in a callback. `matches` is automatically properly typed:
image
2. `type` is not specified, trigger functions only use canned responses. `matches` is never specifically in the trigger so the old rule doesn't apply:
image
3. Someone attempts to write a trigger that does not specify type but attempts to use a non-canned response. tsc won't compile since it can't discriminate the union properly:
image

ui/raidboss/data/02-arr/raid/t1.ts
```diff
@@ -30,8 +31,9 @@ const triggerSet: TriggerSet<Data> = {
       response: Responses.interrupt(),
     },
     {
-      // Indiscriminate Hood Swing
       id: 'T1 Initiated',
+      // Indiscriminate Hood Swing
```
quisquous, line 35:
```suggestion
      // Indiscriminate Hood Swing
      id: 'T1 Initiated',
```

ui/raidboss/data/02-arr/raid/t12.ts
```diff
@@ -141,14 +146,15 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
-      // Chain Of Purgatory
       id: 'T12 Chain',
+      // Chain Of Purgatory
```
quisquous, line 150:
```suggestion
      // Chain Of Purgatory
      id: 'T12 Chain',
```

ui/raidboss/data/02-arr/raid/t2.ts
```diff
@@ -16,25 +15,28 @@ const triggerSet: TriggerSet<Data> = {
...
       id: 'T2 Ballast',
+      type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '4C5', capture: false }),
       suppressSeconds: 3,
       response: Responses.getBehind(),
     },
     {
-      // Allagan Rot
       id: 'T2 Rot',
+      // Allagan Rot
```
quisquous, line 32:
```suggestion
      // Allagan Rot
      id: 'T2 Rot',
```

ui/raidboss/data/02-arr/trial/titan-ex.ts
```diff
@@ -69,15 +68,16 @@ const triggerSet: TriggerSet<Data> = {
   ],
   triggers: [
     {
-      // Doesn't seem like this happens twice, but let's be safe.
       id: 'TitanEx Rock Throw',
+      // Doesn't seem like this happens twice, but let's be safe.
```
quisquous, line 72:
```suggestion
      // Doesn't seem like this happens twice, but let's be safe.
      id: 'TitanEx Rock Throw',
```

types/net_matches.d.ts
```diff
@@ -0,0 +1,12 @@
+import { NetAllFields } from './net_fields';
+import { TriggerTypes } from './net_trigger';
+
+export type Matches = { [s in T]?: string };
+export type ToMatches<Fields> = Record<keyof Fields, string>;
+
+export type NetAllMatches = {
```
quisquous, line 7: This is just a naming suggestion, but feel free to disagree.  I feel like `NetAllMatches` is now used in a bunch of places, but is kind of a long name when paired with square brackets.  What do you think about `NetMatches` and dropping the `All` out of it?

Feel free to disagree and we can go with what's here.  Just a drive-by thought.

ui/raidboss/data/02-arr/raid/t1.ts
valarnin: Don't quote me on this, but I think all of these comments ended up after `id` because I used the automatic-order-fix option? Something to look into in the future I guess?

types/net_matches.d.ts
valarnin: There's a bit of a naming schema going on here, these are all named with the same schema:

```ts
NetAllMatches
NetAllFields
NetAllParams
NetAllProps
```

I feel like if we rename one of them, we should rename all of them?

valarnin: Went ahead and renamed them all, feel free to resolve if that's good for you.

quisquous:
Thanks for the renaming.  :+1:

PR #3138, opened 2021-06-20
fix(timeline): execute timeline trigger with current time

### LINKED ISSUE #3126: BUG: P2 `UCU Bahamut's Claw` is missing in real game
image

```JS
    {
      id: 'UCU Bahamut\'s Claw',
      regex: /Bahamut's Claw x5/,
      beforeSeconds: 5,
      // It's tough to track who this is on, especially for the first one.
      // Both tanks should care about the tankbuster because they can throw
      // mitigation on the other, so just always play this for both tanks.
      condition: Conditions.caresAboutPhysical(),
      suppressSeconds: 1,
    },
```

there is a recent change that rename prop `conditions` to `condition`, but I didn't think it's relative.

Trim21:
should close #3126, this was introduced in https://github.com/quisquous/cactbot/pull/3079

In old implement, all timeline trigger is executed as `timeline.timebase`, not the current time.

the make all `currentTime` become base time, not real current time.

could be tested with dummy:

```JS
const trigger = {
  zoneId: ZoneId.MiddleLaNoscea,
  timeline: [
    `5 "test"`,
    `10 "test"`,
    `15 "test"`,
    `20 "test"`,
    `25 "test"`,
    `30 "test"`,
    `35 "test"`,
  ],
  timelineTriggers: [
    {
      id: 'trim21 test refresh',
      regex: /Death/,
      run() {
        // location.reload();
      },
    },
    {
      id: 'trim21 test',
      regex: /test/,
      preRun(data) {
        data.times = data.times ?? 0;
        data.times++;
      },
      suppressSeconds: 1,
      beforeSeconds: 3,
      infoText(data) {return `test message ${data.times}`;},
    },
  ],
};

Options.Triggers.push(trigger);
```

### DIFF
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -699,6 +699,7 @@ export class Timeline {
       ++this.nextEvent;
     }
   }
+
   private _AddPassedTexts(fightNow: number, currentTime: number): void {
     while (this.nextText < this.texts.length) {
       const t = this.texts[this.nextText];
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -709,16 +709,16 @@ export class Timeline {
         break;
       if (t.type === 'info') {
         if (this.showInfoTextCallback)
-          this.showInfoTextCallback(t.text, this.timebase);
+          this.showInfoTextCallback(t.text, currentTime);
       } else if (t.type === 'alert') {
         if (this.showAlertTextCallback)
-          this.showAlertTextCallback(t.text, this.timebase);
+          this.showAlertTextCallback(t.text, currentTime);
       } else if (t.type === 'alarm') {
         if (this.showAlarmTextCallback)
-          this.showAlarmTextCallback(t.text, this.timebase);
+          this.showAlarmTextCallback(t.text, currentTime);
       } else if (t.type === 'tts') {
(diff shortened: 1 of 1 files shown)


valarnin:
Ah, yeah, that makes sense. Did you test in raidemulator to make sure this didn't have any weird downstream effect as well? The timeline stuff in raidemulator is still a bit fragile unfortunately.

Trim21:
> Ah, yeah, that makes sense. Did you test in raidemulator to make sure this didn't have any weird downstream effect as well? The timeline stuff in raidemulator is still a bit fragile unfortunately.

Not yet.

Trim21:
> Ah, yeah, that makes sense. Did you test in raidemulator to make sure this didn't have any weird downstream effect as well? The timeline stuff in raidemulator is still a bit fragile unfortunately.

raidemulator looks  fine.

Trim21:
I'm not formalir with timeline, didn't know if we should replace all timebase with current time in this function?

valarnin:
> 
> 
> I'm not formalir with timeline, don't know if I should replace all timebase with current time in this function?

Yes. PopupText callbacks expect the current timebase, so updating these would make the code more correct, even if there isn't a problem currently.

quisquous:
Thanks for the fix!

PR #3141, opened 2021-06-20
raidemulator: Convert RaidEmulator class to typescript 

valarnin:
There's some cleanup related to the properties on this class that I'll want to revisit later after the rest of emulator is converted to typescript.


valarnin:
There's an issue with the conversion, converting to draft until I fix it.

PR #3142, opened 2021-06-20
raidemulator: Convert RaidEmulatorPopupText to typescript

valarnin:


ui/raidboss/emulator/EventBus.ts
```diff
@@ -5,7 +5,7 @@ type Scope = object;
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 type Param = any;
 
-type CallbackFunction = (...args: Param) => void;
+type CallbackFunction = ((...args: Param) => void) | ((...args: Param) => Promise<void>);
```
quisquous, line 8:
```suggestion
type CallbackFunction = (...args: Param) => void | Promise<void>;
```

Simplification suggestion.

ui/raidboss/emulator/overrides/RaidEmulatorPopupText.ts
```diff
@@ -0,0 +1,248 @@
...
+    this.scheduledTriggers = this.scheduledTriggers.filter((t) => {
+      return t.expires - currentLogTime > 0;
+    });
+    this.displayedText = this.displayedText.filter((t) => {
+      const remaining = t.expires - currentLogTime;
+      if (remaining > 0) {
+        const elem = t.element.querySelector('.popup-text-remaining');
+        if (!elem)
+          throw new UnreachableCode();
+        elem.textContent = '(' + (remaining / 1000).toFixed(1) + ')';
```
quisquous, line 95:
```suggestion
        elem.textContent = `(${remaining / 1000).toFixed(1)})`;
```

ui/raidboss/emulator/overrides/RaidEmulatorPopupText.ts
```diff
@@ -0,0 +1,248 @@
...
+
+      const enc = currentEnc.encounter;
+      if (!enc || !enc.encounterZoneName || enc.encounterZoneId === undefined)
+        return;
+
+      this._emulatorReset();
+      this.OnChangeZone({
+        type: 'ChangeZone',
+        zoneName: enc.encounterZoneName ?? 'Unknown',
+        zoneID: parseInt(enc.encounterZoneId ?? '', 16),
```
quisquous, line 165:
```suggestion
        zoneName: enc.encounterZoneName,
        zoneID: parseInt(enc.encounterZoneId, 16),
```

ui/raidboss/emulator/overrides/RaidEmulatorPopupText.ts
```diff
@@ -0,0 +1,248 @@
...
+    container.textContent = text;
+    return $ret;
+  }
+
+  addDisplayText($e: HTMLElement, endTimestamp: number): void {
+    const remaining = (endTimestamp - this.emulatedOffset) / 1000;
+    const container = $e.querySelector('.popup-text-remaining');
+    if (!(container instanceof HTMLElement))
+      throw new UnreachableCode();
+    container.textContent = '(' + remaining.toFixed(1) + ')';
```
quisquous, line 241:
```suggestion
    container.textContent = `(${remaining.toFixed(1)})`;
```

ui/raidboss/emulator/EventBus.ts
valarnin: I wasn't sure where `|` operator priority was for type defs here, e.g. which of the following that evaluates to:
```ts
type CallbackFunction = (...args: Param) => (void | Promise<void>);
// or
type CallbackFunction = ((...args: Param) => void) | (Promise<void>);
```

If it's the former, maybe for clarity it makes sense to wrap the return types in parenthesis as in my example?

valarnin: Changed to this:
```ts
type CallbackFunction = (...args: Param) => (void | Promise<void>);
```

quisquous: I am pretty sure that `=>` binds more strongly than `|`.  If you do something like `() => void | () => boolean` tsc will complain at you with `Function type notation must be parenthesized when used in a union type.(1385)` because it looks like it's `() => void | boolean` but is really `() => (void | (() => boolean))`.

PR #3143, opened 2021-06-20
raidemulator: Fix async issue, fix tts still actually triggering tts during analysis

valarnin:
This PR fixes an async issue caused by #3115 (basically, actual cactbot user options weren't being applied).

As well this PR fixes TTS audibly still happening during encounter analysis, which caused my ears to bleed when the entire e12s p2 fight TTS'd for all 8 players at the same time a few minutes ago.


valarnin:
Oh, it's also still TTSing when seeking, let me see if I can fix that too.

valarnin:
The TTS-while-seeking issue should be fixed with the typescript conversion, actually.

PR #3144, opened 2021-06-20
raidboss: Fix tts issues re: #3136

### LINKED ISSUE #3136: E12S Double Aero Finder still has tts
is the purpose of this `tts: null` to make it silent?

This still get tts.

https://github.com/quisquous/cactbot/blob/961a1ba7321f96d5d475b449b75b016a3d0c8b8e/ui/raidboss/data/05-shb/raid/e12s.js#L1607-L1616

I guess the reason is here, we didn't distinguish `null` and `undefined`

https://github.com/quisquous/cactbot/blob/ed708582d715cf304e354b2631f9766b175f12ae/ui/raidboss/popup-text.ts#L1183-L1190

valarnin:
Fixes #3136

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -1184,9 +1184,7 @@ export class PopupText {
       let result = undefined;
       if (triggerHelper.triggerOptions.TTSText) {
         result = triggerHelper.valueOrFunction(triggerHelper.triggerOptions.TTSText);
-      } else if (triggerHelper.trigger.tts ||
-          triggerHelper.trigger.tts === null ||
-          triggerHelper.trigger.tts === '') {
+      } else if (triggerHelper.trigger.tts !== undefined) {
         result = triggerHelper.valueOrFunction(triggerHelper.trigger.tts);
       } else if (triggerHelper.response) {
         const resp: ResponseField<RaidbossData> = triggerHelper.response;
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -1185,6 +1185,7 @@ export class PopupText {
       if (triggerHelper.triggerOptions.TTSText) {
         result = triggerHelper.valueOrFunction(triggerHelper.triggerOptions.TTSText);
       } else if (triggerHelper.trigger.tts !== undefined) {
+        // Allow null/false/NaN/0/'' in this branch.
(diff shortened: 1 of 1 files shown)


valarnin:
The comment below this change indicates that `false` should disable as well, but doesn't mention `''`. Should I update both the `else if` clause and the comment to include both?

ui/raidboss/popup-text.ts
```diff
@@ -1184,7 +1184,9 @@ export class PopupText {
       let result = undefined;
       if (triggerHelper.triggerOptions.TTSText) {
         result = triggerHelper.valueOrFunction(triggerHelper.triggerOptions.TTSText);
-      } else if (triggerHelper.trigger.tts) {
+      } else if (triggerHelper.trigger.tts ||
+          triggerHelper.trigger.tts === null ||
+          triggerHelper.trigger.tts === '') {
```
quisquous, line 1189:
```suggestion
      } else if (triggerHelper.trigger.tts !== undefined) {
        // Allow null or false in this branch.
```

What about this to allow for null, false, and empty string?

valarnin: Would also allow `NaN` and `0`, if both of those are fine, then sure?

quisquous: I can go either way.  I think personally returning either of those here is sort of "undefined behavior" and it's fine if they do weird things.  If you think it's better to be explicit and ignore it, that seems fine too?

valarnin: Updated, with some more clear comments. Sorry for the double commit, I forgot to save the comments before I staged/commited

ui/raidboss/popup-text.ts
```diff
@@ -1184,15 +1184,16 @@ export class PopupText {
...
+        // Allow null/false/NaN/0/'' in this branch.
         result = triggerHelper.valueOrFunction(triggerHelper.trigger.tts);
       } else if (triggerHelper.response) {
         const resp: ResponseField<RaidbossData> = triggerHelper.response;
         if (resp.tts)
           result = triggerHelper.valueOrFunction(resp.tts);
       }
 
-      // Allow false or null to disable tts entirely
+      // Allow falsey values to disable tts entirely
```
quisquous, line 1196: Thanks for the extra comment.

PR #3145, opened 2021-06-20
raidemulator: Convert NetworkLogConverterWorker to typescript

valarnin:
Also removed the Map files in this PR, no point in converting them to TS.


valarnin:
On a related note, I looked into RaidEmulatorOverlayApiHook and it's going to require some additional work to convert, I might need to do a pass over the entire `overlay_plugin_api.ts` file and rework how it types things to let the emulator hook exist in typescript without a ton of eslint rule disabling.

That'll be what I do next, then typescript conversion for that file, then Persistor which is going to be a bit of a lift.

PR #3146, opened 2021-06-20
config: add link to coverage report

quisquous:
Requested by @MyTechnoHunter because this link is hard to find.

Also add support for html "inputs".  Maybe this is a case where we need
some other way to add more text at the beginning of a section, but we
can figure that out if we get more examples.

I left this as the raw link text to make it super clear that it was
a link to click on, and what that link was so you could find it outside
of ACT.

It would be really nice if OverlayPlugin supported something like the
`cactbotChooseDirectory` overlay handler for opening links in a browser
outside of ACT.  I would use it for this and view source if it existed.

### DIFF
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -991,7 +991,7 @@ const templateOptions = {
     {
       id: 'Coverage',
       name: {
-        en: 'Supported content',
+        en: 'Supported content (latest version)',
       },
       type: 'html',
       html: {
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -362,6 +362,10 @@ class DoNothingFuncProxy {
   }
 }
 
+const makeLink = (href) => {
+  return `<a href="${href}" target="_blank">${href}</a>`;
+};
+
 class RaidbossConfigurator {
   constructor(cactbotConfigurator) {
     this.base = cactbotConfigurator;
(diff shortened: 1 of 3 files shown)


quisquous:
Looks like this: 
image

MaikoTan:
I think if someone using a elder version of cactbot, the content listed in the webpage is not compatible with their local version. Maybe something like `Supported Contents Currently` or `Supported Contents in Latest Version` make sense?

quisquous:
> Maybe something like `Supported Contents Currently` or `Supported Contents in Latest Version` make sense?

Done, thanks!  I always forget that.

ui/raidboss/raidboss_config.js
```diff
@@ -984,6 +988,22 @@ const templateOptions = {
...
   options: [
+    {
+      id: 'Coverage',
+      name: {
+        en: 'Supported content (latest version)',
+      },
+      type: 'html',
+      html: {
+        // TODO: it'd be nice if OverlayPlugin could open links on the system outside of ACT.
+        en: makeLink('https://quisquous.github.io/cactbot/util/coverage/coverage.html?lang=en'),
```
MaikoTan, line 999: I think it would be better to show link text as "Open link in new window" instead of an URL?

quisquous: The problem is that when you open a link in ACT, there is no url bar.  I thought it was better to show the url in case you wanted to open it in a browser or copy it or know where the link was.

MaikoTan: Oh, that make sense.
So, only if OverlayPlugin supports "Open in system browser" would solve the problem?
I can recall that you can run shell command like "start <url>" to invoke system browser, is it possible to be built in cactbot/OverlayPlugin?

quisquous: I'm not sure about what the best way to invoke the system browser is.  But, I do think that approach makes the most sense to me.  Add an overlay handler to open a system browser to OverlayPlugin and then switch cactbot to use that.

PR #3147, opened 2021-06-20
jobs: turn on notifications by default

quisquous:
I think more people want this than not, and those that don't can turn it
off in the options.

### DIFF
--- a/ui/jobs/jobs_config.ts
+++ b/ui/jobs/jobs_config.ts
@@ -238,7 +238,7 @@ UserConfig.registerOptions('jobs', {
           'Wenn der Countdown nahe 0 ist.': 'threshold',
         },
       },
-      default: 'expired',
+      default: 'threshold',
     },
   ],
 });
--- a/ui/jobs/jobs_config.ts
+++ b/ui/jobs/jobs_config.ts
@@ -217,7 +217,7 @@ UserConfig.registerOptions('jobs', {
         de: 'Dot/etc. boxen blinken bis zu n mal wenn im Kampf und dot ist nicht aktiv. (-1: deaktiviert, 0: ohne Limit)',
       },
       type: 'integer',
-      default: -1,
+      default: 5,
     },
     {
       id: 'NotifyExpiredProcsInCombatSound',
@@ -238,7 +238,7 @@ UserConfig.registerOptions('jobs', {
           'Wenn der Countdown nahe 0 ist.': 'threshold',
         },
       },
-      default: 'disabled',
+      default: 'expired',
     },
   ],
 });


quisquous:
@rcj (and @Echoring) do you have opinions on this? Mostly I see people asking for sounds for `jobs` all the time and so I think being slightly flashier/noisier by default isn't that bad since people can turn this off.

Echoring:
In my thought this is fine. Even someone has asked for tts output in jobs module.

rcj:
I don't mind.

Regarding TTS, I have intentions to add that to Jobs. I'm currently thinking about how to best expose this in the configuration UI as the user probably wants to customize the TTS output per skill. Same might go for the threshold. So I'm wondering if some kind of UI where you can configure per skill how it should behave (show a timerbox, with sound, with TTS, threshold, etc) would be beneficial.

Maybe initially we could have individual sections in the configuration UI per job to group job specific configuration. 

Ideally the Jobs overlay would be way more configurable. What I would like to do eventually is to be able to place the different UI elements individually at different locations (drag/drop them to where you want them to be), so you could add n containers which could contain m timerbox or other objects which could be placed anywhere on the screen.

ui/jobs/jobs_config.ts
```diff
@@ -238,7 +238,7 @@ UserConfig.registerOptions('jobs', {
           'Wenn der Countdown nahe 0 ist.': 'threshold',
         },
       },
-      default: 'disabled',
+      default: 'expired',
```
rcj, line 241: Should the default be 'threshold' instead? Normally you want to reapply e.g. a dot before it falls off.

Echoring:
> Regarding TTS, I have intentions to add that to Jobs. I'm currently thinking about how to best expose this in the configuration UI as the user probably wants to customize the TTS output per skill. Same might go for the threshold. So I'm wondering if some kind of UI where you can configure per skill how it should behave (show a timerbox, with sound, with TTS, threshold, etc) would be beneficial.
> 
> Maybe initially we could have individual sections in the configuration UI per job to group job specific configuration.
> 
> Ideally the Jobs overlay would be way more configurable. What I would like to do eventually is to be able to place the different UI elements individually at different locations (drag/drop them to where you want them to be), so you could add n containers which could contain m timerbox or other objects which could be placed anywhere on the screen.

Sounds great! In fact I'm also considering about allowing user custom setting for elements in jobs module, but your destination seems far better.

ui/jobs/jobs_config.ts
quisquous: Good idea, thanks!

PR #3149, opened 2021-06-21
util: Rework overlay API types to make backing types more accessible

valarnin:
This PR reworks the way that the overlay API event types are typed out to make request/response type/response promise type more accessible.

I tried to get the `callOverlayHandler` function to properly auto-type the parameter and return value but for some reason it wasn't working, I'm not sure why, so I updated `user_config.ts` and `util.ts` calls to `callOverlayHandler` to pass the call type in.

This rework is required to convert RaidEmulatorOverlayApiHook to typescript.

Verified raidemulator and basic raidboss functionality in-game (general triggers) but didn't have time to run any content tonight.

### DIFF
--- a/types/event.d.ts
+++ b/types/event.d.ts
@@ -351,51 +351,90 @@ export interface PluginCombatantState {
   Heading: number;
 }
 
-export type GetCombatantsCall = {
+type SubscribeHandler = (msg: {
+  call: 'subscribe';
+  events: string[];
+}) => void;
+
+type GetCombatantsHandler = (msg: {
   call: 'getCombatants';
   ids?: number[];
   names?: string[];
   props?: string[];
+}) => { combatants: PluginCombatantState[] };
+
+type CactbotReloadOverlaysHandler = (msg: {
+  call: 'cactbotReloadOverlays';
+}) => void;
+
+type CactbotLoadUserHandler = (msg: {
+  call: 'cactbotLoadUser';
+  source: string;
+  overlayName: string;
+}) => { detail: CactbotLoadUserRet };
+
+type CactbotRequestPlayerUpdateHandler = (msg: {
+  call: 'cactbotRequestPlayerUpdate';
+}) => void;
+
+type CactbotRequestStateHandler = (msg: {
+  call: 'cactbotRequestState';
+}) => void;
+
+type CactbotSayHandler = (msg: {
+  call: 'cactbotSay';
+  text: string;
+}) => void;
+
+type CactbotSaveDataHandler = (msg: {
+  call: 'cactbotSaveData';
+}) => void;
+
+type CactbotLoadDataHandler = (msg: {
+  call: 'cactbotLoadData';
+  overlay: string;
+}) => ({ data: SavedConfig } | undefined);
+
+type CactbotChooseDirectoryHandler = <T>(msg: {
+  call: 'cactbotChooseDirectory';
+}) => ({ data: T } | undefined);
+
+export type OverlayHandlerAll = {
+  'subscribe': SubscribeHandler;
+  'getCombatants': GetCombatantsHandler;
+  'cactbotReloadOverlays': CactbotReloadOverlaysHandler;
+  'cactbotLoadUser': CactbotLoadUserHandler;
+  'cactbotRequestPlayerUpdate': CactbotRequestPlayerUpdateHandler;
+  'cactbotRequestState': CactbotRequestStateHandler;
+  'cactbotSay': CactbotSayHandler;
+  'cactbotSaveData': CactbotSaveDataHandler;
+  'cactbotLoadData': CactbotLoadDataHandler;
+  'cactbotChooseDirectory': CactbotChooseDirectoryHandler;
 };
 
-export type GetCombatantsRet = { combatants: PluginCombatantState[] };
+export type OverlayHandlerTypes = keyof OverlayHandlerAll;
 
-export type IOverlayHandler = {
-  // OutputPlugin build-in
-  (msg: {
-    call: 'subscribe';
-    events: string[];
-  }): Promise<null>;
-  (msg: GetCombatantsCall): Promise<GetCombatantsRet>;
-  // TODO: add OverlayPlugin build-in handlers
-  // Cactbot
-  // TODO: fill up all handler types
-  (msg: {
-    call: 'cactbotReloadOverlays';
-  }): Promise<null>;
-  (msg: {
-    call: 'cactbotLoadUser';
-    source: string;
-    overlayName: string;
-  }): Promise<{ detail: CactbotLoadUserRet }>;
-  (msg: {
-    call: 'cactbotRequestPlayerUpdate';
-  }): Promise<null>;
-  (msg: {
-    call: 'cactbotRequestState';
-  }): Promise<null>;
-  (msg: {
-    call: 'cactbotSay';
-    text: string;
-  }): Promise<null>;
-  (msg: {
-    call: 'cactbotSaveData';
-  }): Promise<null>;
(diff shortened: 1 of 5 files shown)


types/event.d.ts
```diff
@@ -351,51 +351,90 @@ export interface PluginCombatantState {
...
+export type OverlayHandlerAnyResponse = OverlayHandlerResponses[OverlayHandlerTypes];
+
+export type OverlayHandlerFuncs = {
+  [call in OverlayHandlerTypes]:
+      (msg: Parameters<OverlayHandlerAll[call]>[0]) => OverlayHandlerResponses[call];
+};
+
+export type IOverlayHandler =
+    <T extends OverlayHandlerTypes>(msg: OverlayHandlerRequests[T]) =>
+    Promise<OverlayHandlerResponseTypes[T]>;
```
quisquous, line 440: It's not the prettiest, but I think you can say:
```typescript
export type IOverlayHandler =
  OverlayHandlerFuncs['subscribe'] &
  OverlayHandlerFuncs['getCombatants'] &
  OverlayHandlerFuncs['cactbotReloadOverlays'] &
  OverlayHandlerFuncs['cactbotLoadUser'] &
  OverlayHandlerFuncs['cactbotRequestPlayerUpdate'] &
  OverlayHandlerFuncs['cactbotRequestState'] &
  OverlayHandlerFuncs['cactbotSay'] &
  OverlayHandlerFuncs['cactbotSaveData'] &
  OverlayHandlerFuncs['cactbotLoadData'] &
  OverlayHandlerFuncs['cactbotChooseDirectory'];
```

TypeScript has a note about using infer on things in contravariant positions (aka function parameters?) to get intersection behavior but I couldn't quite get it to work with a generic + infers.

valarnin: This whole thing only works right now due to this casting mess in `overlay_plugin_api.ts` which I'd really like to get rid of, but as it stands now without inference working I'll have to do this same mess of casting in RaidEmulatorOverlayApiHook as well:

```ts
export const callOverlayHandler: IOverlayHandler = (
    _msg: { [s: string]: unknown },
    // eslint-disable-next-line @typescript-eslint/no-explicit-any
): Promise<any> => {
  init();
  if (callOverlayHandlerOverride) {
    return callOverlayHandlerOverride(
      _msg as Parameters<IOverlayHandler>[0],
    ) as Promise<unknown>;
  }
  return callOverlayHandlerInternal(_msg as Parameters<IOverlayHandler>[0]);
};
```

Neither the PR as it exists or your suggestion works to fix this issue. I'll poke around at it again today.

valarnin: After some more poking around at it, I decided that the best idea was to keep the terrible syntax constrained to just the `overlay_plugin_api.ts` file. So I reworked the way overrides work so that casting isn't needed elsewhere to override a handler.

Ultimately I think the problem comes down to Promises, for some reason they really don't play well together and I can't figure out why.  For reference, see the `setCallOverlayHandlerOverride` function's casting of override.

valarnin: If you're not attached to the way the current `callOverlayHandler` function works, I can probably make this work globally without weird casting and such, but I'd have to change the function signature significantly to do so I believe.

quisquous: Yeah, I'm not attached to it.  I think the way it is is a conversion of the way it was.  If there's a way that it can be less weird for TypeScript, that seems good to me.

Personally, I think it makes sense for you (and everyone) to have spent a bunch of time getting existing triggers typed properly, because we have a lot of them and changing trigger format seems like a big thing (for many reasons).  However, in comparison I think we should be more skeptical about whether existing forms of other code are correct and if there's some way to rewrite them to be better for TypeScript (or in general) if we run into having to do contortions to get them to work.

valarnin: My proposal here then is basically:

1. Leave the existing code in place but don't export it, instead just assign it to `window.callOverlayHandler` for backwards compatability
2. Set up a new overlay handler calling API

quisquous: Seems reasonable to me.

It does make me wonder if we need a process for deprecating stuff people are using in user triggers though.  I'm not sure what that would look like or how we could even really detect it.  (Add feature for raising ACT notifications => grep user files => notify once per release of deprecated code??? 😬 )

valarnin: I think if that's an option, maybe it makes sense to leave entire current setup in `main` in place, but mark it as deprecated. Move all the cactbot internal code over to a new system, and we can have the old handlers raise the warnings in ACT directly via a callback, rather than having to try to scan user file code?

quisquous: The suggestion to scan is mostly a difference of checking at load time vs run time.  If we only raise warnings at runtime, then a user would have to use that trigger to get a warning.  What if they have something for like the current tier and they don't go back and do anything about it until a year from now when they redo it on blu?

valarnin: What are the odds that someone sees that error and actually fixes the problem immediately when they aren't actively running that content?

valarnin: I'm just worried about trying to scan user triggers for something like that and missing it. I'd rather a concrete way to warn as well if possible.

valarnin:
Closing this in favor of the option discussed in the comments.

types/event.d.ts
quisquous: I'm sorry to reopen this, but I am still confused what is wrong with https://github.com/quisquous/cactbot/commit/7244d17e34e93456bc43e29a1bbb45727569952b.  It seems like all of the types are inferred correctly here.  Is there an additional issue that I'm missing?

valarnin: On mobile, sorry if names are wrong.

The problem comes when trying to use the override in `RaidEmulatorOverrideAPIHook` when it's converted to typescript. Basically it doesn't like the function parameter and return types due to promises for some reason. I tried for a few hours to get it to work and it flat out wouldn't without a cast to any and back for some reason.

I can try again tomorrow with the changes in the commit you referenced and see if that fixes it, but I'm not sure if it will, since it has to do with the override function/inference I think.

quisquous: Yeah, I'd be curious to see.  Please link a commit with what's broken so I can see it.

Mostly your other patch feels like a lot of change, and I wanted to understand better what was wrong with the previous approach.

valarnin: https://github.com/valarnin/cactbot/commit/a280986612b3baec25e51a94c0a2ca77942cc3af

Can you get the `async call` function to work without a bunch of typecasting/without `Promise<any>`?

image
image

quisquous: Here's two thoughts:
(1) https://github.com/quisquous/cactbot/commit/3d32863c09ddcf32b8eda189d984a53c22345af8
(2) https://github.com/quisquous/cactbot/commit/7b5ae646d296756d25c788b7326655edca67c56f

I'm sorry for the pushback here and sorry I said "oh yeah, a new api is fine" only to see it and think it felt like a lot.  Really I just wanted to understand for myself what the issue was here so I knew what the tradeoff was to handle a new api.

valarnin: I'd be more in favor of (2) just because if there's another handler we need to override in the future I'd rather not have to revisit it again.

quisquous: If that sounds good to you, then let's go with the more minimal change.  Do you want to put that up as a PR? Should I?

valarnin: If you've got time go ahead, otherwise I won't be able to get to it until tomorrow after noon

quisquous: Put it up in #3258.  I took a trip to your wizard library and found an arcane book that let me generalize the incantation too.  :)

PR #3150, opened 2021-06-21
Feature(Raidboss): show default value instead of (default)

Echoring:
Sometimes I find a trigger not show long enough, but I turn to config only to find it `(default)`.
What is default value? I have to view source code to know.
So this feature enable user to know actual default value without source code.

### DIFF
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -554,7 +554,7 @@ class RaidbossConfigurator {
           div.appendChild(input);
           input.type = 'text';
           input.step = 'any';
-          input.placeholder = trig.durationSeconds || 3;
+          input.placeholder = `${trig.durationSeconds || 3}${this.base.translate(kMiscTranslations.valueDefault)}`;
           input.value = this.base.getOption('raidboss', 'triggers', trig.id, optionKey, '');
           const setFunc = () => {
             const val = validDurationOrUndefined(input.value) || '';
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -554,7 +554,7 @@ class RaidbossConfigurator {
           div.appendChild(input);
           input.type = 'text';
           input.step = 'any';
-          input.placeholder = `${trig.durationSeconds || 3} ${this.base.translate(kMiscTranslations.valueDefault)}`;
+          input.placeholder = `${trig.durationSeconds || ''} ${this.base.translate(kMiscTranslations.valueDefault)}`;
           input.value = this.base.getOption('raidboss', 'triggers', trig.id, optionKey, '');
           const setFunc = () => {
             const val = validDurationOrUndefined(input.value) || '';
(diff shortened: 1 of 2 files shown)


ui/raidboss/raidboss_config.js
```diff
@@ -554,7 +554,7 @@ class RaidbossConfigurator {
           div.appendChild(input);
           input.type = 'text';
           input.step = 'any';
-          input.placeholder = this.base.translate(kMiscTranslations.valueDefault);
+          input.placeholder = `${trig.durationSeconds || 3}${this.base.translate(kMiscTranslations.valueDefault)}`;
```
MaikoTan, line 557:
```suggestion
          input.placeholder = `${trig.durationSeconds || 3} ${this.base.translate(kMiscTranslations.valueDefault)}`;
```

Add a space would be more clear.

quisquous, line 557: There's a couple of reasons that I said default (mostly because it's easier and I was being lazy <_<).
* `durationSeconds` can be a function (that could depend on `data` and `matches`), and I'm not sure what you output for that
* default trigger duration can be set differently for alarm/alert/info, and so it's not necessarily 3 and may depend on what the trigger outputs.  (What if it outputs both info and alert and you have set those to different duration lengths?)

You could say:

if the trigger has no `durationSeconds` property and all of the alarm/alert/info options are set to the same value, then use that value.  Otherwise, if `durationSeconds` is a number, display that number.  Otherwise, display default.  This would cover many cases, but still avoid the trickier ones.

Echoring: Oh sorry I don't know default duration can be changed.
I think only show number for non default duration will be fine?

quisquous: Yeah, I think showing `trigger.durationSeconds` if `typeof trigger.durationSeconds === 'number'` sounds good to me.  Is that what you're saying?

Echoring: Seems checking with `typeof` is better here. Done.

ui/raidboss/raidboss_config.js
```diff
@@ -554,7 +554,10 @@ class RaidbossConfigurator {
           div.appendChild(input);
           input.type = 'text';
           input.step = 'any';
-          input.placeholder = this.base.translate(kMiscTranslations.valueDefault);
+          if (typeof trig.durationSeconds === 'number')
+            input.placeholder = `${trig.durationSeconds} ${this.base.translate(kMiscTranslations.valueDefault)}`;
```
quisquous, line 558:
```suggestion
            input.placeholder = `${trig.durationSeconds}`;
```

What about just showing the number? The placeholder should be a different color.

Echoring: Oh, this is @MaikoTan suggests before. I think a single number only is a little empty in box.

quisquous: I think it looks ok (although the CSS change makes it look emptier)?

image

Echoring: Fine. This css change is actually for de, for I find that default in de is a little longer to show.

PR #3152, opened 2021-06-21
i18n(eureka, jobs): Fix multi-language support on eureka, small translation for jobs Korean

Jaehyuk-Lee:
This PR fixes something that is almost a bug.

### DIFF
--- a/ui/jobs/jobs_config.ts
+++ b/ui/jobs/jobs_config.ts
@@ -217,7 +217,7 @@ UserConfig.registerOptions('jobs', {
         de: 'Dot/etc. boxen blinken bis zu n mal wenn im Kampf und dot ist nicht aktiv. (-1: deaktiviert, 0: ohne Limit)',
         ja: '戦闘中でDoT/バフが切ったらprocボックスをｎ回点滅させる(-1：無効、0：無限回数)',
         cn: '战斗中模块监控的重要DoT/Buff中断时令对应计时器闪烁N次（-1：禁用，0：无限闪烁）',
-        ko: '도트가 꺼지면 프록 박스를 n번 깜빡하게 합니다. (-1: 비활성화, 0: 무한)',
+        ko: '도트나 버프가 꺼지면 프록 박스를 n번 깜빡하게 합니다. (-1: 비활성화, 0: 무한)',
       },
       type: 'integer',
       default: 5,
@@ -229,7 +229,7 @@ UserConfig.registerOptions('jobs', {
         de: 'Spiele einen Alarm Sound wenn eine dot/etc. box im Kampf inaktiv wird.',
         ja: '戦闘中でDoT/バフが切ったら音を鳴らす',
         cn: '战斗中模块监控的重要DoT/Buff中断时播放提示音',
-        ko: '도트가 꺼지면 소리로 알림을 줍니다.',
+        ko: '도트나 버프가 꺼지면 소리로 알림을 줍니다.',
       },
       type: 'select',
       options: {
--- a/ui/jobs/jobs_config.ts
+++ b/ui/jobs/jobs_config.ts
@@ -215,6 +215,7 @@ UserConfig.registerOptions('jobs', {
       name: {
         en: 'Flash procs boxes of inactive dots/etc. up to n times while in combat. (-1: disabled, 0: infinite)',
         de: 'Dot/etc. boxen blinken bis zu n mal wenn im Kampf und dot ist nicht aktiv. (-1: deaktiviert, 0: ohne Limit)',
+        ko: '도트가 꺼지면 프록 박스를 n회 깜빡하게 합니다. (-1: 비활성화, 0: 무한)',
       },
       type: 'integer',
       default: -1,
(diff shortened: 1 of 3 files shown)


Jaehyuk-Lee:
Found more texts to translate.

Jaehyuk-Lee:
Field Note data-mined from [here English](https://github.com/xivapi/ffxiv-datamining/blob/master/csv/MYCWarResultNotebook.csv) and [Korean](https://github.com/Ra-Workspace/ffxiv-datamining-ko/blob/04cc10c9d8794c38f2e301a7570ea19ec1c43437/csv/MYCWarResultNotebook.csv).

Jaehyuk-Lee:
Korean doesn't need `shortLabel`, every full names are not too long in Korean. But, the option `this.options.CompleteNamesSTQ` is default `false`, which makes default to show "shortLabel" version. So, it's showing English names for default. People would never think `CompleteNamesSTQ` `false` option makes to show English instead of Korean or Japanese, etc except French.

* "Prefer complete names for Skirmishes/Critical Engagements" false (which is default) -> change to true
Honeycam 2021-06-21 20-36-23
when false, shows English (because there's no Korean data)
when true, shows Korean (because there's Korean data)

The ba9c73635f793dadecd2440a97b46f09bc3bc9dd commit seems unfriendly for multi-language support. 
`shortLabel` part problem is associated with `label`, so it could be not actually made sure that it supports multi-language well. (mostly it is okay with just using English if it's empty in other languages) I would temporarily add `shortLabel` for Korean, but I think that's waste of resources. I tried to use `label` instead of `shortLabel` when it's empty, but the translation is done with `TransByDispLang` function, and adding translation exception inside of function seems bad. Then, I didn't add an exception there.
This problem is for every other languages, I've just copy and paste the `label` data to `shortLabel` in ab5fa112491bfaf1f61fc822f4bc36ddae7efeae, since I don't know German, Japanese, Chinese.

ui/jobs/jobs_config.ts
```diff
@@ -237,6 +239,11 @@ UserConfig.registerOptions('jobs', {
           'Wenn der Countdown 0 erreicht.': 'expired',
           'Wenn der Countdown nahe 0 ist.': 'threshold',
         },
+        ko: {
+          '비활성화': 'disabled',
+          '카운트 다운이 0초일 때': 'expired',
+          '카운트 다운이 0초에 가까워 질때': 'threshold',
```
MaikoTan, line 245: I think "0초에 가깝다" is not very clear in this sentence, maybe translate it as "when it reaches the threshold(閾値)" just like [me and Echoring do](https://github.com/quisquous/cactbot/pull/3151/files) would be more accurate?

quisquous:
Sorry about not thinking about translations on the field notes.  Thanks for adding that.

> The ba9c736 commit seems unfriendly for multi-language support.

Ack, my apologies on this too.  I agree that pasting duplicates is really not great, but without listing `ko:` in the `shortLabel` section, the `find-translations` script will point out that it's missing too unless we somehow modify that to know that these are partial.  So, it's a little unfortunate either way.

What about only copying the languages that you know are ok as "short" into `shortLabel`, and then changing this code so that missing short labels don't turn into English?

```javascript
    // Short labels only exist for Save-The-Queen content
    // Changes names' length depending on users options
    // If no strings are available, the english short ones will be the default ones
    if (this.zoneInfo.treatNMsAsSkirmishes) {
      if (this.options.CompleteNamesSTQ)
        name.innerText = this.TransByDispLang(nm.label);
      if (!name.innerText)
        name.innerText = this.TransByDispLang(nm.shortLabel);
      // If the short label is not set, fall back to the full label.
      if (!name.innerText)
        name.innerText = this.TransByDispLang(nm.label);
    } else {
      name.innerText = this.TransByDispLang(nm.label);
    }
```

ui/jobs/jobs_config.ts
Jaehyuk-Lee: I think threshold value would change by each jobs? I thought if there's no exact time written, it would rather better write ambiguous word. People may try to find what exact threshold is, from cactbot readme or somewhere.

Jaehyuk-Lee:
@quisquous The code you wrote doesn't work, because from `name.innerText = this.TransByDispLang(nm.shortLabel);`, `name.innerText` is always set. If `nm.shortLabel` is unset in Korean, `TransByDispLang` function uses English version. So, `name.innerText` is set English even though there's no Korean data.

ui/jobs/jobs_config.ts
MaikoTan: Yeah, that's true.
Then I thought maybe "when the counter is almost ending" would be another choice?
FYI, which is called "終了間際" in Japanese or "即將結束" in Chinese.

Jaehyuk-Lee: "when the counter is almost ending" is better I agree, I'll see if there's better one.

Echoring: By the way, some thresholds may be a little long...

Jaehyuk-Lee: I understood, I have Korean translation in mind.

Jaehyuk-Lee:
used `this.options.DisplayLanguage` directly then checked if it's `undefined`. This works well without shortlabel translation.

quisquous:
> used `this.options.DisplayLanguage` directly then checked if it's `undefined`. This works well without shortlabel translation.
> 
> see [dfb7ea1](https://github.com/quisquous/cactbot/commit/dfb7ea1cea56765f5a37949faee0d7986456f3c7), but `this.DisplayLanguage` is mistake, it was fixed later.

Sorry about the bad suggestion.  What you wrote was what I had intended, thanks.

PR #3154, opened 2021-06-21
feat(util): add interactive command line interface

MaikoTan:
timelineTranslate

As https://github.com/quisquous/cactbot/pull/2603#issuecomment-802670848
Make a bunch of scripts such as `gen_*.ts` (in the future) and `translate_timeline.ts`, etc in `util` directory into a single entry.


util/index.ts
```diff
@@ -0,0 +1,107 @@
...
+    type: 'list',
+    name: 'action',
+    message: 'What do you want to do?',
+    choices: [
+      'Generate common data files',
+      'Translate Raidboss timeline',
+      'Find Raidboss missing translations',
+    ],
+  }]).then((answer: Answers) => {
+    if (answer.action === 'Generate common data files')
```
quisquous, line 33: What about using choice objects, where instead of `'Generate common data files'` (and then repeating this entire English string, which may be updated), you use `{ name: 'Generate common data files', key: 'generate' }`.  I expect also these strings would get translated at some point too, so having them a little bit more abstract will help will that.

util/index.ts
```diff
@@ -0,0 +1,107 @@
...
+
+declare module 'inquirer' {
+  interface QuestionMap<T> {
+    fuzzypath: FuzzyPathQuestionOptions<T>;
+  }
+}
+
+const dataFilesMap = {
+  'effect_id.ts': generateEffectIds,
+} as const;
```
quisquous, line 18: If you're not going to use the keys as static types, what about this, so you don't need the `answers.choice as keyof typeof dataFilesMap` below?

```suggestion
const dataFilesMap: { readonly [filename: string]: Promise<void> } = {
  'effect_id.ts': generateEffectIds,
};
```

util/index.ts
```diff
@@ -0,0 +1,107 @@
...
+        'de',
+        'fr',
+        'ja',
+        'cn',
+        'ko',
+      ],
+    },
+  ]).then((answers: Answers) => {
+    if (answers.timeline && answers.locale)
+      return translateTimeline(answers as { timeline: string; locale: Lang });
```
quisquous, line 76:
```suggestion
    const timeline = answers.timeline;
    const locale = answers.locale;
    if (timeline && locale)
      return translateTimeline({ timeline: timeline; locale: locale });
```

Alternatively, just call `return translateTimeline(answers);` and have the `translateTimeline` function handle missing params?

quisquous (commented):
I think my biggest concern here is that I don't want to remove being able to call these non-interactively.  I think interactive helpers are great for one-off/infrequent usages, but bad for repeated/scripted usages.

I think as long as the original scripts are still able to be called from the command line OR you make it possible to call this script with parameters (e.g. `npm run util -- --generate`) or something, I think I'll be good.  I just don't want to lose that use case to make other things more convenient.

util/find_missing_translations.js
```diff
@@ -160,5 +160,7 @@ const run = async (args) => {
   }
 };
 
-const args = parser.parseArgs();
-run(args);
+if (fileURLToPath(import.meta.url) === process.argv[1]) {
```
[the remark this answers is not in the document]
MaikoTan: I think if there are npm scripts like `npm run util -- --generate`, then it is better to have `"find-translations": "npm run util -- --generate",` in `package.json`'s `scripts` section, as we don't need to run the file directly.

util/index.ts
```diff
@@ -0,0 +1,107 @@
...
+      type: 'fuzzypath',
+      name: 'timeline',
+      message: 'Input a valid timeline filename: ',
+      rootPath: 'ui/raidboss/data',
+    },
+    {
+      type: 'list',
+      name: 'locale',
+      message: 'Select a locale: ',
+      choices: [
```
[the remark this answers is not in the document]
MaikoTan: Yes, thanks

util/find_missing_translations.js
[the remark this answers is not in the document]
quisquous: That sounds good to me.  I just want a way to run things from the command line non-interactively still, even if they all go through util.

MaikoTan:
There are still few type errors, I would try to fix it soon. By the way, are there some documents about `argparse` v1.x? I went to the npm page but it redirect to the argparse python version, and clearly not consistent with the version we are using.

quisquous:
> There are still few type errors, I would try to fix it soon. By the way, are there some documents about `argparse` v1.x? I went to the npm page but it redirect to the argparse python version, and clearly not consistent with the version we are using.

I think it is meant to be exactly the same as the Python version, so there's not separate documentation for it. 
 https://github.com/nodeca/argparse/blob/70fc26eb5a9a829d6ddcceb6b4c87802218227fd/doc/port_difference.md tries to explain the differences between the two.

util/index.ts
```diff
@@ -0,0 +1,188 @@
...
+  if (!(propertyName in obj))
+    return;
+
+  const prop = (obj as { [s: string]: unknown })[propertyName] as T;
+
+  if (typeof prop === 'undefined' || prop === null)
+    return;
+
+  return prop;
+};
```
quisquous, line 31:
```suggestion
const getArgument = <T>(obj: unknown, propertyName: string): T | undefined => {
  if (typeof obj !== 'object' || obj === null)
    return;
  if (!(propertyName in obj))
    return;

  const data = obj as { [key: string]: T };
  // TODO: it'd be nice to check typeof data here too?
  return data[propertyName];
};
```

Also, does this get used for anything but strings? If no, maybe changing `T` to be `string` and checking `typeof data`?

package.json
```diff
@@ -20,8 +20,10 @@
     "test": "node_modules/.bin/mocha",
     "lint-staged": "node_modules/.bin/lint-staged",
     "coverage-report": "node --loader ts-node/esm util/gen_coverage_report.js",
-    "find-translations": "node --loader ts-node/esm util/find_missing_translations.js",
-    "translate-timeline": "node --loader ts-node/esm util/translate_timeline.ts",
+    "util": "node --loader ts-node/esm util/index.ts",
+    "find-translations": "npm run util -- findTranslations",
```
quisquous, line 24: `npm run find-translations -- -l fr -f 05-shb` finds quite a few translations before this patch and nothing with this patch.

package.json
```diff
@@ -20,8 +20,10 @@
...
+    "translate-timeline": "npm run util -- translateTimeline",
```
quisquous, line 25: ### bug 1 (file selection error)

```
$ npm run translate-timeline

> cactbot@ translate-timeline C:\Users\enne\cactbot
> npm run util -- translateTimeline


> cactbot@ util C:\Users\enne\cactbot
> node --loader ts-node/esm util/index.ts "translateTimeline"

(node:29084) ExperimentalWarning: --experimental-loader is an experimental feature. T
his feature could change at any time
? Input a valid timeline filename:  ui\raidboss\data\02-arr\raid\t11.txt
? Select a locale:  fr
Couldn't find 'ui\raidboss\data\02-arr\raid\t11.txt', aborting.
```

### bug 2 (no output)

This command works before this patch, but not with this patch.

```
$ npm run translate-timeline -- -l fr -t t11

> cactbot@ translate-timeline C:\Users\enne\cactbot
> npm run util -- translateTimeline "-l" "fr" "-t" "t11"


> cactbot@ util C:\Users\enne\cactbot
> node --loader ts-node/esm util/index.ts "translateTimeline" "-l" "fr" "-t" "t11"

(node:1596) ExperimentalWarning: --experimental-loader is an experimental feature. Th
is feature could change at any time
```

quisquous (commented):
There's also a bunch of no-unsafe-return/no-unsafe-call/no-unsafe-assignment lint errors too when I rebased this patch.  (Sorry!)

util/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    type: 'list',
+    name: 'action',
+    message: 'What do you want to do?',
+    choices: Object.values(actionChoices),
+    default: getArgument<string>(args, 'action'),
+    when: () => typeof getArgument(args, 'action') !== 'string',
+  }]).then((answer: Answers) => {
+    const action = getArgument(answer, 'action') || getArgument<string>(args, 'action');
+    if (action === actionChoices.generate.value)
+      return generateDataFiles(args);
```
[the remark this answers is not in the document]
MaikoTan: It should be fine to both `return` or not `return`, but I think it is better to let it knows that the function returns a `Promise` so that should be an async function?

util/index.ts
```diff
@@ -0,0 +1,188 @@
...
+    }
+  });
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+const findMissingTranslationsFunc = (args: any) => {
+  return inquirer.prompt([
+    {
+      type: 'fuzzypath',
+      name: 'filter',
```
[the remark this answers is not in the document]
MaikoTan: We can change it not to be a `fuzzypath` but an `input` to fetch normal string input, also set a `default` value to `''`.

package.json
MaikoTan: Bug 2 fixed, that is because I forgot to fetch arguments from `args` when it has values.
But I can't reproduce Bug 1..

package.json
MaikoTan: Should be fixed, can you try again?

quisquous: Seems fixed, thanks!

package.json
quisquous: Bug 2 is fixed for me.  Bug 1 still happens.

I don't know why Bug 1 is not happening for you.  It looks like `findTriggersFile` gets `ui\raidboss\data\02-arr\raid\t11.txt` as a parameter, and it will never find that as a filename.  How does this function work for you?

util/index.ts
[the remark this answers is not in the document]
quisquous: Ok!

util/index.ts
```diff
@@ -0,0 +1,176 @@
...
+  }]).then(async (answers: Answers) => {
+    if (typeof answers.choice === 'string' && answers.choice in dataFilesMap)
+      return await dataFilesMap[answers.choice]?.();
+  });
+};
+
+const translateTimelineFunc = (args: unknown): Promise<void> => {
+  return inquirer.prompt([
+    {
+      type: 'fuzzypath',
```
quisquous, line 128:
```suggestion
      type: 'fuzzypath',
      excludeFilter: (path: string) => !path.endsWith('.txt'),
```

util/index.ts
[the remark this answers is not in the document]
quisquous: What about: `type: 'fuzzypath'` and `suggestOnly: true`? This lets you hit enter but still gives you fuzzy matching.  However if you hit enter in this case, then filter is null, so some adjustments need to be made both in allowing the function call to be made and handling null (currently only undefined counts as "all" in `findMissingTranslations`).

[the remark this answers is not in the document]
MaikoTan: Oh, you're right. I forgot that it has this option!

package.json
MaikoTan: I don't know, either. It runs properly on my side.

I can see a difference that I run it on a Linux machine, and you run on Windows?

<details><summary>Images (Click to Expand)</summary>

Snipaste_2021-07-19_08-33-26
Snipaste_2021-07-19_08-33-51

</details>

panicstevenson:
```diff
-  shortName = shortName.replace(/\.(?:[jt]s|txt)$/, '');
+  shortName = shortName.replace(/\.(?:[jt]s|txt)$/, '').split(path.sep).join(path.posix.sep);
```
Fixes this for me and appears to be backwards compatible with the old script invocation for both Git Bash and Command Prompt from Windows.

I'm assuming this is a no-op for Linux given what the code does, but still might be worth a quick test.

MaikoTan: Looks like the code you wrote would replace all the path separators to the POSIX one (`/`)? I have no idea, but why is it compatible with Windows?

quisquous: Thanks, that works for me on Windows!

PR #3164, opened 2021-06-22
eslint: add indent rule for TypeScript

quisquous:
This is Trim21's #3102 with a workaround extended from:
https://stackoverflow.com/questions/59851672/eslint-indent-and-ignorenodes-trouble-getting-ast-selectors-to-work-correctl/59852368#59852368

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -98,6 +98,12 @@ const rules = {
         'parameters': 2,
       },
       'ignoreComments': false,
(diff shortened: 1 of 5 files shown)


panicstevenson (approved):
Looks good to me. I wish there were a way to throw 'TSIntersectionType' et all into something that we could just be like `'TSIntersectionType': 1`, but I don't see that as possible.

PR #3167, opened 2021-06-22
Chore: dungeon timelines improvement

Jaehyuk-Lee:
Rolled back
~~I think this is better than~~

~~This.~~
K-006

And, I wonder why there's `(latest version)` written. That would be obviously latest version if you see on the web. Does this mean that this is later version than the release version? (The code is written before it's been released.)

### DIFF
--- a/ui/raidboss/data/05-shb/dungeon/matoyas_relict.txt
+++ b/ui/raidboss/data/05-shb/dungeon/matoyas_relict.txt
@@ -83,7 +83,7 @@ hideall "--sync--"
 3086.5 "Minced Meat" sync /:Mother Porxie:5911:/ window 50,10
 3093.6 "--untargetable--"
 
-3096.6 "Huff And Puff" sync /:591A:Mother Porxie/ window 96.6,5 duration 34.3
+3096.6 "Huff And Puff" sync /:591A:Mother Porxie/ window 96.6,5 duration 33.7
 3101.6 "Buffet" sync /:Aeolian Cave Sprite:5926:/
 3107.3 "Buffet" sync /:Aeolian Cave Sprite:5926:/
 3112.8 "Buffet" sync /:Aeolian Cave Sprite:5926:/
--- a/ui/raidboss/data/05-shb/dungeon/matoyas_relict.txt
+++ b/ui/raidboss/data/05-shb/dungeon/matoyas_relict.txt
@@ -83,7 +83,7 @@ hideall "--sync--"
 3086.5 "Minced Meat" sync /:Mother Porxie:5911:/ window 50,10
 3093.6 "--untargetable--"
 
-3096.6 "Huff And Puff" sync /:591A:Mother Porxie/ window 96.6,5 duration 49.70
+3096.6 "Huff And Puff" sync /:591A:Mother Porxie/ window 96.6,5 duration 34.3
 3101.6 "Buffet" sync /:Aeolian Cave Sprite:5926:/
 3107.3 "Buffet" sync /:Aeolian Cave Sprite:5926:/
 3112.8 "Buffet" sync /:Aeolian Cave Sprite:5926:/
--- a/ui/raidboss/data/05-shb/alliance/the_copied_factory.js
+++ b/ui/raidboss/data/05-shb/alliance/the_copied_factory.js
@@ -792,6 +792,7 @@ export default {
   timelineReplace: [
     {
       'locale': 'de',
+      'missingTranslations': true, // needs 'Back' and 'Front(?!al)' translation
       'replaceSync': {
         '9S-Operated Flight Unit': '9S\' Flugeinheit',
         '9S-Operated Walking Fortress': '9S\' mehrbeinig(?:e|er|es|en) Panzer',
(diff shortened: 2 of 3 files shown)


MaikoTan:
In fact, we have discussed before https://github.com/quisquous/cactbot/pull/3146#discussion_r655104777

Jaehyuk-Lee:
Oh, I didn't recognize that it wasn't opening on the web browser.

MaikoTan:
And yes, the content coverage is exported from the `main` branch, so it would always newer than any release.
(But if you are using cactbot via `gh-pages` URL in OverlayPlugin, you could always use the "latest version")

Jaehyuk-Lee:
Guess 3651193 fixes #3169. I'm not sure I don't really touch timeline.

ui/raidboss/data/05-shb/dungeon/matoyas_relict.txt
```diff
@@ -83,7 +83,7 @@ hideall "--sync--"
 3086.5 "Minced Meat" sync /:Mother Porxie:5911:/ window 50,10
 3093.6 "--untargetable--"
 
-3096.6 "Huff And Puff" sync /:591A:Mother Porxie/ window 96.6,5 duration 49.70
+3096.6 "Huff And Puff" sync /:591A:Mother Porxie/ window 96.6,5 duration 34.3
```
valarnin, line 86: I think usually the duration time is based on cast time? I don't have a relict log handy for reference, but can you confirm the cast time on this `Huff And Puff` cast maybe?

It's possible as I indicated in the issue that instead of shortening the duration, instead the enrage time needs pushed back and more `Buffet` casts need added to fill.

valarnin: `Buffet` is the wind AoE that the add casts repeatedly.

Jaehyuk-Lee: Oh, that is? So `Huff And Puff` is waiting for execution until you kill the wind? Guess `Huff And Puff Enrage?` has no cast-bar. And... if we are gonna make sure about duration, we should run this many time to see if enrage happens when wind dies 1 second before the enrage or 2 seconds before or etc.

valarnin: The order of operations is:

1. Untargetable. Boss jumps/disappears and starts casting `Huff And Puff`
2. Add spawns.
3. Add spams `Buffet` AoEs until it dies or Huff And Puff cast finishes (enrage)

So there's two possibilities here:
1. The `Huff And Puff` duration is correct, but the amount of `Buffet` AoEs and the timing of enrage is wrong
2. The `Huff And Puff` duration is incorrect, everything else is correct

This should be verifiable by checking the log file for the fight for the duration of that `Huff And Puff` cast.

Jaehyuk-Lee: ah.. so the enrage time can be wrong, not the duration of `Huff And Puff`.

valarnin: Ah. I *did* have an old log file sitting around with a Relict run.

```
20|2021-06-01T22:56:37.3010000-04:00|4001AA14|Mother Porxie|591A|Huff And Puff|4001AA14|Mother Porxie|49.70||49c62518d45c4710d129a79f9a874b24
```

49.70 does appear to be the cast time for the spell, so it seems like possibility 1 above (enrage time is wrong) is correct?

Jaehyuk-Lee: Umm.. I can't really read the logs, I cannot understand that.
You mean the duration is right, and the enrage is wrong?

MaikoTan: In fact, the gimmick here is
1. Boss fly high
1. A Sprite appears
1. Kill the Sprite.
1. The Sprite dies and cast wind, bring players fly
1. Players could see where the Boss is, and run there for a knockback when fell on the floor

If players didn't kill the Sprite in a certain time, then they couldn't fly and see where the Boss is. So players would probably be hurt by the AoE, which is called "Enrage" here.

MaikoTan: And is there a way to "sync" when the Sprite dies, and remove the running timeline of "Huff And Puff"?

Jaehyuk-Lee: > And is there a way to "sync" when the Sprite dies, and remove the running timeline of "Huff And Puff"?

I've never seen timeline sync with something die.

valarnin: Could sync on the removed combatant or was defeated line for the sprite? But the enrage line's sync actually syncs for the end of the cast regardless, so that sync is fine for either outcome. Just need to adjust the time for that sync to line up with the duration (`3096.6 + 49.7 = 3146.3`, delta of 16 seconds?), adjust the rest of the timeline after that point to match, then fill in the space between the last `Buffet` and the enrage with more `Buffet`s

MaikoTan: But sync to the ending skill of the Sprite seems make sense, I can find that it is called ["Explosion"](https://xivapi.com/Action/20020?columns=Name,Name_ja)

Jaehyuk-Lee: Oh yeah that makes sense.

Jaehyuk-Lee:
umm I guess I should have changed this PR to draft. #3169 problem is not fixed yet.

PR #3170, opened 2021-06-22
i18n: add new config ja/cn/ko translation

MaikoTan:
@Echoring Please review.

### DIFF
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -995,7 +995,7 @@ const templateOptions = {
       id: 'Coverage',
       name: {
         en: 'Supported content (latest version)',
-        ja: 'コンテンツ完成度一覧 (最新バージョン)',
+        ja: '対応コンテンツ一覧 (最新バージョン)',
         cn: '支持副本一览 (最新版本)',
       },
       type: 'html',
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -995,6 +995,8 @@ const templateOptions = {
       id: 'Coverage',
       name: {
         en: 'Supported content (latest version)',
+        ja: 'コンテンツ完成度一覧 (最新バージョン)',
+        cn: '副本覆盖度 (最新版本)',
       },
       type: 'html',
       html: {
--- a/ui/raidboss/raidboss_config.js
+++ b/ui/raidboss/raidboss_config.js
@@ -996,7 +996,7 @@ const templateOptions = {
       name: {
         en: 'Supported content (latest version)',
         ja: '対応コンテンツ一覧 (最新バージョン)',
-        cn: '支持副本一览 (最新版本)',
+        cn: '支持副本一览 (含未发布更新)',
       },
       type: 'html',
       html: {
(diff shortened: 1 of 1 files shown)


ui/raidboss/raidboss_config.js
```diff
@@ -995,6 +995,8 @@ const templateOptions = {
       id: 'Coverage',
       name: {
         en: 'Supported content (latest version)',
+        ja: 'コンテンツ完成度一覧 (最新バージョン)',
+        cn: '副本覆盖度 (最新版本)',
```
Echoring, line 999:
```suggestion
        cn: '支持副本一览 (最新版本)',
```

ui/raidboss/raidboss_config.js
```diff
@@ -995,6 +995,8 @@ const templateOptions = {
       id: 'Coverage',
       name: {
         en: 'Supported content (latest version)',
+        ja: '対応コンテンツ一覧 (最新バージョン)',
+        cn: '支持副本一览 (最新版本)',
```
Jaehyuk-Lee, line 999:
```suggestion
        cn: '支持副本一览 (最新版本)',
        ko: '지원하는 컨텐츠 (릴리즈버전보다 최신)',
```

MaikoTan: You quoted "Newer than release"?
I wonder if we should add in ja/cn too?

Jaehyuk-Lee: I don't know... It's very hard to describe this "commit-release" step of publishing to non-developers. I just tried to notice that this list may contain that are not released yet.

MaikoTan: Oh yeah. And especially for CN/KO server which is outdated compare to the intl server. So there might be some "un-implemented content" included.
I think I would keep the Japanese translation remain, and ping @Echoring for suggestion again.

Echoring: hmmm...... maybe we can use a more common word like `Developing version included` or `Beta version included`？

Jaehyuk-Lee: `Developing version included` looks good.

Jaehyuk-Lee: or let's use `<br>` and describe it long like
```
Supported content
(not-yet released contents included)
```

Jaehyuk-Lee: Or maybe we could add some description on the `coverage.html` page

MaikoTan: Hmm, I would agree to add a description on the HTML page.
As for the CN translation, I think `not-yet released contents included` is good, I can abbreviate it into "含未发布更新" which has only 6 characters.

Jaehyuk-Lee: Imo, most Korean users are aware that the international version has more contents updated, they'll understand what English named ones are like "The Tower at Paradigm's Breach" in Korean page. I don't think Korean need something like that.

MaikoTan: Well, I mainly play in JP, so I'm not very clear that if most CN players are all the same.
If there is a description, I would translate it as soon as possible. ^_^

PR #3172, opened 2021-06-22
plugin: Fetch remote FateWatcher opcodes

KattTails:
This should ready for an initial review, any and all feedback is welcome.

I'm open to ideas on if a fallback should be added in case of failure to fetch the remote json. Considering the nature of ACT and FFXIV it would be expected that there wouldn't be connection problems and that the remote file is always available. On the other hand, a fallback to a locally stored copy of the opcode.json wouldn't be hard to implement and would alleviate any connection issues. 

I'm also open to suggestions on the appropriate location for the FATEOpcodes.json file. My thought at the moment is to store it in the `ui\eureka\` folder so it is included in Cactbot releases. This way it could be used as a local fallback if we decide to implement the option mentioned above, without many other changes. That being said, it isn't a file exclusive for the eureka overlay and so might be better suited somewhere else. 

Before a final merge the remote url will need updating to the final location, I've set it to a temporary repo just for initial testing.

### DIFF
--- a/plugin/CactbotEventSource/FateWatcher.cs
+++ b/plugin/CactbotEventSource/FateWatcher.cs
@@ -45,6 +45,7 @@ namespace Cactbot {
     private static readonly string remoteOpcodeURL = "https://katttails.github.io/TestRepo/FATEOpcodes.json";
 
     bool opcodesLoaded = false;
+    List<string> errors = new List<string>();
     private Region OpCodes;
 
     public struct AC143Category {
@@ -68,6 +69,7 @@ namespace Cactbot {
     }
 
     public struct CEOpcodeList {
+      public uint version;
       public Region intl;
       public Region cn;
       public Region ko;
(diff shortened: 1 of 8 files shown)


MaikoTan:
What if download fail? Is there a fallback method or let user download the file manually and place it into some specific place?

valarnin:
I'm not really familiar with the fate watcher code, but is there caching for the downloaded opcodes at all? Given how github likes to randomly go down as seen with general plugin updates and such, it's probably not a good idea to add a constant hard dependency on GH being up?

quisquous:
One thing you could do is save the values as part of the cactbot config, like `cactbotSaveData` does for overlays and cactbot config.

I think what's not clear to me is what to do with that caching other than use it as a fallback.  Is there a way to know when to go look?  Or some way to throttle?  Maybe it's fine to fetch when the plugin starts every time.

KattTails:
> I think what's not clear to me is what to do with that caching other than use it as a fallback. Is there a way to know when to go look? Or some way to throttle? Maybe it's fine to fetch when the plugin starts every time.

I was thinking of just checking once per ACT launch. With regards to the fallback, I thought about adding a version number to the json and just comparing that with the locally stored values. If the remote is greater than replace, but also if the local copy is greater then allow that to override too.

The only reason I'm hesistant to add it is because I've never personally experienced any connection issues to github/GH pages. If it's a concern for people then I can understand the need for it but for an always-online project such as ACT

quisquous:
> The only reason I'm hesistant to add it is because I've never personally experienced any connection issues to github/GH pages. If it's a concern for people then I can understand the need for it but for an always-online application such as ACT it doesn't necessarily make sense?

I think we might need this sort of thing:
* try to get latest github version
* fallback to cached version
* if no cached version, fallback to version shipped with the release

I thought I remembered @Trim21 or @Jaehyuk-Lee (sorry for ping, but it'd be good to know this more concretely) saying something about github not being available for everybody or having bad connectivity?

Jaehyuk-Lee:
If Github get down, I don't think it lasts very long. The recent [fastly related problem](https://theconversation.com/fastly-global-internet-outage-why-did-so-many-sites-go-down-and-what-is-a-cdn-anyway-162371) maked many overlay stop working. Meanwhile, there had been a fire on the server which hosts some of [ffxiv-teamcraft](https://github.com/ffxiv-teamcraft/ffxiv-teamcraft) database. The Github was recovered fast compare to the ffxiv-teamcraft server problem. Github is big, so I think it's reliable. I remember it was recovered in 2-3 hours.

If we have backup methods, it would be better of course. But I don't think we have to implement it. If we were making parser for example, it would definitely better to make a backup. But, this is only for CE, and even though it doesn't work, people won't feel uncomfortable that much.

Trim21:
> I thought I remembered @Trim21 or @Jaehyuk-Lee (sorry for ping, but it'd be good to know this more concretely) saying something about github not being available for everybody or having bad connectivity?

yes, in fact most users in China have a hard time connecting to GitHub. and I also don't want to set a proxy for my act.

Jaehyuk-Lee:
Oh, ummm that could be a problem if it happens often. I though China doesn't block Github connection or something.

MaikoTan:
I don't think that China government would block Github, there are some anti-government contents on github and got blocked when accessing them in China truly, but that is Github block those content on purpose when user access them from China.
Theoretically, I think many user cannot access `github.io` because of their DNS of ISP and somewhat that aws just losing connections in some ragion of China.
I have lived in Guangzhou and Shenzhen with China Mobile network, and I merely met connection issue on Github after I manually set IP address to `githubusercontent.com` and some other related domains in `/etc/hosts`.

All in all, I think this discussion is off topic. And I would glad to see there is a button to update `opcodes.json` manually. Another thought is that make the "update url" configurable in OverlayPlugin's configuration json or something, so that they Chinese publisher can simply modify the url to their own site.

Jaehyuk-Lee:
@MaikoTan ummm so putting the exact IP address instead of putting domain name would fix or solve this?

Trim21:
> ummm so putting the exact IP address instead of putting domain name would fix or solve this?

Maybe using https://www.jsdelivr.com/ is a better option.

MaikoTan:
> @MaikoTan ummm so putting the exact IP address instead of putting domain name would fix or solve this?

I don't think so, many http server redirect user requests based on their host field in the http request. And setting a IP address would restrict user to access the specific ragion of Github's CDN, for example, a Australian user would get a faster and stabler connection when connect to the CDN in Singapore than connect to one in England.

Jaehyuk-Lee:
Ah, it works that way.

MaikoTan:
> > ummm so putting the exact IP address instead of putting domain name would fix or solve this?
> 
> Maybe using https://www.jsdelivr.com/ is a better option. (than hard code the IP)
> 
> 

Yeah, I'll agree to use jsdelivr too. It is much reliable in China.

KattTails:
I've had another pass on this taking into consideration the feedback given. 

I've added an option to disable the remote updates altogether and an option to configure an alternate source url for the opcode json. The remote file is also cached now in the local config (thanks for the suggestion!) and updated when necessary. This also allows users to override it completely if they wish.

ui/config/config.js
```diff
@@ -393,6 +394,22 @@ export default class CactbotConfigurator {
     parent.appendChild(div);
   }
 
+  buildUrl(parent, opt, group) {
```
KattTails, line 397: I'd like another set of eyes on this, I've attempted to mimic the buildFloat code and from some basic testing it appears to work but I'm not really familiar with the underlying system here.

ui/eureka/FATEOpcodes.json
```diff
@@ -0,0 +1,35 @@
+{
```
KattTails, line 1: This doesn't need to be kept in the ui/eureka/ folder, but I'm not sure on the best place. Thoughts welcome.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -7,97 +7,73 @@
...
-      0x935,
-      0x936,
-      0x93E
-    );
-
-    private struct CEDirectorOPCodes {
-      public CEDirectorOPCodes(int size_, int opcode_) { this.size = size_; this.opcode = opcode_; }
-      public int size;
-      public int opcode;
+    private static readonly string remoteOpcodeURL = "https://katttails.github.io/TestRepo/FATEOpcodes.json";
```
KattTails, line 45: TODO: This needs adjusting before the final merge based on where exactly we're going to be putting the json.

ui/config/general_config.js
```diff
@@ -103,5 +103,31 @@ UserConfig.registerOptions('general', {
...
+      id: 'UseCustomOpcodeSource',
+      name: {
+        en: 'Use custom opcode source',
+      },
+      type: 'checkbox',
+      default: false,
+      debugOnly: true,
+    },
+    {
+      id: 'CustomOpcodeSourceUrl',
```
quisquous, line 124: Do we need both "use custom opcode source" and "custom opcode source url"? Could these be the same with "blank=don't use"?

ui/eureka/FATEOpcodes.json
quisquous: I think `resources/fate_opcodes.json` might be a reasonable place, since we have a lot of other data there.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -132,26 +124,96 @@ public struct CEDirectorData {
...
+
+      localOpcodeList = JsonConvert.DeserializeObject<CEOpcodeList>(Config.OverlayData["options"]["opcodes"].ToString(),
+          new JsonSerializerSettings {
+            Error = delegate (object sender, Newtonsoft.Json.Serialization.ErrorEventArgs args) {
+              errors.Add(args.ErrorContext.Error.Message);
+              args.ErrorContext.Handled = true;
+            }
+          });
+
+      if (errors.Count > 0) {
```
quisquous, line 198: Is it possible to fall back to the local file here? In other words, I feel like a cactbot release should update this and even if github is down it should work.

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -132,26 +124,96 @@ public struct CEDirectorData {
...
+        client_.LogError("There were errors encountered parsing the local opcodes:\r\n" +
+          "{0}" +
+          "FATE/CE data is unavailable.", String.Join("\n", errors));
+        return;
+      }
+
       if (language == "ko")
-        region_ = "ko";
+        OpCodes = localOpcodeList.ko;
```
quisquous, line 206: Ok, I know cactbot is not consistent about capitalization, but can I request lowercase camelCase for variables and CapitalCamelCase for class names for some attempt at consistency?

plugin/CactbotEventSource/FateWatcher.cs
```diff
@@ -132,26 +124,96 @@ public struct CEDirectorData {
...
+        json = opcodeTask.GetAwaiter().GetResult();
+      }
+      return json;
+    }
+
+    public FateWatcher(CactbotEventSource client, CactbotEventSourceConfig Config, string language) {
       client_ = client;
+      CEOpcodeList localOpcodeList = new CEOpcodeList();
+
+      if (!Config.DisableAutomaticOpcodeUpdates) {
```
quisquous, line 140: This is a pretty long function.  Is it possible to have a `TryToGetRemoteOpcodes()` function that returns null or the remote opcodes to wrap some of this?

ui/config/config.js
quisquous: I haven't tested it locally, but this looks fine to my eyes, thanks!

quisquous (commented):
Thanks again for this! Overall looks fine to me.  I have a few small/medium-sized bits of feedback.

KattTails:
Sorry for abandoning this, it was 99% done and then I got ill and by the time I felt well enough to work on it again it had been so long that it was just easier to pretend it didn't exist. 

It'll return some time in the future after I've had another go at it, I've had some ideas on how to implement it a bit better.

quisquous:
> Sorry for abandoning this, it was 99% done and then I got ill and by the time I felt well enough to work on it again it had been so long that it was just easier to pretend it didn't exist.

No worries at all! If you send an updated version in the future, I'd be happy to look.  If you want me to put up a PR myself, I can try to take a look eventually too.

I'm sorry to hear about your health, and I hope things have improved some since then for you.  Bodies really are the worst sometimes (or all the time).

PR #3173, opened 2021-06-23
util: Rework overlay API to allow overrides in typescript and make reliant code more type-safe

valarnin:
As discussed various places, mainly on #3149 but also in a few other places, the existing `overlay_plugin_api.ts` implementation is causing problems with converting raidemulator to typescript, surrounding typing the override code.

This PR rewrites the overlay plugin API code to be fully properly typed in both directions in/out of the API. There's some type mangling, mostly via `... as never` casts, which still don't really feel great, but everything here's functional for all cases I've tested (raidemulator, raidboss, jobs).

I've also completed the typescript conversion of `RaidEmulatorOverlayApiHook` on a sub-branch to confirm that this PR fully resolves the issues I encountered.


quisquous:
I haven't forgotten this patch, but there's a lot to digest.

valarnin:
Closing in favor of the minimal changes from #3258

PR #3181, opened 2021-06-24
raidemulator: Switch Encounter.initialTimestamp from getter to property

### LINKED ISSUE #3163: Emulator displays weird Pull Duration during import
image

(The blue is me highlighting it.  The bug is the text.)

When importing the log from https://github.com/quisquous/cactbot/issues/3158

valarnin:
Fixes #3163, caused due to the getter not transferring properly through the websocket API

### DIFF
--- a/ui/raidboss/emulator/data/Encounter.ts
+++ b/ui/raidboss/emulator/data/Encounter.ts
@@ -42,6 +42,7 @@ export default class Encounter {
   duration = 0;
   playbackOffset = 0;
   language: Lang = 'en';
+  initialTimestamp = Number.MAX_SAFE_INTEGER;
 
   constructor(
     public encounterDay: string,
@@ -110,16 +111,14 @@ export default class Encounter {
         this.initialOffset = 0;
     }
 
+    this.initialTimestamp = this.startTimestamp + this.initialOffset;
+
     const firstLine = this.logLines[this.firstLineIndex];
 
     if (firstLine && firstLine.offset)
       this.playbackOffset = firstLine.offset;
 
     this.startStatus = [...startStatuses].sort().join(', ');
-  }
-
-  public get initialTimestamp() : number {
-    return this.startTimestamp + this.initialOffset;
   }
 
   shouldPersistFight(): boolean {


ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -118,10 +121,6 @@ export default class Encounter {
     this.startStatus = [...startStatuses].sort().join(', ');
   }
 
-  public get initialTimestamp() : number {
-    return this.startTimestamp + this.initialOffset;
-  }
-
   shouldPersistFight(): boolean {
     return isValidTimestamp(this.firstPlayerAbility) && isValidTimestamp(this.firstEnemyAbility);
   }
```
quisquous, line 126: Should this function be a property too?

valarnin: This one is invoked worker-side so it doesn't need changed.

quisquous: Mostly I'm trying to ask if there's a better way to prevent this sort of thing happening in the future, like having no functions on the class (or some TypeScript way to make this happen??)

valarnin: I googled a bit and I don't see a typescript-side way to prevent this. Maybe you could do it with some eslint rules or something but I'm not sure.

We could add comments around the code for worker, encounter, and combatanttracker that only hard properties will get transferred, I guess?

valarnin: Actually, this shouldn't be a problem once Persistor is converted to typescript since that conversion will also involve moving to Dexie which will assign the prototypes to the objects properly, so getters and functions will work as normal. There's a stopgap as well that can be done, as follows:

```diff
diff --git a/ui/raidboss/emulator/data/Encounter.ts b/ui/raidboss/emulator/data/Encounter.ts
index 71d32f0e..6debc6d3 100644
--- a/ui/raidboss/emulator/data/Encounter.ts
+++ b/ui/raidboss/emulator/data/Encounter.ts
@@ -25,6 +25,12 @@ const isValidTimestamp = (timestamp: number) => {
   return timestamp > 0 && timestamp < Number.MAX_SAFE_INTEGER;
 };

+export const isEncounter = (obj: unknown): obj is Encounter => {
+  if (!(obj instanceof Object))
+    return false;
+  return 'logLines' in obj;
+};
+
 export default class Encounter {
   private static readonly encounterVersion = 1;
   public id?: number;
@@ -125,8 +131,8 @@ export default class Encounter {
     return isValidTimestamp(this.firstPlayerAbility) && isValidTimestamp(this.firstEnemyAbility);
   }

-  upgrade(version: number): boolean {
-    if (Encounter.encounterVersion <= version)
+  upgrade(): boolean {
+    if (Encounter.encounterVersion <= this.version)
       return false;

     const repo = new LogRepository();
@@ -140,4 +146,15 @@ export default class Encounter {

     return true;
   }
+
+  static prototypeObject(obj: unknown): void {
+    if (!isEncounter(obj))
+      return;
+
+    Object.setPrototypeOf(obj, Encounter.prototype);
+
+    obj.logLines.forEach((l) => {
+      Object.setPrototypeOf(l, LineEvent.prototype);
+    });
+  }
 }
diff --git a/ui/raidboss/emulator/data/Persistor.js b/ui/raidboss/emulator/data/Persistor.js
index 5129f6b5..7e5cd0e5 100644
--- a/ui/raidboss/emulator/data/Persistor.js
+++ b/ui/raidboss/emulator/data/Persistor.js
@@ -1,7 +1,7 @@
 import EmulatorCommon from '../EmulatorCommon';
 import EventBus from '../EventBus';
 import PersistorEncounter from './PersistorEncounter';
-import Encounter from './Encounter';
+import Encounter, { isEncounter } from './Encounter';

 export default class Persistor extends EventBus {
   constructor() {
@@ -65,7 +65,6 @@ export default class Persistor extends EventBus {
         resolver = res;
       });
       const encounter = EmulatorCommon.cloneData(baseEncounter, []);
-      delete encounter.combatantTracker;
       const encountersStorage = this.encountersStorage;
       let req;
       if (encounter.id === null) {
@@ -96,19 +95,16 @@ export default class Persistor extends EventBus {
         const req = encountersStorage.get(id);
         req.addEventListener('success', (ev) => {
           const enc = req.result;
-          const ret = new Encounter(enc.encounterDay,
-              enc.encounterZoneId,
-              enc.encounterZoneName,
-              enc.logLines);
-          ret.id = enc.id;
+          if (!isEncounter(enc))
+            return;
+          Encounter.prototypeObject(enc);
           // Check for encounter upgrade, re-save encounter if it's upgraded.
-          if (ret.upgrade(enc.version)) {
-            this.persistEncounter(ret).then(() => {
-              res(ret);
+          if (enc.upgrade(enc.version)) {
+            this.persistEncounter(enc).then(() => {
+              res(enc);
             });
           } else {
-            ret.initialize();
-            res(ret);
+            res(enc);
           }
         });
       } else {
```

quisquous: Mostly my concern here was about how to prevent future regressions.  It sounds like you have an idea for how to solve that, so that addresses what I was worried about, thanks!

PR #3183, opened 2021-06-25
raidemulator: Convert EncounterTab to typescript

valarnin:


valarnin:
There's some repeated `querySelector` helpers here and in my next PR, after these are all landed I'll put in a separate PR to move those helpers to EmulatorCommon instead, I just don't want to hold up these PRs on another one.

ui/raidboss/emulator/ui/EncounterTab.ts
```diff
@@ -0,0 +1,285 @@
...
+
+    const zoneMap = this.currentZone ? this.encounters[this.currentZone] : undefined;
+
+    if (!zoneMap)
+      return;
+
+    const dateMap = this.currentDate ? zoneMap[this.currentDate] : undefined;
+
+    if (!dateMap)
+      return;
```
quisquous, line 195:
```suggestion
    if (!this.currentZone || !this.currentDate)
      return;

    const zoneMap = this.encounters[this.currentZone];
    if (!zoneMap)
      return;
  
    const dateMap = zoneMap[this.currentDate];
    if (!dateMap)
      return;
```

What about something more like this to avoid the unnecessary ternary?

valarnin: This was the first of these three that I converted and I think it's a bit obvious I improved the code after this one 😉

quisquous:
Whoops, sorry.  Where is my lint-for-suggestions... <_<

PR #3185, opened 2021-06-25
raidemulator: Convert EmulatedPartyInfo to typescript

valarnin:


ui/raidboss/emulator/ui/EmulatedPartyInfo.ts
```diff
@@ -2,23 +2,112 @@ import EmulatorCommon from '../EmulatorCommon';
 import EventBus from '../EventBus';
 import Tooltip from './Tooltip';
 import Util from '../../../../resources/util';
+import RaidEmulator from '../data/RaidEmulator';
+import { UnreachableCode } from '../../../../resources/not_reached';
+import AnalyzedEncounter, { PerspectiveTrigger } from '../data/AnalyzedEncounter';
+
+const jobOrder = [
```
quisquous, line 10: What do you think about a `BattleJob` subset of `Job` so that you could type this to require all `BattleJob` to be listed here, and a more general `isBattleJob` / `isJob` helper like we have an `isLang` helper?

If we do that, is there some TypeScript way to ensure that all of `BattleJob` is listed here?

ui/raidboss/emulator/ui/EmulatedPartyInfo.ts
```diff
@@ -50,192 +139,214 @@ export default class EmulatedPartyInfo extends EventBus {
...
-    stateID = stateID || combatant.getState(0);
 
     const State = combatant.getState(stateID);
     if (State === undefined)
       return;
 
+    const display = this.displayedParty[id];
+
+    if (!display)
+      throw new UnreachableCode();
```
quisquous, line 294: Why is this `throw new UnreachableCode` but all the other exceptions above are `return`? (Also, if you're going to assign and then check a variable, I find it easier to read with no newline between like you do above.)

ui/raidboss/emulator/ui/EmulatedPartyInfo.ts
valarnin: Maybe it makes sense to split the jobs out entirely in the other direction, such that this is in a `job.ts` file:

```ts
const noGaugeBattleJobs = [...] as const; // these are GLA/etc that don't have job gauges
const gaugeBattleJobs = [...] as const;
const battleJobs = [...noGaugeBattleJobs, ...gaugeBattleJobs] as const;
const gatheringJobs = [...] as const;
const craftingJobs = [...] as const;
const nonBattleJobs = [...gatheringJobs, ...craftingJobs] as const;
const jobs = [...battleJobs, ...nonBattleJobs] as const;

export type NoGaugeBattleJob = typeof noGaugeBattleJobs[number];
// repeat for others

export const isNoGaugeBattleJob = (job?: string): job is NoGaugeBattleJob => {
  const jobStrs: readonly string[] = noGaugeBattleJobs;
  if (!job)
    return false;
  return jobStrs.includes(job);
};
// repeat for others
```

(noGaugeBattleJobs is split out for JobDetail reasons)

ui/raidboss/emulator/ui/EmulatedPartyInfo.ts
valarnin: The first `return;` was correct, the rest should have been `UnreachableCode`. This did lead me to find a pointless call to `updateCombatantInfo`, cleaned up in latest commit.

ui/raidboss/emulator/ui/EmulatedPartyInfo.ts
quisquous: Hmm.  I think when you put it like this, it subjectively feels like too much.  I think the reason that I am for having `Job` and `Role` be specific types in the first place is both so that triggers can type check `data.job === 'PLD'` style things, and also so that we can have things like `Record<Job, etc>` where we know that everything is filled in and if more jobs are added we know where to go find all the places to update.

However, there's definitely a tradeoff in that using `Job` over `string` adds some code friction to using these types.  Having even more types makes going between `BattleJob` and `NoGaugeBattleJob` require even more checking and friction, and I'm not sure it adds enough value.

Maybe if this is the logical conclusion then I'll take back my original suggestion and say the code is good as-is.

valarnin: I don't think this'll be as significant as you think it is? It could be simplified to three types to accomplish the same thing as far as current code is concerned:

1. GaugeBattleJob - used for JobDetail
2. BattleJob - Superset of GaugeBattleJob and remaining battle jobs, used for original suggestion above
3. Job - All jobs, as it exists now

If these types are built correctly, it's possible to do this code:

```ts
const j1: GaugeBattleJob = 'PLD';
const j2: BattleJob = j1;
const j3: Job = j1;
const j4: Job = j2;
```

e.g. you can expand the typing without any additional checks, but you can't narrow the type unless you check.

quisquous: Hmmm! I think I would almost rather fill out the the jobs that are missing for JobDetail with something like `'GLD': undefined;` than have a separate type for it, since it's likely not going to be used outside that file.  What do you think about that?

valarnin: This is already handled by the parent type:

```ts
type PlayerChangedJobDetails<T> = {
  job: T;
  jobDetail: JobDetail[T];
} | {
  job: Job;
  jobDetail: null;
}
```

It's `null` because that matches the actual return type from the overlay.

quisquous: Ok, sure `s/undefined/null/`.  I guess I'm trying to say that I don't think a separate type here adds much, and am proposing to cut back to two.

valarnin: I can see it either way, because the separate type for jobs with a gauge would hopefully allow us to ensure that JobDetail requires a mapped entry for each gauge job, though I'm blanking on the syntax for that right now.

valarnin: I looked at splitting Job out such that:

```ts
export type BattleJob = ...;
export type Job = 'NONE' | BattleJob | ...;
```

But if we want the typeguard functions, the file needs renamed and moved, which will impact a ton of other files to update imports. I think that's a bit much for this PR and maybe it should be done as a separate PR?

Also, the only good compiler-time way to ensure that the array has all of the required values is to do something like converting it to a type and use this: https://gist.github.com/aleclarson/a5e5efefb907a22aef64a00f8d7c7416

Otherwise things like duplicate job entries (e.g. two 'BLU' values in the order array) could cause the number to line up when testing a `jobOrder.length` check (which would otherwise work fine at compile-time since the array is declared `... as const`). And we can't convert to a `Set` and then check the length because only array literals can be declared `as const`, so the `.size` property or the `.length` property of any derived array will be typed as `number` instead of a const `18` or whatever.

So I guess to summarize:
1. Making this change as part of this PR is probably going to make this PR too big
2. Asserting that the order array contains the requisite entries at compile time requires some weird syntax

quisquous: Sounds good.  Thanks for pushing back that this doesn't make much sense in the end.

PR #3186, opened 2021-06-25
raidboss: Convert T5 to TypeScript

panicstevenson:


panicstevenson:
Normally I'd make the changes to the types in a separate PR (which I can still do), but I wanted the rationale to at least be nearby in case there were things that looked obviously bad. @valarnin @quisquous for thoughts if there's a way to avoid making those changes here, or if it's fine.

valarnin:
I think that the `watchCombatant` function's implementation is wrong here instead, and it should have just returned a `Promise<void>` and called `res()`?

There's no case where it would return false anyways, and the logic behind promises in PopupText doesn't care about the return type of the promise, it just cares that `trigger.promise` returns a promise at all.

The case could be argued instead that `trigger.promise` should always return a `Promise<boolean> | undefined`, and PopupText be updated to check the resolver value to determine if execution should continue, but I don't think there's ever a case where we want to deal with a Promise that has a unioned type.

quisquous:
> I think that the `watchCombatant` function's implementation is wrong here instead, and it should have just returned a `Promise<void>` and called `res()`?
> 
> There's no case where it would return false anyways, and the logic behind promises in PopupText doesn't care about the return type of the promise, it just cares that `trigger.promise` returns a promise at all.
> 
> The case could be argued instead that `trigger.promise` should always return a `Promise<boolean> | undefined`, and PopupText be updated to check the resolver value to determine if execution should continue, but I don't think there's ever a case where we want to deal with a Promise that has a unioned type.

Yeah, agreed with all of this.  I would add that I think rejecting the promise is a more natural way to prevent further execution.

ui/raidboss/data/02-arr/raid/t5.ts
```diff
@@ -1,14 +1,58 @@
...
+      monitoringHP: false,
+      hpThresholds: [0.85, 0.55, 0.29],
+      currentPhase: 0,
+    };
+  },
   triggers: [
+    {
+      id: 'T5 Twintania Phase Change Watcher',
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ source: 'Twintania' }),
```
quisquous, line 29: I'm a little bit hesitant to have a netregex that just matches all abilities, but doing it this way does save writing the trigger three times.  This is probably the best way to do it.

PR #3187, opened 2021-06-25
raidboss: Convert ARR Dungeons to TypeScript

panicstevenson:


ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.ts
```diff
@@ -67,12 +78,14 @@ export default {
       // The pack with the boss is 3 Violet Backs, not parsing for them prevents the trigger
       // from activating early when you pick up the Headgate Key and the boss and adds spawn.
       id: 'Brayflox Normal Pelican Adds',
+      type: 'AddedCombatant',
```
panicstevenson, line 81: Confirming assumptions if addedCombatantFull should be type AddedCombatant

ui/raidboss/data/02-arr/dungeon/haukke_manor.ts
```diff
@@ -66,13 +75,14 @@ export default {
     {
       // Void Lamp Spawn
       id: 'Haukke Normal Void Lamps',
+      type: 'GameLog',
```
panicstevenson, line 78: Confirming assumptions if message should be type GameLog

ui/raidboss/data/02-arr/dungeon/brayfloxs_longstop.ts
valarnin: Yes. Functionally they're the same behind the scenes as far as return fields, so it makes no sense to split them into separate types.

Same for GameLog below.

If the typing wasn't correct, this would error in tsc.

However, given that this trigger doesn't use data/matches/output directly (it's all done with canned helpers) you should be able to leave the type out?

PR #3189, opened 2021-06-26
fix: type of disabled trigger and `beforeSeconds` field

Trim21:
try to catch up upstream change in my user config files.

Find some problems:

```TS
  timelineTriggers: [
    {
      id: 'T8 timeline example',
      regex: /11/,
      infoText: () => 'hello',
    },
   { 
      id: "A Trigger I don't want",
      disabled: true,
    }
  ],
  triggers: [
    {
      id: 'E12S Double Aero Finder',
      disabled: true,
    },
]
```

first one can't compile because tsc says it missing `beforeSeconds` property. 

Don't know why it's required in timelineTrigger.

the second and third can't compile because tsc says it missing `regex` or `netRegex` property.

This PR set `beforeSeconds` optional and add a union on trigger type which make it can be disabled.

I'm not sure how do you think about this union.
I think built-in triggers won't have any disabled trigger, but it could be in user triggers, so I'm not sure it's ok to put this union here.

But it's not likely to have another trigger set type called something like `UserTriggerSet`

```TS
{
  triggers?: (NetRegexTrigger<Data> | { id: string; disabled: true })[];
  timelineTriggers?: (TimelineTrigger<Data> | { id: string; disabled: true })[];
}
```


quisquous:
> I think built-in triggers won't have any disabled trigger, but it could be in user triggers, so I'm not sure it's ok to put this union here.

This is the difference I am trying to express with `NetRegexTrigger` (a built-in cactbot trigger, with stricter type defintiions) and `LooseTrigger` (a user trigger, which has everything optional).   So, I think `LooseTriggerSet` already covers what you're proposing in `UserTriggerSet`, unless I'm misunderstanding?

I'm not sure I understand what you're trying to do with this:
```typescript
    {
      id: 'E12S Double Aero Finder',
      disabled: true,
    },
```

...is this a common pattern for you?

Trim21:
> So, I think `LooseTriggerSet` already covers what you're proposing in `UserTriggerSet`, unless I'm misunderstanding?

`LooseTriggerSet` is not a generic type, `data` didn't have any type information except `RaidbossData`

quisquous:
> > So, I think `LooseTriggerSet` already covers what you're proposing in `UserTriggerSet`, unless I'm misunderstanding?
> 
> `LooseTriggerSet` is not a generic type, `data` paramater didn't have any type information except `RaidbossData`

`LooseTriggerSet` is meant to be used by PopupText which doesn't know anything about the typing of data other than `RaidbossData`.  It's not intended to be used in trigger files itself, if that makes sense.

My thought is that if you are writing triggers in TypeScript, you should use the stricter `NetRegexTrigger` and `TriggerSet` types.  If cactbot is handling user triggers in PopupText, it should use the `LooseTriggerSet` which makes fewer assumptions.

Trim21:
> My thought is that if you are writing triggers in TypeScript, you should use the stricter `NetRegexTrigger` and `TriggerSet` types.

Yes, that's what I want to do.

And met this problem when disabling triggers. 

Maybe I should maintain my own copy of `TriggerSet`?

quisquous:
> I'm not sure I understand what you're trying to do with this:
> ...is this a common pattern for you?

This is what I'm trying to ask above.  What is the goal of disabling a trigger that isn't doing anything?  What are you trying to do?

Trim21:
Here is another example:

I have a TEA trigger for limit-cut.

This limit-cut trigger overwrite built-in limit-cut trigger, and with some other solution different with built-in trigger.

Then there is a new trigger `TEA Wormhole Puddle`, which I have implemented in my limit-cut trigger, so there is no need to enable the built-in version.

quisquous:
Oh!  Thank you for the explanation; I just couldn't understand why you wanted this.  I think I had imagined that most people would disable triggers through the config UI, but it is more clean to disable them in the trigger file in some ways.

That seems reasonable to me.  My one suggestion is can you make a `DisabledTrigger` type and use that instead of repeating {id, disabled}?

Trim21:
Yes, you just remaid me that the most important reason: these triggers didn't need to be disabled if this trigger file is not loaded, so they should be disabled in the trigger file not config UI.

I'll add that.

quisquous:
(Sorry for my confusion! Thanks for patiently explaining it to me!)

valarnin:
re: `beforeSeconds`, I'm working on a PR to convert the 00-misc/test.js file to typescript and there are timeline triggers there that just flat out don't have `beforeSeconds`, and there's code in place to default to 0 if not specified, so that seems correct.

Issue #3188, opened 2021-06-25
Unable to Convert Triggers with Complex Responses


panicstevenson:
In attempting to convert the following trigger to TypeScript:
```ts
{
  id: 'ShivaEx Sword Phase',
  type: 'Ability',
  netRegex: NetRegexes.ability({ source: 'Shiva', id: '993', capture: false }),
  netRegexDe: NetRegexes.ability({ source: 'Shiva', id: '993', capture: false }),
  netRegexFr: NetRegexes.ability({ source: 'Shiva', id: '993', capture: false }),
  netRegexJa: NetRegexes.ability({ source: 'シヴァ', id: '993', capture: false }),
  netRegexCn: NetRegexes.ability({ source: '希瓦', id: '993', capture: false }),
  netRegexKo: NetRegexes.ability({ source: '시바', id: '993', capture: false }),
  response: (data, _matches, output) => {
    // cactbot-builtin-response
    output.responseOutputStrings = {
      swordTankSwap: {
        en: 'Sword (Tank Swap)',
        de: 'Schwert (Tankwechsel)',
        fr: 'Épée (Tank Swap)',
        ja: '剣 (スイッチ)',
        cn: '剑（换T）',
        ko: '검 (탱커 교대)',
      },
      sword: {
        en: 'Sword',
        de: 'Schwert',
        fr: 'Épée',
        ja: '剣',
        cn: '剑',
        ko: '검',
      },
    };
    if (data.role === 'tank') {
      if (data.currentTank && data.slashing && data.slashing[data.currentTank])
        return { alertText: output.swordTankSwap!() };
    }

    return { infoText: output.sword!() };
  },
  run: (data) => data.soonAfterWeaponChange = true,
},
```
We receive this output:
```sh
$ tsc
ui/raidboss/data/02-arr/trial/shiva-ex.ts(116,7): error TS2322: Type '(data: Data, _matches: Matches & ToMatches<NetAbilityFields>, output: Output) => { alertText: string; infoText?: undefined; } | { ...; }' is not assignable to type 'ResponseField<Data, Matches> | ResponseField<Data, Matches & ToMatches<NetAbilityFields>> | ... 21 more ... | undefined'.
  Type '(data: Data, _matches: Matches & ToMatches<NetAbilityFields>, output: Output) => { alertText: string; infoText?: undefined; } | { ...; }' is not assignable to type 'ResponseFunc<Data, Matches>'.
[...]

PR #3190, opened 2021-06-26
fix: trigger `response` type

Trim21:
close: #3188

tsc is happy with the example in that PR and my user conf files.

@panicstevenson can you test it with your TS trigger?

I have test it with this example: https://gist.github.com/Trim21/8ed357968b6d8be416b3f2a197dba2fe

hope this contains all condition

(and indent and formatting problems are driving me crazy, why some part of this file is indented with 4 spaces? we should just add prettier with typescript......)

### DIFF
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -161,13 +161,13 @@ export type TriggerSet<Data extends RaidbossData> = {
 // Less strict type for user triggers + built-in triggers, including deprecated fields.
 export type LooseTimelineTrigger = Partial<TimelineTrigger<RaidbossData>>;
 
-export type LooseTrigger = Partial<BaseTrigger<RaidbossData, 'None'> &
+export type LooseTrigger = Partial<
+  BaseTrigger<RaidbossData, 'None'> &
   PartialRegexTrigger &
-  PartialNetRegexTrigger<'None'>>;
+  PartialNetRegexTrigger<'None'>
+>;
 
-export type LooseTriggerSet =
-  Exclude<Partial<TriggerSet<RaidbossData>>, 'triggers' | 'timelineTriggers'>
-  & {
+export type LooseTriggerSet = Exclude<Partial<TriggerSet<RaidbossData>>, 'triggers' | 'timelineTriggers'> & {
   /** @deprecated Use zoneId instead */
   zoneRegex?: RegExp | { [lang in Lang]?: RegExp };
   triggers?: LooseTrigger[];
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -50,11 +50,15 @@ export type TriggerFunc<Data extends RaidbossData, MatchType extends Matches, Re
 
 // The output from a response function (different from other TriggerOutput functions).
 export type ResponseOutput<Data extends RaidbossData, MatchType extends Matches> = {
-  infoText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>>;
-  alertText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>>;
-  alarmText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>>;
-  tts?: TriggerFunc<Data, MatchType, PartialTriggerOutput<Data, MatchType>>;
-};
+  infoText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>> |
+    TriggerOutput<Data, MatchType>;
+  alertText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>> |
+    TriggerOutput<Data, MatchType>;
+  alarmText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>> |
+    TriggerOutput<Data, MatchType>;
+  tts?: TriggerFunc<Data, MatchType, PartialTriggerOutput<Data, MatchType>> |
+    TriggerOutput<Data, MatchType>;
+} | undefined;
 // The type of a response trigger field.
 export type ResponseFunc<Data extends RaidbossData, MatchType extends Matches> =
     (data: Data, matches: MatchType, output: Output) => ResponseOutput<Data, MatchType>;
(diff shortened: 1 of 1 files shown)


valarnin:
My patch from the same issue had two fixes:

1. Change `extends Matches` to `extends NetAnyMatches` - This is a constraining change to make the templated type conform to the stricter typing of the known data set
2. Change `ResponseOutput` properties from `TriggerFunc` to `TriggerField` - This is the exact same as your patch except mine didn't duplicate the type defs, instead using the already-created union

I only had one outstanding question on my patch, basically is this valid format for a trigger:

```ts
{
  id: 'ShivaEx Sword Phase',
  type: 'Ability',
  netRegex: NetRegexes.ability({ source: 'Shiva', id: '993', capture: false }),
  response: (data, _matches, output) => {
    // cactbot-builtin-response
      sword: {
        en: 'Sword',
        de: 'Schwert',
        fr: 'Épée',
        ja: '剣',
        cn: '剑',
        ko: '검',
      },
    };

    return { infoText: (_data, _matches, output) => output.sword!() };
  },
  run: (data) => data.soonAfterWeaponChange = true,
},
```

e.g. can `infoText` on the `response` return object be a function? If not, my patch should have been this instead:

```diff
diff --git a/types/trigger.d.ts b/types/trigger.d.ts
index 32ecde8f..86ad5dee 100644
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -1,5 +1,5 @@
 import { Lang, NonEnLang } from '../resources/languages';
-import { NetMatches, Matches } from '../types/net_matches';
+import { NetMatches, NetAnyMatches } from '../types/net_matches';
 import { TimelineReplacement, TimelineStyle } from '../ui/raidboss/timeline';

 import { RaidbossData } from './data';
@@ -35,31 +35,31 @@ export type Output = {
 };

 // The output of any non-response raidboss trigger function.
-export type TriggerOutput<Data extends RaidbossData, MatchType extends Matches> =
+export type TriggerOutput<Data extends RaidbossData, MatchType extends NetAnyMatches> =
     undefined | null | LocaleText | string | number | boolean |
     ((d: Data, m: MatchType, o: Output) => TriggerOutput<Data, MatchType>);

 // Used if the function doesn't need to return an en key
-export type PartialTriggerOutput<Data extends RaidbossData, MatchType extends Matches> =
+export type PartialTriggerOutput<Data extends RaidbossData, MatchType extends NetAnyMatches> =
     undefined | null | Partial<LocaleText> | string | number | boolean |
     ((d: Data, m: MatchType, o: Output) => PartialTriggerOutput<Data, MatchType>);

 // The type of a non-response trigger field.
-export type TriggerFunc<Data extends RaidbossData, MatchType extends Matches, Return> =
+export type TriggerFunc<Data extends RaidbossData, MatchType extends NetAnyMatches, Return> =
     (data: Data, matches: MatchType, output: Output) => Return;

 // The output from a response function (different from other TriggerOutput functions).
-export type ResponseOutput<Data extends RaidbossData, MatchType extends Matches> = {
-  infoText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>>;
-  alertText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>>;
-  alarmText?: TriggerFunc<Data, MatchType, TriggerOutput<Data, MatchType>>;
-  tts?: TriggerFunc<Data, MatchType, PartialTriggerOutput<Data, MatchType>>;
+export type ResponseOutput<Data extends RaidbossData, MatchType extends NetAnyMatches> = {
+  infoText?: TriggerOutput<Data, MatchType>;
+  alertText?: TriggerOutput<Data, MatchType>;
+  alarmText?: TriggerOutput<Data, MatchType>;
+  tts?: PartialTriggerOutput<Data, MatchType>;
 };
 // The type of a response trigger field.
-export type ResponseFunc<Data extends RaidbossData, MatchType extends Matches> =
+export type ResponseFunc<Data extends RaidbossData, MatchType extends NetAnyMatches> =
     (data: Data, matches: MatchType, output: Output) => ResponseOutput<Data, MatchType>;

-export type ResponseField<Data extends RaidbossData, MatchType extends Matches> =
+export type ResponseField<Data extends RaidbossData, MatchType extends NetAnyMatches> =
     ResponseFunc<Data, MatchType> | ResponseOutput<Data, MatchType>;

 export type TriggerAutoConfig = {
@@ -74,7 +74,7 @@ export type TriggerAutoConfig = {

 // Note: functions like run or preRun need to be defined as void-only as (confusingly)
 // it is not possible to assign `(d: Data) => boolean` to a void | undefined, only to void.
-export type TriggerField<Data extends RaidbossData, MatchType extends Matches, Return> =
+export type TriggerField<Data extends RaidbossData, MatchType extends NetAnyMatches, Return> =
   [Return] extends [void] ? TriggerFunc<Data, MatchType, void> :
   TriggerFunc<Data, MatchType, Return | undefined> | Return | undefined;
```

I still feel that mine is the more correct fix for this problem.

Trim21:
I think change `extends NetAnyMatches` didn't fix anything. matches of `response` are always typed.

I just find that `TriggerFunc|TriggerOutput` is just `TriggerField` ...

panicstevenson:
> @panicstevenson can you test it with your TS trigger?

Yep! Seems happy for the Shiva Ex triggers I had mentioned. Thanks!

> (and indent and formatting problems are driving me crazy, why some part of this file is indented with 4 spaces? we should just add prettier with typescript......)

I think the issue here is that we don't have `@typescript-estlint/indent`, so things like types, interfaces, etc. don't have their indentation enforced. Enabling the rule and running it, there were 396 problems currently in the codebase. I haven't inspected each one individually, but they range from good/what I'd expect:

```ts
export type LooseTriggerSet = Exclude<Partial<TriggerSet<RaidbossData>>, 'triggers' | 'timelineTriggers'> & {
  /** @deprecated Use zoneId instead */
  zoneRegex?: RegExp | { [lang in Lang]?: RegExp };
  triggers?: LooseTrigger[];
  timelineTriggers?: LooseTimelineTrigger[];
}
```
to a little odd:
```ts
export type LooseTrigger = Partial<
BaseTrigger<RaidbossData, 'None'> &
PartialRegexTrigger &
PartialNetRegexTrigger<'None'>
>;
```
Although it seems like nothing complains if that's on one line:
```ts
export type LooseTrigger = Partial<BaseTrigger<RaidbossData, 'None'> & PartialRegexTrigger & PartialNetRegexTrigger<'None'>>;
```

Trim21:
maybe just add a formatter...
Let's not use eslint as a formatter anymore

panicstevenson:
> maybe just add a formatter...
> Let's not use eslint as a formatter anymore

Which formatter? We can't use Prettier, and that's the most commonplace one.

panicstevenson:
Other formatters I've tried are https://github.com/dprint/dprint, https://github.com/vvakame/typescript-formatter, and using VSCode's language options, but really none of them have what we need to do what we all want. Prettier is the only one that does that, and it messes with the quotes; if we can maybe solve the quotes or just never format triggers (which seems less than ideal...) then we can use Prettier probably?

quisquous:
> I think the issue here is that we don't have `@typescript-estlint/indent`, so things like types, interfaces, etc. don't have their indentation enforced. Enabling the rule and running it, there were 396 problems currently in the codebase. I haven't inspected each one individually, but they range from good/what I'd expect:

How do you feel about https://github.com/quisquous/cactbot/pull/3164?

quisquous:
> I only had one outstanding question on my patch, basically is this valid format for a trigger.

In my opinion, yes.  I think for the sake of understandability, the return type of `response.infoText` should be identical to the return type of `infoText`.  So, since `infoText` supports functions then `response.infoText` should as well.

PR #3198, opened 2021-06-27
raidboss: Fix typing of timeline property on trigger sets

valarnin:
This PR fixes some issues with the typing of the `timeline` property on trigger sets, which I encountered while converting `00-misc/test.js` to typescript.

### DIFF
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -136,7 +136,8 @@ export type TimelineTrigger<Data extends RaidbossData> = BaseTrigger<Data, 'None
 };
 
 // Because timeline functions run during loading, they only support the base RaidbossData.
-export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
+export type TimelineFunc = (data: RaidbossData) => TimelineField | undefined;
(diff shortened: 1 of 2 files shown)


types/trigger.d.ts
```diff
@@ -136,7 +136,8 @@ export type TimelineTrigger<Data extends RaidbossData> = BaseTrigger<Data, 'None
 };
 
 // Because timeline functions run during loading, they only support the base RaidbossData.
-export type TimelineFunc = string | string[] | ((data: RaidbossData) => TimelineFunc);
+export type TimelineFunc = (data: RaidbossData) => TimelineField | undefined;
+export type TimelineField = string | (string | TimelineFunc)[];
```
quisquous, line 140:
```suggestion
export type TimelineFunc = (data: RaidbossData) => TimelineField;
export type TimelineField = string | (string | TimelineFunc)[] | TimelineFunc | undefined;
```

I am not sure that there's an example, but there's no reason that `timeline` couldn't be a single throwaway func either.  Mostly, I am suggesting it so that this parallels `TriggerFunc`/`TriggerField`.  What do you think?

valarnin: Sounds good to me.

PR #3200, opened 2021-06-27
raidboss: Convert test triggerset to typescript

valarnin:
There was for some strange reason a local version of `ShortName` here which I also cleaned up since it seemed kind of pointless?


quisquous:
> There was for some strange reason a local version of `ShortName` here which I also cleaned up since it seemed kind of pointless?

I don't remember why this was there.  Thanks for cleaning it up.

PR #3204, opened 2021-06-27
ci: fix auto-label on pure renames

quisquous:
See:
https://github.com/quisquous/cactbot/runs/2923687004?check_suite_focus=true

### DIFF
--- a/.github/scripts/auto-label.cjs
+++ b/.github/scripts/auto-label.cjs
@@ -155,10 +155,10 @@ const getChangedFiles = async (github, identifier) => {
  * @param {string} patch
  */
 const parseChangedLang = (patch) => {
-  const set = new Set();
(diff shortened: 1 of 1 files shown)


.github/scripts/auto-label.cjs
```diff
@@ -156,6 +156,9 @@ const getChangedFiles = async (github, identifier) => {
  */
 const parseChangedLang = (patch) => {
   const set = new Set();
+  // For pure renames, there is no patch.
+  if (!patch)
+    return set;
   for (const lang of validLanguages) {
```
Trim21, line 162: Should return empty array I think.

quisquous: Oops, you're totally right.  Thanks!

PR #3205, opened 2021-06-27
raidboss: Fix trigger regex with no matching groups not being passed as matches

### LINKED ISSUE #3203: raidboss.js "matches" returns empty object
## Description

After updating cactbot to the latest version, custom triggers I've written in my raidboss.js file are no longer working. Logging the return data, I'm seeing `matches` is an empty object instead of the regex capture info.
```js
Options.Triggers = [
  {
    zoneRegex: /.*/,
    triggers: [
      {
        regex: /.*/,
          console.log(matches); // Logs {}
          return;
        }
      }
    ]
  }
];
```

valarnin:
Fixes #3203
The old pre-typescript logic was:

```ts
if (matches.groups)
  matches = matches.groups;
```

`matches` was then passed to triggers. When converting to typescript, after hard types were added for these, this exact logic was no longer possible and was erroneously dropped instead of reproduced.

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -887,6 +887,10 @@ export class PopupText {
     // so triggers can do things like matches.target.
     if (matches && matches.groups)
       groups = matches.groups;
+    // If there are no matching groups, reproduce the old js logic where
+    // groups ended up as the original RegExpExecArray object
+    else if (matches)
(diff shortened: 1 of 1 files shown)


valarnin:
This fix is incomplete, let me correct real quick...

quisquous:
Thanks for the fix!

PR #3206, opened 2021-06-27
raidemulator: Convert Persistor to typescript, use Dexie

valarnin:
Edit: Nevermind, export DB works as a full export/import again, though old format exports aren't compatible.


quisquous:
I think in the future I'd say...please don't convert to TypeScript and change all of the functionality at the same time.

valarnin:
> I think in the future I'd say...please don't convert to TypeScript and change all of the functionality at the same time.

This is the one case where I think it was justified. Making this work without Dexie would have involved a lot of forced casting anyways, e.g. pulling objects from IndexedDB will always return `any` and it would have to be cast back to `Encounter`/`PersistorEncounter`. I looked at doing it in two steps, I can't recall off the top of my head but there were at least half a dozen eslint-disable comments required just for loading the encounter, let alone what would have been needed for the export/import functionality.

This was really less of a conversion and more of a rewrite of the functionality entirely that just happened to be going from js to ts.

PR #3208, opened 2021-06-27
raidboss: Convert general triggerset to typescript

valarnin:


ui/raidboss/data/00-misc/general.ts
```diff
@@ -1,23 +1,26 @@
...
+import { TriggerSet } from '../../../../types/trigger';
 
-const caresAboutTankStuff = (data) => {
+const caresAboutTankStuff = (data: RaidbossData) => {
   return data.role === 'tank' || data.role === 'healer' || data.job === 'BLU';
 };
 
 // Triggers for all occasions and zones.
-export default {
+const triggerSet: TriggerSet<RaidbossData> = {
```
panicstevenson, line 11: For files without their own Data, I've been doing

```ts
export type Data = RaidbossData;
```

I don't think I have a preference, but we should standardize on one.

panicstevenson (approved):

ui/raidboss/data/00-misc/general.ts
valarnin: I think it makes more sense to always export a `Data` type after giving it some thought, so that if there are additions later downstream modules will be aware of them. I'll update this in a bit.

quisquous: Yeah, I think Trim21 requested this on some of my initial TypeScript conversions so that it was easier to write user triggers in TypeScript that extended existing triggers.

PR #3209, opened 2021-06-27
lint: add no-extra-parens rule

quisquous:
In playing around with prettier, if we don't want prettier to add parens everywhere, we probably want this rule to maintain (most of) the current behavior.

### DIFF
--- a/ui/raidboss/emulator/EventBus.ts
+++ b/ui/raidboss/emulator/EventBus.ts
@@ -36,7 +36,7 @@ export default class EventBus {
       const events: EventMapEntry[] = this.listeners[event] ??= [];
       if (callback !== undefined)
         events.push({ event: event, scope: scope, callback: callback });
-      ret.push(...(this.listeners[event] ?? []));
+      ret.push(...this.listeners[event] ?? []);
     }
     return ret;
   }
--- a/util/logtools/split_log.js
+++ b/util/logtools/split_log.js
@@ -225,7 +225,7 @@ const printCollectedFights = (collector) => {
       leftExtendStr(row, lengths, 2) + ' ' +
       leftExtendStr(row, lengths, 3) + ' ' +
       rightExtendStr(row, lengths, 4) +
-      (row[5] ? (' ' + '[' + row[5] + ']') : ''));
+      (row[5] ? ' ' + '[' + row[5] + ']' : ''));
   }
 };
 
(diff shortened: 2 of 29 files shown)


util/logtools/split_log.js
```diff
@@ -225,7 +225,7 @@ const printCollectedFights = (collector) => {
       leftExtendStr(row, lengths, 2) + ' ' +
       leftExtendStr(row, lengths, 3) + ' ' +
       rightExtendStr(row, lengths, 4) +
-      (row[5] ? (' ' + '[' + row[5] + ']') : ''));
+      (row[5] ? ' ' + '[' + row[5] + ']' : ''));
```
panicstevenson, line 228: Not a logical error or anything, but can this be updated to:
```suggestion
      (row[5] ? ` [${row[5]}]` : ''));
```

panicstevenson (approved):

ui/raidboss/emulator/EventBus.ts
```diff
@@ -36,7 +36,7 @@ export default class EventBus {
       const events: EventMapEntry[] = this.listeners[event] ??= [];
       if (callback !== undefined)
         events.push({ event: event, scope: scope, callback: callback });
-      ret.push(...(this.listeners[event] ?? []));
+      ret.push(...this.listeners[event] ?? []);
```
valarnin, line 39: This is the only change here I'm iffy on. I assume this is still functionally equivalent due to operator precedence but unless you're really familiar with operator precedence here, it looks like it could be different.

quisquous: Oof, yeah.  This is not the best.

I think mostly I put this up, because if we go with #3211, then the options are either (1) use prettier's parentheses (which really wants to add parens on everything, esp stuff like `run: (data) => (data.putSomeParensOnThis = 3),` or (2) add an eslint rule to only have parens on some things.  I don't think there's an option in eslint to support this particular use case, so we'd have to rewrite to an if statement if we wanted more clarity.  Unfortunate.

valarnin: I think it's fine to rewrite this as an if case, I just wouldn't want to come back to it in 6 months like this and wonder what the original intent was.

quisquous: Definitely agreed.  If I end up landing this, I will fix that up beforehand.  I think I will only land this if we go with #3211.  If we go with #3212 instead, then I will close this.

quisquous:
Closing this in favor of #3212

PR #3193, opened 2021-06-26
style: add prettier

Trim21:
this is a draft pr for #3191

only file with `@prettier` will be formatted and checked by prettier

PR #3211, opened 2021-06-27
lint: add prettier-eslint

quisquous:
I wanted to see what this looks like across the whole codebase and in case anybody else wants to take a look.  (Probably?) it doesn't make sense to land all of the formatting changes in one PR.  This includes #3209 and #3207 as some cleanup prior to converting. This is an alternative to #3193, because I wanted to try out prettier-eslint myself (sorry).

Missing stuff:
* convert users of programmatic eslint api to prettier-eslint
* add prettier-eslint plugin to VSCode
* probably need to do something about resource/ files which now generate with lint errors

# Pros

On the positive side, personally, I think the results of this look ok.  I think functionally this is the same as @Trim21's #3193 patch in that it runs prettier and then eslint, but using prettier-eslint/prettier-eslint-cli rather than a custom formatter.  Please take a look and tell me if you think anything looks awful, because anything that changes here will also change in Trim21's patch so it'd be nice to get this right.  Most of the changes are prettier being prettier in a lot of cases, but I think the changes are fine/better/not awful.

# Cons

On the negative side, WOOF this is slow, about three times slower.  Go make a coffee.

### existing eslint
```
$ time npm run lint

> cactbot@ lint C:\Users\enne\cactbot
> eslint --max-warnings=0 ./


real    0m53.033s
user    0m0.015s
sys     0m0.105s
```

### prettier-eslint

This patch.  I get this time whether or not the files are formatted, so there's no hypothetical "one time cost" where it would be faster in the future.

```
$ time npm run lint

> cactbot@ lint C:\Users\enne\cactbot
> prettier-eslint --write **/*.{js,ts,cjs,mjs}

success formatting 186 files with prettier-eslint
279 files were unchanged

real    3m14.008s
user    0m0.045s
sys     0m0.076s
```

### Trim21's formatter prettier-eslint
This is unfair because it has to invoke node every time, but I didn't know how else to compare it.  This is with removing the `requirePragma` lines and fixing up the `.prettierignore` to match the one in this patch so that it doesn't format bad files.  I could run this again by hacking in the file list to `formatter.cjs` so it doesn't have to leave `node`, but I would expect it's going to be on par with prettier-eslint, because it's doing the same work.

```
$ time (find eslint/ util/ ui/ resources/ types/ -type f | grep "\.\(js\|ts\|mjs\|cjs\)$" | xargs -i node util/formatter.cjs {})


real    37m40.762s
user    0m2.009s
sys     0m5.499s
```

# Questions

* Is this too slow to inflict on everybody? This seems like a lot.

* If the issue is that we really want better TypeScript formatting, do we only run prettier-eslint on `.d.ts` files and leave the rest, and hope we don't have too many complicated line-wrapping type definitions in `.ts` files? (This is awkward, but would certainly be faster.)

* Can we get husky to only apply prior to pushing vs committing?

* Is dprint a better option? Prior to landing any formatter, I'd like to investigate https://github.com/dprint/dprint more, especially with the speed question.


quisquous:
Closing this in favor of #3212

PR #3213, opened 2021-06-28
set `locale` of TimelineReplacement to `Lang`

Trim21:
```diff
diff --git a/ui/raidboss/timeline.ts b/ui/raidboss/timeline.ts
index 0746284360..809210bd7e 100644
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -75,7 +75,7 @@ const activeText = {
 };
 
 export type TimelineReplacement = {
-  locale: string;
+  locale: Lang;
   missingTranslations?: boolean;
   replaceSync?: { [regexString: string]: string };
   replaceText?: { [timelineText: string]: string };
```

### DIFF
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -75,7 +75,7 @@ const activeText = {
 };
 
 export type TimelineReplacement = {
-  locale: string;
+  locale: Lang;
   missingTranslations?: boolean;
   replaceSync?: { [regexString: string]: string };
   replaceText?: { [timelineText: string]: string };


valarnin:
Makes sense to me. Another spot that might make sense to update:

data.d.ts:8, `ShortLocale`, `string` => `Lang`

Will need some additional updates in `user_config.ts`, basically store off in a temp var and check `isLang` before assigning.

quisquous:
> Makes sense to me. Another spot that might make sense to update:
> 
> data.d.ts:8, `ShortLocale`, `string` => `Lang`
> 
> Will need some additional updates in `user_config.ts`, basically store off in a temp var and check `isLang` before assigning.

That seems reasonable, but also seems like a separate change.

PR #3214, opened 2021-06-29
raidboss: Require initData if data type contains non-optional properties

valarnin:
This PR changes TriggerSet to require initData if the data type passed in contains non-optional properties that are not part of the RaidbossData type.

Example:

image

### DIFF
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -143,7 +143,11 @@ export type DataInitializeFunc<Data extends RaidbossData> = () => Omit<Data, key
 
 export type DisabledTrigger = { id: string; disabled: true };
 
-type Required<Type> =
+// This helper takes all of the properties in Type and checks to see if they can be assigned to a
+// blank object, and if so excludes them from the returned union. The `-?` syntax removes the
+// optional modifier from the attribute which prevents `undefined` from being included in the union
+// See also: https://www.typescriptlang.org/docs/handbook/2/mapped-types.html#mapping-modifiers
+type RequiredFieldsAsUnion<Type> =
   {
     [key in keyof Type]-?: Record<string, never> extends Pick<Type, key> ? never : key
   }[keyof Type];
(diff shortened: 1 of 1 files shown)


types/trigger.d.ts
```diff
@@ -154,9 +159,19 @@ export type TriggerSet<Data extends RaidbossData> = {
...
+// If Data contains required properties that are not on RaidbossData, require initData
+export type TriggerSet<Data extends RaidbossData> = BaseTriggerSet<Data> &
+  (Required<RaidbossData> extends Required<Data> ? Required<Data> extends Required<RaidbossData> ?
+  {
+    initData?: DataInitializeFunc<Data>;
+  } : {
+    initData: DataInitializeFunc<Data>;
+  } : {
+    initData: DataInitializeFunc<Data>;
+  });
```
quisquous, line 173: I think you have an extra ternary.  My understanding of `extends` is that `X extends Y` means X can be assigned to Y.  In this case, the types are unions, and so really you just want to check that `Data` does not have any extra properties that `RaidbossData` doesn't have.  `export type TriggerSet<Data extends RaidbossData>` already verifies that `Data` has all the properties that `RaidbossData` has, I think?

```suggestion
export type TriggerSet<Data extends RaidbossData> = BaseTriggerSet<Data> &
  (Required<Data> extends Required<RaidbossData> ?
  {
    initData?: DataInitializeFunc<Data>;
  } : {
    initData: DataInitializeFunc<Data>;
  });
```

types/trigger.d.ts
```diff
@@ -143,7 +143,12 @@ export type DataInitializeFunc<Data extends RaidbossData> = () => Omit<Data, key
 
 export type DisabledTrigger = { id: string; disabled: true };
 
-export type TriggerSet<Data extends RaidbossData> = {
+type Required<Type> =
+  {
+    [key in keyof Type]-?: Record<string, never> extends Pick<Type, key> ? never : key
```
quisquous, line 148: Can this get a comment as to what it's doing? I don't think I've ever seen `-?` syntax before.

types/trigger.d.ts
valarnin: Yeah, this makes sense. I think it might also be a good idea to store `type RequiredRaidbossDataFields = Required<RaidbossData>` as well, since this could cause some compile-time performance impact once a ton of triggers are converted?

quisquous: I do not know enough about tsc to know if "storing" saves work here.  My instinct is that this is probably premature optimization, especially for a once-per-triggerset sort of type.

types/trigger.d.ts
valarnin: Comment added, I think it explains things pretty thoroughly?

PR #3218, opened 2021-06-30
lint: add a few more no-unsafe-etc rules

quisquous:


panicstevenson:
Are these not covered by TypeScript itself?

quisquous:
> Are these not covered by TypeScript itself?

Are they? What makes you say that?

quisquous:
> > Are these not covered by TypeScript itself?
> 
> Are they? What makes you say that?

OH they're all in https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended-requiring-type-checking.ts

PR #3226, opened 2021-07-02
lint: add no-unsafe-argument

quisquous:


ui/raidboss/emulator/data/NetworkLogConverter.worker.ts
```diff
@@ -1,17 +1,26 @@
...
 
 ctx.addEventListener('message', (msg) => {
+  // TODO: should this just throw an exception? Or return an error?
+  if (!(msg.data instanceof ArrayBuffer)) {
+    ctx.postMessage({
+      type: 'done',
+    });
+    return;
+  }
+
```
valarnin, line 17: This is one of those should-not-happen situations, so it should probably at least log an UnreachableCode to console.

Maybe it makes more sense to postMessage an error instead, and modify raidemulator.js to handle the error case?

quisquous: I think posting an error would be more change than I want in this PR, since there aren't other examples of handling that.  I'll at least log `UnreachableCode` for now with a TODO.

PR #3230, opened 2021-07-03
raidemulator: Convert raidemulator.js to typescript

valarnin:
There was some weirdness with importing NetworkLogConverterWorker in typescript that I had to work around. Couldn't figure out a better way that didn't involve disabling eslint rules.


ui/raidboss/raidemulator.ts
```diff
@@ -345,41 +373,48 @@ document.addEventListener('DOMContentLoaded', async () => {
...
     n.addEventListener('click', (e) => {
+      if (!(e instanceof MouseEvent))
+        return;
+      if (!(e.currentTarget instanceof HTMLElement))
+        return;
       // Find the parent modal from the close button and close it
       let target = e.currentTarget;
       while (!target.classList.contains('modal') && target !== document.body)
-        target = target.parentElement;
+        target = target.parentElement || target;
```
quisquous, line 417:
```suggestion
        target = target.parentElement ?? target;
```

ui/raidboss/emulator/data/NetworkLogConverter.worker.ts
```diff
@@ -83,3 +102,9 @@ ctx.addEventListener('message', (msg) => {
     type: 'done',
   });
 });
+
+// For typescript to pick up the proper typings here, have to do some weird casting for export
+const expVar = {};
+const exp = expVar as typeof Worker & (new () => Worker);
```
quisquous, line 108:
```suggestion
const exp = expVar as (new () => Worker);
```

I think you don't need the intersection, at least.

However, what's actually confusing to me is that it feels like `types/worker.d.ts` should cover this.  Why is the module declaration not handling the type when importing this file? Is it because there's an additional export? Is it not applying?

ui/raidboss/raidemulator.ts
```diff
@@ -421,4 +456,9 @@ document.addEventListener('DOMContentLoaded', async () => {
     emulatedWebSocket: emulatedWebSocket,
     timelineUI: timelineUI,
   };
+};
+
+// eslint-disable-next-line @typescript-eslint/no-misused-promises
```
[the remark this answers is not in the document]
valarnin: This one and the one below were leftover from the conversion process that I forgot to remove.

I wish something like this was part of built-in eslint... https://github.com/mysticatea/eslint-plugin-eslint-comments/blob/master/docs/rules/no-unused-disable.md

ui/raidboss/emulator/data/NetworkLogConverter.worker.ts
valarnin: I think it might be because the file actually exists,  so it's taking priority from the actual file instead of the `types/worker.d.ts` declaration, but I'm not sure. The "fix" for this is to use the `import ... from 'worker-loader!...';` syntax, but then eslint complains about using custom loader syntax.

I removed the `typeof worker` intersection since it does seem to work fine without it, but if you can spot a way to make this work fully I'm all for it.

quisquous: Yeah, I can't figure out how to make it work.  I think it's because the file both exists but is also something that looks like it can be imported (vs manifest.txt which does exist too but isn't a normal import type).  Should `worker.d.ts` be removed, then?

Could you also make the comment explain a bit more, like: `// Export a variable that is the same type that webpack's worker-loader will generate.  This value will not be used.`

valarnin: Since emulator doesn't work from file URIs anyways, I think this can be significantly simplified and cleaned up since we don't need to inline the worker. Check latest commit.

Tested it as working with `npm run start` at least, working fine there.

quisquous: Oh nice!

PR #3245, opened 2021-07-05
refactor: prefix setTimeout with window global

MaikoTan:
For
1) get the correct return type
2) this should be run in browser
3) differentiate with user-defined functions (although there isn't corrently)

### DIFF
--- a/ui/jobs/components/drk.ts
+++ b/ui/jobs/components/drk.ts
@@ -4,9 +4,9 @@ import { kAbility } from '../constants';
 import { computeBackgroundColorFrom } from '../utils';
 
 let resetFunc: ((bars: Bars) => void) = (_bars: Bars) => undefined;
-let tid1: ReturnType<typeof setTimeout>;
-let tid2: ReturnType<typeof setTimeout>;
-let tid3: ReturnType<typeof setTimeout>;
+let tid1: number;
+let tid2: number;
+let tid3: number;
 
 export const setup = (bars: Bars): void => {
   const bloodBox = bars.addResourceBox({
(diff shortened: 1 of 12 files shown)


ui/jobs/components/drk.ts
```diff
@@ -4,9 +4,9 @@ import { kAbility } from '../constants';
 import { computeBackgroundColorFrom } from '../utils';
 
 let resetFunc: ((bars: Bars) => void) = (_bars: Bars) => undefined;
-let tid1: ReturnType<typeof setTimeout>;
-let tid2: ReturnType<typeof setTimeout>;
-let tid3: ReturnType<typeof setTimeout>;
+let tid1: number;
```
quisquous, line 7: I had this same conversation several times with @Trim21 in https://github.com/quisquous/cactbot/pull/3001#discussion_r647138823 (and earlier in that PR).

I agree that we are only going to run this in the browser though, so this could/should just be `number`.

quisquous:
This looks good to me, although your other patch conflicts with it.

Trim21:
I have some other words about these issues that types are different in browser and node.

`console.assert` only accept boolean as it's first parameters, but we have a `Node` in our tsconfig. so `console.assert` could accept anything. maybe that's a problem, too

ui/jobs/components/drk.ts
MaikoTan: It turns out that we may need to differentiate/separate tsconfig for files run in node and browser, but are there some files run in both node and browser?

quisquous: I think my general feeling is that files are either:
* browser only (such as this file, which is dealing with html and window.setTimeout)
* node only (like utilities, which depend on fs or path or node things)
* both (which don't depend on either)

I don't know of code that we're using in node and browser that is using node/browser specifics.

quisquous: Mmmm ok I think that's not technically true.  There is stuff like `timeline.ts` that has some browser stuff in it that we use from some node files, but we don't use any of the browser code, but it does get imported.

quisquous:
I think this is still good to land for now.  @Trim21 do you think so, or do you think this needs to be changed?

Trim21:
LGTM

quisquous:
> LGTM

Thanks! Sometimes it's hard to tell between "this is something on my mind to fix later" or "this should get fixed here".

PR #3250, opened 2021-07-05
oopsy: convert top level file to TS

quisquous:


ui/oopsyraidsy/oopsyraidsy.ts
```diff
@@ -1,41 +1,46 @@
...
-import oopsyFileData from './data/oopsy_manifest.txt';
-
 import '../../resources/defaults.css';
 import './oopsy_common.css';
 
 UserConfig.getUserConfigLocation('oopsyraidsy', defaultOptions, () => {
   const options = { ...defaultOptions };
-  let listView;
-  let mistakeCollector;
+  let initListView: OopsyListView | undefined;
```
panicstevenson, line 20: This feels weird, is the `| undefined` really necessary if it's a let?

panicstevenson (approved):

ui/oopsyraidsy/oopsyraidsy.ts
quisquous: Oh you're right, thanks!

PR #3251, opened 2021-07-05
oopsy: remove event field from all triggers

quisquous:
The event types came before NetRegexes and named match groups
and were trying to do similar things.  Now that we have them,
they're redundant and make the code a lot more complicated.
They also have different names for the fields, which creates
confusions and errors.

This will also make oopsy code more consistent with raidboss.

Obviously, this is a breaking change for user triggers.

### DIFF
--- a/docs/OopsyraidsyGuide.md
+++ b/docs/OopsyraidsyGuide.md
@@ -140,7 +140,7 @@ mistake: (data, matches) => {
 If this following trigger is used, then if a player dies without taking any other damage, the log would show ":skull: Chippy: Doom Debuff" instead of assigning it to the last damage the player took before this trigger, which might incorrectly look more like ":skull: Chippy: Auto (3034/38471)".
 
 ```javascript
-deathReason: function(event, data, matches) {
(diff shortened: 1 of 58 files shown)


docs/OopsyraidsyGuide.md
```diff
@@ -233,7 +185,7 @@ For example, if you want to store a map of which players have doom or not, that
 ```javascript
 {
   netRegex: NetRegexes.gainsEffect({ effect: 'Doom' }),
-  run: function(_e, data, matches) {
+  run: function(data, matches) {
```
panicstevenson, line 188: (Can this be made an arrow function?)

panicstevenson (approved):
Cursory review since I'm assuming something else tested the syntax.

docs/OopsyraidsyGuide.md
quisquous: (Done, thanks!)

quisquous:
> Cursory review since I'm assuming something else tested the syntax.

I did some manual testing of both buffs and a few fights.

PR #3253, opened 2021-07-06
oopsy: convert simple triggersets to TS

quisquous:


ui/oopsyraidsy/data/02-arr/trial/ifrit-nm.ts
```diff
@@ -1,7 +1,10 @@
 import ZoneId from '../../../../../resources/zone_id';
+import { SimpleOopsyTriggerSet } from '../../../../../types/oopsy';
+
+export { OopsyData as Data } from '../../../../../types/data';
 
 // Ifrit Story Mode
-export default {
+const triggerSet: SimpleOopsyTriggerSet = {
```
panicstevenson, line 7: Did we want to have these be `OopsyTriggerSet<Data>` so that in theory you don't have to update the types/imports if you end up needing data?

panicstevenson (approved):

ui/oopsyraidsy/data/02-arr/trial/ifrit-nm.ts
quisquous: (Why disregard? That seems like a reasonable suggestion?)

panicstevenson: Oh, because after this PR I went to the no data PR and that one looks the same, but then there's simple which I feel like is maybe intentionally simple? I figure for the reason of not wanting to check triggers everywhere on the backend?

quisquous: I can see an argument that having some things be `SimpleOopsyTriggerSet` is more confusing and less consistent vs everything being `OopsyTriggerSet<Data>`.  I don't think there's any backend reason, I just had this type lying around and it seemed convenient.

panicstevenson: Yea, I'm definitely pro-everything-`OopsyTriggerSet<Data>` if it's all the same.

quisquous: Ok, made that change before landing this.

PR #3254, opened 2021-07-06
oopsy: convert triggersets that don't use data to TS

quisquous:


ui/oopsyraidsy/data/05-shb/trial/wol-ex.ts
```diff
@@ -42,10 +47,11 @@ export default {
     },
     {
       id: 'WOLEx Tower',
+      type: 'Ability',
       netRegex: NetRegexes.ability({ id: '4F04', capture: false }),
       mistake: {
         type: 'fail',
-        reason: {
+        text: {
```
panicstevenson, line 54: Why is this text here but reason in the case of `WOLEx Missed Interrupt` below?

panicstevenson (approved):

ui/oopsyraidsy/data/05-shb/trial/wol-ex.ts
quisquous: This turned into `text` because `mistake` here is an `OopsyMistake` and `reason` is not a field.  Below, an anonymous object is being returned as an `OopsyMistake`, and `{ type: 'fail', reason: 'foobar' } extends OopsyMistake` and so it's not an error.

I could add `reason?: never` to `OopsyMistake`.  I could also just rename `reason` to `text` while I'm making breaking changes.  That might be better (although I'd do it in a separate PR).

panicstevenson: I'm pro the latter with this batch of breakage

PR #3257, opened 2021-07-06
fix(config): display English when display language set to default

### LINKED ISSUE #3256: Cactbot language is not set properly.
K-003

As you see in the log, the parser language is "Korean". But Cactbot UI is shown in English, and this includes Alerts.

MaikoTan:
Fixes #3256


quisquous:
Thank you so much for the quick fix!!

PR #3260, opened 2021-07-06
[wip] hacky raidboss js -> ts converter script

quisquous:
Example usage:
node convert_raidboss_to_ts.js ui/raidboss/data/03-hw/dungeon/the_vault.js

This is not meant to be checked in, but maybe somebody else will find this useful.  You still need to go back and fix up the type of `Data` in each file, but it does take care of a lot of manual fixups.


quisquous:
Everything's converted, so this is no longer needed.

PR #3262, opened 2021-07-06
raidboss: add Doma Castle timeline

rachelwalsh:
Couple things I'm not happy with on Grynewaht:
- The exact ordering on the simultaneous Delay-Action Charge and Clean Cut varies from pull to pull, so I just combined them in the timeline to avoid any missed syncs. The name is a little long, though, and `Cut` gets cut off with default timeline bar width. Any preference on abbreviation here, or even just leaving it as is?
- `make_timeline.py` inserted a fair number of `--sync--` entries for ID 20B0, during testing I couldn't see anything visible corresponding to this, but it was consistent across pulls. Unsure if these should be removed or not.

### DIFF
--- a/ui/raidboss/data/04-sb/dungeon/doma_castle.txt
+++ b/ui/raidboss/data/04-sb/dungeon/doma_castle.txt
@@ -8,42 +8,42 @@ hideall "--sync--"
 0 "Start" sync / 00:0839:The Third Armory will be sealed off/
 7.4 "Cermet Pile" sync /:Magitek Rearguard:209D:/
 15.0 "Garlean Fire" sync /:Magitek Rearguard:209E:/
-30.4 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-35.4 "Mines" sync /03:........:Added new combatant Rearguard Mine/
-37.6 "Cermet Pile" sync /:Magitek Rearguard:209D:/
-40.6 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-43.2 "Garlean Fire" sync /:Magitek Rearguard:209E:/
-48.7 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-53.7 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-60.7 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+30.6 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+35.6 "Rearguard Mines" #sync /03:........:Added new combatant Rearguard Mine/
+37.8 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+40.8 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+43.4 "Garlean Fire" sync /:Magitek Rearguard:209E:/
+48.9 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+53.9 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+60.9 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+68.5 "Rearguard Mines" #sync /03:........:Added new combatant Rearguard Mine/
 
-68.3 "Mines" sync /03:........:Added new combatant Rearguard Mine/
-70.5 "Cermet Pile" sync /:Magitek Rearguard:209D:/
-73.5 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-76.2 "Garlean Fire" sync /:Magitek Rearguard:209E:/
-81.7 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-85.8 "Cermet Pile" sync /:Magitek Rearguard:209D:/
-88.7 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-91.4 "Garlean Fire" sync /:Magitek Rearguard:209E:/
-97.0 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-98.5 "Mines" sync /03:........:Added new combatant Rearguard Mine/
-101.1 "Cermet Pile" sync /:Magitek Rearguard:209D:/
-104.1 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-106.8 "Garlean Fire" sync /:Magitek Rearguard:209E:/
-112.3 "Magitek Ray" sync /:Rearguard Bit:20A1:/
-115.4 "Cermet Pile" sync /:Magitek Rearguard:209D:/
-118.4 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+70.7 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+73.7 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+76.4 "Garlean Fire" sync /:Magitek Rearguard:209E:/
+81.9 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+86.0 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+88.9 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+91.6 "Garlean Fire" sync /:Magitek Rearguard:209E:/
+97.2 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+98.7 "Rearguard Mines" #sync /03:........:Added new combatant Rearguard Mine/
+101.3 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+104.3 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+107.0 "Garlean Fire" sync /:Magitek Rearguard:209E:/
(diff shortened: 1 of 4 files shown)


valarnin:
`20B0` would be the boss's autoattack, see:
https://xivapi.com/Action/8368

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
```diff
@@ -0,0 +1,137 @@
...
+98.5 "Mines" sync /03:........:Added new combatant Rearguard Mine/
+101.1 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+104.1 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+106.8 "Garlean Fire" sync /:Magitek Rearguard:209E:/
+112.3 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+115.4 "Cermet Pile" sync /:Magitek Rearguard:209D:/
+118.4 "Magitek Ray" sync /:Rearguard Bit:20A1:/
+
+131.1 "Mines" sync /03:........:Added new combatant Rearguard Mine/ jump 68.3
+132.0 "Cermet Pile"
```
quisquous, line 38: The difference between Mines and Cermet Pile at 68.3 is 2.2 seconds, but is 0.9 seconds at 131.1.  Is `Cermet Pile` the loop and not `Mines`? What does `test_timeline` say about the amount it jumps when you run it against several logs?

Also, this it the tiniest of code review nits, but for loops can you run `node --loader ts-node/esm util/timeline_adjust.mjs -f ui/raidboss/data/04-sb/dungeon/doma_castle.txt -a 62.8` and then use the times from `68.3` that have been moved forward 62.8 seconds (131.1 - 68.3) in time so that the loop lookahead has consistent times with where it's jumping?  In the future I'd like to just add a `loop` command to timeline so that loops don't have to be manual, but we're not there yet.

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
```diff
@@ -0,0 +1,137 @@
...
+158.2 "Magitek Ray"
+160.1 "Mines"
+
+### Magitek Hexadrone
+# -ii 20A4 20A6 20A7 2447
+1000 "Start" sync /00:0839:The Training Grounds will be sealed off/ window 1000,0
+1010.6 "Circle Of Death" sync /:Magitek Hexadrone:20A2:/
+1018.9 "2-Tonze Magitek Missile" sync /:Magitek Hexadrone:20A3:/
+1024.1 "Hexadrone Bits" sync /03:........:Added new combatant Hexadrone Bit/
+1029.6 "Bits Activate"
```
quisquous, line 54: Is there an ability name for the bits activating or the damage you take? It's fine if no, mostly just curious.  :)

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
```diff
@@ -0,0 +1,137 @@
...
+2202.4 "Gunsaw" duration 5.0 sync /:Hypertuned Grynewaht:20AA:/
+2215.7 "Delay-Action Charge/Clean Cut" sync /:Hypertuned Grynewaht:20AD:/
+2218.2 "Hexadrone Bits" sync /03:........:Added new combatant Hexadrone Bit/
+2223.7 "Bits Activate"
+2228.7 "Chainsaw" duration 5.5 sync /:Hypertuned Grynewaht:20A8:/
+2236.4 "Thermobaric Charge" sync /:Hypertuned Grynewaht:20AF:/
+2242.8 "Chainsaw" duration 5.5 sync /:Hypertuned Grynewaht:20A8:/
+
+2257.0 "--sync--" sync /:Hypertuned Grynewaht:20B0:/
+# Loop begins here, but the jump is attached to the Delay-Action/Clean Cut to preserve the duration timer for Gunsaw
```
quisquous, line 130: Ah, thanks for the comment! Sorry about the duration not preserving.  We should fix that too.  ;;

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
```diff
@@ -0,0 +1,137 @@
...
+1181.3 "Circle Of Death" sync /:Magitek Hexadrone:20A2:/ jump 1083.9
+1187.4 "Hexadrone Bits"
+1192.9 "Bits Activate"
+1195.0 "Magitek Missiles"
+1203.8 "2-Tonze Magitek Missile"
+1216.0 "Magitek Missiles"
+
+
+### Hypertuned Grynewaht
+# -ii 20A9 20AB 20AC 20AE 20A7 2447
```
quisquous, line 90: > 20B0 would be the boss's autoattack, see:
> https://xivapi.com/Action/8368?pretty=true

Is that true? I'm not sure what I see what makes that an auto.  It also only happens much less frequently than I would expect autos to happen.

> make_timeline.py inserted a fair number of --sync-- entries for ID 20B0, during testing I couldn't see anything visible corresponding to this, but it was consistent across pulls. Unsure if these should be removed or not.

Personally, I think these are fine to leave in, as (in this case) they might help the timeline stay "on track" with syncs because they're infrequent and show up before major mechanics.

If you'd like to remove it, can you add it to the `-ii` line too?

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
```diff
@@ -0,0 +1,137 @@
...
+2000 "Start" sync /00:0839:The Hall Of The Scarlet Swallow will be sealed off/ window 2000,0
+2009.4 "Chainsaw" duration 5.5 sync /:Hypertuned Grynewaht:20A8:/
+2022.7 "Delay-Action Charge" sync /:Hypertuned Grynewaht:20AD:/
+2023.1 "Gunsaw" duration 5.0  sync /:Hypertuned Grynewaht:20AA:/
+2045.4 "Thermobaric Charge" sync /:Hypertuned Grynewaht:20AF:/
+2048.8 "--sync--" sync /:Hypertuned Grynewaht:20B0:/
+2057.0 "Gunsaw" duration 5.0  sync /:Hypertuned Grynewaht:20AA:/
+2070.3 "Delay-Action Charge/Clean Cut" sync /:Hypertuned Grynewaht:20AD:/
```
quisquous, line 98: Is there a different ability id for `Delay-Action Charge` and `Clean Cut`?

> The exact ordering on the simultaneous Delay-Action Charge and Clean Cut varies from pull to pull, so I just combined them in the timeline to avoid any missed syncs.

Is this one of those situations where he will either do one or the other mechanic, but it's a single ability and it happens at the same time regardless?  (Similar to Shield Omen / Sword Omen here https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/eureka/delubrum_reginae.txt#L338.)

> The name is a little long, though, and Cut gets cut off with default timeline bar width. Any preference on abbreviation here, or even just leaving it as is?

In general, I'd suggest leaving the full ability names in the timeline itself.  This helps make translations easier because they can translate full ability names and not have to worry about English abbreviations.  You can add a `timelineReplace` section like https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/eureka/delubrum_reginae.js#L1421-L1430 to replace them dynamically with something shorter.  Maybe, `Delayed Charge/Clean Cut` or something else?

quisquous (commented):
Ohmygosh, thank you for this! I have a bunch of comments (sorry), but this is a really high quality timeline submission and is very nearly ready to be merged.

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
rachelwalsh: There isn't. Activating the tripwire has an ID, but that doesn't actually log until a second or so after triggering it. This timestamp is just my best guess at when the hitbox becomes active.

ui/raidboss/data/04-sb/dungeon/doma_castle.txt
rachelwalsh: It's two abilities that happen at almost the same time (0.2 seconds apart, in variable order) In play, it's essentially a single mechanic resolution (dodge chakram ground AoEs while not stacking the Delay-Action player-AoEs) but it comes from two different actors and they aren't perfectly synced up.

quisquous: Oh! In that case, I'd add two separate timeline lines at the same time and comment out one of the syncs.  If one of the syncs is on `Grynewaht`, then I'd keep that one and comment out the other.  In general, other actors can definitely be inconsistent.

rachelwalsh:
Alright, timelines have all been tightened up. Looks like none of the add spawn timers are terribly consistent (shocking!) so I removed the syncs on them for Rearguard. Unfortunately they're probably a necessary evil for Hexadrone/Grynewaht to accurately show tether activation times.

I also tried to convert the empty trigger file to TypeScript, local testing worked fine but I may have missed something for that.

quisquous:
Thanks for all the changes!  It all looks great!

PR #3273, opened 2021-07-07
Update drowned_city_of_skalla.js

Avanna13:
better french translation


quisquous:
Oops, I missed that this specified it was being merged into the `triggers` branch and not `main as it should have been.  I'll fix it.

PR #3276, opened 2021-07-07
raidboss: Add 03-hw/raid initData

panicstevenson:


test/helper/test_trigger.js
```diff
@@ -176,7 +176,8 @@ const testTriggerFile = (file) => {
             }
             // Built-in response functions can be safely called once.
             const output = new TestOutputProxy(trigger, {});
-            currentTriggerFunction = currentTriggerFunction({}, {}, output);
+            const data = triggerSet.initData ? triggerSet.initData() : {};
```
panicstevenson, line 179: Without this, you get the following:
```
  1) trigger test
       ui/raidboss/data/03-hw/raid/a10s.ts
         has valid matches and output parameters:
     TypeError: Cannot read property 'length' of undefined
```

PR #3277, opened 2021-07-08
raidboss: convert e7s to TS

quisquous:


ui/raidboss/data/05-shb/raid/e7s.ts
```diff
@@ -407,13 +434,16 @@ export default {
         if (data.boundless.light === data.boundless.dark) {
           if (matches.target === data.me)
             return;
-          return { infoText: output.text({ player: data.ShortName(matches.target) }) };
+          return { infoText: output.text!({ player: data.ShortName(matches.target) }) };
         }
+
+        // Commenting out this return makes the error go away.
         return Responses.stackMarkerOn();
```
quisquous, line 441: 🧙 ☎️ @valarnin

valarnin: I'm a bit baffled about why this is happening but this alternative is a workaround while I'm researching:

```ts
        return Responses.stackMarkerOn()(data, matches, output);
```

valarnin: Minimal repro shows that it's only a problem when returning a function in `response`:

```ts
    {
      id: 'test',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing(),
      response: (_data: Data, _matches: NetMatches['StartsUsing'], _output: Output): ResponseFunc<Data, NetMatches['StartsUsing']> => {
        const t = '' as unknown as ResponseFunc<Data, NetMatches['StartsUsing']>;
        return t;
      },
    },
```

valarnin: This adjustment in `trigger.d.ts` fixes the issue I believe, if you want to include it in this PR?

```ts
export type ResponseFunc<Data extends RaidbossData, MatchType extends NetAnyMatches> =
    (data: Data, matches: MatchType, output: Output) =>
    (ResponseOutput<Data, MatchType> | ResponseFunc<Data, MatchType>);
```

quisquous: Aaaah thank you.  I stared at this for a while, and went down a rabbit hole of thinking `Matches` was incorrect based on the compiler errors.  I'll just include this in the "convert shb raids" patch.

PR #3278, opened 2021-07-08
raidboss: Enforce Non-TriggerOutput Return Defined

panicstevenson:
For all non-(Partial)TriggerOutputs, ensure the return value is defined
if the field is defined. The exception being conditions, potentially
only for the time being if we decide that all trigger data booleans
should be defined within initData.

This only applies to built in triggers, as popup text still handles the
undefined case for user triggers and overrides.


valarnin:
These changes make sense to me.

Given that PopupText is handling the case for user triggers, it makes sense to me to enforce returning the right type in conditions as well.

ui/raidboss/data/03-hw/raid/a11s.ts
```diff
@@ -258,7 +258,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'HeadMarker',
       netRegex: NetRegexes.headMarker({ id: '00(?:4F|5[0-6])' }),
       condition: Conditions.targetIsYou(),
-      durationSeconds: (data) => data.limitCutDelay,
+      durationSeconds: (data) => data.limitCutDelay ?? 0,
```
quisquous, line 261:
```suggestion
      durationSeconds: (data) => data.limitCutDelay ?? 3,
```

Maybe if something has gone awry, it should still show up in general? I think previously returning `undefined` here meant "use the default".

quisquous (commented):
I think this seems good to me, and also seems fine for `condition` if that makes sense too.  I think there's some argument for condition supporting `undefined` making optional `boolean` fields on `data` easier to handle, but only supporting `boolean` might make things more clear too.  To sum up, if y'all think it's good, then I think that change sounds fine to me too.

ui/raidboss/data/03-hw/raid/a11s.ts
panicstevenson: Hmmm; yea, looking at the code that definitely appears to be the case. Is there a point to enforcing a 0 second duration? As in, should the code here stay the same, but then in popup text we change it from `duration === undefined` to `!duration` so that 0 isn't supported?
I feel hesitant because that's a change to functionality, but at the same rate then something like the above would not honor option changes.

quisquous: Yeah, I'm not sure.  In some ways I think "0 = use default" feels more arbitrary than "undefined = use default" if I think about it more.  I am also skeptical that `durationSeconds: 0` is something that somebody is depending on.

panicstevenson: Well, `!duration` would include both undefined and zero scenarios, but I do agree that `undefined` is more immediately obvious than falsy.

quisquous: So, coming back to this a million years later, I guess my suggestion is to also allow `undefined` on `durationSeconds`.

panicstevenson: > So, coming back to this a million years later, I guess my suggestion is to also allow `undefined` on `durationSeconds`.

I take this to mean `(delay|duration|suppress)Seconds`, is that your intent as well? (Sorry `sound`!)

quisquous: Sorry that I don't feel strongly on this.  I think `delaySeconds` and `suppressSeconds` are slightly different in that the default is zero, and so `undefined` behaves the same as `0`.  The spirit of this change seemed to be to make fewer things allow `undefined`.  I think there are also few cases where `delaySeconds` or `suppressSeconds` use a variable, and so it's probably very infrequent that somebody would have to do `data.someAbilityDelay ?? 0`.  So, I think it might be ok to both be more strict here or to allow the behavior prior to this patch where undefined was ok.  I lean very weakly towards being strict, but honestly don't care.

And yeah, I think `sound` should allow `undefined` for the same reason that I proposed `durationSeconds` have it, in that it has a reasonable default that maybe you want to only optionally override.

panicstevenson: I added a comment for both durationSeconds and sound; I didn't add one for condition since maybe we want to remove undefined, but not now. I also didn't really want to commit to a TODO, but maybe it's more confusing the way I have it currently.

Issue #3156, opened 2021-06-22
Emulator playback bug with _onTriggerInternalDelaySeconds


quisquous:
```
ReferenceError: Cannot access 'ret' before initialization
   at RaidEmulatorPopupText._onTriggerInternalDelaySeconds
```

It looks like the code is referring to `ret` inside of its own construction.  I'm not sure how to easily disentangle this, so filing a bug.

"repro" /airquotes

```diff
diff --git a/ui/raidboss/data/04-sb/raid/o3s.js b/ui/raidboss/data/04-sb/raid/o3
s.js
index c024f271..9e75ec36 100644
--- a/ui/raidboss/data/04-sb/raid/o3s.js
+++ b/ui/raidboss/data/04-sb/raid/o3s.js
@@ -57,8 +57,9 @@ export default {

         data.holyTargets = data.holyTargets || [];
         data.holyTargets.push(matches.target);
-        return data.holyTargets.length === 4;
+        return data.holyTargets.length === 2;
       },
+      delaySeconds: 0.5,
       alarmText: (data, _matches, output) => {
         if (data.holyTargets[1] !== data.me)
           return '';
```
[DeltascapeV30Savage_20210617_045039_2m3s_wipe.log](https://github.com/quisquous/cactbot/files/6690769/DeltascapeV30Savage_20210617_045039_2m3s_wipe.log)

valarnin:
Would this work? I can't test at the moment:

```ts
    let ret: Promise<void>;
    // Disable prefer-const due to needing `ret` defined before using it in promise
    // eslint-disable-next-line prefer-const
    ret = new Promise<void>((res, rej) => {
      this.scheduledTriggers.push({
        expires: this.emulatedOffset + (delay * 1000),
        promise: ret,
        resolver: res,
        rejecter: rej,
      });
    });
    return ret;
```

PR #3286, opened 2021-07-09
raidemulator: Add support for watchCombatant

valarnin:
This one ended up being significantly more complicated than I was hoping.

This PR:

- Adds an override for watchCombatant for raidemulator to use
- Modifies the way PopupTextAnalysis tracks trigger promise resolution to not block waiting for a promise to resolve
- Refactors the way RaidEmulatorOverlayApiHook handles getCombatants to let other classes call it
- Adds a new class to override watchCombatant

Technically these 4 could be separate sequential PRs but considering how small 1/2/3 would be, I didn't think it was really worth splitting them up.

~~There's some sort of bug with resolved timestamp display, e.g. this trigger actually resolved at 1:05 into the pull:~~

image

~~So this isn't quite ready to merge yet (or that bug can be fixed in a later PR), I just wanted to get it open for feedback.~~

**edit**

Issue with resolved offset should be fixed now.

### DIFF
--- a/ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts
+++ b/ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts
@@ -1,7 +1,5 @@
-import Util, { WatchCombatantParams } from '../../../../resources/util';
+import Util, { WatchCombatantFunc, WatchCombatantParams } from '../../../../resources/util';
 import { OverlayHandlerRequests, OverlayHandlerResponseTypes } from '../../../../types/event';
-import AnalyzedEncounter from '../data/AnalyzedEncounter';
-import CombatantTracker from '../data/CombatantTracker';
 import RaidEmulator from '../data/RaidEmulator';
 
 import RaidEmulatorOverlayApiHook from './RaidEmulatorOverlayApiHook';
@@ -22,46 +20,50 @@ export default class RaidEmulatorWatchCombatantsOverride {
   private watches: Watch[] = [];
 
   constructor(private emulator: RaidEmulator, private overlayHook: RaidEmulatorOverlayApiHook) {
-    Util.setWatchCombatantOverride((params, func) => {
-      // To avoid having to undefined-check the params declare watch as partial here and just cast
-      // to non-partial when pushing to array since all props are set here properly but can't be
-      // set in the initializer
-      const watch: Partial<Watch> = {
+    const func: WatchCombatantFunc = (params, callback) => {
+      let watch: Watch;
+
+      // `watch` must be defined before the object creation, otherwise browser JS will throw the
+      // following error:
+      // Uncaught (in promise) ReferenceError: can't access lexical declaration 'watch'
+      // before initialization
+      // eslint-disable-next-line prefer-const
+      watch = {
         lastCheck: 0,
         params: params,
         cancel: false,
         start: 0,
-        func: func,
+        func: callback,
         msg: {
           call: 'getCombatants',
           ...params,
         },
+        res: () => {/* empty */},
+        rej: () => {/* empty */},
+        promise: new Promise<void>((res, rej) => {
+          watch.res = res;
+          watch.rej = rej;
+        }),
       };
 
-      watch.promise = new Promise<void>((res, rej) => {
-        watch.res = res;
-        watch.rej = rej;
-      });
-      this.watches.push(watch as Watch);
+      this.watches.push(watch);
       return watch.promise;
-    }, this.clear.bind(this));
+    };
+
+    Util.setWatchCombatantOverride(func, this.clear.bind(this));
 
     this.emulator.on('tick', () => {
-      const curEnc = this.emulator.currentEncounter;
-      const tracker = curEnc?.encounter.combatantTracker;
       const timestamp = this.emulator.currentLogTime;
 
-      if (!curEnc || !tracker || !timestamp)
+      if (timestamp === undefined)
         return;
 
-      this.tick(curEnc, tracker, timestamp);
+      this.tick(timestamp);
     });
   }
 
   public tick(
-      curEnc: AnalyzedEncounter,
-      tracker: CombatantTracker,
-      timestamp: number,
+    timestamp: number,
   ): void {
     for (const watch of this.watches) {
       if (watch.cancel)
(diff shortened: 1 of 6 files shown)


ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts
```diff
@@ -0,0 +1,89 @@
...
+          call: 'getCombatants',
+          ...params,
+        },
+      };
+
+      watch.promise = new Promise<void>((res, rej) => {
+        watch.res = res;
+        watch.rej = rej;
+      });
+      this.watches.push(watch as Watch);
```
quisquous, line 45: I'm not sure what you mean about undefined checks.

```suggestion
    Util.setWatchCombatantOverride((params, func) => {
      const watch: Watch = {
        lastCheck: 0,
        params: params,
        cancel: false,
        start: 0,
        func: func,
        msg: {
          call: 'getCombatants',
          ...params,
        },
        res: () => {/* empty */},
        rej: () => {/* empty */},
        promise: new Promise<void>((res, rej) => {
          watch.res = res;
          watch.rej = rej;
        }),
      };

      this.watches.push(watch);
```

ui/raidboss/emulator/overrides/RaidEmulatorOverlayApiHook.ts
```diff
@@ -37,53 +38,62 @@ export default class RaidEmulatorOverlayApiHook {
...
+      else if (hasNames && names.includes(combatant.name))
+        combatants.push(combatantState);
+    }
+    // @TODO: Move this to track properly on the Combatant object
+    combatants.forEach((c) => {
+      const lines = curEnc.encounter.logLines
+        .filter((l) => l.decEvent === 3 && isLineEventSource(l) && parseInt(l.id, 16) === c.ID);
+      const baseLine = lines[0];
+      if (baseLine) {
+        const line = baseLine as LineEvent0x03;
```
quisquous, line 89: Not for this patch, but maybe line events should have `isLineEvent0x03` type guard functions?

ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts
valarnin: Doesn't this run into the issue of `watch` not being set when the promise attempts to set `watch.res`?

ui/raidboss/emulator/overrides/RaidEmulatorOverlayApiHook.ts
valarnin: I was thinking of revisiting LineEvent and making `decEvent` and/or `hexEvent` a union type, and replacing all instances of `LineEvent` with a `LineEventAny` union of all `LineEvent` types, so that just checking the `decEvent`/`hexEvent` properties is sufficient for ts to narrow the type to the right subclass.

ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts
quisquous: If that's true, then `let watch: Watch; watch = { etc` should do the trick, right?

valarnin: I'd expect eslint to complain about that but I'm not able to check right now (still updating my linux install). If not, I'll correct it tomorrow.

quisquous: FWIW, typescript did not complain at me about the original, but that doesn't mean that there's not a bug there too.

valarnin: Yeah, eslint complains about that syntax.

image

The problem with the original suggestion is I think the same thing we ran into in #3156, but I still need to test to be sure.

valarnin: Had a chance to test the original suggestion as well, it throws the same error from the bug I linked, as I thought it would:
```
Uncaught (in promise) ReferenceError: can't access lexical declaration 'watch' before initialization
```

valarnin:
Okay. Something has gone really wrong with that attempt to rebase. Let me see if I can fix this...

ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts
quisquous: I think it is better to disable eslint than to assert something (possibly false in the future) to TypeScript.  That said, will https://eslint.org/docs/rules/prefer-const#ignorereadbeforeassign fix that error?

valarnin: No, unfortunately, setting that won't fix the eslint error in the case of `let watch; watch = ...`.

And the original suggestion's error is flat out a browser js error, not an eslint/ts error.

quisquous: Oh, you're right.  I misread that eslint.

That said, I still think it is better to disable eslint's prefer-const for that line than to add a type assertion.

valarnin: Trying to clean up some of these old PRs. I still need to test the latest changes, but I think this is what you were suggesting?

valarnin: Hmm, after testing, even with these changes it's still throwing an exception.
image

valarnin: Attempt number 3. This one works in testing and doesn't need the eslint disable, but it does require more indenting.

quisquous: Mmm, I was trying to say that you could use `let` here if you needed to for correctness, but just disable the eslint about using const, vs using a `watch as Watch` type assertion.  I can't see what the old code was to leave a snippet, but that's what I meant.

As a general, rule, I think narrow eslint disabling > broad eslint disabling > type assertions as preferred mechanisms for dealing with issues.  Type assertions just feel like working around TypeScript and can hide issues in the future during refactoring.  That's really what I was trying to get at.

valarnin: Yeah, I tried using `let` but it still errored at runtime, I'm guessing due to execution priority (it was executing the promise's anonymous function parameter before actually assigning the generated object to the `watch` variable).

valarnin: For reference, here's the previous iteration where `let` was still erroring:
https://github.com/quisquous/cactbot/blob/254afec6c6271f6132dda33653743cd12e65c5d4/ui/raidboss/emulator/overrides/RaidEmulatorWatchCombatantsOverride.ts#L23-L51

quisquous: > Yeah, I tried using let but it still errored at runtime

Huh!!

Thanks for the link.

quisquous:
Thank you for getting back to this!

PR #3289, opened 2021-07-11
build: convert custom loader to TS

MaikoTan:
@Trim21 Miss Auto-Label didn't tag it as `build`?

### DIFF
--- a/webpack/loaders/manifest-loader.ts
+++ b/webpack/loaders/manifest-loader.ts
@@ -1,4 +1,3 @@
-/* eslint-disable no-invalid-this */
 import fs from 'fs';
 
 import webpack from 'webpack';
--- a/webpack/loaders/manifest-loader.ts
+++ b/webpack/loaders/manifest-loader.ts
@@ -1,8 +1,9 @@
-'use strict';
+/* eslint-disable no-invalid-this */
+import fs from 'fs';
 
-const fs = require('fs');
+import webpack from 'webpack';
 
-module.exports = function(content, map, meta) {
+export default function(this: webpack.LoaderContext<never>, _content: string): string {
   this.cacheable(true);
   const contents = fs.readFileSync(this.resourcePath).toString();
   const lines = contents.trim().split(/\s*[\r\n]+\s*/);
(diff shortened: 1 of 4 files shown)


webpack/loaders/manifest-loader.ts
```diff
@@ -1,26 +1,27 @@
-'use strict';
+/* eslint-disable no-invalid-this */
```
MaikoTan, line 1: Not sure should we add an override in `.eslintrc`?

Trim21:
We don't have build label yet

webpack/loaders/manifest-loader.ts
```diff
@@ -1,26 +1,27 @@
...
+import fs from 'fs';
 
-const fs = require('fs');
+import webpack from 'webpack';
 
-module.exports = function(content, map, meta) {
+export default function(this: webpack.LoaderContext<never>, _content: string): string {
   this.cacheable(true);
```
quisquous, line 7: I tested this locally and it works, but it looks like you're changing the parameters here from (content) to (this, content)? Can you help me understand why this works? Is the `this` parameter the thing you're calling `this.cacheable` on (and if so, should it be renamed to context)?

MaikoTan: See https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#specifying-the-type-of-this-for-functions

> By default the type of `this` inside a function is `any`. Starting with TypeScript 2.0, you can provide an explicit `this` parameter. `this` parameters are fake parameters that come first in the parameter list of a function:

So it can't be renamed.

webpack/loaders/manifest-loader.ts
quisquous: I think we should turn off no-invalid-this in typescript files and use @typescript/no-invalid-this instead which should be able to handle this code without a disable?

webpack/loaders/manifest-loader.ts
quisquous: Oh, thank you! I didn't know about this feature.

webpack/loaders/manifest-loader.ts
MaikoTan: I agree, should I add it in this PR or you'll do a new one?

quisquous: I think you doing it in this PR or another PR is fine so that you can verify that it removes the need for this disable.

MaikoTan: Done, you're right, this rule is the correct one.

PR #3291, opened 2021-07-12
Fix spelling error in Yiazmat fight

DorkmasterFlek:
While hilarious, this does result in the TTS saying "move to pose-tive" which is confusing.


quisquous:
Haha, thanks for the fix.  <3

PR #3293, opened 2021-07-12
[WIP] jobs: Convert jobs component to class style

MaikoTan:
This PR aims to refactor the jobs module, make jobs component become a "real" component, also makes it easier for further TypeScript conversion.

- [ ] Extract job-specific mechanics from `Bars` into their own component.
- [ ] Extract player information/data into a new class, which maintain these as `BuffTracker` does.
- [ ] Get rid of `XxxxFuncMap`, prefer to use `onXxxEvent` to make it more like event-driven, and improve readability （and performance?).
- [x] Convert all job components to use the new style.

------
Also ping @Echoring


ui/jobs/jobs.js
```diff
@@ -766,9 +782,10 @@ class Bars {
     if (this.buffTracker)
       this.buffTracker.clear();
     // Reset job-specific ui
-    const reset = getReset(this.job);
-    if (reset)
-      reset.bind(null, this)();
+    // const reset = getReset(this.job);
+    // if (reset)
+    //   reset.bind(null, this)();
+    this.jobComponent?.reset();
```
valarnin, line 788: Is there a reason to keep this commented out code around (same with above lines 361-363)?

Additionally, just wanted to double check that the CEF version in use by OverlayPlugin supports the `?` operator in this context?

ui/jobs/components/util.js
```diff
@@ -0,0 +1,93 @@
+// import ResourceBar from '../../../resources/resourcebar';
```
valarnin, line 1: Is there a reason to not just make this file typescript to start with?

ui/jobs/jobs.js
MaikoTan: This is still work in progress, when I finish all the conversion, I'll delete these commented lines. Thanks for pointing out.

ui/jobs/components/util.js
MaikoTan: I wanted to write TypeScript, but there are typing issues cannot be fixes right now, so commented out temporary.

quisquous:
This is a lot of changes.  Is it possible to separate these out into smaller PRs?

MaikoTan:
> This is a lot of changes. Is it possible to separate these out into smaller PRs?

Yes, but I have to keep every separate PR works perfectly as before (this PR still does not work now). So I think I can convert most of them and then split the changes into separate patches.

ui/jobs/components/base.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  addTimerBar(options: {
+    id: string;
+    fgColor?: string;
+  }): TimerBar {
+    return this.bars.addTimerBar(options);
+  }
+
+  // In order to provide full type definitions for inherited components work,
+  // we need to disable these rules.
```
quisquous, line 57: Is this something that `abstract` would fix?

ui/jobs/player.ts
```diff
@@ -0,0 +1,99 @@
+import { EventResponses } from '../../types/event';
+import { Job } from '../../types/job';
+import { NetMatches } from '../../types/net_matches';
+
+import { calcGCDFromStat } from './utils';
+
+export type Stats = Partial<
```
quisquous, line 7: Would it be cleaner to not have `Partial` here (and `Required` in all of the functions) and instead have `stats?: Stats` on `Player`?

ui/jobs/components/base.ts
MaikoTan: Non unfortunately.
I tried before but the `abstract` modifier requires inherited classes to implement all abstract methods, that usually a job component need several of them only instead.

Another approach that don't need to disable rules is to have a `ComponentInterface` that provides types definitions for all the methods but marks them as optional, and then all classes would `implement` it.

ui/jobs/player.ts
MaikoTan: Sounds good, I'll do that!

ui/jobs/components/base.ts
valarnin: Yeah, because these are all concrete classes in the downstream .ts files, something like this should work:

```ts
interface ComboComponent {
  /**
   * (override) called on combo state changed
   */
  onCombo(skill: string): void;
}

// ...

if (someComponent instanceof ComboComponent)
  someComponent.onCombo(...);
```

Alternatively, you could leave both rules enabled and define the functions as:

```ts
  /**
   * (override) called on combo state changed
   */
  onCombo(_skill: string): void {
    // This space intentionally left blank
  }
```

MaikoTan: If so, I have thought. How about using [Mixins](https://www.typescriptlang.org/docs/handbook/mixins.html) and inherit methods that are needed only?

valarnin: I don't think mixins really make sense here since we don't have any default behavior or extra properties that would benefit from being inherited here, everything needs to be redefined downstream anyways.

Mixins are more of a way to handle multiple class inheritance, rather than what's going on here.

MaikoTan: Yeah, that's why I mentioned it.
For example, the method `onCombo` that you mentioned has basically the same code across components, mixins can be a good way to replenish the lacking of "multiple inherits" in this case. What do you think?

Additionally, I am thinking to move code that related to specific jobs out of `jobs.js`. Such as `mpTick` for BLM, or `cpBar` for crafters, and so on. Do you know how to deal with these various mechanics perfectly?

PR #3294, opened 2021-07-12
build: accelarate typescript build

MaikoTan:
by adding fork-ts-checker-webpack-plugin

Before:
real    1m28.067s
user    1m9.672s
sys     0m8.042s

After:
real    0m47.489s
user    1m7.175s
sys     0m6.496s

-------
EDIT: Found that remove progress plugin make it faster

real    0m8.470s
user    0m5.978s
sys     0m0.987s

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3001,6 +3001,12 @@
       "integrity": "sha1-x57Zf380y48robyXkLzDZkdLS3k=",
       "dev": true
     },
+    "at-least-node": {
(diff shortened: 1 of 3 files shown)


quisquous:
Oh wow, this is SO much better.  Thank you!!!

JLGarber:
Definitely agree with quisquous, this is substantially better. Thanks for figuring it out!

Issue #3296, opened 2021-07-13
T9 - Problems


DCRSakio:
In T9 (Nael van Darnus)
At the Ice/Fire/Lightning Dragon-Phase the RaidAlert goes crazy as heall and displays the whole code (and with TTS even speaks it for severeal minutes).

I'm using German Overlay/Callouts

valarnin:
Does this issue reproduce in raidemulator? Can you upload a network log file for the fight?

panicstevenson:
Running this through raidemulator, I get:
```
{id: "T9 Dragon Marks", netRegex: /^(?:20)\|(?:[^|]*)\|(?:[^|]*)\|(?:Nael Deus Darnus)\|(?:7E6)\|/i, netRegexDe: /^(?:20)\|(?:[^|]*)\|(?:[^|]*)\|(?:Nael Deus Darnus)\|(?:7E6)\|/i, netRegexFr: /^(?:20)\|(?:[^|]*)\|(?:[^|]*)\|(?:Nael Deus Darnus)\|(?:7E6)\|/i, netRegexJa: /^(?:20)\|(?:[^|]*)\|(?:[^|]*)\|(?:ネール・デウス・ダーナス)\|(?:7E6)\|/i, …} TypeError: Cannot read property '0' of undefined
    at infoText (t9.js:343)
    at Object.valueOrFunction (popup-text.ts:993)
    at PopupTextAnalysis._addTextFor (popup-text.ts:1327)
    at PopupTextAnalysis._onTriggerInternalInfoText (popup-text.ts:1190)
    at triggerPostPromise (popup-text.ts:920)
    at triggerPostDelay (popup-text.ts:952)
    at PopupTextAnalysis.OnTriggerInternal (popup-text.ts:959)
    at PopupTextAnalysis.OnTrigger (PopupTextAnalysis.ts:105)
    at PopupTextAnalysis.onEmulatorLog (PopupTextAnalysis.ts:159)
    at async AnalyzedEncounter.analyzeFor (AnalyzedEncounter.ts:216)
```

```js
    {
      id: 'T9 Dragon Marks',
      netRegex: NetRegexes.startsUsing({ id: '7E6', source: 'Nael Deus Darnus', capture: false }),
      netRegexDe: NetRegexes.startsUsing({ id: '7E6', source: 'Nael Deus Darnus', capture: false }),
[...]

PR #3297, opened 2021-07-14
raidboss: Fix T9 Fire In Callout

panicstevenson:


quisquous:
Regardless of whether this is the issue in #3296, I think this is still a good fix, thanks.

PR #3298, opened 2021-07-14
Raidboss: Add Suzaku Normal oopsy, timeline, and triggers

JLGarber:
This wasn't too difficult to figure out, although getting through phase 1 solo is a pain. It's not realistic to survive the 4x Scarlet Lady explosion without invulns, and it's not yet possible to kill a Scarlet Plume solo, so the trick is to use cooldowns to burst the Ladies on the first set, and invuln the explosions on the second. (Also, although they use a tether visual, they do not generate tether lines in the log.)

There's probably stuff we can do with Scarlet Hymn (DDR during the intermission) but I'm not terribly interested in pursuing it either for Raidboss or Oopsy. It's very clear without artificial aid when someone has failed their dance steps.

Finally, I'm surprised/disappointed that Ruthless Refrain disappears altogether after the first Incandescent Interlude.


quisquous:
Thank you for this, this all looks great.  Thanks for writing this in TypeScript too.

I think the DDR is definitely not worth the time to do.  :P

PR #3300, opened 2021-07-14
Allow additional alert output for TTS+Sound+Text

20133774:
Not sure if I've done this request correctly since it's my first one but:
please allow the use of TT+Sound+Text at once for raidboss alerts, i've been using "Text and Sound" for a while but would love to implement TTS instead of making custom triggers for individual actions.
I would switch to Text + TTS but I'm used to hearing the sounds and they help me quite a lot


MaikoTan:
Are you trying to request a feature or contributing a new feature?
`Pull Request` is for contributing, so maybe you should turn to `Issues` and close this one.

PR #3311, opened 2021-07-16
refactor(jobs): direct compute MNK lightning stack

MaikoTan:
After patch 5.4 update, MNK's Greased Lightning is not
binding on GCD/Combo anymore.
Nowadays it is traits which only be affected by player's current level,
we can directly compute the corresponding `stacks`, so
move the computing code into `calcGCDFromStat` function.

------
By the way, @quisquous should/can I remove the related code from `FFXIVProcessIntl.cs`, etc?
For example, like this?
```diff
-   [StructLayout(LayoutKind.Explicit)]
-   public struct PugilistJobMemory {
-     [FieldOffset(0x00)]
-     public ushort lightningMilliseconds;
-
-     [FieldOffset(0x02)]
-     public byte lightningStacks;
-   };
-
    [StructLayout(LayoutKind.Explicit)]
    public struct MonkJobMemory {
-     [FieldOffset(0x00)]
-     public ushort lightningMilliseconds;
- 
-     [FieldOffset(0x02)]
-     public byte lightningStacks;
      // I don't know if this `Offset` is correct though
      [FieldOffset(0x03)]
      public byte chakraStacks;

-     [NonSerialized]
-     [FieldOffset(0x04)]
-     private byte _lightningTimerState;
-
-     public bool lightningTimerFrozen {
-       get {
-         return (_lightningTimerState > 0);
-       }
-     }
    };
```

### DIFF
--- a/ui/jobs/bar.d.ts
+++ b/ui/jobs/bar.d.ts
@@ -22,7 +22,6 @@ export interface Bars {
     circleOfPower: boolean;
     presenceOfMind: boolean;
     paeonStacks: number;
-    lightningStacks: number;
     huton: boolean;
     shifu: boolean;
     museStacks: 1 | 2 | 3 | 4;
--- a/ui/jobs/components/mnk.js
+++ b/ui/jobs/components/mnk.js
@@ -15,17 +15,6 @@ export function setup(bars) {
     classList: ['mnk-color-chakra'],
   });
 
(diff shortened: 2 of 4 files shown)


quisquous:
> By the way, @quisquous should/can I remove the related code from FFXIVProcessIntl.cs, etc?

Please do!

PR #3303, opened 2021-07-14
raidboss: Fix a8n aoe callout spam

valarnin:
I've gotten this one a couple times in roulette the last week, and when the aoe fires it's certainly noticeable because it fires once for each player that's alive 🙃 

There's no `startsUsing` line for this one so I added `suppressSeconds` instead. I'm not sure if any of the other `ability` triggers here have the same problem, since I've never seen most of them in roulettes.

PR #3313, opened 2021-07-17
Raidboss: Reduce annoyance in A8n

JLGarber:
(This will conflict with and possibly replace #3303.)

As I noted on Valarnin's pull request, this is one of those things I kept meaning to do and kept forgetting. I wasn't able to get a group for this to test, but I did run it through solo, and it didn't *feel* annoying. Short Needle and Hydrothermal Missile correctly called only once for each position on the timeline.

I went ahead and commented out the Hydrothermal syncs that were within 4 seconds of each other. Technically they should have been okay, but I figured it was best to be safe.

No idea what's going on with the end-of-file newline. I would have said it was a Unix-vs-Windows thing, but I'm the only person who's touched this file. Regardless, it shouldn't be a serious issue.


quisquous:
Thanks for the fixes!

PR #3315, opened 2021-07-17
Build: turn on babel cache & use babel for TS building

MaikoTan:
- Turn on babel cache

It is helpful to enable babel cache when in rebuilding (e.g. running webpack-dev-server), which would significantly decrease the building time.

- Use babel for TypeScript building

Although `ts-loader` (`tsc` inside) and `babel-loader` (`babel` inside) would basically do the same thing in TypeScript transform, `babel-loader` removes type notations only, and doesn't do the type-checking things, so it would be faster than `ts-loader` (even with `transpileOnly` flag) in most cases. Additionally, we now need `babel-loader` to ensure code available always in Chrome v75 of current CEF, so why not combine them in one loader?

- How is TypeScript type checking now

Nowadays we are using `fork-ts-checker-webpack-plugin` for type checking, which runs `tsc` in a separated process and report errors asynchronously (in production mode), that would not block webpack building (that `ts-loader` will).

### DIFF
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -6,10 +6,9 @@
         "after": true
       }
     ],
-    // nullish coalescing operator not supported in OverlayPlugin's Chrome M75
-    // TODO: change to ES2020 once OverlayPlugin rolls CEF past M80.
-    "target": "ES2019",
-    "module": "ES2020",
+    // Thanks to babel-loader, we don't need to set it as ES2019.
+    "target": "ESNext",
+    "module": "ESNext",
     "allowJs": true,
     "strict": true,
     "declaration": false,
--- a/tsconfig.base.json
+++ b/tsconfig.base.json
@@ -7,7 +7,8 @@
       }
     ],
     // Thanks to babel-loader, we don't need to set it as ES2019.
-    "target": "ESNext",
+    // But we need to set it as ES2020 to ensure mocha can run correctly.
+    "target": "ES2020",
     "module": "ESNext",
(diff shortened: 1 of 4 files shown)


tsconfig.base.json
```diff
@@ -6,10 +6,10 @@
...
       }
     ],
-    // nullish coalescing operator not supported in OverlayPlugin's Chrome M75
-    // TODO: change to ES2020 once OverlayPlugin rolls CEF past M80.
-    "target": "ES2019",
-    "module": "ES2020",
+    // Thanks to babel-loader, we don't need to set it as ES2019.
+    // But we need to set it as ES2020 to ensure mocha can run correctly.
+    "target": "ES2020",
+    "module": "ESNext",
```
quisquous, line 12: What do you get out of `ESNext`? Mostly wondering why you're adjusting this one as well.

quisquous (commented):
This seems good to me.  We still have a workflow that runs `tsc` as well, so I'm not worried about losing testing here.

tsconfig.base.json
MaikoTan: Currently `esnext` and `es2020` is the same, you mean better to keep it remained?

quisquous: Oh, I see.  Yeah, I was mostly wondering why you were changing it here.  I think I'd rather stick to ES2020 and then change it explicitly when we change it.

MaikoTan: Okay, done.

PR #3316, opened 2021-07-17
jobs: remove dynamic threshold of storm's eye

MaikoTan:
WAR in ShB don't need to track Storm's Eye dynamically,
not only because the duration of Storm's Eye can be 60s nowadays,
but there are also many skills would extend the duration of Storm's Eye.

So we can just set the threshold to `GCD * 3 + 1`,
which is also consistent with other jobs.


quisquous:
Thanks for the cleanup!

PR #3317, opened 2021-07-18
Raidboss/Oopsy: Add Drowned City of  Skalla dungeon package

JLGarber:
Nothing particularly interesting on this one. We could probably eliminate the last block on boss 2, and instead do a forward and back sync against timestamp 1200, but I figured explicit was better to start with.

It's possible we could do a better set of triggers to account for interactions between the boss 3 head markers, but I don't think it's really worth the effort. "You have X" is standard enough anyway.


quisquous:
This all looks great, thank you as always!

PR #3321, opened 2021-07-18
raidboss: earlier callouts for wolex summon phases

quisquous:
BLM/WHM is Specter Of Light before to the limit, and DRK/BRD is To The
Limit before Specter Of Light.  Add additional timeline syncs (but leave
the previous as a catch-all) and fix up triggers to reflect this.

Additionally, every limit break can be used in both phases, so fix up
timeline entries for this.

### DIFF
--- a/ui/raidboss/data/05-shb/trial/wol-ex.txt
+++ b/ui/raidboss/data/05-shb/trial/wol-ex.txt
@@ -103,12 +103,13 @@ hideall "Limit Break"
 517.3 "--middle--" sync /:Warrior Of Light:4F45:/
 
 # BLM/WHM
-522.5 "--sync--" sync /:Warrior Of Light:4F37:/ window 5,5 jump 1522.8
+519.6 "--sync--" sync /:4F73:Warrior Of Light/ window 5,5 jump 1519.5
+522.6 "Specter -> BLM/WHM" #sync /:Warrior Of Light:4F73:/
 527.8 "--sync--" sync /:4F3D:Spectral Black Mage/ window 40,40 jump 1527.8
-528.6 "Specter -> BLM/WHM" #sync /:Warrior Of Light:4F3[456]:/
 
 # DRK/BRD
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/05-shb/trial/wol-ex.txt
```diff
@@ -103,19 +103,20 @@ hideall "Limit Break"
 517.3 "--middle--" sync /:Warrior Of Light:4F45:/
 
 # BLM/WHM
+522.5 "--sync--" sync /:Warrior Of Light:4F37:/ window 5,5 jump 1522.8
 527.8 "--sync--" sync /:4F3D:Spectral Black Mage/ window 40,40 jump 1527.8
-528.6 "Limit -> BLM/WHM" #sync /:Warrior Of Light:4F3[56]:/
+528.6 "Specter -> BLM/WHM" #sync /:Warrior Of Light:4F3[456]:/
```
panicstevenson, line 108: "Specter" is non-actionable, right?

panicstevenson (approved):

ui/raidboss/data/05-shb/trial/wol-ex.txt
quisquous: Oh hmm, I am not sure that I know what you're asking.  Specter is the cast that summons the warriors of light that then do stuff.

However, I think you're right that I can't change "Limit" to "Specter" and I should put that text on the actual "Specter" ability.  If I adjust everything by like 0.1s here and there, this makes all of the tells happen at exactly the same time which feels much nicer then how it used to be.

panicstevenson (approved):

ui/raidboss/data/05-shb/trial/wol-ex.txt
panicstevenson: Eh, I was mostly thinking that people might be confused that Specter means "something is happening", but coming back to this I don't think that's any different from what is already there, especially if it's the wrong non-actionable move; so, tl;dr disregard.

PR #3322, opened 2021-07-18
raidboss: fix DRS avowed sword triggers

quisquous:


panicstevenson (approved):
Assuming the removed console.error is debug logging

PR #3326, opened 2021-07-19
jobs: fix DRK darkside timer box bugs

quisquous:


ui/jobs/components/drk.ts
```diff
@@ -35,7 +35,7 @@ export const setup = (bars: Bars): void => {
       }
     }
 
-    const oldSeconds = darksideBox.duration ?? 0 - darksideBox.elapsed;
+    const oldSeconds = (darksideBox.duration ?? 0) - darksideBox.elapsed;
```
MaikoTan, line 38: I found that this line can be 

```suggestion
    const oldSeconds = darksideBox.value;
```

quisquous: Done, but with an extra simplification too.

PR #3334, opened 2021-07-20
add ja translation for zadnor based on #3333

Akurosia:


quisquous:
Thanks for handling that!

MaikoTan:
I think there are few names can be shorten, or I'll create a new PR later.

PR #3341, opened 2021-07-24
jobs: convert jobs.css into stylus

MaikoTan:


panicstevenson:
What's the advantage to using Stylus over SCSS?

I would imagine if we had a need to move away from CSS 3 we would use SCSS, but I don't know of all the benefits/differences.

MaikoTan:
> What's the advantage to using Stylus over SCSS?
> 
> 
> 
> I would imagine if we had a need to move away from CSS 3 we would use SCSS, but I don't know of all the benefits/differences.

In fact, I don't very clear with the difference between them, either.
Why I choose Stylus instead of SCSS is that I thought the programmatic statements of Stylus is much clearer than SASS, and Stylus also includes some feature that SASS doesn't have. Whatever the css-preprocessor is, they actually do the same thing, basically. So I don't think it is a big deal to choose one?

quisquous:
> Thanks for your contribution!🚀
> 
> Currently your title is: jobs: convert jobs.css into stylus,
> but it should be in the format of `scope: description`.

I fixed this in #3342 if you rebase/merge onto main.

panicstevenson:
> So I don't think it is a big deal to choose one?

I think that's fair; I'm not against a pre-processor for CSS personally, I just have never heard of Stylus before and was really only aware of SCSS. I don't know which one has the bigger market presence but maybe we should adopt whichever one there (similar to why we went with Webpack, lots of people know it)

MaikoTan:
Yep, if we follow the most popular one, we should choose `Sass/Scss`.
I can recall that there are statistical data back in 2019 told that `Sass/Scss` > `LESS` > `Stylus` > `PostCSS` > Others (I can't find the link now tho). And `Sass` got even 50%~ popularity.
I can change the code to use `sass` instead, which is not difficult.

quisquous:
Do you have links for a "why folks use a CSS pre-processor"? I agree that jobs.css needs to be cleaned up, but it's not clear to me that a preprocessor is the solution there.  Are there things you wish you could be doing that you can only do with a preprocessor?  This is just me not quite understanding what the underlying goal is here.

MaikoTan:
Ah, I read a Chinese article that describes `Sass` so I learnt CSS pre-processor. ⊙﹏⊙∥

Technically a pre-processor would provide things like `variables`, `number/units computing`, `programmatical feature like conditional/looping/...`, `extends/inheritance/nesting`, `functions/mixins`, `CSS module management`, `interoperation with JS`, and so on that pure CSS has not.
These features would help developers to reduce duplications and improve maintenance.
Also, things like nesting would let us easier to read stylesheet files, for example,
```stylus
.parent {
  /* blahblah */
  .child {
    /* blahblahblah */
  }
}
```
is much clearer (and you can easily know the nesting relationship) than
```stylus
.parent {
  /* blahblah */
}
.parent .child {
  /* blahblahblah */
}
```
Especially there are always extremely long names of the latter one.

------

Currently, I have just converted few styles to use the syntaxes from `Stylus`, so it might look a little useless in this case?
Mostly I would like to consult with folks that HOW is it like and SHOULD/COULD we use it (also which should we use I guess?).

quisquous:
I think it would help motivate me to want to do something like this if I could understand what problems you were seeing and wanted to solve.

Mostly, I have to admit that I am somewhat skeptical of adding additional tools that people have to understand.  I agree that a CSS preprocessor could add additional functionality and be more expressive, but the downside is that it is yet another layer that people have to learn about and know in order to make changes.  In other words, people have to learn CSS and now another thing as well; additionally the build pipeline a little slower and more complicated too.  So, I think for me I would like additions like that to have a good reason to do it.

If it helps, I didn't switch to webpack for a long time for exactly this reasoning.  Previously, it was much simpler to not use webpack, but there were a bunch of things (imports, typescript) that made switching to webpack a good decision.  However, it also makes everything a LOT more complicated and adds more friction to development.  It's a trade-off, as everything is.

PR #3282, opened 2021-07-09
raidboss: convert 05-shb/eureka to TS

quisquous:

PR #3348, opened 2021-07-25
raidboss: fix DRS sword brand direction

### LINKED ISSUE #3339: DR:S Trinity Avowed Sword and Brand Raidboss calls not working.
## Description

7/23/21

During the Trinity Avowed boss in Delubrum Reginae (Savage) the Raid Boss plugin has stopped accounting for sword cleaves AND brands. So it will call the sword safe spot, but not where to stand cardinal from the safe spot. And during the second half of the fight, the callouts don't account for brand debuff when telling you where to go.
Confirmed running latest version.

quisquous:
Fixes #3339.

Also broken in #3282, sorry!


quisquous:
Just going to go ahead and merge this so it can get tested.

PR #3350, opened 2021-07-25
Raidboss + Oopsy: Add A6N package

JLGarber:
Not much to say here. This is a very simple encounter with nothing interesting going on anywhere.


MaikoTan:
well, the bot is saying that the "scope" should be lowercase, as quis said that helps them to write the release change log, i.e. you should change it like "raidboss/oopsy: Add A6N package" or something.

quisquous:
I fixed up the multiple scopes issue in #3351.  I merged this because everything else looks good.

PR #3354, opened 2021-07-26
raidboss: delubrum reginae normal + savage KR

Bing-su:


Bing-su:
`'node_modules' is not recognized as an internal or external command` caused in my local environment 😢

quisquous:
> `'node_modules' is not recognized as an internal or external command` caused in my local environment 😢

Have you run `npm install`?

Bing-su:
yes. and this [stackoverflow](https://stackoverflow.com/questions/30710550/node-js-npm-modules-installed-but-command-not-recognized/36168581#36168581) answer also don't work.

PR #3356, opened 2021-07-26
lint: force typescript files to use override keyword

MaikoTan:
Enabled `noImplicitOverride` flag to force an `override` keyword when overriding a method or property,
this feature needs TypeScript version 4.3, so I upgrade it incidentally, also the `@typescript-eslint` plugin.


valarnin (approved):
This looks good to me. There's some unfortunate extra churn related to the `properCaseConvertedLine` property which I plan on eliminating eventually.

panicstevenson (approved):

quisquous:
Thanks, this is a really great thing to have turned on.

PR #3357, opened 2021-07-26
ci: trigger lint-pr-title when pr edited/reopened

MaikoTan:
Noticed in #3356 that the bot didn't run the workflow again when I successfully changed the title.

### DIFF
--- a/.github/workflows/lint-pr-title.yaml
+++ b/.github/workflows/lint-pr-title.yaml
@@ -2,7 +2,7 @@ name: Lint PR Title
 
 on:
   pull_request_target:
-    types: [ opened, synchronize ]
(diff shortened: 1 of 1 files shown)


quisquous:
Ah, thank you, you read my mind! I was looking for these actions but couldn't find them when I looked quickly yesterday.

MaikoTan:
> Ah, thank you, you read my mind! I was looking for these actions but couldn't find them when I looked quickly yesterday.

https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#pull_request

Events that trigger Github Actions are based on webhook, so their names and behaviours are the same as webhook payloads.

PR #3359, opened 2021-07-27
raidemulator: Force combatant sig states for line events

### LINKED ISSUE #3349: raidemulator GetCombatants errors when seeking
Repro:
* Import https://github.com/quisquous/cactbot/issues/3339#issuecomment-886120461
* Open Trinity Avowed
* (make a coffee)
* Notice only `Gleaming Arrow: expected 6, got 4` error in log (I'll look at this)
* Click on 8 minutes in timeline
* Hit play (?)
* See a bunch of unexpected GetCombatants errors in the log

```
delubrum_reginae_savage.ts:2102 Avowed Avatar: expected at least 3 combatants got 0
delubrum_reginae_savage.ts:2358 Gleaming Arrow: expected 6, got 0
delubrum_reginae_savage.ts:2102 Avowed Avatar: expected at least 3 combatants got 0
delubrum_reginae_savage.ts:2358 Gleaming Arrow: expected 6, got 0
delubrum_reginae_savage.ts:2102 Avowed Avatar: expected at least 3 combatants got 0
delubrum_reginae_savage.ts:2358 Gleaming Arrow: expected 6, got 0
```

valarnin:
Fixes #3349

Might also fix the second issue mentioned, at least it's not returning `???` anymore, but I've never done this fight and have no idea if the callouts are correct.

I had to add multiple line events to the 'force significant states' mapping, the `20` event only fixed the original `4000478C` combatant. I included logic for target line events though I'm not sure of any specific cases where they would apply yet.

### DIFF
--- a/ui/raidboss/emulator/data/CombatantTracker.ts
+++ b/ui/raidboss/emulator/data/CombatantTracker.ts
@@ -1,4 +1,5 @@
 import { Lang } from '../../../../resources/languages';
+import { UnreachableCode } from '../../../../resources/not_reached';
 import PetNamesByLang from '../../../../resources/pet_names';
 
 import Combatant from './Combatant';
@@ -111,6 +112,22 @@ export default class CombatantTracker {
     this.mainCombatantID = this.enemies.sort((l, r) => {
       return (eventTracker[r] ?? 0) - (eventTracker[l] ?? 0);
     })[0];
+
+    // Always treat the last line involving a combatant as a significant state
+    for (const id in this.combatants) {
+      const combatant = this.combatants[id];
+
+      if (!combatant)
+        throw new UnreachableCode();
+
+      // Get the last state timestamp
+      const state = combatant.latestTimestamp;
+
+      if (state === -1 || combatant.significantStates.includes(state))
+        continue;
+
+      combatant.significantStates.push(state);
+    }
   }
 
   addCombatantFromLine(line: LineEventSource): void {


valarnin:
Hmm, this only fixes the initial analysis and not the playback somehow, which I'm a bit baffled by. I'll investigate this further.

valarnin:
Alright, this should be 100% correct now.

ui/raidboss/emulator/data/Combatant.ts
```diff
@@ -96,7 +110,9 @@ export default class Combatant {
     const oldStateJSON = JSON.stringify(this.states[lastSignificantStateTimestamp]);
     const newStateJSON = JSON.stringify(this.states[timestamp]);
 
-    if (lastSignificantStateTimestamp !== timestamp && newStateJSON !== oldStateJSON)
+    if (lastSignificantStateTimestamp !== timestamp && (
+      newStateJSON !== oldStateJSON ||
+      this.isForcedState(lineEvent)))
```
quisquous, line 115: Are there line types for which we don't want to push a significant state if it's the same state? Saying differently, should this just be `if (lastSignificantStateTimestamp !== timestamp)` with no other qualifications?

valarnin: Significant states add a fair amount of overhead downstream since it has to loop over the array fairly often.

Realistically, it might make sense instead to just track a `lastState` property, and then push that as a significant state after processing the encounter lines, but that felt like a less accurate way to handle the issue?

quisquous: I'm not sure I understand your proposal to track a `lastState` property.  Is that different than me suggesting that you update the time of the previous state if the two states are the same?  (I'm still not sure I understand why that wouldn't work.)

valarnin: Let's assume you have lines with timestamps of `[1, 3, 5, 7, 9]`.

`1` is the `03` added combatant line and is treated as significant.
`3` is a line for an unrelated combatant and is ignored.
`5` is a `starts using` line, and is treated as significant because the position/heading of the combatant has changed.
`7` is a `hit single target` line, and is treated as not significant due to position/heading/etc being identical.
`9` is another `starts using` line, and is treated as not significant due to position/heading/etc being identical.

Of those, `[1, 5]` are counted as significant with current `main` code.

When pulling this information for the `getCombatants` override, calling it at timestamp `3` will give the same state as `1`, since `1` was the last significant state. Calling it at timestamp `5` will give the state at `5` since it was the last significant state. `7` and `9` won't return the combatant, since there are no states which match criteria (e.g. this combatant might have been killed or otherwise removed).

With the current PR code, `[1, 5, 7, 9]` would be counted as significant. This adds additional overhead, since we now have to track the two additional states, but calling `getCombatants` for the `1` and `5` timestamps still returns the correct data. Additionally, now that there are significant states tracked further up, calling for `7` or `9` will now return the combatant.

In the case of tracking `lastState`, ultimately you end up with `[1, 5, 9]` as significant. This will also return the correct data for `7` and `9`, with slightly less overhead.

In the case of updating, ultimately you end up with `[1, 9]` as significant. If this were the case, calling `getCombatants` at timestamp `5` would return the state at `1`, since it's the previous state that's significant, which is not correct data.

quisquous: Re: updating.  Sorry, my misconception was that we got the next combatant state, and not the previous combatant state.

> Calling it at timestamp 5 will give the state at 5 since it was the last significant state. 7 and 9 won't return the combatant, since there are no states which match criteria (e.g. this combatant might have been killed or otherwise removed).

AHA.  Thank you for the detailed explanation <3.  This was the real piece that I was missing.  To restate, the issue here is that these mobs disappear.  If they had done anything else to generate a significant state, then getCombatants would have found them.  If that's true, then your `lastState` idea sounds good to me.

Thank you!

quisquous: Do you think it makes sense to do the `lastState` change in this PR or as a follow-up PR?

valarnin: `lastState` would replace the changes that are already in this PR, essentially they're two different ways to fix the same issue.

quisquous: If significant states add a lot of overhead, then doing `lastState` instead sounds like a better approach to implement than this one.  Do you disagree?  I'm not sure I understand your concerns about it being a less accurate method.

valarnin: Adjusted to accomplish the same as we discussed in latest commit, without having to track anything additional.

Issue #3275, opened 2021-07-07
Future Features for Raid Emulator


valarnin:
Now that the raid emulator typescript conversion is complete, I wanted to solicit input on future features/ideas for the emulator.

I'll maintain a task list here with features/status/requester.

- [ ] N/A - Translation support - WIP
- [ ] valarnin - Map/position display during playback - Not started yet
- [ ] valarnin - Shield display on party info - Not sure if possible
- [ ] quisquous - Encounter analysis for testing purposes - Not started yet
  - quisquous suggested this shortly before the typescript conversion process started, being able to pass logs through the raid emulator analysis process to get information on executed triggers for testing, IIRC?

quisquous:
Hmm.  I feel like my opinions of what the raid emulator should be for diverge from what you think it should be.

For me, my view of the goal of raid emulator is to test raidboss triggers, in other words testing what raidboss would have emitted during a fight.  My uses of the tool are:
* 98% reading the analysis of trigger output for a particular player (or console.log lines I've added to debug)
* 2% real time playback (ultimates + one fight an expansion) to verify tricky timings / colliding outputs

(For what it's worth, I would like to add oopsy emulation at some point, but I don't think it needs the same real time nature that raidboss emulation does and was planning to do it in a "headless" way that just generated the events from a log.  I don't think it would make sense to add it to raidemulator.)

I think I've mentioned this before, but I still don't find the hp/mp bars useful.  I know you said it helped you track down a mitigation issue, but I would strongly argue that you can do that from looking at video, fflogs, or just the logs themselves more efficiently in most cases.  Further things in the "replayer" space like shielding, maps, positioning, etc are in this similar "neat idea but I have a hard time seeing the goal" category.  In particular, I think there are already tools that do this really well like the fflogs replayer, which also has lb, dps meters, effects, cast bars.  See: [example](https://www.fflogs.com/reports/F42Yg8tkmapKPhMX#fight=22&view=replay).  What problem are you trying to solve that requires recreating that entire tool?  Sorry, I'm not trying to be argumentative here, but I just feel like there's a very large disconnect between our goals here and I have a hard time understanding what you're aiming to achieve.

[...]

PR #3360, opened 2021-07-28
raidemulator: Performance optimizations

valarnin:
As discussed in #3275, there are a few different ways that performance for raidemulator's encounter analysis can be improved, mostly surrounding the sheer number of regexes that need to be executed.

This PR goes for a different option than the ones discussed in that issue, instead choosing to cache the results of a regex execution against a given line so that it can be looked up on subsequent analysis. This results in a noticeable performance improvement, 5-20%.

Additional improvements to other components have increased that performance gain to 10-30% on average, mainly consisting of switching from `forEach` to `for ... of`, and eliminating pointless excess looping over log lines.

The DRS Trinity Avowed encounter is a statistical outlier, going on average from 2m 42s to load for me, down to on average 1m 9s.

I'm opening this as draft since it's still a bit untested (in particular I haven't tested anything related to importing logs, which is potentially impacted due to the change in `Encounter.ts`).

### DIFF
--- a/ui/raidboss/emulator/data/Combatant.ts
+++ b/ui/raidboss/emulator/data/Combatant.ts
@@ -72,9 +72,17 @@ export default class Combatant {
   pushPartialState(timestamp: number, props: Partial<CombatantState>): void {
     if (this.states[timestamp] === undefined) {
       // Clone the last state before this timestamp
-      const stateTimestamp = this.significantStates
-        .filter((s) => s < timestamp)
-        .sort((a, b) => b - a)[0] ?? this.significantStates[0];
+      let stateTimestamp = this.significantStates[0] ?? timestamp;
+      // It's faster to start at the last timestamp and work our way backwards
+      // since realistically timestamp skew is only a couple ms at most
+      // Additionally, because cloning a 3000+ element array a few thousand times is slow,
+      // don't for-in over a reverse of the array
+      for (let i = this.significantStates.length - 1; i >= 0; --i) {
+        const ts = this.significantStates[i] ?? 0;
+        if (ts <= timestamp)
+          stateTimestamp = ts;
+      }
+
       if (stateTimestamp === undefined)
         throw new UnreachableCode();
       const state = this.states[stateTimestamp];
@@ -93,10 +101,8 @@ export default class Combatant {
       this.significantStates[this.significantStates.length - 1];
     if (!lastSignificantStateTimestamp)
       throw new UnreachableCode();
-    const oldStateJSON = JSON.stringify(this.states[lastSignificantStateTimestamp]);
-    const newStateJSON = JSON.stringify(this.states[timestamp]);
 
-    if (lastSignificantStateTimestamp !== timestamp && newStateJSON !== oldStateJSON)
+    if (lastSignificantStateTimestamp !== timestamp)
       this.significantStates.push(timestamp);
   }
 
--- a/ui/raidboss/emulator/data/CombatantTracker.ts
+++ b/ui/raidboss/emulator/data/CombatantTracker.ts
@@ -16,59 +16,36 @@ export default class CombatantTracker {
   others: string[] = [];
   pets: string[] = [];
   mainCombatantID?: string;
-  initialStates: { [id: string]: Partial<CombatantState> } = {};
   constructor(logLines: LineEvent[], language: Lang) {
     this.language = language;
     this.firstTimestamp = Number.MAX_SAFE_INTEGER;
     this.lastTimestamp = 0;
     this.initialize(logLines);
-    // Clear initialStates after we initialize, we don't need it anymore
-    this.initialStates = {};
   }
 
   initialize(logLines: LineEvent[]): void {
-    // First pass: Get list of combatants, figure out where they
-    // start at if possible
+    this.firstTimestamp = logLines[0]?.timestamp ?? 0;
+    this.lastTimestamp = logLines.slice(-1)[0]?.timestamp ?? 0;
+
+    const eventTracker: { [key: string]: number } = {};
+
     for (const line of logLines) {
       this.firstTimestamp = Math.min(this.firstTimestamp, line.timestamp);
       this.lastTimestamp = Math.max(this.lastTimestamp, line.timestamp);
 
-      if (isLineEventSource(line))
-        this.addCombatantFromLine(line);
-
-      if (isLineEventTarget(line))
-        this.addCombatantFromTargetLine(line);
-    }
-
-    // Between passes: Create our initial combatant states
-    for (const id in this.initialStates) {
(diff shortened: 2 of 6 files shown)


ui/raidboss/emulator/data/CombatantTracker.ts
```diff
@@ -116,6 +91,7 @@ export default class CombatantTracker {
     initState.maxHp = initState.maxHp ?? extractedState.maxHp;
     initState.mp = initState.mp ?? extractedState.mp;
     initState.maxMp = initState.maxMp ?? extractedState.maxMp;
+    initState.targetable = initState.targetable ?? extractedState.targetable ?? true;
```
quisquous, line 94: I'm not sure I understand the additions to combatant tracker.  Are these a different feature vs a performance optimization?

ui/raidboss/emulator/data/Encounter.ts
```diff
@@ -56,14 +56,14 @@ export default class Encounter {
...
-    this.logLines.forEach((line, i) => {
+    for (const line of this.logLines) {
       if (!line)
         throw new UnreachableCode();
 
       let res: MatchStartInfo | MatchEndInfo | undefined =
           EmulatorCommon.matchStart(line.networkLine);
       if (res) {
-        this.firstLineIndex = i;
+        this.firstLineIndex = this.logLines.indexOf(line);
```
quisquous, line 66: This is premature optimation, but surely keeping some `lineIdx` variable on the side is cheaper than looking through all the log lines to find this line?

ui/raidboss/emulator/data/PopupTextAnalysis.ts
```diff
@@ -81,9 +81,14 @@ export class Resolver {
   }
 }
 
+export type LineRegExpCache = Map<LineEvent, Map<ProcessedTrigger, RegExpExecArray | false>>;
```
quisquous, line 84: All the line matching seems very reasonable to me!

One idle thought I was having (and I'm not sure how complicated this would look like) but would be to do a first pass where you match all lines against all (trigger regex / timeline regex) during log processing.  My assumption is that most lines won't match and so then you'd have 10% (vast overestimate) lines to store in idb.  I don't know how you'd detect if somebody added more triggers or changed regexes though, and so maybe this isn't a plausible approach.  I guess you could do it once after loading here but that's not quite the same level of savings.

ui/raidboss/emulator/data/Combatant.ts
```diff
@@ -93,10 +101,8 @@ export default class Combatant {
       this.significantStates[this.significantStates.length - 1];
     if (!lastSignificantStateTimestamp)
       throw new UnreachableCode();
-    const oldStateJSON = JSON.stringify(this.states[lastSignificantStateTimestamp]);
-    const newStateJSON = JSON.stringify(this.states[timestamp]);
 
-    if (lastSignificantStateTimestamp !== timestamp && newStateJSON !== oldStateJSON)
+    if (lastSignificantStateTimestamp !== timestamp)
```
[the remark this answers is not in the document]
valarnin: I'm planning to change to a different way of tracking if the state has changed, I just haven't included it in this PR yet. Converting both states to JSON and comparing them is pretty slow, especially considering we're already checking the individual properties of the two states.

ui/raidboss/emulator/data/CombatantTracker.ts
valarnin: This is more of a bugfix. The property was already being tracked elsewhere but was missing being set here.

It could be done in a separate PR, I guess?

ui/raidboss/emulator/data/Encounter.ts
valarnin: Lines already track an index but it's indexed against the original log file, not the encounter, which doesn't really make sense. Maybe that should be re-indexed after finalizing the encounter?

ui/raidboss/emulator/data/PopupTextAnalysis.ts
valarnin: By "During log processing" do you mean during the initial log import? Because if that's the case, I don't really think it would make sense in general. That would break combatant tracking, which would break getCombatants triggers.

quisquous: Mmm, that's fair.  I always forget how many lines are needed for that.

ui/raidboss/emulator/data/Encounter.ts
quisquous: Sorry, I'm not sure I understand your suggestion.  I was more just suggesting a micro-optimization here.  `indexOf` is a walk through all of the log lines vs the previous code which just used an incremented variable.  If you don't think that's useful, then ignore this suggestion.

ui/raidboss/emulator/data/CombatantTracker.ts
quisquous: Sorry to be pedantic, but by "this" in "this is more of a bugfix", do you mean only this line? By "the additions to combatant tracker", I meant this line and the other changes to this file.  Are those a different feature as well?  Sorry for the questions, but there was just a bunch of different changes in this PR and it's hard for me to back solve what the motivation for each of them was.

ui/raidboss/emulator/data/Combatant.ts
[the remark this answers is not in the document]
quisquous: Ok, I will take another look in a future version of the PR then.

ui/raidboss/emulator/data/CombatantTracker.ts
valarnin: This addition on line 94 that you've highlighted above is specifically a bugfix.

The other changes to this file are to facilitate only looping over log lines once instead of having to loop over them twice.

ui/raidboss/emulator/data/Encounter.ts
valarnin: During log file parsing during import, the process is basically:

1. Convert log file to `LineEvent`s
2. Split `LineEvent`s into `Encounter`s

During the convert portion this code runs, indexing the `LineEvent`s:
https://github.com/quisquous/cactbot/blob/6a6688cb74b15e9418119e27cfcb380085cb4311/ui/raidboss/emulator/data/NetworkLogConverter.ts#L24-L27

This indexes lines based on their position in the original log file. It should probably be moved entirely, and index them based on their position in the encounter instead, e.g. just before creating the `new Encounter` here:
https://github.com/quisquous/cactbot/blob/6a6688cb74b15e9418119e27cfcb380085cb4311/ui/raidboss/emulator/data/NetworkLogConverter.worker.ts#L27-L29

valarnin:
Closing this now. I have one more follow up PR for this, pending 4311 being merged (due to conflicts).

PR #3361, opened 2021-07-28
raidboss: Add targetability lines to O10n

JLGarber:
What it says on the tin. I have been annoyed by the jump before Exaflare on multiple jobs, but *no more*.

(I also added a count and duration to Protostar since that's moderately helpful to know. Happy to split that out if needed, but I figured a single-line change was small enough to roll up here. Maybe I should add numbers to the Exaflares while I'm at it?)


JLGarber:
You're a cute robot, cactbotbot.

quisquous:
Thanks!

PR #3355, opened 2021-07-26
resources: support position fields for startsUsing

quisquous:

Issue #3338, opened 2021-07-23
UWU Titan Jump Direction


zedekblue:
## Summary

UWU Phase 3, Titan jumps to a random cardinal.
It is determined by which way he faces before jumping, and it is ridiculously brief and hard to notice. Would be great if its possible to have a simple trigger added, much like e12s p2 @ apocs.

Akurosia:
Titan will always jump in the cardinal direction he is looking right befor he jumps

zedekblue:
> 
> 
> Titan will always jump in the cardinal direction he is looking right befor he jumps

He will change direction, it is not baitable.
Titan will turn away from the main tank and look in the direction for only a fraction of a second.

quisquous:
If it is only for a fraction of a second and there is no cast bar, I think this would be tricky to write a trigger for.  Is there a log line that shows which direction Titan is going?

valarnin:
Note that the way the e12s callout works is by delaying for half a second after she starts casting Cataclysm to allow her to fully turn, and even then it doesn't work unless she's reasonably centered (though that could be fixed with some more advanced logic). Cataclysm is a 5.7s cast, for reference.

https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/raid/e12s.ts#L1180

I'm not familiar with the UWU timings, but if the jump is so fast that there can't be a half second delay to allow the boss to fully turn like there is in e12s, it's not really going to be possible I think?

zedekblue:
> 
> 
> If it is only for a fraction of a second and there is no cast bar, I think this would be tricky to write a trigger for. Is there a log line that shows which direction Titan is going?

I am no expert, and this is my first time sifting through a log, but I do see this line right as he is jumping in my ACT log:

`20|2021-07-22T01:49:50.4580000-06:00|40022C3F|Titan|2B66|Geocrush|40022C3F|Titan|2.70|114|100|0|-1.570868||c12700b1032810f60b78cbd66d540d08`

4 seconds later he lands and the damage goes out, and you get 8 lines of this:
[...]

PR #3362, opened 2021-07-28
util: Add type checking for NetRegexes parseHelper fields

valarnin:
I noticed when writing the example UWU trigger in #3338 that when support for the position/heading fields for startsUsing were added in #3355, the underlying type backing NetRegexes weren't updated.

This PR adds the fields to the type, and also adds proper typing of the `fields` parameter for `parseHelper` so that if new field properties are added in the future, they'll need to be added to the underlying type.

### DIFF
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -28,34 +28,17 @@ const keysThatRequireTranslation = [
 ];
 
 type ParseHelperField<
-  T extends TriggerTypes,
-  T2 extends NetFieldsReverse[T],
-  field extends keyof T2
+  Type extends TriggerTypes,
+  Fields extends NetFieldsReverse[Type],
+  Field extends keyof Fields
 > = {
-  field: T2[field] extends string ? T2[field] : never;
+  field: Fields[Field] extends string ? Fields[Field] : never;
   value?: string;
 };
 
 type ParseHelperFields<T extends TriggerTypes> = {
   [field in keyof NetFieldsReverse[T]]: ParseHelperField<T, NetFieldsReverse[T], field>;
 };
-
-const test: ParseHelperFields<'StartsUsing'> = {
-  0: { field: 'type', value: '20' },
-  1: { field: 'timestamp' },
-  2: { field: 'sourceId' },
-  3: { field: 'source' },
-  4: { field: 'id' },
-  5: { field: 'ability' },
-  6: { field: 'targetId' },
-  7: { field: 'target' },
-  8: { field: 'castTime' },
-  9: { field: 'x' },
-  10: { field: 'y' },
-  11: { field: 'z' },
-  12: { field: 'heading' },
-};
-test;
 
 const parseHelper = <T extends TriggerTypes>(
   params: { timestamp?: string; capture?: boolean } | undefined,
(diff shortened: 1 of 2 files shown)


resources/netregexes.ts
```diff
@@ -31,37 +27,70 @@ const keysThatRequireTranslation = [
...
+  5: { field: 'ability' },
+  6: { field: 'targetId' },
+  7: { field: 'target' },
+  8: { field: 'castTime' },
+  9: { field: 'x' },
+  10: { field: 'y' },
+  11: { field: 'z' },
+  12: { field: 'heading' },
+};
+test;
```
quisquous, line 58: Was this leftover from debugging?

resources/netregexes.ts
```diff
@@ -31,37 +27,70 @@ const keysThatRequireTranslation = [
   'line',
 ];
 
+type ParseHelperField<
+  T extends TriggerTypes,
+  T2 extends NetFieldsReverse[T],
```
quisquous, line 32: If there's only one `T` then `T` is (sometimes) ok on its own.  If you have two, then please give them real names for readability (`Type` and `Fields`, maybe?)

resources/netregexes.ts
```diff
@@ -31,37 +27,70 @@ const keysThatRequireTranslation = [
...
+      validFields.push(field.field);
   }
+
   Regexes.validateParams(params, funcName, ['capture', ...validFields]);
 
   // Find the last key we care about, so we can shorten the regex if needed.
   const capture = Regexes.trueIfUndefined(params.capture);
   const fieldKeys = Object.keys(fields);
-  let maxKey;
+  let _maxKey: string;
```
quisquous, line 79: I think underscore on a variable doesn't mean much to me.  Maybe `_maxKey` => `maxKeyStr` to differentiate what it is? (I guess similar `_key` => `keyStr` below for consistency?)

types/net_fields.d.ts
```diff
@@ -77,6 +77,10 @@ type NetStartsUsingFields = Fields & {
   targetId: 6;
   target: 7;
   castTime: 8;
+  x: 9;
+  y: 10;
+  z: 11;
+  heading: 12;
```
quisquous, line 83: Thanks for catching this.  I thought there was something like this file, but my brain was like "maybe that was just in some iteration of raidemulator typescriptization that got removed" because I couldn't find it.

types/net_fields.d.ts
```diff
@@ -306,6 +310,16 @@ export type NetFields = {
   'ActionSync': NetActionSyncFields;
   'StatusEffect': NetStatusEffectFields;
   'None': Fields;
-}
+};
+
+type Reverse<T> = {
```
quisquous, line 315: Can you leave a comment as to what `Reverse` is used for (so folks don't have to puzzle through TypeScript generics to figure it out).

quisquous (commented):
Thanks for catching this! I have some small naming quibbles, but looks great.

resources/netregexes.ts
valarnin: Yeah, this was leftover from me trying to figure out some weirdness with the `ParseHelperField` type.

resources/netregexes.ts
valarnin: Yeah, I named it with the `_` to match the `_key` variable that was already there. Is there maybe an eslint rule to catch this as well, since we're already using `_` to denote parameters that aren't used?

quisquous: https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/naming-convention.md looks plausible (but also complicated).  It'd be nice to just enforce camel case as a starting point.

PR #3370, opened 2021-08-01
raidboss: Add targetability lines to Omega normal

JLGarber:
Figured I might as well complete the set.


quisquous:
Thanks for following up on this!

PR #3372, opened 2021-08-01
raidemulator: change LineEvent to use netlog_defs

quisquous:
Other than 0x19 (was defeated), I left the variables named the same
in the LineEvents to make this change more minimal.

I think it would be worth following up and doing some renaming:
* probably most LineEvent fields should match netregex for consistency?
* changing logDefinitions fields to match netRegex (or add additional names)?
* LineEventSource should probably use sourceId / source
* probably hp vs currentHp vs maxHp should be fixed up in netregexes
  * but this breaks back compat if anybody is using them, so awkward?
  * maybe netregexes could add extra capturing groups to alias fields?


ui/raidboss/emulator/data/network_log_converter/LineEvent0x03.ts
```diff
@@ -1,31 +1,12 @@
...
-  name: 3,
-  jobIdHex: 4,
-  levelString: 5,
-  ownerId: 6,
-  worldId: 7,
-  worldName: 8,
-  npcNameId: 9,
-  npcBaseId: 10,
-  currentHp: 11,
-  maxHpString: 14,
```
quisquous, line 19: :eyes:

PR #3373, opened 2021-08-02
raidemulator: Make LineEvent0x1A not an ability line

quisquous:


quisquous:
@valarnin I'm not sure if this is right, but this was very confusing to see.  In particular, are there really effect ids listed in CombatantJobSearch mixed in with ability ids?

valarnin:
I think this logic was copied over from original raid emulator. I didn't want to change the logic here initially because I didn't fully understand how everything slotted together. I think it's fine to remove this now 👍

PR #3381, opened 2021-08-07
docs: update README to not mention HP option

quisquous:
All of the triggers have been updated to not use this setting
any more, and so the recommendation to use it can be removed.


MaikoTan (approved):
cn lgtm.

PR #3388, opened 2021-08-13
Fix linting issue with dependabot

valarnin:
This should hopefully fix the lint PR title issue in #3387, by allowing `scope(subscope): title`.

### DIFF
--- a/.github/scripts/lint-pr-title.cjs
+++ b/.github/scripts/lint-pr-title.cjs
@@ -140,7 +140,9 @@ const checkTitle = async (octokit, owner, repo, pullNumber) => {
   });
   const { title, user } = pullRequest;
   const userName = user?.login;
-  const m = /^(?<prefix>(?:[\w\[\]]*\s+)!?)?(?<scope>[^:]+):\s?(?<description>.+)$/.exec(title);
+  const m =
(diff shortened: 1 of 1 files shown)


valarnin:
The irony here is not lost on me.

quisquous:
Hmm.  A couple of thoughts.  Minorly, this won't fix it, because `deps` is not a scope.  More importantly though, I'd rather just make dependabot commits use the message `build: bump whatever deps` rather than `build(deps): bump whatever deps` and not enable other people too make titles like `raidboss(oopsy): added stuff`.

Alternate solutions: (1) ignore dependabot in the script entirely (2) make cactabotbot update the title directly when it's dependabot.

valarnin:
I'm not familiar with dependabot's config or anything, so I'm not sure if there's a way to adjust, but if there is that would be the better fix, I agree.

quisquous:
https://dependabot.com/docs/config-file/#commit_message maybe?

valarnin:
Neither of these seem to apply, since the mentioned config files don't already exist.

I'm really not sure how dependabot is even configured on the repo since the documentation I can find indicates that either `.dependabot/config.yml` or `.github/dependabot.yml` should exist in order for it to be active.

quisquous:
Do you see a way to configure dependabot that doesn't involve checking in a config file?

quisquous:
Looks like this was fixed by #3522.

PR #3395, opened 2021-08-15
eureka: TypeScript conversion

quisquous:


ui/eureka/eureka.ts
```diff
@@ -43,42 +152,42 @@ const defaultOptions = {
...
-          return ' pour ' + Math.ceil(min) + ' min ';
+          return ` pour ${Math.ceil(min)} min `;
         }
         return ' pour ???';
       },
       ja: (nowMs, stopTime) => {
         if (stopTime) {
           const min = (stopTime - nowMs) / 1000 / 60;
-          return ' 終わるまであと ' + Math.ceil(min) + ' 分 ';
+          return ` 終わるまであと ${Math.ceil(min)} 分 `;
```
MaikoTan, line 176:
```suggestion
          return ` 終わるまであと${Math.ceil(min)}分 `;
```
It is a little weird to have spaces between numbers and `分`(minute), also there are no spaces needed in Japanese technically as they didn't affect the meaning, so just delete them here and everywhere.

quisquous: Yeah, I was just trying to replicate whatever was there before.  Thanks for taking a look.  It seemed weird to me too.

PR #3400, opened 2021-08-18
raidboss: Fix issue with initial timeline load + user config

### LINKED ISSUE #3399: Custom Timeline Additions Not Showing After Restarting ACT/Game While Instanced
## Description

Recently I've been having an issue with my raidboss custom timeline additions. They work just fine as long as ACT is open when we start the instance, but if I have to restart ACT or the game while we're in the instance, I no longer see any of my extra timeline data. This wouldn't be a problem but when we're progging I frequently add new events to the timeline and have to restart ACT or reload cactbot to update the timeline... but then I don't get any of my events at all unless we reinstance.
When I originally started using cactbot in patch 5.3 reloading cactbot did not have this behavior, it would load my updated timeline as expected even without reinstancing. An update at some point changed this, I'm not sure how/why.

## Additional information

I will add some example screenshots later today once we go back into TEA to prog. See below for an example of my timeline additions from the raidboss file in the user folder.

`Options.Triggers = [
{
    zoneRegex: /(^The Epic of Alexander \(Ultimate\)$)/,
    timeline: function(data) {
      if (data.job != 'DNC')
        return;
      return [
        '82 "👋 Samba"',
        '215 "👋 Dance"',
        '290 "👋 Samba"',
        '355 "👋 blue orange purple green"',
        '664 "👋 Dance"',
        '710 "👋 Samba"',
	'alarmtext "👋 Samba" before 5 "👋 SAMBA"',
	'alarmtext "👋 Dance" before 2 "👋 DANCE"',
      ]},
  },
];`

valarnin:
Fixes #3399

This was caused by #3081

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -565,6 +565,10 @@ export class PopupText {
 
     const orderedTriggers = new OrderedTriggerList();
 
+    // Some user timelines may rely on having valid init data
+    // Don't use `this.Reset()` since that clears other things as well
+    this.data = this.getDataObject();
+
     // Recursively/iteratively process timeline entries for triggers.
     // Functions get called with data, arrays get iterated, strings get appended.
     const addTimeline = (function(this: PopupText, obj: TimelineField | TimelineFunc | undefined) {


quisquous:
Thanks for all the debugging and the quick fix!

PR #3409, opened 2021-08-24
oopsy: change icons to fontawesome

quisquous:
I wanted to try to get back to the original flavor of (colored) win7 emoji with oopsy icons.  I find it very hard to parse the win10 colored emoji when I glance over sometimes.

Original win7 icons:
image

Current win10 icons: (death, fail, warn, pull, potion, wipe, good, damage, heal)
image

Proposed fontawesome icons:
image

Feel free to counterpropose from https://fontawesome.com/v5.15/icons?d=gallery&p=9&m=free or with some other license compatible svg icon.

### DIFF
--- a/ui/oopsyraidsy/oopsy_common.css
+++ b/ui/oopsyraidsy/oopsy_common.css
@@ -5,8 +5,20 @@
 
 .mistake-icon {
   text-align: center;
-  text-shadow: -1px 0 1px grey, 0 1px 1px grey;
+  text-shadow: -1px 0 2px grey, 0 1px 2px grey;
   width: 20px;
+}
+
+.mistake-icon::before {
+  display: inline-block;
+  font-style: normal;
+  font-variant: normal;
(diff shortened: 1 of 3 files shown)


valarnin:
For heal, maybe:

image

For damage, maybe:

image

quisquous:
image

image

Theoretically I like `first-aid` better, but I played with it some and in the live list with the glow-y shadow, `first-aid` loses some of its detail over `plus-square`.

`bomb` is nice too, but I worry that it's another circle, with both `ban` and `skull` being pretty circular too.  The two arrows (and the knife in the old version) are all much more distinctive shapes that they are easier to mentally parse, imo.  If fontawesome ever adds crossed swords / knife / etc, I think I'd be happy to switch to that.

quisquous:
I think I'm gonna just land this as-is.  These are all overridable in user css, and so I mostly just want to provide good defaults.  Thanks for all the suggestions.

PR #3410, opened 2021-08-25
oopsy: move buff tracking into code rather than triggers

quisquous:
The goal here is to extend what is being done with missed buffs.  The
primary goal is to also track which players are alive or dead, and to
then ignore missed buffs on dead players.  However, the later goal is to
add death reporting, where it can say when a player died if they missed
any buffs in the past as extra information.

So, take all of the buff tracking code, and move it into special classes
to handle that in a way that can be extended more than triggers.

Also, because I'd like to eventually create some sort of oopsy
emulation, this new MissedBuffCollector doesn't use setTimeouts and
instead uses log lines for timing things.  It makes things a little bit
more convoluted though, and so I'm not sure this is the best way to
handle it.


ui/oopsyraidsy/damage_tracker.ts
```diff
@@ -141,15 +163,40 @@ export class DamageTracker {
...
+    // Ignore game log lines, which don't track milliseconds.
+    if (type !== '00') {
+      let timestampCallback = this.timestampCallbacks[0];
+      while (timestampCallback) {
+        const timeField = splitLine[1];
+        if (!timeField)
+          break;
+        const thisTimestamp = new Date(timeField).getTime();
+        if (thisTimestamp < timestampCallback.timestamp)
+          break;
```
valarnin, line 176: Is there any guarantee that callbacks are sorted by timestamp?

quisquous: Ah!  I appreciate you catching that.

PR #3418, opened 2021-08-27
raidboss: Misc FR translations

mooondark:


VanSlanzar:
Thanks for translations. I have no time at the moments.

quisquous:
> Sorry I don't get it. Can you tell me why I get this error ?

You should be able to run `npm run lintfix` and it will format things as needed without you needing to do it manually.  If that doesn't work, I'm happy to help get it working for you.

PR #3422, opened 2021-08-28
oopsy: don't report dead players for missed buffs

quisquous:


ui/oopsyraidsy/effect_tracker.ts
```diff
@@ -266,7 +294,13 @@ export class EffectTracker {
     for (const id of collected.targetIds)
       gotBuffMap[id] = true;
 
-    const missedIds = this.partyTracker.partyIds.filter((id) => !gotBuffMap[id] && id !== emptyId);
+    const missedIds = this.partyTracker.partyIds.filter((id) => {
+      // Filter out any empty ids here.
+      if (id === emptyId)
+        return false;
```
panicstevenson, line 300: This is unrelated to the dead people thing, right?

panicstevenson (approved):

ui/oopsyraidsy/effect_tracker.ts
quisquous: Yeah.  This check was there before.

panicstevenson: Oh, I'm bad at reading. I read the two checks as the same two checks below. Obviously that doesn't make sense.

PR #3424, opened 2021-08-29
raidboss: Fix *_only.html pages to also include OVERLAY_WS

### LINKED ISSUE #3403: `Alerts only` and `Timeline only` html redirects don't include OVERLAY_WS
## Description

The helper HTML pages for alerts only and timeline only don't include the OVERLAY_WS param when redirecting, which makes generated WS URLs in OverlayPlugin's WSServer tab not work.

valarnin:
Fixes #3403

### DIFF
--- a/ui/raidboss/raidboss_alerts_only.html
+++ b/ui/raidboss/raidboss_alerts_only.html
@@ -1,1 +1,10 @@
-<meta http-equiv="refresh" content="0; url=raidboss.html?timeline=0&alerts=1" />
+<script type="text/javascript">
+  let redirUrl = 'raidboss.html?timeline=0&alerts=1';
+  if (typeof window !== 'undefined') {
+    const wsUrl = /[\?&](OVERLAY_WS=[^&]+)/.exec(window.location.href);
+    if (wsUrl) {
+      redirUrl += '&' + wsUrl[1];
+    }
+  }
+  window.location.href = redirUrl;
+</script>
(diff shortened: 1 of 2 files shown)


ui/raidboss/raidboss_alerts_only.html
```diff
@@ -1 +1,10 @@
-<meta http-equiv="refresh" content="0; url=raidboss.html?timeline=0&alerts=1" />
\ No newline at end of file
+<script type="text/javascript">
+  let redirUrl = 'raidboss.html?timeline=0&alerts=1';
+  if (typeof window !== 'undefined') {
+    const wsUrl = /[\?&](OVERLAY_WS=[^&]+)/.exec(window.location.href);
```
quisquous, line 4: Would it make sense to preserve all the params minus timeline and alerts via `URLSearchParams` or something?

valarnin: Possibly, though I don't know of any occasion where any other params would happen. This is only needed because the WS URL generator appends OVERLAY_WS and that causes problems for people that want to use it for streaming or XL BrowserHost.

If people are adding their own params to the URL, they're probably capable of using the raidboss.html page with params instead?

quisquous: Sure, agreed that it's unlikely.  It just seemed better to not have to parse urls manually and to handle future edge cases at the same too.  I think this is an improvement for now, so will merge.

PR #3427, opened 2021-09-02
raidboss: Delay Diamond Weapon callouts

valarnin:
As pointed out in a reddit thread, the east/west callouts in Diamond Weapon happen before the visual cue, which wasn't intended and isn't really fair. Update them to coincide with the visual callouts.


quisquous:
Hmm.  I'm not sure I want to get into a policy of trying to line up abilities with what people perceive to be the beginning of the visual clue.  I think these are extremely bread and butter starts using triggers that anybody could write with even an act custom trigger.  I'm inclined to leave this as-is.

valarnin:
I bring this up not just because of the reddit thread, but also because of the previous conversation on #2329

quisquous:
I think what's primarily over the line about #2329 is that it's not something that anybody could call out given the information on screen.  I think secondarily, it's also a complicated trigger that is not "just" a startsUsing trigger on something in a log line that anybody with basic trigger writing knowledge could write.

I think when writing triggers, I would rather have people think about what the best time to call something out is to be helpful (e.g. not conflicting with other triggers, not so soon it's forgotten, or not so late that it's a scramble to address) rather than trying to evaluate and police on a per trigger basis what is "fair".  My inclination here is to say no to including triggers by default that are over that line, and leave everything else in the grey area as-is.

Tsunayoshi:
Tbh i am happy about the early callout. Its enough time to call it for ps5 ppl and to react. If u use cactbot then the time for "non cheating play" is already gone. So from my view, this does not matter at all anymore.
Beside that i changed the callouts to red and blue because E and W just confuse me xD

PR #3432, opened 2021-09-11
raidboss: remove use of CaresAbout from triggers

quisquous:
There have been numerous requests and confusion over time around this
issue.  In particular, it's not clear to most people why sometimes
triggers show up and sometimes they don't (causing them to file bugs or
ask in discord).

This was added in the long ago past when there was no config UI and it
was harder to disable triggers.  Now that it's easy to just click off
anything somebody dislikes, avoiding being too noisy is much less of
a goal.

I considered the alternative of adding an option for this, but I think
there are already WAY too many options in cactbot and I think this would
be confusing.  I'd rather just funnel everybody towards the easier to
explain path of "if you don't like it, turn it off in the UI".


panicstevenson (approved):
I like this idea, I think that this is something that gets asked a lot, and there's people that want "all callouts no matter what". However, I don't feel like that's really possible, since there's not a great way in conditional outputs to know what to say as a "general raidcaller" (and who wants to update all the triggers to do that?). I would be open if someone else had an idea of how to do this, but I think from a maintenance standpoint this is by far the easiest, and allows users to customize it how they'd like, with a small expense of now they might see a physical/magical tankbuster they weren't seeing before.

EDIT: Other people feel free to chime in too if they have thoughts!

quisquous:
This has fallen off the first page, so I'm going to assume that nobody else has opinions here.  I'll go ahead and merge this.

PR #3446, opened 2021-09-14
i18n: added 5.5 KR translations

Akurosia:


quisquous:
Thank you, as always!!

Issue #3292, opened 2021-07-12
Suggestion: Switch package manager to Yarn 2


MaikoTan:
[Yarn 2 (Berry)](https://yarnpkg.com/) brings many features:

- 🔌 [Plug'n'Play](https://yarnpkg.com/features/pnp). To get rid of the extremely biggest `node_modules`, it also make node scripts run faster. _moon, earth, sun, black hole, node_modules._

- 🏎️ Faster installation. Not only because of the [Zero-installs](https://yarnpkg.com/features/zero-installs) feature, but Yarn installs packages in parallel, however npm does this one by one, usually Yarn is faster than npm several times.

- 🔩 Cross-platform scripts. Yarn includes a simple shell interpreter, make scripts runs consistently across platforms, no matter which OS/Shell you're running from. _It is time to get rid of `node_modules/.bin/blahblah` prefixes!_

There are also disadvantages:

- 🔧 The first one is obvious: you should install Yarn manually and globally, it is not bundled like npm.

- 🐛 The [Plug'n'Play feature is not compatible with all tools](https://yarnpkg.com/features/pnp#compatibility-table), or only be supported by the new version, so it might be painful to make a bunch of tools work.

- 😫 Many IDE/Editor would assume that you are running npm by default, although you can configure it manually, for example in [VSCode ESLint plugin](https://github.com/Microsoft/vscode-eslint#settings-options).

------

Feel free to comment your thoughts! (づ｡◕‿‿◕｡)づ

quisquous:
I don't know much about yarn, but I know it's widely used.  I think really my hesitation is that this creates more friction because it's another dependency to install.  Faster sounds nice, though.  :)

MaikoTan:
Yeah, Yarn is widely used. Mostly I like its modern and fashion CLI. It is also bundled in `setup-node` action in Github Actions, so I think the most important thing is that how many cactbot developers supports it?

JLGarber:
From a "developer-user" standpoint, my primary concern is "how much work will it take to make our existing tools work with Yarn?" I'm completely unfamiliar with Yarn, so anything that needs to be modified will block me completely until Someone Else is able to convert tools. This isn't necessarily a problem, since I would assume part of the process of moving to Yarn would be that conversion step, but it is something to be aware of. 

[...]

PR #3447, opened 2021-09-16
build: Yarn Berry Playground

MaikoTan:
Per #3292 

### Plug'n'Play
PnP feature is enabled, but there are still errors with mocha and else.

### Offline Cache
As yarn suggests, the `cache` (aka packages from npm) is bundled and uploaded to Git, although the size is much smaller than `node_modules`:
```bash
$ du -sh .yarn
190M    .yarn

$ du -sh .yarn/cache
61M     .yarn/cache

$ du -sh node_modules
507M    node_modules
```

------

## How to try
Switch to this branch and then run `yarn` first, yarn would extract cache for you, and then replace all `npm run blahblah` as `yarn run blahblah`


MaikoTan:
The Github Actions CI failed because the `--immutable` flag for `yarn install` command was default to `true` in a CI environment
> If the `--immutable` option is set (defaults to `true` on CI), Yarn will abort with an error exit code if the lockfile was to be modified (other paths can be added using the `immutablePatterns` configuration setting). For backward compatibility we offer an alias under the name of `--frozen-lockfile`, but it will be removed in a later release.
And somewhat that the `checksum` of packages from registry on Github Actions CI are not equal to those in the local lockfile, so cause the fail.
But even I run the command `yarn install --immutable` on my computer, it works fine. I don't know what's wrong with Github Actions, maybe change the registry to npmjs would solve the problem?

@JLGarber Can you try this PR and give me some feedback? I believe it is runnable now.

PR #3456, opened 2021-09-17
jobs: reduce regex matching via logDefinition

MaikoTan:
Also convert massive "if-else" to "switch-case" for readability.

This PR should improve the performance a bit?

### DIFF
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -77,7 +77,7 @@ class Bars {
     this.inCombat = false;
     this.combo = undefined;
     this.comboTimer = undefined;
-    this.regexes = new RegexesHolder(this.options.ParserLanguage);
+    this.regexes = undefined;
     this.partyTracker = new PartyTracker();
 
     this.skillSpeed = 0;
@@ -852,6 +852,12 @@ class Bars {
   }
 
   _onPlayerChanged(e) {
+    if (this.me !== e.detail.name) {
+      this.me = e.detail.name;
+      // setup regexes prior to the combo tracker
+      this.regexes = new RegexesHolder(this.options.ParserLanguage, this.me);
+    }
+
     if (!this.init) {
       this.combo = ComboTracker.setup(this._onComboChange.bind(this));
       this.init = true;
@@ -957,26 +963,26 @@ class Bars {
   }
 
   _onNetLog(e) {
-    if (!this.init)
+    if (!this.init || !this.regexes)
       return;
     const line = e.line;
+    const log = e.rawLine;
 
     const type = line[logDefinitions.None.fields.type];
 
     switch (type) {
       case logDefinitions.GameLog.type: {
-        const message = line[logDefinitions.GameLog.fields.line];
-        const m = this.regexes.countdownStartRegex.exec(message);
+        const m = this.regexes.countdownStartRegex.exec(log);
         if (m && m.groups?.time) {
           const seconds = parseFloat(m.groups.time);
           this._setPullCountdown(seconds);
         }
-        if (this.regexes.countdownCancelRegex.test(message))
+        if (this.regexes.countdownCancelRegex.test(log))
           this._setPullCountdown(0);
-        if (/:test:jobs:/.test(message))
+        if (/:test:jobs:/.test(log))
           this._test();
         if (Util.isCraftingJob(this.job))
-          this._onCraftingLog(message);
+          this._onCraftingLog(log);
         break;
       }
 
--- a/ui/jobs/jobs.js
+++ b/ui/jobs/jobs.js
@@ -28,6 +28,7 @@ import {
   computeBackgroundColorFrom,
   doesJobNeedMPBar,
   makeAuraTimerIcon,
+  normalizeLogLine,
   RegexesHolder,
 } from './utils';
 import { getReset, getSetup } from './components/index';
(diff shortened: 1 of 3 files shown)


ui/jobs/utils.ts
```diff
@@ -1,76 +1,43 @@
...
-    this.countdownCancelRegex = getCurrentRegex(LocaleNetRegex.countdownCancel);
+    this.cordialRegex = /20(017FD|F5A3D|F844F|0420F|0317D)/;
+
+    this.countdownStartRegex = localeRegexes.countdownStart;
+    this.countdownCancelRegex = localeRegexes.countdownCancel;
     this.craftingStartRegexes = [
-      LocaleNetRegex.craftingStart,
-      LocaleNetRegex.trialCraftingStart,
-    ].map(getCurrentRegex);
+      localeRegexes.craftingStart,
```
quisquous, line 28: The if/then => switch changes look fine, but I'm not sure I understand what the goal here is.  It seems like you could have left the original `getLocaleRegex` code and your "matching via logDefinition" purpose would still be met?

I think if you don't like the ordering of line => lang => regex and would rather index on lang => line => regex, I'd rather change `LocaleRegexesObj` to do that rather than having `LocaleRegexesObj` and `getLocaleRegexes` as two separate ways of building the same information, if that makes sense.  (And I think that would make more sense as a separate change than to put in here, unless I'm misunderstanding.)

ui/jobs/jobs.js
```diff
@@ -961,118 +957,142 @@ class Bars {
...
-      if (
-        line[2] === this.lastAttackedDotTarget &&
-        line[4] === 'DoT' &&
-        line[5] === '0'
-      ) {
-        // 0 if not field setting DoT
-        this.updateDotTimerFuncs.forEach((f) => f());
+
+      case logDefinitions.NetworkDoT.type: {
+        // line[2] is dotted target id.
```
quisquous, line 1083:
```suggestion
        // line[fields.id] is dotted target id.
```

ui/jobs/utils.ts
```diff
@@ -214,3 +181,17 @@ export const makeAuraTimerIcon = (
...
+
+export const normalizeLogLine = <Fields extends NetAnyFields>(
+  line: string[],
+  fields: Fields,
+): Partial<Record<keyof Fields, string>> => {
+  return new Proxy({}, {
+    get(_target, property) {
+      if (typeof property === 'string' && property in fields) {
+        const fieldKey = fields[property as keyof Fields];
+        return line[fieldKey as unknown as number];
```
quisquous, line 193:
```suggestion
        const looseFields: { [prop: string]: number } = fields;
        const fieldKey: number | undefined = looseFields[property];
        if (fieldKey)
          return line[fieldKey];
```

Suggestion for fewer type assertions.

ui/jobs/utils.ts
MaikoTan: Yeah, you are right. But I think it looks like it is redundant to match/extract something like `timeline` or `type` that we don't need, and that I think it should improve performance a bit? Because extracting items from an array is just `O(1)`.

I wanted to replace the whole `RegexSet` in `translation.ts` to `getLocaleRegexes` mentioned above, but there are not only `jobs` module relies on them, so I leave them still alive as this PR changed `jobs` module only. So it is not an ordering thing? I am just changing the regex match range/scope to the message only....

quisquous: Sorry, by `timeline` and `type`, what are you referring to?

And yeah, other things use `LocalRegexesObj` / `RegexSet`, so those would need to be changed as well.  That's why I was suggesting to do that in another PR where you could change all of them at once to something more convenient.

MaikoTan: Oh, sorry, it is `timestamp` and `code`....
I am referring that when we are using the RegExp that generated from NetRegexes, the full regex would be like (in jp):
```js
/^(?<type>(?:00))\|(?<timestamp>(?:[^|]*))\|(?<code>(?:[^|]*))\|(?<name>(?:[^|]*))\|(?<line>(?:戦闘開始まで(?<time>-?[0-9]+(?:[.,][0-9]+)?(?:E-?[0-9]+)?)秒！ （(?<player>.*?)）[^|]*?))\|/
```
There are many match groups, but we need only is the `line` group, others like `timestamp` or `code` are not required in this situation, right? I think as we have the `logDefinition` that records all items' positions already, we can directly extract the `line` part via an array indexing (`log[fields.line]`), instead to do a longer regex matching.

quisquous: Sorry for any misunderstanding, but I think we are talking about different things.

I am trying to say that I think your patch is doing two different (unrelated) things:
(1) remove a bunch of regexes in favor of a proxy / `logDefinitions`, change `if` to `switch (e.g. `YouGainEffectRegex`)
(2) change the initialization of other regexes that you haven't removed (e.g. `countdownStartRegex`)

I am saying that (1) should be this PR on its own and is all good.  I don't have any more feedback on that.  I think what you are responding to above in the previous comment is about (1), if I understand correctly.  Is it possible to make this PR only do (1) and not (2), and then I would merge it?

The (2) part is what I was trying to get at with these comments.  I was mostly saying that it feels like you are adding a second method to build/access `LocaleRegexesObj` / `RegexSet` and I think having both `buildLocaleRegexes` and `getLocaleRegexes` is somewhat redundant.  If you want to change how `translations.ts` works, then I think it'd make sense to do that in a separate PR and change other parts of the code that use it.

MaikoTan: Ah, I see your point. Sorry for the unrelated responses.
Sure, I can try to separate the (2) things into a new PR later.

PR #3457, opened 2021-09-18
ci: add version to package.json

MaikoTan:
If no this field, cactbot cannot be imported as a dependency
when we want to use its definitions in other repositories.

Also change create-tag.yml to read version from package.json and
write into AssemblyInfo.cs files.
So that you can easily run `npm version patch` to increase
 the last number of the version, or `npm version minor` to increase the
 second number and commit automatically.
 And then let Github Actions helps you to do the rest things.

------

I am not very familiar with bash shell script, hope I didn't make mistakes. ( ･ั﹏･ั)


.github/workflows/create-tag.yml
```diff
@@ -17,23 +17,20 @@ jobs:
...
       - name: Checkout Code
         uses: actions/checkout@v2
-      - name: Parse Plugin Version
+      - name: Parse Version
         id: parse-version
         run: |
-          # Parse new tag from AssemblyFileVersion
-          VERSION="$(cat plugin/CactbotEventSource/Properties/AssemblyInfo.cs | sed -ne 's/.*AssemblyFileVersion.*"\([0-9]*\.[0-9]*\.[0-9]*\)\.0".*/\1/p')"
+          # Parse new tag from package.json
+          VERSION="$(cat package.json | sed -ne 's/"version": "\([0-9]*\.[0-9]*\.[0-9]*\)",/\1/p')"
```
quisquous, line 24: I think the logic above is not quite correct.  This workflow is looking for versions in the .cs files, but also writes the .cs files??

If you'd like to add the tag to package.json, that seems good.  What about leaving this script mostly as-is and adding an additional check that all three versions (instead of all two) versions match before creating a tag? I think it's an error not to have all of these match.

.github/workflows/create-tag.yml
```diff
@@ -17,23 +17,20 @@ jobs:
...
-            echo "Matching assembly file version values found."
-          else
-            echo "Different assembly file version values found. Skipping tag creation..."
-            exit 1
-          fi
+          # Write version to both AssemblyInfo files
+          sed -i "s/\(.*Version.*\)\"\([0-9]*\.[0-9]*\.[0-9]*\.0\)\"/\1\"${VERSION}.0\"/g" plugin/CactbotEventSource/Properties/AssemblyInfo.cs plugin/CactbotOverlay/Properties/AssemblyInfo.cs
+
+          # Output version
+          echo "VERSION=$VERSION" >> $GITHUB_ENV
```
quisquous, line 30: Why did you change TAG_NAME to VERSION?

.github/workflows/create-tag.yml
MaikoTan: Sorry, maybe I should clarify my purpose first.
It can read version number from `package.json` and then write into .cs files.

When you publish a release, you should always change the version number in two `AssemblyInfo.cs` files, and then let Github Actions run this workflow and release it, right?
In fact, `npm version` is a command that can increase version number properly and do a git commit (also `git-tag` it if you like), so I think it should be easier to run a command like `npm version patch` and push, then Github Actions would help you to change the version numbers in `AssemblyInfo.cs` files and create the corresponding tag.

MaikoTan: Ah! ⊙﹏⊙∥
I didn't create a git commit that includes .cs files after I changed their version numbers...

quisquous: Yes, but in this patch there would be two commits that say "Bump version number"? One from the `package.json` commit and then one from the workflow, right?

I guess my feeling is that this isn't that hard to do, and that adding a script here makes this more complicated rather than easier?

I am supportive of having a version number in `package.json`.  That sounds great.  If you felt like this needed a script and automation, then I think I would rather have a local script (e.g. `npm run bumpversion 0.23.7`) than a workflow.

MaikoTan: Done!
There is a hook script in running [`npm version` command](https://docs.npmjs.com/cli/v7/commands/npm-version) called "version", so I can add a js script to sync version number from package.json to AssemblyInfo.cs files.

After that, you can just `npm version patch` or `npm version minor` to increase the version number and automatically commit the bumping change, or something like `npm version 0.22.1` to specify a number, more detail in [npm's doc](https://docs.npmjs.com/cli/v7/commands/npm-version)

.npmrc
```diff
@@ -0,0 +1,3 @@
+# set git commit when bump version
+message="Bump version to %s"
```
quisquous, line 2: Can this file be removed?

quisquous:
Thank you!  The script works great.  I never knew you could hook npm version like that.  That's neat!  I have one more question and then this can be merged.

.npmrc
MaikoTan: Don't you like an extra config file for npm?
if you didn't set the `message`, the git commit message would fallback to version number only like `0.23.6` only when you bump versions.
If you still need a message says `bump version to blahblah`, you can set this option globally via running `npm config set <key> <value>`, or just add a `-m <message>` when you are bumping versions.

quisquous: > Don't you like an extra config file for npm?

:joy:

Sorry, I don't understand what message there is.  Running `npm version patch` seems to modify the files directly but does not `git add` or `git commit` them.  Is it supposed to do something else?  Even if I `npm version patch -m "message"` it does not commit.

MaikoTan: That's weird. What version are your nodejs and npm?
According to [npm's doc](https://docs.npmjs.com/cli/v7/commands/npm-version), when you run `npm version` command in a git repo (which should be a clean one), it would automatically create a commit and tag it.

quisquous: npm version is 6.14.8.  nodejs is 14.15.1.

MaikoTan: Yeah, I am using npm v7.x , and I can reproduce.
Let me figure out why and how to deal with it later.
```bash
$ nvm use 14.15.1
Now using node v14.15.1 (npm v6.14.8)

$ npm install -g npm@6.14.8
/home/***/.nvm/versions/node/v14.15.1/bin/npm -> /home/***/.nvm/versions/node/v14.15.1/lib/node_modules/npm/bin/npm-cli.js
/home/***/.nvm/versions/node/v14.15.1/bin/npx -> /home/***/.nvm/versions/node/v14.15.1/lib/node_modules/npm/bin/npx-cli.js
+ npm@6.14.8
updated 1 package in 5.345s

$ node -v
v14.15.1

$ npm -v
6.14.8
$ cat package.json | grep version
  "version": "0.23.6",
    "version": "node util/sync_version.cjs"

$ git log --oneline | head -3
b5935e4a chore: add script sync version number
865cf1ac chore: dont tag with npm
e5fb4b4f chore: add version in package.json

$ npm version minor
v0.24.0

> cactbot@0.24.0 version /workspaces/cactbot
> node util/sync_version.cjs



   ╭────────────────────────────────────────────────────────────────╮
   │                                                                │
   │      New major version of npm available! 6.14.8 → 7.24.0       │
   │   Changelog: https://github.com/npm/cli/releases/tag/v7.24.0   │
   │               Run npm install -g npm to update!                │
   │                                                                │
   ╰────────────────────────────────────────────────────────────────╯

$ cat package.json | grep version
  "version": "0.24.0",
    "version": "node util/sync_version.cjs"

$ git log --oneline | head -3
b5935e4a chore: add script sync version number
865cf1ac chore: dont tag with npm
e5fb4b4f chore: add version in package.json
```

MaikoTan: Ah, I know it.
That's because I set `git-tag-version` option as false in `.nvmrc`, and this option behaves differently in npm v6.x and v7.x:

- v6.x
> git-tag-version
> - Default: true
> - Type: Boolean
> Commit and tag the version change.
- v7.x
> git-tag-version
> - Default: true
> - Type: Boolean
> Tag the commit when using the npm version command.

So....can I just remove the `create-tag.yml` and let npm create a tag for you? Only you should remember to push tags as well.

quisquous: Yeah, removing `create-tag.yml` seems reasonable.  If I change `git-tag-version=true`, I get the commit properly.  Nice find!

However, the commit does not include the assembly files (which are unstaged).  Does `util/sync_version.cjs` need to `git add` or `git commit -a --amend -C HEAD` them?

MaikoTan: Done!
Yeah, `npm` suggest to run an explicit `git add .` too, so I add it.

quisquous: Thank you! I wasn't sure what the right thing to do was.  This works great.  I made one last edit and will merge this.  I changed `git add .` to `git add -u` so that it didn't add any loose untracked files in the repo, only tracked changed files.

PR #3465, opened 2021-09-22
resources: change export type of resources/util.ts as named

MaikoTan:
as named export

There are two export approaches in an es module:
1. Named export
2. Default export

Currently `resources/util.ts` use the 2. default export,
although the statement `import Util from 'path/to/util.ts'` is clean
and easy to recognize,
to export an object makes webpack cannot `tree-shake` it correctly.

Therefore, I think we should change it to named export style,
perhaps also the `resources/condition.ts` file

BREAKING CHANGES:
`import Util from 'path/to/resources/util'` is not available anymore,
should change to `import * as Util from 'path/to/resources/util'`,
or `import { xxx } from 'path/to/resources/util'`

### DIFF
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -26,7 +26,7 @@ import { Responses as _Responses } from './responses';
 const Responses = _Responses;
 import _Outputs from './outputs';
 const Outputs = _Outputs;
-import _Util from './util';
+import * as _Util from './util';
 const Util = _Util;
 import _ZoneId from './zone_id';
 const ZoneId = _ZoneId;
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -26,8 +26,52 @@ import { Responses as _Responses } from './responses';
 const Responses = _Responses;
 import _Outputs from './outputs';
 const Outputs = _Outputs;
-import * as _Util from './util';
-const Util = _Util;
+import {
+  canAddle,
+  canCleanse,
+  canFeint,
+  canSilence,
+  canSleep,
+  canStun,
+  clearWatchCombatants,
+  getAllRoles,
+  isCasterDpsJob,
+  isCombatJob,
+  isCraftingJob,
+  isDpsJob,
+  isGatheringJob,
+  isHealerJob,
+  isMeleeDpsJob,
+  isRangedDpsJob,
+  isTankJob,
+  jobEnumToJob,
+  jobToJobEnum,
+  jobToRole,
+  watchCombatant,
+} from './util';
+const Util = {
+  canAddle,
+  canCleanse,
+  canFeint,
+  canSilence,
+  canSleep,
+  canStun,
+  clearWatchCombatants,
+  getAllRoles,
+  isCasterDpsJob,
+  isCombatJob,
+  isCraftingJob,
+  isDpsJob,
+  isGatheringJob,
+  isHealerJob,
+  isMeleeDpsJob,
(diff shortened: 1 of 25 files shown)


MaikoTan:
emmmm...
Why can't I use the exclamatory notation to mark a breaking change.

quisquous:
> emmmm...
> Why can't I use the exclamatory notation to mark a breaking change.

I don't know! It looks from the code like you should.

Also, why is this a breaking change? It looks like user triggers will not have to change because of this, so I'm not sure what is breaking.

MaikoTan:
> Also, why is this a breaking change? It looks like user triggers will not have to change because of this, so I'm not sure what is breaking.

The import style?
you cannot use the original one but should use `* as Util` or just import part of it.

quisquous:
> > Also, why is this a breaking change? It looks like user triggers will not have to change because of this, so I'm not sure what is breaking.
> 
> The import style?
> you cannot use the original one but should use `* as Util` or just import part of it.
> This breaks some code that imports `util.ts`.

Yeah, I guess my question is really a bigger question of "what is a breaking change, for cactbot".  I have been thinking about it as "does this break user triggers that may have been using old code". In this case, since `Util.canStun()` etc still exist, it doesn't feel like a breaking change under that rule.

I don't think changing an import style is breaking in the same way that I don't think changing a function parameter or a function name or something like that (as long as it's outside of user triggers).  Otherwise, almost everything is a breaking change.  What do you think?

resources/user_config.ts
```diff
@@ -26,7 +26,7 @@ import { Responses as _Responses } from './responses';
 const Responses = _Responses;
 import _Outputs from './outputs';
 const Outputs = _Outputs;
-import _Util from './util';
+import * as _Util from './util';
```
quisquous, line 29: Would you mind changing this to be explicit imports? I think I should add eslint to not have `import *` for consistency.

Also, I think in this case, it would be better to only import the functions that used to exist and put them in a structure called `Util` so that we can only include functions in user triggers that we want to.

MaikoTan: Are you saying that should be changed to something like:
```ts
import {
  isTankJob,
  isHealerjob,
  // ...
} from './util';
const Util = {
  isTankJob,
  isHealerjob,
  // ...
};
```
?

quisquous: Yeah.  It's a little wordy, but I think it's better to be explicit than include all future additions to `Util`.

MaikoTan:
Ah, I see your point!
So you think that this PR only apply the change inside cactbot and/or break someone(developers)'s code that imports cactbot's type directly; for an end user, the change is insensitive or ignorable, so it should not be a "break change"?
I agreed, I may think about this change at the first time as a developer.ಠ_ಠ

quisquous:
> Ah, I see your point!
> So you think that this PR only apply the change inside cactbot and/or break someone(developers)'s code that imports cactbot's type directly; for an end user, the change is insensitive or ignorable, so it should not be a "break change"?
> I agreed, I may think about this change at the first time as a developer.ಠ_ಠ

Yeah, I think of a breaking change as "breaking your public api" or as semver puts it "making an incompatible api change".  But, the real question is "what is cactbot's public api".  At the moment, for me, I think cactbot's public api is only end user triggers.  I think if there are developers who have projects that are importing cactbot, at the moment it is up to them to handle changes as there are no promises that internal details will stay the same from release to release.

PR #3467, opened 2021-09-23
raidboss: Extends caresAboutAOE, changes some AOE callouts

xiashtra:
Extends caresAboutAOE to include physical ranged jobs with party mitigation abilities and other jobs with personal mitigation abilites.

Adds canTactician for physical ranged jobs with party mitigation abilities:
	BRD: Troubadour (ARC base class does not get this ability)
	MCH: Tactician
	DNC: Shield Samba

Adds canPersonalMitigation for jobs with personal mitigation abilities:
	MNK: Riddle of Earth (PGL base class does not get this ability)
	SAM: Third Eye
	THM, BLM: Manaward

Adjusts some AOE callout conditions for consistency.


quisquous:
Thanks for this PR! I am thinking about going the opposite direction and merging #3432, which removes these conditions rather than extending them.  (Or, it removes CaresAbout, and I'll handle the manual conditions after that in a similar way.)  I think a lot of people are confused why some triggers don't happen on certain jobs (and they feel like it's broken, because they know there should be a tankbuster/aoe callout in some places) and so I'm thinking of just showing them for everybody.  Feel free to add any thoughts there, if you'd like.

xiashtra:
> Thanks for this PR! I am thinking about going the opposite direction and merging #3432, which removes these conditions rather than extending them. (Or, it removes CaresAbout, and I'll handle the manual conditions after that in a similar way.) I think a lot of people are confused why some triggers don't happen on certain jobs (and they feel like it's broken, because they know there should be a tankbuster/aoe callout in some places) and so I'm thinking of just showing them for everybody. Feel free to add any thoughts there, if you'd like.

Between tanks, healers, BLU, canAddle, and what I've added above, all but two jobs would get a notification on almost every raidwide, so removing the conditionals altogether makes just as much sense.  Especially with the changes coming in Endwalker to Feint and Addle, when either/both ability will be an appropriate response to almost any raidwide attack.

If you are going to remove these conditionals, I will just re-focus my efforts to cleaning up the inconsistencies (I came across several instances that used hardcoded job comparisons instead of CaresAbout conditionals).

xiashtra:
No longer applicable after https://github.com/quisquous/cactbot/pull/3432 was merged.

PR #3468, opened 2021-09-23
resources: bug fix of overlay_plugin_api.ts

nyaoouo:


nyaoouo:
1. add `try... catch...` in `processEvent`, preventing an exception occurred then take down all the follows
2. check index of callback correctly at `removeOverlayListener`, preventing callback at index 0 cant be remove
3. add `decodeURIComponent` at `init`, ensure that the link address can be get correctly

MaikoTan:
For getting the `OVERLAY_WS` parameter, I think it is better to use `URLSearchParams`:
```js
const urlParams = new URLSearchParams(window.location.search);
const wsURL = urlParams.get('OVERLAY_WS');
```

nyaoouo:
[``](url)

> For getting the `OVERLAY_WS` parameter, I think it is better to use `URLSearchParams`:
> 
> ```js
> const urlParams = new URLSearchParams(window.location.search);
> const wsURL = urlParams.get('OVERLAY_WS');
> ```

thx, add as commit

resources/overlay_plugin_api.ts
```diff
@@ -115,7 +122,7 @@ export const removeOverlayListener: IRemoveOverlayListener = (event, cb): void =
     const list = subscribers[event];
     const pos = list?.indexOf(cb as VoidFunc<unknown>);
 
-    if (pos && pos > -1)
+    if (pos !== undefined && pos > -1)
```
quisquous, line 125: Thanks! This definitely looks like a bug.

resources/overlay_plugin_api.ts
```diff
@@ -189,10 +196,12 @@ export const init = (): void => {
...
 
   if (typeof window !== 'undefined') {
-    wsUrl = /[\?&]OVERLAY_WS=([^&]+)/.exec(window.location.href);
-    if (wsUrl) {
+    wsUrl = urlParams.get('OVERLAY_WS');
+    if (wsUrl !== null) {
       const connectWs = function() {
-        ws = new WebSocket(wsUrl?.[1] as string);
+        if(wsUrl === null)
+          throw new TypeError('wsUrl is null, a string is required');
```
quisquous, line 203:
```suggestion
```

I don't think `wsUrl` can be null here.

quisquous (commented):
Thanks for the fix.  It looks like you'll need to run `npm run lintfix` to fix up spacing errors.

resources/overlay_plugin_api.ts
MaikoTan: I think it can be, when the user didn't run cactbot overlays in ACT/ngld's OverlayPlugin and/or didn't provide the parameter?
```suggestion
        if(wsUrl === null)
          throw new TypeError('Didn\'t found OVERLAY_WS parameter. If this overlay is not running in ACT or ngld\'s OverlayPlugin, make sure you have provided OVERLAY_WS parameter.');
```

quisquous: I agree with everything you said, but there's a `if (wsUrl) {` line above it that prevents the `connectWs` function from even being created if there if `wsUrl` is null.  This is also the only place that `wsUrl` is set.

MaikoTan: Ah, correct.
Sorry I misreaded it. ಠ︵ಠ

nyaoouo: by the way, this type checking in is only used to pass the typescript compile test but no any actual usage, because the function is defined in the global namespace so the compiler may think it is dangerous🤣

MaikoTan: Maybe should be change to something like:
```ts
    wsUrl = urlParams.get('OVERLAY_WS');
    if (wsUrl !== null) {
      const connectWs = function(wsUrl: string): void {
        ws = new WebSocket(wsUrl);

        // ...
        ws.addEventListener('close', () => {
          queue = null;

          console.log('Trying to reconnect...');
          // Don't spam the server with retries.
          window.setTimeout(() => {
            connectWs(wsUrl);
          }, 300);
        });
      };

      connectWs(wsUrl);
    } else {
      // ...
```

MaikoTan: Also remember to run `npm run lintfix`.

resources/overlay_plugin_api.ts
```diff
@@ -54,14 +54,15 @@ type VoidFunc<T> = (...args: T[]) => void;
...
-let wsUrl: RegExpExecArray | null = null;
+let wsUrl: string | null = null;
 let ws: WebSocket | null = null;
 let queue: (
   | { [s: string]: unknown }
   | [{ [s: string]: unknown }, ((value: string | null) => unknown) | undefined]
 )[] | null = [];
 let rseqCounter = 0;
 const responsePromises: Record<number, (value: unknown) => void> = {};
+const urlParams = new URLSearchParams(window.location.search);
```
quisquous, line 65: `window` may not be defined if this is running from node.  (See the errors in post process.)  What about moving the `const urlParams` declaration down to where it is used on line 199?

Also, would you mind moving the `wsUrl` declaration down to line 199 as well? It could be `const wsUrl`.

quisquous: There's no rush, but just wanted to make sure this comment wasn't missed.  Do you have time to fix this issue?  You should be able to reproduce the error by running `npm run test` locally.

nyaoouo: A commit is pushed

PR #3471, opened 2021-09-24
oopsy: add logDefinition refactoring

MaikoTan:
- change massive if-else to switch-case
- stop using hard-coded log type, use logDefinition's instead
- change target name extraction in `WasDefeated` to using logDefinition's field


quisquous:
Thanks!!

PR #3476, opened 2021-09-25
oopsy: add suzaku extreme

quisquous:


ui/oopsyraidsy/data/04-sb/trial/suzaku-ex.ts
```diff
@@ -0,0 +1,78 @@
...
+    'SuzakuEx Ashes To Ashes': '32D0', // Scarlet Lady add, raidwide explosion if not killed in time
+    'SuzakuEx Well Of Flame': '32E1', // targeted wide line aoe
+    'SuzakuEx Hotspot': '32E2', // platform fire when the runes are activated
+  },
+  damageFail: {
+    'SuzakuEx Immolate': '32E6', // Tower mechanic failure on Incadescent Interlude (party failure, not personal)
+    'SuzakuEx Phantom Flurry': '32DE', // "phantom half" final hit of Phantom Flurry
+  },
+  soloWarn: {
+    'SuzakuEx Scathing Net': '3243', // stack marker
```
JLGarber, line 34: This one *might* be worth handling as a trigger, since it can be handled with Diamondback on Blue Mage.

ui/oopsyraidsy/data/04-sb/trial/suzaku-ex.ts
```diff
@@ -0,0 +1,78 @@
...
+import ZoneId from '../../../../../resources/zone_id';
+import { OopsyData } from '../../../../../types/data';
+import { OopsyTriggerSet } from '../../../../../types/oopsy';
+import { playerDamageFields } from '../../../oopsy_common';
+
+// TODO: Rekindle (32E0) hitting multiple adds, any other players,
+// or hitting adds before they've been killed once is a mistake.
+// However, there is a point where it should hit one add in the
+// beginning. This also happens later in the fight too, where it
+// should not hit anybody else.
```
JLGarber, line 11: As a stop-gap, maybe set a phase-change trigger and at least warn of shared Rekindle markers during the second half? Or if we want to handle it really properly, check the EntityID for shares and see whether both are players?

ui/oopsyraidsy/data/04-sb/trial/suzaku-ex.ts
quisquous: Should I just remove it, then?

ui/oopsyraidsy/data/04-sb/trial/suzaku-ex.ts
quisquous: Unfortunately, I think it's really hard to see if "both" are shares, because if all 4 people hit themselves and a bird and 1 extra person gets hit by it, then they're all `0x16`, and how do you know which one of those 9 lines is the failure?

I'm kind of inclined to just leave this as an unhandled TODO.

JLGarber: Mmph, I see what you mean. It would end up being "check sources on each instance of damage, and for each source, at most one target can be a player". Not worth it. I do think it's worth handling post-intermission though?

ui/oopsyraidsy/data/04-sb/trial/suzaku-ex.ts
JLGarber: Oh, right, that's why you left it as a warning and not a failure? It's probably fine then. It just happened to be fresh in my head since I ran it with some buddies on Wednesday.

PR #3484, opened 2021-09-28
raidboss: add ifrit dialog translations

MaikoTan:


ui/raidboss/data/02-arr/trial/ifrit-nm.txt
```diff
@@ -26,6 +26,7 @@ hideall "--sync--"
 126.1 "Incinerate"
 
 # 70%
+# Ifrit:Succumb to the inferno!
 200.0 "--sync--" sync /00:0044:Ifrit:Succumb/ window 200,0
```
quisquous, line 30:
```suggestion
200.0 "--sync--" sync /00:0044:Ifrit:Succumb to the inferno/ window 200,0
```

What about changing the sync here to be the line itself rather than adding a comment (and then adjusting the translations as well)?

MaikoTan: Okay!

PR #3495, opened 2021-10-02
eureka: Fix flagRegex ignoring certain lines

KattTails:
The current implementation only allows lines that have text both before and after the map link to be shown on the map, match[1] and match[4] should be allowed to be empty.

Examples that currently don't show flags: 
```
00:000b:Player NameServer:Eureka Pagos ( 10.3  , 25.4 ) rez plz
00:000b:Player NameServer:pop Eureka Pagos ( 14.5  , 18.5 )
```

### DIFF
--- a/ui/eureka/eureka.ts
+++ b/ui/eureka/eureka.ts
@@ -764,8 +764,8 @@ class EurekaTracker {
         this.AddFlag(
           parseFloat(match[2]),
           parseFloat(match[3]),
-          match[1] as string,
-          match[4] as string,
+          match[1] ?? '',
+          match[4] ?? '',
         );
       }
 
--- a/ui/eureka/eureka.ts
+++ b/ui/eureka/eureka.ts
@@ -760,7 +760,7 @@ class EurekaTracker {
     for (const log of e.detail.logs) {
       const flagRegex = this.TransObjectByParserLang(this.options.Regex, 'gFlagRegex');
       let match = flagRegex.exec(log);
-      if (match && match[1] && match[2] && match[3] && match[4])
+      if (match && match[2] && match[3])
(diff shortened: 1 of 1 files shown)


KattTails:
I don't know typescript so I'm not really sure what I'm doing here. Sorry if it's bad :(

ui/eureka/eureka.ts
```diff
@@ -760,8 +760,14 @@ class EurekaTracker {
...
       const flagRegex = this.TransObjectByParserLang(this.options.Regex, 'gFlagRegex');
       let match = flagRegex.exec(log);
-      if (match && match[1] && match[2] && match[3] && match[4])
-        this.AddFlag(parseFloat(match[2]), parseFloat(match[3]), match[1], match[4]);
+      if (match && match[2] && match[3]) {
+        this.AddFlag(
+          parseFloat(match[2]),
+          parseFloat(match[3]),
+          match[1] as string,
+          match[4] as string,
```
MaikoTan, line 768: Mostly we don't use `as`-cast

```suggestion
          match[1] ?? '',
          match[4] ?? '',
```

If you are not familiar with TypeScript or JavaScript:
If the symbol value before `??` is `undefined` or `null` then use the symbol after it.
in other words, `A ?? B` is equal to `(A === undefined || A === null) ? B : A`

MaikoTan: Also, you can run `npm run lintfix` after you do the changes, ESLint would try its best to automatically fix lint errors.

KattTails: Thanks! I was initially just doing this through the github text editor 😅 but I did end up using --fix for the third attempt..

quisquous:
Thanks for the fix and the reviewing!

PR #3506, opened 2021-10-04
i18n: add ja misc translations

MaikoTan:


ui/raidboss/emulator/translations.ts
```diff
@@ -284,12 +339,14 @@ const emulatorLabels: Translation = {
...
     fr: 'Masquer les entrées sautées',
+    ja: 'スキップした項目を隠す',
     cn: '隐藏已跳过项目',
   },
   ' label[for=hideCollector]': {
     en: 'Hide Collectors',
     de: 'Verstecke "Sammel" Einträge',
     fr: 'Masquer les entrées "Collecté"',
+    ja: 'コレクターを隠す',
     cn: '隐藏收集器',
```
MaikoTan, line 350: @quisquous After talking with @ShadyWhite , we think we might translate wrong on this text, can you explain what is "collectors" means?

quisquous: Collectors are triggers that don't output anything.

MaikoTan: Oh, that's it. I'll open a new PR to fix it!

PR #3510, opened 2021-10-04
resource: change conditions.ts to named export

MaikoTan:
Same as https://github.com/quisquous/cactbot/pull/3465
Change the export style to named export

### DIFF
--- a/resources/conditions.ts
+++ b/resources/conditions.ts
@@ -4,23 +4,23 @@
 import { RaidbossData as Data } from '../types/data';
 import { TargetedMatches } from '../types/trigger';
 
-export default {
-  targetIsYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: TargetedMatches) => data.me === matches?.target;
-  },
-  targetIsNotYou(): (data: Data, matches: TargetedMatches) => boolean {
-    return (data: Data, matches: TargetedMatches) => data.me !== matches?.target;
-  },
-  caresAboutAOE(): (data: Data) => boolean {
-    return (data: Data) =>
-      data.role === 'tank' || data.role === 'healer' || data.CanAddle() || data.job === 'BLU';
-  },
-  caresAboutMagical(): (data: Data) => boolean {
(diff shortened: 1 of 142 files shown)


quisquous:
Positively, I really appreciate the move to named exports and your changes in this area.  I think in retrospect, I should have made everything be a named export when we switched to ESM modules.  Personally, I'd like to remove all of the default exports and any `import * from` and have everything be named imports and exports.

What I'm less sure about is making these free-floating functions.  I think it doesn't really matter that much for `Conditions` which is 90% `targetIsYou`, but following the pattern of `Util` and `Conditions` here, I don't want to `import { ability, abilityFull, addedCombatant, gainsEffect, losesEffect } from NetRegexes;` in a ton of trigger files.    For functions like this, I think there's some added readability value in saying `NetRegexes.ability()` vs `ability()`.  I think it's especially true given that `NetRegexes` and `Regexes` share function names.

(This makes me think that there's a bug here already for user triggers in this area.  Users who look at a trigger [like this](https://github.com/quisquous/cactbot/blob/triggers/02-arr/trial/levi-ex.js#L140) and then try to copy out `isRangedDpsJob` will not have their code work properly.  They need to say `Util.isRangedDpsJob`.)

I'm leaning towards saying that I'd prefer having a named export, but having that export be a class called `Conditions` with a bunch of static functions.  What do you think about that?

MaikoTan:
You're right, I didn't think that it cause problems like that, maybe we should revert #3465 as well?

MaikoTan:
> I do think the named export part here is fine. Like, if we had `import { Conditions } from ...` instead of `import Conditions from ...`, for what it's worth.
> 
> I think reverting #3465 or changing it to export a static `Util` class are both options here.

Maybe we can add an `index.ts` file in `resources/` that import all modules in `resources/` and export them, then you can just `import { Conditions } from '/path/to/resources';`?
And even `import { Conditions, NetRegexes, Utils } from '/path/to/resources';` which decrease lines that every trigger files have.

PR #3518, opened 2021-10-07
raidboss/oopsy: Baelsar's Wall

JLGarber:
The first boss's timeline is an approximation at best. Timings vary based on how fast enemies are killed and whether the user is solo or not. Given that the rotation is only three mechanics, and nothing is a surprise, it's *probably* fine.


JLGarber:
Eep, I can't believe I forgot the `Hideall` and `Reset` lines.

PR #3520, opened 2021-10-09
ci: Only Run Workflows On Push to Main

panicstevenson:
Stop running GitHub workflows twice for pushes to branches other than
main that also result in a pull request.


quisquous:
Thank you!

JLGarber:
This is great, thanks for the fix!

PR #3522, opened 2021-10-09
ci: add dependabot.yml to set commit prefix

quisquous:
This is an alternative to #3388.


quisquous:
I think this is correct.  Let's see if it works.

PR #3550, opened 2021-10-16
jobs: convert ui/jobs/jobs.js to typescript

quisquous:
This also removes the temporary Bars interface from bar.d.ts
and uses Bars directly from the jobs import.


quisquous:
Yeah, it happens at runtime, but typescript has no problems.  You should see it immediately if you run with this PR.

MaikoTan:
Looks like the code that defines custom element is not running before element creation, i.e: something like `window.customElements.define('timer-box', TimerBox);`

PR #3591, opened 2021-10-29
jobs: custom elements add create function

MaikoTan:


quisquous:
Does `npm run test` when you run it locally?

I think the issue is that raidboss imports `timerbar` and node doesn't know what to do with `HTMLElement`.

MaikoTan:
> Does `npm run test` when you run it locally?

Yes, `npm run test` locally reports the same error, either.
Then I found that it might be the code in `test/helper/test_timeline.js` that imports the `Timeline` class in `ui/raidboss/timeline.ts` which has a `TimelineUI` instance that imports `TimerBar`.
Maybe we should separate the timeline parsing code out of the `Timeline` class for testing code and utilities like `find_missing_timeline_translations.js` or something else.

quisquous:
> > Does `npm run test` when you run it locally?
> 
> Yes, `npm run test` locally reports the same error, too. Then I found that it might be the code in `test/helper/test_timeline.js` that imports the `Timeline` class in `ui/raidboss/timeline.ts` which has a `TimelineUI` instance that imports `TimerBar`. Maybe we should separate the timeline parsing code out of the `Timeline` class for testing code and utilities like `find_missing_timeline_translations.js` or something else.

Yeah, I think having a different PR to separate the timeline parsing from the timeline showing code would make this possible.

PR #3593, opened 2021-10-30
resources: revert resources/util.ts default export

MaikoTan:
But change to export an object expression
instead of a const global variable

### DIFF
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -1,81 +1,19 @@
-// TODO: Fix import/order
-/* eslint-disable import/order */
-import { isLang, Lang } from './languages';
 import { BaseOptions } from '../types/data';
 import { CactbotLoadUserRet, SavedConfig, SavedConfigEntry } from '../types/event';
 import { LocaleText } from '../types/trigger';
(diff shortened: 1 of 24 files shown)


resources/user_config.ts
```diff
@@ -1,81 +1,19 @@
...
-  jobEnumToJob,
-  jobToJobEnum,
-  jobToRole,
-  watchCombatant,
-} as const;
-import _ZoneId from './zone_id';
-const ZoneId = _ZoneId;
-import _ZoneInfo from './zone_info';
-const ZoneInfo = _ZoneInfo;
+import Conditions from './conditions';
```
quisquous, line 5: Is this a direct revert? I think this is incorrect.  This needs to be:
```
import _Conditions from './conditions';
const Conditions = _Conditions;
```

...as before, or it won't work for evals because of the way Webpack renames variables.

MaikoTan: Oops, sorry, that's part of another branch change.

MaikoTan: Reverted

PR #3594, opened 2021-10-30
i18n: update shinryu-ex cn

ShadyWhite:


quisquous:
Thank you for the quick translations!

PR #3595, opened 2021-10-30
raidboss: add triggers for Meng-Zi/Xun-Zi Deploy Armaments

jacob-keller:
During the Xun-Zi/Meng-Zi robot fight of the Tower at Paradigm's Rift,
the two robots use some line aoe attacks. These attacks, "Deploy
Armaments" can either be a single big line down the middle of the arena,
or two lines down the sides of the arena.

Add support to trigger an alert on where the safe area is for these
attacks. Use the heading of the skills to determine whether the middle
line attack should be avoided by going north/south or east/west.

TODO:
- [ ] Figure out how to handle near-overlapping messages (especially when they activate simultaneously

### DIFF
--- a/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
+++ b/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
@@ -6,21 +6,20 @@ import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
+// Object representing a "Deploy Armaments" attack.
+interface DeployArmaments {
+  sides: boolean;
+  vertical: boolean;
+  finishedTime: number;
+}
+
 export interface Data extends RaidbossData {
   busterTargets?: string[];
   cloneLunge?: boolean;
   seedTargets?: string[];
   seenSphere?: boolean;
   signalCount?: number;
-  deployArmaments?: number[];
-}
-
-// Enumeration of possible Deploy Armaments attacks
-enum DeployArmaments {
-  VerticalMiddle,
-  VerticalSides,
-  HorizontalMiddle,
-  HorizontalSides,
+  deployArmaments?: DeployArmaments[];
 }
 
 // TODO:
@@ -497,26 +496,26 @@ const triggerSet: TriggerSet<Data> = {
     //       Always appears simultaneously with the other bosses abilities
     //       and 2x 6079 casts.
     //
-    // To handle overlap, we have one trigger which captures all of the
-    // events into an array. Then, a delayed trigger processes all of the
-    // triggers and determines what to message players.
+    // Because these attacks overlap, we use one trigger to collect the
+    // active attacks, and a second trigger to display a message. Finally, a
+    // third trigger cleans up finished attacks after the ability is cast.
     {
       id: 'Paradigm Meng-Zi/Xun-Zi Deploy Armaments Collect',
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: ['5C00', '5C01', '5C03', '5C04'] }),
       run: (data, matches) => {
         data.deployArmaments ??= [];
-        if (matches.id === '5C00' || matches.id === '5C01') {
-          if (matches.heading === '1.570772')
-            data.deployArmaments.push(DeployArmaments.VerticalMiddle);
-          else if (matches.heading === '-4.792213E-05')
-            data.deployArmaments.push(DeployArmaments.HorizontalMiddle);
-        } else if (matches.id === '5C03' || matches.id === '5C04') {
-          if (matches.heading === '1.570772')
-            data.deployArmaments.push(DeployArmaments.VerticalSides);
-          else if (matches.heading === '-4.792213E-05')
-            data.deployArmaments.push(DeployArmaments.HorizontalSides);
-        }
+
+        // Convert the heading into 0=N, 1=E, 2=S, 3=W
+        const direction = Math.round(2 - 2 * parseFloat(matches.heading) / Math.PI) % 4;
+
+        const obj: DeployArmaments = {
+          sides: matches.id === '5C03' || matches.id === '5C04',
+          finishedTime: Date.parse(matches.timestamp) + parseFloat(matches.castTime) * 1000,
+          vertical: direction === 0 || direction === 2,
+        };
+
+        data.deployArmaments.push(obj);
       },
     },
     {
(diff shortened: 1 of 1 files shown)


jacob-keller:
I've tested this with the emulator and scanning a dozen logs of this fight. These are definitely the correct IDs. The major challenge is finding the best way to store up the data in order to handle when the two bosses overlap or nearly overlap their casts.

Ideally it would nice to delay messaging until we know whether you should go to the middle, corners, or N/S middle, or E/W middle. I have a pretty good way to calculate it, but I'm not familiar with how to handle multiple network log lines at once.

jacob-keller:
I marked this as WIP still because I don't think we can merge it until we find a way to handle over-lapping AOEs properly.

quisquous:
> I marked this as WIP still because I don't think we can merge it until we find a way to handle over-lapping AOEs properly.

Yeah, this was exactly why I didn't write this in the first place.   I know the first two are Xun-Zi on its own, and then it's both 3 seconds apart, both 3 seconds apart, both together.  I have never seen both alive after that and I haven't ever seen this phase last for too long so I don't know how it loops into the future. 

One possibility is to have one trigger (which catches starts using for all 4 possibilities) record which any and all possibilities it sees, and another trigger (which has the same regex) delay 0.3/0.5s and then call out anything that it has seen.  That way if there's only one, you can call them sequentially, and if there's two you can call them together?

jacob-keller:
> > I marked this as WIP still because I don't think we can merge it until we find a way to handle over-lapping AOEs properly.
> 
> Yeah, this was exactly why I didn't write this in the first place. I know the first two are Xun-Zi on its own, and then it's both 3 seconds apart, both 3 seconds apart, both together. I have never seen both alive after that and I haven't ever seen this phase last for too long so I don't know how it loops into the future.
> 
> One possibility is to have one trigger (which catches starts using for all 4 possibilities) record which any and all possibilities it sees, and another trigger (which has the same regex) delay 0.3/0.5s and then call out anything that it has seen. That way if there's only one, you can call them sequentially, and if there's two you can call them together?

I tried something like this but I wasn't able to get the delay working. It seemed like in the emulator at least that the "delay" just delayed when the message appeared and not the actual processing of the run event. I may have done it wrong though. I'll give it another shot later today.

My original idea was to basically keep track of "safe" areas using a bitmask and have each noticed attack clear out the bits that it damges, and then have the delayed trigger check this bitmask for the combinations of whats safe or not.

I think though, that if i just accumulate them and then do the bitmask calculation in the later function it might work! I'll give that a shot.

jacob-keller:
Ok I got a collector/trigger setup working. It might need some delays tweaked. I *think* there is enough time if we increase the delay to just over 3 seconds we could handle the staggered overlaps. The main question is whether those staggered triggers give enough time for players to make it to safety. For now I have it just collecting only for the true simultaneous captures.

It seems to work well enough in the emulator on logs I have. Thoughts?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -468,6 +477,144 @@ const triggerSet: TriggerSet<Data> = {
...
+          else if (matches.heading === '-4.792213E-05')
+            data.deployArmaments.push(DeployArmaments.HorizontalSides);
+        }
+      },
+    },
+    {
+      id: 'Paradigm Meng-Zi/Xun-Zi Deploy Armaments Trigger',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: ['5C00', '5C01', '5C03', '5C04'], capture: false }),
+      delaySeconds: 0.25,
```
jacob-keller, line 526: Specifically, we could do one of I think two things here:

1) just increase the delay to 3 or 3.05 or 3.1 seconds. This would collapse the 3-second staggered casts into one trigger, but at the cost of making it potentially not warn you about one section with enough time to move for the first AOE.

2) have it print immediately to move, but follow up with a "move here" warning for the 2nd AOE. To capture this, we could keep track of the time for each buff, and when we get a "using" skill, we could clear that attack from the list. This would allow printing something like "go middle!" followed by "Go N/S middle" or "Go E/W" by "Go Corners"

Basically the (2) idea is that we'd track casts until the AOE actually cleared. It's a bit more work, and I'm not sure how much benefit it has.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -468,6 +477,144 @@ const triggerSet: TriggerSet<Data> = {
...
+        },
+        nsMiddle: {
+          en: 'Go N/S Middle',
+        },
+        corners: {
+          en: 'Go Corners',
+        },
+        oops: {
+          en: 'Avoid line AOEs',
+        },
```
jacob-keller, line 615: For clarity, this is just here as a fallback in the weird case that we get something which we don't understand

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -12,6 +12,15 @@ export interface Data extends RaidbossData {
   seedTargets?: string[];
   seenSphere?: boolean;
   signalCount?: number;
+  deployArmaments?: number[];
+}
+
+// Enumeration of possible Deploy Armaments attacks
+enum DeployArmaments {
```
quisquous, line 19: Because users can look at transpiled triggers via the triggers branch, I think it'd be better to just use a JavaScript object here and not a TypeScript enum, if you don't mind.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
quisquous: I agree that (2) seems like the better approach if you wanted to do something like this.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -468,6 +477,144 @@ const triggerSet: TriggerSet<Data> = {
...
+    // events into an array. Then, a delayed trigger processes all of the
+    // triggers and determines what to message players.
+    {
+      id: 'Paradigm Meng-Zi/Xun-Zi Deploy Armaments Collect',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: ['5C00', '5C01', '5C03', '5C04'] }),
+      run: (data, matches) => {
+        data.deployArmaments ??= [];
+        if (matches.id === '5C00' || matches.id === '5C01') {
+          if (matches.heading === '1.570772')
```
quisquous, line 510: Are they always in one of two spots?  I usually try to avoid matching exact floating point number strings.  Maybe they're the same now, but maybe they'll be slightly different in the future?

In the past for headings, I've done stuff like [this](https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.ts#L255) to turn heading into a 0=N, 1=E, 2=S, 3=W integer.  I think that will be more robust in the future.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -468,6 +477,144 @@ const triggerSet: TriggerSet<Data> = {
...
+        else if (safeSpots === horizontalSides)
+          return output.nsSides!();
+        else if (safeSpots === nsMiddle)
+          return output.nsMiddle!();
+        else if (safeSpots === ewMiddle)
+          return output.ewMiddle!();
+        else if (safeSpots === corners)
+          return output.corners!();
+
+        return output.oops!();
```
quisquous, line 591: The bitmask is a nice generalization, but it feels a bit like overkill and it's a little hard to follow and make sure everything is correct.  What do you think about only using a series of if statements?  I think if there were more cases, a bitmask approach would cut down on combinatoric conditionals, but in this case there's 1-2 abilities, and many combinations are mutually exclusive, so I don't think that there's that many cases to cover manually.

```suggestion
        if (!data.deployArmaments)
          return output.oops!();

        if (data.deployArmaments.includes(DeployArmaments.VerticalMiddle)) {
          if (data.deployArmaments.includes(DeployArmaments.HorizontalMiddle))
            return output.corners!();
          // etc...
        }
        // etc...
```

Re: callout.  I also am not sure I would be able to know what "E/W Middle" vs "E/W Sides" meant on the fly.  I think "E/W Middle" is a subset of "E/W Sides" if I squint?  If these two mobs always appear in the same place, what about using cardinal directions, e.g. "go west", or "go west middle" (?), or "go northwest", or "go center".  (Assuming they're north and west.  It probably makes sense to go between them for melee.)

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
quisquous: Makes sense to me!

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: They were always in these two spots for every log I checked. But I like the more robust pattern.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: Do you have an example for this?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: I think I agree. that we can stick to a simpler approach where we only care about the 4 corners closest to the two bosses. That seems reasonable. It's hard to call out "move east and stay in the middle"

The two enemies always are in either the north end of the platform, or the west end, never anywhere else. The only two that are a bit weird are the north middle and west middle areas. I'll see if I can come up with something.

jacob-keller: In regards to simpler setup, ya I think we can mostly make it easier by just simplifying the overall attempt to care about only 4 corners.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
quisquous: Yeah, it's less about 2 vs 4 spots and more about not trusting a floating point string to be exactly that same string into the future.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: I ended up with "Go to North Boss" and "Go to West Boss" for those spots. That makes the most sense. When only a single line is detected, it is "Go North" or "Go West".

Will have the new version up in a bit.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -468,6 +476,140 @@ const triggerSet: TriggerSet<Data> = {
...
+        }
+        // other combinations shouldn't be possible
+        return output.oops!();
+      },
+      outputStrings: {
+        center: {
+          en: 'Go to Center',
+        },
+        northBoss: {
+          en: 'Go to North Boss',
```
quisquous, line 579: Go to north boss sounds good!  Thanks for wordsmithing this.  :)

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -468,6 +476,140 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: NetRegexes.ability({ id: ['5C00', '5C01', '5C03', '5C04'] }),
+      run: (data, matches) => {
+        if (!data.deployArmaments)
+          return;
+
+        // Get time of current cast
+        const now = Date.parse(matches.timestamp);
+
+        // Filter out any attacks that have completed
+        data.deployArmaments = data.deployArmaments.filter((e) => e.finishedTime > now);
```
quisquous, line 610: What do you think about moving this to the start of `Paradigm Meng-Zi/Xun-Zi Deploy Armaments Trigger`? 

Mostly my concern is, what if this trigger happens before the finishedTime and this doesn't get cleaned up? I think if you were filtering out armaments based on their ability id that'd be one thing, but doing it on time seems a little shaky.  Additionally (but minorly), if you did this at the start of one of the other two triggers above, you wouldn't need this third trigger.

quisquous (commented):
This looks great.  I have one more tiny small thought, and then this is good to go.  Thanks for all the improvements, I think this is great!  I really appreciate the follow-up to get to triggers that were skipped the first time around.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: Ah. Yea that makes sense. When processing a new message there's no reason to care about any that are done when this event is going off. There's no advantage to cleaning htem up earlier and this guarantees we clean them up just before they need to. It avoids any weird pitfalls with logging. Ok cool!

quisquous:
🎉

PR #3604, opened 2021-10-31
util: Typescriptify encounter_tools

JLGarber:
This is the first in a series of changes to our timeline generation/testing ecosystem that will (or at least we hope) unify those utilities with the rest of the project. Rather than using naively re-implemented Python parsing for log lines, we can just drop the existing Typescript definitions in and use those to implement `make/test_timeline`.

I'm still not completely sure what I'm doing with Typescript here, but I did at least manage to clear the lint errors. I was mostly trying to convert the existing `encounter_tools.js` prototype for this particular PR. This isn't in any way intended to be final or polished, simply at a point where I've changed what I can and now need some input for what might need further changes and additions before an actual release.

I'm a little annoyed that we have to import `ZoneInfo` for just one check, but I'm not sure there's a better way to ensure "yes, we are in a combat instance".

### DIFF
--- a/util/logtools/encounter_tools.ts
+++ b/util/logtools/encounter_tools.ts
@@ -90,7 +90,7 @@ export class EncounterFinder {
 
   skipZone(): boolean {
     // We don't want combat from the overworld or from solo instanced duties.
-  // However, if we can't find zone info, we explicitly don't skip it.
+    // However, if we can't find zone info, we explicitly don't skip it.
     if (!this.currentZone.zoneId || !this.zoneInfo)
       return false;
     const content = this.zoneInfo.contentType;
--- a/util/logtools/encounter_tools.ts
+++ b/util/logtools/encounter_tools.ts
@@ -1,25 +1,41 @@
-import NetRegexes from '../../resources/netregexes';
-import ZoneInfo from '../../resources/zone_info';
 import ContentType from '../../resources/content_type';
-
+import NetRegexes from '../../resources/netregexes';
+import { data as ZoneInfo, ZoneInfoType } from '../../resources/zone_info';
+import { LocaleText } from '../../types/trigger';
 import { commonReplacement, syncKeys } from '../../ui/raidboss/common_replacement';
+
 
 // TODO: add some error checking that a zone has been found before a fight.
 // This can happen on partial logs.
 
+
 export class EncounterFinder {
+  currentZone: LocaleText | null;
+  currentFight: LocaleText | null;
+  currentSeal: boolean;
+  zoneInfo: ZoneInfoType | null;
+  zoneId: number | null;
+
+  haveWon: boolean;
+  haveSeenSeals: boolean;
+
+  regex: Record<string, NetRegexes>;
+
+  sealRegexes: Array<NetRegexes>;
+  unsealRegexes: Array<NetRegexes>;
   constructor() {
     this.currentZone = null;
     this.currentFight = null;
-    this.currentSeal = null;
+    this.currentSeal = false;
     // May be null for some zones.
     this.zoneInfo = null;
+    this.zoneId = null;
 
     this.haveWon = false;
     this.haveSeenSeals = false;
 
     this.regex = {
-      changeZone: NetRegexes.changeZone(),
+      changeZone: NetRegexes.changeZone({ id: '*' }),
       cactbotWipe: NetRegexes.echo({ line: 'cactbot wipe.*?' }),
       win: NetRegexes.network6d({ command: '40000003' }),
       wipe: NetRegexes.network6d({ command: '40000010' }),
@@ -34,28 +50,35 @@ export class EncounterFinder {
     this.unsealRegexes = [];
 
     const sealReplace = commonReplacement.replaceSync[syncKeys.seal];
-    for (const lang in sealReplace) {
-      const line = sealReplace[lang].replace('$1', '(?<seal>.*?)') + '.*?';
+    if (!sealReplace)
+      throw new Error('missing seal regex');
+    for (const regexString of Object.values(sealReplace)) {
+      const line = `${regexString?.replace('$1', '(?<seal>.*?)')}.*?`;
       this.sealRegexes.push(NetRegexes.message({ line: line }));
     }
 
     const unsealReplace = commonReplacement.replaceSync[syncKeys.unseal];
-    for (const lang in unsealReplace) {
-      const line = '.*?' + unsealReplace[lang] + '.*?';
+    if (!unsealReplace)
+      throw new Error('missing unseal regex');
+    for (const lang of Object.values(unsealReplace)) {
+      const line = `*.?${lang}.*?`;
       this.unsealRegexes.push(NetRegexes.message({ line: line }));
     }
   }
 
-  skipZone() {
-    if (!this.zoneInfo)
+  skipZone(): boolean {
+    if (this.zoneId === null || this.zoneInfo === null)
       return false;
-    const contentType = this.zoneInfo.contentType;
-    if (!contentType)
+    const info = this.zoneInfo;
+    if ((info[this.zoneId]?.contentType) === null)
+      return false;
+    const content = info?.[this.zoneId]?.contentType;
+    if (!content)
       return false;
 
     // There are some seal messages in older raids, but not consistently.
     // Therefore, we can't require them.
-    const keepTypes = [
+    const keepTypes: Array<number> = [
       ContentType.Dungeons,
       ContentType.Eureka,
       ContentType.Raids,
@@ -64,13 +87,14 @@ export class EncounterFinder {
       ContentType.DeepDungeons,
     ];
 
-    return !keepTypes.includes(contentType);
+    return !keepTypes.includes(content);
   }
 
-  process(line) {
-    let m;
+  process(line: string): void {
+    let m: RegExpMatchArray | null;
 
     m = line.match(this.regex.changeZone);
+    // m = NetRegexes.changeZone(line);
     if (m) {
       if (this.currentZone === m.groups.name) {
         // Zoning into the same zone, possibly a d/c situation.
@@ -104,7 +128,7 @@ export class EncounterFinder {
     if (this.skipZone())
       return;
 
-    m = line.match(this.regex.cactbotWipe);
+    m = new RegExp(this.regex.cactbotWipe);
     if (m) {
       if (this.currentFight !== null) {
         this.onEndFight(line, this.currentFight, { ...m.groups, endType: 'CactbotWipe' });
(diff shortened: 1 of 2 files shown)


util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+  startTime?: Date;
+  endTime?: Date;
+  endType?: string;
+  sealName?: string;
+};
+
+export class EncounterFinder {
+  currentZone: ZoneEncInfo;
+  currentFight: FightEncInfo;
+  currentSeal: string | null;
```
quisquous, line 36: Maybe prefer `undefined` to `null` here unless there's a good reason? (The cases where it makes sense to me are examples like `document.createElement` that can return `null`.)

```suggestion
  currentSeal?: string;
```

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+  zoneInfo: {
+    readonly exVersion: number;
+    readonly contentType?: number;
+    readonly name: LocaleText;
+    readonly offsetX: number;
+    readonly offsetY: number;
+    readonly sizeFactor: number;
+    readonly weatherRate: number;
+  } | null;
```
quisquous, line 45: What about avoiding type duplication (and null vs undefined) and saying:

```suggestion
  zoneInfo?: typeof ZoneInfo[number];
```

Alternatively, export `ZoneInfoType` from `ZoneInfo` (which would require a script change)` and import that directly?

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+  constructor() {
+    this.currentZone = {};
+    this.currentFight = {};
+    this.currentSeal = null;
+    // May be null for some zones.
+    this.zoneInfo = null;
+
+    this.haveWon = false;
+    this.haveSeenSeals = false;
```
quisquous, line 77: Could you move these to be done in the member declaration, e.g. `currentZone: ZoneEncInfo = {};`

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+
+    this.regex = {
+      changeZone: NetRegexes.changeZone({ id: '*' }),
```
quisquous, line 80: I think `'*'` is incorrect?

```suggestion
      changeZone: NetRegexes.changeZone(),
```

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+      cactbotWipe: NetRegexes.echo({ line: 'cactbot wipe.*?' }),
+      win: NetRegexes.network6d({ command: '40000003' }),
+      wipe: NetRegexes.network6d({ command: '40000010' }),
+      commence: NetRegexes.network6d({ command: '4000000[16]' }),
+      playerAttackingMob: NetRegexes.ability({ sourceId: '1.{7}', targetId: '4.{7}' }),
+      mobAttackingPlayer: NetRegexes.ability({ sourceId: '4.{7}', targetId: '1.{7}' }),
+    };
+
+    // TODO: consider also doing this with lang? But lang only deals with Regexes???
+    // Sorry, this is quite a bit of a hack.
```
quisquous, line 90: Feel free to not make this change here (it's not really a conversion, but an improvement), but I think this whole section could be simplified by using `LocaleNetRegex`.

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+    if (!unsealReplace)
+      throw new Error('missing unseal regex');
+    for (const lang of Object.values(unsealReplace)) {
+      const line = `*.?${lang}.*?`;
+      this.unsealRegexes.push(NetRegexes.message({ line: line }));
+    }
+  }
+
+  skipZone(): boolean {
+    if (!this?.currentZone?.zoneId || this.zoneInfo === null)
```
quisquous, line 112: No need for `this?` here and elsewhere.

```suggestion
    if (!this.currentZone?.zoneId || this.zoneInfo === null)
```

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+      return false;
+    const info = this.zoneInfo;
```
quisquous, line 114: If you move this statement to be the first line of the function, and then check `info === null` (or undefined) vs `this.zoneInfo === null)`, then the rest of the code in the function will know that `info` is defined.

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+      ContentType.UltimateRaids,
+      ContentType.DeepDungeons,
+    ];
+
+    return !keepTypes.includes(content);
+  }
+
+  process(line: string): void {
+    const cZ = this.regex.changeZone.exec(line);
+    if (cZ?.groups) {
```
quisquous, line 137: This is a take it or leave it suggestion, but I've found it useful elsewhere to do this sort of pattern:

```typescript
    const cZ = this.regex.changeZone.exec(line)?.groups;
    if (cZ) {
```

...and then everywhere in the block, I know that `cZ` is defined and don't have to `?.` it everywhere or repeat `.groups`.

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+      // If we changed zones, we have definitely stopped fighting.
+      // Therefore we can safely initialize everything.
+      if (this.currentFight.startTime)
+        this.initializeFight();
+      if (this.currentZone.startTime)
+        this.initializeZone();
+
+      this.haveWon = false;
+      this.haveSeenSeals = false;
+      this.zoneInfo = ZoneInfo[parseInt(cZ.groups.id, 16)] || null;
```
quisquous, line 153:
```suggestion
      this.zoneInfo = ZoneInfo[parseInt(cZ.groups.id, 16)] ?? null;
```

... or alternatively just `this.zoneInfo = ZoneInfo[parseInt(cZ.groups.id, 16)` if you make it `undefined` instead of `null`.

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+    return new Date(Date.parse(matches.timestamp));
+  }
+}
+
+export class EncounterCollector extends EncounterFinder {
+  zones: Array<ZoneEncInfo>;
+  fights: Array<FightEncInfo>;
+  lastZone: ZoneEncInfo;
+  lastFight: FightEncInfo;
+  lastSeal: string | null;
```
quisquous, line 255: Same thing with `undefined` vs `null` here, as well as initializing members in their declaration.

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+
+  onEndZone(line: string, matches: NetMatches['ChangeZone']): void {
+    this.lastZone.endLine = line;
+    this.lastZone.endTime = this.dateFromMatches(matches);
+    this.zones.push(this.lastZone);
+    this.lastZone = {};
+  }
+
+  override onStartFight(line: string, name: string, matches: NetMatches['Ability' | 'GameLog']): void {
+    const id = this.lastZone.zoneId || 0;
```
quisquous, line 283:
```suggestion
    const id = this.lastZone.zoneId ?? 0;
```

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,313 @@
...
+  }
+
+  onEndZone(line: string, matches: NetMatches['ChangeZone']): void {
+    this.lastZone.endLine = line;
+    this.lastZone.endTime = this.dateFromMatches(matches);
+    this.zones.push(this.lastZone);
+    this.lastZone = {};
+  }
+
+  override onStartFight(line: string, name: string, matches: NetMatches['Ability' | 'GameLog']): void {
```
quisquous, line 282: I ... didn't know you could do `NetMatches['Ability' | 'GameLog']` vs `NetMatches['Ability'] | NetMatches['GameLog']`.  Neat!

quisquous (commented):
Thanks for this!  Did you test this on splitting any logs?

It's definitely a bit hard to read the diff.  One thing I've found useful for reviewing is to have one commit where you just do a rename, and then put other changes on top of that.  Without that, git can think that you're removing one file and adding an unrelated file and so the diff doesn't exist.  I'm not saying it makes any sense to go back and do that now, but just as a thought for the future.

util/logtools/encounter_tools.ts
JLGarber: That would definitely be better, but I think you're right that that's best left for another time.

util/logtools/encounter_tools.ts
JLGarber: It occurs to me that there's no real reason to declare a helper variable here. `this.zoneInfo` is just as easily human-readable.

util/logtools/encounter_tools.ts
JLGarber: I don't know it myself, it's just a guess, and it satisfies the linter.

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,283 @@
...
+    for (const lang of Object.values(unsealReplace)) {
+      const line = `*.?${lang}.*?`;
+      this.unsealRegexes.push(NetRegexes.message({ line: line }));
+    }
+  }
+
+  skipZone(): boolean {
+    if (!this.currentZone.zoneId || !this.zoneInfo)
+      return false;
+    if (!(this.zoneInfo.contentType))
```
quisquous, line 91: tiny nit: unnecessary parens

```suggestion
    if (!this.zoneInfo.contentType)
```

quisquous (commented):
Looks good other than one tiny thing.

JLGarber:
I did run a basic test with `split_log`. It dumped a whole bunch of `41` lines that it correctly didn't know how to handle, but the split function *seemed* okay. I'll definitely want to consider how we might move some of that functionality over to `encounter_tools`, since it really shouldn't live directly inside a specific utility!

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,283 @@
...
+    commence: CactbotBaseRegExp<'ActorControl'>;
+    playerAttackingMob: CactbotBaseRegExp<'Ability'>;
+    mobAttackingPlayer: CactbotBaseRegExp<'Ability'>;
+  };
+
+  sealRegexes: Array<CactbotBaseRegExp<'GameLog'>> = [];
+  unsealRegexes: Array<CactbotBaseRegExp<'GameLog'>> = [];
+
+  initializeZone(): void {
+    this.currentZone = {};
```
JLGarber, line 55: It occurs to me, maybe we should re-initialize `this.zoneInfo` here as well? In theory, `skipZone` should be called only in places where `zoneInfo` is correctly populated, but would it be bad to be explicit about it?

util/logtools/encounter_tools.ts
JLGarber: Can't be completely sure, but I didn't see any errors arising from this while testing with `split_log.js`, so I think we can assume it works for now.

util/logtools/encounter_tools.ts
quisquous: Yeah, I think that makes sense.  Good idea!

util/logtools/encounter_tools.ts
```diff
@@ -0,0 +1,290 @@
...
+      throw new Error('missing unseal regex');
+    for (const lang of Object.values(unsealReplace)) {
+      const line = `*.?${lang}.*?`;
+      this.unsealRegexes.push(NetRegexes.message({ line: line }));
+    }
+  }
+
+  skipZone(): boolean {
+    // We don't want combat from the overworld or from solo instanced duties.
+  // However, if we can't find zone info, we explicitly don't skip it.
```
quisquous, line 93: I'm not sure why dprint didn't complain about this comment spacing but

```suggestion
    // However, if we can't find zone info, we explicitly don't skip it.
```

PR #3605, opened 2021-10-31
raidboss: add trigger for Bloody Sweep

jacob-keller:
Hansel and Gretel use "Tandem Assault: Bloody Sweep" as a combo move
which will either sweep the outsides and mainly the south half of the
arena, or sweep the insides hitting mainly the north half of the arena.
Players need to move south or north depending on the position of Hansel
and Gretel. Sometimes, the two bosses will use "Transference" and switch
places.

Add a trigger which detects the relevant IDs and issues an alert on
whether ot move south or north. We can be clever by using a few facts:

* Hansel is left handed, and Gretel is right handed
* The ID indicates which position Hansel or Gretel will be at start
* The cast time can determine whether or not they switch places

Using this information we can use a single strigger which determines the
side of the arena which is safe.

### DIFF
--- a/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
+++ b/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
@@ -427,12 +427,8 @@ const triggerSet: TriggerSet<Data> = {
         return output.north!();
       },
       outputStrings: {
-        north: {
-          en: 'Go North',
-        },
-        south: {
-          en: 'Go South',
-        },
+        north: Outputs.north,
+        south: Outputs.south,
       },
     },
     {
--- a/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
+++ b/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
@@ -24,7 +24,6 @@ export interface Data extends RaidbossData {
 
 // TODO:
 //   Update Knave knockback directions to instead use cardinals
-//   Hansel and Gretel Bloody Sweep
 //   Hansel and Gretel Stronger Together Tethered
 //   Hansel & Gretel Passing Lance
 //   Hansel & Gretel Breakthrough
@@ -401,6 +400,40 @@ const triggerSet: TriggerSet<Data> = {
       netRegexCn: NetRegexes.startsUsing({ id: '5C7[34]', source: ['韩塞尔', '格雷特'], capture: false }),
       netRegexKo: NetRegexes.startsUsing({ id: '5C7[34]', source: ['헨젤', '그레텔'], capture: false }),
       response: Responses.aoe(),
+    },
+    {
+      id: 'Paradigm Hansel/Gretel Bloody Sweep',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '5C5[4567]', source: ['Hansel', 'Gretel'] }),
+      delaySeconds: (_data, matches) => parseFloat(matches.castTime) - 5,
+      durationSeconds: 5,
+      suppressSeconds: 1,
+      alertText: (_data, matches, output) => {
+        if (matches.id === '5C54' || matches.id === '5C55') {
+          // Hansel is West and Gretel is East
+          if (parseFloat(matches.castTime) > 12) {
+            // Hansel and Gretel will switch places
+            return output.north!();
+          }
+          // Hansel and Gretel stay in same position
+          return output.south!();
+        }
+        // Gretel is West and Hansel is East
+        if (parseFloat(matches.castTime) > 12) {
+          // Hansel and Gretel will switch places
+          return output.south!();
+        }
+        // Hansel and Gretel stay in same position
+        return output.north!();
+      },
+      outputStrings: {
+        north: {
+          en: 'Go North',
+        },
+        south: {
+          en: 'Go South',
+        },
+      },
     },
     {
       id: 'Paradigm Red Girl Cruelty',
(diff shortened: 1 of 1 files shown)


jacob-keller:
I've looked through the network logs i have and this seems to be the pattern. I've confirmed it against a video capture along with the live video, though I don't have live video for every combination.

We could get overly complicated with combining multiple triggers (i.e. check for the Tandem network ability that indicates the red/blue DNA lines, or look at heading/position of the AOE markers). However, I think this simple check works in practice. I chose 12 as the cast time cutoff since the full cast when they swap is over 12 seconds and the cast for when they don't swap is ~8 seconds.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -402,6 +401,40 @@ const triggerSet: TriggerSet<Data> = {
...
+        if (parseFloat(matches.castTime) > 12) {
+          // Hansel and Gretel will switch places
+          return output.south!();
+        }
+        // Hansel and Gretel stay in same position
+        return output.north!();
+      },
+      outputStrings: {
+        north: {
+          en: 'Go North',
```
quisquous, line 431: Do you think this needs to be "Go North" vs "North"? You could say `north: Outputs.north` here, or we could add a `goNorth` to `Outputs` as well?

quisquous (commented):
Ooooooh nice find about the ids and the cast times!!

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -402,6 +401,40 @@ const triggerSet: TriggerSet<Data> = {
...
+      id: 'Paradigm Hansel/Gretel Bloody Sweep',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '5C5[4567]', source: ['Hansel', 'Gretel'] }),
+      delaySeconds: (_data, matches) => parseFloat(matches.castTime) - 5,
+      durationSeconds: 5,
+      suppressSeconds: 1,
+      alertText: (_data, matches, output) => {
+        if (matches.id === '5C54' || matches.id === '5C55') {
+          // Hansel is West and Gretel is East
+          if (parseFloat(matches.castTime) > 12) {
```
quisquous, line 414: Also, in your comment you said the cast times were 12 and 8? I assume you mean something bigger than 12? Or, how does this work?

jacob-keller: yea, I can double check the values here real quick. Maybe its better to just use the average between the two.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: I think just saying "north" is fine.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: it.s 7.7 seconds and 12.7 seconds. I can use half way of 10 instead for clarity

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: Alternatively, we just used "Go North" in the meng-zi/xun-zi fight so maybe that is a good idea too

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -402,6 +401,36 @@ const triggerSet: TriggerSet<Data> = {
       netRegexKo: NetRegexes.startsUsing({ id: '5C7[34]', source: ['헨젤', '그레텔'], capture: false }),
       response: Responses.aoe(),
     },
+    {
+      id: 'Paradigm Hansel/Gretel Bloody Sweep',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '5C5[4567]', source: ['Hansel', 'Gretel'] }),
+      delaySeconds: (_data, matches) => parseFloat(matches.castTime) - 5,
```
quisquous, line 408: Sorry, I didn't see this before, but is there a reason to delay this?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
```diff
@@ -402,6 +401,36 @@ const triggerSet: TriggerSet<Data> = {
...
+      durationSeconds: 5,
+      suppressSeconds: 1,
+      alertText: (_data, matches, output) => {
+        if (matches.id === '5C54' || matches.id === '5C55') {
```
quisquous, line 412: Would you mind also leaving a comment for a future reader about the two times it can be, so `10` can be explained?

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: Not particularly other than the alert starting super early relative to the actual ability cast (8 seconds or 12 seconds) which means the alert will have disappeared a long time before. But.. I guess that's only more important if you need to do something precisely on time... now that I think about it just moving ahead of time is fine, so I'll go ahead and remove this.

ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
jacob-keller: Will fix.

quisquous: Thanks, this is great.  It's in the commit message, but having a comment makes this kind of hard-won detail more obvious.  :)

PR #3607, opened 2021-11-01
util: fix some type error

MaikoTan:


types/net_matches.d.ts
```diff
@@ -1,7 +1,7 @@
 import { NetFields } from './net_fields';
 import { TriggerTypes } from './net_trigger';
 
-export type Matches = { [s in T]?: string };
+export type Matches = { [s: string]: string };
```
MaikoTan, line 4: I am not very sure if this is the right case, but apparently, the type `T` in here is not defined, so cause the error.
@valarnin can you do a review?

types/net_trigger.d.ts
```diff
@@ -1,12 +1,12 @@
...
   groups?: NetMatches[T];
-}
+};
 
-export class CactbotBaseRegExp<T extends TriggerTypes> extends RegExp {
-  exec(string: string): CactbotRegExpExecArray<T> | null;
-}
+export type CactbotBaseRegExp<T extends TriggerTypes> = RegExp & {
+  exec: (string: string) => CactbotRegExpExecArray<T> | null;
+};
```
MaikoTan, line 10: It is weird that here define classes and `extends` an interface (theoretically an interface should only be `implement`-ed when you using it in a class)
so I change to a `type` expression and do a union operation.

quisquous:
Just for my own understanding, what is the type error that you're fixing?

MaikoTan:
> Just for my own understanding, what is the type error that you're fixing?

> `types/data.d.ts(50,32)`: error TS2304: Cannot find name 'NetMatches'.
Didn't import `NetMatches`.

> `types/net_matches.d.ts(4,31)`: error TS2304: Cannot find name 'T'.
Type `T` is not defined.

> `types/net_trigger.d.ts(4,69)`: error TS2689: Cannot extend an interface 'RegExpExecArray'. Did you mean 'implements'?
A class cannot `extends` an interface.

types/net_matches.d.ts
valarnin: I believe that this was originally:

```ts
export type Matches<T> = { [s in T]?: string };
```

I can't recall why it's not defined like this now.

MaikoTan: But then you should change all `Matches` to `Matches<T>` everywhere.
Maybe `export type Matches<T = string> = { [s in T]?: string };` is the intended?

valarnin: Right, it's coming back to me a bit now. `Matches` was originally an extract of field properties but was changed to be the generic any valid object handler. So the change in this PR is fine/accurate here.

valarnin: Actually, slight adjustment. This should instead be as follows, to match the original intention (e.g. properties might not be defined):
```ts
export type Matches = { [key: string]: string | undefined };
```

MaikoTan: Okay, I'll do that.

types/net_trigger.d.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export class CactbotRegExpExecArray<T extends TriggerTypes> extends RegExpExecArray {
+export interface CactbotRegExpExecArray<T extends TriggerTypes> extends RegExpExecArray {
   groups?: NetMatches[T];
 }
 
-export class CactbotBaseRegExp<T extends TriggerTypes> extends RegExp {
-  exec(string: string): CactbotRegExpExecArray<T> | null;
+export interface CactbotBaseRegExp<T extends TriggerTypes> extends RegExp {
+  exec: (string: string) => CactbotRegExpExecArray<T> | null;
```
valarnin, line 9: This looks correct in general, however to account for the `undefined` change in net_matches.d.ts, this adjustment is needed:
```ts
  groups?: T extends 'None' ? never : NetMatches[T];
```

MaikoTan: Sure

MaikoTan:
But I still don't know why are these errors not reported in cactbot itself, I found these errors when I am [trying to import cactbot as a submodule for my extension cactbot-highlight](https://github.com/MaikoTan/cactbot-highlight/pull/70).
Although I fixed it by [using babel instead of tsc to transpile typescript files](https://github.com/MaikoTan/cactbot-highlight/pull/71/commits/94c4eb77d36a971e7a405375077b55322a0b7e45), I am very very curious about the reason. (/ω＼*)

quisquous:
Thanks for all the review and fixing this all up.

> But I still don't know why are these errors not reported in cactbot itself

Yeah, that was really my question above.  It wasn't clear to me why this wasn't causing issues eslewhere.

valarnin:
To clarify re: why this is erroring in this specific case but not on compile or anything in the actual cactbot project. In `tsconfig.base.json`:

```json
    "skipLibCheck": true,
```

See also: https://stackoverflow.com/questions/64042204/typescript-is-not-detecting-errors-for-nonexistent-type-names-in-declaration-d

PR #3613, opened 2021-11-03
util: fix type errors

MaikoTan:
Mentioned in https://github.com/quisquous/cactbot/pull/3607#issuecomment-957517587 by @valarnin  

There are still errors caused by `fork-ts-checker-webpack-plugin`, currently typing of this plugin relies on webpack@v4, but we are now using webpack@v5 which migrated to typescript and changed exported types names that is not compatiable with v4.
Although `fork-ts-checker-webpack-plugin` fixed this by dropping webpack v4 support in 7.0.0, but its v7 version is still in alpha.


types/manifest.d.ts
```diff
@@ -1,14 +1,14 @@
...
+    [filename: string]: import('./trigger').LooseTriggerSet | string;
   }
   const raidbossFileData: RaidbossFileData;
   export default raidbossFileData;
 }
 
 declare module '*/oopsy_manifest.txt' {
   export interface OopsyFileData {
-    [filename: string]: LooseOopsyTriggerSet;
+    [filename: string]: import('./oopsy').LooseOopsyTriggerSet;
```
quisquous, line 11: Thank you for this syntax.  I wrestled with getting this import right in the past but could never figure it out.

.vscode/settings.json
```diff
@@ -28,5 +28,5 @@
       */
     "editor.formatOnSaveMode": "file"
   },
-  "typescript.tsdk": "node_modules\\typescript\\lib"
+  "typescript.tsdk": "node_modules/typescript/lib"
```
quisquous, line 31: Is this unrelated to `skipLibCheck`?

ui/raidboss/timeline.ts
```diff
@@ -1086,14 +1086,14 @@ export class TimelineController {
   constructor(
     protected options: RaidbossOptions,
     protected ui: TimelineUI,
-    raidbossDataFiles: { [filename: string]: string },
+    raidbossDataFiles: { [filename: string]: string | LooseTriggerSet },
```
quisquous, line 1089: Can this be `raidbossDataFiles: { [filename: string]: RaidbossFileData },` to avoid duplicating the definition of `RaidbossFileData`? (Maybe that means `trigger.d.ts` has to export the `RaidbossFileData` type?)

quisquous (commented):
I'm not usually keen on alpha versions of anything, but I think in this case if there's not a lot of changes that come along with using it, then I think it's fine.  If it breaks in the future, we can either turn back on `skipLibCheck` temporarily and revert back to a non-alpha version or we can stop using `fork-ts-checker-webpack-plugin` if we had to.

MaikoTan:
> I'm not usually keen on alpha versions of anything, but I think in this case if there's not a lot of changes that come along with using it, then I think it's fine. If it breaks in the future, we can either turn back on `skipLibCheck` temporarily and revert back to a non-alpha version or we can stop using `fork-ts-checker-webpack-plugin` if we had to.

As we have typescript checking in lint workflow already, I think it is safe to just remove this plugin?

types/manifest.d.ts
MaikoTan: I was just trying, as I know `import()` can be used in JSDoc comment that typescript knows. Glad to see it works in code as well. 😁

.vscode/settings.json
MaikoTan: Oops, that might be changed when I changed the typescript version in VSCode.
I always develop cactbot on a Linux machine, the backslash `\` is not compatible with POSIX paths; but vice versa, to use slash `/` on a windows machine didn't cause problems. So I think we can keep this setting as is?

ui/raidboss/timeline.ts
MaikoTan: True, I think to export `RaidbossFileData` type is better.

quisquous:
> As we have typescript checking in lint workflow already, I think it is safe to just remove this plugin?

I mean true, but I think these are different steps.  One is webpack that is used for both the live dev server and for build time, and the other is just lint workflow for submitting.  I agree that the webpack build and the lint workflow are the same, but I (personally?) think it is useful to know when there is a TypeScript error when writing things live and having it reload.  I think I would lose a lot of time to silent TypeScript errors if they weren't printed out as a part of the dev server.

MaikoTan:
I add back fork-ts-checker plugin and the skipLibCheck flag, I don't know what is a better solution.
Sorry for waiting so long.
Besides, I cannot run `npm run test` on my machine, which reports ERR_MODULE_NOT_FOUND just like https://github.com/quisquous/cactbot/issues/3616 says.

quisquous:
> I add back fork-ts-checker plugin and the skipLibCheck flag, I don't know what is a better solution. Sorry for waiting so long.

Ok, how about I merge this and we can go from here.  I think one option is to use the plugin alpha? Other options include seeing if we can add some type assertion / override for the type of the plugin so that it's only that one object that has no type but everything else is checked?

quisquous:
Sorry, I reverted this because it broke `npm run process-triggers`.  I'm not sure what in particular broke, and it's weird that it passed the test post-process step, because that runs `npm run process-triggers`!

MaikoTan:
Found
```log
ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts:494:13 - error TS2367: This condition will always return 'false' since the types 'string' and 'number | undefined' have no overlap.

494         if (matches[1] === data.puddle)
                ~~~~~~~~~~~~~~~~~~~~~~~~~~

ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts:498:13 - error TS2367: This condition will always return 'false' since the types 'string' and 'number | undefined' have no overlap.

498         if (matches[1] === data.puddle)
                ~~~~~~~~~~~~~~~~~~~~~~~~~~

ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts:503:13 - error TS2367: This condition will always return 'false' since the types 'string' and 'number | undefined' have no overlap.

503         if (matches[1] === data.puddle)
                ~~~~~~~~~~~~~~~~~~~~~~~~~~


Found 3 errors.
```

PR #3614, opened 2021-11-04
raidboss: add basic triggers for Angra Mainyu

jacob-keller:
Angra Mainyu is the first boss of The World of Darkness alliance raid.
This boss has a few instant-death mechanics, including some basic
mechanics that are very easy to add triggers for.

Add a trigger for the "Mortal Gaze" mechanic. When this is cast, players
should look away. If they fail to look away, they acquire the Doom
debuff and will die in 10 seconds unless they reach a cleansing circle.
Add an alert for the look away, and an alarm for when doom is acquired.

Add a trigger for the "Double Vision" attack. This attack is cast
periodically, and damages players based on their debuff. The whole arena
lights up half red and half white. If the player is hit by the red area
they gain Ireful, and if they are hit by the white area they gain
Sullen. Repeated hits of the same color do increasing damage until it is
fatal. Keep track of the current debuff and warn the player of which
color to stand in based on this.

In the future, it would be nice to add some alerts for Level 100 Flare,
Level 150 Doom, and possibly the roulette mechanic. These are more
complex and have been skipped for now.

### DIFF
--- a/ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
+++ b/ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
@@ -64,7 +64,6 @@ const triggerSet: TriggerSet<Data> = {
           // Stand in white half to switch to Sullen
           return output.white!();
         }
-        return output.vision!();
       },
       outputStrings: {
         red: {
@@ -72,9 +71,6 @@ const triggerSet: TriggerSet<Data> = {
         },
         white: {
           en: 'Stand in White Half',
-        },
-        vision: {
-          en: 'Double Vision',
         },
       },
     },
--- a/ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
+++ b/ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
@@ -1,0 +1,103 @@
+import Conditions from '../../../../../resources/conditions';
+import NetRegexes from '../../../../../resources/netregexes';
+import { Responses } from '../../../../../resources/responses';
+import ZoneId from '../../../../../resources/zone_id';
+import { RaidbossData } from '../../../../../types/data';
+import { TriggerSet } from '../../../../../types/trigger';
+
+export interface Data extends RaidbossData {
+  sullenDebuff?: boolean;
+  irefulDebuff?: boolean;
+}
+
+// TODO:
+//  Angra Mainyu
+//    Add Level 100 Flare
+//    Add Level 150 Doom
+//    Add Roulette?
+//    Add info text for add spawns?
+//  Five-Headed Dragon
+//  Howling Atomos
+//  Cerberus
+//  Cloud of Darkness
+
+const triggerSet: TriggerSet<Data> = {
+  zoneId: ZoneId.TheWorldOfDarkness,
+  triggers: [
+    {
+      id: 'Angra Mainyu Gain Sullen',
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: '27c' }),
+      condition: Conditions.targetIsYou(),
+      run: (data) => data.sullenDebuff = true,
+    },
+    {
+      id: 'Angra Mainyu Lose Sullen',
+      type: 'LosesEffect',
+      netRegex: NetRegexes.losesEffect({ effectId: '27c' }),
+      condition: Conditions.targetIsYou(),
+      run: (data) => data.sullenDebuff = false,
+    },
+    {
+      id: 'Angra Mainyu Gain Ireful',
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: '27d' }),
+      condition: Conditions.targetIsYou(),
+      run: (data) => data.irefulDebuff = true,
+    },
+    {
+      id: 'Angra Mainyu Lose Ireful',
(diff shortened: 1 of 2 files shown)


jacob-keller:
I have no idea how to write a timeline file, so this currently lacks any timeline data. Just the basic triggers to help remind when to avoid mechanics.

ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
```diff
@@ -0,0 +1,103 @@
...
+          return output.red!();
+        } else if (data.irefulDebuff) {
+          // Stand in white half to switch to Sullen
+          return output.white!();
+        }
+        return output.vision!();
+      },
+      outputStrings: {
+        red: {
+          en: 'Stand in Red Half',
```
quisquous, line 71: I forget which is front and which is back, but can these be "Get In Front (red)" sort of thing?

ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
```diff
@@ -0,0 +1,103 @@
...
+      netRegex: NetRegexes.startsUsing({ id: 'CC8', source: 'Angra Mainyu', capture: false }),
+      alertText: (data, _matches, output) => {
+        if (data.sullenDebuff) {
+          // Stand in red half to switch to Ireful
+          return output.red!();
+        } else if (data.irefulDebuff) {
+          // Stand in white half to switch to Sullen
+          return output.white!();
+        }
+        return output.vision!();
```
quisquous, line 67: Would it be better to say nothing here if you don't have a debuff? I'm not sure that "Double Vision" is really actionable.

jacob-keller: So the main reason I had this was for testing to make sure things were actually appearing. In my log, a lot of players die and then don't get any output without it because they don't have the debuff anymore.

My main concern with printing nothing is that it would potentially feel weird like "aren't i supposed to do something here?" I guess we could also code it so that for DPS it starts with the rear and for Tank it starts with the front side if you have no debuff?

ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
jacob-keller: I'll check which side is which and can fix this.

jacob-keller:
Now I say front or behind, (confirmed on gites, the white/light/blue side is always front).

I also captured a mainTank string and use that to allow Double Vision to default main tank to start on white and everyone else start on red. I think that helps since it means people don't get a weird "why does this not pop up sometimes?" vibe.

ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
```diff
@@ -0,0 +1,111 @@
...
+          return output.red!();
+        } else if (data.irefulDebuff) {
+          // Stand in front of boss in the white half to switch to Sullen
+          return output.white!();
+        } else if (data.mainTank === data.me) {
+          // Main tank should stand in front
+          return output.white!();
+        }
+        // Everyone else generally stand behind
+        return output.red!();
```
quisquous, line 78: Hmm.  Sorry to quibble, but I think tracking the main tank is overthinking it.  The boss is frozen in place during this mechanic and I'm not sure that there's a reason for anybody to stand in any particular place on the first one.

What about "stand in either color" as a no debuff text? I think that's better than "double vision".

quisquous:
> I have no idea how to write a timeline file

Also, I am *MORE* than happy to walk you (or anybody) through how to make timeline files, especially for old content.  (I don't think it needs to be part of this PR, but just for future reference.)  There is a [timeline guide](https://github.com/quisquous/cactbot/blob/main/docs/TimelineGuide.md), but the short of it is that you run `util/make_timeline`.  Then you massage the result into a timeline by ignoring noisy/inconsistent abilities, find loops, and space out each fight in a zone across time.

ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
jacob-keller: Yea that is probably overkill. I'm also happy to just leave the message off. I guess its not that critical it was more of a thought that it might feel weird not to get a pop up at all. The more I think about it, the more keeping it simple seems fine. It would be weird to have to go "oh I guess I don't do anything special". So I'll just go ahead and remove this.

jacob-keller:
> > I have no idea how to write a timeline file
> 
> Also, I am _MORE_ than happy to walk you (or anybody) through how to make timeline files, especially for old content. (I don't think it needs to be part of this PR, but just for future reference.) There is a [timeline guide](https://github.com/quisquous/cactbot/blob/main/docs/TimelineGuide.md), but the short of it is that you run `util/make_timeline`. Then you massage the result into a timeline by ignoring noisy/inconsistent abilities, find loops, and space out each fight in a zone across time.

is there discord or some other place we could discuss this? I got a bit started but I think I am at a loss as to how to properly go past the first stage of just generating the timeline automatically..

quisquous:
The discord is the FFXIV ACT Discord (https://discord.gg/hK523Pj).  Feel free to ask in #overlay-discussion or find me by nickname and DM me there.

PR #3623, opened 2021-11-06
oopsy/test: add player id to PlayerChangedDetail

quisquous:


quisquous:
It's a little bit weird that `PlayerChangedDetail` has id that's a string, but the id is the decimal string (not a number) and not the hex string.  I wonder if either the plugin should have an id that's a decimal number or a hex string.

PR #3638, opened 2021-11-12
i18n: fix oopsy config kr translation

Bing-su:
🙄


quisquous:
Oh whoops, sorry for not catching this!

PR #3645, opened 2021-11-14
config: change param types of getOption/setOption

quisquous:
This will make it easier to add the right types and make the
parameters required.


MaikoTan:
This would solve the problem but looks weird that you should write the default value in front of other parameters.
What about make the function accept something like `getOption(obj, path, defaultValue)` when path can be `string | string[]` and defaultValue is optional.
e.g:
```js
getOption(obj, 'foo')
getOption(obj, ['foo', 'bar'])
getOption(obj, ['foo', 'bar'], 'default')
```

quisquous:
> This would solve the problem but looks weird that you should write the default value in front of other parameters. 

Thanks for the suggestion, done!

PR #3646, opened 2021-11-14
raidboss: fix TEA type error

MaikoTan:
Found in #3613 that cause `process-triggers` failed.

EDIT:
Looks like this configuration
https://github.com/quisquous/cactbot/blob/e617b03118893f68b875f94968e819c13b06696e/tsconfig.json#L18
that cause this error not report by `npm run tsc-no-emit`,
If I comment this line and run `tsc`, errors in this PR would be reported normally,
But from [the description of this parameter in typescript official website](https://www.typescriptlang.org/tsconfig#noUncheckedIndexedAccess), I didn't found any related information that cause the problem.


quisquous:
Thanks for the fix!!

PR #3649, opened 2021-11-15
util: convert gen_pet_names to ts

MaikoTan:
Changed the mechanism that reads from coinach to read from xivapi.
(Maybe we need a general function to fetch data from xivapi? As it is used in gen_fisher_data or else)
At least runs well on my side.

```bash
npm run generate

> cactbot@0.23.10 generate /workspaces/cactbot
> npm run util -- generate


> cactbot@0.23.10 util /workspaces/cactbot
> node --loader ts-node/esm util/index.ts "generate"

(node:7785) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
? Which data file do you want to generate? pet_names
wrote: resources/pet_names.ts
```


quisquous:
> (Maybe we need a general function to fetch data from xivapi? As it is used in gen_fisher_data or else)

Maybe we need a general function that can switch between raw xivapi / raw csv / saint coinach which all should have the same data format.  And then we could switch the source of data in the future more easily?

PR #3651, opened 2021-11-16
util: re: fix d.ts files type errors

MaikoTan:
This PR re-post changes from #3613 which is reverted by #3639


quisquous:
Thanks for sorting through the errors! Sorry for reverting before, it just wasn't obvious what was wrong and I wanted to get the triggers branch back.  Hopefully that won't happen again now that process-triggers fails when it should.  O:)

PR #3652, opened 2021-11-16
util: Fix encounter_tools after Typescriptification

JLGarber:
Sorry this is so extensive! It turns out I didn't do anywhere near enough investigation on how the JS version of the utility actually *worked* before converting it. This restores it to (presumably) exact parity with the pre-Typescript version. Comparing output from this against the JS version, it's exactly the same for all More Than One logs I tried.

One thing that I think we should do in future is extract the various different operation points of `process()` out into helper functions. It's *really* long, and I found myself constantly scrolling up and down to verify what had or hadn't yet happened.

Perhaps the most significant change I made here was mostly eliminating the distinction between the data properties of `Finder` vs `Collector`. Thus, all functions, overridden or not, will refer to `current(Zone|Fight)` rather than `last(Zone|Fight)`. This eliminates a source of annoying bugs in the `process()` function, where there are references to one or the other, which leads to initialization functions hitting the wrong properties, or just not activating at all.

I noted it in-line, but if we don't explicitly check for seals when ending combat, stuff gets really weird. If there are seals present in the current zone, we want to use *only* seals for starting/stopping combat, since those are the only encounters we care about.

When assembling the `sealRegexes` array, we have to explicitly add in the check for whether `s.seal` is defined. This match is an element of the bare regex itself, and is *not* present in the `netLog` definition:
`/^(?<type>(?:00))\|(?<timestamp>(?:[^|]*))\|(?<code>(?:0839))\|(?<name>(?:[^|]*))\|(?<line>(?:(?<seal>.*?) will be sealed off.*?))\|/i`
 (The `name` property in the `netLog` definition is always the empty string.) This is less than ideal, but it's what we have to work with for now.

There's definitely more to be done, but I think this is a good stopping point in terms of "fix it but try not to make too many actual changes".

### DIFF
--- a/util/logtools/encounter_tools.ts
+++ b/util/logtools/encounter_tools.ts
@@ -21,6 +21,7 @@ type ZoneEncInfo = {
 type FightEncInfo = {
   name?: string;
   zoneId?: number;
+  zoneName?: string;
   startLine?: string;
   endLine?: string;
   startTime?: Date;
@@ -83,7 +84,7 @@ export class EncounterFinder {
     if (!unsealReplace)
       throw new Error('missing unseal regex');
     for (const lang of Object.values(unsealReplace)) {
-      const line = `*.?${lang}.*?`;
+      const line = `.*?${lang}.*?`;
       this.unsealRegexes.push(NetRegexes.message({ line: line }));
     }
   }
@@ -91,7 +92,7 @@ export class EncounterFinder {
   skipZone(): boolean {
     // We don't want combat from the overworld or from solo instanced duties.
     // However, if we can't find zone info, we explicitly don't skip it.
-    if (!this.currentZone.zoneId || !this.zoneInfo)
+    if (!this.zoneInfo)
       return false;
     const content = this.zoneInfo.contentType;
     if (!content)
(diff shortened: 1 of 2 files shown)


util/logtools/encounter_tools.ts
```diff
@@ -144,58 +144,55 @@ export class EncounterFinder {
 
     // We are in a combat zone, so we next check for victory/defeat.
     // If either is found, end the current encounter.
-    const cW = this.regex.cactbotWipe.exec(line);
+    // However, *if* seals are present, we want to use them to end the current encounter,
+    // rather than using combat status.
```
quisquous, line 148: > I noted it in-line, but if we don't explicitly check for seals when ending combat, stuff gets really weird. If there are seals present in the current zone, we want to use only seals for starting/stopping combat, since those are the only encounters we care about

I'm not sure I understand why.  What's wrong with ending with either combat or seals?

JLGarber: "Really weird" might be an overstatement, (that's mostly from my memory of automating `make_timeline` in the first place,) but `split_log` throws an error if we don't handle seals this way. I would have to dig through a little to see exactly why,  but interleaving starts and stops this way can leave us with encounters whose start or end times aren't defined.

quisquous:
Thanks for following up here! This looks reasonable to me, and if there's cleanup that makes sense to do after this, please have at.

PR #3658, opened 2021-11-20
util: Initial splitting-out TS timeline argparse

JLGarber:
Tested and fully working (at least as far as I saw) with `split_log`. This isn't final by any means, I'm mostly doing this pull request now to get input on what else is needed.

I don't like the games I had to play with `argsObject`, but there's not much that can be done when the library itself returns `any` from a core method. It works, at least.

The help text is a little too specific to `split_log` right now. We'll need to make changes there. I would *like* to have a unified timeline utility that contains `make/test/adjust`, but I don't think `split_log` belongs with those operations.

When none of the required arguments are supplied, the error message is
`split_log: error: one of the arguments -lf/--search-fights -lz/--search-zones -fr/--fight-regex -zr/--zone-regex -a/--adjust is required`

If 2+ are supplied, it is
`split_log: error: argument "-lz/--search-zones": Not allowed with argument "-lf/--search-fights"`
(And so forth for whatever combination.) *Unfortunately*:

> Note that currently mutually exclusive argument groups do not support the title and description arguments of add_argument_group().

We work with what we have.

The changes to `split_log` beyond simple removal of the argument definitions are to accommodate the slight differences in the Typescript implementation of `argparse`. (Or maybe I just don't get it.) In the Python version, we would specify a `default` property for an argument. Here, if we specify a `default`, that value is *always* applied if that argument is not specified. That is,
```
-f { default: foo }
-t { default: bar }
[invoke] split_log -f
// foo bar
[invoke] split_log -f test.log -t
// test.log bar
```
Instead, we have to specify `constant`, and then be much more explicit about checking the values of the relevant parsed arguments.

The `zone_name` changes are due to two issues in `encounter_tools`. One, `name` is not a consistent property between fight objects and zone objects. We'll probably want to normalize that. Second, `zoneId` doesn't seem to be consistently/correctly populated. (Dzemael Darkhold populates as 3, while Ul'Dah populates as 82.) I didn't test this much, but we'll want to look at it in future. (I don't think it's terribly necessary to resolve that before we eventually land this one though.)


util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,88 @@
+import argparse from 'argparse';
+
+type argsObject = {
```
quisquous, line 3: (a pile of tiny nits, but types should be capitalized.  And maybe I'd bikeshed and call this `TimelineArgs`? And also export this?)

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,88 @@
...
+    }
+    if (numFileArgs === 0) {
+      console.error('Error: Must specify at least one of -f or -t');
+      process.exit(-1);
+    }
+    return args;
+  }
+}
+
+export default TimelineParse;
```
quisquous, line 88: Also, I'll probably try to switch most things over to not exporting by default, so I think this would be better as `export class TimelineParse` above.

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,88 @@
...
+  });
+  // At least one non-file argument must be selected from this group.
+  fileGroup = this.parser.addArgumentGroup({
+    title: 'File Args',
+    description: 'A file and/or a timeline must be selected before performing operations',
+  });
+  // Exactly one argument should be selected from this group.
+  requiredGroup = this.parser.addMutuallyExclusiveGroup({ required: true });
+  args: argsObject;
+  constructor() {
```
quisquous, line 26: Can you add a newline before the constructor too?

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.fileGroup.addArgument(['-t', '--timeline'], {
+      help: 'The filename of the timeline to test against, e.g. ultima_weapon_ultimate',
+    });
+    this.parser.addArgument(['--force'], {
+      nargs: 0,
+      help: 'Overwrite files when exporting',
+    });
+    this.requiredGroup.addArgument(['-lf', '--search-fights'], {
+      nargs: '?',
+      constant: -1,
```
quisquous, line 39: Is it possible to not have magic constants here? Could this be `null` or even better `undefined`?  I'm not sure I understand why you need to check for both `-1` and `null` below.  Could you explain?

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.requiredGroup.addArgument(['-a', '--adjust'], {
+      nargs: '?',
+      constant: 0,
+      type: 'float',
+      help: 'Adjust all entries in a timeline file by this amount',
+    });
+    this.args = this.validate(this.parser.parseArgs() as argsObject);
+  }
+
+  validate(args: argsObject): argsObject {
```
quisquous, line 72: I think maybe `validate` should be a `void` function, as it's not changing or modifying `args`.

util/logtools/split_log.js
```diff
@@ -113,10 +56,15 @@ const generateFileName = (fightOrZone) => {
   else
     seal = '';
 
-  const idToZoneName = {};
-  for (const zoneName in ZoneId)
-    idToZoneName[ZoneId[zoneName]] = zoneName;
-  const zoneName = idToZoneName[zoneId];
+  let zoneName;
+  if (fightOrZone.zoneName || fightOrZone.name) {
+    zoneName = fightOrZone.zoneName || fightOrZone.name;
```
quisquous, line 61:
```suggestion
    zoneName = fightOrZone.zoneName ?? fightOrZone.name;
```

Also, is this an unrelated change that just improves names when there's not a zone name? I'm not sure I understand why the logic change is here.  Just trying to clarify my understanding.

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,88 @@
...
+    });
+    this.requiredGroup.addArgument(['-fr', '--fight-regex'], {
+      nargs: '?',
+      constant: '*',
+      type: 'string',
+      help: 'Export all fights that match this regex',
+    });
+    this.requiredGroup.addArgument(['-zr', '--zone-regex'], {
+      nargs: '?',
+      constant: '*',
```
quisquous, line 59: What does constant `'*'` mean?

util/logtools/arg_parser.ts
JLGarber: What I was thinking, at least, is that whatever is consuming `TimelineParse` might as well have the common validations performed before, rather than manually validating everything in each utility. Maybe something like this in the constructor would be preferable?

```
this.args = this.parser.parseArgs() as argsObject;
validate(this.args);
```

util/logtools/split_log.js
JLGarber: This is related to what I noted about the discrepancy between `ZoneEncInfo` and `FightEncInfo`. The property `zoneName` is present only on `FightEncInfo`, which is probably a mistake that we should fix separately. (That is, we likely should be explicit about *what each name is*, rather than have *anything* just called a generic "name".)

That being said, you're completely correct that nullish coalescing here is better. I'm not sure why I didn't go with that originally.

util/logtools/arg_parser.ts
JLGarber: That shouldn't be there. I think I intended for it to be `.?`, but that's *probably* not a good idea.

util/logtools/arg_parser.ts
JLGarber: I'm not completely sure whether it's possible. We need a constant defined in order for bare flags to work *at all*, and the Typescript version of `argparse` treats falsy constants as if the bare flags weren't actually present. (That is, if `--foo` has no constant definition, or if that constant is falsy, there is no difference between `[invoke]` and `[invoke] --foo`.)

That being said, you're entirely correct that if the user is not printing the list of encounters/zones, we could simplify the subsequent check down to `if(args[arg])`

util/logtools/arg_parser.ts
JLGarber: It actually occurs to me, should we rename this as something like `LogUtilArgParse`? I know it's long, but it's significantly more specific about what it's actually used for.

quisquous: Sounds good!

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,87 @@
...
+  validate(args: TimelineArgs): TimelineArgs {
+    // We can't enforce 1+ arguments in a non-exclusive group,
+    // so we have to do it manually here.
+    let numFileArgs = 0;
+    for (const opt of ['file', 'timeline']) {
+      if (args[opt] !== null)
+        numFileArgs++;
+    }
+    if (numFileArgs === 0) {
+      console.error('Error: Must specify at least one of -f or -t');
```
quisquous, line 80: Reading this validate function, I'm not actually sure I understand this error.  Don't you need `-t` with -`f`?  Or, said differently, isn't `-f` mandatory but `-t` optional?

util/logtools/arg_parser.ts
```diff
@@ -0,0 +1,87 @@
...
+    this.requiredGroup.addArgument(['-a', '--adjust'], {
+      nargs: '?',
+      constant: 0,
+      type: 'float',
+      help: 'Adjust all entries in a timeline file by this amount',
+    });
+    this.args = this.validate(this.parser.parseArgs() as TimelineArgs);
+  }
+
+  validate(args: TimelineArgs): TimelineArgs {
```
quisquous, line 71: > I don't like the games I had to play with argsObject, but there's not much that can be done when the library itself returns any from a core method. It works, at least.

For what it's worth, you could make this a [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards), e.g.

```typescript
const args = this.parser.parseArgs();
if (isTimelineArgs(args))
  this.args = args; // args is implicitly TimelineArgs inside this conditional due to type guard
if (!this.args)
  process.exit(-1);

// ...and then elsewhere...

isTimelineArgs(args: unknown): args is TimelineArgs {
  // TODO: first make sure args is of type object and not an array

  const validArgs: Required<TimelineArgs> {
    file: '',
    // fill out rest of args here
  };
  for (const [key, value] of Object.entries(validArgs)) {
    if (key in args && typeof args[key] !== typeof value)
      return false;
  }
  return true;
}

util/logtools/arg_parser.ts
JLGarber: No, because adjusting timelines with `-a` should not require a log file.

quisquous: Oh, I see! I think that makes me think that this sort of validation should go to the script itself, since they each need different things.  Adjusting requires `-t`.  Testing requires `-f` (or fflogs) and `-t`.  Making requires `-f` (or fflogs).

JLGarber: I suppose? I sort of feel like every log utility is guaranteed to want at least one of a log file or a timeline file, so it makes sense to validate for that in the common log utility arguments script. But I'm not that attached to it, we can certainly do validation on a per-script basis.

util/logtools/arg_parser.ts
JLGarber: I'll poke at this once I finish up with the timeline conversions. This suggestion would definitely be less potentially fallible than what I did.

JLGarber:
I'm back to poking at this one. At this point, I think I'll probably leave the `TimelineArgs` type definition as it is for now. I'm having some trouble grasping how to get from Point A to Point B on the type guard implementation, and in theory making this change in the future should be a simple drop-in replacement.

As far as argument addition and validation, I don't think we really settled how to handle that. Are we okay with all timeline arguments being exported from `LogUtilArgParse` but then narrowed down in validation on a per-script basis?

Once we figure out the validation and get this landed, I think our next steps will probably be to move a lot of the utility functions from `split_log` over to `encounter_tools`. From there, we can re-implement the ``FFLogs helper library in Typescript and then re-implement the other timeline utilities.

quisquous:
> As far as argument addition and validation, I don't think we really settled how to handle that. Are we okay with all timeline arguments being exported from LogUtilArgParse but then narrowed down in validation on a per-script basis?

:+1:

JLGarber:
Closing in favor of #4335

PR #3665, opened 2021-11-21
util: coinach add helper function that sorts keys

MaikoTan:
The original approach that calls
`JSON.stringify(data, Object.keys(data).sort(), 2)` works well
 in some situation, but it would
only sort the first dimention of the object,
and the worst thing is that it would drop nested object.
For example, call the function above with a object
like `{ foo: { bar: 1 } }`, the actual output would be
'{"foo":{}}' that lost nested object.

So the ideal way is to recursively sort keys before passing
to `JSON.stringify` function.

### DIFF
--- a/util/coinach.ts
+++ b/util/coinach.ts
@@ -154,6 +154,23 @@ export class CoinachWriter {
     return '.';
   }
 
+  sortObjByKeys(_obj: unknown): unknown {
+    const obj = _obj as Record<string, unknown>;
+    const out = Object
+      .keys((obj))
+      .sort()
+      .reduce((acc: Record<string, unknown>, key) => {
+        const nested = obj[key] as Record<string, unknown>;
+        if (nested && typeof nested === 'object' && !Array.isArray(nested))
+          acc[key] = this.sortObjByKeys(nested);
+        else
+          acc[key] = nested;
+
+        return acc;
+      }, {});
+    return out;
+  }
+
   async write(
     filename: string,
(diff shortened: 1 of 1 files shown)


util/coinach.ts
```diff
@@ -154,6 +154,23 @@ export class CoinachWriter {
     return '.';
   }
 
+  sortObjByKeys(_obj: unknown): unknown {
+    const obj = _obj as Record<string, unknown>;
```
quisquous, line 158: Is it possible to do this without type assertion? `sortObjByKeys` could be called with a `number` or something that you can't call Obect.keys on.

I don't think there's a good way to go from unknown to an indexable type.  (You can get to `object` or `{}` but there's lint banning those.)  You could add a type guard though, which seems slightly better than a type assertion.

```typescript
const isObject = (obj: unknown): obj is { [key: string]: unknown } => {
  return obj !== null && typeof obj === 'object' && !Array.isArray(obj);
};

  sortObjByKeys(obj: unknown): unknown {
    if (!isObject(obj))
      return;
    const out = Object
      .keys((obj))
      .sort()
      .reduce((acc: typeof obj, key) => {
        const nested = obj[key];
        if (isObject(nested))
          acc[key] = this.sortObjByKeys(nested);
        else
          acc[key] = nested;

        return acc;
      }, {});
    return out;
  }
```

MaikoTan: Done.
I have no better solution either...

PR #3675, opened 2021-11-24
ci: bump node version to 16

Trim21:
nodejs 16 has been the latest lts version, and it's packed with a new npm version which require a v2 version of lock file.

also bump version of `ts-node` and `webpack` to make it work with new node loader api, maybe wait for #3676 get merged.

(there are also some change about dprint config to make it work better with vscode)


quisquous:
Thanks for this.  I'll take a look at this PR this weekend when I have more time.

Could you split out the dprint changes into their own PR? Are they unrelated to this node version change?

Trim21:
I just revert the changes about dprint.

.github/workflows/build-artifact.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     runs-on: windows-latest
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v2
```
quisquous, line 31: Please leave these as v1.  See: https://github.com/quisquous/cactbot/pull/3679#issuecomment-981128556.

.github/workflows/test.yml
```diff
@@ -23,12 +23,15 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node_version: [14, 16]
```
quisquous, line 28: Do we need to run tests on multiple versions of node?

quisquous (commented):
It looks like a bunch of the changes in #3676 are also in this PR.  I will look at #3676 first.

.github/workflows/test.yml
Trim21: maybe not, then we should also bump the version of `package.json#engines.node` I think.

tsconfig-for-webpack.json
```diff
@@ -1,7 +1,25 @@
...
+      // Globs are also supported with the same behavior as tsconfig "include"
+      "webpack/**/*": "cjs"
+    }
+  },
   "compilerOptions": {
-    "module": "commonjs",
-    "target": "es5",
+    "plugins": [
+      {
+        "transform": "@zoltu/typescript-transformer-append-js-extension/output/index.js",
```
quisquous, line 15: Can this file use `extends` like `tsconfig.json` does, so it can share these settings with tsconfig.json and tsconfig.base.json?

.github/workflows/test.yml
quisquous: Sorry, I'm not sure I understand why this conversation was resolved.  I was imagining that you would remove the `matrix`, only test against 16, and update the `engines` line too?

tsconfig-for-webpack.json
```diff
@@ -1,7 +1,16 @@
 {
+  "extends": "./tsconfig.base.json",
```
quisquous, line 2: Should this extend `tsconfig.json`? I'm not sure I understand what the difference is between the "base" version and that one.

tsconfig-for-webpack.json
```diff
@@ -1,7 +1,16 @@
...
+      // Globs are also supported with the same behavior as tsconfig "include"
+      "webpack/**/*": "cjs"
+    }
+  },
   "compilerOptions": {
-    "module": "commonjs",
-    "target": "es5",
-    "esModuleInterop": true
+    "moduleResolution": "node",
+    "skipLibCheck": false
```
quisquous, line 14: Can you explain why the compiler options are changing here and why we need extra settings for `webpack/*`?

I also think that we should `skipLibCheck: true` here until we also make it true in `tsconfig.base.json` so that we do the lib check everywhere.  I would prefer not to only fail in webpack and also fail when building.  I would suggest to skip the lib check now, and we can have another PR that turns it on everywhere.

tsconfig-for-webpack.json
Trim21: I think `base` mean to be extended.

maybe ask @MaikoTan here

tsconfig-for-webpack.json
Trim21: > 

in node14, it works fine if a cjs try to require a esm(.mjs or .ts file), but it doesn't work in node 16(or I say ts-node v0.10)

`target` and  `module` set to `"es2020“` or `”esnext“` is recommand by ts-node for configuraing webpack.

quisquous: Thanks for all the explanation, I really appreciate it.  That all sounds good to me.  :+1:

package.json
```diff
@@ -9,9 +9,9 @@
...
   "scripts": {
-    "build": "cross-env TS_NODE_PROJECT=\"tsconfig-for-webpack.json\" webpack --config webpack/webpack.prod.ts",
-    "build-gh-pages": "cross-env TS_NODE_PROJECT=\"tsconfig-for-webpack.json\" webpack --config webpack/webpack.ghpages.ts",
-    "start": "cross-env TS_NODE_PROJECT=\"tsconfig-for-webpack.json\" webpack serve --config webpack/webpack.dev.ts",
+    "build": "cross-env NODE_OPTIONS=\"--loader ts-node/esm\" TS_NODE_PROJECT=\"tsconfig-for-webpack.json\" webpack --config webpack/webpack.prod.ts",
+    "build-gh-pages": "cross-env NODE_OPTIONS=\"--loader ts-node/esm\" TS_NODE_PROJECT=\"tsconfig-for-webpack.json\" webpack --config webpack/webpack.ghpages.ts",
+    "start": "cross-env NODE_OPTIONS=\"--loader ts-node/esm\" TS_NODE_PROJECT=\"tsconfig-for-webpack.json\" webpack serve --config webpack/webpack.dev.ts",
     "fetch-deps": "node --loader ts-node/esm ./util/fetch_deps.ts",
     "tsc-no-emit": "node_modules/.bin/tsc --noEmit",
     "lint": "node_modules/.bin/eslint --max-warnings=0 ./",
```
quisquous, line 17: I think you need to `s/node_modules\/.bin\///` here.  Things like `npm run lint` don't work with node v16 if it says `node_modules/.bin/eslint` instead of just `eslint` on its own.

Trim21: Current config works fine both on my computer and ci.

quisquous: Sorry, I don't know what's wrong, but the current config doesn't work for me.

Does it work if you remove `node_modules/.bin/` as well? If so, then maybe that's a solution that will more consistently work for everybody.  I know in the past somebody was suggesting we should remove that path anyway.

quisquous:
```
enne@thuse MINGW64 ~/cactbot (Trim21-npm-lock-v2)
$ npm run test

> cactbot@0.23.10 test
> node_modules/.bin/mocha

'node_modules' is not recognized as an internal or external command,
operable program or batch file.
```

quisquous: Haha, thank you.  I'm sorry, I don't know why it works for you but not for me!

Trim21:
Something error with webpack, taking too much time.

quisquous:
Hmm, looking at https://github.com/quisquous/cactbot/actions/workflows/build-artifact.yml, it seems like this patch periodically doesn't finish the `npm run build` step, although I wonder if it's a github issue as it doesn't even let me see the logs in progress.  I think I'm going to land this for now and see how it goes.

Trim21:
If a step is running, you can only see new log of this step from web. Full log is only available after this step is finished.

Trim21:
result in running forever

```text
> cactbot@0.23.11 build
> cross-env NODE_OPTIONS="--loader ts-node/esm" TS_NODE_PROJECT="tsconfig-for-webpack.json" webpack --config webpack/webpack.prod.ts

(node:2412) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(node:1632) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(node:4812) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
(node:2412) ExperimentalWarning: --experimental-loader is an experimental feature. This feature could change at any time
```

quisquous:
You're saying that even locally it sometimes runs forever?

Trim21:
> You're saying that even locally it sometimes runs forever?

no, never. I mean in CI.

PR #3676, opened 2021-11-24
build: bump webpack and its friend newer newer

MaikoTan:
- In webpack v5 should use `webpack serve` with `webpack-dev-server@v4` instead of calling `webpack-dev-server` directly
- `worker-loader` was no longer used since https://github.com/quisquous/cactbot/commit/2f1f02cb1dee6bd012b11e587fcac00eb1e75512 but isn't removed from `package.json`
- since `webpack-dev-server` is bumped to v4, apply migrations based on https://github.com/webpack/webpack-dev-server/blob/master/migration-v4.md


quisquous:
Thanks for this.  I'll take a look at this PR this weekend when I have more time.

quisquous:
The config page throws errors for me with this patch.  Steps:
* npm start
* open ACT
* go to ACT -> Plugins -> Cactbot
* observe log

I get continuous:
```
WebSocketClient.js:13 WebSocket connection to 'ws://0.0.0.0:8080/ws' failed: Error in connection establishment: net::ERR_ADDRESS_INVALID WebSocketClient @ WebSocketClient.js:13
index.js:558 [webpack-dev-server] Event index.js:558
```

This seems potentially similar to #3663 in terms of it being websocket related, although I see that you updated the config to use the new format to reflect that patch, so maybe not?

quisquous:
Sorry, false alarm.  It looks like this is happening without your patch as well.

quisquous:
> Sorry, false alarm. It looks like this is happening without your patch as well.

Sorry for the repeated confusion.

I went back several times back and forth, and I don't know why it looked like this was happening without your patch.  I'm going to say that maybe I didn't close ACT between tests.  I am reasonably confident that it is caused by this patch.  Sorry for the confusion, but I was testing a lot of things today.

Are you seeing these errors?

webpack/webpack.config.ts
```diff
@@ -75,9 +75,13 @@ export default (
...
-      contentBase: path.join(__dirname, '../dist'),
-      writeToDisk: true,
-      disableHostCheck: true,
+      static: {
+        directory: path.join(__dirname, '../dist'),
+      },
+      devMiddleware: {
+        writeToDisk: true,
+      },
+      allowedHosts: 'all',
```
quisquous, line 84:
```suggestion
      allowedHosts: 'all',
      client: {
        webSocketURL: {
          hostname: 'localhost',
        },
      },
```

This looks like the fix to get the config page to not throw errors and support hot reloading again.  I found this from [this doc](https://webpack.js.org/configuration/dev-server/) and played with several options until I found something that worked.

MaikoTan:
Well, I have only tested pages in Chrome with the OVERLAY_WS parameter, and didn't see any errors.
I would do the test tomorrow because I am on the train right now.

MaikoTan:
Yeah, it was okay on my side, too.

PR #3677, opened 2021-11-24
lint: imporve editor config

Trim21:
this PR make dprint work better for vscode users


.vscode/settings.json
```diff
@@ -19,6 +19,21 @@
...
     "dist/**": true
   },
+  "[jsonc]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[json]": {
+    "editor.defaultFormatter": "dprint.dprint"
+  },
+  "[yaml]": {
+    "editor.defaultFormatter": null
```
Trim21, line 29: for some prettier extension users (like myself)

package.json
```diff
@@ -117,6 +117,7 @@
     "*.{js,cjs,mjs}": "eslint --max-warnings=0",
     "*.ts": "eslint --max-warnings=0 --plugin tsc --rule 'tsc/config: [2, {configFile: \"./tsconfig.json\"}]'",
     "*.md": "markdownlint",
+    "*.json": "dprint fmt",
```
quisquous, line 120: Should eslint also handle json via dprint? Having it here will handle it via husky, but nothing will catch it later.

quisquous (commented):
Is there a way to add a `dprint.path` config variable that points to the relative npm installed version of dprint so that cactbot devs don't have to install dprint and add dprint to their global PATH variable? (It looks like maybe that dprint is out of date, though.  But, it would be nice to use the same version of dprint in vscode and in cactbot cli for consistency.)

package.json
Trim21: not sure if eslint recognize json file

quisquous: I think there are some plugins?

Alternatively, maybe we can remove this line until eslint is configured to handle it? Otherwise, this is autoformatting at the lint-staged step (which can be skipped) and CI won't catch json.

Trim21: Don't think it's necessary, Jason is a pretty simple language. I add this just because dprint has the ability to format json.

quisquous: It's both not changing commit in hooks, but also the CI lint step not throwing errors if json isn't formatted.

Trim21:
> Is there a way to add a `dprint.path` config variable that points to the relative npm installed version of dprint so that cactbot devs don't have to install dprint and add dprint to their global PATH variable? (It looks like maybe that dprint is out of date, though. But, it would be nice to use the same version of dprint in vscode and in cactbot cli for consistency.)

Not sure, I need to test if workspace path variable works here.

quisquous:
> > Is there a way to add a `dprint.path` config variable that points to the relative npm installed version of dprint so that cactbot devs don't have to install dprint and add dprint to their global PATH variable? (It looks like maybe that dprint is out of date, though. But, it would be nice to use the same version of dprint in vscode and in cactbot cli for consistency.)
> 
> Not sure, I need to test if workspace path variable works here.

Ok, if it works, that'd be great.  If not, then no worries.  I was just curious if there was a way to make this more convenient.

MaikoTan:
> If it must be a absolute path without variable support, developers'll have to install it global or remove this settings.json (name it as `settings.example.json` or something else, I'd like we take this way, make devs be able to configure their vscode workshop more flexibility)

要不寫個NPM hook脚本（比如説install），如果找不到`.vscode/settings.json`文件就把這個實例文件複製過去？
What about to write an NPM hook script (`install` for example), if there is no a `.vscode/settings.json` file, then copy this example file to that location?

Trim21:
> If it must be a absolute path without variable support, developers'll have to install it global or remove this settings.json (name it as `settings.example.json` or something else, I'd like we take this way, make devs be able to configure their vscode workshop more flexibility)
> 
> 要不寫個NPM hook脚本（比如説install），如果找不到`.vscode/settings.json`文件就把這個實例文件複製過去？ What about to write an NPM hook script (`install` for example), if there is no a `.vscode/settings.json` file, then copy this example file to that location?

IMO a example.json is enough

.vscode/settings.example.json
```diff
@@ -19,6 +19,18 @@
     "**/node_modules": true,
     "dist/**": true
   },
+  "[jsonc]": {
```
quisquous, line 22: I still think this is better as settings.json for now.  Re: disabling yaml, I think that's fine for now and I think a future answer is to add a yaml formatter to the editor settings, the lint CI, and the lint-staged steps.

.editorconfig
```diff
@@ -10,8 +10,11 @@ root = true
 insert_final_newline = true
 indent_style = space
 
-[*.{js,json,cjs,mjs,ts}]
+[*.{js,json,cjs,mjs,ts,yml,yaml}]
 indent_size = 2
 
+[*.{yaml,yml}]
+quote_type = single
```
MaikoTan, line 17: It is noted that the single quotes and double quotes behave differently in yaml as double allows escaped characters but single ignore escape and print it as-is.
Is this option affects when we are using escape in double quotes?

Trim21: It's for prettier, too

MaikoTan: 沒，我只是提示一下，畢竟yaml在這方面似乎有點嚴格。
No I just notice it if it may broke something. As Yaml is a bit stricky on these things.

.vscode/settings.example.json
quisquous: This is still the sticking point for me here.  I am not convinced that moving to a separate file is worth the friction that it creates yet.  If there was more than just this one thing, maybe I would feel differently, but at this point I think having a single file that auto-updates for devs is better.

quisquous:
Wait, there were a lot of good things in this PR though other than that.  Are you going to open another?

Trim21:
feel free&nbsp;to&nbsp;cherry pick&nbsp;the changes you like&nbsp;




------------------&nbsp;Original&nbsp;------------------
From: ***@***.***&gt;; 
Date: 2021年12月22日(星期三) 凌晨4:11
To: ***@***.***&gt;; 
Cc: ***@***.***&gt;; "State ***@***.***&gt;; 
Subject: Re: [quisquous/cactbot] lint: imporve editor config (PR #3677)





 
Wait, there were a lot of good things in this PR though other than that.  Are you going to open another?

quisquous:
Ok.  I'm sorry if I created too much of a pain for you in this review that you felt like you just had to drop it.  There's a lot of good things here, and I just wasn't convinced by part of them.  It's possible also that I just don't understand the reasoning here too.

Trim21:
No, just AFK for some time wait for CN server 6.0, not spending much time on FF ( and cactbot ) these days.

PR #3679, opened 2021-11-24
ci: bump actions/checkout to v2

Trim21:
image

image

`actions/checkout@v1` will fetch all history commits and tags, and it's very slow.

test ci takes only 1m37s with v2


quisquous:
I think we are deliberately on v1 because of the issue linked in https://github.com/quisquous/cactbot/pull/1604.

PR #3682, opened 2021-11-25
jobs: full refactoring for job module

MaikoTan:
A full refactoring for the jobs module, hope it makes easier in 6.0 jobs module renewing.

### Constructure

Class `Bars` was moved into `bars.ts`, and those code related to player is moved to a new class `Player`, code related to log parsing or events data from C# plugin is moved to a new class `JobsEventEmitter`.

Now the structure is like:

|file|class|feature|
|:-:|:-:|:-:|
|jobs.ts|none|entry of jobs module, initialize several classes|
|bars.ts|Bars|helds HTML DOM nodes, expose some API that can add or modify HTML nodes|
|event_emitter.ts|JobsEventEmitter|register overlay plugin listener and provides event subscription, do some pre-process on event data. Many of other classes subscribe event from it.|
|player.ts|Player|mentaining player data, this class is also an event emitter that expose some API to ease event subscribing. When player's status changed, the corresponding event emittes.|
|components/index.ts|ComponentManager|the entry class of jobs module, register events and invoke methods from `Bars` and other job components.|
|component/base.ts|BaseComponent|the father class of all jobs components, held instances of `Player` and `Bars` and so on, register events and provide methods for overriding.|

### Event Handling

I imported `eventemitter3` which acts just like the `event` module of NodeJS does. So that I can split out several blocks that do different things, would improve the readability a bit I think.

### Job Components

Now all "components" are classes that inherit `BaseComponent` in `ui/jobs/component/base.ts`, `BaseComponent` would register the most popular events and expose several `onXxx` methods that can be overridden by derived classes.
Job Components can also access `this.player`, `this.bars` or else for their exposed API or events.

### User Customisation (Incomplete)

I didn't find a proper way to provide user customisation.
My feeling is like to expose a field in `Options`, for example:
```ts
// ui/jobs/jobs_options.ts
export type JobsOptions = {
  // ...
  OverrideComponent: (job: Job, override: (Component: typeof BaseComponent) => BaseComponent) => void;
  // ...
}

// user/jobs.js
Options.OverrideComponent('BLM', (BLMComponent) => {
  return class OverrideBLMComponent extends BLMComponent {
    // hide buffs list and hp bar
    shouldShow = { buffsList: false, hpBar: false };

    // overrides others, like access `this.player.on('mp', ({ mp }) => {});` to get alert on every mp updates
  }
});
```

The `this.shouShow` is typed here (although I didn't implemented it completely):

https://github.com/MaikoTan/cactbot/blob/f7c44e3c8b9a71432effd1dc80546eae51e085e4/ui/jobs/components/base.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L11-L19

https://github.com/MaikoTan/cactbot/blob/f7c44e3c8b9a71432effd1dc80546eae51e085e4/ui/jobs/components/index.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L175-L184

Users can access `this.bars` `this.player` `this.emitter` for most data in-game or subscribe events, but there are problems such as that user cannot turn off boxes they don't like because now those are written in the constructors, or access global variables like `kAbility` `EffectId`.

------

## Note

I have only caster and healer at the max level, so I can only make sure that components for casters and healers works properly for now.
I'll appreciate if there is someone test it with all jobs.


quisquous:
This is a lot of changes and is really hard to review.  I will try to take a look at this and test this during the next week, but it is a lot to understand.  I tried to say this on your other PR, but in general it is much easier to break changes like this up into smaller refactorings where possible rather than trying to do everything in one PR that changes everything.

MaikoTan:
I think I have only imported a new dependency `eventemitter3` that provides APIs just like `event` from Node native?
I refactored this module but basically didn't change its features, so maybe you can checkout codes and run it locally, rather than checking the compare view on Github.
I try my best to explain that the roles of files I modified in the module and their features in the description.

quisquous:
I will try to take a look today.  I have all jobs at 80 so I should be able to test everything.

ui/jobs/bars.ts
```diff
@@ -0,0 +1,784 @@
...
+  _setupJobContainers(job: Job, shows: ShouldShows): void {
+    const shouldShows = {
+      buffList: true,
+      pullBar: true,
+      hpBar: true,
+      mpBar: true,
+      cpBar: false,
+      gpBar: false,
+      mpTicker: false,
+      ...shows,
```
quisquous, line 98: I'm not sure I understand this `ShouldShows` data flow.  Doesn't the call from `components/index.ts` set all of these values, so maybe `_setupJobContainers` should just take a `job` parameter and make the `Util` calls itself? Or, maybe just `const shouldShows = shows;` (or rename the parameter)?

ui/jobs/jobs.css
```diff
@@ -1137,3 +1143,8 @@
     opacity: 0;
   }
 }
+
+/** hide bars except hp-bar and mp-bar */
```
quisquous, line 1147: This is a little bit of what I'm talking about in terms of "why is this patch so big".  This is both a components/event emitter patch, but also adds additional functionality like checking for pvp zones.  I think you could have added the pvp changes before or after the refactoring patch, and it would have made this patch easier to read because it is smaller.

ui/jobs/event_emitter.ts
```diff
@@ -0,0 +1,231 @@
...
+  // party changed
+  'party': (party: Party[]) => void;
+  // zone changing
+  'zone/change': (id: number, name: string, info?: typeof ZoneInfo[number]) => void;
+  // battle events
+  'battle/in-combat': (info: { game: boolean; act: boolean }) => void;
+  'battle/wipe': () => void;
+  'battle/target': (target?: { name: string; distance: number; effectiveDistance: number }) => void;
+  // triggered when effect gains or loses
+  'effect/gain': (effectId: string, info: Partial<ToMatches<NetFields['GainsEffect']>>) => void;
```
quisquous, line 28: I think `Partial<ToMatches<NetFields[string]>>` can be written as `Partial<NetMatches[string]>` or maybe even better use your new `PartialFieldMatches<string>` helper?

ui/jobs/player.ts
```diff
@@ -0,0 +1,481 @@
...
+      this.jobDetail = data.jobDetail;
+      this.emit('job-detail', data.job, data.jobDetail);
+    }
+
+    this.emit('player', this);
+  }
+
+  private processPlayerStatsLogLine(line: string[]): void {
+    const matches = normalizeLogLine(line, logDefinitions.PlayerStats.fields);
+    // const stat = Object
```
quisquous, line 390: Is this old code and the comment should be removed?

ui/jobs/event_emitter.ts
```diff
@@ -0,0 +1,231 @@
...
+    this.targets = [];
+
+    this.registerListeners();
+  }
+
+  private registerListeners(): void {
+    this.player.on('effect/gain', (id, { sourceId, targetId }) => {
+      if (
+        targetId?.startsWith('4') &&
+        parseInt(sourceId ?? '0', 16) === this.player.id &&
```
quisquous, line 179: What about storing the hex id of the player instead (or in addition)? I feel like there are a lot of `parseInt(x, 16)` around that could be simpler if we kept the hex id around.

quisquous (commented):
I haven't tested yet, and I'm still reading, but here are some initial comments.

ui/jobs/components/nin.ts
```diff
@@ -1,123 +1,161 @@
...
 
-  let mudraTriggerCd = true;
-  // Ninjutsu's cooldown begins to countdown at the first mudra.
-  bars.onYouGainEffect(EffectId.Mudra, () => {
-    if (!mudraTriggerCd)
-      return;
+    this.ninki = this.bars.addResourceBox({
+      classList: ['nin-color-ninki'],
+    });
+    this.hutonBox = this.bars.addProcBox({
```
quisquous, line 28: Both huton and ninjutsu both appear to not work, but it appears that it was broken before your patch too.

ui/jobs/components/brd.ts
```diff
@@ -1,162 +1,197 @@
...
+    this.repertoireBox.innerText = '';
     if (jobDetail.songName === 'Minuet') {
-      repertoireBox.innerText = jobDetail.songProcs.toString();
-      repertoireBox.parentNode.classList.add('minuet');
-      songBox.fg = computeBackgroundColorFrom(songBox, 'brd-color-song.minuet');
-      songBox.threshold = 5;
-      repertoireBox.parentNode.classList.remove('full');
+      this.repertoireBox.innerText = jobDetail.songProcs.toString();
+      this.repertoireBox.parentNode.classList.add('minuet');
+      this.songBox.fg = computeBackgroundColorFrom(this.songBox, 'brd-color-song.minuet');
```
quisquous, line 101: I don't see this box ever proc.  However, this also happens without your patch, so it's not an error here, but a bug for later.

quisquous:
I also once saw:
```
overlay_plugin_api.ts:93 TypeError: Cannot read properties of undefined (reading 'toString')
    at MNKComponent.onJobDetailUpdate (mnk.ts:53)
    at EventEmitter.wrapper (player.ts:274)
    at EventEmitter.emit (index.js:203)
    at Player.emit (player.ts:483)
    at Player.processPlayerChangedEvent (player.ts:386)
    at JobsEventEmitter.<anonymous> (player.ts:179)
    at JobsEventEmitter.emit (index.js:181)
    at event_emitter.ts:49
    at overlay_plugin_api.ts:91
    at Array.forEach (<anonymous>)
```

... but was unable to reproduce.  Maybe there's some race where the job changes and we try to evaluate the job detail based on the wrong job so it has undefined fields? I think this is minor, but something worth thinking about if you have any ideas.

MaikoTan:
> I also once saw:
> 
> ```
> 
> overlay_plugin_api.ts:93 TypeError: Cannot read properties of undefined (reading 'toString')
> 
>     at MNKComponent.onJobDetailUpdate (mnk.ts:53)
> 
>     at EventEmitter.wrapper (player.ts:274)
> 
>     at EventEmitter.emit (index.js:203)
> 
>     at Player.emit (player.ts:483)
> 
>     at Player.processPlayerChangedEvent (player.ts:386)
> 
>     at JobsEventEmitter.<anonymous> (player.ts:179)
> 
>     at JobsEventEmitter.emit (index.js:181)
> 
>     at event_emitter.ts:49
> 
>     at overlay_plugin_api.ts:91
> 
>     at Array.forEach (<anonymous>)
> 
> ```
> 
> 
> 
> ... but was unable to reproduce.  Maybe there's some race where the job changes and we try to evaluate the job detail based on the wrong job so it has undefined fields? I think this is minor, but something worth thinking about if you have any ideas.

So it is happened on a job change? I think I have code that compares job name before emitting the job detail, but I am not very sure if there are bugged conditions or something.
But I afraid that I can't test more as the FFXIV server is going to mentain for a whole day at 18:00 today (japan time), and I have bought Early Access which means I cannot access 5.0 after mentainance any more. Maybe I can only rebase this after 6.0 launched

quisquous:
> > I also once saw:
> > ```
> > 
> > overlay_plugin_api.ts:93 TypeError: Cannot read properties of undefined (reading 'toString')
> > 
> >     at MNKComponent.onJobDetailUpdate (mnk.ts:53)
> > 
> >     at EventEmitter.wrapper (player.ts:274)
> > 
> >     at EventEmitter.emit (index.js:203)
> > 
> >     at Player.emit (player.ts:483)
> > 
> >     at Player.processPlayerChangedEvent (player.ts:386)
> > 
> >     at JobsEventEmitter.<anonymous> (player.ts:179)
> > 
> >     at JobsEventEmitter.emit (index.js:181)
> > 
> >     at event_emitter.ts:49
> > 
> >     at overlay_plugin_api.ts:91
> > 
> >     at Array.forEach (<anonymous>)
> > ```
> > 
> > 
> >     
> >       
> So it is happened on a job change? I think I have code that compares job name before emitting the job detail, but I am not very sure if there are bugged conditions or something.

Yeah, it happened on a job change, but only once.

> But I afraid that I can't test more as the FFXIV server is going to mentain for a whole day at 18:00 today (japan time), and I have bought Early Access which means I cannot access 5.0 after mentainance any more. Maybe I can only rebase this after 6.0 launched

Ack.  I am sorry it took me so long to get to this.  I was out of town and also wanted to finish getting everything ready for the alpha plugin changes as well.

quisquous:
I do think the crafting CP bar and the missing mp bar colors need to get fixed before this is merged, but the one off `toString` error seems fine to just see if we ever see it again.

package.json
```diff
@@ -128,7 +127,9 @@
     "argparse": "^1.0.10",
     "dexie": "^3.0.3",
     "dexie-export-import": "^1.0.0",
+    "eventemitter3": "^4.0.7",
     "inquirer": "^8.1.1",
-    "inquirer-fuzzy-path": "^2.3.0"
+    "inquirer-fuzzy-path": "^2.3.0",
+    "lodash": "^4.14.170"
```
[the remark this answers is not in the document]
MaikoTan: In fact the lodash is already in the `devDependencies`, I just move it into the `dependencies` that can be pulled into production build. If you are talking about the size that adding lodash size, I don't think it is even an issue, because that according to the doc, webpack should be smart enough to do the `tree-shaking` to bundle the imported function itself only, not the whole bunch of lodash.
And yes, I agree that `lodash` is another things to learn, although it is widely used nowadays, so that I have only added it into the "fundamental" code, and try to keep `job components` easier to read or use.

ui/jobs/components/index.ts
```diff
@@ -1,66 +1,335 @@
...
+
+  constructor(o: ComponentInterface) {
+    this.bars = o.bars;
+    this.ee = o.emitter;
+    this.options = o.options;
+    this.partyTracker = o.partyTracker;
+    this.player = o.player;
+
+    this.shouldShows = {};
+    this.zoneId = -1;
```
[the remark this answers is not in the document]
MaikoTan: Sure, I might make mistakes in this class as it completes when I am tired. Sorry for that

ui/jobs/bars.ts
MaikoTan: Hmm, my thought is that (as I wrote in the PR description) every job component has a `shouldShows` (or as you suggested above that should be `shouldShow`) and can override the default settings in `Bars._setupJobContainers` to control those widgets showing or not, but still, I have not completed it in a properly.
If the `shouldShow` field is designed as a non-static (I forgot what is it called in English sorry!) property, it is available only when the instance of the class created(constructed?), but for now "job components" initialise many "boxes" or "bars" in their constructors, which should be created after "hp-bar"/"mp-bar" or something else.
But if `shouldShow` field is designed as a static property, I don't know how to provide user customisation, it cannot be something like `BLMComponent.showShow = { /* blahblah */ }` which confused?

ui/jobs/jobs.css
MaikoTan: Well, the PvP checking logic is included before, what I do was to change the programmatic bars hiding to using CSS to hide them, which I think is much clear. Although a potential cost is that the logic split out into js and css.

ui/jobs/event_emitter.ts
MaikoTan: Yes, I might forget to change them to the new helper type after I added it.

ui/jobs/player.ts
MaikoTan: Oops!
The code commented is the clear way to do the conversion (but a bit hard to understand), but then I should fight with typescript type checking as it lost types after the `filter` part, so I am not very sure which is better, comparing the "assigning-every-field-we-need" approach below.

ui/jobs/event_emitter.ts
MaikoTan: In fact, my fealing is to store all id in decimal would be better.
I have already trapped in several times that comparing with hex numbers in string that one is in lower cases and the other one is upper cases so cause bugs, at least all in decimal make it consistent?

ui/jobs/components/base.ts
```diff
@@ -0,0 +1,110 @@
...
+    this.player.onStatChange(this.player.job, this.onStatChange.bind(this));
+    if (this.player.job !== 'NONE') {
+      const job = this.player.job as keyof JobDetail;
+      this.player.onJobDetailUpdate(job, this.onJobDetailUpdate.bind(this));
+    }
+    this.player.onCombo(this.onCombo.bind(this));
+  }
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  onCombo(id: string | undefined, combo: ComboTracker): void {
```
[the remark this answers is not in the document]
MaikoTan: Okay, so no more `eslint-disable`

ui/jobs/components/base.ts
```diff
@@ -0,0 +1,110 @@
...
+      this.inCombat = game;
+    });
+    this.player.onYouGainEffect(this.onYouGainEffect.bind(this));
+    this.player.onYouLoseEffect(this.onYouLoseEffect.bind(this));
+    this.player.onMobGainsEffectFromYou(this.onMobGainsEffectFromYou.bind(this));
+    this.player.onMobLosesEffectFromYou(this.onMobLosesEffectFromYou.bind(this));
+    this.player.onUseAbility(this.onUseAbility.bind(this));
+    this.player.onStatChange(this.player.job, this.onStatChange.bind(this));
+    if (this.player.job !== 'NONE') {
+      const job = this.player.job as keyof JobDetail;
```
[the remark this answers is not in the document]
MaikoTan: It might be I think? But I should do more test to ensure how it should be changed.

ui/jobs/player.ts
```diff
@@ -0,0 +1,481 @@
...
+    [
+      K in keyof NetMatches['PlayerStats'] as string extends K
+        ? never
+        : number extends K
+        ? never
+        : K
+    ]: number;
+  },
+  'type' | 'timestamp' | 'job'
+>;
```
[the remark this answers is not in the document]
MaikoTan: Nit!

ui/jobs/components/index.ts
```diff
@@ -1,66 +1,335 @@
...
+    });
+  }
+
+  private _onPartyWipe(): void {
+    this.buffTracker?.clear();
+    // Reset job-specific ui
+    this.component?.reset();
+  }
+
+  private _onCraftingLog(message: string): void {
```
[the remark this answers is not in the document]
MaikoTan: Ah, I see. It looks like I make mistakes on the condition that pass to `classLists.toggle` function cause the bug.

ui/jobs/bars.ts
```diff
@@ -0,0 +1,784 @@
...
+      if (inCountdown && this.options.PlayCountdownSound) {
+        const audio = new Audio('../../resources/sounds/freesound/sonar.webm');
+        audio.volume = 0.3;
+        void audio.play();
+      }
+    }
+  }
+
+  setJobsContainerVisibility(hide?: boolean): void {
+    const container = document.getElementById('jobs-container');
```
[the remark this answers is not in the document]
MaikoTan: Nice idea. I think it even is worth writing it in the HTML directly and just calling `document.getElementById` in the constructor to store the reference for its methods.

ui/jobs/event_emitter.ts
quisquous: In the next version of the ffxiv plugin, all ids will be uppercase in all log lines (network and parsed).  So, I think it will be consistent in the future, if that helps reassure you.

It just seems like if logs lines have a hex id, then it's more straightforward to compare a hex id against the log line rather than doing a hex->decimal conversion on every line.  That's maybe premature optimization, though, but I think it also makes the code simpler (and prevents any bugs where somebody forgot to include the `, 16` part of `parseInt`).

MaikoTan: Wow, that is really good news.
maybe I can call them `id` and a `idHex`(in uppercase), and use the `idHex` for comparing for the future.

ui/jobs/player.ts
quisquous: I think assigning-every-field is ok.  It's definitely not as satisfying, but it's hard to convince typescript that you've set everything without explicitly setting everything.

ui/jobs/components/index.ts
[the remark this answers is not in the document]
MaikoTan: https://github.com/MaikoTan/cactbot/blob/a81ac0727f4aa3335a53ea3d8a916ea819f72b5f/ui/jobs/components/index.ts?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L333-L334

Looks like I wrote the wrong comment that confused me, sorry!

ui/jobs/bars.ts
```diff
@@ -0,0 +1,784 @@
...
+      this.o.pullCountdown.duration = seconds;
+      if (inCountdown && this.options.PlayCountdownSound) {
+        const audio = new Audio('../../resources/sounds/freesound/sonar.webm');
+        audio.volume = 0.3;
+        void audio.play();
+      }
+    }
+  }
+
+  setJobsContainerVisibility(hide?: boolean): void {
```
quisquous, line 776: If the parameter is hide, maybe this should be called `setJobsContainerHidden` to avoid confusion? (or maybe it should be `setJobsContainerVisible(show?: boolean)`?

ui/jobs/player.ts
MaikoTan: Yeah, you said it.
Okay, I'll remove the commented code here.

ui/jobs/bars.ts
MaikoTan: Hmm, I think to change it as `show` would be better.

ui/jobs/components/index.ts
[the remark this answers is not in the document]
quisquous: Your new code does show the cp bar when starting to craft.  However, it also shows the cp bar when switching to a crafting job.  I think maybe `_setupJobContainers` needs to initially hide the cp bar?

MaikoTan:
I'll try to fix the mana-bar issue later, as it needs more debugging things; and for the ninja/bard issue, I asked Echoring to test them on the Chinese server.

ui/jobs/bars.ts
```diff
@@ -0,0 +1,784 @@
...
+    // Update color based on the astral fire/ice state
+    let colorTag = 'mp-tick-color';
+    if (data.umbralStacks < 0)
+      colorTag = 'mp-tick-color.ice';
+    if (data.umbralStacks > 0)
+      colorTag = 'mp-tick-color.fire';
+    this.o.mpTicker.fg = computeBackgroundColorFrom(this.o.mpTicker, colorTag);
+  }
+
+  _updateMana(data: {
```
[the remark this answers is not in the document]
quisquous: I did some debugging and I see what the issue is.  The previous code was "if far / else if low / else if medium / else normal" and the current code has "if far else normal" and "if low / else if medium / else normal" and in the changed logic the two "else normals" collide with each other.  The 'battle/target' event handler overwrites the low/medium color from the 'mp' event handler.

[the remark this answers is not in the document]
MaikoTan: Ah I see!
I think it is better to extract the colour computing logic to a new function and then let it invoked by both events with the player data. But I have no idea where or how to store the `far` checking result or the `distance` between player and target, should it be a property of the class just as before?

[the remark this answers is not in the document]
quisquous: Yes, I agree.  I think it has to be a property of the class so that both events can call the new function and know about it.

quisquous:
Thanks for all the work.  Obviously I can't test the mp bar change, but it looks good to me.  I'm going to merge this now so we can move forward with this and new changes can come on top of this (rather than anybody else trying to submit changes before this is merged).

PR #3685, opened 2021-11-26
jobs: fix create func didn't receive falsy value

MaikoTan:
the function checks parameters with `if (param)`
which didn't work if we assign it as a falsy value
such as `0` or `''`(empty strings)
So change it as `if (typeof param !== 'undefined' && param !== null)`
would fix this.

### DIFF
--- a/resources/resourcebar.ts
+++ b/resources/resourcebar.ts
@@ -51,33 +51,33 @@ export default class ResourceBar extends HTMLElement {
     if (!o)
       return element;
 
-    if (typeof o.value !== 'undefined' && o.value !== null)
+    if (typeof o.value === 'string')
       element.value = o.value;
-    if (typeof o.maxvalue !== 'undefined' && o.maxvalue !== null)
+    if (typeof o.maxvalue === 'string')
       element.maxvalue = o.maxvalue;
-    if (typeof o.lefttext !== 'undefined' && o.lefttext !== null)
+    if (typeof o.lefttext === 'string')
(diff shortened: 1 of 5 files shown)


resources/resourcebar.ts
```diff
@@ -51,33 +51,33 @@ export default class ResourceBar extends HTMLElement {
     if (!o)
       return element;
 
-    if (o.value)
+    if (typeof o.value !== 'undefined' && o.value !== null)
```
quisquous, line 54: Should these all be `typeof o.value === 'string'`?

quisquous (commented):
Whoops.  Thank you for catching this.

resources/resourcebar.ts
MaikoTan: I think the null check is still needed as the typescript type check only ensure the type-safe when you are developing, if there is a js file that invoke this function, an explicit null-check in runtime would be better so we won't just pass null to functions that might cause potential un-expected bihaviors.

quisquous: I think if your goal is to be defensive, then `if (typeof o.value === 'string')` style might be better, because it would prevent other types as well as null?

MaikoTan: Sure.
I just changed these code via a regex matching without changing them one by one particularly as they are not all strings but numbers or booleans.
I'll change them manually later.

MaikoTan: Done.

PR #3691, opened 2021-11-26
resources: add html lang attribute

MaikoTan:
change some `.lang-xx` selector to use `:lang()` pseudo-selector

Chrome seems to apply some optimisation based on the `lang` attributes, the text would show prettier when using the correct value, checked in CJK (Chinese/Japanese/Korean) characters.
This makes texts still prettier even deleting the font mentioned in #3684 .


resources/defaults.css
```diff
@@ -21,7 +21,7 @@ body {
   will-change: content;
 }
 
-.lang-ja .text {
+:lang(ja) .text {
```
quisquous, line 24: Should this be `ja-JP`?

util/coverage/coverage.js
```diff
@@ -369,7 +369,7 @@ document.addEventListener('DOMContentLoaded', () => {
   // Allow for `coverage.html?lang=de` style constructions.
   const params = new URLSearchParams(window.location.search);
   const lang = params.get('lang') ? params.get('lang') : 'en';
-  document.body.classList.add(`lang-${lang}`);
+  document.documentElement.lang = lang === 'cn' ? 'zh-CN' : lang;
```
quisquous, line 372: What about a shared function that this and `UserConfig` can call to get the right mapping from cactbot lang to locale, possibly in `resources/languages.ts`?

I think you missed a few, as well:
```
$ git grep lang- -- *.css
resources/defaults.css:.lang-ja .text {
resources/defaults.css:.lang-cn .text {
ui/raidboss/raidboss.css:#timeline-container.lang-fr {
ui/raidboss/raidboss.css:#timeline-container.lang-de {
ui/raidboss/raidboss.css:#timeline-container.lang-ko {
ui/raidboss/raidboss.css:.lang-ko .timer-bar {
```

resources/defaults.css
MaikoTan: As[MDN](https://developer.mozilla.org/ja/docs/Web/HTML/Global_attributes/lang) says, `:lang(ja)` matches all valid language code that prefixed by `ja`, so it is okay here.

As for Chinese, Simplified Chinese and Traditional Chinese use different characters set and different words, so it is nessessary to differentiate `zh-CN` and `zh-TW` or something else.

util/coverage/coverage.js
MaikoTan: Oh, sure!
I'll take a look later.

quisquous: Also thinking about this,l this is a breaking change if users are depending on `lang-cn` existing as a style.  I think we could either keep the old CSS as well or alternatively I could write a release note for this.

MaikoTan: True.
Maybe I can say something like
```css
:lang(zh-CN) .text,
/* old one */
.lang-cn .text {
  /* blahblah */
}
```

resources/user_config.ts
```diff
@@ -264,7 +264,9 @@ class UserConfig {
       else
         options.DisplayLanguage = options.ParserLanguage || 'en';
 
-      document.body.classList.add(`lang-${options.DisplayLanguage}`);
+      // TODO: remove this later?
```
quisquous, line 267:
```suggestion
      // TODO: left for now as backwards compatibility with user css.  Remove this later?
```

resources/defaults.css
```diff
@@ -21,6 +21,7 @@ body {
   will-change: content;
 }
 
+:lang(ja) .text,
 .lang-ja .text {
```
quisquous, line 25:
```suggestion
:lang(ja) .text {
```

Sorry for any confusing comment.  I think internal to cactbot we should standardize on one format.  I think `lang(ja)` makes sense to me.  Can cactbot CSS just use `lang(ja)` style?

My earlier comment was me saying that we still needed to keep adding the `lang-ja` class, which you also did.  However, I don't think we need to reference it in cactbot itself.

MaikoTan: Yeah, I think I misunderstand your thought, thanks!

PR #3673, opened 2021-11-24
raidboss: Automated alpha conversion of ARR timelines

JLGarber:
Overall I'm fairly happy with this. I'll need to update the script to strip the trailing dot from `addedCombatant` lines, but apart from that it looks relatively clean on a 15-second inspection.

The following errors were encountered and will be fixed manually: 

```
t7: [
    '300.0 "--sync--" sync /:Melusine HP at 79%/ window 300,0',
    '800.0 "--sync--" sync /:Melusine HP at 59%/ window 800,0'
  ],
  t9: [
    '503.3 "Golem Meteors" sync /1B:........:[^:]*:....:....:0007:/ duration 11
window 505,0',
    '582.3 "Meteor 1/6" duration 7 sync /1B:........:[^:]*:....:....:000[9A]:/ w
indow 83,0',
    '598.3 "Golem Meteors" sync /1B:........:[^:]*:....:....:0007:/ duration 11
window 80,80'
  ],
  cape_westwind: [
    '199.0 "--sync--" sync /00:0044:[^:]*:My shields are impregnable/ window 200
,0',
    '584.3 "--sync--" sync /00:0044:[^:]*:Your defeat will bring/ window 600,0'
  ],
  'ifrit-nm': [
    '200.0 "--sync--" sync /00:0044:Ifrit:Succumb to the inferno/ window 200,0',
    '300.0 "--sync--" sync /00:0044:Ifrit:Surrender thyself to the fires of judg
ment/ window 300,0'
  ],
  'titan-nm': [ '703.6 "Landslide" sync /Titan:285:/' ]
```

PR #3694, opened 2021-11-27
test: Temporarily permit old/new seal formats in timelines

JLGarber:
Per <https://github.com/quisquous/cactbot/pull/3673#issuecomment-980494383>, this should permit passing tests for all timeline seal/unseal entries that use the new log line format. Note the error message doesn't include the old format, since we shouldn't be using it for any new work.

### DIFF
--- a/test/helper/test_timeline.js
+++ b/test/helper/test_timeline.js
@@ -309,7 +309,7 @@ const testTimelineFiles = (timelineFiles) => {
             // we should come back here and make the doubled colon non-optional.
             if (regex.includes('is no longer sealed')) {
               assert(
-                /00:0839::?\.\*is no longer sealed/.exec(regex),
+                /00:0839::?.*is no longer sealed/.exec(regex),
                 `${timelineFile}:${sync.lineNumber} 'is no longer sealed' sync must be exactly '00:0839::.*is no longer sealed'`,
               );
             } else if (regex.includes('will be sealed')) {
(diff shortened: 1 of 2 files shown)


test/helper/test_timeline.js
```diff
@@ -303,15 +303,19 @@ const testTimelineFiles = (timelineFiles) => {
...
             const regex = sync.regex.source;
+            // FIXME: This test will currently accept log lines with or without a second colon,
+            // as "0839:" or "0839::".
+            // Once we have completely converted things for 6.0,
+            // we should come back here and make the doubled colon non-optional.
             if (regex.includes('is no longer sealed')) {
               assert(
-                regex.includes('00:0839:.*is no longer sealed'),
-                `${timelineFile}:${sync.lineNumber} 'is no longer sealed' sync must be exactly '00:0839:.*is no longer sealed'`,
+                /00:0839::?\.\*is no longer sealed/.exec(regex),
```
quisquous, line 312:
```suggestion
                /00:0839::?.*is no longer sealed/.exec(regex),
```

JLGarber: Oh, that's fair, I guess there might be edge cases where we do want to permit specific unseal lines, as opposed to the One True Normalized one.

quisquous: Oh oh oh, you're right.  I think it should be like it was before I made a mistaken edit.  Sorry, I shouldn't send code review comments when i'm tired.

JLGarber: It is a *good* thing for all of us that tomorrow is Saturday and not Monday!

PR #3697, opened 2021-11-27
util: Alpha Python timeline utility updates

JLGarber:
It's all bad, but it seemed to work on the alpha logs I was given. I'm not particularly attached to anything I've implemented here, this was mostly an attempt to get *something* in place for later improvement.

For consistency with converted old logs, I figured it made most sense to format `make_timeline` output the same way, as `1[56]`.

### DIFF
--- a/util/encounter_tools.py
+++ b/util/encounter_tools.py
@@ -263,7 +263,7 @@ def is_tl_line_begincast(poss_match):
 
 def is_tl_line_buff(poss_match):
     return re.search(
-        r"\s1A:(?P<effectId>(?:\.\{0,4\}|.{0,4})):(?P<effect>(?:[^:]*?|\[\^:\]\*?)):\[\^:\](?:\+|\*):\.\{8\}:(?P<source>(?:[^:]*?)):\.\{8\}:(?P<target>(?:[^:]*))",
+        r"\s1A:(?P<effectId>(?:[^:]*?|\[\^:\]\*?)):(?P<effect>(?:[^:]*?|\[\^:\]\*?)):(?P<duration>(?:\[\^:\]\*|[^:]*)):(?P<sourceId>(?:\[\^:\]\*|[^:]*)):(?P<source>(?:\[\^:\]\*|[^:]*)):(?P<targetId>(?:\[\^:\]\*|[^:]*)):(?P<target>(?:\[\^:\]\*|[^:]*)):?",
         poss_match,
     )
 
--- a/util/encounter_tools.py
+++ b/util/encounter_tools.py
@@ -256,23 +256,34 @@ def is_tl_line_syncmatch(line):
 
 
 def is_tl_line_begincast(poss_match):
-    return re.search(r"^:([0-9A-F\[\]\(\)\|]+):([^:]+)", poss_match)
+    return re.search(r"\s14:(?P<source>[^:]+):(?P<id>[0-9A-F\(\)\|]+)", poss_match)
 
 
 def is_tl_line_buff(poss_match):
-    return re.search(r"1A:.......:(.+) gains the effect of (.+)( from)?", poss_match)
+    return re.search(
+        r"\s1A:(?P<effectId>(?:\.\{0,4\}|.{0,4})):(?P<effect>(?:[^:]*?|\[\^:\]\*?)):\[\^:\](?:\+|\*):\.\{8\}:(?P<source>(?:[^:]*?)):\.\{8\}:(?P<target>(?:[^:]*))",
+        poss_match,
+    )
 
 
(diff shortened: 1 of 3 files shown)


util/encounter_tools.py
```diff
@@ -256,23 +256,34 @@ def is_tl_line_syncmatch(line):
...
 def is_tl_line_log(poss_match):
-    return re.search(r"^\s*00:([0-9]*):(.*$)", poss_match)
+    return re.search(
+        r"\s00:(?P<id>[0-9]*):(?P<entity>[^:]*):(?P<message>[^\/]*)\s?\/\s", poss_match
+    )
 
 
 def is_tl_line_adds(poss_match):
-    return re.search(r"Added new combatant (.*$)", poss_match)
+    return re.search(r"\s*03:........:(?P<entity>[^:]*):", poss_match)
```
quisquous, line 280: Why do some of these lines have `\s*` and others just `\s`? I think it should just be a single space in all cases, right?

util/make_timeline.py
```diff
@@ -371,7 +371,7 @@ def main(args):
         if entry["line_type"] == "34":
             output_entry = '{position:.1f} "{targetable}"'.format(**entry)
```
quisquous, line 372: Are these targetable lines still ok?

util/test_timeline.py
```diff
@@ -39,38 +39,69 @@ def load_timeline(timeline):
...
             entry["special_type"] = False
 
             begincast_match = e_tools.is_tl_line_begincast(sync_match.group(1))
             if begincast_match:
                 entry["special_type"] = "begincast"
                 entry["special_line"] = "20"
-                entry["cast_id"] = begincast_match.group(1)
-                entry["caster_name"] = begincast_match.group(2)
+                entry["cast_id"] = begincast_match.group("id")
+                entry["caster_name"] = begincast_match.group("source")
```
quisquous, line 57: Did you verify that all of these log types: cast, begincast, buff, log, adds, headmarker all match properly? Mostly just looking for "have each of these conditional blocks been run at least once and seem to work" level of testing.

util/make_timeline.py
JLGarber: These don't include syncs ever, it just writes it in as `1000.0 "--targetable--"`. (When I get to the Typescript re-write, I may add the ability to sync these optionally, but it's not worth trying to add that to the Python version here.)

util/encounter_tools.py
JLGarber: It should indeed.

util/test_timeline.py
```diff
@@ -236,11 +262,7 @@ def test_match(event, entry):
         elif entry["special_type"] == "battlelog" and event.startswith(entry["special_line"]):
             # Matching this format generically:
             # 00|2019-01-12T18:08:14.0000000-05:00|0839||The Realm of the Machinists will be sealed off in 15 seconds!|
-            log_match = re.search(
-                "^00\|[^\|]*\|{}\|[^\|]*\|{}".format(entry["logid"], entry["line"]),
-                event,
-                re.IGNORECASE,
-            )
+            log_match = re.search(entry["regex"], event, re.IGNORECASE,)
```
JLGarber, line 265: *This* little annoyance right here was what made `gainsEffect` syncs perpetually fail. We can't rely on the casing of the `effectId` in the timeline file to match that in the network log. ~~It took me "only" half an hour of "WHY WON'T THIS WORK? IT MATCHES!" to debug~~

util/test_timeline.py
JLGarber: Verified now. SO much nonsense.

util/test_timeline.py
```diff
@@ -224,9 +252,7 @@ def test_match(event, entry):
         elif entry["special_type"] == "applydebuff" and event.startswith(entry["special_line"]):
             # Matching this format generically:
             # |Dadaluma Simulation|0.00|E0000000||4000AE96|Guardian
-            buff_match = re.search(
-                "\|{}\|([^\|]*\|){{4}}{}".format(entry["buff_name"], entry["buff_target"]), event
-            )
+            buff_match = re.search(entry["regex"], event, re.IGNORECASE)
```
JLGarber, line 255: *This* little annoyance right here was what made `gainsEffect` syncs perpetually fail. We can't rely on the casing of the `effectId` in the timeline file to match that in the network log. ~~It took me "only" half an hour of "WHY WON'T THIS WORK? IT MATCHES!" to debug~~

quisquous: Oh nooo.

If it makes you feel any better, everything is uppercase in the alpha plugin (but we probably need to keep this indefinitely to support old network log files).

PR #3704, opened 2021-11-29
build: fix webpack freeze in ci

Trim21:
I think this is a issue cased by `fork-ts-checker-webpack-plugin`.

We have another workflow running tsc on our code, remove it from webpack.


MaikoTan:
我之前也這麼做過呢，但是quis說[這個程式「在開發時非常有用」](https://github.com/quisquous/cactbot/pull/3613#issuecomment-961587648)，所以我又加回去了。
我覺得至少在開發階段，我們可以保留這個外掛，比如說僅將其置入`webpack.dev.ts`。
I did this before, but quis said that [this plugin is useful in developing](https://github.com/quisquous/cactbot/pull/3613#issuecomment-961587648), so I added it back finally.
I think at least in developing, you can keep this plugin remaining in the repo, for example, add it in `webpack.dev.ts`.

Trim21:
it's also very slow. It need 25s to do type checking but only 5s to compile web asserts.

maybe we can use this: https://www.npmjs.com/package/is-ci

quisquous:
Alternatively, if it's the ts checker plugin, then maybe we do need to `skipLibCheck: true` here?

That said, if the plugin isn't useful to anybody else, we can remove it and if I feel like I'm really missing it, I can add it back.

Trim21:
I think the value of `skipLibCheck`  is not a relative here.

MaikoTan:
我覺得引入is-ci是個不錯的方式，畢竟我們在CI裏已經有相應的TypeScript類型檢查了，應該可以直接去掉這個步驟。
I think it is good to import `is-ci` for this, as that we have already TypeScript type-checking in CI, so this step can be skipped.

webpack/webpack.config.ts
```diff
@@ -51,6 +53,33 @@ export default (
...
+          // only raidboss for now though.
+          from: 'ui/*/skins/**/*',
+          noErrorOnMissing: true,
+        },
+      ],
+    }),
+  ];
+
+  if (!isCI)
+    plugins.push(new ForkTsCheckerWebpackPlugin());
```
MaikoTan, line 81: 不如用`unshift`，把這外挂放最前面，反正它是一個新綫程，越早啓動越好。
I think `unshift` is better (to put the plugin on top), as it is running in a dedicated thread, it is better to invoke it as earlier as possible.

Trim21: I guess it realy starts on a webpack event, not creating insntance.

quisquous:
Is this ready to merge?

Trim21:
Yes

quisquous:
Thanks for debugging this and figuring out a fix!

PR #3711, opened 2021-12-01
raidboss: remaining alpha timeline conversions

JLGarber:
No issues fixing these up. I didn't touch the `21` lines in the Shadowbringers Eureka files since they looked to be correctly formatted already.

I *think* the line ID for "Enemy Gains Buff" is`2AAE`, but I figured it was safer to just make the Gubal Hard lines universal, since that's basically the only place this happens.

### DIFF
--- a/ui/raidboss/data/04-sb/eureka/eureka_hydatos.txt
+++ b/ui/raidboss/data/04-sb/eureka/eureka_hydatos.txt
@@ -6,7 +6,7 @@ hideall "--Reset--"
 hideall "--sync--"
 
 0.0 "--Reset--" sync / 00:0839::.*is no longer sealed/ window 100000 jump 0
-0 "--Reset--" sync /04:........:Removing combatant Proto Ozma\.  Max HP: \d{8}\./ window 10000 jump 0
+0.0 "--Reset--" sync / 04:........:Proto Ozma:/ window 100000 jump 0
 
 
 #####################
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/04-sb/eureka/eureka_hydatos.txt
```diff
@@ -16,7 +16,7 @@ hideall "--sync--"
 
 # Note: This checks that Art's auto hits a target close enough for you to see.
 # You can see Owain's auto abilities, but the target is listed as empty.
-1002.5 "--sync--" sync / 1[56]:[^:]:Art:3956:/ window 1500,0
+1002.5 "--sync--" sync / 1[56]:[^:]*:Art:3956:[^:]*:[^:]*:/ window 1500,0
```
quisquous, line 19: Mmm sorry, but I think that's not it either.  (And you also need to handle the Owain sync below as well.)

The format is `[timestamp] ActionEffect 15:[sourceId]:[source]:[id]:[ability]:[targetId]:[target]:`

So, I think you'd need: `1002.5 "--sync--" sync / 1[56]:[^:]*:Art:3956:[^:]*:[^:]+:[^:]+:/ window 1500,0`

JLGarber: That's fine, I'm just starting some scheduled  in-game stuff so I'll be back at this a little later tonight.

PR #3715, opened 2021-12-02
raidboss: uwu ifrit nails

chalkpe:
alert safe zone for woken ifrit dashes, based on reverse-Z order strategy for now
maybe needs to adjust which is safe by the order nail was killed?


quisquous:
Thanks for this!  Yeah, ideally I think this nail trigger should track nail deaths and not nail added combatants so that it can handle reverse-Z and forward-Z (or accidents, regardless of strategy).

If you wanted to make this even fancier, I think it could say:
* which intercardinal to go to for the first set of dashes
* which way to rotate for the dashes
* how far to rotate, depending on where awoken ifrit is

(but not required! I think having a dash safe zone that knows about nail death order would be plenty if that's all you wanted to do)

PR #3719, opened 2021-12-02
resources/jobs: add new EW job enums

quisquous:
I don't know what the actual enum values are, but I just wanted to see what was needed typescript-wise to add a new job.


MaikoTan:
Confirm that RPR is 39 and Sage is 40

PR #3727, opened 2021-12-03
i18n: add EW text localisation

MaikoTan:


util/coverage/coverage.ts
```diff
@@ -94,6 +97,9 @@ const exVersionToShortName: { [exVersion: string]: LocaleText } = {
   },
   '4': {
     en: 'EW',
+    de: 'EW',
+    fr: 'EW',
+    ja: '暁月',
```
MaikoTan, line 102:
```suggestion
    ja: '暁月',
    cn: '6.X',
```

Folks in Chinese would still using patch number for the abbreviaion, so this should be okay at this point,

PR #3733, opened 2021-12-03
raidboss: Add some basic Tower of Zot triggers

valarnin:
Haven't tested, WIP, etc etc. Just putting it out there to make sure directory structure lines up with what's expected.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
+++ b/ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
@@ -9,11 +9,17 @@ export type Data = RaidbossData;
 const triggerSet: TriggerSet<Data> = {
   zoneId: 0x3B8, // @TODO: change to zoneid entry when available
   triggers: [
-    // Tankbuster shared by 2nd and 3rd bosses
     {
-      id: 'ToZ Isitva Siddhi',
+      id: 'ToZ Cinduruva Isitva Siddhi',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: '62A9', source: ['Sanduruva', 'Cinduruva'], capture: true }),
+      netRegex: NetRegexes.startsUsing({ id: '62A9', source: 'Cinduruva', capture: true }),
+      condition: Conditions.caresAboutPhysical(),
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'ToZ Sanduruva Isitva Siddhi',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '62C0', source: 'Sanduruva', capture: true }),
       condition: Conditions.caresAboutPhysical(),
       response: Responses.tankBuster(),
     },
--- a/ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
+++ b/ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
@@ -13,7 +13,7 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'ToZ Isitva Siddhi',
       type: 'StartsUsing',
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
```diff
@@ -0,0 +1,31 @@
...
+export type Data = RaidbossData;
+
+const triggerSet: TriggerSet<Data> = {
+  zoneId: 0x3B8, // @TODO: change to zoneid entry when available
+  triggers: [
+    // Tankbuster shared by 2nd and 3rd bosses
+    {
+      id: 'Isitva Siddhi',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '62A9', source: '(?:Sanduruva|Minduruva)', capture: true }),
```
JLGarber, line 16:
```suggestion
      netRegex: NetRegexes.startsUsing({ id: '62A9', source: ['Sanduruva, 'Minduruva'], capture: true }),
```

Not that it matters *that* much, but it's probably better not to use raw regex syntax within regex helpers.

valarnin:
Sorry about the force push, was still sorting out some issues with the new windows install and that caused all sorts of weird issues on my end with the git tree.

This is technically functional now, validated in-zone and via raidemulator. I'll mark it as Ready for review, feel free to merge or to wait until there's other triggers available.

JLGarber:
Oh, sorry, I should have mentioned it previously. Tank busters no longer take conditions, they're now intended for all users. (Individuals can turn them off as they like.)

ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
```diff
@@ -0,0 +1,37 @@
+import Conditions from '../../../../../resources/conditions';
+import NetRegexes from '../../../../../resources/netregexes';
+import { Responses } from '../../../../../resources/responses';
+import { RaidbossData } from '../../../../../types/data';
+import { TriggerSet } from '../../../../../types/trigger';
+
+export type Data = RaidbossData;
+
+const triggerSet: TriggerSet<Data> = {
+  zoneId: 0x3B8, // @TODO: change to zoneid entry when available
```
quisquous, line 10: This should already exist!

valarnin: My local tree was pre-EW-update when I wrote these originally.

ui/raidboss/data/06-ew/dungeon/the_tower_of_zot.ts
```diff
@@ -0,0 +1,37 @@
...
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '62C0', source: 'Sanduruva', capture: true }),
+      condition: Conditions.caresAboutPhysical(),
+      response: Responses.tankBuster(),
+    },
+    // Trash mob tankbuster
+    {
+      id: 'ToZ Wheel',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '5E52', source: 'Zot Roader', capture: true }),
```
quisquous, line 28: It looks like there's no startsUsing for this attack, it's just an uncasted ability.

valarnin: Hmm, I guess it'll need to be dropped from triggers then. I wanted to include it due to how hard it hits unmitigated but with no startsCasting I'm not sure if there's a feasible way to warn for it.

JLGarber: I'll try to generate timelines for this tonight and we'll see whether we can do a timeline trigger instead.

JLGarber: That isn’t a boss, and I am sad now. T_T

PR #3736, opened 2021-12-04
plugin: add sage job detail

MaikoTan:
I am not very sure if it is correct because I don't know C# and memory things, but it works well on my side.

### DIFF
--- a/plugin/CactbotEventSource/FFXIVProcessIntl.cs
+++ b/plugin/CactbotEventSource/FFXIVProcessIntl.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Linq;
 using System.Collections.Generic;
 using System.Runtime.InteropServices;
@@ -704,7 +704,7 @@ namespace Cactbot {
     [StructLayout(LayoutKind.Explicit)]
     public struct SageJobMemory {
       [FieldOffset(0x00)]
-      public ushort addersgallGauge_bits; // the addersgall gauge, from 0 to 19999.
+      public ushort addersgallGauge; // the addersgall gauge, from 0 to 19999.
 
       [FieldOffset(0x02)]
       public byte addersgall;
@@ -714,13 +714,6 @@ namespace Cactbot {
 
       [FieldOffset(0x04)]
       private byte eukrasia_bits;
-
-      public int addersgallGauge {
-        get {
-          // translate the bits to a value from 0 to 100.
-          return (int)((float)addersgallGauge_bits / 20000 * 100);
-        }
-      }
 
       public bool eukrasia {
         get {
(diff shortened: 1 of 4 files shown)


plugin/CactbotEventSource/FFXIVProcessIntl.cs
```diff
@@ -698,5 +700,33 @@ public struct SamuraiJobMemory {
...
+      [FieldOffset(0x03)]
+      public byte addersting;
+
+      [FieldOffset(0x04)]
+      private byte eukrasia_bits;
+
+      public int addersgallGauge {
+        get {
+          // translate the bits to a value from 0 to 100.
+          return (int)((float)addersgallGauge_bits / 20000 * 100);
```
quisquous, line 721: Do you think it's worth sending the whole value, rather than truncating here?

MaikoTan: you mean the original value from memory?

MaikoTan: well, it can be also converted in js side, I don't care where it is. sure

quisquous: Yeah, that sounds good!

plugin/CactbotEventSource/FFXIVProcessIntl.cs
```diff
@@ -698,5 +700,26 @@ public struct SamuraiJobMemory {
...
+      public ushort addersgallGauge; // the addersgall gauge, from 0 to 19999.
+
+      [FieldOffset(0x02)]
+      public byte addersgall;
+
+      [FieldOffset(0x03)]
+      public byte addersting;
+
+      [FieldOffset(0x04)]
+      private byte eukrasia_bits;
```
quisquous, line 716: Are the other bits used for anything, or can you pass the full byte? If you need to do the mask, can you mark this field as `[NonSerialized]`?

MaikoTan: I did only found that when you have an eukrasia, it is 1, otherwise 0.
I don't really know C# and even don't know what it is, do you mean like this:
```suggestion
      [NonSerialized]
      [FieldOffset(0x04)]
      private byte eukrasia_bits;
```

plugin/CactbotEventSource/FFXIVProcessIntl.cs
```diff
@@ -698,5 +700,26 @@ public struct SamuraiJobMemory {
...
+
+      public bool eukrasia {
+        get {
+          return (eukrasia_bits & 0x1) != 0;
+        }
+      }
```
quisquous, line 722:
```suggestion
      [FieldOffset(0x04)]
      private byte eukrasia;
```

If it's always 1 or 0, then I would suggest doing the above.  That way there's a single `eukrasia` field that we send.  No need to do the `0x1` mask part.

PR #3738, opened 2021-12-04
resources: update endwalker effect ids

MaikoTan:
I didn't get gen_effect_id.ts to work, but I commented the ESLint part and let it write the text that is not linted to file and then lint it manually.

Oops, this patch break Samurai and Dancer, as there are changes


quisquous:
image

In the future could you try to make each PR pass CI? It's very hard for me to merge things when there are multiple patches that don't pass.  Ways you can do this differently in the future are:

* if two patches must be submitted together (dancer/samurai/effects), then put them in the same PR (feel free to separate them into different commits)
* if one patch depends on another patch but not vice versa, then you can change github to say you are merging into the other PR (vs merging into quisquous:main)

I will pull these and make sure they are ok locally and merge them, if so.

MaikoTan:
Ah, you are right. I'll do that in the future!

PR #3742, opened 2021-12-04
plugin: renew 6.0 summoner job detail

MaikoTan:
Still no idea what is 0x04-0x05 is.
But seems like that we need is already here?

### DIFF
--- a/ui/jobs/components/smn.ts
+++ b/ui/jobs/components/smn.ts
@@ -1,7 +1,7 @@
 import TimerBox from '../../../resources/timerbox';
 import { JobDetail } from '../../../types/event';
 import { ResourceBox } from '../bars';
-import { kAbility } from '../constants';
+import { kAbility, patch5xEffectId } from '../constants';
 import { PartialFieldMatches } from '../event_emitter';
 import { computeBackgroundColorFrom } from '../utils';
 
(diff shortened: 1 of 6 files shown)


MaikoTan:
Summoner are basically renewed fully, how can I get them still work on CN/KR but also keep TypeScript happy with new parameters in job details? @quisquous

quisquous:
> Summoner are basically renewed fully, how can I get them still work on CN/KR but also keep TypeScript happy with new parameters in job details? @quisquous

Maybe a discriminated union, where you have both sets of fields { oldFields } | { newFields } but only one set of them is true?

MaikoTan:
Added a blank SMNComponent for now, maybe Echoring or me would do this later.

ui/jobs/components/smn.ts
```diff
@@ -74,13 +79,13 @@ export class SMNComponent extends BaseComponent {
   }
 
   override onYouGainEffect(id: string, matches: PartialFieldMatches<'GainsEffect'>): void {
-    if (id === EffectId.FurtherRuin) {
+    if (id === '4BC') { // Further Ruin changed id in 6.0, EffectId.FurtherRuin is the 6.0 version.
```
quisquous, line 82: Could you add `FurtherRuin5x: 4BC` into ui/jobs/constants?

MaikoTan: Done!

PR #3759, opened 2021-12-07
raidemulator: Endwalker fixes

valarnin:
WIP PR for EW fixes.

Current fixes:
- Fix wrong ID reference in LineEvent0x17.ts
- Fix missing job icons for SGE/RPR

Known issues pending fix:
- Missing Reaper/Sage abilities for job detection


valarnin:
I'm not sure if there's a dump of RPR/SGE ability IDs and effect IDs available currently to reference for the `CombatantJobSearch.ts` file?

quisquous:
> I'm not sure if there's a dump of RPR/SGE ability IDs and effect IDs available currently to reference for the `CombatantJobSearch.ts` file?

Offhand, I don't know of any other than looking through action/status tables.

Can we get the jobs from the added combatant line instead?

valarnin:
Jobs are assigned based on this code in `CombatantTracker.ts`:

```ts
    if (isLineEventJobLevel(line)) {
      combatant.job = this.combatants[line.id]?.job ?? line.job;
      combatant.level = this.combatants[line.id]?.level ?? line.level;
    }

    if (isLineEventAbility(line)) {
      if (!combatant.job && !line.id.startsWith('4') && line.abilityId !== undefined)
        combatant.job = CombatantJobSearch.getJob(line.abilityId);
    }
```

This covers all cases where we have an entire pull, but will not cover e.g.
1. Zone in
2. Start ACT
3. Pull

Unless that's changed at some point since emulator was first written and on ACT start it always prints all combatants.

quisquous:
> This covers all cases where we have an entire pull, but will not cover e.g.
> 
> 1. Zone in
> 2. Start ACT
> 3. Pull
> 
> Unless that's changed at some point since emulator was first written and on ACT start it always prints all combatants.

No, that has not changed.  However, I could argue that it's an edge case that we don't have to support, either?

valarnin:
I'd argue that we don't need to support it, but it should probably be handled more gracefully than what happens now, e.g. player detection and job detection is visibly broken but otherwise it continues to run?

quisquous:
> I'd argue that we don't need to support it, but it should probably be handled more gracefully than what happens now, e.g. player detection and job detection is visibly broken but otherwise it continues to run?

Maybe those encounters are just dropped when importing? (With a console warning for debugging?)

PR #3760, opened 2021-12-07
resources: run gen tools for EW

Akurosia:


quisquous:
Can you revert the change to zone_info.ts?

PR #3764, opened 2021-12-08
Eureka: Add a BA Portal Timer to Hydatos

KattTails:
This adds a new timer to the Hydatos map for tracking the blue/red BA portals in the same style as the weather info 

Preview image:
image
image

I've also implemented a new check for nm.startTimerOnKill, which allows NMs/CEs such as Ovni, Castrum and the Dalriada to be tracked more accurately as a result.

If you want to test it quickly you can use this url: `https://katttails.github.io/cactbot/ui/eureka/eureka.html`
however do note that I've been doing some other work on top of this PR so there are some other unrelated code changes as well, but they shouldn't affect anything (I hope).

### DIFF
--- a/ui/eureka/eureka.css
+++ b/ui/eureka/eureka.css
@@ -76,6 +76,7 @@
 }
 
 #label-time > div,
+#label-portal-timer > div,
 #label-weather-container > div > div {
   display: inline-block;
 }
@@ -251,6 +252,15 @@
   border-bottom: 2px dashed rgba(255, 255, 255, 0.3);
 }
 
+#portal-timer-container {
+  position: absolute;
+  right: 0;
+  font-size: 16px;
+  font-weight: normal;
+  padding: 5px;
+  border: 2px dashed rgba(255, 255, 255, 0.3);
+}
+
 .aspect-ratio-anemos #label-container {
   top: 0%;
 }
@@ -264,6 +274,10 @@
 }
 
 .aspect-ratio-hydatos #label-container {
+  top: 90%;
+}
+
+.aspect-ratio-hydatos #portal-timer-container {
   top: 90%;
 }
 
--- a/ui/eureka/eureka.html
+++ b/ui/eureka/eureka.html
@@ -1,4 +1,4 @@
-﻿<html lang="en-US">
+﻿<html>
 <head>
   <meta charset="utf-8" />
   <title>Cactbot Eureka</title>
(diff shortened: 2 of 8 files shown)


KattTails:
cactbot bot doesn't like capitals :(

ui/eureka/eureka_config.ts
```diff
@@ -193,5 +193,13 @@ UserConfig.registerOptions('eureka', {
         options['RefreshRateMs'] = seconds * 1000;
       },
     },
+    {
+      id: 'showBAPortalTimer',
+      name: {
+        en: 'Show Timer for BA red/blue Portals',
```
quisquous, line 199: I think it'd be fine to just always show portal timers.  I don't think this needs an option.

ui/eureka/eureka.css
```diff
@@ -251,6 +252,15 @@
   border-bottom: 2px dashed rgba(255, 255, 255, 0.3);
 }
 
+#portal-timer-container {
```
quisquous, line 255: What do you think about combining these with weather (like on top of the current weather?) just to avoid having too many boxes spread about?

KattTails: I decided on creating a separate box for it because the weather container is already pretty tall compared to the Hydatos map  and I wanted to keep it as compact as possible.
The layout of the weather box is also pretty fixed, there's always the time of day and then 5 weathers visible, whereas the portal timer is only going to be popping up when it's relevant to be shown. I think that having the box resize and an extra element appearing/disappearing would be pretty distracting personally, but I'm open to others' thoughts on this.

ui/eureka/eureka_config.ts
KattTails: Fair enough! I was 50/50 about it in the first place and only added it at the last minute.

ui/eureka/eureka.css
quisquous: Ok, that seems reasonable!

ui/eureka/eureka.html
```diff
@@ -1,4 +1,4 @@
-﻿<html lang="en-US">
+﻿<html>
```
quisquous, line 1:
```suggestion
﻿<html lang="en-US">
```

ui/eureka/eureka_translations.ts
```diff
@@ -11,6 +11,18 @@ export const bunnyLabel: LocaleText = {
   ko: '토끼',
 };
 
+const sec2time = (timeInMs: number) => {
```
quisquous, line 14: is this `ms2time` and not `sec2time`? or, maybe I misunderstand what "sec" means here?

PR #3766, opened 2021-12-08
eureka: Add an option to only import new NMs from tracker

KattTails:
Just skips over entries from the tracker if we already have a pop time in the overlay, because it's probably more accurate.

Needs translations.

### DIFF
--- a/ui/eureka/eureka.ts
+++ b/ui/eureka/eureka.ts
@@ -104,6 +104,7 @@ const defaultEurekaConfigOptions = {
   FlagTimeoutMs: 90,
   CompleteNamesSTQ: false,
   EnrichedSTQ: false,
+  TrackerOnlyImportMissing: false,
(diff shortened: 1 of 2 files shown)


quisquous (commented):
I think I'm a little confused by this.  There's already `respawnTimeMsLocal` vs `respawnTimeMsTracker` variables.  And then on line ~650, local takes precedence over tracker always.  (I think that's how it should be.)

```typescript
      if (nm.respawnTimeMsLocal)
        respawnMs = nm.respawnTimeMsLocal;
      else if (nm.respawnTimeMsTracker)
        respawnMs = nm.respawnTimeMsTracker;
```

KattTails:
Oops. This PR is totally pointless then sorry! I didn't notice that section in the code and just assumed it'd be used everywhere.

PR #3767, opened 2021-12-08
eureka: Bozja map fixes

KattTails:
A friend pointed out to me that the bozja maps didn't lock the aspect ratio, so could be stretched out depending on the overlay window. Fixing that lead me to realise the weather info panel had been hidden all this time.. Oops!

This PR just adds some aspect ratio correction, adjusts the weather info box to fit neatly and adds the missing weathers for Bozja.

### DIFF
--- a/ui/eureka/eureka.css
+++ b/ui/eureka/eureka.css
@@ -288,7 +288,9 @@
 }
 
 .aspect-ratio-zadnor #label-container {
-  top: 90%;
+  top: 20%;
+  left: unset !important;
+  right: 0;
 }
 
 .aspect-ratio-hydatos #portal-timer-container {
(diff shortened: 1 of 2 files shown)


quisquous:
Do you think we should hide the weather in Bozja? I don't think there's anything that's affected by it, right? I had it for Eureka because there were some NMs where it mattered.

Akurosia:
aren't sprite related to weather even in bozja?

KattTails:
> aren't sprite related to weather even in bozja?

Yeah, the sprites are weather based and a great source for some of the fragments via reflect farming.

I think it should be shown, personally.

quisquous:
Ohhhhhh, ok ok! I do all my "farming" from the market board, so I didn't realize that mattered.

PR #3769, opened 2021-12-08
plugin: update RDM/BRD job details

MaikoTan:
learnt the new method from the Dalamud project.

### Note
Didn't test 'cause in hotfixes maintainance, I'll test it tonight.


MaikoTan:
Tested, it works.

PR #3774, opened 2021-12-09
jobs: WAR surging tempest timer fix

AmenneHolelane:
it's not consistent at all, the actual displayed timer is usually off the ingame buff timer. included as much info as i could about how long it actually lasts but at least it displays now and is mostly accurate

### DIFF
--- a/ui/jobs/components/war.ts
+++ b/ui/jobs/components/war.ts
@@ -55,11 +55,6 @@ export class WARComponent extends BaseComponent {
     if (skill)
       this.comboTimer.duration = 15;
   }
-  // for some undocumented reason, neither storm's eye or mythril tempest combo function exactly as the tooltip says.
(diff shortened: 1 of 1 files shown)


AmenneHolelane:
for some undocumented reason, neither storm's eye or mythril tempest combo function exactly as the tooltip says. When initially applied, mythril tempest sets the duration to 30.5 seconds rounded down. Storm's Eye sets the duration to 30+1.8 or something. The initial duration of eye isn't exactly 32 seconds but it's really close, the buff only reads 30 seconds tho the duration is longer
Increasing the buff timer via an aditional eye/tempest adds 30 seconds as you'd expect. final duration is always +1.8ish or +0.5 seconds depending on which skill was used however overcapping the buff will always set the duration to 60.5 seconds, no matter which skill is used

quisquous:
Yeah, when I use this, it looks like the Storm's Eye buff is +1s over the box.  I'll merge this patch for now and will see if we can follow up to make this better.  Thanks for the change.

PR #3794, opened 2021-12-10
raidboss: 2nd EW extreme timeline touchups

quisquous:


ui/raidboss/data/06-ew/trial/hydaelyn-ex.txt
```diff
@@ -1,25 +1,31 @@
-# Hydaelyn Extreme
+### Hydaelyn Extreme
+# -p 65BC:336.2 65BB:1433.6
+# -ii 65A7 6F12 65BA 6507 65AA 6596 6690 65BE 686C 6C92
```
quisquous, line 3: @Legends0 This is what I meant about `-ii` stuff.  You can pass `-ii` parameters to ignore specific abilities when calling `make_timeline` rather than just manually trimming them.  I've started listing out what abilities those are when making timelines so that others can easily reproduce the timeline in the future.  `-p` is to add a phase sync to particular time for a particular ability. `-it` will give you targetable/untargetable lines for particular mob names too.

ui/raidboss/data/06-ew/trial/hydaelyn-ex.txt
```diff
@@ -152,54 +168,57 @@ hideall "--sync--"
...
+2097.7 "Lightwave" sync / 1[56]:[^:]*:Hydaelyn:669[345]:/ jump 2206.7
 2117.2 "Echoes x5" #sync / 1[56]:[^:]*:Hydaelyn:65B5:/
 2125.0 "Bright Spectrum" #sync / 1[56]:[^:]*:Hydaelyn:65B9:/
 2125.0 "Dichroic Spectrum" #sync / 1[56]:[^:]*:Hydaelyn:65B8:/
 2137.3 "Crystallize" #sync / 1[56]:[^:]*:Hydaelyn:659A:/
 
+
 # Lightwave 3
-2206.7 "Lightwave" #sync / 1[56]:[^:]*:Hydaelyn:6693:/
+2206.7 "Lightwave" sync / 1[56]:[^:]*:Hydaelyn:669[345]:/
```
quisquous, line 194: I always saw the first lightwave as 6693, but I saw others as 669[345], so I left this one as is possibly being any of them, as missing this sync would be bad.  It looks like there's only three values though: https://xivapi.com/search?indexes=action&string=Lightwave&columns=Name,ID&pretty=true

ui/raidboss/data/06-ew/trial/hydaelyn-ex.txt
```diff
@@ -49,27 +57,33 @@ hideall "--sync--"
...
+1433.6 "Exodus" sync / 1[56]:[^:]*:Hydaelyn:65BB:/ window 2000,10
 1448.4 "Exodus" sync / 1[56]:[^:]*:Hydaelyn:662B:/
+1453.7 "--targetable--"
 1468.8 "Halo" sync / 1[56]:[^:]*:Hydaelyn:65A5:/
 
+
 # Lightwave 1
 1476.9 "Lightwave" sync / 1[56]:[^:]*:Hydaelyn:6693:/
-1488.7 "Light of the Crystal" sync / 1[56]:[^:]*:Hydaelyn:65A6:/
+1488.7 "Light of the Crystal" #sync / 1[56]:[^:]*:Hydaelyn:65A6:/
```
quisquous, line 84: These were just too close in time, and `test_timeline` showed that the second one would resync.

ui/raidboss/data/06-ew/trial/hydaelyn-ex.txt
Legends0: Makes sense. When making this timeline, I forgot that the lightwaves have different patterns such as which side will hit first.

ui/raidboss/data/06-ew/trial/hydaelyn-ex.txt
Legends0: Thanks for testing this.

PR #3798, opened 2021-12-11
raidboss: Zodiark Ex triggers

valarnin:
Still extremely WIP. Only covers phase one, not extensively tested yet. Probably lots of logic mistakes still.

No clue how to handle the star patterns, I see nothing in logs that could help yet. If there's a similar pattern with IDs as there was for quetzalcoatl, then maybe I could make that work, but I haven't identified one yet.

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
@@ -1,14 +1,119 @@
 import NetRegexes from '../../../../../resources/netregexes';
+import { UnreachableCode } from '../../../../../resources/not_reached';
+import Outputs from '../../../../../resources/outputs';
+import { callOverlayHandler } from '../../../../../resources/overlay_plugin_api';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
+import { PluginCombatantState } from '../../../../../types/event';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  sigilMap: { [id: string]: 'Blue' | 'Red' };
+  storedSigils: { x: number; y: number; id: string }[];
+  storedTetherTargets: { [id: string]: PluginCombatantState };
+  storedPythons: PluginCombatantState[];
+  // For Quetz, mechanics which only use two of them always use the two with the highest ID
+  storedQuetzalcoatls: PluginCombatantState[];
+  storedBehemoths: PluginCombatantState[];
+  exoterikosCounter: number;
+  paradeigmaCounter: number;
+  astralFlowCounter: number;
+  adikiaCounter: number;
+}
+
+const directionOutputStrings = {
+  northeast: Outputs.northeast,
+  north: Outputs.north,
+  northwest: Outputs.northwest,
+  west: Outputs.west,
+  southeast: Outputs.southeast,
+  south: Outputs.south,
+  southwest: Outputs.southwest,
+  east: Outputs.east,
+  combo: {
+    en: '${dir1} / ${dir2}',
+  },
+};
+
+const pythonLineOutputStrings = {
+  '0': Outputs.num1,
+  '1': Outputs.num2,
+  '2': Outputs.num3,
+  '3': Outputs.num4,
+  'row': {
+    en: 'Row ${first}/${second}',
+  },
+  'col': {
+    en: 'Col ${first}/${second}',
+  },
+  'singleRow': {
+    en: 'Row ${first}',
+  },
+};
+
+const refreshCombatantsPromise = async (data: Data) => {
+  const combatants = await callOverlayHandler({
+    call: 'getCombatants',
+  });
+  data.storedBehemoths = [];
+  data.storedPythons = [];
+  data.storedQuetzalcoatls = [];
+  for (const combatant of combatants.combatants) {
+    const id = combatant.ID;
+    if (id === undefined)
+      continue;
+    if (combatant.BNpcID === 14386)
+      data.storedBehemoths.push(combatant);
+    if (combatant.BNpcID === 14387)
+      data.storedPythons.push(combatant);
+    if (combatant.BNpcID === 14388)
+      data.storedQuetzalcoatls.push(combatant);
+    if (combatant.BNpcID === 13712)
+      data.sigilMap[id.toString(16).toUpperCase()] = 'Red';
+    if (combatant.BNpcID === 13713)
+      data.sigilMap[id.toString(16).toUpperCase()] = 'Blue';
+  }
+  data.storedBehemoths.sort((l, r) => {
+    return (l.ID ?? 0) - (r.ID ?? 0);
+  });
+  data.storedPythons.sort((l, r) => {
+    return (l.ID ?? 0) - (r.ID ?? 0);
+  });
+  data.storedQuetzalcoatls.sort((l, r) => {
+    return (l.ID ?? 0) - (r.ID ?? 0);
+  });
+};
+
+const getPythonLine = (python: PluginCombatantState): number => {
+  let coordinate = python.PosX;
+  if (python.PosX < 80 || python.PosX > 120)
+    coordinate = python.PosY;
+
+  if (coordinate < 85)
+    return 0;
+  if (coordinate < 100)
+    return 1;
+  if (coordinate < 115)
+    return 2;
+  return 0;
+};
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.TheMinstrelsBalladZodiarksFall,
   timelineFile: 'zodiark-ex.txt',
+  initData: () => ({
+    sigilMap: {},
+    storedSigils: [],
+    storedTetherTargets: {},
+    storedBehemoths: [],
+    storedQuetzalcoatls: [],
+    storedPythons: [],
+    exoterikosCounter: 0,
+    paradeigmaCounter: 0,
+    astralFlowCounter: 0,
+    adikiaCounter: 0,
+  }),
   triggers: [
     {
       id: 'ZodiarkEx Ania',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -19,6 +124,307 @@ const triggerSet: TriggerSet<Data> = {
       netRegexJa: NetRegexes.startsUsing({ id: '67EF', source: 'ゾディアーク' }),
       response: Responses.tankBuster(),
     },
+    {
+      id: 'ZodiarkEx Blue Tether',
```
quisquous, line 128: Blue (Purple?) is the cone? and red is the half arena?

Is it possible to add more differentiation between the two? It looks like right now that if you had a west red tether and an east blue tether that they would both say "east", but the safe spots for those are different.

Should the conal tether say "NE/SE" or "East Cone" or something else?

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -19,6 +124,307 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegexJa: NetRegexes.startsUsing({ id: '67BF', source: 'ゾディアーク' }),
+      condition: (data) => data.paradeigmaCounter === 0,
+      delaySeconds: 0.1,
+      alertText: (data, _matches, output) => {
+        ++data.paradeigmaCounter;
+        return output.underQuetz!();
+      },
+      outputStrings: {
+        underQuetz: {
+          en: 'Under NW Quetzalcoatl',
```
quisquous, line 299: Thanks for picking one.  :+1:

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -19,6 +124,307 @@ const triggerSet: TriggerSet<Data> = {
...
+      outputStrings: directionOutputStrings,
+    },
+    {
+      id: 'ZodiarkEx Paradeigma 1',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '67BF', source: 'Zodiark' }),
+      netRegexDe: NetRegexes.startsUsing({ id: '67BF', source: 'Zodiark' }),
+      netRegexFr: NetRegexes.startsUsing({ id: '67BF', source: 'Zordiarche' }),
+      netRegexJa: NetRegexes.startsUsing({ id: '67BF', source: 'ゾディアーク' }),
+      condition: (data) => data.paradeigmaCounter === 0,
```
quisquous, line 291: This is maybe a silly request, but do you think it's worth having the `paradeigmaCounter` match the `id` here? It seems harder to get wrong if they're the same.

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -19,6 +124,307 @@ const triggerSet: TriggerSet<Data> = {
...
+    },
+    {
+      id: 'ZodiarkEx Paradeigma 2',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '67BF', source: 'Zodiark' }),
+      netRegexDe: NetRegexes.startsUsing({ id: '67BF', source: 'Zodiark' }),
+      netRegexFr: NetRegexes.startsUsing({ id: '67BF', source: 'Zordiarche' }),
+      netRegexJa: NetRegexes.startsUsing({ id: '67BF', source: 'ゾディアーク' }),
+      condition: (data) => data.paradeigmaCounter === 1,
+      delaySeconds: 0.1,
```
quisquous, line 311: Oh wow, 0.1?

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -19,6 +124,307 @@ const triggerSet: TriggerSet<Data> = {
...
+    {
+      id: 'ZodiarkEx Paradeigma 3',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '67BF', source: 'Zodiark' }),
+      netRegexDe: NetRegexes.startsUsing({ id: '67BF', source: 'Zodiark' }),
+      netRegexFr: NetRegexes.startsUsing({ id: '67BF', source: 'Zordiarche' }),
+      netRegexJa: NetRegexes.startsUsing({ id: '67BF', source: 'ゾディアーク' }),
+      condition: (data) => data.paradeigmaCounter === 2,
+      delaySeconds: 0.1,
+      run: (data) => ++data.paradeigmaCounter,
```
quisquous, line 342: Do you think it's worth just having one "Paradeigma Counter" trigger up top that handles this in all places, vs conditionally incrementing everywhere?

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -1,14 +1,119 @@
...
+  },
+};
+
+const pythonLineOutputStrings = {
+  '0': Outputs.num1,
+  '1': Outputs.num2,
+  '2': Outputs.num3,
+  '3': Outputs.num4,
+  'row': {
+    en: 'Row ${first}/${second}',
```
quisquous, line 45: I'm a little bit skeptical of Row/Column outputs being something that will be meaningful to a lot of people.

What do you think about some form like one of these:
* North all the way
* Slightly north of center
* Just east of center
* All the way west

For the later paradeigma with snakes, then maybe something like "NE (lean S)", since folks will already be constrained to a small area.

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
valarnin: That's the one that every single PF I've been in goes to.

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
valarnin: There's some wonky timing issues here because the `=== 0` trigger will fire and then resolve before this one fires, so then this one fires for the same mechanic.

ui/raidboss/data/06-ew/trial/zodiark-ex.ts
valarnin: Wouldn't that run into the same race condition as the reason I added the 0.1 second delay?

quisquous: If anything, I think it would address it, if I'm understanding this correctly.  I think there are two ways to remove the `delaySeconds: 0.1`.  One is to reverse order your paradeigmas in the file so that they can never chain into each other by having 0 fire, increment, and then 1 fire.  The other is to have a single counter incrementing trigger at the top above everything (so that you only increment once), and then have all of the following triggers just check (but not increment).  Since they would all be checking different values, only one would fire.

valarnin:
I've revisited this multiple times over the last few days and I'm struggling with a way to translate this fight fully to triggers given it's so stateful and the trigger for each mechanic varies. Does anyone have any ideas on how to set this up other than my original idea of just incrementing a counter for each potential trigger mechanic and pacing it out that way?

quisquous:
> I've revisited this multiple times over the last few days and I'm struggling with a way to translate this fight fully to triggers given it's so stateful and the trigger for each mechanic varies. Does anyone have any ideas on how to set this up other than my original idea of just incrementing a counter for each potential trigger mechanic and pacing it out that way?

How many different types of mechanic triggers are you dealing with? There's paradeigma, but what else is triggering? Can you give a rough overview of the structure of what's needed here?

I think in the past if there's like mechanicA and mechanicB, and they're on different triggers and they're interleaved, usually I'd have a series of triggers like:
* counterA: increments data.countA when A happens
* counterB: increments data.countB when B happens
* A1: runs on A and data.countA === 1
* A2: runs on A and data.countA === 2
* etc
* B1: runs on B and data.countB === 1
* etc

valarnin:
Pulling from the timeline, off the top of my head, combined mechanics are decided on the corresponding startsCasting for:
- 69.6 "Esoterikos 2" (Cone side, north or south)
- 93.9 "Algedon" (Quetzalcoatl side, away from diagonal sweep)
- 135.6 "Astral Flow" (Cone side, away from Python lines)
- 166.0 "Adikia" (Under boss, away from Python lines)
- 190.6 "Esoteric Pattern 1" (2 and 3 as well, add phase, two are north by blue cone, one is south center)
- 374.8 "Triple Esoteric Ray" (Dodge into the first ray after it fires)
- 406.1 "Astral Flow" (North side under Behemoth, adjust for Python lines and fire line)
- 446.4 "Algedon" (East or west side depending on sigil, away from diagonal sweep)
- 475.8 "Astral Flow" (Northwest Quetzalcoatl, north or south side depending on fire line, adjust to one extreme for Python positions)
- 508.3 "Trimorphos Exoterikos" (Not 100% sure how to write the trigger for this yet, since it's three random sigils around the room and requires starting positioning immediately)
- 606.8 "Astral Flow" (NE/NW depending on sigil, adjust for Python lines)
- 648.4 "Triple Esoteric Ray" (NE/N/NW depending on cone and green lines, dodge into the line)
- 678.8 "Astral Flow" (Under north Behemoth, adjust for fire line)
- 717.0 "Trimorphos Exoterikos" (Same as the one at 508.3)
- 771.0 "Astral Flow" (Under NW Quetzalcoatl, adjust for fire and python lines)

quisquous:
> Pulling from the timeline, off the top of my head, combined mechanics are decided on the corresponding startsCasting for:

So, Algedon, Adikia, Astral Flow, Trimorphos Exoterikos, Triple Esoteric Ray, and some one-off Esoteric X.  Foof, that's a lot of separate things.

I think it'd be a lot to write out (and maybe it doesn't feel elegant), but I think having a counter and a single trigger for all counts (or separate triggers per count, if that makes more sense) might be a way to get this started.  Once everything is laid out some better organization will appear.  And if not, then maybe being explicit about each variation is the only way to handle this.

The only other thing I can think is that if you have a bunch of places where you're calling "getCombatants" and then handling some combo of birds, beasts, fire walls, esoteric things, rotations, maybe you could have a single solver that says "here are the safe spots for these things" and then reuse that? I'm not sure what that would look like in more detail, but that might be an approach to consider too, if there's some more general logic that can be reused.

amki:
I've tested this pr over some zodiark encounters, unfortunately the pythons as well as the quetzalcoatls are off. Looks like there is a logic error somewhere.

valarnin:
Pythons are a logic error that I've resolved locally. Quetzalcoatls are a timing issue that I haven't quite gotten right yet.

amki:
I've started work on this @ https://github.com/amki/cactbot/tree/zodiark-ex which is heavily inspired by your ideas.

I already managed to track the state of all quetz, pythons and sigils (read: if they are currently visible and contributing to the fight) but the correct calls for arena turning are still missing

valarnin:
@amki If you want to take over writing the triggers for this fight, that's fine by me. I took a step back after clearing 60ish times to avoid getting burnout, and I'm busy with holiday stuff for probably the next 3-4 days.

amki:
No idea if I can actually do it but I'm giving it my best shot. :)

amki:
I have almost everything handled so I'm going to put in a new pull request because there will be a metric megaton of discussion because I've gone completely off the rails.

valarnin:
Closing in favor of  #3933

PR #3800, opened 2021-12-11
jobs: Temporarily fake kLevelMod data to avoid error

### LINKED ISSUE #3788: Jobs: GCD calculation can not work properly
When you change to job above level 80, console will output an error: `Error: This code shouldn't be reached`
If this job module contains something using GCD calculation, it will continously output this error when using it. 
Job below level 80 won't recive this error, but GCD calculation also works not properly.
Seems this error happens because kLevelMod have not updated yet.

Echoring:
Fixes #3788
https://github.com/quisquous/cactbot/blob/54303910154b6ae5847780f1dbac616ce629a381/ui/jobs/utils.ts#L142
[0,0] will cause a 0 division error, so make it 9999 to make unknown gcd about 2.5
382 at lv81 and 400 at lv90 are confirmed, others are assumed.

### DIFF
--- a/ui/jobs/constants.ts
+++ b/ui/jobs/constants.ts
@@ -469,4 +469,17 @@ export const kLevelMod = [
   [376, 3034],
   [378, 3164],
   [380, 3300],
+  // FIXME: The following are the assumed values added to avoid errors.
+  // 9999 will make unknown gcd about 2.5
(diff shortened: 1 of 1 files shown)


MaikoTan (approved):

quisquous:
Thank you!!

PR #3801, opened 2021-12-11
i18n: (raidboss) add EW kr translations

Bing-su:


ui/raidboss/data/06-ew/trial/hydaelyn-ex.ts
```diff
@@ -20,23 +20,27 @@ const storedMechanicsOutputStrings = {
...
     en: 'Party Stack',
+    ko: '쉐어',
   },
 };
 
 const crystallizeOutputStrings = {
   ...storedMechanicsOutputStrings,
   crystallize: {
     en: 'Crystallize: ${name}',
+    ko: 'Crystallize: ${name}', // FIXME
```
quisquous, line 35:
```suggestion
```

If you can't translate it, I think it's better to leave it.

Bing-su: you're right. I fixed it while rebasing, please check it.

quisquous: Thanks! Sorry for the small correction, but I just wanted more to let you know for the future.  :)

PR #3810, opened 2021-12-12
i18n: will EW translation ever stop

Akurosia:


quisquous:
> i18n: will EW translation ever stop

can't stop, won't stop

PR #3818, opened 2021-12-12
jobs: refactor procs boxes in flex layout

MaikoTan:
Per https://github.com/quisquous/cactbot/pull/1856#issuecomment-706589184
So we can abandon to use so many css selectors and calculate the distances.
Add a new div as a container for proc boxes, so now the construction is:
image
As my test when using caster and healer, it should look the same as before:
BLM:
image


quisquous:
Thank you! This will make it much easier in the future.

PR #3823, opened 2021-12-13
jobs: fix effect/gain/you dont trigger self gains buffs

MaikoTan:
This bug would break something like BuffTracker that
relies on the event, so buffs would not correctly trigger on the event
when the player gains some effect but when the others gains effect
from the player.

This patch fixed this. But I am a bit worrying whether the name is
clear enough for what is it used for.


quisquous:
> This patch fixed this. But I am a bit worrying whether the name is clear enough for what is it used for.

The name is clear to me, for what it's worth.

PR #3830, opened 2021-12-13
jobs: refactor combo tracker as an event emitter

MaikoTan:
Also fix a bug that combo timer would just stop after 15 seconds, because there is a `setTimeout` called with the old 15000ms.

### DIFF
--- a/ui/jobs/combo_tracker.ts
+++ b/ui/jobs/combo_tracker.ts
@@ -1,4 +1,7 @@
+import { EventEmitter } from 'eventemitter3';
+
 import { kAbility, kComboBreakers, kComboBreakers5x } from './constants';
+import { Player } from './player';
 
 type StartMap = {
   [s: string]: {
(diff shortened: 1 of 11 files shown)


ui/jobs/combo_tracker.ts
```diff
@@ -65,146 +93,27 @@ export default class ComboTracker {
...
-      kAbility.Fountain,
-    ]);
-    comboTracker.AddCombo([
-      kAbility.Windmill,
-      kAbility.Bladeshower,
-    ]);
+    const comboTracker = new ComboTracker({
+      player: player,
+      comboBreakers: breakers,
+      comboDelayMs: is5x ? 15000 : 30000,
```
quisquous, line 114: Is it possible to not duplicate this time between `base.ts` and `combo_tracker.ts`?

MaikoTan: Yeah, I am just wondering should I put it in constants.ts or just in ComponentManager.

quisquous: I think you could put both values of constants into constants.ts, and then just pick one in each place based on is5x? Maybe there's a better idea, but that's one thought.  I just want to have the numbers in one place.

MaikoTan: I have the same thought.
But I am thinking "pass is5x to ComboTracker" vs "just pass comboDelayMs that assigned outside to ComboTracker", which is better.

quisquous: Looks good to me, thanks!

PR #3831, opened 2021-12-13
jobs: WAR: Re-support ShB

Echoring:
#3774 forgot those regions still in ShB.


quisquous:
> #3774 forgot those regions still in ShB.

Whooooops.  Thank you!

PR #3834, opened 2021-12-13
jobs: AST: Draw and Sealbox change

Echoring:
Draw is chargeable now.
In EW you can no longer ensure that you can gain 3 different seals due to change of redraw and minor arcana.
So change this box to show seal amount instead of kinds, any 3 seals will make this box green.

### DIFF
--- a/ui/jobs/components/ast.ts
+++ b/ui/jobs/components/ast.ts
@@ -101,7 +101,10 @@ export class ASTComponent extends BaseComponent {
         this.combustBox.duration = 18;
         break;
       case kAbility.Draw:
-        this.drawBox.duration = 30 + this.drawBox.value;
+        if (this.is5x)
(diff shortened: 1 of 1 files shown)


ui/jobs/components/ast.ts
```diff
@@ -90,7 +101,7 @@ export class ASTComponent extends BaseComponent {
         this.combustBox.duration = 18;
         break;
       case kAbility.Draw:
-        this.drawBox.duration = 30;
+        this.drawBox.duration = 30 + this.drawBox.value;
```
quisquous, line 104: Does this need to be behind an `if (this.is5x) {`?

Echoring: You are right. I think this will not make a difference at first, but then I come to "Lost Font of Skill"...

PR #3836, opened 2021-12-13
jobs: NIN: Fix huton and ninjutsu

Echoring:
Seems they have been broken for a long time for the change of #3545


quisquous:
> Seems they have been broken for a long time for the change of #3545

Thank you for fixing all my bugs!

PR #3842, opened 2021-12-14
util: add G'raha to list of ignored names

quisquous:


JLGarber (approved):
Ignoring our best boy. Whatever sadness will Square visit upon us next?

PR #3843, opened 2021-12-14
raidboss: Stigma Dreamscape initial timeline and triggers

JLGarber:
Lots of stuff that we don't know yet in the timeline, but unfortunately even with randoms I haven't been able to get logs long enough to tell me what we need. However, the data present in the timelines is accurate against the More Than One logs I tried.


ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.ts
```diff
@@ -0,0 +1,257 @@
...
+        text: {
+          en: 'Guided Missile on YOU',
+        },
+      },
+    },
+    {
+      id: 'Dreamscape Mustard Bomb',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '632B', source: 'Proto-Omega' }),
+      response: Responses.tankBuster(),
```
JLGarber, line 99: I haven't tested whether any of the busters in here splash, since I've only run this with randoms.

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.ts
```diff
@@ -0,0 +1,257 @@
...
+    },
+    {
+      id: 'Dreamscape Assault Cannon',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '63AB', source: 'Arch-Lambda', capture: false }),
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Go to second dash side',
```
JLGarber, line 108: I don't like this wording. 

The pattern *seems* to be that the boss dashes in the shape of a figure 4. The second dash *should* always be on the same side of the first dash as the end of the fourth dash. But it's possible there are other patterns I haven't seen.

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.ts
```diff
@@ -0,0 +1,257 @@
...
+      type: 'HeadMarker',
+      netRegex: NetRegexes.headMarker({ id: ['004F', '0050', '0051', '0052'] }),
+      condition: Conditions.targetIsYou(),
+      alertText: (_data, matches, output) => {
+        const limitCutNumber = limitCutNumberMap[matches.id];
+        return output.text!({ num: limitCutNumber });
+      },
+      outputStrings: {
+        text: {
+          en: '#${num} laser on YOU!',
```
JLGarber, line 123: This might need rewording to indicate that the laser originates from the paired head marker and not from the boss?

JLGarber:
One thing that might be improvable is to account for the combination of Mindjack and Electromagnetic Release. Currently we tell the user about each and just sort of expect them to handle it.

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.txt
```diff
@@ -0,0 +1,125 @@
...
+
+65.6 "Iron Kiss" sync / 1[56]:[^:]*:Mark II Guided Missile:6327:/
+73.0 "Electric Slide" sync / 1[56]:[^:]*:Proto-Omega:632A:/
+84.2 "Mustard Bomb" sync / 1[56]:[^:]*:Proto-Omega:632B:/
+96.4 "Side Cannons" sync / 1[56]:[^:]*:Proto-Omega:(6320|6321):/ window 10,10
+104.6 "Guided Missile" sync / 1[56]:[^:]*:Proto-Omega:6326:/
+111.7 "Chemical Missile" sync / 1[56]:[^:]*:Proto-Omega:6328:/
+116.8 "Burn" sync / 1[56]:[^:]*:Proto-Omega:6329:/
+
+# TODO This is likely inaccurate past Mustard Bomb.
```
quisquous, line 29: Do you think it's worth guessing that this is a loop for now?

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.txt
```diff
@@ -0,0 +1,125 @@
...
+2151.6 "Rush" sync / 1[56]:[^:]*:Proto-rocket Punch:642D:/ window 30,30
+2157.5 "Fire Breath?" sync / 1[56]:[^:]*:Hybrid Dragon:642E:/
+2157.5 "Proto-wave Cannon?" sync / 1[56]:[^:]*:Omega Frame:(642A|642B):/
+
+# TODO: Still not a loop, more data is needed
+2168.1 "Mindhack" sync / 1[56]:[^:]*:Stigma-4:6430:/ window 30,30
+2172.2 "--sync--" sync / 1[56]:[^:]*:Stigma-4:(6431|6433):/
+2177.3 "Self-Destruct?"
+2182.2 "Electromagnetic Release" sync / 1[56]:[^:]*:Stigma-4:(6432|6434):/
+2192.3 "Multi-AI Takeover" sync / 1[56]:[^:]*:Stigma-4:6C4B:/
```
quisquous, line 118: Is this a loop with the previous Multi-AI Takeover maybe?

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.txt
```diff
@@ -0,0 +1,125 @@
...
+2067.0 "--sync--" sync / 1[56]:[^:]*:Stigma-4:6433:/
+2077.0 "Electromagnetic Release" sync / 1[56]:[^:]*:Stigma-4:6434:/
+2085.1 "--sync--" sync / 1[56]:[^:]*:Stigma-4:6431:/
+2095.1 "Electromagnetic Release" sync / 1[56]:[^:]*:Stigma-4:6432:/
+2108.2 "Mindhack" sync / 1[56]:[^:]*:Stigma-4:6430:/ window 30,30
+2113.4 "--sync--" sync / 1[56]:[^:]*:Stigma-4:(6431|6433):/
+2123.4 "Electromagnetic Release" sync / 1[56]:[^:]*:Stigma-4:(6432|6434):/
+2132.5 "Atomic Ray" sync / 1[56]:[^:]*:Stigma-4:6436:/
+2142.7 "Multi-AI Takeover" sync / 1[56]:[^:]*:Stigma-4:6C4B:/
+2148.4 "Touchdown?" sync / 1[56]:[^:]*:Hybrid Dragon:68F9:/
```
quisquous, line 108: Why does touchdown get a question mark? (and firebreath above too?)

JLGarber: Because each AI Takeover is either a Hybrid Dragon spawn *or* an Omega Frame spawn, and their abilities are different, but the timings on everything else are otherwise the same. They don’t alternate either, although I don’t *think* I’ve seen a run with all Dragons or Frames.

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.ts
```diff
@@ -0,0 +1,257 @@
...
+      // then immediately begins casting Fire Breath in a cone across the arena.
+      // If the player is not already in motion by the time Fire Breath begins,
+      // they are likely to be hit.
+      id: 'Dreamscape Touchdown',
+      type: 'AddedCombatant',
+      netRegex: NetRegexes.addedCombatant({ name: 'Hybrid Dragon' }),
+      condition: (data) => data.lastBoss,
+      infoText: (_data, matches, output) => {
+        // The arena is a 50x50 square, with (0,0) in the exact center.
+        const isEast = parseFloat(matches.x) > 5;
```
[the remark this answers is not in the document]
JLGarber: You’re right, this one’s probably safe to just check negative/positive. Usually we don’t have a safe (0,0) origin, so it’s better in those cases to compare against known “safe” values.

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.ts
```diff
@@ -0,0 +1,257 @@
...
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6B9D', source: 'Arch-Lambda' }),
+      response: Responses.tankBuster(),
+    },
+    {
+      // Dragons spawn outside the last boss, but those ones don't matter.
+      // Ensure that we don't say anything until the player has engaged the last boss.
+      id: 'Dreamscape Last Boss',
+      type: 'GameLog',
+      netRegex: NetRegexes.message({ line: 'A-4 Headquarters will be sealed off.*?', capture: false }),
```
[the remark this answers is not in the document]
JLGarber: That’s fair, I was mostly following patterns from previous work that used triggers like this. Maybe Stigma-4’s auto-attack?

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.txt
JLGarber: Probably not? The Proto-Wave Cannon at 2157.3 is followed up by Mindhack, while the one at 2207.1 is followed by Atomic Ray, and it's incredibly unlikely those are alternative mechanics.

ui/raidboss/data/06-ew/dungeon/stigma_dreamscape.txt
JLGarber: I suppose there's no reason not to, you're right.

JLGarber: Hmm, looking at this more, I'm not completely sure this *is* the loop. I had left it just as it was generated for the initial version, but for the loop I went and adjusted the previous section to be the lookahead, and it's definitely got different timings.

I think this is probably safe to merge for now, but we definitely need more data.

quisquous:
Could you fix the conflicts too?

PR #3847, opened 2021-12-14
raidboss: fix 2nd EW extreme Crystallize->Beacon->Aureole

### LINKED ISSUE #3846: EW ex trial: crystalize > beacon> aureole callouts are wrong
The callout after the beacons says e.g. "Sides > Healer groups" while as the healer groups were before the beacon dashes and will not happen after the aureole cast again.

releastes to #3820

quisquous:
Fixes #3846.


quisquous:
image

PR #3849, opened 2021-12-15
jobs: fix only buff option not hide procs

### LINKED ISSUE #3848: Is it possible to add a setting to turn off these new boxes?
image

I updated just a few minutes ago and now these boxes for No Mercy, Bloodfest, and Gnashing Fang are on my screen permanently. I can understand why they are useful to some people, but I do not wish to have them.

Could a toggle be added?

I would like to just go back to have raid buffs and food being tracked only.

MaikoTan:
Fixes #3848

### DIFF
--- a/ui/jobs/jobs.css
+++ b/ui/jobs/jobs.css
@@ -34,8 +34,8 @@
   font-size: 12px;
 }
 
-.justbuffs #bars,
-.justbuffs #procs-container {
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for the quick fix! Sorry for not thinking about this earlier when I looked at the other patch.

PR #3853, opened 2021-12-15
jobs: minor improvements for proc boxes

MaikoTan:
- make proc boxes wrap when there are more than 4 boxes
- fix a typo: notfy -> notify
- stop using `[job]-proc` id everywhere, use class instead. technically HTML only allows one element per one id.


ui/jobs/jobs.css
```diff
@@ -363,11 +363,10 @@ div.justbuffs div#procs-container {
   top: 146px;
   left: 100px;
   display: flex;
-  flex-direction: row;
-  justify-content: space-between;
-  align-items: baseline;
+  flex-flow: row wrap;
```
quisquous, line 366: Right now, 4 is the most proc boxes anything has and so nothing will wrap, right?

I think I am a little bit unsure here, because already we extend the buffs out to the right, and so it'd make sense to me visually to just extend the proc boxes further right if we have more than four as well, rather than taking up extra space below.  What's the reasoning behind wrapping vs extending?

MaikoTan: Yes, currently we have only 4 at the most.
This is requested by Echoring that she said some jobs might have more than 4 boxes in the future, so I add this feature. I think extending is also okay in this case, just needs more CSS tweaking. In my opinion, if there are more than 5 or 6 boxes at the bottom, would be a bit asymmetric(?) so I support wrapping moreover. What is your opinion instead?

quisquous: I think it's just *personally* I don't have any room for boxes underneath on my screen, and so my initial preference is to continue extending in the direction it's already going.

How weird does it look extended with 5 or 6?  Is it possible to adjust the flex parameters and so it looks good?

MaikoTan: image
Although this example would be a little extreme, I think it would be better to wrap them rather than extend.
Or what about adding an extra option to control this behaviour for users?

quisquous: I'm kind of opposed to adding options unless it's really necessary.  We don't really have the testing set up to handle combinations of options, and things get forgotten (e.g. "proc boxes only") when writing other features.

If you want to wrap it, let's just wrap it.  I think not wrapping looks fine to me and I'll just adjust my own CSS to do that.

MaikoTan: Maybe we need e2e tests or something in the future?
And yeah, since cactbot gets more and more complicated, it is hard to cover all the tests when changing something.

Sorry I was just very busying at both work and home in the past few months, so I seldom contributed to cactbot. Let me fetch and merge the newest main branch first and then talk about the detail of this PR.

quisquous: Yeah, I think better testing would be good, but I also think it's good not to have a million supported options either.  Too many options means that it just makes it harder to create new features or changes (even if it is tested).  I am happy to sort of "support myself" for this option rather than create a burden on the code.

Sorry for not landing this earlier, I think it said "draft" and so it seemed like you weren't ready to land.  Maybe I should have asked.

MaikoTan: As you said, I not going to add more changes in this PR, see if these changes satify you.
Not only because I have no much time to do the coding and testing recently, But I also have been several month that I didn't touch cactbot or FF. :-(

PR #3858, opened 2021-12-16
raidboss: basic Ktisis Hyperboreia triggers

quisquous:


quisquous:
@JLGarber Re: timeline.
* I found more Pyric breath combinations, and the first double combo is not guaranteed either.
* After that, test_timeline ran cleanly, with no missing Pyric Spheres either
* I'm not 100% sure on the Ladon jump as two of the abilities have an extra +1s when I run test, which sometimes is supicious
* Also saw through Hermes 2224.6 (the first Aero IV in the loop block) and that was clean, but not enough to have total surety on the loop

ui/raidboss/data/06-ew/dungeon/ktisis_hyperboreia.txt
```diff
@@ -77,23 +82,26 @@ hideall "--sync--"
 1077.4 "Pyric Sphere" sync / 1[56]:[^:]*:Pyric Sphere:6490:/
 1079.8 "--sync--" sync / 1[56]:[^:]*:Ladon Lord:6485:/
 1082.5 "Pyric Sphere" sync / 1[56]:[^:]*:Pyric Sphere:6491:/
-1088.9 "Pyric Breath #1" #sync / 1[56]:[^:]*:Ladon Lord:(6486|6487|648A|648B):/
-1091.0 "Pyric Breath #2" #sync / 1[56]:[^:]*:Ladon Lord:(6486|6487|648A|648B):/
+1081.9 "--rotate--"
```
JLGarber, line 85: The tiniest of nits, but perhaps swap the Rotate and Pyric Sphere lines to maintain timestamp order, if the Rotate cast isn't intended to *be* the Rotate?

JLGarber (approved):

ui/raidboss/data/06-ew/dungeon/ktisis_hyperboreia.txt
quisquous: Ah, thanks! Yeah, the boss rotates as the sphere lines go off.

PR #3868, opened 2021-12-16
raidboss: Correct Dreamscape dragon direction trigger

JLGarber:
Per a report on the XIV plugin Discord server, this trigger seems to call only West, regardless of the dragon's actual spawn location. I'm not having much luck tracking down the issue, so for now we'll just tell the user to get behind the dragon the moment it spawns.


quisquous:
What if the issue is addedCombatantFull??

JLGarber:
```javascript
defaultParams('AddedCombatant', [
        'type',
        'timestamp',
        'id',
        'name',
      ]),
```
Aaaaaaaa.

I’ll correct that when I’m at my computer in half an hour or so. I completely forgot that was a thing.

PR #3874, opened 2021-12-17
jobs: WAR: fix Surging Tempest timer

### LINKED ISSUE #3778: jobs: Surging Tempest timer is inconsistent
Filing an issue here so this doesn't get lost.  See comments on #3774.

* Initial application of Storm's Eye is not exactly 30s (instead is ~31.1 or so, sometimes varying??)
* Mythril Tempest and Storm's Eye refreshing add a different amount of time
* Storm's Eye refreshing adds a slightly different amount of time than Storm's Eye initial

Echoring:
Fixes #3778 
Some hypothesis to the question in #3778:
> Initial application of Storm's Eye is not exactly 30s (instead is ~31.1):

Storm's Eye will apply Surging Tempest at once with a log says you gain Surging Tempest for 30s,
but this buff duration will not go down until Storm's Eyes' damage is dealed.
(if you use other ability to cancel its animation, duration will still be locked to designed time, this make things very complex when use Inner Release in this case)
When refreshing, this duration lock will not apply.
Mythril Tempest do not have buff lock time. 

> I found that the timer and box were consistent, but if I refreshed then they got offset by like ~0.5s?

There is a log line delay of 0.5s for most buff.

### DIFF
--- a/ui/jobs/components/war.ts
+++ b/ui/jobs/components/war.ts
@@ -71,7 +71,7 @@ export class WARComponent extends BaseComponent {
     if (skill === kAbility.MythrilTempest &&
       (this.eyeBox.duration === null || this.eyeBox.duration === 0)) {
       if (!this.is5x)
-        this.eyeBox.duration = 30 + 0.5;
+        this.eyeBox.duration = 30;
         this.initial = true;
         setTimeout(() => {
           this.initial = false;
--- a/ui/jobs/components/war.ts
+++ b/ui/jobs/components/war.ts
@@ -69,8 +69,8 @@ export class WARComponent extends BaseComponent {
     // See: https://github.com/quisquous/cactbot/issues/3778
     //
     // Here's a hack to at least get the initial application to be better.
-    const bonus = this.eyeBox.duration === 0 ? 1.1 : 0;
-    this.eyeBox.duration = duration + bonus;
+    const bonus = this.eyeBox.duration === 0 ? 1.6 : 0;
+    this.eyeBox.duration = duration + bonus - 0.5;
   }
   override onYouLoseEffect(id: string): void {
     // TODO: delete StormsEye after every region launch 6.0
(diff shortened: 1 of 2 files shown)


quisquous:
Hmmm.  This is a lot of heuristics.  I played a bunch with warrior and even though the time is off by a little bit, it was still very playable.  Because the gains effect line always has the full time (vs relative times that might accumulate errors over a long playtime), it's always off by a little bit and never off by a lot.  I'm a little inclined to keep this simpler even if wrong.

If you'd like to make this better, can I ask that you investigate looking at corresponding the 0x25 EffectResult lines with the 0x15 ActionEffect lines? For example, these two lines share the sequence `0006BB0F`, and so it seems likely that this is when the effect actually starts.

```
21|2021-12-18T13:14:29.4990000-08:00|1084E23D|Paprika Rika|2D|Storm's Eye|4000118D|Striking Dummy|46710003|15B80000|A3D|340000|A0F|A758000|11B|2D8000|0|0|0|0|0|0|0|0|2121774|2134350|0|10000|||-632.72|-845.13|36.00|1.57|60298|68221|10000|10000|||-630.73|-844.70|36.00|-1.78|0006BB0F|0|b9b135cbbdf33780
37|2021-12-18T13:14:31.1510000-08:00|1084E23D|Paprika Rika|0006BB0F|60298|68221|10000|10000|0||-630.73|-844.70|36.00|-1.78|1500|0|0|01|03000A75|EB5F0000|41F00000|1084E23D|f62d36b41dd85860
```

However, Ravahn mentions [here](https://discord.com/channels/551474815727304704/551476873717088279/919706647780278302) that sometimes the `EffectResult` is dropped, so it may be sufficient to say "if you ever see an `EffectResult` for this player, assume surging tempest has been applied" rather than checking sequence numbers.

Echoring:
Well, if you want to keep it simpler, I think I can remove Inner Release after Storm's eye at once support and it will be simple but more accurate than now. (only not accurate about 0.5s when Inner Release after Storm's eye at once)

PR #3876, opened 2021-12-18
jobs: fix component should reset on init

MaikoTan:
According to @Echoring , components should reset when it is initialized.
Otherwise, some variable is not valid value.

### DIFF
--- a/ui/jobs/components/ast.ts
+++ b/ui/jobs/components/ast.ts
@@ -47,6 +47,8 @@ export class ASTComponent extends BaseComponent {
     this.sealBox = this.bars.addResourceBox({
       classList: ['ast-color-seal'],
     });
+
+    this.reset();
   }
 
   override onJobDetailUpdate(jobDetail: JobDetail['AST']): void {
--- a/ui/jobs/components/blm.ts
+++ b/ui/jobs/components/blm.ts
@@ -75,6 +75,8 @@ export class BLMComponent extends BaseComponent {
     this.xenoTimer = this.bars.addResourceBox({
       classList: ['blm-xeno-timer'],
     });
+
+    this.reset();
   }
 
   override onUseAbility(id: string): void {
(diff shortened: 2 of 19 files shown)


ui/jobs/components/index.ts
```diff
@@ -214,6 +214,7 @@ export class ComponentManager {
 
       // initialize components
       this.component = this.getJobComponents(job);
+      this.component.reset();
```
quisquous, line 217: I'm not sure I understand this.  This function returns a constructed object.  If those objects are not valid, we should fix it in the constructors of those objects.

MaikoTan: But this way we don't need to duplicate those codes in the reset function in every constructor, so I think putting it here is good.

quisquous: What about calling `this.reset()` in the constructor then? This is an abstract thought, but I feel that in general if you create an object it should be initialized, because that way it can't ever be used incorrectly.

MaikoTan: But then you should write this.reset() in every components...

quisquous: Yes, but I think repeating is less bad? I think it's just wrong in general to have a class where calling `new` leaves you with a half-finished object.  I think it's just a bad practice, because then somebody in the future can call `new ThingComponent` and then be surprised that they have to do an extra step.

MaikoTan: Oh, I see your point. but I think it is better to initialize those values in the "box"-es itself instead. Maybe in another PRs.
Sorry I am a bit busy these days, and not react immediately, I'll tweak some more things and then can be merged.

quisquous: No worries! I know everybody's busy these days.  :)

I'm not sure I understand what you mean by "box"-es here, but I'll take a look when you tweak some things.

MaikoTan:
@quisquous So I think this patch would be like this to merge, some more tweaking would be done after I complete my other PRs later in a new one?

PR #3888, opened 2021-12-21
oopsy: add final trial

quisquous:


JLGarber (approved):
I'm kind of amazed how small the pool of mistake mechanics is for such a long encounter. Square put a lot of work into this one!

PR #3893, opened 2021-12-21
raidboss: Add p1n triggers

amki:
This adds triggers for the new pandæmonium raid's first boss.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p1n.ts
+++ b/ui/raidboss/data/06-ew/raid/p1n.ts
@@ -22,6 +22,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.startsUsing({ id: '6DA3', source: 'Erichthonios', capture: false }),
       response: Responses.goRight(),
     },
+    /* Aetherflail sends startsUsing Gaoler's Flail as well so don't double tell
     {
       id: 'P1N Aetherflail Right',
       type: 'StartsUsing',
@@ -34,6 +35,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.startsUsing({ id: '65E0', source: 'Erichthonios', capture: false }),
       response: Responses.goRight(),
     },
+    */
     {
       id: 'P1N Warder\'s Wrath',
       type: 'StartsUsing',
@@ -44,6 +46,12 @@ const triggerSet: TriggerSet<Data> = {
       id: 'P1N Shining Cells',
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '65E9', source: 'Erichthonios', capture: false }),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'P1N Slam Shut',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '65EA', source: 'Erichthonios', capture: false }),
       response: Responses.aoe(),
     },
     {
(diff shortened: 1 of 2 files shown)


Akurosia:
Zones are generated by ./uril/gen_zone_id_and_info.py

Akurosia:
new zoneNames will be
```
  'AsphodelosTheFirstCircle': 1002,
  'AsphodelosTheFourthCircle': 1008,
  'AsphodelosTheSecondCircle': 1004,
  'AsphodelosTheThirdCircle': 1006,
```

amki:
Thanks for supplying the zoneIds I guess that tool needs some kind of game data it reads via SaintCoinach that I don't have...

This seems to be ready for inclusion then, theoretically one could figure out where the objects for Intemperance are (NW/NE/SW/SE) but I don't intend to, knowing which one has to hit you is more than enough imo.

ui/raidboss/data/06-ew/raid/p1n.ts
```diff
@@ -0,0 +1,116 @@
...
+      netRegex: NetRegexes.startsUsing({ id: '65DF', source: 'Erichthonios', capture: false }),
+      response: Responses.goLeft(),
+    },
+    {
+      id: 'P1N Aetherflail Left',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '65E0', source: 'Erichthonios', capture: false }),
+      response: Responses.goRight(),
+    },
+    */
```
quisquous, line 39: I think no need for commented out trigger blocks.  I think if you wanted to leave a note on the Gaoler's Flail triggers above as a comment that `Also happens during Aetherflail Right (65DF)` or something, that would be a better way to leave a note.

```suggestion
```

quisquous (commented):
Thanks for sending triggers! This is super appreciated.

Could you remove the commented out triggers and then I'll merge this?

quisquous:
> Thanks for supplying the zoneIds I guess that tool needs some kind of game data it reads via SaintCoinach that I don't have...

Yeah, exactly.  It requires installing SaintCoinach so that it can read game data files.

quisquous:
(oops sorry, misclicked close and not comment!)

ui/raidboss/data/06-ew/raid/p1n.ts
amki: Sure, I changed it in 198dfa351cc0138c29a91ee5515c1580f4b6d9ae

quisquous:
I did a run with this earlier, and so here's some offhand thoughts for you (or others, if they want to pick them up) for some minor improvements to make this feel better.

* Get hit by red/blue should happen three times in a row and then not be called out on the fourth
* Right/Left + Fire/Light are both alerts and happen at roughly the same time.  I think these could be collected into one trigger, e.g. store whichever one happens first on `data` and then call them out together, e.g. `Right + Fire`
* There is a Right/Left during Get hit by Red/Blue, and if we really wanted to be fancy, we could say "Get Blue -> Right" or something like that to indicate which is going off first.

quisquous:
(Thanks again!!!)

amki:
> I did a run with this earlier, and so here's some offhand thoughts for you (or others, if they want to pick them up) for some minor improvements to make this feel better.
> 
>     * Get hit by red/blue should happen three times in a row and then not be called out on the fourth

I noticed this as well but it didn't always happen for me. I could easily count up to 3, ignore 4 and then reset count (because the mechanic repeats) but this wouldn't work if the 4th call is skipped the first time and I'm not sure why/how often this happens.

>     * Right/Left + Fire/Light are both alerts and happen at roughly the same time.  I think these could be collected into one trigger, e.g. store whichever one happens first on `data` and then call them out together, e.g. `Right + Fire`

I thought about this but I can't make it call out on the right/left because there is one iteration fire/light without right/left. I guess a construct that calls out on the first fire/light and then only notes which one it was and only calls out on right/left could work...

>     * There is a Right/Left during Get hit by Red/Blue, and if we really wanted to be fancy, we could say "Get Blue -> Right" or something like that to indicate which is going off first.

I agree this one is a little ugly, if the second boss doesn't break me I'll come back and fix this and the one before.

PR #3906, opened 2021-12-22
raidboss: p2n trigger

amki:
Adds triggers for the second pandæmonium normal boss.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p2n.ts
+++ b/ui/raidboss/data/06-ew/raid/p2n.ts
@@ -163,33 +163,10 @@ const triggerSet: TriggerSet<Data> = {
       // Aoe from outside the arena
       id: 'P2N Dissociation',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: '6806', source: 'Hippokampos', capture: false }),
+      netRegex: NetRegexes.startsUsing({ id: '6806', source: 'Hippokampos', capture: true }),
       delaySeconds: 1,
-      promise: async (data) => {
-        const callData = await callOverlayHandler({
-          call: 'getCombatants',
-        });
-        if (!callData || !callData.combatants || !callData.combatants.length) {
-          console.error('Dissociation: failed to get combatants: ${JSON.stringify(callData)}');
-          return;
-        }
-        // This is the right dissociationHippo, according to position.
-        const hippos = callData.combatants.filter((c) => c.BNpcID === 14441);
-        console.log(hippos);
-        if (hippos.length !== 1) {
-          console.error('Dissociation: There is not exactly one dissociationHippo?!', hippos);
-          console.log(callData.combatants);
-          data.dissociationHippo = undefined;
-          return;
-        }
-        data.dissociationHippo = hippos[0];
-      },
-      alertText: (data, _matches, output) => {
-        if (!data.dissociationHippo) {
-          console.error('Dissociation: No boss actor found. Did the promise fail?');
-          return 'unknown';
-        }
-        const xcord = data.dissociationHippo.PosX;
+      alertText: (data, matches, output) => {
+        const xcord = parseFloat(matches.x);
         if (xcord === 110)
           return output.e!();
         if (xcord === 90)
--- a/ui/raidboss/data/06-ew/raid/p2n.ts
+++ b/ui/raidboss/data/06-ew/raid/p2n.ts
@@ -31,6 +31,8 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'P2N Spoken Cataract',
       type: 'StartsUsing',
+      // create value for radiants that are readible (no hardcoded variables) -> we dont need FD
+      // Comment for what IDs are there for! TO DO!!!
       netRegex: NetRegexes.startsUsing({ id: ['67FD', '67F8', '67F7', '67F9'], source: 'Hippokampos', capture: true }),
       delaySeconds: 1,
       promise: async (data) => {
@@ -96,6 +98,106 @@ const triggerSet: TriggerSet<Data> = {
             return 'body south, head east';
           if (bodyHeading === 'east')
             return 'body east, head north';
+        }
+      },
+      outputStrings: {
+        text: {
+          en: 'Tank Laser on YOU',
+          de: 'Tank Laser auf DIR',
+          fr: 'Tank laser sur VOUS',
+          ja: '自分にタンクレーザー',
+          cn: '坦克射线点名',
+          ko: '탱 레이저 대상자',
+        },
+      },
+    },
+    {
+      id: 'P2N Sewage Deluge',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '67F6', source: 'Hippokampos', capture: false }),
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Aoe get on grid',
+        },
+      },
+    },
+    {
+      // Spread aoe marker on some players, not all
+      id: 'P2N Tainted Flood',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6808', source: 'Hippokampos', capture: false }),
+      response: Responses.spread(),
+    },
+    {
+      // Drops aoe zones beneath you -> run to dodge (on everyone)
+      // Not working currently properly -> On you not working.
+      id: 'P2N Sewage Erruption',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '680D', source: 'Hippokampos', capture: false }),
+      response: Responses.spread(),
+    },
+    {
+      // Shortly After Sewage Erruption, not every Sewage Erruption but every Predatory Sight, maybe one trigger for both
+      id: 'P2N Predatory Sight',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '680A', source: 'Hippokampos', capture: false }),
+      response: Responses.doritoStack(),
+    },
+    {
+      // Raidwide knockback -> dont get knocked into slurry
+      id: 'P2N Shockwave',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6807', source: 'Hippokampos', capture: false }),
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Knockback stay on grid',
+        },
+      },
+    },
+    {
+      // Aoe from outside the arena
+      id: 'P2N Dissociation',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6806', source: 'Hippokampos', capture: true }),
+      delaySeconds: 1,
+      promise: async (data) => {
+        const callData = await callOverlayHandler({
+          call: 'getCombatants',
+        });
+        if (callData.combatants)
+          console.log(callData.combatants);
+        if (!callData || !callData.combatants || !callData.combatants.length) {
+          console.error('SpokenCataract: failed to get combatants: ${JSON.stringify(callData)}');
+          return;
+        }
+        // This is the real hippo, according to hp.
+        const hippos = callData.combatants.filter((c) => c.BNpcID === 14441);
+        if (hippos.length !== 1) {
+          console.error('SpokenCataract: There are more than one Hippo?!?: ${JSON.stringify(hippos)}');
+          return;
+        }
+        data.bodyActor = hippos[0];
+      },
+      alertText: (data, matches, _output) => {
+        // return 'body' + data.bodyActor?.Heading;
+        if (!data.bodyActor?.PosX) {
+          console.error('SpokenCataract: No boss actor found. Did the promise fail?');
+          return;
+        }
+        const xcord = data.bodyActor?.PosX;
+        let newHead = null;
+
+        if (xcord === 110)
+          newHead = 'east';
+        if (xcord === 90)
+          newHead = 'west';
+        if (matches.id === '6806') {
+          if (newHead === 'east')
+            return 'West';
+          if (newHead === 'west')
+            return 'East';
         }
       },
       outputStrings: {
--- a/ui/raidboss/data/06-ew/raid/p2n.ts
+++ b/ui/raidboss/data/06-ew/raid/p2n.ts
@@ -5,7 +5,9 @@ import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  bodyHeading?: string;
+}
 
 const triggerSet: TriggerSet<Data> = {
   // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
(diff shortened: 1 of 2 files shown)


amki:
Everything should be working correctly as far as I know and this should be ready for inclusion if there aren't any style blunders. :3

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,206 @@
...
+        if (matches.id === '67F8') {
+          if (bodyHeading === 'north')
+            return 'body north, head south';
+          if (bodyHeading === 'west')
+            return 'body west, head east';
+          if (bodyHeading === 'south')
+            return 'body south, head north';
+          if (bodyHeading === 'east')
+            return 'body east, head west';
+        }
```
JLGarber, line 79: @quisquous would need to have final say here, but rather than an If Tower, what about something like
```javascript
return {
  north: 'body north, head south',
  west: 'body west, head east',
  south: 'body south, head north',
  east: 'body east, head west',
}[bodyHeading];
```
Both for vertical real estate and for closer association between heading and output?

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,206 @@
...
+      // Spread aoe marker on some players, not all
+      id: 'P2N Tainted Flood',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6809', source: 'Hippokampos', capture: true }),
+      condition: (data, matches) => matches.target === data.me,
+      response: Responses.spread(),
+    },
+    {
+      // Drops aoe zones beneath you -> run to dodge (on everyone)
+      // Not working currently properly -> On you not working.
```
JLGarber: I'm not completely sure I understand what this means. In my logs, it looks like Hippokampus casts Sewage Eruption on itself, and then casts it at unknown entities on the floor under players, not *at* players. Do you have a way to determine which players it is targeting even so? (I ask because that would be useful elsewhere.)

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,206 @@
...
+      netRegex: NetRegexes.startsUsing({ id: '680D', source: 'Hippokampos', capture: false }),
+      suppressSeconds: 5,
+      response: Responses.spread(),
+    },
+    {
+      // Shortly After Sewage Erruption, not every Sewage Erruption but every Predatory Sight, maybe one trigger for both
+      id: 'P2N Predatory Sight',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '680A', source: 'Hippokampos', capture: false }),
+      response: Responses.doritoStack(),
```
JLGarber, line 144: It's probably best just to leave them separated, unless there were specific combinations you saw where it could be more difficult to handle them together.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,206 @@
...
+      },
+    },
+    {
+      id: 'P2N Sewage Deluge',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '67F6', source: 'Hippokampos', capture: false }),
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Aoe get on grid',
```
JLGarber, line 119:
```suggestion
          en: 'Aoe--Get on grid',
```

Minor style nit, feel free to ignore.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,206 @@
...
+    },
+    {
+      // Raidwide knockback -> dont get knocked into slurry
+      id: 'P2N Shockwave',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6807', source: 'Hippokampos', capture: false }),
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Knockback stay on grid',
```
JLGarber, line 154:
```suggestion
          en: 'Knockback--Stay on grid',
```

Same as the other, feel free to ignore.

JLGarber (commented):
Thanks so much for putting these all together. I really wasn't looking forward to the triggers for this one, so I appreciate you doing the hard part.

Are you moving to P3 next? I was going to start in on that this week, but if you're going to do triggers for all of Pandaemonium, I'll leave that all to you and stick with timelines.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,206 @@
...
+    {
+      id: 'P2N Murky Depths',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '680F', source: 'Hippokampos', capture: false }),
+      response: Responses.aoe(),
+    },
+    {
+      id: 'P2N Doubled Impact',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '680E', source: 'Hippokampos', capture: true }),
```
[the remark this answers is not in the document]
amki: Oh I didn't know that. I'll change that.

ui/raidboss/data/06-ew/raid/p2n.ts
amki: This is an old version, the handling has since changed dramatically :3

ui/raidboss/data/06-ew/raid/p2n.ts
amki: This has also been fixed in a later version

xziino: Fixed in the latest version!

ui/raidboss/data/06-ew/raid/p2n.ts
amki: This as well

xziino:
Yup, plan is to do all the triggers for Pandaemonium and we are currently working on p3.  Would be really nice to have timelines soon as well!
In case we bump into any major probelms we let you know (in case of no further progress)

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in 2fe82019d5146ac7edffbd41251b7b6aaca4113d.

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in 2fe82019d5146ac7edffbd41251b7b6aaca4113d.

ui/raidboss/data/06-ew/raid/p2n.ts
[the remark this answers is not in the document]
amki: Fixed in 2fe82019d5146ac7edffbd41251b7b6aaca4113d.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+      id: 'P2N Dissociation',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6806', source: 'Hippokampos' }),
+      alertText: (_data, matches, output) => {
+        const xcord = parseFloat(matches.x);
+        if (xcord === 110)
+          return output.w!();
+        if (xcord === 90)
+          return output.e!();
+      },
```
JLGarber, line 169: Ah, this is nice. Generally it's best to avoid using `getCombatants` where possible, since it's not replayable in the emulator, but I wasn't sure whether there was nonsense with actor movements between the `startsUsing` and `ability` lines that would require it here. Glad this is workable!

amki: I started out like that because the log line's positions for Spoken Cataract are completely wrong but turns out in this case the log position is fine.

I guess it would be necessary to find the npc that will later cast the spell to call it out earlier but I've testetd that there is ample time to adjust after the current call.

JLGarber:
Excellent, I'll just start right in on timelines then.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+import { TriggerSet } from '../../../../../types/trigger';
+
+export interface Data extends RaidbossData {
+  bodyActor?: PluginCombatantState;
+}
+
+const triggerSet: TriggerSet<Data> = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  zoneId: ZoneId.AsphodelosTheSecondCircle,
+  timelineFile: 'p2n.txt',
```
quisquous, line 17:
```suggestion
```

If there's no timeline file, can you remove this line? Or add a blank file?

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+import { RaidbossData } from '../../../../../types/data';
+import { PluginCombatantState } from '../../../../../types/event';
+import { TriggerSet } from '../../../../../types/trigger';
+
+export interface Data extends RaidbossData {
+  bodyActor?: PluginCombatantState;
+}
+
+const triggerSet: TriggerSet<Data> = {
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
quisquous, line 15: Rather than an eslint disable, I think you're missing a ZoneId import here.  Look at other trigger files for examples.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+        },
+      },
+    },
+    {
+      // Aoe from head outside the arena
+      id: 'P2N Dissociation',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6806', source: 'Hippokampos' }),
+      alertText: (_data, matches, output) => {
+        const xcord = parseFloat(matches.x);
```
quisquous, line 164: Rather than using exact coordinates, can you use a comparison? If the center is 100, compare > 100 and < 100, rather than exactly equal to 110 or 90.  I think in general I would like to avoid exact positional comparisons in case for some reason the head gets placed at 109.99 or 90.0001 or something.

Slight naming nit, but call this `xCoord` and not `xcord`?

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+          return;
+        }
+        data.bodyActor = hippos[0];
+      },
+      alertText: (data, matches, output) => {
+        if (!data.bodyActor) {
+          console.error('SpokenCataract: No boss actor found. Did the promise fail?');
+          return;
+        }
+        const heading = data.bodyActor?.Heading;
```
quisquous, line 58: This is a minor comment, but do you need the `?` here if you've already checked `data.bodyActor` above? If typescript is complaining, can you store this in its own variable, e.g. `const bodyActor = data.bodyActor` and then say `if (!bodyActor) etc` so that you don't need this? I think logically it's not possible for `data.bodyActor` to be undefined here, but it makes it harder to read if it looks like it might.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+        let bodyHeading = null;
+
+        if (heading < -3.0 || heading > 3.0)
```
quisquous, line 61: Rather than an if-else tower, could you do something like https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.ts#L255?

In other words something more like:
```
// N = 0, E = 1, etc
const heading =  Math.round(2 - 2 * parseFloat(data.bodyActor.Heading) / Math.PI) % 4;
```

The reasoning here is (1) I think this is a lot more condensed, (2) probably this sort of thing should be a utility function because we've used something like it enough and so using the same formula will make it easier to consolidate them later and (3) I think it's better to just round fully to closest direction in case things are slightly off more than the 0.1 radians that you've given a tolerance for.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+    {
+      id: 'P2N Doubled Impact',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '680E', source: 'Hippokampos' }),
+      response: Responses.sharedTankBuster(),
+    },
+    {
+      id: 'P2N Spoken Cataract',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: ['67F8', '67F7', '67F9'], source: 'Hippokampos' }),
```
quisquous, line 34: I'm a little surprised that there are only three here.  Is there a missing one for the other direction?

https://xivapi.com/search?indexes=action&string=Spoken%20Cataract&columns=Name,ID&pretty=true says there are five ability ids here, which I might expect if there's four directions for self-casting and one for damage (67FD).

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -0,0 +1,178 @@
...
+        if (heading < -1.56 && heading > -1.58)
+          bodyHeading = 'west';
+        if (heading > -0.1 && heading < 0.1)
+          bodyHeading = 'south';
+        if (heading > 1.56 && heading < 1.58)
+          bodyHeading = 'east';
+
+        if (matches.id === '67F8') {
+          if (bodyHeading === 'north')
+            return output.nw!() + '/' + output.ne!();
```
quisquous, line 72: I should probably add this to the contributing/raidboss docs, but even though it's technically valid to return a string here, as a general rule I try not to do any string concatenation of outputs and instead use the output strings themselves.  For example, Hydaelyn uses a `combo` output string here: https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/06-ew/trial/hydaelyn-ex.ts#L73 that takes in the two parameters of the mechanics.  The reasoning here is that it then gives users full flexiblity to override the output in the UI.

Another alternative here is to say "North" and "North Intercards" (or "North Corners")? I find that a little bit easier to parse than NW/NE, but maybe you feel differently?

quisquous (commented):
Thanks so much y'all for these.  I really appreciate all the contributions.  This one is definitely more complicated that p1n, so I have a few more comments here.

ui/raidboss/data/06-ew/raid/p2n.ts
amki: 67FD is the one the body casts, I ask OverlayPlugin for the bosses position already so I don't care about the 67FD cast. (The heading value of the log line is wrong because of timing)

The head can never be aligned with the boss but only left/back/right hence 3 IDs.

I could add 67FD to the trigger but the execution would never yield any information and thus be ignored. (Think about the CPU cycles)

ui/raidboss/data/06-ew/raid/p2n.ts
amki: I saw that line from hydaelin-ex actually but I have no idea how it works. Maybe my JavaScript fu is too weak, I'll try and look into it.

I guess intercards or corners would indeed be better but I just didn't think of it. I'll change it once I find out what "..." does and how these objects end up inside each other?!?

ui/raidboss/data/06-ew/raid/p2n.ts
amki: My geometry is (also...) a little weak so I haven't come up with such an elegant formula but if you say it works it is much better indeed.

> I think it's better to just round fully to closest direction in case things are slightly off more than the 0.1 radians that you've given a tolerance for.

Yes and no, had I started with this it would have been an even harder to debug error that boss headings were off 90° sometimes (because closest at time of log line emission wasn't always closest to actual spell direction) while they are fine the rest of the time.

ui/raidboss/data/06-ew/raid/p2n.ts
quisquous: > The head can never be aligned with the boss but only left/back/right hence 3 IDs.

Hmm, ok.  The fifth id is 0x67FE, but I don't see 0x67FE in my logs at all, so maybe that's unused/savage.

> (Think about the CPU cycles)

:joy:

ui/raidboss/data/06-ew/raid/p2n.ts
quisquous: > My geometry is (also...) a little weak so I haven't come up with such an elegant formula but if you say it works it is much better indeed.

You don't have to trust me! You can put in some numbers into a javascript console and try out the formula and see if it would give you the same answers you already have.

> closest at time of log line emission wasn't always closest to actual spell direction

Yeah, log lines sometimes have stale positions.  It's very frustrating.

ui/raidboss/data/06-ew/raid/p2n.ts
quisquous: `...` is "spread syntax": https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax.  I was just using it so that I could have a couple of different sets of outputstrings that built on top of each other.

I don't think the spread syntax is important here.  A better example than Hydaelyn is: https://github.com/quisquous/cactbot/blob/38e1c8ca5c6422be44bf831b38a86f9a1f5937c7/ui/raidboss/data/03-hw/trial/sophia-ex.ts#L416-L439 where you need one output string to be the combining version and can then pass in the parameters when you call `output.multiple({ dir1: etc, dir2: etc })`.

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in a8562bbece06d2fd3a33b1cbac9473feb52ee86b

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in a8562bbece06d2fd3a33b1cbac9473feb52ee86b

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in a8562bbece06d2fd3a33b1cbac9473feb52ee86b

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in a8562bbece06d2fd3a33b1cbac9473feb52ee86b

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Fixed in a8562bbece06d2fd3a33b1cbac9473feb52ee86b

ui/raidboss/data/06-ew/raid/p2n.ts
amki: Changed output to corner style in a8562bbece06d2fd3a33b1cbac9473feb52ee86b

quisquous:
Thank you!!!

PR #3920, opened 2021-12-24
raidboss: slight Panda normal adjustments

quisquous:


ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -132,13 +132,7 @@ const triggerSet: TriggerSet<Data> = {
...
       netRegexFr: NetRegexes.startsUsing({ id: '67F6', source: 'Hippokampos', capture: false }),
       netRegexJa: NetRegexes.startsUsing({ id: '67F6', source: 'ヒッポカムポス', capture: false }),
-      alertText: (_data, _matches, output) => output.text!(),
-      outputStrings: {
-        text: {
-          en: 'Aoe--Get on grid',
-          de: 'AoE--Geh auf die Gitter',
-        },
-      },
+      response: Responses.aoe(),
```
quisquous, line 135: I think the "get on grid" part is a little bit much.  There's a ground aoe many seconds later that indicates that you should stand on the grid (or the corners that aren't grids), and so I think just saying "aoe" is enough here.

ui/raidboss/data/06-ew/raid/p2n.ts
```diff
@@ -180,13 +174,7 @@ const triggerSet: TriggerSet<Data> = {
...
       netRegexFr: NetRegexes.startsUsing({ id: '6807', source: 'Hippokampos', capture: false }),
       netRegexJa: NetRegexes.startsUsing({ id: '6807', source: 'ヒッポカムポス', capture: false }),
-      alertText: (_data, _matches, output) => output.text!(),
-      outputStrings: {
-        text: {
-          en: 'Knockback--Stay on grid',
-          de: 'Rückstoß--Bleib auf den Gittern stehen',
-        },
-      },
+      response: Responses.knockback(),
```
quisquous, line 177: Similarly, I think "don't get knocked into bad stuff" felt fairly obvious even the first time through, and I think it'd be better to not complicate this.  I don't think we'd say "knockback--don't get hit into spikes" for p4n.

ui/raidboss/data/06-ew/raid/p3n.ts
```diff
@@ -14,21 +14,13 @@ const triggerSet: TriggerSet<Data> = {
...
-      response: Responses.getInThenOut(),
-    },
-    {
-      id: 'P3N Experimental Fireplume 2',
-      type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: '6696', source: 'Phoinix', capture: false }),
-      netRegexDe: NetRegexes.startsUsing({ id: '6696', source: 'Phoinix', capture: false }),
-      netRegexFr: NetRegexes.startsUsing({ id: '6696', source: 'Protophénix', capture: false }),
-      netRegexJa: NetRegexes.startsUsing({ id: '6696', source: 'フェネクス', capture: false }),
+      // This is Experimental Fireplume (6696) into Fireplume (6697), which starts casting 5s later.
```
quisquous, line 19: 6696 is 11 seconds ahead of time.  6697 is 6 seconds ahead (which is maybe long too, but better).

ui/raidboss/data/06-ew/raid/p3n.ts
```diff
@@ -14,21 +14,13 @@ const triggerSet: TriggerSet<Data> = {
...
   triggers: [
     {
-      id: 'P3N Experimental Fireplume 1',
+      id: 'P3N Experimental Fireplume Out',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: '6698', source: 'Phoinix', capture: false }),
-      netRegexDe: NetRegexes.startsUsing({ id: '6698', source: 'Phoinix', capture: false }),
-      netRegexFr: NetRegexes.startsUsing({ id: '6698', source: 'Protophénix', capture: false }),
-      netRegexJa: NetRegexes.startsUsing({ id: '6698', source: 'フェネクス', capture: false }),
-      response: Responses.getInThenOut(),
```
quisquous, line 23: I think "in then out" is technically true, but is not a raid call that made much sense to me.  It's really more like "get middle and then run out to where the third explosion was while dodging the other outside explosions".

I think this is just a "dodge visible telegraphs that don't need a trigger" sort of mechanic, and so suggesting that we remove it.

ui/raidboss/data/06-ew/raid/p3n.ts
JLGarber: Given the size of the ground effect, I don't think it's unreasonable to have a 6-second warning. From the center, it's more than a GCD to get to safety, and for casters it can be impossible to get out if they have less time.

ui/raidboss/data/06-ew/raid/p3n.ts
JLGarber: It's kind of unfortunate Optical Sight isn't a more common mechanic. I kind of feel like this does warrant a trigger just based on how fast the ground effects resolve, but I'm not totally sure how to word it, and I think `inThenOut` doesn't best convey what should happen. I don't know. It's a tricky question!

quisquous: Yeah, it kind of feels like one of those "this isn't something you can explain in the moment" sort of situations.

ui/raidboss/data/06-ew/raid/p2n.ts
Legends0: Arms length / surecast is up during these casts so just kb is fine.

ui/raidboss/data/06-ew/raid/p3n.ts
Legends0: I prefer the early call to slide cast into position.

ui/raidboss/data/06-ew/raid/p3n.ts
Legends0: I'm okay with its removal. The benefit of keeping it and/or just setting get middle/start middle is that the other version of get out is called where as here you would use silence or visually looking at the boss to tell what's upcoming.

ui/raidboss/data/06-ew/raid/p3n.ts
quisquous: Ok, then maybe "Get out soon" instead of "Get out" which feels more immediate?

ui/raidboss/data/06-ew/raid/p3n.ts
quisquous: "Start Middle" seems reasonable.

ui/raidboss/data/06-ew/raid/p3n.ts
Legends0: Get out soon, although longer, would feel better. May want to have it leave the text up for extra time as it is early?

Legends0: I see it is "Out of middle soon" vs Get out soon", so the get out being shorter would be preferred.

quisquous: I think of "get out" as "get away from the boss" in this case it is "get away from the middle of the room" which feels different to me.  I added a duration, thanks for that suggestion.

quisquous:
Thanks for all the comments and feedback!

PR #3946, opened 2021-12-30
jobs: AST: Minor Arcana support and Sign improve

Echoring:
- Fix a mixture of Arcana and Minor Arcana in `Plugin`. This will fix `cardBox` error when you held a minor arcana. 
- Support Minor Arcana analysis in memory.
- Split EW AST from 5.x component.
- Change charactor to symbol `○` `×`, for `O` is too similar to `0`. (I'm not sure how it looks like in other language's PC?)
- Change Sign count box to show Minor Arcana. `↑` refers to Lord, and `＋` refers to Lady. (Is there any better symbols like sword and wings?)
- Add a stack-like gauge to show Sign. (When filling from left to right, users are more likely to feel full, but in-game gauge is from right to left)
ast

### DIFF
--- a/plugin/CactbotEventSource/FFXIVProcessIntl.cs
+++ b/plugin/CactbotEventSource/FFXIVProcessIntl.cs
@@ -718,6 +718,7 @@ namespace Cactbot {
         Celestial = 3,
       }
 
+      [NonSerialized]
       [FieldOffset(0x04)]
       private byte _heldCard;
 
--- a/plugin/CactbotEventSource/FFXIVProcessIntl.cs
+++ b/plugin/CactbotEventSource/FFXIVProcessIntl.cs
@@ -707,6 +707,8 @@ namespace Cactbot {
         Spear = 4,
         Ewer = 5,
         Spire = 6,
+        Lord = 0x70,
+        Lady = 0x80,
       }
 
       public enum Arcanum : byte {
@@ -717,7 +719,7 @@ namespace Cactbot {
       }
 
       [FieldOffset(0x04)]
-      private Card _heldCard;
+      private byte _heldCard;
 
       [NonSerialized]
       [FieldOffset(0x05)]
@@ -733,7 +735,13 @@ namespace Cactbot {
 
       public string heldCard {
         get {
-          return _heldCard.ToString();
+          return ((Card)(_heldCard & 0xF)).ToString();
+        }
+      }
+
+      public string crownCard {
+        get {
+          return ((Card)(_heldCard & 0xF0)).ToString();
         }
       }
 
--- a/ui/jobs/components/ast.ts
+++ b/ui/jobs/components/ast.ts
@@ -62,9 +62,9 @@ export class AST5xComponent extends BaseComponent {
     if (card in cardsMap)
       cardParent.classList.add(cardsMap[card].bonus);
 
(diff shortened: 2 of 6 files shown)


ui/jobs/components/ast.ts
```diff
@@ -62,34 +62,148 @@ export class ASTComponent extends BaseComponent {
...
-      else
-        this.sealBox.parentNode.classList.remove('ready');
+    // Show how many kind of seals you already have
+    // Turn green when you have all 3 kinds of seal
+    const sealCount = new Set(seals).size;
+    this.sealBox.innerText = sealCount.toString();
+    if (sealCount === 3)
+      this.sealBox.parentNode.classList.add('ready');
+    else
+      this.sealBox.parentNode.classList.remove('ready');
```
MaikoTan, line 82: This part can be simplified with `classList.toggle`

ui/jobs/components/ast.ts
```diff
@@ -62,34 +62,148 @@ export class ASTComponent extends BaseComponent {
...
+    else if (sign.includes(cardsMap[card].seal))
+      this.cardBox.innerText = '×';
+    else
+      this.cardBox.innerText = '○';
+
+    for (let i = 0; i < 3; ++i) {
+      this.signs[i]?.classList.remove('Solar', 'Lunar', 'Celestial');
+      const asign = sign[i];
+      if (asign)
+        this.signs[i]?.classList.add(asign);
```
MaikoTan, line 206:
```suggestion
    this.signs.forEach((elem, i) => {
      elem.classList.remove('Solar', 'Lunar', 'Celestial');
      const asign = sign[i];
      if (asign)
        elem.classList.add(asign);
    });
```
so no more `this.signs?.`

ui/jobs/components/ast.ts
```diff
@@ -62,34 +62,148 @@ export class ASTComponent extends BaseComponent {
...
+    // Minor Arcana, ↑ = lord, + = lady
+    const minor = jobDetail.crownCard;
+    this.minorBox.parentNode.classList.toggle('lord', minor === 'Lord');
+    this.minorBox.parentNode.classList.toggle('lady', minor === 'Lady');
+    if (minor === 'Lord')
+      this.minorBox.innerText = '↑';
+    else if (minor === 'Lady')
+      this.minorBox.innerText = '＋';
+    else
+      this.minorBox.innerText = '';
```
MaikoTan, line 181: This might be simplified as this better.

```suggestion
    this.minorBox.innerText = minorMap[minor] ?? '';
```

and add this to the top:
```ts
const minorMap = {
  'Lord': '↑',
  'Lady': '＋',
}
```

MaikoTan (approved):
some nits

plugin/CactbotEventSource/FFXIVProcessIntl.cs
```diff
@@ -717,7 +719,7 @@ public enum Arcanum : byte {
       }
 
       [FieldOffset(0x04)]
-      private Card _heldCard;
+      private byte _heldCard;
```
quisquous, line 722: Should this be `[NonSerialized]`?

Echoring: I'm not sure how `[NonSerialized]` works, but seems you are right.

PR #3962, opened 2022-01-03
raidboss: improve zenos protean text

jacob-keller:
- raidboss: trigger on gain doom, not losing doom
- raidboss: improve Zenos "protean" tell

### DIFF
--- a/ui/raidboss/data/04-sb/dungeon/ala_mhigo.ts
+++ b/ui/raidboss/data/04-sb/dungeon/ala_mhigo.ts
@@ -96,7 +96,7 @@ const triggerSet: TriggerSet<Data> = {
       alertText: (_data, _matches, output) => output.text!(),
       outputStrings: {
         text: {
-          en: 'Protean',
+          en: 'Protean (Spread Out)',
           de: 'Um den Boss verteilen',
           fr: 'Changement',
           ja: 'ボスを基準として散開',
--- a/ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
+++ b/ui/raidboss/data/02-arr/alliance/the_world_of_darkness.ts
@@ -87,8 +87,8 @@ const triggerSet: TriggerSet<Data> = {
     },
     {
       id: 'Angra Mainyu Gain Doom',
-      type: 'LosesEffect',
-      netRegex: NetRegexes.losesEffect({ effectId: 'd2' }),
+      type: 'GainsEffect',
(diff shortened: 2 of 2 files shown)


ui/raidboss/data/04-sb/dungeon/ala_mhigo.ts
```diff
@@ -96,7 +96,7 @@ const triggerSet: TriggerSet<Data> = {
       alertText: (_data, _matches, output) => output.text!(),
       outputStrings: {
         text: {
-          en: 'Protean',
+          en: 'Protean (Spread Out)',
```
quisquous, line 99: Protean means something different to me than spread.  I understand that not everybody knows what this means, but I think it's better to call things consistently across multiple fights.  Folks can change it in the cactbot ui if they don't like it.

Jessidhia: “Protean”, however, is meaningless to people who didn’t raid Alex…

jacob-keller:
That was also my argument. If you don't participate in specific savage
raids from which the terminology arose it makes no sense and left myself
and a friend confused.

I agree that spread out isn't really correct either though.

ui/raidboss/data/04-sb/dungeon/ala_mhigo.ts
quisquous: I hear you, but I think there's not a better word for "spread out, but in relation to your angle to the boss" vs "spread out, in relation to each other" than "protean".  I think they are different and it's better to differentiate what the mechanic actually is and have a little confusion, and if folks want to call it something different then the they can use the cactbot ui to change it.

jacob-keller:
@quisquous on a related note, during the level 87 dungeon there is an attack on the final boss which is probably mis-labeled as "spread" when its a protean (line attak originating from the boss targeting every player).

This further adds evidence that we're inconsistent. I personally feel that a compromise like "protean spread" or similar would be easier to understand for new players and include the term to differentiate it...

quisquous:
I think ultimately my feeling is that communication is hard, and that sometimes the first time you see a mechanic it might be confusing what cactbot is asking you to do.  I think it's better to use consistent language everywhere and then the next time people see "protean" they'll know what it's asking them to do, and if they don't like it, then they can change it in the cactbot config.

Re: 87 boss, sure we could change that to protean, although it's kind of more of a "protean laser" vs the typical "protean (cone)" since it has a width that a normal protean doesn't have.

PR #3965, opened 2022-01-04
raidboss: fix o7n ultros typo

quisquous:


JLGarber (approved):
Definitely a typo, [according to XIVAPI](https://xivapi.com/search?indexes=status&string=ultros+simulation&columns=Name,Name_de,Name_fr,Name_ja,ID).

PR #3966, opened 2022-01-04
raidboss: Tower of Babil--Add Barnabas timeline fork

JLGarber:
When I generated this timeline, every single one of the 15+ attempts I did had Scrapline first. Naturally it was only this week that I saw the other possibility. (Quisquous had sent me an example a while ago, but I forgot about it when I was finishing stuff up.)

Tested against More Than One log, but not in-game due to maintenance.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/the_tower_of_babil.txt
+++ b/ui/raidboss/data/06-ew/dungeon/the_tower_of_babil.txt
@@ -12,31 +12,75 @@ hideall "--sync--"
 
 # -ii 6076 61DE
 
+# Common opener
 0.0 "--sync--" sync / 00:0839::Magitek Servicing will be sealed off/ window 0,1
-8.9 "Ground and Pound" sync / 1[56]:[^:]*:Barnabas:6247:/ window 8.9,5
+5.7 "--sync--" sync / 14:[^:]*:Barnabas:6247:/ window 5.7,5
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/dungeon/the_tower_of_babil.txt
```diff
@@ -12,31 +12,75 @@ hideall "--sync--"
...
+332.4 "Electromagnetic Release" sync / 1[56]:[^:]*:Barnabas:62F1:/
+
+342.7 "Thundercall" sync / 1[56]:[^:]*:Barnabas:62ED:/ window 42.7,5 jump 400.0
+349.2 "Rolling Scrapline"
+349.5 "Shock"
+359.8 "Shocking Force"
+368.4 "Ground and Pound 1"
+374.4 "Ground and Pound 2"
+
+# Common Thhundercall block, looping back to one of the two blocks.
```
quisquous, line 68:
```suggestion
# Common Thundercall block, looping back to one of the two blocks.
```

Does it always alternate after that, or is it totally random?

JLGarber: It looks like it's a strict alternation of Scrapline --> Pound or Pound --> Scrapline *within the magnet block, but I think the timeline as it's written should handle any possibility?

quisquous: :+1:

PR #3986, opened 2022-01-07
raidboss: p3s timings for add phases

Legends0:
Based on some of my logs.


Legends0:
Could probably smoosh the gaps so to include the timings after the enrages. It might be that even if we clear the adds, we have to still wait.

Legends0:
A note regarding Experimental Fireplumes for the timeline. I would like some timing for the 66BE's fireplume as it's helpful to know when to run back in or out. The trigger for the get out is good. For the 66C0's Fireplumes, I find running out on the 2nd one, towards the second one to be the better timing.

Legends0:
Brightened Fire takes some getting used to, I'd probably have it call out a swap for players +4 over the current Brightened Fire. There is some leeway for this movement. I find having to mentally count each number as it impacts.

quisquous:
Before this gets merged, can you undo your changes to the german translations?

> Could probably smoosh the gaps so to include the timings after the enrages. It might be that even if we clear the adds early, we have to still wait.

I think it's fine to have jumps here.

> A note regarding Experimental Fireplumes for the timeline. I would like some timing for the 66BE's fireplume as it's helpful to know when to run back in or out. The trigger for the get out is good. For the 66C0's Fireplumes, I find running out on the 2nd one, towards the second one to be the better timing.

Agreed that the "out" needs a timeline line.  I left a comment about the timing but never added them in like they are in normal mode.

I am not sure what you're suggesting about the "rotate" version.  I run either towards the first or second explosion and either seem fine.

> Brightened Fire takes some getting used to, I'd probably have it call out a swap for players +4 over the current Brightened Fire. There is some leeway for this movement. I find having to mentally count each number as it impacts.

I am not sure what you're suggesting here either.  For prog we just stacked everybody (sorry healers) and had everybody not move.  I think "swap" is maybe strat specific.  I could see it calling out 1 2 3 4 5 6 7 8 with info text if you really wanted.

quisquous:
Oh, I see what happened.  You removed the duplicate, but I didn't realize it was a duplicate.  Then I merged an actual removal, and now it's conflicting.  Sorry.

Legends0:
> I am not sure what you're suggesting here either. For prog we just stacked everybody (sorry healers) and had everybody not move. I think "swap" is maybe strat specific. I could see it calling out 1 2 3 4 5 6 7 8 with info text if you really wanted.

My group was seeing 1hkos when we tried stacking the pairs, theoretically the debuff should fall off before the second hit but I guess there may have been some delay.

quisquous:
> My group was seeing 1hkos when we tried stacking the pairs, theoretically the debuff should fall off before the second hit but I guess there may have been some delay.

Yeah, it should fall off just in time if you everybody only gets hit by the correct two.

PR #3987, opened 2022-01-07
i18n: added and correction of translation P1s-p3s

VanSlanzar:
There was a large duplication of German translations. I fixed this and corrected the translation which was in Japanese (De and Fr).


mooondark (approved):

quisquous:
Thanks for the fixes and the translations!

PR #3999, opened 2022-01-08
raidboss: fix p2s dissociation callout

Legends0:
In my previous testing it was outputting East when it was West, but based on other findings it was actually outting always East as it outputs West always without this change. 682D is the unknown spellid that always occurs after 682C. I believe it is tied to the head here and may yield better results.


Legends0:
On second thought, that may be the right coords, but could be calling out on the hit.

elsorume:
My log file does not contain StartCasting lines with id 682D, only ActionEffect lines with this id. But I found StartCasting lines with id 682E. It also always occurs after 682C, but has x coord 90 or 110. Seems like this is what you need.

quisquous:
I think I might put up a PR to remove this for now, and then when this is tested in game, we can merge this? What do you think?

Legends0:
> I think I might put up a PR to remove this for now, and then when this is tested in game, we can merge this? What do you think?

The timeline probably needs updated to with the timing for this action. That sounds good. It doesn't appear in fflogs. I think I need to change the netregex as well.

panicstevenson:
The current iteration at the time of writing seems to work properly for me

quisquous:
Thanks for the fix (and thanks for chiming in that it worked, Panic!)

PR #4005, opened 2022-01-09
raidboss: p3s fledgling flights and death's toll triggers

Legends0:
I did not see a 006F marker and this is assuming death's toll markers are the same ids. I got the id's using the function provided and comparing to video + log I had for directions.

This solves Death's Toll rather than just stating the count. ~~There is also the possibility of merging the two Death's Toll triggers by having the condition be in the output and using the log of a gainsEffect to set the boolean.~~ EDIT: merged


Legends0:
Hope the delay there is adequate, it could be sooner but don't want to have people move too early from placing the eyes.

Legends0:
Both Fledgling callouts work in this iteration, but the death's toll debuff callout does not. I am guessing something is wrong in the infotext.

Legends0:
Seems to all work when I ran it tonight.

Legends0:
Trigger does fire after the event ends as well, so that will need to be fixed.

Legends0:
I think I worked out a fix.

It is possible to miss the effect such as being dead and then still need to place an Ashen Eye as there is time to be rezzed after the debuffs go out. Their sacrifice will not be forgotten.

quisquous:
Sorry, I am short on time, I'll try to take a look at this tomorrow.

Legends0:
My attempt at fixing the extra call broke it.

ui/raidboss/data/06-ew/raid/p3s.ts
```diff
@@ -375,6 +383,35 @@ const triggerSet: TriggerSet<Data> = {
...
+            '01': output.outCardinals!(),
+            '02': output.outIntercards!(),
+            '04': output.middle!(),
+          }[matches.count];
+        }
+      },
+      outputStrings: {
+        middle: Outputs.middle,
+        outIntercards: {
+          en: 'Out + Intercards',
```
quisquous, line 408: Suggesting to put intercards first, because otherwise "Out + Intercards" and "Out + Cardinals" read very similarly.

```suggestion
          en: 'Intercards + Out',
```

Alternatively, I've also seen people do cardinals and inside (but not middle) for 2s, so I wonder a little bit if this should just call 1, 2, 4 and folks can use the UI to set a more specific string if they want? What do you think.

quisquous:
I gave this a test tonight and everything looked good, just a small wording question.

ui/raidboss/data/06-ew/raid/p3s.ts
Legends0: I never thought about using the UI, namely because I have not used it for the most part. On one hand just numbers makes it easier for a person to translate to something else.

PR #4006, opened 2022-01-09
raidboss: fix EW hunts wrong ZoneId

### LINKED ISSUE #3994: RAIDBOSS: Ultima Thule Triggers set to Elpis incorrectly
## Description

Under RaidBoss -> Endwalker | There are two that is set as Elpis. The second Elpis should be set as Ultima Thule. Looking at the trigger source, it appears that the ZoneId is also set as Elpis, therefore, the calls do not happen when hunting in Ultima Thule.

xiashtra:
Fixes #3994 incorrect zone id for Ultima Thule hunts.


quisquous:
Thanks for the fix!

PR #4007, opened 2022-01-09
resources: Convert Simple HTML Text to InnerText

panicstevenson:
Convert simple text values from using innerHTML to innerText as it has a
significant impact on performance; the values being updated should only
be numbers or percentages and shouldn't need HTML values, even for
potentially custom content.


quisquous:
Thank you for all the investigation on this!

PR #4012, opened 2022-01-10
i18n: p3s/p4s missing translations

Akurosia:


Akurosia:
since i am not that far in the raids and some data is not visible in data, i cant translate following entries:
```
        'Demigod Double': '',
        'Directional Shift': '',
        'Farsight': '',
        'Nearsight': '',
```

PR #4020, opened 2022-01-11
raidboss: smileton fix duplicate knockback call

Legends0:
The knockback call during the first boss is called out more than once simultaneously.

This is caused by the trigger checking for two spellids that go off simultaneously.

673E targets Face
6C5E targets players

suppressSeconds is one fix... an alternative would be to remove one of spellids? Looks like they both are always cast.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/smileton.ts
+++ b/ui/raidboss/data/06-ew/dungeon/smileton.ts
@@ -26,10 +26,10 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'Smileton Face Off My Lawn',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: ['673E'], source: 'Face', capture: false }),
-      netRegexDe: NetRegexes.startsUsing({ id: ['673E'], source: 'Fratze', capture: false }),
-      netRegexFr: NetRegexes.startsUsing({ id: ['673E'], source: 'Visage Imperturbable', capture: false }),
-      netRegexJa: NetRegexes.startsUsing({ id: ['673E'], source: 'フェイス', capture: false }),
+      netRegex: NetRegexes.startsUsing({ id: '673E', source: 'Face', capture: false }),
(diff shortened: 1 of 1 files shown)


Legends0:
The other issue also with this boss is that there is a knockback call same time as get red/blue, which could potentially be resolved by tracking tethers (not sure if in logs), calling the color early or delaying color call.

ui/raidboss/data/06-ew/dungeon/smileton.ts
```diff
@@ -26,10 +26,10 @@ const triggerSet: TriggerSet<Data> = {
...
       id: 'Smileton Face Off My Lawn',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: ['6C5E', '673E'], source: 'Face', capture: false }),
-      netRegexDe: NetRegexes.startsUsing({ id: ['6C5E', '673E'], source: 'Fratze', capture: false }),
-      netRegexFr: NetRegexes.startsUsing({ id: ['6C5E', '673E'], source: 'Visage Imperturbable', capture: false }),
-      netRegexJa: NetRegexes.startsUsing({ id: ['6C5E', '673E'], source: 'フェイス', capture: false }),
+      netRegex: NetRegexes.startsUsing({ id: ['673E'], source: 'Face', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ id: ['673E'], source: 'Fratze', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ id: ['673E'], source: 'Visage Imperturbable', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ id: ['673E'], source: 'フェイス', capture: false }),
```
quisquous, line 32:
```suggestion
      netRegex: NetRegexes.startsUsing({ id: '673E', source: 'Face', capture: false }),
      netRegexDe: NetRegexes.startsUsing({ id: '673E', source: 'Fratze', capture: false }),
      netRegexFr: NetRegexes.startsUsing({ id: '673E', source: 'Visage Imperturbable', capture: false }),
      netRegexJa: NetRegexes.startsUsing({ id: '673E', source: 'フェイス', capture: false }),
```

PR #4023, opened 2022-01-13
raidboss: Fix incorrect response for Violent Discharge

alphaONE2:
Violent Discharge, used by The Big Cheese in Smileton, is a room-wide AoE. This was probably a copy/paste error from Right Disassembler.


quisquous:
Thanks for the fix!

PR #4031, opened 2022-01-16
raidboss: p4s part two headmarkers and safe spot triggers

Legends0:
~~Currently WIP. I may add additional triggers for the acts here.~~

~~Currently this may call out Act 1 and Act's "north/south first" or "east/west first" safe spot during the tethers going out, but relies on combatant order being what determines the position of the thorns which may or may not be static Not yet working, but the maps should be correct..~~ Thorn mapping to be verified in other Locales as it currently depends on them matching the boss name.


Legends0:
Currently these don't work. I am guessing the headmarker that is first in the instance is wrong on top of the combatant numbers are off that it registered into the array as there were no matches.

Legends0:
I'm not sure how we want to handle curtain call. I think the generic requirement is that no back-to-back of the same role will be followed, but not sure if it's possible to have tank/healers at 12 and dps at 6 or some other strange timing strategy. Right now it just does dps tank/healer dps tank/healer etc.

kshman:
I'm tested lastest version. No Safe Sports appears. Thorns Collector seems like not working
image

Below is yesterday version
image

Thank you so much!!!

Legends0:
I somewhat recall there being discussion that the callOverlayHandler stuff doesn't work in raidemulator... but maybe I am misremembering. I don't fully understand what the overlayhandler does other than we can get more up to date information from it regarding combatants than we do in the network logs. I have written with it a few times now, so I kind of know what sort of things it can do and/or things it would be required for.

For the thorns, the alternative approach would be for cactbot to record the addcombatants outside of combat and use that in the next combat (before instance reset), but I am not aware of a feature in cactbot to do that. It may have that option, I just haven't used it.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -653,6 +1161,22 @@ const triggerSet: TriggerSet<Data> = {
...
       },
     },
+    {
+      id: 'P4S Curtain Call Debuffs',
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'AF4', capture: true }),
+      condition: (data) => Conditions.targetIsYou() && data.act === 'curtain',
+      delaySeconds: (data, matches) => {
+        const duration = parseFloat(matches.duration);
+        return (data.role === 'dps' ? duration - 12 : duration - 6);
```
quisquous, line 1171: Hmm, I think this whole trigger is a little strat specific.  I could see a warning at ~5 seconds left to break your tether, because at that point you just have to go or it will explode, but 12 and 6 role-specific seems a little too specific to particular strategies about who goes first.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -512,6 +804,110 @@ const triggerSet: TriggerSet<Data> = {
...
+    },
+    {
+      id: 'P4S Thorns Collector',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6A0C', source: 'Hesperos', capture: false }),
+      netRegexDe: NetRegexes.startsUsing({ id: '6A0C', source: 'Hesperos', capture: false }),
+      netRegexFr: NetRegexes.startsUsing({ id: '6A0C', source: 'Hespéros', capture: false }),
+      netRegexJa: NetRegexes.startsUsing({ id: '6A0C', source: 'ヘスペロス', capture: false }),
+      promise: async (data, _matches, _output) => {
+        // Collect all Hesperos entities up front
```
quisquous, line 834: Should this just be done by `matches.source` instead of having a separate block of names?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -571,6 +1034,51 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: NetRegexes.tether({ id: '00AD', source: 'Hesperos' }),
+      netRegexDe: NetRegexes.tether({ id: '00AD', source: 'Hesperos' }),
+      netRegexFr: NetRegexes.tether({ id: '00AD', source: 'Hespéros' }),
+      netRegexJa: NetRegexes.tether({ id: '00AD', source: 'ヘスペロス' }),
+      condition: (data) => data.act === '3',
+      // Tethers come out East or West (0 seconds), (3s) Middle knockack, (6) Opposite Cardinal
+      suppressSeconds: 7,
+      infoText: (data, matches, output) => {
+        const thorn = (data.thornIds ??= []).indexOf(parseInt(matches.sourceId, 16));
+        const thornMap: { [thorn: number]: string } = {
```
quisquous, line 1049: I'm not sure you need both thornMap and thornMapDirs.  I think you could call `output[thornMapDirs[thorn]]!()` and use `thornMapDirs` to handle both.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -562,6 +958,73 @@ const triggerSet: TriggerSet<Data> = {
...
+    {
+      id: 'P4S Color Headmarkers',
+      type: 'HeadMarker',
+      netRegex: NetRegexes.headMarker({}),
+      condition: Conditions.targetIsYou(),
+      response: (data, matches, output) => {
+        // cactbot-builtin-response
+        output.responseOutputStrings = {
+          purpleTether: {
+            en: 'Purple Tether',
```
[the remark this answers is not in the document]
Legends0: Looks like you're right, and I also could not find a case where the tank would get marked and not the healer.

```
27|9:35:03||Healer 1|0000|E000|016F|
27|9:35:03||Tank 1|0000|E000|016F|
35|9:35:03||Healer 1||Tank 1|E000|0000|00AC|

27|9:35:03||DPS 1|0000|E000|016F|
27|9:35:03||DPS 2|0000|E000|016F|
35|9:35:03||DPS 1||DPS 2|E000|0000|00AC|

27|9:35:03||Healer 2|0000|E000|016D|
35|9:35:03||Healer 2||Tank 2|E000|0000|00AC|

27|9:35:03||DPS 3|0000|E000|016E|
27|9:35:03||DPS 4|0000|E000|016E|
35|9:35:03.9040000-05:00||DPS 3||DPS 4|E000|0000|00AC|
```

So something more would need to be done to give a callout for the purple tether-paired player there.

Some more info:
The purple headmarker goes out during Act 2, Act 4 and Curtain Call. Curtain Call is all purple, and there is the debuff that is more important for the mechanic.
Green headmarker goes out during Act 2 and Finale. The Finale callout one probably isn't necessary as it's all green tethers and I used a different trigger just to state what to do that seems to work fine (may need a durationSeconds added).
Orange headmarker goes out during Act 2.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: Good catch. I was able to do so.

ui/raidboss/data/06-ew/raid/p4s.ts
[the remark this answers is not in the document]
quisquous: The other thing to keep in mind re: headmarkers is that phase 1 will have a different `firstHeadmarker` than phase 2, and so I think there will need to be some logic to handle that.  E12S has the same issue.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: How about a customizable timing with default of 6?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -1147,7 +1147,17 @@ const triggerSet: TriggerSet<Data> = {
       condition: (data) => Conditions.targetIsYou() && data.act === 'curtain',
       delaySeconds: (data, matches) => {
         const duration = parseFloat(matches.duration);
-        return (data.role === 'dps' ? duration - 12 : duration - 6);
+
+        // Check if custom duration was set in options
+        let breakTime = parseFloat(typeof data.options.cactbotp4sCurtainCallBreakTime === 'string' ? data.options.cactbotp4sCurtainCallBreakTime : '6' );
```
quisquous, line 1152: I know there were a couple of options previously, but I think I would really like to move to just "strat specific things belong in user triggers" and not try to support a pile of options.

To me, it would make sense to just have the debuff call out Group 1 / Group 2 / Group 3 / Group 4, and then folks can figure it out from there.

Legends0: I think this can be done. I currently have it waiting to call out until 12s on the debuff and changed to infoText as the player may or may not need to do something right away.

It could call out as soon as debuffs went out (or in addition). I chose not to call out at start because it would duplicate the callout (although we could handle it such as only groups 2-4 or 3/4 get a reminder), and also did not want to leave the text up for too long (42s at the worst).

Legends0: For a break tether callout at 6 for better or worse if a callout were implemented I'd want it to check that the debuff is still on the player so it's not calling it out when the player already broke the tether... would that necessitate OverlayHandler?

ui/raidboss/data/06-ew/raid/p4s.ts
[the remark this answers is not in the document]
Legends0: The first headmarker in phase 2 is that purple headmarker.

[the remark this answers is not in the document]
Legends0: Thinking about it... we could work it out such that we do not need to be know the headmarker's true ID as the headmarker IDs we do see in the logs can be sorted to determine which is which and then recorded for a match later in the fight. The blue one could be predicted during Act 1.

ui/raidboss/data/06-ew/raid/p4s.ts
quisquous: I think it's better to call out right at the beginning so people know.  Partially it's that group 1 needs to be ready to go (they could get an alarm too?), and also that the tanks need to know if they need to tank swap too.  Personally, I think the "when to break" is less necessary and a more damage dependent timing?  I think people will be able to break faster as gear improves too.

I am not sure how to figure out if a tether has broken on a particular player or not.

ui/raidboss/data/06-ew/raid/p4s.ts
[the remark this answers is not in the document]
quisquous: I think the headmarker offset should be the same across the entire zone, so you should be able to determine the true offset from a single log that has a door boss kill in it.  I know this is theoretical, but if possible I'd prefer to use true headmarker ids in the chance that eventually we can figure out a more built in way to calculate correct headmarkers.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: There is a limit to how fast the tethers can be broken due to the vulnerability up that occurs, which we could actually use to possibly indicate a break tether during the groups for the second person to know they could break.

ui/raidboss/data/06-ew/raid/p4s.ts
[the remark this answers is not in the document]
Legends0: Ok, so looking back at my changes, I actually had the right values at one point but the problem I ran into is as you described with E12S issue. For here it's also tricky because there are three headmarkers that go out simultaneously and thus we will probably want to delay to grab all three, sort for one of them to start and go from there.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: Oh I see how we can use that magic vulnerability up even more. We just have a counter that counts them, and just say break tether to the person next in line once the previous is gone.

Legends0: Hmmm, looking at it, I think the call breaks one I wrote may best to just call when group 2, 3, 4 can start than to call their break

Legends0: One note on adding a tank swap... would need to know if the data.role has aggro of the boss which I am not sure at the moment where to get that info.

1unie: Quick question for Curtain Call, how difficult would it be to have a trigger that would call the order of the breaks using party members' names? I feel like that's a trigger that would help groups stay afloat during that mechanic. Or perhaps is it too much info?

Legends0: As it is currently, each group gets there own tailored call out instead of everyone receiving 4 callouts. It's possible to add the names of each player to the groups, but i think this would require an additional trigger for collecting those names and then adding a tiny delay for the collection to complete. I do not see a whole lot of benefit to this unless you were to be running some non-standard comp where you couldn't get the role info or you perhaps had deaths and an uneven amount of debuffs went out.

The later check of which group is next also could go out to everyone as an infotext instead of the individual players. There wouldn't need to be an extra trigger for there as we could just use the extra trigger created by the initial check. Adding the names to the group 2-4 calls is the same.

Legends0: From a raid leader stand point, I could see that knowing which players are to break would help  but this is going to be a lot of info.

quisquous: Yeah, I think that sounds like a personal trigger to me as well.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -108,18 +161,26 @@ const triggerSet: TriggerSet<Data> = {
       regex: /Kothornos Kick/,
       beforeSeconds: 5,
       infoText: (data, _matches, output) => {
-        if (data.role === 'tank' || data.role === 'healer')
-          return;
-        return output.text!();
+        let jumpDir;
+        if (data.jumpDir1 === 'east')
```
quisquous, line 165: Subjectively, I think the bait call comes out late.  It seems like the directions are reversed as well.

https://www.twitch.tv/videos/1268415201?t=2h23m24s

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -899,6 +1186,53 @@ const triggerSet: TriggerSet<Data> = {
         },
       },
     },
+    {
+      id: 'P4S Curtain Call Debuffs',
```
quisquous, line 1190: https://www.twitch.tv/videos/1268415201?t=2h27m28s

* Tank tether swap seems wrong (we don't need to swap)
* Tank tether callout has ??? in it
* Group call is wrong (I'm group 1, not 4)
* Group call happens twice

What about moving the tank callout to a future PR, and keeping this as just a simple group number?

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: I see where the jump directions go wrong. The value stored in jumpDir1 is the same as the starting direction, so it has to start out flipped. The ternary operator is working based on it being undefined at start to determine the direction as well.

I don't think we can really add more time outside of .3 seconds as that is when you would have to be in the thorn stack otherwise. However, it would be possible we add more time for the first kick and have a shorter time for 2nd.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: > Group call is wrong (I'm group 1, not 4)
Not sure how that happened, data.curtainCallGroup should have evaluated to 1 and only been set when target is you... the output[1] should evaluate to the num1... is Conditions.targetIsYou() not functioning?

> Tank tether callout has ??? in it
I guess something went wrong getting Purple healer headmarker...

> Group call happens twice
I'm not sure how the call can happen twice... the condition is target is you and you only gain the affect once?

Legends0: I'm hoping it's just something I messed up with the return lines as I can't explain what else would have caused that output. The ??? output clearly means some sort of syntax was wrong on my part.

The rest I am guessing is just the bad return line causing multiple output, but the 4 is a weird one, the log would show "12.00", which should parseFloat to 12.00 and then subtract 2, gets 10 then ceiling of that is still 10 and then /10 is 1. The only way I could see it getting 4 is that it also ran for some other party member as the tank role would always get the tankswap line.

I have simplified the trigger now. I think to do a tank swap callout it needs to know who has highest aggro, but I don't know if that is something we can get. Looking at other cactbot encounters, such feature was implemented in a9s by just tracking autos the whole fight. Not sure if that's ok to do here (could at least lock it to curtain call with condition). I think if you had been the main tank at the time you'd want to swap there when you go out to break the orb?

Legends0: Regarding the headmarkers. I think I see where I may have made a mistake. I assumed 3 headmarkers would be enough, likely need all of them as there is no guarantee on the order in the log they will appear? In other words, I checked for purple tether at 0 index, but if the 3 headmarkers it picked first did not include purple (could have been all orange for instance), then it would fail. No idea how to handle missed headmarkers here, but likely needs to keep gathering, sort once in the other trigger and just hope purple is guaranteed? Unless someone gets some logs with just 1 or 2 people alive ser what happens ha.

Legends0: Reviewing some pulls I have, not sure if it is consistent but I have a fight where 1 dps was dead prior to headmarkers, so with 7/8 up it did not distribute a green headmarker.

Legends0:
based on what i am seeing from output, this condition is bugged

quisquous:
> EDIT: 2nd callout for curtain call works, but the first call out is still double calling somehow

Ok, thanks.  I'll take a look in raidemulator and see if I can see anything obvious.

Legends0:
Is it that I left out matches in the function that Conditions.targetIsYou bugs out?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -796,6 +849,102 @@ const triggerSet: TriggerSet<Data> = {
       netRegexJa: NetRegexes.startsUsing({ id: '6A2D', source: 'ヘスペロス', capture: false }),
       response: Responses.aoe(),
     },
+    {
+      id: 'P4S Act',
```
quisquous, line 853:
```suggestion
      id: 'P4S Act Tracker',
```

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -846,6 +995,89 @@ const triggerSet: TriggerSet<Data> = {
...
+          en: '${dir1}/${dir2} first',
+        },
+        north: Outputs.north,
+        east: Outputs.east,
+        south: Outputs.south,
+        west: Outputs.west,
+      },
+    },
+    {
+      id: 'P4S Color Headmarker Tracker',
```
quisquous, line 1033:
```suggestion
      id: 'P4S Color Headmarker Collector',
```

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -27,7 +31,14 @@ export interface Data extends RaidbossData {
   wellShiftKnockback?: boolean;
   beloneCoilsTwo?: boolean;
   bloodrakeCounter?: number;
+  act?: string;
+  colorHeadmarkerIds?: number[];
```
quisquous, line 35:
```suggestion
  colorHeadmarkerIds: number[];
```

You have to check this being undefined in a lot of places.  I think it'd be simpler to always initialize it.

If you add this above the `triggers: [` entry:
```typescript
  initData: () => {
    return {
      colorHeadmarkerIds: [],
    };
  },
```

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -846,6 +995,89 @@ const triggerSet: TriggerSet<Data> = {
...
+        data.colorHeadmarkerIds.push(id);
+      },
+    },
+    {
+      id: 'P4S Color Headmarkers',
+      type: 'HeadMarker',
+      netRegex: NetRegexes.headMarker({}),
+      condition: (data) => data.act !== undefined,
+      // Delay some for headmarkers to be gathered
+      delaySeconds: (data) => (data.decOffset) ? 0 : 0.3,
```
quisquous, line 1051:
```suggestion
      delaySeconds: (data) => data.decOffset ? 0 : 0.3,
```

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -72,16 +83,50 @@ const roleOutputStrings = {
...
   // If we naively just check !data.decOffset and leave it, it breaks if the first marker is 00DA.
   // (This makes the offset 0, and !0 is true.)
-  if (typeof data.decOffset === 'undefined')
-    data.decOffset = parseInt(matches.id, 16) - firstHeadmarker;
+  if (typeof data.decOffset === 'undefined') {
+    if (typeof data.act === 'undefined') {
+      data.decOffset = parseInt(matches.id, 16) - firstHeadmarker;
+    } else {
+      data.colorHeadmarkerIds ??= [];
+      data.decOffset = (data.colorHeadmarkerIds[0] ?? 0) - purpleMarker;
```
quisquous, line 127: I'm not sure what this code is doing.  Doesn't `P4S Color Headmarkers` set this too?  I think only one place should do it.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -111,21 +156,29 @@ const triggerSet: TriggerSet<Data> = {
...
-        if (data.role === 'tank' || data.role === 'healer')
-          return;
-        return output.text!();
+        let jumpDir;
+        if (data.jumpDir1 === 'east')
+          jumpDir = !data.kickTwo ? output.west!() : output.east!();
+        if (data.jumpDir1 === 'west')
+          jumpDir = !data.kickTwo ? output.east!() : output.west!();
+
+        return output.text!({ dir: jumpDir });
```
quisquous, line 167: `popup-text.ts:350 Trigger P4S Kothornos Kick has non-string param value dir.` I think this is getting called with undefined dir somehow.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -111,21 +156,29 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'P4S Kothornos Kick',
       regex: /Kothornos Kick/,
-      beforeSeconds: 5,
+      beforeSeconds: 5.3,
```
quisquous, line 159: This is a 5 second cast.  Should this be a normal trigger and not a timeline trigger?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -796,6 +849,102 @@ const triggerSet: TriggerSet<Data> = {
...
+        // trigger will not work, so just resume promise here
+        if (combatantData === null) {
+          console.error(`Hesperos: null data`);
+          return;
+        }
+        if (!combatantData.combatants) {
+          console.error(`Hesperos: null combatants`);
+          return;
+        }
+        if (combatantData.combatants.length < 39) {
```
quisquous, line 901: In raid emulator, there's only ~22 on one log that I have.  Do you really need 39?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -855,6 +1087,44 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegexDe: NetRegexes.tether({ id: '00AD', source: 'Hesperos' }),
+      netRegexFr: NetRegexes.tether({ id: '00AD', source: 'Hespéros' }),
+      netRegexJa: NetRegexes.tether({ id: '00AD', source: 'ヘスペロス' }),
+      condition: (data) => data.act === '3',
+      // Tethers come out East or West (0 seconds), (3s) Middle knockack, (6) Opposite Cardinal
+      suppressSeconds: 7,
+      infoText: (data, matches, output) => {
+        const thorn = (data.thornIds ??= []).indexOf(parseInt(matches.sourceId, 16));
+        const num = parseInt(matches.sourceId, 16);
+        console.error(`Act Two: Found thorn index of ${thorn} and id ${num} with ${matches.sourceId}`);
```
quisquous, line 1103: Should this only print out if there's an error?

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: The second one requires timeline trigger or duplicate trigger with delay as the second one is instant cast

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: Every log I have has 39. It needs the last 8 npcs that get loaded, at least for just the current thorns callouts. I hadn't seen a 22 combatant log, but I guess they exist then?

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: hmmm, I used the undefined of it as a false input. Perhaps that's bad design?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -899,6 +1169,51 @@ const triggerSet: TriggerSet<Data> = {
...
       },
     },
+    {
+      id: 'P4S Curtain Call Debuffs',
+      // Durations could be 12s, 22s, 32s, and 42s
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'AF4', capture: true }),
+      condition: (data) => {
+        return (Conditions.targetIsYou() && data.act === 'curtain');
+      },
```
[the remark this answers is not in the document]
Legends0: Ah, thanks.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -899,6 +1169,51 @@ const triggerSet: TriggerSet<Data> = {
...
+      response: (data, matches, output) => {
+        // cactbot-builtin-response
+        output.responseOutputStrings = curtainCallOutputStrings;
+
+        data.curtainCallGroup = (Math.ceil(parseFloat(matches.duration)) - 2) / 10;
+
+        if (parseFloat(matches.duration) < 13)
+          return { alarmText: output.groupTethers!({ num: data.curtainCallGroup }) };
```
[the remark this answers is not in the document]
Legends0: I tried to have alarm and different wording to put the emphasis on time for the group to start versus just the info text at start.

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -899,6 +1169,51 @@ const triggerSet: TriggerSet<Data> = {
...
+        return { infoText: output.tetherGroup!({ num: data.curtainCallGroup }) };
+      },
+    },
+    {
+      id: 'P4S Curtain Call Reminders',
```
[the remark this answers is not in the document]
Legends0: I can definitely see the alarm triggering someone to move without them being the one of the two to required to move.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: Woops, left that one in from copy paste.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: It isn't run, I'll remove that. The phase 2 setting of decOffset is occuring outside of the function here.

ui/raidboss/data/06-ew/raid/p4s.ts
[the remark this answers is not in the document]
Legends0: I've removed the 'tethers' word.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: Also of note, will need to rework the other thorns triggers when the array of combatants is not 39. It could be done to just have the value of data.thornIds be the last 8 in the list instead of all Hesperos combatants, but then I am not certain where the others went if it's less than 39. Perhaps there is a locale difference in some of these Hesperos combatants that I am not seeing?

Legends0: I have added the splicing. I think it's possible to just gather all the combatants instead of Hesperos only if what happens is some of these thorns in other Locales have different names, but I think this needs to be verified in other Locales?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -913,6 +1175,51 @@ const triggerSet: TriggerSet<Data> = {
...
+      // Alarms for the other groups
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'B7D', capture: true }),
+      condition: (data) => data.act === 'curtain',
+      preRun: (data) => data.curtainCallTracker = (data.curtainCallTracker ?? 0) + 1,
+      delaySeconds: (_data, matches) => parseFloat(matches.duration),
+      suppressSeconds: 1,
+      alarmText: (data, _matches, output) => {
```
quisquous, line 1206:
```suggestion
      infoText: (data, _matches, output) => {
```

Maybe infoText (or alertText) instead of alarmText for the reminder?

ui/raidboss/data/06-ew/raid/p4s.ts
```diff
@@ -111,21 +152,29 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'P4S Kothornos Kick',
       regex: /Kothornos Kick/,
-      beforeSeconds: 5,
+      beforeSeconds: 5.3,
       infoText: (data, _matches, output) => {
-        if (data.role === 'tank' || data.role === 'healer')
-          return;
-        return output.text!();
+        let jumpDir = '';
```
quisquous, line 157: I'm not sure this fixes whatever the issue was.  If `data.jumpDir1` is somehow not east or west, then should this trigger not print out anything? (or print out some other `Bait jump` with no dir output string? I'm not sure if grammatically it can just have a blank dir and have that work in all languages)

quisquous (commented):
Thanks for all the back and forth.  These are my last comments here and then this can get merged.

ui/raidboss/data/06-ew/raid/p4s.ts
Legends0: Could be possible to move the {dir} somewhere else in that language such that it would make sense, but it's easier to just have a second output in that situation than require the extra thought.

PR #4042, opened 2022-01-18
plugin: update CE opcode for cn 5.58

sandtechnology:
Thanks for @Loskh, the opcode was found by his tips : D


quisquous:
Thank you!

PR #4050, opened 2022-01-19
raidboss: p3s add callout for add phase tethers

amki:
This adds callouts for the tethers that go out in p3s' second part of the add phase.

It expects that the adds are being put on cardinal positions (that is how they spawn afterall), it could be extended to cover intercardinals as well but that'd be a lot more complicated and I've never seen anyone do this(?)

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p3s.ts
+++ b/ui/raidboss/data/06-ew/raid/p3s.ts
@@ -373,7 +373,7 @@ const triggerSet: TriggerSet<Data> = {
       },
       alertText: (data, matches, output) => {
         ++data.sunbirdTetherCounter;
-        if (data.sunbirdTetherCounter < 8) {
+        if (data.sunbirdTetherCounter <= 8) {
           // Save tether
           if (matches.source === 'Sunbird') {
             for (const s of data.sunbirds) {
@@ -392,7 +392,8 @@ const triggerSet: TriggerSet<Data> = {
             // If it's not from a sunbird it's p2p
             data.p2pTethers.push({ source: matches.source, target: matches.target });
           }
-          return;
+          if (data.sunbirdTetherCounter < 8)
+            return;
         }
         // All 8 tethers collected, make the call
         // Start with own name
@@ -400,7 +401,7 @@ const triggerSet: TriggerSet<Data> = {
         // If we're tethered to a player look for this player's sunbird tether instead
         for (const t of data.p2pTethers) {
           if (t.target === lookupName) {
-            lookupName = t.target;
+            lookupName = t.source;
             break;
           }
         }
--- a/ui/raidboss/data/06-ew/raid/p3s.ts
+++ b/ui/raidboss/data/06-ew/raid/p3s.ts
@@ -351,24 +351,6 @@ const triggerSet: TriggerSet<Data> = {
           console.error('SunbirdTether: There is not exactly four Sunbirds?!?: ${JSON.stringify(sunbirds)}');
           return;
         }
-        /*
-        console.log('------------');
-        for (const s of sunbirds) {
-          if (!s.ID) {
-            console.error('SunbirdTether: Sunbird without ID? ', s);
-            return;
-          }
-          if (s.PosY < 93)
-            console.log('sunbird ', '(', s.ID.toString(16).toUpperCase(), ')', s, 'is north');
-          else if (s.PosY > 107)
-            console.log('sunbird ', '(', s.ID.toString(16).toUpperCase(), ')', s, 'is south');
-          else if (s.PosX < 93)
-            console.log('sunbird ', '(', s.ID.toString(16).toUpperCase(), ')', s, 'is west');
-          else if (s.PosX > 107)
-            console.log('sunbird ', '(', s.ID.toString(16).toUpperCase(), ')', s, 'is east');
-        }
-        console.log('------------');
-        */
         data.sunbirds = sunbirds;
       },
       alertText: (data, matches, output) => {
@@ -409,22 +391,22 @@ const triggerSet: TriggerSet<Data> = {
           if (s.name === lookupName) {
             switch (s.pos) {
               case 0:
-                return 'Tethered North Sunbird';
+                return output.n!();
               case 1:
-                return 'Tethered East Sunbird';
+                return output.e!();
               case 2:
-                return 'Tethered South Sunbird';
+                return output.s!();
               case 3:
-                return 'Tethered West Sunbird';
+                return output.w!();
             }
           }
         }
-        return output.text!();
       },
       outputStrings: {
-        text: {
-          en: 'Tether!',
-        },
+        n: Outputs.north,
+        e: Outputs.east,
+        s: Outputs.south,
+        w: Outputs.west,
       },
     },
     {
--- a/ui/raidboss/data/06-ew/raid/p3s.ts
+++ b/ui/raidboss/data/06-ew/raid/p3s.ts
@@ -13,9 +13,7 @@ export interface Data extends RaidbossData {
   deathsToll?: boolean;
   deathsTollPending?: boolean;
   sunbirdTethers: { name: string; pos: number }[];
-  p2pTethers: { source: string; target: string }[];
   sunbirds: PluginCombatantState[];
-  sunbirdTetherCounter: number;
   decOffset?: number;
 }
 
@@ -42,8 +40,6 @@ const triggerSet: TriggerSet<Data> = {
     return {
       sunbirds: [],
       sunbirdTethers: [],
-      p2pTethers: [],
-      sunbirdTetherCounter: 0,
     };
   },
   triggers: [
@@ -335,9 +331,9 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
-      id: 'P3S Sunbird Tether',
+      id: 'P3S Sunbird Tether Collect',
       type: 'Tether',
-      netRegex: NetRegexes.tether({ id: '0039' }),
+      netRegex: NetRegexes.tether({ id: '0039', source: 'Sunbird' }),
       promise: async (data) => {
         const callData = await callOverlayHandler({
           call: 'getCombatants',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p3s.ts
```diff
@@ -320,6 +334,89 @@ const triggerSet: TriggerSet<Data> = {
...
+          console.error('SunbirdTether: failed to get combatants: ${JSON.stringify(callData)}');
+          return;
+        }
+        const sunbirds = callData.combatants.filter((c) => c.BNpcID === 13635);
+        if (sunbirds.length !== 4) {
+          console.error('SunbirdTether: There is not exactly four Sunbirds?!?: ${JSON.stringify(sunbirds)}');
+          return;
+        }
+        data.sunbirds = sunbirds;
+      },
```
Legends0, line 355: Does this need to run on every regex match?

amki: No idea if the situation changes because getCombatants can't be emulated correctly by the raidemulator. I definitely need it before first tether to populate the list.

Also I don't know of a way to limit the promise execution with a condition. Is that even possible? Best I could do is immediately resolve the promise if it's the wrong trigger.

ui/raidboss/data/06-ew/raid/p3s.ts
```diff
@@ -320,6 +334,89 @@ const triggerSet: TriggerSet<Data> = {
         },
       },
     },
+    {
+      id: 'P3S Sunbird Tether',
+      type: 'Tether',
+      netRegex: NetRegexes.tether({ id: '0039' }),
+      promise: async (data) => {
+        const callData = await callOverlayHandler({
+          call: 'getCombatants',
```
quisquous, line 343: Is there some way to limit this call to not have it return every single combatant? Maybe by name, or npc id, or even if you already have the previous ids somewhere?

ui/raidboss/data/06-ew/raid/p3s.ts
```diff
@@ -320,6 +334,89 @@ const triggerSet: TriggerSet<Data> = {
...
+              }
+            }
+          } else {
+            // If it's not from a sunbird it's p2p
+            data.p2pTethers.push({ source: matches.source, target: matches.target });
+          }
+          if (data.sunbirdTetherCounter < 8)
+            return;
+        }
+        // All 8 tethers collected, make the call
```
quisquous, line 374: What if somebody is dead?

ui/raidboss/data/06-ew/raid/p3s.ts
quisquous: You could accumulate all of the tethers in a separate trigger, and then add a `condition` here.  I think it depends on when you can get the information for the birds.  Ideally you get the bird information as soon as possible (maybe as soon as they become targetable, maybe there's an added combatant line and you don't need get combatants?), and then when tethers appear many seconds later, you can make calls from there.

Raidemulator should be able to handle getCombatants here, imo.  The birds are around and have plenty of data about where they are while people are hitting them.

ui/raidboss/data/06-ew/raid/p3s.ts
```diff
@@ -320,6 +334,89 @@ const triggerSet: TriggerSet<Data> = {
...
+        data.sunbirds = sunbirds;
+      },
+      alertText: (data, matches, output) => {
+        ++data.sunbirdTetherCounter;
+        if (data.sunbirdTetherCounter <= 8) {
+          // Save tether
+          if (matches.source === 'Sunbird') {
+            for (const s of data.sunbirds) {
+              if (s.ID && s.ID.toString(16).toUpperCase() === matches.sourceId) {
+                const pos = Math.round(2 - 2 * Math.atan2(s.PosX - 100, s.PosY - 100) / Math.PI) % 4;
```
quisquous, line 363: I know most folks do cards, but I think it'd be worth handling intercards here too just in case.

ui/raidboss/data/06-ew/raid/p3s.ts
amki: Good question. I think all 8 tethers always go out but no call would make real sense because you will die either way.
grafik
I _think_ this is the 8th tether retargeting because of a dead person unfortunately this would delay the call until the retargeting occured

ui/raidboss/data/06-ew/raid/p3s.ts
amki: I don't really know how to filter this. I see calls to getCombatants mostly not being filtered at all, sometimes by id but these have multiple so that'd be really complicated and I found https://github.com/quisquous/cactbot/blob/cf43f8fb13ae4f33cd1a271696f8df9dbd1c995d/ui/raidboss/data/05-shb/raid/e9s.ts#L842 doing it with a hard coded name array.

I could hardcode it but I have no idea what they are called in cn so this would break if locale is cn. Imo investing the additional memory copied to not depend on locale is worth it but your call obviously.

ui/raidboss/data/06-ew/raid/p3s.ts
amki: Depending on how tether retargeting works I have to refresh the sunbird's position when it tethers. I think if the bird has no more tethers but a charge left it just picks a new tether target as happened here: grafik

No idea if it's actually possible to do anything about it because I remember the charges being in very rapid succession.

valarnin:
I just wrote this same thing for my static, so I'll just leave this here for inspiration:

```js
    {
      id: 'P3S - Add Tethers',
      netRegex: NetRegexes.tether({id: '0039', capture: true}),
      promise: async (data, matches) => {
        data.sunbirds = data.sunbirds ?? [];

        data.sunbirds.push({
          source: matches.source,
          sourceId: matches.sourceId,
          target: matches.target,
          targetId: matches.targetId,
        });

        if (data.sunbirds.length < 8)
          return;

        const birds = data.sunbirds.filter(bird => bird.sourceId[0] === '4');

        const birdData = await callOverlayHandler({
          call: 'getCombatants',
          ids: birds.map(bird => bird.sourceId).map(id => parseInt(id, 16)),
        });

        if (birdData === null) {
          console.error(`Add Tethers: null data`);
          return;
        }
        if (!birdData.combatants) {
          console.error(`Add Tethers: null combatants`);
          return;
        }
        if (birdData.combatants.length !== 4) {
          console.error(`Add Tethers: expected 4, got ${birdData.combatants.length}`);
          return;
        }

        birdData.combatants.forEach((c) => {
          const bird = birds.find(b => parseInt(b.sourceId, 16) === c.ID);
          bird.x = c.PosX;
          bird.y = c.PosY;
          bird.quadrantIdx = Math.round(4 - 4 * (Math.atan2(bird.x - 100, bird.y - 100) - (Math.PI / 8)) / Math.PI) % 8;
        });
      },
      infoText: (data) => {
        if (data.sunbirds.length < 8)
          return;

        const directions = {
          0: 'N',
          1: 'NE',
          2: 'E',
          3: 'SE',
          4: 'S',
          5: 'SW',
          6: 'W',
          7: 'NW',
        };

        const meTether = data.sunbirds.find(b => b.target === data.me);
        const parentTether = data.sunbirds.find(b => b.targetId === meTether.sourceId);

        if (parentTether === undefined) {
          return directions[(meTether.quadrantIdx + 4) % 8];
        }

        return directions[parentTether.quadrantIdx];
      },
    }
```

ui/raidboss/data/06-ew/raid/p3s.ts
valarnin: > No idea if the situation changes because getCombatants can't be emulated correctly by the raidemulator. I definitely need it before first tether to populate the list.

raidemulator handles getCombatants fine for this case, for future reference. See my other version of the trigger, which works fine in raidemulator and produces the correct results according to VOD checking.

amki: No offense meant against the raidemulator I just don't know how it emulates the combatant data and since it burned me on zodiark my trust in the data is low.

I _guess_ it uses all log lines to retrack combatant positions? In that case I totally agree the data is fine.

amki:
I tested my version and it works flawlessly in the 10 pulls we did.

ui/raidboss/data/06-ew/raid/p3s.ts
amki: This would reduce the leeway tanks have to put an add "north" significantly. Can the config GUI do checkboxes or something to let the user choose between two options for this?

quisquous:
Out of curiosity, is there a reason you both went for the getCombatants approach vs using the positions from AddedCombatant? I was mostly just thinking about how to handle the case where there are missing tethers.

> No idea if it's actually possible to do anything about it because I remember the charges being in very rapid succession. This would also block using the AddCombatant line which looks to have sane coordinates attached to it.

I'm not sure I understand why tether retargeting would block this approach.  If somebody is dead, then (likely) somebody will die from an extra charge, but I think it's still worth telling everybody where to go as if people aren't dead.

amki:
> Out of curiosity, is there a reason you both went for the getCombatants approach vs using the positions from AddedCombatant? I was mostly just thinking about how to handle the case where there are missing tethers.

It is the obvious approach imo. When you (or better when I) start to think about how to implement this mechanic I go from: 
- Ok, I get a tether and need to know where I have to go. How do I find out?
- Look where the tether comes from! Where does it come from?
- Sunbird or Player then Sunbird! Get where the Sunbird is right now then! => getCombatants

Also AddCombatant is (at least a little?) unforeseeable in its usefulness. Will it be emitted or another creature reused? Are the sunbirds the same as the first iteration and AddCombatant is only called for them?

My new approach will also work with missing tethers by the way because it just awaits the tether hitting you and then (and only then) looks up the path.

> I'm not sure I understand why tether retargeting would block this approach. If somebody is dead, then (likely) somebody will die from an extra charge, but I think it's still worth telling everybody where to go as if people aren't dead.

Because the Sunbird might not have it's AddCombatant position anymore when the tether retargets. As you can see the retarget happens 9.2 seconds after the first tethers, by then the sunbirds will have moved and the stale call might be wrong?

PR #4062, opened 2022-01-23
raidboss: p3s sunbird tether callouts

quisquous:
This is an alternate approach to #4050, combining amki and valarnin's
work from that patch with some extra tweaks.

In particular, this uses AddedCombatant over getCombatants to make it
easier to debug and differentiates bird tethers from player tethers.

Co-authored-by: valarnin <valarnin@gmail.com>
Co-authored-by: amki <amki@amki.eu>

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p3s.ts
+++ b/ui/raidboss/data/06-ew/raid/p3s.ts
@@ -258,9 +258,9 @@ const triggerSet: TriggerSet<Data> = {
       id: 'P3S Sunbird Tether',
       type: 'Tether',
       // There is no need for a delay here, because all of the tethers are ordered:
-      //   Sunbird => Player1
+      //   SunbirdA => Player1
       //   Player1 => Player2
-      //   Sunbird => Player3
+      //   SunbirdB => Player3
       //   Player3 => Player4
       // ...therefore if this tether has the current player as a target, then we
       // will have seen the Sunbird => Player tether previously if it exists in the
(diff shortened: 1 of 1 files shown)


quisquous:
@amki Sorry, I started taking a closer look at #4050 and I had a bunch of comments and thought maybe it'd just be briefer if I wrote my own.  I took what you wrote and what valarnin wrote and some of my own opinions and wrote this.

I think retargeting is not worth handling.  We should call out an initial tether on folks and then not call any other tethers that appear at that point.

I know you had concerns about all eight directions vs just four, but in logs that I tested, eight directions worked just fine and seems more general in case anybody does that.

I think I was also worried about generalities about whether tethers started 0039 or 0001.

amki:
> @amki Sorry, I started taking a closer look at #4050 and I had a bunch of comments and thought maybe it'd just be briefer if I wrote my own.

Ouch.

I agree with all your points though. ;)

ui/raidboss/data/06-ew/raid/p3s.ts
```diff
@@ -229,6 +238,90 @@ const triggerSet: TriggerSet<Data> = {
         north: Outputs.north,
       },
     },
+    {
+      id: 'P3S Sunbird Tether Collector',
+      type: 'Tether',
+      // 0039 when pink, 0001 when stretched purple.
+      // TODO: in general, it seems like the tethers are picked to start unstretched,
+      // but plausibly you could create a scenario where one starts stretched?
+      netRegex: NetRegexes.tether({ id: ['0039', '0001'] }),
```
amki, line 247: You could narrow this to source: 'Sunbird' later 'Sunshadow's spawn that also create a 0001 tether.

Doesn't make any functional difference though.

quisquous: Hmm, I don't think I can narrow it, because it could be between players too.  I think right now if a player somehow doesn't get a tether during add phase, then it'll fall into the "no bird" error output and not print anything on screen.  I can make a note that it might happen.

quisquous:
> Ouch.

Sorry!  I really appreciate the PR, and wasn't trying to invalidate it.  I just had a bunch of opinions and it seemed like it would be easier to just mash everybody's suggestions together rather than trying to ask for everything individually, if that makes sense?

amki:
No worries I'm just offering ideas, no strings attached :)

ui/raidboss/data/06-ew/raid/p3s.ts
quisquous: Ohhhh, you're right, thanks for explaining.  Updated.

PR #4076, opened 2022-01-29
i18n: Ultima_weapon_ultimate fr

VanSlanzar:
For `Grand Whirlwind`, i couldn't find this skill. Is this something you created for the timeline or that it was an error when creating the timeline? Otherwise, i made this correction, because i found this skill `Great Whirlwind`. Tell me if this is wrong and I will correct my modification.

**This image is from the FFLOGS on UWU.**
Screenshot_7


quisquous:
I checked my logs, and it sure looks like Great Whirlwind a few years ago too.  Thanks for the fix!

VanSlanzar:
It's always a pleasure to contribute for everyone.

PR #4088, opened 2022-01-31
i18n: add missing transaltions - Misc test fr

VanSlanzar:
@quisquous or @Akurosia

For these 3 regex, I'm not sure. Is there a way to check them?
```
netRegexFr: NetRegexes.echo({ line: 'cactbot langue.*?', capture: false }),
netRegexFr: NetRegexes.echo({ line: 'cactbot test de réponse.*?', capture: false }),
netRegexFr: NetRegexes.echo({ line: 'cactbot montre d\'essai.*?', capture: false }),
```

### DIFF
--- a/ui/oopsyraidsy/oopsy_live_list.ts
+++ b/ui/oopsyraidsy/oopsy_live_list.ts
@@ -19,6 +19,7 @@ const kCopiedMessage = {
 const errorMessageEnableACTWS = {
   en: 'Plugins -> OverlayPlugin WSServer -> Stream/Local Overlay -> Start',
   de: 'Plugins -> OverlayPlugin WSServer -> Stream/Local Overlay -> Start',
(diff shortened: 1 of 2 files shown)


quisquous:
Hmm.  I'm not sure what you're asking.  These are just commands in game, e.g. `/echo cactbot langue`.

That said, I'm not sure that these *should* have translations? I think these are just debugging commands to test things out, and so probably they should be the same in all languages.  What do you think, @Akurosia?

Akurosia:
i would be fine in either scenario,
Best thing would be if both the english and the translated commnd would work^^'
But i think with the current implementation just english would be the best to avoid confusion (but tbh most users will never use any of these commands)

VanSlanzar:
You can merged. I understood what the regex wanted to do. It's OK.

VanSlanzar:
Thank you for your respons. That helped me.

PR #4092, opened 2022-01-31
jobs: SGE: Improvement

Echoring:
- Change adderTimer from bar to number box, and add exceed alert color.
Things of healing resource need more accurate gauge.
- Hide Adder Gauge below lv 45
Due to unknown reason, if you sync to below lv45, addersgall is not availble but memory still says you have 3 addersgall.
- Add HP and MP number display
- Add Phlegma Timer
- Fix missing threholds and reset

### DIFF
--- a/ui/jobs/components/sge.ts
+++ b/ui/jobs/components/sge.ts
@@ -1,16 +1,17 @@
 import EffectId from '../../../resources/effect_id';
-import TimerBar from '../../../resources/timerbar';
 import TimerBox from '../../../resources/timerbox';
 import { JobDetail } from '../../../types/event';
+import { ResourceBox } from '../bars';
 import { kAbility } from '../constants';
 import { PartialFieldMatches } from '../event_emitter';
 
 import { BaseComponent, ComponentInterface } from './base';
 
 export class SGEComponent extends BaseComponent {
-  adderTimer: TimerBar;
+  stacksContainer: HTMLDivElement;
   addersgallStacks: HTMLDivElement[];
   adderstingStacks: HTMLDivElement[];
+  adderTimerBox: ResourceBox;
   eukrasianDosis: TimerBox;
   rhizomata: TimerBox;
   lucidDream: TimerBox;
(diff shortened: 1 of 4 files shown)


ui/jobs/components/sge.ts
```diff
@@ -93,29 +114,33 @@ export class SGEComponent extends BaseComponent {
...
   override onStatChange({ gcdSpell }: { gcdSpell: number }): void {
     this.eukrasianDosis.valuescale = gcdSpell;
     this.eukrasianDosis.threshold = gcdSpell + 1;
+    this.phlegma.valuescale = gcdSpell;
+    this.phlegma.threshold = gcdSpell + 1;
+    this.rhizomata.valuescale = gcdSpell;
+    this.rhizomata.threshold = gcdSpell + 1;
     this.lucidDream.valuescale = gcdSpell;
     this.lucidDream.threshold = gcdSpell + 1;
+    // Due to unknown reason, if you sync to below lv45,
```
quisquous, line 134: Thanks for the comment!

ui/jobs/components/sge.ts
```diff
@@ -76,10 +82,25 @@ export class SGEComponent extends BaseComponent {
...
   override onUseAbility(id: string): void {
-    if (id === kAbility.LucidDreaming)
-      this.lucidDream.duration = 60;
-    else if (id === kAbility.Rhizomata)
-      this.rhizomata.duration = 90;
+    switch (id) {
+      case kAbility.Phlegma:
+      case kAbility.Phlegma2:
+      case kAbility.Phlegma3:
+        if (!this.multipleCall) { // Avoid multiple call in AOE
```
quisquous, line 89: I think you'd have to pass in the entire matches here, but you can use the `matches.targetIndex` field to only trigger on the first target of the aoe rather than using a countdown.  I think this is fine for now, but maybe something to fix up if you wanted?

quisquous:
Thanks for all the jobs improvements!!

ui/jobs/components/sge.ts
Echoring: @MaikoTan said he is trying to import a t
throttle function in util to simplify this. I'm not sure which means is better? I've not thought about how to use targetIndex to do this

quisquous: Every enemy hit by an ability gets a different target index.  You can see this in logs if you look at aoe ability lines.  So you can check the match and only consider targetIndex of 0 when doing things like this.

MaikoTan: Echoring came to me and asked how to restrict a function so can only be invoked once in some secondes, so I answered "throttle function" 😂.
Looks like in this case it can be solved with the `targetIndex` that mentioned above. That's great !

PR #4097, opened 2022-02-02
jobs: Improve multi-call preventer

Echoring:
As discussed in #4092, use `matches.targetIndex` instead.


quisquous:
Thanks!  That makes things a lot simpler.  :)

PR #4100, opened 2022-02-03
eureka: detect _ in the trackerRegex

Jessidhia:
It's possible for the tracker id to be generated with an underscore.

Example: https://ffxiv-eureka.com/v3y_9B
image


quisquous:
Thanks for the fix!

PR #4118, opened 2022-02-13
docs: Log guide updates for 6.x

xpdota:
- Added some 6.x changes (like the target count on 21/22 lines)
- Clarified ability damage/effect flags
- Clarified what buff "stacks" really mean
- Clarified buff refresh/overwrite behavior
- Added note about sequential/obfuscated headmarkers
- Added info about how 37-lines work
- Added note about HP in 37/38/39 lines
- Added more reasons to not use chat lines (as if we needed more)
- Added note about fake combatants

### DIFF
--- a/docs/LogGuide.md
+++ b/docs/LogGuide.md
@@ -415,6 +415,8 @@ There are a number of reasons to avoid basing triggers on game log lines:
 - inconsistent text (gains effect vs suffers effect, begins casting vs readies, you vs player name)
 - often vague (the attack misses)
 - can change spelling at the whim of SquareEnix
+- requires the relevant type of line to be enabled in the player's battle log settings
+- is the most likely to require variants for other languages
 
 Instead, the recommendation is to base your triggers on ACT log lines that are not type `0x00`.
 Prefer using [NetworkBuff](#line26) line instead of "suffers the effect" game log lines.
@@ -557,7 +559,8 @@ ACT Log Line Examples:
 
 This combatant may be invisible and fake.  The real ones have more HP.
 For example, at the start of Deltascape V2.0 you will see messages like the
-latter 5 examples above.
+latter 5 examples above. In addition, anything with an NPC ID of 9020 (0x233C)
+is almost certainly fake.
 
 In heavy zones (e.g. Eureka), combatants may be culled if there are too many
 things nearby.
@@ -760,26 +763,33 @@ to include both possibilities.
 
 Ground AOEs that don't hit anybody are considered [NetworkAOEAbility](#line22) lines.
 
+There are two fields on 21/22 lines that provide information about the number of targets affected. The `targetCount`
+field indicates the number of targets. The `targetIndex` field indicates which target this particular line refers to.
+For example, for a 21-line, you would see a `targetIndex` of 0, and a `targetCount` of 1. For an AoE ability that hits
+three targets, all three lines would have a `targetCount` of 3, but the `targetIndex` would be 0, 1, and 2 for the three
+lines. Thus, you can determine all of the targets hit by a particular ability usage by collecting 21/22-lines until you
+see one for which `targetCount - 1 == targetIndex`.
+
 <!-- AUTO-GENERATED-CONTENT:START (logLines:type=Ability&lang=en-US) -->
 
 #### Structure
 
 ```log
 Network Log Line Structure:
-21|[timestamp]|[sourceId]|[source]|[id]|[ability]|[targetId]|[target]|[flags]|[damage]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[targetCurrentHp]|[targetMaxHp]|[targetCurrentMp]|[targetMaxMp]|[?]|[?]|[targetX]|[targetY]|[targetZ]|[targetHeading]|[currentHp]|[maxHp]|[currentMp]|[maxMp]|[?]|[?]|[x]|[y]|[z]|[heading]|[sequence]|[targetIndex]
+21|[timestamp]|[sourceId]|[source]|[id]|[ability]|[targetId]|[target]|[flags1]|[damage1]|[flags2]|[damage2]|...|[flags8]|[damage8]|[targetCurrentHp]|[targetMaxHp]|[targetCurrentMp]|[targetMaxMp]|[?]|[?]|[targetX]|[targetY]|[targetZ]|[targetHeading]|[currentHp]|[maxHp]|[currentMp]|[maxMp]|[?]|[?]|[x]|[y]|[z]|[heading]|[sequence]|[targetIndex]|[targetCount]
 
 ACT Log Line Structure:
-[timestamp] ActionEffect 15:[sourceId]:[source]:[id]:[ability]:[targetId]:[target]:[flags]:[damage]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[?]:[targetCurrentHp]:[targetMaxHp]:[targetCurrentMp]:[targetMaxMp]:[?]:[?]:[targetX]:[targetY]:[targetZ]:[targetHeading]:[currentHp]:[maxHp]:[currentMp]:[maxMp]:[?]:[?]:[x]:[y]:[z]:[heading]:[sequence]:[targetIndex]
+[timestamp] ActionEffect 15:[sourceId]:[source]:[id]:[ability]:[targetId]:[target]:[flags1]:[damage1]:[flags2]:[damage2]:...[flags8]:[damage8]:[targetCurrentHp]:[targetMaxHp]:[targetCurrentMp]:[targetMaxMp]:[?]:[?]:[targetX]:[targetY]:[targetZ]:[targetHeading]:[currentHp]:[maxHp]:[currentMp]:[maxMp]:[?]:[?]:[x]:[y]:[z]:[heading]:[sequence]:[targetIndex]:[targetCount]
 ```
 
 #### Regexes
 
 ```log
 Network Log Line Regex:
-^(?<type>(?:2[12]))\|(?<timestamp>(?:[^|]*))\|(?<sourceId>(?:[^|]*))\|(?<source>(?:[^|]*))\|(?<id>(?:[^|]*))\|(?<ability>(?:[^|]*))\|(?<targetId>(?:[^|]*))\|(?<target>(?:[^|]*))\|(?<flags>(?:[^|]*))\|(?<damage>(?:[^|]*))\|(?:[^|]*\|){14}(?<targetCurrentHp>(?:[^|]*))\|(?<targetMaxHp>(?:[^|]*))\|(?<targetCurrentMp>(?:[^|]*))\|(?<targetMaxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<targetX>(?:[^|]*))\|(?<targetY>(?:[^|]*))\|(?<targetZ>(?:[^|]*))\|(?<targetHeading>(?:[^|]*))\|(?<currentHp>(?:[^|]*))\|(?<maxHp>(?:[^|]*))\|(?<currentMp>(?:[^|]*))\|(?<maxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<x>(?:[^|]*))\|(?<y>(?:[^|]*))\|(?<z>(?:[^|]*))\|(?<heading>(?:[^|]*))\|(?<sequence>(?:[^|]*))\|(?<targetIndex>(?:[^|]*))\|
+^(?<type>(?:2[12]))\|(?<timestamp>(?:[^|]*))\|(?<sourceId>(?:[^|]*))\|(?<source>(?:[^|]*))\|(?<id>(?:[^|]*))\|(?<ability>(?:[^|]*))\|(?<targetId>(?:[^|]*))\|(?<target>(?:[^|]*))\|(?<flags1>(?:[^|]*))\|(?<damage1>(?:[^|]*))\|(?:[^|]*\|){14}(?<targetCurrentHp>(?:[^|]*))\|(?<targetMaxHp>(?:[^|]*))\|(?<targetCurrentMp>(?:[^|]*))\|(?<targetMaxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<targetX>(?:[^|]*))\|(?<targetY>(?:[^|]*))\|(?<targetZ>(?:[^|]*))\|(?<targetHeading>(?:[^|]*))\|(?<currentHp>(?:[^|]*))\|(?<maxHp>(?:[^|]*))\|(?<currentMp>(?:[^|]*))\|(?<maxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<x>(?:[^|]*))\|(?<y>(?:[^|]*))\|(?<z>(?:[^|]*))\|(?<heading>(?:[^|]*))\|(?<sequence>(?:[^|]*))\|(?<targetIndex>(?:[^|]*))|(?<targetCount>(?:[^|]*))\|
 
 ACT Log Line Regex:
-(?<timestamp>^.{14}) (?:ActionEffect|AOEActionEffect) (?<type>(?:15|16)):(?<sourceId>(?:[^:]*)):(?<source>(?:[^:]*)):(?<id>(?:[^:]*)):(?<ability>(?:(?:[^:]|: )*?)):(?<targetId>(?:[^:]*)):(?<target>(?:[^:]*)):(?<flags>(?:[^:]*)):(?<damage>(?:[^:]*))(?::[^:]*){14}:(?<targetCurrentHp>(?:[^:]*)):(?<targetMaxHp>(?:[^:]*)):(?<targetCurrentMp>(?:[^:]*)):(?<targetMaxMp>(?:[^:]*))(?::[^:]*){2}:(?<targetX>(?:[^:]*)):(?<targetY>(?:[^:]*)):(?<targetZ>(?:[^:]*)):(?<targetHeading>(?:[^:]*)):(?<currentHp>(?:[^:]*)):(?<maxHp>(?:[^:]*)):(?<currentMp>(?:[^:]*)):(?<maxMp>(?:[^:]*))(?::[^:]*){2}:(?<x>(?:[^:]*)):(?<y>(?:[^:]*)):(?<z>(?:[^:]*)):(?<heading>(?:[^:]*)):(?<sequence>(?:[^:]*)):(?<targetIndex>(?:[^:]*))(?:$|:)
+(?<timestamp>^.{14}) (?:ActionEffect|AOEActionEffect) (?<type>(?:15|16)):(?<sourceId>(?:[^:]*)):(?<source>(?:[^:]*)):(?<id>(?:[^:]*)):(?<ability>(?:(?:[^:]|: )*?)):(?<targetId>(?:[^:]*)):(?<target>(?:[^:]*)):(?<flags1>(?:[^:]*)):(?<damage1>(?:[^:]*))(?::[^:]*){14}:(?<targetCurrentHp>(?:[^:]*)):(?<targetMaxHp>(?:[^:]*)):(?<targetCurrentMp>(?:[^:]*)):(?<targetMaxMp>(?:[^:]*))(?::[^:]*){2}:(?<targetX>(?:[^:]*)):(?<targetY>(?:[^:]*)):(?<targetZ>(?:[^:]*)):(?<targetHeading>(?:[^:]*)):(?<currentHp>(?:[^:]*)):(?<maxHp>(?:[^:]*)):(?<currentMp>(?:[^:]*)):(?<maxMp>(?:[^:]*))(?::[^:]*){2}:(?<x>(?:[^:]*)):(?<y>(?:[^:]*)):(?<z>(?:[^:]*)):(?<heading>(?:[^:]*)):(?<sequence>(?:[^:]*)):(?<targetIndex>(?:[^:]*)):(?<targetCount>(?:[^:]*))(?:$|:)
 ```
 
 #### Examples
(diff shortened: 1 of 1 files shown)


xpdota:
Also, if you're okay with links to competing projects, I can put in links to reference implementations of some of that stuff, like the flag/damage parsing, buff tracking, or sequence ID tracking.

Akurosia:
https://github.com/quisquous/cactbot/pull/4118/files#diff-af3eb20d6c5ae8fe60c2919a9d350a5641b16e9fe862cfd61e10d79f22af183aR779

Are you sure its always multiple flags and damage numbers?
As far as i understood it, these are always pairs of values [e.g. flags and damage] but can also be different pairs that are not purly for damage

docs/LogGuide.md
```diff
@@ -557,7 +559,8 @@ ACT Log Line Examples:
 
 This combatant may be invisible and fake.  The real ones have more HP.
 For example, at the start of Deltascape V2.0 you will see messages like the
-latter 5 examples above.
+latter 5 examples above. In addition, anything with an NPC ID of 9020 (0x233C)
+is almost certainly fake.
```
quisquous, line 563: Do you have examples of this? I haven't seen anything specific related to fake NPC ids before.

docs/LogGuide.md
```diff
@@ -760,26 +763,33 @@ to include both possibilities.
 
 Ground AOEs that don't hit anybody are considered [NetworkAOEAbility](#line22) lines.
 
+There are two fields on 21/22 lines that provide information about the number of targets affected. The `targetCount`
```
quisquous, line 766: Can you format this with semantic linebreaks rather than wrapping at some post 80 column limit? Same comment for all of the other changes to this file as well.

See: https://github.com/quisquous/cactbot/blob/main/CONTRIBUTING.md#markdown-guidelines

docs/LogGuide.md
```diff
@@ -760,26 +763,33 @@ to include both possibilities.
...
+field indicates the number of targets. The `targetIndex` field indicates which target this particular line refers to.
+For example, for a 21-line, you would see a `targetIndex` of 0, and a `targetCount` of 1. For an AoE ability that hits
```
quisquous, line 768: You should only use a hyphen on compound nouns that are being used as an adjective, e.g. two-party system or 10-minute enrage.  "21-line" isn't being used as an adjective and so shouldn't have a hyphen, in my opinion.  You do this in a bunch of other places in this PR, which I haven't listed.

```suggestion
For example, for a 21 line, you would see a `targetIndex` of 0, and a `targetCount` of 1. For an AoE ability that hits
```

docs/LogGuide.md
```diff
@@ -760,26 +763,33 @@ to include both possibilities.
...
+three targets, all three lines would have a `targetCount` of 3, but the `targetIndex` would be 0, 1, and 2 for the three
+lines. Thus, you can determine all of the targets hit by a particular ability usage by collecting 21/22-lines until you
```
quisquous, line 770:
```suggestion
lines. Thus, you can determine all of the targets hit by a particular ability usage by collecting 21/22 lines until you
```

docs/LogGuide.md
```diff
@@ -838,47 +848,71 @@ Network ability lines are a combination of raw network data
...
 (e.g. the `66480` current hp value and `-653.0394` x position).
 
-This means that there's a number of caveats going on to handling all the data in these lines.  The raw network data is subject to change over time from ff14 servers.  Also, the data from memory may be slightly stale and out of date
+This means that there's a number of caveats going on to handling all the data in these lines. The raw network data is
+subject to change over time from ff14 servers. Also, the data from memory may be slightly stale and out of date
+
+#### Action Effects
+
+Each ability may have one or more effects. These are indicated by the flagsX and damageX fields, between `targetName`
+and `targetCurHp`. There are eight pairs, each with a 'flags' field and a 'damage' field. The damage field is not
```
quisquous, line 857:
```suggestion
and `targetCurrentHp`. There are eight pairs, each with a 'flags' field and a 'damage' field. The damage field is not
```

docs/LogGuide.md
```diff
@@ -838,47 +848,71 @@ Network ability lines are a combination of raw network data
...
-and whether the positional was hit.
+Some of these flags also indicate whether the ability is part of a combo or not and whether the positional was hit.
 However, these values do not seem to be consistent between jobs.
 
-For example, the flags for successful trick attack are `28710.03`.
-The `.` here represents 0-3 as the trick may crit, dh, both, or neither.
-The flags for a missed trick attack are `710.03`.
-Thus, there's a `0x28700000` mask applied here when the positional is correct,
-which was determined via experimentation.
+For example, the flags for successful trick attack are `28710.03`. The `.` here represents 0-3 as the trick may crit,
```
quisquous, line 905: Why did you reformat this section?

docs/LogGuide.md
```diff
@@ -838,47 +848,71 @@ Network ability lines are a combination of raw network data
...
-Please send pull requests to this document so it can be shared!
+If you care about specific ability flags, you likely have to do this research yourself. Please send pull requests to
+this document so it can be shared!
 
 #### Ability Damage
 
 Damage bitmasks:
-    0x1000 = hallowed, no damage
-    0x4000 = "a lot" of damage
+0x1000 = hallowed, no damage 0x4000 = "a lot" of damage
```
quisquous, line 915: ?

docs/LogGuide.md
```diff
@@ -760,26 +763,33 @@ to include both possibilities.
...
+see one for which `targetCount - 1 == targetIndex`.
+
 <!-- AUTO-GENERATED-CONTENT:START (logLines:type=Ability&lang=en-US) -->
 
 #### Structure
 
 ```log
 Network Log Line Structure:
-21|[timestamp]|[sourceId]|[source]|[id]|[ability]|[targetId]|[target]|[flags]|[damage]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[?]|[targetCurrentHp]|[targetMaxHp]|[targetCurrentMp]|[targetMaxMp]|[?]|[?]|[targetX]|[targetY]|[targetZ]|[targetHeading]|[currentHp]|[maxHp]|[currentMp]|[maxMp]|[?]|[?]|[x]|[y]|[z]|[heading]|[sequence]|[targetIndex]
+21|[timestamp]|[sourceId]|[source]|[id]|[ability]|[targetId]|[target]|[flags1]|[damage1]|[flags2]|[damage2]|...|[flags8]|[damage8]|[targetCurrentHp]|[targetMaxHp]|[targetCurrentMp]|[targetMaxMp]|[?]|[?]|[targetX]|[targetY]|[targetZ]|[targetHeading]|[currentHp]|[maxHp]|[currentMp]|[maxMp]|[?]|[?]|[x]|[y]|[z]|[heading]|[sequence]|[targetIndex]|[targetCount]
```
quisquous, line 779: Did you edit this manually? If so, please edit `resources/netlog_defs.ts` and then rerun `util/gen_log_guide.ts`.

I think also you shouldn't rename the first `flags` as there is code (possibly user code) using those field names.

Alternatively you could just leave it as-is and only add `targetCount`.

docs/LogGuide.md
```diff
@@ -838,47 +848,71 @@ Network ability lines are a combination of raw network data
...
+This means that there's a number of caveats going on to handling all the data in these lines. The raw network data is
+subject to change over time from ff14 servers. Also, the data from memory may be slightly stale and out of date
+
+#### Action Effects
+
+Each ability may have one or more effects. These are indicated by the flagsX and damageX fields, between `targetName`
+and `targetCurHp`. There are eight pairs, each with a 'flags' field and a 'damage' field. The damage field is not
+necessarily always damage. For example, for a buff application, it indicates which buff ID is about to be applied.
+
+This means that damage is not necessarily the first flag. If you are simply trying to determine how much damage a
```
quisquous, line 860: Do you have examples of this not being true? I feel like I've seen the damage being first almost always.

docs/LogGuide.md
```diff
@@ -1122,6 +1139,25 @@ You cannot count on the time remaining to be precise.
...
 
+The 'stacks' value may also indicate other information about the buff. For example, Mudra will show different stack
+values for different combinations of Mudra. The only way to be sure that you are getting a "real" stack value is by
+cross-referencing with game data (such as via Xivapi) to see what the game says the max stack amount for that ability
+is (0 is the max for most abilities).
+
+#### Refreshes, Overwrites, and Deaths
+
+If a buff is refreshed early, you will get another 26-line. You will not get a 30-line indicating that the existing
+buff has been removed. The same is true for when stacks are consumed or added.
```
quisquous, line 1150: This is not always true.  I have seen plenty of examples where you lose 4 stacks and gain 5 stacks when transitioning from 4 to 5.  I can dig some up if you need them.  I think light rampant lightsteeped stacks do this, for example.

docs/LogGuide.md
```diff
@@ -1172,9 +1208,13 @@ although the fields before the `id` infrequently have non-zero values.
 Note: It's unclear when the head markers disappear.
 Maybe one of these fields is a duration time? It's not clear what either of these unknown values mean.
 
-Also, this appears to only be true on later fights.
-Turn 5 fireball and conflag headmarkers are actions from Twintania and not `NetworkTargetIcon` lines.
-It seems likely this was implemented later and nobody wanted to break old content by updating it to use newer types.
+Also, this appears to only be true on later fights. Turn 5 fireball and conflag headmarkers are actions from Twintania
```
quisquous, line 1211: Why is this reformatted?

quisquous (commented):
> Also, if you're okay with links to competing projects, I can put in links to reference implementations of some of that stuff, like the flag/damage parsing, buff tracking, or sequence ID tracking.

I'm not sure what you mean by "competing" but I think links to example implementations sound helpful for folks to understand how log lines work.

docs/LogGuide.md
xpdota: It is all manual, I do not have a local environment for any of this stuff. I am not intending to make any code changes, just purely docs since the information isn't specific to cactbot.

docs/LogGuide.md
xpdota: https://discord.com/channels/192732719879290880/192732719879290880/926929235766030346 (fflogs discord)

docs/LogGuide.md
quisquous: If you aren't able to run that script, then can you undo all of your changes to autogenerated code blocks? (see: `<!-- AUTO-GENERATED-CONTENT:START (logLines:type=Ability&lang=en-US) -->`)

docs/LogGuide.md
xpdota: A few of the existing examples show that already. This is just a generalized version of the "special case shifts" section that was there before.

xpdota:
@Akurosia It is not necessarily damage, but previously the field was just called "damage". "flags" isn't the best name either, since it's more of "type". I'm not really sure what the best name would be, since it's more or less specific to whatever type the flags indicate it should be. "effect" might be confused with "status effect". Maybe "type" and "rawvalue", since the value usually needs some kind of processing?

docs/LogGuide.md
quisquous: Fair, but I haven't seen the "special case shifts" in a while and was thinking about removing it entirely.  Most of those examples are ancient at this point.

docs/LogGuide.md
xpdota: Maybe the guide should just be split between the "general log guide" (since everyone, including me, links this whenever someone asks about log lines) and cactbot-specific stuff? I agree that someone referencing it for cactbot specifically would want to see information that is relevant to cactbot, but it seems most people just reference it for general log lines, for which it would be outdated. It's the closest thing there is to general documentation about log lines.

docs/LogGuide.md
```diff
@@ -1172,9 +1209,13 @@ although the fields before the `id` infrequently have non-zero values.
...
+and not `NetworkTargetIcon` lines. It seems likely this was implemented later and nobody wanted to break old content by
+updating it to use newer types.
+
+Note that newer content uses 'sequential headmarkers' - every headmark is offset by a per-instance value. You will need
+to wait until you see the first headmarker in the instance, and then use this as an offset by which to adjust all the
+other IDs you see.
```
JLGarber, line 1218: Might be worth noting here that this is only the case for Extreme+ content, and thus it's relatively simple to get marker IDs from normal versions of encounters.

docs/LogGuide.md
quisquous: Sorry, it's just that it doesn't make sense to make manual updates to an autogenerated section.  The next time anybody runs that script it will clobber all of your changes.

docs/LogGuide.md
```diff
@@ -1122,6 +1140,25 @@ You cannot count on the time remaining to be precise.
...
+
+Most debuffs allow one player to place the debuff on each target. For some, such as Trick Attack, only one can be on
+the enemy at a time. If a buff is overwritten, a 30-line will be generated for the status effect that got bumped.
+
+Thus, it is sufficient to track buffs using a combination of caster, target, and status effect ID. A refresh or stack
+change will have the same caster, target, and status effect, while an overwrite will generate a 30-line anyway.
+
+When a unit dies, you will get 30-lines for buffs that were removed by it dying, so there is no need to handle that
+specially in order to have accurate buff tracking.
+
```
ayyaruq, line 1161: This is somewhat misleading. The StatusRemove events in ACT are fabricated by ACT, as well as some of the StatusAdd events. The source of truth is the StatusList event.

docs/LogGuide.md
ayyaruq: I'm not sure this is even correct, the overhead icons are VFX animations from the LockOn EXD. Their values are references to the specific VFX. The duration is available, but you need to parse the actual AVFX file to get the duration. There's also a separate actor control event for explicit removals, but ACT doesn't parse it.

docs/LogGuide.md
```diff
@@ -1706,14 +1747,36 @@ at the moment the action is "locked in".
...
+#### Tracking Ability Resolution
+
+Unfortunately, this is not particularly clean. For one, while the server does tell the game when an action is resolved,
+it does not tell the game when an action will not resolve (ghosting). However, the caster dying or target becoming
+untargetable is usually a decent indicator that something will not resolve.
+
+Note that AoE abilities may have the same sequence ID for all targets hit. Thus, you need to key off of both the
+sequence ID, *and* the target.
+
+In addition, you don't always get a 37-line. For example, a heal that is pure overheal will not generate one.
```
ayyaruq, line 1768: Just track action events without a sequence confirmation? You can do this reliably without any dodgy hacks.

docs/LogGuide.md
```diff
@@ -1706,14 +1747,36 @@ at the moment the action is "locked in".
...
+
+#### HP Values
+
+Sometimes, only the current HP is present, rather than current and max. In this case, it should be assumed that the max
+HP is unchanged.
+
+The current HP value is actually an update to the target's HP. In other words, it is the target's HP after this ability
+resolves, and thus is the most up to date HP value for the target. Line 38 and 39 also work this way, while other lines
+pull HP data from memory, and do not necessarily indicate HP changes in the first place. Thus, if you only care about
+HP, then you should ideally use only 37/38/39 lines.
```
ayyaruq, line 1778: EffectResult and StatusList always contain both the current and max resources. UpdateHPMP (line type 39) contains both as well but only the current ones are correct, the rest come from ACT memory reading.

docs/LogGuide.md
ayyaruq: It's almost never true, just a consequence of how ACT's status manager logs events.

docs/LogGuide.md
```diff
@@ -1122,6 +1140,25 @@ You cannot count on the time remaining to be precise.
 In rare cases, the time will already have counted down a tiny bit.
 This matters for cases such as ucob Nael phase doom debuffs.
 
+The 'stacks' value may also indicate other information about the buff. For example, Mudra will show different stack
+values for different combinations of Mudra. The only way to be sure that you are getting a "real" stack value is by
+cross-referencing with game data (such as via Xivapi) to see what the game says the max stack amount for that ability
+is (0 is the max for most abilities).
```
ayyaruq, line 1146: The actual use for the field is more generic, it can be used for stacks but other purposes too. It's on a per-status case tho. Side note, `xivapi` is just `xivapi`, all lowercase.

docs/LogGuide.md
ayyaruq: The majority of BNPC Base IDs aren't used by "real" actors, but you also can't rely on the ID - battle NPC actors are identified by the combination of their Base+Model IDs. While 9020 is likely one of these cases, so are like 80% of the other IDs. You can't actually rule them out without the model ID as well though.

docs/LogGuide.md
```diff
@@ -838,47 +848,72 @@ Network ability lines are a combination of raw network data
...
+necessarily always damage. For example, for a buff application, it indicates which buff ID is about to be applied.
+
+This means that damage is not necessarily the first flag. If you are simply trying to determine how much damage a
+particular ability did, there is no specific field you can look at - you must parse all of them and find the one that
+indicates damage.
 
 #### Ability Flags
 
-Damage bitmasks:
+The 'flags' field for each pair of values can be further broken down.
```
ayyaruq, line 866: I think you misunderstand how action effects work. There is no `damage` field, it's 4 generic params and a flags field, and the usage of *any* of those fields or flags depends entirely on the category of the effect. There are currently 63 effect categories, not just the 2 you mention.

ayyaruq (commented):
Just some minor feedback on usage of some fields. I think there's a bit of a knowledge gap between the author and the data. NARI has this documented in more detail for the underlying packet structures against the ACT formatted log lines. Some of the event names need updating too I guess, but that's more on ACT for recently changing them and Cactbot falling behind.

docs/LogGuide.md
xpdota: True, though the solution for that would probably be to just add the necessary documentation to the StatusEffectList event. The log guide as it exists currently has nothing useful for 38 lines.

docs/LogGuide.md
xpdota: But how do you differentiate between ghosted and pending if you don't track them?

docs/LogGuide.md
xpdota: Good to know, thanks. *Probably* not the biggest issue since max HP/MP updates are less common than current.

docs/LogGuide.md
xpdota: Yes, that's why I corrected this part of the documentation. Calling it "damage" is more or less just a holdover form the old version of it.

docs/LogGuide.md
ayyaruq: When you have both events, the sequence IDs correlate. Sequence IDs without a correlated result are ghosted. You can watch for events with a 3 second buffer (maximum event interval) and if there's no result you know it was ghosted.

xpdota: Oh, I see what you mean, but that still means you're incorrect for that window if something dies before actions resolve. Probably not a huge deal, but the goal behind updating this doc is to better answer the question of "how do I take a log and convert it back to a reasonably abstracted game state".

ayyaruq: If the target dies, you still get the sequence followup, the actor ID is just set to 0xE0000000 which is the game internal way of representing an actor that's no longer instantiated, unless I forgot how this works but that should be correct.

xpdota: I don't think that's accurate. Easy enough to test on WHM. Find a low level mob that would get oneshot by assize. Cast Holy  then immediately assize so that the Holy doesn't have a chance to resolve properly.
```
# Holy snapshot
21|2022-04-08T23:29:52.5510000-07:00|108AA260|Wynn Dohz|6504|Holy III|40029A6C|Hornet Swarm|750003|F5B0000|E|20000|1B|65048000|0|0|0|0|0|0|0|0|0|0|217|217|0|10000|||181.32|364.43|-32.38|0.00|50437|50437|9600|10000|||181.79|363.91|-32.36|-2.82|00095A40|0|1|8228a170257c3adf
# Assize heal snapshot
22|2022-04-08T23:29:53.1760000-07:00|108AA260|Wynn Dohz|DF3|Assize|108AA260|Wynn Dohz|10004|32680000|B|1F40000|0|0|0|0|0|0|0|0|0|0|0|0|50437|50437|9600|10000|||181.79|363.91|-32.36|-2.82|50437|50437|9600|10000|||181.79|363.91|-32.36|-2.82|00095A42|0|2|0d6958009f133a27
# Assize damage snapshot
22|2022-04-08T23:29:53.1760000-07:00|108AA260|Wynn Dohz|DF3|Assize|40029A6C|Hornet Swarm|750003|2B750000|0|0|0|0|0|0|0|0|0|0|0|0|0|0|217|217|0|10000|||181.32|364.43|-32.38|0.00|50437|50437|9600|10000|||181.79|363.91|-32.36|-2.82|00095A42|1|2|ea150610e59dce66
# Heal resolves
37|2022-04-08T23:29:53.8400000-07:00|108AA260|Wynn Dohz|00095A42|50437|50437|10000|10000|0||181.79|363.91|-32.36|-2.82|1800|0|BF3D|0|c249768f640e5a2b
# Damage resolves
37|2022-04-08T23:29:53.9730000-07:00|40029A6C|Hornet Swarm|00095A42|0||||||181.32|364.43|-32.38|0.00|b50a4b060fc13a33
```
In addition, even if it did generate a 37-line with E0... as the target, you still wouldn't know which ones snapshotted or ghosted on a per-target basis.

xpdota:
@ayyaruq I don't know why it only lets me reply directly to some comments, so here's the rest

> the overhead icons are VFX animations from the LockOn EXD

I'm talking about specifically the headmarker lines. Cactbot is already doing the whole headmarker offset thing.

> It's almost never true, just a consequence of how ACT's status manager logs events.

Hmm? Refreshing a buff before it falls off generates another 26, I'm not sure what you mean by this. Yes, it may be a consequence of how it manages it, but I'm not sure what you're getting at there.

> The majority of BNPC Base IDs aren't used by "real" actors

Sure, it only works one way. 9020 implies fake, but non-9020 doesn't imply it's real.

PR #4121, opened 2022-02-14
raidboss: predict last 2 EmEx swords to give more time

jacob-keller:
During Emerald Ex phase 2, the boss will use an attack called Tertius
Terminus Est. This attack spawns several swords which perform cross or
plus aoes in sequence. Players need to predict which spots will explode
last and move near them and away from the first explosions in order to
safely avoid the AOEs.

The current implementation does not prompt players anything until all
6 swords have dropped. This gives players little time to figure out
what is safe and where to move after the first aoe.

It is seven seconds, but this can still leave players rushing and
mis-reading the queue.

We know that swords will always spawn in all 4 corners (along with two
in the middle). We can predict the last two sword spots by intersecting
the first 4 with the expected full set. This allows us to give the same
information a few seconds earlier, which may give players enough time to
orient themselves.

Additionally, provide an alert when the initial Tertius Terminus Est
cast is started. Doing this allows players to be primed to look for the
swords earlier, and not be caught off guard.

### DIFF
--- a/ui/raidboss/data/05-shb/trial/emerald_weapon-ex.ts
+++ b/ui/raidboss/data/05-shb/trial/emerald_weapon-ex.ts
@@ -351,12 +351,19 @@ const triggerSet: TriggerSet<Data> = {
       netRegexJa: NetRegexes.startsUsing({ source: 'エメラルドウェポン', id: '55CC', capture: false }),
       netRegexCn: NetRegexes.startsUsing({ source: '绿宝石神兵', id: '55CC', capture: false }),
       netRegexKo: NetRegexes.startsUsing({ source: '에메랄드 웨폰', id: '55CC', capture: false }),
+      alertText: (_data, _matches, output) => output.text!(),
       run: (data) => delete data.tertius,
+      outputStrings: {
+        text: {
+          en: 'Watch for Swords',
+        },
+      },
     },
     {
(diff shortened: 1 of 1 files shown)


jacob-keller:
I've tested this manually (by pointing cact at the localhost raidboss.html) and at least personally very much prefer the earlier callout. It gives me more time to process which areas of the screen I need to parse to figure out safe spots. Its a little complex but its not that bad and the extra 2 or so seconds has saved my life a few runs.

jacob-keller:
> Thanks, this looks great! I'm going to merge this because I think it's a good improvement.
> 
> Even more thoughts for the future that I've had is that I think whenever it's the knockback version, it's always NE/SE -> Middle, and whenever it's the expanding circle version it's always Middle -> NE/SE. I am 95% sure that all cases are NE/SE -> Middle or Middle -> NE/SE and that you never have to go the other way. I've seen some diagrams that say otherwise, but like ~60 runs or something never seemed to have them.

I am not 100% certain of this either. Unfortunately most of my runs were unsynced so we either skipped this or panicked at this. I think you're right tho because having to go the other way would be really really difficult to pull off with the knockback

quisquous:
I'd be happy to send you a bunch of logs if you want to analyze them.

PR #4122, opened 2022-02-15
raidboss: fix overriding trigger index 0 from user file

alphaONE2:
The first trigger registered by the system (usually "General Provoke") gets index `0`. When attempting to override this trigger from a user file, the index would be looked up and checked for existence via an implicit conversion to `boolean`. However `0` converts to `false`, so a new trigger would always be created, making it impossible to override this trigger.

This fix explicitly checks both the trigger id and index against `undefined`, instead of using an implicit conversion.


quisquous:
Thank you for the fix!!

PR #4124, opened 2022-02-16
raidboss: add p4s thorn direction

quisquous:


Legends0:
comment on line 18 could be removed with this commit I think.
`// TODO: Act 4 Blue/Purple Tether strategy?`

Legends0:
If this were to be added, I think id: 'P4S Act 2 Color Tether' trigger either needs to be removed or merged as it fires simultaneously.

Also to note, In trying this out I get that it is calling the thorn the player is tethered to.  Purple tether may benefit from having the text left open longer as most strats I see set some order for breaking purple tethers based on coordinate.

quisquous:
> Purple tether may benefit from having the text left open longer as most strats I see set some order for breaking purple tethers based on coordinate.

Good thought, thanks!  I ended up writing a more strat-specific thing in <https://github.com/quisquous/cactbot-user/blob/master/raidboss/06-ew/p4s-act4.js> fwiw.

quisquous:
> If this were to be added, I think id: 'P4S Act 2 Color Tether' trigger needs to be removed or merged as it fires simultaneously.

I think it's "P4S Color Headmarkers" that is firing simultaneously.  I've removed that one.

PR #4139, opened 2022-02-18
raidboss: add trigger for Bloody Sweep (#3605)

jacob-keller:
Hansel and Gretel use "Tandem Assault: Bloody Sweep" as a combo move in
which the two bosses stand on the edge of the arena and each swipe
a large angled chunk of the platform simultaneously.

Hansel and Gretel will teleport to the side of the platform and will
sweep with their spears. Hansel is left handed and will hit the left
half while Gretel will hit the right half. Depending on which side of
the arena each is on, the safe spot to avoid getting hit will either be
between the two bosses or on the opposite side of the platform.

This can be detected using the relevant IDs, as determining whether the
spears are on the outside or inside of the arena is based on which of
the two IDs that Hansel and Gretel open the attack with.

To further complicate this, sometimes Hansel and Gretel will use
Transference after starting the Bloody Sweep but before it finishes
casting. This causes them to swap places, reversing the safe spot. This
is shown to players using a red/blue twisting tether. However, we can
also detect it via the cast time. A shorter cast time of 7.7 seconds
indicates no swap, while a longer 12.7 second cast time indicates
a Transference swap.

Using this logic we can alert the player whether the safe spot is
inbetween the two bosses, or on the opposite side.

Note that a previous implementation assumed that Hansel and Gretel
always teleported to the north side of the arena. This is incorrect, and
they appear to have multiple different teleport spots. However, using
the simpler "Move between" and "Move opposite" tells is clear as the
player can always orient relative to their starting sweep positions.

### DIFF
--- a/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
+++ b/ui/raidboss/data/05-shb/alliance/the_tower_at_paradigms_breach.ts
@@ -24,7 +24,6 @@ export interface Data extends RaidbossData {
 
 // TODO:
 //   Update Knave knockback directions to instead use cardinals
-//   Hansel and Gretel Bloody Sweep
 //   Hansel and Gretel Stronger Together Tethered
 //   Hansel & Gretel Passing Lance
 //   Hansel & Gretel Breakthrough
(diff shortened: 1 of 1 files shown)


jacob-keller:
I did some further testing of this code recently, and realized that most of the assumptions are correct. The only mistake was that the bloody sweep may not start north/south. Instead, we need to use "opposite" and "between" as a way to alert players where the safe spot is. This seems to match most of the existing parses I have proper data on, and worked in my most recent test run :)

PR #4142, opened 2022-02-18
raidboss: Rename netlog def Ability to AbilityFull

valarnin:
This is a follow up PR to #4141 for option A. This passes `npm run test` and some quick basic testing in raidemulator, but I don't have time right now to do any more extensive testing.

Might cause a weird merge issue since it also contains the same fix as the aforementioned PR.

### DIFF
--- a/resources/netlog_defs.ts
+++ b/resources/netlog_defs.ts
@@ -295,9 +295,9 @@ const logDefinitions = {
     },
     canAnonymize: true,
   },
-  Ability: {
+  AbilityFull: {
     type: '21',
-    name: 'Ability',
+    name: 'AbilityFull',
     messageType: 'ActionEffect',
     fields: {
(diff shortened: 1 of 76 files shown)


quisquous:
> I don't have any concrete numbers on performance differences between the two to make any solid conclusions. I think that ditching the partial match version is more clean from a code standpoint, though.

Yeah, I didn't have any concrete numbers when I added it either <_< and it's been too long to remember why I thought it was important.  I also think it's cleaner to not have two versions that people have to think about.

> If we pursue option B, does it make sense to mark abilityFull as deprecated to avoid having it called from TS-land, so it's only available for backward compat with user triggers?

Yeah, that's a great thought.  I always forget about mark as deprecated.

valarnin:
Closing in favor of option B merged PR.

PR #4143, opened 2022-02-18
raidboss: Refactor ability vs abilityFull

valarnin:
Per discussion on #4142, this matches "option B".

Should probably be a follow up PR to make linting check for `@deprecated` and error out, along with related cleanup?


quisquous:
Thank you!!

PR #4161, opened 2022-02-23
oopsy: don't mark abilities with one target as shares

quisquous:


quisquous:
This unfortunately needs to wait to be merged until #4196 is reverted.

PR #4164, opened 2022-02-24
raidboss: add basic triggers for Sephirot Normal mode

jacob-keller:
The sephirot normal mode boss does not have any triggers or data. This
boss has many mechanics without cast bars, and low visual tells.

Add some basic triggers to alert the player of the major mechanics like
the stack marker, earth shaker, spreading, etc. Also add a trigger to
warn about where knockbacks will land during Pillar of Mercy and to kill
the add that pops up before the three-arm wave (as failing to kill it in
time will be a wipe).

### DIFF
--- a/ui/raidboss/data/03-hw/trial/sephirot.ts
+++ b/ui/raidboss/data/03-hw/trial/sephirot.ts
@@ -37,38 +37,12 @@ const triggerSet: TriggerSet<Data> = {
       response: Responses.earthshaker(),
     },
     {
+      // The coordinates for skill are inconsistent and can't be used to
+      // reliably determine the position of the knockback.
       id: 'Sephirot Pillar of Mercy',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: '16EA', source: 'Sephirot' }),
-      alertText: (_data, matches, output) => {
-        // The pillars spawn in slightly randomized positions, typically
-        // either in the west side, center, or east side. The X position
-        // range for the arena is -20 to 20.
-        //
-        // TODO: Sometimes the coordinates for the Pillar of Mercy seem to be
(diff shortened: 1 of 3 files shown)


jacob-keller:
I titled this WIP because I found some bugs while testing it related to the X and Y positions displayed for the Pillar of Mercy attack. The coordinates usually indicate the target location, but in some cases I seem to get garbage data with incorrect x and y coordinates. This results in being unable to properly detect the knockback origin.

I still struggle to figure out how the timeline stuff works too :(

jacob-keller:
i'm not sure how to go about debugging the x/y coordinate issue.

ui/raidboss/data/03-hw/trial/sephirot.ts
```diff
@@ -0,0 +1,93 @@
...
+      netRegex: NetRegexes.startsUsing({ id: '16EA', source: 'Sephirot' }),
+      alertText: (_data, matches, output) => {
+        // The pillars spawn in slightly randomized positions, typically
+        // either in the west side, center, or east side. The X position
+        // range for the arena is -20 to 20.
+        //
+        // TODO: Sometimes the coordinates for the Pillar of Mercy seem to be
+        // wrong (an xPos of -10 for a central knockback or +12.5 for a west
+        // knockback). It's unclear if this is a bug in the parser or
+        // cactbot.
```
quisquous, line 51: For mobs that spawn and immediately start casting, sometimes they are moved into place right before they cast, and the position is wrong because it hasn't been updated in the plugin memory yet.

If this is the "blue knockback puck", I think you could just say "knockback" on its own (or nothing).

jacob-keller: The casts come from Sephirot, (but he is in big set piece mode), so I am not sure that's it. But I'll stick with just knockback and simplify it to avoid this.

PR #4168, opened 2022-02-26
resources: revert: add targetCount to 0x15/0x16 lines

quisquous:
Reverts quisquous/cactbot#4147


quisquous:
This is not enough.  I'll upload a manual patch.

PR #4169, opened 2022-02-26
resources: temporarily remove targetIndex/targetCount

### LINKED ISSUE #4165: raidboss: issues with v0.24.20
Always thank you for maintaining project. Again, here are some problem reports which were only tested in KR client. (It has had no file change since Jan 11th, 2022(v5.58), and would not be changed before v6.0 - May 10th.)

## Description
After updating to v0.24.20, I found some SB ultimates triggers did not work. It looks similar to #3415 but I'm not sure that they are KR specific problems. I saw **UWU Titan Gaols** and so UWU Titan Bomb Failure triggers do not work. Also, many community members said that in UCoB, **UCU Nael Fireball 2** trigger outputs only fireOutBeInIt no matter what ice debuff count is. Another debuff triggers, like **UCU Nael Thunderstruck** and doom counts also seem not to work. I remeber that they had no problems on Feb 15th when I was in UCoB, so I guess it has occured after v0.24.20. May you examine whether they are global cactbot issue or because of parser version as before, or something else?

## Additional information
https://drive.google.com/file/d/1sm-BEdCRNB-5tR7Miukq_ZON6int_1O0/view?usp=sharing
The UCoB log is provided by my friend.

+Minor problem: Encounters in my previous UCoB log(on Feb 15th) and UCoB log in the link above cannot be loaded in raidemulator.

quisquous:
Fixed #4165.  See discussion there for more details.

Because all fields are mandatory, adding new fields supported only on the "international" version of the ffxiv plugin breaks the Korean and Chinese versions which do not have these fields.  This causes regexes to not match.  `abilityFull` has been broken since it was added in the 6.x release, but because this is not used frequently it was likely not noticed.

Once both of these move to 6.x this patch can be reverted.  Alternatively, if we add support for versioning before then, we can add this back.


quisquous:
cc @Echoring for jobs changes

PR #4171, opened 2022-03-02
i18n: add missing translations Fr death_report

VanSlanzar:
- Again problem regex. I hope my fix is correct. If no, help me plz.
Screenshot_6

### DIFF
--- a/ui/oopsyraidsy/death_report.ts
+++ b/ui/oopsyraidsy/death_report.ts
@@ -222,6 +222,7 @@ export class DeathReport {
     const text = Translate(this.lang, {
       en: `Gain: ${effectName}`,
       de: `Erhalten: ${effectName}`,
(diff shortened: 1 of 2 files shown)


quisquous:
Does that match what you see in game? The error is that these two translations should be the same:

* https://github.com/quisquous/cactbot/blob/8eda2b4a4bca3bf3048a17d5449cf3ae1158a303/ui/raidboss/data/04-sb/trial/byakko-ex.ts#L242
* https://github.com/quisquous/cactbot/blob/8eda2b4a4bca3bf3048a17d5449cf3ae1158a303/ui/raidboss/data/04-sb/trial/byakko-ex.ts#L347

VanSlanzar:
Good ^^. It's ok for merged.

PR #4187, opened 2022-03-14
ci: Remove Version Number from Release Folder

panicstevenson:
Remove the version number from the release folder such that the release
structure will be cactbot-version.zip which contains cactbot/cactbot as
the nested folders. This will make it so in the future the folder under
ACT/Plugins for any new installs will be cactbot without a version number.

Tested with my fork by updating the GitHub Actions as well as the
VersionChecker code to point to my fork directly.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
       # Gets the numerical output of the git tag and assigns it to an environment variable
       - name: Get Tag Version
         run: |
-          echo "CACTBOT_RELEASE=cactbot-${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV
+          echo "RELEASE_VERSION=${GITHUB_REF#refs/tags/v}" >> $GITHUB_ENV
         shell: bash
       - name: Check dependencies cache
         id: cache-dependencies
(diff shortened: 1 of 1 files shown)


quisquous:
@MyTechnoHunter FYI

We considered the patch the Aditu sent, but I have had a lot of bad experiences with Windows and filesystems, and didn't want to depend on moving files and deleting files as part of the constructor.  It also meant that if we included that code, it would have to be run on every install forever into the future.  I'm somewhat more inclined to have a "fix once, never think about again" sort of solution than more code that could go awry.  No disrespect to Aditu; it's a really neat solution, but it's just more risk than I'm comfy with.

I also thought about staging some change where if you're on cactbot version X+1 then it strips 2 directories and if you're on cactbot version X+2 then it strips 1, but I don't think there's any way to wait until everybody has updated from cactbot version X to cactbot version X+1 to get the initial version of that conditional code.

So, this PR is the "low brow" approach to fixing it.  The version directory is renamed from `cactbot-0.25.23` or whatever to just `cactbot`.  For people who already have a version directory, it will continue to exist and be updated just fine.  However, new "get plugin" installs will give people a `cactbot/cactbot` directory instead of `cactbot-0.25.23/cactbot`.  That way if they ever re-get plugins, it will go into the same place and it'll prevent the "why is my cactbot not updating" class of bugs that this caused.  This approach doesn't disrupt existing users and will make future installs work better.

PR #4188, opened 2022-03-14
raidboss: Refactor log defs to support multiple versions

valarnin:
As discussed in #4165 and elsewhere, this is an attempt at addressing the issue of log format drift between versions (international, cn, ko).

This is still very much a WIP, and should function exactly as the current release as far as logic. The goal ultimately for this PR is to remove all of the default types (`lang extends Lang = 'en'` => `lang extends Lang`, remove default export for `netlog_defs.ts`).

This PR doesn't tackle extracting the version from log files for raidemulator or handling old log files in raidemulator, and in fact raidemulator still treats all log files as `en`.

For now, I'm looking mostly for feedback on the changes in `netlog_defs.ts` as that drives the changes elsewhere. Changes elsewhere are still extremely WIP.

### DIFF
--- a/resources/netlog_defs.ts
+++ b/resources/netlog_defs.ts
@@ -30,8 +30,9 @@ export type LogDefinition = {
   optionalFields?: readonly number[];
 };
 export type LogDefinitionMap = { [name: string]: LogDefinition };
+type LogDefinitionVersionMap = { [version: string]: LogDefinitionMap };
 
-const logDefinitions = {
+const latestLogDefinitions = {
   GameLog: {
     type: '00',
     name: 'GameLog',
@@ -812,12 +813,17 @@ const logDefinitions = {
   },
 } as const;
 
+export const logDefinitionsVersions = {
+  'latest': latestLogDefinitions,
+} as const;
+
 // Verify that this has the right type, but export `as const`.
-const assertLogDefinitions: LogDefinitionMap = logDefinitions;
+const assertLogDefinitions: LogDefinitionVersionMap = logDefinitionsVersions;
 console.assert(assertLogDefinitions);
 
-export type LogDefinitions = typeof logDefinitions;
+export type LogDefinitions = typeof logDefinitionsVersions['latest'];
 export type LogDefinitionTypes = keyof LogDefinitions;
+export type LogDefinitionVersions = keyof typeof logDefinitionsVersions;
 
 export type ParseHelperField<
   Type extends LogDefinitionTypes,
@@ -832,4 +838,4 @@ export type ParseHelperFields<T extends LogDefinitionTypes> = {
   [field in keyof NetFieldsReverse[T]]: ParseHelperField<T, NetFieldsReverse[T], field>;
 };
 
-export default logDefinitions;
+export default logDefinitionsVersions['latest'];
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -1,7 +1,12 @@
 import { NetParams } from '../types/net_props';
 import { CactbotBaseRegExp } from '../types/net_trigger';
 
-import logDefinitions, { LogDefinitionTypes, ParseHelperFields } from './netlog_defs';
+import {
+  logDefinitionsVersions,
+  LogDefinitionTypes,
+  LogDefinitionVersions,
+  ParseHelperFields,
+} from './netlog_defs';
 import Regexes from './regexes';
 
 const separator = '\\|';
@@ -23,19 +28,20 @@ const keysThatRequireTranslation = [
 ];
 
 const defaultParams = <
-  T extends keyof typeof logDefinitions,
->(type: T, include?: string[]): Partial<ParseHelperFields<T>> => {
-  include ??= Object.keys(logDefinitions[type].fields);
+  T extends LogDefinitionTypes,
+  V extends LogDefinitionVersions,
+>(type: T, version: V, include?: string[]): Partial<ParseHelperFields<T>> => {
+  include ??= Object.keys(logDefinitionsVersions[version][type].fields);
   const params: { [index: number]: { field: string; value?: string } } = {};
 
-  for (const [prop, index] of Object.entries(logDefinitions[type].fields)) {
+  for (const [prop, index] of Object.entries(logDefinitionsVersions[version][type].fields)) {
     if (!include.includes(prop))
       continue;
     const param: { field: string; value?: string } = {
       field: prop,
     };
     if (prop === 'type')
-      param.value = logDefinitions[type].type;
+      param.value = logDefinitionsVersions[version][type].type;
 
     params[index] = param;
   }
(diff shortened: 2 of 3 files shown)


resources/netlog_defs.ts
```diff
@@ -812,24 +814,36 @@ const logDefinitions = {
...
   Field extends keyof Fields,
+  L extends Lang = 'en',
 > = {
   field: Fields[Field] extends string ? Fields[Field] : never;
   value?: string;
 };
 
-export type ParseHelperFields<T extends LogDefinitionTypes> = {
-  [field in keyof NetFieldsReverse[T]]: ParseHelperField<T, NetFieldsReverse[T], field>;
+export type ParseHelperFields<T extends LogDefinitionTypes[L], L extends Lang = 'en'> = {
```
quisquous, line 845: (I don't think I've seen a template like this with extends and an equal.  Is this a default template parameter?)

resources/netregexes.ts
```diff
@@ -129,6 +131,7 @@ const parseHelper = <T extends LogDefinitionTypes>(
 };
 
 export default class NetRegexes {
+  public static lang: Lang = 'en';
```
quisquous, line 134: I agree with this approach too.  I think it's kind of sad to have a static variable here controlling things, but it seems like it would be obnoxious to change all NetRegexes call to explicitly take in a language variable.  And how would that even work for things like trigger files?  I think this is probably the best approach given all the tradeoffs?

resources/netlog_defs.ts
```diff
@@ -29,9 +31,9 @@ export type LogDefinition = {
   // a list of fields that are ok to not appear (or have invalid ids)
   optionalFields?: readonly number[];
 };
-export type LogDefinitionMap = { [name: string]: LogDefinition };
+export type LogDefinitionMap = { [lang in Lang]: { [name: string]: LogDefinition } };
 
-const logDefinitions = {
+const intlLogDefinitions = {
```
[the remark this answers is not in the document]
valarnin: The reason I didn't opt to go for by-version to begin with is that it would necessitate some sort of casting and version translation, e.g. detect parser version `2.6.4.1` on lang 'en' => use defs version `2.6.2.0`.

This would necessitate having a strict type, e.g:
```ts
const logDefinitions = {
  '2.2.1.0': {
    GameLog: {/*...*/},
  },
  '2.6.2.0': {
    GameLog: {/*...*/},
  },
} as const;
type ParserVersion = keyof typeof logDefinitions;
// Take a parser version number, such as 2.6.4.1, and a lang, such as 'en', and return the correct version, e.g. '2.6.2.0'
const toVersion = (version: string, lang: Lang): ParserVersion => {};
```

It's not feasible to map every single parser version to a key on the defs object, that'll cause issues with the template expansion of the net_fields.d.ts file, which expands out to everything else, due to the template possibilities exceeding built-in typescript limits (which exist for compiler performance reasons, IIRC).

resources/netregexes.ts
valarnin: The regexes/netregexes.ts files could instead export an instance of Regexes/NetRegexes but at that point it's basically the same thing (singleton vs static class).

resources/netlog_defs.ts
valarnin: Yes, this has been supported since TS 2.3, as far as I can see, it's just non-existent in their actual documentation:

https://github.com/Microsoft/TypeScript/wiki/Roadmap#23-april-2017

It's generally not a good idea for a number of reasons related to compatibility between definitions, which is why I stated that the goal before merging this PR is to remove the default type specifiers.

An example of bad compatibility, adding a default value to a function parameter for a template property with a default value:
image

resources/netlog_defs.ts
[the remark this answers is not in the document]
quisquous: > It's not feasible to map every single parser version to a key on the defs object

Yeah, definitely agreed.  My expectation is that we'd only have keys for parser versions where the logs changed (so right now we'd really only have two keys).

> The reason I didn't opt to go for by-version to begin with is that it would necessitate some sort of casting and version translation, e.g. detect parser version 2.6.4.1 on lang 'en' => use defs version 2.6.2.0.

I think in the future this would be true if we wanted to make the emulator and tools handle this, or if we wanted to somehow do this based on the actual plugin version vs the parsed language version.  In the short term, I could imagine a simple parser language -> version lookup and it could get smarter from there.  (This is a minor point, but I also don't think language is an input to this? My understanding is that the Korean and Chinese users are using an older parser version at the moment, and so parser version is sufficient to determine the defs.)

> `const toVersion = (version: string, lang: Lang): ParserVersion => {};`

Yeah, I think this is exactly what I was expecting.  This doesn't bother me too much, because I don't think it proliferates in cactbot code too much, unless I am misunderstanding.  I imagine that the templated code would change language->version in this PR and then a follow up PR would call this `toVersion` function (with a simple implementation, maybe only based on lang) and set static variables appropriately.

Could you expand on your concerns here, re: "solution isn't as clean/elegant for the majority of the cactbot code"?

valarnin:
I have poked at this a few more times and I don't think this approach will work at all with either a language-based option or a version-based option. Any attempt to implement this at the `logDefinitions` level results in type errors in trigger files as soon as there's a difference in keys between versions.

A functioning approach may be instead to always treat triggers as the latest version from a typescript/typing standpoint, and instead map into version differences in the `parseHelper` function(s). I'm still not sure exactly how to go about implementing this for raidemulator either.

quisquous:
Thanks for continuing to take a look at this.  I was going to check in and see if you were still investigating this.  Much appreciated!!

> A functioning approach may be instead to always treat triggers as the latest version from a typescript/typing standpoint

Yeah, I think from a static perspective, this is the only way I can imagine it working.  Since you can't have `matches` returning different types based on a dynamic property, then it must return the same `static` type in all cases.  I think it would make sense to have the static typing be the latest version but `parseHelper` and `exec` function dynamically handle different versions.  Ideally, the `exec` function for `Regexes` and `NetRegexes` could fill in any missing fields, and then calling trigger code can behave as if it's always there without having to worry about it.

valarnin:
I haven't done any extensive testing but here's a second pass attempt. Much simpler approach, haven't done the raidemulator side yet, only latest version definition for now.

I think the actual version number might need to come from the CS side, as a return value from `cactbotLoadData` maybe? Since it's completely possible that the overlay is loaded/reloaded after the version number log line has been emitted.

Also, maybe pull in https://www.npmjs.com/package/compare-versions for the helper function to map plugin version to logdef version?

quisquous:
Thanks for all the refactoring.  This looks good to me.  Do you think this needs more testing, or do you feel like it's ready to go?

quisquous:
> Thanks for all the refactoring. This looks good to me. Do you think this needs more testing, or do you feel like it's ready to go?

Seems fine enough to me.  o7

PR #4196, opened 2022-03-19
raidboss: add options for gamepad rumble

### LINKED ISSUE #4055: Possibility of rumble/force feedback for raidboss timeline
Been using cactbot for a long time now and I appreciate all the work gone into this over the years

Having recently gone deaf a few months ago and relying on cactbot more than I have before, it'd be nice to have the option for rumble alerts as I'm unable to hear sound alerts for raidboss anymore.

quisquous:
Adds options to enable/disable rumble, along with per-severity settings
for rumble.  There are not per-trigger settings for this as that seemed
like overkill, but if folks really want it, feel free to submit a PR.

Fixes #4055.  (Thanks, @valarnin!!)


ui/raidboss/raidboss_options.ts
```diff
@@ -94,6 +94,16 @@ const defaultRaidbossConfigOptions = {
   InfoSoundVolume: 1,
   LongSoundVolume: 1,
   PullSoundVolume: 1,
+  RumbleEnabled: false,
+  InfoRumbleDuration: 400,
+  InfoRumbleWeak: 0.5,
+  InfoRumbleStrong: 0,
```
panicstevenson, line 100: Is there any information as to what a constitutes a 'weak' vs 'strong' rumble?
For instance, I'm surprised that any strong number is lower than a weak number because I would have assumed they're a spectrum or sine wave that the rumble goes through.

panicstevenson (approved):

ui/raidboss/popup-text.ts
```diff
@@ -1194,6 +1212,30 @@ export class PopupText {
...
+      if (!gp)
+        return;
+
+      // `vibrationActuator` is supported in CEF but is not in the spec yet.
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const gamepad: any = gp;
+
+      // eslint-disable-next-line max-len
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,@typescript-eslint/no-unsafe-call
+      void gamepad?.vibrationActuator?.playEffect(gamepad.vibrationActuator.type, {
```
panicstevenson, line 1230: For multiple triggers in a row, what does this look like? Does it reset the vibration effect and start over? Do they overlap?

ui/raidboss/raidboss_options.ts
valarnin: "Strong" and "Weak" are general conventions. Gamepads can also just have two motors that have the same strength, at which point it's less "Strong"/"Weak" and more "Left"/"Right".

From the MDN documentation:
```
strongMagnitude: Rumble intensity of the low-frequency (strong) rumble motors, normalized to the range between 0.0 and 1.0.
weakMagnitude: Rumble intensity of the high-frequency (weak) rumble motors, normalized to the range between 0.0 and 1.0.
```

For instance, in most shooters, the "strong" motor will fire for explosions with a high magnitude, with the "weak" motor firing at low magnitude, whereas for gunfire the "weak" motor is used for most types of gun.

ui/raidboss/popup-text.ts
valarnin: Also from the MDN documentation:
```
Note: A new call to playEffect() overrides a previous ongoing call.
```

ui/raidboss/popup-text.ts
```diff
@@ -1194,6 +1212,30 @@ export class PopupText {
     this._addTextFor('info', triggerHelper);
   }
 
+  _onTriggerInternalRumble(triggerHelper: TriggerHelper): void {
+    if (!this.options.RumbleEnabled)
+      return;
+
+    // getGamepads returns a "GamePadList" which isn't iterable.
+    [...navigator.getGamepads()].forEach((gp) => {
+      if (!gp)
```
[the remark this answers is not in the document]
valarnin: From MDN documentation:

```
Elements in the array may be null if a gamepad disconnects during a session, so that the remaining gamepads retain the same index. 
```

ui/raidboss/popup-text.ts
quisquous: For a single response with both alarm and info, only the alarm rumble will play.  For sequential triggers it will override, as valarnin says above.

I think this is probably fine enough.  Half of the point of a rumble (or sound) is to say "hey look listen" and then partially for the severity itself to indicate extra information.  I would argue that multiple sounds playing is also pretty muddy and we could think about avoiding that too as a more general case of behavior, however neither of them of those feels required to fix here and it'd be "good enough" without that.

I can leave a comment about this behavior, for sure.

quisquous: Other options here could be including a timestamp for when the current rumble will end and not starting a new rumble until that rumble is done? But that could create a priority inversion of info -> alarm where the alarm wouldn't play.

panicstevenson: Reading the Chromium source code it looks like you get some form of a callback, but even then I don't know how much we'd care to figure all of that out; simply looking at if there's a callback wouldn't be enough and we'd have to store and inspect them. I think following a general principle of "Don't call out multiple things at once, and, if you have to, then call them out in ascending priority order" is already somewhat adhered to, so maybe it just ends up being okay.

quisquous: Yeah, `playEffect` returns a promise when it is `'complete'`, but not clear to me if the callback occurs when the entire effect is done or just when it's submitted.  That's an option too.

Unless you feel strongly, I think my inclination here is to do the simplest thing and when we find a bad case, we can think more about how to handle it.

PR #4203, opened 2022-03-21
raidboss/config: add ui for timeline text enable/renaming

quisquous:
This adds a new "Edit Timeline" section to the config ui for any trigger set that has a timeline.  On a per text-basis, it allows for hiding all instances of a particular text or renaming it.  The primary use cases are to for folks who think the timeline is too messy or who want to rename things on the timeline (e.g. `Elegant Evisceration` to `Tank Buster`) or some such.

image

### DIFF
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -118,12 +118,12 @@ export class Timeline {
   public ui?: TimelineUI;
 
   constructor(
-    private zoneId: number,
     text: string,
     replacements: TimelineReplacement[],
     triggers: LooseTimelineTrigger[],
     styles: TimelineStyle[],
     options: RaidbossOptions,
+    private zoneId: number,
   ) {
     this.options = options || {};
     this.replacements = replacements;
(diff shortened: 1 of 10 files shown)


ui/raidboss/timeline.ts
```diff
@@ -150,7 +151,14 @@ export class Timeline {
...
 
   private LoadFile(text: string, triggers: LooseTimelineTrigger[], styles: TimelineStyle[]): void {
-    const parsed = new TimelineParser(text, this.replacements, triggers, styles, this.options);
+    const parsed = new TimelineParser(
+      text,
+      this.replacements,
+      triggers,
+      styles,
+      this.options,
+      this.zoneId,
```
panicstevenson, line 160: Very nit, if you want: weird to have zoneId be last here when it comes before text everywhere else?

panicstevenson (approved):

ui/raidboss/timeline.ts
quisquous: Mmm in this case it's because it's optional but text is not optional.  It's also not optional in the other paths.  Do you think I should put it last elsewhere?

panicstevenson: No, I feel very not strongly, just something I saw

quisquous: Done, all the same! :+1:

PR #4211, opened 2022-03-25
util: upgrade ts-node for native ESM support

MaikoTan:
ts-node has native ESM support in [v10.7.0 version](https://github.com/TypeStrong/ts-node/releases/tag/v10.7.0),
although according to the [related PR](https://github.com/TypeStrong/ts-node/pull/1655) says,
they are just spawn a child process that passed `--loader ts-node/esm`
options to node itself.
But no more painful `ExperimentalWarning: --experimental-loader is an experimental feature.` message in the output.

Also, upgrade `typescript` and `typescript-eslint` to the newer version.

------
By the way, seems like `ts-node` has added `swc` transpiler several versions ago,
and according to `swc` or `ts-node`, that this transpiler is significantly faster a lot.
But appearly `swc` has no ability to check the typescript type.
So I am wondering whether or not we can switch to this one?


MaikoTan:
`npm run test` is still using the `loader` option for node with `ts-node/esm`, but seems like mocha has hacked the node `require` to do its trick to support ESM which is not compatible with `ts-node` currently.
I tried several solution to make mocha work with `ts-node`'s ESM support but the only one that worked is the weirdo `ts-node node_modules/.bin/mocha`.(＠_＠;)
So the only continuing reporting `ExperimentalWarning: --experimental-loader is an experimental feature.` is the test script. Folks in the `ts-node` org said that is a rabbit hole. ┑(￣Д ￣)┍

quisquous:
Thanks for this.  Sorry I haven't merged it yet, I wanted to look at this locally a bit more.

quisquous:
> But appearly swc has no ability to check the typescript type.
> So I am wondering whether or not we can switch to this one?

What do you mean, by this?

MaikoTan:
> 

I mean that although `ts-node` is using `tsc` to transpile TypeScript code by default, but it can also use `swc` to transpile.
according to `swc` or `ts-node`, `swc` is much faster than `tsc`. You see, every time we run those TypeScript-written scripts, it always take about 1 second to transpile code. If we set the `swc` as the transpiler in `ts-node`, it takes much shorter time before running the script actually, but `swc` transpiler doesn't do any type-checking things, so I am not very sure whether it is worth to switch transpiler to `swc` in this case.

quisquous:
Ah, ok!  I think it'd be nice to have type checking for the dev server as well as at least *some* type checking for all the scripts (even if it's a separate step), but if there's some option to run swc in some places but not others and not lose any type checking coverage, that seems like an option.

quisquous:
Overall though, I think this looks good.  swc changes can come in some other PR.  Thanks!

PR #4219, opened 2022-03-27
jobs: WAR: more timer box

Echoring:
war
- add Upheaval and Inner Release Timer.
- minor name change

Due to #3853 changes, those jobs with less than 3 box may appear weird like following so add some boxes.
weird


quisquous:
> Due to https://github.com/quisquous/cactbot/pull/3853 changes, those jobs with less than 3 box may appear weird like following so add some boxes.

We could also change the CSS to behave differently when there's fewer boxes, but I think more boxes here is good.

PR #4221, opened 2022-03-29
build: use esbuild to accelerate build step

MaikoTan:
Before:
```
webpack 5.70.0 compiled with 2 warnings in 18487 ms                                                                                                                           

real    0m25.407s
user    0m40.822s
sys     0m3.779s
```

After:
```
webpack 5.64.4 compiled with 2 warnings in 2681 ms

real    0m5.872s
user    0m7.895s
sys     0m0.671s
```

Well, data tells all.
The only concern is that `esbuild` is still in beta, not in stable level, and release rapidly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,6 +53,7 @@
         "css-loader": "^6.5.1",
         "css-minimizer-webpack-plugin": "^3.0.1",
         "dprint": "^0.15.3",
+        "esbuild-loader": "^2.18.0",
         "eslint": "^7.14.0",
         "eslint-config-google": "^0.9.1",
         "eslint-import-resolver-typescript": "^2.4.0",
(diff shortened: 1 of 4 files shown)


quisquous:
> Well, data tells all.

WOW :tada:

> The only concern is that esbuild is still in beta, not in stable level, and release rapidly.

I mean, we don't have to take releases as fast as they appear if we don't want them.  So, I don't see "release rapidly" as a particular issue.

Re: "still in beta".  I think what makes this an easier choice for me is that it is just esbuild-loader.  It seems like it would be easy to undo this patch in the future if for some reason esbuild stopped being supported or working.  If this was an entire switch from webpack to esbuild, it would be more of a long-term risk that I'd want to think about more.  However, just esbuild-loader is different.

Other than concerns about the risk of a new dependency, do you see any reason not to merge this?

quisquous:
npm start without patch:
webpack 5.64.4 compiled successfully in 108741 ms
webpack 5.64.4 compiled successfully in 118455 ms
webpack 5.64.4 compiled successfully in 119013 ms

npm start with patch:
webpack 5.64.4 compiled successfully in 106576 ms
webpack 5.64.4 compiled successfully in 112268 ms
webpack 5.64.4 compiled successfully in 114843 ms

Maybe my computer  is just very slow and esbuild doesn't fix whatever makes it slow?

MaikoTan:
> Maybe my computer  is just very slow and esbuild doesn't fix whatever makes it slow?

I think it didn't supposed to compile such that long. May I ask which OS and shell are you using? Also what is your node js and npm version?

What I am using is Debian 11 and bash, with node 16.14.0 and npm 8.3.1, and this patch works well on my side.

quisquous:
> > Maybe my computer  is just very slow and esbuild doesn't fix whatever makes it slow?
> 
> I think it didn't supposed to compile such that long. May I ask which OS and shell are you using? Also what is your node js and npm version?

I'm using Windows 10, and mingw64 shell.  npm is 8.10 and node is 16.12.0.  I'm running `npm start` to get those times.

trim21:
and `ts-node/esm` is broken at node v16.17.0...

MaikoTan:
Found that the `ForkTsCheckerWebpackPlugin` was added in #3294 that intend to accelerate ...

trim21:
back then we are still using ts-loader...

PR #4222, opened 2022-03-30
raidboss: extend duration for diamond ex platform switch

jacob-keller:
When the Diamond Weapon is going to swipe a platform, an info message is
presented to inform the user of which side to move to and what attack
will follow this.

The default 5 second duration causes the info message to be gone well
before the attacks happen. Extend this duration to 15 seconds, lining up
better with the overall mechanic. This is ok because no other alerts
happen during this time.

The longer duration allows users to see what is happening more easily,
giving time to notice it and response without needing to memorize it for
10 seconds while focusing on other aspects of the fight.


quisquous:
Thanks for the improvement! I think that makes a lot of sense.

jacob-keller:
I've got a few other diamond ex improvements but they need more work. Hope to post them soon :)

PR #4233, opened 2022-04-05
raidboss: Ultima EX timeline and minimal triggers

JLGarber:
Ultima is a mostly-typical ARR-style encounter. It consists of four phases each with its own rotation, with each activating in turn at HP thresholds. Phases 1/2/4 are reasonably well-behaved and presented basically no problems working out. Phase 3 appears to be incredibly convoluted almost beyond belief.

 Triggers first though. As is normal for ARR-era encounters, the vast majority of Ultima's abilities have no cast times. This means most of our triggers will have to be timeline-driven. This is mostly not an issue, but we will discuss the exceptions later. For now, the major important triggers are in place. (Raidwide, tank swap, and off-tank cleave.)

The major issue experienced with generating this timeline was in phase 3. At 64.9% HP, Ultima enters its Ifrit phase, where it (mostly) alternates ability blocks centered on Eruption or Crimson Cyclone. Most of the ridiculousness is noted in-line in the comments, but it is *astonishing* how much variability Square seems to have put into what is likely the shortest of the phases! I haven't yet been able to fully test whether this variability is fully random or whether it might be some HP-based thresholds within the phase. (Unfortunately it can take 2-3 minutes of waiting around within phase 3 t o see whether Ultima does something or not, so it's not easy to test.)

Phase 4 also deserves a mention. It looks like this encounter uses 21 log lines instead of 27 for the Aetheroplasm headmarkers in phase 4. (I'll be adding those into the PR, both timeline and triggers, once maintenance is over and I can test them.) This would fit in with the lack of 27 log lines in other ARR-era content.

The transition to phase 4 is indicated by the first cast of Aetheric Boom, and the rotation starts fresh from there. None of the Aetheric Boom casts appear to delay the timeline by more than a few seconds. The appearance of the Ultima Bits similarly causes no delay. The gunship crashes, inflicting proximity damage with Freefire, seem to interrupt the rotation and postpone whatever they interrupted by about 12 seconds. All this is to say, we need to sync phase 4 much more aggressively than we normally would. Fortunately Tank Purge occurs only once per rotation, so we do have an if-all-else-fails stopgap.

Finally, Homing Lasers needs special mention. There's a point in the phase 3 rotation at 732 where a shorter "bridge" block between two standard ability blocks is indicated by a Homing Lasers usage. Unfortunately, we will have no way to definitively warn the player if and only if Homing Lasers is used here. I think we'll just have to accept the possibility of a phantom tank cleave callout here and hope for the best. 

Realistically, most groups should clear phase 3 long before any of this becomes a problem, but we do need to do our best because of how timeline-dependent the trigger set is.

### DIFF
--- a/ui/raidboss/data/02-arr/trial/ultima-ex.ts
+++ b/ui/raidboss/data/02-arr/trial/ultima-ex.ts
@@ -32,7 +32,7 @@ const triggerSet: TriggerSet<Data> = {
     {
       // At 5 stacks of Aetheroplasm, the target begins taking massive damage.
       id: 'Ultima EX Viscous Aetheroplasm',
-      netRegex: NetRegexes.gainsEffect({ effectId: '171', count: '4' }),
+      netRegex: NetRegexes.gainsEffect({ effectId: '171', count: '04' }),
       response: Responses.tankBusterSwap(),
     },
   ],
--- a/ui/raidboss/data/02-arr/trial/ultima-ex.ts
+++ b/ui/raidboss/data/02-arr/trial/ultima-ex.ts
@@ -4,17 +4,26 @@ import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  plasmTargets?: Array<string>;
+  boomCounter: number;
+}
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.TheMinstrelsBalladUltimasBane,
   timelineFile: 'ultima-ex.txt',
+  initData: () => {
+    return {
+      plasmTargets: [],
+      boomCounter: 0,
+    };
+  },
   timelineTriggers: [
     {
       // Early Callout for Tank Cleave
       id: 'Ultima EX Homing Lasers',
       regex: /Homing Lasers/,
-      beforeSeconds: 5,
+      beforeSeconds: 4,
       alertText: (_data, _matches, output) => output.text!(),
       outputStrings: {
         text: {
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/02-arr/trial/ultima-ex.txt
```diff
@@ -0,0 +1,287 @@
...
+797.3 "--sync--" sync / 14:[^:]*:The Ultima Weapon:5F4:/ window 2.5,0
+800.0 "Eruption x5" duration 8 #sync / 1[56]:[^:]*:The Ultima Weapon:5F4:/
+810.6 "Tank Purge" sync / 1[56]:[^:]*:The Ultima Weapon:5EA:/ window 15,15
+813.9 "Homing Lasers" sync / 1[56]:[^:]*:The Ultima Weapon:5E1:/
+819.2 "Ceruleum Vent" sync / 1[56]:[^:]*:The Ultima Weapon:5E0:/
+822.5 "Viscous Aetheroplasm" sync / 1[56]:[^:]*:The Ultima Weapon:5DF:/
+
+825.1 "--sync--" sync / 14:[^:]*:The Ultima Weapon:5F1:/ window 15,15 jump 697.3 # Radiant  Plume
+827.8 "Radiant Plume?"
+829.8 "Homing Lasers?" sync / 1[56]:[^:]*:The Ultima Weapon:5E1:/window 10,10 jump 900
```
quisquous, line 209:
```suggestion
829.8 "Homing Lasers?" sync / 1[56]:[^:]*:The Ultima Weapon:5E1:/ window 10,10 jump 900
```

ui/raidboss/data/02-arr/trial/ultima-ex.ts
```diff
@@ -0,0 +1,41 @@
...
+    {
+      id: 'Ultima EX Tank Purge',
+      netRegex: NetRegexes.startsUsing({ id: '5EA', source: 'The Ultima Weapon', capture: false }),
+      response: Responses.bigAoe(),
+    },
+    {
+      // At 5 stacks of Aetheroplasm, the target begins taking massive damage.
+      id: 'Ultima EX Viscous Aetheroplasm',
+      netRegex: NetRegexes.gainsEffect({ effectId: '171', count: '04' }),
+      response: Responses.tankBusterSwap(),
```
quisquous, line 36: I think this isn't "tank buster swap" and more just "tank swap"?

quisquous (commented):
Are there other triggers you think should be added here? (Or should we merge this first and then add more?)

Mostly thinking of:
* are there phase 1 in/out things?
* aetheric boom (at least maybe "pop orbs (some word describing the pattern)"?)
* freefire places to run or maybe it's obvious from the ground and we can skip it?

Mostly curious what you think is needed here, trigger-wise.

JLGarber:
> Are there other triggers you think should be added here? (Or should we merge this first and then add more?)
> Mostly thinking of:
     - are there phase 1 in/out things?
     - aetheric boom (at least maybe "pop orbs (some word describing the pattern)"?)
     - freefire places to run or maybe it's obvious from the ground and we can skip it?

I think it's mostly there, yes. You hit on the major ones already, the only other one I would probably add is a warning for "You have been targeted by a homing Aetheroplasm", which I still have to add/test.

Aetheric Boom, yes, you're right, we should add a Boom counter and notify which it is based on that. Should be simple enough. Freefire is the standard proximity markers, and we haven't been completely consistent with whether or not we call those. Ordinarily I would say we shouldn't bother, but given that the modern convention is to call stuff and let the user mute it if needed, it might be better to see whether there's consistency in the locations.

I'll be adding some trigger updates later, but unfortunately I've run into a configuration issue of some sort that's making the test suite fail any new trigger additions. Once I get that worked out I'll send in some trigger updates.

JLGarber:
Okay, this is untested in-game since I had a surprise mount farm tonight, but I think this is close to the final trigger set at least. I wasn't able to find any data on where the Freefire proximity AoEs come from, since the Assault Craft that actually cast thoseare spawned at the start, and there aren't any `startsUsing` lines. I haven't mathed out whether Freefire comes at a set time after Booms, or whether it's on a separate HP threshold, but I can look for that.

Other than that, I think it's Good Enough. There will inevitably be minor tweaks to the phase 3 timeline, but it should function for 90% of what people need.

ui/raidboss/data/02-arr/trial/ultima-ex.ts
```diff
@@ -0,0 +1,126 @@
...
+      plasmTargets: [],
+      boomCounter: 0,
+    };
+  },
+  timelineTriggers: [
+    {
+      // Early Callout for Tank Cleave
+      id: 'Ultima EX Homing Lasers',
+      regex: /Homing Lasers/,
+      beforeSeconds: 4,
```
quisquous, line 26:
```suggestion
      beforeSeconds: 5,
```

Unless there's a timeline reason, for consistency do 5 seconds before knockback in case somebody wants to arm's length? (if it works)

ui/raidboss/data/02-arr/trial/ultima-ex.ts
```diff
@@ -0,0 +1,126 @@
...
+      outputStrings: {
+        0: {
+          en: 'Boom Orbs: 1x at cardinals',
+        },
+        1: {
+          en: 'Boom Orbs: 1x N/S, 2x E/W',
+        },
+        2: {
+          en: 'Boom Orbs: 2x at intercardinals',
+        },
```
quisquous, line 120: Minor brevity suggestion, which you can take or leave.

```suggestion
      outputStrings: {
        0: {
          en: 'Orbs: Cardinals',
        },
        1: {
          en: 'Orbs: Cardinals (N/S first)',
        },
        2: {
          en: 'Orbs: Intercardinals',
        },
```

These "strings" 0, 1, 2 are shown to the user in the config ui.  Would it make sense to start at 1, or use `boom${data.boomCounter}` so the strings are `boom1`, `boom2`, etc?

ui/raidboss/data/02-arr/trial/ultima-ex.ts
```diff
@@ -0,0 +1,126 @@
...
+    {
+      // At 5 stacks of Viscous Aetheroplasm, the target begins taking massive damage.
+      id: 'Ultima EX Viscous Aetheroplasm',
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: '171', count: '04', capture: false }),
+      condition: (data) => data.role === 'tank',
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Tank Swap Now--4 Stacks',
```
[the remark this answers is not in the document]
JLGarber: Ah, rightright, I was looking under `responses` and forgot that bare `outputs` are also a thing.

ui/raidboss/data/02-arr/trial/ultima-ex.ts
JLGarber: This isn't the knockback, this is the splashing off-tank buster. However, I need to do some additional checking, because one video I saw suggested that this is targeted on anyone not in melee range, not the person who has second threat.

ui/raidboss/data/02-arr/trial/ultima-ex.ts
JLGarber: You're right, I think I'll combine those just to make it easier for everyone.

ui/raidboss/data/02-arr/trial/ultima-ex.ts
quisquous: Sorry this comment was meant for line 37 below on Vulcan Burst.

ui/raidboss/data/02-arr/trial/ultima-ex.ts
```diff
@@ -0,0 +1,126 @@
+import NetRegexes from '../../../../../resources/netregexes';
+import outputs from '../../../../../resources/outputs';
```
quisquous, line 2: (this needs to be Outputs or it won't work in user files when eval'd)

PR #4235, opened 2022-04-07
raidemulator: Fix error in PopupTextAnalysis

### LINKED ISSUE #4214: Raidemulator behaves unexpectedly on preRun/alertText split
See patch in #4213.  If you run that patch against the log in 4212, then it works properly.

However, if you keep the previous `preRun`/`alertText` split by applying the patch below, then it will emit many alert texts for `TEA Trine Initial`.

```diff
diff --git a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts b/ui/raidboss/
data/05-shb/ultimate/the_epic_of_alexander.ts
index 302bef930..33cff3d92 100644
--- a/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts
+++ b/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts
@@ -2768,7 +2768,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegexJa: NetRegexes.ability({ source: 'パーフェクト・アレキサンダー', id: '
488F' }),
       netRegexCn: NetRegexes.ability({ source: '完美亚历山大', id: '488F' }),
       netRegexKo: NetRegexes.ability({ source: '완전체 알렉산더', id: '488F' }),
-      alertText: (data, matches, output) => {
+      preRun: (data, matches) => {
         // Looking for (100, 92), (100, 100), or (100, 108).
         const x = Math.round(parseFloat(matches.x));
         const y = Math.round(parseFloat(matches.y));
@@ -2787,7 +2787,8 @@ const triggerSet: TriggerSet<Data> = {
         if (!thisTrine)
           return;
         data.trine.push(thisTrine);
-
+      },
+      alertText: (data, _matches, output) => {
         // Call out after two, because that's when the mechanic is fully known.
         data.trine ??= [];
         if (data.trine.length !== 2)
```

My understanding of how `preRun` and `alertText` should work is that there should be zero difference in a non-async trigger (one without `delaySeconds` or `promise`) whether these are split, and they should run synchronously.

valarnin:
Fixes #4214.

Logic error occurred only when multiple triggers were matched the same line and those triggers depended on each other's output.

Since resolved status is being checked per-line, the call to `this.checkResolved` between the two loops was no longer needed, instead it only needs to run at the end of the outer `for` loop to resolve older triggers.

### DIFF
--- a/ui/raidboss/emulator/data/PopupTextAnalysis.ts
+++ b/ui/raidboss/emulator/data/PopupTextAnalysis.ts
@@ -130,7 +130,6 @@ export default class PopupTextAnalysis extends StubbedPopupText {
           suppressed: false,
           executed: false,
         });
-        this.triggerResolvers.push(resolver);
 
         this.OnTrigger(trigger, r, logObj.timestamp);
 
@@ -140,9 +139,12 @@ export default class PopupTextAnalysis extends StubbedPopupText {
           if (this.callback)
             this.callback(logObj, resolver.triggerHelper, resolver.status, this.data);
         });
-      }
 
-      await this.checkResolved(logObj);
+        const isResolved = await resolver.isResolved(logObj);
+
+        if (!isResolved)
+          this.triggerResolvers.push(resolver);
+      }
 
       for (const trigger of this.netTriggers) {
         const r = trigger.localNetRegex?.exec(logObj.networkLine);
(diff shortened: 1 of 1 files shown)


quisquous:
Thank you for tracking this down!

PR #4244, opened 2022-04-09
raidboss: Ultima EX updates

JLGarber:
Live testing and some video watching shook out a couple of mistakes.

Homing Lasers seems to be similar to Liquid Hell, although there's some dispute over whether it's on the farthest target or "if 1+ targets are out of melee, target one at random, else target random in melee". I'm not sure whether maybe we should just have a "Homing Lasers!" warning and leave it at that?

Diffractive Laser's addition and Homing Aetheroplasm's fix should be self-explanatory.

Should we maybe offset the callouts for the Boom orbs vs the Boom knockback? They are different severity levels, so there's no overlap in that sense, but they do come up at the same time.

The timeline desync in phase 3 that I noted in a previous conversation was simply from a sometimes-present Viscous Aetheroplasm. I don't exactly like this way of handling it, but there aren't a lot of more graceful ways to do it. (And we've used this "solution" for Ravana EX previously anyway.)

### DIFF
--- a/ui/raidboss/data/02-arr/trial/ultima-ex.txt
+++ b/ui/raidboss/data/02-arr/trial/ultima-ex.txt
@@ -80,7 +80,7 @@ hideall "--sync--"
 318.6 "Homing Lasers" sync / 1[56]:[^:]*:The Ultima Weapon:5E1:/
 323.9 "Ceruleum Vent" sync / 1[56]:[^:]*:The Ultima Weapon:5E0:/
 327.2 "Viscous Aetheroplasm?" #sync / 1[56]:[^:]*:The Ultima Weapon:5DF:/
-332.4 "--sync--" sync / 14:[^:]*:The Ultima Weapon:5F1:/window 10,2.5
+332.4 "--sync--" sync / 14:[^:]*:The Ultima Weapon:5F1:/ window 10,2.5
 335.1 "Radiant Plume" sync / 1[56]:[^:]*:The Ultima Weapon:5F1:/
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/02-arr/trial/ultima-ex.txt
```diff
@@ -71,38 +71,41 @@ hideall "--sync--"
...
-340.8 "Crimson Cyclone" sync / 1[56]:[^:]*:Ultima Ifrit:5F5:/
-342.0 "Radiant Plume" sync / 1[56]:[^:]*:The Ultima Weapon:5F1:/
-347.6 "Homing Lasers" sync / 1[56]:[^:]*:The Ultima Weapon:5E1:/
-
-355.5 "Eruption x5" duration 8 #sync / 1[56]:[^:]*:The Ultima Weapon:5F4:/
-360.0 "Ceruleum Vent" sync / 1[56]:[^:]*:The Ultima Weapon:5E0:/
-363.4 "Viscous Aetheroplasm" sync / 1[56]:[^:]*:The Ultima Weapon:5DF:/
-372.1 "Tank Purge" sync / 1[56]:[^:]*:The Ultima Weapon:5EA:/ window 15,15
-375.5 "Homing Lasers" sync / 1[56]:[^:]*:The Ultima Weapon:5E1:/
+327.2 "Viscous Aetheroplasm?"# sync / 1[56]:[^:]*:The Ultima Weapon:5DF:/
```
quisquous, line 82:
```suggestion
327.2 "Viscous Aetheroplasm?" #sync / 1[56]:[^:]*:The Ultima Weapon:5DF:/
```

PR #4246, opened 2022-04-10
docs: Add documentation for TransformTts customization

valarnin:
Not sure if there's a need for more verbiage around the description, or a more complete example maybe via switch/case matching some default TTS outputs?


quisquous:
Thanks! I think this is good enough, and we can point people at it.

PR #4247, opened 2022-04-11
raidboss: zodiark ex

quisquous:
This is #3933, rebased, and with a few minor fixes from running it on top.

Closes #3933.

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
@@ -8,17 +8,12 @@ import { PluginCombatantState } from '../../../../../types/event';
 import { TriggerSet } from '../../../../../types/trigger';
 
 export interface Data extends RaidbossData {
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -264,31 +264,32 @@ const triggerSet: TriggerSet<Data> = {
...
-          case 2:
-            return output.lookforpython!();
-          case 3:
-            return output.text!();
-          default:
-            return output.lookforpython!();
-        }
+        if (data.adikiaCounter % 2 === 1)
+          return output.lookforpython!();
+        return output.text!();
```
JLGarber, line 285: I'm probably confused here, and this will *work*, but why are we counting them? There are only two, the first with snakes and the second without. Maybe the counter could instead just be a boolean that gets set after the first one?

quisquous: Oh, thanks.  I didn't even think about it and just rewrote the code as-is.

PR #4270, opened 2022-04-13
i18n: EW ex trial 3 translations

Akurosia:


quisquous:
@valarnin Is there a way to not use `blueStarNames` and collect ids from combatants instead? What if there are players named those things?

valarnin:
> @valarnin Is there a way to not use `blueStarNames` and collect ids from combatants instead? What if there are players named those things? Sorry for only noticing this belatedly here as Akurosia pointed this out to me as making it harder to translate things.

There's no way to do a collector setup for these unfortunately, the only other real option would be to just not check the combatant name at all, e.g. `netRegex: NetRegexes.startsUsing({ id: '6FFA', capture: true }),`. There should never be a case where a player is using that ability ID so it shouldn't really matter regardless?

Also I think this trigger can be simplified to not require the call to `getCombatants` using the same `heading` technique used for the 4-star trigger, that trigger came after this one.

PR #4272, opened 2022-04-14
raidboss: Endsinger Ex star trigger tweaks

valarnin:
I mentioned these changes in a reply to #4270

Basically, remove the star name requirement entirely (change 1) and remove the need for calls to `getCombatants` (change 2).

### DIFF
--- a/ui/raidboss/data/06-ew/trial/endsinger-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/endsinger-ex.ts
@@ -87,17 +87,13 @@ const getAoEOrbSafeDir = (posX: number, posY: number, output: Output): string | 
 };
 
 const getStarPositionFromHeading = (heading: string) => {
-  switch (parseFloat(heading)) {
-    case 0.79: // SE
-      return [114, 114];
-    case -2.36: // NW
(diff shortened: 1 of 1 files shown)


panicstevenson (approved):

ui/raidboss/data/06-ew/trial/endsinger-ex.ts
```diff
@@ -106,6 +86,51 @@ const getAoEOrbSafeDir = (posX: number, posY: number, output: Output): string |
   return output.nw!();
 };
 
+const getStarPositionFromHeading = (heading: string) => {
+  switch (parseFloat(heading)) {
```
quisquous, line 90: Sorry for not saying this the first time around, but would you mind doing some floating point comparison here (and also on the zero version)? Or doing something more like the "round a heading into an intercardinal direction" approach here? I can buy that this all works now, but mostly I'd prefer to future proof this around level of precision and not doing a string comparison against a floating point number.

valarnin: Updated in latest commit, I think this is what you were suggesting?

PR #4279, opened 2022-04-15
i18n: zodiark-ex-cn-fix

ShadyWhite:
In this trigger, the number of stacks is not fixed 6 times, but incremental, should it be modified?
https://github.com/quisquous/cactbot/blob/b6dafe50d41371ba2b49de49eb8aa6fd0f829b4b/ui/raidboss/data/06-ew/trial/zodiark-ex.ts#L88-L100

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
@@ -97,7 +97,7 @@ const triggerSet: TriggerSet<Data> = {
         text: {
           en: 'Stack x6',
           de: 'Sammeln x6',
-          cn: '6次分摊',
+          cn: '多重分摊',
         },
       },
     },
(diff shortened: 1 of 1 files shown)


quisquous:
I think it's not good for one language to be translated differently.

It looks like it starts at 6 and goes to 9 and never goes above 9.  Do you want to add a counter and a number in the trigger?

ShadyWhite:
> I think it's not good for one language to be translated differently.
> 
> It looks like it starts at 6 and goes to 9 and never goes above 9. Do you want to add a counter and a number in the trigger?

It is not necessary to do so, but only to modify the alert text to "Multiple Stacks".

quisquous:
> > I think it's not good for one language to be translated differently.
> > It looks like it starts at 6 and goes to 9 and never goes above 9. Do you want to add a counter and a number in the trigger?
> 
> It is not necessary to do so, but only to modify the alert text to "Multiple Stacks".

Hmm, this sounds worse to my ears in English.  "Stack x6" is imperative, e.g. "go stack six times" which is in line with other imperative "go left" or "go right" sorts of trigger texts.  If you don't want to make this change, I'll get to it eventually.

PR #4280, opened 2022-04-15
i18n: Better cn config

### LINKED ISSUE #4274: Discussion: Chinese Translations for "Alert" in Raidboss Module
This is the following up for https://github.com/quisquous/cactbot/pull/4264 that changes the translation in raidboss config.
But there are ambiguities founded in the Chinese translations for the word "Alert".

## Meaning

As far as I know, the "Alert" in the config refers to two different things:

1. The flying text that coming up, and/or the TTS / sound that let user what is coming or how can they do.
2. A importance level of the "Alert" in 1) , by distinguished as "info", "alert" and "alarm".

### For 1)

There are several translations in different configs, some are "警报"/"警告" (which are ambiguitied with 2), and not very comprehensive), some are "提示" or "提示文本" (which are partially translated, but "提示" might be a good word for this meaning).

So these should be consist.

https://github.com/quisquous/cactbot/blob/a055416ecf67b8276fcb334b10b26db265d9d98e/ui/raidboss/raidboss_config.ts#L1442-L1451

https://github.com/quisquous/cactbot/blob/a055416ecf67b8276fcb334b10b26db265d9d98e/ui/raidboss/raidboss_config.ts#L1500-L1509
(This is a wrong translation as the option not only control the "text")

### For 2)
The current translations for these three level are "信息", "警告" and "警报", basically the literal translation. Translating "info" to 信息 would not lead to problem, but many Chinese cannot distinguish "警告" and "警报", also these two words in Chinese is not commonly used to describe an importance level (Instead, Chinese would add an extra word like "警告级别"(Alert level)).

https://github.com/quisquous/cactbot/blob/a055416ecf67b8276fcb334b10b26db265d9d98e/ui/raidboss/raidboss_config.ts#L206-L237

------

cc @sandtechnology @ShadyWhite @Echoring

ShadyWhite:
Fix #4274 .These changes should make it more intuitive for Chinese players to distinguish the importance level. I hope this change can be mentioned in the release note, so that Chinese players can know these changes.
Please review this. @MaikoTan @sandtechnology @Echoring


ui/raidboss/raidboss_config.ts
```diff
@@ -209,7 +209,7 @@ const kDetailKeys = {
       de: 'alarm',
       fr: 'alarme',
       ja: '警報',
-      cn: '警报文本',
+      cn: '致命级文本',
```
MaikoTan, line 212:
```suggestion
      cn: '致命级',
```

cause it is not only refer to "text", but also TTS.

ui/raidboss/raidboss_config.ts
```diff
@@ -1762,7 +1762,7 @@ const templateOptions: OptionsTemplate = {
         de: 'Alarm-Text Anzeigedauer (in Sekunden)',
         fr: 'Durée d\'affichage du texte d\'alarme (secondes)',
         ja: '警報テキスト表示時間の長さ (秒)',
-        cn: '警报级别文字显示持续时间 (秒)',
+        cn: '致命级别文字显示持续时间 (秒)',
```
MaikoTan, line 1765:
```suggestion
        cn: '致命级文字显示持续时间 (秒)',
```

be consistent with the top one.

also, "文字显示持续时间" might be too long for users. I think it makes sense to just "致命级文本显示时长"

ui/raidboss/raidboss_config.ts
```diff
@@ -1801,7 +1801,7 @@ const templateOptions: OptionsTemplate = {
         de: 'Alarm Lautstärke (0-1)',
         fr: 'Volume de l\'alarme (0-1)',
         ja: '警報音声の音量 (0-1)',
-        cn: '警报级别提示音的音量 (0-1)',
+        cn: '致命级别提示音的音量 (0-1)',
```
MaikoTan, line 1804:
```suggestion
        cn: '致命级提示音量 (0-1)',
```

may be this make sense.

MaikoTan:
So after discussing with @ShadyWhite , we decided to change the importance level of alerts to be "致命", "严重" and "信息", and describe it with the suffix "级".
What do you think, @Echoring ?

Echoring:
I think we'd better add a subject to these level, like "致命级警报"、“严重级警报”、“信息级警报”？
By the way, I think those level can be "致命-重要-一般" to make more sense？Or use "提示" instead of "警报"？

MaikoTan:
I'd prefer "提示" more than "警报", also agree "一般" instead of "信息".
But "严重" vs "重要", I'd prefer the previous one I think ?

Echoring:
emmm, seems "严重" is actually better and consistence. I agree.

Echoring (approved):

MaikoTan (approved):

sandtechnology (approved):

ShadyWhite:
It is good to merge now.

MaikoTan (approved):

MaikoTan:
Good to merge, thank you @ShadyWhite !! (o゜▽゜)o❤️

quisquous:
> I hope this change can be mentioned in the release note, so that Chinese players can know these changes.

I've started leaving all changes in the release notes and not editing out translations.

Issue #4294, opened 2022-04-18
ReadyCheck alert bug


yimo0908:
## Description

The `D.Va_-_Game_on.webm` will be played as I say "发起了准备确认" in the echo channel.

## Additional information

This will happen when say "发起了准备确认" in other channels, such as party fc etc. .

PR #4298, opened 2022-04-19
raidboss: prevent chat trigger ready check

MaikoTan:
fixes ReadyCheck alert bug #4294
thank yimo for finding the such message channel code

Co-authored-by: yimo <53830825+yimo0908@users.noreply.github.com>


quisquous:
Thanks for the fix!

PR #4302, opened 2022-04-19
raidboss: Fix Endsinger Ex 5head collector logic

valarnin:
5head mechanic is the 5 IDs in a row 19/20 times. It's the 1/20 that gets ya.

### DIFF
--- a/ui/raidboss/data/06-ew/trial/endsinger-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/endsinger-ex.ts
@@ -296,32 +296,25 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
-      id: 'EndsingerEx 5Head Collector',
+      id: 'EndsingerEx 5Head Initial Direction',
       type: 'GainsEffect',
       netRegex: NetRegexes.gainsEffect({ effectId: '891', capture: true }),
       condition: (data) => data.headPhase === 5,
       delaySeconds: 0.5,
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/trial/endsinger-ex.ts
```diff
@@ -364,6 +357,30 @@ const triggerSet: TriggerSet<Data> = {
       netRegexJa: NetRegexes.startsUsing({ id: ['6FFC', '7006', '7009', '700A'], source: '終焉を謳うもの', capture: true }),
       netRegexCn: NetRegexes.startsUsing({ id: ['6FFC', '7006', '7009', '700A'], source: '讴歌终结之物', capture: true }),
       condition: (data) => data.headPhase === 5,
+      // Do not need delaySeconds here, heads have been spawned for 5+ seconds
```
quisquous, line 360: Thanks for the comment!

quisquous:
Thanks for the fix! Foof...that makes me nervous about other id-related things.  Did that id get added and used as something else or just entirely skipped?

valarnin:
> Thanks for the fix! Foof...that makes me nervous about other id-related things. Did that id get added and used as something else or just entirely skipped?

Honestly I have no clue. Might have been an invisible actor or something, but the missing ID wasn't in the network log file at all, so if it was actually used it wasn't in any log lines that are normally generated.

PR #4307, opened 2022-04-20
raidemulator: Refactor state tracking for better performance

valarnin:
I mentioned changing the way state comparisons work in #3360. This is that change.

This is part one of breaking that old PR up into separate PRs instead.

The logic here is to only have to stringify the state once on creation, and then use that cached value until the combatant tracker state has been fully created.

Also shifted the logic added in #3359 to sit in the Combatant class since it makes more sense to live there alongside the new logic.

### DIFF
--- a/ui/raidboss/emulator/data/CombatantState.ts
+++ b/ui/raidboss/emulator/data/CombatantState.ts
@@ -11,6 +11,8 @@ export default class CombatantState {
   mp: number;
   maxMp: number;
 
+  // This is a temporary variable used during CombatantTracker initialization and is `delete`d
+  // after the combatant states have been determined to keep memory usage low.
   json?: string;
 
   constructor(
--- a/ui/raidboss/emulator/data/CombatantState.ts
+++ b/ui/raidboss/emulator/data/CombatantState.ts
@@ -11,6 +11,8 @@ export default class CombatantState {
   mp: number;
   maxMp: number;
 
+  json?: string;
+
   constructor(
     posX: number,
     posY: number,
@@ -31,6 +33,7 @@ export default class CombatantState {
     this.maxHp = maxHp;
     this.mp = mp;
     this.maxMp = maxMp;
+    this.json = JSON.stringify(this);
   }
 
   partialClone(props: Partial<CombatantState>): CombatantState {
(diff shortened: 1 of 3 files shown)


ui/raidboss/emulator/data/CombatantState.ts
```diff
@@ -31,6 +33,7 @@ export default class CombatantState {
     this.maxHp = maxHp;
     this.mp = mp;
     this.maxMp = maxMp;
+    this.json = JSON.stringify(this);
```
quisquous, line 36: Is `json` just being used for comparison? Would it make sense to have an `isEqual` function instead of maintaining an optional field?

Mostly I think it makes it more complex to try to think about when this field is valid or not.  Another alternative is to leave more comments about when this is valid or not?

valarnin: I tested the following:

1. JSON.stringify as it exists in `main` currently. Very slow since it has to stringify multiple times per line in the encounter, per combatant included in that line
2. Direct comparison of properties. Faster, but still a bit slow, and there's no way to enforce checking all properties without an `as` type assertion. Also still slower than comparing json strings
3. Encapsulating `this.json` and comparing via helper method. Fast, but with no way to remove the `json` property, uses up a lot of extra memory
4. Same as 3, but add a finalize to state as well, calling the function instead of just calling delete is also a significant overhead (that's 21k function calls on my test 9-minute encounter)

Maybe it makes sense to make `fast-deep-equal` a direct dependency and use that instead, but it's still not going to be as fast as comparing json strings, and the point of the original PR is to improve performance.

quisquous: That all makes sense.  I think mostly it's just looking at the "I am comparing the json fields of these two objects and carefully destroying it some of the time, how do I make sure I don't destroy it at the wrong time" worry about having to hold this field so carefully.

Do you think it'd be worth asserting that the old state has a json when you compare it? Or adding some comments at least about when it's valid and when it's not?

valarnin: I opted to do both. I think the comment I added makes sense?

quisquous: Thanks!

PR #4308, opened 2022-04-21
raidemulator: Update LineEvent indexing

valarnin:
This is part two of breaking #3360 up into smaller PRs. Update the indexing logic to be encounter-relative instead of file-relative.

I specifically avoided using `lines.forEach` or any other direct iteration method like that as even a 9-minute fight is 42k lines long which adds a noticeable overhead when having to recreate arrays. Calling `Object.entries` as used elsewhere is even slower.

### DIFF
--- a/ui/raidboss/emulator/data/NetworkLogConverter.worker.ts
+++ b/ui/raidboss/emulator/data/NetworkLogConverter.worker.ts
@@ -33,10 +33,6 @@ ctx.addEventListener('message', (msg) => {
         line.index = i;
       }
 
-      lines = lines.map((l, i) => {
-        l.index = i;
-        return l;
-      });
       const enc = new Encounter(day, zoneId, zoneName, lines);
       enc.initialize();
       if (enc.shouldPersistFight()) {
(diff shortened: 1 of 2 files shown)


valarnin:
Fixed extra lines, vscode running in WSL can have some pretty nasty file save delay apparently.

ui/raidboss/emulator/data/NetworkLogConverter.worker.ts
```diff
@@ -25,6 +25,14 @@ ctx.addEventListener('message', (msg) => {
   localLogHandler.on(
     'fight',
     (day: string, zoneId: string, zoneName: string, lines: LineEvent[]) => {
+      // Index the lines for this encounter
+      for (let i = 0; i < lines.length; ++i) {
+        const line = lines[i];
+        if (!line)
+          throw new UnreachableCode();
+        line.index = i;
```
quisquous, line 33: Maybe grep is failing me, but where does `index` get used?

valarnin: It's going to be used in an upcoming PR for the breaking up of 3360, and it's useful for debugging purposes in general (which is why it was included originally).

PR #4310, opened 2022-04-21
i18n: ultima KO translations

pyeonch:
Sorry for misc, I didn't notice that ultima-un is arrived. XD


quisquous:
If it's helpful, you can always run `npm run util -- findTranslations -l=ko -f=ui` to find any translations that are missing.

pyeonch:
That's what I looking for, I'll try it later. Thank you!

Issue #4287, opened 2022-04-16
International ffxiv version with Chinese parse language does not work


miller0929:
Hi,

I'm new to cactbot. I have used ACT and dps metre for a while. I installed cactbot yesterday via ACT (I have updated everything yesterday and today). I tried to add raidboss (combined alerts and timeline), as well as add alters only and timeline only separately. Alerts seem to be working (I'm getting text and voice alerts in actual raids after changing the settings). However, in either way, I cannot get the timeline to show. Also, I expected to see the timeline and alerts when testing in Summerford Farms after doing /countdown 5, but neither showed. 

Any idea on what might have gone wrong? Thanks.

quisquous:
It's most probably something chat log related.  Either you have "enable chat log for privacy", or you have a difference between the language the ffxiv plugin is in and the parser language for cactbot, or you have a dalamud plugin like chattyper that's messing with the chat log, or least likely that somehow you don't have those chat log messages enabled in any chat window.

For this sort of "setup" support, you might also want to ask in the ffxiv act discord.

miller0929:
It makes sense. My client is in English, but I have used another github tool to replace the in-game text with another language. However, changing cactbot's language in the config doesn't seem to help in my case. I'll also ask in the act discord. Thank you.

quisquous:
> I have used another github tool to replace the in-game text with another language

Yeah, that would most likely be the issue.  Which tool is this?

miller0929:
https://github.com/GpointChen/FFXIVChnTextPatch-GP
At the time that I started to install cactbot, I was on [6.1 4/15] simplified chinese with 4k font pack. I just updated to [6.1 4/16 ] simplified chinese with 4k font pack. I tested in Summerford Farms again (even after re-adding cactbot, as well as changing cactbot's languages), timeline still didn't show. 
One alternative could be: I leave the client's English as original, change Japanese to Chinese without the tool, display Japanese in the game (so i see Chinese), hoping the background works in English still, so that cactbot may read the data in the format it needs. It may work, but I'm not very keen to mess around. Cactbot's alerts is helping me a lot right now.

LTY0503:
I have the same problem. Just happen in this new update.

In english ACT setting, log:
Info: Version: intl
Info: LoadAddons: CactbotOverlay.dll: Initialized Cactbot.PluginLoader
Info: Found enmity memory for 6.1.

In chinese ACT setting, log:
Info: Version: cn
[...]

PR #4313, opened 2022-04-22
plugin: fixSomeChineseBug

tssailzz8:
fix https://github.com/quisquous/cactbot/issues/4287


tssailzz8:
I don't know how to solve this dependency

plugin/CactbotEventSource/CactbotEventSource.cs
```diff
@@ -268,14 +268,31 @@ public override void SaveConfig(IPluginConfig config)
       // Log this for now as there will likely be a lot of questions, re: user directories.
       if (Config.UserConfigFile != null)
         LogInfo(Strings.CactbotUserDirectory, Config.UserConfigFile);
-
+      var gameRegion = Machina.FFXIV.Headers.Opcodes.OpcodeManager.Instance.GameRegion;
```
quisquous, line 271: Is there a reason to check the language below? Should this always use only the game region instead of game region + language?

plugin/CactbotEventSource/CactbotEventSource.csproj
```diff
@@ -52,28 +52,28 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Advanced Combat Tracker">
-      <HintPath>..\ThirdParty\ACT\Advanced Combat Tracker.exe</HintPath>
+      <HintPath>E:\Advanced Combat Tracker\Advanced Combat Tracker.exe</HintPath>
```
quisquous, line 55: And as Akurosia says, you should revert all the changes to csproj files to fix the build errors.

plugin/CactbotEventSource/CactbotEventSource.cs
tssailzz8: This is to load different opcode

tssailzz8: it can also ues in ngld https://github.com/ngld/OverlayPlugin/blob/1a4232e189204c15f07a57cdeb48afa8b6c37990/OverlayPlugin.Core/EventSources/EnmityEventSource.cs#L73

quisquous: Sorry, I was trying to suggest something more like:

```cs
if (gameRegion==Machina.FFXIV.GameRegion.Chinese) {
  ffxiv_ = new FFXIVProcessCn(this);
} else if (gameRegion==Machina.FFXIV.GameRegion.Korean) {
  ffxiv_ = new FFXIVProcessKo(this);
} else {
  ffxiv_ = new FFXIVProcessIntl(this);
}
```

...without checking the language setting in the ffxiv act plugin.  If we have the game region, I don't think the language matters anymore.  Does that make sense?

plugin/CactbotOverlay/CactbotOverlay.csproj
```diff
@@ -50,17 +50,14 @@
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
-    <Reference Include="Advanced Combat Tracker, Version=3.4.5.266, Culture=neutral, PublicKeyToken=a946b61e93d97868, processorArchitecture=MSIL">
-      <SpecificVersion>False</SpecificVersion>
-      <HintPath>..\ThirdParty\ACT\Advanced Combat Tracker.exe</HintPath>
+    <Reference Include="Advanced Combat Tracker">
+      <HintPath>E:\Advanced Combat Tracker\Advanced Combat Tracker.exe</HintPath>
```
quisquous, line 54: Can you also revert these changes to this csproj file?

tssailzz8:
in the first it is rely on Machina.FFXIV.dll https://github.com/ravahn/machina
so i use reflection to not use it

plugin/CactbotOverlay/CactbotOverlay.csproj
quisquous: Sorry, I don't think you should change the csproj.  Can you revert your changes to this file?

tssailzz8:
i am not good at git,maybe  i pr it again

quisquous:
Sorry! You should be able to do: `git checkout main plugin/CactbotOverlay/CactbotOverlay.csproj`

PR #4316, opened 2022-04-22
raidboss: Fix raidboss not obeying `/e end`

### LINKED ISSUE #4189: Cacbot not obeying ACT "/e end".
## Description

ACT has a command called "/e end", that forces the encounter to end. Cacbot does not follow it. If someone starts a countdown and for whatever reason, the countdown ends and people dont start the fight (lets say, someone went afk), the only way to reset the timeline/calls is to wipe. Would be great if it also would reset when /e end would be issued.

## Additional information

image

valarnin:
This is a naive attempt to fix #4189. Basically, add `/e end` detection to WipeDetector and add OnWipe handling to PopupText.

This should also technically allow the removal of every `--Reset--` in every timeline, though I haven't tested this in-game at all yet (I can't actually build the dll file with my current setup).

I'd like to request review and testing by someone that can actually build the plugin dll, especially since I haven't worked with the C# source (and I'm pretty rusty on C# in general, I rarely have need of it).

### DIFF
--- a/plugin/CactbotEventSource/WipeDetector.cs
+++ b/plugin/CactbotEventSource/WipeDetector.cs
@@ -24,6 +24,9 @@ namespace Cactbot {
         if (log.IndexOf(" 00:0038:cactbot wipe", StringComparison.Ordinal) != -1) {
           WipeIt();
         }
+        if (log.IndexOf(" 00:0038:end", StringComparison.Ordinal) != -1) {
+          WipeIt();
+        }
         if (wipe_regex_.IsMatch(log)) {
           WipeIt();
         }
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -513,6 +513,9 @@ export class PopupText {
     addOverlayListener('onInCombatChangedEvent', (e) => {
       this.OnInCombatChange(e.detail.inGameCombat);
     });
+    addOverlayListener('onPartyWipe', () => {
+      this.OnInCombatChange(false);
+    });
     addOverlayListener('onLogEvent', (e) => {
       this.OnLog(e);
     });


quisquous:
https://github.com/quisquous/cactbot/actions/runs/2210017754 should have the built plugin if you wanted to test this yourself.

plugin/CactbotEventSource/WipeDetector.cs
```diff
@@ -24,10 +24,13 @@ class WipeDetector {
         if (log.IndexOf(" 00:0038:cactbot wipe", StringComparison.Ordinal) != -1) {
           WipeIt();
         }
+        if (log.IndexOf(" 00:0038:end", StringComparison.Ordinal) != -1) {
```
quisquous, line 27: This looks good to me!

(The wipe detector used to have a lot more logic (looking at hps and seeing if you come back to life without limit break or raise) but is much simpler these days.  In my mind this is all a bit vestigial and probably we should just do this in overlays that care about it rather than having a special event.  I think this is fine for now, this is all just musing about code that kind of doesn't make a ton of sense anymore.)

valarnin: So, finally got around to testing this, and it seems like WipeDetector just flat out doesn't work for any of the three wipe types. It's likely completely broken, but with no way to build locally I therefore can't debug locally to verify. I did test all three types and nothing triggered the `onPartyWipe` event.

Instead, it probably makes sense to just move this logic to `popup-text.ts:866` where it's already checking for `cactbot wipe`, as well as relevant updates in oopsy's `damage_tracker.ts` and pullcounter's `pullcounter.ts`, then remove the cactbot c# event entirely? If that's the case, I'll close this PR and open that as a different PR since it would have a different scope.

quisquous: Hmm!  Maybe it never got updated for log line format changes and nobody ever noticed.  I think removing the C# makes a lot of sense and moving it all to the relevant ts code.

valarnin:
Closing in favor of discussed alternate change.

PR #4322, opened 2022-04-23
raidboss: Sephirot EX updates

JLGarber:
Updated the timeline and added new triggers.

The timeline had some issues with sync timings after the first rotation. I fixed the initial Yesod cast and added windows elsewhere to mitigate any other possible jumps. Everything else timeline-related should be self-explanatory.

The positions for the Ein Sof puddles in phase 1 can technically be detected so that we can call an actual direction for moving the boss on Fiendish Rage and Ain + Ratzon. It's a good bit of work to do so because the puddles constantly cast as they expand, and it's just a lot of work to make that functional. Maybe in a later update, but it's good enough as it is here.

Some triggers have full translations because I copied them from the normal mode version.

Earth shakers do proximity falloff damage in this encounter. I'm not sure how best to communicate that to the player.

I'm not really happy about driving the Yesod bait triggers entirely off the timeline, but there's no good way to set the `pillarsActive` flag with ability regexes. I suppose we could count the Earth Shaker casts and use a delay to ensure that the Yesod immediately after doesn't count, but it seems like too much work. The solution as implemented *works*, but if something better is suggested I'll gladly switch it out.

### DIFF
--- a/ui/raidboss/data/03-hw/trial/sephirot-ex.ts
+++ b/ui/raidboss/data/03-hw/trial/sephirot-ex.ts
@@ -1,13 +1,322 @@
+import conditions from '../../../../../resources/conditions';
+import NetRegexes from '../../../../../resources/netregexes';
+import Outputs from '../../../../../resources/outputs';
+import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  mainTank?: string;
+  phase: number;
+  force?: string;
+  shakerTargets?: string[];
+  pillarActive: boolean;
+}
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.ContainmentBayS1T7Extreme,
   timelineFile: 'sephirot-ex.txt',
-  triggers: [],
+  initData: () => {
+    return {
+      phase: 1,
+      pillarActive: false,
+    };
+  },
+  timelineTriggers: [
+    {
+      id: 'SephirotEx Tiferet',
+      regex: /Tiferet/,
+      beforeSeconds: 4,
+      suppressSeconds: 2, // Timeline syncs can otherwise make this extra-noisy
+      response: Responses.aoe(),
+    },
+    {
+      id: 'SephirotEx Triple Trial',
+      regex: /Triple Trial/,
+      beforeSeconds: 4,
+      response: Responses.tankCleave(),
+    },
+    {
+      id: 'SephirotEx Ein Sof Rage',
+      regex: /Ein Sof \(4 puddles\)/,
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Move to safe quadrant',
+        },
+      },
+    },
+    {
+      id: 'SephirotEx Ein Sof Ratzon',
+      regex: /Ein Sof \(1 puddle\)/,
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Bait toward puddle',
+        },
+      },
+    },
+    {
+      id: 'SephirotEx Yesod Bait',
+      regex: /Yesod/,
+      beforeSeconds: 6,
+      alertText: (data, _matches, output) => {
+        if (data.pillarActive)
+          return output.withPillar!();
+        return output.noPillar!();
+      },
+      outputStrings: {
+        noPillar: {
+          en: 'Bait Yesod',
+        },
+        withPillar: {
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -1,13 +1,322 @@
...
+      phase: 1,
+      pillarActive: false,
+    };
+  },
+  timelineTriggers: [
+    {
+      id: 'SephirotEx Tiferet',
+      regex: /Tiferet/,
+      beforeSeconds: 4,
+      suppressSeconds: 2, // Timeline syncs can otherwise make this extra-noisy
```
quisquous, line 31: Yeah, I've been wondering if we just need some built-in logic to auto-suppress timeline triggers.

ui/raidboss/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -1,13 +1,322 @@
...
+        if (data.me === data.mainTank)
+          return output.noStack!();
+        return output.stack!();
+      },
+      outputStrings: {
+        noStack: {
+          en: 'Main tank--Do\'nt Stack!',
+        },
+        stack: {
+          en: 'Stack with your group',
```
quisquous, line 140: Group Stacks?

ui/raidboss/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -1,13 +1,322 @@
...
+      condition: (data) => data.phase === 1,
+      suppressSeconds: 10,
+      alertText: (data, _matches, output) => {
+        if (data.me === data.mainTank)
+          return output.noStack!();
+        return output.stack!();
+      },
+      outputStrings: {
+        noStack: {
+          en: 'Main tank--Do\'nt Stack!',
```
quisquous, line 137: What about just `Don\'t Stack` or `Avoid Stacks`?

ui/raidboss/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -1,13 +1,322 @@
...
+        return output.might!();
+      },
+      outputStrings: {
+        might: {
+          en: 'Pass tether; avoid puddles + greens',
+        },
+        magic: {
+          en: 'Grab a tether; get front',
+        },
+      },
```
quisquous, line 253: Suggesting a direction as the first thing to say here.

```suggestion
      outputStrings: {
        might: {
          en: 'Get Away, Avoid Puddles + Tethers',
        },
        magic: {
          en: 'Go Front; Get Tether',
        },
      },
```

ui/raidboss/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -1,13 +1,322 @@
...
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Bait toward puddle',
+        },
+      },
+    },
+    {
+      id: 'SephirotEx Yesod Bait',
+      regex: /Yesod/,
```
quisquous, line 62: Re: pillar presence, the only other thing I can think is to add a counter (and maybe a suppress too, for safety) here?  I'm not sure if that's better or worse, but it's the only alternative I could come up with.

quisquous:
> Earth shakers do proximity falloff damage in this encounter. I'm not sure how best to communicate that to the player.

I don't feel like I ever moved out for these.  If it won't kill somebody at full health, it's probably not worth saying anything about (but could leave a comment).  If it will kill somebody if they're not far enough out then maybe some sort of "earthshaker (max melee)" kind of text?

ui/raidboss/data/03-hw/trial/sephirot-ex.ts
JLGarber: Right, that was the other option I considered here, but I *really* don't like using counters unless there's literally no other way. There are 6 Yesods per cycle, and we want to special-case the third one. I *guess* we could start the counter after seeing two Yesods and then check `counter % 6 === 0`, but that only save lines of code, and it wouldn't be super clear to read.

JLGarber:
I ran this minimum ilevel with a group a couple weeks ago, and I know I consistently died if I stayed inside the hitbox. It's very possible the healers weren't being good about keeping the targets at 100%, but it's probably better to have everyone do their best to make it survivable.

ui/raidboss/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -1,13 +1,324 @@
...
+      condition: (data) => data.phase === 1,
+      suppressSeconds: 10,
+      alertText: (data, _matches, output) => {
+        if (data.me === data.mainTank)
+          return output.noStack!();
+        return output.stack!();
+      },
+      outputStrings: {
+        noStack: {
+          en: 'Do\'nt Stack!',
```
quisquous, line 137:
```suggestion
          en: 'Don\'t Stack!',
```

PR #4334, opened 2022-04-25
plugin: Remove WipeDetector

valarnin:
Also updates raidboss, pullcounter, oopsy, jobs to match all relevant wipe conditions.

I didn't update jobs to check `cactbot wipe` or `end` echo lines, since it doesn't seem to care regardless?


quisquous:
Thank you!

PR #4335, opened 2022-04-26
util: Argparse separation updates

JLGarber:
This is a cleaned-up (I hope) version of the mess in #3658.

~~It appears that when we create an instance of `LogUtilArgParse` after exporting + importing it, the various values for any unused args are set to `null` rather than their default values. This means we have to shift how we validate the various flags. (That is, if we `invoke split_log -f logfile.log -lf`, all arguments apart from `-f` and `-lf` will be `null` rather than their defined constant values.`-lf` will be -1.) It's very possible I'm missing a better way to handle all this, but as it stands this all seems to *work*.~~

~~`--force` is special-cased to be true if present, since if we don't define it as constant, it will instead have a value of `force[]` rather than `true`.~~

Disregard this because I'm bad at Argparse  and forgot that I changed over to using `constant` over `default`.

To expand on the reason for preferring `constant` over `defaultValue`, the Argparse documentation has it that [`constant` values are passed into the `args` object only if that flag is present on the command line](https://docs.python.org/3/library/argparse.html#const), while `default` is always passed through. Specifically:

> For positional arguments with [nargs](https://docs.python.org/3/library/argparse.html#nargs) equal to ? or *, the default value is used when no command-line argument was present:

Because we need to use `nargs = '?'`, I think in our context it makes more sense for flags not present to present `null` values, and to have their default values only if the flag is actually there. This way we don't have to handle magic numbers in the same step as checking for which flags are present.

### DIFF
--- a/util/logtools/split_log.js
+++ b/util/logtools/split_log.js
@@ -26,7 +26,7 @@ for (const opt of ['search_fights', 'search_zones', 'fight_regex', 'zone_regex']
     numExclusiveArgs++;
 }
 if (numExclusiveArgs !== 1)
-  printHelpAndExit('Error: Must specify exactly one of -lf, -lz, -fr\n');
+  printHelpAndExit('Error: Must specify exactly one of -lf, -lz, or -fr\n');
 if (args['fight_regex'] === -1)
   printHelpAndExit('Error: -fr must specify a regex\n');
 if (args['zone_regex'] === -1)
--- a/util/logtools/split_log.js
+++ b/util/logtools/split_log.js
@@ -5,7 +5,7 @@ import Splitter from './splitter';
 import { EncounterCollector } from './encounter_tools';
 import ZoneId from '../../resources/zone_id';
 import argparse from 'argparse';
-import TimelineParse from './arg_parser';
+import { TimelineParse } from './arg_parser';
 
 // TODO: add options for not splitting / not anonymizing.
 const timelineParse = new TimelineParse();
@@ -58,7 +58,7 @@ const generateFileName = (fightOrZone) => {
 
   let zoneName;
   if (fightOrZone.zoneName || fightOrZone.name) {
-    zoneName = fightOrZone.zoneName || fightOrZone.name;
+    zoneName = fightOrZone.zoneName ?? fightOrZone.name;
   } else {
     const idToZoneName = {};
     for (const zoneName in ZoneId)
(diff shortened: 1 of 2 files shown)


JLGarber:
Alright, I think this is probably close to where we need this to be.

Once we land this and start the re-implementation of the log tools in TypeScript, are we planning to move the new utilities to the `util` folder, or will they permanently live in a subfolder? I don't care much either way, but if we're going to move them we'll want to do it as soon as we can. (It's not impossible to browse the history of moved files, but it really feels like a pain.)

quisquous:
> I don't care much either way, but if we're going to move them we'll want to do it as soon as we can. (It's not impossible to browse the history of moved files, but it really feels like a pain.)

I don't feel like this is a huge issue to me, personally? `git log --follow` does a really good job, and it's pretty obvious when you hit a move and you need to add that arg.  I think the anonymizing / etc utilities could still live in their own folder? I was going to write an html front end for them so folks could drag and drop and split eventually.

util/logtools/split_log.js
```diff
@@ -314,25 +282,25 @@ const writeFile = (outputFile, startLine, endLine) => {
 
   const collector = await makeCollectorFromPrepass(logFileName);
 
-  if (args['search_fights'] === null) {
+  if (args['search_fights'] === -1) {
```
quisquous, line 285: For my own understanding, why did all the nulls change to -1?

JLGarber: Because when used in combination with `nargs = '?'`, a `defaultValue` will always be passed to the consuming script even if the flag was not actually present at the command line:

```
node --loader ts-node/esm split_log -f log.log -lf 1
// console.log(JSON.stringify(args));
{"file":"log.log","force":null,"search_fights":1,"search_zones":-1,"fight_regex":-1,"zone_regex":-1}

node --loader ts-node/esm split_log -f log.log -lf -lz
// console.log(JSON.stringify(args));
{"file":"log.log","force":null,"search_fights":null,"search_zones":null,"fight_regex":-1,"zone_regex":-1}
```

It Seems To Me (and I recognize that it's opinion) that it doesn't make a lot of sense to have nulls indicate that a flag may actually *be* present but without a value present at the command line.

Conversely, if we set `constant` instead, the default value will be passed to the consuming script if and only if the flag is present. A null value in this case means explicitly that the flag was not used, while a default value explicitly means "a magic number is present, handle it accordingly".

```
 node --loader ts-node/esm split_log -f log.log -lf
// console.log(JSON.stringify(args));
{"file":"log.log","timeline":null,"force":null,"search_fights":-1,"search_zones":null,"fight_regex":null,"zone_regex":null,"adjust":null}
```

quisquous: Thanks for the explanation! I get it now.

PR #4343, opened 2022-04-26
raidboss: DSR doorboss WIP triggers and timeline

valarnin:
Extremely WIP, only up to doorboss headmarkers so far, and a wall of placeholder triggers for all the various abilities and statuses in the fight that I'm pruning as I work through the log file I have available.

Only tested in raidemulator so far.

Timeline is also incomplete, not cleaned up, and doesn't have the command at the top yet.

Feel free to steal parts of what I have here to supplement what you're working on, otherwise I'll update this as I make progress on triggers.

PR #4352, opened 2022-04-27
raidboss: add some basic DSR triggers

quisquous:
Partially borrowed from #4343.

Tested in raid emulator.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -1,35 +1,202 @@
+import NetRegexes from '../../../../../resources/netregexes';
+import { UnreachableCode } from '../../../../../resources/not_reached';
+import Outputs from '../../../../../resources/outputs';
+import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { NetMatches } from '../../../../../types/net_matches';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export interface Data extends RaidbossData {
-  decOffset?: number;
-}
+type Phase = 'doorboss' | 'thordan';
 
-// TODO
-const firstDecimalMarker = 0;
+export interface Data extends RaidbossData {
+  phase: Phase;
+  decOffset?: number;
+  seenEmptyDimension?: boolean;
+}
 
 // Due to changes introduced in patch 5.2, overhead markers now have a random offset
 // added to their ID. This offset currently appears to be set per instance, so
 // we can determine what it is from the first overhead marker we see.
-const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker']) => {
+const headmarkers = {
+  // vfx/lockon/eff/lockon6_t0t.avfx
+  'hyperdimensionalSlash': '00EA',
+  // vfx/lockon/eff/r1fz_firechain_01x.avfx through 04x
+  'firechainCircle': '0119',
+  'firechainTriangle': '011A',
+  'firechainSquare': '011B',
+  'firechainX': '011C',
+  // vfx/lockon/eff/r1fz_skywl_s9x.avfx
+  'skywardLeap': '014A',
+} as const;
+
+const firstMarker: { [phase in Phase]: string } = {
+  'doorboss': headmarkers.hyperdimensionalSlash,
+  'thordan': headmarkers.skywardLeap,
+} as const;
+
+const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker'], firstDecimalMarker?: number) => {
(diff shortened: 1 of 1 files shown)


quisquous:
@valarnin would you mind taking a look over this if you have any time? I saw your automatic collection of headmarkers, but I couldn't get it to work reliably when I added more headmarker triggers in.  I left these as "headmarker on current player only" as a simple way to get *something* up.

valarnin:
Do you mean that they work in doorboss but not in Thordan? Or do you mean that the headmarker tracker fails when only looking at doorboss?

quisquous:
> Do you mean that they work in doorboss but not in Thordan? Or do you mean that the headmarker tracker fails when only looking at doorboss?

I am not sure.  It was probably a logic error on my end and I opted to just make things simpler (when I hadn't tested them in instance) rather than add more nuance.

I think it was that I added some clearing of headmarkers (because that needs to happen somewhere?) and I think I messed it up, so it was probably just me.  `*<_<*`

valarnin:
My original version (and your copy) both suffer an issue if the party elects to continue the fight from doorfight to Thordan, where the headmarker IDs won't line up unless they were all originally gathered that way in the first place.

Additionally, the construct for `firstMarker` currently assumes that every `phase` has a checkpoint, which isn't really accurate either.

valarnin:
Here's a proof of concept/example that fixes assumed checkpoints and should fix the continue issue. Also contains a global headmarker collector, as an example, with automatic cleanup at each phase transition.

```ts
  phase: Phase;
  decOffset?: number;
  seenEmptyDimension?: boolean;
  markerCollector: { [name: string]: string };
}

// Due to changes introduced in patch 5.2, overhead markers now have a random offset
// added to their ID. This offset currently appears to be set per instance, so
// we can determine what it is from the first overhead marker we see.
const headmarkers: { [phase in Phase]?: { [headmarker: string]: string } } = {
  'doorboss': {
    // vfx/lockon/eff/lockon6_t0t.avfx
    'hyperdimensionalSlash': '00EA',
    // vfx/lockon/eff/r1fz_firechain_01x.avfx through 04x
    'firechainCircle': '0119',
    'firechainTriangle': '011A',
    'firechainSquare': '011B',
    'firechainX': '011C',
  },
  'thordan': {
    // vfx/lockon/eff/r1fz_skywl_s9x.avfx
    'skywardLeap': '014A',
  },
} as const;

const firstMarker: { [phase in Phase]?: string } = {
  'doorboss': headmarkers.doorboss.hyperdimensionalSlash,
  'thordan': headmarkers.thordan.skywardLeap,
} as const;

const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker'], firstDecimalMarker?: number) => {
    // This must be set the first time this function is called in DSR Headmarker Tracker.
    if (firstDecimalMarker === undefined)
      throw new UnreachableCode();
    data.decOffset = parseInt(matches.id, 16) - firstDecimalMarker;
  }
};

  zoneId: ZoneId.DragonsongsRepriseUltimate,
  timelineFile: 'dragonsongs_reprise_ultimate.txt',
  initData: () => {
    return {
      phase: 'doorboss',
      markerCollector: {},
    };
  },
  triggers: [
    {
      id: 'DSR Phase Tracker',
      type: 'StartsUsing',
      // 62D4 = Holiest of Holy
      // 63C8 = Ascalon's Mercy Concealed
      netRegex: NetRegexes.startsUsing({ id: ['62D4', '63C8'], capture: true }),
      run: (data, matches) => {
        const currentPhase = data.phase;
        switch (matches.id) {
          case '62D4':
            data.phase = 'doorboss';
            break;
          case '63C8':
            data.phase = 'thordan';
            break;
        }
        if (data.phase !== currentPhase) {
          if (firstMarker[data.phase])
            data.decOffset = undefined;
          data.markerCollector = {};
        }
      },
    },
    {
      id: 'DSR Headmarker Tracker',
      type: 'HeadMarker',
      condition: (data) => data.decOffset === undefined,
      // Unconditionally set the first headmarker here so that future triggers are conditional.
      run: (data, matches) => {
        const firstHeadmarker: number = parseInt(firstMarker[data.phase], 16);
        const id = getHeadmarkerId(data, matches, firstHeadmarker);
        const headmarker = headmarkers[data.phase][id];
        data.markerCollector[matches.target] = headmarker ?? id;
      },
    },
```

quisquous:
I don't understand what problem you're trying to fix here, sorry. The marker ids are the correct (real) ids from the `Lockon.csv` table.  Even if that were not true, headmarker offsets are only changed when you zone in and so as long as you only set `decOffset` once (and set it correctly), then everything should work out.

I think I am still inclined to land something simpler and then can fix it up to be more complicated afterwards.

valarnin:
Yeah, it's fine to land this now and clean it up later, just wanted to make sure the potential issue was known.

quisquous:
> Yeah, it's fine to land this now and clean it up later, just wanted to make sure the potential issue was known.

I still don't understand what this issue is that you're seeing here? I think the triggers will still work as written if you continue from phase 1 into Thordan.

PR #4357, opened 2022-04-28
plugin: fix broken singleton mode in Machina.FFXIV

VergilGao:
Hello, if you want get the ```Machina.FFXIV.Headers.Opcodes.OpcodeManager.Instance```, just try opcode_manager_type.GetProperty("Instance").GetValue(***null***) instead of create a new instance of the singleton.


MaikoTan:
Thanks for the fixing !
You may change the title as "plugin: fix broken singleton mode in Machina.FFXIV" then you can pass the CI checking.

VergilGao:
> Thanks for the fixing ! You may change the title as "plugin: fix broken singleton mode in Machina.FFXIV" then you can pass the CI checking.

OK, i changed the title.

PR #4360, opened 2022-04-28
raidboss: add more DSR Thordan triggers

quisquous:


valarnin:
Re: KB dir for Ser Adelphel:
1. Store combatant ID the first time `Holiest of Holy` is used
2. `getCombatants` against that ID when the KB ability starts. The NPC has already moved before the start of the KB ability.

quisquous:
> Re: KB dir for Ser Adelphel:
> 
> 1. Store combatant ID the first time `Holiest of Holy` is used
> 2. `getCombatants` against that ID when the KB ability starts. The NPC has already moved before the start of the KB ability.

I'm pretty sure you can do it on nameToggle 01 as an even earlier time too.

PR #4362, opened 2022-04-28
raidboss: DSR Strength and Sanctity additional triggers

Legends0:
- Calls out location Ser Adelphel moves to for charge.
- Calls out safe spots for Spiral Thrusts.
- Calls out location Thordan moves to after Spiral Thrusts.
- Calls out clock direction to move first for Sanctity.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -264,7 +264,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegexCn: NetRegexes.ability({ id: '63D3', source: '骑神托尔丹', capture: false }),
       netRegexKo: NetRegexes.ability({ id: '63D3', source: '기사신 토르당', capture: false }),
       condition: (data) => data.phase === 'thordan',
-      delaySeconds: 4.7,
+      delaySeconds: 4.4,
       promise: async (data) => {
         // Collect Ser Vellguine (3636), Ser Paulecrain (3637), Ser Ignasse (3638) entities
         const vellguineLocaleNames: LocaleText = {
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -204,7 +204,7 @@ const triggerSet: TriggerSet<Data> = {
     },
     {
       id: 'DSR Adelphel Charge Start Direction',
-      // 63C4 Is Ser Adelphel's Holy Bladedance, casted once during the encounter
+      // 62D2 Is Ser Adelphel's Holy Bladedance, casted once during the encounter
       // Adelphel is in position about 29~30 seconds later
       type: 'Ability',
       netRegex: NetRegexes.ability({ id: '62D2', source: 'Ser Adelphel' }),
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -316,27 +316,20 @@ const triggerSet: TriggerSet<Data> = {
           spiralThrusts.push(matchedPositionToDir(combatant));
         }
 
-        if (spiralThrusts.length !== 3) {
-          console.error(`Spiral Thrusts: expected 3 combatants got ${spiralThrusts.length}`);
+        const [thrust0, thrust1, thrust2] = spiralThrusts;
+        if (thrust0 === undefined || thrust1 === undefined || thrust2 === undefined)
           return;
-        }
 
         // Array of dirNums
         const dirNums = [0, 1, 2, 3, 4, 5, 6, 7];
 
         // Remove where the knights are at and where they will go to
-        if (spiralThrusts[0] || spiralThrusts[0] === 0) {
-          delete dirNums[(spiralThrusts[0] + 4) % 8];
-          delete dirNums[spiralThrusts[0]];
-        }
-        if (spiralThrusts[1] || spiralThrusts[1] === 0) {
-          delete dirNums[(spiralThrusts[1] + 4) % 8];
-          delete dirNums[spiralThrusts[1]];
-        }
-        if (spiralThrusts[2] || spiralThrusts[2] === 0) {
-          delete dirNums[(spiralThrusts[2] + 4) % 8];
-          delete dirNums[spiralThrusts[2]];
-        }
+        delete dirNums[(thrust0 + 4) % 8];
+        delete dirNums[thrust0];
+        delete dirNums[(thrust1 + 4) % 8];
+        delete dirNums[thrust1];
+        delete dirNums[(thrust2 + 4) % 8];
+        delete dirNums[thrust2];
 
         // Remove null elements from the array to get remaining two dirNums
         dirNums.forEach((dirNum) => {
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -564,7 +564,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegexCn: NetRegexes.ability({ id: '63E1', source: '骑神托尔丹', capture: false }),
       netRegexKo: NetRegexes.ability({ id: '63E1', source: '기사신 토르당', capture: false }),
       condition: (data) => data.phase === 'thordan',
-      delaySeconds: 4.3,
+      delaySeconds: 4.4,
       promise: async (data) => {
         // The side the white knight, Ser Janlennoux (3635), is onrelative to
         // east/west can be used to determine where both knights will dash.
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -398,7 +398,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegexCn: NetRegexes.ability({ id: '63C4', source: '骑神托尔丹' }),
       netRegexKo: NetRegexes.ability({ id: '63C4', source: '기사신 토르당' }),
       condition: (data) => (data.phase === 'thordan' && (data.thordanJumpCounter = (data.thordanJumpCounter ?? 0) + 1) === 2),
-      delaySeconds: 2,
+      delaySeconds: 0.5,
       promise: async (data, matches) => {
         // Select King Thordan
         let thordanData = null;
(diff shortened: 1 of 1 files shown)


Legends0:
To call out sooner requires overlay.

quisquous:
https://www.twitch.tv/videos/1469757826?t=3h6m41s (this is 20:13)

Here's a bunch of other times on my stream where I had this patched in:
```
2022-04-28T20:07:38.8570000-07:00
2022-04-28T20:09:19.8100000-07:00
2022-04-28T20:11:46.8710000-07:00
2022-04-28T20:13:14.8290000-07:00
2022-04-28T20:28:10.3790000-07:00
2022-04-28T20:29:52.6250000-07:00
2022-04-28T20:31:20.4610000-07:00
2022-04-28T20:34:01.9800000-07:00
2022-04-28T20:36:51.9580000-07:00
2022-04-28T20:39:34.2270000-07:00
2022-04-28T20:41:10.3020000-07:00
2022-04-28T20:43:40.2300000-07:00
2022-04-28T20:50:07.3960000-07:00
2022-04-28T20:51:14.3280000-07:00
```

I think there's something awry here? Here's a network log: https://drive.google.com/file/d/1_z9sO9Bl8jXq8gSuYym8C6L235FJw2hE/view?usp=sharing

Legends0:
> https://www.twitch.tv/videos/1469757826?t=3h6m41s (this is 20:13)

The original from before I added parenthesis in aa4d13b was sometimes outputting the wrong numbers. I think after that fix it worked for the remainder of the night in my tests.

I have a refactored version that I'll test tomorrow that will use overlayplugin that can hopefully get earlier calls. I'll go ahead and work out its lint. From the logs, the startsUsing is roughly 6.6 seconds after Strength of the Ward is cast which is the spell that summons the knights into position. So just need to determine what amount of delaySeconds is needed for when we can first get the position.

Legends0:
I see what's happening. dir can be 0/1 and it's deleted by filter of boolean

Legends0:
I just realized there are a few strats for the sanctity direction, this is coded for the 2 move pattern, the 1 move pattern would be opposite?

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -229,6 +251,218 @@ const triggerSet: TriggerSet<Data> = {
...
+          delete dirNums[spiralThrusts[0]];
+        }
+        if (spiralThrusts[1] || spiralThrusts[1] === 0) {
+          delete dirNums[(spiralThrusts[1] + 4) % 8];
+          delete dirNums[spiralThrusts[1]];
+        }
+        if (spiralThrusts[2] || spiralThrusts[2] === 0) {
+          delete dirNums[(spiralThrusts[2] + 4) % 8];
+          delete dirNums[spiralThrusts[2]];
+        }
```
quisquous, line 339: What are these conditionals trying to do? If you're just trying to satisfy typescript, maybe this instead? (Please give this suggestion a whirl in raid emulator or something, as I'm just typing it, but I think it's just a little odd to say `if (spiralThrusts[0] || spiralThrusts[0] === 0) {`

```suggestion
        const [thrust0, thrust1, thrust2] = spiralThrusts;
        if (thrust0 === undefined || thrust1 === undefined || thrust2 === undefined)
          return;

        // Array of dirNums
        const dirNums = [0, 1, 2, 3, 4, 5, 6, 7];

        // Remove where the knights are at and where they will go to
        delete dirNums[(thrust0 + 4) % 8];
        delete dirNums[thrust0];
        delete dirNums[(thrust1 + 4) % 8];
        delete dirNums[thrust1];
        delete dirNums[(thrust2 + 4) % 8];
        delete dirNums[thrust2];
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -229,6 +251,218 @@ const triggerSet: TriggerSet<Data> = {
...
+        northeast: Outputs.northeast,
+        east: Outputs.east,
+        southeast: Outputs.southeast,
+        south: Outputs.south,
+        southwest: Outputs.southwest,
+        west: Outputs.west,
+        northwest: Outputs.northwest,
+        unknown: Outputs.unknown,
+        thordanLocation: {
+          en: 'Boss: ${dir}',
```
quisquous, line 462: Small suggestion to put direction first, and call Thordan Thordan.

```suggestion
          en: '${dir} Thordan',
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -257,6 +491,85 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: NetRegexes.ability({ id: '63E1', source: 'King Thordan', capture: false }),
+      netRegexDe: NetRegexes.ability({ id: '63E1', source: 'Thordan', capture: false }),
+      netRegexFr: NetRegexes.ability({ id: '63E1', source: 'Roi Thordan', capture: false }),
+      netRegexJa: NetRegexes.ability({ id: '63E1', source: '騎神トールダン', capture: false }),
+      netRegexCn: NetRegexes.ability({ id: '63E1', source: '骑神托尔丹', capture: false }),
+      netRegexKo: NetRegexes.ability({ id: '63E1', source: '기사신 토르당', capture: false }),
+      condition: (data) => data.phase === 'thordan',
+      delaySeconds: 4.3,
+      promise: async (data) => {
+        // Only need to know one of the knights locations, Ser Janlenoux (3635)
```
quisquous, line 506: Can you add some comments about how this works? I kind of expected that you'd have to look at heading here and not position.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -63,6 +67,24 @@ const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker'], firstDec
...
+  const y = combatant.PosY - 100;
+  const x = combatant.PosX - 100;
+
+  // During Thordan, knight dives start at the 8 cardinals + numerical
+  // slop on a radius=23 circle.
+  // N = (100, 77), E = (123, 100), S = (100, 123), W = (77, 100)
+  // NE = (116.26, 83.74), SE = (116.26, 116.26), SW = (83.74, 116.26), NW = (83.74, 83.74)
+  //
+  // Starting with northwest to favor sorting between north and south for
+  // Advanced Relativity party splits.
```
quisquous, line 82:
```suggestion
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: The side the white knight, Ser Janlennoux, is on (relative to east/west) can be used to determine where both knights will dash.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: Problem I was having is I was pulling from an index of the spiralThrust array, which could potentially be null and so I had added a simple if spiralThrusts[0] as a check for it. However an oversight in that the integer value of 0 evaluates to false and so I added an exception. At the time I couldn't see a cleaner fix for it, was making modifications between pulls haha.

quisquous: Haha, I know that feeling.  :)

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
quisquous: What if he lands on south or north? I've seen intercards and cards here (but I have no example of him on south or north, despite looking through a number of videos).

quisquous: Mostly I'm asking because we do calls based on the way the knight closest to the drk is facing.  e.g. if you're on the drk and looking mid the closest knight is looking left (either diagonal towards you or away from you) then you're going left/clockwise, and I think that would solve the north/south issue, unless there's some fixed "if it's north it's always left" sort of limited possibility?

Unless it's really the case that it can never be north or south??

Legends0: In all my pulls, I have not seen north or south be spawned, but that just means they spawned more towards the intercards than the other cases. I haven't exactly looked where the white knight can spawn.

Legends0: To add, it should call out unknown if it spawned north/south.

Legends0: some further testing, this only works for the 2 move strat. the 1 move strat will probably need heading?

Legends0: Ok, I looked at the fflogs data. The two gladiators spawn in one of two positions: West (95, 100) or East (105, 100). They only look in two directions as well. Ser Janlenoux only ever looks towards the north (northeast or northwest), and Ser Adelphel only looks towards the south (southwest or southeast). So I think if it calls out wrong I either have the wrong one of three Ser Janlenoux or the delaySeconds needs adjustment?

Legends0: I believe it's resolved now. Testing for if I need to increase delay more... but several pulls and it's been solid.
Could opt for changing cw/ccw to left/right and that way either 1-move or 2-move strats can use it without alterations (they just change direction they'd be looking), but that is more vague.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -257,6 +553,92 @@ const triggerSet: TriggerSet<Data> = {
...
+          0: output.unknown!(),
+          1: output.unknown!(), // north position
+          2: output.unknown!(),
+          3: output.counterclock!(),
+          4: output.unknown!(),
+          5: output.unknown!(), // south position
+          6: output.unknown!(),
+          7: output.clockwise!(),
+          8: output.unknown!(),
+        };
```
quisquous, line 628: I think if east and west are the only valid positions, then I think it's better to just compare x position against the center rather than get a direction (6 of which are invalid).

I am not worried about multiple strategies here, as in this case if it's backwards then folks can change the text in the cactbot config ui easily.  I think clockwise is still better than left though, because folks are always confused about "is it left about looking in or looking out" etc etc.

quisquous (commented):
As a request for the future, is it possible to break up these triggers into multiple PRs rather than adding more and more to the same PR?  (It doesn't have to be done here, this is fine.  But just saying future PRs could be smaller.)  That way if one trigger has questions or confusion (on my part) then the rest can get merged without being held up by changes to one.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -229,6 +320,211 @@ const triggerSet: TriggerSet<Data> = {
...
+      id: 'DSR Dragon\'s Rage',
+      // 63C4 Is Thordan's --middle-- action, thordan jumps again and becomes untargetable, shortly after the 2nd 6C34 action
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: '63C4', source: 'King Thordan' }),
+      netRegexDe: NetRegexes.ability({ id: '63C4', source: 'Thordan' }),
+      netRegexFr: NetRegexes.ability({ id: '63C4', source: 'Roi Thordan' }),
+      netRegexJa: NetRegexes.ability({ id: '63C4', source: '騎神トールダン' }),
+      netRegexCn: NetRegexes.ability({ id: '63C4', source: '骑神托尔丹' }),
+      netRegexKo: NetRegexes.ability({ id: '63C4', source: '기사신 토르당' }),
+      condition: (data) => (data.phase === 'thordan' && (data.thordanJumpCounter = (data.thordanJumpCounter ?? 0) + 1) === 2),
```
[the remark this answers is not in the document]
Legends0: I'll change to one of the Heavy Impact rings. Will require using similar overlayHandler to 4367, I need to test out the timing for delaySeconds.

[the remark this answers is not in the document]
Legends0: In testing out, I can't seem to get the data for Ser Hermenost any earlier than this trigger.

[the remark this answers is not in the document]
Legends0: I think the Ser Hermenost data I am getting may be even later than the King Thordan data as it seems to be available when the Dragon's Rage cast starts...?

To use the Heavy Impact with King Thordan names may require additional code to either get the boss' id in another trigger or use LocaleText. If there are multiple King Thordan comabatants, it shouldn't be too difficult to get the boss combatant. With that would just need to re-validate the delaySeconds required, but it's probably around 7.7s.

quisquous:
Tried this tonight, everything seemed great, thanks! I'll probably follow up and switch "West Adelphel" to say "East Knockback" or something so it's phrased as the positive rather than the negative thing to avoid.

PR #4367, opened 2022-05-01
raidboss: DSR Thordan location callout.

Ganom:
This is a pretty simple addition. It will call out Thordan's location before he spawns during `Strength of the Ward` by referencing the knight that spawns early in his opposite position. Ideally this trigger allows players to be able to move quicker with the knowledge of where Thordan will re-appear. The trigger its using doesn't actually matter, its just for proper timing on the overlay call.

The hard coded x/y checks are pretty nasty, I'm open to any recommendations on how to make that less ugly. I tested this on 10 of my logs tonight and it had correct predictions for each, but IIRC Overlay Handler calls can be unreliable on the emulator. Real world testing probably required.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -1,9 +1,11 @@
 import NetRegexes from '../../../../../resources/netregexes';
 import { UnreachableCode } from '../../../../../resources/not_reached';
 import Outputs from '../../../../../resources/outputs';
+import { callOverlayHandler } from '../../../../../resources/overlay_plugin_api';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
+import { PluginCombatantState } from '../../../../../types/event';
 import { NetMatches } from '../../../../../types/net_matches';
 import { TriggerSet } from '../../../../../types/trigger';
 
@@ -21,6 +23,7 @@ export interface Data extends RaidbossData {
   phase: Phase;
   decOffset?: number;
   seenEmptyDimension?: boolean;
+  knights?: PluginCombatantState[];
 }
 
 // Due to changes introduced in patch 5.2, overhead markers now have a random offset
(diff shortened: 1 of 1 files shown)


Ganom:
I believe this is similair to pr: https://github.com/quisquous/cactbot/pull/4362 specifically `DSR Dragon's Rage` trigger, not sure which is more useful. I'll leave that up to you all.

Ganom:
Turns out, moving the trigger to `63D6 - Heavy Impact 1` allows you to fine thordans spot up to 10 seconds before he comes down!

Legends0:
> I believe this is similair to pr: #4362 specifically `DSR Dragon's Rage` trigger, not sure which is more useful. I'll leave that up to you all.

The trigger in that referenced PR named `DSR Dragon's Rage` does call out Thordan's location, but it uses Thordan's real combatant, and so the soonest it can call out is when Thordan appears +- some delay in OverlayHandler updating combatant data.

Looking back at my VoDs, I do see that the Marauder knight, Ser Hermenost, spawns right after the Spiral Thrusts and then Thordan appears to spawn on the opposite side each time. I never realized that was also a way to determine where Thordan will be. I think switching to Ser Hermenost is a good idea since the player can also realistically get that info at that time as well. Also looks like it does not currently compete with any other callout, although I think given it is some time before Thordan appears, maybe it would need some durationSeconds added just to match up to a little after the boss appears.

I think there are multiple Ser Hermenost combatants, at least that's what fflogs is showing. The one thing that I don't know you can get in the raidemulator is how much delaySeconds is needed to get the position data once Ser Hermenost is in position, and in my experience getting that value has required trial and error.

Legends0:
I have code written to switch to using Ser Hermenost, would just need to determine how much delaySeconds it's going to need which I can get later.

Would you mind if I committed this change to my PR?

> The hard coded x/y checks are pretty nasty, I'm open to any recommendations on how to make that less ugly.
Regarding hard coded checks here: I am using some code I've borrowed from previous commits (both myself and others) that just does some trigonometry on X / Y coordinates which avoids the issue with getting minor fluctuations in floating point values across runs and also allows us to do modular arithmetic to get card/intercard values and then use a hashmap to quickly translate output strings.

Typically it's like this for the 8 cards/intercards
```
const matchedPositionToDir = (combatant: PluginCombatantState) => {
  // Positions are moved up 100 and right 100
  const y = combatant.PosY - 100;
  const x = combatant.PosX - 100;

  // During Thordan, knight dives start at the 8 cardinals + numerical
  // slop on a radius=23 circle.
  // N = (100, 77), E = (123, 100), S = (100, 123), W = (77, 100)
  // NE = (116.26, 83.74), SE = (116.26, 116.26), SW = (83.74, 116.26), NW = (83.74, 83.74)
  //
  // Starting with northwest to favor sorting between north and south for
  // Advanced Relativity party splits.
  // Map NW = 0, N = 1, ..., W = 7

  return (Math.round(5 - 4 * Math.atan2(x, y) / Math.PI) % 8);
};
```

There are similar code for headings too that get commonly used. I think it's also been proposed at one point to have some sort of standard function in cactbot for these calculations.

quisquous:
> Would you mind if I committed this change to my PR?

This is just a general suggestion, but I think it's better to split changes up so that PRs can land earlier and not keep getting added to incrementally preventing all the earlier changes from landing until those changes are also resolved.

Legends0:
The one that is giving off the position of Thordan isn't giving me the data until Thordan is visible... but maybe I am just getting the wrong combatant.

Legends0:
To add, I am just getting 1 combatant in overlayHandler, so maybe some difference with raidemulator makes it possible to get the data early? It's odd though why I can't get the position data earlier as theoretically a player could.

Ganom:
> To add, I am just getting 1 combatant in overlayHandler, so maybe some difference with raidemulator makes it possible to get the data early? It's odd though why I can't get the position data earlier as theoretically a player could.

Oof, that's not good. I wonder why that is, I'm about to start raiding so i'll try and get some data as well.

PR #4373, opened 2022-05-02
i18n: ko DSR translations

pyeonch:


quisquous:
Sorry, can you resolve these conflicts here?

PR #4378, opened 2022-05-02
raidemulator: Fix NaNs and extra states

valarnin:
Noticed some NaNs while debugging DSR triggers. When I was verifying those were fixed, also noticed extra states caused by the optimization PR.


valarnin:
Fixed this merge conflict with the github UI since I don't want to nuke what I have locally right now, hopefully it didn't break anything.

PR #4382, opened 2022-05-04
raidboss: Add support for optional netlog fields

### LINKED ISSUE #3713: Support StatusEffectExplicit regex having variable length number of fields
Regexes/NetRegexes parseHelper functions assume that all fields must exist and have no support for optional fields.  There are a number of line types that have optional fields.  One example is party, which there is no named regex for.  There's also statusEffectExplicit which has some number of triplets of data.

I think we need to add some new field to netlog_defs that says "after this index, fields may or may not exist", and have Regexes and NetRegexes just return undefined for those.  This probably will run into issues with `NetMatches` typing which assumes everything exists, but maybe it's possible to creatively write a regex such that any extra fields are empty strings.

This came up because I have a user trigger that looks for `data4`, which no longer exists as it's not guaranteed.  However, if I include `data4` then any line that only has 3 fields will not match the regex, which also seems bad.

We don't need data0 through data450 or whatever, but it'd be nice to at least support data0 through data5 and make it possible to add more in the future if needed.

valarnin:
Fixes #3713.

Also managed to clean up an `x as y` cast in both files as part of this.

Examples with this change:

```js
// NetRegexes.statusEffectExplicit({ data4: '00', capture: true });
/^(?<type>(?:38))\|(?<timestamp>(?:[^|]*))\|(?<targetId>(?:[^|]*))\|(?<target>(?:[^|]*))\|(?<jobLevelData>(?:[^|]*))\|(?<hp>(?:[^|]*))\|(?<maxHp>(?:[^|]*))\|(?<mp>(?:[^|]*))\|(?<maxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<x>(?:[^|]*))\|(?<y>(?:[^|]*))\|(?<z>(?:[^|]*))\|(?<heading>(?:[^|]*))\|(?<data0>(?:[^|]*))\|(?<data1>(?:[^|]*))\|(?<data2>(?:[^|]*))\|(?<data3>(?:[^|]*))\|(?<data4>(?:00))\|/i;
// NetRegexes.statusEffectExplicit({ data1: '00', capture: true });
/^(?<type>(?:38))\|(?<timestamp>(?:[^|]*))\|(?<targetId>(?:[^|]*))\|(?<target>(?:[^|]*))\|(?<jobLevelData>(?:[^|]*))\|(?<hp>(?:[^|]*))\|(?<maxHp>(?:[^|]*))\|(?<mp>(?:[^|]*))\|(?<maxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<x>(?:[^|]*))\|(?<y>(?:[^|]*))\|(?<z>(?:[^|]*))\|(?<heading>(?:[^|]*))\|(?<data0>(?:[^|]*))\|(?<data1>(?:00))\|/i;
// NetRegexes.statusEffectExplicit({ maxHp: '00', capture: true });
/^(?<type>(?:38))\|(?<timestamp>(?:[^|]*))\|(?<targetId>(?:[^|]*))\|(?<target>(?:[^|]*))\|(?<jobLevelData>(?:[^|]*))\|(?<hp>(?:[^|]*))\|(?<maxHp>(?:00))\|(?<mp>(?:[^|]*))\|(?<maxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<x>(?:[^|]*))\|(?<y>(?:[^|]*))\|(?<z>(?:[^|]*))\|(?<heading>(?:[^|]*))\|/i;
// NetRegexes.statusEffectExplicit({ data4: '00' });
/^(?<type>(?:38))\|(?<timestamp>(?:[^|]*))\|(?<targetId>(?:[^|]*))\|(?<target>(?:[^|]*))\|(?<jobLevelData>(?:[^|]*))\|(?<hp>(?:[^|]*))\|(?<maxHp>(?:[^|]*))\|(?<mp>(?:[^|]*))\|(?<maxMp>(?:[^|]*))\|(?:[^|]*\|){2}(?<x>(?:[^|]*))\|(?<y>(?:[^|]*))\|(?<z>(?:[^|]*))\|(?<heading>(?:[^|]*))\|(?<data0>(?:[^|]*))\|(?<data1>(?:[^|]*))\|(?<data2>(?:[^|]*))\|(?<data3>(?:[^|]*))\|(?<data4>(?:00))\|/i;
// NetRegexes.statusEffectExplicit({ data4: '00', capture: false });
/^(?:38)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*\|){2}(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:00)\|/i;
// NetRegexes.statusEffectExplicit({ maxHp: '00', capture: false });
/^(?:38)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:[^|]*)\|(?:00)\|/i;

// NetRegexes.statusEffectExplicit({ data4: '........', capture: true }).exec('38|2022-04-26T08:49:42.3420000-07:00|40003B54||005A5A00|51357|53616|10000|10000|0||||||8E02|F9|0|29310030|4472C95C|10001234|06867907b260d524')
Array(19) [ "38|2022-04-26T08:49:42.3420000-07:00|40003B54||005A5A00|51357|53616|10000|10000|0||||||8E02|F9|0|29310030|4472C95C|", "38", "2022-04-26T08:49:42.3420000-07:00", "40003B54", "", "005A5A00", "51357", "53616", "10000", "10000", ... ];
// NetRegexes.statusEffectExplicit({ data4: '........', capture: true }).exec('38|2022-04-26T08:49:42.3420000-07:00|40003B54||005A5A00|51357|53616|10000|10000|0||||||8E02|F9|0|06867907b260d524')
null
```

### DIFF
--- a/resources/netlog_defs.ts
+++ b/resources/netlog_defs.ts
@@ -711,7 +711,7 @@ const latestLogDefinitions = {
     },
     firstUnknownField: 20,
     canAnonymize: true,
-    optionalFields: [15, 16, 17, 18, 19, 20],
+    optionalFields: [18, 19, 20],
   },
   NetworkUpdateHP: {
     type: '39',
--- a/resources/netlog_defs.ts
+++ b/resources/netlog_defs.ts
@@ -28,8 +28,8 @@ export type LogDefinition = {
   playerIds?: { [fieldIdx: number]: number | null };
   // a list of fields that are ok to be blank (or have invalid ids)
   blankFields?: readonly number[];
-  // a list of fields that are ok to be completely missing including separator
-  optionalFields: readonly number[];
+  // this field and any field after will be treated as optional when creating capturing regexes
+  firstOptionalField: (number | undefined);
 };
 export type LogDefinitionMap = { [name: string]: LogDefinition };
 type LogDefinitionVersionMap = { [version: string]: LogDefinitionMap };
@@ -66,7 +66,7 @@ const latestLogDefinitions = {
         },
       },
     },
-    optionalFields: [],
+    firstOptionalField: undefined,
   },
   ChangeZone: {
     type: '01',
@@ -80,7 +80,7 @@ const latestLogDefinitions = {
     },
     lastInclude: true,
     canAnonymize: true,
-    optionalFields: [],
+    firstOptionalField: undefined,
   },
   ChangedPlayer: {
     type: '02',
@@ -97,7 +97,7 @@ const latestLogDefinitions = {
     },
     lastInclude: true,
     canAnonymize: true,
-    optionalFields: [],
+    firstOptionalField: undefined,
   },
   AddedCombatant: {
     type: '03',
@@ -131,7 +131,7 @@ const latestLogDefinitions = {
       6: null,
     },
     canAnonymize: true,
-    optionalFields: [],
+    firstOptionalField: undefined,
   },
   RemovedCombatant: {
     type: '04',
@@ -159,7 +159,7 @@ const latestLogDefinitions = {
       6: null,
     },
     canAnonymize: true,
-    optionalFields: [],
+    firstOptionalField: undefined,
   },
   PartyList: {
     type: '11',
@@ -220,32 +220,7 @@ const latestLogDefinitions = {
       25: null,
       26: null,
     },
-    optionalFields: [
-      3,
-      4,
-      5,
-      6,
-      7,
-      8,
-      9,
-      10,
-      11,
-      12,
-      13,
-      14,
-      15,
-      16,
-      17,
-      18,
-      19,
-      20,
-      21,
-      22,
-      23,
-      24,
-      25,
-      26,
-    ],
+    firstOptionalField: 3,
     canAnonymize: true,
     lastInclude: true,
   },
@@ -276,7 +251,7 @@ const latestLogDefinitions = {
     },
     canAnonymize: true,
     lastInclude: true,
-    optionalFields: [],
+    firstOptionalField: undefined,
   },
   StartsUsing: {
     type: '20',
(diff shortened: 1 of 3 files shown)


resources/netlog_defs.ts
```diff
@@ -26,8 +26,10 @@ export type LogDefinition = {
   };
   // map of indexes from a player id to the index of that player name
   playerIds?: { [fieldIdx: number]: number | null };
-  // a list of fields that are ok to not appear (or have invalid ids)
-  optionalFields?: readonly number[];
+  // a list of fields that are ok to be blank (or have invalid ids)
+  blankFields?: readonly number[];
+  // a list of fields that are ok to be completely missing including separator
+  optionalFields: readonly number[];
```
quisquous, line 32: Hmm, should this be "firstOptionalField"? Mostly in the sense that if one field is missing, then you can't have any fields after that one, right? (or if you do, then how do you differentiate [valid1, missing2, valid3] from [valid1, valid3]?

quisquous:
(Also, thank you! You are really shortening the bug backlog lately, and it is super appreciated!)

resources/netlog_defs.ts
valarnin: The original implementation instead wrapped the entire field including the separator in an optional `(?:field)?` which allowed this, but it wouldn't work ultimately because that would match the fields you're describing.

The bigger issue here is that the property has to be specified on every definition or else it's not accessible without a specific hardcoded definition name, so what should be put in that property if there's no optional fields? And it can't be an optional property, TS won't detect the union that way. It could maybe be specified as `firstOptionalField: undefined`, or `null`, but it has to be present.

PR #4386, opened 2022-05-05
util: enable vscode file nesting feature

MaikoTan:
[file nesting](https://code.visualstudio.com/updates/v1_64#_explorer-file-nesting) is a new feature of vscode that would merge related files into a logical "folder" that makes workspace apperance better.

Before:
image
After:
image

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -39,5 +39,16 @@
     "tsconfig.json": "tsconfig.*.json,tsconfig-for-webpack.json",
     ".gitignore": ".gitattributes",
     "README.md": "CONTRIBUTING.md,CODE_OF_CONDUCT.md,LICENSE"
+  },
+  "explorer.experimental.fileNesting.enabled": true,
+  "explorer.experimental.fileNesting.patterns": {
+    "*.ts": "${capture}.js",
+    "*.js": "${capture}.js.map, ${capture}.min.js, ${capture}.d.ts",
+    "*.jsx": "${capture}.js",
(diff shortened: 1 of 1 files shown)


JLGarber:
It is likely I am missing something here, but this is still an experimental feature, correct? I am on VSCode 1.66.2 and I had to enable it as 
```
  "explorer.experimental.fileNesting.enabled": true,
  "explorer.experimental.fileNesting.patterns": {
  // settings
}
```

After enabling it as `experimental` it worked perfectly. I personally have no issues with enabling an experimental feature that might be this useful, but we should definitely be aware if we decide to do it.

MaikoTan:
I found that I am using the insider preview version of VSCode that the file nesting feature can be enabled without the experimental tag. (╯°□°）╯︵ ┻━┻

Well, may be should this PR get merged later when this feature come into stable?

quisquous:
Could we enable both experimental and not experimental and delete experimental later? Are there errors if it's not supported, or would this silently just keep working as-is?

MaikoTan:
> Could we enable both experimental and not experimental and delete experimental later? Are there errors if it's not supported, or would this silently just keep working as-is? 

Will work as is.
Only when the user open the `settings.json` the VSCode would says "Whoa dude, I have no idea with these configs!".

MaikoTan:
https://code.visualstudio.com/updates/v1_67#_explorer-file-nesting
The file nesting feature now comes stable in v1.67 released just now. ლ(╹◡╹ლ)

PR #4387, opened 2022-05-05
build: Update stylelint and stylelint-config-standard

valarnin:
Also adds a stylelintfix run target

Updating stylelint fixes the following build warning:
```
npm WARN deprecated @stylelint/postcss-markdown@0.36.2: Use the original unforked package instead: postcss-markdown
```

However, updating stylelint necessitates updating stylelint-config-standard due to removed rules.

Updating stylelint-config-standard pulls in new rules to the default config, most importantly:
- kebab-case for classes and ids
- quotes around url parameters
- color specifications must use "modern" syntax

Various other files were updated to conform to the new rules. RaidEmulator changes are tested good, but jobs changes could use a double-check (AST and MNK).


quisquous:
> Various other files were updated to conform to the new rules. RaidEmulator changes are tested good, but jobs changes could use a double-check (AST and MNK).

Looks good to me.  Thanks for mentioning what needed a closer look too!

PR #4395, opened 2022-05-05
build: Update webpack and related packages

valarnin:
Updating webpack caused some tests to break for some reason, or maybe they were broken by the typescript update since I didn't test `npm run build` with that change. Regardless, updated chai and fixed the tests. They were breaking with a compile error:
```
TS2775: Assertions require every name in the call target to be declared with an explicit type annotation.
```

This PR fixes the following two warnings:

```
npm WARN deprecated querystring@0.2.0: The querystring API is considered Legacy. new code should use the URLSearchParams API instead.
npm WARN deprecated uuid@3.4.0: Please upgrade  to version 7 or higher.  Older versions may use Math.random() in certain circumstances, which is known to be problematic.  See https://v8.dev/blog/math-random for details.
```

Tested the following targets (as well as the resulting builds of raidboss and raidemulator), which I believe covers all potential changes:

```
npm run build
npm run build-gh-pages
npm run test
npm run start
```


valarnin:
The lint error is being caused by:

```
/cactbot/webpack/webpack.config.ts
  4:8  error  No default export found in imported module "copy-webpack-plugin"  import/default
```

Adding this rule fixes it, but I feel like we should restrict it to only node_modules imports somehow?
```js
  'import/default': 'off',
```

quisquous:
> Adding this rule fixes it, but I feel like we should restrict it to only node_modules imports somehow?

Yeah, I'm not sure how to restrict it so that it doesn't apply to imports that are coming from node_modules.

I would also be ok just disabling it for that line with a comment, and if this comes up in more instances we can reconsider how to handle it.

quisquous:
Running the coverage report now appears to OOM: https://github.com/quisquous/cactbot/runs/6315854299?check_suite_focus=true

valarnin:
> Running the coverage report now appears to OOM: https://github.com/quisquous/cactbot/runs/6315854299?check_suite_focus=true

I dug into this, and it's happening *after* the file is done executing. Throwing a console.log both at the end of the IIFE and at the end of the file itself, both log, then the entire script pauses for a significant time, and then it starts spewing warnings, and then it OOMs. I'll dig further today, it might be ts-node related.

PR #4396, opened 2022-05-06
build: Update various packages to resolve upstream vulns

valarnin:
There's only one remaining possibly-resolvable build warning at this point and I have no clue how to fix it, so someone else can tackle it.

```
[BABEL] Note: The code generator has deoptimised the styling of .../cactbot/ui/fisher/static-data.ts as it exceeds the max of 500KB.
```

Slight tweak to `CONTRIBUTING.md` to account for markdownlint rule change with updated package.

Also fixed the path of the `markdownlint` command for ThirdParty, I guess this was missed at some point?


MaikoTan:
> There's only one remaining possibly-resolvable build warning at this point and I have no clue how to fix it, so someone else can tackle it.
> 
> ```
> [BABEL] Note: The code generator has deoptimised the styling of .../cactbot/ui/fisher/static-data.ts as it exceeds the max of 500KB.
> ```

That is reported by babel, you can tweak the [compact option](https://babeljs.io/docs/en/options#compact) to a larger value, the default is 500KB.

valarnin:
The question is, should the value be increased, or is there a way to not have it try to compact that file given that it's a static data file anyways?

MaikoTan:
> The question is, should the value be increased, or is there a way to not have it try to compact that file given that it's a static data file anyways?

We have now the TerserPlugin in webpack enabled who will minify the files right? Maybe it would make sense to just turn off the compact option, what do you think?

PR #4398, opened 2022-05-06
util: Make encounter info property categories explicit

JLGarber:
When working with `EncounterCollectors` during imports, it's a real pain to deal with properties that may or may not be present on a `FightEncInfo` object vs a `ZoneEncInfo` object. Going forward, there should be no more ambiguity.

(This is a stepping-stone change that came up while I was doing some other work with `encounter_tools`, and I figured it would be best to consider it entirely separately.)

### DIFF
--- a/util/logtools/encounter_tools.ts
+++ b/util/logtools/encounter_tools.ts
@@ -10,7 +10,7 @@ import { commonReplacement, syncKeys } from '../../ui/raidboss/common_replacemen
 // This can happen on partial logs.
 
 type ZoneEncInfo = {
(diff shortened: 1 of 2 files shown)


JLGarber:
Huh. That's a new error. `test` came back clean locally, so I'm not sure whether this is my issue or just the build server being itself.

quisquous:
> Huh. That's a new error. `test` came back clean locally, so I'm not sure whether this is my issue or just the build server being itself.

Mmm it was "caused" by #4395 and fixed by #4401.

PR #4401, opened 2022-05-06
util: Fix coverage report error by not linting autogen file

valarnin:
This is more of a workaround, but the autogenerated file is legible enough without the linting being applied.

I traced it back to specifically an eslint issue by trial-and-error commenting stuff out.

Someone else should pull this PR down and verify the fix works properly as well, since I ended up updating a ton of packages related to eslint trying to get this to work without removing the linting.

`npm run process-triggers` ran fine. I'm not sure how to run the coinach.ts file so I can't check that case.

Minimal repro for the issue for future reference (save as `test.ts`, run via `npx ts-node test.ts`):

```ts
import eslint from 'eslint';

const outputFileName = 'coverage/coverage_report.ts';

const str = `export const test = true;\n`;

(async () => {
  const dprintLinter = new eslint.ESLint({ fix: true });

  // Get the default config for output file, remove dprint
  const config = JSON.parse(
    JSON.stringify(await dprintLinter.calculateConfigForFile(outputFileName)),
  ) as eslint.Linter.Config<eslint.Linter.RulesRecord>;
  config.plugins = config.plugins?.filter((p) => p !== 'dprint');
  delete config?.rules?.['dprint/dprint'];

  const linter = new eslint.ESLint({ fix: true, useEslintrc: false, baseConfig: config });
  await linter.lintText(str, { filePath: outputFileName });
})().catch((e) => {
  console.error(e);
  process.exit(1);
});
```


quisquous:
Ahhhhhhhh, of course.  Thank you!

dprint+eslint has some pathological (n^2??) issues when formatting large arrays.  This is why there's a format without dprint first and then a format with dprint when it's closer to what dprint wants.  I don't know why it suddenly started falling over here, but this feels like something that was on the threshold of OOM and your changes pushed it over the edge.

> npm run process-triggers ran fine. I'm not sure how to run the coinach.ts file so I can't check that case.

`ts-node util/gen_effect_id.ts` and `ts-node util/gen_pet_names.ts` both ran fine for me, so I think those arrays are smaller than the coverage report.

valarnin:
The example repro `test.ts` file is linting a single line of text so I don't think it's a size issue. It's probably some eslint plugin that needs updated but doesn't have an update available yet, but I couldn't find any way of getting more info out of tsc/node/etc to track down the root cause.

quisquous:
> The example repro `test.ts` file is linting a single line of text so I don't think it's a size issue. It's probably some eslint plugin that needs updated but doesn't have an update available yet, but I couldn't find any way of getting more info out of tsc/node/etc to track down the root cause.

Oh quite right, sorry for missing that.  That's wild.

PR #4406, opened 2022-05-06
util: Move utility functions to Encounter Tools

JLGarber:
This is another stepping-stone for continued changes to our log utilities. I'm not really attached to having these functions live in `encounter_tools`, I mostly placed them here to get them *out* of `split_log`. I'm not really attached to much here in general, really. In large part the changes made here are "satisfy the compiler but otherwise maintain most of the logic".

### DIFF
--- a/util/logtools/encounter_tools.ts
+++ b/util/logtools/encounter_tools.ts
@@ -241,7 +241,7 @@ export class EncounterFinder {
   }
 }
 
-export class EncounterCollector extends EncounterFinder {
+class EncounterCollector extends EncounterFinder {
   zones: Array<ZoneEncInfo> = [];
   fights: Array<FightEncInfo> = [];
   lastSeal?: string;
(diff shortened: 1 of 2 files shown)


util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
     this.lastSeal = undefined;
   }
 }
+
+class TLUtilFunctions {
+  timeFromDate(date: Date | undefined): string {
+    if (date)
+      return date.toISOString().slice(11, 19);
+    return 'Unknown_Time';
+  }
```
JLGarber, line 295: Here and for most of the other string-slice-and-dice functions for filenames, I figured that since we have type information available on our `EncounterInfo` types, we might as well handle potentially undefined properties within the utility functions themselves rather than attempting to massage the data before passing it to the functions. That is, I feel like it makes more sense to do this:

```typescript
const obj: SomeEncounterInfo = { SomeData };
const date = TLUtilFuncs.dayFromDate(obj.startTime); // We know that startTime is either Date or undefined already, so just use it
```

As opposed to 
```typescript
const obj: SomeEncounterInfo = { SomeData };
let date; // We have no idea whether startTime is defined, so we have to check for it explicitly, oh no.
if (obj.startTime)
    date = TLUtilFuncs.dayFromDate(obj.startTime);
else
    date = 'Unknown_Date';
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+      if (row[dateIdx] !== lastDate) {
+        lastDate = row[dateIdx];
+        console.log(lastDate);
+      }
+      const row0 = this.leftExtendStr(row[0], lengths[0]);
+      const row2 = this.leftExtendStr(row[2], lengths[2]);
+      const row3 = this.leftExtendStr(row[3], lengths[3]);
+      const row4 = this.rightExtendStr(row[4], lengths[4]);
+
+      console.log(`  ${row0})   ${row2}   ${row3}  ${row4}`);
```
JLGarber, line 410: Here and for `printCollectedFights`, I'm definitely not proposing that we leave this forever, by any means, but it does *work*, and it's not *that* slow. (Also I don't want to actually have to think about the logic right now. >_>)

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+      return '';
+    if (!length)
+      return str;
+    while (str.length < length)
+      str += ' ';
+    return str;
+  }
+
+  generateFileName(fightOrZone: FightEncInfo | ZoneEncInfo): string {
+    const zoneName = fightOrZone.zoneName || 'Unknown_Zone';
```
JLGarber, line 344: Given that we always have the `zoneName` available on our encounter objects, it Seems To Me that it makes more sense to just use that, rather than deal with the look-it-up process we had in place previously. I think I mentioned to you elsewhere that I was having difficulty getting types to line up when looking up zone names by zoneId. I did get it to work, but this approach is just better.

quisquous:
```suggestion
    const zoneName = fightOrZone.zoneName ?? 'Unknown_Zone';
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
     this.lastSeal = undefined;
   }
 }
+
+class TLUtilFunctions {
+  timeFromDate(date: Date | undefined): string {
```
quisquous, line 291:
```suggestion
  timeFromDate(date?: Date): string {
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+    if (date)
+      return date.toISOString().slice(11, 19);
+    return 'Unknown_Time';
+  }
+
+  dayFromDate(date: Date | undefined): string {
```
quisquous, line 297:
```suggestion
  dayFromDate(date?: Date): string {
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+    if (date)
+      return date.toISOString().slice(0, 10);
+    return 'Unknown_Date';
+  }
+
+  durationFromDates(start: Date | undefined, end: Date | undefined): string {
```
quisquous, line 303:
```suggestion
  durationFromDates(start?: Date, end?: Date): string {
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+    if (!(start && end))
```
quisquous, line 304:
```suggestion
    if (start === undefined || end === undefined)
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+  toProperCase(str: string): string {
+    return str.split(' ').map((str) => {
+      const cap = str[0]?.toUpperCase();
+      const lower = str.slice(1);
+      if (cap)
+        return `${cap} ${lower}`;
+    }).join(' ');
+  }
+
+  leftExtendStr(str: string | undefined, length: number | undefined): string {
```
quisquous, line 323:
```suggestion
  leftExtendStr(str?: string, length?: number): string {
```

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+    if (!str)
```
quisquous, line 324:
```suggestion
    if (str === undefined)
```

Since empty string is false...

util/logtools/encounter_tools.ts
```diff
@@ -286,3 +286,191 @@ export class EncounterCollector extends EncounterFinder {
...
+      return '';
+    if (!length)
```
quisquous, line 326:
```suggestion
    if (length === undefined)
```

Since 0 is false...

quisquous (commented):
Here's some initial comments and I didn't leave a comment for all of these, but I think in general:
* prefer "x?: type" over "x: type | undefined"
* check explicitly against undefined when you mean that to avoid accidentally hitting those cases on valid-but-falsy values

util/logtools/encounter_tools.ts
JLGarber: Ack. Yes. I should have remembered that after the incidents from exactly that with TEA!

JLGarber:
I'm not attached to using the built-in `pad` functions, but it finally clicked in my head "ohright, there's a builtin for that", and so I figured I might as well just toss it in quickly.

util/logtools/split_log.js
```diff
@@ -2,13 +2,14 @@ import fs from 'fs';
...
 import Splitter from './splitter';
-import { EncounterCollector } from './encounter_tools';
+import { EncounterCollector, TLUtilFunctions } from './encounter_tools';
 import ZoneId from '../../resources/zone_id';
 import argparse from 'argparse';
 import { LogUtilArgParse } from './arg_parser';
 
 // TODO: add options for not splitting / not anonymizing.
 const timelineParse = new LogUtilArgParse();
+const TFuncs = new TLUtilFunctions();
```
quisquous, line 12: Is there any state on `TLUtilFunctions`? Should all of its functions be static (e.g. regexes)?  And then you wouldn't need to make an object here, and could just refer to it off of the class itself.

PR #4409, opened 2022-05-07
raidboss: Add Japanese translations for DSU

vanderheim:
Added jp translations for some of the mechanics in DSU. Not all mechanics were translated, just the more important ones.


MaikoTan (approved):

quisquous:
It looks like you're probably missing a comma here.

quisquous:
TypeScript has some strict trigger types so throws a lot of errors, but it looks like it should be `ja` and not `jp` everywhere.

quisquous:
Thank you!  Sorry for the test hassle.

Issue #4389, opened 2022-05-05
[DSR] Spiral Thurst - Wrong Callouts


InitialDet:
It seems inconsistent, but sometimes when the safe spots are NE and SW, the callouts given are East and West (not safe lol). Most callouts were right but this pattern of wrong callouts happened more than once.

I'm not sure if it happens in other patterns.

quisquous:
Hmmm, I haven't seen this for any pattern, the spiral thrust callouts have been correct.  Do you have a network log where this happened?

InitialDet:
[Network_26501_20220504.zip](https://github.com/quisquous/cactbot/files/8628900/Network_26501_20220504.zip)

The first one I remember happened at 21:45pm (my time), the second time at 22:49pm

InitialDet:
This might be one of the times

`20|2022-05-04T21:45:22.8940000-03:00|40029F0A|Ser Ignasse|63D4|Spiral Thrust|40029F0A|Ser Ignasse|5.70|100.00|123.00|0.00|-3.14|1c7a7536b59e9466`

`20|2022-05-04T21:45:22.8950000-03:00|40029F0D|Ser Paulecrain|63D4|Spiral Thrust|40029F0D|Ser Paulecrain|5.70|77.00|100.00|0.00|1.57|d0f50ff8da614941`

`20|2022-05-04T21:45:22.8950000-03:00|40029F0E|Ser Vellguine|63D4|Spiral Thrust|40029F0E|Ser Vellguine|5.70|116.26|116.26|0.00|-2.36|b49e4c4bafa0e9a6`

levian05:
Had the same issue. Safe spots were N/S, but the callouts given were E/W(not safe)

valarnin:
InitialDet, your three line example shows the knights at south (X:100,Y:123), west (X:77,Y:100), and southeast(X:116.26,Y:116.26) respectively, making the safe spots northeast and southwest.

This is reflected when loading that encounter in raidemulator:

![image](https://user-images.githubusercontent.com/6119598/166932269-e692add7-89b0-4182-81ba-062b4fc3a495.png)

Two players got hit during that pull, standing at (X:106.38,Y:88.70) and (X:110.24,Y:86.43). They were hit by Ser Ignasse, who started south and moved north.

I checked the pull immediately after that one, and it has the knights at south, southwest, and west, making the safe spots northwest and southeast,

---

levian05, you have two different pulls in your example lines, as follows:
[...]

PR #4414, opened 2022-05-07
raidboss: delay Spiral Thrust callout by 0.5s

quisquous:
An attempt to address #4389 by adding some more time to account
for lag.


Legends0:
I did try this out, it's about the same as when people in pf would be calling it out. Still plenty of time to get in position with this additional time.

PR #4408, opened 2022-05-07
raidboss: DSRU Adjust timing of baited protean move call

valarnin:
As was discussed previously, adjust the timing of the move call for the baited protean to be closer to where it actually snapshots.

PR #4417, opened 2022-05-08
raidboss: DSR Skyblind move callout

Legends0:
I have it at 0.2 before as it felt better to account a bit for reaction time. The snapshot is on end of the effect, so there is plenty of time to move.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -11,7 +11,6 @@ import { NetMatches } from '../../../../../types/net_matches';
 import { LocaleText, TriggerSet } from '../../../../../types/trigger';
 
 // TODO: Ser Adelphel left/right movement after initial charge
-// TODO: "move" call after you take your Brightwing cleave?
 // TODO: Meteor "run" call?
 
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -355,6 +354,16 @@ const triggerSet: TriggerSet<Data> = {
...
       },
     },
+    {
+      id: 'DSR Skyblind',
+      // 631A Skyblind (2.2s cast) is a targetted ground aoe where A65 Skyblind
+      // effect expired on the player.
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'A65' }),
+      condition: Conditions.targetIsYou(),
+      delaySeconds: (_data, matches) => parseFloat(matches.duration) - 0.2,
```
quisquous, line 364: For similar reasons to https://github.com/quisquous/cactbot/pull/4408#issuecomment-1120243311, in general I don't want to be in the business of trying to account for reaction times.  I'd rather have the call happen when you're supposed to do the thing, even if that adds a little latency into the process.

You've got a lot of time to move out of the baited circle here too (vs the baited protean above), so I think no need for the reaction time adjustment here.

PR #4435, opened 2022-05-10
raidboss: DSR DFG Place, Stack, and Soak Callouts

Legends0:
This is a work in progress. I have tested the stack/soak callouts that they at least seem to flow without too much interruption with other TTS.

What this PR adds:
- Adds Dive position callout for each player based on their debuff
- Adds Move callout for player after their bait snapshots
- Adds Stack 1 callout for num2s and num3s
- Adds Stack 2 callout for num1 and num2s; Note that this has different timings based on dive 2/bait 2
- Adds Tower position callout for each player based on their debuff for Tower 1, and dive positioning for Towers 2 and 3
- Adjusts Darkdragon Dive Single Tower for second bait callout to be 'Bait => Stack with num2' and dives for num3s
- Adjusts DFG Dir Collect to include all debuffs for purposes of dive/stack/tower prediction

Note: Dive and Tower callouts are not paired with Bait/Stack as to not have the player move out too early.

Essentially the timeline for DFG:
03:28.246: cactbot calls out number **NEW** with Stack North for 2s and 3s
03:33.985: cactbot calls out up/down arrow or circle with arrows/without
03:35.407: cactbot calls out InThenOut or OutThenIn (11secs to get ready)
03:38:056: **NEW** num1s get Dive Positions, based on debuffs
03:42.968: Lash/Gnash 1 cast end
03:43.056: first towers are placed
03:43.056: **NEW** num1 that was middle gets Stack North
03:43.234: Eye of Tyrant stack hit
03:43.234: **NEW** num3s get callout for tower soaks based on debuffs (in sync with future dive position) with in/out reminder
03:46.661: In/Out and cactbot calls out next mechanic **NEW** with "... then Bait" for num3s
03:47.151: 1st towers appear
03:48.007: **NEW** num2s get Dive Positions, based on debuffs
03:49.642: 1st towers are soaked
03:49.642: cactbot calls out bait
03:49.775: Out/In
03:52.269: 1st bait ends **NEW** cactbot tells the targetted player to move
03:53.070: 2nd towers are placed
03:53.070: **NEW** num1s that were not middle get their respective callout for backleft/backright Tower 2, and num2s get Stack => Tower 3 (11.6 sec adv notice)
03:56.936: cactbot calls out InThenOut or OutThenIn (11secs to get ready)
03:57.161: 2nd towers appear
03:59.650: 2nd towers are soaked
03:59.650: cactbot calls out bait, **NEW** then stack (it's assumed that num1 players would get this, but not required)
03:59.947:  **NEW** num3s get Dive Position, based on debuffs (aligned with soak position); they would not have moved much since the 1st bait so this could be seen as a reminder from the bait callout ~10s ago
04:02.273: second bait ends **NEW** cactbot tells the targetted player to move
04:04.053: 3rd towers are placed
04:04.496: Lash/Gnash 2 cast end
04:04.764: Eye of Tyrant 2 stack hit
04:04.764: **NEW** num1 gets middle tower callout, num2 gets left/right towers, based on dive position, with in/out; a reminder
04:08.142: 3rd towers appear
04:08.186: In/Out and cactbot calls out next mechanic **NEW** Adds "... then Bait" for num1 (who was middle) and num2s
04:10.634: 3rd towers soaked
04:10.634: cactbot calls out bait (it's assumed that num1 that soaked middle and num2 players would get this, but not required) 
04:11.300: Out/In
04:13.255: 3rd bait ends **NEW** cactbot tells the targetted player to move
04:29.907: enumerations begin

The positioning is currently defaulted the following:
- North party stack
- High Jump South

There also only seems to be one order possible to resolve the mechanic due to timing of physical vulnerabilities and fire resistance down II vulnerabilities preventing strategies that do not follow the specific order based on 1, 2, 3. Positioning could possibly be different. The choice of North stack is based on current PF strategy: https://ff14.toolboxgaming.space/?id=580858020801561&preview=1 .


Legends0:
While it is possible to give the num1 player that is not doing the num2 towers a stack call, in order for cactbot to _know_ it's a solution it seems it has to check who took the num2 towers. Realistically that num1 player would have ~2.2s to get in place, which, if they weren't already there, will likely be failing the mechanic already...

This is probably only an issue for the case where all 1s are circles as we could use an educated guess that when 1s have spineshatter/elusive they are very likely going to want to do the same side tower.

Legends0:
Question would be whether or not to have a 'Stack/Soak?' callout to the Num1s for that second set of towers for completeness and maybe that be its own trigger to be disabled on/off.

Also possible to re-work these triggers around timing of tower placement and player's Physical Vulnerability with Fire Resistance Down II vulnerability in an attempt to break out of the required number order, but that seems much more complex and it doesn't look to me like there is another order.

Physical Vulnerabilities come in 7s and 2.96s timings. Fire Resistance Down II vulnerabilities come in 12s and 11.96s timings.

Legends0:
I aim to start testing the rest of this out tomorrow.

quisquous:
> Question would be whether or not to have a 'Stack/Intercard Tower?' callout to the Num1s for that second set of towers for completeness and maybe that be its own trigger to be disabled on/off.

Is it possible to figure out who the "side 1s" are and tell them to "soak tower 2" and only give a "do one of these things??" style response if we can't figure that out? I know that's assuming the strategy here, but maybe that's fine unless some other competing strategy comes up.

> Also possible to re-work these triggers around timing of tower placement and player's Physical Vulnerability with Fire Resistance Down II vulnerability in an attempt to break out of the required number order, but that seems much more complex and it doesn't look to me like there is another order.

Yeah, I talked to some folks and it looks like you can't take two towers in a row (the debuff just barely extends too long) or your own tower.  Therefore, I think the order that towers are taken are fairly fixed other than splitting the 1s in terms of who soaks 2s and who soaks a 3.

Legends0:
> Is it possible to figure out who the "side 1s" are and tell them to "soak tower 2" and only give a "do one of these things??" style response if we can't figure that out? I know that's assuming the strategy here, but maybe that's fine unless some other competing strategy comes up.
> 
I think you're right, it should possible to know who the side 1s are by looking at the targetX and targetY fields of the '670E', '670F', '6710' spells and thus give them the call to soak that same side during soak tower 2.

Legends0:
I think I've mapped out the tower soaks. Just need to implement the inverse for the num3 players as need to collect their Darkdragon Dive targetX, targetY and basically do the same math.

I've opted to try and estimate based on distances between the three players which is taking which position to try and support either north/south patterns as at least the spine/elusive are likely to be the furthest apart. It's too much to handle other possible positions and this itself is more complex than just comparing x,y inside a defined box as a measure to say which location they soaked. This would also bring up the issue where it's not necessarily a wipe if a tower is placed out in the middle of nowhere, but that could lead to some odd call outs when relying solely on previous position. Possible to put back the fire debuff tracking to eliminate that bit for the one player on tower 3 and for the cases that we know the debuffs.

quisquous:
> The choice of North stack

I've heard tell that the stack mechanics picks somebody in the northern half of the arena, which is why the stack has to be there.

Legends0:
One thing I am finding is that the sooner the 2s and 3s get out of the 1s positions, the easier it is for the 1s to choose a position (regardless of whether or not it's correct spine/elusive/circle). So I am opting to include Stack with the num callout. I don't know if the 1s necessarily need anything additional at that time since party is usually south at the time anyways, 1s wouldn't need to move yet.

Legends0:
Still needs some work, but the numbers it is getting appear to be at least working.

Some notes, being the num1 and getting the Stack => Tower 3 call is a bit jarring. Since the default TTS is fast, it you basically focus more on the Tower call and may forget the stack since it's so early.

Legends0:
There is another strat going around for this phase involving 3s and 1s pre-positioning:
https://ff14.toolboxgaming.space/?id=442165647202561&preview=1#1

Legends0:
I think after 7727935, it is getting close to ready. I noticed that I started getting the wrong values on the subsequent dives. I forgot to reset the collector since I am basically assuming the first three (or two and 1 null) are the ones that were just hit.

quisquous:
Re: your raidplan link (that maybe you deleted).  That's what we do, fwiw.  The 3s and 1s preposition in case it's all circles, and then 3s collapse north before the stack.  (Our 2s also preposition left and right north.)  We haven't had any issues with confusion, as they just don't stand on top of each other and it's pretty easy to glance around and see where your partners are.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+          console.error(`DFG Tower 2 and 3: missing number: ${JSON.stringify(data.diveFromGraceNum)}`);
+          return;
+        }
+
+        // With two Cartesian Coordinates, calculate distance
+        const distance = (a?: number[], b?: number[]) => {
+          if (
+            a === undefined || a[0] === undefined || a[1] === undefined ||
+            b === undefined || b[0] === undefined || b[1] === undefined
+          ) {
```
quisquous, line 871:
```suggestion
        const distance = (a: number[], b: number[]) => {
          if (
            a[0] === undefined || a[1] === undefined ||
            b[0] === undefined || b[1] === undefined
          ) {
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+          posB = [];
+        if (posC === undefined)
+          posC = [];
+        const posAX = posA[0] ?? 0;
+        const posBX = posB[0] ?? 0;
+        const posCX = posC[0] ?? 0;
+
+        // If undefined keys, map to non-player names
+        if (nameA === undefined)
+          nameA = '???';
```
quisquous, line 897: Can you make this be `output.unknown!()` rather than a literal string?

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+          // Get Distances of assumed "Triangle" pattern created by 3 players
+          const distanceAB = distance(posA, posB);
+          const distanceAC = distance(posA, posC);
+          const distanceBC = distance(posB, posC);
+          const distances = [distanceAB, distanceAC, distanceBC];
+
+          // Sort the distances, largest to smallest
+          const sorted = distances.sort((a, b) => b - a);
+
+          // Assuming the two players furthest apart are the left/right players
```
quisquous, line 914: Sorry for misunderstanding, but why look at distance vs using the x coordinate to figure out left vs right?

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -32,6 +32,12 @@ export interface Data extends RaidbossData {
   diveFromGraceNum: { [name: string]: number };
   // mapping of 1, 2, 3 to whether that group has seen an arrow.
   diveFromGraceHasArrow: { [num: number]: boolean };
+  diveFromGraceDir: { [name: string]: string };
+  // mapping of player name to cartesian (x, y)
+  diveFromGracePositions: { [name: string]: number[] };
+  diveFromGraceTowerCounter?: number;
+  eyeOfTheTyrantCounter?: number;
+  diveFromGracePreviousPosition: { [num: string]: string };
```
quisquous, line 40:
```suggestion
  diveFromGracePreviousPosition: { [num: string]: 'middle' | 'west' | 'east' };
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+          return output.text!();
+        }
+        if (data.eyeOfTheTyrantCounter === 1 && num === 1)
+          return output.baitThenStack!({ num: output.num2!() });
+        return output.text!();
+      },
       outputStrings: {
+        num2: Outputs.num2,
+        baitThenStack: {
+          en: 'Bait => Stack with #${num}',
```
quisquous, line 1128: I think `Bait => Stack` is plenty here.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+              return output.spineshatterTower2!();
+            if (data.diveFromGracePreviousPosition[data.me] === 'east')
+              return output.elusiveTower2!();
+          }
+        }
+        // Third Dive, on num3s
+        if (data.diveFromGraceTowerCounter === 3) {
+          // Num3s need to move out of tower 3
+          if (num === 3)
+            return output.move!();
```
quisquous, line 988: Everybody has to move out of their tower (and are usually doing it as a part of the out/in).  Personally, I think we could skip this call.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+        }
+      },
+      run: (data) => {
+        // Empty the collector
+        for (const player in data.diveFromGracePositions)
+          delete data.diveFromGracePositions[player];
+      },
```
quisquous, line 995:
```suggestion
      run: (data) => data.diveFromGracePositions = {},
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+      outputStrings: {
+        move: Outputs.moveAway,
+        stackNorth: {
+          en: 'Stack North',
+        },
+        spineshatterTower2: {
+          en: 'Backright Tower',
+        },
+        spineshatterTower3: {
```
quisquous, line 1004: Sorry for this tiny request, but can these strings be called "westTower3" etc (here and in other places)? I think that would make it easier for folks editing in the config ui, since the output string keys are what appears there.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+          en: 'Stack => Right Tower',
```
quisquous, line 1005: Can this be `Stack => West Tower` (I think right = west here?) Right/Left are confusing as folks might be facing a different way.  I think better for the built-in triggers to be more absolute and if folks want left/right they can edit the text.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+        },
+        elusiveTower2: {
+          en: 'Backleft Tower',
```
quisquous, line 1008: Similarly, can this say "west" or "east"? I'm confused why elusive is "left" as it's the right flank of the boss? Same comments elsewhere about changing left/right to east/west.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -818,14 +827,306 @@ const triggerSet: TriggerSet<Data> = {
...
+        }
+      },
+      run: (data) => data.eyeOfTheTyrantCounter = 1,
+      outputStrings: {
+        num3: Outputs.num3,
+        circleTower: {
+          en: 'South Tower',
+        },
+        circleTowers: {
+          en: '${num} Cardinal Towers',
```
quisquous, line 1055: Maybe just `Towers` or `Towers (all circles)` or `Get Tower`? I'm not sure the number part here adds a lot.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -928,6 +1235,80 @@ const triggerSet: TriggerSet<Data> = {
...
+          en: 'Intercard Dive',
+        },
+        diveSpineshatter: {
+          en: 'West Dive, Face Boss',
+        },
+        diveElusive: {
+          en: 'East Dive, Face Out',
+        },
+        diveSpineshatter2: {
+          en: 'Backright Dive, Face East',
```
quisquous, line 1305: So, I have seen different people do different strats here and face west.  Could this be "up arrow dive" and folks can rewrite it if they want to have something more specific?

I think that would then make "circle dive" a more natural call for the 2s and 1s where it's unclear.  I had to read "intercard dive" a few times to figure out what that meat.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -32,6 +32,12 @@ export interface Data extends RaidbossData {
   diveFromGraceNum: { [name: string]: number };
   // mapping of 1, 2, 3 to whether that group has seen an arrow.
   diveFromGraceHasArrow: { [num: number]: boolean };
+  diveFromGraceDir: { [name: string]: string };
```
quisquous, line 35: Do you think it would be worth making this `'circle' | 'up' | 'down'` rather than having to remember what `AC3` is elsewhere?

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -848,17 +1149,20 @@ const triggerSet: TriggerSet<Data> = {
...
-            return output.num3!();
+            return output.stackNorthNum!({ num: output.num3!() });
         }
       },
       outputStrings: {
         num1: Outputs.num1,
         num2: Outputs.num2,
         num3: Outputs.num3,
+        stackNorthNum: {
+          en: '${num}, Stack North',
```
quisquous, line 1164: I did mention the prepositioning, but I think this call is fine even with a prepositioning strat for 3s.  It's a reminder that they need to make their way north.

quisquous (commented):
Thanks for so much working through this and testing all of it.  I have a few small logic questions and a bunch of wording massaging.

I know generally we don't do strat-specific stuff, I think this is one where there really isn't a ton of options on who to have take which towers and I also think this is something that a lot of folks could use extra calls for.

The only strat variations I've seen are:
* looking west vs east
* prepositioning 3s vs having them stack immediately
* early prog safety strats where prepositioning took priority over arrows, where even if you were south with an up arrow, you'd go to the wall and face in and always place your tower on your prepositioned side

I think with the changes I suggested below (e.g. saying "up arrow dive" for this) would still cover this last case correctly.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: The distance is actually just to get the circle player's position for the 1's and 3's. At the time it seemed like a simple solution to handle orientations that were of any cardinal as I was not sure if the stack was required to be in a specific cardinal. Any cardinal such as ENW, NWS, ESW, NSE.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: These ones I have opted to include the north in it just to differentiate it between the locations of the other towers.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: I have changed it to Towers (all circles).

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
quisquous: Yeah, "northwest" is great, thanks.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: There was a mistake in one of the comparisons.

```
          const distanceAB = distance(posA, posB);
          const distanceAC = distance(posA, posC);
          const distanceBC = distance(posB, posC);
          const distances = [distanceAB, distanceAC, distanceBC];
```
These will give us the distances between players, which is important because by sorting distances and then looking back by comparing distanceAB, distanceAC, and distanceBC to the sorted values, we can determine which players were left/right and south player through elimination, assuming a triangle-like pattern.

quisquous: I guess I still don't understand why distances need to be used here.  It seems like you could be able to sort based on x coordinate, and then there would be a lot less logic here.

Legends0: I think pretty much the same thing can be done with sorting posX and taking the middle value to get who was middle, which will work for n/s.

Legends0:
Because of various ways groups may do pre-positioning for 3s/1s, I think the bait call out for 3s should be adjusted to also include where their position should be after the bait. Some groups may have the 3s do the baits where they pre-positioned, others have the 3s swap prior to the baits.

quisquous:
> Because of various ways groups may do pre-positioning for 3s/1s, I think the bait call out for 3s should be adjusted to also include where their position should be after the bait. Some groups may have the 3s do the baits where they pre-positioned, others have the 3s swap prior to the baits.

Would it just say "Bait (up arrow)" or "Bait (all circles)" sort of thing?

Legends0:
> Would it just say "Bait (up arrow)" or "Bait (all circles)" sort of thing?

I currently have it using the same language as the dive callout.

Legends0:
Another thing I just realized. Players often have issues of knowing when the bait has ended... so potentially could move the then x portion to a trigger that tracks start of 670A cast. Difficult part being I think this is a cast on self and so would need to collect who the players are that got targeted by 6711.

Legends0:
Seems pretty good for now. Is it worth adding comments for the relative (left/right) that I am using?

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -829,14 +847,301 @@ const triggerSet: TriggerSet<Data> = {
...
+            }
+            return output.baitThenCirclesDive!();
+          }
+        }
+        return output.text!();
+      },
+      run: (data) => data.waitingForGeirskogul = true,
       outputStrings: {
+        baitThenStack: {
+          en: 'Bait => Stack',
```
quisquous, line 1131: I got a "Bait => Stack" as the call for my 3 tower bait as a south 1 circle with arrow.  I'll see if I can repro in the emulator and see what's wrong.

quisquous (commented):
I think overall this is good.  There's a lot of infoText going on and I wonder if some of these things should be combined (e.g. "Stack => In" "Out => Bait"), but I think this is really close.  Let me take a closer look through video examples and see if I have any other feedback.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: That seems odd. Maybe a order of operations problem with data.eyeOfTheTyrantCounter = data.eyeOfTheTyrantCounter ?? 0 + 1 in the timeline trigger where it is always 1.

Legends0: typescript sim shows that to be the case, so I pushed parenthesis that resolves the trigger not incrementing properly.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -136,7 +136,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'DSR Eye of the Tyrant Counter',
       regex: /Eye of the Tyrant/,
       beforeSeconds: 1,
-      run: (data) => data.eyeOfTheTyrantCounter = data.eyeOfTheTyrantCounter ?? 0 + 1,
+      run: (data) => data.eyeOfTheTyrantCounter = (data.eyeOfTheTyrantCounter ?? 0) + 1,
```
quisquous, line 140: Oh yeah, that would definitely do it.

Legends0:
> I think overall this is good. There's a lot of infoText going on and I wonder if some of these things should be combined (e.g. "Stack => In" "Out => Bait"), but I think this is really close. Let me take a closer look through video examples and see if I have any other feedback.

Aside from these triggers, I do believe that the current in/out callouts need to be addressed. The second one is good, but the first is so early and with these other triggers, it's easy to forget the order.

Legends0:
I think maybe track the first in/out in a data variable, recall it for these calllouts:
04:04.764: NEW num1 gets middle tower callout, num2 gets left/right towers, based on dive position; a reminder
03:43.234: NEW num3s get callout for tower soaks based on debuffs (in sync with future dive position)

Something like In/Out + Dive/Soak ...

The other cases I don't recall being too much of a problem.

quisquous:
From the 19:02:52 pull in the log that I sent you, DRK pov:
image

I think the `Dive From Grace Tower 3` is incorrect?

Same log, AST pov:
image

I think `Stack => East Tower` prior to `Out, then in` prior to `Dive (all circles)` seems like a weird order.

Same log, SMN pov:
image

I am not sure what `DSR Dive From Grace Tower 1` is saying.  I think this should not say west or east here.  If anything, it should say `Up Arrow Tower (Out)`? I think also `Bait => Up Arrow Dive` is really early.  They have to first move out of the geirskogul and then wait like 10-15 seconds before anything happens.  Maybe it should just say `Bait` because there's so much time?

Legends0:
> 
> I think the `Dive From Grace Tower 3` is incorrect?
> 

It's likely calling for the num1s that were not middle since I left out the check on that in (num != 3)...

Legends0:
> . If anything, it should say `Up Arrow Tower (Out)`? I think also `Bait => Up Arrow Dive` is really early. They have to first move out of the geirskogul and then wait like 10-15 seconds before anything happens. Maybe it should just say `Bait` because there's so much time?

I have adjusted the first tower callout there to match up/down Arrow instead of east/west.

The early bait callout I think should be left as separate variables for easy modification in case player there is doing a swap before the bait, but maybe we just default the strings to match the output.text!(). I can rename the varaibles accordingly ie: bait3...

Legends0:
> 
> I think `Stack => East Tower` prior to `Out, then in` prior to `Dive (all circles)` seems like a weird order.
> 

I'll take a look at this next.

Legends0:
> > I think the `Dive From Grace Tower 3` is incorrect?
> 
> It's likely calling for the num1s that were not middle since I left out the check on that in (num != 3)...



> > I think the `Dive From Grace Tower 3` is incorrect?
> 
> It's likely calling for the num1s that were not middle since I left out the check on that in (num != 3)...

I may have also missed what is deemed incorrect as I did not look at position info from the log.

The In/Out portion to call could be questionable? At the time of the callout, it is correct in that my design was to call out where to pre-position before the tower appears and not get knocked before the soak snapshots. The goal is to say they should be standing in the tower, outside the hitbox.
The soak does not end until 0.7s before the followup. However, if they get the first in/out wrong, they are likely to miss the tower. There is a callout for the followup during the soak that will have them move while in the tower.

PR #4437, opened 2022-05-11
raidboss: Remove knockback call from zodiark algedon

amki:
You are only being knocked if you stand in the beam which the call aims to avoid.

For other mechanics "knockback" is used when there is an unavoidable knockback (that can or can't be suppressed by anti knockback abilities) but in this case the knockback doesn't happen if you don't stand in avoidable bad stuff, regardless it always confuses me.

Hence I propse removing it.


amki:
Hm, ok. I just thought it was accidental.

PR #4440, opened 2022-05-11
jobs: Fix AST/MNK sign/beastchakra cannot remove color

### LINKED ISSUE #4438: AST Job color are not working correctly
Hello,
since few versions, seal colors in the job UI are not working as expected.

After using many seals, color becomes light blue for all indicators (and do not change anymore).

It's better to see it live, I've made a video [https://www.youtube.com/watch?v=B-_Jag0DsoY](here). You'll see that after a certain use of seal, all indicators are light blue.

Echoring:
Due to the change of #4387 some upper case id in css changes to lower case, become inconsistence with codes in js.
Fixes #4438


quisquous:
Thanks for the fix!

PR #4443, opened 2022-05-12
raidboss/raidemulator: Fix issues with initData handling

### LINKED ISSUE #4425: raidemulator: initData is only called when loading an encounter
## Description

Using raidemulator, it seems the initData function specified by user-triggers is only ran when initially loading an encounter. When "seeking" within the enounter or starting from the beginning, initData is not called again.

## Additional information

If initData is only called once, then this means any triggers which rely on initData to "reset state" will fail until reloading the encounter or changing character. For example, let's say a user has a data object that exists to collect a list of players with a specific status, and then a trigger is used later in the encounter based on that list. That list needs to be reset every pull in order for the trigger to work as intended. When emulating the encounter, the list does not get reset when "rewinding", so the trigger will only work as expected the first time that trigger fires.

My current work around is to have an "initialization trigger" that performs the same actions as initData, but this feels redundant since it's only there for the raidemulator.

As an aside, this is my first time using the raidemulator, so if there is something I am missing or it's likely to be on my end, feel free to close this issue.

valarnin:
Fixes #4425.
While testing the fix I noticed an issue with initData being called multiple times instead. Turns out it wasn't ever being cleared which I'm sure lead to some memory leakage slowly over time, as the initData functions built up and populated the data object with more and more zones' worth of initial state.

### DIFF
--- a/ui/raidboss/emulator/overrides/RaidEmulatorPopupText.ts
+++ b/ui/raidboss/emulator/overrides/RaidEmulatorPopupText.ts
@@ -138,9 +138,11 @@ export default class RaidEmulatorPopupText extends StubbedPopupText {
     emulator.on('midSeek', async (line: LineEvent) => {
       await this.doUpdate(line.timestamp);
     });
-    emulator.on('preSeek', () => {
+    emulator.on('preSeek', (seekTimestamp: number) => {
       this.seeking = true;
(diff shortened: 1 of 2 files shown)


quisquous:
Good catch on that needing to be cleared, thanks!

PR #4444, opened 2022-05-12
util: Add typescript version of test_timeline.ts

valarnin:
Still very WIP. Should technically work, except Timeline requires HTMLElement which doesn't exist in node. I tried various options (`jsdom`, `jsdom-global`, `html-element`, `@happy-dom/global-registrator`) but none of them seemed to work.

I'm pretty out of ideas on how to resolve it.

If anyone wants to take over this, feel free to take what I have and make it work.


valarnin:
I've finally had success in getting jsdom-global to work, so I can proceed in working on this PR now 😃

quisquous:
You should also feel free to abstract Timeline from its us of HtmlElement too? Like, it could take in some interface with functions to call that do all the HtmlElement stuff, and then test_timeline doesn't need to know anything about it.  I'm not sure if that makes things better or worse, but please consider that as an option too.

The idea of reusing Timeline rather than reimplementing it in test_timeline is really good though.  If we can have the same feature set of the old tool while sharing the internals, that'd be great.

valarnin:
The fix to get it working is basically just:
1. `npm i --save-dev jsdom-global`
2. Add `-r jsdom-global/register` to the `ts-node` call in package.json for util

That seems like a better option in general, rather than trying to untangle the dependency tree so that no `.ts` file that references a DOM object is loaded.

valarnin:
Progress. Still a bit derpy though.
image

valarnin:
Much better.
Old:
image
New:
image
Here's an example before I added the window checking for syncs, showing what a drift fail and drift warning look like.
image
Output columns are `delta | line number | line contents`. Thoughts on if anything else is useful for output here?

quisquous:
Ooooh this is nice.

Tiny request is that I think of the `delta` as "how much you have to add to current time to get to this line", so +376 is a jump forward 376, vs your -376 planar prison "we got to this line, but the real time was 376 behind".  I realize this is six of one half dozen of the other, but it would help me personally when looking at this.

valarnin:
This is rebased off of #4476, so the changed files will look like a mess until that PR is merged and I can rebase off of `main` again.

Relevant changes are in commit https://github.com/quisquous/cactbot/pull/4444/commits/92c976cca45e9527b16f630c520adaeae035e59b

A quick summary:

- Added `jsdom-global` as a dev dependency
- Adjusted `npm run util` to pre-require the `jsdom-global/register` module so that DOM is available in node
- Fixed a bug in `timeline_parser.ts` which was causing syncs to have an ID one higher than their associated event, and which was causing the last user-added event to have the same ID as the first default timeline event
- Added import and register calls to `util/index.ts`
- The entirety of the `util/test_timeline.ts` file

Not included in this PR:
- `-lf`/`--search_fights` support
- FFLogs report support
- `-s`/`-e` parameters require a valid date timestamp instead of a shorthand timestamp, e.g. the parameter must be passable to `new Date()`, and the exact semantics around what lines are included may vary slightly from the python implementation due to this version checking the date timestamp rather than a string match against the date

quisquous:
> Re: delta, it's currently calculated as delta that the log line was off compared to the timeline sync. I guess I look at it as "you need to adjust the timeline entry by this amount to match what was in the log"?

Yeah, I think both are valid ways of looking at it.  I'm just being selfishly opinionated here because I've stared at make timeline for so long that it looks wrong to have it the other way.  If you feel strongly we could add an option or something.

valarnin:
> If you feel strongly we could add an option or something.

I don't feel strongly one way or the other, it's just an inversion of the delta as it exists to check against the sync window (as indicated by the change in the most recent commit).

Ultimately it's the same number with a different prefixed sign, it was just an extra step for the output display that I didn't really think was needed originally.

valarnin:
There are the following outstanding issues on this. I'm not sure if it's worth delaying merging since I'm not sure how much time I'll have to work on this for the next week or so, given that it's functional as-is.

1. There's no "x time since last sync" check/warning (see `test_timeline.py` line 363)
2. There's no notificcation for jumps (see `test_timeline.py` line 396)
3. There's no support for [`-lf`/`--search_fights`, `-r`/`--report`], as indicated previously
4. `-s`/`-e` are looking for a different format, but I guess this might just be counted as a breaking change instead?
5. It doesn't skip/suppress missed syncs before the first sync that was seen (see screenshot below)

image

quisquous:
> There are the following outstanding issues on this. I'm not sure if it's worth delaying merging since I'm not sure how much time I'll have to work on this for the next week or so, given that it's functional as-is.

Thanks! I gave this a quick spin, and other than it being painful to find and specify `-s` and `-e`, it seems good to go.  And given that the other script is still there, no reason not to merge this now.

> There's no "x time since last sync" check/warning (see test_timeline.py line 363)

I think this is fine.  We could probably drop this, imo.

> There's no notification for jumps (see test_timeline.py line 396)

I think that's also fine. 

On a similar thought, I've been thinking some about how to better integrate test_timeline with timeline practices.  Like, I wonder a little if it makes sense not to count events (both initially, but also mid-pull for phases) as missed if there's a jump.  I also wonder about if we should look for `/\?$/` in the text and consider that only "maybe missed" or somehow different than "missed" in a red color.

> There's no support for [-lf/--search_fights, -r/--report], as indicated previously

`-lf` is probably *the* highest priority missing feature here.  Specifying timestamps manually is really painful.

> -s/-e are looking for a different format, but I guess this might just be counted as a breaking change instead?

That's fine to have a breaking change here (but the prompt probably could use an example of what the correct format looks like).

> It doesn't skip/suppress missed syncs before the first sync that was seen (see screenshot below)

This is "nice to have" but not a big deal.

package.json
```diff
@@ -22,7 +22,7 @@
     "test": "mocha",
     "lint-staged": "lint-staged",
     "coverage-report": "ts-node util/gen_coverage_report.ts",
-    "util": "ts-node util/index.ts",
```
quisquous, line 25: My other feedback here is that it's kinda unfortunate that you have to run this through `npm run util` too and any other way of running it is a mistake.  Other scripts can be run by running the script directly and this is the only weird one.

Is there some way that you can do this jsdom-global dance inside the script itself?

valarnin:
> `-lf` is probably *the* highest priority missing feature here. Specifying timestamps manually is really painful.

I poked at this originally for a bit, but then saw that #4460 had some stuff related to this functionality so decided to wait until that PR is merged, then ran out of free time to potentially contribute for a bit.

> That's fine to have a breaking change here (but the prompt probably could use an example of what the correct format looks like).

I struggled a bit with an example valid format because it accepts anything that the `new Date` constructor accepts, which is... [a lot](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/parse). Technically `2022-05-17 16:33:25` is also a valid example.
I think it's better to include the date anyways, there have been multiple cases in the past where I've gone to make a timeline and had issues with the python script because I had more than a day's worth of logs in a single file, and it only picked up the first day's encounter, which was completely unrelated.

package.json
valarnin: It should be run-able directly with e.g. `npx ts-node -r jsdom-global/register ...`. The jsdom-global can be auto-require'd by ts-node or by node itself (if you're somehow working with a compiled `.js` version of the script).

As for doing the dance in the file itself, from everything I tested it's not possible due to the way tsc is compiling the script (essentially HTMLElement is getting referenced before `jsdom-global/register` can load, even if `jsdom-global/register` is the first import in the script).

The more glaring issue is with all of the files registered to the `util/index.ts` helper, basically none of them can be run directly, they all have to be run through `util/index.ts`, not just this script. I'm pretty sure there's a way to write them so that they can be (by checking process.args for if it's being run directly, and calling the appropriate function), but maybe open an issue for that so it doesn't get lost?

PR #4447, opened 2022-05-12
config: Fix config page not showing proper checkbox state

valarnin:


quisquous:
Thanks for all the debugging and catching this mistake.

PR #4449, opened 2022-05-12
lint/raidboss: Clean up conditionals in raidboss

valarnin:


ui/raidboss/popup-text.ts
```diff
@@ -1089,7 +1089,7 @@ export class PopupText {
   // Set defaults for triggerHelper object (anything that won't change based on
   // other trigger functions running)
   _onTriggerInternalHelperDefaults(triggerHelper: TriggerHelper): void {
-    if (triggerHelper.triggerAutoConfig) {
+    {
```
valarnin, line 1092: This scoping feels a bit bad to be honest, maybe the const variables should just be renamed?

ui/raidboss/popup-text.ts
```diff
@@ -1266,20 +1266,21 @@ export class PopupText {
...
         const resp: ResponseField<RaidbossData, Matches> = triggerHelper.response;
-        if (resp.tts)
+        if (resp.tts !== undefined)
           result = triggerHelper.valueOrFunction(resp.tts);
       }
 
       // Allow falsey values to disable tts entirely
       // Undefined will fall back to defaultTTSText
       if (result !== undefined) {
+        // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
valarnin, line 1283: This is a case where we actually want to allow the falsey behavior, as indicated by the comment above, so I elected to disable the rule for this line instead.

ui/raidboss/raidboss_config.ts
```diff
@@ -1440,17 +1438,15 @@ const flattenTimeline = (
 const userFileHandler: UserFileCallback = (
   name: string,
   files: { [filename: string]: string },
-  baseOptions: BaseOptions,
+  baseOptions: (BaseOptions & Partial<RaidbossOptions>),
```
valarnin, line 1441: This is more of a stop-gap until the TODO comment below can be handled, but it's more accurate than the `as` cast which was here before.

quisquous: I think this isn't true in practice, though.  I think the options really is a full `RaidbossOptions` and not a partial one.  There's default values for all fields.

I don't think this was true in the past, pre-typescript.  But it is true now, e.g. the default export from `raidboss_options.ts`.

ui/raidboss/timeline_parser.ts
```diff
@@ -395,7 +395,7 @@ export class TimelineParser {
     replaceLang: Lang,
     isGlobal: boolean,
   ): string {
-    if (!this.replacements)
+    if (this.replacements.length === 0)
```
quisquous, line 398: I think this is an incorrect logic change.  An empty array is truthy, and if you early out here you won't get the common replacements below.

ui/raidboss/raidboss_config.ts
```diff
@@ -1544,28 +1540,32 @@ const processPerZoneTimelineConfig = (options: RaidbossOptions, savedConfig: Sav
...
+      enableEntry !== undefined && typeof enableEntry === 'object' && !Array.isArray(enableEntry)
+    ) {
       for (const [key, value] of Object.entries(enableEntry)) {
         if (typeof value === 'boolean' && !value)
           (timelineConfig.Ignore ??= []).push(key);
       }
     }
 
-    if (replaceEntry && typeof replaceEntry === 'object' && !Array.isArray(replaceEntry)) {
+    if (
```
quisquous, line 1559: I think this is somewhat redundant.  If `replaceEntry` is an object, then it is not undefined.  Same with `enableEntry` and `zoneEntry` and `timeline` above.

ui/raidboss/raidboss_config.ts
```diff
@@ -1195,7 +1195,7 @@ class RaidbossConfigurator {
     const evalTrigger = (trig: LooseTrigger, key: OutputKey, data: RaidbossData) => {
       try {
         const result = this.valueOrFunction(trig[key], data, kFakeMatches, fakeOutputProxy);
-        if (!result)
+        if (result === null || result === undefined)
```
quisquous, line 1198: I should probably get rid of all this code.  This was all here to give an idea of trigger output prior to output strings existing, and it doesn't show up if there are output strings.  Not for this PR obviously, but foof this is old and super hacky.

ui/raidboss/raidboss_config.ts
valarnin: Typescript doesn't like it which is why I didn't go for it to begin with.
image
(same error if type is `BaseOptions & RaidbossOptions`)

quisquous: Yeah, agreed that you can't do this without a type assertion.  I was just saying that it *is* a `RaidbossOptions`, regardless of what TypeScript knows about it.

Ok, I think removing the type assertion is fine here and we can clean up the conditionals if this TODO ever gets fixed.

ui/raidboss/popup-text.ts
```diff
@@ -1077,7 +1077,7 @@ export class PopupText {
 
   _onTriggerInternalCondition(triggerHelper: TriggerHelper): boolean {
     const condition = triggerHelper.triggerOptions.Condition ?? triggerHelper.trigger.condition;
-    if (condition) {
+    if (condition !== undefined && condition !== false) {
```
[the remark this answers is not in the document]
valarnin: Yeah, I just tested this against main, that's a fun unintentionally found bug 🙃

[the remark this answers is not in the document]
valarnin: This is the rewrite I'm thinking makes the most sense for this:
```ts
  _onTriggerInternalCondition(triggerHelper: TriggerHelper): boolean {
    const condition = triggerHelper.triggerOptions.Condition ?? triggerHelper.trigger.condition;
    if (condition === undefined || condition === true)
      return true;
    else if (condition === false)
      return false;

    const conditionFuncReturn = condition(this.data, triggerHelper.matches, triggerHelper.output);
    if (conditionFuncReturn === true)
      return true;
    return false;
  }
```

The question is, since `conditionFuncReturn` is of type `boolean | undefined`, what do we do with `undefined`, e.g. what do we do if the condition function returns undefined? Right now, `undefined` is treated as `false`, e.g. abort the trigger because condition failed. The rewritten function above treats it as the same, but maybe a comment to indicate that it's intentional behavior is warranted?

ui/raidboss/timeline_parser.ts
valarnin: So then this check literally did nothing and can be removed entirely? Because `this.replacements` isn't nullable/optional, it's always an array of some sort, so the condition always evaluates to false?

ui/raidboss/popup-text.ts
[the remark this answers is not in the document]
quisquous: Yeah, I think falsey values can be treated as `false` here with a comment.

ui/raidboss/raidboss_config.ts
quisquous: I'm suggesting `typeof replaceEntry === 'object' && !Array.isArray(replaceEntry)` is sufficient here.  If `replaceEntry === undefined` then `typeof replaceEntry === 'object'` already covers that case, I think?

valarnin:
Hopefully that didn't break anything, that was an ugly rebase/merge due to the conflict in `timeline_parser.ts`.

Azruine:
seems like this PR produces error in raid emulator
got following logs in raid emulator console:

popup-text.ts:1465 Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'currentHP')
    at RaidEmulatorPopupText.getDataObject (popup-text.ts:1465:19)
    at new PopupText (popup-text.ts:498:22)
    at new StubbedPopupText (StubbedPopupText.ts:13:5)
    at new RaidEmulatorPopupText (RaidEmulatorPopupText.ts:39:5)
    at raidEmulatorOnLoad (raidemulator.ts:189:21)

ui/raidboss/popup-text.ts
```diff
@@ -1448,7 +1448,7 @@ export class PopupText {
 
   getDataObject(): RaidbossData {
     let preserveHP = 0;
-    if (this.data && this.data.currentHP)
+    if (this.data.currentHP)
       preserveHP = this.data.currentHP;
```
JLGarber, line 1452: This should possibly be `this?.data.currentHP`? Raidboss is currently breaking on the hosted version because of `this.data` not being defined at this point.

```
[5/15/2022 6:06:37 PM] Info: Raidboss: BrowserConsole: Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'currentHP') (Source: https://quisquous.github.io/cactbot/ui/raidboss/raidboss.bundle.js, Line: 15057)
```

Legends0:
It broke the timeline parser and raidboss for me as well, but the 4478 pr fixed it.

PR #4450, opened 2022-05-12
lint/raidboss: Clean up conditionals in raidboss triggers

valarnin:


ui/raidboss/data/05-shb/eureka/delubrum_reginae.ts
```diff
@@ -326,10 +326,6 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`Iron Splitter: null data`);
           return;
         }
-        if (!seekerData.combatants) {
-          console.error(`Iron Splitter: null combatants`);
-          return;
-        }
```
valarnin, line 332: This check seems to exist randomly for only some of the getCombatants calls and is pointless, the actual handler in OverlayPlugin always returns an array here.

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.ts
```diff
@@ -2337,7 +2319,7 @@ const triggerSet: TriggerSet<Data> = {
 
         // Calculate which adjacent zone to go to, if needed
         let adjacentZone = null;
-        if (effectiveTemperature && adjacentZones) {
+        if (effectiveTemperature !== 0) {
```
valarnin, line 2322: I'm not 100% sure on this change, it seems right at a glance but I've got no context for this fight since I've never done DR or DRS.

ui/raidboss/data/05-shb/eureka/delubrum_reginae.ts
quisquous: Yeah, there's a lot of boiler plate here that probably shouldn't be copy pasta'd.  I sometimes wonder if we should have a get combatants promise helper.  Something that has the expected number of things that are returned, maybe even stores it in `data.combatantData` or something so all the logic moves out of the promise.  I don't know.  Just thinking out loud.

ui/raidboss/data/00-misc/test.ts
```diff
@@ -310,7 +310,7 @@ const triggerSet: TriggerSet<Data> = {
             }
             return false;
           }
-          console.log(`test watch: me = ${me ? 'true' : 'false'}; no dummies`);
+          console.log(`test watch: me = undefined; ${dummies.length} dummies`);
```
quisquous, line 313: Should this output reflect if me is defined?

ui/raidboss/data/03-hw/trial/sophia-ex.ts
```diff
@@ -240,7 +240,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegexCn: NetRegexes.startsUsing({ id: '1988', source: '信徒其一' }),
       netRegexKo: NetRegexes.startsUsing({ id: '1988', source: '제1신도' }),
       infoText: (data, matches, output) => {
-        if (Conditions.targetIsYou())
+        if (Conditions.targetIsYou()(data, matches))
```
quisquous, line 243: Sometimes I wonder if `Conditions` were a mistake.  I think this is more readable than the double parens, but feel free to disagree:

```suggestion
        if (data.me === matches.target)
```

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1139,7 +1139,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegexJa: NetRegexes.abilityFull({ source: ['アイスクロウ', 'サンダーウィング', 'ライトファング', 'ダークテイル', 'ファイアホーン'], id: ['26C6', '26C7', '26CA', '26C9', '26C5'] }),
       netRegexCn: NetRegexes.abilityFull({ source: ['冰爪', '雷翼', '光牙', '暗尾', '火角'], id: ['26C6', '26C7', '26CA', '26C9', '26C5'] }),
       netRegexKo: NetRegexes.abilityFull({ source: ['얼음발톱', '번개날개', '빛의 송곳니', '어둠의 꼬리', '화염뿔'], id: ['26C6', '26C7', '26CA', '26C9', '26C5'] }),
-      condition: (data, matches) => !data.seenDragon || !(matches.source in data.seenDragon),
+      condition: (data, matches) => !(matches.source in data.seenDragon),
```
quisquous, line 1142: Hmmm this does seem like a logic change here in that previously this would have always run unconditionally (unintentionally).  I thiiiink it should still work though?

ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.ts
quisquous: Yeah, this is correct.  There's two cleaves, and the first time through (and sometimes the second?) you can resolve it by being hit by the first cleave (and then not needing the second).

ui/raidboss/data/05-shb/raid/e9s.ts
```diff
@@ -842,7 +842,7 @@ const triggerSet: TriggerSet<Data> = {
           names: [cloudOfDarknessLocaleNames[data.parserLang]],
         });
 
-        if (bossData === null || !bossData.combatants || !bossData.combatants.length)
+        if (bossData === null || !bossData.combatants.length)
```
quisquous, line 845:
```suggestion
        if (!bossData?.combatants.length)
```

ui/raidboss/data/05-shb/raid/e9s.ts
```diff
@@ -867,7 +867,7 @@ const triggerSet: TriggerSet<Data> = {
           names: [cloneOfDarknessLocaleNames[data.parserLang]],
         });
 
-        if (cloneData === null || !cloneData.combatants || !cloneData.combatants.length)
+        if (cloneData === null || !cloneData.combatants.length)
```
quisquous, line 870:
```suggestion
        if (!cloneData?.combatants.length)
```

ui/raidboss/data/03-hw/trial/sophia-ex.ts
JLGarber: It's worth noting that this was written shortly after `Conditions` were introduced and before we had really solidified what we wanted out of them. Given that the usage here is inside a popup text element, rather than as a `Condition` property, I agree that the suggestion is the way we should take this.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
valarnin: The conditional here was a no-op, as outlined elsewhere `(!{}) === false`, it looks like it was just an `undefined` check before, not relying on false-y empty object behavior.

quisquous: Oh, you're right you're right.  I was misreading this as `true || etc` and not `false || etc`, sorry.

PR #4451, opened 2022-05-12
lint/raidemulator: Clean up conditionals in raidemulator

valarnin:


ui/raidboss/emulator/ui/Tooltip.ts
```diff
@@ -124,18 +127,6 @@ export default class Tooltip {
...
-      return;
-
-    Tooltip.templates = {
-      top: Tooltip.getTemplate('top'),
-      right: Tooltip.getTemplate('right'),
-      bottom: Tooltip.getTemplate('bottom'),
-      left: Tooltip.getTemplate('left'),
-    };
-  }
-
```
valarnin, line 138: I think this could originally race condition back when raidemulator was javascript, but that won't happen now that it's all in typescript.

PR #4452, opened 2022-05-12
lint: Clean up conditionals in misc other files

valarnin:


resources/user_config.ts
```diff
@@ -37,6 +37,7 @@ const ZoneInfo = _ZoneInfo;
 // Convince TypeScript and eslint that these are used.  TypeScript doesn't have a great way
 // to disable individual rules, so this is safer than disabling all rules.
 console.assert(
+  // eslint-disable-next-line @typescript-eslint/strict-boolean-expressions
```
valarnin, line 40: I wasn't really sure what the best way to handle this would be. Break it apart into 8 different `console.assert` calls?

I think the comment above covers why the eslint disable is required as well.

test/helper/test_trigger.ts
```diff
@@ -581,11 +581,12 @@ const testTriggerFile = (file: string) => {
           const funcStr = func.toString();
           const keys: string[] = [];
 
-          dynamicOutputStringAccess = dynamicOutputStringAccess || /\boutput\[/.test(funcStr);
+          dynamicOutputStringAccess = dynamicOutputStringAccess ||
+            /(?:\boutput\[|\boutput[^.\r\n])/.test(funcStr);
```
valarnin, line 585: Added `output`-passed-as-function-argument to this regex, which covers a bunch of other cases that were getting excluded due to bad logic below.

test/helper/test_trigger.ts
```diff
@@ -607,7 +608,7 @@ const testTriggerFile = (file: string) => {
 
         // Responses can have unused output strings in some cases, such as ones
         // that work with and without matching.
-        if (!response && !dynamicOutputStringAccess) {
+        if (!dynamicOutputStringAccess) {
```
valarnin, line 611: This logic was faulty and causing a lot of cases to not get tested properly. `(!{}) === false`

ui/raidboss/data/06-ew/raid/p1n.ts
```diff
@@ -89,8 +89,8 @@ const triggerSet: TriggerSet<Data> = {
       type: 'GainsEffect',
       netRegex: NetRegexes.gainsEffect({ effectId: ['AB3', 'AB4'], capture: true }),
       condition: Conditions.targetIsYou(),
-      alertText: (_data, _matches, _output) => {
-        return _matches.effectId === 'AB3' ? _output.red!() : _output.blue!();
+      alertText: (_data, _matches, output) => {
+        return _matches.effectId === 'AB3' ? output.red!() : output.blue!();
```
valarnin, line 93: Renamed parameters here and below to not be underscore-prefixed, so that they would properly get picked up by test_trigger.ts.

util/fetch_deps.ts
```diff
@@ -211,16 +211,15 @@ export const main = async (updateHashes = false): Promise<void> => {
...
 
-      for (const [oldCache, newCache] of Object.entries(hashUpdateMap))
-        data.replace(oldCache, newCache);
+    for (const [oldCache, newCache] of Object.entries(hashUpdateMap))
+      data.replace(oldCache, newCache);
+
+    await fs.writeFile(depsPath, data);
 
-      await fs.writeFile(depsPath, data);
-    }
```
valarnin, line 223: It might make more sense to check `Object.keys().length` instead of removing this check entirely?

util/gen_coverage_report.ts
```diff
@@ -251,7 +249,7 @@ const buildTotals = (coverage: Coverage) => {
       totalsByContentType.raidboss++;
       totals.overall.raidboss++;
     }
-    if (thisCoverage?.oopsy?.num ?? 0 > 0) {
+    if ((thisCoverage?.oopsy?.num ?? 0) > 0) {
```
valarnin, line 252: This logic was completely broken, prior to this change it evaluated to:
```ts
    if (thisCoverage?.oopsy?.num ?? (0 > 0)) {
```
See:
image

resources/responses.ts
```diff
@@ -137,14 +137,20 @@ const combineFuncs = function(
...
   } else {
     obj[text1] = (data: Data, matches: TargetedMatches, output: Output) => {
-      return func1(data, matches, output) || func2(data, matches, output);
+      const func1Value = func1(data, matches, output);
+      if (func1Value !== undefined && func1Value !== null && func1Value === true)
+        return true;
+      const func2Value = func2(data, matches, output);
+      if (func2Value !== undefined && func2Value !== null && func2Value === true)
+        return true;
+      return func1Value !== false || func2Value !== false;
```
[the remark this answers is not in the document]
quisquous: Mmm, I think this is meant to be `return func1(data, matches, output) ?? func2(data, matches, output);` and not a boolean value.

resources/user_config.ts
quisquous: Yeah, I think the comment covers this.  If the individual lines wouldn't need an eslint disable, that seems like one avenue if you like that better, but if they all would need eslint disable this seems best.

test/helper/test_trigger.ts
quisquous: Wow, thanks for fixing this.

util/fetch_deps.ts
quisquous: Yeah, I think that'd make sense.

util/gen_coverage_report.ts
quisquous: Oops!

resources/user_config.ts
valarnin: It only needs an eslint-disable because the variables are being `&&`-ed together. I think breaking this into multiple asserts is better in general since it makes it easier to add/remove them as needed later?

quisquous: Sounds good to me.

PR #4454, opened 2022-05-12
lint: Clean up conditionals in misc UI components

valarnin:


valarnin:
As is generally the case for my PRs, since I don't really use the UI components outside raidboss, raidemulator, and oopsy, these changes should probably have extra scrutiny/testing.

ui/oopsyraidsy/oopsyraidsy_config.ts
```diff
@@ -235,14 +235,14 @@ const templateOptions: OptionsTemplate = {
       return;
     const triggers = savedConfig['triggers'];
 
-    if (!triggers || typeof triggers !== 'object' || Array.isArray(triggers))
+    if (triggers === undefined || typeof triggers !== 'object' || Array.isArray(triggers))
```
quisquous, line 238:
```suggestion
    if (typeof triggers !== 'object' || Array.isArray(triggers))
```

ui/oopsyraidsy/oopsyraidsy_config.ts
```diff
@@ -365,17 +365,15 @@ const templateOptions: OptionsTemplate = {
 const userFileHandler: UserFileCallback = (
   name: string,
   _files: { [filename: string]: string },
-  baseOptions: BaseOptions,
+  baseOptions: BaseOptions & Partial<OopsyOptions>,
```
quisquous, line 368: Same comment here as on the RaidbossOptions.  I don't think this is Partial.

ui/oopsyraidsy/oopsyraidsy_config.ts
valarnin: I think this was meant to be `!== null`, because:
image
Does that make sense instead?

ui/jobs/components/ast.ts
```diff
@@ -97,9 +97,7 @@ export class ASTComponent extends BaseComponent {
     // Show whether you already have this seal
     // ○ means it's OK to play this card
     // × means you'd better redraw if possible
-    if (!cardsMap[card])
-      this.cardBox.innerText = '';
-    else if (sign.includes(cardsMap[card].seal))
+    if (sign.includes(cardsMap[card].seal))
```
Echoring, line 100: Cannot remove O/X after play the card. Sign color also not updated until next draw.
With the following error:
```
TypeError: Cannot read properties of undefined (reading 'seal')
    at ASTComponent.onJobDetailUpdate (ast.ts:100)
```

Echoring (commented):
Seems some part of jobs UI getting broken.
I'm not sure what if `mpTicker` `leftBuffDiv` `Component` not exist (can them not exist?)

ui/oopsyraidsy/oopsyraidsy_config.ts
MaikoTan: What about to add helper functions like `isObject` and `isNullish` so you don't need to repeat so many times. Or consider to import lodash-es (so that webpack would do the tree-shaking things).

ui/jobs/event_emitter.ts
```diff
@@ -121,15 +121,11 @@ export class JobsEventEmitter extends EventEmitter<EventMap> {
...
-        effectiveDistance: target.EffectiveDistance,
-      });
-    } else {
-      this.emit('battle/target');
-    }
+    this.emit('battle/target', {
+      name: target.Name,
+      distance: target.Distance,
+      effectiveDistance: target.EffectiveDistance,
+    });
```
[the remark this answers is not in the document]
valarnin: If this is the case, then the underlying event definition in `event.d.ts` is wrong and should be corrected instead, which will then require the null check here as appropriate.

Said underlying event definition is also missing entries for `Focus` and `Hover`, if we're updating it we should probably update it to fully match the spec?

ui/jobs/components/ast.ts
valarnin: The typing of `cardsMap` doesn't allow `undefined`, this seems like a bug in general.

```ts
const cardsMap = {
  'Balance': { 'bonus': 'melee', 'seal': 'Solar' },
  'Bole': { 'bonus': 'range', 'seal': 'Solar' },
  'Arrow': { 'bonus': 'melee', 'seal': 'Lunar' },
  'Ewer': { 'bonus': 'range', 'seal': 'Lunar' },
  'Spear': { 'bonus': 'melee', 'seal': 'Celestial' },
  'Spire': { 'bonus': 'range', 'seal': 'Celestial' },
} as const;
```

If this change is breaking functionality then something related to either `cardsMap` or `card` is not defined correctly and should be updated. I'm guessing maybe `JobDetail['AST'].heldCard` should be optional or allow null?

ui/jobs/event_emitter.ts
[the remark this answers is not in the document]
MaikoTan: Agree to update the definition, also I think the null check here is needed.

ui/jobs/components/ast.ts
MaikoTan: This is not a matter of `cardsMap`. If player do not held a card, card will be `'None'`, and `cardsMap[card]` returns `undefined`. If plugin remains unchanged then there should be a condition.
By the way, I think `jobDetail.heldCard` and `jobDeail.crownCard` should return `null/undefined` when no card.

valarnin: They should map to `'None'` when no card is held, if I'm reading the corresponding C# code properly? So the proper typedef would be:
```ts
  'AST': {
    heldCard: 'None' | 'Balance' | 'Bole' | 'Arrow' | 'Spear' | 'Ewer' | 'Spire';
    crownCard: 'None' | 'Lord' | 'Lady';
    arcanums: ('Solar' | 'Lunar' | 'Celestial')[];
  };
```

Corresponding C# code just toString's the enum:
```cs
      public string heldCard {
        get {
          return ((Card)(_heldCard & 0xF)).ToString();
        }
      }

      public string crownCard {
        get {
          return ((Card)(_heldCard & 0xF0)).ToString();
        }
      }
```

ui/jobs/event_emitter.ts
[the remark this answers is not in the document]
valarnin: See latest commit, event typedef is updated to match the C# definition, and null check is in place.
Distance is string and was just working because of type coercion before, so I elected to add `parseFloat`s to match the expected typing.

image

ui/jobs/components/ast.ts
Echoring: Seems both problem solved. Thank you!

valarnin:
@Echoring With regards to the other comments/concerns:

1. mpTicker - The removed false-y check was against `const mpTicker = TimerBar.create()` which always returns a valid object (`TimerBar extends HTMLElement`)
2. leftBuffDiv - The arguments to the constructor here are checked for valid values before being passed in, hence why the constructor's parameters aren't optional/nullable. See `ui/jobs/components/index.ts:233`.
3. Component - `ComponentMap` has all jobs defined and mapped to a valid component object, so the check for `Component` to be valid is redundant. `Record<Job, typeof BaseComponent>` requires that all jobs have an entry in the map.

ui/oopsyraidsy/oopsyraidsy_config.ts
quisquous: > I think this was meant to be !== null

I think `triggers` cannot be null, fwiw.

> What about to add helper functions like isObject and isNullish so you don't need to repeat so many times

Unless I am misunderstanding, a lot of these checks are here for TypeScript.  In other words, if you remove the `Array.isArray` check here, then TypeScript will complain.  If you call a function that says `isObject` instead, that isn't enough information for TypeScript to deduce that it's not an array.

valarnin: In addition to what quisquous said above, the end goal of these PRs is to remove nullish/falsey checks and instead be explicit about what values are being checked, to avoid issues like #4447, or the change being discussed here: https://github.com/quisquous/cactbot/pull/4450#discussion_r871907682, so it makes no sense to just move the nullish/falsey checks to elsewhere and have the underlying potential for bugs continue.

quisquous:
This looks good to me.  @Echoring, @MaikoTan did y'all want to take another look at these changes?

Echoring (approved):

MaikoTan:
Looks good to me. Thanks!

trim21:
`AggressionStatus` is added in this PR, what's it's type?

valarnin:
As with most questions like this, the answer is upstream in OverlayPlugin.

https://github.com/OverlayPlugin/OverlayPlugin/blob/5416039b970a3ea986f7ed52c63d6f7084b8f899/OverlayPlugin.Core/MemoryProcessors/EnmityMemoryCommon.cs#L82-L96

PR #4455, opened 2022-05-12
test: fix test for missing regex translations

quisquous:
The NetRegexes flag needs to be set prior to import when it is
used for evaluation.  This was broken due to a final change to #4433
without being properly retested.

Also, some small cleanups while here.

This is working towards #1306 as well.


quisquous:
Landing this promptly so that no errors creep in here.

PR #4460, opened 2022-05-14
util: Typescriptify make_timeline

JLGarber:
Lots and lots of nonsense, but at a bare minimum, it can produce a basic timeline. We have duplicated (for the most part) the logic of our Python version with some small improvements here and there. Speed is comparable between the two versions. (We should be able to get more speed in future with some improvements to `EncounterCollector`.)

All `ignore` flags should work, as should `-it` and `-oc`.  

A *lot* of the weird-looking declarations are to satisfy the compiler. Most of them are "the heck with it, *now* it's guaranteed defined". This is particularly true in interactions with `args`, since we can't guarantee any specific one is populated from Typescript's point of view. (Also there are very possibly remnants from VSCode's debugger, since it sometimes has subtle behavior differences from a non-integrated shell.)

There's still plenty left to do on this, but this should be a reasonable foundation to build on.

### DIFF
--- a/util/logtools/make_timeline.ts
+++ b/util/logtools/make_timeline.ts
@@ -166,14 +166,14 @@ if (args['fight_regex'] === '-1')
 if (args['zone_regex'] === '-1')
   printHelpAndExit('Error: -zr must specify a regex\n');
 
-if (args.file && !(args.start && args.end)) {
+if (args.file && numExclusiveArgs !== 1 && !(args.start && args.end)) {
   printHelpAndExit(
     'Error: -s and -e must be specified if not using one of -lf, -lz, -fr, or -zr\ ',
   );
 }
-if (!(typeof args.start === 'string' && timeStampValidate(args.start)))
+if (args.start && !(typeof args.start === 'string' && timeStampValidate(args.start)))
   printHelpAndExit('Error: Timestamp must be entered exactly as HH:MM:SS.DDD, as 12:34:56:.789');
-if (!(typeof args.end === 'string' && timeStampValidate(args.end)))
+if (args.end && !(typeof args.end === 'string' && timeStampValidate(args.end)))
   printHelpAndExit('Error: Timestamp must be entered exactly as HH:MM:SS.DDD, as 12:34:56:.789');
 
 const makeCollectorFromPrepass = async (fileName: string) => {
@@ -237,11 +237,11 @@ const extractTLEntries = (
   });
   file.on('line', (line) => {
     const lineTimeStamp = line.substring(14, 26);
-    if ([start, end].includes(lineTimeStamp))
+    if (start === lineTimeStamp)
       started = start === lineTimeStamp;
     if (started)
       lines.push(line);
-    if (!started && lines.length > 0)
+    if (end === lineTimeStamp)
       file.close();
   });
 
@@ -296,15 +296,72 @@ const extractTLEntries = (
   return entries;
 };
 
-const assembleTimelineStrings = (entries: TimelineEntry[], args: TLArgs): string[] => {
+const ignoreTimelineAbilityEntry = (entry: TimelineEntry, args: ExtendedArgs): boolean => {
+  const abilityName = entry.abilityName;
+  const abilityId = entry.abilityId;
+  const combatant = entry.combatant;
+  // Ignore auto-attacks named "attack"
+  if (abilityName?.toLowerCase() === 'attack')
+    return true;
+
+  // Ignore abilities from NPC allies.
+  if (combatant && ignoredCombatants.includes(combatant))
+    return true;
+
+  // Ignore abilities by name.
+  if (abilityName && args['ignore-ability'] && args['ignore-ability'].includes(abilityName))
+    return true;
+
+  // Ignore abilities by ID
+  if (abilityId && args['ignore-id'] && args['ignore-id'].includes(abilityId))
+    return true;
+
+  // Ignore combatants by name
+  if (combatant && args['ignore-combatant'] && args['ignore-combatant'].includes(combatant))
+    return true;
+
+  // If only-combatants was specified, ignore all combatants not in the list.
+  if (combatant && args['only-combatant'] && !args['only-combatant'].includes(combatant))
+    return true;
+  return false;
+};
+
+const findTimeDifferences = (lastTimeDiff: number): { diffSeconds: number; drift: number } => {
+  let diffSeconds = lastTimeDiff % 1000000;
+  const diffMicroSeconds = lastTimeDiff - diffSeconds;
+  let drift = 0;
+
+  // Find the difference in tenths of a second.
+  const diffTenthSeconds = diffMicroSeconds / 100000;
+
+  // Adjust full-second difference.
+  diffSeconds += diffTenthSeconds;
+
+  // Round up a tenth of a second.
+  if (diffMicroSeconds > 60000) {
+    diffSeconds += 0.1;
+  } else if (diffMicroSeconds > 50000) {
+    // Round up, warning of exceptional drift.
+    diffSeconds += 0.1;
+    drift = diffMicroSeconds - 100000;
+  } else if (diffMicroSeconds > 40000) {
+    // Round down, warning of exceptional drift
+    drift = diffMicroSeconds;
+  } else {
+    // If <20ms then there's no need to adjust sec or drift
+    true;
+  }
+  return { diffSeconds: diffSeconds, drift: drift };
+};
+
+const assembleTimelineStrings = (
+  entries: TimelineEntry[],
+  start: string,
+  args: ExtendedArgs,
+): string[] => {
   const assembled: string[] = [];
-  // const phases: {[name: string]: string } = {};
-  // if (args.phase && args.phase is typeof string[]) {
-  //   for (const phase of args.phase)
-  // }
-  const timelinePosition = 0;
-  const lastAbilityTime = entries[0]?.time;
-  const ignoredAbilities = args['ignore-ability'] as string[];
+  let lastAbilityTime = parseInt(start);
+  let lastEntry: TimelineEntry = { time: start ?? 'Unknown', lineType: 'None' };
   if (entries[0]?.zoneSeal) {
     const zoneMessage = TLFuncs.toProperCase(entries[0].zoneSeal.seal);
     const tlString = `0.0 "--sync--" sync / 00:0839::${zoneMessage} will be sealed off/ window 0,1`;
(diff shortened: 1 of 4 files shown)


util/logtools/make_timeline.ts
```diff
@@ -0,0 +1,512 @@
...
+  }
+};
+
+const makeTimeline = async () => {
+  if (args.file) {
+    const collector = await makeCollectorFromPrepass(args.file);
+    if (args['search_fights'] === -1) {
+      TLFuncs.printCollectedFights(collector);
+      process.exit(0);
+    }
```
JLGarber, line 473: It occurs to me that if we drop support for manual start/end timestamps (or even if we don't) we could eliminate the need to extract lines a second time. I quickly hacked something together on a new branch to test this, and it appears to work okay. We could do something awful like:

```typescript
  if (args.file) {
    const store = (args.search_fights && (args.search_fights > 0)) as boolean;
    const collector = await makeCollectorFromPrepass(args.file, store);
    if (args['search_fights'] === -1) {
      TLFuncs.printCollectedFights(collector);
      process.exit(0);
    }
// etc
```

And have this store an array of lines to each fight in the `EncounterCollector`. It might end up being slightly impractical if someone's logs go over 2 GB, but that's a them problem at that point.

I don't have things set up to benchmark it, but it has to be faster overall since we're completely removing a file access and a loop based on it. The number of lines to loop over in `extractTLEntries` should still end up being the same.

util/logtools/make_timeline.ts
```diff
@@ -0,0 +1,512 @@
...
+  'Moonstone Carbuncle',
+  'Rook Autoturret',
+  'Bishop Autoturret',
+  'Demi-Bahamut',
+  'Demi-Phoenix',
+  'Earthly Star',
+  'Seraph',
+  'Automaton Queen',
+  'Bunshin',
+  'Esteem',
```
valarnin, line 73: For pet names, maybe it makes sense to pull them in from the auto-generated file instead of duplicating them here?

util/logtools/make_timeline.ts
```diff
@@ -0,0 +1,512 @@
...
+const extractRawLines = async (fileName: string, start: string, end: string): Promise<string[]> => {
+  const lines: string[] = [];
+  const file = readline.createInterface({
+    input: fs.createReadStream(fileName),
+  });
+  let started = false;
+  for await (const line of file) {
+    // This will fail on lines with 3-digit identifiers,
+    // but that's okay because those will never be start lines.
+    const lineTimeStamp = line.substring(14, 26);
```
valarnin, line 232: For this logic (and elsewhere in the file), RaidEmulator's ParseLine could be leveraged instead (and maybe it's time for that functionality to move out of the RaidEmulator subset of code and into `resources` or something).

valarnin:
With regards to args, argparse through typescript can handle this properly but it requires a bit different syntax. Example:

```ts
import argparse, { Namespace } from 'argparse';

interface TestArgs {
  'include_targetable': string[] | null;
}

class ExtendedArgsNS extends Namespace implements TestArgs {
  'include_targetable': string[] | null;
}

const parser = new argparse.ArgumentParser({
  addHelp: true,
});

parser.addArgument(['--include_targetable', '-it'], {
  nargs: '+',
  help: 'Set this flag to include "34" log lines when making the timeline',
});

const ns = new ExtendedArgsNS({});

parser.parseArgs(undefined, ns);

console.log(ns);
console.log(ns['include_targetable']);
```
Output:
```
// with `-it 1 2 3 4`
ExtendedArgsNS { include_targetable: [ '1', '2', '3', '4' ] }
[ '1', '2', '3', '4' ]
// with no arguments
ExtendedArgsNS { include_targetable: null }
null
```

Also note that argparse is automatically converting those arguments' dashes to underscores, so your current code's `-it` parameter probably doesn't work.

```ts
import argparse from 'argparse';

interface TestArgs {
  'include-targetable': string[] | null;
}

const parser = new argparse.ArgumentParser({
  addHelp: true,
});

parser.addArgument(['--include-targetable', '-it'], {
  nargs: '+',
  help: 'Set this flag to include "34" log lines when making the timeline',
});

const ns = parser.parseArgs() as TestArgs;

console.log(ns);
console.log(ns['include-targetable']);
```
Output:
```
Namespace { include_targetable: [ '1', '2', '3', '4' ] }
undefined
```

util/logtools/make_timeline.ts
JLGarber: Definitely does, I totally forgot that existed. The generated listing isn't complete, but there certainly is no reason to duplicate the ones that it does contain.

util/logtools/make_timeline.ts
JLGarber: Yes, you're right. I haven't really looked at the internals of the emulator much, and this was intended as a foundational mostly 1-1 conversion, but looking through `EmulatorCommon` there's a lot of time handling that could be brought into resources and removed from several other places. I think it might be better to do that as a follow-up after landing this (and maybe your `test` implementation as well.)

util/logtools/make_timeline.ts
valarnin: If the generated list isn't complete then the tool to generate it should probably be updated?

JLGarber:
Thanks for the explanation. The Typescript implementation has a lot less documentation than I would like, and it's tough to find information on some of the differences there without doing a lot of exploration. (It doesn't help that VSCode's Intellisense defaults to `args['dash-prop']` syntax.)

util/logtools/make_timeline.ts
JLGarber: Probably so. It looks like that file hasn't been refreshed since 6.0, so that's probably part of it. As for adding NPCs, I would have to check through the XIVAPI listings to see where they live.

quisquous: This has mostly just been hand-edited by me over time to account for pets and for trust names.  I will say that Estinien is problematic because it skips the Estinien Resentment cast in DSR.

JLGarber: That's unfortunate. I guess maybe we could add support for overriding the ignore list? That, or we could just not ignore Estinien and have to remember to manually ignore him for all future dungeons we run with trusts?

quisquous: Yeah, I wouldn't overthink it here.  It's a minor inconvenience we can sort out later with either of those alternatives, I think.

JLGarber:
This didn't show up locally on `test`, unfortunately.

I think this is about as far as I can take things without moving a bunch of things around. I've noted most of the TODOs in-line. Currently selecting fights by index works, and gives substantially similar results to our Python implementation.

@valarnin I think you're right, we can definitely leverage a lot of the emulator code for this. I think moving the time-string parsing functions to a resource file makes sense, and should be relatively simple. `ParseLine` would be similarly useful, but it might be a little more complex to move.

util/logtools/make_timeline.ts
```diff
@@ -0,0 +1,475 @@
...
+        entries.push(targetEntry);
+      }
+      continue;
+    }
+
+    // At this point, only ability lines are left.
+    if (ability) {
+      // Cull non-enemy lines
+      // TODO: Handle this using the raid eumlator's line parsing functionality.
+      if (ability.sourceId.substring(0, 3) !== '400')
```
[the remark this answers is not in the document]
valarnin: `id[0] === '4'` is used in a ton of places. This looks like a direct port of the original python logic, it should probably be adjusted to match logic elsewhere in the project.
```python
            # We're looking for enemy casts or enemies becoming targetable/untargetable.
            # These lines will start with 21, 22, or 34, and have an NPC ID (400#####)
            # If none of these apply, skip the line

            if not line[0:2] in ["21", "22", "34"] or not line[37:40] == "400":
                continue
```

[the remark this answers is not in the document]
JLGarber: Yes, I don't remember why we settled on that at the time. I'm not at all attached to it.

util/logtools/make_timeline.ts
```diff
@@ -0,0 +1,475 @@
...
+    return entries;
+  }
+};
+
+const ignoreTimelineAbilityEntry = (entry: TimelineEntry, args: ExtendedArgs): boolean => {
+  const abilityName = entry.abilityName;
+  const abilityId = entry.abilityId;
+  const combatant = entry.combatant;
+
+  const ia = args.ignore_ability as string[] | null;
```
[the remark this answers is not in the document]
JLGarber: This could be a little messy to work around. I think we would need to revisit how we're typing `args` generally, since `argParse` just hands`Any` to us. The best I can think of  is "top-level check that any non-null args are of type X" on a per-script basis. We won't be able to export a validation function like this alongside the basic `TimelineArgs` type since a runtime check isn't going to be good enough.

All that is to say, any such changes we make are going to cascade through a few different files in the `logtools` folder. I'm fine doing that, and it definitely *needs* to be done, but it's going to mean this PR contains a little more than just the addition of `make_timeline`. Alternatively, we can put this one on hold and I can go through and do that cascading cleanup first.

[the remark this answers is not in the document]
valarnin: Line 123 where you call argparse.parse, that will accept a pre-created namespace object and populate that object with the parameters instead. See:
https://github.com/quisquous/cactbot/blob/c764a17fc1f92e2d73c5b3b0c1f68f9ad613bafa/util/index.ts#L42-L48
https://github.com/quisquous/cactbot/blob/c764a17fc1f92e2d73c5b3b0c1f68f9ad613bafa/util/index.ts#L54
https://github.com/quisquous/cactbot/blob/c764a17fc1f92e2d73c5b3b0c1f68f9ad613bafa/util/index.ts#L87-L90
https://github.com/quisquous/cactbot/blob/c764a17fc1f92e2d73c5b3b0c1f68f9ad613bafa/util/index.ts#L56

This doesn't really directly offer type safety itself, but it at least doesn't require the `... as ...` casts.

Here's a full diff which makes the requisite adjustments. It compiles with `npm run tsc-no-emit`.

```diff
diff --git a/util/logtools/arg_parser.ts b/util/logtools/arg_parser.ts
index 4c932fb21..fc1dd3eb7 100644
--- a/util/logtools/arg_parser.ts
+++ b/util/logtools/arg_parser.ts
@@ -1,14 +1,13 @@
 import argparse from 'argparse';
 
 type TimelineArgs = {
-  [index: string]: string | number | boolean | undefined;
-  'file'?: string;
-  'force'?: boolean;
-  'search_fights'?: number;
-  'search_zones'?: number;
-  'fight_regex'?: string;
-  'zone_regex'?: string;
-  'adjust'?: number;
+  'file': string | null;
+  'force': boolean | null;
+  'search_fights': number | null;
+  'search_zones': number | null;
+  'fight_regex': string | null;
+  'zone_regex': string | null;
+  'adjust': number | null;
 };
 
 class LogUtilArgParse {
diff --git a/util/logtools/make_timeline.ts b/util/logtools/make_timeline.ts
index e1f237c00..bd524ed00 100644
--- a/util/logtools/make_timeline.ts
+++ b/util/logtools/make_timeline.ts
@@ -1,6 +1,8 @@
 import fs from 'fs';
 import readline from 'readline';
 
+import { Namespace } from 'argparse';
+
 import NetRegexes from '../../resources/netregexes';
 import PetData from '../../resources/pet_names';
 import { NetMatches } from '../../types/net_matches';
@@ -37,18 +39,36 @@ type TimelineEntry = {
 };
 
 type ExtendedArgs = TimelineArgs & {
-  [index: string]: string | string[] | number | boolean | undefined;
-  'output-file'?: string;
-  'start'?: string;
-  'end'?: string;
-  'ignore-id'?: string[];
-  'ignore-ability'?: string[];
-  'ignore-combatant'?: string[];
-  'only-combatant'?: string[];
-  'phase'?: string;
-  'include-targetable'?: string[];
+  'output_file': string | null;
+  'start': string | null;
+  'end': string | null;
+  'ignore_id': string[] | null;
+  'ignore_ability': string[] | null;
+  'ignore_combatant': string[] | null;
+  'only_combatant': string[] | null;
+  'phase': string | null;
+  'include_targetable': string[] | null;
 };
 
+class ExtendedArgsNamespace extends Namespace implements ExtendedArgs {
+  'file': string | null;
+  'force': boolean | null;
+  'search_fights': number | null;
+  'search_zones': number | null;
+  'fight_regex': string | null;
+  'zone_regex': string | null;
+  'adjust': number | null;
+  'output_file': string | null;
+  'start': string | null;
+  'end': string | null;
+  'ignore_id': string[] | null;
+  'ignore_ability': string[] | null;
+  'ignore_combatant': string[] | null;
+  'only_combatant': string[] | null;
+  'phase': string | null;
+  'include_targetable': string[] | null;
+}
+
 // Some NPCs can be picked up by our entry processor.
 // We list them out explicitly here so we can ignore them at will.
 const ignoredCombatants = PetData['en'].concat([
@@ -120,7 +140,8 @@ timelineParse.parser.addArgument(['--include_targetable', '-it'], {
   nargs: '+',
   help: 'Set this flag to include "34" log lines when making the timeline',
 });
-const args = timelineParse.parser.parseArgs() as ExtendedArgs;
+const args = new ExtendedArgsNamespace({});
+timelineParse.parser.parseArgs(undefined, args);
 
 const printHelpAndExit = (errString: string): void => {
   console.error(errString);
@@ -134,7 +155,8 @@ if (args.file === null)
 if (!args.file?.includes('.log'))
   printHelpAndExit('Error: Must specify an FFXIV ACT log file, as log.log\n');
 let numExclusiveArgs = 0;
-for (const opt of ['search_fights', 'search_zones']) {
+const exclusiveArgs = ['search_fights', 'search_zones'] as const;
+for (const opt of exclusiveArgs) {
   if (args[opt] !== null)
     numExclusiveArgs++;
 }
@@ -266,10 +288,10 @@ const ignoreTimelineAbilityEntry = (entry: TimelineEntry, args: ExtendedArgs): b
   const abilityId = entry.abilityId;
   const combatant = entry.combatant;
 
-  const ia = args.ignore_ability as string[] | null;
-  const ii = args.ignore_id as string[] | null;
-  const ic = args.ignore_combatant as string[] | null;
-  const oc = args.only_combatant as string[] | null;
+  const ia = args.ignore_ability;
+  const ii = args.ignore_id;
+  const ic = args.ignore_combatant;
+  const oc = args.only_combatant;
   // Ignore auto-attacks named "attack"
   if (abilityName?.toLowerCase() === 'attack')
     return true;
```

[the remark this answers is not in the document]
JLGarber: Well, that's great. Thanks for taking the time to put that together. I wasn't anticipating having much time tonight because of reclears, so that's a lovely savings.

Issue #4475, opened 2022-05-15
Alerts and Timelines are not working anymore on last build


mooondark:
With Fr language timeline and alerts are no more working since 0.26.7. I've rolled back on 0.26.6 and it's working again.

quisquous:
I don't have any issues with the parser language set to French and trying out things in Summerford Farms.  Can you be more specific about what's not working? @Akurosia are you having any issues in German?

mooondark:
I'm on P4S right now, I've no calls and no timelines with 0.26.7.
Triggernometry is working fine & mopimopi too.

I'll go to Summerford after the raid

mooondark:
Fixed in 0.26.8

PR #4478, opened 2022-05-15
raidboss: Fix issue with data init in PopupText

valarnin:
Fixes https://github.com/quisquous/cactbot/pull/4449#issuecomment-1127046058

This seems like a bit of a code smell issue in general, and is probably something that tsc or a linter should catch?

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -1462,7 +1462,8 @@ export class PopupText {
 
   getDataObject(): RaidbossData {
     let preserveHP = 0;
-    if (this.data.currentHP)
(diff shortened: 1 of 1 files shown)


quisquous:
I wonder if the P4S issue in #4475 is this as well.  @mooondark do you see console errors that look like it could be this?

quisquous:
> This seems like a bit of a code smell issue in general, and is probably something that tsc or a linter should catch?

Agreed, but I think the constructor is just always an indeterminate state with the guarantees only happening at the end, and sometimes you can have fields that need to depend on other fields when setting up state.  I'm not sure what pattern would fix this in the general case.

PR #4479, opened 2022-05-16
raidboss: DSR Wyrmsbreath 2 Boiling and Freezing

Legends0:
The second Wyrmsbreath in phase 6 utilizes Boiling and Freezing debuffs. I am not sure if it's possible to get these debuffs by failing the first Wyrmsbreath, but the debuffs do not appear in a kill log at that time that I looked at and so I did not think to add a check for that.

This also brings back Deep Freeze and Pyretic debuffs from old. Deep Freeze looks to be unavoidable, but Pyretic will require an advance notice such that the player is not moving when Boiling expires as there is a brief period of time between Cauterize hit and Boiling expiration where Pyretic is on the player.

I am not certain on the exact timing of when the dragons are in position, but it looks like it's at least random left/right which dragon is where and I am not sure if the cardinal they choose is also random. Position could be done either with Cauterize cast position and collection trigger or modifying this trigger with OverlayHandler in the future which I think the dragons may jump to position prior to starting the cast.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -12,6 +12,7 @@ import { LocaleText, TriggerSet } from '../../../../../types/trigger';
 
 // TODO: Ser Adelphel left/right movement after initial charge
 // TODO: Meteor "run" call?
+// TODO: Wyrmsbreath 2 cardinal positions for Cauterize and adjust delay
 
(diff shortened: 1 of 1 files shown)


quisquous:
Does this also occur in the first wyrmsbreath too? Should it be suppressed there somehow?

Legends0:
> Does this also occur in the first wyrmsbreath too? Should it be suppressed there somehow?
I looked at quite a few 12:30 wipes during this phase and do not see these debuffs in the first wyrmsbreath. None of the kill logs have it during the first one, so it looks like they only happen in the 2nd.

PR #4486, opened 2022-05-16
docs: update docs for removing netRegexDe etc

quisquous:
Follow up to #4474.


MaikoTan:
https://github.com/quisquous/cactbot/blob/9bcb0b954b50f655cccb035e6e42bc334fe2b775/docs/zh-CN/CactbotCustomization.md?plain=1#L218

Here is a `regexCn` that is not removed.

Also a quick question, as far as I can see, after these several PRs, the `regex`-es and `netRegex`-es can be translated based on `timelineReplace`, is it right?
Then what if users write a new trigger that has `name` which is not exist in the original files, should user also provides the `timelineReplace` section includes the `name` and its translation?

PR #4489, opened 2022-05-18
raidboss: DSR Hallowed Wings and Plume

Legends0:
The callout may be rather lengthy.

Lef/Right, Near/Far from Hraesvelgr (Tankbusters)


Legends0:
Hmmm... this will need a Cauterize check from Nidhogg to get quadrant.

PR #4501, opened 2022-05-19
i18n: DSR update Japanese & Korean

kshman:


ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -322,25 +325,29 @@ const triggerSet: TriggerSet<Data> = {
           en: 'Red Circle',
           de: 'Roter Kreis',
           fr: 'Cercle rouge',
-          ko: '빨강 원형징',
+          ja: '赤 ○',
```
MaikoTan, line 328: Some TTS engine didn't recognize these symbols, i think it would be better just マル, 三角 and 四角.
Or add a extra `tts` entry in those trigger to make them consistent.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -322,25 +325,29 @@ const triggerSet: TriggerSet<Data> = {
...
+          ko: '빨강 ○',
```
Bing-su, line 329: 이 부분 번역은 제가 했는데, MaikoTan님 말대로 TTS를 고려해서 한국어로 쓰여졌습니다. 괜찮다면 다시 고쳐도 될까요?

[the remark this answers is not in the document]
pyeonch: That symbols doesn't work on korean TTS, and also font is too small to read.
I think need to revert this, @quisquous @kshman

pyeonch: 이부분 번역 빙수님말씀대로 TTS고려해서 한글로 번역했었습니다. 실제로 트라이할때 오늘 써보니 가독성도 좋지 않아보였으며 en de fr의 번역과도 다른 방향이라 원래대로 되돌리는 게 좋을 것 같습니다. 

image

kshman:
고쳐서 PR하시는 거는 이견이 없습니다. 
다만,  그런 이유로 고치려면,  tts 태그를 쓰는게 좋을것 같습니다.

image
이 케이스의 경우 일본어 haruka는 그다지 문제가 없는데, 한국어 heami는 '='를 읽습니다. 이런 문제도 있어요.

pyeonch:
https://github.com/quisquous/cactbot/blob/0df2bb29a48b4985b81aac9dab3d577b0a70674f/ui/raidboss/popup-text.ts#L1363
말씀하신 =>는 그리고로 대체됩니다

MaikoTan:
> https://github.com/quisquous/cactbot/blob/0df2bb29a48b4985b81aac9dab3d577b0a70674f/ui/raidboss/popup-text.ts#L1363
> 
> 
> 말씀하신 =>는 그리고로 대체됩니다


https://github.com/quisquous/cactbot/blob/0df2bb29a48b4985b81aac9dab3d577b0a70674f/ui/raidboss/popup-text.ts#L1361
Looks like japanese translate '=>' as 'や' which is not consistent, either. 
Suggest to change to 'そして'

> 그리고 제가 쓴 번역 tts유저와 문자 유저 둘다 쓰기 좋았다고 생각하는데요.. 뭐가 부족해서 왜바꾸셨는지 그리고 왜 tts태그를 써야하는지 설명 부탁드려도될까요?

I agree to add an extra tts entry for this. I like the symbols on the screen that is easily recognized. But on the other hand, some TTS engines just ignore these symbols, which would cause problems when someone didn't use system TTS.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
MaikoTan: For the "font too small" issue, I think you can open a new PR to add a korean font, just like I do in #2042 or #2056 .

pyeonch:
Well, I don't think symbols is more easily recognize and readable, but it's very subjective topic, let's talk about other one.

At first, to support this symbol, we have to fix korean font issue, TTS doesn't work issue. 

And  only ja,ko, cn(maybe?) use symbol text is not sounds good for consistent, if other en, de, fr texts use symbol too, they have to fix font issue : ○△■ goes small in Arial font. 

I think it's better way for revert this and use old localization (means literally "red circle token" etc, normally recognize, TTS can read), if someone want to use symbol, they can change easily using cactbot user config tab.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
pyeonch: image
Perfectly it helps bigger.. I'll take that into consideration. Thank you!

MaikoTan:
> Well, respect you but I don't think symbols is more easily recognize and readable, but it's very subjective topic,

Yeah, you're right. That's very subjective.

> I think it's better way for revert this and use old localization (means literally "red circle token" etc, normally recognize, TTS can read), if someone want to use symbol, they can change easily using cactbot user config tab.

Yeah, that's a way, too.
But when user change the literal text to symbols in the cactbot config page, while accidently his TTS didn't recognize those symbols, then TTS still would not work in this case.

------

So far here are 3 issues:
- [ ] change the replacement text in ja/ko of the arrow symbol (=>).
https://github.com/quisquous/cactbot/blob/0df2bb29a48b4985b81aac9dab3d577b0a70674f/ui/raidboss/popup-text.ts#L1361-L1362
- [ ] add a korean font.
- [ ] fix the ○△■X symbols, either revert to literal text or add a `tts` entry.

PR #4506, opened 2022-05-20
i18n: DSR fix Japanese & Korean for TTS 

kshman:
Modify texts for support TTS #4501


MaikoTan (approved):
Japanese part looks good to me ! 👍

PR #4507, opened 2022-05-20
raidboss: changed DCoS final boss custom text to preset

BerkeBat:
As far as I can see this is pretty much a clean cut, boss turns to player and does frontal type mechanic. When I first saw the message I was trying to figure out what an eye laser was and where I should be looking.

Only concern I have is that the boss does it multiple times so the previous custom message had 'avoid eye laser**s**' whereas it would now just say 'Away From Front' 🤷‍♂️.


quisquous:
Thanks for the improvement! This is an old trigger.

Issue #4510, opened 2022-05-21
Detailed error indication on `utils/coinach.py`


Bing-su:
## Summary

Detailed error indication on `coinach.py`

-------

last week, I tried to run `utils/gen_*.py`s to make korean data, but failed with ambiguous errors.

like this:
```shell
❯ python util\gen_fisher_data.py
Traceback (most recent call last):
  File "D:\repos\cactbot\util\gen_fisher_data.py", line 347, in <module>
    append_special_place_names(places)
  File "D:\repos\cactbot\util\gen_fisher_data.py", line 266, in append_special_place_names
    coin = coinach.CoinachReader()
  File "D:\repos\cactbot\util\coinach.py", line 57, in __init__
    if not os.path.isfile(os.path.join(self.ffxiv_path, _FFXIV_EXE)):
  File "C:\ProgramData\Miniconda3\lib\ntpath.py", line 78, in join
    path = os.fspath(path)
TypeError: expected str, bytes or os.PathLike object, not NoneType
```

Upon investigation, there are three errors.

1. handling NoneType error
The example above is this. this occurs when ffxiv or coinach path is not in default paths and not given.

2. UnicodeDecodeError
https://github.com/quisquous/cactbot/blob/a558dafcfae452aa04e5b48eb6a3f0ed072d4a40/util/coinach.py#L83-L87

`raw_output` is not always encoded with utf-8. It follows user's windows locale. (in my case, it's `cp949` (korean))

3. CoinachError
In fact, it is natural, but when run this program with Chinese or Korean client to generate global server data, an error occurs,
https://github.com/quisquous/cactbot/blob/a558dafcfae452aa04e5b48eb6a3f0ed072d4a40/util/coinach.py#L93-L94
and it is zero successes.

PR #4511, opened 2022-05-21
util: coinach.py detail error indication

Bing-su:
#4510 

```py
try:
    output = raw_output.decode("utf8")
except UnicodeDecodeError:
    output = str(raw_output).replace("\\r", "\r").replace("\\n", "\n")
```

to infer encoding from bytes, you need a package like [`chardet`](https://github.com/chardet/chardet). I did it this way because it's not important.

### DIFF
--- a/util/coinach.py
+++ b/util/coinach.py
@@ -52,10 +52,12 @@ class CoinachReader:
                     self.ffxiv_path = p
                     break
 
-        if not os.path.isfile(os.path.join(self.coinach_path, _COINACH_EXE)):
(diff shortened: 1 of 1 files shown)


quisquous:
Thanks for the fixes! One possibility here is to use `get_intl_table` from csv_util.py instead of coinach if you wanted to look into changing it to use that instead.  Also, fwiw, when I run `gen_fisher_data.py` I don't get any updated changes.  I think everything is up to date from when I ran it last.

PR #4520, opened 2022-05-25
raidboss: add DSR eyes phase

quisquous:


ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -964,6 +995,109 @@ const triggerSet: TriggerSet<Data> = {
...
+    },
+    {
+      id: 'DSR Spear of the Fury Limit Break',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '62E2', source: 'Ser Zephirin', capture: false }),
+      condition: (data) => data.role === 'tank',
+      // This is a 10 second cast, and (from video) my understanding is to
+      // hit tank LB when the cast bar gets to the "F" in "Fury", which is
+      // roughly 2.8 seconds before it ends.
+      delaySeconds: 10 - 2.8,
```
JLGarber, line 1088: I feel like we had a conversation about some sort of issue with using functions in `delaySeconds` properties, but I can't remember what it was. Is that what you're going for here?

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -964,6 +995,109 @@ const triggerSet: TriggerSet<Data> = {
...
+      suppressSeconds: 1,
+      response: Responses.bigAoe('alert'),
+    },
+    {
+      id: 'DSR Right Eye Reminder',
+      type: 'StartsUsing',
+      // If the Right Eye is dead and the Left Eye gets the aoe off, then the Right Eye
+      // will be revived and you shouldn't forget about it.
+      netRegex: NetRegexes.startsUsing({ id: '68BD', source: 'Left Eye' }),
+      delaySeconds: (_data, matches) => parseFloat(matches.castTime),
```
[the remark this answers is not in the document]
quisquous: I don't think I had really thought it through.  Maybe it was mostly that there's only one startsUsing but there's 10 ability uses.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
quisquous: Yeah, cast times are 0.3 seconds shorter in the log lines than they are in practice.  You can see that if it says 2.7 cast time, it'll be exactly 3 seconds between cast and ability.

JLGarber: I guess I never looked closely enough for that. Makes sense then.

Legends0:
May need a check that this is in nidhogg phase as the nidhogg2 phase has eyes as well. For some events though, if you were to get to them I think it's a wipe anyways.

quisquous:
> May need a check that this is in nidhogg phase as the nidhogg2 phase has eyes as well. For some events though, if you were to get to them I think it's a wipe anyways.

Yeah, I think probably we should take another pass through what needs to reset during a phase change in case folks want to do some ridiculous loop for fun.

PR #4522, opened 2022-05-26
util: Move + centralize core date/string handling 

JLGarber:
(*Ping @valarnin, since you probably have the best idea of whether anything I've touched here breaks things.*)

Mostly self-explanatory. I changed a couple of date/time function names to make What They Do slightly easier to follow, but apart from that it's largely simple movement. I'm not strongly attached to the choices I made here, so please offer all the criticisms. In particular, I'm not really sure about all the names I assigned here. (Should I have/have not have aliased the imports? Are some of the function names too long? etc)

(It's noted in-line, but I absolutely do not intend `getTZOffsetFromLogLine` to permanently use naive string parsing to extract a timestamp. Once we get the line parser disentangled from the emulator I'll gladly swap over to using that, but I don't know the emulator well enough to attempt that in this submission.)

### DIFF
--- a/resources/datetime.ts
+++ b/resources/datetime.ts
@@ -1,0 +1,82 @@
+import StringFuncs from './stringhandlers';
+
+// For performance reasons to prevent re-calculating this every single line,
+// store already calculated values
+const tzOffsetMap: { [key: string]: number } = {};
+
+export default class DateTimeFuncs {
+  static getTimezoneOffsetMillis(timeString: string): number {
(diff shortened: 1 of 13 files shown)


valarnin:
This looks good at first glance. I'll see if I can pull it down and do some testing.

valarnin:
Pulled down this branch and tested, works fine and handles the edge cases I remember off the top of my head.

PR #4523, opened 2022-05-26
lint: Handle nullable empty string checks in misc files

valarnin:
This feels like it's right on the edge of too much for one commit, but it lines up with what we previously discussed?


quisquous:
> This feels like it's right on the edge of too much for one commit

Files Changed: 92 ☠️ 

I think we have very different definitions of what's too much for one commit.  I'll try to work through this and hopefully github will save my work.

valarnin:
If it's too many then I'm fine breaking it up. I never really look at the count of files, rather I look at the count of lines changed.

quisquous:
Ok, I'm at like 21/92 files and it's going slowly because each callsite requires some digesting.  I'll try to get through the rest soon.

util/logtools/make_timeline.ts
```diff
@@ -357,7 +357,7 @@ const assembleTimelineStrings = (
   let lastAbilityTime = (start.getTime());
   let timelinePosition = 0;
   let lastEntry: TimelineEntry = { time: lastAbilityTime.toString(), lineType: 'None' };
-  if (fight.sealName) {
+  if (fight.sealName !== undefined && fight.sealName.length > 0) {
```
quisquous, line 360:
```suggestion
  if (fight.sealName !== undefined) {
```

`sealName` is already checked for empty in encounter tools.  If set, then it's either valid or undefined.

util/logtools/make_timeline.ts
```diff
@@ -368,11 +368,11 @@ const assembleTimelineStrings = (
   // If the user entered phase information,
   // process it and store it off.
   const phases: { [name: string]: number } = {};
-  if (args.phase) {
+  if (args.phase !== null && args.phase.length > 0) {
```
quisquous, line 371:
```suggestion
  if (args.phase !== null) {
```

Length check is unnecessary here as the for loop will do that below.

util/find_missing_translations.ts
```diff
@@ -102,7 +102,7 @@ const parseJavascriptFile = (file: string, inputLocales: Lang[]) => {
 
     // If we're not inside an object, keep looking for the start of one.
     const openMatchValue = openMatch?.[1];
-    if (!openMatch || !openMatchValue)
+    if (!openMatch || openMatchValue === undefined || openMatchValue.length === 0)
```
quisquous, line 105:
```suggestion
    if (!openMatch || openMatchValue === undefined)
```

I think this is a bug that if you have a curly brace with no whitespace before it, we wouldn't find missing translations in that block.  I think we shouldn't early out here, but this is a deliberate change over what was there before.

util/find_missing_translations.ts
```diff
@@ -116,7 +116,7 @@ const parseJavascriptFile = (file: string, inputLocales: Lang[]) => {
 
         const openStr = openMatch[2];
 
-        if (!openStr)
+        if (openStr === undefined || openStr.length === 0)
```
quisquous, line 119:
```suggestion
        if (openStr === undefined)
```

This match is always of at least length 1 because it includes the curly brace.

util/find_missing_translations.ts
```diff
@@ -162,7 +162,7 @@ const findMissingTranslations = async (filter: string, locale: Lang): Promise<vo
         let str = file;
         if (line)
           str += `:${line}`;
-        if (label)
+        if (label !== undefined && label.length > 0)
```
quisquous, line 165:
```suggestion
        if (label !== undefined)
```

I don't think any callers of this special case empty string, so no need to handle it specially.

util/gen_coverage_report.ts
```diff
@@ -79,7 +79,7 @@ const processRaidbossCoverage = async (manifest: string, coverage: Coverage) =>
     const triggerSet = (await import(triggerFileName)).default as LooseTriggerSet;
 
     let timelineContents: string | undefined = undefined;
-    if (triggerSet.timelineFile) {
+    if (triggerSet.timelineFile !== undefined && triggerSet.timelineFile.length > 0) {
```
quisquous, line 82:
```suggestion
    if (triggerSet.timelineFile !== undefined) {
```

I don't think any trigger set special cases the empty string here, and I think we shouldn't handle it specially.

util/gen_log_guide.ts
```diff
@@ -500,7 +500,7 @@ const config: markdownMagic.Configuration = {
       const structureLogLine = ParseLine.parse(logRepo, structureNetworkArray.join('|'));
       let structureLog = structureLogLine?.convertedLine;
 
-      if (!structureLog)
+      if (structureLog === undefined || structureLog.length === 0)
```
quisquous, line 503:
```suggestion
      if (structureLog === undefined)
```

I think this is only intended to be an undefined check for TypeScript and I don't think `convertedLine` will ever be empty.

util/gen_log_guide.ts
```diff
@@ -544,7 +544,7 @@ ${structureLog}
 ${translate(language, titles.networkLogLineRegexes)}
 ${regexes.network}
 `;
-        if (regexes.logLine) {
+        if (regexes.logLine !== undefined && regexes.logLine.length > 0) {
```
quisquous, line 547:
```suggestion
        if (regexes.logLine !== undefined) {
```

This is never empty in the structure above.

util/logtools/encounter_tools.ts
```diff
@@ -158,7 +158,7 @@ export class EncounterFinder {
       // Therefore we can safely initialize everything.
       if (this.currentFight.startTime)
         this.onEndFight(line, cZ, 'Zone Change');
-      if (this.currentZone.zoneName)
+      if (this.currentZone.zoneName !== undefined && this.currentZone.zoneName.length > 0)
```
quisquous, line 161: I don't think we should special case `zoneName`.  It comes from the log line.

```suggestion
      if (this.currentZone.zoneName !== undefined)
```

util/logtools/encounter_tools.ts
```diff
@@ -172,7 +172,11 @@ export class EncounterFinder {
     }
 
     // If no zone change is found, we next verify that we are inside a combat zone.
-    if (this.skipZone() || !this.currentZone.zoneName)
+    if (
+      this.skipZone() ||
+      this.currentZone.zoneName === undefined ||
+      this.currentZone.zoneName.length === 0
```
quisquous, line 178:
```suggestion
```

util/logtools/encounter_tools.ts
```diff
@@ -358,7 +362,7 @@ class TLFuncs {
     return str.split(' ').map((str) => {
       const cap = str[0]?.toUpperCase();
       const lower = str.slice(1);
-      if (cap)
+      if (cap !== undefined && cap.length > 0)
```
quisquous, line 365:
```suggestion
      if (cap !== undefined)
```

Unnecessary, I think?

util/logtools/encounter_tools.ts
```diff
@@ -384,10 +388,10 @@ class TLFuncs {
     const dateStr = TLFuncs.dayFromDate(fightOrZone.startTime).replace(/-/g, '');
     const timeStr = TLFuncs.timeFromDate(fightOrZone.startTime).replace(/:/g, '');
     const duration = TLFuncs.durationFromDates(fightOrZone.startTime, fightOrZone.endTime);
-    let seal;
+    let seal: string | undefined;
     if ('sealName' in fightOrZone)
       seal = fightOrZone['sealName'];
-    if (seal)
+    if (seal !== undefined && seal.length > 0)
```
quisquous, line 394: The `sealName` comes from the match groups on the `will be sealed` log line.  I think we don't need to treat empty as a special case here.

```suggestion
    if (seal !== undefined)
```

util/logtools/encounter_tools.ts
```diff
@@ -445,15 +449,13 @@ class TLFuncs {
       const startTime = fight.startTime ? TLFuncs.timeFromDate(fight.startTime) : 'Unknown Start';
       const fightDuration = TLFuncs.durationFromDates(fight.startTime, fight.endTime) ?? 'Unknown Duration';
       let fightName = 'Unknown Encounter';
-      if (fight.sealName)
+      if (fight.sealName !== undefined && fight.sealName.length > 0)
```
quisquous, line 452:
```suggestion
      if (fight.sealName !== undefined)
```

util/logtools/encounter_tools.ts
```diff
@@ -445,15 +449,13 @@ class TLFuncs {
...
         fightName = fight.sealName;
-      else if (fight.fightName)
+      else if (fight.fightName !== undefined && fight.fightName.length > 0)
         fightName = fight.fightName;
 
-      if (!seenSeal && fight.sealName)
-        seenSeal = true;
-       else if (seenSeal && !fight.sealName)
-        seenSeal = false;
+      if (fight.sealName === undefined || fight.sealName.length === 0)
```
quisquous, line 457:
```suggestion
      if (fight.sealName === undefined)
```

util/logtools/encounter_tools.ts
```diff
@@ -445,15 +449,13 @@ class TLFuncs {
       const startTime = fight.startTime ? TLFuncs.timeFromDate(fight.startTime) : 'Unknown Start';
       const fightDuration = TLFuncs.durationFromDates(fight.startTime, fight.endTime) ?? 'Unknown Duration';
       let fightName = 'Unknown Encounter';
-      if (fight.sealName)
+      if (fight.sealName !== undefined && fight.sealName.length > 0)
         fightName = fight.sealName;
-      else if (fight.fightName)
+      else if (fight.fightName !== undefined && fight.fightName.length > 0)
```
quisquous, line 454:
```suggestion
      else if (fight.fightName !== undefined)
```

`fightName` comes from the zone or seal, so shouldn't be empty.

quisquous (commented):
Ok, here's everything but `ui/raidboss/...`.  Sorry this took a while, but I had to look at everything in context.  This is a lot of changes for one patch because each line took a lot of understanding.

I think in general, I'd rather have `undefined` mean "unset" rather than attempt to special case empty strings everywhere.  There are some places where I think it makes sense to check if something is empty, but I think those are fewer.  I think also, I'd rather check once and then have the rest of the pipeline using that value not have to care about empty strings.

Mostly it's just an argument about cognitive overhead when reading the code for me.  Having to care about empty strings in more places means more things to think about and makes code harder to read and understand.

valarnin:
> I think in general, I'd rather have undefined mean "unset" rather than attempt to special case empty strings everywhere. There are some places where I think it makes sense to check if something is empty, but I think those are fewer. I think also, I'd rather check once and then have the rest of the pipeline using that value not have to care about empty strings.

I agree that having to check for empty strings everywhere isn't great, but I think that anything user-specified needs to as a general rule of thumb?

util/logtools/encounter_tools.ts
JLGarber: Realistically, this whole function should *probably* be switched over to use Valarnin's implementation that's living in `resources/string_handlers` now. I'm not sure how I missed doing that during the move.

quisquous:
> > I think in general, I'd rather have undefined mean "unset" rather than attempt to special case empty strings everywhere. There are some places where I think it makes sense to check if something is empty, but I think those are fewer. I think also, I'd rather check once and then have the rest of the pipeline using that value not have to care about empty strings.
> 
> I agree that having to check for empty strings everywhere isn't great, but I think that anything user-specified needs to as a general rule of thumb?

I think it depends on where it's coming from.  It is a url parameter or a text field? Sure, we should empty check.  Is it an option set directly in a user file? Probably no.

JLGarber:
I can't speak to the majority of the changes, but I'll add my agreement  that we don't need to check string length in the log tool scripts. None of that is user-generated apart (potentially) from the output file name, and it's way better to see possible runtime exceptions than to just silently swallow those (decidedly rare) cases where a zero-length whatever comes in from a `NetRegex` match.

quisquous:
Is it possible to split out the `ui/oopsyraidsy/` and `ui/raidboss/` stuff into separate patches? I think that's the bits I didn't look at, and it would make this patch more manageable.

valarnin:
Status update on this PR: I should have the free time to start working on this in a couple days.

I'll take the following actions:

1. Rebase local branch to latest upstream/main, resolve conflcits
2. Split the other changes off into separate PRs as requested
3. Add a third new PR to make the change to linting rules so it can be merged separately after the others have been merged
4. Force push this PR's branch to fix the pulled out PRs and the rebase/merge conflicts
5. Make the suggested changes

util/gen_effect_id.ts
```diff
@@ -74,7 +74,7 @@ const makeEffectMap = (table: Table<'#', 'Name'>) => {
   const map = new Map<string, string>();
   for (const [id, effect] of Object.entries(table)) {
     const rawName = effect['Name'];
-    if (!rawName)
+    if (rawName === undefined || rawName.length === 0)
```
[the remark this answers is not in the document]
valarnin: The check on line 81 doesn't cause issues because it's not a nullable string (`cleanName` returns type `string`, not `string | undefined`).

I think that maybe we need a different eslint rule to cover this case?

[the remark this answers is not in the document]
valarnin: (this suggestion has been implemented but I'm leaving this unresolved to discuss the line 81 question)

valarnin:
All suggestions relevant to this PR specifically have been addressed.

There are suggestions for oopsy code changes that I'll address over on that PR instead. I've left them open here so I can track them easier.

util/gen_effect_id.ts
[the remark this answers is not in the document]
quisquous: Are you suggesting that we have an eslint rule that says "if you want to check for an empty string you need to do it explicitly via `someString === ''` and not say `!someString`?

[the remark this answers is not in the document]
valarnin: You were asking why the check on line 81 wasn't causing issues, and it's because `name` can't be `undefined` or `null`. If you feel like it *should* be throwing a warning or error, that might be the `allowString` parameter here which defaults to `true`:

https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/strict-boolean-expressions.md#allowstring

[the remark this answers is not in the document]
quisquous: Ah, sorry sorry.  Once you explained it a few comments earlier, that made sense to me why it wasn't throwing.  Given the eslint rule that you're enabling now, it might be a good thing to follow up with.  I think it would certainly clarify these checks.

This is a weak "sounds good to enable this rule as a followup" from me, if you feel strongly that it's not a good idea, I'm happy to leave this as-is too.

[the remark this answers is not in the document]
valarnin: I can play around with enabling it after the rest of the changes are merged and see how big the impact is, I think that's probably the best option.

PR #4524, opened 2022-05-26
plugin: update bozja intl opcode for 6.11h

KattTails:
tested and working.

Sorry for forgetting about this! 😭


quisquous:
Thank you! I also always forget about it!

PR #4526, opened 2022-05-26
raidboss: DSR DFG minor improvements

quisquous:
This is a series of opinionated commits to add some minor improvements to #4435.

In particular:
* merges the "dive" callout into the In+Out (or In+Out followup) calls
* uses alert in more places
* more specialized logic for two baiters/placers
* simplifies the position sorting logic

without this PR side 1:
image

with this PR side 1:
image

without this PR 2:
image

with this PR 2:
image

(One other thought is maybe the "1 (all circles)" call should be combined with "Dive (circles) => Out" a second later? Or is it worth having the 1-2 second warning there but a duplicate call? Should there be another out/in reminder after the stack or a dive, e.g. dive => out, out, in vs just dive => out, in (10 seconds later)? Also, the `Dir Group` call is a personal trigger, in case there's confusion.)

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -921,27 +921,27 @@ const triggerSet: TriggerSet<Data> = {
         if (matches.id === '6715') {
           data.diveFromGraceLashGnashKey = 'in';
           if (data.eyeOfTheTyrantCounter === 1 && num === 3)
-            return output.inOutThenBait!({ inout: output.in!() });
+            return output.inOutAndBait!({ inout: output.in!() });
           if (data.eyeOfTheTyrantCounter === 2) {
             if (num === 2 || (num === 1 && data.diveFromGracePreviousPosition[data.me] === 'middle'))
-              return output.inOutThenBait!({ inout: output.in!() });
+              return output.inOutAndBait!({ inout: output.in!() });
           }
           return output.in!();
         }
         data.diveFromGraceLashGnashKey = 'out';
         if (data.eyeOfTheTyrantCounter === 1 && num === 3)
-          return output.inOutThenBait!({ inout: output.out!() });
+          return output.inOutAndBait!({ inout: output.out!() });
         if (data.eyeOfTheTyrantCounter === 2) {
           if (num === 2 || (num === 1 && data.diveFromGracePreviousPosition[data.me] === 'middle'))
-            return output.inOutThenBait!({ inout: output.out!() });
+            return output.inOutAndBait!({ inout: output.out!() });
         }
         return output.out!();
       },
       outputStrings: {
         out: Outputs.out,
         in: Outputs.in,
-        inOutThenBait: {
-          en: '${inout} => Bait',
+        inOutAndBait: {
+          en: '${inout} + Bait',
         },
       },
     },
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -1327,78 +983,388 @@ const triggerSet: TriggerSet<Data> = {
...
+
+        const inout = output[data.diveFromGraceLashGnashKey]!();
+
+        if (data.eyeOfTheTyrantCounter === 1) {
+          // Tower 1 soaks only based on debuffs.
+          if (num === 3) {
+            if (data.diveFromGraceDir[data.me] === 'circle') {
+              if (data.diveFromGraceHasArrow[3])
+                return output.southTower1!({ inout: inout });
+              return output.circleTowers1!({ inout: inout });
```
panicstevenson, line 1176: No way this fits, right?
```suggestion
              return data.diveFromGraceHasArrow[3] ? output.southTower1!({ inout: inout }) : output.circleTowers1!({ inout: inout });
```

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -1327,78 +983,388 @@ const triggerSet: TriggerSet<Data> = {
...
+        upArrowTower3: {
+          en: 'Up Arrow Tower (${inout})',
         },
-        upArrowDive: {
-          en: 'Up Arrow Dive',
+        downArrowTower3: {
+          en: 'Down Arrow Tower (${inout})',
+        },
+        westTower3: {
+          en: 'West Tower (${inout})',
```
panicstevenson, line 1236: Do these have a way to be like:
```suggestion
          en: Outputs.west.en + ' Tower (${inout})',
```

panicstevenson (approved):
I personally like "1 (all circles)" since it helps me grasp more of the situation (like, more than just if I'm a circle/arrow, but that all 1s are circles so a different number has arrows), and I think having the in/out reminders be in more places would also be helpful.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
quisquous: I mean, it maybe does because the line length gets long.  I think I just avoid that when there's parameters because I find the original easier to read than a really long line.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
quisquous: Unfortunately, I don't think this helps anything, because everything still needs a translation.

quisquous:
> I personally like "1 (all circles)" since it helps me grasp more of the situation (like, more than just if I'm a circle/arrow, but that all 1s are circles so a different number has arrows)

Ok.  I thought saying "1 (all circles)" once and then followup just "circles" would be fine, but I've edited this to say "all circles" everywhere.

> and I think having the in/out reminders be in more places would also be helpful.

Yeah, I'll followup with this.  I want to land this first because I think it's good to go.  I fixed a missing param for two towers with arrows and an in/out thing, but it felt really good last night when I tried it.

PR #4535, opened 2022-05-28
i18n: ko translations update

Bing-su:
- zodiark-ex

nobody in Korea goes northwest instead of northeast. 😢

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zodiark-ex.ts
@@ -71,7 +71,7 @@ const triggerSet: TriggerSet<Data> = {
           en: 'Under NW Quetzalcoatl',
           de: 'Unter NW Quetzalcoatl',
           cn: '站在左上 (西北) 鸟',
-          ko: '북동쪽 새 밑으로',
+          ko: '북서쪽 새 밑으로', // 한섭은 보통 북동쪽으로 가지만, 번역 일관성을 위해 이렇게 함
         },
       },
     },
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/06-ew/trial/zodiark-ex.ts
```diff
@@ -71,7 +71,7 @@ const triggerSet: TriggerSet<Data> = {
           en: 'Under NW Quetzalcoatl',
           de: 'Unter NW Quetzalcoatl',
           cn: '站在左上 (西北) 鸟',
-          ko: '북서쪽 새 밑으로',
+          ko: '북동쪽 새 밑으로',
```
quisquous, line 74: In general I think we want the translations to be the same.  I think this one is a little weird because it's prescribing a set position.  Can you leave a comment that Korean folks go a different direction and that this translation is deliberately different?

Bing-su: If consistency in translation is more important, I'll.

quisquous: If everybody goes NE and not NW, I think that's fine to have in Korean.  I just think that it needs a comment so that if anybody looks at this and goes "why is this translation different" then there's an explanation.

Bing-su: it looks good.

quisquous: Yeah, my general rule is that I don't want translations to be different, as if somebody was playing the intl version but had their output text in Korean, I don't want it to tell them something different than what everybody else is doing.  In this case, it doesn't really matter (and folks have plenty of time to figure it out) and so I'm ok with this one case of a disagreement.

PR #4550, opened 2022-05-30
Update p3s Cinderwing correction direction in p3s.js

IcyGrain:
in p3s, most time team are in front of boss, so the Cinderwing correct direction should be left for left/right for right


quisquous:
Sorry, for consistency, if cactbot ever says "go left" it's boss relative, meaning "go to the bosses' left flank".  I know that this can be confusing sometimes, but overall in terms of what folks expect from cactbot, I'd rather have it be consistent than people have to guess from fight to fight what "left" means.

I think if you would like this to be different, then you can change the output text in the cactbot config ui.

IcyGrain:
> 

thanks for your reply, i will cancel the PR

PR #4553, opened 2022-05-30
raidboss/oopsy: Add A3n package

JLGarber:
I've had this sitting around for a while and for some reason I never bothered to finish it up. Most of it is copypasta from A3s, but I made a few modifications to better fit normal mode.


quisquous:
Thank you!

PR #4563, opened 2022-06-06
i18n: DSR japanese translations

resqnet:
Corrected Japanese name


quisquous:
@Akurosia can you look at this (since maybe it's a script issue on your end?)

Issue #4564, opened 2022-06-06
Incorrect call on DSR P6S Cauterize (Dives)


Ryphael:
A couple of friends and I have found out that cauterise on P6 in Dragonsong Reprise during the part where u need to be dived by the opposite dragon that you got tethered from is wrong. It's constantly calling out for nidhogg regardless of whether you're getting hit by Hraes or Nidhogg.

quisquous:
Could you attach a network log here?

Ryphael:
Hmm.. my logs exceed the limit of 25mb, even after splitting it's still too huge to upload

quisquous:
Can you zip it or upload it to some other site and link it here?

Ryphael:
https://www.file.io/zHwi/download/mHmHEbEc9xAi

Done, hopefully this allows you to get it

PR #4566, opened 2022-06-06
raidboss: DSR reverse p6 hallowed wings left/right

quisquous:
Looking at fflogs replays and the log from #4564,
it appears that the "left wing" and "right wing" comments
are backwards.

We could consider saying "north" and "south" here to be
absolute, but I think most folks will be looking at Hraesvelgr
to see the wing and head, so left/right is pretty natural.


quisquous:
cc @Legends0

Legends0:
The comments I intended to write as respective to the boss's left/right and not as the player is looking and for the output to be the opposite. Perhaps I mixed something up.

xpdota:
> The comments I intended to write as respective to the boss's left/right and not as the player is looking and for the output to be the opposite. Perhaps I mixed something up.

Yeah, it's a double negative. Boss's right wing => player's left side bad => player's right side safe. The original comment is correct, the calls just need to be flipped.

PR #4567, opened 2022-06-07
plugin: update bozja intl opcode for 6.15

KattTails:
Untested due to the other networking changes rendering the FFXIV_ACT_Plugin unusable, but the opcode should be correct.

Might need further work based on the extent of the changes required.


KattTails:
Update: I just tested it with the beta plugin and it worked as expected! 🎉

PR #4575, opened 2022-06-08
util: TypeScriptify util/logtools/split_log.ts

quisquous:


util/logtools/split_log.ts
```diff
@@ -1,46 +1,77 @@
...
   console.error(str);
   exitCode = 1;
 };
 
-const writeFile = (outputFile, startLine, endLine) => {
-  return new Promise((resolve, reject) => {
+const writeFile = (
+  outputFile: string,
+  startLine: string | undefined,
+  endLine: string | undefined,
```
JLGarber, line 64: It occurs to me that for the future, we could probably rewrite this to be something like (not production-ready!)
```typescript
const writeFile = (
    outputFile: string,
    collector: EncounterCollector,
    indexList: number[],
): void => {
  const lines: string[] = [];
  for (const fight of collector.fights) {
    if (indexList.includes(collector.fights.indexOf(fight)))
      lines.concat(fight.lines);
  }

  // Process lines as usual
}
```

and then set `processLine(line, true)` in our collector prepass. This would save us an additional file access when doing the actual line processing. (We might also want to add a hardcoded index to each fight/zone in `encounter_tools` as well in that case, although that might be slightly more complex.)

PR #4578, opened 2022-06-08
util/docs: remove adjust_timeline.ts

quisquous:
The Cactbot Highlight timeline adjuster is much more convenient
than using the timeline_adjust script.  Therefore, just remove it
so there's only one supported path here.

Note: the translated TimelineGuide is not updated, and additionally
probably the TimelineGuide should be rewritten as it's more
longwinded than it should be and Cape Westwind doesn't exist anymore.


JLGarber:
Overall I think you're right, but I do have the tiny concern that this basically forces any timeline contributors to use an editing environment that supports this plugin. I don't know whether this will actually affect any current contributors, but it does bear consideration maybe?

quisquous:
No, I think that's a fair point, but I think given how few people there are making timelines, I'd rather have one supported tool than multiple tools (some of which could get out of date).  I think if somebody was like "vim only!" they could probably make their own plugin at that point.

PR #4581, opened 2022-06-09
util: Restore correct date formatting in encounter_tools

JLGarber:
This resolves bugs I introduced when moving functions around in #4522. I'm not sure how I missed seeing the Glaringly Obvious Misformatting, but after removing the unnecessary replacements from `printZone/Fight`, all seems to be back to normal. 

This also fixes a bug with getting milliseconds where the logic was exactly backward from what it should have been, and therefore we were getting nonsense offsets. This somehow didn't show up obviously anywhere.

(Thanks quisquous for the bug report!)


quisquous:
Thank you!!

PR #4585, opened 2022-06-09
util: Add basic FFLogs support to make_timeline.ts

JLGarber:
It works for the More Than One reports I tried it on. Usage is (or should be) identical to the original Python implementation. I haven't tested all the combinations of arguments, so it's entirely possible I left in some invalid combinations.

On the error codes, I tried to keep it to "a missing or invalid encounter is -2, actual script errors are -1". If we should make that more granular, I'm happy to make those changes.

Testing On My Machine seems to be not costly at all, with a solid half-hour debugging session not even reaching 100 V1 API points.

### DIFF
--- a/util/logtools/fflogs.ts
+++ b/util/logtools/fflogs.ts
@@ -3,10 +3,17 @@ import fetch from 'node-fetch';
 // The FFLogs API returns more data than just this,
 // but it's only status information for source and target,
 // and currently nothing in the timeline utilities needs this.
-type ffLogsEventEntry = {
+export type ffLogsEventEntry = {
   timestamp: number;
   type: string;
-  sourceID: number;
+  sourceID?: number;
+  source?: {
+    name: string;
+    id: number;
+    guid: number;
+    type: string;
+  };
+  sourceInstance?: number;
   sourceIsFriendly: boolean;
   targetID: number;
   targetIsFriendly: boolean;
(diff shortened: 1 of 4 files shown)


util/logtools/fflogs.ts
```diff
@@ -18,48 +25,146 @@ type ffLogsEventEntry = {
   fight: number;
 };
 
-type fflogsEventResponse = {
+export type fflogsFightListEntry = {
```
quisquous, line 28: Are these all fields from fflogs itself (with mismatched camel case and snake case?)

JLGarber: Unfortunately yes:
```
{
    "id": 1,
    "boss": 2038,
    "start_time": 21087,
    "end_time": 348888,
    "name": "Rhalgr",
    "zoneID": 1054,
    "zoneName": "Aglaia",
    "size": 24,
    "difficulty": 100,
    "kill": true,
    "partial": 2,
    "inProgress": false,
    "standardComposition": true,
    "hasEcho": false,
    "bossPercentage": 1,
    "fightPercentage": 1,
    "lastPhaseAsAbsoluteIndex": 0,
    "lastPhaseForPercentageDisplay": 0,
    "maps": [
        {
            "mapID": 755,
            "mapName": "Heaven of Lightning",
            "mapFile": "m-z5r1-z5r1.01.jpg"
        }
    ]
}
```

This is the structure for each object in the `events` array that's returned. You can see the full fight response at https://www.fflogs.com:443/v1/report/fights/G48Jt7bcfKVPwyBh?api_key=YOUR KEY HERE

PR #4592, opened 2022-06-14
raidboss: DSR earlier Strength Thordan call

### LINKED ISSUE #4400: [DSR] Finding Thordan faster in DSR Dragon's Rage
For trigger "DSR Dragon\'s Rage", it is looking for Thordan and calling out the location of Thordan, as in this phase of battle we need to find Thordan as fake north. However, "Ser Hermenost" will be casting "Conviction" right opposite Thordan about 2 seconds before Thordan appears. Finding Thordan fast will be really helpful in this phase. 
The example network log line is 
`21|2022-05-05T22:16:03.6900000+08:00|40006AD4|Ser Hermenost|63DE|Conviction|40006AD4|Ser Hermenost|1B|63DE8000|0|0|0|0|0|0|0|0|0|0|0|0|0|0|6920000|6920000|10000|10000|||100.00|77.00|0.00|0.00|6920000|6920000|10000|10000|||100.00|77.00|0.00|0.00|000046BB|0|1|bdd6720fe25a938c`
(I know this is not ACT log which is used to do the match, just for your reference)

quisquous:
Use Ser Adelphel and Ser Janlenoux to find Thordan rather than
waiting for Thordan to appear.  Because this is very close
in time to the "move" trigger it is deliberately silent.

Fixes #4400.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -29,8 +29,6 @@ export interface Data extends RaidbossData {
   adelphelDir?: number;
   brightwingCounter: number;
   spiralThrustSafeZones?: number[];
-  thordanJumpCounter?: number;
-  thordanDir?: number;
   sanctityWardDir?: string;
   thordanMeteorMarkers: string[];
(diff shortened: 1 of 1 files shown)


Legends0:
I tried this out with tts, it's still good.

PR #4593, opened 2022-06-14
raidboss: DSR add no doom trigger for Death

quisquous:
cc @g0shu


g0shu:
Do you think it's worth having different severity for doom/no doom here?

PR #4603, opened 2022-06-22
raidboss: DSR Individual Sacred Sever hits in timeline

Legends0:
This adds the initial hit to the timeline for the one and two sword marker groups during sanctity. This helps with landing a heal between the swords before movement is required to dodge orbs in daylight strategy.


Legends0:
Could potentially add the 3rd and 4rth hit, but I don't think those are necessary and even adding a line for the first hit seems a bit unnecessary as the Server is 0.1s within the hit, but I did it for completeness of the two first hits.

Legends0:
I need to make an edit, the Server is the one that does the hit.

quisquous:
The reason the duration was there was because it was covering multiple hits.  Do you think it'd make sense to just list the 4 hits out individually at this point?

Legends0:
> The reason the duration was there was because it was covering multiple hits. Do you think it'd make sense to just list the 4 hits out individually at this point?

I have added the other two hits as it is easier to see it coming than interpreting the two duration bars. This could also help for reaction if something needs to be swift cast or timed before the 2nd hit.

PR #4608, opened 2022-06-25
lint: Handle nullable empty string checks in Oopsy

valarnin:
This is being split off from #4523

### DIFF
--- a/ui/oopsyraidsy/oopsy_live_list.ts
+++ b/ui/oopsyraidsy/oopsy_live_list.ts
@@ -157,7 +157,7 @@ export class DeathReportLive {
 
     const damageElem = document.createElement('div');
     damageElem.classList.add('death-row-amount');
-    if (amountClass)
+    if (amountClass !== undefined && amountClass.length > 0)
       damageElem.classList.add(amountClass);
     if (amount !== undefined)
       damageElem.innerText = amount;
@@ -272,9 +272,11 @@ export class OopsyLiveList implements MistakeObserver {
 
     const iconClass = m.type;
     const blame = m.name ?? m.blame;
-    const blameText = blame ? ShortNamify(blame, this.options.PlayerNicks) + ': ' : '';
+    const blameText = (blame !== undefined && blame.length > 0)
+      ? ShortNamify(blame, this.options.PlayerNicks) + ': '
+      : '';
     const translatedText = Translate(this.options.DisplayLanguage, m.text);
-    if (!translatedText)
+    if (translatedText === undefined || translatedText.length === 0)
       return;
 
     const time = GetFormattedTime(this.baseTime, timestamp);
@@ -336,7 +338,9 @@ export class OopsyLiveList implements MistakeObserver {
     div.addEventListener('click', () => {
       const mistakeText = div.childNodes[1]?.textContent ?? '';
       const mistakeTime = div.childNodes[2]?.textContent;
-      const str = mistakeTime ? `[${mistakeTime}] ${mistakeText}` : mistakeText;
+      const str = (mistakeTime !== undefined && mistakeTime !== null && mistakeTime.length > 0)
+        ? `[${mistakeTime}] ${mistakeText}`
+        : mistakeText;
       const el = document.createElement('textarea');
       el.value = str;
       document.body.appendChild(el);
--- a/ui/oopsyraidsy/oopsy_summary_list.ts
+++ b/ui/oopsyraidsy/oopsy_summary_list.ts
@@ -82,7 +82,7 @@ export class OopsySummaryTable implements MistakeObserver {
 
   OnMistakeObj(m: OopsyMistake): void {
     const longName = m.name ?? m.blame;
-    if (!longName)
+    if (longName === undefined || longName.length === 0)
       return;
     const name = ShortNamify(longName, this.options.PlayerNicks);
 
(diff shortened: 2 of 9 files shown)


ui/oopsyraidsy/player_state_tracker.ts
```diff
@@ -408,7 +412,7 @@ export class PlayerStateTracker {
     this.collector.OnMistakeObj(timestamp, mistake);
 
     const targetId = mistake.reportId;
-    if (!targetId || !IsPlayerId(targetId))
+    if (targetId === undefined || targetId.length === 0 || !IsPlayerId(targetId))
```
quisquous, line 415:
```suggestion
    if (targetId === undefined || !IsPlayerId(targetId))
```

`IsPlayerId` covers this.

ui/oopsyraidsy/player_state_tracker.ts
```diff
@@ -486,7 +490,7 @@ export class PlayerStateTracker {
 
   OnHoTDoT(_line: string, splitLine: string[]): void {
     const targetId = splitLine[logDefinitions.NetworkDoT.fields.id];
-    if (!targetId || !this.IsInParty(targetId))
+    if (targetId === undefined || targetId.length === 0 || !this.IsInParty(targetId))
```
quisquous, line 493:
```suggestion
    if (targetId === undefined || !this.IsInParty(targetId))
```

`IsInParty` covers this one too.

ui/oopsyraidsy/player_state_tracker.ts
```diff
@@ -549,7 +553,7 @@ export class PlayerStateTracker {
 
     const missedNames = missedIds.map((id) => {
       const name = this.partyTracker.nameFromId(id);
-      if (!name) {
+      if (name === undefined || name.length === 0) {
```
quisquous, line 556:
```suggestion
      if (name === undefined) {
```

I think we shouldn't bother special-casing empty strings coming from added combatant lines.

ui/oopsyraidsy/oopsy_live_list.ts
```diff
@@ -272,9 +272,11 @@ export class OopsyLiveList implements MistakeObserver {
 
     const iconClass = m.type;
     const blame = m.name ?? m.blame;
-    const blameText = blame ? ShortNamify(blame, this.options.PlayerNicks) + ': ' : '';
+    const blameText = (blame !== undefined && blame.length > 0)
```
quisquous, line 275: I realize this is not what's there before, but I think it's weird logic to skip the name because the name is blank.  What about the blank name turning into `???` and undefined using `''` instead of `': '`?

ui/oopsyraidsy/oopsy_live_list.ts
```diff
@@ -272,9 +272,11 @@ export class OopsyLiveList implements MistakeObserver {
...
+      ? ShortNamify(blame, this.options.PlayerNicks) + ': '
+      : '';
     const translatedText = Translate(this.options.DisplayLanguage, m.text);
-    if (!translatedText)
+    if (translatedText === undefined || translatedText.length === 0)
```
quisquous, line 279:
```suggestion
    if (translatedText === undefined)
```

The text is coming from the the trigger itself, and I think if it's weirdly blank we should still show a mistake line rather than special casing.

ui/oopsyraidsy/oopsy_summary_list.ts
```diff
@@ -82,7 +82,7 @@ export class OopsySummaryTable implements MistakeObserver {
 
   OnMistakeObj(m: OopsyMistake): void {
     const longName = m.name ?? m.blame;
-    if (!longName)
+    if (longName === undefined || longName.length === 0)
```
quisquous, line 85: Same comment as in the other file, but I think it'd make more sense to just have it be `???` here rather than skipping?

ui/oopsyraidsy/oopsy_summary_list.ts
```diff
@@ -194,7 +194,7 @@ export class OopsySummaryList implements MistakeObserver {
     pullDiv.innerText = `Pull ${this.pullIdx}`;
     headerDiv.appendChild(pullDiv);
     const zoneDiv = document.createElement('div');
-    if (this.zoneName)
+    if (this.zoneName !== undefined && this.zoneName.length > 0)
```
quisquous, line 197:
```suggestion
    if (this.zoneName !== undefined)
```

Arguably the zone name should never be empty?

ui/oopsyraidsy/oopsy_summary_list.ts
```diff
@@ -216,9 +216,11 @@ export class OopsySummaryList implements MistakeObserver {
   OnMistakeObj(timestamp: number, m: OopsyMistake): void {
     const iconClass = m.type;
     const blame = m.name ?? m.blame;
-    const blameText = blame ? `${ShortNamify(blame, this.options.PlayerNicks)}: ` : '';
+    const blameText = (blame !== undefined && blame.length > 0)
```
quisquous, line 219: Same comments here as on the other naming questions.

ui/oopsyraidsy/oopsy_summary_list.ts
```diff
@@ -216,9 +216,11 @@ export class OopsySummaryList implements MistakeObserver {
...
+      ? `${ShortNamify(blame, this.options.PlayerNicks)}: `
+      : '';
     const text = Translate(this.options.DisplayLanguage, m.text);
-    if (!text)
+    if (text === undefined || text.length === 0)
```
quisquous, line 223:
```suggestion
    if (text === undefined)
```

`text` is coming from triggers and should never be an empty string.

ui/oopsyraidsy/oopsy_summary_list.ts
```diff
@@ -277,7 +279,7 @@ export class OopsySummaryList implements MistakeObserver {
 
       const damageElem = document.createElement('div');
       damageElem.classList.add('death-row-amount');
-      if (event.amountClass)
+      if (event.amountClass !== undefined && event.amountClass.length > 0)
```
quisquous, line 282:
```suggestion
      if (event.amountClass !== undefined)
```

`amountClass` here is all internal and so shouldn't ever be empty.

PR #4609, opened 2022-06-25
lint: Handle nullable empty string checks in Raidboss

valarnin:
This is being split off from #4523

### DIFF
--- a/ui/raidboss/browser_tts_engine.ts
+++ b/ui/raidboss/browser_tts_engine.ts
@@ -1,4 +1,4 @@
-import { Lang } from '../../resources/languages';
+import { isLang, Lang } from '../../resources/languages';
 
 class TTSItem {
   readonly text: string;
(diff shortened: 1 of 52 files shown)


valarnin:
The failing test is because of the following lines in the a7s timeline:

https://github.com/quisquous/cactbot/blob/a081e237cda8de6ab112d15c98073af9975edade/ui/raidboss/data/03-hw/raid/a7s.txt#L87-L96

I guess the question is, should they be counted as an error?

Additionally these lines, including the first one since it has a blank name, will also cause the test to fail.

https://github.com/quisquous/cactbot/blob/a081e237cda8de6ab112d15c98073af9975edade/ui/raidboss/data/03-hw/raid/a7s.txt#L172-L181

quisquous:
> I guess the question is, should they be counted as an error?

I don't think this is an error.

PR #4611, opened 2022-06-25
raidboss: add x2 indiactor to double Mercy Swords

jacob-keller:
When Trinity Seeker does the 4 Mercy Swords mechanic, sometimes the
callout is something like "Right > Left". This callout doesn't clearly
indicate that the player should wait two swords before moving. Improve
this by adding a x2 indicator.

We don't need a x4 indicator for the rare cases where the safe spot is
always the same, since the player doesn't need to move once in position.

### DIFF
--- a/ui/raidboss/data/05-shb/eureka/delubrum_reginae.ts
+++ b/ui/raidboss/data/05-shb/eureka/delubrum_reginae.ts
@@ -192,7 +192,7 @@ const triggerSet: TriggerSet<Data> = {
           ko: '왼쪽',
         },
         double: {
-          en: '${dir1} > ${dir2}',
+          en: '${dir1} x2 > ${dir2} x2',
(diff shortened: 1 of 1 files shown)


quisquous:
Hmm, I think this is a case of "you need to understand how the mechanic works, and "right 2x" makes it harder to read (when you do) in my opinion.  I know it's subjective, but I think the callout should reflect what a raid caller would say here, and I think a raid caller would say "right > left".

I think if somebody doesn't like this, they could change it to "Right > Right > Left > Left" or "Right 2x > Left 2x" in the cactbot config ui?

jacob-keller:
Well a shot caller is likely to time the "right" and "left" so that you don't run too early but it's a bit hard to time when to put that on screen.

I am fine just modifying the text for my own use if this isn't useful to everyone.

Thanks for the discussion!

PR #4612, opened 2022-06-26
raidemulator: Change emulator display to hide pre-pull time

### LINKED ISSUE #4548: raidemulator: Adjust encounter duration display to consider offset in the encounter listing
## Summary
raidemulator encounter listing
For encounters that have a very large offset (minutes), it becomes difficult to find the encounters I am looking for at first glance. One of the reasons I use raidemulator is for data collection/analysis of certain mechanics with user triggers, so it is important for me to find pulls that reached a certain point (phase). Loading long encounters seems to take at least a few seconds in DSR, so I want to minimize how many unnecessary checks I do to get the actual duration of the pull. Because the "duration" field in the UI doesn't consider offset, it is basically impossible to see the length of a pull without at least clicking on the encounter row and doing some mental math to manually subtract the offset. Alternatively, I could just upload the log to fflogs to see the start time of all the relevant pulls, but this is more hoops than I feel should be necessary.

I didn't include it in my image above, but if the current behavior is important, then another idea would be to add another field/line for Pull Duration. I am not familiar enough with how others use raidemulator to understand the need to know the duration that is currently displayed instead of subtracting offset first.

valarnin:
Assuming I understood the issue correctly, this should fix #4548

Could use some additional testing by another pair of eyes, as stated elsewhere this is a bit of a complicated mess to adjust.

Also this PR fixes a bug I noticed where triggers with a delaySeconds parameter were getting additional time added to their resolvedOffset due to leftover code from a previous way of calculating the resolvedOffset value.

### DIFF
--- a/ui/raidboss/emulator/data/AnalyzedEncounter.ts
+++ b/ui/raidboss/emulator/data/AnalyzedEncounter.ts
@@ -213,14 +213,11 @@ export default class AnalyzedEncounter extends EventBus {
       if (!perspective || !triggerHelper)
         throw new UnreachableCode();
 
-      const delay = currentTriggerStatus.delay ?? 0;
-
       perspective.triggers.push({
         triggerHelper: triggerHelper,
         status: currentTriggerStatus,
         logLine: log,
-        resolvedOffset: (log.timestamp - this.encounter.startTimestamp) +
-          (delay * 1000),
+        resolvedOffset: (log.timestamp - this.encounter.startTimestamp),
       });
     };
     popupText.triggerResolvers = [];
--- a/ui/raidboss/emulator/ui/EncounterTab.ts
+++ b/ui/raidboss/emulator/ui/EncounterTab.ts
@@ -55,8 +55,8 @@ export default class EncounterTab extends EventBus {
         const zone = enc.zoneName;
         // ?? operator here to account for old encounters that don't have the property
         const encDate = DTFuncs.timeStringToDateString(enc.start, enc.tzOffsetMillis ?? 0);
-        const encTime = DTFuncs.timeToTimeString(enc.start, enc.tzOffsetMillis ?? 0);
-        const encDuration = DTFuncs.msToDuration(enc.duration);
+        const encTime = DTFuncs.timeToTimeString(enc.start + enc.offset, enc.tzOffsetMillis ?? 0);
+        const encDuration = DTFuncs.msToDuration(enc.duration - enc.offset);
         const zoneObj = this.encounters[zone] = this.encounters[zone] || {};
         const dateObj = zoneObj[encDate] = zoneObj[encDate] || [];
         dateObj.push({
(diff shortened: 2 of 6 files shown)


valarnin:
Man, PR title naming feels like playing code golf sometimes with the 60-character limit.

ui/raidboss/emulator/data/AnalyzedEncounter.ts
```diff
@@ -213,14 +213,11 @@ export default class AnalyzedEncounter extends EventBus {
...
 
-      const delay = currentTriggerStatus.delay ?? 0;
-
       perspective.triggers.push({
         triggerHelper: triggerHelper,
         status: currentTriggerStatus,
         logLine: log,
-        resolvedOffset: (log.timestamp - this.encounter.startTimestamp) +
-          (delay * 1000),
+        resolvedOffset: (log.timestamp - this.encounter.startTimestamp),
```
quisquous, line 220: Sorry for the confusion on my part, but I don't think I understand this change.  I could understand adjusting for the encounter's offset, but I don't understand how you can take out the individual trigger's delay.  Could you explain the change to me?

ui/raidboss/emulator/ui/EncounterTab.ts
```diff
@@ -55,8 +55,8 @@ export default class EncounterTab extends EventBus {
         const zone = enc.zoneName;
         // ?? operator here to account for old encounters that don't have the property
         const encDate = DTFuncs.timeStringToDateString(enc.start, enc.tzOffsetMillis ?? 0);
-        const encTime = DTFuncs.timeToTimeString(enc.start, enc.tzOffsetMillis ?? 0);
-        const encDuration = DTFuncs.msToDuration(enc.duration);
+        const encTime = DTFuncs.timeToTimeString(enc.start + enc.offset, enc.tzOffsetMillis ?? 0);
+        const encDuration = DTFuncs.msToDuration(enc.duration - enc.offset);
```
quisquous, line 59: Do you think it'd make sense to have a `PersistorEncounter` getter property for `trimmedDuration` and `trimmedStart` (or some other name)?

quisquous (commented):
Sorry to take so long to look at this.  I'll give this a spin myself and see if I have any other feedback.  Here's my only comments from looking at the code.

ui/raidboss/emulator/data/AnalyzedEncounter.ts
quisquous: Oh sorry, I can't read.  I see your comment, re: `Also this PR fixes a bug I noticed where triggers with a delaySeconds parameter were getting additional time added to their resolvedOffset due to leftover code from a previous way of calculating the resolvedOffset value.`

ui/raidboss/emulator/ui/EncounterTab.ts
quisquous: I'll just merge this patch for now and this can be a followup if I or you are interested in it.

PR #4616, opened 2022-06-28
raidboss: DSR add Great Wyrmsbreath tank triggers

quisquous:


quisquous:
This looks good in the emulator, so I'll probably land it unless there's any comments.

PR #4624, opened 2022-07-01
Un-commenting durations in DSR timeline

xpdota:
It looks like some syncs were commented out, but the durations got accidentally commented out along with them.


quisquous:
Thanks!

PR #4643, opened 2022-07-13
raidboss: A10n raid package

JLGarber:
I needed a break from high-level content, so I sat down and picked at this one over the weekend. It's mostly self-explanatory. I don't know whether there's a HP% push before the intermission, but the sync windows should handle that.

I did some naming nonsense on the Single Charge abilities since the damage actually resolves on the Conklops/Whooshdrops ability, and nobody is going to remember what that ability *is* from the name anyway.

I haven't tested whether the Gobbieboom add explosion is a genuine enrage or just a healer mechanic, but given that I've literally never seen it even when it was current content, I don't think it matters.

The log doesn't seem to have information on players activating the traps.


quisquous:
Thanks for going back and doing this! Looks great to me.

Issue #4530, opened 2022-05-27
oopsy: data.party.isTank no longer works after #4499


quisquous:
#4499 tracks who is in the party/alliance but I forgot that I didn't do anything to track which role was which role.  Just filing this to remember to fix this up at some point.

This is slightly irritating, but only applies to p4n, TEA, and (soon) DSR oopsy.  I think rather than implementing a "detect players from abilities" sort of line, I'd rather hold off in case we can come up with a cleaner way to implement this (e.g. player event turns into a new log line or something), which would clean up all the code.

PR #4649, opened 2022-07-17
oopsy: Sephirot EX

JLGarber:
Largely self-explanatory. I could use criticism on the Force Against mistake text, since even after I cut it down to what it is now, it's still really long.

I'm not sure what we want to do with the non-tank + no-debuff tower trigger, since #4530 is still floating around. I would be okay removing it, but it's going to be moderately important once this is released for Unreal.

### DIFF
--- a/ui/oopsyraidsy/data/03-hw/trial/sephirot-ex.ts
+++ b/ui/oopsyraidsy/data/03-hw/trial/sephirot-ex.ts
@@ -1,0 +1,152 @@
+import NetRegexes from '../../../../../resources/netregexes';
+import ZoneId from '../../../../../resources/zone_id';
+import { OopsyData } from '../../../../../types/data';
+import { OopsyTriggerSet } from '../../../../../types/oopsy';
+
+export interface Data extends OopsyData {
+  force?: { [name: string]: string };
+}
+
+const triggerSet: OopsyTriggerSet<Data> = {
+  zoneId: ZoneId.ContainmentBayS1T7Extreme,
+  damageWarn: {
+    'SephirotEX Yesod': '157E', // Snapshot floor spikes
+    'SephirotEX Ain': '1569', // Half-arena baited frontal
+    'SephirotEX Ein Sof': '156F', // Expanding green puddles
+    'SephirotEX Fiendish Wail': '1577', // Raidwide if tower is missed
+  },
+  damageFail: {
+    'SephirotEX Pillar Of Mercy': '1581', // Standing in the blue puddles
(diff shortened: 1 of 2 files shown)


ui/oopsyraidsy/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -0,0 +1,152 @@
...
+      id: 'SephirotEX Spirit',
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: '157B', source: 'Sephirot' }),
+      condition: (data, matches) => data?.force?.[matches.target] === '3ED',
+      mistake: (_data, matches) => {
+        return {
+          type: 'warn',
+          blame: matches.target,
+          text: {
+            en: 'Took green; orange debuff',
```
quisquous, line 82: What about just calling all of these custom warnings just the ability name? I think the fact that it gets listed as a mistake means "you shouldn't have done this" and matches what oopsy does elsewhere.

ui/oopsyraidsy/data/03-hw/trial/sephirot-ex.ts
```diff
@@ -0,0 +1,152 @@
...
+          },
+        };
+      },
+    },
+    {
+      // Only tanks should take towers without a Force debuff.
+      id: 'SephirotEX Fiendish Wail Non-Tank',
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: '1576', source: 'Sephirot' }),
+      condition: (data, matches) => !data.party.isTank(matches.target) && data?.force?.[matches.target] === undefined,
```
quisquous, line 122: I think for now just leave this as-is and assume that we'll (I'll?) fix this sometime soon.

Also, how should this handle BLU?

quisquous (commented):
Sorry for the extremely belated review here.  Thank you for taking a look at this!

ui/oopsyraidsy/data/03-hw/trial/sephirot-ex.ts
JLGarber: Sure, that makes sense. My concern was that the ability names are kinda ambiguous (Life Force/Spirit), but I guess that's true (for non-Hebrew speakers) of basically all the Kabbalistic terms in the encounter anyway.

ui/oopsyraidsy/data/03-hw/trial/sephirot-ex.ts
JLGarber: Fair point, we can just not call Blue Mages if they take towers without a debuff. Fortunately that's the only role-based trigger we'll need for this encounter!

PR #4655, opened 2022-07-21
oopsy: add Lost City of Amdapor Hard

quisquous:


JLGarber (approved):
I didn't get to test this one live due to time constraints, but the boss abilities all look good from a brief look.

PR #4660, opened 2022-07-28
raidboss: DSR add Mortal Vow callout

Legends0:
Adds a callout ~~5~~ **3.7** seconds before Mortal Vow occurs.
Will call out Soon (Spread), who has it when it is time to pass, and if it is on you when it is time to pass.

I think 5 seconds is enough time, perhaps it could be shorter. I find it helpful to have a callout even if healer as it helps as a reminder to avoid whoever has it when the time is coming since it can be hectic learning this part of the fight so late.

**EDIT:** Changed to 3.7 as last call should happen after touchdown to avoid triggering early movement into bosses.


Legends0:
The last mortal vow pass happens 4.3 seconds after touchdown... might 5 seconds be too early for that call.

Legends0:
Maybe the on you could be changed to just Mortal Vow pass?

PR #4661, opened 2022-07-30
raidboss: fix some ew hunts triggers

Bing-su:
## Yilan Mini Light
[yilan 2022-07-30 11-24-21.webm](https://user-images.githubusercontent.com/37621276/181867570-f30920a8-fb59-453c-aa1f-718fcb0d1054.webm)

Not all players gain directional march effect. so I think it would be better if there was an alarm to go outside

## Fan Ail Divebomb
[Fan Ail 2022-07-30 11-17-32.webm](https://user-images.githubusercontent.com/37621276/181867682-0a712161-68f8-4e4a-b4af-9a63c93951a9.webm)

In fact, the direction of Fan Ail is fixed when the divebomb marker appears, so you only need to avoid front of that.

### DIFF
--- a/ui/raidboss/data/06-ew/hunts/thavnair.ts
+++ b/ui/raidboss/data/06-ew/hunts/thavnair.ts
@@ -145,6 +145,13 @@ const triggerSet: TriggerSet<Data> = {
       condition: (data) => data.inCombat,
       response: Responses.getBehind(),
     },
+    {
+      id: 'Hunt Yilan Mini Light',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6A5F', source: 'Yilan', capture: false }),
+      condition: (data) => data.inCombat,
+      response: Responses.getOut(),
+    },
   ],
   timelineReplace: [
     {
--- a/ui/raidboss/data/06-ew/hunts/ultima_thule.ts
+++ b/ui/raidboss/data/06-ew/hunts/ultima_thule.ts
@@ -82,29 +82,9 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'Hunt Fan Ail Divebomb',
       type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: '6AED', source: 'Fan Ail' }),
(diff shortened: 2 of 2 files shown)


ui/raidboss/data/06-ew/hunts/thavnair.ts
```diff
@@ -145,6 +145,13 @@ const triggerSet: TriggerSet<Data> = {
       condition: (data) => data.inCombat,
       response: Responses.getBehind(),
     },
+    {
+      id: 'Hunt Yilan Mini Light',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '6A5F', source: 'Yilan', capture: false }),
+      condition: (data) => data.inCombat,
+      response: Responses.getOut(),
```
quisquous, line 153: Do you think it's worth only doing this if you don't have a march debuff?

ui/raidboss/data/06-ew/hunts/ultima_thule.ts
```diff
@@ -82,29 +82,9 @@ const triggerSet: TriggerSet<Data> = {
...
-        },
-        divebombMarker: {
-          en: 'Away from Divebomb Marker',
-          de: 'Weg von dem Sturzflug-Marker',
-          fr: 'Éloignez-vous de la bombe plongeante',
-          cn: '躲开俯冲点名',
-          ko: '초록징 피하기',
-        },
-      },
+      response: Responses.awayFromFront(),
```
quisquous, line 87: I disagree a little here? It's away from where the divebomb marker appeared, not from the front of Fan Ail.  I think it behaves like t5 divebomb markers.

ui/raidboss/data/06-ew/hunts/thavnair.ts
Bing-su: No, I think it's better to call always

ui/raidboss/data/06-ew/hunts/ultima_thule.ts
Bing-su: Unlike twintania, Fan Ail always turns to where the divebomb marker appeared.

quisquous: Oh, you're totally right.  Thanks for letting me know.

PR #4663, opened 2022-08-04
raidboss: dsr fix p5 wrath/doth targetable

Legends0:
Order was wrong in the timeline which results in the targetable message not appearing in timeline.


quisquous:
I'm a little confused here.  The order in a timeline file is just for convenience, and everything should be ordered by time when running.

Legends0:
> I'm a little confused here. The order in a timeline file is just for convenience, and everything should be ordered by time when running.

I may just not have noticed the targetable message at the time then and wrongly thought something was wrong. I'll close and recheck that it's there.

PR #4669, opened 2022-08-09
i18n: Add CN translations for 6.1

fakedestinyck:


ui/raidboss/data/06-ew/trial/endsinger-ex.ts
```diff
@@ -734,9 +734,22 @@ const triggerSet: TriggerSet<Data> = {
...
+        'Befoulment': '脓液弹',
+        'Benevolence': '博爱',
+        'Despair Unforgotten': '绝望侵蚀：记录事件',
+        'Diairesis': '分离',
+        'Eironeia': '反讽',
+        'Elegeia Unforgotten': '哀歌：记录事件',
         'Elenchos': '反诘',
+        'Endsong\'s Aporrhoia': '流溢：绝望轮唱',
+        'Endsong(?!\')': '绝望轮唱',
         'Fatalism': '宿命',
```
quisquous, line 746:
```suggestion
        '(?<! )Fatalism': '宿命',
```

Because `Theological Fatalism` and `Fatalism` try to translate the same phrase, you need to differentiate them like this.

quisquous:
Hmm, it looks like #4670 is doing the same thing as this PR but has a bunch more translations.  Thanks for sending this, but I'm going to close this PR in favor of that one just so @ShadyWhite doesn't have to untangle conflicts to land both of these.

PR #4670, opened 2022-08-09
i18n: CN 6.1

ShadyWhite:
#4669 is OK to close.
Late but not too late.
虽迟但到。
Check plz. @MaikoTan @tssailzz8


quisquous:
Ok, this looks like it covers everything from #4669 and then some.  I will close that one in favor of this PR.  Can you also resolve the conflicts here, and let me know when this is ready to merge?

ShadyWhite:
Ready to merge.

Akurosia:
Where did you got the translations for e.g. `Flames of Ascalon`? And would there be a way to get them or better a log with debug lines so i can get all obfuscated information for fragonsong?

PR #4677, opened 2022-08-19
raidboss: DSR p7 Gigaflare direction to run

Legends0:
Calls out clockwise or counterclockwise based on the atan2 of determinant and dot product of first and second gigaflare NetworkStartsCasting locations. Negative angle should mean one direction, positive angle should mean the other.

Possible we could also replace this with relative to boss since I imagine most groups are facing the boss?

~~To be tested.~~ It's probably okay now. I only ran with it once and didn't have a problem with in/out tts over this tts, but it's easy enough to test out with raidsim or fflogs data and typescript sim.


Legends0:
Possible this also calls out the first safe spot such as Southwest => Clockwise or Southwest => Left.

Legends0:
Another note, this currently happens 1 second before an in/out call, and given TTS lengths it likely leads to multiple callouts at once.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -2500,6 +2502,55 @@ const triggerSet: TriggerSet<Data> = {
...
+          const first = data.firstGigaflare;
+          const second = data.secondGigaflare;
+          delete data.firstGigaflare;
+          delete data.secondGigaflare;
+
+          if (first[0] === undefined || first[1] === undefined || second[0] === undefined || second[1] === undefined) {
+            console.error(`Gigaflare: missing coordinates`);
+            return;
+          }
+          // Compute atan2 of determinant and dot product to get rotational direction
```
quisquous, line 2531: !!!

quisquous (commented):
> Possible this also calls out the first safe spot such as Southwest => Clockwise or Southwest => Left.

I think this could say "Back left -> Clockwise" sort of thing.  They are definitely boss relative.

Let me know if you have a chance to test this out, as I may or may not have a chance in the next week or two.  I'm not sure that I know this math well enough to know if there are edge cases to worry about.

ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
Legends0: > 
> I think this could say "Back left -> Clockwise" sort of thing. They are definitely boss relative.
> 
> Let me know if you have a chance to test this out, as I may or may not have a chance in the next week or two. I'm not sure that I know this math well enough to know if there are edge cases to worry about.

From my checks in typescript sim (pulling x,y from fflogs reports) and raidsim of my own logs it seems to at least be correct. I'll likely be able to do a test in game for how it feels on Sunday.

Here are some more details on how the algorithm works: https://stackoverflow.com/a/16544330 . I think the only way for a 0 value to be returned is if the two coordinates are the same, which shouldn't be possible.

As for getting relative to boss direction for a Back Left => clockwise like call, I think I would need to add tracking of the 6D99 cast to get facing. I know the x,y appear to be accurate, but I haven't really looked at if the facing value there is accurate.

Legends0:
I've added the starting location to the callout. The only oddity I ran into was switching the value of X and Y, normally it's cosine that gets X and sine that gets Y, but I am guessing since the Y value is flipped (North is negative) the X,Y calculation here had to be flipped.

Another thing to note, it's a radius 14 for the gigaflares from boss. Calculate the point the boss would be facing to on this "circle", then used the same math for determining clock/counterclock between this point and the first gigaflare to know where to move to behind the boss. Since the heading of the boss and the angle of the gigaflares may not be exact data, I used an alternate method for determining the front/back one that looks to be consistent.

Legends0:
The math for if it's front is not correct, but the counter/clockwise portion does seem to be. I need to come up with a better method for front.

Legends0:
I see, the X and Y seem to be flipped for the atan2 to method to work. The reason the others work is probably because the angles are just relative and since both are calculated the same way it produces a result that works.

quisquous:
Sorry for leaving this PR for so long.  For some reason I thought it still needed testing?  Is this ready to go once the lint issues are solved?

Legends0:
> Sorry for leaving this PR for so long. For some reason I thought it still needed testing? Is this ready to go once the lint issues are solved?

It was good with all the values I tested.

I think the lint checker is actually broken though, the fix it is doing does not make sense to me. It wants to add a comma inside the conditional...

quisquous:
> I think the lint checker is actually broken though, the fix it is doing does not make sense to me. It wants to add a comma inside the conditional...

I think the statement is so long that it thinks it's an item in a list?

You could always break it into smaller statements that don't need to be wrapped, and then there'd be no need for a comma either.

Legends0:
Ok, I broke the conditional into statements, this removes a set of parenthesis since order is now preserved in the statement instead which the variable can then be operated.

Legends0:
It should be ready to go, and this is verifiable in raid emulator as well. I think it was last pending that you wanted to trial run it, but never had the time to get around to it.

PR #4678, opened 2022-08-19
raidboss: DSR fix#4664 track p6/p7 by ability trigger

### LINKED ISSUE #4664: raidboss: DSR phase tracker misses P6
## Description
During P6 of DSR, `data.phase` remains set to `thordan2`. I was expecting it to update to `nidhogg2`.

## Additional information
This issue happens for me both in the raidboss emulator and in-game. That said, I don't think this technically broke anything - the existing P6 triggers don't have a phase-check condition afaik.

My best guess at the root cause is that the DSR Phase Tracker matches a `StartsCasting` log line, which doesn't seem to exist for 6667. ([link to trigger](https://github.com/quisquous/cactbot/blob/triggers/06-ew/ultimate/dragonsongs_reprise_ultimate.js#L166)).

Legends0:
fixes #4664

Removes the P6 and P7 regex captures from DSR Phase Tracker as they are not casted abilities.
Adds DSR Phase Tracker P6/P7 to capture network abilities for the p6/p7 ids.
Changes regex to match Shockwave as indicator for p7.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -241,7 +241,7 @@ const triggerSet: TriggerSet<Data> = {
       // 6708 = Final Chorus
       // 62E2 = Spear of the Fury
       // 6B86 = Incarnation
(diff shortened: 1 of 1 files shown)


quisquous:
Thank you!

PR #4688, opened 2022-08-23
plugin: update bozja intl opcode for v6.2

KattTails:
Tested working. Sorry for the delay, had to get it manually.


quisquous:
Thank you as always!

PR #4689, opened 2022-08-24
raidboss: barbariccia-ex add some triggers and timeline

Legends0:
~~The timeline needs more work, it does not account for the patterns that exist (I think there are just two?)~~

EDIT: I've added some of the regex from logs I could find to match the possible values, at least for the first phase.


Legends0:
Timeline needs to be retimed too. Original timeline based on earlier log version than what is now live.

Legends0:
I'm almost certain that the animation the boss is doing could be indicated by the spellid, but I don't have footage to compare with logs, so will leave that for someone else to solve.

copenblend (approved):

ui/raidboss/data/06-ew/trial/barbariccia-ex.txt
```diff
@@ -1,7 +1,315 @@
...
+
+439.9 "Void Aero IV" sync / 1[56]:[^:]*:Barbariccia:7570:/
+447.1 "Raging Storm" sync / 1[56]:[^:]*:Barbariccia:7572:/
+
+# Transition 3 - Second Curling Iron
+455.4 "Curling Iron" sync / 1[56]:[^:]*:Barbariccia:75B2:/
+460.7 "Voidstrom" duration 2 #sync / 1[56]:[^:]*:Barbariccia:7577:/
+463.7 "Curling Iron" sync / 1[56]:[^:]*:Barbariccia:75B3:/
+
+468.8 "Brutal Rush 1" sync / 1[56]:[^:]*:Barbariccia:7583:/
```
quisquous, line 213: Did you run this through test_timeline? I'm surprised this worked with all of the repeated ids here next to each other with overlapping syncs.  I think for something like these, maybe it'd make sense to just have a duration? I don't think having all of these listed is as helpful as something like sanctity meteor counts.

Legends0: I never ran it through test timeline, but I think it did work in game, or worked enough.
I was thinking it should be changed to duration, it's odd for two spells, but for the most part I think it would be more functional as a duration since one part is the hit and the other is the line.

valarnin:
These are all untested but at a glance:

746[456789] = under?
748[9ABCDE] = line?

quisquous:
Ok thanks, I'll land this as a first pass and we can make improvements from here.

PR #4690, opened 2022-08-24
jobs: 6.2 changes

Echoring:
- GNB: Bloodfest cooldown changes to 120s.
- DRK: Living Shadow duration is reduced. 
And I think it's not necessary to trace its duration, for there is no action to make the Shadow attack quicker.
The duration-cooldown structure is lengthy, I'm thinking about use a util func to instead them like
```
export const showDuration = (o: {
  tid: number;
  timerbox: TimerBox;
  duration: number;
  cooldown: number;
  threshold: number;
  activecolor: string;
  deactivecolor: string;
}): number => {
  o.timerbox.duration = o.duration;
  o.timerbox.threshold = o.duration;
  o.timerbox.fg = computeBackgroundColorFrom(o.timerbox, o.activecolor);
  o.tid = window.setTimeout(() => {
    o.timerbox.duration = o.cooldown - o.duration;
    o.timerbox.threshold = o.threshold;
    o.timerbox.fg = computeBackgroundColorFrom(o.timerbox, o.deactivecolor);
  });
  return o.tid;
```
maybe later for more test and adjust.

### DIFF
--- a/ui/jobs/components/drk.ts
+++ b/ui/jobs/components/drk.ts
@@ -16,7 +16,6 @@ export class DRKComponent extends BaseComponent {
   delirium: TimerBox;
   livingShadow: TimerBox;
   tid1 = 0;
-  tid2 = 0;
 
   constructor(o: ComponentInterface) {
     super(o);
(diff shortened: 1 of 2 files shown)


quisquous:
> The duration-cooldown structure is lengthy, I'm thinking about use a util func to instead them like

Yeah, I think this would be a nice simplification in the future.  Good thought!

PR #4691, opened 2022-08-24
raidboss/oopsy: update actor control wipe in code/docs

quisquous:
In 6.2, the "fade in" actor control changed from
`40000010` to `4000000F`.  This commit changes code and
docs, and will follow up with a timeline change after.


panicstevenson (approved):

trim21:
why don't use same wipe regex from util/logtools/encounter_tools.ts?

PR #4431, opened 2022-05-10
jobs/test: remove pre-6.x workarounds

quisquous:

PR #4692, opened 2022-08-24
raidboss: update actor control wipe for timelines

quisquous:
This is the timeline change for #4691.
To avoid having to redo this change in the future,
change the timelines to the desired future 6.2 code,
and use the same "back compat replacement" strategy
that we used before #4431 removed them.


quisquous:
Tested this manually by printing out the regex from timelines.  It looks good to go.

Issue #4682, opened 2022-08-23
6.2 content additions planning


JLGarber:
With more of us contributing here, it Seems To Me to make sense to do a bit of a roll call for those of us who do regular raidboss/oopsy work so we don't step over each other. I don't really care which piece of content I do aside from "I'd prefer not to do the trial".

### Complete

- MSQ Dungeon — quisquous (initial timeline/triggers/oopsy complete)
- MSQ Trial -- panicstevenson (initial timeline/triggers/oopsy complete)
- EX Trial -- Legends0 (initial timeline/triggers/oopsy complete)
- Pandaemonium 5 — panicstevenson (initial timeline/triggers/oopsy complete)
- Pandaemonium 6 -- JLGarber and quisquous (initial timeline/triggers/oopsy complete)
- Pandaemonium 7 -- JLGarber (initial timeline/triggers/oopsy complete)
- Pandaemonium 8 -- quisquous (initial timelintriggers/oopsy complete)
- Sephirot Unreal -- JLGarber (initial timeline/triggers/oopsy complete)

I'm envisioning each of us noting what we're going to focus on, possibly sharing logs to whoever has a piece of content that we're running. I'll start by claiming the Unreal, since I already did the EX version. I'll also pick some other stuff up in a day or so if not everything is in progress, since the Unreal should be a simple find/replace for the most part.

I know of at least @valarnin @Legends0 as well as @quisquous who regularly do Raidboss stuff. Maybe @panicstevenson as well. Others please also add in your preference on content once we see stuff released!

valarnin:
Thoughts:

1. Chime in here when we start working on something. Follow up as needed
   - Kind of clutters up a single issue
   - Makes it hard to track comments by content
2. Use this issue as a reference, open a separate issue mentioning this issue when we start working on the content
   - Lets this issue serve as a common tracking point for the patch
   - Can reference the separate issue in the PR for auto-closing
   - Problems/thoughts/brainstorming tracked per content, easier to follow and chime in if someone has feedback/help

Additionally, I think a super early PR that adds skeleton/placeholder files and folders for timeline and trigger files would be a good idea to avoid conflicts in manifest?

Legends0:
I agree with having skeleton files in place. Makes it easier to setup and start testing since I often don't have the zone ids.

[...]

PR #4696, opened 2022-08-25
raidboss: ex4 switch impact trigger

Legends0:
Workaround for a possible timeline trigger issue with tracking Impact. This uses different spell that is also related to the knockback and has a cast time. Using the "standard" 5s timing, although the spell cast is almost the same as the knockback immunity duration.


Legends0:
This was brought up in #4682 discussion by @xpdota

PR #4708, opened 2022-08-26
raidboss: Initial P5N Timeline and Triggers

panicstevenson:


JLGarber (approved):
Yeah, the "dodge to the starting position when safe" bits are *tough* to call concisely. Maybe we can call actual quadrants once we have time to look for position data on the crystals?

panicstevenson:
> Yeah, the "dodge to the starting position when safe" bits are _tough_ to call concisely. Maybe we can call actual quadrants once we have time to look for position data on the crystals?

Yea, that's kind of my thought as well. I am definitely looking for better placeholders if anyone can think of them though. It's going to be somewhat important to get this information before savage though.

PR #4716, opened 2022-08-27
raidboss: Add Sephirot Unreal

JLGarber:
Largely a matter of find/replace. I finished up the timeline and triggers prior to an attempt at running it, and everything seemed to work okay. The "change to Earth Shakers" that was mentioned in one of the live letters is simply that the targets are now *also* clearly marked with stretchy tethers that turn purple when extended far enough.

Oopsy is untested, but the IDs all match, so I'm comfortable releasing it as it stands.

### DIFF
--- a/ui/oopsyraidsy/data/06-ew/trial/sephirot-un.ts
+++ b/ui/oopsyraidsy/data/06-ew/trial/sephirot-un.ts
@@ -10,29 +10,29 @@ export interface Data extends OopsyData {
 const triggerSet: OopsyTriggerSet<Data> = {
   zoneId: ZoneId.ContainmentBayS1T7Unreal,
   damageWarn: {
-    'SephirotEX Yesod': '76AB', // Snapshot floor spikes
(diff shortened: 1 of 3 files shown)


ui/oopsyraidsy/data/06-ew/trial/sephirot-un.ts
```diff
@@ -1,11 +1,146 @@
...
 
-export type Data = OopsyData;
+export interface Data extends OopsyData {
+  force?: { [name: string]: string };
+}
 
 const triggerSet: OopsyTriggerSet<Data> = {
   zoneId: ZoneId.ContainmentBayS1T7Unreal,
+  damageWarn: {
+    'SephirotEX Yesod': '76AB', // Snapshot floor spikes
```
quisquous, line 13: Can you rename these to `SephirotUn`?

JLGarber: aaaaa. I knew I'd forgotten something.

PR #4728, opened 2022-08-29
i18n: p8n translation

Akurosia:
'missingTranslations': true, is only set as i get following error without it;

Y:\Github\cactbot_server\ui\raidboss\data\06-ew\raid\p8n.txt:84 sync " 1[56]:[^:]*:[^:]*:6723:"


quisquous:
Mmm, ok.  Probably the script needs to be updated to account for that.

PR #4732, opened 2022-08-30
raidboss: P6N timeline

quisquous:


JLGarber (approved):
Thanks for taking over on this. I was completely lost due to broken FFLogs data and couldn't see that clean loop right there.

PR #4740, opened 2022-08-30
raidboss: Initial P5S Triggers

panicstevenson:


ui/raidboss/data/06-ew/raid/p5s.ts
```diff
@@ -1,13 +1,210 @@
...
+      outputStrings: {
+        busterOnYou: Outputs.tankBusterOnYou,
+        busterOnTarget: Outputs.tankBusterOnPlayer,
+      },
+    },
+    {
+      id: 'P5S Double Rush',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '771B', source: 'Proto-Carbuncle', capture: false }),
+      response: Responses.knockback(),
```
Legends0, line 104: This is a 5.7s cast, followed by a 2nd one 2.1s later, does this catch both?

ui/raidboss/data/06-ew/raid/p5s.ts
```diff
@@ -1,13 +1,210 @@
...
+        return output.busterOnTarget!({ player: data.ShortName(data.target) });
+      },
+      outputStrings: {
+        busterOnYou: Outputs.tankBusterOnYou,
+        busterOnTarget: Outputs.tankBusterOnPlayer,
+        tankSwap: Outputs.tankSwap,
+      },
+    },
+    {
+      // Maybe too noisy?
```
Legends0, line 83: I was thinking just output.tankBusters for Venomous Gale would be fine for healer role.

panicstevenson: Sure! It has to be convoluted because they don't cast on a target. I'll add a note about that as well.

ui/raidboss/data/06-ew/raid/p5s.ts
panicstevenson: image

Just the first one.

ui/raidboss/data/06-ew/raid/p5s.ts
```diff
@@ -1,13 +1,212 @@
...
+      run: (data, matches) => {
+        const index = convertAbilityIdToTopazRayIndex(matches.id);
+        if (data.topazRays[index] === undefined)
+          data.topazRays[index] = [];
+        const direction = convertCoordinatesToDirection(parseFloat(matches.x), parseFloat(matches.y));
+        data.topazRays[index]?.push(direction);
+      },
+    },
+    {
+      id: 'P5S Topaz Ray',
```
quisquous, line 157: How confident are you on this one? Is it worth landing the more sure stuff first and then landing this later after direct testing?

ui/raidboss/data/06-ew/raid/p5s.ts
quisquous: This should have a delay so it covers both, right?

ui/raidboss/data/06-ew/raid/p5s.ts
panicstevenson: As sure as I can be without video, honestly. The logic is pretty sound and it gives 4 results, but I can't confirm them otherwise. Is there a way to make it experimental?

quisquous: No, just remove it and add it locally and then send another PR later.

PR #4741, opened 2022-08-30
raidboss: p5s draft timeline

Legends0:
I don't have a network log for this to be able to test it. Based on fflogs reports.

Targetable/untargetable lines need to be adjusted. Enumerations I don't think are accounted for in the timeline.


quisquous:
Haha we both uploaded one at the same time.  Amazing.  I'm going to commit mine because I've run it against the tester.  Feel free to clean it up and merge in anything that makes sense.  I haven't seen a log with enrage.

There's a network log in the #contribute-logs channel of discord.

PR #4744, opened 2022-08-30
raidboss: P5S Topaz Rays Callouts

panicstevenson:


ui/raidboss/data/06-ew/raid/p5s.ts
```diff
@@ -115,6 +141,72 @@ const triggerSet: TriggerSet<Data> = {
       durationSeconds: 5,
       response: Responses.getBackThenFront(),
     },
+    {
+      id: 'P5S Topaz Ray Collect',
```
valarnin, line 145: What does using a collector here gain over just putting this in the below trigger's `run` instead?

Legends0:
In my testing, this does not seem to work for all patterns.

quisquous:
Personally, I would also suggest that this only call out one adjacent safe spot per callout.  If I were gonna be biased, I'd say start with SW/SE if possible too, since one of those is always a safe place to start.

valarnin:
For reference this was my scratch attempt at this trigger when I was progging p5s, I think it has two of the directions inverted (unless I fixed that and don't remember doing so), but it all fires off of the 79FF that the crystals themselves cast.

```js
    {
      id: 'Custom - P5S - Memory Game',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: '79FF' }),
      infoText: (data, matches) => {
        data.memoryGameCount ??= 1;
        data.memoryGame ??= [];
        data.memoryGame.push(matches);
        const count = data.memoryGameCount < 3 ? 2 : 3;
        if (data.memoryGame.length >= count) {
          const regions = {
            NW:1,
            NE:1,
            SE:1,
            SW:1,
          };
          for (const match of data.memoryGame) {
            let quadrant = '';
            if (parseFloat(match.y) > 100)
              quadrant += 'S';
            else
              quadrant += 'N';
            if (parseFloat(match.x) > 100)
              quadrant += 'E';
            else
              quadrant += 'W';
            delete regions[quadrant];
          }
          data.memoryGameCount++;
          data.memoryGame = [];
          return Object.keys(regions).join('|');
        }
      },
    },
```

panicstevenson:
> In my testing, this does not seem to work for all patterns. It will sometimes not call anything.

This seems due to sometimes stale positional data in the log itself. I'll run with it calling OverlayPlugin and report back.

panicstevenson:
> For reference this was my scratch attempt at this trigger when I was progging p5s, I think it has two of the directions inverted (unless I fixed that and don't remember doing so), but it all fires off of the 79FF that the crystals themselves cast.

This might be the better approach, the only concern is that `79FF` happens quite often. Additionally, I think you can run into the above issue with the stale position data above as well, so it would need to be converted to getCombatants.

It might be worth taking `79FF` and expanding it to include all secondary intercardinal directions so that it can solve all the permutations in one trigger, then have Ruby Glow preRun clear the positions and initiate a collector.

ui/raidboss/data/06-ew/raid/p5s.ts
panicstevenson: I guess there is none as long as they return if the array's length isn't expected. I added the safeguards after writing the skeleton. However, I think the next iteration requires getCombatants, so it will still need a collector.

quisquous:
Did you say you had tested this and it's ready to land? It looks fine to me, but just wanted to check about the state of things.

panicstevenson:
> Did you say you had tested this and it's ready to land? It looks fine to me, but just wanted to check about the state of things.

I tested one iteration previous to this one live and it worked very well. Since then, I cleaned up the triggers and made it into a single callout. It's hard to test since sometimes the emulator doesn't get perfect data either, but on those pulls it doesn't show anything and I have some pulls that have the correct callout (confirmed with video recordings as well). I never had a pull running it live that had bad data from getCombatants.

PR #4751, opened 2022-09-01
raidboss: p5s add targetable/untargetable lines

Legends0:
The targetable line seems to be +- .1s depending on if relative to 7708 or relative to first 7A03, I chose to make it relative to the first 7A03.
Untargetable is relative to the last 7A03.

I retrieved the lines from make_timeline.py.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p5s.ts
+++ b/ui/raidboss/data/06-ew/raid/p5s.ts
@@ -156,6 +156,7 @@ const triggerSet: TriggerSet<Data> = {
   timelineReplace: [
     {
       'locale': 'de',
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/06-ew/raid/p5s.ts
```diff
@@ -156,6 +156,7 @@ const triggerSet: TriggerSet<Data> = {
   timelineReplace: [
     {
       'locale': 'de',
+      'missingTranslations': true,
```
quisquous, line 159:
```suggestion
```

I don't think you should need this line.  Targetable and untargetable should be automatically translated.

PR #4752, opened 2022-09-01
raidboss: p5s add Ruby Glow aoe callout

Legends0:
Captures both spellids similar to the normal mode trigger.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p5s.ts
+++ b/ui/raidboss/data/06-ew/raid/p5s.ts
@@ -41,6 +41,13 @@ const triggerSet: TriggerSet<Data> = {
       response: Responses.aoe(),
     },
     {
+      id: 'P5S Ruby Glow',
+      type: 'StartsUsing',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p5s.ts
```diff
@@ -40,6 +40,13 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.startsUsing({ id: '7720', source: 'Proto-Carbuncle', capture: false }),
       response: Responses.aoe(),
     },
+    {
+      id: 'P5S Ruby Glow',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '76F[34]', source: 'Proto-Carbuncle', capture: false }),
+      suppressSeconds: 1,
```
quisquous, line 47:
```suggestion
      netRegex: NetRegexes.startsUsing({ id: '76F3', source: 'Proto-Carbuncle', capture: false }),
```

`76F3` is always paired with `76F4`, so only need one here.  `76F4` is the damage and `76F3` is the self-targeted ability, I think?  I don't think there's an instance where `76F4` happens on its own.

PR #4758, opened 2022-09-01
raidboss: p6s Exchange of Agonies headmarkers

Legends0:
This is a continuation of the previous headmarkers in PR #4757

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -164,7 +164,7 @@ const triggerSet: TriggerSet<Data> = {
           cn: '集合放月环',
           ko: '도넛 장판, 쉐어',
         },
-        goCorner: {
+        spreadCorner: {
           en: 'Spread Corner',
         },
       },
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -102,8 +102,6 @@ const triggerSet: TriggerSet<Data> = {
           case '0168':
           case '016E':
             return output.donut!();
-          default:
-            return output.unknown!();
         }
       },
       outputStrings: {
@@ -119,7 +117,6 @@ const triggerSet: TriggerSet<Data> = {
         goCorner: {
           en: 'Spread Corner',
         },
-        unknown: Outputs.unknown,
       },
     },
     {
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -85,7 +85,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'HeadMarker',
       netRegex: NetRegexes.headMarker({}),
       condition: (data, matches) => {
-        return data.me === matches.target && !(/(00(?:4F|5[0-6]))|013E|0148|0065/).test(getHeadmarkerId(data, matches));
+        return data.me === matches.target;
       },
       infoText: (data, matches, output) => {
         const correctedMatch = getHeadmarkerId(data, matches);
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -127,6 +127,7 @@ const triggerSet: TriggerSet<Data> = {
           ja: '${num}番',
           cn: '#${num}',
           ko: '${num}번째',
+        },
       },
     },
     {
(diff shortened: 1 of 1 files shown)


valarnin:
I've been using these scratch headmarker triggers, for reference. The number in comments on markers was count at that offset in ms into the pull.
```js
const headmarkerIds = {
  '0143': 'Stack', // Stack marker; 51390 x2, 276236 x2, 538819 x2
  '0054': '1', // LC 1; 63787
  '0055': '2', // LC 2; 63787
  '0056': '3', // LC 3; 63787
  '0057': '4', // LC 4; 63787
  '0058': '5', // LC 5; 63787
  '0059': '6', // LC 6; 63787
  '005A': '7', // LC 7; 63787
  '005B': '8', // LC 8; 63787
  '0169': 'Out', // 105496 x2, 430453 x2, out?
  '0173': 'In', // 105496 x2, 430453, in?
  '016F': 'Out', // 105496, 430453 x2, out?
  '016D': 'In', // 105496, 430453 x2, in?
  '016A': 'Out', // 105496, 217336 x2, out?
  '016C': 'In', // 105496, 217336 x2, in?
  '006A': 'Spread', // 294863 x8, spread markers
  '014D': 'Spread', // 402277 x8, 538819 x2, spread markers
  '0168': 'In', // 430453, in?
};

const firstDecimalMarker = 0x0143;

const getHeadmarkerId = (data, matches) => {
    data.decOffset = parseInt(matches.id, 16) - firstDecimalMarker;
  }
};

// ...

  triggers: [
    {
      id: 'Custom - P6S - Headmarker Collector',
      type: 'HeadMarker',
      run: (data, matches) => getHeadmarkerId(data, matches),
    },
    {
      id: 'Custom - P6S - Headmarkers',
      type: 'HeadMarker',
      infoText: (data, matches) => headmarkerIds[getHeadmarkerId(data, matches)] ?? '???',
    },
  ],
```

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -48,6 +80,51 @@ const triggerSet: TriggerSet<Data> = {
...
+      type: 'HeadMarker',
+      netRegex: NetRegexes.headMarker({}),
+      condition: (data, matches) => {
+        return data.me === matches.target && !(/(00(?:4F|5[0-6]))|013E|0148|0065/).test(getHeadmarkerId(data, matches));
+      },
+      infoText: (data, matches, output) => {
+        const correctedMatch = getHeadmarkerId(data, matches);
+        switch (correctedMatch) {
+          case '0163':
+          case '0167':
```
quisquous, line 94: Somebody else kindly sent me:

163    Stack
167    Stack
169    Stack

164    Spread
165    Spread
16A    Spread

166    Dynamo
168    Dynamo
16E    Dynamo

Are these missing? It's possible some of them are from the second application, in which case maybe the callout should just be "stack" and not "stack donut"?

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -48,6 +80,51 @@ const triggerSet: TriggerSet<Data> = {
...
+        donut: {
+          en: 'Stack Donut',
+          de: 'Sammeln Donut',
+          fr: 'Packez-vous, donut',
+          ja: '頭割り',
+          cn: '集合放月环',
+          ko: '도넛 장판, 쉐어',
+        },
+        goCorner: {
+          en: 'Go To Corner',
```
quisquous, line 118: This is a tiny trigger readability request, but can this be `Spread Corner` or `Spread in Corner` or some such, so that the first word of the callout is either "stack" or "spread"?

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -48,6 +80,51 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.startsUsing({ id: '7881', source: 'Hegemone', capture: false }),
       response: Responses.goSides(),
     },
+    {
+      id: 'P6S Exchange of Agonies Markers',
+      type: 'HeadMarker',
+      netRegex: NetRegexes.headMarker({}),
+      condition: (data, matches) => {
+        return data.me === matches.target && !(/(00(?:4F|5[0-6]))|013E|0148|0065/).test(getHeadmarkerId(data, matches));
```
quisquous, line 88: I'm not sure why you need a test here too.

What about: (1) condition is just data.me===matches.target and (2) drop the unknown from the switch statement and just return silently if it's not one of the correct markers?

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: Ah, it's possible I was missing some headmarkers as I only checked three logs.

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: It looked like I was missing some headmarkers so I put an unknown catch, but your previous comment should mean we have them all now.

PR #4759, opened 2022-09-01
raidboss: P6s initial timeline

JLGarber:
I built this based on a log from quisquous. This has received no live testing, and I haven't taken the time to match everything 1-1 with the video I was handed alongside, but it passed `test_timeline` against every encounter in that log.

Given that my group hasn't started on this yet, I don't know what timeline elements are superfluous. Unholy Darkness and Dark Dome in particular are probably in need of some cleanup. I also didn't distinguish any potential 100% split or shared Synergy busters.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p6s.txt
+++ b/ui/raidboss/data/06-ew/raid/p6s.txt
@@ -16,9 +16,9 @@ hideall "--sync--"
 42.0 "--sync--" sync / 1[56]:[^:]*:Hegemone:7854:/
 48.6 "Unholy Darkness" sync / 1[56]:[^:]*:Hegemone:(786D|786E|7891):/
 54.6 "Polyominous Dark IV" sync / 1[56]:[^:]*:Hegemone:7867:/
-54.7 "Exocleaver 1" sync / 1[56]:[^:]*:Hegemone:7869:/
+54.7 "Exocleaver 1" sync / 1[56]:[^:]*:Hegemone:(7869|786B):/
 56.8 "Unholy Darkness" sync / 1[56]:[^:]*:Hegemone:7892:/
-57.2 "Exocleaver 2" sync / 1[56]:[^:]*:Hegemone:786B:/
+57.2 "Exocleaver 2" sync / 1[56]:[^:]*:Hegemone:(7869|786B):/
 71.2 "Pathogenic Cells x8" sync / 1[56]:[^:]*:Hegemone:7864:/ duration 15
 91.9 "--sync--" sync / 1[56]:[^:]*:Hegemone:784C:/
 97.7 "Aetherial Exchange" sync / 1[56]:[^:]*:Hegemone:784D:/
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/06-ew/raid/p6s.txt
```diff
@@ -1,7 +1,121 @@
...
+
+0.0 "--sync--" sync /Engage!/ window 0,1
+9.5 "--sync--" sync / 14:[^:]*:Hegemone:7860:/ window 10,5
+14.2 "Hemitheos's Dark IV" sync / 1[56]:[^:]*:Hegemone:7860:/
+28.0 "Synergy" sync / 1[56]:[^:]*:Hegemone:(7887|788A):/
+35.1 "--sync--" sync / 1[56]:[^:]*:Hegemone:784C:/
+42.0 "Aetheric Polyominoid" sync / 1[56]:[^:]*:Hegemone:7866:/ window 15,15
+42.0 "--sync--" sync / 1[56]:[^:]*:Hegemone:7854:/
+48.6 "Unholy Darkness" sync / 1[56]:[^:]*:Hegemone:(786D|786E|7891):/
+54.6 "Polyominous Dark IV" sync / 1[56]:[^:]*:Hegemone:7867:/
```
valarnin, line 18: This line is the line cleaves from the floor markers, not the raidwide/aoe, so it probably shouldn't be called the same thing in timeline even if the cast is the same name?

quisquous: The raidwide is `Hemitheos's Dark IV`, right?

ui/raidboss/data/06-ew/raid/p6s.txt
```diff
@@ -1,7 +1,121 @@
...
+141.4 "Hemitheos's Dark IV" sync / 1[56]:[^:]*:Hegemone:7860:/ window 30,30
+153.2 "Transmission" sync / 1[56]:[^:]*:Hegemone:7861:/
+159.0 "--sync--" sync / 1[56]:[^:]*:Hegemone:784C:/
+166.1 "Choros Ixou 1" sync / 1[56]:[^:]*:Hegemone:(7881|7882|7883|7884):/
+168.0 "Chelic Claw/Reek Havoc" sync / 1[56]:[^:]*:Parasitos:(79EE|79EF):/
+169.2 "Choros Ixou 2" sync / 1[56]:[^:]*:Hegemone:(7881|7882|7883|7884):/
+179.0 "Aetherial Exchange" sync / 1[56]:[^:]*:Hegemone:784D:/
+185.7 "Polyominoid Sigma" sync / 1[56]:[^:]*:Hegemone:7868:/ window 30,30
+185.7 "--sync--" sync / 1[56]:[^:]*:Hegemone:7854:/
+195.3 "Dark Dome 1" sync / 1[56]:[^:]*:Hegemone:788B:/
```
quisquous, line 39: For consistency, can you do the same thing with all 3 `Dark Dome`? The first ability is a castbar that leaves baited circles on the floor and the second ability is when those baited circles explode.  It's not really the same thing twice, so 1 and 2 numbering doesn't quite make sense here.

Maybe the second one should be `--circles--` or `--dark dome--` or something along those lines?

quisquous:
I think landing something here is better than nothing.  I'm going to leave the comments here as suggestions for follow-ups, and will merge this for now.

PR #4769, opened 2022-09-01
Update p5s.js

Berserk1993:


quisquous:
I think you're looking for <https://github.com/quisquous/cactbot/blob/main/docs/CactbotCustomization.md#changing-trigger-text-with-the-cactbot-ui>

PR #4770, opened 2022-09-01
raidboss: p6s Exocleaver Move + Healer Groups

Legends0:
Adds callout for the Exocleaver mechanic and the healer group headmarkers (013E).

Notes:
- ~~Unholy Darkness stack headmarkers are 3.8s before first Exocleavers~~ EDIT: Unholy Darkness stack headmarkers come out same time as first Exocleavers
- Exchange of Agonies headmarkers are ~~7~~3s before second Exocleavers

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -84,9 +84,10 @@ const triggerSet: TriggerSet<Data> = {
     },
     {
       id: 'P6S Exocleaver Move',
-      type: 'StartsUsing',
-      netRegex: NetRegexes.startsUsing({ id: ['7869', '786B'], source: 'Hegemone' }),
-      delaySeconds: (_data, matches) => parseFloat(matches.castTime),
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: ['7869', '786B'], source: 'Hegemone', capture: false}),
+      // Supress until after second Exocleaver in the set
+      suppressSeconds: 4,
       response: Responses.moveAway(),
     },
     {
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -87,10 +87,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: ['7869', '786B'], source: 'Hegemone' }),
       delaySeconds: (_data, matches) => parseFloat(matches.castTime),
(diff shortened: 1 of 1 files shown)


Legends0:
Ah, actually need to make modification, the headmarkers and the start casting on first protean are same time.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -69,6 +70,32 @@ const triggerSet: TriggerSet<Data> = {
...
+          return output.protean!();
+        return output.healerGroupsProtean!();
+      },
+      run: (data) => data.secondExocleavers = true,
+      outputStrings: {
+        healerGroupsProtean: {
+          en: 'Healer Groups + Protean',
+        },
+        protean: {
+          en: 'Protean',
```
quisquous, line 90: This is kind of an extreme nit here about wording, but I think this is not really a protean in the sense that the boss is shooting a conal at every player.  When I think protean, I think "people need to be in clock positions".  I think if you wanted to prompt the players here, you could do a "Healer Groups" but then do a Responses.moveAway() after the first hit of the excocleaver to remind people not to plant.

Legends0: Makes sense.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -69,6 +70,25 @@ const triggerSet: TriggerSet<Data> = {
...
+      run: (data) => data.secondExocleavers = true,
+      outputStrings: {
+        healerGroups: Outputs.healerGroups,
+      },
+    },
+    {
+      id: 'P6S Exocleaver Move',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: ['7869', '786B'], source: 'Hegemone' }),
+      delaySeconds: (_data, matches) => parseFloat(matches.castTime),
```
quisquous, line 89: `castTime` in the log line is 0.3s shorter than the actual cast time.  Do you think it'd be safer to use `Ability` instead of `StartsUsing` with a `suppressSeconds`?

Legends0: I think it snapshots on the cast, but it doesn't hurt to change. Also looking at some logs, it might always be 786B + 786A x6 followed by 786B x6, so might be possible to update timeline and remove the 7869 trigger match and just use 786B without supressSeconds.

quisquous: Yeah, I'm just saying that `parseFloat(matches.castTime)` is not the end of the cast, it's 0.3s before the end of the cast.  If you look at a log file with a cast and an ability usage, the cast time is always 0.3s shorter than when the ability shows up.

PR #4771, opened 2022-09-02
raidboss: p6 fix Unholy Darkness spellid

Legends0:
This timing is tied to 013E stackmarkers with Unholy Darkness 7892


JLGarber:
Ah, thank you, I was building this one without having been in myself, so separating those out was a pain. Good to know exactly why `test_timeline` was complaining there until I threw everything in.

PR #4772, opened 2022-09-02
raidboss: p6s predation bait triggers

Legends0:
It calls out where to go (left/right) and order of your bait. There is a collection trigger here that can be used to create additional triggers for the other mechanics related to the debuffs.

EDIT: Now also has In reminder and out added for baits.


ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -190,6 +192,67 @@ const triggerSet: TriggerSet<Data> = {
...
+          16.00: output.fourthBait!(),
+        };
+        const dir = data.predationDebuff === 'CF7' ? output.left!() : output.right!();
+        return output.text!({ dir: dir, bait: AetheronecrosisMap[parseFloat(matches.duration)] });
+      },
+      outputStrings: {
+        text: {
+          en: '${dir}, ${bait}',
+        },
+        left: Outputs.left,
```
quisquous, line 230: Just for super clarity, can you say "Left (Wing Side)" and "Right (Snake Side)", both here and in Ptera Ixou? (I think that's which side is which, please double check.)

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -190,6 +192,67 @@ const triggerSet: TriggerSet<Data> = {
...
+      durationSeconds: (_data, matches) => {
+        const duration = parseFloat(matches.duration);
+        return duration === 20 ? duration - 0.1 : duration + 16 - 0.1;
+      },
+      infoText: (data, matches, output) => {
+        const AetheronecrosisMap: { [duration: number]: string } = {
+          20.00: output.firstBait!(),
+          8.00: output.secondBait!(),
+          12.00: output.thirdBait!(),
+          16.00: output.fourthBait!(),
```
quisquous, line 221: I am not sure if everybody is doing this, but that's definitely the strat that I used too.  Do you think it'd be worth calling "bait near" or something when it's your turn to, after the aoes go off?

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -190,6 +192,67 @@ const triggerSet: TriggerSet<Data> = {
...
+        return output.text!({ dir: dir, bait: AetheronecrosisMap[parseFloat(matches.duration)] });
+      },
+      outputStrings: {
+        text: {
+          en: '${dir}, ${bait}',
+        },
+        left: Outputs.left,
+        right: Outputs.right,
+        firstBait: {
+          en: 'First Bait',
```
quisquous, line 233: Can you put the time in here too? e.g. "20: First Bait" or "8: Second Bait" or "First Bait (20s)" or something? Mostly because I know people can use the times for positioning and so I think both things are important.

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: Yea, I was going to write an in at the time to go in and then out after the 4s/hit, but never got around to it.

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: Yea, I think that is correct.

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: I'll go with time in parenthesis as is looks cleaner with direction (wing/snake side), bait (time)

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: Regarding strat, I think it's only possible to do baits in this order due to vulnerability up.

Legends0: I have added an in trigger and out trigger.

In trigger for first bait is based on the Dual Predation cast time, but realistically they could be in as soon as they get their debuff.

The other triggers are based on the hits, which means it is safe for the player to move in without getting aoe'd by the previous predation. This comes at a risk that the trigger will still fire if the player has a vulnerability up from making a mistake...

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -190,6 +200,143 @@ const triggerSet: TriggerSet<Data> = {
...
+      delaySeconds: 0.1,
+      durationSeconds: (_data, matches) => {
+        const duration = parseFloat(matches.duration);
+        // First Dual Predation is 3.7s before expiration
+        // Remaining Dual Predations are 12.3s (second), 12.4s (third/fourth)
+        return duration === 20 ? duration - 3.8 : duration + 12.3;
+      },
+      infoText: (data, matches, output) => {
+        const AetheronecrosisMap: { [duration: number]: string } = {
+          20.00: output.firstBait!(),
```
quisquous, line 229: This is not safe, sometimes log lines get written out slightly late and the effect has already started, e.g. this could be `19.98` or something and not `20.00`.  I think you need to do something more like https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts#L701-L716 where you do a more rough comparison against times.  Sorry to make it more complicated, just trying to safeguard against issues I've seen in the past.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -190,6 +200,143 @@ const triggerSet: TriggerSet<Data> = {
...
+          en: 'In (Fourth Bait)',
+        },
+      },
+    },
+    {
+      id: 'P6S Predation Out',
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: ['787A', '787B'], source: 'Hegemone' }),
+      condition: Conditions.targetIsYou(),
+      alertText: (_data, _matches, output) => output.out!(),
```
quisquous, line 321:
```suggestion
      infoText: (_data, _matches, output) => output.out!(),
```

Maybe infoText to go along with "move" and to not have everything be alerts back to back here.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -190,6 +200,143 @@ const triggerSet: TriggerSet<Data> = {
...
+        };
+        const dir = data.predationDebuff === 'CF7' ? output.left!() : output.right!();
+        return output.text!({ dir: dir, bait: AetheronecrosisMap[parseFloat(matches.duration)] });
+      },
+      outputStrings: {
+        text: {
+          en: '${dir}, ${bait}',
+        },
+        left: {
+          en: 'Left (Wing Side)',
```
panicstevenson, line 242: Is the boss always facing North here? Should these be West/East instead?

Legends0: The boss always faces North, so whether relative to boss facing or true north, both will work. I just tend to think boss side to look at.

ui/raidboss/data/06-ew/raid/p6s.ts
Legends0: Thanks for the heads up. I've modified it to compare to <= duration and > 16 where 20 is used.

quisquous:
Thanks for the triggers and all the changes.  It's super appreciated.

ui/raidboss/data/06-ew/raid/p6s.ts
quisquous: If you want to follow up and rename, go for it.

PR #4780, opened 2022-09-03
raidboss: p7s Bough of Attis (arrows) callout

Legends0:
Since the Close and Far ones are static, they may be better done with a timeline trigger for pre-positioning.


quisquous:
This is a tiny nit, but do you think this is worth making "left" or "right" since everybody is always south before these two arrows, facing north?

Legends0:
> This is a tiny nit, but do you think this is worth making "left" or "right" since everybody is always south before these two arrows, facing north?

I think it does work to say left/right. I'll update it since it also has response for it.

PR #4781, opened 2022-09-03
raidboss: p7s Roots of Attis Triggers

Legends0:
These provide insight into which platform players should choose as the Roots of Attis spell will be breaking bridges between platforms.

The second and third occurrence will certainly lead to deaths if the wrong platforms are chosen prior to the cast finishing. I'm not sure if they should be upgraded to an alert.


Legends0:
Probably needs some sort of delay for the first one as it is real early.

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -47,6 +48,43 @@ const triggerSet: TriggerSet<Data> = {
...
     },
+    {
+      // First breaks north bridge for upcoming South Knockback Spreads
+      // Second breaks remaining bridges, Separate Healer Groups
+      // Third breaks all bridges, Bait on Empty Platform
+      id: 'P7S Roots of Attis 1',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '780E', source: 'Agdistis', capture: false }),
+      condition: (data) => data.rootsCounter === undefined,
+      delaySeconds: 0.1, // Slight delay to prevent duplicate callout
```
quisquous, line 59: Sorry, can you explain what this is doing? Is it possible to do things in a more deterministic way without delays?

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -47,6 +48,43 @@ const triggerSet: TriggerSet<Data> = {
...
+    },
+    {
+      id: 'P7S Roots of Attis 3',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '780E', source: 'Agdistis', capture: false }),
+      condition: (data) => data.rootsCounter === false,
+      infoText: (_data, _matches, output) => output.baitSoon!(),
+      outputStrings: {
+        baitSoon: {
+          en: 'Bait on Empty Platform Soon',
```
quisquous, line 84: Is it possible to have this call out which platform to go to?

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -10,6 +10,7 @@ import { TriggerSet } from '../../../../../types/trigger';
 export interface Data extends RaidbossData {
   purgationDebuffs: { [role: string]: { [name: string]: number } };
   purgationDebuffCount: number;
+  rootsCounter?: boolean;
```
quisquous, line 13: This is a little bit hard for me to follow.  It's named "counter" but you're checking against both `false` and `undefined` and `true` separately?

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -47,6 +48,43 @@ const triggerSet: TriggerSet<Data> = {
...
+      id: 'P7S Roots of Attis 2',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '780E', source: 'Agdistis', capture: false }),
+      condition: (data) => data.rootsCounter,
+      delaySeconds: 0.1, // Slight delay to prevent duplicate callout
+      infoText: (_data, _matches, output) => output.separateHealerGroups!(),
+      run: (data) => data.rootsCounter = false,
+      outputStrings: {
+        separateHealerGroups: {
+          en: 'Separate Healer Groups',
```
quisquous, line 72: Maybe:
```suggestion
          en: 'Healer Group Platforms',
```

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: It's just a way to count to three with a boolean (1 byte) rather than int, but it is less intuitive. Alternatively it could be renamed and just flip odd/even with the boolean.

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: The spell is the same id for each event. I'm not sure how cactbot reads trigger order, but when it was ordered as is, the first trigger would fire, set the value, second trigger fire sees the updated value, leading to multiple triggers firing "simultaneously".

It might be possible to reverse the order in the file (3 => 2 => 1) that way the 1 trigger is read last.

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: I'm not sure. Visually it is because you see which platforms have an egg spawn on them. I have not checked which npcIds spawn on which platforms, when that is logged, and if OverlayPlugin is needed and can obtain earlier data.

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: I've re-ordered it, looks like the reorder of 3, 2, 1 solves the problem.

quisquous: Trigger order is deterministic, top to bottom.  The only thing that changes ordering is any asynchronicity like delaySeconds or promise.

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: I changed the name to be more understandable as secondRoots. Still utilizing the undefined for the first roots.

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: I looked into it some more, there are Forbidden Fruit npcs that I am guessing are used, they map 1:1 with what I've seen for the Immature npcs.

11377 => Immature Minotaur (11380)
11376 => Immature Stymphalide (11379)
11375 => Immature Io (11378)

Legends0: Checking on spell cast for '7811', the forbidden fruits haven't moved. FFlogs seems to catch them as Agdistis. I might try adding some delay as I do see later '7811' casts have updated positions.

Legends0: A delay of 0.5s is enough to grab the data. It's probably available slightly sooner, but for latency reasons I think 0.5s has worked well.

Legends0: It seems either the delay is not consistent between phases, or the Forbidden Fruit combatant data is not what we see visually. I tried looking at Forbidden Fruit 7 here and the position data I am getting is not accurate. While the positions do change, they are not where the birds spawn. I might try and check the Agdistis entities next.

Legends0: Something I did just notice, the eggs do not appear until after the 7811 spell completes, so maybe that too could have been the problem.

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -101,6 +102,41 @@ const triggerSet: TriggerSet<Data> = {
...
+      },
+    },
+    {
+      // First breaks north bridge for upcoming South Knockback Spreads
+      // Second breaks remaining bridges, Separate Healer Groups
+      // Third breaks all bridges, Bait on Empty Platform
+      id: 'P7S Roots of Attis 1',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '780E', source: 'Agdistis', capture: false }),
+      condition: (data) => data.secondRoots === undefined,
```
quisquous, line 137: Sorry for the nit, but I think it makes it more confusing to follow when you check both `data.secondRoots` and `data.secondRoots === undefined`.  Personally, I think if you have an optional boolean, then undefined/false should be equivalent.  Could you just make this a roots counter and increment when you see roots? I think this would make it easier to read.

Legends0: Ok, I've updated it to be a number.

ui/raidboss/data/06-ew/raid/p7s.ts
Legends0: I have it working now, just checking for delay. Egg data is correct by 1s after 7811 ability.

PR #4784, opened 2022-09-03
raidboss: p7s Inviolate Bonds/Purgation Reminders

Legends0:
Adds a headmarker-based callout for the Inviolate Bonds and Inviolate Purgation mechanics utilizing previous triggers. **EDIT:** This effectively gives a 6s notice, which by then the player should already be headed to or on the platform.

Note: Previous triggers are relying on 4/4 support/dps party and purgation relies on 8-player party, so these may not function properly in non-standard party...

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p7s.ts
+++ b/ui/raidboss/data/06-ew/raid/p7s.ts
@@ -10,7 +10,7 @@ import { TriggerSet } from '../../../../../types/trigger';
 
 export interface Data extends RaidbossData {
   decOffset?: number;
-  previousBondsDebuff?: string;
+  bondsDebuff?: string;
   purgationDebuffs: { [role: string]: { [name: string]: number } };
   purgationDebuffCount: number;
   purgationEffects?: string[];
@@ -146,7 +146,7 @@ const triggerSet: TriggerSet<Data> = {
         };
 
         // Store debuff for reminders
-        data.previousBondsDebuff = matches.effectId;
+        data.bondsDebuff = (matches.effectId === 'CEC' ? 'spread' : 'stackMarker');
 
         const longTimer = parseFloat(matches.duration) > 9;
         if (longTimer)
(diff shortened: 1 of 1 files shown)


Legends0:
I do not see what currently would prevent a non-standard party from completing this, so maybe the original triggers should be updated to support that. Also this currently does not alter the durationSeconds of the original triggers.

Legends0:
Callout could say 'new platform' for the subsequent events in addition to what it has now for the ones that require switching platforms.

quisquous:
> Note: Previous triggers are relying on 4/4 support/dps party and purgation relies on 8-player party, so these may not function properly in non-standard party...

Yes, I agree.  I think it will work in most cases, but it would definitely be more robust if it only checked your personal debuffs.  Then you could even say "stack (no debuff)" and "stack (on you)" kinda thing too.

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -134,19 +138,47 @@ const triggerSet: TriggerSet<Data> = {
...
       durationSeconds: 20,
-      response: (_data, matches, output) => {
+      response: (data, matches, output) => {
         // cactbot-builtin-response
         output.responseOutputStrings = {
           stackThenSpread: Outputs.stackThenSpread,
           spreadThenStack: Outputs.spreadThenStack,
         };
 
+        // Strore debuff for reminders
```
quisquous, line 148:
```suggestion
        // Store debuff for reminders
```

ui/raidboss/data/06-ew/raid/p7s.ts
```diff
@@ -134,19 +138,47 @@ const triggerSet: TriggerSet<Data> = {
...
+      // These happen 6s before cast
+      type: 'HeadMarker',
+      netRegex: NetRegexes.headMarker({}),
+      suppressSeconds: 1,
+      infoText: (data, matches, output) => {
+        const correctedMatch = getHeadmarkerId(data, matches);
+        if (correctedMatch === '00A6' && data.purgationDebuffCount === 0) {
+          switch (data.previousBondsDebuff) {
+            case 'CEC':
+              data.previousBondsDebuff = 'D45';
```
quisquous, line 169: Why do you need to set this again here?

Rather than using effect ids, could `P7S Inviolate Bonds` just set some `data.finalBondsDebuff = 'spread';` and then this trigger could read it? That seems a little bit more straight forward than remembering the previous as an effect id and then having to switch it here.

Legends0: I think it's leftover from an original thought to use this same variable for purgation, but I instead used effects there since it had what I needed.

Legends0: The trigger is used for both the first debuff and the second debuff, so it needs to check which output to use.

PR #4791, opened 2022-09-04
raidboss: p7s add chaser baits

Legends0:


Legends0:
Maybe Away from Bridge and/or get together would be a better phrase to say?

panicstevenson:
This feels kind of not necessary? Maybe if we called which direction you'd be going? Unsure, but it feels like you're just dodging a floor AoE here (plus a bird)

quisquous:
Is this already covered by #4781 ?

Legends0:
It is a later call out., but I can agree that it can be handled by adjusting what #4781, or even the callout in the egg collection. I may this (after closing the pr) plus delay or timing of the aoe Puddle if I can get that to call out the platform to run towards.

PR #4793, opened 2022-09-04
raidboss: p7s Egg Tracking + Forbidden Fruits 1,4,6-10

Legends0:
The eggs spawn after ability 7811 completes, but it takes time for the data to be available. I tested with 0.5s and 1s, both seemed to work. It may be available shorter than 0.5s, but latency wise 0.5s is safer than a lower number.

Currently calls out Left/Right for Forbidden Fruit 1,  then the two platforms for Forbidden Fruit 6, and Left/Right or South for Forbidden Fruit 7.


Legends0:
It's simple enough to test out Forbidden Fruit 1. This is not tested for Forbidden Fruit 7, but I think at least the data is correct? Need to confirm the birds that are used in Forbidden Fruit 7.

Legends0:
The forbidden fruit 6 I added isn't always correct...

valarnin:
I don't suppose you know which eggs are being used for 8/9/10 (harvests)?

Legends0:
> I don't suppose you know which eggs are being used for 8/9/10 (harvests)?
> 
> Edit:
> 
> Nevermind, looks like it's a 1:1 mapping of fruit ID = immature ID + 1, e.g. fruit `4004BC65` = immature add `4004BC64`.

I'll have to check my logs, but I think I have the data logged for it, I'm just not too sure what to do with it. Also, I have not checked if the headings of the eggs matches the birds when they spawn. Was just going to have a hatchedEggs array for those.

Legends0:
Oh yea, if it happens to be simpler such ad take the id of hatchedegg us always just 1 less than it's fruit, then I can just extrapolate what the egg was. I have all harvest positions logged with their ID from the tethers PR. I guess that would also mean no need for index other than it beats search times...

valarnin:
Looks pretty accurate to me.

image

Legends0:
I think for FF4 a callout at eggs for the bull's position could be helpful for orientation as non tethers need to Bait cleave opposite, between the minotaurs, and others would know where bull is if they get bull tether.

For Death, orientation around the bull on the platform is used in conjunction with role, so a callout for where that bull egg is could be helpful for preposition.

Legends0:
War Harvest as a few strats. The two prevailing strays I see will orient based on the bird between the two minotaurs and have each tether go to their own circle platform, the others just have players go directly across from their tether.

For death, there are also two static position strats (although the original I think was incorrect for one of the patterns). These still orient based on the bull on the platform though so seems that one should call that egg.

Legends0:
It seems to work for forbidden fruit 4. I think two of them it is incorrect and the last I would need to check again.

Legends0:
Ok, I realize I can just use the birds for Death as well to reduce code. For War, I realize I based it on the birds on accident but should have used the Minotaurs.

The minotaurs spawn a these locations in War, and from the two minotaurs could get the platform between them:
(107.14, 104.13) (4) SE Minotuar
(100, 91.75) (1) N Minotaur
(92.86, 104.13) (6) SW Minotaur

Legends0:
Ok I think I fixed up the issues while keeping what was working still working. Will have time to try it out some more later on in the week or next week.

PR #4795, opened 2022-09-05
i18n: p6s & p7s FR translations

mooondark:


mooondark:
This bot is so strict !

PR #4800, opened 2022-09-06
i18n: missing p7s and p8s translations

Akurosia:


quisquous:
(Can you fix this test issue?)

PR #4805, opened 2022-09-08
config: New skin for timeline and alert text

Jwidea:
I use this with my material-ui and delvui. I wanna to have something similar to Lippe but closer to what I have, so maybe there are ppl want to have sth like this as well.

unknown2


quisquous:
Thanks!!

PR #4811, opened 2022-09-09
raidboss: Add Option to Order Timeline Bottom-to-Top

panicstevenson:


panicstevenson:
Small demo gif:
Demo

panicstevenson:
Converting to draft since there's a small bug with KeepExpiredTimerBarsForSeconds not being honored if the timerbar is pushed out of the window.

panicstevenson:
To expand upon the above: I believe the bug is instead the fact that keepAlive bars are not actually a part of the grid. I think there needs to be a small refactor to make that work more in line with expectations before landing anything here.

quisquous:
> To expand upon the above: I believe the bug is instead the fact that keepAlive bars are not actually a part of the grid. I think there needs to be a small refactor to make that work more in line with expectations before landing anything here.

I ended up doing this refactoring in #5869.

panicstevenson:
To the two daredevils that eagerly upthumb'd the original post, your nearly 14-month wait is almost over.

PR #4813, opened 2022-09-09
i18n: Corrected one p6s machanic CN translation.

richackard:
Corrected the Chinese translation of p6s `Synergy` mechanic.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -80,7 +80,7 @@ const triggerSet: TriggerSet<Data> = {
           de: 'Geteilter Tankbuster',
           fr: 'Séparez les Tankbusters',
           ja: '2人同時タンク強攻撃',
-          cn: '双T分散死刑',
+          cn: '分散死刑',
           ko: '따로맞는 탱버',
         },
       },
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -80,7 +80,7 @@ const triggerSet: TriggerSet<Data> = {
           de: 'Geteilter Tankbuster',
           fr: 'Séparez les Tankbusters',
           ja: '2人同時タンク強攻撃',
-          cn: '坦克分摊死刑',
+          cn: '双T分散死刑',
```
MaikoTan, line 83: I think the original one is enough for readability, is there any reason to change it to "雙T" instead?

ShadyWhite, line 83:
```suggestion
          cn: '双T死刑',
```

ShadyWhite: I think '双T死刑' is enough in this situation.

ShadyWhite (changes requested):

ui/raidboss/data/06-ew/raid/p6s.ts
MaikoTan: Suggest: "分散死刑" by @Echoring as it is the only word that tanks should remember.

MaikoTan:
@Akurosia Hey bro, I don't know deutsch but it looks like there are something wrong in the deutsche translation of this trigger either, which should announce tank to take the tank busters separately, but currently it looks like saying sharing?

PR #4818, opened 2022-09-09
i18n: update ko p7s, p8s translate

Bing-su:
Thanks always for the updates


Bing-su:
#4817 I'll add it when it's merged

PR #4821, opened 2022-09-10
raidboss: p8s Blazing Footfalls directions

Legends0:
Adds a callout for all and a reminder.

It looked like the targetX and targetY data of the 793D and 793C spells could be used to resolve this without OverlayPlugin.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -1077,8 +1077,13 @@ const triggerSet: TriggerSet<Data> = {
             en: '${dir}',
           },
           trailblazeKnockback: {
+            en: '${dir} Knockback',
+          },
+          trailblazeKnockbackTo: {
             en: '${dir1} Knockback to ${dir2}',
           },
+          left: Outputs.left,
+          right: Outputs.right,
           north: Outputs.north,
           east: Outputs.east,
           south: Outputs.south,
@@ -1137,13 +1142,35 @@ const triggerSet: TriggerSet<Data> = {
 
           // Second trailblze should call torch location
           if (data.footfallsOrder[data.trailblazeCount] === 'impact') {
-            // If missing torch data, target to the middle
-            const knockbackTo = (data.trailblazeTorchSafeZone ? output[data.trailblazeTorchSafeZone]!() : dirToCard[(dir + 2) % 4]);
-            return { alertText: output.trailblazeKnockback!({ dir1: dirToCard[dir], dir2: knockbackTo }) };
+            if (
+              (data.trailblazeTorchSafeZone === 'east' && dir === 0) ||
+              (data.trailblazeTorchSafeZone === 'west' && dir === 2)
+            )
+              return { alertText: output.trailblazeKnockbackTo!({ dir1: dirToCard[dir], dir2: output.left!() }) };
+            if (
+              (data.trailblazeTorchSafeZone === 'west' && dir === 0) ||
+              (data.trailblazeTorchSafeZone === 'east' && dir === 2)
+            )
+              return { alertText: output.trailblazeKnockbackTo!({ dir1: dirToCard[dir], dir2: output.right!() }) };
+
+            // Unable to determine direction, output only knockback
+            return { alertText: output.trailblazeKnockback!({ dir1: dirToCard[dir] }) };
           }
+
           // Output torch location during crush
-          if (data.footfallsOrder[data.trailblazeCount] === 'crush' && data.trailblazeTorchSafeZone)
-            return { infoText: output[data.trailblazeTorchSafeZone]!() };
+          if (data.footfallsOrder[data.trailblazeCount] === 'crush' && data.trailblazeTorchSafeZone) {
+            if (
+              (data.trailblazeTorchSafeZone === 'east' && dir === 0) ||
+              (data.trailblazeTorchSafeZone === 'west' && dir === 2)
+            )
+              return { infoText: output.right!() };
+            if (
+              (data.trailblazeTorchSafeZone === 'west' && dir === 0) ||
+              (data.trailblazeTorchSafeZone === 'east' && dir === 2)
+            )
+              return { infoText: output.left!() };
+            // Unable to determine direction, no output
+          }
         }
       },
       run: (data) => data.trailblazeCount++,
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -919,7 +919,7 @@ const triggerSet: TriggerSet<Data> = {
             en: '${dir}',
           },
           directionsKnockback: {
-            en: '${dir} Knockback',
+            en: 'Knockback ${dir1} to ${dir2}',
           },
           north: Outputs.north,
           east: Outputs.east,
@@ -958,8 +958,10 @@ const triggerSet: TriggerSet<Data> = {
           if (data.trailblazeCount === 0 && data.footfallsOrder[data.trailblazeCount] === 'crush')
             return { infoText: output.directionsPush!({ dir: dirToCard[data.footfallsDirs[1]] }) };
           // Call Knockback direction if Impact
-          if (data.footfallsOrder[data.trailblazeCount] === 'impact')
-            return { alertText: output.directionsKnockback!({ dir: dirToCard[dir] }) };
+          if (data.footfallsOrder[data.trailblazeCount] === 'impact') {
+            const knockbackTo = (dir + 2) % 4;
+            return { alertText: output.directionsKnockback!({ dir1: dirToCard[dir], dir2: dirToCard[knockbackTo] }) };
+          }
         }
       },
       run: (data) => data.trailblazeCount++,
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -1012,7 +1012,7 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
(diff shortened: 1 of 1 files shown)


Legends0:
I forgot, but this does not change the first crush/impact, that I think would need it's own trigger and if it's anything like Cataclysm was in e12s, it will need OverlayPlugin to get accurate data.

Legends0:
Strange how the raidsimulator has this only outputing once, but in game it duplicates.

Legends0:
This probably also needs a reminder callout such as ${dir} Knockback if it's impact between the trailblaze since the stack mechanic after first trailblaze can mean players are not centered.

Legends0:
May also be better if callouts are relative to center...

Legends0:
Ok, so I think the callouts are at least correct, but the wording and/or timing of the callouts may still need improvement.

quisquous:
If it's helpful, the callouts I make are:
* black line to C (for initial partner/light party stack)
* knockback C to A
* black line to A
* run right/left

OR

* black line to C (for initial partner/light party stack)
* WAIT => line to A
* black line to A (repeated, once crush goes off)
* knockback left/right

(obviously A/C here are just cardinal directions, but that's what I call out)

Legends0:
I like the adding black line emphasis. Early on I see a lot of players mistake which line to be at.
After several calls I feel that emphasis on the direction first is important. There is an additional reminder for after the push if they need to adjust for subsequent knockback/trailblaze and I think there is time to do so.

I do like adding the concept into the initial call.

I haven't done any work on the torch flames, but I could look at that or leave it for someone else to add.

Legends0:
If this used OverlayPlugin (since now I guess the required delay is no longer there) it may sound better during the sequence of events if used as there won't be as much of a delay although that would be more difficult to implement.

Legends0:
Some things todo:

* Sooner calls via OverlayPlugin during the tells by capturing heading of the invisible combatant, currently relies on AOEAbility regex which is not available until after the tell dissapears
* Calls black line to go to while knockback/crush is still happening which can cause player to inadvertantly clip; Either a delay may be warranted or some other cautionary message such as `Wait => ${dir} Black Line`
* Knockback second direction should check future line location in first trailblaze and use it's direction as target, for torches it would need torch data which is to be done
* There can be a duplicate call at the end.

Legends0:
Ok I added basic torch callout here. It could be acted more on to get relative to the knockback position and/or opposite the crush spot so that it stated left/right (as in the todo), but I think this works for the moment.

Legends0:
Some updates regarding use of OverlayPlugin. The invisible combatant does not turn 0.3s into the cast of the Impact/Crush, the cast is still 1.7s long, but It might actually only be turning when it 'teleports' to the target location, which would mean we couldnt do much with the heading data. We may not be able to get earlier call outs like I had thought.

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -941,6 +960,221 @@ const triggerSet: TriggerSet<Data> = {
...
+            // Only need to call this out if there is an upcoming pushback
+            if (data.footfallsOrder[data.trailblazeCount] === 'crush')
+              return { infoText: output.trailblaze!({ dir: dirToCard[data.footfallsDirs[1]] }) };
+
+            // Call future push location if knockback
+            if (data.footfallsOrder[data.trailblazeCount] === 'impact')
+              return { infoText: output.trailblazeKnockback!({ dir1: dirToCard[dir], dir2: dirToCard[data.footfallsDirs[1]] }) };
+          }
+
+          // First trailblaze may require moving to new cardinal during Crush/Impact
```
kshman, line 1131: First trailblaze again here 😊

Legends0: Oh thanks for that. Did not notice I pasted that wrong.

Legends0:
Hmmm, I think if it ends in crush, the left/right is wrong.

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -941,6 +960,201 @@ const triggerSet: TriggerSet<Data> = {
...
+        // cactbot-builtin-response
+        output.responseOutputStrings = {
+          trailblaze: {
+            en: '${dir}',
+          },
+          trailblazeKnockback: {
+            en: '${dir} Knockback',
+          },
+          trailblazeKnockbackTo: {
+            en: '${dir1} Knockback to ${dir2}',
```
quisquous, line 1083: This is super nitpicky, but it sounds a little bit weird to my ears to say "south knockback to right" vs "south knockback right". Would you mind having a `trailblazeKnockback`, a `trailblazeKnockbackToDir`, and a `trailblazeKnockbackSide` output string here, so the left/right doesn't get the word "to" in there?

Feel free to disagree (maybe this is just something I should customize personally), but I also feel like the final knockback doesn't need a dir1 in it and could just say "Knockback Right" or "Knockback Left".  It should be pretty clear that you have just gotten knocked back to south, that the knockback is coming from south.

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -941,6 +960,201 @@ const triggerSet: TriggerSet<Data> = {
...
+          // First trailblaze may require moving to new cardinal during Crush/Impact
+          if (data.trailblazeCount === 0) {
+            // Call move to next push back side if Crush
+            // Only need to call this out if there is an upcoming pushback
+            if (data.footfallsOrder[data.trailblazeCount] === 'crush')
+              return { infoText: output.trailblaze!({ dir: dirToCard[data.footfallsDirs[1]] }) };
+
+            // Call future push location if knockback
+            if (data.footfallsOrder[data.trailblazeCount] === 'impact')
+              return { infoText: output.trailblazeKnockback!({ dir1: dirToCard[dir], dir2: dirToCard[data.footfallsDirs[1]] }) };
```
quisquous, line 1128: This output string doesn't take dir1 and dir2.  I thought there was a test for this, but maybe it doesn't work for responses.

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -941,6 +960,201 @@ const triggerSet: TriggerSet<Data> = {
...
+          }
+
+          // Second trailblze should call torch location
```
quisquous, line 1131:
```suggestion
          // Second trailblaze should call torch location
```

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -941,6 +960,201 @@ const triggerSet: TriggerSet<Data> = {
...
+          // Dir is flipped for crush, thus matching knockback direction
+          if (
+            (data.trailblazeTorchSafeZone === 'east' && dir === 0) ||
+            (data.trailblazeTorchSafeZone === 'west' && dir === 2)
+          ) {
+            if (data.footfallsOrder[data.trailblazeCount] === 'impact')
+              return { alertText: output.trailblazeKnockbackTo!({ dir1: dirToCard[dir], dir2: output.left!() }) };
+            if (data.footfallsOrder[data.trailblazeCount] === 'crush')
+              return { infoText: output.left!() };
```
quisquous, line 1140: Can this say "Run Left" to differentiate from "Knockback Left"?

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -412,6 +421,16 @@ const triggerSet: TriggerSet<Data> = {
         if (safe0 === undefined)
           return;
 
+        // Blazing Foothills will have 4 safe spots
+        // However, it will only be East or West
+        if (data.trailblazeCount === 1) {
+          if (safe2 === 'outsideEast')
+            data.trailblazeTorchSafeZone = 'east';
+          if (safe2 === 'outsideWest')
+            data.trailblazeTorchSafeZone = 'west';
```
quisquous, line 430: I agree that what you have is correct, but I think it's better to just check everything for safety.  The other option is to filter out corners from safeKeys and check if there's only one key and it's outside east/west.

```suggestion
          if (safeKeys.length !== 3)
            return;
          if (safe0 === 'cornerNE' && safe1 === 'cornerSE' && safe2 === 'outsideEast')
            data.trailblazeTorchSafeZone = 'east';
          if (safe0 === 'cornerNW' && safe1 === 'cornerSW' && safe2 === 'outsideWest')
            data.trailblazeTorchSafeZone = 'west';
```

quisquous (commented):
Looks good in general.  Just have a few wording changes.

ui/raidboss/data/06-ew/raid/p8s.ts
Legends0: >Would you mind having a trailblazeKnockback, a trailblazeKnockbackToDir, and a trailblazeKnockbackSide output string here, so the left/right doesn't get the word "to" in there?

I think I understand what you're asking for here.

It's likely not recoverable if the player is pushed back to the wrong side during the second trailblaze as they are very likely to have used their immunity for the earlier mechanic... although I haven't even tested that this knockback from the wall can be surecasted/armslengthed (It looks like it can?).

quisquous: I didn't even think you could use immunity on this.

Really my main point is that: "knockback to left" is awkward to my ears; it should be "knockback left" (preferable) or "knockback to the left" (too long, in my opinon).  So, maybe differentiate if impact is the final mechanic ("knockback left") vs the penultimate one ("W knockback to S") via different output strings like `trailblazeKnockbackSide` and `trailblazeKnockbackToDir`, respectively.  I guess you could drop the "to" on the "N/E/S/W" one, but that'd certainly sound awkward if you replaced N/E/S/W with marker names? Not sure.

quisquous:
Thank you!!

PR #4822, opened 2022-09-10
raidboss: p8s initial crush/impact direction

Legends0:
Adds direction based on OverlayPlugin Heading.

Current delay is what was safe for e12s, it may be possible that the boss here turns sooner...

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -12,7 +12,6 @@ import { TriggerSet } from '../../../../../types/trigger';
 
 // TODO: call out gorgon spawn locations
 // TODO: call out shriek specifically again when debuff soon? (or maybe even gaze/poison/stack too?)
-// TODO: crush/impact directions during 2nd beast phase
 // TODO: make the torch call say left/right during 2nd beast
 // TODO: better vent callouts
 // TODO: initial tank auto call on final boss as soon as boss pulled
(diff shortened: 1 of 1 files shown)


Legends0:
ok this misses because of race condition it seems. The collector has the right direction, but the other triggers fire before it gets the data.

Legends0:
Using `${dir} (Away)` may be confusing... Perhaps it should just say the direction?

Legends0:
Apparently 0.5s is sometimes not enough time for the boss to finish spinning...

quisquous:
> Apparently 0.5s is sometimes not enough time for the boss to finish spinning...

I think we need a release with https://github.com/OverlayPlugin/OverlayPlugin/pull/50 in it to make this better.

That said, personally speaking, despite using markers and directions for many things, this boss is huge and I find "follow" and "away" the most important part of the call here because I think it's pretty obvious which way the boss is facing (vs taking a second to reorient to directions).  I think if you wanted to have directions, maybe "Follow to A" or "Away to C" seems like it has the more important information up front.

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -753,31 +753,86 @@ const triggerSet: TriggerSet<Data> = {
...
-      alertText: (_data, _matches, output) => output.text!(),
-      outputStrings: {
-        text: {
-          en: 'Follow Jump',
-          de: 'Sprung folgen',
-          ja: '近づく',
-          ko: '보스 따라가기',
-        },
+      netRegex: NetRegexes.startsUsing({ id: ['7A04', '7A05'], source: 'Hephaistos' }),
+      delaySeconds: 0.5,
```
quisquous, line 759: Would you mind testing if this delay can go away with the new version of OverlayPlugin? If we don't have to delay this call, I think it would be better.

Legends0: Yea, I'll need to get a few pulls in to get the worst culprit which is something like boss facing north turns 275 degrees (lol).

Legends0: I only had a few pulls in tonight to test it with no delay, did not notice any issues.

PR #4824, opened 2022-09-10
raidboss: p8s add hints about which tower to take

quisquous:


nsetton:
Not sure if you're aware but in HC2, the people who need to merge for towers are actually fixed. Long alpha and long beta will always have to merge (to make the wind concepts that need to fuse with ifrits at the end of the phase) and similarly long gamma and the unused debuff (can be either unused alpha or unused beta, but never unused gamma) will have to fuse to take the final set of towers. This makes that mechanic deterministic unlike HC1 where there's some randomness depending on which tower spawns.

quisquous:
> Not sure if you're aware but in HC2, the people who need to merge for towers are actually fixed. Long alpha and long beta will always have to merge (to make the wind concepts that need to fuse with ifrits at the end of the phase) and similarly long gamma and the unused debuff (can be either unused alpha or unused beta, but never unused gamma) will have to fuse to take the final set of towers. This makes that mechanic deterministic unlike HC1 where there's some randomness depending on which tower spawns.

My understanding was that the first tower was either water or lightning and that wind is always in the final set.  Is that not true?

nsetton:
So because there have to be two wind towers in the final set, you know that long alpha and long beta must always merge (to form wind). This also means that no matter which color the other two towers are (they can be water or lightning yes), the correct two people to merge are long gamma and whoever was unused from the very first set of alpha/beta/gamma debuffs. This means that in essence, you don't actually need to pay attention to the colors: the movement is static every time. Here are two diagrams to better illustrate (towers can obv be diagonal or left/right but this is just a matter of communicating who gets which): 
image
image

quisquous:
Yeah, I think there's a lot of room for improvement here.  This callout is still good for the previous three towers, though.

PR #4829, opened 2022-09-12
raidboss: p8s add End of Days first/second row

quisquous:
This is probably ok, but it's also like ~2s after the first visual so it feels late.


Legends0:
Worked pretty well for me.

PR #4830, opened 2022-09-13
raidboss: p8s add 2nd/3rd vent calls

quisquous:


quisquous:
Hmm, this sometimes gives me wrong answers, even with the new OverlayPlugin version.  I think it needs more investigation.

quisquous:
I think if somebody wants to investigate this more / faster than I can, the best bet is to record video and console log both `data.ventCasts` and `unsafeSpots` for each call.  It's possible there's a logic error here or it's a timing issue or something else is happening unexpectedly.  I won't get back into p8s until next week.

cc @Legends0 maybe if this is something you're interested in

Legends0:
I did try out the p7s egg tracker with no delay and I assume this behaves the same, if so you still need delaySeconds.

quisquous:
> I did try out the p7s egg tracker with no delay and I assume this behaves the same, if so you still need delaySeconds.

Was that before the OverlayPlugin change?

Legends0:
> > I did try out the p7s egg tracker with no delay and I assume this behaves the same, if so you still need delaySeconds.
> 
> Was that before the OverlayPlugin change?

I tested without delay both before and after the OverlayPlugin change, it continues to need the delay.

Legends0:
I can confirm this pr can do a wrong tell. I'll try with some delay.

Legends0:
It seems adding 0.3s delay makes it significantly worse.

Legends0:
Adding 0.5s delay seems to do the trick. Will try more samples this weekend.

quisquous:
Thanks for the testing.  I was wondering if I messed up the logic or it was just a timing issue.

Legends0:
Seems to still get wrong, with delay in another instance.

Legends0:
I am testing this out with a delay of 0.6s as that still seems workable. 0.5s sometimes still has an occasional incorrect call for me such as SW / SW. Might be due to the time it takes for heading to snap to correct direction.

Legends0:
So with 0.6s I have not seen the SW / SW calls, but I think it's still doing wrong calls on corners, like it called the opposite corners or something? Unfortunately, I do not have a vod of this.

Legends0:
I'm not sure if this is the case, but what can happen is that you need delay to get appropriate heading data but by adding the delay you could get positional data that matches a different location on the map.

Legends0:
Changed the later trigger to:
```
    {
      id: 'P8S Suneater Cthonic Vent Later',
      type: 'Ability',
      netRegex: NetRegexes.ability({ id: ['7923', '7924'] }),
      condition: (data, matches) => {
        data.ventCasts.push(matches);
        return data.ventCasts.length === 2;
      },
      delaySeconds: 0.5,
      promise: async (data: Data) => {
        data.combatantData = [];

        const ids = data.ventCasts.map((m) => parseInt(m.sourceId, 16));
        data.combatantData = (await callOverlayHandler({
          call: 'getCombatants',
          ids: ids,
        })).combatants;
      },
        if (data.combatantData.length !== 2)
          return;

        const unsafeSpots = [];
        for (const c of data.combatantData) {
          const originalPos = positionTo8Dir(c);
          const heading = headingTo8Dir(c.Heading);
          console.error(`Vent Later: Vent with heading ${heading} and position ${originalPos}`);
          // There's maybe some way to do this more generally, but I don't see it.
          // Also, if this fails for some reason, it will just not call anything below.
          if ((originalPos === 7 && heading === 2) || (originalPos === 3 && heading === 0) || (originalPos === 5 && heading === 1)) {
            // Going towards NE
            unsafeSpots.push(1);
          } else if ((originalPos === 1 && heading === 4) || (originalPos === 5 && heading === 2) || (originalPos === 7 && heading === 3)) {
            // Going towards SE
            unsafeSpots.push(3);
          } else if ((originalPos === 3 && heading === 6) ||  (originalPos === 1 && heading === 5) ||  (originalPos === 7 && heading === 4)) {
            // Going towards SW
            unsafeSpots.push(5);
          } else if ((originalPos === 5 && heading === 0) || (originalPos === 1 && heading === 6) || (originalPos === 3 && heading === 7)) {
            // Going towards NW
            unsafeSpots.push(7);
          }
        }

        return ventOutput(unsafeSpots, output);
      },
      run: (data) => data.ventCasts = [],
      outputStrings: ventOutputStrings,
    },
```

In other words, just took out the diagonal check (maybe it was wrong?) and treated the ids the same way.

Legends0:
I setup a PR that changes the code to use the same method of determining direction regardless of ability id for the later vents. I think that will make it easier to troubleshoot or may have fixed a problem in the diagonal case.

PR #4837, opened 2022-09-14
raidboss: p8s fix initial impact/crush

### LINKED ISSUE #4835: P8S initial crush/impact direction incorrect
When I ran last night, it said "Away to east".  The boss was facing north, then rotated counter-clockwise to face south and the call came midway through rotation.

Does the boss always face north or south? It kind of feels that way.  I think also the boss recenters to face north during Reforged Reflection before casting footprint.  Does that mean if it's no longer facing north then it's south?  Or are there other rotations here?

cc @Legends0

Legends0:
Originally needed delay to get correct data even after OverlayPlugin update, however it turns out this mechanic only has results of 0 or 1 and since the boss starts at 0, we can safely assume non-0 is south.

Fixes #4835

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -904,31 +904,17 @@ const triggerSet: TriggerSet<Data> = {
         data.crushImpactSafeZone = cardinal;
       },
       infoText: (data, matches, output) => {
-        const dirs: { [dir: number]: string } = {
-          0: 'north',
-          1: 'east',
-          2: 'south',
-          3: 'west',
-        };
-        if (data.crushImpactSafeZone === undefined)
-          return;
-
-        // BEGIN TEMPORARY HACK
-        // This trigger calls out the wrong directions sometimes, so disable until
-        // it can be fixed.
-        if (matches.id === '7A05')
-          return output.crush!();
-        else if (matches.id === '7A04')
-          return output.impact!();
-        // END TEMPORARY HACK
-
-        // Check if dir is valid, else output generic
-        const dir = dirs[data.crushImpactSafeZone];
-        if (dir === undefined) {
-          if (matches.id === '7A05')
+        if (data.crushImpactSafeZone === undefined) {
+          // Failed to get data, return generic result
+           if (matches.id === '7A05')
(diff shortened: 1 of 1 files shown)


Legends0:
One comment... it could use some further testing such as what the heading value is for North, if it's always a value, then we don't need to do atan2 and have the current risk that it captures a value that when mod 4 snaps to 0 (such as boss is more north than east/west in its turn).

quisquous:
> One comment... it could use some further testing such as what the heading value is for North, if it's always a value, then we don't need to do atan2 and have the current risk that it captures a value that when mod 4 snaps to 0 (such as boss is more north than east/west in its turn).

Yeah this was exactly my thought too.  Maybe just check the heading and see if it's "very close" to north rather than rounding to a cardinal.

Legends0:
He faces exact North, 3.14159274, each time it's north.

quisquous:
> He faces exact North, 3.14159274, each time it's north. Will check for > 3.14 I guess.

I think "greater than" is strictly wrong here, because it might turn one way and be >3.14 while turning.  Can you check for Math.abs(heading - 3.14) < epsilon or something, e.g. 0.1?

Legends0:
> > He faces exact North, 3.14159274, each time it's north. Will check for > 3.14 I guess.
> 
> I think "greater than" is strictly wrong here, because it might turn one way and be >3.14 while turning. Can you check for Math.abs(heading - 3.14) < epsilon or something, e.g. 0.1?

Ok, I'll change it to that.

When the boss was turning south, I recorded these values:
-2.14292955
-1.98264849
-2.38800049
-1.91409552

So at least in this small sample, it tends to turn negative.

Legends0:
hmm.. I don't think this quite works with epsilon as 0.1...

quisquous:
> So at least in this small sample, it tends to turn negative.

Yeah, that's fair.  Mostly I just don't trust ffxiv to be consistent for this sort of thing, and want to cover all the cases.

Legends0:
I think this can still call wrong "Away to North" when it should have been south.

Legends0:
Ah, I got what the problem is, epsilon needs to be larger because in that one case:
 Got heading of -3.13727832

PR #4841, opened 2022-09-17
raidboss: p8s Limitless Desolation Callouts

Legends0:
Adds three callouts:
- A silent counter
- An audible Bait and Tower number
- Soak Tower when safe to move

The Counter is based on Tyrant's Fire III (75F0). The Soak Tower is based on Tyrant's Flare II (7A88) puddles, it could have been just a delayed trigger from the Tyrant's Fire III, but possibly this is a cleaner solution for timings?

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -45,6 +45,9 @@ export interface Data extends RaidbossData {
   concept: { [name: string]: InitialConcept };
   splicer: { [name: string]: Splicer };
   alignmentTargets: string[];
+  burstCounter: number;
+  myTower?: number;
+  flareCounter: number;
   inverseMagics: { [name: string]: boolean };
   deformationTargets: string[];
 }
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -1732,6 +1737,57 @@ const triggerSet: TriggerSet<Data> = {
...
+      id: 'P8S Tyrant\'s Fire III Bait then Tower',
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: '75F0', source: 'Hephaistos' }),
+      condition: Conditions.targetIsYou(),
+      durationSeconds: 7.9,
+      infoText: (data, _matches, output) => output.text!({ num: data.burstCounter }),
+      run: (data) => data.myTower = data.burstCounter,
+      outputStrings: {
+        text: {
+          en: '${num}: Bait near Tower ${num}',
```
quisquous, line 1771:
```suggestion
      alertText: (data, _matches, output) => output.text!({ num: data.burstCounter }),
      run: (data) => data.myTower = data.burstCounter,
      outputStrings: {
        text: {
          en: '${num}',
```

I think this should be an alert to not overlap with all the infoText going on.  I also think this is a "you need to understand the mechanic".  Just the number is enough.

Legends0: Works for me.

PR #4843, opened 2022-09-18
raidboss: Allow Skins to Override Timerbar Elements

panicstevenson:
Using [parts](https://developer.mozilla.org/en-US/docs/Web/CSS/::part), you can override parts of custom elements. Hopefully this will alleviate some of the necessary touching of TypeScript components and allow users to completely override via CSS in their user raidboss.css file instead.

### DIFF
--- a/resources/timerbar.ts
+++ b/resources/timerbar.ts
@@ -348,83 +348,6 @@ export default class TimerBar extends HTMLElement {
           text-align: right;
           padding: 0px 0.4em 0px 0.4em;
         }
-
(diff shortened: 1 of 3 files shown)


quisquous:
This looks great, and is really nice cleanup.  Thank you for investigating and digging this up.

jpdied:
Is there an example of how a user would override using raidboss.css? I noticed Lippe skin changed and now the timeline and text are colored based on the intensity and I personally prefer it back to the original as some of the colors make it harder to see.

PR #4847, opened 2022-09-19
resources: add source to netlog_defs.ts

quisquous:
Also, clean up some comments while I'm here.

The raidemulator will currently generate all non-ffxiv plugin lines with "Unknown" prefix which should not match anything.  A followup fix could prevent it from generating these lines at all (making converted line optional or skipping it later), but it shouldn't cause any problems to have it here.

### DIFF
--- a/ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
+++ b/ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
@@ -15,8 +15,10 @@ const unknownLogMessagePrefix = 'Unknown';
 
 const logMessagePrefix: { [type: string]: string } = {};
 const logDefsGeneric: LogDefinitionMap = logDefinitions;
-for (const def of Object.values(logDefsGeneric))
-  logMessagePrefix[def.type] = def.messageType;
+for (const def of Object.values(logDefsGeneric)) {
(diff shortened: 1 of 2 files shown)


ui/raidboss/emulator/data/network_log_converter/LineEvent.ts
```diff
@@ -47,6 +49,7 @@ export default class LineEvent {
 
   prefix(): string {
     const timeString = DTFuncs.timeToTimeString(this.timestamp, this.tzOffsetMillis, true);
+    // TODO: should raidemulator not convert lines that don't come from the ffxiv plugin?
```
quisquous, line 52: FYI @valarnin

valarnin: I think it's a bit more complicated than that, because to replicate the behavior exactly as it current works in ACT, we'd have to also not emit log line events for the converted versions of lines. We'd need to make the converted line optional, then if it's undefined during playback, skip it for emitting lines.

quisquous: Yeah, my thoughts were along those same lines.  I think it is ok for now as no regexes should match, but maybe something to clean up later.

PR #4853, opened 2022-09-20
raidboss: make P7S inviolate purgation not use roles

quisquous:
Sometimes roles don't work and so this can cause bugs. This only considers individual debuffs.

Borrowed/modified from valarnin's comment here:
https://github.com/OverlayPlugin/OverlayPlugin/issues/60#issuecomment-1250426209


Legends0:
This worked for me tonight.

PR #4856, opened 2022-09-22
raidboss: simplify trigger by define api

MaikoTan:
For now, a raidboss trigger was structured like
```ts
import NetRegexes from '../../../../resources/netregexes';
import ZoneId from '../../../../resources/zone_id';
import { RaidbossData } from '../../../../types/data';
import { TriggerSet } from '../../../../types/trigger';

  // ...
}

  zoneId: ZoneId.Xxx,
  triggers: [],
};

export default triggerSet;
```

I think it is redundant to declare a `triggerSet` variable then export it as default just because you can not export it with a type in one line.
Recently I am learning VueJS and found that their [`defineComponent`](https://github.com/vuejs/core/blob/fbd697a4b6c238147d661026836314862fd09221/packages/runtime-core/src/apiDefineComponent.ts#L188) API is extremely useful for this use case.
So I changed the trigger type a little bit and add a tiny helper function called `defineTriggerSet` which reflectly return the object eaten by itself.

As in this patch, a trigger set can be structured like
```ts
import { defineTriggerSet } from '../../../../resources/api_define_trigger_set';
import NetRegexes from '../../../../resources/netregexes';
import ZoneId from '../../../../resources/zone_id';

// Or just omit the type parameter, as `export default defineTriggerSet({})`
// if there are no any extra parameters would be feed to `data`.
// The type parameter can also be omitted if there is `initData` function
// with fully typed return value defined
export default defineTriggerSet<{
  someExtraDataParams: string;
}>({
  zoneId: Zone.Xxx,
  initData: () => {
    return {
      someExtraDataParams: 'string',
    };
  },
  triggers: [],
});
```

I hope it would be easier to use and reduce several import line for types.

------

@valarnin  pinging if you have time to review this.

### DIFF
--- a/resources/api_define_trigger_set.ts
+++ b/resources/api_define_trigger_set.ts
@@ -1,0 +1,23 @@
+import { RaidbossData } from '../types/data';
+import { TriggerSet } from '../types/trigger';
+
+export function defineTriggerSet<InitData>(
+  set:
+    & TriggerSet<RaidbossData & InitData>
+    & (
+      Record<string, never> extends InitData ? Record<never, never>
+        : { initData: () => InitData }
+    ),
+): TriggerSet<RaidbossData>;
+
+export function defineTriggerSet(
+  set: Omit<TriggerSet<RaidbossData>, 'initData'>,
+): TriggerSet<RaidbossData>;
+
+/* eslint-disable func-style */
+export function defineTriggerSet<Data extends RaidbossData>(
+  set: TriggerSet<Data>,
+): TriggerSet<Data> {
+  return set;
+}
+/* eslint-enable func-style */
--- a/resources/api_define_trigger_set.ts
+++ b/resources/api_define_trigger_set.ts
@@ -8,7 +8,7 @@ export function defineTriggerSet<InitData>(
       Record<string, never> extends InitData ? Record<never, never>
         : { initData: () => InitData }
     ),
-): TriggerSet<RaidbossData>;
+): TriggerSet<RaidbossData & InitData>;
 
 export function defineTriggerSet(
   set: Omit<TriggerSet<RaidbossData>, 'initData'>,
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -187,14 +187,11 @@ export type BaseTriggerSet<Data extends RaidbossData> = {
 };
 
 // If Data contains required properties that are not on RaidbossData, require initData
-export type TriggerSet<Data extends RaidbossData> =
+export type TriggerSet<Data extends RaidbossData = RaidbossData> =
   & BaseTriggerSet<Data>
-  & (RequiredFieldsAsUnion<Data> extends RequiredFieldsAsUnion<RaidbossData> ? {
+  & {
     initData?: DataInitializeFunc<Data>;
-  }
-    : {
-      initData: DataInitializeFunc<Data>;
-    });
+  };
 
 // Less strict type for user triggers + built-in triggers, including deprecated fields.
 export type LooseTimelineTrigger = Partial<TimelineTrigger<RaidbossData>>;
--- a/ui/raidboss/data/00-misc/general.ts
+++ b/ui/raidboss/data/00-misc/general.ts
@@ -1,16 +1,14 @@
+import { defineTriggerSet } from '../../../../resources/api_define_trigger_set';
 import NetRegexes from '../../../../resources/netregexes';
 import ZoneId from '../../../../resources/zone_id';
 import { RaidbossData } from '../../../../types/data';
-import { TriggerSet } from '../../../../types/trigger';
-
-export type Data = RaidbossData;
 
 const caresAboutTankStuff = (data: RaidbossData) => {
   return data.role === 'tank' || data.role === 'healer' || data.job === 'BLU';
 };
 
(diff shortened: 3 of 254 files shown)


resources/api_define_trigger_set.ts
```diff
@@ -0,0 +1,23 @@
...
+  set: Omit<TriggerSet<RaidbossData>, 'initData'>,
+): TriggerSet<RaidbossData>;
+
+/* eslint-disable func-style */
+export function defineTriggerSet<Data extends RaidbossData>(
+  set: TriggerSet<Data>,
+): TriggerSet<Data> {
+  return set;
+}
+/* eslint-enable func-style */
```
valarnin, line 23: Is there a reason that something like this wouldn't work instead, and not require disabling the eslint rule?
```ts
export defineTriggerSet<Data extends RaidbossData>(
  set: TriggerSet<Data>,
): TriggerSet<Data> => {
  return set;
}
```

MaikoTan: When to declare a function with overrides, I think this approach would be much clear. Every overrides would be regarded as a function declaration.
```ts
export function defineTriggerSet(set): TriggerSet<Data>;
export function defineTriggerSet<...>(set: TriggerSet<...>): TriggerSet<Data>;
export function defineTriggerSet<Data>(set: TriggerSet<Data>): TriggerSet<Data> {
  // ...
}
```

Rather than regarded as a type?
```ts
export const defineTriggerSet: {
  <InitData>(
    set:
      & TriggerSet<RaidbossData & InitData>
      & (
        Record<string, never> extends InitData ? Record<never, never>
          : { initData: () => InitData }
      ),
  ): TriggerSet<RaidbossData>;

  (set: Omit<TriggerSet<RaidbossData>, 'initData'>): TriggerSet<RaidbossData>;
} = <Data extends RaidbossData>(
  set: TriggerSet<Data>,
): TriggerSet<Data> => {
  return set;
};
```

MaikoTan:
> I think that part of the original design here was to allow the `Data` type to be used elsewhere, e.g. in maybe oopsy or in test suites, but it's never been leveraged. It would still be possible to extract that type from the default export but it's a bit more ugly, as such I think I'd still prefer to leave the previous `export interface Data extends RaidbossData`.

I see, so that is why are there every file export a `type Data = Raidboss`?
I didn't found it was used everywhere so I thought it is redundant.

But it is also possible to extract `Data` type from the trigger set such as:
```ts
type ExtractData<TS> = TS extends TriggerSet<infer Data> ? Data : never;
```

trim21:
> I see, so that is why are there every file export a type Data = Raidboss?

I need to import them as a ts user trigger writter to know what properties are already used 😄

quisquous:
Re: explicitly defining `Data`.  I agree that you can pull `Data` back out with `ExtractData` as you say above, but I think that's cumbersome.  My opinion here is that if a type is being used by another file, that type should generally be exported at the top level and not extracted from other types.

On top of that, I think inferring types can lead to some awkward code.  One example: `combatantData: PluginCombatantState[];`.  If `initData` was the only place where this type was inferred, I'd have to write `combatantData: [] as PluginCombatantState[]`.  I think having to use `as` type coercion is bad practice, because it's the programmer overriding the type checker (when the programmer can get it wrong).

I think if the `Data` part of this patch is reverted, then the only remaining part is simplifying the `export default triggerSet`.  Personally, I don't think that's really enough of a simplification to merit a function here.  I think if there were ~3-5 lines of boilerplate that everybody had to copy and paste in every trigger, a function would be great.  I think if we needed to do some special processing that required a function, a function would be great.  But, in this case, it's just a single straightforward line.

Sorry to push back, but if in the end this is just to eliminate a single `export default triggerSet` line, then I don't think this change is worth it.

MaikoTan:
> On top of that, I think inferring types can lead to some awkward code. One example: `combatantData: PluginCombatantState[];`. If `initData` was the only place where this type was inferred, I'd have to write `combatantData: [] as PluginCombatantState[]`. I think having to use `as` type coercion is bad practice, because it's the programmer overriding the type checker (when the programmer can get it wrong).

It can also be `defineTriggerSet<{ combatantData: PluginCombatantState[] }>(...)`.
But yeah, I will agree that `as`-coercion could make people get it wrong.

> I think if the `Data` part of this patch is reverted, then the only remaining part is simplifying the `export default triggerSet`. Personally, I don't think that's really enough of a simplification to merit a function here. I think if there were ~3-5 lines of boilerplate that everybody had to copy and paste in every trigger, a function would be great. I think if we needed to do some special processing that required a function, a function would be great. But, in this case, it's just a single straightforward line.
> 
> Sorry to push back, but if in the end this is just to eliminate a single `export default triggerSet` line, then I don't think this change is worth it.

That is fair. I am trying not to change the trigger set structure a lot, otherwise may it causing unexpected bugs.
If you think it is not worth, I don't mind just to close it.

quisquous:
Yeah, sorry.  I think there are just too many constraints here to make this clean up worth it.  I appreciate you thinking about how to try to simplify things.  I think in a general sense it's tough to change too much with triggers because of backwards compat issues and because of how many different ways that they are used.

trim21:
we can add a `cast` method to do typechecking on `export default`?

PR #4862, opened 2022-09-22
raidboss: fix skin colors not working properly

### LINKED ISSUE #4857: Raidboss: Lippe skin colors overwrittten by last update
Is there an example of how a user would override using raidboss.css? I noticed Lippe skin changed and now the timeline and text are colored based on the intensity and I personally prefer it back to the original as some of the colors make it harder to see.

_Originally posted by @jpdied in https://github.com/quisquous/cactbot/issues/4843#issuecomment-1254620818_

quisquous:
`rgba(r g b / a%)` needs a slash in it.

Closes #4859.
Fixes #4857.

### DIFF
--- a/ui/raidboss/skins/jwidea/jwidea.css
+++ b/ui/raidboss/skins/jwidea/jwidea.css
@@ -47,7 +47,7 @@ timer-bar::part(text-container) {
 }
 
 .info-text {
-  color: rgba(255 255 255 100%);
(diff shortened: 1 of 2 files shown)


quisquous:
I tried upgrading stylelint and it still didn't catch this.  I don't see any obvious rules here either.

panicstevenson (approved):
(oops, thanks!)

Issue #4868, opened 2022-09-24
CSS styles not applied to WSServer URLs? 


Theletha:
The overlay is not showing the correct themed CSS styles when using WSServer URLs:

`file://C:/Users/Thele/AppData/Roaming/Advanced%20Combat%20Tracker/Plugins/cactbot-0.16.11/cactbot/ui/raidboss/raidboss_timeline_only.html?OVERLAY_WS=ws://127.0.0.1:10501/ws`
`file://C:/Users/Thele/AppData/Roaming/Advanced%20Combat%20Tracker/Plugins/cactbot-0.16.11/cactbot/ui/raidboss/raidboss_alerts_only.html?OVERLAY_WS=ws://127.0.0.1:10501/ws`

Previously it has showed the correct one for the timeline only but never with the alerts as you can see 
One overlay is from ACT (Top) and the other one is via WSServer URL (Bottom)

![167521530-e4c13363-40fd-450f-bc75-008a029712f3](https://user-images.githubusercontent.com/10131942/192090609-dd53564c-100f-4082-bd6c-e37ea710f805.png)

but now even the timeline has reverted back to the default theme 

![asU2T4B](https://user-images.githubusercontent.com/10131942/192090683-61e211e4-d9dd-46b2-918a-a53050445f97.png)

Is there a way for me to change the theme for the WSServer URLs?

panicstevenson:
I think I see the issue:
This happened with the application of https://github.com/quisquous/cactbot/pull/4843, which is a more proper version of allowing skins to work without having to put the custom CSS within the timerbars themselves.
It looks like there is code to apply CSS that has always been failing with `OVERLAY_WS=` due to the forward slashes, so CSS has just never been getting applied but it didn't really matter since skins were essentially hardcoded.
I'll try to figure out a permanent change and something temporary so you can use skins before the next cactbot release.

panicstevenson:
Probably not worth it, but if you needed to make this change locally, you can edit `raidboss.bundle.js` locally (`%APPDATA%/Roaming/Advanced Combat Tracker/Plugins/cactbot-0.16.11/cactbot/ui/raidboss/raidboss.bundle.js` for you) line 3798:
```js
  handleSkin(skinName) {
    if (!skinName || skinName === 'default') return;
    let basePath = document.location.toString();
    const slashIdx = basePath.lastIndexOf('/');
    if (slashIdx !== -1) basePath = basePath.substr(0, slashIdx);
[...]

PR #4870, opened 2022-09-24
raidboss: Fix Skins CSS for OVERLAY_WS

panicstevenson:
For users using the `OVERLAY_WS` query parameter, the value could contain forward slashes, which causes a failure to load skin CSS. This wasn't an issue when the CSS was directly in timerbar.ts.

Fixes https://github.com/quisquous/cactbot/issues/4868

### DIFF
--- a/resources/user_config.ts
+++ b/resources/user_config.ts
@@ -401,15 +401,8 @@ class UserConfig {
   handleSkin(skinName: string) {
     if (!skinName || skinName === 'default')
       return;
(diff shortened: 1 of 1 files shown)


Theletha:
Should i try to change it to backwards slashes instead?

Theletha:
ok its working when i did that. amazing. wonder if i need to do the code swap too since its still in a changed state?

PR #4871, opened 2022-09-24
build: Upgrade ts-loader Package

panicstevenson:
Recent versions (>9.3.0) of `ts-loader` implicitly disable the compiler options for declaration and declarationMap when the Fork TS Checker Webpack Plugin is being used (non-CI environments).

This improves my local build times from ~40 seconds to ~15 seconds. There are faster options overall (esbuild-loader was one potential option), but this seemed fairly low-risk.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -80,7 +80,7 @@
         "stylelint": "^14.8.2",
         "stylelint-config-standard": "^25.0.0",
         "tar-fs": "^2.1.1",
-        "ts-loader": "^8.0.17",
+        "ts-loader": "^9.3.0",
         "ts-node": "^10.8.0",
         "tsconfig-paths": "^3.9.0",
         "ttypescript": "^1.5.12",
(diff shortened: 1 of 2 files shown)


trim21:
bottleneck of local building time is type checking (tsc), upgrading `ts-loader` won't help.

https://github.com/quisquous/cactbot/pull/4221#issuecomment-1255775247

quisquous:
Even if it's not the primary bottleneck, I think no reason not to upgrade it if Panic sees some perf improvements.

quisquous:
@trim21 for me at least, this sped things up by 2x

```diff
-webpack 5.72.0 compiled successfully in 133084 ms
+webpack 5.72.0 compiled successfully in 50205 ms
```

trim21:
yes, I forget that dev-server don't need to wait for tsc fork plugin.

trim21:
I guess we can replace babel-loader with tsc-loader back in webpack.prod.ts to get both type checking and TypeScript transpiling faster ( and remove tsc fork plugin), considering not ts-loader is faster than tsc.

Issue #3000, opened 2021-06-03
Option to use job name instead of player name


Jaehyuk-Lee:
## Summary

I know why triggers are calling player name, not the job name. You can have two or more player with same job.
But usually when you go latest raid content, you normally have one player per job. I think there's quite a few people who wants to know the job name, not nickname. I often hear that if there is any way to called by job name instead of nickname from community, especially from those who uses PF/RF without static.

Jaehyuk-Lee:
I've briefly thought how to apply it, it seems not an easy one. I guess the player name is loaded directly from the log. So, there should be some table relating player with job.

valarnin:
You can get any player in the party's job via `data.party.details`, for instance I have this in one of my custom triggers:

```typescript
let playerJob;
data.party.details.forEach((p) => {
  if (p.name === partner)
    playerJob = Util.jobEnumToJob(p.job);
});
```

My question is, how on earth would you code this? A post-trigger filter to replace player names with jobs or something?

Jaehyuk-Lee:
> My question is, how on earth would you code this? A post-trigger filter to replace player names with jobs or something?
I guess so. I do think this would need a lot effort.

valarnin:
Would it make more sense to do this based on indexed roles or something so it works even with duplicates? E.g.
```
Tank 1
Tank 2
Healer 1
Healer 2
DPS 1
DPS 2
DPS 3
DPS 4
```

I don't know if there's a way to get the party list in the order it's actually displayed in-client.

Jaehyuk-Lee:
Oh... I understand what you mean. If we can get party member's list order, this can be different by each player, so we should get it from the client data, that's not a bad idea.
But I guess that's harder than just calling player's job.

valarnin:
It's possible that it's the natural order of the party list returned by the FFXIV parser plugin, I'd have to check the decomp or do some tests in-game to be sure?

valarnin:
No, looks like that's a bust, party order in partychanged event is random from what it appears? It also doesn't fire that event mid-fight at all, only when someone is added to or removed from the party.

panicstevenson:
> My question is, how on earth would you code this? A post-trigger filter to replace player names with jobs or something?

Something that pushes everyone's <name, job> to shortNames temporarily upon entering an instance? This would be especially terrible in the alliance raids when you're asked to stack on 'Red Mage (6)'.

MaikoTan:
[...]

PR #4873, opened 2022-09-25
Triggers?

AngelSeed:
https://gyazo.com/cba21f677d2de90c39f06516b3ffbe0e
For example, is it possible to change this (${player}) to the job notation instead of the player name?
If you're a player, the player name will be different every time you're attacking as a stray.


trim21:
duplicated with https://github.com/quisquous/cactbot/issues/3000

PR #4874, opened 2022-09-25
build: remove timeline loader

trim21:
timeline is now parsed at runtime and currently `timeline-loader` do nothing, so no need to keep `timeline-loader`, just use `raw-loader`

also, `raw-loader` is deprecated in webpack5.


trim21:
this change should trigger build workflow…

quisquous:
Thanks for the cleanup, looks good to me!

Issue #3247, opened 2021-07-05
Emulator worker.js is named as `379.bundle.js` in gh-pages branch


trim21:
## Description

I happen to find there is  a new file in gh-pages branch.

I checked the commit log, it's the original service worker js

NetworkLogConverter.worker.bundle.worker.js → 379.bundle.js

We should keep it's old filename I think.

Or some filename to tell what it's doing. Current filename is confusing

valarnin:
I'm really not sure why it's named like that in the gh-pages branch. When running with webpack dev server, the name is a bit ugly but at least it's obvious what it is:

![image](https://user-images.githubusercontent.com/6119598/124676324-4dfac580-de8c-11eb-8627-351e3e3a802d.png)

Maybe it's a dev vs prod build difference?

(this was caused by #3230, for reference. Since it doesn't need to bundle the js file for inlineing, it doesn't get a manually assigned bundle name I think)

trim21:
it's named by chunk name rule `[id].js`, and its id was 379, and curretnly 378.

So there are 2 solution for this:

1. add webpack config `{ optimization: { chunkIds: 'named' } }`
2. set a chunk name when importing service worker `new Worker(/* webpackChunkName: emulator.worker" */ ... )`

PR #4875, opened 2022-09-25
build: stabilize emulator service worker's output filename

trim21:
close https://github.com/quisquous/cactbot/issues/3247

currently the final file of emulator's service worker script is named by chunk name `[id].bundle.js`, and its id is curretnly 378, so webpack will generate a service worker file with meanless filename `378.bundle.js` in `dist/` .

set `webpackEntryOptions: { filename: "ui/raidboss/raidemulator.worker.js" }` so it will be generated to `dist/ui/raidboss/raidemulator.worker.js`


quisquous:
Thank you!!

Issue #4865, opened 2022-09-23
gh-pages branch is too big


trim21:
I have a 1 year ago cactbot repo (head 16700eb4 raidemulator: Convert RaidEmulatorAnalysisTimelineUI to typescript (\# 3106) - <valarnin> 1 year, 3 months ago)

And today I pull it I found I need to pull 1.67GB files

```text
remote: Enumerating objects: 22376, done.
remote: Counting objects: 100% (3747/3747), done.
remote: Compressing objects: 100% (197/197), done.
remote: Total 22376 (delta 3596), reused 3679 (delta 3542), pack-reused 18629
Receiving objects: 100% (22376/22376), 1.67 GiB | 6.29 MiB/s, done.
Resolving deltas: 100% (14320/14320), completed with 855 local objects.
```

Then I found currently github api says cactbot git repo's size is 1.74 GB (less than 100 MB 1 year ago).

also, it take actions/checkout 2min to clone the repo in github actions...

![image](https://user-images.githubusercontent.com/13553903/192562362-1901228a-35b7-41cb-8314-5e01526ca095.png)

MaikoTan:
maybe [`git maintenance`](https://github.com/git/git/blob/master/Documentation/git-maintenance.txt) command is useful here.

trim21:
> maybe [`git maintenance`](https://github.com/git/git/blob/master/Documentation/git-maintenance.txt) command is useful here.

need a github admin to run it on github server, not locally.

trim21:
```text
# git clone https://github.com/quisquous/cactbot.git --branch main --depth=10000000 cactbot-test
Cloning into 'cactbot-test'...
remote: Enumerating objects: 61899, done.
remote: Counting objects: 100% (163/163), done.
remote: Compressing objects: 100% (135/135), done.
remote: Total 61899 (delta 47), reused 132 (delta 22), pack-reused 61736
Receiving objects: 100% (61899/61899), 63.52 MiB | 9.00 MiB/s, done.
Resolving deltas: 100% (47011/47011), done.
```

```text
# git clone https://github.com/quisquous/cactbot.git --branch gh-pages --depth=10000000 cactbot-test
Cloning into 'cactbot-test'...
remote: Enumerating objects: 55939, done.
remote: Counting objects: 100% (204/204), done.
remote: Compressing objects: 100% (134/134), done.
Receiving objects:  22% (12509/55939), 909.39 MiB | 10.25 MiB/s

```

[...]

PR #4883, opened 2022-09-27
ci: upgrade actions checkout to v3

trim21:
bug of `actions/checkout@v2` is fixed in v3 I think, using v3 for a while, didn't get wrong commit checkout.

And it only fetch depth=1 by default to avoid fetching all commits https://github.com/quisquous/cactbot/issues/4865 , take 2m to 2s.


MaikoTan:
nice catch!

PR #4884, opened 2022-09-27
raidboss: Ensuring timelines have a non-chatlog initial sync

xpdota:
From https://github.com/xpdota/event-trigger/issues/130

The problem is that if a user enables "Hide chatlog for privacy" in the parsing plugin OR is simply chainpulling with no countdown, and a timeline doesn't have any valid initial sync other than that (i.e. sync with a large enough window to be valid for 0.0), then the timeline will just never start.

I looked at also just using OP's combat start to start the timeline, but this doesn't necessarily work for timelines where the actual contents don't start at 0.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p5n.txt
+++ b/ui/raidboss/data/06-ew/raid/p5n.txt
@@ -7,7 +7,7 @@ hideall "--sync--"
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
 
 0.0 "--sync--" sync /Engage!/ window 0,1
-10.3 "--sync--" sync / 1[56]:[^:]*:Proto-Carbuncle:76D6:/ window 10,10
+10.3 "--sync--" sync / 1[56]:[^:]*:Proto-Carbuncle:76D6:/ window 11,10
 19.1 "Searing Ray" sync / 1[56]:[^:]*:Proto-Carbuncle:76D7:/
 28.2 "Ruby Glow" sync / 1[56]:[^:]*:Proto-Carbuncle:76D4:/
 31.4 "--sync--" sync / 1[56]:[^:]*:Proto-Carbuncle:76D6:/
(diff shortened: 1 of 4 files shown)


ui/raidboss/data/06-ew/raid/p7n.txt
```diff
@@ -9,7 +9,7 @@ hideall "--sync--"
 # -ii 77F6 77F7 77FA 77FD 77FF 7800 7804 7806 7807 780A 783E 7841
 
 0.0 "--sync--" sync /Engage!/ window 0,1
-3.0 "--sync--" sync / 1[56]:[^:]*:Agdistis:78E3:/
+3.0 "--sync--" sync / 1[56]:[^:]*:Agdistis:78E3:/ window 3,5
```
quisquous, line 12: Autos have inconsistent timing.  They are ok to start a timeline, but you should add windows to them, because it'll catch the next auto.  This timeline already has a `window 16,5` below.  If you wanted to make it earlier, you could add a starts using line.

quisquous (commented):
Thanks for the patch!

ui/raidboss/data/06-ew/raid/p7n.txt
xpdota: Ok, tightened down the windows on those.

ui/raidboss/data/06-ew/raid/p5n.txt
```diff
@@ -7,7 +7,7 @@ hideall "--sync--"
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
 
 0.0 "--sync--" sync /Engage!/ window 0,1
-10.3 "--sync--" sync / 1[56]:[^:]*:Proto-Carbuncle:76D6:/ window 11,10
+10.3 "--sync--" sync / 1[56]:[^:]*:Proto-Carbuncle:76D6:/ window 11,1
```
quisquous, line 10:
```suggestion
10.3 "--sync--" sync / 1[56]:[^:]*:Proto-Carbuncle:76D6:/ window 11,10
```

This isn't an auto.

PR #4886, opened 2022-09-27
build: transpile triggers with ts-node api

trim21:
remove a TODO in `util/process_triggers_folder.ts`

>  // Generate javascript from typescript.
>  // TODO: replace this with programatic use of TypeScript API.

not extractly using typescript, but calling ts-node's transpile api, so type checking is also removed ( still covered by other tsc ).

This PR also make this script able to be executed in any cwd.


util/process_triggers_folder.ts
```diff
@@ -109,16 +115,30 @@ const lint = async (filename: string, lines: string[]) => {
...
-  const originalFilename = outputFileToOrigFile(filename);
-  let lines = originalContents.split(/[\r\n]+/);
+const tsc = tsNode.create({
+  transpileOnly: true,
+  project: path.join(__dirname, '../tsconfig.json'),
+  // compilerOptions: {
+  //   target: 'ES2020',
+  //   declaration: false,
+  //   declarationMap: false,
+  // },
```
quisquous, line 125:
```suggestion
```

Should this comment be removed?

util/process_triggers_folder.ts
```diff
@@ -159,51 +179,40 @@ const processFile = async (filename: string) => {
...
   // Overwrite the file.
-  fs.writeFileSync(filename, contents);
+  fs.mkdirSync(path.dirname(distFilePath), { recursive: true });
+  fs.writeFileSync(distFilePath.replace('.ts', '.js'), contents);
 
   // Copy timeline file if present
-  const importPath = path.join(
-    '..',
-    originalFilename.replace(path.extname(originalFilename), '.ts'),
+  const importPath = ('../ui/raidboss/data/' + path.relative(root, originalFilename)).replace(
```
quisquous, line 192: Thanks for fixing this.  I have no idea how this ever worked locally without the slash replacement.

trim21: yes, it didn't work...

trim21:
It looks like eslint is not necessary here? It can also to be removed.

quisquous:
> It looks like eslint is not necessary here? It can also to be removed.

I'm not sure.  I felt like it was there for a reason, but I'd have to look at a diff to believe it.  I think this is fine to land and eslint can be removed separately.

trim21:
> > It looks like eslint is not necessary here? It can also to be removed.
> 
> I'm not sure. I felt like it was there for a reason, but I'd have to look at a diff to believe it. I think this is fine to land and eslint can be removed separately.

https://github.com/trim21/temp-repo-show-diff/commit/a0520596737a6d1a93457fada9ebd46de2145e43

quisquous:
Yeah, looks like it's there for a reason, then.  I'd prefer that the transpiled triggers look as close to the original as possible.

PR #4891, opened 2022-09-28
raidboss: fix Thavnair/Mare Lamentorium hunt triggers

### LINKED ISSUE #4889: Some Endwalker A-Rank Hunt Triggers Not Working
This will inevitably require me to collect more information if someone doesn't have a fix or an acknowledgment that this is a known issue but ever since I ran multiple hunt trains on Tuesday, I've noticed that a lot of Endwalker A Rank hunts which formerly had callouts no longer do.

Two that come to my mind, off the top of my head, are Surgriva in Thavnair and Lunatender in Mare Lamentorum, for example.

I still see these hunts in Cactbot's settings but the callouts/raid warnings no longer play for me on these and some of the other A-Ranks for some reason, unfortunately. I still have raid warnings and callouts for some other A-rank hunts and all other content in Cactbot so I don't know why it's only affecting these encounters.

quisquous:
Fixes #4889.

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -833,6 +833,9 @@ export class PopupText {
 
     this.inCombat = inCombat;
 
+    // Update in combat
+    this.data = this.getDataObject();
+
     if (!this.resetWhenOutOfCombat)
       return;
 


quisquous:
This needs testing, but I'm pretty sure this is the issue.

PR #4892, opened 2022-09-28
ci: simplify config

trim21:
github actions like `setup-node@v3` and `setup-python@v3` has cache feature now so no need to add a seprated cache action.

It's also possible to keep all `setup-node@v3` and `npm ci` in one file.


.github/actions/setup-js-env/action.yml
```diff
@@ -0,0 +1,13 @@
...
+  using: "composite"
+  steps:
+    - uses: actions/setup-node@v3
+      with:
+        node-version: '16'
+        cache: 'npm'
+
+    - name: Install Dependencies
+      shell: bash
+      run: npm ci --ignore-scripts
```
MaikoTan, line 13: It is also possible to include `install` `cache` `lint` `test` actions into a compose action, and controls by `inputs` (with `with` keyword)

trim21: that's too complex I think

quisquous:
@panicstevenson Do you have any time to look at this?

panicstevenson (approved):
I don't have the availability to test this currently, but LGTM on the surface

PR #4896, opened 2022-09-30
i18n: add 6.1 ko translations

Bing-su:
Worked on #4894.

- alzadaals_legacy.txt

I've checked it several times, and this seems to be `Toxin Shower`.
<https://xivapi.com/Action/28508>


quisquous:
Yeah, it looks like it should be Toxin and not Toxic.  You're correct.  Can you rebase this patch and I'll land that change?

Bing-su:
Done. thanks!

PR #4900, opened 2022-10-01
raidboss: minor Sephirot trigger changes

quisquous:


JLGarber (approved):

JLGarber:
Good catch on the 1 vs 4 puddles in the timeline, I definitely didn't consider that carefully enough when writing this.

PR #4903, opened 2022-10-02
resources: support for mapeffects

wexxlee:
Added support in netlogs_defs and regexes/netregexes for new MapEffect (257) lines, along with some updates to LogGuide.  

There are dependencies both on the ACT FFXIV Plugin being updated to v2.6.6.7 (which adds processing support for these lines), and on an update to OverlayPlugin to  re-add support for these lines that was disabled in v0.19.6.  Marking this as draft for now until these dependencies are resolved.

PR #4907, opened 2022-10-03
raidboss: add triggers for P6S polyominoids & cachexia 2

wexxlee:
Added new triggers to support all of the Aetherial Polyominoid / Polyominoid Sigma mechanics in P6S.  I converted this trigger from my raidboss.js user file, and had to clean up a lot of TypeScript foo, so I would appreciate other eyes on this.

I've tested pretty heavily using Raid Emulator, and I'll be testing with a local build this week.

Marked as draft for now pending further testing.  Also depends on #4903 support for MapEffect lines.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p6s.ts
+++ b/ui/raidboss/data/06-ew/raid/p6s.ts
@@ -1,9 +1,11 @@
 import Conditions from '../../../../../resources/conditions';
 import NetRegexes from '../../../../../resources/netregexes';
 import Outputs from '../../../../../resources/outputs';
+import { callOverlayHandler } from '../../../../../resources/overlay_plugin_api';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
+import { PluginCombatantState } from '../../../../../types/event';
 import { NetMatches } from '../../../../../types/net_matches';
 import { TriggerSet } from '../../../../../types/trigger';
 
@@ -16,6 +18,10 @@ export interface Data extends RaidbossData {
   aetheronecrosisDuration: number;
   predationCount: number;
   predationDebuff?: string;
+  polyInstance: number;
+  tileTethers: NetMatches['Tether'][];
+  mapEffects: NetMatches['MapEffect'][];
+  combatantData: PluginCombatantState[];
 }
 
 // Due to changes introduced in patch 5.2, overhead markers now have a random offset
@@ -42,6 +48,10 @@ const triggerSet: TriggerSet<Data> = {
       pathogenicCellsCounter: 0,
       aetheronecrosisDuration: 0,
       predationCount: 0,
+      polyInstance: 0,
+      tileTethers: [],
+      mapEffects: [],
+      combatantData: [],
     };
   },
   triggers: [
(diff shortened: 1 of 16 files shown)


ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +95,259 @@ const triggerSet: TriggerSet<Data> = {
...
+        data.tileTethers.push(matches);
+      },
+    },
+    {
+      id: 'P6S Polyominoid MapEffect Collect',
+      type: 'MapEffect',
+      netRegex: NetRegexes.mapEffect({ flags: ['00020001', '00400020'] }),
+      run: (data, matches) => {
+        // location '00' is center/mapwide and won't be used
+        // for determining mechanic resolution
```
valarnin, line 112: More specifically, location `00` triggers the black lines outlining the tiles in the arena.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +95,259 @@ const triggerSet: TriggerSet<Data> = {
...
+      type: 'Tether',
+      netRegex: NetRegexes.tether({ id: '00CF' }),
+      run: (data, matches) => {
+        data.tileTethers.push(matches);
+      },
+    },
+    {
+      id: 'P6S Polyominoid MapEffect Collect',
+      type: 'MapEffect',
+      netRegex: NetRegexes.mapEffect({ flags: ['00020001', '00400020'] }),
```
valarnin, line 109: I dislike these two flags being repeated here and then below, feels like a "magic numbers" situation. I'd rather see a `const crossTile = ...; const plusTile = ...;` at the top of the file and then using those below so it's easier to tell at a glance what these correspond to.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +95,259 @@ const triggerSet: TriggerSet<Data> = {
...
+        insideNE: {
+          en: 'Inside NE',
+        },
+        insideSE: {
+          en: 'Inside SE',
+        },
+        insideSW: {
+          en: 'Inside SW',
+        },
+        // Corner tiles will never be safe for any version of Poly,
```
valarnin, line 322: This comment is incorrect, since poly 6/cache2 has safe corner spots. It would be more accurate to say that the trigger never calls out corner safe spots?

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: Thanks - I never paid attention to that,  Will update the comment.

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: Agreed - will update that too.

wexxlee:
Tested yesterday (before today's fixes/changes re: Poly 5 & Chorus Ixou), and no issues.  Will be testing again today or Thursday, and aiming to mark this as ready by Friday.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +102,235 @@ const triggerSet: TriggerSet<Data> = {
...
+              return;
+            return output.combo!({ dir1: output[safe0]!(), dir2: output[safe1]!() });
+          case 2: // one inside safe spot
+            if (safeTiles.length !== 1 || output[safe0] === undefined)
+              return;
+            return output.single!({ dir1: output[safe0]!() });
+          case 3: // two inside safe spots
+            if (safeTiles.length !== 2 || safe1 === undefined || output[safe0] === undefined || output[safe1] === undefined)
+              return;
+            return output.combo!({ dir1: output[safe0]!(), dir2: output[safe1]!() });
```
quisquous, line 271: Similar to case 1, I wonder if this one should also say "healer groups (NW/SE)" or something.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +102,235 @@ const triggerSet: TriggerSet<Data> = {
...
+      },
+      outputStrings: {
+        combo: {
+          en: 'Safe Tiles: ${dir1} / ${dir2}',
+        },
+        single: {
+          en: 'Safe Tile: ${dir1}',
+        },
+        polyCachexia: {
+          en: 'Safe Tiles: Intercards of ${dir1}',
```
quisquous, line 312: I understand what you're getting at here, but I personally find this wording very hard to parse.  I certainly could not call this out and have anybody understand it.  Maybe this is just me?

I was wondering it'd make sense to just say "inside east" or "outside west" depending on which debuff you had, since at least being in the right column is the most important part here, and it's pretty easy to figure out where to go from there.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +102,235 @@ const triggerSet: TriggerSet<Data> = {
...
+          case 2: // one inside safe spot
+            if (safeTiles.length !== 1 || output[safe0] === undefined)
+              return;
+            return output.single!({ dir1: output[safe0]!() });
+          case 3: // two inside safe spots
+            if (safeTiles.length !== 2 || safe1 === undefined || output[safe0] === undefined || output[safe1] === undefined)
+              return;
+            return output.combo!({ dir1: output[safe0]!(), dir2: output[safe1]!() });
+          case 4: // lots of safe spots, so give generic warning
+            return output.polyAvoid!();
```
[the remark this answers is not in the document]
wexxlee: Yeah, I think you're right.  I think when I first wrote this, I was aiming to have a call for every Poly mech for completeness, but in hindsight, it's really not needed.  I'm actually leaning toward doing an "if poly4/return" at the top of the trigger now, just to avoid going through all the safe-spot calculations unnecessarily.

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -85,6 +102,235 @@ const triggerSet: TriggerSet<Data> = {
...
+
+        const poly6Map: { [l: string]: string } = {
+          insideNW: 'insideSE',
+          insideNE: 'insideSW',
+          insideSE: 'insideNW',
+          insideSW: 'insideNE',
+        };
+
+        switch (data.polyInstance) {
+          case 1: // four safe spots: two inside, two outside.  call only the inside ones.
```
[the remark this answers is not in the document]
wexxlee: Interesting - the way my various groups do it and the way I've seen the few times I've done it in PF is that one healer group takes the north inside tile, and the other takes the south.  But if you're seeing it resolve mostly east/west splits, then I think that's what we should call (and either way, calling inside west/outside east gives more fulsome info).  And yes, completely agree with combining it with exocleaver.

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: > Similar to case 1, I wonder if this one should also say "healer groups (NW/SE)" or something.

Yep, totally agree.  Oddly though, as you say that,  I don't remember ever seeing a healer group alert during that mechanic, or during the final poly (poly8) where that too is combined with a healer group stacks.  I've noticed it a couple times, but keep forgetting to bring it up/do something about it.

Looking at the triggers and my logs, I _think_ it's because the P6S Exocleaver Healer Groups trigger sets data.secondExocleavers = true (so the trigger doesn't execute twice for the second exocleaver hit) - but that var is not ever reset, so the trigger doesn't refire on subsequent occurrences of the unholy dark mechanic.  So I think the fix would be to set data.secondExocleavers = false in the second trigger (P6S Exocleaver Move), so the initial trigger is able to fire again the next time the mechanic occurs.  Let me know if that sounds right.

Happy to fix this as part of modifying the exocleaver trigger anyway for this PR - but lmk if you prefer to do it as a separate PR, and I can submit that tonight after raid.

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: > I understand what you're getting at here, but I personally find this wording very hard to parse. I certainly could not call this out and have anybody understand it. Maybe this is just me?

To be honest, it's a pretty difficult mechanic to call, because there are four safe squares that are non-adjacent, and each needs to be utilized by a different person/group to successfully resolve the mechanic.  Combine that with the fact that there are different strats (aoe N stack S vs. aoe S stack N), and there's a lot of room for confusion.

I hesitate though about doing "inside east"/"outside west" though, because we use the "inside x / outside x" nomenclature in p8s to call adjacent pairs of safe tiles (and now will be doing the same with poly 1 & 8), so i think that risks even more confusion.  And at the end of the day, being in the wrong tile, even if on the right side, is at least a damage down if not death.

Two other options to consider.  Since we do know which side (e/w) the player needs to be on, we could call it as:
1. Go North/South of ${dir1} (and have $dir1 be the tile in between the N/S safe tiles on the correct side.  This might be more clear.  Downside is that it's not a call that you could call out universally for the entire raid, as it will only apply to 4 people.
2. Call the two safe tiles directly (${dir1} and ${dir2}) on the safe side.  This is even more direct, but (a) it means using "Outside WNW" (or whatever) as part of the call, and personally, I hate how nasty that is, and (b) same problem as above in only applying to 4 people.

Just not sure the right way to concisely communicate the info here, both for the individual to resolve and for people doing callouts.

wexxlee: 3. I suppose the third option here (which is what one of my groups does) is to call the inside safe tile (there will only be one) as the "anchor tile", and everyone knows the rest of the safe tiles based on that (inside tile, opposite corner, and then the other two).  But again, not sure how to concisely communicate that in a call.

ui/raidboss/data/06-ew/raid/p6s.ts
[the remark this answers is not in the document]
wexxlee: > Do you think it'd be worth combining this with `P6S Exocleaver Healer Groups` too, e.g. "Healer groups (inside west/outside east)" or however we'd word it.

Actually, testing this tonight, I think the combined call "Healer Groups (Inside North/Outside South)" is a bit too long/wide for a single output on screen.  I think it might make sense to either have the inside north/outside south still called by the poly trigger, or to have the Exocleaver call do a response: and put the healer groups up as alertText and the directions up as infoText.  Thoughts?

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: Ok, sleep + caffeine + more thought, and here's (yet another) option but I think it's the best of all worlds.

- infoText (~8:52) when the mapeffect lines are sent that calls the inside safe tile and the corner safe tile (e.g. "Safe Tiles: Inside NE / Corner SW".
- alertText (~8:57) when the two stack markers & two aoe markers appear that calls out the side the player should be on and whether they should stack or spread based on what headmarker they get (or none)  (e.g. "Stack East" or "Spread West")
- cach2 resolves (~9:02)

I like this approach because it calls the just enough of the safe tiles, and early enough, for players to find at least one safe tile on the correct side, and it then gives a player-specific instruction to allow them to resolve whether they should be in the "north safe" or "south safe" tile on their east/west side (based on whatever strat their group is doing).  What do you think?

quisquous: You know at 8:52 which side you're on, so it could just call the safe spots on your side, e.g. `Safe Tiles: Inside NE, Outside SSE`? Otherwise I think this is good.

ui/raidboss/data/06-ew/raid/p6s.ts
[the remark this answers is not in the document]
quisquous: I am personally not sure what people do.  I'm in a static, but I also looked at some toolboxes and everybody seemed to split E/W.  I think that's fine to keep it split up.  I think even if both stacks are on the inside side, saying "inside E/outside W" will still be clear where to go.

I guess my other feedback is that if it's not going to be combined, then I think "safe tiles: X" is a bit verbose, probably especially for folks who are using TTS.  What do you think about dropping the "safe tiles" part, or do you think it's confusing without it?

ui/raidboss/data/06-ew/raid/p6s.ts
quisquous: Yeah, that's exactly what I was thinking originally.  Although, in the other comment you say this is too much information in one line, so maybe it needs to be split up still.

ui/raidboss/data/06-ew/raid/p6s.ts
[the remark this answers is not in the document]
wexxlee: > What do you think about dropping the "safe tiles" part, or do you think it's confusing without it?

Good point - I use the text alerts not tts, so wasn't thinking about that.  I think we can drop "safe tiles" and people will be able to get quickly what it refers to, especially as it fires during the mapeffects and before other mechanics start happening.

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: Works for me.

ui/raidboss/data/06-ew/raid/p6s.ts
wexxlee: > Yeah, that's exactly what I was thinking originally. Although, in the other comment you say this is too much information in one line, so maybe it needs to be split up still.

Totally fair - and yeah, I do think it should be split up.  So giving this more thought, and with where we've landed for the other poly calls, I think it makes sense to have this poly trigger do infoText with the safe tiles (as mapeffects render), and then do a separate trigger with alertText for the healer groups when boss applies unholy darkness (but excluding the first instance during poly1 with exocleavers, since we already have a trigger for that that includes the exos).  Realize this is a little different than what I said above, but (1) it avoids having to turn this whole trigger into a response trigger just to handle dual text for these mechanics, (2) allows the safe tiles to be called early to give people time before the healer group trigger fires, and (3) keeps it as consistent as possible with how we're handling most of the other polys (safe tiles in infotext, other mechs in alerttext).

quisquous: Yeah, sounds good.

wexxlee:
@quisquous Should be ready for you!  Thanks for all the support & help!

quisquous:
I think this is all good.  I think my only minor followup comment is that poly6 maybe should still call out the two safe spots immediately so folks can kinda preposition better, and then narrow it down to one with the chorus ixou call.

PR #4905, opened 2022-10-02
docs: add remaining new log lines from OverlayPlugin

quisquous:

PR #4908, opened 2022-10-03
resources: rename field in MapEffect line

quisquous:
Mentioned by @valarnin:
https://github.com/quisquous/cactbot/pull/4905#discussion_r985769682

### DIFF
--- a/docs/LogGuide.md
+++ b/docs/LogGuide.md
@@ -2232,20 +2232,20 @@ This can also include things like:
 
 ```log
 Network Log Line Structure:
-257|[timestamp]|[instance]|[id]|[location]|[data0]|[data1]
+257|[timestamp]|[instance]|[flags]|[location]|[data0]|[data1]
 
 ACT Log Line Structure:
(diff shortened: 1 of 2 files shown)


quisquous:
also @wexxlee just FYI

wexxlee:
> also @wexxlee just FYI

100% agree with @valarnin that it's a bitmask.  My only thought with labeling it as 'id', rather than by its datatype, was that it's functionally the field that identifies which effect is rendered, and 'id' is a label used for similar purposes in other log lines (thinking specifically of NetworkTargetIcon).  But either way is fine with me - I'll update #4907 if/once this is merged.

wexxlee:
Oh, and if the field name does change, there probably should also be an update to the log guide, as the field is mentioned in the description text.

quisquous:
> Oh, and if the field name does change, there probably should also be an update to the log guide, as the field is mentioned in the description text.

Oh, thanks.  I forgot about that.  `*<_<*`

PR #4918, opened 2022-10-07
raidboss: fix entering combat resetting `data`

quisquous:
Calling `getDataObject` returns a `RaidbossData` that does not have `initData` called on it.  Prior to this patch, when entering combat, `data` would be set to a new `data` object but without `initData` called on it.

This isn't a larger issue because if `resetWhenOutOfCombat` is true, then `this.Reset()` is called again in `SetInCombat`, fixing the issues.  This only is an issue for the s-rank hunts in Thav and Mare that have this flag set.

To avoid this problem in the future, always have `getDataObject` return a properly initialized object.  To fix the short-term issue, update `this.data.inCombat` directly without calling `getDataObject` (oops).

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -832,9 +832,7 @@ export class PopupText {
       return;
 
     this.inCombat = inCombat;
-
-    // Update in combat
-    this.data = this.getDataObject();
+    this.data.inCombat = inCombat;
 
     if (!this.resetWhenOutOfCombat)
       return;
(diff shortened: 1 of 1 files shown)


panicstevenson (approved):

trim21:
Sorry, I just say this change, don't know if it's too late to give opinion.

I think it would be better to set base keys of `RaidbossData` after we merge all `dataInitializers` so them won't get overwriteed accidentally

quisquous:
Seems reasonable to me.  I guess we don't have anything in the type that says that the keys can't be the same (I guess we could do that too).  Do you want to put up a PR?

quisquous:
And late opinions are more than fine! I appreciate the thoughts.

PR #4921, opened 2022-10-08
raidboss: add trigger for P6S transmission

wexxlee:


ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -471,6 +473,34 @@ const triggerSet: TriggerSet<Data> = {
...
+      },
+      infoText: (_data, matches, output) => {
+        return matches.effectId === 'D48' ? output.forwardCleave!() : output.backwardCleave!();
+      },
+      outputStrings: {
+        forwardCleave: {
+          en: 'Cleave in Front (Look Away)',
+        },
+        backwardCleave: {
+          en: 'Cleave Behind (Face Boss)',
```
panicstevenson, line 500: Just because English is quirky and "look" and "face" can mean the same thing here but wildly different things in other contexts, can we standardize these two parentheses to have more similar verbiage, ie. "Look Out" / "Look In" or similar? Or does it feel like there's value in having them be different enough that they don't bleed together mentally during raid?

quisquous: Another suggestion is "Front Cleave" and "Rear Cleave" to keep it short.

wexxlee: Thanks - didn't think about the linguistic nuance.  I struggled a bit with how prescriptive to be on where to look, since, e.g., some groups have all players face the boss (at least during trans1) and just put snakes in front/wings behind, so everyone can face the boss..  I think "Front Cleave" and "Rear Cleave" is a good solution - and if people want a specific instruction about where to aim it, they can always modify it in config. :-)

ui/raidboss/data/06-ew/raid/p6s.ts
```diff
@@ -471,6 +473,34 @@ const triggerSet: TriggerSet<Data> = {
...
+      condition: Conditions.targetIsYou(),
+      preRun: (data, matches) => {
+        // 1st transmission has 11s duration, 2nd has 25s duration
+        // in either case, trigger should fire 3s before debuff expires
+        const duration = parseFloat(matches.duration);
+        data.transmissionAlertDelay = duration > 3 ? duration - 3 : 0;
+      },
+      delaySeconds: (data) => {
+        return data.transmissionAlertDelay;
+      },
```
quisquous, line 491: Not sure why you need a variable here.  I think you can do this just in the function itself.

```suggestion
      delaySeconds: (data, matches) => {
        // 1st transmission has 11s duration, 2nd has 25s duration
        // in either case, trigger should fire 3s before debuff expires
        return parseFloat(matches.duration) - 3;
      },
```

wexxlee: Wasn't sure at the time if I would also be using it to manipulate the timing on the cachexia2 call to prevent too many calls from firing at the same time, but I think it should be okay.

quisquous:
Your other p6s patch conflicts with this one.  Would you mind fixing the conflicts, and then I'll land this?

wexxlee:
> Your other p6s patch conflicts with this one. Would you mind fixing the conflicts, and then I'll land this?

Done!

PR #4926, opened 2022-10-11
plugin: CE opcode for cn 6.15

ShadyWhite:
Contains some cn translation fixes before 6.2, due to few changes, do not open PR separately.


quisquous:
> due to few changes, do not open PR separately.

I think in the future I'd appreciate two PRs.  I think PRs should only be doing one thing, even if that thing is small.  Doing two separate things can make it confusing (in changelog, in history, etc).

PR #4928, opened 2022-10-12
raidboss: p8s_vents remove ability requirement

Legends0:
Switches second vents to use all 8 directions and same method regardless of ability id.

I think there may be something wrong with the ability id part of the current PR and this PR changes it to be ability agnostic.

It's possible to merge the two triggers (initial and later), but there would need to be extra code to check the first one since we need to ignore heading on that trigger so I have kept the two triggers.


Legends0:
This is to be merged with #4830

Legends0:
Re-created the PR under 4929 as this PR had incorrect author settings.

PR #4929, opened 2022-10-12
raidboss: p8s_vents possible fix

Legends0:
Switches second vents to use all 8 directions and same method regardless of ability id.

I think there may be something wrong with the ability id part of the current PR and this PR changes it to be ability agnostic.

It's possible to merge the two triggers (initial and later), but there would need to be extra code to check the first one since we need to ignore heading on that trigger so I have kept the two triggers.


Legends0:
To be merged to #4830

Legends0:
oh, regarding b29bd31, it was that way in the code I ran with, but I think it should have remained untouched from the automerge given the order in which the triggers occur.

ui/raidboss/data/06-ew/raid/p8s.ts
```diff
@@ -75,9 +75,9 @@ export const positionTo8Dir = (combatant: PluginCombatantState) => {
   return Math.round(4 - 4 * Math.atan2(x, y) / Math.PI) % 8;
 };
 
-export const headingTo4Dir = (heading: number) => {
-  // N = 0, E = 1, etc
-  return Math.round(2 - 2 * heading / Math.PI) % 4;
+export const headingTo8Dir = (heading: number) => {
+  // Dirs: N = 0, N = 1, ..., W = 7
```
wexxlee, line 79: If N = 0, should NE = 1?  And would NW = 7?  Or are these being done in pairs (e.g. N=[0,1], E=[2,3], etc)?

Legends0: oh it's a typo in the comment, NE is 1

Legends0: thanks

wexxlee: Np.  I think you might still have a typo on W = 7 though.

Legends0: oh man, that's embarrassing.  I've fixed it to be NW.

wexxlee:
Tested this a bunch tonight and it worked every time.

quisquous:
Yeah, this seems good to me too, thanks.  Maybe I was wrong re: the ability ids.  Super appreciate you making this work.

PR #4934, opened 2022-10-15
i18n: update ko p6s translations

Bing-su:
Cap 2022-10-15 12-49-19-154
Cap 2022-10-15 12-49-23-878

is this right? It's confusing...


wexxlee:
> is this right? It's confusing...

Yes, that looks right.  Sorry for the confusion.

PR #4937, opened 2022-10-16
build/ci: update to nodejs 18 LTS

trim21:
node18 will be lts 2022-10-25

lots of warning about unstable node flags disappear 😄


trim21:
And it looks like we don't need ttypescriot anymore, I'll test if it's ok to remove it

MaikoTan:
is there possible to have a property as "lts" or something to get node.js always at the LTS version, rather than set it every iteration.

quisquous:
> is there possible to have a property as "lts" or something to get node.js always at the LTS version, rather than set it every iteration?

I would rather bump it manually after we're sure that it works rather than have it automatically update it.  I know that's extra work, but it'd probably save some headaches for some future person trying to do local cactbot stuff for the first time.

trim21:
> is there possible to have a property as "lts" or something to get node.js always at the LTS version, rather than set it every iteration?

`node-version： "lts/*"`

trim21:
> And it looks like we don't need ttypescriot anymore, I'll test if it's ok to remove it

it's not worth it.

trim21:
chrome95 and node18 support all es2021 features, we can imporve our ts target too

trim21:
18 is LTS now https://nodejs.org/en/

quisquous:
Thanks.  I will try to take a look at getting this merged in the next few days.

PR #4939, opened 2022-10-16
raidboss: set base raidboss data after `initData`

trim21:


trim21:
... get annoying whitespace diff again

PR #4943, opened 2022-10-18
resources: remove unused file

trim21:


trim21:
we need a new `types` or `type` scope.

quisquous:
> we need a new `types` or `type` scope.

Yeah, that's possible.  I think it just hasn't come up very much.  Usually when changing types, it's because you're doing it for some other purpose and not just cleanup.  I'm not sure what the best scope here is.  "build" is an option too, but this is good enough.

Issue #4953, opened 2022-10-19
drop `eslint-plugin-dprint`


trim21:
`eslint-plugin-dprint` is using a embed ts plugin `typescript-0.32.4.wasm`, and it has been pretty old.

It can't format some of our ts files already

eg: 

https://github.com/quisquous/cactbot/blob/0c1c65a8b40992a4fa19c5099620e24e1ad19cfe/ui/jobs/components/blu.ts#L6-L11

You can try change dprint config in .eslintrc.js and reformat it with eslint, nothing will change because when you formatting ts with this plugin, you will get this:

```text
Error formatting C:/Users/Trim21/proj/cactbot/ui/jobs/components/blu.ts. Message: Line 31, column 12: Unexpected token `onUseAbility`. Expected * for generator, private key, identifier or async

    override onUseAbility(id: strin
             ~~~~~~~~~~~~
Had 1 error(s) formatting.
```

quisquous:
Do we just need to upgrade? https://github.com/dprint/dprint-plugin-typescript/blob/b8589d04f456283907fc2a5632f4bd30280d1354/dprint.json

quisquous:
I'll think about this.  We went through a bunch of options to setting on eslint+dprint and I don't want to remove it entirely without having an alternative in mind first.

trim21:
Another problem caused by this plugin, eslint `parserOptions.ecmaVersion` 2022 require eslint8, and doesn't work with this plugin.

PR #4954, opened 2022-10-19
lint: remove `eslint-plugin-dprint`

trim21:
close https://github.com/quisquous/cactbot/issues/4953

dprint plugin is removed so some style error can't be fixed by eslint automatically

I try to config dprint to make less different, but this is the best I can get.

something major: 

## trigger branch

Just run dprint on `dist` dir after all files are transpiled, don't run eslint on each file.

## import order

dprint now support sorting importing, it works fine with `import/order`

## different line width

this is not changed, triggers files still have 300 max line witdh.

This would be much more simpler is we use same max length width in trigger and non-trigger files, considering we don't have `netRegex{cn/de/jp}` anymore.


package.json
```diff
@@ -115,10 +121,18 @@
   "lint-staged": {
     "*.css": "stylelint",
     "*.{js,cjs,mjs}": "eslint --max-warnings=0",
-    "*.ts": "eslint --max-warnings=0 --plugin tsc --rule 'tsc/config: [2, {configFile: \"./tsconfig.json\"}]'",
+    "**/!(raidboss/data|oopsyraidsy/data)/*.ts": [
+      "dprint check"
+    ],
```
quisquous, line 129: Is this the only way to handle multiple dprint config files?

Does that mean if I run `dprint check` from the command-line, it'll reformat trigger files incorrectly?

package.json
```diff
@@ -14,8 +14,14 @@
     "start": "webpack serve --config webpack/webpack.dev.ts",
     "fetch-deps": "ts-node ./util/fetch_deps.ts",
     "tsc-no-emit": "tsc --noEmit",
-    "lint": "eslint --max-warnings=0 ./",
-    "lintfix": "eslint --max-warnings=0 --fix ./",
+    "lint": "run-s lint:**",
+    "lint:dprint": "dprint check",
+    "lint:dprint:trigger": "dprint --config=dprint.trigger.json check",
+    "lint:eslint": "eslint --max-warnings=0 ./",
```
quisquous, line 20: Do we need to teach VSCode about dprint as well? Currently it knows about eslint and so there's errors when things are formatted incorrectly.  Is it possible to do this with dprint too?

util/process_triggers_folder.ts
```diff
@@ -75,46 +75,6 @@ const changeExportToPush = (lines: string[]) => {
...
-
-  // Run without dprint first, because dprint is slow on unformatted files.
-  const linter = new eslint.ESLint({
-    fix: true,
-    useEslintrc: false,
-    baseConfig: config,
-  });
-  const contents = lines.join('\n');
-  const eslintResults = await linter.lintText(contents, { filePath: filename });
-  const eslintLintResult = eslintResults[0];
```
quisquous, line 95: With this patch, files are formatted with dprint and then eslint.  Shouldn't the triggers folder be formatted the same way to match? Why remove eslint?

ui/dps/rdmty/dps.js
```diff
@@ -1,5 +1,5 @@
-import { defaultOptions, InitDpsModule } from '../dps_common';
 import UserConfig from '../../../resources/user_config';
+import { defaultOptions, InitDpsModule } from '../dps_common';
```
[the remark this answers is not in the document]
trim21: ok, I'll set this file to be ignored.

package.json
trim21: No it won't. actually `dprint check` will ignore trigger files. and just like `dprint fmt`.

dprint doesn't support config override like prettier https://prettier.io/docs/en/configuration.html#configuration-overrides

package.json
trim21: I didn't using vscode for a lont time.....

util/process_triggers_folder.ts
trim21: This patch remove eslint in fact.

Do you think the diff of triggers branch is important?

package.json
trim21: And I think dprint extension of vscode doesn't support multiple config file either.

package.json
trim21: maybe adding a `format` script

util/process_triggers_folder.ts
quisquous: I personally think it's nice to have the triggers branch look like cactbot code as much as possible.

dprint.json
```diff
@@ -7,6 +7,7 @@
   "excludes": [
     "**/node_modules",
     "**/ui/raidboss/data",
+    "**/ui/dps",
```
quisquous, line 10: I think it's not all of ui/dps, it's just `ui/dps/rdmty/dps.js`

trim21: Oh, I'll just add a dprint ignore comment to this file in this case.

util/process_triggers_folder.ts
trim21: I think they are same, only without ts types

this is diff of 06-ew/ultimate/dragonsongs_reprise_ultimate before and after processing

```diff
diff --git a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
index 44497b924..41035615d 100644
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -1,67 +1,4 @@
-import Conditions from '../../../../../resources/conditions';
-import NetRegexes from '../../../../../resources/netregexes';
-import { UnreachableCode } from '../../../../../resources/not_reached';
-import Outputs from '../../../../../resources/outputs';
-import { callOverlayHandler } from '../../../../../resources/overlay_plugin_api';
-import { Responses } from '../../../../../resources/responses';
-import ZoneId from '../../../../../resources/zone_id';
-import { RaidbossData } from '../../../../../types/data';
-import { PluginCombatantState } from '../../../../../types/event';
-import { NetMatches } from '../../../../../types/net_matches';
-import { LocaleObject, LocaleText, TriggerSet } from '../../../../../types/trigger';
-
-// TODO: Ser Adelphel left/right movement after initial charge
-// TODO: Meteor "run" call?
-// TODO: Wyrmsbreath 2 cardinal positions for Cauterize and adjust delay
-// TODO: Trigger for Hallowed Wings with Hot Tail/Hot Wings
-// TODO: Phase 6 Resentment callout?
-
-type Phase = 'doorboss' | 'thordan' | 'nidhogg' | 'haurchefant' | 'thordan2' | 'nidhogg2' | 'dragon-king';
-
-const playstationMarkers = ['circle', 'cross', 'triangle', 'square'] as const;
-type PlaystationMarker = typeof playstationMarkers[number];
-
-export interface Data extends RaidbossData {
-  combatantData: PluginCombatantState[];
-  phase: Phase;
-  decOffset?: number;
-  seenEmptyDimension?: boolean;
-  adelphelId?: string;
-  firstAdelphelJump: boolean;
-  adelphelDir?: number;
-  brightwingCounter: number;
-  spiralThrustSafeZones?: number[];
-  sanctityWardDir?: string;
-  sanctitySword1?: string;
-  sanctitySword2?: string;
-  thordanMeteorMarkers: string[];
-  // mapping of player name to 1, 2, 3 dot.
-  diveFromGraceNum: { [name: string]: number };
-  // mapping of 1, 2, 3 to whether that group has seen an arrow.
-  diveFromGraceHasArrow: { [num: number]: boolean };
-  diveFromGraceDir: { [name: string]: 'circle' | 'up' | 'down' };
-  diveFromGraceLashGnashKey: 'in' | 'out' | 'unknown';
-  // mapping of player name to x coordinate
-  diveFromGracePositions: { [name: string]: number };
-  diveFromGraceTowerCounter?: number;
-  eyeOfTheTyrantCounter: number;
-  diveFromGracePreviousPosition: { [num: string]: 'middle' | 'west' | 'east' };
-  waitingForGeirskogul?: boolean;
-  stackAfterGeirskogul?: boolean;
-  diveCounter: number;
-  // names of players with chain lightning during wrath.
-  thunderstruck: string[];
-  hasDoom: { [name: string]: boolean };
-  deathMarker: { [name: string]: PlaystationMarker };
-  addsPhaseNidhoggId?: string;
-  hraesvelgrGlowing?: boolean;
-  nidhoggGlowing?: boolean;
-  hallowedWingsCount: number;
-  spreadingFlame: string[];
-  entangledFlame: string[];
-  mortalVowPlayer?: string;
-}
-
+const playstationMarkers = ['circle', 'cross', 'triangle', 'square'];
 // Due to changes introduced in patch 5.2, overhead markers now have a random offset
 // added to their ID. This offset currently appears to be set per instance, so
 // we can determine what it is from the first overhead marker we see.
@@ -88,27 +25,23 @@ const headmarkers = {
   'skywardSingle': '000E',
   // vfx/lockon/eff/bahamut_wyvn_glider_target_02tm.avfx
   'cauterize': '0014',
-} as const;
-
-const playstationHeadmarkerIds: readonly string[] = [
+};
+const playstationHeadmarkerIds = [
   headmarkers.firechainCircle,
   headmarkers.firechainTriangle,
   headmarkers.firechainSquare,
   headmarkers.firechainX,
-] as const;
-
-const playstationMarkerMap: { [id: string]: PlaystationMarker } = {
+];
+const playstationMarkerMap = {
   [headmarkers.firechainCircle]: 'circle',
   [headmarkers.firechainTriangle]: 'triangle',
   [headmarkers.firechainSquare]: 'square',
   [headmarkers.firechainX]: 'cross',
-} as const;
-
-const firstMarker = (phase: Phase) => {
+};
+const firstMarker = (phase) => {
   return phase === 'doorboss' ? headmarkers.hyperdimensionalSlash : headmarkers.skywardTriple;
 };
-
-const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker'], firstDecimalMarker?: number) => {
+const getHeadmarkerId = (data, matches, firstDecimalMarker) => {
   // If we naively just check !data.decOffset and leave it, it breaks if the first marker is 00DA.
   // (This makes the offset 0, and !0 is true.)
   if (data.decOffset === undefined) {
@@ -122,38 +55,31 @@ const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker'], firstDec
   // since we know all the IDs that will be present in the encounter.
   return (parseInt(matches.id, 16) - data.decOffset).toString(16).toUpperCase().padStart(4, '0');
 };
-
 // Calculate combatant position in an all 8 cards/intercards
-const matchedPositionTo8Dir = (combatant: PluginCombatantState) => {
+const matchedPositionTo8Dir = (combatant) => {
   // Positions are moved up 100 and right 100
   const y = combatant.PosY - 100;
   const x = combatant.PosX - 100;
-
   // During Thordan, knight dives start at the 8 cardinals + numerical
   // slop on a radius=23 circle.
   // N = (100, 77), E = (123, 100), S = (100, 123), W = (77, 100)
   // NE = (116.26, 83.74), SE = (116.26, 116.26), SW = (83.74, 116.26), NW = (83.74, 83.74)
   //
   // Map NW = 0, N = 1, ..., W = 7
-
   return (Math.round(5 - 4 * Math.atan2(x, y) / Math.PI) % 8);
 };
-
 // Calculate combatant position in 4 cardinals
-const matchedPositionTo4Dir = (combatant: PluginCombatantState) => {
+const matchedPositionTo4Dir = (combatant) => {
   // Positions are moved up 100 and right 100
   const y = combatant.PosY - 100;
   const x = combatant.PosX - 100;
-
   // During the vault knights, Adelphel will jump to one of the 4 cardinals
   // N = (100, 78), E = (122, 100), S = (100, 122), W = (78, 100)
   //
   // N = 0, E = 1, S = 2, W = 3
-
   return (Math.round(2 - 2 * Math.atan2(x, y) / Math.PI) % 4);
 };
-
-const triggerSet: TriggerSet<Data> = {
+Options.Triggers.push({
   zoneId: ZoneId.DragonsongsRepriseUltimate,
   timelineFile: 'dragonsongs_reprise_ultimate.txt',
   initData: () => {
@@ -191,7 +117,7 @@ const triggerSet: TriggerSet<Data> = {
       regex: /Resentment/,
       beforeSeconds: 5,
       condition: (data) => data.phase === 'nidhogg',
-      infoText: (_data, _matches, output) => output.text!(),
+      infoText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'aoe + dot',
@@ -211,10 +137,10 @@ const triggerSet: TriggerSet<Data> = {
       durationSeconds: 3.7,
       infoText: (data, _matches, output) => {
         if (data.me === data.mortalVowPlayer)
-          return output.vowOnYou!();
+          return output.vowOnYou();
         if (data.mortalVowPlayer)
-          return output.vowOn!({ player: data.mortalVowPlayer });
-        return output.vowSoon!();
+          return output.vowOn({ player: data.mortalVowPlayer });
+        return output.vowSoon();
       },
       outputStrings: {
         vowOnYou: {
@@ -254,7 +180,6 @@ const triggerSet: TriggerSet<Data> = {
       run: (data, matches) => {
         // On the unlikely chance that somebody proceeds directly from the checkpoint into the next phase.
         data.brightwingCounter = 1;
-
         switch (matches.id) {
           case '62D4':
             data.phase = 'doorboss';
@@ -301,7 +226,7 @@ const triggerSet: TriggerSet<Data> = {
       condition: (data) => data.decOffset === undefined,
       // Unconditionally set the first headmarker here so that future triggers are conditional.
       run: (data, matches) => {
-        const firstHeadmarker: number = parseInt(firstMarker(data.phase), 16);
+        const firstHeadmarker = parseInt(firstMarker(data.phase), 16);
         getHeadmarkerId(data, matches, firstHeadmarker);
       },
     },
@@ -322,7 +247,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '62DA', source: 'Ser Grinnaux', capture: false }),
       alertText: (data, _matches, output) => {
-        return data.phase !== 'doorboss' || data.seenEmptyDimension ? output.in!() : output.inAndTether!();
+        return data.phase !== 'doorboss' || data.seenEmptyDimension ? output.in() : output.inAndTether();
       },
       run: (data) => data.seenEmptyDimension = true,
       outputStrings: {
@@ -363,7 +288,7 @@ const triggerSet: TriggerSet<Data> = {
       alertText: (data, matches, output) => {
         const id = getHeadmarkerId(data, matches);
         if (id === headmarkers.hyperdimensionalSlash)
-          return output.slashOnYou!();
+          return output.slashOnYou();
       },
       outputStrings: {
         slashOnYou: {
@@ -398,7 +323,6 @@ const triggerSet: TriggerSet<Data> = {
           call: 'getCombatants',
           ids: [parseInt(matches.id, 16)],
         });
-
         // if we could not retrieve combatant data, the
         // trigger will not work, so just resume promise here
         if (adelphelData === null) {
@@ -410,7 +334,6 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`Ser Adelphel: expected 1 combatants got ${adelphelDataLength}`);
           return;
         }
-
         // Add the combatant's position
         const adelphel = adelphelData.combatants.pop();
         if (!adelphel)
@@ -419,14 +342,14 @@ const triggerSet: TriggerSet<Data> = {
       },
       infoText: (data, _matches, output) => {
         // Map of directions, reversed to call out KB direction
-        const dirs: { [dir: number]: string } = {
-          0: output.south!(),
-          1: output.west!(),
-          2: output.north!(),
-          3: output.east!(),
-          4: output.unknown!(),
+        const dirs = {
+          0: output.south(),
+          1: output.west(),
+          2: output.north(),
+          3: output.east(),
+          4: output.unknown(),
         };
-        return output.adelphelLocation!({
+        return output.adelphelLocation({
           dir: dirs[data.adelphelDir ?? 4],
         });
       },
@@ -454,13 +377,13 @@ const triggerSet: TriggerSet<Data> = {
       alertText: (data, matches, output) => {
         const id = getHeadmarkerId(data, matches);
         if (id === headmarkers.firechainCircle)
-          return output.circle!();
+          return output.circle();
         if (id === headmarkers.firechainTriangle)
-          return output.triangle!();
+          return output.triangle();
         if (id === headmarkers.firechainSquare)
-          return output.square!();
+          return output.square();
         if (id === headmarkers.firechainX)
-          return output.cross!();
+          return output.cross();
       },
       outputStrings: {
         circle: {
@@ -506,7 +429,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.ability({ id: '6319', capture: false }),
       // One ability for each player hit (hopefully only two??)
       suppressSeconds: 1,
-      infoText: (data, _matches, output) => output[`dive${data.brightwingCounter}`]!(),
+      infoText: (data, _matches, output) => output[`dive${data.brightwingCounter}`](),
       run: (data) => data.brightwingCounter++,
       outputStrings: {
         // Ideally folks can customize this with who needs to run in.
@@ -551,7 +474,7 @@ const triggerSet: TriggerSet<Data> = {
       delaySeconds: 5,
       promise: async (data) => {
         // Collect Ser Vellguine (3636), Ser Paulecrain (3637), Ser Ignasse (3638) entities
-        const vellguineLocaleNames: LocaleText = {
+        const vellguineLocaleNames = {
           en: 'Ser Vellguine',
           de: 'Vellguine',
           fr: 'sire Vellguine',
@@ -559,8 +482,7 @@ const triggerSet: TriggerSet<Data> = {
           cn: '圣骑士韦尔吉纳',
           ko: '성기사 벨긴',
         };
-
-        const paulecrainLocaleNames: LocaleText = {
+        const paulecrainLocaleNames = {
           en: 'Ser Paulecrain',
           de: 'Paulecrain',
           fr: 'sire Paulecrain',
@@ -568,8 +490,7 @@ const triggerSet: TriggerSet<Data> = {
           cn: '圣骑士波勒克兰',
           ko: '성기사 폴르크랭',
         };
-
-        const ignasseLocaleNames: LocaleText = {
+        const ignasseLocaleNames = {
           en: 'Ser Ignasse',
           de: 'Ignasse',
           fr: 'sire Ignassel',
@@ -577,20 +498,16 @@ const triggerSet: TriggerSet<Data> = {
           cn: '圣骑士伊尼亚斯',
           ko: '성기사 이냐스',
         };
-
         // Select the knights
-        const combatantNameKnights: string[] = [];
+        const combatantNameKnights = [];
         combatantNameKnights.push(vellguineLocaleNames[data.parserLang] ?? vellguineLocaleNames['en']);
         combatantNameKnights.push(paulecrainLocaleNames[data.parserLang] ?? paulecrainLocaleNames['en']);
         combatantNameKnights.push(ignasseLocaleNames[data.parserLang] ?? ignasseLocaleNames['en']);
-
         const spiralThrusts = [];
-
         const knightCombatantData = await callOverlayHandler({
           call: 'getCombatants',
           names: combatantNameKnights,
         });
-
         // if we could not retrieve combatant data, the
         // trigger will not work, so just resume promise here
         if (knightCombatantData === null) {
@@ -602,17 +519,13 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`Spiral Thrust: expected 3 combatants got ${combatantDataLength}`);
           return;
         }
-
         for (const combatant of knightCombatantData.combatants)
           spiralThrusts.push(matchedPositionTo8Dir(combatant));
-
         const [thrust0, thrust1, thrust2] = spiralThrusts;
         if (thrust0 === undefined || thrust1 === undefined || thrust2 === undefined)
           return;
-
         // Array of dirNums
         const dirNums = [0, 1, 2, 3, 4, 5, 6, 7];
-
         // Remove where the knights are at and where they will go to
         delete dirNums[(thrust0 + 4) % 8];
         delete dirNums[thrust0];
@@ -620,32 +533,31 @@ const triggerSet: TriggerSet<Data> = {
         delete dirNums[thrust1];
         delete dirNums[(thrust2 + 4) % 8];
         delete dirNums[thrust2];
-
         // Remove null elements from the array to get remaining two dirNums
         dirNums.forEach((dirNum) => {
           if (dirNum !== null)
-            (data.spiralThrustSafeZones ??= []).push(dirNum);
+            (data.spiralThrustSafeZones ?? (data.spiralThrustSafeZones = [])).push(dirNum);
         });
       },
       infoText: (data, _matches, output) => {
-        data.spiralThrustSafeZones ??= [];
+        data.spiralThrustSafeZones ?? (data.spiralThrustSafeZones = []);
         if (data.spiralThrustSafeZones.length !== 2) {
           console.error(`Spiral Thrusts: expected 2 safe zones got ${data.spiralThrustSafeZones.length}`);
           return;
         }
         // Map of directions
-        const dirs: { [dir: number]: string } = {
-          0: output.northwest!(),
-          1: output.north!(),
-          2: output.northeast!(),
-          3: output.east!(),
-          4: output.southeast!(),
-          5: output.south!(),
-          6: output.southwest!(),
-          7: output.west!(),
-          8: output.unknown!(),
+        const dirs = {
+          0: output.northwest(),
+          1: output.north(),
+          2: output.northeast(),
+          3: output.east(),
+          4: output.southeast(),
+          5: output.south(),
+          6: output.southwest(),
+          7: output.west(),
+          8: output.unknown(),
         };
-        return output.safeSpots!({
+        return output.safeSpots({
           dir1: dirs[data.spiralThrustSafeZones[0] ?? 8],
           dir2: dirs[data.spiralThrustSafeZones[1] ?? 8],
         });
@@ -681,7 +593,7 @@ const triggerSet: TriggerSet<Data> = {
         // Therefore, use names.  There should be exactly one of each.
         // TODO: maybe we need data function to do this sort of translating so it's
         // not duplicating the timelineReplace section below.
-        const names: LocaleObject<string[]> = {
+        const names = {
           en: ['Ser Adelphel', 'Ser Janlenoux'],
           de: ['Adelphel', 'Janlenoux'],
           fr: ['sire Adelphel', 'sire Janlenoux'],
@@ -689,7 +601,6 @@ const triggerSet: TriggerSet<Data> = {
           cn: ['圣骑士阿代尔斐尔', '圣骑士让勒努'],
           ko: ['성기사 아델펠', '성기사 장르누'],
         };
-
         const combatantNames = names[data.parserLang] ?? names['en'];
         data.combatantData = (await callOverlayHandler({
           call: 'getCombatants',
@@ -705,40 +616,36 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`DragonsRage: wrong length: ${JSON.stringify(data.combatantData)}`);
           return;
         }
-
         // Ser Adelphel and Ser Janlenoux appear on the field in two spots.  It is random which
         // side they are on.  Thordan appears opposite of them.  If they are SW and SE, then
         // Thordan is N.  They will always be two spaces apart.
         let d1 = matchedPositionTo8Dir(c1);
         let d2 = matchedPositionTo8Dir(c2);
-
         // Adjust for wrapping around n=0/8 so that we can average the two points below.
         if (d2 === 6 && d1 === 0 || d2 === 7 && d1 === 1)
           d1 += 8;
         if (d1 === 6 && d2 === 0 || d1 === 7 && d2 === 1)
           d2 += 8;
-
         // After the above adjustment to handle modular math wrapping,
         // d1 and d2 should be exactly two spaces apart.
         if (d2 - d1 !== 2 && d1 - d2 !== 2) {
           console.error(`DragonsRage: bad dirs: ${d1}, ${d2}, ${JSON.stringify(data.combatantData)}`);
           return;
         }
-
         // Average to find the point between d1 and d2, then add 4 to find its opposite.
         const thordanDir = (Math.floor((d1 + d2) / 2) + 4) % 8;
-        const dirs: { [dir: number]: string } = {
-          0: output.northwest!(),
-          1: output.north!(),
-          2: output.northeast!(),
-          3: output.east!(),
-          4: output.southeast!(),
-          5: output.south!(),
-          6: output.southwest!(),
-          7: output.west!(),
+        const dirs = {
+          0: output.northwest(),
+          1: output.north(),
+          2: output.northeast(),
+          3: output.east(),
+          4: output.southeast(),
+          5: output.south(),
+          6: output.southwest(),
+          7: output.west(),
         };
-        return output.thordanLocation!({
-          dir: dirs[thordanDir] ?? output.unknown!(),
+        return output.thordanLocation({
+          dir: dirs[thordanDir] ?? output.unknown(),
         });
       },
       outputStrings: {
@@ -768,7 +675,7 @@ const triggerSet: TriggerSet<Data> = {
       alertText: (data, matches, output) => {
         const id = getHeadmarkerId(data, matches);
         if (id === headmarkers.skywardTriple)
-          return output.leapOnYou!();
+          return output.leapOnYou();
       },
       outputStrings: {
         leapOnYou: {
@@ -810,7 +717,7 @@ const triggerSet: TriggerSet<Data> = {
         // Thus we can just use one knight and get whether it is east or west for location.
         // Callout assumes starting from DRK position, but for east/west (two-movement strategy)
         // you can reverse the cw/ccw callout.
-        const janlenouxLocaleNames: LocaleText = {
+        const janlenouxLocaleNames = {
           en: 'Ser Janlenoux',
           de: 'Janlenoux',
           fr: 'sire Janlenoux',
@@ -818,11 +725,9 @@ const triggerSet: TriggerSet<Data> = {
           cn: '圣骑士让勒努',
           ko: '성기사 장르누',
         };
-
         // Select Ser Janlenoux
         let combatantNameJanlenoux = null;
         combatantNameJanlenoux = janlenouxLocaleNames[data.parserLang];
-
         let combatantDataJanlenoux = null;
         if (combatantNameJanlenoux) {
           combatantDataJanlenoux = await callOverlayHandler({
@@ -830,7 +735,6 @@ const triggerSet: TriggerSet<Data> = {
             names: [combatantNameJanlenoux],
           });
         }
-
         // if we could not retrieve combatant data, the
         // trigger will not work, so just resume promise here
         if (combatantDataJanlenoux === null) {
@@ -842,21 +746,19 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`Ser Janlenoux: expected at least 1 combatants got ${combatantDataJanlenouxLength}`);
           return;
         }
-
         // Sort to retreive last combatant in list
-        const sortCombatants = (a: PluginCombatantState, b: PluginCombatantState) => (a.ID ?? 0) - (b.ID ?? 0);
+        const sortCombatants = (a, b) => (a.ID ?? 0) - (b.ID ?? 0);
         const combatantJanlenoux = combatantDataJanlenoux.combatants.sort(sortCombatants).shift();
         if (!combatantJanlenoux)
           throw new UnreachableCode();
-
         // West (95, 100) or East (105, 100).
         if (combatantJanlenoux.PosX < 100)
-          data.sanctityWardDir = output.clockwise!();
+          data.sanctityWardDir = output.clockwise();
         if (combatantJanlenoux.PosX > 100)
-          data.sanctityWardDir = output.counterclock!();
+          data.sanctityWardDir = output.counterclock();
       },
       infoText: (data, _matches, output) => {
-        return data.sanctityWardDir ?? output.unknown!();
+        return data.sanctityWardDir ?? output.unknown();
       },
       run: (data) => delete data.sanctityWardDir,
       outputStrings: {
@@ -885,9 +787,9 @@ const triggerSet: TriggerSet<Data> = {
       alarmText: (data, matches, output) => {
         const id = getHeadmarkerId(data, matches);
         if (id === headmarkers.sword1)
-          return output.sword1!();
+          return output.sword1();
         if (id === headmarkers.sword2)
-          return output.sword2!();
+          return output.sword2();
       },
       outputStrings: {
         sword1: {
@@ -920,13 +822,11 @@ const triggerSet: TriggerSet<Data> = {
           data.sanctitySword2 = matches.target;
         else
           return;
-
         if (data.sanctitySword1 === undefined || data.sanctitySword2 === undefined)
           return;
-
         const name1 = data.ShortName(data.sanctitySword1);
         const name2 = data.ShortName(data.sanctitySword2);
-        return output.text!({ name1: name1, name2: name2 });
+        return output.text({ name1: name1, name2: name2 });
       },
       // Don't collide with the more important 1/2 call.
       tts: '',
@@ -962,21 +862,19 @@ const triggerSet: TriggerSet<Data> = {
         const [p1, p2] = data.thordanMeteorMarkers.sort();
         if (data.thordanMeteorMarkers.length !== 2 || p1 === undefined || p2 === undefined)
           return;
-
         const p1dps = data.party.isDPS(p1);
         const p2dps = data.party.isDPS(p2);
-
         if (p1dps && p2dps)
-          return output.dpsMeteors!({ player1: data.ShortName(p1), player2: data.ShortName(p2) });
+          return output.dpsMeteors({ player1: data.ShortName(p1), player2: data.ShortName(p2) });
         if (!p1dps && !p2dps)
-          return output.tankHealerMeteors!({ player1: data.ShortName(p1), player2: data.ShortName(p2) });
-        return output.unknownMeteors!({ player1: data.ShortName(p1), player2: data.ShortName(p2) });
+          return output.tankHealerMeteors({ player1: data.ShortName(p1), player2: data.ShortName(p2) });
+        return output.unknownMeteors({ player1: data.ShortName(p1), player2: data.ShortName(p2) });
       },
       outputStrings: {
         tankHealerMeteors: {
           en: 'Tank/Healer Meteors (${player1}, ${player2})',
           de: 'Tank/Heiler Meteore (${player1}, ${player2})',
-          fr: 'Météores Tank/Healer (${player1}, ${player2})', // FIXME
+          fr: 'Météores Tank/Healer (${player1}, ${player2})',
           ja: 'タンヒラ 隕石 (${player1}, ${player2})',
           cn: 'T/奶 陨石 (${player1}, ${player2})',
           ko: '탱/힐 메테오 (${player1}, ${player2})',
@@ -984,7 +882,7 @@ const triggerSet: TriggerSet<Data> = {
         dpsMeteors: {
           en: 'DPS Meteors (${player1}, ${player2})',
           de: 'DDs Meteore (${player1}, ${player2})',
-          fr: 'Météores DPS (${player1}, ${player2})', // FIXME
+          fr: 'Météores DPS (${player1}, ${player2})',
           ja: 'DPS 隕石 (${player1}, ${player2})',
           cn: 'DPS 陨石 (${player1}, ${player2})',
           ko: '딜러 메테오 (${player1}, ${player2})',
@@ -1006,7 +904,7 @@ const triggerSet: TriggerSet<Data> = {
       alertText: (data, matches, output) => {
         const id = getHeadmarkerId(data, matches);
         if (id === headmarkers.meteor)
-          return output.meteorOnYou!();
+          return output.meteorOnYou();
       },
       outputStrings: {
         meteorOnYou: Outputs.meteorOnYou,
@@ -1016,7 +914,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'DSR Broad Swing Right',
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '63C0', source: 'King Thordan', capture: false }),
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Behind => Right',
@@ -1031,7 +929,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'DSR Broad Swing Left',
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '63C1', source: 'King Thordan', capture: false }),
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Behind => Left',
@@ -1052,15 +950,15 @@ const triggerSet: TriggerSet<Data> = {
         if (id === headmarkers.dot1) {
           data.diveFromGraceNum[matches.target] = 1;
           if (matches.target === data.me)
-            return output.num1!();
+            return output.num1();
         } else if (id === headmarkers.dot2) {
           data.diveFromGraceNum[matches.target] = 2;
           if (matches.target === data.me)
-            return output.stackNorthNum!({ num: output.num2!() });
+            return output.stackNorthNum({ num: output.num2() });
         } else if (id === headmarkers.dot3) {
           data.diveFromGraceNum[matches.target] = 3;
           if (matches.target === data.me)
-            return output.stackNorthNum!({ num: output.num3!() });
+            return output.stackNorthNum({ num: output.num3() });
         }
       },
       outputStrings: {
@@ -1124,15 +1022,13 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`DFGYou: missing number: ${JSON.stringify(data.diveFromGraceNum)}`);
           return;
         }
-
         if (data.diveFromGraceDir[data.me] === 'up')
-          return output.upArrow!({ num: num });
+          return output.upArrow({ num: num });
         else if (data.diveFromGraceDir[data.me] === 'down')
-          return output.downArrow!({ num: num });
-
+          return output.downArrow({ num: num });
         if (data.diveFromGraceHasArrow[num])
-          return output.circleWithArrows!({ num: num });
-        return output.circleAllCircles!({ num: num });
+          return output.circleWithArrows({ num: num });
+        return output.circleAllCircles({ num: num });
       },
       outputStrings: {
         circleAllCircles: {
@@ -1174,51 +1070,45 @@ const triggerSet: TriggerSet<Data> = {
       durationSeconds: 5,
       alertText: (data, matches, output) => {
         const key = matches.id === '6712' ? 'out' : 'in';
-        const inout = output[key]!();
+        const inout = output[key]();
         data.diveFromGraceLashGnashKey = key;
-
         const num = data.diveFromGraceNum[data.me];
         if (num !== 1 && num !== 2 && num !== 3) {
           console.error(`DSR Gnash and Lash: missing number: ${JSON.stringify(data.diveFromGraceNum)}`);
           return inout;
         }
-
         // Special case for side ones baiting the two towers.
         if (data.eyeOfTheTyrantCounter === 1 && num === 1 && data.diveFromGracePreviousPosition[data.me] !== 'middle')
-          return output.baitStackInOut!({ inout: inout });
-
+          return output.baitStackInOut({ inout: inout });
         // Filter out anybody who needs to be stacking.
         const firstStack = data.eyeOfTheTyrantCounter === 0 && num !== 1;
         const secondStack = data.eyeOfTheTyrantCounter === 1 && num !== 3;
         if (firstStack || secondStack)
-          return output.stackInOut!({ inout: inout });
-
+          return output.stackInOut({ inout: inout });
         // Call out all circles.
         if (!data.diveFromGraceHasArrow[num]) {
           return {
-            1: output.circlesDive1!({ inout: inout }),
-            2: inout, // this shouldn't happen, but TypeScript wants this key
-            3: output.circlesDive3!({ inout: inout }),
+            1: output.circlesDive1({ inout: inout }),
+            2: inout,
+            3: output.circlesDive3({ inout: inout }),
           }[num];
         }
-
         // Remaining people are arrow dives.
         if (num === 1) {
           if (data.diveFromGraceDir[data.me] === 'circle')
-            return output.southDive1!({ inout: inout });
+            return output.southDive1({ inout: inout });
           if (data.diveFromGraceDir[data.me] === 'up')
-            return output.upArrowDive1!({ inout: inout });
+            return output.upArrowDive1({ inout: inout });
           if (data.diveFromGraceDir[data.me] === 'down')
-            return output.downArrowDive1!({ inout: inout });
+            return output.downArrowDive1({ inout: inout });
         } else if (num === 3) {
           if (data.diveFromGraceDir[data.me] === 'circle')
-            return output.southDive3!({ inout: inout });
+            return output.southDive3({ inout: inout });
           if (data.diveFromGraceDir[data.me] === 'up')
-            return output.upArrowDive3!({ inout: inout });
+            return output.upArrowDive3({ inout: inout });
           if (data.diveFromGraceDir[data.me] === 'down')
-            return output.downArrowDive3!({ inout: inout });
+            return output.downArrowDive3({ inout: inout });
         }
-
         // If things have gone awry, at least output something.
         return inout;
       },
@@ -1345,11 +1235,9 @@ const triggerSet: TriggerSet<Data> = {
             ko: '${inout} => 아래 화살표',
           },
         };
-
         const key = matches.id === '6715' ? 'in' : 'out';
-        const inout = output[key]!();
+        const inout = output[key]();
         data.diveFromGraceLashGnashKey = key;
-
         const num = data.diveFromGraceNum[data.me];
         if (num === undefined) {
           // Don't print a console error here because this response gets eval'd as part
@@ -1357,23 +1245,21 @@ const triggerSet: TriggerSet<Data> = {
           // TODO: maybe have a better way to know if we're in the middle of testing?
           return { infoText: inout };
         }
-
         if (data.eyeOfTheTyrantCounter === 1) {
           if (num === 2) {
             if (!data.diveFromGraceHasArrow[num])
-              return { alertText: output.circlesDive2!({ inout: inout }) };
+              return { alertText: output.circlesDive2({ inout: inout }) };
             if (data.diveFromGraceDir[data.me] === 'up')
-              return { alertText: output.upArrowDive2!({ inout: inout }) };
+              return { alertText: output.upArrowDive2({ inout: inout }) };
             if (data.diveFromGraceDir[data.me] === 'down')
-              return { alertText: output.downArrowDive2!({ inout: inout }) };
+              return { alertText: output.downArrowDive2({ inout: inout }) };
           } else if (num === 3) {
-            return { alertText: output.inOutAndBait!({ inout: inout }) };
+            return { alertText: output.inOutAndBait({ inout: inout }) };
           }
         } else if (data.eyeOfTheTyrantCounter === 2) {
           if (num === 2 || (num === 1 && data.diveFromGracePreviousPosition[data.me] === 'middle'))
-            return { alertText: output.inOutAndBait!({ inout: inout }) };
+            return { alertText: output.inOutAndBait({ inout: inout }) };
         }
-
         // Otherwise, just tell the remaining people the dodge.
         return { infoText: inout };
       },
@@ -1404,22 +1290,20 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`DFG Tower 1 Reminder: missing number: ${JSON.stringify(data.diveFromGraceNum)}`);
           return;
         }
-
-        const inout = output[data.diveFromGraceLashGnashKey]!();
-
+        const inout = output[data.diveFromGraceLashGnashKey]();
         if (data.eyeOfTheTyrantCounter === 1) {
           // Tower 1 soaks only based on debuffs.
           if (num === 3) {
             if (data.diveFromGraceDir[data.me] === 'circle') {
               if (data.diveFromGraceHasArrow[3])
-                return output.southTower1!({ inout: inout });
-              return output.circleTowers1!({ inout: inout });
+                return output.southTower1({ inout: inout });
+              return output.circleTowers1({ inout: inout });
             } else if (data.diveFromGraceDir[data.me] === 'up') {
-              return output.upArrowTower1!({ inout: inout });
+              return output.upArrowTower1({ inout: inout });
             } else if (data.diveFromGraceDir[data.me] === 'down') {
-              return output.downArrowTower1!({ inout: inout });
+              return output.downArrowTower1({ inout: inout });
             }
-            return output.unknownTower!({ inout: inout });
+            return output.unknownTower({ inout: inout });
           }
           // Folks who could be placing a two tower here just get
           // the normal "in" or "out" below, and the Lash Gnash Followup
@@ -1429,21 +1313,20 @@ const triggerSet: TriggerSet<Data> = {
           const pos = data.diveFromGracePreviousPosition[data.me];
           if (num === 1) {
             if (pos === 'middle')
-              return output.southTower3!({ inout: inout });
+              return output.southTower3({ inout: inout });
             // Already soaked the number two towers.
             if (pos === 'east' || pos === 'west')
               return;
           } else if (num === 2) {
             if (pos === 'west')
-              return output.westTower3!({ inout: inout });
+              return output.westTower3({ inout: inout });
             if (pos === 'east')
-              return output.eastTower3!({ inout: inout });
+              return output.eastTower3({ inout: inout });
           } else if (num === 3) {
             return;
           }
-          return output.unknownTower!({ inout: inout });
+          return output.unknownTower({ inout: inout });
         }
-
         // Anybody who isn't placing or taking a tower here just gets a "in/out" reminder.
         return inout;
       },
@@ -1549,7 +1432,6 @@ const triggerSet: TriggerSet<Data> = {
           console.error(`DFG Tower 1 and 2: missing number: ${JSON.stringify(data.diveFromGraceNum)}`);
           return;
         }
-
         // Sorted from west to east, and filled in with unknown if missing.
         const [nameA, nameB, nameC] = [
           ...Object.keys(data.diveFromGracePositions).sort((keyA, keyB) => {
@@ -1559,11 +1441,10 @@ const triggerSet: TriggerSet<Data> = {
               return 0;
             return posA - posB;
           }),
-          output.unknown!(),
-          output.unknown!(),
-          output.unknown!(),
+          output.unknown(),
+          output.unknown(),
+          output.unknown(),
         ];
-
         // Dive 1 and Dive 3 have 3 players
         if (data.diveFromGraceTowerCounter !== 2) {
           data.diveFromGracePreviousPosition[nameA] = 'west';
@@ -1573,15 +1454,14 @@ const triggerSet: TriggerSet<Data> = {
           data.diveFromGracePreviousPosition[nameA] = 'west';
           data.diveFromGracePreviousPosition[nameB] = 'east';
         }
-
         if (num === 1 && data.diveFromGraceTowerCounter === 2) {
           if (data.diveFromGracePreviousPosition[data.me] === 'west')
-            return output.northwestTower2!();
+            return output.northwestTower2();
           if (data.diveFromGracePreviousPosition[data.me] === 'east')
-            return output.northeastTower2!();
+            return output.northeastTower2();
           if (data.diveFromGracePreviousPosition[data.me] === 'middle')
             return;
-          return output.unknownTower!();
+          return output.unknownTower();
         }
       },
       run: (data) => data.diveFromGracePositions = {},
@@ -1620,14 +1500,14 @@ const triggerSet: TriggerSet<Data> = {
         const num = data.diveFromGraceNum[data.me];
         if (!num) {
           console.error(`DFG Dive Single Tower: missing number: ${JSON.stringify(data.diveFromGraceNum)}`);
-          return output.text!();
+          return output.text();
         }
         // To condense messages, two tower baiters get this call during the gnash and lash.
         if (data.diveFromGraceTowerCounter === 2) {
           data.stackAfterGeirskogul = true;
           return;
         }
-        return output.text!();
+        return output.text();
       },
       run: (data) => data.waitingForGeirskogul = true,
       outputStrings: {
@@ -1651,10 +1531,10 @@ const triggerSet: TriggerSet<Data> = {
           return;
         // Two tower baiters need to quickly get to the stack here.
         if (data.stackAfterGeirskogul) {
-          const inout = output[data.diveFromGraceLashGnashKey]!();
-          return output.stackInOut!({ inout: inout });
+          const inout = output[data.diveFromGraceLashGnashKey]();
+          return output.stackInOut({ inout: inout });
         }
-        return output.move!();
+        return output.move();
       },
       run: (data) => {
         delete data.waitingForGeirskogul;
@@ -1690,7 +1570,7 @@ const triggerSet: TriggerSet<Data> = {
       // Since dives are usually blue people dropping off their blue tether
       // to a red person (who needs to run in), make the blue tether
       // the higher severity one.
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Blue',
@@ -1707,7 +1587,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.tether({ id: '0034' }),
       condition: (data, matches) => matches.source === data.me,
       // See note above on Right Eye Blue Tether.
-      infoText: (_data, _matches, output) => output.text!(),
+      infoText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Red',
@@ -1725,7 +1605,7 @@ const triggerSet: TriggerSet<Data> = {
       // One cast for each dive.  68C3 is the initial cast/self-targeted ability.
       // 68C4 is the damage on players.
       suppressSeconds: 1,
-      infoText: (_data, _matches, output) => output.text!(),
+      infoText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Dives Soon',
@@ -1744,7 +1624,7 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: NetRegexes.ability({ id: '68C4', source: 'Nidhogg', capture: false }),
       // One ability for each player hit.
       suppressSeconds: 1,
-      infoText: (data, _matches, output) => output[`dive${data.diveCounter}`]!(),
+      infoText: (data, _matches, output) => output[`dive${data.diveCounter}`](),
       run: (data) => data.diveCounter++,
       outputStrings: {
         dive1: Outputs.num1,
@@ -1768,7 +1648,7 @@ const triggerSet: TriggerSet<Data> = {
       // will be revived and you shouldn't forget about it.
       netRegex: NetRegexes.startsUsing({ id: '68BD', source: 'Left Eye' }),
       delaySeconds: (_data, matches) => parseFloat(matches.castTime),
-      infoText: (_data, _matches, output) => output.text!(),
+      infoText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Kill Right Eye',
@@ -1789,7 +1669,7 @@ const triggerSet: TriggerSet<Data> = {
       // hit tank LB when the cast bar gets to the "F" in "Fury", which is
       // roughly 2.8 seconds before it ends.
       delaySeconds: 10 - 2.8,
-      alarmText: (_data, _matches, output) => output.text!(),
+      alarmText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'TANK LB!!',
@@ -1806,7 +1686,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'Ability',
       netRegex: NetRegexes.ability({ id: '6B8B', source: 'Vedrfolnir', capture: false }),
       suppressSeconds: 1,
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Twisters',
@@ -1823,7 +1703,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'Tether',
       netRegex: NetRegexes.tether({ id: '0005' }),
       condition: Conditions.targetIsYou(),
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Tether on YOU',
@@ -1843,7 +1723,7 @@ const triggerSet: TriggerSet<Data> = {
         const id = getHeadmarkerId(data, matches);
         // The Wrath of the Heavens skyward leap is a different headmarker.
         if (id === headmarkers.skywardSingle)
-          return output.leapOnYou!();
+          return output.leapOnYou();
       },
       outputStrings: {
         leapOnYou: {
@@ -1867,7 +1747,7 @@ const triggerSet: TriggerSet<Data> = {
       delaySeconds: 2.5,
       alertText: (data, matches, output) => {
         if (data.me === matches.target)
-          return output.text!();
+          return output.text();
       },
       run: (data, matches) => data.thunderstruck.push(matches.target),
       outputStrings: {
@@ -1891,9 +1771,9 @@ const triggerSet: TriggerSet<Data> = {
       infoText: (data, _matches, output) => {
         // In case somebody wants to do some "go in the order cactbot tells you" sort of strat.
         const [fullName1, fullName2] = data.thunderstruck.sort();
-        const name1 = fullName1 ? data.ShortName(fullName1) : output.unknown!();
-        const name2 = fullName2 ? data.ShortName(fullName2) : output.unknown!();
-        return output.text!({ name1: name1, name2: name2 });
+        const name1 = fullName1 ? data.ShortName(fullName1) : output.unknown();
+        const name2 = fullName2 ? data.ShortName(fullName2) : output.unknown();
+        return output.text({ name1: name1, name2: name2 });
       },
       // Sorry tts players, but "Thunder on YOU" and "Thunder: names" are too similar.
       tts: null,
@@ -1916,7 +1796,7 @@ const triggerSet: TriggerSet<Data> = {
       alarmText: (data, matches, output) => {
         const id = getHeadmarkerId(data, matches);
         if (id === headmarkers.cauterize)
-          return output.diveOnYou!();
+          return output.diveOnYou();
       },
       outputStrings: {
         diveOnYou: {
@@ -1935,13 +1815,13 @@ const triggerSet: TriggerSet<Data> = {
       preRun: (data, matches) => data.hasDoom[matches.target] = true,
       alertText: (data, matches, output) => {
         if (data.me === matches.target)
-          return output.doomOnYou!();
+          return output.doomOnYou();
       },
       infoText: (data, _matches, output) => {
         const dooms = Object.keys(data.hasDoom).filter((x) => data.hasDoom[x]);
         if (dooms.length !== 4 || dooms.includes(data.me))
           return;
-        return output.noDoom!();
+        return output.noDoom();
       },
       outputStrings: {
         doomOnYou: {
@@ -1970,35 +1850,31 @@ const triggerSet: TriggerSet<Data> = {
         const marker = playstationMarkerMap[id];
         if (marker === undefined)
           return;
-
         data.deathMarker[matches.target] = marker;
-
         if (data.me !== matches.target)
           return;
-
         // Note: in general, both circles should always have Doom and both crosses
         // should not have Doom.  If one doom dies, it seems that crosses are
         // removed and there's a double triangle non-doom.  If enough people die,
         // anything can happen.  For example, in P1 tanks can get circles if enough people are dead.
-
         if (data.hasDoom[data.me]) {
           if (marker === 'circle')
-            return output.circleWithDoom!();
+            return output.circleWithDoom();
           else if (marker === 'triangle')
-            return output.triangleWithDoom!();
+            return output.triangleWithDoom();
           else if (marker === 'square')
-            return output.squareWithDoom!();
+            return output.squareWithDoom();
           else if (marker === 'cross')
-            return output.crossWithDoom!();
+            return output.crossWithDoom();
         } else {
           if (marker === 'circle')
-            return output.circle!();
+            return output.circle();
           else if (marker === 'triangle')
-            return output.triangle!();
+            return output.triangle();
           else if (marker === 'square')
-            return output.square!();
+            return output.square();
           else if (marker === 'cross')
-            return output.cross!();
+            return output.cross();
         }
       },
       outputStrings: {
@@ -2079,34 +1955,31 @@ const triggerSet: TriggerSet<Data> = {
       alarmText: (data, _matches, output) => {
         if (data.deathMarker[data.me] !== undefined)
           return;
-
         const seenMarkers = Object.values(data.deathMarker);
         const markers = [...playstationMarkers].filter((x) => !seenMarkers.includes(x));
-
         const [marker] = markers;
         if (marker === undefined || markers.length !== 1)
           return;
-
         // Note: this will still call out for the dead doom person, but it seems better
         // in case they somehow got insta-raised.
         if (data.hasDoom[data.me]) {
           if (marker === 'circle')
-            return output.circleWithDoom!();
+            return output.circleWithDoom();
           else if (marker === 'triangle')
-            return output.triangleWithDoom!();
+            return output.triangleWithDoom();
           else if (marker === 'square')
-            return output.squareWithDoom!();
+            return output.squareWithDoom();
           else if (marker === 'cross')
-            return output.crossWithDoom!();
+            return output.crossWithDoom();
         } else {
           if (marker === 'circle')
-            return output.circle!();
+            return output.circle();
           else if (marker === 'triangle')
-            return output.triangle!();
+            return output.triangle();
           else if (marker === 'square')
-            return output.square!();
+            return output.square();
           else if (marker === 'cross')
-            return output.cross!();
+            return output.cross();
         }
       },
       outputStrings: {
@@ -2187,9 +2060,8 @@ const triggerSet: TriggerSet<Data> = {
         const myMarker = playstationMarkerMap[id];
         if (myMarker === undefined)
           return;
-
         // Find person with the same mark.
-        let partner: string | undefined = undefined;
+        let partner = undefined;
         for (const [player, marker] of Object.entries(data.deathMarker)) {
           if (player !== data.me && marker === myMarker) {
             partner = player;
@@ -2198,23 +2070,20 @@ const triggerSet: TriggerSet<Data> = {
         }
         if (partner === undefined)
           return;
-
         // If a circle ends up with a non-doom or a cross ends up with a doom,
         // I think you're in serious unrecoverable trouble.  These people also
         // already need to look and adjust to the other person, vs the triangle
         // and square which can have a fixed position based on doom vs non-doom.
         if (myMarker === 'circle' || myMarker === 'cross')
           return;
-
         // I think circles fill out with doom first, so it should be impossible
         // to have two doom triangles or two doom squares as well.
         if (data.hasDoom[data.me] || data.hasDoom[partner])
           return;
-
         if (myMarker === 'triangle')
-          return output.doubleTriangle!({ player: data.ShortName(partner) });
+          return output.doubleTriangle({ player: data.ShortName(partner) });
         if (myMarker === 'square')
-          return output.doubleSquare!({ player: data.ShortName(partner) });
+          return output.doubleSquare({ player: data.ShortName(partner) });
       },
       outputStrings: {
         // In case users want to have triangle vs square say something different.
@@ -2256,10 +2125,9 @@ const triggerSet: TriggerSet<Data> = {
             ko: '흐레스벨그 탱버',
           },
         };
-
         if (data.role === 'tank')
-          return { alertText: output.tanksApart!() };
-        return { infoText: output.hraesvelgrTankbuster!() };
+          return { alertText: output.tanksApart() };
+        return { infoText: output.hraesvelgrTankbuster() };
       },
     },
     {
@@ -2291,10 +2159,9 @@ const triggerSet: TriggerSet<Data> = {
             ko: '니드호그 탱버',
           },
         };
-
         if (data.role === 'tank')
-          return { alertText: output.tanksApart!() };
-        return { infoText: output.nidTankbuster!() };
+          return { alertText: output.tanksApart() };
+        return { infoText: output.nidTankbuster() };
       },
     },
     {
@@ -2327,12 +2194,11 @@ const triggerSet: TriggerSet<Data> = {
             ko: '쉐어 탱버',
           },
         };
-
         if (!data.hraesvelgrGlowing || !data.nidhoggGlowing)
           return;
         if (data.role === 'tank')
-          return { alertText: output.sharedBuster!() };
-        return { infoText: output.sharedBuster!() };
+          return { alertText: output.sharedBuster() };
+        return { infoText: output.sharedBuster() };
       },
       run: (data) => {
         delete data.hraesvelgrGlowing;
@@ -2354,7 +2220,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: ['6D41', '6D43'], source: ['Hraesvelgr', 'Nidhogg'], capture: false }),
       suppressSeconds: 2,
-      alertText: (_data, _matches, output) => output.groups!(),
+      alertText: (_data, _matches, output) => output.groups(),
       outputStrings: {
         groups: {
           en: 'Healer Groups',
@@ -2388,44 +2254,38 @@ const triggerSet: TriggerSet<Data> = {
       durationSeconds: 6,
       promise: async (data) => {
         data.combatantData = [];
-
         // TODO: implement Hot Tail/Hot Wing combination here
         if (data.hallowedWingsCount !== 1)
           return;
-
         // If we have missed the Nidhogg id (somehow?), we'll handle it later.
         const id = data.addsPhaseNidhoggId;
         if (id === undefined)
           return;
-
         data.combatantData = (await callOverlayHandler({
           call: 'getCombatants',
           ids: [parseInt(id, 16)],
         })).combatants;
-
         if (data.combatantData.length === 0)
           console.error(`Hallowed: no Nidhoggs found`);
         else if (data.combatantData.length > 1)
           console.error(`Hallowed: unexpected number of Nidhoggs: ${JSON.stringify(data.combatantData)}`);
       },
       alertText: (data, matches, output) => {
-        const wings = (matches.id === '6D23' || matches.id === '6D24') ? output.left!() : output.right!();
+        const wings = (matches.id === '6D23' || matches.id === '6D24') ? output.left() : output.right();
         let head;
         const isHeadDown = matches.id === '6D23' || matches.id === '6D26';
         if (isHeadDown)
-          head = data.role === 'tank' ? output.tanksNear!() : output.partyFar!();
+          head = data.role === 'tank' ? output.tanksNear() : output.partyFar();
         else
-          head = data.role === 'tank' ? output.tanksFar!() : output.partyNear!();
-
+          head = data.role === 'tank' ? output.tanksFar() : output.partyNear();
         const [nidhogg] = data.combatantData;
         if (nidhogg !== undefined && data.combatantData.length === 1) {
           // Nidhogg is at x = 100 +/- 11, y = 100 +/- 34
-          const dive = nidhogg.PosX < 100 ? output.forward!() : output.backward!();
-          return output.wingsDiveHead!({ wings: wings, dive: dive, head: head });
+          const dive = nidhogg.PosX < 100 ? output.forward() : output.backward();
+          return output.wingsDiveHead({ wings: wings, dive: dive, head: head });
         }
-
         // If something has gone awry (or this is the second hallowed), call out what we can.
-        return output.wingsHead!({ wings: wings, head: head });
+        return output.wingsHead({ wings: wings, head: head });
       },
       outputStrings: {
         // The calls here assume that all players are looking at Hraesvelgr, and thus
@@ -2503,7 +2363,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'DSR Nidhogg Hot Wing',
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '6D2B', source: 'Nidhogg', capture: false }),
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           // Often cactbot uses "in" and "out", but that's usually hitbox relative vs
@@ -2520,7 +2380,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'DSR Nidhogg Hot Tail',
       type: 'StartsUsing',
       netRegex: NetRegexes.startsUsing({ id: '6D2D', source: 'Nidhogg', capture: false }),
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Outside',
@@ -2544,8 +2404,8 @@ const triggerSet: TriggerSet<Data> = {
       delaySeconds: (_data, matches) => parseFloat(matches.duration) - 6,
       infoText: (_data, matches, output) => {
         if (matches.effectId === 'B52')
-          return output.hraesvelgr!();
-        return output.nidhogg!();
+          return output.hraesvelgr();
+        return output.nidhogg();
       },
       outputStrings: {
         nidhogg: {
@@ -2576,7 +2436,7 @@ const triggerSet: TriggerSet<Data> = {
       delaySeconds: (_data, matches) => parseFloat(matches.duration) - 1,
       // Player will have Pyretic for about 3s before hit by Cauterize
       durationSeconds: 4,
-      infoText: (_data, _matches, output) => output.text!(),
+      infoText: (_data, _matches, output) => output.text(),
       outputStrings: {
         text: {
           en: 'Stop',
@@ -2595,7 +2455,6 @@ const triggerSet: TriggerSet<Data> = {
       preRun: (data, matches) => {
         if (matches.effectId === 'AC6')
           data.spreadingFlame.push(matches.target);
-
         if (matches.effectId === 'AC7')
           data.entangledFlame.push(matches.target);
       },
@@ -2604,12 +2463,11 @@ const triggerSet: TriggerSet<Data> = {
           return;
         if (data.entangledFlame.length < 2)
           return;
-
         if (data.spreadingFlame.includes(data.me))
-          return output.spread!();
+          return output.spread();
         if (data.entangledFlame.includes(data.me))
-          return output.stack!();
-        return output.nodebuff!();
+          return output.stack();
+        return output.nodebuff();
       },
       outputStrings: {
         spread: {
@@ -2658,7 +2516,7 @@ const triggerSet: TriggerSet<Data> = {
       condition: (data, matches) => data.me === matches.target && data.role === 'tank',
       alertText: (_data, matches, output) => {
         if (parseFloat(matches.duration) > 10)
-          return output.text!();
+          return output.text();
       },
       outputStrings: {
         text: {
@@ -2684,7 +2542,7 @@ const triggerSet: TriggerSet<Data> = {
       delaySeconds: 2.5,
       alertText: (_data, matches, output) => {
         if (parseFloat(matches.duration) > 10)
-          return output.text!();
+          return output.text();
       },
       outputStrings: {
         text: {
@@ -3366,6 +3224,4 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
   ],
-};
-
-export default triggerSet;
+});
```

trim21: And I tested with/without eslint, doesn't make any difference

package.json
trim21: In this case I suggest we drop 300 line width in triggers files as we no longer need multiple line for netRegex/netRegexCn/netRegexKo and so on and in the future will replace netRegex lone line with it's param.

quisquous: Yeah, I was having a similar thought.  Let me look through scripts to make sure nothing is trying to parse trigger files as text and will break if things go on different lines.

trim21: I tried `yarn util findTranslations` and it works fine, don't know if any other script may break.

trim21: I know `.github/scripts/auto-label.cjs` may break if we add parentheses to these value that got splited in new line...

util/process_triggers_folder.ts
quisquous: Ok, if it makes no difference with eslint, we can remove it here.

One question though is: why are there so many empty lines removed?

trim21: I guess tsc remove them

package.json
quisquous: Yeah, I looked through everything, and it seems like everything is doing a dynamic `import` of trigger files these days and not parsing as text (phew).  So, I think this should be ok.

Can you explain your concern about `auto-label`? It seems like it should be ok to me.

trim21: And we can write a formatter by our own but I think it's too much... https://github.com/dprint/js-formatter

trim21: > 

we are actually using json5 to parse the ast result into js object to avoid executing any code, adding parentheses breaks that.

quisquous: Why are there parentheses? I don't understand.

trim21:
```ts
    {
      'locale': 'fr',
      'replaceSync': {
        'This humanoid prototype can perfectly replicate': (
          'Ce modèle exceptionnel est basé sur les non moins exceptionnelles divinités guerrières'
        ),
        'Utilizing our data on Sophia': (
          'Quoi de mieux pour dominer les divinités guerrières que de retourner leur puissance contre elles?♪',
         )
      },
```

if someone do line break like this (

quisquous: Oh, I see.  But this could already be an issue today, right?

What about an extra test that the section is valid json5 so that it's obvious when this happens?

quisquous: Also, if you don't mind, could you update the line length to 100 in a separate PR, land that first, and then land this patch? I feel like changing the line length will be very large, and it will make this PR very unwieldy to look at further changes.

trim21: Yes, but I don't think that's very necessary, let's do that when this actually happen ...

trim21:
we can set line-width to 120, only need to format 36 triggers, and could be included in this pr

quisquous:
> we can set line-width to 120, only need to format 36 triggers, and could be included in this pr, another option.

Doesn't this reformat a lot of the other code that's not just in triggers? One of the reasons that I picked 300 (and not 10000) was that eslint with the dprint plugin would try to fit everything onto one line if it could.

trim21:
> > we can set line-width to 120, only need to format 36 triggers, and could be included in this pr, another option.
> 
> Doesn't this reformat a lot of the other code that's not just in triggers? One of the reasons that I picked 300 (and not 10000) was that eslint with the dprint plugin would try to fit everything onto one line if it could.

not very much

trim21:
maybe after https://github.com/quisquous/cactbot/pull/4970 is merged

trim21:
may considring add eslint rule [no-extra-parens](https://eslint.org/docs/latest/rules/no-extra-parens) if new version dprint will remove them.

trim21:
I also create a branch with 120 line width https://github.com/trim21/cactbot/tree/style-width-120

trim21:
This pr only format 40+ files including the triggers.

I think it's not too much change to merge now.

trim21:
> > we can set line-width to 120, only need to format 36 triggers, and could be included in this pr, another option.
> 
> Doesn't this reformat a lot of the other code that's not just in triggers? One of the reasons that I picked 300 (and not 10000) was that eslint with the dprint plugin would try to fit everything onto one line if it could.

I may misunderstand this comment. It doesn't need to change many trigger file with these width now

trim21:
any thoughts?

dprint.json
```diff
@@ -0,0 +1,20 @@
+{
+  "$schema": "https://dprint.dev/schemas/v0.json",
+  "incremental": true,
+  "includes": [
+    "**/*.{ts,cjs,mjs,js,json}"
+  ],
+  "excludes": [
+    "**/node_modules",
+    "./dist/",
+    "./npm-package/",
```
quisquous, line 10:
```suggestion
    "./npm-package/",
    "./plugin/ThirdParty/",
    "./publish/",
```

quisquous (commented):
I think this seems good overall.

I'd be fine with line length 100 for all files, but I'd like that to be a separate PR.

Overall my biggest concern with having both dprint and eslint is that I don't want them to fight.  It seems possible that you could write some code A that dprint would change to B and then eslint would change back to A.  In that case, the code is good but it would not pass the lint check.  (This could not happen before when dprint was a plugin, because dprint+eslint was one step.)  It doesn't seem like this is the case anywhere today (running dprint fmt without eslint makes no changes to the codebase), but it seems possible that it could happen.  It might be nice to move more formatting things into dprint if possible and have eslint be more "lint" and less "formatting".

I am still concerned with VSCode support.  Have you looked at that at all?

Should the eslint no extra parens rule be removed now as well?

trim21:
I have changed eslint config to disable all conflicting rules.

And I think vacode dprint will work fine with single dprint config.

And what do you mean by separated PR?

quisquous:
> I have changed eslint config to disable all conflicting rules, we can remove it now if you don't like it

Either is ok.  You could remove it here if it's not doing anything or you could remove it later if there's other rules that we should also think about removing that dprint is already doing.

> And I think vacode dprint will work fine with single dprint config.

Have you tried it or is this something I need to go look into myself?

> And what do you mean by separated PR?

A separate PR to change the line length for the trigger files / combine the dprint config files.

trim21:
> Have you tried it or is this something I need to go look into myself?

yes, but only single dprint config, otherwise trigger files not get formatted well.

trim21:
I just add config for vscode

trim21:
It's ok now, I think.

trim21:
> It might be nice to move more formatting things into dprint if possible and have eslint be more "lint" and less "formatting".

There are still several formatting rules in eslint but working well with dprint so I didn't edit them.

PR #4955, opened 2022-10-19
build: merge tsconfig files

trim21:
no need to keep a separated tsconfig for webpack config file, merge it into `tsconfig.json`


quisquous:
Thanks, this is great cleanup.

PR #4961, opened 2022-10-20
resources: common net regex

trim21:
This should work perfectly fine


quisquous:
Thanks for the fixup.

PR #4964, opened 2022-10-21
i18n: update ko translations

Bing-su:
_rsv data remain in English


Akurosia:
also if you are able to enter DSU on ko client, enable the debug option in FFXIV plugin with log all lines before entering. Afterwards i could decrypt the _rsv_ names

Bing-su:
fix that

PR #4967, opened 2022-10-24
raidboss: dungeons Add Another Sil'dihn Subterrane

Legends0:
This is some basic stuff I've put together so far.


xiashtra:
Since ASS and ASSS have the same mechanics, it is possible to reuse all triggers for both zones by adding the savage ability IDs.  That may cause complications with the timeline files however.  @quisquous thoughts?

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,482 @@
...
+  timelineFile: 'another_sildihn_subterrane.txt',
+  initData: () => {
+    return {
+      soapCounter: 0,
+      beaterCounter: 0,
+      mightCasts: [],
+    };
+  },
+  triggers: [
+    // ---------------- Silkie ----------------
```
xiashtra, line 40:
```suggestion
    // ---------------- first trash ----------------
    {
      id: 'ASS Atropine Spore',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7960', '7978'], source: 'Aqueduct Belladonna', capture: false }),
      response: Responses.getIn(),
    },
    {
      id: 'ASS Frond Affront',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7961', '7979'], source: 'Aqueduct Belladonna', capture: false }),
      response: Responses.lookAway(),
    },
    {
      id: 'ASS Deracinator',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7962', '797A'], source: 'Aqueduct Belladonna', capture: false }),
      response: Responses.tankBuster(),
    },
    {
      id: 'ASS Left Sweep',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7964', '797C'], source: 'Aqueduct Kaluk', capture: false }),
      response: Responses.goRight(),
    },
    {
      id: 'ASS Right Sweep',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7963', '797B'], source: 'Aqueduct Kaluk', capture: false }),
      response: Responses.goLeft(),
    },
    {
      id: 'ASS Creeping Ivy',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7965', '797D'], source: 'Aqueduct Kaluk', capture: false }),
      response: Responses.getBehind(),
    },
    {
      id: 'ASS Honeyed Left',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['795B', '7973'], source: 'Aqueduct Udumbara', capture: false }),
      response: Responses.goRight(),
    },
    {
      id: 'ASS Honeyed Right',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['795C', '7974'], source: 'Aqueduct Udumbara', capture: false }),
      response: Responses.goLeft(),
    },
    {
      id: 'ASS Honeyed Front',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['795D', '7975'], source: 'Aqueduct Udumbara', capture: false }),
      response: Responses.getBehind(),
    },
    {
      id: 'ASS Arboreal Storm',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7957', '796F'], source: 'Aqueduct Dryad', capture: false }),
      response: Responses.getOut(),
    },
    // ---------------- Silkie ----------------
```

xiashtra: NB: this suggestion includes ASS and ASS Savage IDs

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,482 @@
...
+        text: {
+          en: 'aoe + bleed',
+          de: 'AoE + Blutung',
+          fr: 'AoE + Saignement',
+          ja: 'AOE + 出血',
+          ko: '전체 공격 + 도트',
+        },
+      },
+    },
+    // ---------------- Gladiator of Sil'dih ----------------
```
xiashtra, line 190:
```suggestion
    // ---------------- second trash  ----------------
    {
      id: 'ASS Infernal Pain',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7969', '7981'], source: 'Sil\'dihn Dullahan', capture: false }),
      infoText: (_data, _matches, output) => output.text!(),
      outputStrings: {
        text: {
          en: 'aoe + bleed',
          de: 'AoE + Blutung',
          fr: 'AoE + Saignement',
          ja: 'AOE + 出血',
          ko: '전체 공격 + 도트',
        },
      },
    },
    {
      id: 'ASS Blighted Gloom',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7966', '797E'], source: 'Sil\'dihn Dullahan', capture: false }),
      response: Responses.getOut(),
    },
    {
      id: 'ASS King\'s Will',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['7968', '7980'], source: 'Sil\'dihn Dullahan', capture: false }),
      infoText: (_data, _matches, output) => output.text!(),
      outputStrings: {
        text: {
          en: 'big autos',
        },
      },
    },
    {
      id: 'ASS Hells\' Nebula',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['796C', '7984'], source: 'Aqueduct Armor', capture: false }),
      infoText: (_data, _matches, output) => output.text!(),
      outputStrings: {
        text: {
          en: 'HP to 1',
        },
      },
    {
      id: 'ASS Infernal Weight',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['796B', '7983'], source: 'Aqueduct Armor', capture: false }),
      infoText: (_data, _matches, output) => output.text!(),
      outputStrings: {
        text: {
          en: 'aoe + heavy',
        },
      },
    },
    {
      id: 'ASS Dominion Slash',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['796A', '7982'], source: 'Aqueduct Armor', capture: false }),
      response: Responses.getBehind(),
    },
    // ---------------- Gladiator of Sil'dih ----------------
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,482 @@
...
+  initData: () => {
+    return {
+      soapCounter: 0,
+      beaterCounter: 0,
+      mightCasts: [],
+    };
+  },
+  triggers: [
+    // ---------------- Silkie ----------------
+    {
```
xiashtra, line 41:
```suggestion
    {
      id: 'ASildihn Soap\'s Up',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: ['775A', '777D'], source: 'Silkie', capture: false }),
      infoText: (_data, _matches, output) => output.text!(),
      outputStrings: {
        text: {
          en: 'cardinals',
        },
      },
    },
    {
```

xiashtra: NB: this suggestion includes ASS and ASS Savage IDs

Akurosia:
I would not put savage and normal into one file as we usualy seperate them (see ex primals and theire unreal counter parts).

Legends0:
I think the one issue with combining Savage and normal here is the zone id portion only accepts one value? I could be wrong... otherwise I don't particularly see why this couldn't be written to handle both.

Unlike with other instances, the Savage version does not change mechanics of the spells, just the id changes. The lingering issue would be that some of the trigger design relies on a set number of ids for capturing, the triggers would need to be updated to account for multiple options. For instance if a trigger is filtering for a left/right mechanics but in the trigger itself checks only one of the values. A similar issue can happen with the timeline. This also makes it more difficult for translation as far as I am aware since the localized often use the spell ids to translate lines?

xiashtra:
If Unreal and EX are being kept separate then ASS and ASSS should probably be kept separate for consistency, though it does lead to a lot of code duplication and a possibility of desync on changes.

Unreal is kind of a unique case though in that it has a short shelf-life; ASS and ASSS should be a permanent fixture, and this duplication will come up again with future Criterion dungeons.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,482 @@
...
+      },
+    },
+    // ---------------- Shadowcaster Zeless Gah ----------------
+    {
+      id: 'ASS Show of Strength',
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '74AF', source: 'Shadowcaster Zeless Gah', capture: false }),
+      response: Responses.aoe(),
+    },
+  ],
```
xiashtra, line 478:
```suggestion
{
      id: 'ASS Firesteel Fracture',
      type: 'StartsUsing',
      netRegex: NetRegexes.startsUsing({ id: '74AD', source: 'Shadowcaster Zeless Gah', capture: false }),
      response: Responses.tankCleave(),
    },
  ],
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: I don't know if players are doing this, but does cactbot ever handle multiple mobs pulled at once that could cause multiple of these triggers to fire? Some of it I guess is just an if you know the encounter then you know which entities the call is for.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: I think there are similar triggers for handling a mechanic such ad King's Will and Hell's Nebula. I'm not sure if a mechanic such as Infernal's Weight has existed. Infernal's Weight I think targets a random player as well, so might that be accounted for where the followup Cleave is going.

xiashtra: From my logs Infernal Weight targets everyone (no specific target), and the follow-up Dominion Slash has no target, so I don't know if there's a possibility of calling where it's going without looking at mob facing.

Legends0: Ah, well we can adjust the domion slash later if necessary.
I think the Infernal Weight will be better with a aoe + getin call instead so that the followup Slash can be safely avoided.

I would remove the Hell's Nebula trigger, cactbot historically does not call out for these unless there is a requirement to heal up after (often that is a trigger for the spell that requires health threshold and not the white hole spell itself). I'm not sure an exception should be made here as you could potentially be pulling multiple mobs at once. Examples: E12S Hell's Judgment, O4S White Hole.

I would also remove the King's Will trigger. I believe these spells have been reviewed in previous encounters, but no trigger specific to the buff was made? Effectively this buff does behave similarly to a mini tankbuster (albeit multiple ones for the duration) which we have had an instance of in a dungeon boss before, I think it was a Shadowbringers dungeon. So maybe change to something like that if kept I would change to a gainsEffect trigger and set duration to match the buffs timer.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: The tankbuster response needs to have the capture. Could you remove the capture: false?

xiashtra: Sorry about that, I'm grabbing portions from my js test triggers where it's not required.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
xiashtra: Done.  I may circle back on the King's Will trigger later as a tank/heal only heads-up.

Legends0: Thanks!

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: Thanks for the update.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: There is a fizzling Suds cast before this that determines the soaps up ability, but seeing as soaps up only casts once and it's always fizzling Suds, this trigger works.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,599 @@
...
+          return output.stackOnYou!();
+        return output.stackOn!({ player: data.ShortName(matches.target) });
+      },
+      outputStrings: {
+        spread: Outputs.spread,
+        stackOnYou: Outputs.stackOnYou,
+        stackOn: Outputs.stackOnPlayer,
+      },
+    },
+    // ---------------- Shadowcaster Zeless Gah ----------------
```
xiashtra, line 582:
```suggestion
    {
      id: 'ASS Curse of the Monument',
      type: 'Ability',
      netRegex: NetRegexes.ability({ id: '7666', source: 'Gladiator of Sil\'dih', capture: false }),
      response: Responses.breakChains(),
    },
    // ---------------- Shadowcaster Zeless Gah ----------------
```

Legends0: Thanks!

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
xiashtra: Those were my thoughts as well.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,488 @@
...
+        // Filter for unique
+        const mirage2 = (mirage1.x === unfilteredMirage2.x && mirage1.y === unfilteredMirage2.y) ? unfilteredMirage3 : unfilteredMirage2;
+
+        const x1 = parseFloat(mirage1.x);
+        const y1 = parseFloat(mirage1.y);
+        const x2 = parseFloat(mirage2.x);
+        const y2 = parseFloat(mirage2.y);
+
+        const getLine = (x: number) => {
+          if ((x > -46 && x < -43) || (x > -27 && x < -24))
```
quisquous, line 353: FWIW, you could do something like `Math.round(x / 5) * 5` if you wanted to round to the nearest multiple of 5.  That might make a lot of this easier to read.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+
+0.0 "--sync--" sync / 00:0839::The Trial of Knowledge is sealed off/ window 0,1
+12.2 "--sync--" sync / 14:[^:]*:Silkie:7759:/ window 15,15
+14.9 "Fizzling Suds" sync / 1[56]:[^:]*:Silkie:7759:/
+21.0 "Soap's Up" sync / 1[56]:[^:]*:Silkie:775A:/
+22.0 "Fizzling Duster" sync / 1[56]:[^:]*:Silkie:775D:/
+32.2 "Dust Bluster" sync / 1[56]:[^:]*:Silkie:776C:/
+38.4 "Fresh Puff" sync / 1[56]:[^:]*:Silkie:7766:/
+43.5 "Bracing Suds" sync / 1[56]:[^:]*:Silkie:7757:/
+50.1 "Squeaky Clean" sync / 1[56]:[^:]*:Silkie:(7751|7752):/
```
quisquous, line 22:
```suggestion
50.1 "--sync--" sync / 1[56]:[^:]*:Silkie:(7751|7752):/
```

If there's 1 2 3 abilities after, then maybe drop the castbar text to make the timeline easier to read.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+51.6 "Squeaky Clean 1" sync / 1[56]:[^:]*:Silkie:7753:/
+53.3 "Squeaky Clean 2" sync / 1[56]:[^:]*:Silkie:7754:/
+54.8 "Squeaky Clean 3" sync / 1[56]:[^:]*:Silkie:(7755|7756):/
+61.9 "Chilling Suds" sync / 1[56]:[^:]*:Silkie:7758:/
+64.0 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+69.0 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+69.4 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:7760:/
```
quisquous, line 29: This pattern is suspicious to me.  Is 7760 the damage and 775E the self-targeted ability? If so, add 7760 to `-ii` and remove it.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+73.0 "Chilling Duster" sync / 1[56]:[^:]*:Silkie:7763:/
+80.6 "Carpet Beater" sync / 1[56]:[^:]*:Silkie:774F:/
+89.7 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+89.7 "Chilling Duster and Bracing Duster" #sync / 1[56]:[^:]*:Silkie:(7768|7769):/
+100.9 "Total Wash" sync / 1[56]:[^:]*:Silkie:7750:/
+106.0 "Chilling Suds" sync / 1[56]:[^:]*:Silkie:7758:/
+112.1 "Fresh Puff" sync / 1[56]:[^:]*:Silkie:7766:/
+123.1 "Puff and Tumble" duration 0.7 #sync / 1[56]:[^:]*:Silken Puff:7771:/
+129.5 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+129.5 "Fizzling Duster x2 and Chilling Duster x2" #sync / 1[56]:[^:]*:Silken Puff:(776A|7768):/
```
quisquous, line 39: Is this Fizzling or Chilling? Or both at the same time?

If either, please say: `Fizzling Duster x2/Chilling Duster x2` and add an English replacement in `timelineReplace` to make it say `Fizzling/Chilling Duster x2` in English.

If both, please put them on separate lines.

EDIT: same comments for the below "and" duster lines.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+134.6 "Fizzling Suds" sync / 1[56]:[^:]*:Silkie:7759:/
+136.7 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+141.7 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+142.1 "Slippery Soap" #sync / 1[56]:[^:]*:Silkie:7762:/
```
quisquous, line 43: Same comment here as for the previous Slippery Soap.  If 7762 is the damage, please `-ii` and remove this line.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+#~~~~~~~~~~~~~~~~~~~~~~#
+
+# -ii 766F 7670 7676 7677 77BF 7660
+
+1000.0 "--sync--" sync / 00:0839::The Trial of Might will be sealed off/ window 1000,1
+1006.8 "--sync--" sync / 14:[^:]*:Gladiator of Sil'dih:7671:/ window 1007.3,5
+1011.5 "Flash of Steel" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7671:/
+1023.6 "Specter of Might" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7673:/
+1038.3 "Rush of Might" sync / 1[56]:[^:]*:Gladiator of Sil'dih:765B:/
+1040.4 "Rush of Might" sync / 1[56]:[^:]*:Gladiator of Sil'dih:765C:/
```
quisquous, line 96: Same comment here, re: remove the damage line and keep the self-cast line.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+1082.4 "Mighty Smite" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7672:/
+1095.5 "Curse of the Fallen" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7674:/
+1097.6 "--sync--" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7639:/
+1110.2 "--sync--" sync / 1[56]:[^:]*:Gladiator of Sil'dih:77BF:/
+1113.6 "Echo of the Fallen" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7675:/
+1121.3 "Flash of Steel" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7671:/
+1131.5 "--sync--" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7639:/
+1136.6 "Hateful Visage" sync / 1[56]:[^:]*:Gladiator of Sil'dih:766E:/
+1141.7 "Wrath of Ruin" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7663:/
+1150.7 "Gold Flame and Silver Flame" sync / 1[56]:[^:]*:Hateful Visage:(766F|7670):/
```
quisquous, line 112: Same comment here re: multiple abilities.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+1185.7 "Gold Flame and Silver Flame" sync / 1[56]:[^:]*:Hateful Visage:(766F|7670):/
+1187.3 "Nothing beside Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:(768B|768C):/
+1196.4 "Flash of Steel" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7671:/
+1206.6 "--sync--" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7639:/
+1212.7 "Curse of the Monument" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7666:/
+1219.9 "Sundered Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
+1220.5 "Sundered Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
+1221.1 "Sundered Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
+1221.7 "Sundered Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
+1222.3 "Sundered Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
```
quisquous, line 126: This probably should be a duration and not 5 abilities?

Did you run your log through test_timeline? I feel like this would have broken.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+2006.8 "--sync--" sync / 14:[^:]*:Shadowcaster Zeless Gah:74AF:/ window 1007.3,5
+2011.5 "Show of Strength" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74AF:/
+2023.7 "Infern Brand" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:7491:/
+2030.9 "Cryptic Portal" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:7494:/
+2042.0 "Firesteel Strike" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B0:/
+2042.4 "Firesteel Strike" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B1:/
+2043.8 "Firesteel Strike" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B2:/
+2051.8 "Blessed Beacon" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B3:/
+2052.2 "Blessed Beacon" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B4:/
+2054.3 "Blessed Beacon" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B5:/
```
quisquous, line 158: Should these be shortened or some removed?

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,179 @@
...
+2125.7 "Firesteel Strike" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B2:/
+2133.7 "Blessed Beacon" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B3:/
+2134.1 "Blessed Beacon" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B4:/
+2136.3 "Blessed Beacon" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B5:/
+2146.6 "Firesteel Fracture" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74AD:/
+2158.8 "Infern Brand" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:7491:/
+2170.2 "Cryptic Flames" sync / 1[56]:[^:]*:Shadowcaster Zeless Gah:74B6:/
+2173.3 "Cryptic Flames" sync / 1[56]:[^:]*:Infern Brand:74B7:/
+2174.8 "Cryptic Flames" sync / 1[56]:[^:]*:Infern Brand:74B7:/
+2175.7 "Cryptic Flames" sync / 1[56]:[^:]*:Infern Brand:74B7:/
```
quisquous, line 172: duration on 74B6?

quisquous (commented):
Thanks for all of this, it's super appreciated.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,488 @@
...
+        let ring = 3;
+        if (matches.id === '7660')
+          ring = 1;
+        if (matches.id === '7661')
+          ring = 2;
+        return output.text!({ ring: ring });
+      },
+      outputStrings: {
+        text: {
+          en: 'Outside Ring #${ring}',
```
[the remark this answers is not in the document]
Legends0: I've updated it to be more specific.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,488 @@
...
+      // CDA Echoes of the Fallen (Spread)
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'CDA' }),
+      condition: Conditions.targetIsYou(),
+      delaySeconds: (_data, matches) => parseFloat(matches.duration) - 4,
+      response: Responses.spread(),
+    },
+    {
+      id: 'ASS Thunderous Echo Reminder',
+      // CDA Thunderous Echo (Stack)
```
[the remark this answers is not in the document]
Legends0: Ah, seems I typed it wrong here. Fixed.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,488 @@
...
+    {
+      id: 'ASS Slippery Soap',
+      // Happens 5 times in the encounter
+      type: 'Ability',
+      netRegex: NetRegexes.ability({ id: '79FB', source: 'Silkie' }),
+      preRun: (data) => data.soapCounter++,
+      alertText: (data, matches, output) => {
+        if (data.suds === '7757') {
+          // Does not happen on first or third Slippery Soap
+          if (matches.target === data.me)
```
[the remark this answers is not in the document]
Legends0: Correct, but to be consistent I've switched the return statement. I had changed the output strings at one point and placed them wrong in this section.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,488 @@
...
+
+        const mirage1 = data.mightCasts[0];
+        const unfilteredMirage2 = data.mightCasts[1];
+        const unfilteredMirage3 = data.mightCasts[2];
+
+        if (mirage1 === undefined || unfilteredMirage2 === undefined || unfilteredMirage3 === undefined)
+          throw new UnreachableCode();
+
+        // Filter for unique
+        const mirage2 = (mirage1.x === unfilteredMirage2.x && mirage1.y === unfilteredMirage2.y) ? unfilteredMirage3 : unfilteredMirage2;
```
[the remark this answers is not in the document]
Legends0: In my logs I am seeing each spell cast twice. There are 4 casts total. We can ignore the last spell since it's just going to be a duplicate of one of the two. I think getCombatants would not work since this is getting the target location of the aoe. The Gladiator Mirages do cast at the target location as well and only twice, but their target location data is what is not reliable. The boss' 4 casts I have seen are reliable.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Hmm... I was thinking one is the first Cleave and the other is the second Cleave, not damage lines.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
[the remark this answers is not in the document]
Legends0: Ah, there is going to be 4 casts because there are two cleaves happening together.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: I have not run through test_timeline. I have had to use fflogs for generating the timelines as the utility could not find starting time for the encounters (might be bug or user error).

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Yea, looks like there is some paste errors here.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Yea, 7760 is the damage applied after cast end. This ability is a bit weird. It's basically 1 second after Slippery Soap damage goes out, the boss stops moving/tracking the target, so depending on latency player could receive the damage early and be moving to dodge followup mechanic and the boss follows them still.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: I'll modify this later.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: The one here is always x2 of both. The others can vary. I'll have to review what's possible at each...
Soaping Spree 1 (Slippery Soap hits non Chilling Duster):
- Chilling Duster + Fizzling Duster + Bracing Duster
- Chilling Duster x2 + Bracing Duster
Soaping Spree 1 (Slippery Soap misses dusters):
- Bracing Duster x2 + Chilling Duster
- Fizzling Duster x2 + Bracing Duster
- Bracing Duster x2 + Fizzling Duster

Soaping Spree 2:
- 2x Chilling Duster + 2x Fizzling Duster + 1x Bracing Duster (boss)
- 2x Chillind Duster + 2x Fizzling Duster + 1x Chillind Duster (boss)

Soaping Spree 3:
- 2x Fizzling Duster
- 1x Bracing Duster + 1x Fizzling Duster
- 1x Fizzling Duster + 1x Chilling Duster

Soaping Spree 4:
- 0x - 6x Chilling Duster + Bracing Duster (boss)

Soaping Spree 5:
- 4x Chilling Duster +  1x Bracing Duster (boss)

Soaping Spree 5 (if missed any puffs during Slippery soap)
- 0-6x Bracing Duster + 4x Chilling Buster + 1x Bracing Duster (boss)
- 
Soaping Spree 5 (if faced tether wrong way)
- 0-4x Chilling Duster + 0-4x Chilling Duster + 1x Bracing Duster (boss)

Soaping Spree 5 (Failing multiple soaping spree checks):
- 0-7x Bracing Duster + 0-8x Chilling Duster + 1x  Bracing Duster (boss)

I think you can get somewhat lucky and have puffs be on safe side and become Chillind Dusters as well, but I think this last Spree just cascades the explosions. Also, you can have limit break 2 from tank here and survive a bad duster or two.

These can also trigger chain reactions if mechanics are done wrongly. I've seen some weird interactions where dusters go off before Soaping Spree finishes cast.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
quisquous: Oh, if it's first and second, then it should be "Rush of Might (near)" and "Rush of Might (far)" kinda thing? or even "1" and "2" vs just two identically named abilities back to back.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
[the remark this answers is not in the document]
Legends0: Correct, it is two of each ability id. One set at 10.2s cast time, the other at 12.2s.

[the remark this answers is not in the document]
Legends0: I think it's possible to filter off castTime and just grab one.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Makes sense. It could be 1 or 2.
There is also the possibility of stating the safe spot in it, but I think that doesn't work if it happens to be 2/3 or 3/2 pattern.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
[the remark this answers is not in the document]
Legends0: Ah, it's simpler. The two of one ability id will get both positions.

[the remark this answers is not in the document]
Legends0: Looking at this more. I see that the data from the boss can sometimes be wrong.
I have a log where the boss targets (-35, -281) for all the casts, which is not correct. Although this seems really rare. I also have a log where the target locations did not change between north and south, which is also wrong

It might be possible to fix this with OverlayPlugin because the boss is targetting one of its own entities, which maybe is invisible and updated before end of cast...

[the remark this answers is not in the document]
Legends0: Ok I also noticed an oddity, but this is just how the mechanic is, the north/south line 1 here actually does not abide by the < -271 rule. They overlap so a more strict check is needed.

I'll code out a overlayPlugin of this next.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: Using 5 works for x value. For y value, highest integer was 3 that seems to work given the exact floating point values for the positions.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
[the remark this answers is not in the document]
Legends0: This uses overlayPlugin now which will require in game testing to see if it gets the correct values.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,187 @@
...
+299.4 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+303.5 "Bracing Duster / Chilling Duster / Fizzling Duster" sync / 1[56]:[^:]*:Silkie:(7764|7763|7765):/
+303.8 "Soapsud Static?" #sync / 1[56]:[^:]*:Silken Puff:77ED:/
+309.0 "Bracing Suds / Chilling Suds / Fizzling Suds" sync / 1[56]:[^:]*:Silkie:(7757|7758|7759):/
+311.1 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+316.1 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+320.0 "Bracing Duster / Chilling Duster | Fizzling Duster" sync / 1[56]:[^:]*:Silkie:(7764|7763|7765):/
+322.9 "--sync--" sync / 14:[^:]*:Silkie:79E0:/ window 15,5
+332.9 "Total Wash (enrage)" sync / 1[56]:[^:]*:Silkie:79E0:/
+338.1 "Total Wash (enrage)" sync / 1[56]:[^:]*:Silkie:79F9:/
```
quisquous, line 89: Which one of these is the enrage?

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,187 @@
...
+1161.4 "Flash of Steel" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7671:/
+1171.6 "Accursed Visage" sync / 1[56]:[^:]*:Gladiator of Sil'dih:768D:/
+1176.7 "Wrath of Ruin" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7663:/
+1185.7 "Gold Flame x4" sync / 1[56]:[^:]*:Hateful Visage:766F:/
+1185.7 "Silver Flame x4" #sync / 1[56]:[^:]*:Hateful Visage:7670:/
+1187.3 "Nothing beside Remains" sync / 1[56]:[^:]*:Gladiator of Sil'dih:(768B|768C):/
+1196.4 "Flash of Steel" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7671:/
+1206.6 "--sync--" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7639:/
+1212.7 "Curse of the Monument" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7666:/
+1219.9 "Sundered Remains 1" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
```
quisquous, line 130:
```suggestion
1219.9 "Sundered Remains 1" #sync / 1[56]:[^:]*:Gladiator of Sil'dih:7668:/
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
```diff
@@ -0,0 +1,660 @@
...
+      type: 'StartsUsing',
+      netRegex: NetRegexes.startsUsing({ id: '765C', source: 'Gladiator of Sil\'dih' }),
+      delaySeconds: 0.5,
+      promise: async (data, matches) => {
+        if (data.mightCasts.length === 2)
+          data.mightCasts = [];
+
+        // select the Hephaistoss with same source id
+        let gladiatorData = null;
+        gladiatorData = await callOverlayHandler({
```
quisquous, line 333: Can you remove this trigger from this PR so the rest of the patch can be merged and submit it separately?

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: They both are. If you survive the enrage hit somehow, such as rezzing, then 79F9 gets you after. 79F9  probably is a cycled repeat instant cast like the ones we have seen in DRS.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,187 @@
...
+141.7 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+145.7 "Bracing Duster / Fizzling Duster" sync / 1[56]:[^:]*:Silkie:(7764|7765):/
+146.0 "Soapsud Static?" #sync / 1[56]:[^:]*:Silkie:77ED:/
+160.1 "Total Wash" sync / 1[56]:[^:]*:Silkie:7750:/
+166.2 "Fresh Puff" sync / 1[56]:[^:]*:Silkie:7766:/
+172.3 "Eastern Ewers" sync / 1[56]:[^:]*:Silkie:776D:/
+178.6 "Rinse" sync / 1[56]:[^:]*:Eastern Ewer:776F:/ duration 7.3
+188.1 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+188.1 "Bracing Duster?" #sync / 1[56]:[^:]*:Silken Puff:7769:/
+188.1 "Fizzling Duster?" #sync / 1[56]:[^:]*:Silken Puff:776A:/
```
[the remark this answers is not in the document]
Legends0: I am also missing Chilling Duster there... so it may just say Two Dusters or something.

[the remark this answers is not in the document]
quisquous: Yeah `Duster x2` seems reasonable then?

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
Legends0: Sure!

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: I have reason to believe all bosses are like this in this instance since the instance itself does not reset, you get pulled back to the start. Also, you can get rezzed as the encounter resets too and trigger the encounter instantly.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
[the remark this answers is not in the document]
Legends0: I think this is resolved... I added a line for Fizzling Duster and made the question marks here into its own line that gets replaced to 'Duster' as it can be any duster. The only question would be is if the second duster can be a Chilling Duster, which then the line would need to be updated to reflect Chilling Duster / Fizzling Duster.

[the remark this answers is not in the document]
Legends0: Sorry, didn't see your comment there before I made the change.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: For what it's worth, the gladiator's second enrage cast is either 7A32 or 7A33 based of what i see for instant cast Flash of Steel. I'm not sure why there are two. Also not sure if Savage mode has these casts, oddly Show of Strength and Total Wash each only have the one version of instant cast spell and Show of Strength was added in 6.1 but used here.

Legends0:
Some notes on the last boss...

The Ball of Fire are strange. They get added to the instance on first Infern Brand cast, but then are not seen added again until new instance. So I guess between pulls the same combatants are used? This makes it such that the first pull you could track addCombatant, but then the remaining pulls it would not work. So once again, OverlayPlugin would need to be used. I also don't really see anything to show that the ball of fire is affected by the clockwise/counterclockwise mechanic, so not sure how that could be tracked. They may just teleport to their spot as well.
They appear to spawn (03 log):
(289, -120)
(294, -115)
(284, -95)
(289, -90)

Seen them get removed from these spots at end (04 log), and fflogs also seems to indicate they appear at these locations when Burn goes off:
(284, -125)
(279.20, -120) (299, -120) 
(284, -105) (294, -105)
(279, -90) (299, -90)
(284, -85)

The second Infern Band mechanic (with debuffs) looks like it can be resolved to at least pin point where players should start as there are logs for what appear to be the line numbers and band locations as the bands receive 95D effect with 1C2 through 1C9. Not sure which value is 1 and which is 8, but that could be figured out. The bands do get moved around, so to be determined if overlayPlugin would be needed or if the x,y data on the gainsEffect is reliable (will need to look at more log lines).

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,225 @@
...
+54.8 "Squeaky Clean 3" sync / 1[56]:[^:]*:Silkie:(7755|7756):/
+61.9 "Chilling Suds" sync / 1[56]:[^:]*:Silkie:7758:/
+64.0 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+69.0 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+73.0 "Chilling Duster" sync / 1[56]:[^:]*:Silkie:7763:/
+80.6 "Carpet Beater" sync / 1[56]:[^:]*:Silkie:774F:/
+89.7 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+89.7 "Fizzling Duster?" #sync / 1[56]:[^:]*:Silken Puff:776A:/
+89.7 "Bracing Duster?" #sync / 1[56]:[^:]*:Silken Puff:7769:/
+89.7 "Chilling Duster?" #sync / 1[56]:[^:]*:Silken Puff:7768:/
```
quisquous, line 34:
```suggestion
89.7 "Bracing Duster / Chilling Duster / Fizzling Duster" #sync / 1[56]:[^:]*:Silken Puff:(7769|7768|776A):/
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,225 @@
...
+195.2 "Chilling Suds" sync / 1[56]:[^:]*:Silkie:7758:/
+202.3 "Carpet Beater" sync / 1[56]:[^:]*:Silkie:774F:/
+204.4 "--sync--" sync / 1[56]:[^:]*:Eastern Ewer:79FB:/
+209.4 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+213.4 "Chilling Duster" sync / 1[56]:[^:]*:Silkie:7763:/
+225.9 "Bracing Suds" sync / 1[56]:[^:]*:Silkie:7757:/
+233.0 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+233.0 "Chilling Duster?" #sync / 1[56]:[^:]*:Silken Puff:7768:/
+233.0 "Bracing Duster?" #sync / 1[56]:[^:]*:Silken Puff:7769:/
+233.0 "Fizzling Duster?" #sync / 1[56]:[^:]*:Silken Puff:776A:/
```
quisquous, line 63:
```suggestion
233.0 "Bracing Duster / Chilling Duster / Fizzling Duster" #sync / 1[56]:[^:]*:Silken Puff:(7769|7768|776A):/
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,225 @@
...
+262.3 "--sync--" sync / 1[56]:[^:]*:Silkie:(7751|7752):/
+263.8 "Squeaky Clean 1" sync / 1[56]:[^:]*:Silkie:7753:/
+265.5 "Squeaky Clean 2" sync / 1[56]:[^:]*:Silkie:7754:/
+266.9 "Puff and Tumble x4" duration 0.7 #sync / 1[56]:[^:]*:Silken Puff:7771:/
+266.9 "Squeaky Clean 3" sync / 1[56]:[^:]*:Silkie:(7755|7756):/
+276.0 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+276.0 "Chilling Duster x4?" #sync / 1[56]:[^:]*:Silken Puff:7768:/
+276.0 "Bracing Duster" #sync / 1[56]:[^:]*:Silkie:7769:/
+# 276.0 "Fizzling Duster?" #sync / 1[56]:[^:]*:Silken Puff:776A:/ # Possible here, but means a mechanic failed
+# 276.0 "Bracing Duster?" sync / 1[56]:[^:]*:Silken Puff:7769:/ # Possible here, but means a mechanic failed
```
quisquous, line 76: In general, I think it's better to not include failures in the timeline.

```suggestion
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,225 @@
...
+283.1 "Total Wash" sync / 1[56]:[^:]*:Silkie:7750:/
+292.3 "Bracing Suds / Chilling Suds / Fizzling Suds" sync / 1[56]:[^:]*:Silkie:(7757|7758|7759):/
+294.4 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+299.4 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+303.5 "Bracing Duster / Chilling Duster / Fizzling Duster" sync / 1[56]:[^:]*:Silkie:(7764|7763|7765):/
+303.8 "Soapsud Static?" #sync / 1[56]:[^:]*:Silken Puff:77ED:/
```
quisquous, line 82: Is this a failure as well? If so, then remove?

```suggestion
```

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,225 @@
...
+112.1 "Fresh Puff" sync / 1[56]:[^:]*:Silkie:7766:/
+123.1 "Puff and Tumble" duration 0.7 #sync / 1[56]:[^:]*:Silken Puff:7771:/
+129.5 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+129.5 "Fizzling Duster x2" #sync / 1[56]:[^:]*:Silken Puff:776A:/
+129.5 "Chilling Duster x2" #sync / 1[56]:[^:]*:Silken Puff:7768:/
+134.6 "Bracing Suds / Fizzling Suds" sync / 1[56]:[^:]*:Silkie:(7757|7759):/ 
+136.7 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+141.7 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+145.7 "Bracing Duster / Fizzling Duster" sync / 1[56]:[^:]*:Silkie:(7764|7765):/
+146.0 "Soapsud Static?" #sync / 1[56]:[^:]*:Silkie:77ED:/
```
quisquous, line 46: Does this only sometimes happen?

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,225 @@
...
+1011.5 "Flash of Steel" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7671:/
+1023.6 "Specter of Might" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7673:/
+1038.3 "Rush of Might 1" sync / 1[56]:[^:]*:Gladiator of Sil'dih:765B:/
+1040.4 "Rush of Might 2" sync / 1[56]:[^:]*:Gladiator of Sil'dih:765C:/
+1042.8 "Specter of Might" sync / 1[56]:[^:]*:Gladiator of Sil'dih:7673:/
+1057.5 "Rush of Might 1" sync / 1[56]:[^:]*:Gladiator of Sil'dih:765B:/
+1059.6 "Rush of Might 2" sync / 1[56]:[^:]*:Gladiator of Sil'dih:765C:/
+1062.0 "--sync--" sync / 1[56]:[^:]*:Gladiator of Sil'dih:6854:/
+1067.0 "Sculptor's Passion" sync / 1[56]:[^:]*:Gladiator of Sil'dih:766C:/
+1067.3 "Sculptor's Passion" sync / 1[56]:[^:]*:Gladiator of Sil'dih:79F3:/
```
quisquous, line 108: I'm guessing that `79F3` is damage here, right? If so, can you remove it and add it to `-ii` and also remove the other one below?

```suggestion
1067.0 "Sculptor's Passion" sync / 1[56]:[^:]*:Gladiator of Sil'dih:766C:/
```

quisquous (commented):
A few more timeline comments, sorry.

I can write out some better documentation for this if it's helpful, but for consistency with other timelines:
* ability names with a question mark are only used for (1) there's a loop coming up and some of these might happen and some might not or (2) there's a one-off ability that may or may not get used here (like first boss of DRS)
* if A or B might get used, then say "A / B"
* if A and B get used, then say "A" and "B" on different lines
* don't include abilities that only happen when there are failures (caveat: maybe if there's some timing involved to be somewhere or not be somewhere, then include it and comment out the sync, and maybe call it like `--towers--` or something)

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: It is possible to write in some branching for Silkie since we can branch from the Suds cast then rejoin the main timeline.

Legends0: I think branching would be the ideal goal, but I didn't code it in for simplicity.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
quisquous: Personally, I think given that it's `0.3s` later, you can just drop it from the timeline.  If it was an ability like 3-5s later or something, then it would be worth having.  But, because it's already so close to the existing one, this feels like a "castbar + damage, so don't list the damage" kind of case.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
quisquous: I'm not sure this is complicated enough to merit the branching, so keeping it simple for now seems good.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Assuming this is in regards to Soapsud Static, see other comment. However, the Soapsud Static lines are now removed.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Ok, I thought this one out some more.
There will always be a Bracing Duster and a Chilling Duster when resolved correctly. The last one has to be Chilling or Fizzling as two Bracing Dusters results in a guaranteed damage down. I have fixed it to show the "correct" three lines. There will always be two Bracing Dusters prior to making a Chilling Duster with Slippery Soap given the boss always has Bracing Suds here and converts the north duster + east or west.

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
Legends0: Looks like I also missed a line here. The boss always casts Bracing Duster here as it always has Bracing Suds at this point when it does Soaping Spree.

Correctly resolving the Slippery Soap means there should be 6x Chilling Dusters going off here as well.

Legends0:
Some more notes for Silkie.

I'm not sure it's predictable other than based on timeline/time in the encounter, but Silkie has two "Bracing Duster" casts and two "Fizzling Duster" casts.

Fizzling Duster 775D (Silken Puffs also cast this one), cast without Slippery Soap
Bracing Duster 7769 (Silken Puffs also cast this one), cast without Slippery Soap
Chilling Duster 7768 (Silken Puffs also cast this one), cast without Slippery Soap

Fizzling Duster 7765 (Only Silkie casts) with Slippery Soap
Bracing Duster 7764 (Only Silkie casts) with Slippery Soap
Chilling Duster 7763 (Only Silkie casts) with Slippery Soap

ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt
```diff
@@ -0,0 +1,219 @@
...
+64.0 "--sync--" sync / 1[56]:[^:]*:Silkie:79FB:/
+69.0 "Slippery Soap" sync / 1[56]:[^:]*:Silkie:775E:/
+73.0 "Chilling Duster" sync / 1[56]:[^:]*:Silkie:7763:/
+80.6 "Carpet Beater" sync / 1[56]:[^:]*:Silkie:774F:/
+89.7 "Soaping Spree" sync / 1[56]:[^:]*:Silkie:7767:/
+89.7 "Chilling Duster / Fizzling Duster" #sync / 1[56]:[^:]*:Silken Puff:(7768|776A):/
+89.7 "Bracing Duster" #sync / 1[56]:[^:]*:Silken Puff:7769:/
+89.7 "Chilling Duster" #sync / 1[56]:[^:]*:Silken Puff:7768:/
+100.9 "Total Wash" sync / 1[56]:[^:]*:Silkie:7750:/
+106.0 "Bracing Suds / Chilling Suds" sync / 1[56]:[^:]*:Silkie:(7757/7758):/
```
kshman, line 36: it seems like mistyping here
7757/7758 → 7757|7758

Legends0: Nice catch. I've corrected it.

quisquous:
Thanks for all the work on this, it's super appreciated.  I don't know that I'm going to have time to get into this dungeon for a few weeks myself.

PR #4968, opened 2022-10-25
i18n: update ko DSR translations

Bing-su:
from dsr clear youtube


Akurosia:
do you have a log with debug log lines available?

PR #4972, opened 2022-10-25
ci: should fix the npm release permission

trim21:


quisquous:
I guess I don't understand why this will fix anything? There's nothing very useful in `.npmrc`.

trim21:
setup-node will set auth token to `${NODE_AUTH_TOKEN}` in `.npmrc`

https://docs.github.com/cn/actions/publishing-packages/publishing-nodejs-packages#publishing-packages-to-the-npm-registry

trim21:
And I'm not 100% sure that if existing .npmrc will affect this

PR #4974, opened 2022-10-25
ci: fix npm release

trim21:
I tested on another repo, this should work.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -77,6 +77,7 @@ jobs:
 
       - run: node .github/scripts/npm-package.cjs
 
+      - run: cp .npmrc ./npm-package/
       - run: npm publish
         working-directory: npm-package
         env:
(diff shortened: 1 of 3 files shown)


trim21:
this `${NODE_AUTH_TOKEN}` accidentally break yarn but I guess it's not a problem.

quisquous:
How does it break it?

trim21:
```text
~/proj/cactbot (style/indent) # yarn
yarn install v1.22.19
error An unexpected error occurred: "Failed to replace env in config: ${NODE_AUTH_TOKEN}".
info If you think this is a bug, please open a bug report with the information provided in "C:\\Users\\Trim21\\proj\\cactbot\\yarn-error.log".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

not a problem because we're not using yarn, I just run `yarn` by habit.

PR #4979, opened 2022-10-25
ci: fix for yarn user

trim21:
`.github/scripts/npm-package.cjs` will create a `.npmrc ` in `npm-package/.npmrc` so release script will work.


trim21:
it's not actually `ci` scope but `chore` is not allowed

quisquous:
> it's not actually `ci` scope but `chore` is not allowed

Changing workflows feels like `ci` to me.

PR #4980, opened 2022-10-26
raidboss: Fix seal common translation for param netRegex

quisquous:
This would break on the eureka_hydatos version:
`netRegex: { line: '.* will be sealed off.*?', capture: false },`

This patch allows this to be translated properly.  However, this patch doesn't change the trigger itself because more work needs to be done on message vs gamelog etc before using the parameter version.


quisquous:
cc @trim21

PR #4981, opened 2022-10-26
lint: add `no-extra-parens`

trim21:
dprint will format these in new version, so remove them with eslint

https://eslint.org/docs/latest/rules/no-extra-parens

we can remove this after dprint is updated


quisquous:
Maybe `lint` and not `style`?

ui/fisher/fisher-ui.js
```diff
@@ -75,7 +75,7 @@ export default class FisherUI {
       bar.centertext = row.getAttribute('data-fish');
 
       // Step one: fill until the minimum time
-      if ((min && min !== 'undefined') && (max && max !== 'undefined')) {
+      if (min && min !== 'undefined' && (max && max !== 'undefined')) {
```
quisquous, line 78:
```suggestion
      if (min && min !== 'undefined' && max && max !== 'undefined') {
```

I love that it only caught one of the parens here??

trim21:
This PR just change eslintrc and run `lintfix`, no other manually changes.

trim21:
and this also avoid https://github.com/quisquous/cactbot/pull/4954#discussion_r1002784695

PR #4983, opened 2022-10-26
raidboss: p8s unnecessary brackets around 793E

Legends0:
Just removed these unnecessary brackets.


quisquous:
I'm going to hold off on merging this in case I need to revert Trim's megapatch.

PR #4987, opened 2022-10-26
revert: NetRegexes.ability

trim21:
use `NetRegexes.ability` for `Ability` triggers

`NetRegexes.ability` will generate regex with type `2[12]` but `buildRegex('Ability', ...)` generate a regex with type `21`


trim21:
[](url)

Issue #4984, opened 2022-10-26
raidboss: p8s Blazing Footfalls Crush/Impact  triggers not triggering?


Legends0:
Seeing these not trigger today:
![image](https://user-images.githubusercontent.com/5702579/197932122-fc8d90d1-1ad5-4ae8-84c2-5b67f0ac8778.png)

Legends0:
probably related to #4982, since this is with latest commit.

trim21:
Oh yes I saw some trigger missing

PR #4988, opened 2022-10-26
raidboss: fix `NetRegexes.ability` is alias

trim21:
`NetRegexes.ability` match NetworkAbility and NetworkAOEAbility, but `buildRegex('Ability', ...) ` only match NetworkAbility 

Add a function `buildNetRegexForTrigger` to dispatch ability trigger to use `NetRegexes.ability` to build regex

should fix https://github.com/quisquous/cactbot/issues/4984


quisquous:
Good catch! Thanks for the quick fix.  I tested it locally and it seems like it works.

PR #4991, opened 2022-10-27
raidboss: ASS use new netregexes

Legends0:
~~Only tested the new trigger in raid emulator. Need to confirm that it is actually calling the safe spots. The output looks good as in 2 players in one corner and the other 2 players in the opposite corner.~~
Converts ASS triggers to use new netregex format.


Legends0:
I think there can be a trigger written to call out for when to break the Infern Brand line, but I'm not sure yet if we can get which color (orange or blue) to break because we need to get data on where the Arcane Font(s) are.

Legends0:
Telling when to cut here is a bit like P4S Curtain Call in that there is a Magic Vulnerability Up that we _could_ track, but unfortunately there is a second source between the cuts that can also trigger Magic Vulnerability Up which would potentially break the trigger if player is hit by Cast Shadow.

I have currently opted to just have the trigger reduce to infoText and remove the word Cut for cuts 5-8. I can think of some ways to make it safe for a 5th cut, but I do not know a consistent simple method for the last 3 cuts. There's too many variables.

Legends0:
Would it be OK to store time hit by Cryptic Flame, then on the trigger for the player to be able to break their second flame add some delaySeconds to it based off the current time (or logged timestamp)?

xiashtra:
> I guess you mean “add” in title？

the zone is abbreviated "ASS" for "Another Sil'dihn Subterrane".

quisquous:
Sorry, a PR should really do one thing.  Can you split the "new regexes" and "extra triggers" into separate PRs?

quisquous:
> Would it be OK to store time hit by Cryptic Flame, then on the trigger for the player to be able to break their second flame add some delaySeconds to it based off the current time (or logged timestamp)?

I don't know what this mechanic is.  I'm not quite sure what you mean, re: "based off the current time".  Wouldn't that just be a delaySeconds from a trigger?  Happy to discuss more on the PR for just this trigger.

PR #4994, opened 2022-10-27
raidboss: additional triggers for ASS second boss

xiashtra:
@Legends0 I'd appreciate you looking these over so we don't duplicate work


Legends0:
Aside from a Rush of Might trigger utilizing OverlayPlugin, I do not have other triggers in the works for second boss mechanics so no worries about duplicate work there. It's good to see you have a solution to these.

Legends0:
I think for some of the Golden/Silver Flame results, there is only one solution. A more advanced version of the trigger could be similar to our p8s/p6s triggers that say where in the 16 block space to stand.

xiashtra:
This is a "use eyes" trigger right now; it reminds you what debuffs you have but you have to solve the resolution yourself.

There should only be one valid double gold or double silver quadrant per pull, but I believe it's possible to have one or two quadrants for single gold/single silver.

Legends0:
FYI, to do such a trigger based on where the gold/silver are looks like it requires OverlayPlugin because I am seeing invalid heading data in my logs. Positional data is correct, but the heading data is out of date.

PR #4996, opened 2022-10-27
raidboss: ASS add Infern Brand 2 triggers

Legends0:
Calls out the player's brand number which quadrant to start at during Infern Brand 2 and. This is done by checking where the player's Brand debuffs match the orange and blue flame's brand numbers.

First player will receive call to cut based on receipt of First Flame debuff. Once an Infern Brand has lost its brand number (a player cut the flame), the next player in line gets the call to cut their flame number.

Utilizes timestamp of Cryptic Flame hit that applies Magic Vulnerability Up to compare with timestamp of the flame that was cut before player's turn to cut to ensure there is delay added if necessary for a safe quick cut.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
@@ -24,7 +24,7 @@ export interface Data extends RaidbossData {
   myFlame?: number;
   brandEffects: { [effectId: number]: string };
   brandCounter: number;
-  flameCounter: number;
+  myLastCut?: number;
 }
 
 const triggerSet: TriggerSet<Data> = {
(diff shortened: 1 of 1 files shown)


Legends0:
Proposed Solution for handling of Magic Vulnerability Up - dynamic delaySeconds on `ASS Infern Brand 2 First Cuts`:

- A trigger that fires on `Ability` Cryptic Flame where `condition: Conditions.targetIsYou()` will store the timestamp. This ability applies the Magic Vulnerability Up and the ability occurs only on the player who cuts their flame.
- `ASS Infern Brand 2 First Cuts` delaySeconds looks at the timestamp stored from Cryptic Flame trigger and compares to timestamp of its netregex, if it is the player's turn to cut and it has not been 7.96 seconds since they received Cryptic Flame hit, add delay until it has been 7.96 seconds.

quisquous:
Can you explain this mechanic more? I think I'd like to avoid dealing with timestamps as I think it will probably not play well with the raidemulator.  Is this something that could be solved with a trigger when the magic vuln falls off or with some delay related to the magic vuln getting added/removed?

Legends0:
> Can you explain this mechanic more? I think I'd like to avoid dealing with timestamps as I think it will probably not play well with the raidemulator. Is this something that could be solved with a trigger when the magic vuln falls off or with some delay related to the magic vuln getting added/removed?

The mechanic is as such:
1. Players receive ${number} Brand debuff, last 12 seconds and 8 numbered lines (1-4 orange and blue) spawn in a + pattern.
2. Players look at their number and look where their matching numbered lines are.
3. Players have until their brand debuff expires to cross flame lines to get to the corner they need to be for cutting their flame.
4. Players then look at the first set of Arcane Fonts that spawned prior to debuff to know whether they need to cross over orange or blue flame lines to avoid the Arcane Font's line aoe. Typically what will happen is one group will be in the safe spot and the other will be in the aoe if lines are not cut.
5. Once players receive their Flame number (after Brand debuff expired), players must then cut in order and have 22 seconds to do so before their flame expires. Crossing a line that does not match your Flame # results in a long stun, damage, and does not cut the flame.
6. After a player cuts a flame, they get hit by Cryptic Flame ability, which applies 7.96s Magic Vulnerability Up.
7. Around the time players finish cutting the first set of 4 flames, the boss casts Cast Shadow which is a telegraphed pie slice aoe that players move into after it goes off. If player gets hit by this, they also receive Magic Vulnerability Up, which likely kills them if they just cut their flame.
8. Players have dodged Cast Shadow and must now get to the safe spot to dodge the other Arcane Fonts which is done by cutting the last 4 numbered flames of the other color. Technically the players can cut the next flame immediately after they have cut their first flame if they have access to it and the previous of the same color has already been cut, but if they still have the Magic Vulnerability Up from first cut they will die to Cryptic Flame damage received from the cut. So players must wait for their vulnerability to expire prior to cutting the other color and they most likely will not have access to the other line if all 4 of the first color have not been cut.

quisquous:
I assume the magic vuln up from cast shadow is the same id?

What about a "loses magic vuln" trigger for cutting the first line, with a suppress that's long enough to cover the cast shadow and the vuln from cutting the second line?

Legends0:
n

> I assume the magic vuln up from cast shadow is the same id?
> 
> What about a "loses magic vuln" trigger for cutting the first line, with a suppress that's long enough to cover the cast shadow and the vuln from cutting the second line?

The Magic Vulnerability Up are the same id, B7D. Two other mechanics prior to Infern Brand 2 also apply this, but I think it's possible to ignore them with the infern brand counter condition already setup. It could work for the first flame as it is probably a wipe if they did not cut all 4 flames prior to Arcane Font going off, but would not work for subsequent flames as it all depends on the other player's timing for cutting the flames. They may not be able to access their flame by the time the vulnerability expires.

Legends0:
The last sentence there is a bit why it differs from P4s, in P4s there wasn't anything preventing access to breaking the tether. Here, you get stunned if you go too early, so we have both that factor and the magic vulnerability on two separate timers.

xiashtra:
I don't believe getting hit by Cast Shadow actually gives you another Magic Vulnerability Up.  It appears that it only gives a 30s Damage Down (0xB5F).  However, it will most likely overlap with the Magic Vulnerability Up debuff from breaking the lines, and so is usually fatal.

Legends0:
> I don't believe getting hit by Cast Shadow actually gives you another Magic Vulnerability Up. It appears that it only gives a 30s Damage Down (0xB5F). 

Ah, you're right. I also see that the other Cryptic Flames (70D0) happens if the player stepping into a flame that is not their number and that also applies a Magic Vulnerability Up 7.96s + 3s Stun.

There is 13s? to cut 4 flames and get to safety to avoid the Arcane Font's Blazing Benifice line aoe. Getting a stun does mean the player will die to cutting the correct flame, and I think death also results in loss of the buff to be able to cut the flames, so it will lead to an unavoidable Blazing Benifice. Plus I think any uncut wires might trigger explosion as well at end of mechanic.

Legends0:
Ok, I saw that this is not the first trigger to utilize parsing of the timestamp. It does look like it works in raid emulator.

The one catch is that there is still potential for players to now be late on cutting where they could potentially have sacraficed. It could get more complex such as checking timestamp of their debuff against the timestamp of the infern brand cut and see how it compares with the delay that has been added. If the delay is greater than time remaining, I think an exception should be made to drop the delay and just call cut as it means they die, but party may still survive.

Legends0:
This does call wrong after testing it in game. The diagonal needs to be flipped.

PR #4999, opened 2022-10-27
raidboss: ASS missing slippery soap count

### LINKED ISSUE #4992: raidboss: Incorrect ASS triggers
## Description

There are a few issues with the ASS triggers for the first boss (Silkie) merged in #4967:

For the first and third Soaping Spree, the Suds Collector does not account for Silkie losing its suds buff, and therefore the Soaping Spree trigger makes incorrect callouts based on the last seen suds buff.

The Slippery Soap trigger is triggering off of a phantom cast from a hidden actor (0x79FB) instead of the cast from the boss actor (0x775E); for the third Slippery Soap, it fails to execute, since the phantom cast does not happen here.  This results in the "Behind puffs and party (East/West)" callout never being triggered.

@Legends0

Legends0:
As shown in the timeline, the one Slippery Soap here is cast not by Silkie, but the Eastern Ewer. Since the count is missed, the later triggers do not work.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
@@ -138,7 +138,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'ASS Slippery Soap',
       // Happens 5 times in the encounter
       type: 'Ability',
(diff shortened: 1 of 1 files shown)


Legends0:
@xiashtra I'm not sure it needs all the changes in your PR, does this fix work?

Legends0:
Closing, this should be merged into xiashtra's PR.

PR #5000, opened 2022-10-28
lint: format trigger files

trim21:
files in `/ui/` only get formatted. changed files are `dprint.*json` `package.json` and `util/process_triggers_folder.ts`

we can change it back after [dprint support config override](https://github.com/dprint/dprint/issues/574)


trim21:
And I just found somehow I revert change of `util/process_triggers_folder.ts`...

ui/raidboss/data/00-misc/general.ts
```diff
@@ -211,7 +215,8 @@ const triggerSet: TriggerSet<Data> = {
         'has initiated a ready check': 'eine Bereitschaftsanfrage gestellt',
         'You have commenced a ready check': 'Du hast eine Bereitschaftsanfrage gestellt',
         'You poke the striking dummy': 'Du stupst die Trainingspuppe an',
-        'You psych yourself up alongside the striking dummy': 'Du willst wahren Kampfgeist in der Trainingspuppe entfachen',
```
trim21, line 214: dprint doesn't have a config for this, for me I really like this didn't break into new line

quisquous: Yeah, I think it's not great, but I also don't want lines to get longer in non-trigger files either.  I think the fact that it auto-formats and this only tends to happen on really long game log lines infrequently makes it not as bad as it could be.

PR #5004, opened 2022-10-28
raidboss: ASS earlier Infern Brand 2 color call

Legends0:
Will need #5001 fixes first.


Legends0:
> I'm not sure I understand what's changed here. Could you merge/rebase on main?

It should just be adding a check on the number of Arcane Fonts detected to now state the first color in addition to whichever output it would state previously.

Legends0:
I may need to recheck log, as I think the portals might actually spawn a second or two before this callout, but this timing is something like 12s sooner than what I previously had. Would likely need OverlayPlugin to do the soonest callout to check just x,y coord of first one that spawns which I think I'll recode in for use with stating move (if needed) to east/west or north/south for the Arcane Font's aoe after the 1C5 or 1C9 counts expire on the brands.

Legends0:
I think sometimes the addcombatant call can be late, making it miss.

quisquous:
> EDIT: The odd thing is, this happens in game (1 in 20 pulls or so), but when raid emulator goes back and replays it, it does not happen in raid emulator. Seems it only happens on the blue first thing too.

Well, added combatant comes from memory polling in the ffxiv plugin, so it's not surprising that sometimes the line comes in a slightly different place or order.  It's very weird that raidemulator would behave differently though, given that there's no getCombatants here?

quisquous:
Also, does this need more testing, re: that delay? Or is this good to merge?

Legends0:
> Also, does this need more testing, re: that delay? Or is this good to merge?

I likely won't get around to testing more until next week. It happened so rarely I think just the 0.1s I added may work.

PR #5005, opened 2022-10-28
raidboss: ASS add Rush of Might Triggers

Legends0:
Utilizes OverlayPlugin to get more up to date data. Log data was inaccurate, this data was good when I tested.

I tested earlier and the second part was always correct. First part did not call due to wrong rounded y1 values, but that has been fixed so I think it's good. Wording will take some getting used to and there may be reason to make this relative callouts instead of current card/intercad based callouts since it can be difficult to understand at first.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
@@ -414,12 +414,10 @@ const triggerSet: TriggerSet<Data> = {
         if (!gladiator)
           return;
         data.mightCasts.push(gladiator);
-        console.error(`Rush of Might 1: Added Combatant at (${gladiator.PosX}, ${gladiator.PosY})`);
       },
       infoText: (data, _matches, output) => {
         if (data.mightCasts.length !== 2)
           return;
-        console.error(`Rush of Might 1: Received two gladiators, executing.`);
         const mirage1 = data.mightCasts[0];
         const mirage2 = data.mightCasts[1];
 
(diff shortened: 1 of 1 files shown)


Legends0:
So I tested this again tonight. It works very well with the default output if oriented north. It might be that an East/West needs to be flipped for the second set of cleaves, but I'm not to certain. Also possibly the whole intercard portion be removed and it just state the east/west portion as the north/south portion is well understood part of the mechanic for where the mirages are rushing towards. I'm not really sure how to state the side of the line or even if it should be saying line over some other word to describe the distance the mirages will be traveling.

Could just be one of those types of triggers where if you know the mechanic, recognize the output, it no longer becomes an issue as all the important detail is there long in advance. Potentially this needs a durationSeconds with how early it calls it.

quisquous:
Yeah, I am really not sure what text to use for this.  Like trying to explain the difference between NE 3 + NW 1 vs NE 3 + NW 2 seems very subtle.  I wonder if it'd make sense to just say "SE 1 + SW 3" kinda thing? and then you can just find those two lines and go to the intersection? This just feels fairly precise in terms of positioning to say something like "NE east of line" which feels a bit ambiguous.  I haven't done it yet and so it's mostly theoretical, but I'm just trying to think it through.

xiashtra:
These callouts might be too detailed; for anyone that understands the mechanic, the only information you need to know where to stand for the 1/2 or 1/3 combinations is which side (east/west) is dashing farther, and how far (2 or 3 lines) they are dashing.  North/south is extremely obvious, and happens in the same order every pull anyways so shouldn't need a callout.  For the 2/3 combinations, there are safespots relative to the arena floor texture such that you only need to know which side is dashing farther, without any need of an "east/west side of [x] line" callout.  Granted, making a callout relative to a floor texture design (or convenient marker placed on such design) is a whole other challenge.

Legends0:
I've cut out the north/south portion of the first call and changed the side portion to be an outside/inside part if that makes sense. I also split the outputs here so that the player can more easily customize how they want to state the 2/3 part versus the 1/3, 1/2 patterns.

I think further reduction on output could be done on the 2/3 pattern since it's unique for where the line you are basing off of is, but it looks better to me to keep it stating the line number.

quisquous:
Sounds good.  We can always come back and refine the outputs too if there's a better way to phrase these.

PR #5008, opened 2022-10-28
raidboss: add Another Sil'dihn Subterrane (Savage)

xiashtra:
Simple find/replace on IDs.  Timeline will need extra attention.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
@@ -915,6 +915,99 @@ const triggerSet: TriggerSet<Data> = {
       condition: Conditions.targetIsYou(),
       run: (data, matches) => data.myLastCut = Date.parse(matches.timestamp),
     },
+    {
+      id: 'ASS Banishment',
+      // Players receive invisible effect that indicates rotation and direction
+      // of their teleport attached teleport pad
+      //
+      // At the same time, two teleports on North and South are also marked:
+      // one rotates outside the arena, the other rotates towards the inner rows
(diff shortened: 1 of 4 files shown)


xiashtra:
Timelines should be fixed up now, just need to integrate any outstanding ASS fixes/improvements.

quisquous:
Is savage 100% identical to non-savage here?

Do you have a list of all the ids you've searched and replaced? Mostly wondering if there's some way to semi-automate keeping this up to date as changes are being made rather than trying to manually integrate everything.

xiashtra:
> Is savage 100% identical to non-savage here?
> 
> Do you have a list of all the ids you've searched and replaced? Mostly wondering if there's some way to semi-automate keeping this up to date as changes are being made rather than trying to manually integrate everything.

Yes, ASS Savage is mechanically 100% identical to normal.  The differences are: enemies have more HP, enemy attacks do more damage (especially the trash mobs), some abilities no longer have any telegraphs, and there is an overall enrage timer for the entire dungeon.  Also, no resurrections whatsoever (not even the special Variant Raise II that's in normal Criterion), and if the party wipes, the entire dungeon resets (all bosses and trash respawn and you have to start over from the very beginning).

I definitely think there is an opportunity here for programmatically keeping Criterion and Criterion Savage triggers in-sync, especially since we will be getting at least two more of these during Endwalker, according to the roadmap.  It may even be useful for EX => Unreal as well.  Unfortunately, I did not keep an organized list of ID replacements, though I could probably recreate something by hand from what's been done so far.

xiashtra:
@quisquous I'm promoting this from draft to consideration for merge.  It integrates all ASS changes up to the 0.27.21 release.

quisquous:
Thank you!

PR #5012, opened 2022-10-28
build: upgrade tsconfig target to es2022

trim21:
chrome95 and node18 have full support of es2022 now.


util/test_timeline.ts
```diff
@@ -477,7 +477,7 @@ export const registerTestTimeline = (
 };
 
 class TestTimeline extends Timeline {
-  public override ui?: TestTimelineUI | undefined;
+  declare ui?: TestTimelineUI | undefined;
```
trim21, line 480: This is because ts's class property is added before ES standard and have conflct.

tsconfig.json
```diff
@@ -42,6 +42,7 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedIndexedAccess": true,
     "inlineSources": true,
+    "useDefineForClassFields": false,
```
trim21, line 45: This option disable es2022 new feature https://github.com/tc39/proposal-class-fields when emiting

tsconfig.json
```diff
@@ -42,7 +42,7 @@
     "noFallthroughCasesInSwitch": true,
     "noUncheckedIndexedAccess": true,
     "inlineSources": true,
-    "useDefineForClassFields": false,
+    "useDefineForClassFields": true,
```
quisquous, line 45: This is a behavior change though, right? Previously, it was as if this flag was false?

trim21: This is because ts's class property is added before ES standard and have conflct. When targeting es2022 a tsconfig option for compatibility useDefineForClassFields is default true instead of false.

No behavior change.

trim21: I may be not clear enough... set this to true is to keep behavior to not change

.eslintrc.cjs
```diff
@@ -145,9 +145,10 @@ const rules = {
       'allowAllPropertiesOnSameLine': true,
     },
   ],
-  'operator-linebreak': 'error',
+  'operator-linebreak': 'off',
   'prefer-arrow-callback': 'error',
   'prefer-const': 'error',
+  'prefer-promise-reject-errors': 'off',
```
quisquous, line 151: Sorry, but could you explain the eslint changes to me too? I'm not sure I understand why these need to change here.

trim21: I don't know eighrt, I just upgrade eslint and plugins version then these errors are reported

trim21: I guess they are actually broken before the upgrade

trim21: Oh because `eslint-config-google` change both these rule's option

quisquous: It's not clear to me why eslint and eslint-config-google needed to be updated? Did the old version of eslint not have node18 support?

.eslintrc.cjs
```diff
@@ -145,9 +145,10 @@ const rules = {
       'allowAllPropertiesOnSameLine': true,
     },
   ],
-  'operator-linebreak': 'error',
+  'operator-linebreak': 'off',
```
quisquous, line 148: This is the old behavior, rather than just disabling this.

```suggestion
  'operator-linebreak': [
    'error',
    'after',
    {
      'overrides': {
        ':': 'before',
        '?': 'before',
      },
    },
  ],
```

.eslintrc.cjs
trim21: They do, I just upgrade eslint and update its plugin by the way

quisquous: Oh, I guess that's my confusion then, because I was trying to understand why there were eslint changes in the same PR as a tsconfig target patch if they weren't related.

PR #5015, opened 2022-10-28
raidboss: ASS add Banishment triggers

Legends0:
During the final boss, Banishment is the mechanic that happens between Infern Brand 3 and Infern Ward 4. I have seen multiple strategies for this mechanic, some simpler than others.

It looked to me like there was an invisible debuff, B9A, that has its count field that correlates to the portal animation that the player has, so this can be used to make triggers for where to position to solve the mechanic. I compared raid_emulator with positioning I have in a log and created the mapping.

I also added two additional triggers for baiting the wards. I was not sure they are both necessary, I know the first one can be a bit more trickier as it has extra time before the bait and at least two players are likely to have been teleported in such a way that they need to move to not trigger an Infern Wave going in the wrong direction. The second wave has much longer stun lock before it goes off leaving less time to react.

Will be able to test this in game later.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.ts
@@ -912,6 +912,99 @@ const triggerSet: TriggerSet<Data> = {
       condition: Conditions.targetIsYou(),
       run: (data, matches) => data.myLastCut = Date.parse(matches.timestamp),
     },
+    {
+      id: 'ASS Banishment',
+      // Players receive invisible effect that indicates rotation and direction
+      // of their teleport attached teleport pad
+      //
+      // At the same time, two teleports on North and South are also marked:
+      // one rotates outside the arena, the other rotates towards the inner rows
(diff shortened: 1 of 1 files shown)


Legends0:
Hmm... could the bait triggers fire erroneously on player death as they lose the debuffs before expiration? If so they may need to be moved to something else like a timelinetrigger or some way to check time remaining/not from a death...

quisquous:
> It looked to me like there was an invisible debuff, B9A, that has its count field that correlates to the portal animation that the player has, so this can be used to make triggers for where to position to solve the mechanic. I compared raid_emulator with positioning I have in a log and created the mapping.

Amazing, nice work.  :100:

quisquous:
> Hmm... could the bait triggers fire erroneously on player death as they lose the debuffs before expiration? If so they may need to be moved to something else like a timelinetrigger or some way to check time remaining/not from a death...

This is fine.  If somebody dies and a trigger goes off, I think that's ok.  They're already dead, so a "wrong" trigger isn't going to do much here.

PR #5016, opened 2022-10-29
raidboss: ASS Infern Brand 5 Triggers

Legends0:
This adds a set of triggers to call out mechanic sequence in the last boss' Infern Brand 5 mechanic. This is similar to Infern Brand 2, but it adds the baited wards from Banishment to the equation.

This PR essentially:
- Calls where to start for the mechanic (Middle / Out)
- Calls when to cut (Using Magic Vulnerability and checking of previous cut lines as safety measures)
- Calls additional move for the 4rth cut person after the baits
- Calls Baiting of the Ward # (# is somewhat superflous information)
- Calls bait puddles

Limitations:
The main limitation I see with it is I do not have a known method to determine which Infern Ward is active until 0.5s before it goes off, so this utilizes a record of the first flame color and relies on players to have cut the correct colors. It only changes the color output on the 1st cut flame. Alternatively, we could just remove the color portion of the trigger until something better is out, but I figure if the first flame cut is wrong, it's a wipe so this better than nothing.

Code code probably be cleaned up a bit more, it's repeating a lot of output blocks.

I've only run this through raid emulator.
image


Legends0:
This is the last mechanic in the instance/encounter.

It would be nice to have some stuff for the first portals and the portals with Arcane Fonts. Balls of Fire I checked, but could not find out how to get where they will be casting their AOE / rotating.

Arcane Fonts appear to have a 2s cast where you could potentially have a late call by grabbing the target location of 74A6. The Fonts spawn 19s before this, but I don't see any other activity on them or at their location in the log to be able to see where they are teleporting to.

quisquous:
Sorry, this has conflicts now because I merged other things.

PR #5018, opened 2022-10-29
raidboss: fix ASS Soaping Spree with no suds

### LINKED ISSUE #5017: ASS incorrect call for Soaping Spree with no suds buff
The current ASS trigger for Soaping Spree has a bug that causes an incorrect call when Silkie does not have a current suds buff.

xiashtra:
resolves #5017


quisquous:
Thanks for all the fixes and work here.

Issue #5021, opened 2022-10-29
Husky did not prevent checking in dprint-failure files


quisquous:
See: [53a63cc](https://github.com/quisquous/cactbot/pull/5020/commits/53a63cc48b8a0036fda1bc77b8ea828d611eb3c1)

cc @trim21

PR #5022, opened 2022-10-30
lint: add missing ts file extention to husky dprint

trim21:
close https://github.com/quisquous/cactbot/issues/5021


quisquous:
Thank you for the fix!

PR #5023, opened 2022-10-30
ci: upgrade action for deprecation warning

trim21:


trim21:
fix these warnings https://github.com/quisquous/cactbot/actions/runs/3353584707

PR #5024, opened 2022-10-30
raidboss: enhance Tentacle trigger in Alzadaal

### LINKED ISSUE #4985: Add triggers for Ambujam tentacle safe spots in Alzadaal's Legacy
It looks very promising to use map effect lines for the tentacle effects.  Does anybody want to look into this and write some triggers for it? Even sending logs + video on this issue would be very helpful.

```
00|2022-10-26T07:19:10.0000000+03:00|282B||The ambujam uses Tentacle Dig.|8e0bdc23a7258829
257|2022-10-26T07:19:10.6510000+03:00|80030057|00020001|11|00|0000|22937abf77243336
257|2022-10-26T07:19:11.4110000+03:00|80030057|00200004|11|08|0000|2d10b2333a83c8a0
```

wexxlee:
Code has comments with info on the mapeffect lines, specifically the flags/locations and what effects/directions they correspond do.  Tested in raidemulator against earlier investigative pulls and with local instance running in some live pulls.

The direction callouts are not ideal, mostly because the safe spots are weirdly shaped, but I think this is the best way to concisely communicate the info.

Closes #4985.


ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
```diff
@@ -17,12 +46,124 @@ const triggerSet: TriggerSet<Data> = {
...
+          },
+        };
+
+        let cyanLoc = undefined;
+        let scarletLoc = undefined;
+        // location 10 = cyan tentacle path, location 11 = scarlet tentacle path
+        for (const tentacle of data.tentacleEffects) {
+          if (tentacle.location === '10') {
+            if (cyanTentacleLocations[tentacle.flags] !== undefined)
+              cyanLoc = cyanTentacleLocations[tentacle.flags]!;
```
quisquous, line 139:
```suggestion
              cyanLoc = cyanTentacleLocations[tentacle.flags];
```

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
```diff
@@ -17,12 +46,124 @@ const triggerSet: TriggerSet<Data> = {
...
+          } else if (tentacle.location === '11') {
+            if (scarletTentacleLocations[tentacle.flags] !== undefined)
+              scarletLoc = scarletTentacleLocations[tentacle.flags]!;
```
quisquous, line 142:
```suggestion
              scarletLoc = scarletTentacleLocations[tentacle.flags];
```

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
```diff
@@ -17,12 +46,124 @@ const triggerSet: TriggerSet<Data> = {
...
+          NW: 'SE',
+          NE: 'SW',
+          SE: 'NW',
+          SW: 'NE',
+        };
+
+        const tentacleFlag = data.tentacleEffects[0]!.flags;
+        if (tentacleFlag === undefined || scarletTentacleLocations[tentacleFlag] === undefined)
+          return output.default!();
+        const tentacleLocation = scarletTentacleLocations[tentacleFlag]!;
```
quisquous, line 74: In general, I want to try to avoid non-null assertions (`!`) because they could be incorrect.  You can usually get around the need for this by storing things in separate variables so that TypeScript can narrow the type for just that variable.  It can't do type narrowing on a specific index of an object.

The one exception is for `output` since TypeScript doesn't know about proxies, and it would be a giant pain to undefined-check all of those.

```suggestion
        const tentacleFlag = data.tentacleEffects[0]?.flags;
        if (tentacleFlag === undefined)
          return output.default!();
        const tentacleLocation = scarletTentacleLocations[tentacleFlag];
        if (tentacleLocation === undefined)
          return output.default!();
```

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
```diff
@@ -17,12 +46,124 @@ const triggerSet: TriggerSet<Data> = {
...
+            if (scarletTentacleLocations[tentacle.flags] !== undefined)
+              scarletLoc = scarletTentacleLocations[tentacle.flags]!;
+          }
+        }
+        if (
+          cyanLoc === undefined || scarletLoc === undefined ||
+          safeMap[cyanLoc][scarletLoc] === undefined
+        )
+          return output.default!();
+        const [safe0, safe1]: string[] = safeMap[cyanLoc][scarletLoc]!;
```
quisquous, line 150:
```suggestion
        if (cyanLoc === undefined || scarletLoc === undefined)
          return output.default!();
        const safeArray = safeMap[cyanLoc][scarletLoc];
        if (safeArray === undefined)
          return output.default!();
        const [safe0, safe1]: string[] = safeArray;
```

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
```diff
@@ -17,12 +46,124 @@ const triggerSet: TriggerSet<Data> = {
...
+          return output.default!();
+
+        // cyan tentacle ends on cardinal, and scarlet tentacle ends on intercard opposite cyan
+        // safe wedge runs from the cardinal adjacent to through the unoccupied intercard opposite of the cyan tentacle
+        const safeMap: {
+          [key in Cardinal]: { [idx: string]: [safecard: Cardinal, safeintercard: Intercard] };
+        } = {
+          // Cyan location: {possible Scarlet locations: [safe dirs]}
+          N: {
+            SE: ['W', 'SW'],
```
quisquous, line 116: Can these be "WSW" etc (we've used that elsewhere when we need to be more specific)?  I assume that's what you mean? Reference: <https://en.wikipedia.org/wiki/Cardinal_direction>

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
```diff
@@ -1,14 +1,43 @@
...
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  tentacleEffects: NetMatches['MapEffect'][];
+}
+
+export type Cardinal = 'N' | 'E' | 'S' | 'W';
+export type Intercard = 'NW' | 'NE' | 'SE' | 'SW';
+export const tentacleFlags = ['00080004', '00200004', '00800004', '02000004'];
+export type TentacleFlag = typeof tentacleFlags[number];
```
quisquous, line 16: This is just a string, I think.  (You have to say `export const tentacleFlags etc etc as const` to have a non-string type if you wanted to, but that will cause issues elsewhere in places that want strings.)

In other words, I think it's fine that `tentacleFlags` is a `string[]`, and we can probably drop `TentacleFlag` as an explicit type, and also just say `key: string` instead of `key in TentacleFlag` below.  It's not a logic change, but just a way to be more readable.

quisquous (commented):
Thanks! I have one wording comment and a few small TypeScript cleanup comments, but otherwise looks great overall, thank you!

Somehow I never knew that the tentacles always ended on cards and intercards respectively!

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
wexxlee: Not exactly.  The problem is that the safe area is unusually shaped, which creates a safe area at a cardinal (just slightly offset) that spans through the intercardinal.  Here's an image showing the safe area for North-Northeast for example: [https://i.imgur.com/5VyHHQy.png](https://i.imgur.com/5VyHHQy.png)  (boss faces North for this mech).

NNE (or any other secondary intercardinal) is technically safe in this diagram, but so is N (partially) and NE (but far away from the boss).  So my thought was to use the double direction nomenclature (e.g. North / Northeast) to communicate both the safe cardinal (N) and the direction in which the safe zone expanded from that cardinal (NE).  Not ideal, but I think it's more accurate than just calling a secondary intercardinal (like NNE).  That said, totally open to suggestions for improving this.

wexxlee:
> Somehow I never knew that the tentacles always ended on cards and intercards respectively!

Neither did I till I ran it 14 times. 😁

ui/raidboss/data/06-ew/dungeon/alzadaals_legacy.ts
quisquous: Oh! Ok that makes a lot of sense.  Thanks for the explanation.

PR #5030, opened 2022-10-31
raidboss: ASS/S add King's Will trigger

xiashtra:
King's Will is an ability done by trash mobs between the first and second boss in ASS and ASS(S) that gives the mob a 100% physical damage increase for the next 12s.  In Savage, this causes the mob's auto-attacks to increase from ~33k per hit to ~66k per hit, effectively becoming mini-tankbusters.  This trigger provides a heads-up callout to the tank or healer for the incoming damage.

There was a question if a trigger for this ability was useful when merging the initial ASS triggers, so this pull request is made in isolation so it can be rejected if undesired.


quisquous:
I think this is good, but it could probably just say "big autos" with no "on who".  Anybody can feint / addle / heal / mit, so I think it's worth just saying for everybody.

xiashtra:
Simplified the callout to "big autos" for everyone per recommendation.

Issue #4508, opened 2022-05-20
`tsc` execution speed improvements


valarnin:
## Summary

There are a number of code paths that slow `tsc` execution.

Steps taken to check:

1. `npm i -g @typescript/analyze-trace`
2. `npm run tsc-no-emit -- --generateTrace ../cactbot-tracing`
3. `npx analyze-trace ../cactbot-tracing --force-millis 300`

Results:

```
Hot Spots
├─ Check file /mnt/c/Users/valarnin/Source/valarnin-cactbot/ui/raidboss/popup-text.ts (2467ms)
│  └─ Check expression from (line 695, char 70) to (line 697, char 11) (1918ms)
│     └─ Check expression from (line 697, char 11) to (line 697, char 10) (1915ms)
├─ Check file /mnt/c/Users/valarnin/Source/valarnin-cactbot/ui/raidboss/data/05-shb/eureka/delubrum_reginae_savage.ts (721ms)
│  └─ Check variable declaration from (line 95, char 7) to (line 3679, char 2) (633ms)
│     └─ Check expression from (line 95, char 38) to (line 3679, char 2) (509ms)
│        └─ Check expression from (line 174, char 13) to (line 2567, char 4) (326ms)
├─ Check file /mnt/c/Users/valarnin/Source/valarnin-cactbot/ui/raidboss/data/05-shb/ultimate/the_epic_of_alexander.ts (573ms)
│  └─ Check variable declaration from (line 352, char 7) to (line 3191, char 2) (505ms)
│     └─ Check expression from (line 352, char 38) to (line 3191, char 2) (375ms)
│        └─ Check expression from (line 553, char 13) to (line 2701, char 4) (308ms)
├─ Check file /mnt/c/Users/valarnin/Source/valarnin-cactbot/test/helper/test_trigger.ts (540ms)
│  └─ Check deferred node from (line 391, char 34) to (line 562, char 4) (485ms)
│     └─ Check variable declaration from (line 520, char 20) to (line 520, char 28) (407ms)
│        └─ Check expression from (line 520, char 32) to (line 520, char 50) (407ms)
├─ Check file /mnt/c/Users/valarnin/Source/valarnin-cactbot/ui/raidboss/data/05-shb/eureka/zadnor.ts (451ms)
│  └─ Check variable declaration from (line 105, char 7) to (line 2316, char 2) (425ms)
│     └─ Check expression from (line 105, char 38) to (line 2316, char 2) (324ms)
[...]

PR #5038, opened 2022-11-01
build: speedup tsc

trim21:
This is a (almost) type only change, it change the type of `TriggerSet` only. Our core file are using `LooseTriggerSet` and `LooseTrigger` so they are not affected.

This patch make `trigger.type` required for all cactbot built-in net triggers and merge `BaseTrigger` and `PartialNetRegexTrigger` to help tsc narrow trigger type.

about why merging types is explained here: https://github.com/quisquous/cactbot/issues/4508#issuecomment-1298080356

## result

`npx tsc` 35s -> 20s
`npx eslint .` 1m6.862s -> 40.713s
`npm test` 38s -> 20s
`npm run start` no improvement.

---

I know the team of swc is trying to port tsc to rust, wish them good luck…


ui/raidboss/popup-text.ts
```diff
@@ -762,7 +762,7 @@ export class PopupText {
               trigger.localNetRegex = Regexes.parse(localeNetRegex);
               orderedTriggers.push(trigger);
               found = true;
-            } else if (defaultNetRegex) {
+            } else if (defaultNetRegex !== undefined) {
```
trim21, line 765: this trigger eslint error about "always true"

trim21:
trace after this patch:

```text
~/proj/cactbot (speedup-tsc-a-little) # time npm run tsc-no-emit -- --generateTrace ../cactbot-tracing && npx analyze-trace ../cactbot-tracing --force-millis 200

> cactbot@0.27.21 tsc-no-emit
> tsc --noEmit --generateTrace ../cactbot-tracing


real    0m20.240s
user    0m0.046s
sys     0m0.166s
Hot Spots
├─ Check file c:\users\trim21\proj\cactbot\node_modules\typescript\lib\lib.dom.d.ts (366ms)
└─ Check file c:\users\trim21\proj\cactbot\ui\raidboss\data\05-shb\eureka\delubrum_reginae_savage.ts (208ms)
   └─ Check variable declaration from (line 94, char 7) to (line 3712, char 2) (185ms)
      └─ Check expression from (line 94, char 38) to (line 3712, char 2) (169ms)
         └─ Check expression from (line 2601, char 20) to (line 3711, char 4) (126ms)

No duplicate packages found
```

` Check expression from (line 2601, char 20) to (line 3711, char 4) (126ms)` is a huge `timelineReplace`

ui/raidboss/popup-text.ts
trim21: we may need to refactor the code for processing triggers, it would be much easy to maintain if it's pure function.

trim21: And I find the reason why it's not working:

```ts
export type LooseTriggerSet =
  & Exclude<Partial<TriggerSet<RaidbossData>>, 'triggers' | 'timelineTriggers'>
  & {
    /** @deprecated Use zoneId instead */
    zoneRegex?:
      | RegExp
      | { [lang in Lang]?: RegExp };
    triggers?: LooseTrigger[];
    timelineTriggers?: LooseTimelineTrigger[];
  };
```

here it's not `Exclude`, it should be `Omit`

ui/raidboss/popup-text.ts
```diff
@@ -718,7 +718,7 @@ export class PopupText {
           // time later.  This will clobber each time we
           // load this, but that's ok.
           trigger.filename = setFilename;
-          const id = trigger.id;
+          const id = trigger.id ?? trigger.toString();
```
trim21, line 721: here `id` is optional, and it's option in user trigger, I don't know how to handle there...

trim21: And I found it's a much more complicated type issue here https://github.com/quisquous/cactbot/issues/5039

fixing the `LooseTriggerSet` cause many lint error.

trim21:
This PR is ready to merge wither https://github.com/quisquous/cactbot/issues/5039 is fixed or not

quisquous:
Thanks for the investigation and the fix.  That's really smart to combine the types like that to help out tsc.

PR #5045, opened 2022-11-01
raidboss/test: fix `LooseTriggerSet` type

### LINKED ISSUE #5039: wrong type definition
https://github.com/quisquous/cactbot/blob/423dac8971f83a8002a670d54fac6dd2337b9b7a/types/trigger.d.ts#L211-L213

`Exclude` should be `Omit` to remove properties `triggers` and `timelineTriggers`.

Here `Exclude` does nothing, because its correct usage is `type NonEnLang = Exclude<'en' | 'de' | 'fr', 'en'>`

This is not a easy fix and it need many fixes in other files (mostly optional `id` and template string, and `id` need to be optional because of the [user triggers](https://github.com/quisquous/cactbot/blob/main/docs/RaidbossGuide.md#trigger-properties))…

trim21:
close #5039


types/trigger.d.ts
```diff
@@ -190,7 +190,7 @@ export type BaseTriggerSet<Data extends RaidbossData> = {
 };
 
 // If Data contains required properties that are not on RaidbossData, require initData
-export type TriggerSet<Data extends RaidbossData> =
+export type TriggerSet<Data extends RaidbossData = RaidbossData> =
```
quisquous, line 193: What does this change do? Sorry for my lack of knowledge, I just am not sure what `= RaidbossData` does.

trim21: default generic argument

quisquous: Thanks for the explanation.  It's just hard to search for something like that.  I didn't know you could do defaults on generics.

PR #5047, opened 2022-11-01
raidboss: refactor trigger loading

trim21:
this contains the change of #5045 and #5038, please review and merge them first

This PR seprated triggers loading into 3 seprated steps, loading, localizing and overriding for better maintenance


ui/raidboss/popup-text.ts
```diff
@@ -69,14 +64,16 @@ export type ProcessedTrigger = LooseTrigger & {
...
+  & ({ localRegex: RegExp } | { localNetRegex: RegExp });
 
-type ProcessedTriggerSet = LooseTriggerSet & {
-  filename?: string;
-  timelineTriggers?: ProcessedTimelineTrigger[];
-  triggers?: ProcessedTrigger[];
+export type LoadedTriggerSet = LooseTriggerSet & {
+  filename: string;
+  timelineTriggers?: Omit<LooseTrigger, 'netRegex'>[];
+  triggers?: LooseTrigger[];
```
quisquous, line 76: I think this `timelineTriggers` type is incorrect. Shouldn't it be `LooseTimelineTrigger[]`? And if that's true, aren't these both already the same types in `LooseTriggerSet`?

```suggestion
```

ui/raidboss/popup-text.ts
```diff
@@ -817,58 +678,236 @@ export class PopupText {
...
-
-    delete trigger.localRegex;
-    delete trigger.localNetRegex;
-
-    trigger.output = TriggerOutputProxy.makeOutput(
-      trigger,
-      this.options.DisplayLanguage,
-      this.options.PerTriggerAutoConfig,
-    );
+  TriggerSetEnabled(set: LoadedTriggerSet): boolean {
```
quisquous, line 714: Can you call this `IsTriggerSetEnabled`?

ui/raidboss/popup-text.ts
```diff
@@ -640,160 +633,28 @@ export class PopupText {
...
+      if (set.timelineTriggers) {
+        for (const trigger of set.timelineTriggers) {
+          // Timeline triggers are never translated.
+          timelineTriggers.push(this.ProcessTimelineTrigger(trigger));
         }
       }
 
       if (set.overrideTimelineFile) {
+        // TODO: set.filename won't be undefined
         const filename = set.filename ? `'${set.filename}'` : '(user file)';
```
quisquous, line 658: How can `set.filename` be undefined now?

```suggestion
        const filename = set.filename;
```

ui/raidboss/popup-text.ts
```diff
@@ -212,48 +209,40 @@ const textMap: TextMap = {
...
-      const triggerFile = (trigger: ProcessedTrigger) =>
-        trigger.filename ? `'${trigger.filename}'` : 'user override';
-      const oldFile = triggerFile(oldTrigger);
-      const newFile = triggerFile(trigger);
-      console.log(`Overriding '${trigger.id}' from ${oldFile} with ${newFile}.`);
-
-      this.triggers[idx] = trigger;
-      return;
+const handleTriggerOverride = <T extends { id?: string }>(triggers: Array<T>): Array<T> => {
+  console.log('handleTriggerOverride');
```
quisquous, line 213:
```suggestion
```

ui/raidboss/popup-text.ts
```diff
@@ -212,48 +209,40 @@ const textMap: TextMap = {
...
-// Helper for handling trigger overrides.
-//
-// asList will return a list of triggers in the same order as append was called, except:
-// If a later trigger has the same id as a previous trigger, it will replace the previous trigger
-// and appear in the same order that the previous trigger appeared.
-// e.g. a, b1, c, b2 (where b1 and b2 share the same id) yields [a, b2, c] as the final list.
-//
-// JavaScript dictionaries are *almost* ordered automatically as we would want,
-// but want to handle missing ids and integer ids (you shouldn't, but just in case).
-class OrderedTriggerList {
```
quisquous, line 224: What was wrong with OrderedTriggerList?

ui/raidboss/popup-text.ts
```diff
@@ -483,7 +468,10 @@ export class PopupText {
   protected me = '';
   protected job: Job = 'NONE';
   protected role: Role = 'none';
-  protected triggerSets: ProcessedTriggerSet[] = [];
+
+  protected collectedTriggerSets: LoadedTriggerSet[] = [];
+  // protected processedTriggerSets: ProcessedTriggerSet[] = [];
```
quisquous, line 473:
```suggestion
```

ui/raidboss/raidboss_config.ts
```diff
@@ -1416,7 +1416,7 @@ class RaidbossConfigurator {
   }
 }
 
-const flattenTimeline = (
+const pureFlattenTimeline = (
```
quisquous, line 1419: I'm not sure I understand these `pureFlattenTimeline`/`flattenTimelinePlace` (in place?) changes.  The pure function isn't very pure as it's modifying `set`.  What's the goal here?

ui/raidboss/popup-text.ts
trim21: Nothing, just it mix-up the localization step and override step, and not they are separated.

ui/raidboss/raidboss_config.ts
trim21: Good catch, I missed it still modifying set

ui/raidboss/raidboss_options.ts
```diff
@@ -50,7 +52,11 @@ type RaidbossNonConfigOptions = {
   PerTriggerAutoConfig: PerTriggerAutoConfig;
   PerTriggerOptions: PerTriggerOptions;
   PerZoneTimelineConfig: PerZoneTimelineConfig;
+
+  // loaded builtin triggers and user triggers
   Triggers: LooseTriggerSet[];
+  LoadedTriggers: LoadedTriggerSet[];
```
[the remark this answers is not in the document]
trim21: because we set `filename` in `raidboss_config.ts` https://github.com/quisquous/cactbot/blob/4fdec62f69fc076f030e68b6017045d4e46c8bb1/ui/raidboss/raidboss_config.ts#L1456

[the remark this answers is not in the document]
quisquous: What do you think about adding `filename?` to the trigger set type instead? Arguably it's already there, as you can do stuff like <https://github.com/quisquous/cactbot-user/blob/main/raidboss.js#L309>.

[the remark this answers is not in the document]
trim21: sounds a good idea, maybe seprated PR? This refactoring PR may be better to keep behavior not changed.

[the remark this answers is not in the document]
quisquous: I think it's less behavior change to do add that field there, since you don't have to expose extra things in `Options` to pass typed objects around.  Then you can remove that type entirely and just use `LooseTriggerSet`.  The field has arguably been there already, since the code adds it if it's not there.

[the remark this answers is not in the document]
trim21: Ok

ui/raidboss/popup-text.ts
```diff
@@ -817,58 +674,235 @@ export class PopupText {
...
+
+  LocalizeTrigger(
+    trigger: LooseTrigger,
+    set: LoadedTriggerSet,
+    triggerIndex: number,
+    langSuffix: string,
+  ): LocalizedTrigger | undefined {
+    if ('disabled' in trigger && trigger.disabled) {
+      return { id: trigger.id, filename: set.filename, disabled: true };
+    }
```
quisquous, line 807: It seems more complicated to add a disabled type to `LocalizedTrigger`.  What about returning `undefined` here instead, or some other way to skip disabled triggers before this is called?

trim21: disabled trigger may override other trigger so they need to be keeped.

and you are right, need a better aproach here.

quisquous: Oh, nevermind, these need to be included so an override trigger with disabled works.  Got it.

trim21:
Should we add legacy field like `regexDe` to loose trigger?

ui/raidboss/popup-text.ts
quisquous: I guess I don't really understand.  You could replace `ProcessedTrigger` with `LocalizedTrigger` and this code would still work.  Mostly I just know this existing code is working, and so I'd rather not replace it with something new, unless there's a reason to rewrite it.

ui/raidboss/raidboss_config.ts
```diff
@@ -1453,20 +1464,23 @@ const userFileHandler: UserFileCallback = (
   if (!baseOptions.Triggers)
     return;
 
+  baseOptions.Triggers ??= [];
+  baseOptions.LoadedTriggers ??= [];
```
quisquous, line 1468: I still think `LoadedTriggers` is an implementation detail and it doesn't make sense for it to be part of `Options`.  It's not something set by a user or the config.

If `filename` is part of `LooseTriggerSet`, then `Triggers` can just be updated in place as it used to be, right? That's what I was trying to suggest before.

ui/raidboss/raidboss_config.ts
```diff
@@ -1430,15 +1430,26 @@ const flattenTimeline = (
...
     console.log(`ERROR: '${filename}' specifies non-existent timeline file '${timelineFile}'.`);
     return;
   }
 
   // set.timeline is processed recursively.
-  set.timeline = [set.timeline, files[timelineFile]];
+  return [set.timeline, files[timelineFile]];
+};
+
+const flattenTimelinePlace = (
```
quisquous, line 1443: Can this function `delete set.timelineFile` then if you've removed it from the other function, since it's overriding the timeline?

ui/raidboss/popup-text.ts
```diff
@@ -69,14 +64,16 @@ export type ProcessedTrigger = LooseTrigger & {
...
+  & ({ localRegex: RegExp } | { localNetRegex: RegExp } | { disabled: true });
 
-type ProcessedTriggerSet = LooseTriggerSet & {
-  filename?: string;
-  timelineTriggers?: ProcessedTimelineTrigger[];
-  triggers?: ProcessedTrigger[];
+export type LoadedTriggerSet = LooseTriggerSet & {
+  filename: string;
+  timelineTriggers?: LooseTimelineTrigger[];
+  triggers?: LooseTrigger[];
```
quisquous, line 76:
```suggestion
```

Aren't these already defined in `LooseTriggerSet`? I'm not sure I understand why they need to be re-listed here.

quisquous:
> Should we add legacy field like regexDe to loose trigger?

Sure, sounds good!

ui/raidboss/raidboss_config.ts
trim21: This is for trying not to modified object in place 🤔

quisquous: In general, I think "don't modify things in place" is a good approach, but usually that means there needs to be a clear data pipeline or data should be in temporary scoped variables.  I think passing things through `Options` that don't live in `Options` is strictly worse.

(Also, this is already being modified in place on line 1336.)

Since there's already a lot going on here, what do you think about an approach of: (1) keep modifying things in place / keep `Options` as-is without new members (2) follow up to change how triggers are loaded here and maybe return different things from this function, make it better?

ui/raidboss/popup-text.ts
trim21: `LooseTriggerSet` now is `triggers?: (LooseTrigger & LegacyTrigger)[];`

and
```ts
export type LegacyTrigger = {
  regexEn?: RegExp;
  regexDe?: RegExp;
  regexFr?: RegExp;
  regexCn?: RegExp;
  regexJa?: RegExp;
  regexKo?: RegExp;

  // @deprecated
  netRegexEn?: RegExp;
  netRegexFr?: RegExp;
  netRegexDe?: RegExp;
  netRegexCn?: RegExp;
  netRegexJa?: RegExp;
  netRegexKo?: RegExp;
};
```

resources/languages.ts
```diff
@@ -4,6 +4,11 @@ export type Lang = typeof languages[number];
 
 export type NonEnLang = Exclude<Lang, 'en'>;
 
+/**
+ * @deprecated remove this after we don't support `netRegexCn` style triggers
+ */
+export type LegacyLangSuffix = 'En' | 'De' | 'Fr' | 'Ja' | 'Cn' | 'Ko';
```
quisquous, line 10: Where does this get used?

types/trigger.d.ts
```diff
@@ -200,6 +200,26 @@ export type TriggerSet<Data extends RaidbossData = RaidbossData> =
 // Less strict type for user triggers + built-in triggers, including deprecated fields.
 export type LooseTimelineTrigger = Partial<TimelineTrigger<RaidbossData>>;
 
+/**
+ *  @deprecated
+ */
+export type LegacyTrigger = {
```
quisquous, line 206: I think this could be combined into `LooseTrigger`.  `LooseTrigger` is the "user trigger, that could be missing fields, and could have old deprecated fields in it" type.  It's the same reason that `LooseTriggerSet` has `zoneRegex` in it.

ui/raidboss/raidboss_options.ts
```diff
@@ -50,7 +46,10 @@ type RaidbossNonConfigOptions = {
   PerTriggerAutoConfig: PerTriggerAutoConfig;
   PerTriggerOptions: PerTriggerOptions;
   PerZoneTimelineConfig: PerZoneTimelineConfig;
-  Triggers: LooseTriggerSet[];
+
+  // loaded builtin triggers and user triggers
```
quisquous, line 50:
```suggestion
  // user triggers
```

ui/raidboss/popup-text.ts
quisquous: In particular, I think it's a little confusing from a user output error to have the override console messages come in the opposite order that they are overridden.  If you have A B C triggers in order, this will print out C overrides B, and B overrides A, but in reality the load order is A B C.

resources/languages.ts
trim21: `regex${lang as LegacyLangSuffix}` `netRegex${lang as LegacyLangSuffix}`

quisquous: Sorry, I don't see what you're referring to.  Which file is that in?

trim21: it's used to get legacy  `regexDe`/`netRegexDe` property in `popup-text.ts`

quisquous:
```shell
$ git grep LegacyLangSuffix
resources/languages.ts:export type LegacyLangSuffix = 'En' | 'De' | 'Fr' | 'Ja' | 'Cn
' | 'Ko';
```

quisquous: Sorry for the confusion, I just don't see what you mean.  Did you mean to add these changes and they didn't save/commit?

trim21: that's strange, I remembered I used it in `popup-test.ts`, maybe I lost it by mistake.

PR #5051, opened 2022-11-03
raidboss: ASS 2nd trash fix

KiroKai:
Not sure why, but it seems like even though the mob name is Aqueduct Armor in-game it shows up as Sil'dihn Armor in logs (at least over multiple pulls i've noticed.)

[20:40:22.000] ChatLog 00:282B::The Sil'dihn armor readies Hells' Nebula. [20:40:22.743] StartsCasting 14:40014A89:Sil'dihn Armor:796C:Hells' Nebula:40014A89:Sil'dihn Armor:3.700:-53.12:-213.06:517.50:-1.56

[20:40:33.781] StartsCasting 14:40014A89:Sil'dihn Armor:796B:Infernal Weight:40014A89:Sil'dihn Armor:4.700:-63.80:-207.57:517.50:-0.77 [20:40:33.000] ChatLog 00:282B::The Sil'dihn armor readies Infernal Weight.

[20:40:39.908] StartsCasting 14:40014A89:Sil'dihn Armor:796A:Dominion Slash:40014A89:Sil'dihn Armor:3.200:-63.80:-207.57:517.50:-0.77 [20:40:39.000] ChatLog 00:282B::The Sil'dihn armor readies Dominion Slash.

PR #5052, opened 2022-11-03
raidboss: ASS(S) mob name change

xiashtra:
re: #5051 Mob name changed from Aqueduct Armor in 6.25 to Sil'dihn Armor in 6.28.

Translated names may need adjusting too?


quisquous:
Can you remove the other translations and mark it as `'missingTranslations': true` (or look up from data what it should be)? If you change it like this, nobody will know it's broken until they try the content and file a bug.  If you mark it as missing, then scripts will point out that somebody needs to look at it.

KiroKai:
Looked it up on fflogs and it seems like all the other languages got adjusted to Sil'dihn Armor as well.

quisquous:
@KiroKai has the new mob names in #5051

Issue #2584, opened 2021-03-14
Import remote triggers


trim21:
Do you have any thoughts on implementing this feature in cactbot itself?

I didn't know C# well so I'm planning to write a simple wrapper in C# and put most work in another language. 

My thoughts are like npm package, describe a remote repo with a json file, and a `files` filed to describe all content files (without glob support). Other thinga are just like a package manager, when updating, check new version, and replace local file content with new one.

MaikoTan:
1. Performance and Safety.
When developers upload triggers to cactbot this repo, we have CI to enforce the coding style, to ensure safety and performance.
But for a remote trigger, we aren't able to check it in runtime.
Expecially when you try to add triggers from multiple remotes, users who has not very enough knowledge dont't know what is good and what is bad.
I have seen several users who got a bunch of triggernometry triggers, without a zone filter or job filter, then cause ACT stucking. *"ACT sucks", they said.*

Another things that I concerned is safety.
Although JS in browser cannot directly access user' files, it can do some HTTP requests, modifying existing code behaviors, and even 
upload users' private information to somewhere that the user don't want.

1. Trigger executing confliction.
In most content, there are triggers prerequiste others triggers, if we allows multiple remotes, then a couple of conflict would happen.
For example, cactbot has a `"uwu trigger1"`trigger existing. There's a remote A trigger sets overrides the such trigger and set the new property `data.garuda = true`, then add a new custom trigger `"uwu custom A"` to read this `data.garuda` property and do something based on the boolean value. But there are another remote B trigger sets has also a `"uwu trigger1"` overriding, while it didn't set any new property.
Cactbot didn't know anything, it just runs the `"uwu trigger1"` from remote B, and then run the `"uwu custom A"` from remote A, and it gives the wrong answer.

1. Trigger overriding order
As described above, when applying multiple overrides, cactbot don't know which should be activated or which should be disabled, so cause the confliction and confusion.
*"Why can't I add remote triggers from Author Pandragon? There are only a few triggers executed."*

quisquous:
Here are some things that I think would be nice:

1) auto-find cactbot user directory

    This often trips users up, so being able to figure out where this is stored would be a huge win.  This may mean that such a plugin needs to be an OverlayPlugin plugin (like cactbot is) so it can use OverlayPlugin APIs to read the config files.

2) sync from git/github

[...]

PR #5066, opened 2022-11-12
i18n: TEA strategy for ko server

Bing-su:
There was a request to change the TEA trine strategy to fit the Korean server.
This is the only strategy used in the Korean server right now.

tea trine

<https://youtu.be/YXvQvjbVDRI?t=450>

```
g r y r g

y y r g r

r g G-G y
    | |
y g Y-R r

g y r y g
```


quisquous:
I had some requests for this in general too, but my feeling in general is to not have triggers have different instructions in different languages.  So, I don't think I want to merge this, sorry.

I think the options here are:
(1) tell people to disable the built-in trigger
(2) distribute a user file (something like <https://github.com/quisquous/cactbot-user/blob/main/raidboss/06-ew/p4s-fleeting_impulse_number_one.js> thaht can be dropped in `cactbot/user/raidboss/`) with the instructions
(3) have people override the output strings in the cactbot UI to tell them what to do

This is sort of in line with what #2584 wants, to make it easier to import different triggers from other people and would make this sort of situation better in the future.

Bing-su:
I see. thanks.

PR #5074, opened 2022-11-12
util: handle undefined params in make_timeline.ts

### LINKED ISSUE #5073: make_timeline shows help text even with valid parameters
Example:

```shell
$ ts-node util/logtools/make_timeline.ts -f Network*20221110.2*.log -lf 28
Error: Must specify exactly one of -f or -r
```

quisquous:
It seems that entirely unused params are `undefined`, and params without a value (e.g. `-f` on its own) are `null`.

I'm not sure when this changed, but the types are wrong in this file now in multiple places.

Fixes #5073.


quisquous:
I think there's more cleanup that can be done here, but this changes this script from "definitely not working in a number of cases" to "working for at least two files and one fflogs test", so I'm going to land it and iterate from there.

PR #5075, opened 2022-11-13
util: some cleanup of make_timeline.ts

quisquous:


util/logtools/make_timeline.ts
```diff
@@ -464,10 +450,11 @@ const assembleTimelineStrings = (
   let lastEntry: TimelineEntry = { time: lastAbilityTime.toString(), lineType: 'None' };
   if (fight !== undefined && fight.sealName !== undefined) {
     const zoneMessage = SFuncs.toProperCase(fight.sealName);
-    const tlString = `0.0 "--sync--" sync / 00:0839::${zoneMessage} will be sealed off/ window 0,1`;
+    const tlString =
+      `0.0 "--sync--" sync / 00:0839::${zoneMessage} will be sealed off/ window 100000,0`;
     assembled.push(tlString);
   } else {
-    assembled.push('0.0 "--sync--" sync /Engage!/ window 0,1');
+    assembled.push('0.0 "--sync--" sync /Engage!/ window 100000,0');
```
JLGarber, line 457: Are you 100% sure this is doing what you intend here? Or more specifically, can you be 100% sure that this will work correctly if someone pulls early? I'm not familiar enough with `timeline.ts` to know, but my gut feeling is that if combat is already started before the timeline starts, a sync of X,0 will fail.

util/logtools/make_timeline.ts
```diff
@@ -148,42 +136,44 @@ timelineParse.parser.addArgument(['--include_targetable', '-it'], {
...
-if (
-  typeof args.report_id === 'string' &&
-  (typeof args.report_fight !== 'number' || args.report_fight < 0)
-)
-  printHelpAndExit('Error: Must specify a report fight index of 0 or greater');
-if (args.fight_regex === '-1')
-  printHelpAndExit('Error: Timeline generation does not currently support -fr\n');
-if (args.zone_regex === '-1')
-  printHelpAndExit('Error: Timeline generation does not currently support -zr\n');
+const validateArgs = (args: ExtendedArgs): void => {
```
JLGarber, line 145: Oh, of course, sweeping all the validation into a function makes a lot more sense than keeping it in-line.

util/logtools/make_timeline.ts
```diff
@@ -22,10 +22,8 @@ import FFLogs, { ffLogsEventEntry } from './fflogs';
 
 // TODO: Add support for auto-generating loops.
 
-// TODO: Add FFLogs report parsing support.
-
-// TODO: Reinstate support for using raw start/end times
-// rather than force fight indexing.
+// TODO: We could consider re-adding `-s` and `-e` flags for manual start and end times,
+// but it would likely be better to just add support to `-lf` for any missing cases.
```
JLGarber, line 26: Yes, I mostly left that in for edge cases. Sometimes when Ravahn changes stuff across expansions, it's nice not to have to figure out exactly how to update the collection logic if we just want a timeline Right Now. But that's probably not t hat much of a concern going forward.

JLGarber (approved):

util/logtools/make_timeline.ts
quisquous: Well, two things.  One you're correct that this is wrong.  If the end of the window doesn't extend past zero, it won't work at all.

However also, if somebody pulls early (probably) we shouldn't sync at all, since if some other sync has happened that's going to be more reliable.

PR #5083, opened 2022-11-14
raidboss: initial triggers for the Palace of the Dead

xiashtra:
## Description

Adds general triggers for the Palace of the Dead deep dungeon.

## Additional Notes

Includes general triggers that are applicable to all PotD floor sets.  Lays out the file structure for all PotD floor sets, intended to be filled in by future individual pull requests per floor set.

It may be beneficial to generate a synthetic zoneID that captures all of the PotD floor sets for the purpose of the configuration UI.  Currently, the UI gets confused by the multiple zoneIDs for the general triggers, and looks like this:
potd_question


quisquous:
I think we don't need a synthetic zone id here.  I think this is a case where it'd make sense to add a feature where a triggerSet could define an optional custom (translatable) name to be displayed in the config UI.

In other words, add some new field to triggerSet (like `label?: LocaleText`) or something, and then update `raidboss_config.ts` to use this.  I think it should be pretty straightforward to add if you wanted to do that.

ui/raidboss/data/03-hw/deep_dungeon/the_palace_of_the_dead_general.ts
```diff
@@ -0,0 +1,184 @@
...
+      // 441 = HP Penalty (maximum HP reduced)
+      // 442 = Damage Down (damage dealt reduced)
+      // 443 = Haste (weaponskill recast, spell recast, and auto-attack delay reduced)
+      // 444 = Amnesia (can't use abilities)
+      // 445 = HP & MP Boost (maximum HP & MP increased)
+      // 446 = Item Penalty (can't use items or pomanders)
+      // 447 = Sprint Penalty (can't use Sprint)
+      // 448 = Knockback Penalty (knockback and draw-in effects for enemies and players disabled)
+      // 449 = Auto-heal Penalty (out-of-combat passive HP regeneration disabled)
+      // TODO: trigger might overlap when multiple enchantments exist on the same floor
```
JLGarber, line 151: For this, you might want to do a collect --> call --> cleanup construction. You can see a clean example in Sephirot EX/Unreal, starting at line 345. https://github.com/quisquous/cactbot/blob/0e6d23b401f82d48412f3b1f7232de4a7156b55d/ui/raidboss/data/06-ew/trial/sephirot-un.ts#L345-L390

xiashtra:
> I think we don't need a synthetic zone id here. I think this is a case where it'd make sense to add a feature where a triggerSet could define an optional custom (translatable) name to be displayed in the config UI.
> 
> In other words, add some new field to triggerSet (like `label?: LocaleText`) or something, and then update `raidboss_config.ts` to use this. I think it should be pretty straightforward to add if you wanted to do that.

Implemented in #5091.

ui/raidboss/data/03-hw/deepdungeon/the_palace_of_the_dead_general.ts
```diff
@@ -0,0 +1,184 @@
...
+        text: {
+          en: 'Mimic spawned!',
+        },
+      },
+    },
+    {
+      id: 'PotD General Mimic Infatuation',
+      // gives Accursed Pox (43F) if not interrupted
+      type: 'StartsUsing',
+      netRegex: { id: '18FD', source: 'Mimic' },
```
quisquous, line 77: Is there a different ability id for the uninterruptible versions on later floors?

quisquous (commented):
I have to admit, I'm a little not sure of the usefulness of some of these triggers.

I think Cairn of Passage seems good, since you really want to know when to get to the next level and don't want to miss that.  (You could even just put the 'Long' sound on it, since that's usually what gets used for "phase transitions".) Things that can be interrupted, especially pox, definitely.

The rest of it...I'm less sure about the usefulness of calling things out, e.g. hoard, enchantments, cairn of return, gloom, mimics.  I think these things are generally very obvious, and are often happening right after a transition where you have a brief second to take stock of what's going on (vs a trigger during combat where maybe you are preoccupied).  Especially if you're solo, you're not going to miss a mimic or gloom.  Am I thinking about this wrong, and you feel like they're more useful?

xiashtra:
> I have to admit, I'm a little not sure of the usefulness of some of these triggers.
> 
> I think Cairn of Passage seems good, since you really want to know when to get to the next level and don't want to miss that. (You could even just put the 'Long' sound on it, since that's usually what gets used for "phase transitions".) Things that can be interrupted, especially pox, definitely.
> 
> The rest of it...I'm less sure about the usefulness of calling things out, e.g. hoard, enchantments, cairn of return, gloom, mimics. I think these things are generally very obvious, and are often happening right after a transition where you have a brief second to take stock of what's going on (vs a trigger during combat where maybe you are preoccupied). Especially if you're solo, you're not going to miss a mimic or gloom. Am I thinking about this wrong, and you feel like they're more useful?

The floor enchantments/Gloom I agree are probably superfluous since the game does have on-screen text at the start of a floor to indicate them; I was mostly thinking of vision-impaired players that might prefer a TTS callout.

In my own experience, I like having something for duplicate pomanders since they otherwise only show as a system message in the chat box that's easily missed.  I find it makes playing with random party members not in a voice-call smoother.  Similar reasoning for calling mimic spawns; if you are solo obviously you'll notice it, but if you are playing with randoms they can end up spread out into different rooms and it helps to know when someone might need assistance, especially if they are playing a job that has no interrupt.

I'm not particularly attached to any of them so if it seems like too much I can remove the extraneous ones.  Cairn of Return in particular I mostly added for completeness but agree it's not something I'd really use in practice.

quisquous:
> In my own experience, I like having something for duplicate pomanders since they otherwise only show as a system message in the chat box that's easily missed. I find it makes playing with random party members not in a voice-call smoother. Similar reasoning for calling mimic spawns; if you are solo obviously you'll notice it, but if you are playing with randoms they can end up spread out into different rooms and it helps to know when someone might need assistance, especially if they are playing a job that has no interrupt.

That seems reasonable.  I am mostly playing solo in potd so didn't think about these.  Duplicate poms and mimics make sense then as wanting to know what other players are up to that you can't see.

> I'm not particularly attached to any of them so if it seems like too much I can remove the extraneous ones. Cairn of Return in particular I mostly added for completeness but agree it's not something I'd really use in practice.

Yeah, if you're not going to use it, then I think it makes sense to remove.

ui/raidboss/data/03-hw/deep_dungeon/the_palace_of_the_dead_general.ts
xiashtra: Thanks for the idea, but I ended up removing the enchantment triggers per quisquous's request.  I'll keep this pattern in mind for the future.

xiashtra:
I'm not sure why this PR thinks it's touching a bunch of extra files that haven't changed all of the sudden but it's not worth fighting with.  Closing this PR and opening a fresh one.

PR #5085, opened 2022-11-14
raidboss: Chi boss FATE triggers

xiashtra:
Partial triggers for the Chi boss FATE.


quisquous:
Here's a log with some missing ids for you:
[Ultima Thule_20221002_102752.931_3m.zip](https://github.com/quisquous/cactbot/files/10005496/Ultima.Thule_20221002_102752.931_3m.zip)

PR #5103, opened 2022-11-17
raidboss: initial triggers for the Palace of the Dead

xiashtra:
## Description

Adds general triggers for the Palace of the Dead deep dungeon.

## Additional Notes

Includes general triggers that are applicable to all PotD floor sets.

Continuation of #5083


Bing-su:
Thanks for nice triggers. I think pomanders should also be separated per language.

xiashtra:
> Thanks for nice triggers. I think pomanders should also be separated per language.

Updated per recommendation.

PR #5105, opened 2022-11-18
raidboss: more Chi triggers

xiashtra:


ui/raidboss/data/06-ew/hunts/ultima_thule.ts
```diff
@@ -259,6 +256,67 @@ const triggerSet: TriggerSet<Data> = {
...
+      condition: (data) => data.inCombat,
+      response: Responses.aoe(),
+    },
+    {
+      id: 'Hunt Chi Hellburner',
+      // tankbuster aoe around marked target (primary threat)
+      type: 'StartsUsing',
+      netRegex: { id: '6574', source: 'Chi' },
+      condition: (data) => data.inCombat,
+      alertText: (data, matches, output) => {
```
quisquous, line 272: I think this is just `response: Responses.tankCleave('alert'),`

xiashtra: This attack is less of a cleave and more of a circle around the targeted player (you can safely stand behind them as long as you are outside the radius of effect); is cleave appropriate in that instance (I've always associated cleave with a cone in the target's direction but perhaps I'm wrong on the intended meaning)?

quisquous: Ah yeah, I think of cleave as anything that hits more than one target and not a shape, e.g. https://www.fandomspot.com/ffxiv-cleave/.  There's not really a differentiation of "get away from the tank" and "get away from a conal attack pointed at the tank" since in most cases it's just "don't be in front with the tank".

PR #5109, opened 2022-11-19
raidboss: add ShB hunts

quisquous:


quisquous:
This needs a bit more (any) testing and cleanup.

PR #5116, opened 2022-11-22
raidboss: re-add territory replaced by 6.2 revamp

### LINKED ISSUE #5115: De-synced resources for pre-6.2 game due to the changes in 6.2
## Description

Due to the revamp of old dungeons and trials in 6.2, the resources update of 6.2 did an breaking changes for the game pre-6.2: https://github.com/quisquous/cactbot/commit/afb903eac9204ba8c41f1107b8226645ba9338dc

For example, the zone id of The Vault was changed to `1066` from `421`, which does not exist in the pre-6.2, causing the timeline and trigger made for it broken.

This breaking changes could be happened as well in the next game version (6.3), since SE does not complete the support of The Duty Support to all main scenario quest dungeons.

It's there have any way or plan to fix that? Like adding an old id or did an replacement for those values? (Also those files is auto-generated, the changes should be done in the tools generating it, but I'm unsure which tools of this project used).

## Additional information

There are also a question for those revamped dungeons: does the timeline and trigger needs update for the revamped version? If so, how to ensure its compatibility to older game version?

sandtechnology:
fix #5115, old territory id still needed for pre-6.2 game (cn/ko)


xiashtra:
Instead of duplicating the trigger files for the old zoneIds, you can just add them to the existing triggers as an array alongside the 6.2 IDs, along with a TODO to remove them once KO/CN catches up to 6.2.

sandtechnology:
> Instead of duplicating the trigger files for the old zoneIds, you can just add them to the existing triggers as an array alongside the 6.2 IDs, along with a TODO to remove them once KO/CN catches up to 6.2.

For why I want to duplicate those files is that the new version may having difference compare to old version, and when someone want to update it, it will become a pain for compatible with old version, so instead putting 6.1 and 6.2 zone id into the same trigger files, I think duplicating is the more comfortable solution.

quisquous:
I think there are already differences.  I updated the Sohm Al headmarkers in 5e84ec839 and so probably that needs to be reverted for the 6.1 version.  Do you want to undo that in the 6.1 files?

Agreed that if the files are the same we don't need to duplicate, but I suspect that all of these have changes that nobody's updated the files for, so it's probably best to duplicate up front and we'll just remove the duplicates when everything has updated.

sandtechnology:
> I think there are already differences. I updated the Sohm Al headmarkers in [5e84ec8](https://github.com/quisquous/cactbot/commit/5e84ec8393971ed30dbdbcdb21c7134f49b43f0a) and so probably that needs to be reverted for the 6.1 version. Do you want to undo that in the 6.1 files?
> 
> Agreed that if the files are the same we don't need to duplicate, but I suspect that all of these have changes that nobody's updated the files for, so it's probably best to duplicate up front and we'll just remove the duplicates when everything has updated. EDIT: Sorry, that's maybe ambiguously worded. I'm saying I agree that duplicating files is the right approach, since there are likely changes across game versions even if the files aren't updated yet.

Thanks for reminding! I just check those triggers and seems the Sohm AI headmarker is need to update even reverted (Due to the 6.0 headmarker changes? I don't know), so I update it and now this PR should be ready for review again.

quisquous:
Thank you for this!

PR #5124, opened 2022-11-25
plugin: CE opcode for ko 6.15

Bing-su:


quisquous:
The events that this opcode uses aren't used anymore in cactbot, and I don't think anybody else is using them.  The real opcodes live in OverlayPlugin now.  I'll go ahead and remove this event code too so there's less confusion in the future.

PR #5140, opened 2022-12-11
i18n: add palace timeline translations

Akurosia:


mooondark:
@Akurosia , I wonder why the 1st letter in not in capîtal in FR. Is there something in the script you use or maybe the source is wrong.

Is it possible to change this ?
Moon.

Akurosia:
this is how i extract it from saint coinach. As i am not familiar what and when to capitalize a word in french, there would be no way i can make changes that would propperly fit.

mooondark:
It's very easy, only the 1st letter of the sentence should be in capital.

quisquous:
For what it's worth, the `replaceSync` sections are not not case sensitive and are not seen by players.  So, in some senses it shouldn't matter what the capitalization is.

PR #5151, opened 2022-12-20
raidboss: include healers in DSR p6 rng buster callouts

GoldenLumia:
Something I've noticed during reclears is that the healers do not get a buster callout for the RNG busters that occur during Wyrmsbreath 1 and 2, even though it would benefit healers to know where to quickly use their single target mitigation so they have more time to get to the right spot. I did not include healers in the callout for tanks apart, since obviously the healers will have their own positions during those mechanics.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -2281,72 +2281,68 @@ const triggerSet: TriggerSet<Data> = {
       id: 'DSR Great Wyrmsbreath Hraesvelgr Not Glowing',
       type: 'StartsUsing',
       netRegex: { id: '6D34', source: 'Hraesvelgr', capture: false },
-      alertText: (data, _matches, output) => {
+      response: (data, _matches, output) => {
+        // cactbot-builtin-response
+        output.responseOutputStrings = {
+          tanksApart: {
+            en: 'Apart (Hrae buster)',
+            de: 'Auseinander (Hrae-buster)',
+            ja: '離れる (フレスから攻撃)',
+            cn: '分散 (圣龙死刑)',
+            ko: '떨어지기 (흐레스벨그 탱버)',
+          },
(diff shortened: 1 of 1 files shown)


quisquous:
Hmm, I guess I don't totally follow these changes.  It looks like the original code is already (intending?) to call out for non-tanks, e.g.

```typescript
        if (data.role === 'tank')
          return { alertText: output.tanksApart!() };
        return { infoText: output.hraesvelgrTankbuster!() };
```

For healers, it should say "Hrae Tankbuster" there, and for tanks it should say "Apart (Hrae buster)".

Is the issue that `DSR Great Wyrmsbreath Hraesvelgr Glowing` and `DSR Great Wyrmsbreath Nidhogg Glowing` have a condition on them? And the `condition: (data) => data.role === 'tank',` line on both of those triggers should be removed?

GoldenLumia:
I did not write the original code, but speaking from my experience playing healer in that fight I don't think I've ever seen that trigger fire in-game. 

>Is the issue that DSR Great Wyrmsbreath Hraesvelgr Glowing and DSR Great Wyrmsbreath Nidhogg Glowing have a condition on them? And the condition: (data) => data.role === 'tank', line on both of those triggers should be removed?

I was going to ask about this actually when I made my edits. Since that trigger is just passing data I was wondering if it needed a role restriction, but I couldn't tell for certain on that reading the associated docs. 

I can try removing that condition from the Glowing trigger and see if that returns it to expected behavior.

GoldenLumia:
Looks like removing those lines as suggested has returned the intended behavior from a healer POV in-game. Although I believe DPS will also see the trigger, it can help with what to feint/addle so I didn't have any plans to adjust the code further. 

I also was going to suggest adding a ${player} to the buster callout, but that's a bit outside the scope of this PR.

Thanks for the suggestion!

quisquous:
> I also was going to suggest adding a ${player} to the buster callout, but that's a bit outside the scope of this PR.

Feel free to send a follow-up PR if you'd like!

GoldenLumia:
>Feel free to send a follow-up PR if you'd like!

Sure! Will do.

PR #5152, opened 2022-12-20
util/raidboss: automate sildihn replacements

quisquous:


quisquous:
@JLGarber I didn't do this for older unreals, but I think this is something we could use in the future for keeping unreals in sync with each other

JLGarber:
Oh, nice. I like this. It should drastically cut down on the headache of "did I do the copypasta correctly?"

PR #5155, opened 2022-12-24
raidboss: Eden Normal timeline improvements

JLGarber:
Eden 7's change isn't that important, but it's an improvement, so I threw it in.

Eden 2 and 8 have had sync issues since before the start of Endwalker. Voidwalker's update should be self-explanatory. I'm still not sure where the Shiva sync issues come from, but on multiple occasions things have desynced by 3-5 seconds during the first phase. When I originally put this timeline together, player DPS was *far* too low to even think of breaking the timeline, so I never bothered to add safety sync windows.


quisquous:
Thanks for this! I haven't done any normal modes in a long while.

PR #5165, opened 2022-12-28
raidboss: ASS/S change teleport direction to final

quisquous:
This is a subjective change, but I think it's much more useful to know which direction the teleport is going to teleport you, rather than which direction it is pre-rotation.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
@@ -1699,10 +1699,10 @@ const triggerSet: TriggerSet<Data> = {
       // which should have been positioned to teleport across the laser to bait
       // the final ward away from other players
       //
-      // 1CD Blue (Counterclockwise) Teleporting East
-      // 1CE Orange (Clockwise) Teleporting West
-      // 1D2 Orange (Clockwise) Teleporting East
-      // 1D3 Blue (Counterclockwise) Teleporting West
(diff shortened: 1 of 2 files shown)


quisquous:
also cc @Legends0 here for opinions on this change

Legends0:
This works for me. I think it provides additional information without additional complexity.

xiashtra:
This is going to be subjective, but I expect some people will find this to be a confusing change because one of the more popular strategies (JP braindead) is taught by looking at which way your teleport starts pre-rotation (E or W) along with it's color (orange or blue), and not which way it will finish after rotation.

quisquous:
> This is going to be subjective, but I expect some people will find this to be a confusing change because one of the more popular strategies (JP Banish) is taught by looking at which way your teleport starts pre-rotation (E or W) along with it's color (orange or blue), and not which way it will finish after rotation.

Hmm!  How does this strategy work?  The way we did it was "oranges north, blues south" + "if post-rotation is pointing outside then you start on the far outside".

I'm glad I posted this for a conversation.

xiashtra:
> 

it's the same resolution, but instead of thinking where you will be pointing post-rotation, a lot of people only look at where they are pointing pre-rotation (so less mental pre-visualization).  Pre-rotation pointing West end up far N (orange) or far S (blue), and pre-rotation pointing East end up middle-N (orange) or middle-S (blue).  There are several guides that use this way of thinking instead of visualizing the post-rotation, here is one sample:
JPbraindeadWithVisuals

quisquous:
Okeydoke.  Maybe I'll just close this and not merge it then, especially if this is a known strat.  Thanks for the feedback.

PR #5167, opened 2022-12-29
raidboss: ASS/S simplify Rush of Might 1

quisquous:
This is subjective, but I find this style of instructions much easier to read and understand.

I asked a few people about this, and nobody is thinking about absolute north during this mechanic, and so "go east" when everybody is looking at the gladiators requires a lot more thinking.

I also understand that there is some ambiguity in "go left 1, right 2" (is it saying you need to go 2 the left or is the left one blinking 2 so you go 2 to the right, but hopefully the word "go" there makes it more obvious?)

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
@@ -8,7 +8,7 @@ import { callOverlayHandler } from '../../../../../resources/overlay_plugin_api'
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
-import { PluginCombatantState } from '../../../../../types/event';
+import { NetMatches } from '../../../../../types/net_matches';
 import { TriggerSet } from '../../../../../types/trigger';
 
 // TODO: Gladiator triggers for gold/silver location using OverlayPlugin?
@@ -28,8 +28,7 @@ export interface Data extends RaidbossData {
   soapCounter: number;
   beaterCounter: number;
   spreeCounter: number;
-  mightCasts: PluginCombatantState[];
-  mightDir?: string;
+  mightCasts: NetMatches['StartsUsing'][];
   hasLingering?: boolean;
   thunderousEchoPlayer?: string;
   gildedCounter: number;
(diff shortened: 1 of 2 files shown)


quisquous:
cc @Legends0 for wording opinions here; if other folks like the original better, I'm happy to just leave it as-is.  This worked for my brain a lot better.

Legends0:
I don't quite understand how the text output covers all the cases because It's always saying go left, which means either west for the first rush set or east for the second. I thought this mechanic can force the safe spot to be on the right instead of left.

The only other concern I have is that the 2/3 or 3/2 case here where the place to stand is not different enough in phrasing to differentiate it from the other cases. Perhaps the original phrasing had this problem of where to stand as well.

quisquous:
Yeah, the thought here is instead of trying to say precisely where to stand, it's to say how many notches to go down each line.  So "go 1 left 3 right" would mean on the line going front left to back right on the 3rd notch of that line.  In that case you'd be "on the right". You have to be able to form the intersection in your mind and form the intersection from that.

xiashtra:
I think these are over-specific to the point of information overload.  I maintain my thoughts from the previous discussion on this mechanic: https://github.com/quisquous/cactbot/pull/5005#issuecomment-1295919986

The only information needed is which side is dashing farther and how far.  Instead of "left [x] right [y]" it could be "left [z]", etc.  The only pattern that might need a more unique call would be the 2/3 combinations, but again all you really need to know is which side is dashing farther and that it's a 2/3 combo.

quisquous:
I think I disagree with "you only need to know which side is dashing further".  As you say, you have to know if it's a 2/3 or a 1/3.  And if you know that, then you know how far both sides are dashing, and not just the further one.  Just one side is not sufficient here.

Do you feel like "go right 1 left 3" is information overload as well?

xiashtra:
> I think I disagree with "you only need to know which side is dashing further". As you say, you have to know if it's a 2/3 or a 1/3. And if you know that, then you know how far both sides are dashing, and not just the further one. Just one side is not sufficient here.
> 
> Do you feel like "go right 1 left 3" is information overload as well?

I think a call that includes both the words "left" and "right" simultaneously opens up the possibility of confusion in which side a player should move towards, when it could give you the same info with "right behind third line", "left behind second line" etc.

At the very least the call should order the side you need to move towards immediately after the word "go".  As it is currently, it can call "go right 3 left 1" when you should be moving towards the left.

quisquous:
> At the very least the call should order the side you need to move towards immediately after the word "go". As it is currently, it can call "go right 3 left 1" when you should be moving towards the left.

Good idea.  This should be an easy fix to do.

quisquous:
@xiashtra does this look like what you're asking for?

quisquous:
Another option is that I could add all 6 possibilities and people could word them however they want.

quisquous:
Ok, there's all options, and better wording for on the line.

xiashtra:
> Another option is that I could add all 6 possibilities and people could word them however they want.

I think either of these works fine.  Having all 6 options allows for maximum user customization for those that want it.

ssilve1989:
Just throwing my thoughts in here that what was there previously was perfectly fine. I find the new defaults confusing to know if its inside/outside. However looks like I'll just have to customize out of the defaults. 

In general, I find cardinal directions better to use over relative directions because it doesn't assume the direction anyone is facing. Sure, most parties will be looking south for the second rush, but you by no means have to, especially with Legacy movement.

ibenny1011:
how do you customize back to the old one?

PR #5178, opened 2023-01-04
oopsy: more missing buffs

xiashtra:
I think this covers everything that was missing for all current jobs.
This file could use some re-formatting so abilities are sorted by job for easier updating in the future.

### DIFF
--- a/ui/oopsyraidsy/buff_map.ts
+++ b/ui/oopsyraidsy/buff_map.ts
@@ -38,7 +38,7 @@ export const missedEffectBuffMap: readonly MissableEffect[] = [
   {
     // PLD
     id: 'Divine Veil',
-    type: 'heal',
+    type: 'mitigation',
     effectId: '2D7',
     ignoreSelf: true,
     collectSeconds: 2,
@@ -109,9 +109,9 @@ export const missedAbilityBuffMap: readonly MissableAbility[] = [
     abilityId: '4057',
   },
   {
-    // WAR party shield + regen
+    // WAR
     id: 'Shake It Off',
-    type: 'heal',
+    type: 'mitigation',
     abilityId: '1CDC',
   },
   {
@@ -124,7 +124,7 @@ export const missedAbilityBuffMap: readonly MissableAbility[] = [
   {
     // DNC channeled shield
     id: 'Improvised Finish',
-    type: 'heal',
+    type: 'mitigation',
     abilityId: '64BD',
   },
   {
@@ -303,7 +303,7 @@ export const missedAbilityBuffMap: readonly MissableAbility[] = [
   {
     // SCH
     id: 'Succor',
-    type: 'heal',
+    type: 'mitigation',
     abilityId: 'BA',
   },
   {
@@ -315,7 +315,7 @@ export const missedAbilityBuffMap: readonly MissableAbility[] = [
   {
     // SCH
     id: 'Deployment Tactics',
-    type: 'heal',
+    type: 'mitigation',
     abilityId: 'E01',
   },
   {
@@ -333,7 +333,7 @@ export const missedAbilityBuffMap: readonly MissableAbility[] = [
   {
     // SCH
     id: 'Consolation',
-    type: 'heal',
+    type: 'mitigation',
     abilityId: '40A3',
   },
   {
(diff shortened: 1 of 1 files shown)


ui/oopsyraidsy/buff_map.ts
```diff
@@ -47,51 +50,136 @@ export const missedEffectBuffMap: readonly MissableEffect[] = [
...
+    effectId: '8AA',
+    collectSeconds: 45,
+  },
+  {
+    // BRD
+    id: 'The Wanderer\'s Minuet',
+    type: 'damage',
+    effectId: '8A8',
+    collectSeconds: 45,
+  },
```
quisquous, line 73:
```suggestion
```

I would skip these.  There used to be a comment about it but I think that comment got lost when these abilities moved into a different file.

My recollection is that they pulse every three seconds, and it's very noisy when people are being missed.  Collecting for 45 seconds is "did somebody get hit at all during that time period" which isn't right either.

ui/oopsyraidsy/buff_map.ts
```diff
@@ -47,51 +50,136 @@ export const missedEffectBuffMap: readonly MissableEffect[] = [
...
+    // DRK
     id: 'Dark Missionary',
     type: 'mitigation',
     abilityId: '4057',
   },
   {
+    // WAR party shield + regen
     id: 'Shake It Off',
-    type: 'mitigation',
+    type: 'heal',
```
quisquous, line 135:
```suggestion
    type: 'mitigation',
```

Yes it's a heal, but people use it as mitigation.

ui/oopsyraidsy/buff_map.ts
```diff
@@ -34,8 +36,9 @@ export const missedEffectBuffMap: readonly MissableEffect[] = [
     collectSeconds: 15,
   },
   {
+    // PLD
     id: 'Divine Veil',
-    type: 'mitigation',
+    type: 'heal',
```
quisquous, line 41:
```suggestion
    type: 'mitigation',
```

Yeah it's also a heal, but I think most people use it as mitigation.

ui/oopsyraidsy/buff_map.ts
```diff
@@ -100,249 +188,363 @@ export const missedAbilityBuffMap: readonly MissableAbility[] = [
...
+    // AST
+    id: 'Microcosmos',
+    type: 'heal',
+    abilityId: '6513',
+  },
+  {
+    // AST
+    id: 'Lady of Crowns',
+    type: 'heal',
+    abilityId: '',
```
quisquous:
```suggestion
    abilityId: '1D15',
```

quisquous (commented):
The comments about jobs are fine, but I wonder also if this could be better organized overall (in another PR).

I could imagine grouping jobs together like:
```typescript
const pld = {
  effects: [],
  abilities: [],
};
// etc

const missedEffects = [
  ...pld.effects,
  ...war.effects,
  // etc
];
const missedAbilities = [
  ...pld.abilities,
  ...war.abilities,
  // etc
];
```

ui/oopsyraidsy/buff_map.ts
xiashtra: It's a shield, it boosts effective HP.  Mitigations scale with incoming damage, heals and shields are always a fixed amount.

ui/oopsyraidsy/buff_map.ts
xiashtra: Same comment as above.

xiashtra: It might make sense to add `shield` as a `MissableBuffType` instead.

ui/oopsyraidsy/buff_map.ts
quisquous: Sure, but shields are mitigation too?  These types are just for icons and not anything else.  It's more about "what do you consider this ability", and I think divine veil and shake it off are more mitigation/shields and less heals even though they do both.

xiashtra: If you're considering shields and mitigation to be the same then the SCH and SGE abilities are very inconsistent right now.

quisquous: That's fair! I guess I'm thinking of heals as being "top up" and shields/mit as being "prevent damage".  I only ever made oopsy icons for heal and shield and didn't separate into some third category of percent mitigation vs shields.

PR #5186, opened 2023-01-10
plugin: remove FateWatcher and fate/CE events

quisquous:
This removes fates and CE events from cactbot.  These events have not been used in cactbot for a while,
and this functionality is now done through OverlayPlugin log lines.

This is a breaking change for any people who have written their own overlays using these events.


quisquous:
> "Breaking changes can use ! as a prefix to mark a breaking change, e.g. !plugin: change all the apis."

Mmhmm, sure lint script.

PR #5191, opened 2023-01-11
plugin: fix entity position/rotation

quisquous:


quisquous:
Oops, I thought I looked at this before the initial 6.3 release but apparently not.

PR #5194, opened 2023-01-14
raidboss: Preliminary Sophia Unreal timeline/triggers

JLGarber:
This is based on modifications to `sync_files` to parallel what we're doing with Criterion dungeons. Overall it seems to have been successful. I was able to get a small amount of testing in last night before maintenance, and most stuff worked.

Tilt detection is completely broken. I check ed my logs and it *looks* like the scale pan entities are in the correct spots, but for some reason they aren't being collected. I'll have to investigate what's going on there.

The portion of the timeline between the Scales of Wisdom cast and the first set of post-intermission clones is messy. I'll need to do more research to find the appropriate places to sync that. Regardless, after the first set of clones, everything seems to sync nicely.

Apart from those two issues, nothing else is broken, and more importantly I didn't see any actively *wrong* items. I think we're safe to merge this and tackle the remaining issues as we have time. (It might be good to have an aside in the patch notes specifically calling out the deficiencies.)


util/sync_files.ts
```diff
@@ -43,14 +43,23 @@ const zoneReplace: ZoneReplace[] = [
         'ui/oopsyraidsy/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts',
       'ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane.txt':
         'ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.txt',
+
+      'ui/raidboss/data/03-hw/trial/sophia-ex.ts': 'ui/raidboss/data/06-ew/trial/sophia-un.ts',
```
quisquous, line 47: Would you mind adding an extra entry to the `zoneReplace` array rather than adding it into the criterion block directly?

ui/raidboss/data/06-ew/trial/sophia-un.ts
```diff
@@ -0,0 +1,719 @@
...
+      // 3: (-55.0637, 3.523648)
+      // 4: (54.9907, 3.387837)
+      // 5: (54.98699, 9.5779B3)
+      // 6: (54.9907, -3.50686)
+      // 7: (54.99068, -10.14043)
+
+      // Because of this, we need only see one entity use a 21 log line and we can find the rest.
+      id: 'SophiaUN Quasar Setup',
+      type: 'Ability',
+      netRegex: { id: '19A[89]' },
```
quisquous, line 410: This is probably the issue, if I had to guess? This line looks unchanged.

JLGarber: Looks like it, yes. Definitely a good reason to avoid the compact form in the future!

JLGarber:
So it looks like one of the Criterion replacements just happens to match one of the position lines in a comment. I think it's probably fine to leave it as it is for now, and when we fix the tool to avoid this happening in future it will fix itself automagically.

quisquous:
> So it looks like one of the Criterion replacements just happens to match one of the position lines in a comment

Oh? I'm not sure what you're referring to here.

quisquous:
> > So it looks like one of the Criterion replacements just happens to match one of the position lines in a comment
> 
> Oh? I'm not sure what you're referring to here.

Oh! I see, got it.  /o\

PR #5195, opened 2023-01-14
raidboss/oopsy: add Euphrosyne

quisquous:


quisquous:
still a work in progress

JLGarber (approved):

xiashtra:
Nophica: Heavens' Earth (7C22 on boss, 7C23 on targeted players) - three tankbuster aoes, one on each party

quisquous:
Definitely a lot of TODOs still.  The timing on the tetrapagos was bad so I removed that until I can run and test again (it could probably be worded better too).  Hydrostasis also has bad positions the second time around and so probably needs getCombatants.  Also, collecting all the Menphina ids is still a work in progress, and maybe once we have them it'll be easier to do something about all the half room cleaves.

PR #5201, opened 2023-01-15
resources: Update content_list.ts for 6.3

xiashtra:


resources/content_list.ts
```diff
@@ -133,10 +134,10 @@ const contentList: (ZoneIdType)[] = [
   ZoneId.TheBowlOfEmbers,
   ZoneId.TheNavel,
   ZoneId.TheHowlingEye,
-  ZoneId.CapeWestwind,
+  ZoneId.CapeWestwind, // changed to solo duty in 6.1
```
quisquous, line 137: I think we should remove it from here, as in general there's not support for solo duties.

xiashtra: I wasn't sure where CN and KO are and if they need those to stick around for a bit.

quisquous: Ah, ok.  That's fair.  We can remove them eventually, then.

PR #5207, opened 2023-01-16
raidboss: Euphrosyne Matron's Harvest additional ID

xiashtra:
Nophica's Matron's Harvest raidwide has a second ID that is used second and sometimes third in the fight.


quisquous:
Thank you!

PR #5209, opened 2023-01-17
raidboss: enhancement for p8s orogenic deformation

### LINKED ISSUE #5141: P8S Orogenic Deformation - Add name of other Support/DPS player that is also Hit/Not Hit
During the Dominion mechanic in P8Sp2, the Orogenic Deformation spread AOE always hits 2 Support and 2 DPS, which gives them a debuff, resulting in them needing to take the second set of towers. Currently the on-screen callout text for this is "Second Towers" while the players that did not get hit receive "First Towers."

The most common strategy for this mechanic is to assign towers to pairs of players, most commonly role pairs, then assigning towers to each, such as THMR. If both melee players are hit, for instance, one of them would need to flex and take the ranged tower. Flexers are typically assigned at the start of the encounter.

It would be helpful if the trigger could identify and callout the other Support/DPS player that has the matching debuff/no-debuff, much like the Snake Debuff callouts in P8Sp1. Would it be feasible for this to be included?

Thank you

wexxlee:
Closes #5141.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p8s.ts
+++ b/ui/raidboss/data/06-ew/raid/p8s.ts
@@ -66,7 +66,10 @@ export interface Data extends RaidbossData {
   myTower?: number;
   flareCounter: number;
   inverseMagics: { [name: string]: boolean };
-  deformationTargets: string[];
+  deformationHit: string[];
+  deformationNotHit: string[];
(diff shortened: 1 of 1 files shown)


quisquous:
Thank you!

PR #5214, opened 2023-01-19
raidboss: add missing JP translations

kshman:
Always thanks a lot!


ui/raidboss/data/06-ew/trial/sophia-un.ts
```diff
@@ -111,7 +111,7 @@ const triggerSet: TriggerSet<Data> = {
           en: 'Avoid Dash Attack',
           de: 'Ansturm-Angriff ausweichen',
           fr: 'Évitez l\'attaque Charge',
-          ja: '突進に避け',
+          ja: '突進回避',
```
quisquous, line 114: If you are going to change this, can you change sophia-ex.ts instead and run `ts-node util/sync_files.ts` so that these two files stay in sync?

quisquous:
It looks like #5216 merged in your changes because they were both touching the same files (and then did run sync_files).  I think this change should be included now.  Let me know if there was something missing.

kshman:
Yes, this merged into #5216. I checked
So I'll close this PR.

Very thanks!  @quisquous & @Bing-su

PR #5216, opened 2023-01-19
i18n: add missing ko translations

Bing-su:
ASS/S is out of sync, so I ran `util/sync_files.ts`. Please tell me if it was intentional.


quisquous:
Yeah, thank you for running sync_files.  I really should make a test that verifies that it creates no changes.

PR #5223, opened 2023-01-21
[wip] raidboss/oopsy: timeline + triggers for Rubicante

wexxlee:
Going to spend some time tomorrow working on the map effects stuff with Purgation, so marking this as draft for now.   However, the rest should be good to go.

### DIFF
--- a/ui/oopsyraidsy/data/06-ew/trial/rubicante.ts
+++ b/ui/oopsyraidsy/data/06-ew/trial/rubicante.ts
@@ -1,0 +1,39 @@
+import ZoneId from '../../../../../resources/zone_id';
+import { OopsyData } from '../../../../../types/data';
+import { OopsyTriggerSet } from '../../../../../types/oopsy';
+
+export type Data = OopsyData;
+
+const triggerSet: OopsyTriggerSet<Data> = {
+  zoneId: ZoneId.MountOrdeals,
+  damageWarn: {
+    'Rubicante Circle of Purgatory Fiery Expiation': '7CC5', // conal from outer ring
+    'Rubicante Arch Inferno': '7CC8', // center puddle around boss
+    'Rubicante Inferno Devil': '7CCA', // 2x large rotating circle AoEs during Arch Inferno
+    'Rubicante Hellscathe': '7CCF', // small circle AoEs baited on players during Arch Inferno
+    'Rubicante Conflagration': '7CCB', // middle danger bacon during Arch Inferno
+    'Rubicante Radial Flagration': '7CCD', // large fan-style AoE during Arch Inferno
+    'Rubicante Flamesent Ghastly Wind': '7CD6', // targeted conal (adds phase)
+    'Rubicante Flamesent Ghastly Flame 1': '7CD7', // ground circle A0Es (adds phase)
+    'Rubicante Flamesent Ghastly Flame 2': '7CD8', // ground circle A0Es (adds phase)
+    'Rubicante Flamerake 1': '7CDF', // first of three line AoEs (center)
+    'Rubicante Flamerake 2': '7CE0', // second (pair) of three line AoEs (sides of boss)
+    'Rubicante Flamerake 3': '7CE1', // third (pair) of three line AoEs (edges)
+    'Rubicante Scalding Fleet': '7CE9', // baited proteans
+    'Rubicante Sweeping Immolation': '7CE3', // 180 front cleave
+  },
+  shareWarn: {
+    'Rubicante Explosive Pyre': '7CDB', // tank autos w/puddle cleave
+  },
+  shareFail: {
+    'Rubicante Shattering Heat': '7CEB', // solo tankbuster
+    'Rubicante Flamesent Shattering Heat': '7CD5', // solo tankbuster (adds phase)
+    'Rubicante Inferno (spread)': '7CDA', // spread AoE puddles (post-adds)
+    'Rubicante Dualfire': '7D96', // 2x wide cleaving tankbusters
+  },
+  soloFail: {
+    'Rubicante Total Immolation': '7CE5', // party stack during Sweeping Immolation
+  },
+};
+
+export default triggerSet;
--- a/ui/raidboss/data/06-ew/trial/rubicante.ts
+++ b/ui/raidboss/data/06-ew/trial/rubicante.ts
@@ -3,31 +3,42 @@ import Outputs from '../../../../../resources/outputs';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
-import { NetMatches } from '../../../../../types/net_matches';
 import { TriggerSet } from '../../../../../types/trigger';
 
 /**
-Slots:
-00 = Arena fiery or not
-01 = Inner circle
+Map Effects
+
+Locations/Slots:
+00 = Arena visual
 02 = Middle ring
-03 = Outer ring
-04 = Flamespire brand indicator?
+04 = Flamerake "compass"
+05-0C = Cardinals/intercardinals [N,NE...NW] - *starting* direction of Ordeal of Purgation
 
 00 flags:
 00020001 = Fiery
-00080004 = Not fiery
+00080004 = Clear fiery
+00200010 = Rocky terrain after boss death
 
-01/02/03 flags:
+02 flags:
 00020001 = Arrows rotating CW
 00080004 = Clear CW arrows
 00200010 = Arrows rotating CCW
 00400004 = Clear CCW arrows
 
 04 flags:
-00010001 = cardinals safe?
-00200020 = intercards safe?
+00100010 = starts N/S, cleaves NW/SE
+01000100 = starts N/S, cleaves NW/SE (same as above? needs investigation?)
+00020002 = starts N/S, cleaves E/W
 00080004 = clear indicator
+
+05-0C flags:
+00010004 = initial telegraph - straight line in given direction
+00020004 = initial telegraph - CW jagged line (turns 1 dir CW from starting direction) (only used for Purg 2)
+00800004 = clear telegraph of fire line
+00100004 = fire pulse - straight line in given direction
+00200004 = fire pulse - CW jagged line (turns 1 dir CW from starting direction) (paired with CCW arrows)
+00400004 = fire pulse - CCW jagged line (turns 1 dir CCW from starting direction) (paired with CW arrows)
+
  */
 
 // 7CD4 Ghastly Torch during add phase *is* an aoe but is constant and small, so skipped.
@@ -37,17 +48,8 @@ Slots:
 //  - Flamerake - identify initial direction of line cleave?
 
 export interface Data extends RaidbossData {
-  decOffset?: number;
   dualfireTarget: boolean;
 }
-
-// First headmarker is Shattering Heat (tankbuster on MT)
-const firstHeadmarker = parseInt('0156', 16);
-const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker']) => {
-  if (typeof data.decOffset === 'undefined')
-    data.decOffset = parseInt(matches.id, 16) - firstHeadmarker;
-  return (parseInt(matches.id, 16) - data.decOffset).toString(16).toUpperCase().padStart(4, '0');
-};
 
 const triggerSet: TriggerSet<Data> = {
   zoneId: ZoneId.MountOrdeals,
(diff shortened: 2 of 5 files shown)


ui/raidboss/data/06-ew/trial/rubicante.ts
```diff
@@ -0,0 +1,284 @@
...
+//  - Flamerake - identify initial direction of line cleave?
+
+export interface Data extends RaidbossData {
+  decOffset?: number;
+  dualfireTarget: boolean;
+}
+
+// First headmarker is Shattering Heat (tankbuster on MT)
+const firstHeadmarker = parseInt('0156', 16);
+const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker']) => {
```
quisquous, line 46: Does normal mode use headmarker offsets? (In other words, is the first headmarker always literally `0156`?)  If no, can you just use the headmarker ids directly?

Mostly I ask because I don't think there's ever been a normal mode that used headmarker offsets.

ui/raidboss/data/06-ew/trial/rubicante.txt
```diff
@@ -0,0 +1,127 @@
...
+201.6 "--sync--" sync / 1[56]:[^:]*:Rubicante:7CBC:/ # boss reposition at edge
+
+# Phase 2 (Flamerake & Sweeping Immolation)
+986.4 "--sync--"  sync / 14:[^:]*:Rubicante:7CD2:/ window 1000,0
+986.4 "--targetable--"
+1000.0 "Blazing Rapture" sync / 1[56]:[^:]*:Rubicante:7CD2:/
+1013.4 "--untargetable--" # cutscene
+1036.4 "--targetable--"
+1048.5 "Inferno (spread)" sync / 1[56]:[^:]*:Rubicante:7CDA:/
+1062.2 "Flamerake (indicator)" sync / 1[56]:[^:]*:Rubicante:7CDD:/
```
quisquous, line 58:
```suggestion
1062.2 "--sync--" sync / 1[56]:[^:]*:Rubicante:7CDD:/
```

You can take this suggestion or leave it, but given that there's already 1 2 3 in the timeline, maybe the indicator doesn't need to be listed too unless it's a castbar somebody would mit.

ui/raidboss/data/06-ew/trial/rubicante.txt
```diff
@@ -0,0 +1,127 @@
...
+163.1 "Ordeal of Purgation" sync / 1[56]:[^:]*:Rubicante:80E8:/ # middle-disc spin, jagged line (from this point forward)
+171.1 "Fiery Expiation" sync / 1[56]:[^:]*:Circle of Purgatory:7CC5:/
+187.5 "Ordeal of Purgation" sync / 1[56]:[^:]*:Rubicante:80E8:/
+195.5 "Fiery Expiation" sync / 1[56]:[^:]*:Circle of Purgatory:7CC5:/
+201.6 "--untargetable--"
+
+# Adds Phase
+# The adds will intermittently cast various AoEs.
+# All are ground-telegraphed except 7CD4 (Ghastly Torch), a small raid-wide.
+# Because variation in add deaths affects timing, don't sync to them or display timeline.
```
quisquous, line 48: Is this true for normal mode? I think it'd be worth at least listing the first round or two like we do in extreme.

ui/raidboss/data/06-ew/trial/rubicante.txt
```diff
@@ -0,0 +1,127 @@
...
+1286.8 "Soulscald" sync / 1[56]:[^:]*:Rubicante:7CE6:/
+1292.9 "Scalding Fleet" #sync / 1[56]:[^:]*:Rubicante Mirage:7CE9:/
+1299.1 "--sync--" sync / 1[56]:[^:]*:Rubicante:8034:/
+1304.1 "Dualfire" sync / 1[56]:[^:]*:Rubicante:7D95:/
+1318.9 "Sweeping Immolation" sync / 1[56]:[^:]*:Rubicante:7CE3:/
+1319.3 "Total Immolation" sync / 1[56]:[^:]*:Rubicante:7CE5:/ window 30,10000 jump 1114.5
+# pretty sure this is the end of the loop, but didn't go far enough to see it loop around again
+
+# Fake loop
+1332.3 "--middle--"
```
quisquous, line 101: Can you only add ~30s of fake loop abilities?

ui/raidboss/data/06-ew/trial/rubicante.ts
```diff
@@ -0,0 +1,284 @@
...
+      netRegex: { id: '7CDA', source: 'Rubicante' },
+      condition: Conditions.targetIsYou(),
+      response: Responses.spread(),
+    },
+    {
+      id: 'Rubicante Flamerake',
+      type: 'StartsUsing',
+      netRegex: { id: '7CDD', source: 'Rubicante', capture: false },
+      // Spinning compass animation is 6 seconds; damage goes out 2.6 seconds later
+      delaySeconds: 3,
```
quisquous, line 113: What about making this trigger happen on the damage for Flamerake 1 (with a suppress) rather than try to guess the timing? (This is my usual strategy, so that when the call happens it is always safe to move, and never too early.)

ui/raidboss/data/06-ew/trial/rubicante.ts
```diff
@@ -0,0 +1,284 @@
...
+      infoText: (_data, _matches, output) => output.avoid!(),
+      outputStrings: {
+        avoid: {
+          en: 'Avoid line cleave, then in',
```
quisquous, line 117: I think this could just be `Outputs.in`

ui/raidboss/data/06-ew/trial/rubicante.ts
```diff
@@ -0,0 +1,284 @@
...
+    },
+    {
+      id: 'Rubicante Soulscald',
+      type: 'Ability',
+      netRegex: { id: '7CE6', source: 'Rubicante', capture: false },
+      delaySeconds: 3,
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Bait protean, then dodge',
```
quisquous, line 129: I'm not sure this necessarily needs a trigger, because there's ground markers for this that show up with plenty of time to dodge.  If you wanted to have a prompt to get out of the way, maybe just a `moveAway` when the initial damage hits, like this similar mechanic? https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/trial/hades-ex.ts#L61-L67

quisquous (commented):
A few small comments.  Thank you SO much for this.

I'd be happy to land this without the ordeals changes and have that as a follow-up.

ui/oopsyraidsy/data/06-ew/trial/rubicante.ts
```diff
@@ -0,0 +1,39 @@
...
+    'Rubicante Flamesent Ghastly Flame 1': '7CD7', // ground circle A0Es (adds phase)
+    'Rubicante Flamesent Ghastly Flame 2': '7CD8', // ground circle A0Es (adds phase)
+    'Rubicante Flamerake 1': '7CDF', // first of three line AoEs (center)
+    'Rubicante Flamerake 2': '7CE0', // second (pair) of three line AoEs (sides of boss)
+    'Rubicante Flamerake 3': '7CE1', // third (pair) of three line AoEs (edges)
+    'Rubicante Scalding Fleet': '7CE9', // baited proteans
+    'Rubicante Sweeping Immolation': '7CE3', // 180 front cleave
+  },
+  shareWarn: {
+    'Rubicante Explosive Pyre': '7CDB', // tank autos w/puddle cleave
```
quisquous, line 26: Is this an add phase ability? Or when does this happen? I've seen the flying text for this in extreme, but could not tell what it was.

ui/raidboss/data/06-ew/trial/rubicante.ts
wexxlee: > I'd be happy to land this without the ordeals changes and have that as a follow-up.

Sounds good.  I just finished mapping (pun intended?) the rest of the Ordeals map effects - it's a little different for normal mode - and I want to think on a couple of things.  So I'll make changes/address your comments below and mark as ready, and do a follow-up PR once I've landed on an approach for the Ordeals.

ui/raidboss/data/06-ew/trial/rubicante.ts
wexxlee: The mech is a little different in normal mode, so I don't think that would give sufficient warning.  Basically, the timing is:

0.0 Starts casting `7CDD` on self (this corresponds to the compass appearing and spinning)
6.0 Uses `7CDD` on self (corresponds to compass disappearing)
6.8 Starts casting `7CDE` on self (short silent cast)
7.3 Uses `7CDE` on self (corresponds to a flash telegraph of a line cleave through mid - orientation depends on compass)
8.6 Uses `7CDF` (aoe damage applied for flash telegraph loc)
9.6 Uses `7CE0` (aoe damage applied for lines on either side moving outward from original loc)
10.6 Uses `7CE1` (aoe damage applied for lines further out moving outward/edge of arena)

If we triggered off the 21 line (instead of the 20 line) for `7CDD`, it would give only 2.6 seconds to dodge the initial damage, which feels pretty tight - and keying off any later abilityId reduces that further.  

Another option though might be to just eliminate delaySeconds and do an 8-second duration, which would leave the warning up until right before the initial damage goes out?

ui/raidboss/data/06-ew/trial/rubicante.ts
wexxlee: See note above - I think this part of the same conversation.  We could just do this, but the timing would be super-tight because we'd want it to display once it was safe to move in (after the snapshot for the `7CDF` damage and before the `7CE0` damage, which is only 1.0 seconds apart.

wexxlee:
> I'd be happy to land this without the ordeals changes and have that as a follow-up.

Sounds good. I just finished mapping (pun intended?) the rest of the Ordeals map effects - it's a little different for normal mode - and I want to think on a couple of things. So I'll make changes/address your comments below and mark as ready, and do a follow-up PR once I've landed on an approach for the Ordeals.

ui/raidboss/data/06-ew/trial/rubicante.ts
wexxlee: Nope, unnecessary code - we can just use the headmarkerId directly.

ui/oopsyraidsy/data/06-ew/trial/rubicante.ts
wexxlee: Post adds-phase, he becomes "empowered" with the blue wing (similar to P3 in Ex), and Explosive Pyre is his tank auto that does a small puddle cleave around the target.  (Somewhat annoyingly, he uses it only a second or two after total immolation resolves when the party is stacked up... 😆)

ui/raidboss/data/06-ew/trial/rubicante.txt
wexxlee: Looking more closely at this, I think we can probably display the first ~11s of add casts in the timeline.  After that, looking through my logs, things started to get wildly divergent depending on dps, whether the party burned the Greater Flamesent first, etc.  I'll add that in.

ui/oopsyraidsy/data/06-ew/trial/rubicante.ts
quisquous: Should this not be ignored in the timeline, then?

ui/raidboss/data/06-ew/trial/rubicante.ts
quisquous: Ok, thanks for the explanation.  I think this is fine as is!

ui/oopsyraidsy/data/06-ew/trial/rubicante.ts
wexxlee: He uses it pretty often, and it's not a significant amount of damage.  This is what would be added to Phase 2:

```
2050.6 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2068.5 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2071.6 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2084.9 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2087.9 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2091.0 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2102.2 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2117.0 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
```

And this would get added to the loop:

```
2262.1 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2265.2 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2274.3 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2289.6 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2292.6 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2295.7 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
2306.9 "Explosive Pyre" sync / 1[56]:[^:]*:Rubicante:7CDB:/
```

If you don't think that's too spammy, happy to add it in!

quisquous: haha oh gosh, nevermind

PR #5224, opened 2023-01-22
raidboss: add Menphina Midnight Frost / Waxing Claw

quisquous:


quisquous:
@JLGarber @Legends0 idk if y'all are running Euphrosyne on any sort of regular basis, but if you do could you apply this patch and tell me if calls feel right for the "unattached dog" case (and attach any console output if wrong?)? I think the logic is correct (based on some debugging) but I think the dog position isn't correct always? I don't think I have the time to run this another million times, so I will just kinda keep casually debugging this as I go.

Legends0:
I tested this out... it did not call out the wolf's left/right during the slide in mechanic and did not call correct spot for the wolf cleave after... It stated front right, although it was front left. It worked for the first half/works enough i guess.

quisquous:
Belatedly, I can't sort out what's going on with Playful Orbit here, so I just left comments.

PR #5235, opened 2023-01-24
raidboss: add TOP p1 debuff triggers

quisquous:


quisquous:
Just gonna land this, and will fix up any mistakes / revert after some testing in game.

PR #5241, opened 2023-01-26
raidboss: Fix Euphrosyne boss 3 stack + spread call

JLGarber:
The condition for the Rondel trigger can potentially run before `haloneIceDartTargets` is fully populated, leading to a situation where a player is simultaneously instructed to spread and stack. Because `condition` evaluates prior to `delaySeconds`, we have to move the condition in-line and return if it's met.


quisquous:
Thank you!!

PR #5245, opened 2023-01-26
raidboss: add TOP p2 eye location

quisquous:


quisquous:
In retrospect, I'm not sure how useful this trigger is.  I think it's hard to find a good spot for it (as early party synergy is "far/mid symbol with who + in/out) and by the time you're dodging you really need to have a good idea where the eye is already.

Legends0:
> In retrospect, I'm not sure how useful this trigger is. I think it's hard to find a good spot for it (as early party synergy is "far/mid symbol with who + in/out) and by the time you're dodging you really need to have a good idea where the eye is already.

I have been testing this out, honestly so much is going on and the eye is so massive that I haven't paid any attention to this callout.

Legends0:
I have been running with delaySeconds of 5. This calls out as the ground aoes appear and seemed easier to notice than 6. 7 feels too late.

Alternative to the eye position it could call out the perpendicular safe spots, but the shorter string for eye is good.

Legends0:
I think this may need a suppressSeconds. It is calling it out when the eye moves for cosmic memory.

quisquous:
Added more delay and a suppress and will go ahead and land this

PR #5246, opened 2023-01-27
util: fix when the log time is 00

tssailzz8:


quisquous:
Could you describe the bug that you're fixing here a little bit more? I'm not sure I understand the situation.

tssailzz8:
When the battle time is in 00 o'clock, it will be converted to “24:” but the log time is ”00:”

valarnin:
Huh, that's a fun node bug. Has it been reported upstream?

image

quisquous:
Ah, good find, valarnin!  Sorry for not looking at this PR earlier.  I've been busy with ultimate and wanted to look if there was a better way to do this, but valarnin beat me to it.  Can we use that instead?

PR #5258, opened 2023-02-04
raidboss: add p4-p6 initial timeline

Legends0:
Using fflogs logs from Kindred and a few other public logs that have enrage spells and timings to generate an initial timeline for p4-p6.

Patch timings likely need to be updated. Targetable/untargetable were manually added from first hit in logs, so will need to be adjusted.

There are multiple spells with names Wave Cannon, Cosmo Meteor, and Cosmo Arrow spells in p6. I clarified some of the spells in the timeline. Ignoring a spell that's named Inhale (7E51) and the propagating Wave Cannon Exaflares (7BAE).

I was not able to test this.


quisquous:
I think there's a p5 log in discord too, if you need a real log file.

quisquous:
Thanks again for this.  Been busy, will try to take a look soon.

PR #5260, opened 2023-02-09
Add instant reminder for P6S Transmission

super-dough:
In some "full uptime strats" like https://youtu.be/03kPYiaWBV8?t=109, players are required to do pre-position according to the buff they gain. It would be useful if cactbot can remind player instantly whether their buff is forwardCleave or backwardCleave.

### DIFF
--- a/06-ew/raid/p6s.js
+++ b/06-ew/raid/p6s.js
@@ -998,6 +998,35 @@ Options.Triggers.push({
       },
     },
     {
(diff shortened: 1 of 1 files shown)


quisquous:
This is great, but can you merge this into the `main` branch and not into the `triggers` branch? The triggers branch is just a "read only" javascript version for people to reference but the main branch in typescript is the real one.

PR #5262, opened 2023-02-09
raidboss/oopsy: adjust TOP p4 timeline

quisquous:
Some adjustments to #5258

* clean up self-targeted / damage lines to only include self-targeted
* use letters for wave cannon to avoid two things with numbers

I thought about numbering the three stacks as well? But that seems like too many numbered things.


Legends0:
I think it looks good without numbering the stacks.

PR #5263, opened 2023-02-09
raidboss: add monitors/p4 triggers

quisquous:


quisquous:
Some untested triggers

quisquous:
Monitors looked good in game, and p4 stacks looked fine in raid emulator.

PR #5264, opened 2023-02-09
raidboss: p3 add some rot triggers

Legends0:
Adds color tower to go to when they spawn for players with the rot debuffs
Adds Pass rot and Get rot callouts timed on Tower Abilities
Adds Spread callout for effect after soaking a tower (delay could maybe use some adjusting?)

The first pass appears to happen later in the duration than the others, ~~this was based off some kill logs I looked at for timings truncated down to nearest second.~~ **EDIT:** Changed to use ~~Latent Defect~~ Tower Abilities .


ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -772,6 +776,69 @@ const triggerSet: TriggerSet<Data> = {
...
+      run: (data) => {
+        data.bugRot = {};
+        data.latentDefectCount = data.latentDefectCount + 1;
+      },
+      outputStrings: {
+        colorTower: {
+          en: '${color} Tower',
+        },
+        colorTowerDefamation: {
+          en: '${color} Tower /w Defamation',
```
quisquous, line 799:
```suggestion
          en: '${color} Tower Defamation',
```

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -772,6 +776,69 @@ const triggerSet: TriggerSet<Data> = {
...
+        else if (myColor)
+          return output.colorTower!({ color: myColor });
+      },
+      run: (data) => {
+        data.bugRot = {};
+        data.latentDefectCount = data.latentDefectCount + 1;
+      },
+      outputStrings: {
+        colorTower: {
+          en: '${color} Tower',
```
quisquous, line 796:
```suggestion
          en: '${color} Tower Stack',
```

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -772,6 +776,69 @@ const triggerSet: TriggerSet<Data> = {
...
+      type: 'GainsEffect',
+      // D65 Critical Performance Bug (blue)
+      // DC6 Critical Underflow Bug (red)
+      // Debuffs last 27s
+      netRegex: { effectId: ['D65', 'DC6'] },
+      condition: (data, matches) => {
+        data.bugRot[matches.target] = matches.effectId === 'D65' ? 'blue' : 'red';
+        return (matches.target === data.me) && data.latentDefectCount != 3;
+      },
+      delaySeconds: (data, matches) => parseFloat(matches.duration) - (data.latentDefectCount === 0 ? 4 : 7),
```
quisquous: In general, I've heard most "pass rot" calls happen right after latent defect (+defamation / stack explosions).  Do you think it's worth triggering on those abilities instead of trying to find a correct delay?

Also, fwiw, most streams I've seen have the people without rot coming to the people in the towers (who are planting).  Maybe it'd be worth just making this unconditional and telling this to everybody?

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
Legends0: Should this have said Soak instead of Stack?

Legends0: Forgot to mention, the callout only goes to those with the rot debuff. I was think a trigger based on the patch tether + defamation would be for the others but it would be similar output.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
Legends0: Ah, if that's better timing I think this trigger should be turned into a collect for the two bug debuffs and then a trigger on Latent Defect Ability for the pass and maybe a 'Grab Rot' for red/green tether or and a 'Break Tether + Grab Rot' for blue tether (with exception for last blue to just break).  A callout for the red/green for break I think can be done by tracking the loss of the blue tether debuff + delay according to the vuln.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
Legends0: I might be wrong, I haven't tested this in game, only in raidemulator. But it is true I just realized the stack could be an indicator to be near the blue tether which is needed.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
Legends0: Ok I have changed it to the ability.

quisquous:
I don't know if you've fixed this, but the "spread" call seemed to come very late (like 1 second left on rot debuff):
image

Legends0:
> I don't know if you've fixed this, but the "spread" call seemed to come very late (like 1 second left on rot debuff)

This was happening because I had used the debuff obtained from soaking the tower and not the rot. It should be corrected now. This also means the spread trigger could be merged into the rot collect trigger.

quisquous:
> EDIT: I just noticed that the latent defects from the towers get removed when rot expires... is that a check that needs to be done for the rot spread to occur? I suppose practically it may not matter because it would mean a tower was missed.

I'm not actually sure.  My suspicion here is that maybe you would explode if that debuff isn't cleansed by the proper rot explosion, since the fatal damage is called cascading latent defect.  Hello World has never been my strong suit in terms of actually understanding what all the debuffs mean.

Legends0:
Probably could use a "1st towers, 2nd towers, 3rd towers, final towers" call out or integration into these existing tower calls.

quisquous:
I was getting a bunch of `Trigger TOP Latent Defect Tower has missing outputString blue` errors, but it looks like those are probably fixed now?

quisquous:
I'll give this another test with all the changes and then merge if it looks good.  Thanks!

Legends0:
> I was getting a bunch of `Trigger TOP Latent Defect Tower has missing outputString blue` errors, but it looks like those are probably fixed now?

Yea I think I fixed this one.

quisquous:
Looked great to me, thanks!

PR #5266, opened 2023-02-10
raidboss: p3 add some tether triggers

Legends0:
- Adds a break tether callout for when the magic vulnerability up from the Patch spell expires and duration has <4s remaining.
- Adds callout for tether players for which tower to go to

~~Potential for this to be expanded to call the tower to run to with merging of my other PR #5264.~~

I don't really see it necessary to call the first break here since it happens naturally. For the last set I am not too sure on, I only know that the regression that breaks there changes order. ~~Not sure how these triggers impact P5, might need to condition these to only run in P3?~~ **EDIT:** P5 Uses different effectIds for its Local and Remote Regressions, but the same Magic Vulnerability debuff is used with Patch.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -782,6 +782,47 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
+      id: 'TOP Latent Defect Tether Towers',
+      type: 'GainsEffect',
+      // D71 Remote Code Smell (blue)
+      // DAF Local Code Smell(red/green)
+      // Using Code Smell as the regressions come ~8.75s after Latent Defect
+      // Debuffs are 23, 44, 65, and 86s
+      // TODO: Possibly include direction?
+      netRegex: { effectId: ['D71', 'DAF'] },
+      condition: Conditions.targetIsYou(),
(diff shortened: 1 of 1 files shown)


quisquous:
> This adds a break tether callout for when the magic vulnerability up from the Patch spell expires.

I'm less sure about this timing.  Depending on the heal plan, you also have to wait on heals too.  What do you think about a reminder of like 3-4s left on your near red/green tethers? (If you still have the debuff, given that red/green breaks early on the last set usually.)  My worry is mostly people breaking too soon.

Legends0:
There is room to spare, but with how mit planning works, you would want to break as soon as possible to catch the stack/defamation and patches.

Time wise is first tether can go off about 2.1s after received, realistically it's going to take another second to break. So about the 5.1s remaining mark it becomes safe to break.  About 1.9s after this, the rot aoe goes off, but does not do damage to source. It's feasible to set the delay to be 3s before debuff expires since no other magic damage comes out until next towers.

Legends0:
I tested this with the timing based on the vulns and it work pretty well as infotext for first three sets.

At least for my tests, it really only felt that the last set may need some delay. It's nice for awareness of the vuln being gone, but watching the health bars then it almost always needed to wait.

Legends0:
Not sure if the vuln checking is going to be useful for later, but I have left it in.

I think it's possible to reduce the collect/cleanup triggers here but I hadn't quite resolved that on my end.

PR #5268, opened 2023-02-11
oopsy: add some Hello World mistakes

quisquous:


quisquous:
Looks good to me.  I think there's still room for a few more edge cases but this seems to do a good job with most Hello World issues.

PR #5276, opened 2023-02-14
raidboss: Mobile tts fix

RadiantKakas:
Added window.speechSynthesis.getVoices(); in the constructor so the first TTS event would play.

With help from EQAditu#2691, adjusted the const voice to replace all underscores with dashes. Android expects underscores and Windows expects dashes.


ui/raidboss/browser_tts_engine.ts
```diff
@@ -30,6 +30,7 @@ export default class BrowserTTSEngine {
 
   constructor(private cactbotLang: Lang) {
     if (window.speechSynthesis !== undefined)
+	  window.speechSynthesis.getVoices();
       window.speechSynthesis.onvoiceschanged = () => this.initializeVoice();
     else
       console.error('BrowserTTS error: no browser support for window.speechSynthesis');
```
quisquous, line 36:
```suggestion
    if (window.speechSynthesis !== undefined) {
      // https://bugs.chromium.org/p/chromium/issues/detail?id=334847
      window.speechSynthesis.onvoiceschanged = () => this.initializeVoice();
  	  window.speechSynthesis.getVoices();
    } else {
      console.error('BrowserTTS error: no browser support for window.speechSynthesis');
    }
```

style, but also I think you need to set the event before calling getVoices.

ui/raidboss/browser_tts_engine.ts
```diff
@@ -56,7 +57,7 @@ export default class BrowserTTSEngine {
 
     // figure out what TTS engine type we need
     const speechLang = cactbotLangToSpeechLang[this.cactbotLang];
-    const voice = window.speechSynthesis.getVoices().find((voice) => voice.lang === speechLang);
+    const voice = window.speechSynthesis.getVoices().find((voice) => voice.lang.replaceAll('_','-') === speechLang);
```
quisquous, line 60:
```suggestion
    const voice = window.speechSynthesis.getVoices().find((voice) => {
      return voice.lang === speechLang || voice.lang.replaceAll('_', '-') === speechLang);
    });
```

I think this will break desktop if you always replace.  Better to check both the hyphen and the underscore versions so we can support both.

quisquous, line 60:
```suggestion
    const voice = window.speechSynthesis.getVoices().find((voice) => voice.lang.replaceAll('_', '-') === speechLang);
```

EDIT: sorry, nevermind.  This is correct, but lint will probably complain about the formatting.

ui/raidboss/browser_tts_engine.ts
RadiantKakas: Hmm.. I had it before just to get that initial call out of the way before getting into the initialize function.

quisquous: `window.speechSynthesis.onvoiceschanged = () => this.initializeVoice();` is setting an event callback, so isn't calling `initializeVoice` directly here.  I think I'm overthinking this; at the very least it all needs to be in curly braces.

RadiantKakas: Yep, I added the curly braces in my last merge!

quisquous:
> I'm not sure what the error means from the Lint check

It's asking you to use spaces instead of tabs, and the second diff is too long so needs to be put on its own line.

See: https://github.com/quisquous/cactbot/actions/runs/4177417280/jobs/7235291925 for details.

RadiantKakas:
> 

Ohhhh I see, thanks! Fixing now.

quisquous:
Thank you!!!

RadiantKakas:
No problem!! 

Doing some additional testing now that it's live, and It looks like it doesn't load *at all* on iOS. But, I think that would fall under a new issue? Tried in both Safari and Chrome and even the initial page where you choose the character is just blank.

valarnin:
> Doing some additional testing now that it's live, and It looks like it doesn't load _at all_ on iOS. But, I think that would fall under a new issue? Tried in both Safari and Chrome and even the initial page where you choose the character is just blank.

All iOS browsers use webkit behind the scenes, even Chrome or Firefox.

cactbot raidboss uses lookbehind in regexes.

Chromium (which is what OverlayPlugin uses via CEF) has supported this since 2017. Firefox has supported this since 2020.

webkit (and therefore iOS in general, regardless of what browser frontend you use) will finally support this feature in its next release I assume, since it's listed as supported in the current Technology Preview release. No clue when that release will actually happen.

https://caniuse.com/js-regexp-lookbehind

PR #5287, opened 2023-02-18
i18n: add & fix ko translations

Bing-su:
Comments are very helpful.


quisquous:
> Comments are very helpful.

If there's any confusion, please let me know.  I know there's a bunch of oopsy errors for TOP that I don't usually add for other content, and I think those are probably more confusing and less clear than raidboss triggers.

PR #5296, opened 2023-02-28
raidboss: TOP add phase tracking

Legends0:
This has more "phases" than fflogs, but it seems more useful to separate the p5 into multiple parts for different hello/near world behavior.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -8,6 +8,16 @@ import { PluginCombatantState } from '../../../../../types/event';
 import { NetMatches } from '../../../../../types/net_matches';
 import { TriggerSet } from '../../../../../types/trigger';
 
+type Phase =
+  | 'omega'
+  | 'm/f'
+  | 'final-omega'
+  | 'blue-screen'
+  | 'delta-version'
+  | 'sigma-version'
+  | 'omega-version'
+  | 'alpha-omega';
+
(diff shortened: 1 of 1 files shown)


Legends0:
I looked over the output in raidemulator and made some corrections. I think it was the 7B42 had to be changed because it happens around p2 as well.

Delta version has multiple casts, which could be suppressed.

I moved the auto attack that was removed in previous commit to ability which fixes why it was not appearing and added suppression for it, but it's probably still not required to have that be tracked since the fight resets to that value.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -8,6 +8,16 @@ import { PluginCombatantState } from '../../../../../types/event';
...
 
+type Phase =
+  | 'omega'
+  | 'm/f'
+  | 'final-omega'
+  | 'blue-screen'
+  | 'delta-version'
+  | 'sigma-version'
+  | 'omega-version'
+  | 'alpha-omega';
```
quisquous, line 19: Sorry for nitpicking, but these are all pretty wordy.  What about p1, p2, p3, p4, delta, sigma, omega, p6? These correspond to what fflogs (and everybody else I know uses) but still splits out p5 (which I think is a good idea).

Legends0: No worries. It was complicated thinking up a name for it and seeing already what we placed in the timeline, yet another difference in phase names.

PR #5303, opened 2023-03-09
raidboss: TOP P5 delta tether + beyond defense filter

Legends0:
The effects are unique for delta, so there wasn't a need to use phase filter.

I think maybe in the future it would say something more for the blue tether players like blue tether with some description about the world they may be with. For certain strats, players are using near world or distant world as a method to say which blue tether pair goes where.


Legends0:
Would it be better to have two separate triggers for a Delta Beyond Defense and p2 beyond defense?

quisquous:
> Would it be better to have two separate triggers for a Delta Beyond Defense and p2 beyond defense?

I could go either way.  It would mean it isn't possible to turn off p2 but still have p5 since they're the same trigger.  However, it seems pretty minor too.

Legends0:
> EDIT: I'm happy to merge this if you are.

Should be good now. I just ran it in raidemulator just now to fix an issue with the updated beyond defense.

PR #5306, opened 2023-03-10
raidboss: TOP p5 delta swivel cannon

Legends0:
Have not really tested this one out yet, but ran it through one of my encounters in raidemulator to get the right output. ~~It also only needs the 4 directions, but I used an 8 direction function.~~ **EDIT:** Replaced with using eye relative.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -1237,8 +1237,8 @@ const triggerSet: TriggerSet<Data> = {
       },
       outputStrings: {
         north: Outputs.north,
-        south: Outputs.south,
         east: Outputs.east,
+        south: Outputs.south,
         west: Outputs.west,
       },
     },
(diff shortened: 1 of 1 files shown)


quisquous:
Is this how you look at this mechanic? I am not thinking about absolute directions at all at this point, and am doing everything relative to the eye.  (I have the same feeling on monitor direction as well.)

Legends0:
> Is this how you look at this mechanic? I am not thinking about absolute directions at all at this point, and am doing everything relative to the eye. (I have the same feeling on monitor direction as well.)

I was just looking at the beetle each time. I think using the eye as a reference point can work and reduce the number of outputs here. Should it just say something like "Away from Eye" and "Near Eye"? I had some thoughts that maybe the output needed clarity as to what it is such as mentioning swivel, dynamis, or resolve worlds...

quisquous:
Yeah, I think even if you are looking at beetle, "go right" = "go relative north" = "go towards eye" and vice versa.  So, I think having that sort of output would allow for ~most of the customization I suspect people want.  It might just be the way I'm thinking about it, but my suspicion is that most people aren't thinking in absolute directions here.

PR #5317, opened 2023-03-25
raidboss: TOP P5 Omega Dodge Locations

Legends0:
This adds callouts for the safe spots during Run Dynamis: Omega Version.

The pattern used is "Distance, Direction".

There are combatants added once phase has reached the omega version. The combatants when ordered by id give us who will be spawned first. The positional data is known while they are invisible at start of the phase.

This needs to be tested... I have only 1.5 logs.

EDIT: The output for distance is not correct in raidemulator because of WeaponId not being available. So long as it's the same as P2 I think the logic is correct?


quisquous:
Ooh, thank you for writing this!!

I think one big thought I have is that although legs+anything and shield+staff have only a single solution, there's two safe spots for staff+sword that are opposite of each other (plus a third safe spot behind the staff but I don't know why you'd go there).  In this picture, there's "east far" and "west staff" for lack of a better way to put it (maybe even "west staff lean left") where you stand on the third ring and take a tiny step away from omega-F.  In some ways it's a much safer dodge because it puts the "far" pattern much closer to the middle so there's less running for cases of far->far or mid->far.

Ignore the fact in this picture that the staff/sword clones have disappeared and omega has already switched rings, but it should be obvious from the aoes what's going on here: staff_sword_ring3_lean.

Since the safe spots are the direct opposite of each other, I'm not sure how to have the trigger handle both in a way that's useful for users.  I guess it could say "E far / W staff" or something and then users could edit it to say the one they care about? That's kind of awkward.  Do you have any other suggestions here to handle this?

A couple of other followup thoughts that don't have to be done in this PR but just something to consider:
* could the trigger say something like "N or E close" prior to the omega cleave coming out, so people can preposition beforehand?
* do you think it's worth giving a left/right rotation as well, since it's always a 90 degree rotation from one spot to the next? just like a "right -> east mid" kind of thing for the second call?

Legends0:
The calls were meant to represent this diagram from the balance:


> Since the safe spots are the direct opposite of each other, I'm not sure how to have the trigger handle both in a way that's useful for users. I guess it could say "E far / W staff" or something and then users could edit it to say the one they care about? That's kind of awkward. Do you have any other suggestions here to handle this?

It could have an extra output for the case where there are multiple spots so it could some how be overridden separate from the single spot locations. The default could just be left what it is outputting now or adjusted so it's more noticeable from TTS/text that it is editable separately to the other patterns.

> could the trigger say something like "N or E close" prior to the omega cleave coming out, so people can preposition beforehand?

The positions of the entities are known so it could at least say a side the two males are on. I am not sure if their WeaponID is already defined though. I also do not know if the final omega in the center has a WeaponID for which way its ring is so right now the side it is lasering is not known until the cast starts. These would be something requiring testing in game to determine.

> do you think it's worth giving a left/right rotation as well, since it's always a 90 degree rotation from one spot to the next? just like a "right -> east mid" kind of thing for the second call?

I think that is a good idea as it is faster at that point to know left/right. I think it's safe to assume people are looking at the boss for the second direction to go to and that may generally be where the Omega-F that players would be hitting is/was located at spawn.

quisquous:
> The calls were meant to represent this diagram from the balance:

Yeah, that's what I figured.  The staff+sword "far" is on the 7th or even 6th ring, but the staff+sword "mid-ish" from my screenshot is on the 3rd ring on the opposite side.

> > Since the safe spots are the direct opposite of each other, I'm not sure how to have the trigger handle both in a way that's useful for users. I guess it could say "E far / W staff" or something and then users could edit it to say the one they care about? That's kind of awkward. Do you have any other suggestions here to handle this?
> 
> It could have an extra output for the case where there are multiple spots so it could some how be overridden separate from the single spot locations. The default could just be left what it is outputting now or adjusted so it's more noticeable from TTS/text that it is editable separately to the other patterns.

Are you suggesting we have three output strings for staff+sword, e.g.
```
staffSwordCombo: {
  en: '${farText} / ${midText}',
},
staffSwordFar: {
  // etc
},
staffSwordMid: {
  // etc
},
```

... and then users could edit `staffSwordCombo` to just be `${farText}` or `${midText}`?

> > could the trigger say something like "N or E close" prior to the omega cleave coming out, so people can preposition beforehand?
> 
> The positions of the entities are known so it could at least say a side the two males are on. I am not sure if their WeaponID is already defined though. I also do not know if the final omega in the center has a WeaponID for which way its ring is so right now the side it is lasering is not known until the cast starts. These would be something requiring testing in game to determine.

Yeah my thought was to do it on the cast itself, but to have a pre-call before that cast starts.

Legends0:
> Are you suggesting we have three output strings for staff+sword, e.g.
> 
> ```
> staffSwordCombo: {
>   en: '${farText} / ${midText}',
> },
> staffSwordFar: {
>   // etc
> },
> staffSwordMid: {
>   // etc
> },
> ```
> 
> ... and then users could edit `staffSwordCombo` to just be `${farText}` or `${midText}`?
> 

Yea, something like that or just have the return when it's staff+sword be
```
staffSwordCombo: {
  en: '${distance} ${dir}',
},
```

...and then the alertText would use something like:
```
const firstSpot = (!isF1in && !isM1in) ? output.safeSpot!({ distance: distance1, dir: dir1 }) : output.staffSwordCombo!({ distance: distance1, dir: dir1 });
const secondSpot = (!isF1in && !isM1in) ? output.safeSpot!({ distance: distance2, dir: dir2 }) : output.staffSwordCombo!({ distance: distance2, dir: dir2 });

return output.safeSpots!({ first: firstSpot, second: secondSpot });
```

I'll look into some changes.

Legends0:
I've updated it to use separate output for Staff + Sword combo including the additional safe sport near the middle on F's side.

The code originally was only giving an spot related to where M was. It now flips to F-side when close, I'll have to check that simply checking the X or Y is still adequate for F. The F side's Staff + Sword Combo required knowing the exact coordinate in relation to the cleave by the boss, so I compared both PosX and PosY. I used the inter-intercard callout in the direction.

The triggers will need another revisit if the second half works better as a left/right callout.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -1366,6 +1366,368 @@ const triggerSet: TriggerSet<Data> = {
...
+        }
+
+        const firstSpot = output.safeSpot!({ distance: distance1, dir: dir1 });
+        const secondSpot = output.safeSpot!({ distance: distance2, dir: dir2 });
+
+        if (!isM1In && !isF1In && !isM2In && !isF2In) {
+          // Output two locations => two locations
+          const staffSwordFar1 = output.staffSwordFar!({ dir: dir1 });
+          const staffSwordFar2 = output.staffSwordFar!({ dir: dir2 });
+          const staffSwordMid1 = output.staffSwordMid!({ dir: dir3 });
```
quisquous, line 1527: My only feedback here is if you can use a different output string for this "mid".  I think it's a different "mid" then the other one and I (in particular) want to use a different word there to differentiate.  You could call it `midStaff` or something?

quisquous (commented):
Thanks, I'll give this some more testing and see how it goes.

Legends0:
The early call I added likely will want some form of delaySeconds added with adjusting durationSeconds because the adds are not even visible for another 5~10s. I am not sure at which point after Run Dynamis: Omega Version or before Diffuse Wave Cannon the M+F are visible.

Legends0:
Based on quisquos findings, some of the calls are still flipped. Will need to look at the logic more closely.

Legends0:
The logic should be correct now... discovered OverlayPlugin had more precision than AddCombatants log line and I used a strict equals that would fail the logic.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -1381,6 +1381,369 @@ const triggerSet: TriggerSet<Data> = {
...
+              // NE
+              dir3 = output.dirNNW!();
+            } else {
+              // SE
+              dir3 = output.dirSSW!();
+            }
+          }
+        }
+        if (!isM2In && !isF2In) {
+          if (matches.id === '7B9B') {
```
quisquous, line 1503: Belatedly, is this logic incorrect? Should this boolean condition be flipped (i.e. `!== '7B9B'`) because it's for the second set?

I was going through and doing some edits/cleanup so that I could add a few extra output strings that some people requested and saw this.

Legends0: The output not matching the first set I think means the output is correct, but perhaps this was written poorly.

- The matches is based on the first cannon and the comment does not what is happening in the second cannon.


However, I think it is somehow possible to simplify the logic by placing it into a function since  7B9B M1+F1 === 7B9C M2 + F2.

quisquous: Ohhhh I see it's just the comment that's wrong.  I put it into a function in #5334.

Issue #5153, opened 2022-12-21
delayed triggers throw an error if combat ends before they fire


xiashtra:
## Description
Delayed triggers throw an error if combat ends before they fire.

## Additional information
If a trigger with a delay is queued and then combat ends before the `delaySeconds` duration passes, a console error will occur.

## Steps to reproduce
1. Copy the attached test trigger to local user triggers folder
2. Zone in to Sastasha normal, unsynced with a high level character
3. Aggro the Black Bats at the start of the dungeon and let them execute at least one autoattack
4. Kill the Black Bats before the trigger fires and wait for the trigger delay to pass (30s)
5. Observe the error in the ACT console

## Test trigger
[sastasha.zip](https://github.com/quisquous/cactbot/files/10281733/sastasha.zip)

PR #5318, opened 2023-03-25
raidboss: TOP P5 Sigma Staff/Feet Call

Legends0:
The logs seem a bit strange on this one. Omega-M which casts Discharge, is seen as casting the spell but Omega-F is what readies it. Unsure if this is capturing the correct npc model.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -1341,7 +1341,9 @@ const triggerSet: TriggerSet<Data> = {
       // Track from Discharger (7B2E)
       type: 'Ability',
       netRegex: { id: '7B2E', source: 'Omega' },
+      condition: (data) => data.phase === 'sigma',
       delaySeconds: 6,
+      suppressSeconds: 1,
       promise: async (data, matches) => {
         data.combatantData = [];
         data.combatantData = (await callOverlayHandler({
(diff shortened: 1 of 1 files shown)


Legends0:
Last I tested this, it's throwing an 'error stopped' like message when the trigger activates in sigma but trigger actives fine on the p2 Discharger event. I am not quite sure what is causing the error. It's likely happening in either the promise or in the data.combatantData.pop() during runtime.

quisquous:
Hmm.  I don't think `data.combatantData.pop()` could go wrong unless `data.combatant` was undefined.  "Error stopped" sounds a lot more like #5153, where the pull is over before the delayed trigger fires.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -1335,6 +1335,43 @@ const triggerSet: TriggerSet<Data> = {
         return { infoText: output.noDebuff!() };
       },
     },
+    {
+      id: 'TOP Sigma Superliminal/Blizzard',
+      // Omega-M casts Superliminal/Blizzard
+      // Track from Discharger (7B2E)
+      type: 'Ability',
+      netRegex: { id: '7B2E', source: 'Omega' },
```
quisquous, line 1343:
```suggestion
      netRegex: { id: '7B2E', source: 'Omega-M' },
```

Maybe this?

Legends0: I think you're correct here. The P2 is just Omega, but P5 here is Omega-M and the P2 and P5 npc are different sourceIDs.

P2:
```
|4000ED62|Omega|7B2E|Discharger|
```

P5 in same pull:
```
|4000EEB8|Omega-M|7B2E|Discharger|
|4000EEB8|Omega-M|7B2A|Superliminal Steel|4000EEB8|Omega-M|1.200|90.00|100.00|0.00|
```

Legends0:
I dropped the delay to 6.2, I swear I had tested with 6 on my part and it did not work, but it may work at 6.2 if others seeing it less than 6 from the ability going off in the log.

PR #5326, opened 2023-03-30
raidboss: TOP P5 Sigma Omega-M location

Legends0:
Adds callout for the card/intercard Omega-M jumps to.

Fflogs seems to show that the same npc that casts Omega-M moves to the card/intercard, so assuming that is correct this may just need its delaySeconds tested to see how early the data is available.


Legends0:
Ok, so the default timing if it's too early is going to output Northwest. So 2s is too short.

Legends0:
5s does not work, 6s works... further testing for if it can be less delay / how accurate needed.

PR #5327, opened 2023-04-01
raidboss: TOP P5 Sigma Tower Orientation

Legends0:
This uses my TOP P5 Sigma Omega-M location work.

Basically the log indicates that the Omega NPCs that cast 7B74 Wave Cannon will also cast the 7B04, 7B05, and 7B06 Storage Violation spells. The point where they cast this from is seen as the center of the tower. Theoretically, if this position is accurate and available at time of tower spawning we can pull the tower data from the list of these combatants.

Initially this draft is just taking M's location and checking each NPC if it exists at where we think the tower should be, if it does, then we narrow down the two sides the tower's orientation could be to where it is.


Legends0:
This was revealed to not be possible with the combatants as they move within the second the towers expire.

PR #5328, opened 2023-04-03
i18n: Zodiark-ex missing FR strings

mooondark:


mooondark:
Don't know why it failed ... How to force to check again ?

PR #5329, opened 2023-04-03
raidboss: TOP Add delay to P3 Monitors

Legends0:
The current TTS output of both West/East Oversampled Wave Cannon and Oversampled Wave Cannon Loading is unintelligible without some sort of delay added.

I recommend adding delay to the personal monitor callout here. Thoughts? @quisquous


quisquous:
I'm not sure that 0.5s is enough here.  Maybe we should just drop the tts for the east/west monitors? I think that's most easy to see.

Legends0:
Closing as this PR isn't needed anymore.

PR #5330, opened 2023-04-03
raidboss: TOP P5 Omega Earlier Safe Spot Call

Legends0:
There is an AddCombatants that occurs either before or shortly after the 8015 Run Dynamis: Omega Version cast starts. This utilizes OverlayPlugin, which the other two triggers still pull data from.

Previous PR did not have delay and so did not fire if the combatants were added late per the log line. Alternatively, this trigger can fire off a collection of AddCombatants that match the BNPCIds, but some care would need to be accounted for if they happened to be added before the Omega Version cast starts since that is also what triggers the 'omega' phase. Basing on AddCombatants (if this line varies) could also make a durationSeconds more varied. So I have stuck with the 8015 cast for now and may test different delays.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -1487,12 +1487,12 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
-      id: 'TOP Omega Safe Spots',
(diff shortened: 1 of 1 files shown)


quisquous:
Do you feel like this is good to land, or should I wait on merging?

Legends0:
I think it's good enough to land. It can then be adjusted in #5334.

Issue #5335, opened 2023-04-06
TOP Sigma Superliminal/Blizzard making inconsistent calls


quisquous:
No more information than this.  I increased the delay to 7 and sometimes it was right and sometimes it was flipped.  I will take a look at logs and see if I can get more information.

I think we should disable this for now.

xpdota:
Is it possible that the NPC that casts discharger isn't the always the same as the one whose WeaponId needs to be checked 100% of the time? The only difference I see between this logic and mine is how it determines which NPC to check the WeaponId for.

Souma-Sumire:
SuperliminalSteel WeaponId is 11
(I have been verifying in my user.js for over a month)

quisquous:
I think the issue with the original trigger is that the outputs were reversed.

This is incorrect:
```typescript
      outputStrings: {
        optimizedBlizzard: {
          en: 'Follow Laser, Move In',
          ko: '레이저 따라서 안으로',
        },
        superliminalSteel: {
          en: 'Wait First',
          ko: '기다렸다가 이동',
        },
      },
```

Re: should we check for WeaponId 11, I think the answer is no? I always see 11 in all logs.  Here's a snippet of legs vs staff:

legs
```
[20:02:44.882] StartsCasting 14:4001B436:Omega-M:8014:Run: ****mi* (Sigma Version):4001B436:Omega-M:4.700:99.99:99.99:0.00:-3.13
[20:03:10.312] 261 105:Change:4001B436:WeaponId:11
[20:03:16.765] 261 105:Change:4001B436:Radius:5.0100:TransformationId:492:WeaponId:5:ModelStatus:2048
[20:03:17.820] 261 105:Change:4001B436:WeaponId:11:ModelStatus:0:PCTargetID:10967424
[20:03:20.793] AOEActionEffect 16:4001B436:Omega-M:7B2E:Discharger:
[20:03:26.534] 261 105:Change:4001B436:PosX:107.0711:PosY:92.9289:PosZ:0.0000:Heading:-0.7855:WeaponId:4
[20:03:40.745] StartsCasting 14:4001B436:Omega-M:7B2A:Superliminal Steel:4001B436:Omega-M:1.200:107.07:92.93:0.00:-0.79
```

staff
```
[21:03:41.683] StartsCasting 14:4002352C:Omega-M:8014:Run: ****mi* (Sigma Version):4002352C:Omega-M:4.700:99.99:99.99:0.00:-3.12
[...]

PR #5336, opened 2023-04-06
raidboss: disable TOP Sigma Superliminal/Blizzard

quisquous:
See: #5335


quisquous:
@Legends0 I'm going to disable this for now, and we can keep debugging and re-enable once it's more consistent.

PR #5338, opened 2023-04-07
raidboss: TOP adjustments and cleanup for omega pre-dodges

quisquous:
This is #5334 but applied to #5330 as well.


quisquous:
I have tested this locally and it looks good, so I'm going to merge so that it can get some further testing before release.

PR #5350, opened 2023-04-13
raidboss: add trigger set ids

quisquous:
This will make it possible to enable/disable entire triggersets at once, and also make it more reasonable to add config ui to individual trigger sets for options and such.


quisquous:
cc @valarnin because we talked about this

xiashtra:
Would there be an opportunity here to also establish sub-triggersets, essentially trigger dependency chains, so that collectors/cleanup triggers are automatically enabled/disabled when a trigger that requires them to function is enabled/disabled?

valarnin:
> cc @valarnin because we talked about this

I think the only thing missing from our discussion in Discord is the eslint rule to ensure that triggerset IDs are unique? Or it could be a github actions check that runs on PR, either way.

I don't know if you want to include runtime checking for unique triggerset id in this PR or in some other PR, but we talked about that as well.

quisquous:
> Would there be an opportunity here to also establish sub-triggersets, essentially trigger dependency chains, so that collectors/cleanup triggers are automatically enabled/disabled when a trigger that requires them to function is enabled/disabled?

Mmm, right now there's no way to add/remove triggers (or load/unload a triggerset) dynamically at runtime.  This is especially true because later triggersets could override a previous one, so unloading a triggerset would be kind of weird because it might reenable a previously suppressed trigger with a duplicate id? I don't know.  It's not something that feels high on my list to deal with.  What problem are you trying to solve with this? It is a performance optimization?

xiashtra:
> > Would there be an opportunity here to also establish sub-triggersets, essentially trigger dependency chains, so that collectors/cleanup triggers are automatically enabled/disabled when a trigger that requires them to function is enabled/disabled?
> 
> Mmm, right now there's no way to add/remove triggers (or load/unload a triggerset) dynamically at runtime. This is especially true because later triggersets could override a previous one, so unloading a triggerset would be kind of weird because it might reenable a previously suppressed trigger with a duplicate id? I don't know. It's not something that feels high on my list to deal with. What problem are you trying to solve with this? Is it a performance optimization?

I was thinking of the possibility that a user disables a trigger that they don't want or selectively enables a trigger that they do want without enabling any others, without realizing that the trigger they disabled/enabled is required for or requires other triggers to function.  The configuration UI treats every trigger as a self-contained unit, but there are instances where triggers have interdependencies that are not expressed to the end user.

quisquous:
> > > Would there be an opportunity here to also establish sub-triggersets, essentially trigger dependency chains, so that collectors/cleanup triggers are automatically enabled/disabled when a trigger that requires them to function is enabled/disabled?
> > 
> > 
> > Mmm, right now there's no way to add/remove triggers (or load/unload a triggerset) dynamically at runtime. This is especially true because later triggersets could override a previous one, so unloading a triggerset would be kind of weird because it might reenable a previously suppressed trigger with a duplicate id? I don't know. It's not something that feels high on my list to deal with. What problem are you trying to solve with this? Is it a performance optimization?
> 
> I was thinking of the possibility that a user disables a trigger that they don't want or selectively enables a trigger that they do want without enabling any others, without realizing that the trigger they disabled/enabled is required for or requires other triggers to function. The configuration UI treats every trigger as a self-contained unit, but there are instances where triggers have interdependencies that are not expressed to the end user.

At least from the config ui's perspective "disabled" just means "hide this".  The triggers all still run for exactly those reasons.  Advanced use cases where people are overriding triggers / fully disabling triggers are on their own.

quisquous:
> > cc @valarnin because we talked about this
> 
> I think the only thing missing from our discussion in Discord is the eslint rule to ensure that triggerset IDs are unique? Or it could be a github actions check that runs on PR, either way.

Yeah, added a github action test for this.  eslint would be tricky since it'd need to somehow know the full set of triggerset IDs (via some manifest-y thing, as loading all triggersets to do lint seems not great).

> I don't know if you want to include runtime checking for unique triggerset id in this PR or in some other PR, but we talked about that as well.

Yeah, I added a runtime check for this and dropped the triggerset if the id is defined but also a duplicate.

Souma-Sumire:
yes, nice!

PR #5355, opened 2023-04-18
raidboss: correct translation

Echoring:
This is a translation mistake. (misunderstood as maximum melee range)
<del>JP is also mistranslated. I can read but cannot write, @MaikoTan </del> Done.


resources/outputs.ts
```diff
@@ -336,7 +336,7 @@ export default {
     de: 'Raus aus Nahkampf',
     fr: 'Sortez de la mêlée',
     ja: '近接最大レンジ',
```
MaikoTan, line 338:
```suggestion
     ja: '近接レンジ離れ',
```

PR #5357, opened 2023-04-21
config: add config ui for trigger set options

quisquous:
This adds a new `config` entry to trigger sets that can add field entries to the top of the trigger set under the raidboss section in the config ui. Trigger sets must have an id to have any config loaded, saved, or displayed.

Triggers can use `data.triggerSetConfig['optionId']` here to access anything listed in the `config` section by id. The `Data` type for a trigger set can be extended in a way that will type check ids and add field names.

To avoid naming collisions, not all trigger set config values are loaded. Trigger sets get their own config's values loaded implicitly, but if they need to load some other trigger set's values, they can use `loadConfigs` and specify other trigger sets by id.

For backwards compat, `data.triggerSetConfig` contains all saved values (even ones that no longer have a `config` entry). Also, it is possible to use `loadConfigs` to load trigger set ids that no longer exist.

To also support backwards compat, `ConfigEntry.default` can now be a function that reads from options. See the e8s/TEA changes for how this works.

### DIFF
--- a/eslint/cactbot-triggerset-property-order.js
+++ b/eslint/cactbot-triggerset-property-order.js
@@ -30,6 +30,7 @@ module.exports = {
     const raidbossOrderList = [
       'id',
       'zoneId',
+      'config',
       'overrideTimelineFile',
       'timelineFile',
       'timeline',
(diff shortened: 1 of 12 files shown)


quisquous:
@valarnin FYI if you're interested in taking a look

valarnin:
I think this PR covered all the points I raised in our conversation. I won't have time to do more than a cursory examination at the code, but it looks good as far as my use cases are concerned.

Souma-Sumire:
I love this new feature, thanks for developing it.

PR #5358, opened 2023-04-23
docs: add new OverlayPlugin log line information

quisquous:


quisquous:
cc @valarnin if there's anything you wanted to add about either of these line types

docs/LogGuide.md
```diff
@@ -2453,3 +2461,114 @@ Parsed Log Line Examples:
...
+<a name="line261"></a>
+
+### Line 261 (0x105): CombatantMemory
+
+OverlayPlugin has a `getCombatants` function which returns the state of combatants in the game.
+However, it is hard to know when it is safe to call this function
+and know that combatants have moved into position (or changed model or changed heading).
+These lines give more granular information when combatants change their status.
+Please note that this is still polling memory (so timing may be racy)
+and there are some heuristics to not emit too many lines (so data may be imprecise).
```
valarnin, line 2474: It may make sense to link back to OP's code for the heuristics here? Maybe something like:

```md
For more information, see the [class definition](https://github.com/OverlayPlugin/OverlayPlugin/blob/main/OverlayPlugin.Core/MemoryProcessors/Combatant/LineCombatant.cs#L27) in OverlayPlugin
```

docs/LogGuide.md
```diff
@@ -2453,3 +2461,114 @@ Parsed Log Line Examples:
...
+in current savage and ultimate content in the game data itself.
+This is to prevent data mining.
+However, as these ability names need to be displayed by the game itself
+these ability names are sent as network data upon zoning in.
+
+These lines display the currently obfuscated abilities
+for the current zone for the current game locale.
+
+Note that `:` characters are not escaped even if the game abilities have a `:` in them.
+It is recommended to use the network log line format to parse these lines for that reason.
```
valarnin, line 2536: The documentation for this line should absolutely explain that CR and LF characters are escaped to prevent issues with the network log format, and therefore need to be unescaped. There are direct examples of this in the `rsv lines.txt` file I included in the PR comments, e.g.:

```
261|2023-04-21T23:24:06.0630000-04:00|en|00000051|_rsv_35827_-1_1_0_0_S13095D61_E13095D61|Further testing is required.,\r�)�� , assist me with this evaluation.|38151741aad7fe51
```

Additionally, it should probably be stated somewhere that the lines may contain special unicode character sequences that the game client will replace with placeholder values, e.g. in my example line above that gets replaced with the player's first name I believe.

In fact, we may want to modify the replace logic in OP to just escape any non-standard character with the assumption that downstream consumers will need to unescape fully, e.g. the above line might instead read:

```
261|2023-04-21T23:24:06.0630000-04:00|en|00000051|_rsv_35827_-1_1_0_0_S13095D61_E13095D61|Further testing is required.\u0002\u0010\u0001\u0003\u0002,\u000d\u00ef\u00bf\u00bd\u0007\u0002)\u0003\u00ef\u00bf\u00bd\u0002\u0003\u00ef\u00bf\u00bd\u0002 \u0002\u0003, assist me with this evaluation.|38151741aad7fe51
```

docs/LogGuide.md
quisquous: Done, thanks!

docs/LogGuide.md
quisquous: Added some text here.

Re: escaping. I think really it'd depend on who the consumers of these messages are in terms of what made the most sense.  I don't suspect there will be many (any??) consumers of the messages part of this (vs abilities and statuses) so it probably doesn't need too much time spent massaging it unless it starts causing problems or somebody has requests for what it should look like.

docs/LogGuide.md
```diff
@@ -2473,6 +2473,9 @@ These lines give more granular information when combatants change their status.
 Please note that this is still polling memory (so timing may be racy)
 and there are some heuristics to not emit too many lines (so data may be imprecise).
 
+For more information,
+see the [class definition](https://github.com/OverlayPlugin/OverlayPlugin/blob/main/OverlayPlugin.Core/MemoryProcessors/Combatant/LineCombatant.cs#L27) in OverlayPlugin.
+
 There are three types of this line:
 
 - Add: emits all initial fields for this combatant
```
valarnin, line 2481: Just thought of this today, but it may also make sense to note that `Add` lines exclude "default" values, e.g. if it's an `int` value type and the value is `0` on add, it won't write that field?

quisquous: Oh yeah, that's really important to say too, thank you!

PR #5363, opened 2023-04-23
i18n: correct ASS(S) fr translation of Hateful Visage

### LINKED ISSUE #5217: Raidboss: Another Sil'dhin Subterrane some triggers not appearing
## Description

Most of ASS (normal) triggers work fine but looking into the Raidboss list I've noticed some triggers I've never seen through any of my runs. I believe these include the following, but it could be more that I haven't noticed:
- ASS Fresh Puff 2 Bait ({boss} {dir} {puff} trigger)
- ASS Golden Silver Flame ({loc} trigger)

## Additional information

Playing in French, if that matters.

quisquous:
Fixes #5217.

cc @Akurosia who added this in #4990

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
+++ b/ui/raidboss/data/06-ew/dungeon/another_sildihn_subterrane-savage.ts
@@ -2538,7 +2538,7 @@ const triggerSet: TriggerSet<Data> = {
         'Ball of Fire': 'Boule de flammes',
         'Eastern Ewer': 'cruche orientale',
         'Gladiator of Sil\'dih': 'gladiateur sildien',
-        'Hateful Visage': 'Visage de haine',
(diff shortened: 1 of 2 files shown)


Akurosia:
image

As the "Hateful Visage" is the skill and not the bnpc in this case, i think original translation is correct (as i think the bnpc is in a fully different instance)

quisquous:
> image
> 
> As the "Hateful Visage" is the skill and not the bnpc in this case, i think original translation is correct (as i think the bnpc is in a fully different instance)

Looking at the log provided in the issue, in this case it is the bnpc and not the skill name. The `replaceSync` section is for bnpc names / messagesand `replaceText` should be for skills / timeline texts.

Akurosia:
thanks for clarification, i see the issue now, will look to avoid it next time :)

PR #5367, opened 2023-04-23
raidboss: TOP add config option + rotation to omega dodge

quisquous:
Because the rotation direction depends on which side you have dodged to, it's impossible to write this PR without having the user pick which dodge they are doing.

This promotes the dodge direction to a new option (defaulting to far, since that seems to be the PF approach) and adds in the rotation direction as well.

Because there's an option, there's no longer any combo output strings for staff sword, and the correct one is picked.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -38,6 +38,7 @@ export type TetherColor = 'blue' | 'green';
 export type TrioDebuff = 'near' | 'distant';
 
 export interface Data extends RaidbossData {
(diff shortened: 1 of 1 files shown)


quisquous:
cc @Legends0 if you wanted to test this out too

Legends0:
> cc @Legends0 if you wanted to test this out too

I'll try this out next time my group runs.

Legends0:
I had it call left then far E, but that was actually right. Using far dodge.

Legends0:
After resetting my config, it seems to have resolve itself as I haven't been able to replicate the issue.

Legends0:
This works well since the 5372 fix. Haven't had an issue with it.

quisquous:
Ok, I think I'll land this and do a release.  Thanks for all the testing.

PR #5369, opened 2023-04-26
raidboss: Fix ParseHelperType definitions

valarnin:
To quote Discord DMs:

> valarnin: ... is it just me, or is the `ParseHelperType` in `netregexes.ts` just flat out incorrectly defined? I just have no clue how typescript allows it in its current state.
> Consider:

```ts
type ParseHelperType<T extends LogDefinitionTypes> =
  & {
    [field in Extract<keyof NetFieldsReverse[T], string>]?: string;
  }
  & { capture?: boolean };

type EmptyType = ParseHelperType<'Ability'>;

type ParseHelperTypeFixed<T extends LogDefinitionTypes> =
  & {
    [field in keyof NetFields[T]]?: string | string[];
  }
  & { capture?: boolean };

type CorrectType = ParseHelperTypeFixed<'Ability'>;
```

image
image

> valarnin: (in fact, `Extract<keyof NetFieldsReverse[T], string>` evaluates to `never` because all keys of a given `NetFieldsReverse` type entry are numerical and therefore can never be assigned to `string`)

### DIFF
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -1,4 +1,4 @@
-import { NetFieldsReverse } from '../types/net_fields';
+import { NetFields, NetFieldsReverse } from '../types/net_fields';
 import { NetParams } from '../types/net_props';
 import { CactbotBaseRegExp } from '../types/net_trigger';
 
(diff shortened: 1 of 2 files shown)


quisquous:
How did this even work??

valarnin:
> How did this even work??

At a guess, the narrowing to `never` made typescript shrug and say "they probably know what they're doing", but honestly I have no clue. Or maybe it's because the parent types are pulled from a .d.ts file which is evaluated/scoped differently.

PR #5370, opened 2023-04-28
raidboss: TOP Unlimited Wave Cannon Dodge + Clock

Legends0:
I noticed I could use some code I wrote for DSR gigaflares here that should also work for determining the exaflare clock here.

The startsUsing log lines are sometimes incorrect and sometimes correct, so trying to use OverlayPlugin instead ~~but I do not yet know the proper delay needed. This will need 2s delay as that is the time between Wave Cannon exas, but the additional 1s here is for OverlayPlugin delay needed.~~

_Determined that 1.1s is enough_ (For most of my cases, but 1.2s needed for all my cases). This calls before players even see the exas.

This calls the card/intercard that is safe and next to the first exaflare. In the same output, it calls the direction (clock/counterclock) the exaflares spawned in which is where the player will need to run later.


Legends0:
As far as I can tell, I was getting accurate OverlayPlugin data at time of casts. Raidemulator seems to work, but I haven't run the cleaned up version in game. I think it can handle both Unlimited Wave Cannons but I don't yet have a log of both.

Legends0:
This might need another trigger to collect the exaflare positions... the delay window is too small doing it in one trigger like this.

Legends0:
So, in typescript simulation with some data I have from runs the logic works now. Just need to determine the lowest/recommended delay.

Theoretically 1s and some small change is when the casting happens and position data should be available, but historically have had to add .3s to .5s for safety.

Legends0:
This may need an additional reminder for clock/clockwise on count of the puddles as a vocal reminder.

Legends0:
> Has this been pretty reliable for you in practice? I think I had been using a previous version (with 5s delay?? and sometimes incorrect), but I will update to your latest version here and test myself some.
> 
> Personally also, I think if things are so tight that 0.1s can cause errors, it's better to add even more delay. The issue is mostly that the delay is client-side but when the information gets to people's computers can depend on latency. So, even if it works for you, it may not work for others. If this is a call that doesn't need to happen immediately, maybe 1.5 (or more?) would be much safer and reliable for more users.

This has been reliable for me since I changed filter to checking the combatants list for the same id from the collect. Before the issue was overlayPlugin would return results in different orders than I expected. I had assumed it would return a list in the order of IDs I passed in, but that was not the case on later outputs of the lists compared to the collect list so I corrected that which allowed me to verify the rotation and dirs.

The only time I saw a failure at 1.1s was when 7/8 players died, including myself, right before the combatants came out which I think triggered some higher than normal latency and so I upped it to 1.2s on my end. I think 1.5s will be more than enough which I have set it back to.

PR #5377, opened 2023-05-01
raidboss: TOP P6 Add Cosmo Arrow Exa Calls

Legends0:
There are NPCs that are added at the first Cosmo Arrow that then cast 7BA3 Cosmo Arrow. The position that these NPCs spawn in at or where they start casting at can be used to determine if it is in or out exasquare first.

There are 4 NPCs used for the outer exasquare and two NPCs for the inner exasquare.

The second Cosmo Arrow will use these same NPCs.

This calls initial in/out before any visible aoes. The rest are timed with the ability.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -60,6 +60,7 @@ export interface Data extends RaidbossData {
   monitorPlayers: NetMatches['GainsEffect'][];
   deltaTethers: { [name: string]: TetherColor };
   trioDebuff: { [name: string]: TrioDebuff };
+  seenOmegaTethers?: boolean;
   cosmoArrowCount: number;
   cosmoArrowIn?: boolean;
   cosmoArrowExaCount: number;
--- a/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
+++ b/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
@@ -1984,12 +1984,12 @@ const triggerSet: TriggerSet<Data> = {
       type: 'Ability',
       netRegex: { id: '7BA4', source: 'Alpha Omega', capture: false },
       preRun: (data) => data.cosmoArrowExaCount = data.cosmoArrowExaCount + 1,
-      suppressSeconds: 1, // Only capture 1 in the set of casts
       durationSeconds: (data) => {
         if (data.cosmoArrowExaCount === 3 && data.cosmoArrowIn)
           return 5;
         return 3;
       },
+      suppressSeconds: 1, // Only capture 1 in the set of casts
       infoText: (data, _matches, output) => {
         if (data.cosmoArrowIn) {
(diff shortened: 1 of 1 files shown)


Legends0:
Closing this for now, I think the logic might be too tight, so it doesn't not resolve properly. The timing is also off by 1.

Legends0:
Going to recode this based on count of the casts instead of position data since that seems to be inaccurate.

Legends0:
This seems to work now.

ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts
```diff
@@ -1920,6 +1924,115 @@ const triggerSet: TriggerSet<Data> = {
...
+      outputStrings: {
+        in: Outputs.in,
+        inWait2: {
+          en: 'In => Wait 2',
+        },
+        outWait2: {
+          en: 'Out => Wait 2',
+        },
+        cardinalPositionIn: Outputs.moveAway,
+        cardinalPositionOut: {
```
quisquous, line 2031: The rest of this PR looks good and seems good in practice, but I'm not sure I understand what "cardinal in" and "cardinal out" mean here.  I've always just called this "sides" in both cases.

Legends0: Ah, the in/out in the variable name was to represent if it is in first (box pattern) or out first (plus pattern). 

For`output.cardinalPositionOut!()` I chose 'cardinal out' as it is a diagonal outwards movement. Originally I was going to leave out the word 'out', but in practice I felt the tts needed the extra detail as outward movement is needed for this dodge rather then going right into the cardinal (side) position. Sides could work and makes for a quicker call with less syllables, but I put cardinal since some of the players may also be going intercard at that time. The dodge here at this exaflare is safe for everyone to do though as the split into clock positions happens after this movement.

For the in pattern, I chose the generic move call. I think it may need something else, but could not think of something at the time. I think it should have a different wording than the other pattern since the movement is different. It still could be a diagonal out movement, but you are not stepping into the center of the + of two exaflares and instead going into the opening that the two exaflares create. I think It's harder to do this movement. I chose 'Move' because at that point each person just needs to know where to go since that is roughly when the wave cannon is going to go off. If both players are still in the intercard, they will very likely clip eachother, so the prefered movement there is more like a horizontal/to the side movement than a diagonal movement with the player(s) already at the correct row. Technically the next movement after can be in or out, but I preferred in since it places party closer to reach heals etc.

quisquous: Re: "cardinal out".  What about "out + sides" or "sides + out"?

Re: "cardinal in".  I guess "move" is fine there too, it just felt like an odd call among the rest of them that had directions.  I think this could be "sides" as well (esp as you say "the prefered movement there is more like a horizontal/to the side"), but I think it's fine as-is.

Legends0: I've reworded the cardinal out to sides + out. I will leave the in call as default move.

PR #5378, opened 2023-05-03
raidboss: re-enable TOP Sigma Superliminal/Blizzard

### LINKED ISSUE #5335: TOP Sigma Superliminal/Blizzard making inconsistent calls

quisquous:
Switching the output strings and increasing the delay makes this consistent.

Fixes #5335.


Legends0:
I can confirm this was consistent for me today.

PR #5379, opened 2023-05-03
resources: Support for `combatantMemory` lines

valarnin:
This is extremely preliminary support for `combatantMemory` lines.

It's still very rough, but feedback is appreciated.

Outstanding issues:

- [x] ~~Only support for `NetRegexes` so far~~
- [x] ~~Requires that entries passed be passed in the order they will appear in the log line. It's unlikely that these will be allowed out of order due to potential performance issues.~~
- [x] ~~No example in `test.ts` test triggers~~
- [x] ~~Pretty untested in general~~

Given the following:

```ts
// one-liner
NetRegexes.combatantMemory({ 'pair': [{ 'key': 'ModelStatus', 'value': ['1234', '5678'] }, {'key': 'Owner', 'value': '10001234' }] })
// formatted
NetRegexes.combatantMemory({
  'pair': [
      { 'key': 'ModelStatus', 'value': ['1234', '5678'] },
      {'key': 'Owner', 'value': '10001234' },
    ],
  });
```

This results in the following regex:

```js
/^(?<type>261)\|(?<timestamp>[^|]*)\|(?<change>[^|]*)\|(?<id>[^|]*)\|(?:[^|]*\|)*?(?<key>ModelStatus)\|(?<value>(?:1234|5678))\|(?:[^|]*\|)*?(?<key1>Owner)\|(?<value1>10001234)\|/i
```

Broken down into fields:
```
/^
(?<type>261)\|
(?<timestamp>[^|]*)\|
(?<change>[^|]*)\|
(?<id>[^|]*)\|

(?:[^|]*\|)*?

(?<key>ModelStatus)\|
(?<value>(?:1234|5678))\|

(?:[^|]*\|)*?

(?<key1>Owner)\|
(?<value1>10001234)\|

/i
```

I'm unsure of my free time to continue working on this, so if I can't continue to work on this PR, someone else can feel free to take over this effort.


valarnin:
Aside from the order-of-entries issue, this is ready for review.

As stated previously, I don't think the order-of-entries issue will ever really be resolved, so it may make sense for a test to be added, or the `repeatingFields` type definition to be expanded in some manner to support this, but I think that's beyond the scope of this PR.

I've done some testing with both the `Regexes` and `NetRegexes` versions and they both work as anticipated from what I've tested.

quisquous:
>  Requires that entries passed be passed in the order they will appear in the log line. It's unlikely that these will be allowed out of order due to potential performance issues.

I think this doesn't have to be true.  The regex builder could sort them itself (rather than requiring the caller to do so), and we could update OP to sort the fields to match as well.

valarnin:
The problem is sorting it OP-side. Due to the nature of the two different types of checks (time-based vs threshold-based), we can't check properties in order to start with. Instead we would need to sort after accumulating all properties to be written, which is going to have a pretty bad negative impact on performance.

`var changed = new List<FieldInfo>();` could instead be changed to be a dictionary or something, but that would add the penalty of dictionary lookups which is also slow.

quisquous:
Yeah, makes sense to not add additional performance hits to something that is writing a lot of log lines.

I guess that leaves an existential question for this PR? If the fields in the line are not ordered, then it is impossible to write a reasonable regex for multiple fields, right?  If that's true, then I'm not sure what this PR is doing?  Am I misunderstanding something here?

A couple of overlapping thoughts:
* add the full definitions as-is to netlog_defs for completeness, but don't allow regexes?
* or, restrict regexes to a single key/value entry, such as looking for one weapon id (maybe we could handle posX/posY/posZ all in order as a special case?)
* make the log guide generator incorporate this somehow? (Possibly also for the key/value pairs in ability, even if they aren't / shouldn't be included in the general ability regex?)

valarnin:
The log line order is deterministic, with the following simplified logic:

```ts
const positionFields = []; // PosX, etc
const CheckFieldDelay = []; // OwnerID, etc
const AllFields = []; // Every field in Combatant that's not flagged as ignored, in the order they appear in the Combatant struct

let fieldsToWrite = [];
if (position.hasPassedDistanceThreshold)
  fieldsToWrite.push(...positionFields);
else CheckFieldDelay.forEach((field) => {
  if (field.hasPassedTimeThreshold)
    fieldsToWrite.push(field);
});

if (fieldsToWrite.length > 0) {
  AllFields.forEach((field) => {
    if (field.hasChanged)
      fieldsToWrite.push(field);
  });
}
```

---

But yes, I agree that writing a good consistent regex that allows matching multiple fields is complex.

I have a thought about fixed-size arrays that should allow for good performance OP-side, but the question is what order should it be sorted in? Alphabetical sort doesn't really make a lot of sense since it'll split up position and heading, for example. But on the other side, trying to keep non-OP consumers of the line synced with order isn't really trivial if it's not a simple agreed-upon algorithm.

quisquous:
> But on the other side, trying to keep non-OP consumers of the line synced with order isn't really trivial if it's not a simple agreed-upon algorithm.

I think we could publish a list of fields and say "they are in this order".  Alternatively, I don't think there's a ton of human consumers of this, so heading being away from position (but ~mostly in the Add case, since Change will likely only be heading + positions) doesn't seem too awful to me.  Putting position fields first (and the rest alphabetical) also makes some sense, as this is likely the thing that most people care about.

In other words, I don't feel too strongly.  I think any of these options could be fine?

quisquous:
I also wanted to add that we could add any non-alphabetical order into the netlog_defs structure too, both for the logic and for printing into the log guide.

valarnin:
I think sorting is handled properly in https://github.com/OverlayPlugin/OverlayPlugin/pull/211

Once that's merged I'll update this PR to reflect the sorting logic.

valarnin:
I'm... not sure why that test is failing again. It doesn't fail locally and I have no changes that aren't committed/pushed.

```
± npm run test

> cactbot@0.28.19 test
> mocha

(node:6151) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time

  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬]

  6385 passing (1m)

± git status
On branch netlog-kvp
Your branch is up to date with 'origin/netlog-kvp'.

nothing to commit, working tree clean
```

quisquous:
I assume this is due to #5386? If you merge in main, do you see the error locally?

valarnin:
Yeah, it was due to two different PRs I guess (`de` and `cn`)

valarnin:
I have identified one additional issue with this related to capturing groups that I'd like to resolve before merging. Basically, capturing groups are dependent on the order of the sorted keys and not the order that repeating objects are passed in.

quisquous:
> I have identified one additional issue with this related to capturing groups that I'd like to resolve before merging. Basically, capturing groups are dependent on the order of the sorted keys and not the order that repeating objects are passed in.

Do non-named capturing groups get exposed anywhere?

valarnin:
> Do non-named capturing groups get exposed anywhere?

(the original issue was fixed with commit https://github.com/quisquous/cactbot/pull/5379/commits/ea8f54684cbbdcdbc89defc2de6e85fd7cde94b3, but to answer your question...)

No. Indexed capturing groups are available on the base array returned by `exec`. Triggers receive the sub-property `groups` of that array, which contains named capturing groups only.

image

resources/netlog_defs.ts
```diff
@@ -35,6 +35,13 @@ export type LogDefinition = {
   blankFields?: readonly number[];
   // This field and any field after will be treated as optional when creating capturing regexes.
   firstOptionalField: number | undefined;
+  // These fields are treated as repeatable fields
+  repeatingFields?: {
+    startingIndex: number;
+    label: string;
+    names: readonly string[];
+    sortFn?: (L: unknown, R: unknown) => number;
```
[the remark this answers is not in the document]
valarnin: This is technically narrow-able, but since the `L` and `R` parameters can be user-supplied in a user trigger, I felt that defaulting to `unknown` was a better option for safety reasons. If you'd rather handle that check elsewhere, or not enforce the check at all, that's fine.

As for casing, I'll probably just expand to full `left` and `right` instead of `L`, `R`?

ui/raidboss/data/00-misc/test.ts
```diff
@@ -325,10 +327,36 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: NetRegexes.echo({ line: 'cactbot test combatant cast.*?', capture: false }),
+      run: (data) => {
+        data.watchingForCast = true;
+      },
+    },
+    {
+      id: 'Test Combatant Cast',
+      type: 'CombatantMemory',
+      netRegex: NetRegexes.combatantMemory({
+        pair: [{ key: 'IsCasting1', value: '1' }, { key: 'CastBuffID', value: '.*?' }],
```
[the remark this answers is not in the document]
valarnin: I believe they should always show up on a line together, because of the nature of the underlying logic handling in the game client itself, but technically if something changed down the line in the game client and `CastBuffID` no longer always got set back to all `0`s in memory, this would no longer be the case.

resources/netlog_defs.ts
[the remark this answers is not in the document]
quisquous: Yeah, I think that's fine to be safe about input from users (with a comment about why).  Maybe it should throw an exception / print something to the console if it's not a string then? I just think it'd be better handled in the caller of the sort function rather than in every sort function.

Full `left` and `right` sounds fine to me.  `a` and `b` are also common param names to a sort function.

resources/regexes.ts
```diff
@@ -18,32 +20,86 @@ const defaultParams = <
...
 };
+type RepeatingFieldsMap<
+  TBase extends LogDefinitionTypes,
+  TKey extends RepeatingFieldsTypes = TBase extends RepeatingFieldsTypes ? TBase : never,
+> = {
+  // This syntax is a bit weird to make typescript compiler happy
+  [
+    label in RepeatingFieldsDefintions[TKey]['repeatingFields'][
+      'label'
+    ] as `${RepeatingFieldsDefintions[TKey]['repeatingFields']['label']}`
```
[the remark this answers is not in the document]
valarnin: Backticks are a string evaluation syntax just like normal JS/TS strings.

Basically, as far as I'm aware, you can't read the value of a type even when it's const, and extend that into defining the key/property on another type.

Consider assigning the value of a dynamic key to an object in JS:
```js
const someObjectDefinition = { 'key': 'pair' };
const value = 'anything';
const obj = {};
obj[someObjectDefinition.key] = value;
```

In typescript, this last operation isn't possible when defining a type directly (as opposed to assigning to an object). Instead, the code in `regexes.ts`/`netregexes.ts` does this equivalent logic:
```js
const someObjectDefinition = { 'key': 'pair' };
const value = 'anything';
const obj = {};
for (const keyDef in [someObjectDefinition.key]) {
  obj[keyDef] = value;
}
```

In other words, this is read as a string union with only one entry:
```ts
RepeatingFieldsDefintions[TKey]['repeatingFields']['label']
```
This is a loop over that one-element union:
```ts
label in ...
```
And this forces the value to be read as a constant entry so that the actual typedef gets a proper key (instead of an indexer):
```ts
 as `${RepeatingFieldsDefintions[TKey]['repeatingFields']['label']}`
```

---

It would be easier if this was allowed, but it's not valid TS syntax:
```ts
type RepeatingFieldsMap<
  TBase extends LogDefinitionTypes,
  TKey extends RepeatingFieldsTypes = TBase extends RepeatingFieldsTypes ? TBase : never,
> = {
  `${RepeatingFieldsDefintions[TKey]['repeatingFields']['label']}`: {
    [name in RepeatingFieldsDefintions[TKey]['repeatingFields']['names'][number]]:
      | string
      | string[];
  }[];
}
```

quisquous:
Sorry, one last thing, but could you add an example to regex_test.ts and netregex_test.ts as well, just to unit test that the fields (especially the pairs) come back as you'd expect with the right matching names?

resources/regexes.ts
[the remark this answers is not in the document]
quisquous: > And this forces the value to be read as a constant entry so that the actual typedef gets a proper key (instead of an indexer):

It is this part that I don't understand.  It seems to me like all the type-checking is the same with or without that.  If I remove it, the regex still typechecks `pair` vs incorrect `pair2` parameters.  Sorry, I'm just not following this one `as` bit here.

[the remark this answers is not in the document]
valarnin: I tested this and it wasn't working before. I'm not sure now if that was because of desync (due to running cactbot under WSL which causes some weird filewatcher issues) or if it's working now because of other changes that were made between when I originally tested and now.

Either way, I'll make this adjustment with the next push.

resources/netregexes.ts
```diff
@@ -36,32 +38,87 @@ const defaultParams = <
...
+  ]?: {
+    [name in RepeatingFieldsDefintions[TKey]['repeatingFields']['names'][number]]:
+      | string
+      | string[];
+  }[];
+};
+
+type RepeatingFieldsMapType<T extends LogDefinitionTypes> = T extends RepeatingFieldsTypes
+  ? RepeatingFieldsMap<T>
+  : Record<string, unknown>;
```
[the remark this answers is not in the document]
valarnin: This is a combined reply for both this comment and https://github.com/quisquous/cactbot/pull/5379#discussion_r1193194445.

I think you're maybe misunderstanding the underlying additional type here, which is what's causing the confusion. Consider the following more explicit definitions:

```ts
type RepeatingFieldsMapSubtype<
  TBase extends LogDefinitionTypes,
  TKey extends RepeatingFieldsTypes = TBase extends RepeatingFieldsTypes ? TBase : never,
> = TBase extends LogDefinitionTypes ? {
  [name in RepeatingFieldsDefinitions[TKey]['repeatingFields']['names'][number]]:
    | string
    | string[];
} : never;

type RepeatingFieldsMap<
  TBase extends LogDefinitionTypes,
  TKey extends RepeatingFieldsTypes = TBase extends RepeatingFieldsTypes ? TBase : never,
> = TBase extends LogDefinitionTypes ? {
  [
    label in RepeatingFieldsDefinitions[TKey]['repeatingFields']['label']
  ]?: RepeatingFieldsMapSubtype<TKey>[];
} : never;

// ...

    let fieldValue:
      string | string[] | RepeatingFieldsMapSubtype<T>[] | undefined = params[fieldName];

    if (fields[keyStr]?.repeating) {
      if (Array.isArray(fieldValue)) {
        let repeatingArray: RepeatingFieldsMapSubtype<T>[] =
          fieldValue as unknown as RepeatingFieldsMapSubtype<T>[];
```

This requires an even more ugly cast, because without any cast:

image
image
image

As for why `RepeatingFieldsMapSubtype<T>[]` is not in the typedef, I have no clue. It won't recognize any array type added to the `fieldValue` typing beyond `string[]`.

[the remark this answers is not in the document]
valarnin: I have refactored the logic for these types in a way that hopefully makes more sense.

resources/netlog_defs.ts
```diff
@@ -1014,17 +1024,41 @@ const latestLogDefinitions = {
...
     canAnonymize: true,
     firstOptionalField: 4,
     playerIds: {
       3: null,
     },
+    repeatingFields: {
+      startingIndex: 4,
+      label: 'pair',
+      names: ['key', 'value'],
+      sortFn: (left: unknown, right: unknown): number => {
```
quisquous, line 1038: Sorry to restate this, but just trying to reiterate all the open issues on my mind.

Looking for these changes:
* `sortFn: (left: string, right: string): number` as a type in the definition
* ideally then `sortFn: (left: string, right: string) => left.localeCompare(right),` here?
* this helper code to go from unknown => string in the caller in netregexes.ts / regexes.ts (maybe some shared helper function somewhere) which itself can call the string `sortFn` at the bottom

That way future `sortFn` can just write a very trivial sort function rather than having to re-do this helper goop eveyrwhere.

resources/regexes.ts
```diff
@@ -131,26 +203,88 @@ const parseHelper = <T extends LogDefinitionTypes>(
...
+              // All capturing groups have their index appended
+              key + origRepeatingArray.indexOf(rep).toString(),
+              val,
+              defaultFieldValue,
+            );
+          });
+
+          if (fieldRegex.length > 0) {
+            // Prepend a match for any number of fields before matching this one.
+            // TODO: Should we scope this as the following instead?
```
quisquous, line 264: Yeah, I think probably we should.  Probably it's the case that no field name is going to match a field value, but that would be a really hairy bug to sort out later.

resources/regexes.ts
```diff
@@ -131,26 +203,88 @@ const parseHelper = <T extends LogDefinitionTypes>(
...
+
+      repeatingArray.forEach((rep) => {
+        if (typeof rep === 'object' && rep !== null && Object.keys(rep).length > 0) {
+          let fieldRegex = '';
+          // Rather than looping over the keys defined on the object,
+          // loop over the base type def's keys. This enforces the correct order.
+          fields[keyStr]?.repeatingKeys?.forEach((key) => {
+            fieldRegex += separator;
+            if (!(key in rep)) {
+              // If we don't have a value for this key, insert a placeholder
```
quisquous, line 232: When would this happen? Sorry that I'm just not understanding this case.

resources/regexes.ts
```diff
@@ -131,26 +203,88 @@ const parseHelper = <T extends LogDefinitionTypes>(
...
+                // All capturing groups have their index appended
+                key + origRepeatingArray.indexOf(rep).toString(),
+                defaultFieldValue,
+                defaultFieldValue,
+              );
+              return;
+            }
+            // This is an ugly cast. There's probably a better way of narrowing this type but
+            // tsc doesn't like the dozen or so variations I've tried.
+            const val = rep[key as keyof typeof rep];
```
quisquous, line 244:
```suggestion
            const anonRep: { [name: string ]: string | string[] } = rep;
            const val = anonRep[key];
```

This maybe? I'm sure I could create some extra types to not redundantly type `string | string[]` here, but this seemed fine.

ui/raidboss/data/00-misc/test.ts
```diff
@@ -325,10 +327,36 @@ const triggerSet: TriggerSet<Data> = {
         },
       },
     },
+    {
+      id: 'Test Combatant Cast Enable',
```
quisquous, line 331: Just reiterating https://github.com/quisquous/cactbot/pull/5379#issuecomment-1546997059, but could you add a unit test for these lines as well?

resources/netregexes.ts
```diff
@@ -136,21 +193,77 @@ const parseHelper = <T extends LogDefinitionTypes>(
...
-        fieldValue,
-      ) +
-        separator;
+    const fieldName = value.field;
+    const defaultFieldValue = value.value?.toString() ?? matchDefault;
+    let fieldValue: string | string[] | Record<string, unknown>[] | undefined = params[fieldName];
+
+    if (fields[keyStr]?.repeating) {
+      if (Array.isArray(fieldValue)) {
+        let repeatingArray: unknown[] = fieldValue;
```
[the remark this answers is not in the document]
quisquous: Thanks, this is easier to read and more type safe.

resources/netregexes.ts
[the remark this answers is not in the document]
quisquous: Thanks! I appreciate all the changes.

quisquous (commented):
Sorry for the delay in getting back to this.  This required some brain and time to review and I've been pretty short on that the past week.

resources/regexes.ts
quisquous: Hmm, I see.  This was unclear to me.  I think `.*` would be incorrect here, as it would match across multiple fields.  (I mean, arguably we could detect/fix this and it's already an issue for other regexes potentially.)  Can you leave a comment that this is why this case exists, and maybe also put a test case for this?  This was non-obvious to me.

> (Though this raises a question of "maybe we should just optionally match against all possible fields if it's a capture: true regex, even if that results in a really big and potentially slow regex?)

For me the answer is no.  I think if you want all of them, you just pull the line apart yourself.  In terms of match group names, I think it'd be hard to number them in a consistent way.

resources/netlog_defs.ts
quisquous: Hmm, ok.  I guess I assumed we'd always have at least some string keys to sort on here for anything we'd want to match against, but maybe that's not true.  I guess we only have one example so I shouldn't try to oversimplify yet.

resources/regexes.ts
valarnin: I've got a local version right now that given this NetRegexes:
```ts
      NetRegexes.combatantMemory({
        pair: [{ key: 'IsCasting1', value: '1' }, { key: 'CastBuffID', value: '.*?' }],
      })
```
Produces this regex:
```ts
/^(?<type>261)\|(?<timestamp>[^|]*)\|(?<change>[^|]*)\|(?<id>[^|]*)\|(?:(?<AggressionStatuskey>AggressionStatus)\|(?<AggressionStatusvalue>[^|]*)\|)?(?:(?<BNpcIDkey>BNpcID)\|(?<BNpcIDvalue>[^|]*)\|)?(?:(?<BNpcNameIDkey>BNpcNameID)\|(?<BNpcNameIDvalue>[^|]*)\|)?(?:(?<CastBuffIDkey>CastBuffID)\|(?<CastBuffIDvalue>.*?)\|)(?:(?<CastDurationCurrentkey>CastDurationCurrent)\|(?<CastDurationCurrentvalue>[^|]*)\|)?(?:(?<CastDurationMaxkey>CastDurationMax)\|(?<CastDurationMaxvalue>[^|]*)\|)?(?:(?<CastTargetIDkey>CastTargetID)\|(?<CastTargetIDvalue>[^|]*)\|)?(?:(?<CurrentCPkey>CurrentCP)\|(?<CurrentCPvalue>[^|]*)\|)?(?:(?<CurrentGPkey>CurrentGP)\|(?<CurrentGPvalue>[^|]*)\|)?(?:(?<CurrentHPkey>CurrentHP)\|(?<CurrentHPvalue>[^|]*)\|)?(?:(?<CurrentMPkey>CurrentMP)\|(?<CurrentMPvalue>[^|]*)\|)?(?:(?<CurrentWorldIDkey>CurrentWorldID)\|(?<CurrentWorldIDvalue>[^|]*)\|)?(?:(?<Distancekey>Distance)\|(?<Distancevalue>[^|]*)\|)?(?:(?<EffectiveDistancekey>EffectiveDistance)\|(?<EffectiveDistancevalue>[^|]*)\|)?(?:(?<Headingkey>Heading)\|(?<Headingvalue>[^|]*)\|)?(?:(?<IDkey>ID)\|(?<IDvalue>[^|]*)\|)?(?:(?<IsCasting1key>IsCasting1)\|(?<IsCasting1value>1)\|)(?:(?<IsCasting2key>IsCasting2)\|(?<IsCasting2value>[^|]*)\|)?(?:(?<IsTargetablekey>IsTargetable)\|(?<IsTargetablevalue>[^|]*)\|)?(?:(?<Jobkey>Job)\|(?<Jobvalue>[^|]*)\|)?(?:(?<Levelkey>Level)\|(?<Levelvalue>[^|]*)\|)?(?:(?<MaxCPkey>MaxCP)\|(?<MaxCPvalue>[^|]*)\|)?(?:(?<MaxGPkey>MaxGP)\|(?<MaxGPvalue>[^|]*)\|)?(?:(?<MaxHPkey>MaxHP)\|(?<MaxHPvalue>[^|]*)\|)?(?:(?<MaxMPkey>MaxMP)\|(?<MaxMPvalue>[^|]*)\|)?(?:(?<ModelStatuskey>ModelStatus)\|(?<ModelStatusvalue>[^|]*)\|)?(?:(?<MonsterTypekey>MonsterType)\|(?<MonsterTypevalue>[^|]*)\|)?(?:(?<NPCTargetIDkey>NPCTargetID)\|(?<NPCTargetIDvalue>[^|]*)\|)?(?:(?<Namekey>Name)\|(?<Namevalue>[^|]*)\|)?(?:(?<OwnerIDkey>OwnerID)\|(?<OwnerIDvalue>[^|]*)\|)?(?:(?<PCTargetIDkey>PCTargetID)\|(?<PCTargetIDvalue>[^|]*)\|)?(?:(?<PartyTypekey>PartyType)\|(?<PartyTypevalue>[^|]*)\|)?(?:(?<PosXkey>PosX)\|(?<PosXvalue>[^|]*)\|)?(?:(?<PosYkey>PosY)\|(?<PosYvalue>[^|]*)\|)?(?:(?<PosZkey>PosZ)\|(?<PosZvalue>[^|]*)\|)?(?:(?<Radiuskey>Radius)\|(?<Radiusvalue>[^|]*)\|)?(?:(?<Statuskey>Status)\|(?<Statusvalue>[^|]*)\|)?(?:(?<TargetIDkey>TargetID)\|(?<TargetIDvalue>[^|]*)\|)?(?:(?<TransformationIdkey>TransformationId)\|(?<TransformationIdvalue>[^|]*)\|)?(?:(?<Typekey>Type)\|(?<Typevalue>[^|]*)\|)?(?:(?<WeaponIdkey>WeaponId)\|(?<WeaponIdvalue>[^|]*)\|)?(?:(?<WorldIDkey>WorldID)\|(?<WorldIDvalue>[^|]*)\|)?(?:(?<WorldNamekey>WorldName)\|(?<WorldNamevalue>[^|]*)\|)?/i
```

Example screenshot showing regex with highlighting and matching regex:
image

Example screenshot showing what `matches` would be in the context of a cactbot trigger function:
image

I think this makes a lot more sense, and because there's no `.*?` this should also be significantly faster than the solution currently in this PR, since the regex engine should early out each optional capturing group when the `key` part of the line fails to match.

The only thing that doesn't sit well with me is the naming of the capturing groups, `IsCasting1key` and `IsCasting1value` might be a bit confusing and maybe they should be `IsCasting1__key` and `IsCasting1__value` instead?

(this also follows the `capture: true`/`capture: false` logic, so if `capture: false` it won't insert capturing groups)

quisquous: Ooooh, nice.  One observation from me is why do we need a key/value pair for the keys, when the key/value pair for the value has the key already in it.  In other words, `IsCasting: '1',` seems sufficient, and you don't need `IsCastingKey: 'IsCasting', IsCastingValue: '1',`.

valarnin: My concern is conflicting fields, e.g. there's a `type` by default in all lines (`261` in this case), and there's a `type` on `PluginCombatantState` (which this has renamed to `Type` but still).

So I'd be fine with e.g. `matches['pair.Type']` maybe? Or `matches.pair_Type`, or some other variation on that.

valarnin: (but yes, I agree that having the key as a property on `matches` is redundant)

quisquous: `pairType`  is a bit more camelcase-y on the eyes and wouldn't need the square brackets to access.

valarnin: Alright, this look good to you? I'll see if I can get some actual type hinting on matches as well but that's a bit more iffy.

```ts
/^(?<type>261)\|(?<timestamp>[^|]*)\|(?<change>[^|]*)\|(?<id>[^|]*)\|(?:AggressionStatus\|(?<pairAggressionStatus>[^|]*)\|)?(?:BNpcID\|(?<pairBNpcID>[^|]*)\|)?(?:BNpcNameID\|(?<pairBNpcNameID>[^|]*)\|)?(?:CastBuffID\|(?<pairCastBuffID>.*?)\|)(?:CastDurationCurrent\|(?<pairCastDurationCurrent>[^|]*)\|)?(?:CastDurationMax\|(?<pairCastDurationMax>[^|]*)\|)?(?:CastTargetID\|(?<pairCastTargetID>[^|]*)\|)?(?:CurrentCP\|(?<pairCurrentCP>[^|]*)\|)?(?:CurrentGP\|(?<pairCurrentGP>[^|]*)\|)?(?:CurrentHP\|(?<pairCurrentHP>[^|]*)\|)?(?:CurrentMP\|(?<pairCurrentMP>[^|]*)\|)?(?:CurrentWorldID\|(?<pairCurrentWorldID>[^|]*)\|)?(?:Distance\|(?<pairDistance>[^|]*)\|)?(?:EffectiveDistance\|(?<pairEffectiveDistance>[^|]*)\|)?(?:Heading\|(?<pairHeading>[^|]*)\|)?(?:ID\|(?<pairID>[^|]*)\|)?(?:IsCasting1\|(?<pairIsCasting1>1)\|)(?:IsCasting2\|(?<pairIsCasting2>[^|]*)\|)?(?:IsTargetable\|(?<pairIsTargetable>[^|]*)\|)?(?:Job\|(?<pairJob>[^|]*)\|)?(?:Level\|(?<pairLevel>[^|]*)\|)?(?:MaxCP\|(?<pairMaxCP>[^|]*)\|)?(?:MaxGP\|(?<pairMaxGP>[^|]*)\|)?(?:MaxHP\|(?<pairMaxHP>[^|]*)\|)?(?:MaxMP\|(?<pairMaxMP>[^|]*)\|)?(?:ModelStatus\|(?<pairModelStatus>[^|]*)\|)?(?:MonsterType\|(?<pairMonsterType>[^|]*)\|)?(?:NPCTargetID\|(?<pairNPCTargetID>[^|]*)\|)?(?:Name\|(?<pairName>[^|]*)\|)?(?:OwnerID\|(?<pairOwnerID>[^|]*)\|)?(?:PCTargetID\|(?<pairPCTargetID>[^|]*)\|)?(?:PartyType\|(?<pairPartyType>[^|]*)\|)?(?:PosX\|(?<pairPosX>[^|]*)\|)?(?:PosY\|(?<pairPosY>[^|]*)\|)?(?:PosZ\|(?<pairPosZ>[^|]*)\|)?(?:Radius\|(?<pairRadius>[^|]*)\|)?(?:Status\|(?<pairStatus>[^|]*)\|)?(?:TargetID\|(?<pairTargetID>[^|]*)\|)?(?:TransformationId\|(?<pairTransformationId>[^|]*)\|)?(?:Type\|(?<pairType>[^|]*)\|)?(?:WeaponId\|(?<pairWeaponId>[^|]*)\|)?(?:WorldID\|(?<pairWorldID>[^|]*)\|)?(?:WorldName\|(?<pairWorldName>[^|]*)\|)?/i
```

image

quisquous: Yeah, seems reasonable.  Seems wordy, but it's definitely better than trying to special-case `type` or having `type` vs `Type`, and I can't come up with anything better.

valarnin: Success! Just need to do a bit more cleanup and this'll be ready.

image

ui/raidboss/data/00-misc/test.ts
valarnin: Adding tests is giving me some hassle because of the reverse checking, where `regexUtil` checks the returned capture groups to ensure they're valid params or something along those lines.

```
  1) netregex tests
       combatantMemory:
     Error: CombatantMemory: invalid parameter 'pairBNpcID'.  Valid params: ["capture","type","timestamp","change","id","pair"]
      at Function.validateParams (file:///mnt/c/Users/valarnin/Source/cactbot/resources/regexes.ts:694:15)
      at parseHelper (file:///mnt/c/Users/valarnin/Source/cactbot/resources/netregexes.ts:156:11)
      at buildRegex (file:///mnt/c/Users/valarnin/Source/cactbot/resources/netregexes.ts:336:10)
      at Function.combatantMemory (file:///mnt/c/Users/valarnin/Source/cactbot/resources/netregexes.ts:597:12)
      at file:///mnt/c/Users/valarnin/Source/cactbot/test/unittests/netregex_test.ts:415:63
      at default (file:///mnt/c/Users/valarnin/Source/cactbot/test/helper/regex_util.ts:61:34)
      at Context.<anonymous> (file:///mnt/c/Users/valarnin/Source/cactbot/test/unittests/netregex_test.ts:415:5)
      at processImmediate (node:internal/timers:471:21)
```

I'm not sure if it's really feasible to re-jigger the repeating fields defs to not require the `pair` array, I guess I can investigate that tomorrow.

quisquous: Yeah, feel free to special case anything here that needs to be special-cased and leave a TODO or some comments.  Mostly I just wanted some examples that exercise this code path.

test/unittests/netregex_test.ts
```diff
@@ -406,4 +406,33 @@ describe('netregex tests', () => {
...
+  it('combatantMemory', () => {
+    const lines = [
+      '261|2023-05-26T21:37:40.5600000-04:00|Add|40008953|BNpcID|3F5A|BNpcNameID|304E|CastTargetID|E0000000|CurrentMP|10000|CurrentWorldID|65535|Heading|-3.1416|Level|90|MaxHP|69200|MaxMP|10000|ModelStatus|18432|Name|Golbez\'s Shadow|NPCTargetID|E0000000|PosX|100.0000|PosY|100.0000|PosZ|0.0300|Radius|7.5000|Type|2|WorldID|65535|9d9028a8e087e4c3',
+      '261|2023-05-26T21:39:41.2920000-04:00|Change|10001234|CurrentMP|2400|Heading|-2.3613|2f5ff0a91385050a',
+      '261|2023-05-26T21:39:42.7380000-04:00|Remove|40008AA0|f4b30f181245b5da',
+    ] as const;
+    // TODO: regexCaptureTest doesn't handle the repeating fields well,
+    // so don't run it for this test
+    const matches = lines[0].match(NetRegexes.combatantMemory())?.groups;
+    console.log(matches);
```
quisquous, line 418:
```suggestion
```

types/net_matches.d.ts
```diff
@@ -1,11 +1,31 @@
...
 export type Matches = { [s: string]: string };
 export type ToMatches<Fields> = Record<keyof Fields, string>;
 
+export type RepeatingFieldsMatchesExtract<type extends RepeatingFieldsTypes> = {
+  [
+    key in RepeatingFieldsDefinitions[type]['repeatingFields']['possibleKeys'][
+      number
+    ] as `${RepeatingFieldsDefinitions[type]['repeatingFields']['label']}${key}`
+  ]?: string;
+};
```
quisquous, line 15: Whoa, generating property names like this is neat.  I hadn't seen this before.

quisquous (commented):
Is this ready to be merged? Just making sure as we had a bunch of discussion.  I think it looks good to me, but didn't know if you had any other things planned.

valarnin:
> Is this ready to be merged? Just making sure as we had a bunch of discussion. I think it looks good to me, but didn't know if you had any other things planned.

No, I think this is as far as I can really take this for now. I might revisit later but not for a few weeks at least.

PR #5387, opened 2023-05-09
raidemulator: Refactor `CombatantState` object

valarnin:
This PR is in preparation for `combatantMemory` lines.

Because the state tracker will now need to track all possible `PluginCombatantState` properties, it no longer makes sense to track a limited set for most other types of lines and a full state for those specific lines.

As such, the custom properties on `CombatantState` have been removed in favor of just implementing the `PluginCombatantState` interface with the carried over addition of a `targetable` property to maintain functionality.

Additionally, logic for constructing and cloning states was greatly simplified to use `Object.assign` instead.

Some duplicate properties were removed from `Combatant` in favor of the tracked state entries in `CombatantState`, and associated locations were updated to instead read the initial combatant state's version of those properties. This should result in more correct behavior when `getCombatants` is used, because name changes are tracked properly now for example.

The vast majority of the added lines are in a new `resources/world_id.ts` file, required to translate `ID` <=> `Name`. Excluding that file, the initial commit for this PR contains 166 additions and 205 deletions.

This PR needs more testing before being merged.

- I have only tested with US DC names, for example, so I'm not sure if world name lookup will work as expected for `CN`/`KO`/`JP` servers.
- I have not tested for logs that may have weird values for some entries that would potentially result in `parseFloat` returning `NaN`.
- I have only tested recent content (DSRU, P8S, and TOP). Some older content should probably be tested.


ui/raidboss/emulator/data/CombatantState.ts
```diff
@@ -1,76 +1,85 @@
...
-    maxMp: number,
-  ) {
-    this.posX = posX;
-    this.posY = posY;
-    this.posZ = posZ;
-    this.heading = heading;
+  constructor(props: Partial<PluginCombatantState>, targetable: boolean) {
+    Object.assign(this, props);
+
+    // Force these values to something sane in case they're not in `props`
```
quisquous, line 36:
```suggestion
    // Force these values to defaults in case they're not in `props`
```

Should we do this for other fields as well? I guess I don't understand why these get defaulted here but others are `?? 0` elsewhere.

valarnin: These are defaulted here because they're not optional per the `PluginCombatantState` interface itself.

`CombatantState` could instead implement `Partial<PluginCombatantState>`, but that's just kicking this validation responsibility downstream to other locations, leading to duplicate logic.

If other locations are checking these specific values with `?? 0`, that's something that could be cleaned up in those locations.

As for why `PluginCombatantState` has these fields as required, and other fields as optional, that's because I believe that these values were required to always return for `getCombatants` calls. If instead all fields on `PluginCombatantState` should be made required now that it's expected that they should return, then that change can be made, but it's going to make this PR a bit bigger.

resources/world_id.ts
```diff
@@ -0,0 +1,7611 @@
+// TODO: Generate this automatically from `world.csv`?
```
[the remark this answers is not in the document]
valarnin: I'm not as familiar with the logic/code that builds these normally, so I didn't try to tackle this as part of this PR.

resources/world_id.ts
```diff
@@ -0,0 +1,7611 @@
...
+
+export type World = {
+  ID: number;
```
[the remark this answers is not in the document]
valarnin: I used the same keys as the CSV headers. They could be camelCase'd I guess? That's just an additional transform step during the conversion from CSV.

ui/raidboss/emulator/data/CombatantState.ts
quisquous: Ah, ok.  "compatibility with `PluginCombatantState`" is enough of a reason for me, thanks.

valarnin:
I've identified an annoying issue with state tracking that is going to require rewriting the logic for the state calculation a bit.

Basically, because `ID`, `Name`, `Job`, and `Level` are no longer required to be calculated at a top level, they're not present on states between the initial state and the state which they first are calculated on (exclusive), that fails when displaying party information as well as causing issues if `getCombatants` is called for some reason between those two states during analysis or playback.

valarnin:
Issues are resolved, the automatic generator for `world_id.ts` has been added. Ready for re-review if needed.

quisquous:
This all looks good to me, thanks.  Do you feel like it still needs more testing or is this good to land?

valarnin:
There was a bug fix that didn't get committed properly, apparently. More jank with my weird setup using WSL 🙃 

I think this is good to go now.

PR #5395, opened 2023-05-20
raidboss: updates to UCOB

wexxlee:
Added some additional triggers for tank cleaves/busters, some minor cleanup, and new triggers for QM, Blackfire, Heavensfall, and Grand Octet.  Also added clockwise/counterclockwise to resources/outputs, since it tends to come up reasonably often.


ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -721,23 +754,50 @@ const triggerSet: TriggerSet<Data> = {
...
+        thunderOnYou: {
           en: 'Thunder on YOU',
           de: 'Blitz auf DIR',
           fr: 'Foudre sur VOUS',
           ja: '自分にサンダー',
           cn: '雷点名',
           ko: '나에게 번개',
         },
+        thunderOnOthers: {
+          en: 'Thunder on ${players}',
```
quisquous, line 799: Sorry for skepticism, but I'm just not sure why this is helpful information to know if you don't have thunder.  Mostly just don't understand what the collection is for here.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -45,6 +51,26 @@ const resetTrio = (data: Data, trio: string) => {
   data.megaStack = [];
 };
 
+const dirs = ['dirN', 'dirNE', 'dirE', 'dirSE', 'dirS', 'dirSW', 'dirW', 'dirNW'];
```
quisquous, line 54: I think `dirs` is a little too ambiguous of a name for a top level variable.  This is really a map of dir numbers to dir names, so `dirNames` is fine-ish, but `dirNumToName` maybe even more clear elsewhere.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1485,6 +1646,57 @@ const triggerSet: TriggerSet<Data> = {
...
+        data.combatantData = [];
+        data.combatantData = (await callOverlayHandler({
+          call: 'getCombatants',
+          ids: [data.trioSourceIds.nael, data.trioSourceIds.bahamut, data.trioSourceIds.twin],
+        })).combatants;
+      },
+      alertText: (data, _matches, output) => {
+        // Bosses line up adjacent to one another, but don't necessarily have discrete directional positions (based on 8Dir scale).
+        // But we can calculate their position as an angle (relative to circular arena): 0 = N, 90 = E, 180 = S, 270 = W, etc.
+        // Set initial angles to -1 to avoid issues with nullish checks, as 0 is a valid angle and a nullish value.
```
quisquous, line 1667: Could you set these to undefined and check against undefined explicitly on line 1682 rather than using a magic sentinel value here?

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1485,6 +1646,57 @@ const triggerSet: TriggerSet<Data> = {
...
+          if (isClockwise(naelAngle, bahamutAngle))
+            naelPos = isClockwise(naelAngle, twinAngle) ? 'left' : 'middle';
+          else
+            naelPos = isClockwise(naelAngle, twinAngle) ? 'middle' : 'right';
+        }
+        return output.naelPosition!({ dir: output[naelPos]!() });
+      },
+      outputStrings: {
+        naelPosition: {
+          en: 'Nael is ${dir}',
```
quisquous, line 1692:
```suggestion
          en: '${dir} Nael',
```

Just rewording to put the important word first.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1552,6 +1764,97 @@ const triggerSet: TriggerSet<Data> = {
...
+    // For Grand Octet:
+    // After bosses and dragons start spawning, there's no clear log line we can trigger off of to find bosses' position
+    // until it's effectively too late.  The best way to do this seems to be to fire the trigger
+    // with a delay when Bahamut uses Grand Octet before all 3 bosses jump.
+    {
+      id: 'UCU Grand Octet Run & Rotate',
+      type: 'Ability',
+      // Grab mob position data after dragons/bosses are positioned
+      netRegex: { id: '26E7', source: 'Bahamut Prime', capture: false },
+      delaySeconds: 4.8,
```
quisquous, line 1776: Where did this delay come from?

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1552,6 +1764,97 @@ const triggerSet: TriggerSet<Data> = {
...
+
+        for (const mob of data.combatantData) {
+          const mobDirIdx = Math.round(4 - 4 * Math.atan2(mob.PosX, mob.PosY) / Math.PI) % 8;
+          if (mob.ID === data.trioSourceIds.nael)
+            naelDirIdx = mobDirIdx;
+          else if (mob.ID === data.trioSourceIds.bahamut)
+            bahaDirIdx = mobDirIdx;
+        }
+
+        if (!naelDirIdx || !bahaDirIdx)
```
quisquous, line 1798: Can this check against undefined instead of nullish here?

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1552,6 +1764,97 @@ const triggerSet: TriggerSet<Data> = {
...
+          call: 'getCombatants',
+          ids: [data.trioSourceIds.nael, data.trioSourceIds.bahamut],
+        })).combatants;
+      },
+      alertText: (data, _matches, output) => {
+        let naelDirIdx;
+        let bahaDirIdx;
+
+        for (const mob of data.combatantData) {
+          const mobDirIdx = Math.round(4 - 4 * Math.atan2(mob.PosX, mob.PosY) / Math.PI) % 8;
```
quisquous, line 1791: Is this `positionTo8DirOutput`?

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1552,6 +1764,97 @@ const triggerSet: TriggerSet<Data> = {
...
+          return;
+
+        // If Bahamut spaws on a cardinal, the party goes opposite and rotates counter-clockwise; if intercardinal, clockwise.
+        // If Nael is directly opposite Bahamut, the party instead starts one directional position over (same as the rotation direction)
+        // http://clees.me/guides/ucob/
+        let rotationIdxModifier; // this is used to modify the party starting spot in directions[] if Nael is opposite Bahamut
+        let rotationPath;
+
+        const bahaDirStr = dirs[bahaDirIdx];
+        if (!bahaDirStr)
```
quisquous, line 1808:
```suggestion
        if (bahaDirStr === undefined)
```

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1552,6 +1764,97 @@ const triggerSet: TriggerSet<Data> = {
...
+          // if this pushes partyStartIdx beyond the array boundary, wrap around
+          if (partyStartIdx === -1) {
+            partyStartIdx = 7;
+          } else if (partyStartIdx === 8) {
+            partyStartIdx = 0;
+          }
+        }
+        const partyStartDir = dirs[partyStartIdx] ?? 'unknown';
+
+        if (!partyStartDir || !rotationPath)
```
quisquous, line 1832: Can you check undefined here? (Or, can you check over the whole PR too and if there's places where you're looking for undefined, can you check that explicitly instead of a nullish check?)

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
```diff
@@ -1364,6 +1432,99 @@ const triggerSet: TriggerSet<Data> = {
...
+      alertText: (_data, matches, output) => {
+        // Bosses jump, and dive placement is locked once Bahamut starts casting.
+        // Position data is always updated by now, so need to rely on combatant data from OP.
+        // Bahamut will always be on an exact cardinal/intercardinal (w/Nael & Twin on either side)
+        const diveDir = positionTo8DirOutput(parseFloat(matches.x), parseFloat(matches.y));
+        return output.dive!({ dir: output[diveDir]!() });
+      },
+      outputStrings: {
+        dive: {
+          en: 'Diving from ${dir}',
```
quisquous, line 1449:
```suggestion
          en: '${dir} Dive',
```

Just putting the important information first.

quisquous (commented):
Thanks for all these updates! I've got a few small comments here and there, but nothing major.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
wexxlee: Quickly knowing who has thunder (especially when those debuffs overlap with other Nael mechanics during quotes 3 & 4) allows you to identify on screen if those players have recognized that fact and moved away -- or, if that player also has doom, to recognize that they will be moving to cleanse (requiring the rest of the party to possibly adjust position to accommodate).  I found it helpful during prog, but ymmv, and I'm not wedded to it.

quisquous: Yeah, I am mostly worried about noise here especially with dooms going off too.  I do agree that knowing that the main tank has thunder here goes a long way and that was something that we would call out manually.  Ok, I think I've convinced myself this would be useful; I was mostly thinking about the first set.

My only other request here is, can you make two output strings for "single player thunder" and "multiple player thunder" (e.g. ${player1} ${player2}) to match other triggers? I know it looks unnecessary, but sometimes different languages might want to write this out differently, and having player name output be its own thing will make it easier for the future if we ever want to add settings for player => job.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
wexxlee: `positionTo8DirOutput` returns the direction name from `dirNumToName`, but we actually need the array index (or, more precisely, the direction number), because we use the direction number in `naelDirIdx` and `bahaDirIdx` to calculate the safe spot (based on opposite-Bahamut, possible adjustment for Nael).  
If we wanted to use `positionTo8DirOutput` here instead of directly calling `Math.atan2`, we could grab the direction name, but then we'd have to do an `indexOf` lookup to get back to the direction number, which feels clunky.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
wexxlee: This is the amount of time it takes after Bahamut Prime uses Grand Octet (`26E7`) for the trio to appear mid, disappear, and for Bahamut and Nael to then respawn on the perimeter (which gives the position data we need to calculate both the direction to run and rotation - plus a little safety margin for potential network lag.  Tested extensively during prog and I (and others) found this threshold to be reliable.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
wexxlee: Just to make sure I'm understanding the suggestion: right now, there's an existing outputString for `thunderOnYou` (which will trigger if you have thunder), and this PR adds `thunderOnOthers` (which will trigger if you don't have thunder, because two other party members do).  So I think your ask is to just update `thunderOnOthers` to call out player1 and player2 separately, as opposed to being a single passed string?  I'll update it to that, but let me know if I've misunderstood the ask.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
wexxlee: As an aside, given that we seem to use relatively common variants of positionTo8Dir now in various fights, perhaps these (and, optionally, their 4Dir and 16Dir variants?) would be good candidates to add to a function library, so we're not repeating this each time?  If you agree, I'd be happy to submit a PR, but not sure where would be the right place for them to live.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
quisquous: Yeah, keep `thunderOnYou`.  Separate out `thunderOnOthers` into `thunderOnOther` (1 player) and `thunderOnOthers` (2 players) and have `thunderOnOthers` be `Thunder on ${player1}, ${player2}` rather than combined in code.

wexxlee: OK, that's how I pushed it in the latest commit. 😄   Should be good to go.

ui/raidboss/data/04-sb/ultimate/unending_coil_ultimate.ts
quisquous: Maybe it'd be better for `positionTo8DirOutput` to return a number, and you could look up from there and use it in all places? Personally I think it's fine to re-do the number=>string lookup, but the math part would be nice to keep in one spot.

Agreed that it'd be good to put these in common locations.  I think a good common location could just be `util.ts`.  We could make some new functions for "common raidboss functions" or "raidboss helpers" too (since watch combatants might be a better fit in there than util).  One problem is that a couple of fights have conflicting ideas of what 0 means (there's like ~2-3 that want to have 0=NW and most are 0=N) and so those would need to be refactored/renamed as well.

wexxlee: Sounds good.  Let's pause this PR for now, and I'll open up a new one to add some functionality to util.  If & once you're good with it and land it, I can revise this PR to use the new functions and then do a follow-on PR for the other fights.

wexxlee:
Marking as draft for now to work on landing some position-related stuff to `util.ts`.

resources/util.ts
```diff
@@ -227,12 +227,12 @@ export type DirectionOutputIntercard =
   | 'dirNW'
   | 'unknown';
 
-export type DirectionOutput16Map = { [key in DirectionOutput16]: OutputStrings };
-export type DirectionOutput8Map = { [key in DirectionOutput8]: OutputStrings };
-export type DirectionOutputCardinalMap = { [key in DirectionOutputCardinal]: OutputStrings };
-export type DirectionOutputIntercardMap = { [key in DirectionOutputIntercard]: OutputStrings };
+export type DirectionOutput16Map = { [outputString: string]: OutputStrings };
```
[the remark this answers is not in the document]
wexxlee: Yeah... I liked the idea of explicit typing, but it didn't hold up when it came time to actually use them in the `outputStrings{}` in triggers.  Alas.

I do think it's worthwhile still to keep the explicit typing on the lookup arrays (e.g. `output8Dir: DirectionOutput8[]`), but I agree it makes sense to remove them on the outputStrings objects like you suggest.  If that works for you, this is good to go.

quisquous:
Thanks for all the changes and the helper functions!

PR #5404, opened 2023-05-23
raidboss: Initial ZurvanEX package

JLGarber:
I thought I would have more time to test things in-game and then Omega said no.

There's probably a better way to handle the Metal Cutter spam, but suppressing for 10 seconds seemed reasonably okay. Other than that, everything should be straightforward.

### DIFF
--- a/ui/raidboss/data/03-hw/trial/zurvan-ex.ts
+++ b/ui/raidboss/data/03-hw/trial/zurvan-ex.ts
@@ -1,11 +1,14 @@
 import Conditions from '../../../../../resources/conditions';
+import Outputs from '../../../../../resources/outputs';
 import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
 export interface Data extends RaidbossData {
-  mainTank?: string;
+  flameTarget?: string;
+  tetherBuddy?: string;
+  infiniteElement?: 'fire' | 'ice';
 }
 
 const triggerSet: TriggerSet<Data> = {
@@ -17,19 +20,11 @@ const triggerSet: TriggerSet<Data> = {
       id: 'ZurvanEX Metal Cutter',
       regex: /Metal Cutter/,
       beforeSeconds: 4,
-      suppressSeconds: 5,
+      suppressSeconds: 10,
       response: Responses.tankCleave(),
     }
   ],
   triggers: [
-    {
-      id: 'ZurvanEX Main Tank',
-      type: 'Ability',
-      netRegex: { id: '368', source: 'Zurvan' },
-      // We make this conditional to avoid constant noise in the raid emulator.
-      condition: (data, matches) => data.mainTank !== matches.target,
-      run: (data, matches) => data.mainTank = matches.target,
-    },
     {
       id: 'ZurvanEX Wave Cannon Avoid',
       type: 'HeadMarker',
@@ -54,7 +49,7 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'ZurvanEX Wave Cannon Stack',
       type: 'StartsUsing',
-      netRegex: { id: '1C712', source: 'Zurvan' },
+      netRegex: { id: '1C72', source: 'Zurvan' },
       condition: Conditions.targetIsNotYou(), // The target is stunned during this mechanic
       response: Responses.stackMarkerOn(),
     },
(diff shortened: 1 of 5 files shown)


ui/raidboss/data/03-hw/trial/zurvan-ex.txt
```diff
@@ -0,0 +1,226 @@
...
+1527.7 "Soar" sync / 1[56]:[^:]*:Zurvan:1E09:/
+1527.8 "--untargetable--"
+
+
+
+# Enrage at 12 minutes
+1989.7 "--sync--" sync / 14:[^:]*:Zurvan:1E59:/ window 2000,100
+1998.8 "Flame Halberd" sync / 1[56]:[^:]*:Zurvan:1E59:/
+1998.8 "--untargetable--"
+2000.0 "Flame Halberd" sync / 1[56]:[^:]*:Zurvan:1E58:/
```
quisquous, line 226:
```suggestion
1998.8 "Flame Halberd Enrage" sync / 1[56]:[^:]*:Zurvan:1E59:/
1998.8 "--untargetable--"
2000.0 "--sync--" sync / 1[56]:[^:]*:Zurvan:1E58:/
```

ui/raidboss/data/03-hw/trial/zurvan-ex.ts
```diff
@@ -0,0 +1,234 @@
...
+const triggerSet: TriggerSet<Data> = {
+  id: 'ContainmentBayZ1T9Extreme,',
+  zoneId: ZoneId.ContainmentBayZ1T9Extreme,
+  timelineFile: 'zurvan-ex.txt',
+  timelineTriggers: [
+    {
+      id: 'ZurvanEX Metal Cutter',
+      regex: /Metal Cutter/,
+      beforeSeconds: 4,
+      suppressSeconds: 10,
```
quisquous, line 23:
```suggestion
      suppressSeconds: 16,
```

Hmm.  There's a couple of places with four in a row that seems very noisy to call out multiple tank cleave calls with no mechanic in between.  16 seconds make the first phase a little weird (since maybe you want a call for the first one in the second set), but overall I think is better?

A compromise could be to add a counter and do 10 seconds for the first one 16 seconds suppress after that.

ui/raidboss/data/03-hw/trial/zurvan-ex.ts
```diff
@@ -0,0 +1,234 @@
...
+    {
+      id: 'ZurvanEX Cool Flame Call',
+      type: 'HeadMarker',
+      netRegex: { id: '0017' },
+      condition: Conditions.targetIsYou(),
+      alertText: (_data, _matches, output) => output.demonicSpread!(),
+      run: (data, matches) => data.flameTarget ??= matches.target,
+      outputStrings: {
+        demonicSpread: {
+          en: 'Spread -- Don\'t stack!',
```
quisquous, line 102: I think a spread response might be plenty here.  To me "spread" already implies "don't stack".

quisquous (commented):
Thank you!

ui/raidboss/data/03-hw/trial/zurvan-ex.ts
JLGarber: I guess maybe we could phase-track and suppress for 10 during the opener and then 16 after The Purge? That would be easily doable.

ui/raidboss/data/03-hw/trial/zurvan-ex.ts
JLGarber: It does, but this does follow conventions elsewhere in the repository of being explicit about it if there are both spread and stack mechanics resolving at once. Normally I would agree with you if this were a sequence of mechanics, something like Pantokrator in Omega Ultimate, but for a single mechanic like this I think it's useful to remind the user to watch for the stack.

ui/raidboss/data/03-hw/trial/zurvan-ex.ts
quisquous: Yeah either way, either phase track or a counter?

ui/raidboss/data/03-hw/trial/zurvan-ex.ts
quisquous: Where else do we do this? (Should those also just be spread??)

JLGarber: [Alexander Prime](https://github.com/quisquous/cactbot/blob/4cda9ec2d52abf00dc1f6f4802f58d73fdc61c2e/ui/raidboss/data/03-hw/raid/a12n.ts#L71), [Hades](https://github.com/quisquous/cactbot/blob/4cda9ec2d52abf00dc1f6f4802f58d73fdc61c2e/ui/raidboss/data/05-shb/trial/hades.ts#L264) ,  and [Voidwalker](https://github.com/quisquous/cactbot/blob/4cda9ec2d52abf00dc1f6f4802f58d73fdc61c2e/ui/raidboss/data/05-shb/raid/e2n.ts#L134). Granted, Alexander Prime is my work too, so that's less compelling. (There's also a different reason not to stack there, so it's maybe not an exact analogy.)

quisquous: Ok, sounds good! I don't feel that strongly here.

PR #5425, opened 2023-05-26
raidboss: More ex6 triggers

valarnin:
Fix 2nd Terrastorm trigger
Add Gale Sphere triggers
Add Arctic Assault trigger
Add Flames of Eventide trigger

Not 100% sure on the arctic assault trigger yet, I was only able to find a few variations to test against.

This includes both triggers for Gale Sphere that we discussed, open to further suggestions on these. Example screenshot from raidemulator for each of the new triggers:

image
image

### DIFF
--- a/ui/raidboss/data/06-ew/trial/golbez-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/golbez-ex.ts
@@ -96,7 +96,10 @@ const triggerSet: TriggerSet<Data> = {
       regex: /Flames of Eventide 1/,
       beforeSeconds: 0,
       condition: (data) => data.role === 'tank',
-      alarmText: Outputs.tankSwap,
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: Outputs.tankSwap,
+      },
     },
   ],
   triggers: [
--- a/ui/raidboss/data/06-ew/trial/golbez-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/golbez-ex.ts
@@ -8,6 +8,9 @@ import { RaidbossData } from '../../../../../types/data';
 import { PluginCombatantState } from '../../../../../types/event';
 import { TriggerSet } from '../../../../../types/trigger';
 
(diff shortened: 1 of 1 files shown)


valarnin:
There are some issues with these in-instance. Trying to sort them out still.

wexxlee:
> This includes both triggers for Gale Sphere that we discussed, open to further suggestions on these.

It might be helpful to call the clone order a little before calling the safe spots so you don't get stacked info alerts.  You can determine the order the clones will fire based on the Goldez's Shadow combatants.  They each use a different ability indicating which wall they will float to.  IIRC:

844F: North
8450: East
8451: West
8452: South

quisquous:
> There are some issues with these in-instance. Trying to sort them out still.

Yeah, the clone order was correct for me but it felt like E/W N/S directions were (always?) swapped during gale sphere.  N/S seemed ok for the meteors + ice spikes but E/W also was (always?) backwards.

valarnin:
This should be ready for a proper review now.

Perplexity:
Gale Sphere safespots are wrong sometimes.

valarnin:
> Gale Sphere safespots are wrong sometimes.

With the latest version? 6ff8031 or 744567d.

Perplexity:
> > Gale Sphere safespots are wrong sometimes.
> 
> With the latest version? [6ff8031](https://github.com/quisquous/cactbot/commit/6ff8031f7501ee4cbc3ebed0248709576a055b23) or [744567d](https://github.com/quisquous/cactbot/commit/744567d84476ec3c4d1637c22ab39af6dbd4bc64).

Will try actually

Perplexity:
Okay it works with the newer commits, my bad. Good stuff

Perplexity:
Terraform 2 safe spots are wrong, it seems to always be the opposite?

valarnin:
> Terraform 2 safe spots are wrong, it seems to always be the opposite?

The ice spike wall is inverted sometimes during terra 2 I think. I'm investigating that.

Issue #5200, opened 2023-01-15
about make_timeline


tssailzz8:
in docs it say "ts-node util/logtools/make_timeline.ts -f CapeWestwind.log -s 18:42:23.614 -e 18:49:22.934",in fact i should use ''npx ts-node util/logtools/make_timeline.ts -f CapeWestwind.log -s 2023-01-14T21:02:37.718 -e 2023-01-14T21:12:13.622''

quisquous:
Mmm, I'm not sure.  For me, `ts-node` on its own works and I am not sure what setups require `npx`.

valarnin:
`ts-node` will work if it's installed globally, e.g.:
```bash
npm install -g typescript
npm install -g ts-node
```

Otherwise yes, you need to use `npx ts-node ...`

PR #5432, opened 2023-05-27
raidboss/oopsy: add P9N

foophoof:
This is my first attempt at a timeline, so there's probably some issues here.

I've run this with `test_timeline.ts` against 17 different runs that I did and didn't see anything particularly concerning, there's a few abilities that are sometimes about 0.2 seconds off, but for the majority of the logs it's more or less spot on.

I might've gone a little bit overboard with the forking/merging of timelines in the timeline, since the actual casts are about the same either side of the fork, but I decided to leave it in for the first review since I already did the work of splitting it up. Happy to merge it back together, though, if that's preferred, the timeline of this fight is actually quite linear if you ignore the fire vs ice parts.

I did a test run as well with this, and didn't notice anything obviously wrong, though I don't have a good feel for what's good when it comes to triggers and what to include/not include (especially for things like Burst where you get "Move Away" immediately followed by "Stack on X").

I also included a full table of all the action IDs I discovered when going through this. Not sure if you want that or not, but I wrote it for myself in order to make sense of the timeline, so I figured I'd include it in case it was useful.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p9n.ts
+++ b/ui/raidboss/data/06-ew/raid/p9n.ts
@@ -1,3 +1,6 @@
+import Conditions from '../../../../../resources/conditions';
+import Outputs from '../../../../../resources/outputs';
+import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
@@ -8,7 +11,166 @@ const triggerSet: TriggerSet<Data> = {
   id: 'AnabaseiosTheNinthCircle',
   zoneId: ZoneId.AnabaseiosTheNinthCircle,
   timelineFile: 'p9n.txt',
-  triggers: [],
+  triggers: [
+    {
+      id: 'P9N Gluttony\'s Augur',
+      type: 'StartsUsing',
+      netRegex: { id: '8116', source: 'Kokytos', capture: false },
+      response: Responses.aoe(),
+    },
+    {
+      id: 'P9N Global Spell',
+      type: 'StartsUsing',
+      netRegex: { id: '8141', source: 'Kokytos', capture: false },
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'aoe + bleed',
+          de: 'AoE + Blutung',
+          fr: 'AoE + saignement',
+          ja: 'AoE + DoT',
+          cn: 'AOE + 流血',
+          ko: '전체 공격 + 도트',
+        },
+      },
+    },
+    {
+      id: 'P9N Ascendant Fist',
+      type: 'StartsUsing',
+      netRegex: { id: '8131', source: 'Kokytos', capture: true },
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'P9N Pulverizing Pounce',
+      type: 'HeadMarker',
+      netRegex: { id: '00A1' },
+      response: Responses.stackMarkerOn(),
+    },
+    {
+      id: 'P9N Fire III',
+      type: 'HeadMarker',
+      netRegex: { id: '01C5' },
+      condition: Conditions.targetIsYou(),
+      response: Responses.spread(),
+    },
+    {
+      id: 'P9N Charybdis',
+      type: 'StartsUsing',
+      netRegex: { id: '8133', source: 'Kokytos', capture: true },
+      condition: Conditions.targetIsYou(),
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Drop Puddles Outside',
+          de: 'Flächen drausen ablegen',
+          fr: 'Déposez les zones au sol à l\'extérieur',
+          ja: '散開',
+          cn: '散开',
+          ko: '산개',
+        },
+      },
+    },
+    {
+      id: 'P9N Archaic Rockbreaker',
+      type: 'StartsUsing',
+      netRegex: { id: '8128', source: 'Kokytos', capture: false },
+      response: Responses.knockback(),
+    },
+    {
+      id: 'P9N Beastly Roar',
+      type: 'StartsUsing',
+      netRegex: { id: '8138', source: 'Kokytos', capture: false },
+      response: Responses.knockback(),
+    },
+    {
+      id: 'P9N Archaic Demolish',
+      type: 'StartsUsing',
+      netRegex: { id: '812F', source: 'Kokytos', capture: false },
+      alertText: (_data, _matches, output) => output.healerGroups!(),
+      outputStrings: {
+        healerGroups: Outputs.healerGroups,
+      },
+    },
+    {
+      id: 'P9N Front Combination + Inside Roundhouse',
+      type: 'StartsUsing',
+      netRegex: { id: '8148', source: 'Kokytos', capture: false },
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Get Behind and Under',
+        },
+      },
+    },
+    {
+      id: 'P9N Rear Combination + Inside Roundhouse',
+      type: 'StartsUsing',
+      netRegex: { id: '814A', source: 'Kokytos', capture: false },
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Go Front and Under',
+        },
+      },
+    },
+    {
+      id: 'P9N Front Combination + Outside Roundhouse',
+      type: 'StartsUsing',
+      netRegex: { id: '8147', source: 'Kokytos', capture: false },
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Get Behind and Out',
+          de: 'Geh nach Hinten und Raus',
+          fr: 'Passez derrière et extérieur',
+          ja: '後ろの外側へ',
+          cn: '去背后远离',
+          ko: '보스 뒤 바깥쪽으로',
+        },
+      },
+    },
+    {
+      id: 'P9N Rear Combination + Outside Roundhouse',
+      type: 'StartsUsing',
+      netRegex: { id: '8149', source: 'Kokytos', capture: false },
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Go Front and Out',
+        },
+      },
+    },
+    {
+      id: 'P9N Ecliptic Meteor',
+      type: 'StartsUsing',
+      netRegex: { id: '813B', source: 'Kokytos', capture: false },
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Hide behind unbroken meteor',
+          de: 'Hinter einem nicht zerbrochenen Meteor verstecken',
+          fr: 'Cachez-vous derrière le météore intact',
+          ja: '壊れていないメテオの後ろへ',
+          cn: '躲在未破碎的陨石后',
+          ko: '금이 안 간 돌 뒤에 숨기',
+        },
+      },
+    },
+    {
+      id: 'P9N Burst',
+      type: 'StartsUsing',
+      netRegex: { id: '8136', source: 'Comet', capture: false },
+      response: Responses.moveAway(),
+    },
+  ],
+  timelineReplace: [
+    {
+      locale: 'en',
+      replaceText: {
+        'Front Combination/Rear Combination': 'Front/Rear Combination',
+      },
+    },
+  ],
 };
 
 export default triggerSet;
--- a/ui/raidboss/data/06-ew/raid/p9n.ts
+++ b/ui/raidboss/data/06-ew/raid/p9n.ts
@@ -162,6 +162,26 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: { id: '8136', source: 'Comet', capture: false },
       response: Responses.moveAway(),
     },
+    {
+      id: 'P9N Gluttonous Rampage',
+      type: 'HeadMarker',
+      netRegex: { id: '013A', capture: true },
+      alertText: (data, matches, output) => {
+        if (data.me === matches.target)
+          return output.tankbusterOnYouStretchTethers!();
+
+        if (data.role === 'healer' || data.job === 'BLU')
+          return output.tankbusterOn!({ player: data.ShortName(matches.target) });
+      },
+      outputStrings: {
+        tankbusterOnYouStretchTethers: {
+          en: 'Tankbuster on YOU -- stretch tether',
+        },
+        tankbusterOn: {
+          en: 'Tankbuster on ${player}',
+        },
+      },
+    },
   ],
   timelineReplace: [
     {
--- a/ui/raidboss/data/06-ew/raid/p9n.ts
+++ b/ui/raidboss/data/06-ew/raid/p9n.ts
@@ -168,6 +168,7 @@ const triggerSet: TriggerSet<Data> = {
       locale: 'en',
       replaceText: {
         'Front Combination/Rear Combination': 'Front/Rear Combination',
+        'Inside Roundhouse/Outside Roundhouse': 'Inside/Outside Roundhouse',
       },
     },
   ],
--- a/ui/raidboss/data/06-ew/raid/p9n.txt
+++ b/ui/raidboss/data/06-ew/raid/p9n.txt
@@ -14,7 +14,7 @@ hideall "--sync--"
 
 0.0 "--sync--" sync /Engage!/ window 0,1
 7.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/ window 10,10
-18.2 "Gluttony's Augur" sync / 1[56]:[^:]*:Kokytos:8116:/
+18.2 "Gluttony's Augur" sync / 1[56]:[^:]*:Kokytos:8116:/ window 15,15
 
 
 
--- a/ui/raidboss/data/06-ew/raid/p9n.txt
+++ b/ui/raidboss/data/06-ew/raid/p9n.txt
@@ -2,7 +2,10 @@
 #
 # List of all ability IDs seen at the bottom of the file
 #
+# Autos
 # -ii 811C 8115 8127 8140
+# Cast resolutions
+# -ii 8117 8130 8142 813E
 
 hideall "--Reset--"
 hideall "--sync--"
@@ -12,7 +15,6 @@ hideall "--sync--"
 0.0 "--sync--" sync /Engage!/ window 0,1
 7.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/ window 10,10
 18.2 "Gluttony's Augur" sync / 1[56]:[^:]*:Kokytos:8116:/
-18.6 "--sync--" sync / 1[56]:[^:]*:Kokytos:8117:/
 
 
 
@@ -30,7 +32,6 @@ hideall "--sync--"
 90.3 "Fire III" sync / 1[56]:[^:]*:Kokytos:(8120|811E):/
 90.3 "Blizzard III" sync / 1[56]:[^:]*:Kokytos:(811F|8121):/
 99.8 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8141:/
-100.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:8142:/
 103.0 "Disgorge" sync / 1[56]:[^:]*:Kokytos:8124:/
 
 
@@ -42,9 +43,7 @@ hideall "--sync--"
 145.1 "Archaic Rockbreaker 1" sync / 1[56]:[^:]*:Kokytos:812A:/
 150.1 "Archaic Rockbreaker 2" sync / 1[56]:[^:]*:Kokytos:812A:/
 156.7 "Archaic Demolish" sync / 1[56]:[^:]*:Kokytos:812F:/
-157.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:8130:/
 163.8 "Gluttony's Augur" sync / 1[56]:[^:]*:Kokytos:8116:/
-164.2 "--sync--" sync / 1[56]:[^:]*:Kokytos:8117:/
 170.9 "Ascendant Fist" sync / 1[56]:[^:]*:Kokytos:8131:/
 183.2 "Front Combination/Rear Combination" sync / 1[56]:[^:]*:Kokytos:814[8A]:/
 183.4 "Outside Roundhouse" sync / 1[56]:[^:]*:Kokytos:812C:/
@@ -60,7 +59,6 @@ hideall "--sync--"
 221.8 "--sync--" sync / 1[56]:[^:]*:Kokytos:8133:/
 226.9 "Beastly Roar" sync / 1[56]:[^:]*:Kokytos:8138:/
 229.0 "Beastly Bile" sync / 1[56]:[^:]*:Kokytos:813D:/
-231.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:813E:/
 238.1 "Pulverizing Pounce" sync / 1[56]:[^:]*:Kokytos:813F:/
 253.4 "Gluttonous Rampage" sync / 1[56]:[^:]*:Kokytos:8139:/
 253.6 "Ravenous Bite 1" #sync / 1[56]:[^:]*:Kokytos:813A:/
@@ -99,7 +97,6 @@ hideall "--sync--"
 363.8 "Blizzard III" sync / 1[56]:[^:]*:Kokytos:(811F|8121):/
 363.8 "Fire III" sync / 1[56]:[^:]*:Kokytos:(811E|8120):/
 370.5 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8141:/
-371.4 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8142:/
 372.7 "Disgorge" sync / 1[56]:[^:]*:Kokytos:8124:/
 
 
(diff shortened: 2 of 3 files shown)


ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,364 @@
...
+622.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:8133:/
+627.1 "Beastly Roar" sync / 1[56]:[^:]*:Kokytos:8138:/
+629.2 "Beastly Bile" sync / 1[56]:[^:]*:Kokytos:813D:/
+631.2 "--sync--" sync / 1[56]:[^:]*:Kokytos:813E:/
+638.3 "Pulverizing Pounce" sync / 1[56]:[^:]*:Kokytos:813F:/
+640.4 "--sync--" sync / 1[56]:[^:]*:Kokytos:8140:/
+643.5 "--sync--" sync / 1[56]:[^:]*:Kokytos:8140:/
+653.6 "Gluttonous Rampage" sync / 1[56]:[^:]*:Kokytos:8139:/
+653.8 "Ravenous Bite 1" sync / 1[56]:[^:]*:Kokytos:813A:/
+655.9 "Ravenous Bite 2" sync / 1[56]:[^:]*:Kokytos:813A:/
```
foophoof, line 189: These are tankbusters on both the tanks, one after another, and also has distance tethers on them, which should probably have a trigger but I'm not sure what trigger would be best, if there is something similar pre-existing.

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,364 @@
...
+658.1 "Touchdown" sync / 1[56]:[^:]*:Kokytos:8137:/
```
foophoof, line 190: This is a proximity AoE centered on the centre of the arena, is there some trigger message I can reuse for that?

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,360 @@
...
+1957.6 "Touchdown" sync / 1[56]:[^:]*:Kokytos:8137:/
+1958.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:8146:/
+1960.8 "Disgorge" sync / 1[56]:[^:]*:Kokytos:8126:/
+1968.1 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/
+1971.2 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/
+
+
+
+# Mage phase 3
+1979.3 "Ravening (Mage)" sync / 1[56]:[^:]*:Kokytos:8118:/
```
foophoof, line 360: This is as far as my longest log goes, but in Duty Finder in my experience the boss will usually die in the second Mage or Fighter phase, I rarely see the 2nd Behemoth phase.

foophoof:
> I might've gone a little bit overboard with the forking/merging of timelines in the timeline, since the actual casts are about the same either side of the fork, but I decided to leave it in for the first review since I already did the work of splitting it up. Happy to merge it back together, though, if that's preferred, the timeline of this fight is actually quite linear if you ignore the fire vs ice parts.

I tried linearising it just to see what it looked like, and I much prefer that now that I look at it, so I pushed that as a separate commit.

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,259 @@
 ### P9N: Anabaseios: The Ninth Circle
 
+# Ability IDs that appear in this instance:
```
quisquous, line 3: This is nice! I think my only request would be to put this at the end of the timeline rather than the beginning, just so it's easier to scroll to the "real" stuff.

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,259 @@
...
+# "Unaspected" being the state the boss is in before the first Ravening, and between Disgorge and Ravening
+#
+# Ability ID   | UMFB | Description
+# Kokytos:8115 | U    | Auto-attack (unaspected)
+# Kokytos:8116 | U F  | Gluttony's Augur - cast bar
+# Kokytos:8117 | U F  | Gluttony's Augur - damage
+# Kokytos:8118 | U  B | Ravening - move to mage phase
+# Kokytos:8119 |  M   | Ravening - move to fighter phase
+# Kokytos:811A |   F  | Ravening - move to Behemoth phase
+# Kokytos:811C |  M   | Auto-attack (magical, mage)
```
quisquous, line 14: Can you ignore all of the boss autos from the timeline? (They're generally noisy but also have been inconsistent in other instances.)  If you haven't seen it, you can use `-ii 811C` to ignore an id on the command line when using `make_timeline`.  When writing a timeline, I try to put the command line I used at the top (e.g. golbez-ex.txt has a good example) so other folks rerunning `make_timeline` later for whatever reason can not have to manually redo that work.

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,259 @@
...
+
+# Mage Phase 1
+26.4 "Ravening (Mage)" sync / 1[56]:[^:]*:Kokytos:8118:/
+35.5 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
+38.6 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
+41.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:8144:/
+48.4 "Fire III" sync / 1[56]:[^:]*:Kokytos:814E:/
+51.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:8122:/
+55.6 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
+56.3 "Fire III" sync / 1[56]:[^:]*:Kokytos:8120:/
```
quisquous, line 97: Optional if you want to, but I try to only use real ability names in the timeline, but sometimes there's a need to differentiate.  Here, it could make sense to say `Fire III (cast)` and `Fire III (explode)` or something to differentiate that this is only one `Fire III` and not two.  `golbez-ex.txt` does this a bunch as an example.

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,259 @@
...
+351.6 "Sphere Shatter" sync / 1[56]:[^:]*:Ice Sphere:86E[45]:/
+351.6 "Explosion" sync / 1[56]:[^:]*:Fire Sphere:86E[23]:/
+356.0 "Dualspell" sync / 1[56]:[^:]*:Kokytos:811D:/
+359.3 "--sync--" sync / 1[56]:[^:]*:Kokytos:812[23]:/
+361.2 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
+363.8 "Blizzard III" sync / 1[56]:[^:]*:Kokytos:(811F|8121):/
+363.8 "Fire III" sync / 1[56]:[^:]*:Kokytos:(811E|8120):/
+364.4 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
+370.5 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8141:/
+371.4 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8142:/
```
quisquous, line 200: Can you `-ii 8142` as well? `8141` is the cast, and `8142` is the damage on the party.  There's only one `Global Spell` occurring here.

ui/raidboss/data/06-ew/raid/p9n.txt
```diff
@@ -1,6 +1,259 @@
...
+528.1 "Touchdown" sync / 1[56]:[^:]*:Kokytos:8137:/
+528.5 "--sync--" sync / 1[56]:[^:]*:Kokytos:8146:/
+531.3 "Disgorge" sync / 1[56]:[^:]*:Kokytos:8126:/
+538.6 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/
+541.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/
+
+
+
+# Mage phase 3
+549.8 "Ravening (Mage)" sync / 1[56]:[^:]*:Kokytos:8118:/
```
quisquous, line 259: I am fairly sure this is the loop back to mage phase 2 based on my 11 minute log.  I see it go mage 2 and then fighter 2 up into gluttony's augur.  I think it's prooobably safe to assume a loop here.

Usually I look into fflogs in these cases and find the log with the longest time, but I have a hard time figuring out how to search by zones like this.  It might be there, but I'm not sure how to find it.

quisquous (commented):
Thank you for this! Overall I really only have a few small comments, but this looks really great.  Thanks for all the work.

I'd be curious to know what if anything was the hardest part of writing one of these for the first time, in terms of helping other people do it.

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: If it's helpful, here's the lines from `make_timeline` from my log at this point.

```
544.8 "Ravening" sync / 1[56]:[^:]*:Kokytos:8118:/
553.9 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
557.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
559.1 "--sync--" sync / 1[56]:[^:]*:Kokytos:8144:/
565.5 "Iceflame Summoning" sync / 1[56]:[^:]*:Kokytos:86E1:/
570.4 "--sync--" sync / 1[56]:[^:]*:Kokytos:8122:/
570.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
573.8 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
576.4 "--sync--" sync / 1[56]:[^:]*:Ice Sphere:8729:/
576.4 "--sync--" sync / 1[56]:[^:]*:Fire Sphere:8728:/
576.4 "Sphere Shatter" sync / 1[56]:[^:]*:Ice Sphere:86E4:/
576.4 "Explosion" sync / 1[56]:[^:]*:Fire Sphere:86E3:/
581.0 "Iceflame Summoning" sync / 1[56]:[^:]*:Kokytos:86E1:/
585.9 "--sync--" sync / 1[56]:[^:]*:Kokytos:8123:/
588.3 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
591.9 "--sync--" sync / 1[56]:[^:]*:Ice Sphere:872A:/
591.9 "--sync--" sync / 1[56]:[^:]*:Fire Sphere:8727:/
591.9 "Sphere Shatter" sync / 1[56]:[^:]*:Ice Sphere:86E5:/
591.9 "Explosion" sync / 1[56]:[^:]*:Fire Sphere:86E2:/
596.5 "Dualspell" sync / 1[56]:[^:]*:Kokytos:811D:/
599.8 "--sync--" sync / 1[56]:[^:]*:Kokytos:8123:/
601.7 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
604.3 "Blizzard III" sync / 1[56]:[^:]*:Kokytos:8121:/
604.3 "Fire III" sync / 1[56]:[^:]*:Kokytos:811E:/
604.9 "--sync--" sync / 1[56]:[^:]*:Kokytos:811C:/
611.0 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8141:/
611.8 "Global Spell" sync / 1[56]:[^:]*:Kokytos:8142:/
613.1 "Disgorge" sync / 1[56]:[^:]*:Kokytos:8124:/
617.1 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/
620.2 "--sync--" sync / 1[56]:[^:]*:Kokytos:8115:/
625.3 "Ravening" sync / 1[56]:[^:]*:Kokytos:8119:/
635.4 "--sync--" sync / 1[56]:[^:]*:Kokytos:8127:/
643.5 "Archaic Rockbreaker" sync / 1[56]:[^:]*:Kokytos:8128:/
645.1 "Shockwave" sync / 1[56]:[^:]*:Kokytos:8129:/
647.8 "--sync--" sync / 1[56]:[^:]*:Kokytos:8127:/
650.9 "--sync--" sync / 1[56]:[^:]*:Kokytos:8127:/
652.0 "Archaic Rockbreaker" sync / 1[56]:[^:]*:Kokytos:812A:/
654.0 "--sync--" sync / 1[56]:[^:]*:Kokytos:8127:/
657.0 "Archaic Rockbreaker" sync / 1[56]:[^:]*:Kokytos:812A:/
662.1 "Rear Combination" sync / 1[56]:[^:]*:Kokytos:814A:/
662.3 "Outside Roundhouse" sync / 1[56]:[^:]*:Kokytos:812C:/
665.3 "Swinging Kick" sync / 1[56]:[^:]*:Kokytos:812E:/
673.4 "Gluttony's Augur" sync / 1[56]:[^:]*:Kokytos:8116:/
673.8 "Gluttony's Augur" sync / 1[56]:[^:]*:Kokytos:8117:/
```

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: Can do, I saw some conflicting things in different places on whether to include autos or not, so I decided to include them for now and remove them if they weren't needed.

Should I still include the first auto to have something to sync with near the start? Otherwise the first ability I found near the start is Gluttony's Augur around 11-15 seconds in.

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: `-ii` or mark them as `--sync--`? There's a few other cases of this, e.g. `Beastly Bile` which casts on `813D` and damage is on `813E`, and `Archaic Demolish` is `812F`/`8130`. Most of them I marked the damage one as `--sync--`, including some other `Global Spell` casts, I had just missed this one.

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: I found some FFLogs entries (e.g. [like this](https://www.fflogs.com/zone/reports?zone=54&boss=88&difficulty=0&class=Any&spec=Any&keystone=0&kills=0&duration=550)), but I wasn't able to find out how to download a network log from them to actually use those.

foophoof: That log you provided does line up more or less perfectly with a loop from Mage 2, so I think you're right.

Only exception is that I have Fighter 2 hardcoded as having an `Inside Roundhouse`, since in every single log I had, it's `Outside Roundhouse` in Fighter 1 and `Inside Roundhouse` in Fighter 2, so it seems to alternate and always be Outside on odds and Inside on evens.

I could either switch Fighter 2 to be `Inside/Outside Roundhouse` and loop, or I could fill out through Behemoth 3 and _then_ jump back to Mage 1, whichever you think is better.

ui/raidboss/data/06-ew/raid/p9n.txt
JLGarber: Yes, inclusion of the first auto for a safety sync is standard. There have been other Very Specific situations where we include one or two for Reasons, but in general we put them on the ignore list and then specifically include the first one.

ui/raidboss/data/06-ew/raid/p9n.txt
JLGarber: Typically (and this isn't to say any of us have been particularly consistent about it) if it's a dual situation like this we will ignore the cast entirely. "More syncs is better" has some truth to it, but on the flip side, we want the timeline to be as human-readable as possible. Fewer interspersed `"--sync--"` entries would help with that.

(Note that this does not necessarily hold in situations where there's a cast and then *multiple* follow-ups. In that situation we would usually take it case-by-case.)

ui/raidboss/data/06-ew/raid/p9n.txt
JLGarber: > I found some FFLogs entries (e.g. [like this](https://www.fflogs.com/zone/reports?zone=54&boss=88&difficulty=0&class=Any&spec=Any&keystone=0&kills=0&duration=550)), but I wasn't able to find out how to download a network log from them to actually use those.

You can do something like `node --loader ts-node/esm util/logtools/make_timeline.ts -k [YOUR KEY HERE] -r [FFLOGS REPORT ID] -rf [FIGHT NUMBER] -[OTHER ARGS]` to generate a timeline from FFLogs. It's not quite as nice as generating from a local network log, since it's missing some events, but it works for basic stuff.

ui/raidboss/data/06-ew/raid/p9n.txt
JLGarber: @quisquous will have to weigh in for a final decision on this, but it's probably worth doing something like "buster on YOU -- stretch tethers". It's longer than any of us would prefer, but we do want to convey all the information.

ui/raidboss/data/06-ew/raid/p9n.txt
JLGarber: I don't think we have a dedicated response for this. A custom "get to edge" trigger would fit with previous similar mechanics, or maybe `responses.getOut()` would fit if the damage falloff is linear. (If it's not, such as with the P4S Lightning, we would want to be explicit that the user has to be *all* the way out.)

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: > Only exception is that I have Fighter 2 hardcoded as having an Inside Roundhouse, since in every single log I had, it's Outside Roundhouse in Fighter 1 and Inside Roundhouse in Fighter 2, so it seems to alternate and always be Outside on odds and Inside on evens.

If it's just one ability, it's probably better to just keep the timeline simple and have triggers cover the difference.  If'it's a branch with a whole bunch of abilities tied to it, then a branch in the timeline makes more sense.

> You can do something like node --loader ts-node/esm

I think you should just be able to `ts-node` directly these days without the loader args anymore.

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: Yeah, that's my feeling too, is that there's some balance in having the timeline be readable with too many `--sync--`s in it.  Really the balance is "are there enough abilities to keep the timeline on track" vs "is the timeline readable", which for me is just something every couple of seconds or so.  If there's a bigger break with some syncs in the middle, then definitely keep those.  If it's cast and damage, I don't think you need both.

In other words, I would ignore in this case, because the cast a second earlier more than covers it.

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: > Can do, I saw some conflicting things in different places on whether to include autos or not, so I decided to include them for now and remove them if they weren't needed.

For me I usually think:
* keep the first auto (if it's consistent?) to start the timeline
* keep tank autos if they're mechanics (DSR, TOP, maybe Diamond Extreme should)
* `-ii` other autos because they're noisy / not always consistent in older content

The other thing you can do is specifically include the cast time of the first boss ability as well with a larger sync to get things back on track.  The `golbez-ex.txt` file also does this at 9.4 if you want an example.

@JLGarber one thing that had come up previously is maybe just auto-starting timelines (where it makes sense, i.e. raids but not dungeons) when combat starts rather than on engage or auto which would make things more consistent.  Now that we have in combat lines in OverlayPlugin, we could use those instead? Not a thing to do here, but a thing to consider for the future.

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: In a general sense, I feel like players are probably looking at their feet (i.e. don't stand in telegraphs, move away from proximity markers) and haven't written triggers for these.  I think if you wanted to add a trigger for this, I agree  `Responses.getOut()` seems fine to me for this one.

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: Yeah, there's nothing preexisting here; I think this is the first proximity buster.  I think `Buster on YOU -- stretch tethers` is kind of wordy, but I'm not sure what else to say here either.

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: I guess the proximity marker there is pretty standard and you have loads of time to resolve it (and many ways to do so, as far as I can tell the tankbusters at the same time don't even cleave), so I'll probably leave it without a trigger.

ui/raidboss/data/06-ew/raid/p9n.txt
xiashtra: > Usually I look into fflogs in these cases and find the log with the longest time, but I have a hard time figuring out how to search by zones like this. It might be there, but I'm not sure how to find it.

I start [here](https://www.fflogs.com/zone/reports?zone=54&boss=88&difficulty=100&class=Any&spec=Any&kills=0&duration=0&server=0) and increase the "Minimum Pull Length" until I filter down to only a few logs.  From what I can see, [this](https://www.fflogs.com/reports/gC8AwyNma47pP92M#fight=1 ) ~28min log is the longest one so far.  Just have to be careful because sometimes the excessively long pulls are due to an ACT/FFLogs bug mashing two pulls together.

ui/raidboss/data/06-ew/raid/p9n.ts
```diff
@@ -8,7 +11,166 @@ const triggerSet: TriggerSet<Data> = {
...
+      outputStrings: {
+        text: {
+          en: 'aoe + bleed',
+          de: 'AoE + Blutung',
+          fr: 'AoE + saignement',
+          ja: 'AoE + DoT',
+          cn: 'AOE + 流血',
+          ko: '전체 공격 + 도트',
+        },
+      },
```
xiashtra, line 35: @quisquous There's been a lot of "AoE + bleed" this expansion, I think we should promote this up to Responses.

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: Ended up with "cast" and "resolve" just so I could match the wording between Fire III and Blizzard III, and I couldn't think of a better word for Blizzard III than "resolve".

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: Added `-ii`s and ignored all the autos and the cast resolutions that didn't seem useful for syncing.

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: Ignored this with the other `-ii`s as mentioned in the other comment.

ui/raidboss/data/06-ew/raid/p9n.txt
foophoof: Added as "Tankbuster on ${player}" / "Buster on YOU -- stretch tethers".

The ability is unfortunately not actually targeted on the tanks, as far as I can see it's just the tether line and the head marker, so I based it off the head marker. It does have the slight downside that the people who get the "tankbuster on ${player}" trigger will get two of them, about 2 seconds apart, since the tethers and markers are about two seconds apart.

ui/raidboss/data/06-ew/raid/p9n.ts
foophoof: Happy to make a follow-up PR to do that after this is merged.

foophoof:
> I'd be curious to know what if anything was the hardest part of writing one of these for the first time, in terms of helping other people do it.

Nothing too bad, to be honest. I haven't really done much Node/Javascript stuff in the past, so getting that dev environment running and figuring out things like "run `ts-node` with `npx`", but that wasn't too hard to find. Also the guide being a bit out of date, but not too bad, just had to figure out that the output from `test_timeline` was a bit different now.

Overall the hardest was probably figuring out the various conventions, like what abilities to ignore, what to include, what to trigger off of, what triggers to add in the first place, basically the discussion in this PR.

I think I lucked out a bit starting out with this fight since it turned out to be pretty much entirely timer-based, no percentage based phase pushes, and just the single loop. I have a loose understanding of how `window` works for things like loops, but I feel like I'm effectively just "cargo culting" the exact window numbers for now until I understand it at a deeper level.

If I have the spare time/energy, I might take a stab at slightly updating the guide a bit, at least updating things like the command line output being different now. Would probably be nice to have it center around a duty that actually still exists in the game too, but I'm not sure I'm cut out for that level of rewrite until I learn how this all works better 😅

ui/raidboss/data/06-ew/raid/p9n.txt
quisquous: Yeah, fflogs doesn't always have every ability or add.  I think(?) some abilties that can be dodged don't show up if nobody can be hit, but it definitely doesn't have things like head markers or game log lines, etc.  I tend to use fflogs only for testing, verifying loops, and finding enrage but "actual" network log files for making timelines.

quisquous:
This all looks great! Thank you so much for all the work.

> > I'd be curious to know what if anything was the hardest part of writing one of these for the first time, in terms of helping other people do it.
> 
> Nothing too bad, to be honest. I haven't really done much Node/Javascript stuff in the past, so getting that dev environment running and figuring out things like "run `ts-node` with `npx`", but that wasn't too hard to find. Also the guide being a bit out of date, but not too bad, just had to figure out that the output from `test_timeline` was a bit different now.

Re: `npx`, I think this has to do with whether you have installed node globally or not, based on valarnin's comment here: https://github.com/quisquous/cactbot/issues/5200#issuecomment-1445011177.  If you feel up for it, feel free to send any changes or corrections to the `CONTRIBUTING.md` page that would have helped you, otherwise I'll try to get around to mentioning installing ts-node globally at least.

And yeah, I think even running `test_timeline` needs a little bit of finesse as you found out, in terms of what's "missable" (but ok to miss) vs what's concerning and needs to be adjusted.

> Overall the hardest was probably figuring out the various conventions, like what abilities to ignore, what to include, what to trigger off of, what triggers to add in the first place, basically the discussion in this PR.

Ok, makes sense!  These are the kinds of things I should really make more clear in the timeline guide, and at least document the heuristics that we've accumulated over time.  I tried to write some of this in https://github.com/quisquous/cactbot/blob/main/CONTRIBUTING.md#timeline-guidelines but it deserves to be fleshed out more.

Thanks for taking the time to share all of this.

PR #5433, opened 2023-05-27
raidboss/oopsy: Add support for P10N

wexxlee:
There was a little hiccup because the boss has a unicode character in its `en` name (æ).  

Manually escaping the unicode character (\u00e6) seemed to work for both timeline and trigger regex, and using the non-unicode characters (ae) in the trigger and timeline names displayed better in the UI.


ui/raidboss/data/06-ew/raid/p10n.ts
```diff
@@ -1,14 +1,201 @@
...
+      id: 'P10N Ultima',
+      type: 'StartsUsing',
+      netRegex: { id: '827B', source: bossNameUnicode, capture: false },
+      response: Responses.aoe(),
+    },
+    {
+      id: 'P10N Pandaemoniac Meltdown',
+      type: 'StartsUsing',
+      netRegex: { id: '8276', source: bossNameUnicode, capture: false },
+      response: Responses.stackMarker(),
```
JLGarber, line 100: A tiny and ignorable nit: In my logs, at least, it looks like there's an associated `87A2` Ability line on the actual target of this spell. It goes out at the same time as `8276` starts casting. Could we use it instead and switch the response to `stackMarkerOn()`?

wexxlee: Done.  Thanks for the suggestion!

ui/raidboss/data/06-ew/raid/p10n.ts
```diff
@@ -1,14 +1,201 @@
...
+      type: 'HeadMarker',
+      netRegex: { id: '01CE' },
+      condition: (data, matches) => data.me === matches.target,
+      alertText: (data, _matches, output) => {
+        data.silkspitOnYou = true;
+        return output.text!();
+      },
+      outputStrings: {
+        text: {
+          en: 'Away from party and posts',
```
quisquous, line 34: Definitely agree with "away from posts", but isn't it "spread from other webs and posts" (wording not great) but trying to differentiate that the goal (I think, but am really not sure) is that the marked people spread from each other and posts, and the party then spreads from the marked people.

ui/raidboss/data/06-ew/raid/p10n.ts
```diff
@@ -1,14 +1,201 @@
...
+    },
+    {
+      id: 'P10N Pandaemonic Pillars',
+      type: 'StartsUsing',
+      netRegex: { id: '825D', source: bossNameUnicode, capture: false },
+      durationSeconds: 5,
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Soak towers',
```
quisquous, line 62: I think we've called this `Get Towers` in so many places that it should probably be a `Response` or `Output` at this point (not in this PR, but just talking out loud).

ui/raidboss/data/06-ew/raid/p10n.ts
```diff
@@ -1,14 +1,201 @@
...
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  silkspitOnYou: boolean;
+}
+
+const bossNameUnicode = 'Pand\u00e6monium';
```
quisquous, line 11: I'm not sure I understand why you can't use `Pandæmonium` directly everywhere.  What sort of hiccup did that cause?

ui/raidboss/data/06-ew/raid/p10n.ts
```diff
@@ -1,14 +1,201 @@
...
+        soak: {
+          en: 'Soak tower',
+        },
+        avoid: {
+          en: 'Avoid towers',
+        },
+      },
+    },
+  ],
+  timelineReplace: [
```
quisquous, line 167: Can you skip this section if it's empty?

ui/raidboss/data/06-ew/raid/p10n.txt
```diff
@@ -1,6 +1,81 @@
...
+# 8279      'Soul Grasp' extra
+# 825[BC]   'Entangling Web' extra
+# 8266      'Pandaemonic Ray' extra
+# 8269      'Touchdown' extra
+# 8271      'Parted Plumes' blades
+# 827[34]   'Wicked Step' individual towers
+# 826[B-E]  Supplemental AoE damage from 'Harrowing Hell'
+# 86B[5-8]  Supplemental AoE damage from 'Harrowing Hell'
+
+# Note: The boss's name has a Unicode character (�)  which is causing syncs to fail.
```
quisquous, line 19: Hmmmm.  This seems like something that should be fixed.  Can you a file a separate issue for this?

ui/raidboss/data/06-ew/raid/p10n.ts
```diff
@@ -1,14 +1,201 @@
...
+      },
+    },
+    {
+      id: 'P10N Touchdown',
+      type: 'StartsUsing',
+      netRegex: { id: '8268', source: bossNameUnicode, capture: false },
+      alertText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Cross bridge / off center platform',
```
quisquous, line 129:
```suggestion
          en: 'Side Platforms',
```

Or maybe "Get to Side Platform"? Just trying to keep this concise.

ui/raidboss/data/06-ew/raid/p10n.txt
```diff
@@ -1,6 +1,81 @@
...
+# Entry names use "ae" as separate characters to ensure proper UI display.
 
 hideall "--Reset--"
 hideall "--sync--"
 
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
+
+0.0 "--sync--" sync /Engage!/ window 0,1
+2.5 "--sync--" sync / 1[56]:[^:]*:Pand\u00e6monium:827A:/
+10.9 "Silkspit (marked)" sync / 1[56]:[^:]*:Pand\u00e6monium:8259:/
```
quisquous, line 30: If this has a cast, can you add in a `14` line sync with a larger window (e.g. `golbez-ex.txt`) just in case the first auto is inconsistent?

ui/raidboss/data/06-ew/raid/p10n.txt
```diff
@@ -1,6 +1,81 @@
...
+50.1 "Ultima" sync / 1[56]:[^:]*:Pand\u00e6monium:827B:/
+59.9 "Pandaemoniac Pillars" sync / 1[56]:[^:]*:Pand\u00e6monium:825D:/
+67.8 "Cannonspawn" sync / 1[56]:[^:]*:Pand\u00e6monium:8264:/
+74.6 "--sync--" sync / 1[56]:[^:]*:Pand\u00e6monium:87A2:/
+80.0 "Pandaemoniac Meltdown" sync / 1[56]:[^:]*:Pand\u00e6monium:8276:/
+90.1 "Soul Grasp" sync / 1[56]:[^:]*:Pand\u00e6monium:8278:/
+101.3 "Entangling Web" sync / 1[56]:[^:]*:Pand\u00e6monium:8721:/
+114.7 "Silkspit (marked)" sync / 1[56]:[^:]*:Pand\u00e6monium:8259:/
+126.7 "Silkspit (resolves)" sync / 1[56]:[^:]*:Pand\u00e6monium:825A:/
+140.4 "Pandaemoniac Ray East/West" sync / 1[56]:[^:]*:Pand\u00e6monium:826[57]:/
```
quisquous, line 43: I think the "East/West" isn't necessary here if it's just called "Pandaemoniac Ray" in game (I think?).  If there were North/South or In/Out or other variations that would make sense to differentiate in the timeline then extra text is good, but otherwise just adds more non-automatic translation work.

ui/oopsyraidsy/data/06-ew/raid/p10n.ts
```diff
@@ -1,12 +1,118 @@
...
+  gainsEffectWarn: {
+    'P10N Parted Plumes Puddle': 'C05', // Bleeding debuff from standing in center puddle
+  },
+  soloWarn: {
+    'P10N Pandaemoniac Meltdown': '8277', // line stack
+  },
+  triggers: [
+    {
+      // Stacking debuff that auto-increments and upon reaching 5 stacks is fatal.
+      id: 'P10N Binding Soul Share Gain',
```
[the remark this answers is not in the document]
wexxlee: This is from Silkspit.  If you don't spread sufficiently, you can be tethered to player(s)/posts(s) and receive the 'Binding Soul Share' debuff.  Can always change the trigger name to Silkspit, although I think (unverified) that you can also take damage from standing in a Silkspit puddle, so that could be confusing.

ui/oopsyraidsy/data/06-ew/raid/p10n.ts
```diff
@@ -1,12 +1,118 @@
...
+        data.bindingSoulPlayers = (data.bindingSoulPlayers ??= []).filter((p) =>
+          p !== matches.target
+        );
+      },
+    },
+    {
+      // Stacking debuff that auto-increments as you remain in the poison.
+      id: 'P10N Concentrated Poison Gain',
+      type: 'GainsEffect',
+      netRegex: NetRegexes.gainsEffect({ effectId: 'E6C' }),
```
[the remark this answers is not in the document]
wexxlee: Ahh, I was not aware of the `count` property. I think `gainsEffectWarn` works for Concentrated Poison, since stacks increase as you continue to stand in the poison (and should be continually reported as a mistake), but I think using `count` for Binding Soul Share makes sense since the debuff auto-increments and it gets spammy.

ui/raidboss/data/06-ew/raid/p10n.ts
wexxlee: I struggled a bit with wording on this one too.  If you have the marker, you can tether to (a) other marked players (fairly decent length tethers), (b) any unmarked players (close proximity), or (c) any of the ground posts (somewhere in between distance-wise).  Maybe just "spread and do not tether"?

ui/raidboss/data/06-ew/raid/p10n.ts
wexxlee: No time like the present?  I pulled a full set of locale strings from hades-ex, and I'll push this in the next commit.

ui/raidboss/data/06-ew/raid/p10n.ts
quisquous: I do think the "away from posts" is good though, as it's not as obvious (at least it wasn't to me).  "Spread from webs and posts", or maybe "Spread (avoid posts)"?

ui/raidboss/data/06-ew/raid/p10n.ts
wexxlee: The 'æ' character did not display correctly in the timeline or in raidemulator when used in a trigger `id` or timeline entry name (it displayed as a diamond with a question-mark in both cases).  During some testing, I also found that timeline entries would not sync to log lines where that character was in the regex when I intentionally adjusted some timeline entries to try to test syncs.  It also seemed that certain triggers were not firing correctly, although once I made the switch and things started working, I admittedly didn't dig in deeper to find root cause.

ui/oopsyraidsy/data/06-ew/raid/p10n.ts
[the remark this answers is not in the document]
quisquous: Makes sense! I just didn't know what this was offhand.  Could the comment say "Stacking debuff from Silkspit ... etc" to make this more clear?

ui/raidboss/data/06-ew/raid/p10n.ts
JLGarber: I'll toss my vote in for `Spread (avoid posts)` as well. I think it's the best compromise on length while ensuring that anyone reading it will Do The Thing in some way.

PR #5435, opened 2023-05-27
util: support jumps in test_timeline

foophoof:
This avoids having test_timeline report events as being missed when intentionally jumping over them.

I'm not 100% sure about this implementation, my understanding of this code is still pretty surface level, but it seems to work for me on a few logs I tested it against (mostly my new p9n stuff).

Also not sure if I should do `>=` or `>`. The latter still reports a missed entry for the line you're jumping to, e.g. this:

```
100.0 "foo" jump 200.0
150.0 "bar"
200.0 "foo"
201.0 "baz"
```

would output something like this with `>`:

```
       0.000 |    1 | 100.0 "foo" jump 200.0
      Missed |    3 | 200.0 "foo"
       0.000 |    4 | 201.0 "baz"
```

But with `>=` it outputs:

```
       0.000 |    1 | 100.0 "foo" jump 200.0
       0.000 |    4 | 201.0 "baz"
```

I'm not sure exactly what the behaviour should be like if you have multiple entries at the same time that you're jumping to, like:

```
100.0 "foo" jump 200.0
150.0 "bar"
200.0 "foo"
200.0 "baz"
```

### DIFF
--- a/util/logtools/test_timeline.ts
+++ b/util/logtools/test_timeline.ts
@@ -435,6 +435,11 @@ class TestTimeline extends Timeline {
         // Skip text events with no sync.
         if (event.sync === undefined)
           continue;
+        // Skip events when jumping
+        if (
+          lastRecord?.event.sync?.jump !== undefined && lastRecord?.event.sync?.jump >= event.time
+        )
+          continue;
 
         ui.records.push({
           event: event,
--- a/util/logtools/test_timeline.ts
+++ b/util/logtools/test_timeline.ts
@@ -437,7 +437,7 @@ class TestTimeline extends Timeline {
           continue;
         // Skip events when jumping
         if (
-          lastRecord?.event.sync?.jump !== undefined && lastRecord?.event.sync?.jump >= event.time
+          lastRecord?.event.sync?.jump !== undefined && lastRecord?.event.sync?.jump > event.time
         )
           continue;
 
--- a/util/logtools/test_timeline.ts
+++ b/util/logtools/test_timeline.ts
@@ -436,6 +436,9 @@ class TestTimeline extends Timeline {
         if (event.sync === undefined)
           continue;
         // Skip events when jumping
+        // TODO: Instead of marking the destination as `missed',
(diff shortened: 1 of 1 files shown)


quisquous:
Thank you for this!

As for your questions, here's extra context.  There's definitely a strongly-followed cactbot heuristic of any jumps to other phases going from an explicit line to another explicit line (i.e. not some timeline time between other lines), and both lines having the same sync (or at least the destination line implies the same sync even if it's not explicitly listed or is commented out, etc).  (Maybe we need some tests for this, etc.)  In the game, sometimes there's a single ability line and sometimes there's multiple, and so depending on this the destination line will be synced or missed depending.

So, I think I'd consider the destination line as "not missed", and at least for me it's useful to have it printed out, in terms of reading the test timeline output.  To that end, I think `>` is better than `>=` so that all destination entries are listed.

Feel free to leave a TODO for this (so I can merge this now and there can be another PR later), but another option is to add some new record type `jump` (along with `sync` and `missed`) and we could label (or color?) it differently so it doesn't look like something is "wrong" with the timeline.  However, then I wonder about what happens about multiple ability lines in game for the same id where the timeline jumps to 200 and then resyncs to 200, and then what would that be labeled, and do you need extra cases for that, etc.  Mostly just saying maybe there's some edge cases here.

foophoof:
> So, I think I'd consider the destination line as "not missed" in all cases, and at least for me it's useful to have it printed out, in terms of reading the test timeline output. To that end, I think > is better than >= so that all destination entries are listed.

I agree, updated to `>`.

> Feel free to leave a TODO for this (so I can merge this now and there can be another PR later), but another option is to add some new record type jump (along with sync and missed) and we could label (or color?) it differently so it doesn't look like something is "wrong" with the timeline. However, then I wonder about what happens about multiple ability lines in game for the same id where the timeline jumps to 200 and then resyncs to 200, and then what would that be labeled, and do you need extra cases for that, etc. Mostly just saying maybe there's some edge cases here.

Added a TODO with a link to your comment for some added context.

PR #5436, opened 2023-05-27
raidemulator: Remove significantStates

valarnin:
Plus some other minor cleanup related to combatant properties/functions


quisquous:
I think this is definitely simpler overall.  I know I've had questions in the past over significant states as well.

PR #5439, opened 2023-05-28
i18n: update ko translations

Bing-su:
https://github.com/quisquous/cactbot/blob/c245ef6b45f864fc79689f6e5a87cc06cb899e6e/ui/raidboss/data/03-hw/trial/zurvan-ex.ts#L8-L12

From my reading of the code, it looks like `infiniteElement` will always be either `fire` or `ice`, regardless of the display language.


quisquous:
> https://github.com/quisquous/cactbot/blob/c245ef6b45f864fc79689f6e5a87cc06cb899e6e/ui/raidboss/data/03-hw/trial/zurvan-ex.ts#L8-L12
> 
> From my reading of the code, it looks like `infiniteElement` will always be either `fire` or `ice`, regardless of the display language.

Yeah, it looks like we need `fire` and `ice` outputStrings there, and I missed this when looking through it.  cc @JLGarber

JLGarber:
Sorry about that! I was under a small time crunch and wasn't thinking over the ramifications of hardcoding the element values. Thanks for fixing that for us quisquous.

PR #5442, opened 2023-05-28
raidboss: fix P10N Pandaemoniac Meldown

wexxlee:
Small mistake I didn't catch during testing, but changing the Pandaemoniac Meltdown trigger ability id and output means it should look for a 21-line, not the previous 20-line.  Fixed and tested.


quisquous:
Thanks for the fix!

Issue #5437, opened 2023-05-28
raidboss: Issues with unicode characters in timeline entry names and regex syncs


wexxlee:
While working on P10N, I noticed an issue with uncode support in the raidboss timeline functionality:

1. If unicode characters (specifically, æ) are included a timeline entry name, the timeline entry will display the character as (�) in the UI.
2. Regex syncs do not appear to catch the line, and the timeline seems to run without syncing to those lines.

quisquous:
> While working on P10N, I noticed an issue with uncode support in the raidboss timeline functionality:
> 
> 1. If unicode characters (specifically, æ) are included a timeline entry name, the timeline entry will display the character as (�) in the UI.

If I change a timeline entry, I see this:
![image](https://github.com/quisquous/cactbot/assets/462317/1d4ef7dd-80b3-488b-96eb-55e1a67ba6df)

Is this what you mean for this point, and isn't appearing correctly?

quisquous:
> 2. Regex syncs do not appear to catch the line, and the timeline seems to run without syncing to those lines.

As a test, I added this line to `t9.txt` and went in unsynced.

`0.0 "--sync--" sync /Pandæmonium/ window 0,1 jump 693.6`

If I type `/echo Pandæmonium` into chat, it starts the timeline successfully.

wexxlee:
Sorry for the long delayed reply - didn't have a chance to get back in and spend more time looking at this until now.

> Is this what you mean for this point, and isn't appearing correctly?

Sorry, I should have been more accurate.  I actually meant the timeline overlay itself.  Here's what it looks like for me when the unicode character is included in the timeline entry name:
![image](https://github.com/quisquous/cactbot/assets/86693821/d346c7c7-4c8d-4449-828d-37798a6e47ac)

Wondering if perhaps this is just a font issue though, and not indicative of a broader problem with unicode support?

> If I type /echo Pandæmonium into chat, it starts the timeline successfully.

I can still repro the timeline sync issue.  As a test, I ran a local webpack instance and modified the p10n.txt timeline file.  During my first test pull, I changed the regex to use the actual unicode character instead of the escaped one for the first half-dozen entries, e.g.:
```0.0 "--sync--" sync /Engage!/ window 0,1
2.5 "--sync--" sync / 1[56]:[^:]*:Pandæmonium:827A:/
7.9 "--sync--" sync / 14:[^:]*::Pandæmonium:8259:/ window 15,15
[...]

PR #5447, opened 2023-05-29
i18n: 6.4 translations (continued)

Akurosia:
@quisquous I am not 100% sure for p10n if the translated names will work as they are.
When running through the `translate-timeline`, they will return as "Pandæmonium" with the special "æ" or if this has to be escaped as well 

But i cant avoid errors, its either the translate timeline script does not show correct value but `npm run test` will pass OR (as it is right now) `run test` wont pass (`AssertionError: P10N Pandaemonic Pillars:locale de:missing timelineReplace replaceSync for source 'Pandæmonium': expected false to be true`) and translate timeline will display the correct value.

### DIFF
--- a/ui/raidboss/data/03-hw/trial/zurvan-ex.ts
+++ b/ui/raidboss/data/03-hw/trial/zurvan-ex.ts
@@ -239,9 +239,11 @@ const triggerSet: TriggerSet<Data> = {
         },
         fire: {
           en: 'Fire',
+          de: 'Feuer',
         },
         ice: {
           en: 'Ice',
+          de: 'Eis',
         },
         unknown: Outputs.unknown,
       },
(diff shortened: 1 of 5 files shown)


quisquous:
Sorry, I'm not sure I understand what the two "OR" clauses are.  Right now, this PR passes `npm run test`? What have you tried here, and what are the results?

Akurosia:
Currently it passes, BUT it still has missingTranslations for DE as it does not execpt the replacementSync for the pandemonium

quisquous:
Ok, I think for now maybe leave the missingTranslations: true, as it feels like part of #5437 and we can fix that there.

That said, I think you are missing a backslash? Does this work in game? I think if you want to match English, you should say: `'Pand\\\\u00e6monium': 'Pand\\u00e6monium',`.  Doing that makes `npm run util -- translateTimeline -t p10n -l de` and `npm run util -- translateTimeline -t p10n -l en` look the same to me in terms of syncs.

Akurosia:
added the extra /
test run looked fine to me

PR #5448, opened 2023-05-29
raidemulator: Support for combatant memory lines

valarnin:
~~Since this is in a weird state where it requires both current `main` code and the changes in #5379, the PR will show a lot of extra commits that are part of those branches. As such, please only consider commit c78e682 for review.~~

~~I will rebase to `main` again once #5379 is merged.~~

Rebased to `main`.


valarnin:
@quisquous this is ready for review, in case the rebase/comment edit didn't ping you.

PR #5450, opened 2023-05-30
raidboss/oopsy: P11N Initial Package

JLGarber:
It's done, a little later than everyone else. It's still missing the safespot calls for the illusory adds, since I haven't yet been able to figure out what's going on with those. Overall I'm not terribly happy with the textual choices I made here, but the nature of the mechanics here requires a little more than we normally might want.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p11n.ts
+++ b/ui/raidboss/data/06-ew/raid/p11n.ts
@@ -1,14 +1,195 @@
+import Conditions from '../../../../../resources/conditions';
+import { Responses } from '../../../../../resources/responses';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  busterTargets?: string[];
+}
 
 const triggerSet: TriggerSet<Data> = {
   id: 'AnabaseiosTheEleventhCircle',
   zoneId: ZoneId.AnabaseiosTheEleventhCircle,
   timelineFile: 'p11n.txt',
-  triggers: [],
+  triggers: [
+    {
+      id: 'P11N Eunomia',
+      type: 'StartsUsing',
+      netRegex: { id: '81E2', source: 'Themis', capture: false },
+      response: Responses.aoe(),
+    },
+    {
+      id: 'P11N Divisive Ruling In',
+      type: 'StartsUsing',
+      netRegex: { id: '81B4', source: 'Themis', capture: false },
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Line Cleave -> In',
+        },
+      },
+    },
+    {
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/06-ew/raid/p11n.ts
```diff
@@ -1,14 +1,195 @@
...
+      delaySeconds: 0.5,
+      suppressSeconds: 10,
+      infoText: (data, _matches, output) => {
+        const target1 = data.busterTargets?.[0] || '???';
+        const target2 = data.busterTargets?.[1] || '???';
+        return output.text!({ target1: target1, target2: target2 });
+      },
+      outputStrings: {
+        text: {
+          en: 'Tank buster: ${target1}, ${target2}',
```
quisquous, line 128: Hmm.  What about just "tank buster on you" (alert) vs "tank busters" (info, for the people it's not on)? I think that's what we've done elsewhere.

ui/raidboss/data/06-ew/raid/p11n.ts
```diff
@@ -1,14 +1,195 @@
...
+        },
+      },
+    },
+    {
+      id: 'P11N Dike Cleanup',
+      type: 'HeadMarker',
+      netRegex: { id: '00DA', capture: false },
+      delaySeconds: 10,
+      run: (data) => delete data.busterTargets,
```
quisquous, line 137: You could also delete this in the call trigger, I think?

ui/raidboss/data/06-ew/raid/p11n.txt
```diff
@@ -1,6 +1,149 @@
...
+37.8 "Divisive Ruling (out)" sync / 1[56]:[^:]*:Themis:81B5:/
+40.4 "Divine Ruination" sync / 1[56]:[^:]*:Themis:81B6:/
+50.1 "Divisive Ruling (random)" sync / 1[56]:[^:]*:Themis:81B5:/
+52.7 "Divine Ruination/Ripples of Gloom" sync / 1[56]:[^:]*:Themis:(81B6|81B7|81B8):/
+60.9 "Eunomia" sync / 1[56]:[^:]*:Themis:81E2:/
+
+# Dark and Light Intro
+67.0 "--center--" sync / 1[56]:[^:]*:Themis:8235:/ window 30,30
+73.1 "Dark and Light" sync / 1[56]:[^:]*:Themis:81C3:/
+80.2 "Sigils of Discord" sync / 1[56]:[^:]*:Themis:81C4:/ window 76.2,15
```
quisquous, line 45: I'm not sure I understand this `window`, or the other ones below in Upheld Ruling or Styx x5?

There's also a lot of `window 30,30` etc; is this really inconsistent, or just adding guide rails to make sure everything stays on track?

JLGarber: Whoops, I knew I was going to miss something. I had to adjust the entire timeline four seconds later because the log I generated the base timeline from was early.

As for why there are so many safety syncs on an otherwise on-rails timeline, we've had situations before where pushes are discovered later. This is an attempt to get out ahead of that.

Issue #4986, opened 2022-10-26
Migrating new trigger style


trim21:
A tracker for migrating all trigger to new object-netRegex style.

currently undone works:

- [ ] GameLog
  - [ ] message
  - [ ] dialog
  - [ ] echo

---

- [ ] documents
  - [ ] KR
  - [ ] CN
  - [ ] TW

quisquous:
I think other missing pieces here are documentation and, once everything is converted, enforcing this style for cactbot built-in triggers.

trim21:
There is a problem...

```ts
{
      type: 'StartsUsing',
      netRegex: NetRegexes.ability({ id: ['6D99', '6D9A', '6DD2'], source: 'Dragon-king Thordan' }),
}
```

this doesn't cause a type error anymore

quisquous:
> There is a problem...

Is this related to #5038? or, what broke this?

trim21:
> > There is a problem...
> 
> Is this related to #5038? or, what broke this?

no, because `{}` is correct type of `NetParam[T]`

trim21:
it can be 
```
{
  type: 'StartsUsing',
  netRegex: {},
}
```

and therefore it can be 

```
{
  type: 'StartsUsing',
  netRegex: NetRegexes.ability({ id: ['6D99', '6D9A', '6DD2'], source: 'Dragon-king Thordan' }),
}
```

PR #5451, opened 2023-05-30
raidboss: convert NetRegexes.echo/message/dialog to params

quisquous:
Part of #4986.

Closes #5452.


quisquous:
I'm not sure about exposing even more variables into the global namespace for user triggers.  Maybe this should be under `Util`?

Also it sure feels like `gameLogCodes` is a variable vs `NetRegexes` which is more of a static class (and thus capitalized, theoretically), but it sure looks funny next to the rest of the exposed variables too.

quisquous:
Given the lack of response from anybody else here, I'm just going to make a decision.

> I'm not sure about exposing even more variables into the global namespace for user triggers. Maybe this should be under Util?

Moved to Util after some thought, and I'm going to go with this approach over #5452.

PR #5452, opened 2023-05-30
raidboss: add gameLogType field for GameLog triggers

quisquous:
This is part of #4986.

This is an alternative PR to #5451.

In order to switch entirely over to `netRegex: { params }` style triggers, this PR adds a `gameLogType` field for `GameLog` triggers only.  If specified, it will add the right `code`.

Downsides: kind of awkward if `code` is already specified (and this will throw an error)

Other options:
* add a bunch more types so you could say `type: 'GameLogEcho'`, but this seemed much more hairy than was worth and the TriggerTypes type is already very very long
* export the codes from some common location, import them everywhere, but this is even more nonsense to import so that user triggers can eval properly

### DIFF
--- a/resources/netregexes.ts
+++ b/resources/netregexes.ts
@@ -1,6 +1,7 @@
 import { NetFields, NetFieldsReverse } from '../types/net_fields';
 import { NetParams } from '../types/net_props';
 import { CactbotBaseRegExp } from '../types/net_trigger';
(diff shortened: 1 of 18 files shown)


quisquous:
@valarnin This isn't urgent by any means, but I'd definitely take any opinions you have re: #5452 vs #5451 as different approaches here.  I'm feeling very indecisive.

PR #5453, opened 2023-05-30
raidboss: convert NetRegexes functions to params

quisquous:
This is part of #4986.  Similar to #5452.

`Test Combatant Cast Enable` was manually tested.

### DIFF
--- a/ui/raidboss/data/04-sb/eureka/eureka_hydatos.ts
+++ b/ui/raidboss/data/04-sb/eureka/eureka_hydatos.ts
@@ -112,19 +112,22 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'BA West Side',
       type: 'Ability',
-      netRegex: NetRegexes.ability({ id: '3956', source: 'Art', target: '[^:]+', capture: false }),
+      // The two sides are far enough apart that even though you see the autos
+      // from the opposite boss, you don't see the player target (which is blank).
+      // This considers you on the west side if Art autos somebody, and the name isn't blank.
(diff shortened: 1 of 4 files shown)


ui/raidboss/data/04-sb/eureka/eureka_hydatos.ts
```diff
@@ -112,19 +112,22 @@ const triggerSet: TriggerSet<Data> = {
...
       run: (data) => data.side = 'west',
     },
     {
       id: 'BA East Side',
       type: 'Ability',
-      netRegex: NetRegexes.ability({
+      netRegex: {
         id: '3957',
         source: 'Owain',
         target: '[^:]+',
```
quisquous, line 128: Also, I have no idea how this is working given that `[^:]+` is meant for parsed log lines and not network log lines, but I'll change that in another PR.

quisquous:
cc @valarnin FYI

ui/raidboss/data/04-sb/eureka/eureka_hydatos.ts
valarnin: There's no reason this wouldn't work. Assuming this targets players, a player's name can never have a `:` character anyways, so it's functionally identical to `target: '[^|]'`. Maybe that ability can target non-player actors and the colon check is actually important though? I have no clue, I've never even stepped into Eureka.

quisquous: Well, since this is building into a regular expression, I didn't want it to match a blank entry, e.g. `|playerid||` kinda thing, where `[^:]+` matched the final bar.

PR #5454, opened 2023-05-30
raidboss: initial P9S triggers

quisquous:


quisquous:
Untested, but look fine in raidemulator.

PR #5456, opened 2023-05-30
raidboss: P10S basic triggers

quisquous:


wexxlee:
Three status effects that could be added:

DDE - "Dæmoniac Bonds" - spread
DDF - "Duodæmoniac Bonds" - partners (1 dps, 1 support)
E70 - "Tetradæmoniac Bonds" - role groups (1 dps and 1 support get a 4-person stack)

These are applied with varying durations, sometimes > 35s.

quisquous:
Thanks! I'll try to add something for that today, since that's an extremely useful callout.

PR #5461, opened 2023-05-31
i18n: Set of FR translations

mooondark:


Bing-su:
Modify only the `03-hw/trial/zurvan-ex.ts` file,
run `npm run sync-files` to create the modified `zurvan-un.ts` file.
Commit the modified file afterwards.

mooondark:
Thanks!! I reverted my changes. I'll do that next time.

I'm not a dev, so I'm learning slowly how to use github.

PR #5464, opened 2023-05-31
raidboss: more P10S triggers

quisquous:


quisquous:
This needs a little bit more of a look in the raid emulator before I merge it, but seems largely ok.

quisquous:
Double checked a bunch of stuff in the raid emulator and made some fixes.  I'll add the bonds PR separately once I run that in game shortly.

PR #5480, opened 2023-06-01
build: Build manifest automatically

valarnin:
This PR replaces the manually-maintained `*_manifest.txt` files with dynamically-generated-at-build logic.

Tested:
- raidboss.html via both browser through websocket and via overlay
- raidemulator.html via browser through websocket
- coverage report and coverage.html (by the way, coverage report's generated `.ts` file fails linting because of EoL being `\n` instead of `\r\n`)
- oopsyraidsy.html and oopsy_summary.html

Removed the manifest text file tests since they're irrelevant with this PR.

Note that currently the order of the emitted lines from `findManifestFiles` is currently whatever the underlying filesystem returns them in. It might make sense to sort these alphabetically?


quisquous:
Oooh this is exciting, thank you.  I will try to look at this at some point, but it might be a little bit.

valarnin:
FWIW the emitted `*.bundle.js` files are identical other than the changed order of imports I mentioned, and I believe that the only potential impact to loading the trigger sets in a different order would be if there was one trigger set's trigger overriding another trigger set's trigger, which I don't think we currently allow in the `main` cactbot code, only via user triggers, which always load after manifest trigger sets anyways.

quisquous:
We ... theoretically don't allow it.  I think technically the tests are looser than they should be, right now we test:
* triggerset ids must be globally unique
* trigger ids must be unique in a given file
* all the trigger ids must share a common prefix

There is technically no test that two files don't share the same common prefix (I can't remember why I didn't do this, maybe there was some issue) or that trigger ids are globally unique across all files (I can't remember why I didn't do this either).  But, we're probably ok for now.

tl;dr: sorry the testing is sloppy, but I think whatever order is fine for built-in triggers

webpack/loaders/manifest-loader.ts
```diff
@@ -1,11 +1,39 @@
 import fs from 'fs';
+import path from 'path';
 
 import webpack from 'webpack';
 
+// This is duplicated from util/manifest.ts
+// For some reason webpack throws a `require() of ES modules is not supported.` error
+// when trying to import? :shrug:
```
quisquous, line 8: Yeah, I wasn't able to fix this either.  This feels like a lot of the issues we had with mocha / etc in the past.  Good enough for now.

PR #5474, opened 2023-06-01
raidboss: P9S full timeline

xiashtra:
Fixes #5475.

Issue #5440, opened 2023-05-28
raidboss: start timelines from InCombat lines instead of engage/autos


quisquous:
From a random idea in: https://github.com/quisquous/cactbot/pull/5432#discussion_r1208169275

This seems like a better way to start timelines.  It'd also let us start only the timelines where it makes sense to do this, as we can't do this for dungeons or door bosses or bozja style zones.

quisquous:
@JLGarber, what do you think about switching things over to using this, both going forward and going back to old timelines as well?

```diff
-0.0 "--sync--" sync /:Engage!/ window 0,1
+0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
```

There's been some conversation about Ravahn maybe adding these lines into the ffxiv plugin itself for fflogs, and we can switch if that ever happens.  I filed https://github.com/ravahn/FFXIV_ACT_Plugin/issues/308 for this.

JLGarber:
Yes, that seems reasonable. Any content where we can't start from in-combat status should have zone seal lines.

quisquous:
I do think this seems pretty good in p9s/p10s imo.  If somebody feels up for it, we could consider going back and touching up older timelines (and make_timeline) to do this.

PR #5481, opened 2023-06-01
raidboss/p9s: fix ability ID of first auto

foophoof:
I'm not 100% sure if this is right, but I think this was meant to be `814B` instead of `814C`, since the latter is `Gluttony's Augur`, and doesn't seem to make sense to have at 7.2 if we also do `-p 814C:12.2`.

Fixing this also seems to make `test_timeline.ts` sync better on my logs.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p9s.txt
+++ b/ui/raidboss/data/06-ew/raid/p9s.txt
@@ -32,7 +32,7 @@ hideall "--sync--"
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
 
 0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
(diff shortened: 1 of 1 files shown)


foophoof:
Nevermind, I see it's `14:` now, dunno why it kept breaking now.

quisquous:
Yeah, re: #5440 I wanted to skip (inconsistent) first autos / Engage! lines and instead (now that we have InCombat lines) start timelines from there, so I was giving it a try here.  Are these times incorrect for you?  What is test timeline saying?

foophoof:
They seem correct now, at least as of #5474, when I tried to test a bit yesterday the entire timeline was red and marked as missed for me but I can't replicate that now, so I probably just did something wrong yesterday

quisquous:
Ok, phew.  If you do run into it again, please file an issue with the log.  It's very possible there's some issue somewhere too!

PR #5512, opened 2023-06-06
raidboss: P9S timeline adjustments

xiashtra:
Change timeline to display/sync on damage events instead of non-damaging casts where applicable.

Should we consider hiding the `Disgorge` casts from the timeline?  They are non-damaging, hidden casts in-game.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p9s.txt
+++ b/ui/raidboss/data/06-ew/raid/p9s.txt
@@ -1,14 +1,14 @@
 ### P9S: Anabaseios: The Ninth Circle (Savage)
 
-# -p 814C:12.2
-# -ii 814B 814D 8150 815D 816E 815E 8165 8166 820B 820C 8222 8223 817F 8171 8173 817A 8176 8188 8189 8143 8187
+# -p 814D:12.6
+# -ii 814B 814C 8150 815D 816D 815E 8165 8166 820B 820C 8222 8223 817F 8171 8172 8179 8176 8188 8189 8143 8186
 # -it "Kokytos"
 
 # 814B unaspected auto
 # 8150 mage auto
 # 815D fighter auto
-# 814D Gluttony's Augur damage
-# 816E Archaic Demolish damage
+# 814C Gluttony's Augur (cast)
+# 816D Archaic Demolish (cast)
 # 815E Uplift (raise walls)
 # 8165 Inside Roundhouse (telegraph)
 # 8166 Outside Roundhouse (telegraph)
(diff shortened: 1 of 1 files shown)


xiashtra:
To make the timeline consistent with using the damage events vs dummy casts.  For what it's worth, I strongly believe that should be the standard for all timelines, as that information is **_far_** more important.

quisquous:
> To make the timeline consistent with using the damage events vs dummy casts. For what it's worth, I strongly believe that should be the standard for all timelines, as that information is **_far_** more important.

I guess I disagree with this, and it's not the way any of the other timelines have been written.  I'm not saying it's not good to do this in some cases, but this is not my general position.

Definitely if there's wildly different damage abilities or multiple hits (e.g. tankbuster 1 tankbuster 2), it makes sense to put the damage line in the timeline (and possibly skip the self-targeted ability entirely, e.g. p11s).  Or, if the timing is wildly different from the cast, like many seconds away.

But for things like raidwides, my understanding is that the mit is snapshot (very often) on the castbar end, and not the damage 0.5s later, and so it makes sense to put the self-targeted ability in the timeline for mit planning purposes.  Additionally, damage rolls through the party and is not always at exactly the same time and so it makes timeline syncing less accurate.  A single boss-targeted ability is slightly more reliable for timeline syncing.

For example, [in this link](https://www.fflogs.com/reports/2hbZMFpx4rycvXYW#fight=8&type=damage-done&view=events&hostility=1), there's a "prepares" line for Gluttony's Augur) at 0:11.907 and damage at 0:12.530 (and rolling).  The damage is already calculated and done at 0.11:907, which is [coincidentally](https://www.fflogs.com/reports/2hbZMFpx4rycvXYW#fight=8&type=casts&view=events&hostility=1) the same time as the boss self-targeted ability.  So, mit that fell off as 0:12.500 would still apply to the damage taken at 0:12.530 because it was already calculated at 0:11.907.

If we wanted to be very pedantic about timelines, we could look for [0x37 lines](https://github.com/quisquous/cactbot/blob/main/docs/LogGuide.md#line-37-0x25-networkactionsync) and (have the make timeline tool?) line up sequence numbers to figure out which corresponded to which ability lines and then use those "prepares" lines as the timeline times.  However, we'd also have to add other syncs around it (like the self-targeted cast) to sync the timeline, as 0x37 lines can't be easily used for timeline syncing.

I'm being longwinded here to say that neither "self-targeted ability" nor "damage ability" are correct times, but "self-targeted ability" is (1) often more correct wrt when the damage is calculated because it lines up with 0x37 lines and (2) often more precise wrt keeping the timeline on track because it doesn't roll through the party.

PR #5527, opened 2023-06-08
i18n: Add p11s Chinese translations

Jwidea:
Im not sure what is `other far` and `other close` in `Dark and Light`, so i'll leave those two for others. Other than that all good.


quisquous:
"other far" and "other close" are just listing out the people who have the same near/far tether as you, in case you're doing some sort of priority scheme of particular people adjusting based on others

Issue #5405, opened 2023-05-23
Meta: 6.4 content organization


quisquous:
Just to try to prevent stepping on toes, if you're thinking about handling any particular piece of content, please respond here just so other folks know.

- Aetherfont (dungeon): @quisquous 
- Voidcast Dias (trial): ???
- Voidcast Dias Extreme (trial): @valarnin 
- P9N (raid): @foophoof 
- P10N (raid): @wexxlee 
- P11N (raid): @JLGarber 
- P12N (raid): @quisquous
- Zurvan Unreal (unreal): @JLGarber

quisquous:
Also, if anybody has time to split the dungeons that changed in 6.4 out into separate files like we have for `sohm_al61.ts`, that would be appreciated.

valarnin:
I'm looking at ex fight, but my parties aren't hitting enrage so I can't do a full timeline yet.

quisquous:
> I'm looking at ex fight, but my parties aren't hitting enrage so I can't do a full timeline yet.

Partial timeline with a TODO is just fine if you don't see one.  I sometimes dig that up from fflogs too, but idk if there's any there yet.

quisquous:
I'm going to also take a stab at the dungeon.

JLGarber:
I will take care of Zurvan Unreal, yes. I might pick up one of the raid encounters depending on how clears go tonight.

foophoof:
I'd be interested in giving P9N a shot, but this is my first time doing a timeline so if you want someone more experienced to try to get it done faster then that's perfectly fine ^^"

(I'll probably give it a try anyways just to learn how this works, so I guess I'll just PR up what I got if nobody else does it first)

valarnin:
> I'd be interested in giving P9N a shot, but this is my first time doing a timeline so if you want someone more experienced to try to get it done faster then that's perfectly fine ^^

More people helping to develop for the content is always appreciated!

The skeletons for the trigger file and timeline file are already in place, so all you need to do is add to them. See the guides here:
https://github.com/quisquous/cactbot/blob/main/docs/TimelineGuide.md
https://github.com/quisquous/cactbot/blob/main/docs/RaidbossGuide.md

Note that you may need to use `npx ts-node ...` instead of just `ts-node ...` for the helper tools for timeline creation, if you don't have `npx` installed globally.

[...]

PR #5534, opened 2023-06-09
raidboss: timeline & initial triggers for Golbez (normal)

### LINKED ISSUE #5497: No Callouts For Normal Golbez Trial
It doesn't make sense to me that there's callouts for Golbez EX but not Golbez normal. I just did the normal one and there's no callouts at all. Is there a reason why it's not in there yet? Thanks.

wexxlee:
Tested in raidemulator and a couple of pulls in-game.
Per my comment in #5405 I'll do another PR tomorrow with some basic oopsy stuff as well.

Edit:  Closes #5497

### DIFF
--- a/ui/raidboss/data/06-ew/trial/golbez.ts
+++ b/ui/raidboss/data/06-ew/trial/golbez.ts
@@ -1,14 +1,133 @@
+import { Responses } from '../../../../../resources/responses';
+import { DirectionOutputCardinal, Directions } from '../../../../../resources/util';
 import ZoneId from '../../../../../resources/zone_id';
 import { RaidbossData } from '../../../../../types/data';
 import { TriggerSet } from '../../../../../types/trigger';
 
-export type Data = RaidbossData;
+export interface Data extends RaidbossData {
+  galeSphereShadows: DirectionOutputCardinal[];
+}
 
 const triggerSet: TriggerSet<Data> = {
   id: 'TheVoidcastDais',
   zoneId: ZoneId.TheVoidcastDais,
   timelineFile: 'golbez.txt',
-  triggers: [],
+  initData: () => {
+    return {
+      galeSphereShadows: [],
+    };
+  },
+  triggers: [
+    {
+      id: 'Golbez Terrastorm',
+      type: 'Ability',
+      netRegex: { id: '8463', source: 'Golbez', capture: false },
+      infoText: (_data, _matches, output) => output.text!(),
+      outputStrings: {
+        text: {
+          en: 'Avoid meteor',
+        },
+      },
+    },
+    {
+      id: 'Golbez Crescent Blade',
+      type: 'StartsUsing',
+      netRegex: { id: '846B', source: 'Golbez', capture: false },
+      response: Responses.getBehind(),
+    },
+    {
+      id: 'Golbez Void Meteor',
+      type: 'StartsUsing',
+      netRegex: { id: '84AC', source: 'Golbez', capture: true },
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'Golbez Binding Cold',
+      type: 'StartsUsing',
+      netRegex: { id: '84B2', source: 'Golbez', capture: false },
+      response: Responses.bleedAoe(),
+    },
+    {
+      id: 'Golbez Black Fang',
+      type: 'StartsUsing',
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/06-ew/trial/golbez.ts
```diff
@@ -1,14 +1,133 @@
...
+          dir2: output[dir2]!() ?? '??',
+          dir3: output[dir3]!() ?? '??',
+          dir4: output[dir4]!() ?? '??',
+        });
+      },
+      outputStrings: {
+        clones: {
+          en: 'Clones: ${dir1}->${dir2}->${dir3}->${dir4}',
+        },
+        ...Directions.outputStringsCardinalDir,
```
quisquous, line 107:
```suggestion
        return output.clones!({
          dir1: output[dir1]!() ?? output.unknown!(),
          dir2: output[dir2]!() ?? output.unknown!(),
          dir3: output[dir3]!() ?? output.unknown!(),
          dir4: output[dir4]!() ?? output.unknown!(),
        });
      },
      outputStrings: {
        clones: {
          en: 'Clones: ${dir1}->${dir2}->${dir3}->${dir4}',
        },
        unknown: Outputs.unknown,
        ...Directions.outputStringsCardinalDir,
```

In general I'd prefer not to have strings come from anywhere but the outputStrings section (even if it's just ???).

ui/raidboss/data/06-ew/trial/golbez.txt
```diff
@@ -1,6 +1,111 @@
...
+172.8 "--middle--" sync / 1[56]:[^:]*:Golbez:84B8:/
+186.7 "Explosion" sync / 1[56]:[^:]*:Golbez:8499:/
+191.1 "Double Meteor" sync / 1[56]:[^:]*:Golbez:84B4:/
+194.2 "--sync--" sync / 1[56]:[^:]*:Golbez:8475:/
+200.3 "--middle--" sync / 1[56]:[^:]*:Golbez:84B8:/
+205.6 "Gale Sphere" sync / 1[56]:[^:]*:Golbez:844D:/
+225.3 "--sync--" sync / 1[56]:[^:]*:Golbez:844E:/
+234.3 "Void Stardust" sync / 1[56]:[^:]*:Golbez:84A3:/
+238.2 "Void Stardust" sync / 1[56]:[^:]*:Golbez:84A5:/
+248.5 "Void Meteor" sync / 1[56]:[^:]*:Golbez:84AC:/ duration 4
```
quisquous, line 63: Is 84A3 the telegraph, 84A5 the first hit, and 84AC the continuing damage? Or, what are these ids? Mostly wondering if this needs some extra text here or an ignored id.  The extreme says `(preview)` on the telegraph, for what it's worth.

quisquous (commented):
Thank you for this!  I really appreciate filling in missing content.

ui/raidboss/data/06-ew/trial/golbez.ts
wexxlee: > In general I'd prefer not to have strings come from anywhere but the outputStrings section (even if it's just ???).

Got it.  The strange thing is that we shouldn't need to include `unknown: Outputs.unknown` in `outputStrings`, as `Directions.outputStringsCardinalDir` already contains an identical property.  But if I omit it, eslint (specifically, rulesdir/cactbot-output-strings throws an error that 'unknown' is not a property in 'output'.

If we want to continue to spread direction-related (and potentially other/future) outputStrings in this manner, should I open an issue to update that lint rule?  (I can take a look at it, but I'm not very familiar with lint.)

ui/raidboss/data/06-ew/trial/golbez.txt
wexxlee: 84A3 is the initial (2.7 sec.) cast that shows the corner meteor marker.  At the conclusion of the cast, the remaining telegraphs are shown.  84A5 is a longer cast, started at the same time as 84A3, that eventually corresponds to damage from that corner meteor exploding.  84A9 (ignored) is damage from the subsequent meteors.
(84AC is Void Meteor, which is a separate mechanic -- the dual tankbuster.  Yeah, these names aren't confusing at all...)

I added (preview) on 84A3 and a small duration on 84A5 to correspond to the subsequent meteor hits.  It's a slightly longer duration than the extreme, because that's how long the 84A9 abilities take to resolve, but it otherwise matches what we're doing there.

PR #5544, opened 2023-06-11
raidboss: Adjust p11s timeline to fix miscounted Styx hits

valarnin:
All Styx casts hit one more time than was listed.

image


quisquous:
Counting is hard <_<

valarnin:
Counting? Sounds like a problem for the healer. (I'm the healer)

PR #5550, opened 2023-06-12
i18n: ko p9~12s, golbez

Bing-su:
I didn't expect to see 'Defamation' here again... 'Defamation' here is not really the 'Defamation', so I've made it different from TEA.


quisquous:
> I didn't expect to see 'Defamation' here again... 'Defamation' here is not really the 'Defamation', so I've made it different from TEA.

At least in English, we call any "extremely large circle explosion that you need to get far away from the group with" a defamation at this point.  We called them defamations in Hello World and everybody calls them defamations in p9s as well.  What do y'all say?  Is it only defamation in TEA or a12?

Bing-su:
I've asked around, and no one knows what 'defamation (명예형)' is (even those who have cleared TEA).

It was always referred to as the 'large circle' or the 'blue marker'. So, it's true that defamation was not a defamation even in TEA.

PR #5551, opened 2023-06-12
raidboss: Support for objects in trigger outputStrings

valarnin:
This is a preliminary PR to eventually allow more customization of outputStrings outputs by users. Eventually I'd like to see player names replaced with something like a player object with a default toString that returns player name, which could maybe be a helper on the `PartyTracker` object.

I have simulated that extremely basically with the `data` object test trigger added in `test.ts`. Users should be able to customize the `outputStrings` entry for `text` to print e.g. `${data.role}`. I'm looking for feedback/thoughts on if this approach makes sense, or if this idea should be approached in a different way, before I fully flesh this out.

I'm not super happy with the logic in `popup-text.ts`, feels like it could have benefited with a helper func or something rather than switch/casing over type in two separate spots like that, or maybe a recursive function?

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -198,7 +198,7 @@ export default class PartyTracker {
         return retVal;
       if (typeof retVal === 'number')
         return retVal.toString();
-      return shortName;
+      return ret.shortName;
     };
 
     return ret;
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -1,5 +1,6 @@
 import { Party } from '../types/event';
 import { Job, Role } from '../types/job';
+import { OutputStringsParamObject } from '../types/trigger';
 
 import Util from './util';
 
@@ -13,6 +14,21 @@ const emptyRoleToPartyNames = () => {
     none: [],
   };
 };
+
+export type BasePartyMemberParamObject = {
+  role: string;
+  job: Job;
+  id: string;
+  name: string;
+  shortName: string;
+};
+
+export type PartyMemberParamObjectKeys = keyof BasePartyMemberParamObject;
+
+export interface PartyMemberParamObject
+  extends OutputStringsParamObject, BasePartyMemberParamObject {
+  toString: () => string;
+}
 
 export default class PartyTracker {
   details: Party[] = [];
@@ -154,5 +170,37 @@ export default class PartyTracker {
 
   nameFromId(id: string): string | undefined {
     return this.idToName_[id];
+  }
+
+  paramObjectFromName(
+    name: string,
+    defaultValue: keyof PartyMemberParamObject,
+    playerNicks: { [name: string]: string },
+  ): PartyMemberParamObject | undefined {
+    const partyMember = this.details.find((member) => member.name === name);
+    if (!partyMember)
+      return;
+
+    const jobName = Util.jobEnumToJob(partyMember.job);
+    const role = Util.jobToRole(jobName);
+    const ret: PartyMemberParamObject = {
+      id: partyMember.id,
+      job: jobName,
+      role: role,
+      name: name,
+      shortName: Util.shortName(name, playerNicks),
+    };
+
+    // Need to assign this afterwards so it can reference `ret`
+    ret.toString = () => {
+      const retVal = ret[defaultValue];
+      if (typeof retVal === 'string')
+        return retVal;
+      if (typeof retVal === 'number')
+        return retVal.toString();
+      return name;
+    };
+
+    return ret;
   }
 }
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -16,7 +16,7 @@ const emptyRoleToPartyNames = () => {
 };
 
 export type BasePartyMemberParamObject = {
-  role: string;
+  role: Role;
   job: Job;
   id: string;
   name: string;
@@ -184,7 +184,7 @@ export default class PartyTracker {
       ret = {
         id: '???',
         job: 'NONE',
-        role: '???',
+        role: 'none',
         name: name,
         shortName: Util.shortName(name, playerNicks),
         toString: () => Util.shortName(name, playerNicks),
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -178,18 +178,28 @@ export default class PartyTracker {
     playerNicks: { [name: string]: string },
   ): PartyMemberParamObject | undefined {
     const partyMember = this.details.find((member) => member.name === name);
-    if (!partyMember)
-      return;
-
-    const jobName = Util.jobEnumToJob(partyMember.job);
-    const role = Util.jobToRole(jobName);
-    const ret: PartyMemberParamObject = {
-      id: partyMember.id,
-      job: jobName,
-      role: role,
-      name: name,
-      shortName: Util.shortName(name, playerNicks),
-    };
+    let ret: PartyMemberParamObject;
+    if (!partyMember) {
+      // If we can't find this party member for some reason, use some sort of default
+      ret = {
+        id: '???',
+        job: 'NONE',
+        role: '???',
+        name: name,
+        shortName: Util.shortName(name, playerNicks),
+        toString: () => Util.shortName(name, playerNicks),
+      };
+    } else {
+      const jobName = Util.jobEnumToJob(partyMember.job);
+      const role = Util.jobToRole(jobName);
+      ret = {
+        id: partyMember.id,
+        job: jobName,
+        role: role,
+        name: name,
+        shortName: Util.shortName(name, playerNicks),
+      };
+    }
 
     // Need to assign this afterwards so it can reference `ret`
     ret.toString = () => {
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -198,7 +198,7 @@ export default class PartyTracker {
         return retVal;
       if (typeof retVal === 'number')
         return retVal.toString();
-      return name;
+      return shortName;
     };
 
     return ret;
--- a/test/helper/test_trigger.ts
+++ b/test/helper/test_trigger.ts
@@ -643,7 +643,10 @@ const testTriggerFile = (file: string, info: TriggerSetInfo) => {
 
           for (const key of keys) {
             for (const param of outputStringsParams[key] ?? []) {
-              if (!Regexes.parse(`\\b${param}\\s*:`).exec(funcStr)) {
+              if (
+                (!Regexes.parse(`\\b${param}\\s*:`).exec(funcStr)) &&
+                (!Regexes.parse(`\\b${param.split('.')[0] ?? ''}\\s*:`).exec(funcStr))
+              ) {
                 assert.fail(
                   `'${id}' does not define param '${param}' for outputStrings entry '${key}'`,
                 );
--- a/test/helper/test_trigger.ts
+++ b/test/helper/test_trigger.ts
@@ -59,6 +59,7 @@ const getFakeRaidbossData = (triggerSet?: LooseTriggerSet): RaidbossData => {
     CanCleanse: () => false,
     CanFeint: () => false,
     CanAddle: () => false,
+    partyMemberParam: (_name) => undefined,
     ...triggerSet?.initData?.() ?? {},
   };
 };
@@ -643,10 +644,16 @@ const testTriggerFile = (file: string, info: TriggerSetInfo) => {
 
           for (const key of keys) {
             for (const param of outputStringsParams[key] ?? []) {
-              if (
-                (!Regexes.parse(`\\b${param}\\s*:`).exec(funcStr)) &&
-                (!Regexes.parse(`\\b${param.split('.')[0] ?? ''}\\s*:`).exec(funcStr))
-              ) {
+              // Only allow one level of prop specification, e.g. `obj.key`
+              // Do not allow `obj.key.subkey`
+              const paramParts = param.split('.');
+              if (paramParts.length > 2) {
+                assert.fail(
+                  `'${id}' specifies an object key ('${param}') with too many parts`,
+                );
+              }
+              const trimmedParam = paramParts[0] ?? param;
+              if (!Regexes.parse(`\\b${trimmedParam}\\s*:`).exec(funcStr)) {
                 assert.fail(
                   `'${id}' does not define param '${param}' for outputStrings entry '${key}'`,
                 );
--- a/types/data.d.ts
+++ b/types/data.d.ts
@@ -1,5 +1,5 @@
 import { Lang } from '../resources/languages';
-import PartyTracker from '../resources/party';
+import PartyTracker, { PartyMemberParamObject } from '../resources/party';
 import { ConfigValue } from '../resources/user_config';
 
 import { SystemInfo } from './event';
@@ -42,6 +42,7 @@ export interface RaidbossData {
   CanCleanse: () => boolean;
   CanFeint: () => boolean;
   CanAddle: () => boolean;
+  partyMemberParam: (name: string) => PartyMemberParamObject | undefined;
 }
 
 export interface OopsyData {
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -40,7 +40,7 @@ export type OutputStrings = { [outputKey: string]: LocaleText | string };
 export type Output = {
   responseOutputStrings: OutputStrings;
 } & {
-  [key: string]: (params?: { [param: string]: string | number | undefined }) => string;
+  [key: string]: (params?: { [param: string]: string | number | object | undefined }) => string;
 };
 
 // The output of any non-response raidboss trigger function.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -35,12 +35,21 @@ export type ZoneIdType = number | null;
 
 export type OutputStrings = { [outputKey: string]: LocaleText | string };
 
+export type OutputStringsParamObject = {
+  [key: string]: string | number | (() => string);
+  toString: () => string;
+};
+
+export type OuptutStringsParams = {
+  [param: string]: string | number | OutputStringsParamObject | undefined;
+};
+
 // TODO: is it awkward to have Outputs the static class and Output the unrelated type?
 // This type corresponds to TriggerOutputProxy.
 export type Output = {
   responseOutputStrings: OutputStrings;
 } & {
-  [key: string]: (params?: { [param: string]: string | number | object | undefined }) => string;
+  [key: string]: (params?: OuptutStringsParams) => string;
 };
 
 // The output of any non-response raidboss trigger function.
--- a/types/trigger.d.ts
+++ b/types/trigger.d.ts
@@ -36,11 +36,14 @@ export type ZoneIdType = number | null;
 export type OutputStrings = { [outputKey: string]: LocaleText | string };
 
 export type OutputStringsParamObject = {
+  // The indexer type requires that `() => string` be a possible value, or else `toString` can't be
+  // defined. Effectively, typescript sees it as `someParamObject['toString']()`, not as a direct
+  // function call of `someParamObject.toString()`.
   [key: string]: string | number | (() => string);
   toString: () => string;
 };
 
-export type OuptutStringsParams = {
+export type OutputStringsParams = {
   [param: string]: string | number | OutputStringsParamObject | undefined;
 };
 
@@ -49,7 +52,7 @@ export type OuptutStringsParams = {
 export type Output = {
   responseOutputStrings: OutputStrings;
 } & {
-  [key: string]: (params?: OuptutStringsParams) => string;
+  [key: string]: (params?: OutputStringsParams) => string;
 };
 
 // The output of any non-response raidboss trigger function.
--- a/ui/raidboss/data/00-misc/test.ts
+++ b/ui/raidboss/data/00-misc/test.ts
@@ -365,7 +365,11 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'Test OutputStrings',
       type: 'GameLog',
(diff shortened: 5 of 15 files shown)


ui/raidboss/data/00-misc/test.ts
```diff
@@ -356,10 +356,28 @@ const triggerSet: TriggerSet<Data> = {
...
+      infoText: (data, _matches, output) => {
+        const obj = {
+          ...data,
+          toString: () => data.me,
+        };
+        return output.text!({ customData: obj });
+      },
+      outputStrings: {
+        text: {
+          en: 'data = ${customData}, data.job = ${customData.job}',
```
quisquous, line 372: This PR is definitely along similar thoughts I've had in the past and I am super positive on this approach.  Thanks for taking the time to do this; I had no idea this was on your radar.  (I run my own static, so don't ever think about people as "roles" or "jobs" and know everybody by name.  I realize this is probably not at all how a lot of other folks play, but it's why things are as they are.)

I think that the follow-up here would be to add some option for default way to display players (defaulting to nicknames to be consistent with existing code) and then other names, and then we can replace calls to `ShortName` with some new "create a player object" function and pass them to output strings.

One question I do have is: I think I agree with this general approach, in terms of providing an object so that job/role/name/nick etc can all be used by a trigger.  However, are there other uses past players? Mostly just trying to understand what other uses cases might look like or if it's really just "players".

Other questions are: I think we have a lot of testing right now to make sure that the loose coupling between calls to `output` and entries in `outputStrings` all match up.  Is there anything we can do here for testing "sub params" or is the general thought that cactbot would not use them, and selection of sub params would be done via options or user overrides most of the time?

I do think some more investigation needs to happen on https://github.com/OverlayPlugin/OverlayPlugin/issues/60 though.  Cactbot doesn't use a ton of job/role triggers and there's definitely periodic issues with the party info being incorrect.  If they start being optional for all triggers then it will start happening more.

types/trigger.d.ts
```diff
@@ -40,7 +40,7 @@ export type OutputStrings = { [outputKey: string]: LocaleText | string };
 export type Output = {
   responseOutputStrings: OutputStrings;
 } & {
-  [key: string]: (params?: { [param: string]: string | number | undefined }) => string;
+  [key: string]: (params?: { [param: string]: string | number | object | undefined }) => string;
```
quisquous, line 43: Should `object` be more specific here, e.g. `{ [key: string]: string }` (which apparently does still let you assign a function to `toString`, although we could be explicit about that too).

test/helper/test_trigger.ts
```diff
@@ -641,7 +641,10 @@ const testTriggerFile = (file: string, info: TriggerSetInfo) => {
 
           for (const key of keys) {
             for (const param of outputStringsParams[key] ?? []) {
-              if (!Regexes.parse(`\\b${param}\\s*:`).exec(funcStr)) {
+              if (
+                (!Regexes.parse(`\\b${param}\\s*:`).exec(funcStr)) &&
+                (!Regexes.parse(`\\b${param.split('.')[0] ?? ''}\\s*:`).exec(funcStr))
+              ) {
```
quisquous, line 647:
```suggestion
              const firstParam = param.split('.')[0] ?? param;
              if (!Regexes.parse(`\\b${param}\\s*:`).exec(funcStr)) {
```

I think this something along these lines is better than duplicating the regex.

ui/raidboss/popup-text.ts
```diff
@@ -380,17 +380,55 @@ class TriggerOutputProxy {
...
-        switch (typeof str) {
-          case 'string':
-            return str;
-          case 'number':
-            return str.toString();
+      if (params !== undefined) {
+        let prop: string | undefined = undefined;
+
+        if (!(key in params) && key.includes('.')) {
+          const parts = key.split('.');
```
quisquous, line 387: Asking this about here and elsewhere, but is `split` correct? (Asking mostly if there should be a join on the remaining bits or an error check that there's no more `.` in the prop.) What should be do if somebody says `foo.bar.baz`? Error? Should we look for `bar` in the `foo` object (what the code does)? Should we look for `bar.baz` inside the `foo` object?

types/trigger.d.ts
valarnin: This typing (and some other added/adjusted typing) are preliminary, My end goal is to have something like `type OutputStringsObject = <some object definition that requires a `toString` implementation, and whose properties must be string or number>`. I opened this PR early in the dev process to make sure that we were on the same page as far as a direction to take this before I invested too much time/effort into the process.

As for `foo.bar.baz`, that would be solved by a recursive function as I mentioned in the PR comment, but I was unsure if that was a desired behavior or not, and it's a bit more effort, so I didn't implement it yet.

ui/raidboss/popup-text.ts
```diff
@@ -380,17 +380,55 @@ class TriggerOutputProxy {
...
+        if (key in params) {
+          const val = params[key];
+          switch (typeof val) {
+            case 'string':
+              return val;
+            case 'number':
+              return val.toString();
+            case 'object':
+              if (val !== null) {
+                const looseVal = val as { [outputName: string]: unknown };
```
quisquous, line 401: I think this `as` is only needed because you're using `object` as a type (which we probably should disallow in eslint).  I think if it were some sort of `{ [key: string]: string }` or even `{ [key: string]: unknown }` this code would be more straightforward.

ui/raidboss/popup-text.ts
```diff
@@ -380,17 +380,55 @@ class TriggerOutputProxy {
...
+
+        if (!(key in params) && key.includes('.')) {
+          const parts = key.split('.');
+          key = parts[0] ?? '';
+          prop = parts[1];
+        }
+
+        if (key in params) {
+          const val = params[key];
+          switch (typeof val) {
```
quisquous, line 394: I don't see any way to make this much easier.  I do think probably `if` statements are more easy to read than nested `switch` statements.  (And probably also more early outs rather than super nested code.)  But, I think this is fine if there's only two levels here.  If you thought we wanted three, then probably a recursive function is needed that point.

ui/raidboss/data/00-misc/test.ts
valarnin: > I realize this is probably not at all how a lot of other folks play, but it's why things are as they are.

I think the biggest case for role or job-based callouts are for party finder groups. That's been the case every time something like this was requested in Discord, at least.

> I think that the follow-up here would be to add some option for default way to display players (defaulting to nicknames to be consistent with existing code) and then other names, and then we can replace calls to `ShortName` with some new "create a player object" function and pass them to output strings.

A default config option is intended, when this is more fully fleshed out.

> One question I do have is: I think I agree with this general approach, in terms of providing an object so that job/role/name/nick etc can all be used by a trigger. However, are there other uses past players? Mostly just trying to understand what other uses cases might look like or if it's really just "players".

I can't think of any other use cases currently, but I deliberately wrote this preliminary code to not tie in to any sort of explicit structure so that it can support any object type.

> Other questions are: I think we have a lot of testing right now to make sure that the loose coupling between calls to `output` and entries in `outputStrings` all match up. Is there anything we can do here for testing "sub params" or is the general thought that cactbot would not use them, and selection of sub params would be done via options or user overrides most of the time?

It's probably possible, but I haven't looked too far into the testing code in general. Tweaks to the test code in this PR were only to allow the new object structure to pass through, for example, and should be tightened in some manner.

> I do think some more investigation needs to happen on [OverlayPlugin/OverlayPlugin#60](https://github.com/OverlayPlugin/OverlayPlugin/issues/60) though. Cactbot doesn't use a ton of job/role triggers and there's definitely periodic issues with the party info being incorrect. If they start being optional for all triggers then it will start happening more.

I have some thoughts on how to handle this issue better, but debugging it and testing fixes is a pain because of how inconsistent the issue is, which puts it further down my list of things I'd like to try and fix myself.

quisquous: > I have some thoughts on how to handle this issue better, but debugging it and testing fixes is a pain because of how inconsistent the issue is, which puts it further down my list of things I'd like to try and fix myself.

If you need assistance being in a party and disconnecting / etc, I'm happy to come help be part of any debugging.

valarnin:
I think this is ready for general review now. I haven't tested the latest changes (ran out of time, I have another obligation in about 5 minutes), but the underlying code should be relatively complete.

I ended up moving the `ShortName` logic to Util for this rather than duplicate it in a third place, and cleaned up the duplicate code, hence the `oopsy`-related files being updated.

resources/party.ts
```diff
@@ -151,4 +167,36 @@ export default class PartyTracker {
...
   }
+
+  paramObjectFromName(
+    name: string,
+    defaultValue: keyof PartyMemberParamObject,
+    playerNicks: { [name: string]: string },
+  ): PartyMemberParamObject | undefined {
+    const partyMember = this.details.find((member) => member.name === name);
+    if (!partyMember)
+      return;
```
quisquous, line 178: This makes me a little worried in the face of the OverlayPlugin bug.  I think this function should never return `undefined`.  It should at least return what it can (name, shortName) and then elide other fields.  And, ideally, if somebody asks for a property that doesn't exist, it should print a warning and pick ~some property rather than nothing.  (Maybe shortName if that exists, and any key after that? Maybe these objects need a `default: string` field in case somebody or some code asks for a non-existent property and we need to fall back on something?)

Doing this would also solve the TODO in the raidboss `test.ts` where you don't get anything if you're not in a party.

resources/party.ts
```diff
@@ -151,4 +167,36 @@ export default class PartyTracker {
...
+    };
+
+    // Need to assign this afterwards so it can reference `ret`
+    ret.toString = () => {
+      const retVal = ret[defaultValue];
+      if (typeof retVal === 'string')
+        return retVal;
+      if (typeof retVal === 'number')
+        return retVal.toString();
+      return name;
```
quisquous, line 197:
```suggestion
      return shortName;
```

resources/party.ts
```diff
@@ -151,4 +167,36 @@ export default class PartyTracker {
...
+      name: name,
+      shortName: Util.shortName(name, playerNicks),
+    };
+
+    // Need to assign this afterwards so it can reference `ret`
+    ret.toString = () => {
+      const retVal = ret[defaultValue];
+      if (typeof retVal === 'string')
+        return retVal;
+      if (typeof retVal === 'number')
```
quisquous, line 195: Should we include boolean as well just for completeness?

types/trigger.d.ts
```diff
@@ -35,12 +35,21 @@ export type ZoneIdType = number | null;
 
 export type OutputStrings = { [outputKey: string]: LocaleText | string };
 
+export type OutputStringsParamObject = {
+  [key: string]: string | number | (() => string);
+  toString: () => string;
+};
+
+export type OuptutStringsParams = {
```
quisquous, line 43: typo

types/trigger.d.ts
```diff
@@ -35,12 +35,21 @@ export type ZoneIdType = number | null;
 
 export type OutputStrings = { [outputKey: string]: LocaleText | string };
 
+export type OutputStringsParamObject = {
+  [key: string]: string | number | (() => string);
```
quisquous, line 39: Mmmm is there support for `() => string` in the code?

ui/raidboss/raidboss_options.ts
```diff
@@ -93,6 +94,7 @@ const defaultRaidbossConfigOptions = {
   TimelineLanguage: undefined as (Lang | undefined),
   TimelineEnabled: true,
   AlertsEnabled: true,
+  DefaultPlayerLabel: undefined as (PartyMemberParamObjectKeys | undefined),
```
quisquous, line 97: Can this be `undefined`? 

```suggestion
  DefaultPlayerLabel: 'shortName' as PartyMemberParamObjectKeys,
```

resources/party.ts
valarnin: I left off boolean because boolean wasn't a valid type for the underlying params object.

ui/raidboss/raidboss_options.ts
valarnin: Since I'm not as familiar with how the config portion works, I elected to follow the pattern used above for `AlertsLanguage` and `TimelineLanguage`. If that's not required, then would it even need the `as` cast at all, e.g. following the `DefaultAlertOutput` example above instead?

quisquous: Oh, `AlertsLanguage` and `TimelineLanguage` can be `undefined`.  If they are set to default, the `setterFunc` in `raidboss_config.ts` sets them to be `undefined`.  `DefaultPlayerLabel` should never be `undefined`, I think?

types/trigger.d.ts
quisquous: Oh, that makes sense.  Can you leave a comment that `() => string` is only valid for `toString`?

resources/party.ts
quisquous: I guess my more general question was if it should be added there as well, or if you think that's an edge case that we won't use.

resources/party.ts
quisquous: Also wanted to add that there are some times where triggers might call out to "stack on (non-party member)" for things like hunts, and so in this case even if OP was 100% working for party members, we'd still want to have reasonable fallback behavior.

quisquous: One final question (and maybe there's no good answer here) is what to do if somebody enters `${data.name} (${data.role})` (e.g. #5554) but role is unknown.  I guess it would just say name twice?  Do we need a separate `nameAndRole` entry that could handle this? (Or maybe I am overthinking and this is overkill.)

types/trigger.d.ts
```diff
@@ -35,12 +35,24 @@ export type ZoneIdType = number | null;
 
 export type OutputStrings = { [outputKey: string]: LocaleText | string };
 
+export type OutputStringsParamObject = {
+  // The indexer type requires that `() => string` be a possible value, or else `toString` can't be
+  // defined. Effectively, typescript sees it as `someParamObject['toString']()`, not as a direct
+  // function call of `someParamObject.toString()`.
```
quisquous, line 41:
```suggestion
  // `() => string is not supported for general parameters, and only as a value for `toString`.
  // The indexer type requires that `() => string` be a possible value, or else `toString` can't be
  // defined. Effectively, typescript sees it as `someParamObject['toString']()`, not as a direct
  // function call of `someParamObject.toString()`.
```

resources/party.ts
valarnin: I'm holding off on this piece until I can look at the OP issue in general, but I think the fallback logic is fuzzy and I'm not 100% sure what would be the best behavior here.

quisquous: I think we need this regardless, based on my comment re: hunts.

What options are you considering for fallback logic? Or, what do you see as the issues?

valarnin: In general, what fallback should be used? Should `job`/`role` fall back to `name` or `shortName`, or just display `???`? If we're displaying something like `???`, should that be locale-'d e.g. `Outputs.unknown`, and if so, passing current locale to this location is complicated.

quisquous: We could just fall back to `shortName` always for all objects (and then pick any field after that if it doesn't exist?), but in a general sense where objects are not always players, that seems a little weird.

One thought is that the object should put a (required?) `default` or `fallback` field in that specifies what to be used if a user or code specifies a sub param that doesn't exist.    Personally, I think `shortName` is a reasonable choice for that field.  I think if there's always a valid fallback, then there's no need for `???` or locale plumbing.

quisquous: Coming back to this PR, I still think that this is the missing piece here that prevents me from landing this PR. I feel like there needs to be some story around "what happens if a sub param is specified that doesn't exist". I think in general, it would be better to show something reasonable rather than `???` regardless of the OP party bug.

valarnin: This is updated with the latest commit (also rebased to `main`). Hopefully this makes sense and lines up with what you're looking for?

types/data.d.ts
```diff
@@ -42,6 +42,7 @@ export interface RaidbossData {
   CanCleanse: () => boolean;
   CanFeint: () => boolean;
   CanAddle: () => boolean;
+  partyMemberParam: (name: string) => PartyMemberParamObject | undefined;
```
quisquous, line 45: Just to avoid polluting data even more, what do you think about making this be a function on `PartyTracker` since `data.party` is already a thing? I'd be fine with `data.party.member(name)` if you want to keep it brief.

I know this would require passing in nicknames and the default field as well, so maybe it's not ideal. Mostly curious if you think that's polluting party tracker with things it shouldn't know about or if that's cleaner.

resources/party.ts
```diff
@@ -14,6 +15,21 @@ const emptyRoleToPartyNames = () => {
   };
 };
 
+export type BasePartyMemberParamObject = {
+  role: string;
```
quisquous, line 19: Can this be `role: Role`?

resources/party.ts
```diff
@@ -155,4 +171,46 @@ export default class PartyTracker {
...
+    playerNicks: { [name: string]: string },
+  ): PartyMemberParamObject | undefined {
+    const partyMember = this.details.find((member) => member.name === name);
+    let ret: PartyMemberParamObject;
+    if (!partyMember) {
+      // If we can't find this party member for some reason, use some sort of default
+      ret = {
+        id: '???',
+        job: 'NONE',
+        role: '???',
```
quisquous, line 187: Given the fallbacks, do you think it would make sense to have all of these be optional other than the name/shortName. That way if somebody has a trigger that says `${player1.role}` and something has gone awry, they'll get player1's name instead of `???`.

ui/raidboss/data/00-misc/test.ts
```diff
@@ -362,10 +362,31 @@ const triggerSet: TriggerSet<Data> = {
...
+    {
+      id: 'Test OutputStrings',
+      type: 'GameLog',
+      netRegex: {
+        line: 'cactbot test outputStrings',
+        code: Util.gameLogCodes.echo,
+        capture: false,
+      },
+      infoText: (data, _matches, output) => {
+        // TODO: This doesn't work unless you're in a party, because party tracker is empty.
```
quisquous, line 374: Maybe `partyMemberParam` could handle this case? Or is that too much plumbing to get the current player in there too? (oopsy already does this dance in `PlayerStateTracker.OnPartyChanged`)

quisquous (commented):
Thank you!

types/data.d.ts
valarnin: I think the only way I'd be okay with this would be to pass the whole `RaidbossOptions` object in, rather than passing in individual values, and AFAIK that would cause issues with oopsy and jobs since their config options don't have that parameter?

ui/raidboss/data/00-misc/test.ts
valarnin: I feel like if we're going to fix this cactbot-side, we should do so at a higher level e.g. in `popup-text.ts`'s party change handler?

I don't see the point in fixing the problem here when it's also a problem elsewhere, e.g. in any trigger that's role-based which you might encounter while doing content solo unsync'd.

But overall I'd rather fix this upstream in OverlayPlugin or in the base FFXIV_ACT_Plugin.

resources/party.ts
quisquous: I was saying something different; there's only two states, either we know everything or we only know name. I'm suggesting to make all the fields optional and possibly undefined so that in the case where we only know name, we only need to fill in name and shortName. That way if somebody specifies a missing field (that would normally be filled in but maybe it's the OP bug so it's not) then instead of getting '???' or 'NONE' when that's not useful, they would at least get the name. In other words, only fill in real values and never use dummy values or said differently never add something to this object that you wouldn't want displayed in a trigger.

ui/raidboss/data/00-misc/test.ts
quisquous: Makes sense. Let's defer this for now then. Fixing this in OP or FFXIV_ACT_Plugin would definitely be much better.

types/data.d.ts
quisquous: oopsy has a nicknames field. It could arguably grow the default field and so oopsy could say people's roles rather than names. jobs doesn't really have a need for either. So, maybe it could take an optional object that has nick name and default that both `RaidbossOptions` and `OopsyOptions` is shaped like. (Or an object with optional nick name and default?)

quisquous: My apologies if I derailed this PR with these comments. Mostly I just am trying to be cautious about adding unnecessary things to `data` since it's something shared with users. Happy to discuss more if you have time!

I'm guessing you're just very busy these days; please let me know if there's something I can do to help this along, as I'm still really excited about this change and being able to let users have more custom output for players.

valarnin: Yeah, I've been very busy the last month or so, so I haven't had time to circle back around to this PR and figure out how to get things to pass through well.

I'm hoping to have time in the next week or two, but we'll see how work goes 🤷

resources/party.ts
quisquous: Addressed this in 5121b9926bc94eacbdca86b4616cdcc6645fb511

types/data.d.ts
quisquous: Addressed this in c77169acd05e99a6a90b660b393542402219f61b

resources/party.ts
quisquous: Yes, although I also needed to do e9fcc9690cf24e972e8356c4143278d136a91634 in order for the "no prop specified" case to work.

valarnin:
Closing.

PR #5554, opened 2023-06-13
raidboss: add role callout to p11s dark & light tethers

wexxlee:
Per a comment in discord.  Since many strats rely on positioning based on your partner's role, for PUG/PF groups where party members' roles aren't known by name, this added info would be useful.

Since the mechanic is designed to always tether tanks & healers to a DPS, calling out your partner's role as 'dps' to a tank/healer is arguably unnecessary -- but since role pairings could become muddled for non-standard comps or if there's dead bodies (?), it's probably still a good idea.


quisquous:
This seems good.  Once #5551 lands we can revert this and have it use player objects instead, and folks can set an option for what they want.

PR #5555, opened 2023-06-13
raidboss: update ARR/HW/SB timelines to use InCombat lines

wexxlee:
Intending this PR to finish out the remaining work from #5440 .  Marked as draft pending discussion below.

Right now, it includes timeline updates for everything in 02-arr, using 0x104 lines to start the timeline vs. engage or zone-seal messages where it makes sense.  There's a few of the coils fights where there are various pre-boss add encounters, so using 0x104 lines -- at least by themselves -- will produce inaccurate results.  I've added comments where needed.

Before I start tackling later expansions, this issue of using 0x104 lines in encounters with pre-boss adds will become a more common issue, particularly in dungeon and alliance content (some of the alex raids as well, and doubtless there are others).  There's probably at least two options, maybe more, but wanted to discuss approach before I dig in further:

1.   Continue to use engage or zone-seal messages (as-is) where 0x104 lines have the potential to start the timeline early.  (Easiest)
2.   Use a combination of engage/zone-seal messages for the initial timeline encounter, and use 0x104 lines with large windows for later encounters.  While this might be viable in some current content where spacing between bosses and mobs is at least somewhat predictable, it becomes inaccurate as soon as the content can be unsync'd (if not sooner).  It would also require a pretty intensive re-look at older fight content to try to figure out exactly how to space out the timeline entries between combat and what appropriate look-back windows for 0x104 lines would be.

TLDR: leaning toward option 1 since I *think* this is nice-to-have work and not break-fix?


ui/raidboss/data/02-arr/raid/t12.txt
```diff
@@ -10,7 +10,7 @@ hideall "Scorched Pinion"
 
 ### Phase 1: Blackfire/Whitefire
 0.0 "Start"
-0.0 "--sync--" sync /:Engage!/ window 0,1
+0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
 2.5 "--sync--" sync / 1[56]:[^:]*:Phoenix:368:/ window 3,0
```
quisquous, line 14: I think in general (here and other timelines) we could also remove the initial auto (which is often inconsistent) and add a larger sync on the first ability.

quisquous:
> Use a combination of engage/zone-seal messages for the initial timeline encounter, and use 0x104 lines with large windows for later encounters. 

I think this is not possible, as timelines automatically end when out of combat (unless the trigger set is flagged to not do so).  So, if there is an initial encounter and it ends, then the timeline will reset and I don't think what you're suggesting is possible.

That said, I think (1) is the correct answer as well.

I thiiiink there should be no case where we use `Engage` lines anymore.  If a fight could use in combat lines, it doesn't need engage.  There may be places (e.g. t6 or whatever) where we could use Engage but couldn't use in combat, but I think zone sealing is better there if we can.  I'm not sure if there's anything that doesn't have a zone seal and also can't use in combat lines.

wexxlee:
> There may be places (e.g. t6 or whatever) where we could use Engage but couldn't use in combat, but I think zone sealing is better there if we can. I'm not sure if there's anything that doesn't have a zone seal and also can't use in combat lines.

I haven't gone through exhaustively yet to look at all the use cases, but that tracks with what I've seen in the early content: Engage is always paired with zone-seal.  I'll keep an eye out for edge cases where using InCombat or zone-seal exclusively won't work though.

ui/raidboss/data/02-arr/trial/cape_westwind.txt
```diff
@@ -1,6 +1,8 @@
 ### Cape Westwind
 # -ii 0A 2CD 2CE 194 14 -p 474:204 478:610
 
+# Trial removed in Patch 6.1
```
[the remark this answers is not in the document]
wexxlee: IIRC, the only reason we keep old content around is if there's an int'l version that might still use it, right?  And I think all the int'l versions are past 6.1?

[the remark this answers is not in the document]
xiashtra: We might want to keep this one around as long as the Timeline Guide references it.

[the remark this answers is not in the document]
wexxlee: If we did want to keep it for reference purposes (and don't want to do a rewrite of the Timeline Guide right away), we can remove the triggers and timeline from the data folder and the manifest, and readd the timeline into the docs folder (call it sample_timeline.txt or something like that).

wexxlee:
> I think this is not possible, as timelines automatically end when out of combat (unless the trigger set is flagged to not do so).

Been thinking about this more, and it looks like most (all?) timelines hare a reset/jump 0 when ever the zone is no longer sealed, e.g.:

`0.0 "--Reset--" sync / 00:0839::.*is no longer sealed/ window 10000 jump 0`

Would it make sense to either replace (or perhaps just supplement) these with a corresponding InCombat regex for when the player is out of combat?  E.g.:

`0.0 "--Reset--" sync / 104:[^:]*:0($|:)/ window 10000 jump 0`

Especially in multi-zone instances and older content, I'm guessing this might be a case-by-case thing, but in newer raid/trial content where there is only a single combat encounter, it might have some usefulness?  Or just redundant at this point?

quisquous:
Yes true, reset lines are redundant in anything that does not have `resetWhenOutOfCombat: false,` set on the trigger set.  We could remove them at this point too, I think!

ui/raidboss/data/02-arr/trial/cape_westwind.txt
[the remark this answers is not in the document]
quisquous: We can always link to it in git history by commit hash too.  No need to keep it around.

Honestly, that timeline guide needs to be rewritten anyway, so <_<

ui/raidboss/data/03-hw/raid/a3n.txt
```diff
@@ -8,10 +8,8 @@ hideall "--sync--"
 # -ii 1305 12FA
 
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 10000 jump 0
-0.0 "--Reset--" sync / 00:0839::.*is no longer sealed/ window 10000 jump 0
 
-0 "Start" sync /Engage!/ window 0,1
-0.0 "--sync--" sync / 00:0839::Condensate Demineralizer .9 will be sealed off/ window 10000,10000
+0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
```
quisquous, line 12: a3n has pre-boss adds as well

ui/raidboss/data/03-hw/raid/a5s.txt
```diff
@@ -10,9 +10,8 @@ hideall "Relaxant"
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
 
 ### Hummel and Friends
-1000.0 "--sync--" sync / 00:0839::The Clevering will be sealed off/ window 1000,0
-1001.8 "--sync--" sync / 1[56]:[^:]*:Faust:367:/ window 1002,0.5
-1009.8 "Kaltstrahl x1" sync / 1[56]:[^:]*:Faust:16CC:/
+1000.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 1000,1
```
quisquous, line 13: I think it'd be better to leave the zone seal here, rather than incorrectly starting at t=1000 for the second half of the fight.  This is the same as a9s, imo.

ui/raidboss/data/03-hw/raid/a7s.txt
```diff
@@ -3,6 +3,8 @@
 # Shasta's reddit: /u/shastaxc
 # Last Updated: 5/19/2016
 
+# Revised in http://github.com/quisquous/cactbot to use InCombat lines
```
quisquous, line 6: Some day maybe I or somebody else will fix up this timeline.  I suspect the jails are map effect lines? (Or hope?) and so maybe it's possible now.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
```diff
@@ -14,7 +14,7 @@ hideall "--sync--"
 
 0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
 
-0.0 "--sync--" sync /:Engage!/ window 0,1
+0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
 1.8 "--sync--" sync / 1[56]:[^:]*:Sophia:19C3:/ window 1.8,2
```
quisquous, line 18: I think this auto can be removed.

quisquous:
also cc @JLGarber FYI and for any opinions

ui/raidboss/data/02-arr/trial/cape_westwind.txt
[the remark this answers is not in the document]
wexxlee: > We can always link to it in git history by commit hash too. No need to keep it around.
> 
> Honestly, that timeline guide needs to be rewritten anyway, so <_<

Done.  I was already attempting a (modest) enhancement to the timeline guide based on these changes and committed it before I saw your comment.  It should be in the latest two commits, if you have a chance to look.  Otherwise, will work through these most recent comments and then move on to sb/shb/ew.

ui/raidboss/data/03-hw/raid/a7s.txt
wexxlee: It's definitely a style outlier from what we have now. :-)  I'll add a todo in the file since I'm already in the sauce, and maybe someday...

ui/raidboss/data/03-hw/raid/a10n.txt
```diff
@@ -6,10 +6,8 @@
 hideall "--Reset--"
 hideall "--sync--"
 
-0.0 "--Reset--" sync / 00:0839::.*is no longer sealed/ window 100000 jump 0
-
-0.0 "--sync--" sync / 00:0839::The Excruciationator will be sealed off/ window 0,1
-7.0 "Goblin Rush" sync / 1[56]:[^:]*:Lamebrix Strikebocks:1A17:/
+0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
```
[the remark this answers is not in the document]
wexxlee: Oh shoot - I was manually checking all the instances up to a3 through yesterday, and somehow in the last 24 hours, I forgot that step (blaming it on raid brain).  I'll recheck all the hw stuff and fix, and will make sure to do that on the next batch.

wexxlee:
> Yes true, reset lines are redundant in anything that does not have `resetWhenOutOfCombat: false,` set on the trigger set. We could remove them at this point too, I think!

I removed the reset lines based on zone-unseal messages, but I wasn't sure about the 0x33 ones, since that still seems to be the default in most timelines.  It sounds like those are redundant too, so is there any reason to keep them around?  If not, I'll go back and remove those as well (minus any content like Lost City (Hard) and O5N/S that has that property).  We should probably also remove them from the shell that we use for new content too then.

ui/raidboss/data/03-hw/trial/sophia-ex.txt
wexxlee: Wasn't sure if this was an auto since it wasn't on the -ii list; but regardless, since there's a named ability at 12.0, makes sense.

quisquous:
> > Yes true, reset lines are redundant in anything that does not have `resetWhenOutOfCombat: false,` set on the trigger set. We could remove them at this point too, I think!
> 
> I removed the reset lines based on zone-unseal messages, but I wasn't sure about the 0x33 ones, since that still seems to be the default in most timelines. It sounds like those are redundant too, so is there any reason to keep them around? If not, I'll go back and remove those as well (minus any content like Lost City (Hard) and O5N/S that has that property). We should probably also remove them from the shell that we use for new content too then.

Yeah, I think we could remove reset lines in general other than the ones with the reset false property.

ui/raidboss/data/04-sb/raid/o4s.txt
```diff
@@ -3,12 +3,10 @@
 hideall "--Reset--"
 hideall "--sync--"
 
-0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
-
 # -ii 23F2 23F8 2400 240C
 
 # Exfaust
-3.8 "--sync--" sync / 1[56]:[^:]*:Exdeath:23F2:/ window 3.8,0
+0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1
```
quisquous, line 9: o4s has a door boss.  It uses the auto here to start the timeline for Exdeath vs Neo Exdeath.

ui/raidboss/data/04-sb/raid/o6s.txt
```diff
@@ -3,9 +3,10 @@
 hideall "--Reset--"
 hideall "--sync--"
 
-0.0 "--Reset--" sync / 21:........:4000000F:/ window 100000 jump 0
+# Note: Although there is a 18 sec. delay between the chatlog line and the boss being targetable
```
quisquous, line 6: Ah, nice.  Did you go test this, I'm guessing?

wexxlee: > Ah, nice. Did you go test this, I'm guessing?

Yup.  There were a number of other omega fights that were also started based on chatlog lines - o7, o8, o12s, maybe another one or two.  I swapped in the 0x104 sync and saw in-game that it was causing the first sync ability to jump forward by what seemed to be about ~0.5s.  (e.g. a 12.5 sync would actually seem to sync at 12.0, causing the timeline to jump forward by 0.5).  But I reverted and noticed some variability even using the chatlog lines.

I also compared some of the 0x104 timelines using make_timeline against what we already have, and the drift was actually much smaller (~0.1s/0.2s).  Given that, I *think* we're probably okay to keep the timelines as-is and not regenerate them based on the new 0x104 start?  🙏

ui/raidboss/data/04-sb/raid/o4s.txt
wexxlee: > o4s has a door boss. It uses the auto here to start the timeline for Exdeath vs Neo Exdeath. EDIT: I guess technically this will still work as the Neo auto will still resync the timeline. I just think it looks a little weird when the timeline is like "hey here's the door boss jk it's the final boss" and it'd be better to just not use a combat line here.

I was thinking the same as your edit - but yeah, I guess that will look a little awkward.

If we still want to avoid sync's/starts based on autos, and not have those few seconds of UI dystopia, we could use a combat line for both the door and final bosses, and shorten the sync on the second one so that it only looks back 200 seconds instead of all the way to 0.  And then add a jump to 800 on the AddedCombatant line for Neo Exdeath?  (Granted, I may be over-engineering this, but since the door boss issue is going to come up at least 6 more times before this PR is done, figured I'd propose it now...)

quisquous: I can go either way.  I think either of these approaches are good:
* just don't start the timeline until we know for sure
* start some "we're not sure" shared section with "Door Boss Ability?" "Final Boss Ability?" lines and then jump to the right one when we know.

Also, re: before this PR is done, for my own sake of reviewing this, once o4s is sorted out, can I land this one, and then the other ones be done in a separate PR?

ui/raidboss/data/04-sb/raid/o6s.txt
quisquous: Yeah, I think we're ok.  I usually (try to) double check timelines when I write them to make sure they're when combat is starting and not based on the auto or an inconsistent chat log time.  If anything is broken, we can just fix that when it comes up.

JLGarber:
Nothing to add at this point, I think this is a good change.

ui/raidboss/data/04-sb/raid/o4s.txt
wexxlee: > I can go either way. I think either of these approaches are good:
> 
> * just don't start the timeline until we know for sure
> * start some "we're not sure" shared section with "Door Boss Ability?" "Final Boss Ability?" lines and then jump to the right one when we know.
> 
> Also, re: before this PR is done, for my own sake of reviewing this, once o4s is sorted out, can I land this one, and then the other ones be done in a separate PR?

I lean toward your first option.  I know we do the "not sure" stub elsewhere, but the autos occur early enough, and there's a long enough delay before real abilities start, that I think certainty is better than a stub.  I'll revert the o4s stuff and mark the PR as ready.

PR #5562, opened 2023-06-14
raidboss: adjust p10s bonds timing, message severity

quisquous:
Some of the bonds messages (especially bonds1 and bonds4) could use a much earlier warning, so add a custom delay/duration for each of the four different bonds mechanics.

There's also a lot of info that lines up with bonds warnings so bump some priorities around so there's less info overload.


Legends0:
I'm not sure if it's just me but this change makes it more difficult to discern what it's trying to say when this happens (which occurs first):
image

quisquous:
That looks like an error to me.  It should say `Spread => Partners`, then `Partners`.

quisquous:
Yeah ok, there's definitely some bugs here.  Let me see if I can fix them or if I should revert.

quisquous:
Put up #5572

PR #5570, opened 2023-06-15
raidboss: add p10s Jade Passage trigger

YuzukiTsuru:
add basic p10s Jade Passage trigger

It would be nice if the position of the laser could be displayed like 'stand on lines' or 'stand in box' refer to the pattern of the floor. but I can't find a way to do it.


Legends0:
> It would be nice if the position of the laser could be displayed like 'stand on lines' or 'stand in box' refer to the pattern of the floor. but I can't find a way to do it.

I took a quick look at it and created #5574 .

PR #5572, opened 2023-06-15
raidboss: p10s fixups after #5562

quisquous:
In attempting to clean up messages that don't make sense, e.g. "Partners => Role Stack" I removed the condition check that prevented it from being double called out.

Also, move the counter to the reset location.  This is not a logic change, but it makes sense to have the counter on the cast.


quisquous:
@Legends0 please take a look (thank for the bug report)

Legends0:
Looks like this resolves it in raidemulator:
image

PR #5573, opened 2023-06-15
raidboss: add delay to p9s limit cut 'soak tower' alerts

### LINKED ISSUE #5566: Adjust p9s tower timing
Re: #5511 (and me making assumptions about other people's comments in discord), I think the "Soak Tower" call for your tower comes out ALMOST right as the previous tower is being soaked.  It can feel a little confusing because it's so early in the sense of "should I be soaking that one???".  If the soak call were a second later (or triggered off of the tower damage itself?), then maybe it would feel less confusing because the previous tower would be done and it'd be clear it meant "soak the next one".

wexxlee:
Adds a one-second delay to the 2nd through 4th 'soak tower' calls during p9s limit cut.
Closes #5566

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p9s.ts
+++ b/ui/raidboss/data/06-ew/raid/p9s.ts
@@ -13,6 +13,8 @@ export interface Data extends RaidbossData {
   dualityBuster: string[];
   lastDualspellId?: string;
   limitCutNumber?: number;
+  lc1DashPlayer?: number;
+  lc1SoakPlayer?: number;
   combination?: 'front' | 'rear';
   seenChimericSuccession?: boolean;
   levinOrbs: {
@@ -465,15 +467,17 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: { id: '817D', source: 'Kokytos', capture: false },
       condition: (data) => data.limitCutDash === 0,
       alertText: (data, _matches, output) => {
+        delete data.lc1DashPlayer;
+        delete data.lc1SoakPlayer;
         const activePlayers = limitCutPlayerActive[data.limitCutDash];
         if (activePlayers === undefined)
           return;
-        const [dashPlayer, soakPlayer] = activePlayers;
-        if (dashPlayer === undefined || soakPlayer === undefined)
+        [data.lc1DashPlayer, data.lc1SoakPlayer] = activePlayers;
+        if (data.lc1DashPlayer === undefined || data.lc1SoakPlayer === undefined)
           return;
-        if (data.limitCutNumber === dashPlayer)
+        if (data.limitCutNumber === data.lc1DashPlayer)
           return output.dash!();
-        else if (data.limitCutNumber === soakPlayer)
+        else if (data.limitCutNumber === data.lc1SoakPlayer)
           return output.soak!();
         return;
       },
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p9s.ts
```diff
@@ -465,15 +467,17 @@ const triggerSet: TriggerSet<Data> = {
...
       alertText: (data, _matches, output) => {
+        delete data.lc1DashPlayer;
+        delete data.lc1SoakPlayer;
         const activePlayers = limitCutPlayerActive[data.limitCutDash];
         if (activePlayers === undefined)
           return;
-        const [dashPlayer, soakPlayer] = activePlayers;
-        if (dashPlayer === undefined || soakPlayer === undefined)
+        [data.lc1DashPlayer, data.lc1SoakPlayer] = activePlayers;
+        if (data.lc1DashPlayer === undefined || data.lc1SoakPlayer === undefined)
```
quisquous, line 476: I don't understand what the changes to this trigger are doing.  It seems like there's no change here, and the values of dash/soak player are rewritten in other triggers.  Should the changes to this trigger be un-done?

ui/raidboss/data/06-ew/raid/p9s.ts
```diff
@@ -503,16 +507,24 @@ const triggerSet: TriggerSet<Data> = {
...
           return;
-        const [dashPlayer, soakPlayer] = activePlayers;
-        if (dashPlayer === undefined || soakPlayer === undefined)
+        [data.lc1DashPlayer, data.lc1SoakPlayer] = activePlayers;
+      },
+      delaySeconds: (data) => {
+        // delay 'soak tower' call by 1 second to prevent confusion due to ability timing
+        if (data.lc1SoakPlayer !== undefined && data.limitCutNumber === data.lc1SoakPlayer)
+          return 1;
+        return 0;
```
quisquous, line 520:
```suggestion
        return limitCutPlayerActive[data.limitCutDash][1] === data.limitCutNumber ? 1 : 0,
```

Sorry, just looking at this and wondering why we need an extra variable or two on `Data`.  It seems like there's not a need to store anything extra, or everything is already there? Would something like this be sufficient to solve this problem?

wexxlee: > Sorry, just looking at this and wondering why we need an extra variable or two on `Data`. It seems like there's not a need to store anything extra, or everything is already there? Would something like this be sufficient to solve this problem?

I added the additional vars on data to allow preRun to figure out who was getting dashed/soaking, so that could then be used in delaySeconds, and the logic in alertText was then simplified.  But this approach works too; it would just mean moving the preRun logic back down to alertText.

ui/raidboss/data/06-ew/raid/p9s.ts
wexxlee: Given the changes to Later Dash/Tower Combo (see comment below) and using the vars on `data{}`, I just updated this for consistency.  But using the suggested code below and eliminating the need for the vars on data makes this obsolete, so I'll remove it.

quisquous:
Thanks for the simplification; mostly I just don't want to add extra stuff to `Data` if it doesn't have to be there; mostly more state there means more to think about in any trigger, or interactions between trigger, or redundant info not being kept up to date if other fields change.

PR #5576, opened 2023-06-17
raidboss: fix inconsistent "soon" timeline colors

### LINKED ISSUE #627: raidboss: inconsistent "soon" colorings
image

When adding the timeline for ifrit story mode in 8a38ef3, I noticed that in the second phase (hp between 50% and 70%) that the timers ended up being colored weirdly.

I suspect that maybe when syncs happen the timer expires soon value is not reset, and so it's using the old value prior to reset.  This probably doesn't matter except in tight-ish loops like this.

quisquous:
Largely the issue here is that each timeline bar has an id, and the timeout to recolor a bar to be "soon" was looking by id.  If the timeline ever got adjusted due to syncs or jumps, then the soon timer would end up firing for bars at different times.

To fix this, instead have the soon timer refer to the html element directly rather than an indirect id.
Track and clear this (to avoid outstanding useless timeouts) if the bar is ever removed.

While we're here, also refactor the expiring timeout into the same structure.

Fixes #627.

### DIFF
--- a/ui/raidboss/html_timeline_ui.ts
+++ b/ui/raidboss/html_timeline_ui.ts
@@ -1,4 +1,5 @@
 import { Lang, langToLocale } from '../../resources/languages';
+import { UnreachableCode } from '../../resources/not_reached';
 import TimerBar from '../../resources/timerbar';
 import { LooseTimelineTrigger } from '../../types/trigger';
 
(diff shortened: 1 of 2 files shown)


quisquous:
If anybody has any time to just test this and make sure nothing looks weird with timelines, I'd appreciate it, otherwise I'll just run with this next week and land it after that if I don't see anything.  It's probably ok, but this code doesn't get changed much and so just being careful about breaking things.

JLGarber:
I'll try to check this for Tuesday evening assuming Ravahn gets things set up in time.

JLGarber:
Everything looked fine tonight. Nothing went weird on jumps or duration windows that I noticed.

PR #5577, opened 2023-06-17
raidboss: add `forcejump` timeline command

### LINKED ISSUE #5144: raidboss: new timeline keyword `loop`
# Motivation / Proposal

I've talked about this a little bit, but wanted to add a new timeline keyword `loop`.  The idea being that I think one of the cumbersome things about timelines is adding extra times as a "loop lookahead".

For example, p1n:
https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/06-ew/raid/p1n.txt#L71-L80

```txt
# Repeat
360.0 "--middle--" sync / 1[56]:[^:]*:Erichthonios:65F5:/ window 50,50 jump 188.3
370.3 "Aetherflail Left/Aetherflail Right" #sync / 1[56]:[^:]*:Erichthonios:(6DA3|6DA2):/
370.4 "Powerful Fire/Powerful Light" #sync / 1[56]:[^:]*:Erichthonios:(65EC|65ED):/
381.4 "Aetherflail Right/Aetherflail Left" #sync / 1[56]:[^:]*:Erichthonios:(6DA2|6DA3):/
381.5 "Powerful Fire/Powerful Light" #sync / 1[56]:[^:]*:Erichthonios:(65EC|65ED):/
385.9 "--knockback stack--" #sync / 1[56]:[^:]*:Erichthonios:6D15:/
391.2 "Pitiless Flail" #sync / 1[56]:[^:]*:Erichthonios:65E5:/
398.3 "True Holy" #sync / 1[56]:[^:]*:Erichthonios:65E7:/
404.5 "--middle--" #sync / 1[56]:[^:]*:Erichthonios:65F5:/
```

The intention of these blocks is "when you get to 360, jump back to 188.3, and anything you see after 360, show it as if you jumped".  In other words, any time at 360+X would show 188.3+X, for all positive values X.

I think we could replace this with something much more reasonable like the following.

```txt
360.0 "--middle--" sync / 1[56]:[^:]*:Erichthonios:65F5:/ window 50,50 loop 188.3
```

...and have `timeline.ts` handle the "loop lookahead portion" and behave identical to the above portion.

# Questions

As proposed, `loop` can behave exactly as `jump` does, but showing different lookahead.  However, this creates some additional questions.  What if the sync never occurs?  If we hit the same loop ability again with no sync, does it just wrap back around?

I think there are other ways this could behave.  We could just have `360.0 loop 188.3` with no window, and it just jumps back unconditionally as well as does the loop lookahead.  It would sync to anything on the original loop as needed.  (And if that needed a bigger window, we could add one there as well.)

I'm sure there are other ways that this could work too.

Thoughts? cc @JLGarber @valarnin @panicstevenson

quisquous:
Rather than adding a `loop`, add a variant of `jump` called `forcejump` to the timeline language.  If the sync does not occur, the jump will always happen when the time passes. This allows the timeline controller to automatically add loop lookahead rather than manually specifing it.

There's some manual testing of this in the Frankenstein manual `test.txt` test case.

To track this, `nextEvent` now has a whole bunch more state to track where it is in the lookahead so that when adding additional events as timer bars, it doesn't duplicate its efforts.

Additionally, since the same event could be added twice (or more) times on the screen, they need to have different ids (now strings instead of numbers, that include the "jump count" of the number of times we've gone through a lookahead jump to differentiate them from their pre-jump selves). Similarly, they also need different sort keys, although this is kind of a hack for now.

This also adds some grumpy documentation of a bunch of issues found in the code and potential followups.
Sorry that this code is a mess, but I didn't want to rewrite the world at this time.

Closes #5144.

### DIFF
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -123,6 +123,9 @@ it will behave exactly like a normal `jump`.
 If the time passes this line,
 then it will jump as if it had synced exactly on time.
 This is not handled specially in `test_timeline`, which expects the sync to be correct.
+If the `window` extends past the `forcejump` time,
+this "overhang window" will still be respected even after force jumping
+until the next sync or jump occurs.
 
 ### Commands
 
@@ -274,7 +277,7 @@ Here's an example of using cactbot's tools to make a timeline file for Cape West
 This is pretty straightforward and only requires one person to test, so is a good first example.
 
 Note that the Cape Westwind trial was removed in Patch 6.1,
-and the timeline has since been removed from cactbot.  
+and the timeline has since been removed from cactbot.
 However, you can view the original timeline [here](https://github.com/quisquous/cactbot/blob/aa38bdf8f2551a504e1d3f595cd266d3baa193f2/ui/raidboss/data/02-arr/trial/cape_westwind.txt).
 
 ### Run the fight a few times
@@ -913,7 +916,7 @@ the timeline should start when combat begins,
 and should reset on a wipe or when the player is out of combat.
 
 However, in dungeons for example, the player is often in combat
-with mobs before the timeline should begin for the first boss encounter.  
+with mobs before the timeline should begin for the first boss encounter.
 For that matter, there are also several boss encounters in each dungeon.
 In those situations, we need discrete timelines for each boss encounter,
 and each boss's timeline should start only once that boss encounter begins.
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -76,6 +76,8 @@ Every timeline entry begins with the ability time and the ability name.
 
 `Number "String" sync /Regex/ (window Number,Number) (jump Number) (duration Number)`
 
+`Number "String" sync /Regex/ (window Number,Number) (forcejump Number) (duration Number)`
+
 The parentheses here indicate optionality and are not literal parentheses.
 
 Number can be an integer, e.g. `34`, or a float, e.g. `84.381`.
@@ -109,6 +111,18 @@ If you jump to time 0, the timeline will stop playback.
 This is usually used for phase pushes and loops.
 There does not need to be a timeline entry for the time you jump to,
 although it is very common to have one.
+
+`forcejump Number` tells the timeline playback that there will always be a jump here
+regardless of whether the sync is encountered.
+This is intended for loops that will always be taken in an encounter.
+When this is used, no "lookahead" loop unrolling is needed,
+and the timeline will use the `forcejump` destination to list events in the future,
+because it knows that it will always jump there.
+If this line syncs prior to time passing it by,
+it will behave exactly like a normal `jump`.
+If the time passes this line,
+then it will jump as if it had synced exactly on time.
+This is not handled specially in `test_timeline`, which expects the sync to be correct.
 
 ### Commands
 
--- a/ui/raidboss/data/00-misc/test.ts
+++ b/ui/raidboss/data/00-misc/test.ts
@@ -363,6 +363,7 @@ const triggerSet: TriggerSet<Data> = {
   timelineReplace: [
     {
       locale: 'de',
+      missingTranslations: true,
       replaceSync: {
         'You bid farewell to the striking dummy': 'Du winkst der Trainingspuppe zum Abschied zu',
         'You bow courteously to the striking dummy':
@@ -454,6 +455,7 @@ const triggerSet: TriggerSet<Data> = {
     },
     {
       locale: 'cn',
+      missingTranslations: true,
       replaceSync: {
         'You bid farewell to the striking dummy': '.*向木人告别',
         'You bow courteously to the striking dummy': '.*恭敬地对木人行礼',
--- a/ui/raidboss/data/00-misc/test.txt
+++ b/ui/raidboss/data/00-misc/test.txt
@@ -28,3 +28,15 @@ hideall "--sync--"
 25 "Super Tankbuster" sync /:test sync:/ window 30,30
 30 "Dummy Stands Still"
 40 "Death"
+
+50 "--sync--" sync /:test sync2:/ window 100,1 forcejump 100
+
+# Loop test!
+102 "Two"
+103 "Three"
+104 "Four"
+106 "Six"
+110 "Ten" #duration 100
+115 "Fifteen"
+118 "Force Jump Three" sync /:test sync3:/ window 10,10 forcejump 103
+120 "Invisible" sync /:test sync4:/ forcejump 1000
--- a/ui/raidboss/timeline.ts
+++ b/ui/raidboss/timeline.ts
@@ -135,6 +135,7 @@ export class Timeline {
 
   protected activeSyncs: Sync[];
   private activeEvents: Event[];
+  private activeLastForceJumpSync?: Sync;
 
   public ignores: { [ignoreId: string]: boolean };
   public events: Event[];
@@ -232,6 +233,9 @@ export class Timeline {
   }
 
   protected SyncTo(fightNow: number, currentTime: number, _sync?: Sync): void {
+    // If we ever sync somewhere else, then remove any active overhanging windows from force jumps.
+    this.activeLastForceJumpSync = undefined;
+
     // This records the actual time which aligns with "0" in the timeline.
     const newTimebase = new Date(currentTime - fightNow * 1000).valueOf();
     // Skip syncs that are too close.  Many syncs happen on abilities that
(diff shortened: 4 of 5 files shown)


quisquous:
I'll leave this up for another week or so.  If anybody has any opinions here on either top-level direction or implementation details (@JLGarber? @valarnin?), please feel free to chime in (or ask for more time if you want it too!)

I ended up going with `forcejump` rather than `loop` for a couple of reasons.

One is just naming: sometimes the jump is forwards (such as when we have a branch into A and B and then A and B come back together again into a common timeline), and so it's not always a loop jumping backwards (where eventually you'd hit the `loop` / `forcejump` again).  It felt weird to call the former a `loop`.

The other reason is that while implementing the lookahead (and also in discussion on #5144), I realized that it got confusing to try to think about what happens if the time got past the `loop` point without syncing.  I could imagine going forward in time in some virtual timeline space where we map real time back onto the loop itself with modulo math, but then ... how do you know which syncs apply where?  I decided it was easier to just `forcejump` and say "if you ever get to this point, the timeline will just adjust its real time back to where you're going, and any syncs there will still apply".

I think the only downside to this is that if you have a `forcejump window 20,20` or something, it's really always `forcejump window 20,0` as the forward looking part of the sync window will never apply.  I can't think of any examples where we have a loop, but the loop itself is inconsistent.  We do sometimes add big windows, but that's usually as a "just in case" and less a "this has high variability".  However, if we wanted to add such "just in case" windows now, we'd need to do it at both the source and destination of the `forcejump` rather than just in one place.  (We could add timeline linting so force users to have to write `window 20,0` for `forcejump` so that it's clear in the timeline as well? Maybe that's obnoxious.)

I guess one final alternative here is that we could add a fake sync in at the jump destination as well such that after you forcejumped, if you see a belated sync from the forcejump point, it would adjust your timing.  This is ... a bit weird to think about code-wise, but I could do it.

valarnin:
I guess my two points are:

1. Is this necessary before we end up reworking the format and therefore the underlying classes? Seems like potential breakage mid-raid-tier that we'll end up throwing out anyways later isn't really great? I'd like to get transitioned to a new timeline format before 7.0...
2. `We do sometimes add big windows, but that's usually as a "just in case" and less a "this has high variability".`
   - There is variance if bosses are being dragged, especially in older content but even in new trials and stuff there's still some variance.

quisquous:
> Is this necessary before we end up reworking the format and therefore the underlying classes?

Uh, when and who is reworking this format? I haven't seen any momentum or discussion on this front, and personally haven't heard enough pain points to understand the motivation here.

quisquous:
> I think the only downside to this is that if you have a forcejump window 20,20 or something, it's really always forcejump window 20,0 as the forward looking part of the sync window will never apply. I can't think of any examples where we have a loop, but the loop itself is inconsistent. We do sometimes add big windows, but that's usually as a "just in case" and less a "this has high variability". However, if we wanted to add such "just in case" windows now, we'd need to do it at both the source and destination of the forcejump rather than just in one place. (We could add timeline linting so force users to have to write window 20,0 for forcejump so that it's clear in the timeline as well? Maybe that's obnoxious.)
> 
> I guess one final alternative here is that we could add a fake sync in at the jump destination as well such that after you forcejumped, if you see a belated sync from the forcejump point, it would adjust your timing. This is ... a bit weird to think about code-wise, but I could do it.

I went ahead and did this.  There is now an extra synthetic sync after force jumping that contains any "overhang window" time.  If there's ever any sync or jump (or the window expires naturally), the synthetic sync will be removed.

valarnin:
I think my biggest issues with the current format (aside from what you already mentioned) can be summarized as:

1. There's no way to handle logic as part of the timeline. Something like GolbezEx would benefit greatly from some sort of logic control due to the way the stored mechanics work, similarly with e12s's stored mechanics
2. There's no way for a trigger to interact with the timeline (be that a normal `triggers` trigger, or a `timelineTriggers` trigger).
   - By interact, I mean something like:
   - check some conditions, and jump to this time
   - alter the text of this timeline entry based on some condition
   - hide this timeline entry based on some condition
   - adjust the duration of this timer bar based on some condition
3. There's no good way to reference a specific timeline entry. e.g. let's say I want to change the p9s `272.2 "Gluttony's Augur"` entry to be something like `272.2 "Gluttony's Augur (heal check)"`, there's no way to target that specific timeline entry and override it without also clobbering any other entry with the same text of `Gluttony's Augur`
4. The timeline format really doesn't handle multiple different encounters in the same zone very well. e.g:
   - the variant dungeon conversation that was part of the discord conversation you linked
   - something like Delubrum Reginae where parties are handling different sets of mechanics
   - `The Copied Factory`'s `Hobbes` fight

As for discussion about a new format, I remember some other conversation in Discord where some new example formats were thrown about, something about a yml type format and something else that I can't recall off the top of my head. Discord search is just as bad as always though and I can't find them.

xiashtra:
> * jumps to numbers is fragile (we could probably add named labels)

I haven't thought too much about what would be nice to have in the timeline format, but being able to jump to a label instead of a time would definitely be on the list.  It's very easy to modify a timeline and forget to adjust the jumps currently.

It also feels kind of hacky right now the way encounters in the same zone have to be separated by arbitrarily increasing the times of the subsequent encounters.  If there was a way to say, label an encounter block and then jump to that block at the appropriate time, instead of increasing the later encounters by +1000, +2000, etc., like we currently do.  [Bozjan Southern Front](https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/eureka/bozjan_southern_front.txt) and [Zadnor](https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/05-shb/eureka/zadnor.txt) are good examples of where it would be really useful.

quisquous:
I split off an issue in #5619 to discuss further improvements rather than it being trapped in a PR that will be harder to find in the future.

In terms of this issue, my feeling is that the future falls into one of these buckets:
* if nothing is changed about timelines, this is a positive step
* if we incrementally change timelines, this is still a positive step
* if we fundamentally and completely change timelines, my argument is that any new timeline format will either:
  * have a way to convert from old to new format (this seems important to me, but I guess it's not fundamentally required), and if so I think having a "this jump is a loop with lookahead" is important meta-information about timelines that would help with conversion and at worst could just be treated as a `jump`
  * no way to convert and we just have two timeline formats side by side, in which case maybe this is slightly wasted work but it's already done and I suspect that we'll easily get through all of the remaining 7.0 things before any "completely changed timeline" work occurs given that  the discussion is at the "what do we even need" stage

quisquous:
Hey y'all, this has kinda been sitting for ~2.5w without much feedback. Not that there's true urgency, but there were other timeline things I was interested in doing and felt like it made sense to take care of this one first.

If any of y'all are busy but want to take a closer look at that, that's fine. Just let me know what your schedule looks like. Otherwise, I'll likely plan on landing this in the next week barring any major objections.

JLGarber:
Sorry, I don't really have much to offer beyond "this seems to make sense, let's land it and see what breaks". As far as using this to abstract away the whole lookahead process for fixed timelines, I can get behind that.

To be clear, when the comment notes:

> This is intended for loops that will always be taken in an encounter.
> When this is used, no "lookahead" loop unrolling is needed,
> and the timeline will use the `forcejump` destination to list events in the future,

Therefore, to handle something like [Ultima Warrior in Fractal Continuum (hard)](https://github.com/quisquous/cactbot/blob/493b83b293872516d432aee40e2ca81d60cf11e6/ui/raidboss/data/04-sb/dungeon/fractal_continuum_hard.txt#L58), we would still have to use the current unrolled lookahead, right?

quisquous:
> Therefore, to handle something like [Ultima Warrior in Fractal Continuum (hard)](https://github.com/quisquous/cactbot/blob/493b83b293872516d432aee40e2ca81d60cf11e6/ui/raidboss/data/04-sb/dungeon/fractal_continuum_hard.txt#L58), we would still have to use the current unrolled lookahead, right?

Yeah, this doesn't handle that. I think a potential followup here is to add a `maybejump` (or some sort of thing) which would do lookahead text but with question marks and no forced jump.  I thought about doing that here, but it got more complicated so I left this as-is.

wexxlee:
I'm in the same camp as others -- conceptually it sounds like a useful, well-thought-out addition.  It sounds like `forcejump` will be the default for handling straightforward loops (e.g. normal raids & trials), and we'd continue to use `jump` for the more nuanced situations, e.g., multiple branches, multiple encounters in the same zone, etc?  (Not trying to oversimplify or generalize, just trying to articulate the default use cases.)

valarnin:
Sorry about the lack of feedback on my end, I've been sick the past two weeks or so.

I think this is fine to merge as-is as well, and we can continue discussion in the issue you opened for further adjustments to timeline format when I've got the mental capacity for it again.

quisquous:
> Sorry about the lack of feedback on my end, I've been sick the past two weeks or so.

No worries! Hope you feel better!

PR #5580, opened 2023-06-18
raidboss: add triggers for p12s p1 engravements

wexxlee:
This adds support for the Engravement of Souls debuffs that occur during Paradeigma 2, Superchain I, and Paradeigma 3.

Tested the first two mechanics in game.  Was only able to partially test Engravement/Paradeigma 3 (because of role restrictions), but it looks fine in raidemulator.

This could probably benefit from more explicit typing and possibly more error-handling (suggestions welcome).

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -61,6 +61,24 @@ const superchainNpcBaseIdMap: Record<SuperchainMechanic, string> = {
   partners: '16180',
 } as const;
 
+const engravementIdMap = {
+  lightTilt: 'DF8',
+  darkTilt: 'DF9',
+  lightTower: 'DFB',
+  darkTower: 'DFC',
+  lightBeam: 'DFD',
+  darkBeam: 'DFE',
+  crossMarked: 'DFF',
+  xMarked: 'E00',
+};
+const engravementLabelMap = Object.fromEntries(
+  Object.entries(engravementIdMap).map(([k, v]) => [v, k]),
+);
+const engravementBeamIds = [engravementIdMap.lightBeam, engravementIdMap.darkBeam];
+const engravementTowerIds = [engravementIdMap.lightTower, engravementIdMap.darkTower];
+const engravementTiltIds = [engravementIdMap.lightTilt, engravementIdMap.darkTilt];
+const engravement3TheosSoulIds = [engravementIdMap.crossMarked, engravementIdMap.xMarked];
+
 const headmarkers = {
   ...wings,
   // vfx/lockon/eff/tank_laser_5sec_lockon_c0a1.avfx
@@ -114,6 +132,10 @@ export interface Data extends RaidbossData {
   expectedFirstHeadmarker?: string;
   isDoorBoss: boolean;
   phase?: 'superchain1' | 'palladion' | 'superchain2a' | 'superchain2b';
+  engravementCounter: number;
+  engravement1Towers: string[];
+  engravement3TowerType?: string;
+  engravement3TowerPlayers: string[];
   wingCollect: string[];
   wingCalls: ('swap' | 'stay')[];
   superchainCollect: NetMatches['AddedCombatant'][];
@@ -132,6 +154,9 @@ const triggerSet: TriggerSet<Data> = {
   initData: () => {
     return {
       isDoorBoss: true,
+      engravementCounter: 0,
+      engravement1Towers: [],
+      engravement3TowerPlayers: [],
       wingCollect: [],
       wingCalls: [],
       superchainCollect: [],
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -114,6 +132,10 @@ export interface Data extends RaidbossData {
   expectedFirstHeadmarker?: string;
   isDoorBoss: boolean;
   phase?: 'superchain1' | 'palladion' | 'superchain2a' | 'superchain2b';
+  engravementCounter: number;
+  engravement1Towers: string[];
+  engravement3TowerType?: string;
```
quisquous, line 137:
```suggestion
  engravement3TowerType?: 'lightTower' | 'darkTower';
```

Ideally this sort of thing can be more strongly typed when there's only a few values.

You can do this more explicit sort of thing below as well:
```diff
-data.engravement3TowerType = engravementLabelMap[matches.effectId];
+if (matches.effectId === engravementIdMap.lightTower)
+  data.engravement3TowerType = 'lightTower';
+if (matches.effectId === engravementIdMap.darkTower)
+  data.engravement3TowerType = 'darkTower';
```

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+          },
+        };
+
+        const engraveLabel = engravementLabelMap[matches.effectId];
+        if (engraveLabel === undefined)
+          return;
+        const engraveOutputStr = output[engraveLabel]!();
+        if (engraveOutputStr === undefined)
+          return;
+        if (engravementBeamIds.indexOf(matches.effectId) !== -1)
```
quisquous, line 595:
```suggestion
        if (engravementBeamIds.includes(matches.effectId))
```

Please switch `indexOf` to `includes`, here and elsewhere in this PR.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+            en: 'Stack (dark group) - spread later',
+          },
+        };
+
+        const engraveLabel = engravementLabelMap[matches.effectId];
+        if (engraveLabel === undefined)
+          return;
+        const engraveOutputStr = output[engraveLabel]!();
+        if (engraveOutputStr === undefined)
+          return;
```
quisquous, line 594:
```suggestion
```

Calls to `output` always return a `string`, no matter what.  So you can remove this check, here and elsewhere.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: { effectId: Object.values(engravementIdMap) },
+      condition: (data, matches) => data.engravementCounter === 2 && data.me === matches.target,
+      delaySeconds: 5, // wait until first orb resolves so as not to confuse
+      durationSeconds: 7, // keep active until beams go off
+      suppressSeconds: 30,
+      response: (_data, matches, output) => {
+        // cactbot-builtin-response
+        output.responseOutputStrings = {
+          lightBeam: {
+            en: 'Light Beam (stack with Dark group)',
```
quisquous, line 570: I think the "stack with X" is probably unnecessary here? I think this is a "people need to know where to go".  If it was always "dark group left" or something, then I think that'd make sense to say, but when it's just "stack with dark group, wherever that is", I think it's not helpful; folks already have to know the mechanic.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+            en: 'Dark Beam (stack with Light group)',
+          },
+          lightTower: {
+            en: 'Light Tower (stack with light group)',
+          },
+          darkTower: {
+            en: 'Dark Tower (stack with dark group)',
+          },
+          lightTilt: {
+            en: 'Stack (light group) - spread later',
```
quisquous, line 582:
```suggestion
            en: 'Light Tilt => spread later',
```

I think this is less wordy.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+    // The other role group will receive lightTower, darkTower, lightBeam, and darkBeam.
+    // To resolve the Beams during the 2nd orb, lightBeam needs to stack with darkTower and both darkTilts, and vice versa.
+    // After the 3rd orb, lightTower and darkTower will drop their towers, and  darkBeam and lightBeam (respectively) will soak them.
+    // The four Heavensflame players all simultaneously need to spread to drop their AoEs.
+    // Debuffs do change based on mechanic resolution, which can complicate things:
+    // - When a lightTilt player soaks a dark beam, their debuff will change to darkTilt, and vice versa.
+    // - Once the beams detonate, the lightBeam debuff disappears and is replaced with lightTilt (same with dark).
+    // So only use the initial debuff to resolve the mechanic, and use a long suppress to avoid incorrect later alerts.
+    {
+      id: 'P12S Engravement 2 Beam Soak',
```
quisquous, line 559: How does this trigger work timing-wise with the Left CW / Right CCW one before? Should this be combined with that? Like `Left CW => Light Beam` sorta thing?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+        const engraveOutputStr = output[engraveLabel]!();
+        if (engraveOutputStr === undefined)
+          return;
+        if (engravementBeamIds.indexOf(matches.effectId) !== -1)
+          return { alarmText: engraveOutputStr };
+        return { alertText: engraveOutputStr };
+      },
+    },
+    // darkTower/lightTower are 20s, but lightBeam/darkBeam are shorter and swap to lightTilt/darkTilt before the mechanic resolves.
+    // So use a fixed delay rather than one based on effect duration.
```
quisquous, line 601: Do you think we should try to handle weird edge cases, like ... if both tilts get hit by towers, I think they could theoretically swap towers.  Or, if somebody dies, a spread could technically take a tower.  Or, if you do get hit by a tower drop as a tilt, then you shouldn't be told to take a tower (since it's not going to work).

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+    {
+      id: 'P12S Engravement 2 Tower Drop/Soak Reminder',
```
quisquous, line 603: Should this be combined with the In => Out call before? e.g. `In => Out => Spread` or `In => Out => Dark Tower` then `Out => Drop Dark Tower` reminder?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+      type: 'GainsEffect',
+      netRegex: { effectId: engravement3TheosSoulIds },
+      condition: (data, matches) => data.engravementCounter === 3 && data.me === matches.target,
+      alertText: (_data, matches, output) => {
+        const mark = matches.effectId === engravementIdMap.crossMarked ? '+' : 'x';
+        return output.theosDebuff!({ mark: mark });
+      },
+      outputStrings: {
+        theosDebuff: {
+          en: '${mark} on You',
```
quisquous, line 658: Can you have separate output strings for this, in case people want to override it with like "Plus go north" or whatever?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: { effectId: engravement3TheosSoulIds },
+      condition: (data, matches) => data.engravementCounter === 3 && data.me === matches.target,
+      delaySeconds: (_data, matches) => parseFloat(matches.duration) - 4,
+      alertText: (_data, matches, output) => {
+        const mark = matches.effectId === engravementIdMap.crossMarked ? '+' : 'x';
+        return output.dropBait!({ mark: mark });
+      },
+      outputStrings: {
+        dropBait: {
+          en: 'Drop baited ${mark} AoE',
```
quisquous, line 674: Same here, can you have separate output strings?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+        const towerColor = data.engravement3TowerType === 'lightTower'
+          ? output.light!()
+          : output.dark!();
+        if (myEffect === soakers)
+          return output.soakTower!({ color: towerColor });
+        return output.baitCleaves!();
+      },
+      outputStrings: {
+        soakTower: {
+          en: 'Soak ${color} Tower',
```
quisquous, line 802: Unless somebody comes up with some wild uptime strategy that puts people in different spots, would it be possible to tell the dps soaking here to soak the middle tower vs soaking the "non-middle" (idk what to call this) tower as well?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+      alertText: (data, matches, output) => {
+        // lightTower/darkTower support players receive lightTilt/darkTilt once dropping their tower
+        // so exclude them from receiving this alert
+        if (data.engravement3TowerPlayers.indexOf(data.me) !== -1)
+          return;
+
+        const soakMap: { [key: string]: string } = {
+          lightTower: 'darkTilt',
+          darkTower: 'lightTilt',
+        };
```
quisquous, line 785:
```suggestion
        const soakMap = {
          lightTower: 'darkTilt',
          darkTower: 'lightTilt',
        } as const;
```

If `engravement3TowerType` is more strongly typed, then this can just be `as const`.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+        const myEffect = engravementLabelMap[matches.effectId];
+        if (myEffect === undefined || data.engravement3TowerType === undefined)
+          return;
+        const soakers = soakMap[data.engravement3TowerType];
+        if (soakers === undefined)
+          return;
```
quisquous, line 791:
```suggestion
```

With the above typing changes, this can never be `undefined`.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+
+        const soakMap: { [key: string]: string } = {
+          lightTower: 'darkTilt',
+          darkTower: 'lightTilt',
+        };
+        const myEffect = engravementLabelMap[matches.effectId];
+        if (myEffect === undefined || data.engravement3TowerType === undefined)
+          return;
+        const soakers = soakMap[data.engravement3TowerType];
```
quisquous, line 789: Sorry for this tiny nit, but `soakers` to me sounds like an array of people soaking since it's plural.  Could this be called like `soakTiltType`?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -465,6 +490,328 @@ const triggerSet: TriggerSet<Data> = {
...
+            ? output.light!()
+            : output.dark!();
+        let partner;
+        if (towerOnYou) {
+          partner = data.ShortName(data.engravement3TowerPlayers.find((name) =>
+            name !== data.me
+          )) ?? output.unknown!();
+          return { alertText: output.towerOnYou!({ color: towerColor, partner: partner }) };
+        } else if (data.party.isDPS(data.me)) {
+          // DPS want to know tower color to know if they are soaking or baiting add cleaves later
```
quisquous, line 738: I'm always a little hesitant on the `data.party` checks, both because of https://github.com/OverlayPlugin/OverlayPlugin/issues/60 but also because of (far far) future `BLU` stuff.  Could you track who has the tethers, what color they are, and tell them more specifically what they're doing, e.g. bait cleaves later or bait middle dark tower later or bait non-middle(?) light tower later?

quisquous (commented):
Sorry for all the comments! Looks good in general, and thanks for all this work. I think this'll make these mechanics way better.

Mostly just some type-related cleanup and some wording/combining triggers sorts of questions.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -61,6 +61,24 @@ const superchainNpcBaseIdMap: Record<SuperchainMechanic, string> = {
...
+  lightTower: 'DFB',
+  darkTower: 'DFC',
+  lightBeam: 'DFD',
+  darkBeam: 'DFE',
+  crossMarked: 'DFF',
+  xMarked: 'E00',
+};
+const engravementLabelMap = Object.fromEntries(
+  Object.entries(engravementIdMap).map(([k, v]) => [v, k]),
+);
```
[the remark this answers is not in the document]
wexxlee: > Re: typing, you can do this dance. Inverting objects is always a little fraught in TypeScript because of the way Object.keys and Object.entries strips some of them types.

Oh, that is a neat trick.  What do you think about simplifying this by reversing the declarations and using a type annotation on fromEntries - I think that would let us type both objects?  E.g.:

```
type EngravementLabel = 'lightTilt' | 'darkTilt' | 'lightTower' | 'darkTower' | 'darkTower' | 'lightBeam' | 'darkBeam' | 'crossMarked' | 'xMarked';
const engravementLabelMap: { [effectId: string]: EngravementLabel } = {
  DF8: 'lightTilt',
  DF9: 'darkTilt',
  DFB: 'lightTower',
  DFC: 'darkTower',
  DFD: 'lightBeam',
  DFE: 'darkBeam',
  DFF: 'crossMarked',
  E00: 'xMarked',
} as const;
type EngravementIdMap = Record<EngravementLabel, string>;
const engravementIdMap: EngravementIdMap = Object.fromEntries(Object.entries(engravementLabelMap).map(([k, v]) => [v, k])) as EngravementIdMap;
```

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > You can do this more explicit sort of thing below as well:

Since the netRegex is restricted to just the lightTower and darkTower effects (e.g. `engravementTowerIds`), are you generally okay with something more shorthand like:
`data.engravement3TowerType = matches.id === engravementIdMap.lightTower ? 'lightTower' : 'darkTower';`

Or do you prefer separate explicit conditionals?

quisquous: Yeah, that's fine too!

ui/raidboss/data/06-ew/raid/p12s.ts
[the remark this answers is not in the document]
quisquous: I think that's fine too.  I looked at this a bit and couldn't come up with any way to avoid the `as` type casting in either case.  I thought maybe `as` on the keys looked safer (because it was simpler and felt obvious), but I think this is also fine.

You should be able to still say: `type EngravementLabel = typeof engravementLabelMap[keyof typeof engravementLabelMap];` here without duplicating the labels in the first line as well.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > I think the "stack with X" is probably unnecessary here? I think this is a "people need to know where to go". If it was always "dark group left" or something, then I think that'd make sense to say, but when it's just "stack with dark group, wherever that is", I think it's not helpful; folks already have to know the mechanic.

I think I agree with that on the lightTower and darkTower alerts (i.e., I think we can remove '(stack with light/dark group)' from the alert).  For the others though:
- lightTilt and darkTilt just say 'Stack (light group)' or 'Stack (dark group)' - since we don't otherwise call out the color of their debuff, I think we should keep those parentheticals.
- For lightBeam and darkBeam, I do think it's worthwhile keeping '(stack with Dark/Light group)' - there's a lot going on during the mechanic, and I've found that people sometimes see the light debuff, forget that it's beam, and go stack with the light group (=wipe).  I think similar things could happen if we just say "Light Beam" -- people might see the light and forget the important part, e.g. that it's Beam and that they need to stack with the other group.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: Responded above before I saw this.  That works, although I think I would say "Light Stack => spread later".  Tilt is the technical name of the debuff (well, actually Astral Tilt and Umbral Tilt), but I'm honestly not sure how many people know that it's called Tilt, outside from us nerds.

quisquous: Yeah, that's fair; mostly I think there's no "stack" person there, which was a little of my sticking point.  `Light Tilt` or `Light Debuff` (which is maybe vague) sounded better, since the most important part is the "Light" word.

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: Ok, if you think it's worth keeping the parenthetical on top of the different severity, sounds good to me.  Maybe drop the `group` part and just say `(stack with dark)` or `(stack with light)` for more brevity?

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > How does this trigger work timing-wise with the Left CW / Right CCW one before? Should this be combined with that? Like `Left CW => Light Beam` sorta thing?

The timing here is wonky because of the delays on both triggers to line them up with what's happening during the mech.  As is, the rotation call comes a little before the first orb goes off, and the debuff call comes out as the first orb is resolving, so they are about 3.5->4 seconds apart.  The alerts just briefly overlap on screen.

I think we could combine them, but I think we'd want a longer duration so people see their debuff direction while moving toward the second orb (because goldfish brain).  I'll jump in this afternoon and do some testing.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: There's more wonky timing here.  If it's out->in, the tower drop is almost immediately after the donut telegraph disappears, but if it's in->out, it's like 3-4 seconds before the tower drop after the out resolves.

From testing in-game, basing the tower/soak/spread alert on the effect duration felt like the right amount of warning and wasn't confusing (at least to me) while the in/out was happening.  But if you do want to add a reminder for the second half of the mechanic, I think I'd want to test a little more in-game to see if I can find a good timing that works for both in->out and out->in.

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: Yeah, I got some feedback on the existing left/right that it could come earlier since most folks are standing around in place for protean/partners and it'd be helpful to know which direction they were going first.  Longer duration would be good here for any combined trigger exactly as you say (or alternatively a separate trigger so it's `Left => Light Beam` then `Light Beam` reminder).

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: There's **soooo** many weird edge cases because of how the debuffs interact.  But just tackling those you listed:
  - If both Tilts (I think you mean the Beams that get converted to Tilts, right?) get hit by towers, I think they just die from the tower drop.  Have to go back and look at logs, but that's my recollection.  But realistically, it's such a short time in which the tower must be soaked that I don't think an alert alone would let them react quickly enough to swap unless the towers were extremely close together (which in at least the strat I've most commonly seen, they aren't).
   - If the Beams die, a Tilt (spread) could take a tower, but their Soul AoE will go off at the same time, which will kill anyone nearby (inb4 overgearing).  Would we want to alert all possible Tilt players?
    - If you're a Beam and you get hit by a tower, you shouldn't soak it, true -- but the Soak alert goes off at the same time as the Drop Tower alert (4 sec. before) - so it's not really so much that you'll be getting an inaccurate alert as that the alert you got 4 seconds ago is no longer valid because you screwed up the mechanic in the meantime.

I thought about some of these (although not as many as you!) when writing this, but where I ultimately landed was that it would be a lot of extra work to track player deaths and (unexpected) debuff changes during the mechanic and then try to write trigger contingencies for each that also don't conflict with each other.  (Couple that with the fact that I'll never get my static to willingly edge-test these, and I don't know how long PF will tolerate me griefing to test them myself.)  Also, the tower soak time is so small, I'm not sure that alert calls will help recover if people aren't already paying extremely close attention and reacting already, which makes the trigger less useful.

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: I guess I'm suggesting to add the reminder into the in/out call (so it says `In => Out => Spread` or whatever); this call happens way earlier, so there shouldn't be any timing issues?

I think it's a separate issue for different timings here, re: in/out.  There's no "out" call after the "in" goes off right now.  I think if that existed and we wanted to combine them then that'd be one with timing issues.

Am I misunderstanding?

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: Yeah, that's fair.  I feel like we've saved like one (1) pull ever with some quick reacting in this case.  Mostly just curious what if anything could be done here.  Sounds like this is all a "keep it simple for now, and reconsider later if it ever comes up".

I do think tilts can survive a tower drop.  I think they take the same damage the tower person does but their debuff is also swapped to be the same as the tower so they have to take the other tower (if any).

ui/raidboss/data/06-ew/raid/p12s.ts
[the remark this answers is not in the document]
wexxlee: > You should be able to still say: `type EngravementLabel = typeof engravementLabelMap[keyof typeof engravementLabelMap];` here without duplicating the labels in the first line as well.

I tried that initially, and I think it leads to some sort of weird recursion error, which is why I duplicated the labels.  This is probably my still-rough knowledge of typing, but regardless of whether the `type EngravementLabel` definition comes before or after `const engravementLabelMap`, the const declaration throws an error because EngravementLabel is not validly defined at that point (i.e. 'Type 'string' is not assignable to type 'never'').

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Yeah, that's fair; mostly I think there's no "stack" person there, which was a little of my sticking point. `Light Tilt` or `Light Debuff` (which is maybe vague) sounded better, since the most important part is the "Light" word. (Edited to add: `Light Stack` also fine if you don't think it's confusing too and maybe it's just me. Just trying to disambiguate as much as possible and hard to know what other people think!)

Hmm, that's fair given how we use 'Stack' elsewhere.  How about 'Light Group => spread later'?

quisquous: Sounds good to me!

ui/raidboss/data/06-ew/raid/p12s.ts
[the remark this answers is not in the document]
quisquous:
```typescript
const engravementLabelMap = {
  DF8: 'lightTilt',
  DF9: 'darkTilt',
  DFB: 'lightTower',
  DFC: 'darkTower',
  DFD: 'lightBeam',
  DFE: 'darkBeam',
  DFF: 'crossMarked',
  E00: 'xMarked',
} as const;
type EngravementLabel = typeof engravementLabelMap[keyof typeof engravementLabelMap];
type EngravementIdMap = Record<EngravementLabel, string>;
const engravementIdMap: EngravementIdMap = Object.fromEntries(Object.entries(engravementLabelMap).map(([k, v]) => [v, k])) as EngravementIdMap;
```

With the `as const`, you don't have to explicitly type `engravementLabelMap`

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Unless somebody comes up with some wild uptime strategy that puts people in different spots, would it be possible to tell the dps soaking here to soak the middle tower vs soaking the "non-middle" (idk what to call this) tower as well?

I don't think we can tell the DPS which tower to soak, because both DPS are on the same platform, and one towerr will be dropped on the platform and one will be dropped by the support player on the other platform (but on the intersecting corner).  But either DPS can soak either tower, and I've seen both strats (having the diagonal tether soak the center one, or the straight tether soaking the center one), not to mention the 'yolo' strat (which seems to happen 90% of the time anyway).

I also thought about whether we could remind the *support* player in 'P12S Engravement 3 Towers Drop' to drop their tower center/overlapping if both DPS on their side have the same color tether.  And while we can tell whether the East or West support player will need to drop mid (based on the positioning of the adds and their tether colors), we would have to base that call on whether the support player is standing east or west at the time (and since we know the center point, player position (if accurate) should be a 100% reliable way to determine that).

Does `getCombatants` return player data as well, or just NPC data?

quisquous: Ohhhhhhh that's right.  Sorry, I was thinking about this wrong.  They're either soaking towers or baiting, so you can't tell them which.  Yeah, I think the support person should get the call about where to place the tower, and the dps can just sort it out.

We should be able to know which side they're on based on player position from the taking the tower damage if you want a non-`getCombatants` method.

ui/raidboss/data/06-ew/raid/p12s.ts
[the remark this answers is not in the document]
wexxlee: > With the `as const`, you don't have to explicitly type `engravementLabelMap`

I think that ends up breaking a bunch of the lookups against `engravementLabelMap[]` though.  Like `const engraveLabel = engravementLabelMap[matches.effectId];` is now an unsafe assignment of an 'any' value.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: I honestly can't remember, so you might be right.  But yeah, I'll add a 'todo' note so it's there for posterity.

ui/raidboss/data/06-ew/raid/p12s.ts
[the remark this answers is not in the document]
quisquous: The original thing I posted didn't have this because there's no lookups the other way around.  You can fix that by re-typing it after:

```typescript
const engravementLabelMapAsConst = {
  DF8: 'lightTilt',
  DF9: 'darkTilt',
  DFB: 'lightTower',
  DFC: 'darkTower',
  DFD: 'lightBeam',
  DFE: 'darkBeam',
  DFF: 'crossMarked',
  E00: 'xMarked',
} as const;
type EngravementLabel = typeof engravementLabelMapAsConst[keyof typeof engravementLabelMapAsConst];
const engravementLabelMap: { [effectId: string]: EngravementLabel } = engravementLabelMapAsConst;
type EngravementIdMap = Record<EngravementLabel, string>;
const engravementIdMap: EngravementIdMap = Object.fromEntries(Object.entries(engravementLabelMap).map(([k, v]) => [v, k])) as EngravementIdMap;
```

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > I guess I'm suggesting to add the reminder into the in/out call (so it says `In => Out => Spread` or whatever); this call happens way earlier, so there shouldn't be any timing issues?
> 
> I think it's a separate issue for different timings here, re: in/out. There's no "out" call after the "in" goes off right now. I think if that existed and we wanted to combine them then that'd be one with timing issues.
> 
> Am I misunderstanding?

I think it was the second half of your suggestion about doing the `Out => Drop Dark Tower` that tripped me up. 

No issue with adding the => Spread/Soak/Drop to the end of the in/out call as you suggest.  But that trigger fires 7 seconds before the (current) Soak/Drop/Spread triggers fire, and 11 seconds before those soak/drop/spread mechanics resolve.  So I think we'd still need a subsequent alert closer to the actual debuff going off.  Right now, that subsequent one (in this PR) just calls Soak/Drop/Spread, so are you suggesting that it should include the second-half of the in->out direction as well? (In which case I think we'd need to move it earlier and extend the duration)?

ui/raidboss/data/06-ew/raid/p12s.ts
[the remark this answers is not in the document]
wexxlee: Thanks!

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > We should be able to know which side they're on based on player position from the taking the tower damage if you want a non-`getCombatants` method.

Brill.  Working on it.

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: Ah yeah, I'd say leave that second half as a TODO for later, especially if there's timing issues and also because there's not an `Out` call right now to extend into `Out => Drop Dark Tower` so we'd have to add that as well.

So, mostly just suggesting the first half, to add it to the in/out call.

wexxlee:
Closing due to borked rebase.  New PR opened - #5587

PR #5586, opened 2023-06-19
docs: add headmarker guide

quisquous:


quisquous:
Here's the rendered markdown: https://github.com/quisquous/cactbot/blob/4325d3e4513d77aeca88121b4ce458f6fe271bdc/docs/Headmarkers.md

cc @JLGarber who might be interested in this
cc @xiashtra who asked about this recently

JLGarber:
Looks good. Thinking more about that first paragraph, do we maybe anticipate getting the correction functionality inlined in OverlayPlugin sometime In The Next Year, or should I drag out my old work on adding `getHeadmarkerID()` to the `data` object?

quisquous:
> Looks good. Thinking more about that first paragraph, do we maybe anticipate getting the correction functionality inlined in OverlayPlugin sometime In The Next Year, or should I drag out my old work on adding getHeadmarkerID() to the data object?

Uh, I think it's a "it'll get done if somebody wants to make it happen".  I don't know that it's high on my list of stuff to do.

I feel like the data object is a little bit overloaded, and the function doesn't make sense unless `Data` also has `decOffset`.  If you wanted to go this route, I wonder if it make more sense to have like a `HeadmarkerData extends Data` type with `decOffset` and `firstHeadmarker` member in it, and also some `Util.getHeadmarkerId(data: HeadmarkerData, matches)` kinda helper function? Just brainstorming; happy to chat about this more elsewhere.

Issue #5584, opened 2023-06-19
raidboss: P11S lightstream safespot sometimes incorrect


quisquous:
See comments on #5557

my log file (all my log files appear sorted)
```
[21:40:13.596] TargetIcon 1B:4002ADD7:Arcane Cylinder:0000:0000:00B5:0000:0000:0000
[21:40:13.596] TargetIcon 1B:4002ADD8:Arcane Cylinder:0000:0000:00B6:0000:0000:0000
[21:40:13.596] TargetIcon 1B:4002ADD9:Arcane Cylinder:0000:0000:00B5:0000:0000:0000
[21:40:13.640] StartsCasting 14:4002ADD7:Arcane Cylinder:8207:Lightstream:4002ADD7:Arcane Cylinder:7.700:81.38:89.25:0.00:1.05
[21:40:13.640] StartsCasting 14:4002ADD8:Arcane Cylinder:8207:Lightstream:4002ADD8:Arcane Cylinder:7.700:100.00:121.50:0.00:-3.14
[21:40:13.640] StartsCasting 14:4002ADD9:Arcane Cylinder:8207:Lightstream:4002ADD9:Arcane Cylinder:7.700:118.62:89.25:0.00:-1.05
```

log file from #5557
```
[12:16:45.006] TargetIcon 1B:40004F48:Arcane Cylinder:0000:0000:0167:0000:0000:0000
[12:16:45.006] TargetIcon 1B:40004F4A:Arcane Cylinder:0000:0000:0167:0000:0000:0000
[12:16:45.006] TargetIcon 1B:40004F49:Arcane Cylinder:0000:0000:0168:0000:0000:0000
[12:16:45.050] StartsCasting 14:40004F48:Arcane Cylinder:8207:Lightstream:40004F48:Arcane Cylinder:7.700:81.38:89.25:0.00:1.05
[12:16:45.050] StartsCasting 14:40004F4A:Arcane Cylinder:8207:Lightstream:40004F4A:Arcane Cylinder:7.700:118.62:89.25:0.00:-1.05
[12:16:45.050] StartsCasting 14:40004F49:Arcane Cylinder:8207:Lightstream:40004F49:Arcane Cylinder:7.700:100.00:121.50:0.00:-3.14
```

I think my conclusion is that the adds need be sorted by id in order to be in NW, S, NE order

cc @xiashtra

xiashtra:
[...]

PR #5589, opened 2023-06-19
raidboss: re-fix P11S lightstream safespot

### LINKED ISSUE #5588: P11S lightstream no longer working after today's update
## Description

Lightstream callouts are no longer working since I updated cactbot this morning, they were working just fine yesterday.

## Additional information

If available, please include any logs or screenshots depicting the issue you're experiencing.

quisquous:
Fixes a bug from #5585, which itself was fixing #5584.

I don't know why I thought this worked in the emulator. Sorry for the hassle.

Fixes #5588.


Souma-Sumire:
I guess this is because the logs you used during development and testing were from the same session, so the headmarkers would be the same regardless of whether they went through the getHeadmarkerId function?

PR #5591, opened 2023-06-20
raidboss: fix p10s jade passage wrong output

lsl1225:
Just played p10s in games, and it seems the result is opposite, still need someone to take further testing.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p10s.ts
+++ b/ui/raidboss/data/06-ew/raid/p10s.ts
@@ -581,7 +581,7 @@ const triggerSet: TriggerSet<Data> = {
       suppressSeconds: 5,
       infoText: (_data, matches, output) => {
         const y = parseInt(matches.y);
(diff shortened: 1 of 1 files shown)


wexxlee:
Looking back at the original PR (#5574 ) that added this trigger, I think the assumption was that the largest potential y-value for the combatants (114) was lines, but it isn't.  

The bits don't align perfectly with boxes or lines - there some vertical drift from north to south.  Here's a screenshot showing the bits at [88, 92, 96, 100, 104, 108, 112], and you can see the drift:

image

Assuming the positioning of the bits in this screenshot is best described as being on 'lines' (which, I guess, is probably as good as we can get) -- then the trigger should return output.boxes!() to tell users to stand inside the boxes.  Right now, it returns output.lines!(), so I think this PR is a good fix.

Legends0:
As wexxlee mentioned, it's an approximation. I realize in game if you are at the south end, (last row), it ends up being off whether or not it is lines/boxes, so unless you are middle of the arena, it's iffy how much the change is worth. Unsure if there is anything better it could say.

PR #5587, opened 2023-06-19
raidboss: add support for P12S Engravements

wexxlee:
Not sure if I did something dumb while rebasing or Visual Studio did, but it was a messy unwind, and doing a new PR was POLR.  So I'm killing the old PR (#5580 ) and opening a new one.  This includes all revisions incorporating feedback from the initial code review.  Apologies for any added work.

PR #5597, opened 2023-06-22
raidboss: add P12s Engravement 1 Tower trigger

YuzukiTsuru:
Add P12s Engravement 1 Tower triggers. 
> This trigger is based on part of the functionality of #5587 PR, and later merges may involve conflict merges, which I will resolve at that time.

In Engravement 1 (Paradeigma 2), 2 players receive lightTower and 2 players receive darkTower, 2 players need to guide the light beam and 2 players need to guide the dark beam.

The operator of the beam extends the beam directly from the outside. The beam is attenuated until the jagged line disappears.

The people in the tower find the people who have the opposite attribute to the debuff and put them in four places. At NE NW SE SW as a # shape. The position of outside Anthropos is fixed by two situation. `{[97, 75], [125, 97], [103, 125], [75, 103]} and {[103, 75], [125, 103], [97, 125], [75, 97]}`. The Anthropos will cast 'Searing Radiance' for light beam and 'Shadowsear' for dark beam. We use those as a trigger for Tower players place the Tower.

When debuffs expire and towers drop, their debuff changes to lightTilt or darkTilt (same as tower color).
At the same time the towers drop, the 4 tethered players receive lightTilt or darkTilt depending on their tether color.

example


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+          cn: '引导暗激光',
+        },
+      },
+    },
+    {
+      id: 'P12S Engravement 1 Tower Drop Notifier',
+      type: 'GainsEffect',
+      netRegex: { effectId: ['DFB', 'DFC'] },
+      condition: (data) => data.engravementCounter === 1,
+      delaySeconds: 1, // Make sure Drop Tower Pos Tracker is done
```
quisquous, line 634: How close are these in time? Would it make sense to have a collector for `DFB`/`DFC` and do the callout in `P12S Engravement 1 Drop Tower Pos Tracker` as soon as the locations are known?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+          cn: '去 ${pos1}/${pos2} 放暗塔',
+        },
+        lightTower: {
+          en: 'Drop light tower',
+          cn: '放光塔',
+        },
+        darkTower: {
+          en: 'Drop dark tower',
+          cn: '放暗塔',
+        },
```
quisquous, line 691: Are these used?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+
+          return output.darkTowerSide!({
+            pos1: data.engravement1DarkBeamsPos[0],
+            pos2: data.engravement1DarkBeamsPos[1],
+          });
+        }
+      },
+      outputStrings: {
+        lightTowerSide: {
+          en: 'Drop light tower at ${pos1}/${pos2}',
```
quisquous, line 677:
```suggestion
          en: 'Drop light tower ${pos1}/${pos2}',
```

Just making this shorter.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+          cn: '放光塔',
+        },
+        darkTower: {
+          en: 'Drop dark tower',
+          cn: '放暗塔',
+        },
+        northeast: Outputs.northeast,
+        northwest: Outputs.northwest,
+        southeast: Outputs.southeast,
+        southwest: Outputs.southwest,
```
quisquous, line 695: Since there's two directions, can these be `Outputs.dirNE` instead of `Outputs.northeast` (i.e. `NE` instead of `northeast`) just to keep it shorter?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+          return;
+        const y = data.combatantData[0]?.PosY;
+        if (y === undefined)
+          return;
+
+        if (x === 75 && y === 97) {
+          if (matches.id === '82F1')
+            data.engravement1BeamsPosMap.set('NE', 'dark');
+          else
+            data.engravement1BeamsPosMap.set('NE', 'light');
```
quisquous, line 567: Logic simplification request: I think you could set `const color = matches.id === '82F1' ? 'dark' : 'light'` and then simplify a lot of this logic by setting things in the map to `color` instead of needing if statements.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+      },
+      run: (data, matches) => {
+        const x = data.combatantData[0]?.PosX;
+        if (x === undefined)
+          return;
+        const y = data.combatantData[0]?.PosY;
+        if (y === undefined)
+          return;
+
+        if (x === 75 && y === 97) {
```
quisquous, line 563: It sounds like these positions are always in the same place, but I am very hesitant to do equality comparisons on floating point numbers.  Could I suggest:
* === 75 becomes < 80
* === 125 becomes > 120
* === 97 becomes < 100
* === 103 becomes > 100

That way if for whatever reason rounding changes somewhere, then this will continue to work.  Just trying to be resilient into the future.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -508,6 +520,181 @@ const triggerSet: TriggerSet<Data> = {
...
+          if (matches.id === '82F1')
+            data.engravement1BeamsPosMap.set('NE', 'dark');
+          else
+            data.engravement1BeamsPosMap.set('NE', 'light');
+        } else if (x === 97 && y === 75) {
+          if (matches.id === '82F1')
+            data.engravement1BeamsPosMap.set('SW', 'dark');
+          else
+            data.engravement1BeamsPosMap.set('SW', 'light');
+        } else if (x === 103 && y === 75) {
```
quisquous, line 573: This is probably my biggest request here, but I think there are different ways to solve this mechanic.  The PF strat (which you have here) is to for the tether people to take the tower that's in their quadrant, i.e. the person tethered to this NNE (103, 75) add goes CCW to their tower SE, but if the add is NNW (97, 75) they go CW to their tower SW.  I've seen a different strategy where tethers always rotate clockwise, e.g. people tethered to either NNE or NNW always take the SW tower.

This will likely make this a little more complicated, but can you add some triggerset config ui for this, and add three options?
* no strategy (default, just calls out "drop light tower" with no position)
* quadrant strategy (this PR, maybe PF has a better name for this)
* clockwise strategy (tethers always rotate clockwise to take towers)

...and then `P12S Engravement 1 Drop Tower Pos Tracker` will have to check the option and then set slightly different values in pos map.

If you haven't seen the triggerset config ui, it looks like:
* https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts#L40
* https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts#L173-L194
* https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/06-ew/ultimate/the_omega_protocol.ts#L1909

(let me know if I can explain anything else)

quisquous (commented):
Thanks for all of this work! This is exciting!

ui/raidboss/data/06-ew/raid/p12s.ts
YuzukiTsuru: In combat, the buff starts at 1:06s, and then soak tower at 1:16s, with 10s of time to run to position, which is delayed for 1s to allow the tracker above to gather tower information. Where the delay takes into account the network delay so set the display after 1s, I think 9s for run is enough.

image
image

ui/raidboss/data/06-ew/raid/p12s.ts
YuzukiTsuru: No use, just output in Engravement 1 Tower Drop P12S from #5587

ui/raidboss/data/06-ew/raid/p12s.ts
YuzukiTsuru: Great, I'm sorry I didn't know this output during development.

ui/raidboss/data/06-ew/raid/p12s.ts
YuzukiTsuru: The strategy of this implement is based on game8.jp, Which is wildly used in Japan players. I think triggerset config ui is great for this condition. I'm gonna go check on the current solution for this.

YuzukiTsuru: Both tilt and Vertical methods are implemented with reference to the TOP trigger, while providing the option to display only the type of tower, not the position. 

https://github.com/quisquous/cactbot/pull/5597/commits/cb16f2617173a329a7988307b02cbb346a5f09f6

ui/raidboss/data/06-ew/raid/p12s.ts
YuzukiTsuru: Added a new setting, retain this output.

ui/raidboss/data/06-ew/raid/p12s.ts
YuzukiTsuru: I actually tried it, and it seemed like a better change

https://github.com/quisquous/cactbot/blob/edc4fe0647ad6fb893533fe4b8468956aadd49af/ui/raidboss/data/06-ew/raid/p12s.ts#L747-L748

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -194,11 +201,38 @@ export interface Data extends RaidbossData {
...
+      name: {
+        en: 'Paradeigma: Engravement of Souls Tower strategy',
+        cn: '第一次拉线踩塔方法',
+      },
+      type: 'select',
+      options: {
+        en: {
+          'Vertical Beam (Game8)': 'sharp',
+          'Tilt Beam': 'tilt',
+          'Only tower color': 'tower',
```
quisquous, line 216: Hmm, sorry for wording quibbling here, but I don't think these are descriptive enough, and I think "tilt" isn't precise either (you can have the tethers be vertical but then still go clockwise always vs quadrant).

I'm also a little hesitant to prescribe a particular strategy here too, and wonder if `tower` should be the default.  However, I do think that the "nearest quadrant" tower appears to be the aether PF strat as well, so I think either `tower` or `quadrant` could be ok as the default.

Hmm, also this points out maybe a ui issue where if we start having more strategy options I almost need some place to put descriptions of what these are or links to raidplans? For now, I think it's better to be extra wordy.  (I don't know what Game8 is.  If you have a link to a raidplan or something, can you leave it in a comment?)

```suggestion
          'Vertical beam + nearest quadrant tower (Game8)': 'quadrant',
          'Clockwise tower from beam': 'clockwise',
          'No strategy: just call tower color': 'tower',
```

If we end up with more of these strategy decisions, maybe in the future we need a notification or the cactbot config ui could prompt people that they have options they need to set as well.  This is somewhat of a new feature.

quisquous: There's also Inumaru "add-relative" strat too, but that could be a follow-up if anybody else wanted to add it: https://www.reddit.com/r/ffxiv/comments/1469jxz/p12s_para_2_first_tether_set_inumaru_strat/

YuzukiTsuru: Will, in Japan and China we often use game8's raidplan, links follow 
https://game8.jp/ff14/534748

And for the 'clockwise' raidplan, I saw it in a Youtube video:
https://www.youtube.com/watch?v=Q4Qy3CRMdQc&ab_channel=Rinon

Considering that there are many kinds of raidplans, it is currently possible to set the default to 'tower'.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -604,30 +638,179 @@ const triggerSet: TriggerSet<Data> = {
...
+        },
+      },
+    },
     {
       id: 'P12S Engravement 1 Tower Drop',
       type: 'GainsEffect',
       netRegex: { effectId: engravementTowerIds },
       condition: (data) => data.engravementCounter === 1,
-      durationSeconds: (_data, matches) => parseFloat(matches.duration),
+      delaySeconds: 0.1, // Make sure Drop Tower Pos Tracker is done
```
quisquous, line 744: If I look at the log times, I see this:
```
// Anthropos move into position
[19:12:18.911] 261 105:Change:4001EA0D:Heading:0.0000:PosX:97.0000:PosY:75.0000:PosZ:0.0000
// tethers appear
[19:12:22.471] Tether 23:4001EA0D:Anthropos:etc
// P12S Engravement 1 Tower Drop + 0.1s delay
[19:12:22.471] StatusAdd 1A:DFC:etc
// P12S Engravement 1 Drop Tower Pos Tracker + async getCombatants
[19:12:22.516] StartsCasting 14:4001EA0D:Anthropos:82F1:etc
```

Sorry, it's the delay here that makes me a little nervous.  It's possible that there might be some local (or network, if this is a remote overlay) delay in getting combatant data, and so you're hoping that 0.1s is longer than 0.04s + the time it takes to do 4x calls to `getCombatants`.  This seems like risky timing to me.

My original suggestion was this:
* add a new Anthropos collector trigger that records all the ids of any Anthropos tethers
* assume that 1A gains effect lines come after 23 tether lines always (this seems consistent to me)
* P12S Engravement 1 Tower Drop trigger then waits for the GainsEffect on the player, and does a getCombatants with the Anthropos ids, and does the work of "P12S Engravement 1 Drop Tower Pos Tracker" and "P12S Engravement 1 Tower Drop" in a single trigger

...that way there's no need for a heuristic delay (that might be wrong) and the call comes out as soon as the information about the combatants is ready.

YuzukiTsuru: This looks more reasonable, let me see how to implement it.

YuzukiTsuru: Change P12S Engravement 1 Drop Tower Pos Tracker to the `[19:12:22.471] Tether 23:4001EA0D:Anthropos:etc`, Because you need to capture the locations of the two Tether, I just handle them separately.

https://github.com/quisquous/cactbot/pull/5597/commits/095f49e847fa5c465dfd2f4cd63a209223827cee

quisquous: I'm sorry, but I still think this is a race.  This is now:
```

Comment (quisquous) [reply ui/raidboss/data/06-ew/raid/p12s.ts]: I'm sorry but I still think this is a race.  If I am reading this correctly, this is now:

```
// Anthropos move into position
[19:12:18.911] 261 105:Change:4001EA0D:Heading:0.0000:PosX:97.0000:PosY:75.0000:PosZ:0.0000
// P12S Engravement 1 Drop Tower Pos Tracker + async getCombatants
[19:12:22.471] Tether 23:4001EA0D:Anthropos:etc
// P12S Engravement 1 Tower Drop (hope that the async calls finish?)
[19:12:22.471] StatusAdd 1A:DFC:etc
[19:12:22.516] StartsCasting 14:4001EA0D:Anthropos:82F1:etc
```

There are 4 async calls there that need to be done instantaneously for the StatusAdd line to have the information there?

I'm trying to propose this structure:

`P12S Engravement 1 Tether Tracker`
* a new trigger, on the tether
* records all of the tether `matches.sourceId` into some new `data.engravement1TetherIds`

`P12S Engravement 1 Tower Drop` (combine logic from `Tower Drop` and `Tower Pos Tracker`)
* condition: only run if the status effect is on you
* promise: call get combatants with all of the `data.engravement1TetherIds`
* alertText:
  * for each combatant, do the logic from `P12S Engravement 1 Drop Tower Pos Tracker` to set the `data.engravement1BeamsPosMap`
  * then do the rest of the normal logic for Tower Drop to use that map

The timing in the PR right now is:
* tethers, call 4 async functions
* immediately after, status effects, hope async functions are done??

This timing is:
* tethers collect ids synchronously
* immediately after, status effects call 1 async function, alert text will run naturally when async function is done (no matter when it completes)

Does that make sense?

Comment (YuzukiTsuru) [reply ui/raidboss/data/06-ew/raid/p12s.ts]: This is a good method, I did not know that getCombatants can query multiple data, so I wrote it separately. but for now this is great way to deal with it.

https://github.com/quisquous/cactbot/pull/5597/commits/b60f413ff26c4c5ee41972a46e2f801fb43d77f3

Comment (YuzukiTsuru) [reply ui/raidboss/data/06-ew/raid/p12s.ts]: This is a good method, I didn't know that getCombatants can take multiple objects at once, so writing them separately caused the problem, Sorry for that.

https://github.com/quisquous/cactbot/pull/5597/commits/b60f413ff26c4c5ee41972a46e2f801fb43d77f3

Comment (quisquous) [reply ui/raidboss/data/06-ew/raid/p12s.ts]: Thank you, this is perfect.

For what it's worth, I think the "multiple objects" part is minor, but it's the asynchronous delay that is the major issue.  If the asynchronous call took (for example) an entire second to complete for some reason (weird network delays, very slow computer, bugs) then the previous triggers would not work because the `Tower Drop` trigger would run before `Tower Pos Tracker` had set any values in the map (because the asynchronous function hadn't completed yet).  With this latest PR change, no matter how long the asynchronous function takes, the trigger logic will still work.

File: ui/raidboss/data/06-ew/raid/p12s.ts [hunk 10]
```diff
@@ -221,7 +221,7 @@ const triggerSet: TriggerSet<Data> = {
           '仅提示塔颜色': 'tower',
         },
       },
-      default: 'sharp',
+      default: 'quadrant',
```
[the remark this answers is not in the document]
quisquous: I will write some release notes about this so people know about it.  If it causes problems, we can switch the default to `tower` in a followup.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -604,30 +642,191 @@ const triggerSet: TriggerSet<Data> = {
...
+            return;
+          const y = data.combatantData[index]?.PosY;
+          if (y === undefined)
+            return;
+
+          const ids = data.engravement1TetherIds[index];
+          if (ids === undefined)
+            return;
+
+          const tempColor = data.engravement1TetherPlayers[ids.toString(16).toUpperCase()];
```
quisquous, line 719:
```suggestion
        for (const combatant of data.combatantData) {
          const x = combatant.PosX;
          const y = combatant.PosY;
          const combatantId = combatant.ID;
          if (combatantId === undefined)
            return;

          const tempColor = data.engravement1TetherPlayers[combatantId.toString(16).toUpperCase()];
```

`data.combatantData` is (unfortunately) not sorted in the order as the ids provided, and so you can't use the same index to figure out the color.

PR #5600, opened 2023-06-23
Update zurvan-un.js

ish-edge:
add script for Ja


xiashtra:
Instead of editing the Javascript directly, you'll need to edit the Typescript files that the Javascript is generated from.

quisquous:
> Instead of editing the Javascript directly, you'll need to edit the Typescript files that the Javascript is generated from.

Yeah, and even more specifically you shouldn't edit the `triggers` branch, you need to edit the `main` branch.

PR #5603, opened 2023-06-23
raidboss: update Bardam's Mettle for 6.4 changes

### LINKED ISSUE #5535: Raidboss issue during Bardams Mettle
The warning to run away from boss when you have a tether on the first boss goes off repeatedly instead of being 1 warning, also comes back when the boss charges at you

xiashtra:
Resolves #5535

### DIFF
--- a/util/logtools/make_timeline.ts
+++ b/util/logtools/make_timeline.ts
@@ -94,6 +94,10 @@ const ignoredCombatants = PetData['en'].concat([
   'Mikoto',
   'Liturgic Bell',
   'Zero',
+  'Gosetsu',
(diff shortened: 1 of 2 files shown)


util/logtools/make_timeline.ts
```diff
@@ -94,6 +94,10 @@ const ignoredCombatants = PetData['en'].concat([
   'Mikoto',
   'Liturgic Bell',
   'Zero',
+  'Gosetsu',
+  'Mol Youth',
```
quisquous, line 98: Thanks for this too!

PR #5608, opened 2023-06-24
raidboss: update Doma Castle for 6.4 duty Support

xiashtra:
Triggers were ok for this one, but there were changes to the boss timelines and a non-targetable actor name change.

Bosses will now walk center when using certain actions, making it very easy to induce timeline drift by dragging the boss off-center.  The first boss is particularly prone to timeline de-sync; however, most groups will likely not see it if they are not intentionally moving the boss more than necessary.


quisquous:
Thank you! I super appreciate fixing up this old content.

PR #5612, opened 2023-06-25
raidboss: update Ala Mhigo for 6.4

xiashtra:
Timeline changes, primarily to the second boss.

### DIFF
--- a/ui/raidboss/data/04-sb/dungeon/ala_mhigo.txt
+++ b/ui/raidboss/data/04-sb/dungeon/ala_mhigo.txt
@@ -3,17 +3,18 @@ hideall "--sync--"
 
 # Guard Scorpion
 # -ii 2049 204A 204B 204C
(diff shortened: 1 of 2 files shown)


quisquous:
Are there any oopsy changes, here or in the other 6.4 updates?

xiashtra:
I haven't been checking for oopsy changes, but from what I saw of the timelines, the action ids are the same, just the timing changed for some things or some actions were removed (e.g., no more claws chasing you to your body in Ala Mhigo 2nd boss).

Biggest impact to oopsy would be the second boss of Castrum Abania since it's an all new encounter.

quisquous:
Ok ok! If you have any time to look while you have logs and are looking through it, I'd definitely appreciate the oopsy end of things too, since you also have the freshest knowledge on what exists or not.  It doesn't have to be perfect, but even just "these things don't exist anymore" -> remove and "here's some new ability ids" -> add (or a TODO) would be good.

PR #5616, opened 2023-06-26
util: make_timeline - sort ignored combatants

xiashtra:
Sorts the ignored combatants list in `make_timeline.ts` for easier future maintenance.  No entries added or removed.


quisquous:
Thanks, this was definitely getting unwieldy.

I wonder if we need some better method of detecting npc combatants, recording their ids, and then removing them that way.

xiashtra:
The Duty Support npcs are considered pets of the player, so a way to detect and exclude pets would automatically catch most of this list.

PR #5618, opened 2023-06-26
raidboss: Endsinger Extreme improvements

quisquous:
* handle head direction during star knockback
* combine Elenchos inside/outside with other mechanics
* Telomania call
* chains calls
* timeline numbering


quisquous:
Mostly tested, but probably needs a little bit more to make sure I didn't break anything during the last bits of cleanup.

PR #5625, opened 2023-06-28
raidboss: add p12s superchain2b

wexxlee:
Has been tested in game and in emulator, but could use some additional validation and feedback, particularly on alert timings and collisions.

One odd thing is that for the Sample Tiles combatants/triggers, the trigger works fine in game, but the emulator never  return position data (via getCombatants) for the tiles.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -236,6 +236,7 @@ export interface Data extends RaidbossData {
   superchain2aFirstDir?: 'north' | 'south';
   superchain2aSecondDir?: 'north' | 'south';
   superchain2aSecondMech?: 'protean' | 'partners';
+  superchain2bFirstDir?: 'north' | 'south';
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -1984,14 +1977,270 @@ const triggerSet: TriggerSet<Data> = {
...
+        if (x > 100) {
+          data.superchain2bSecondDir = 'east';
+          return output.combined!({ dir: output.east!(), mechanic: mechanicStr });
+        }
+        data.superchain2bSecondDir = 'west';
+        return output.combined!({ dir: output.west!(), mechanic: mechanicStr });
+      },
+      outputStrings: {
+        combined: {
+          en: '${dir} (Side) => ${mechanic} After',
```
quisquous, line 2044: Can this say left/west => mechanic, since most folks are standing north/south, looking at the boss and a left/right call could helpful.

I also personally think having this call earlier would be better so folks know which direction they can already start moving since the donut is big.

PR #5620, opened 2023-06-26
raidboss: more P12S second phase

kshman:
I made this PR for someone who can make more better triggers.
If this PR is not enough for purpose, please discard this.
Thank you.

Added:
* Palladian Grasp
* Gaiaochos
* The Classical Concepts
* Caloric Theory

PR #5628, opened 2023-06-29
raidboss: refactoring of P12S second phase #5620

quisquous:
* rewrote the Ultima Blow tethers to not depend on roles (since people can be dead and it can go weird)
* used ids in more places
* fixed headmarker id
* updated English text
* changed some alert severities (also by making some things responses)
* rewrote the classical 2 tracking which was wrong for the final player

### DIFF
--- a/resources/outputs.ts
+++ b/resources/outputs.ts
@@ -102,6 +102,9 @@ export default {
     cn: '双T扇形死刑',
     ko: '광역 탱버',
   },
+  tankBusterCleavesOnYou: {
+    en: 'Tank Cleaves on YOU',
+  },
   avoidTankCleave: {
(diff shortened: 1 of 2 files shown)


quisquous:
Sorry about rewriting the caloric 2 changes, but I found that the counting was not correct and it told the final person to pass the puddle (which is also incorrect).  I rewrote it to say "pass fire" for 1-7 and then "move" for 7-8 once it has been passed.

image

quisquous:
@kshman Apologies if this is too heavy-handed of an edit of your original PR.  If there's stuff I changed that I broke or that you think is wrong, please discuss here or send another PR.  I'm going to merge this now just to get these changes in, but I'm happy to revisit any of this.

PR #5637, opened 2023-07-01
raidboss: improvements to p12s pangenesis

### LINKED ISSUE #5630: Add P12S Pangenesis strat options
The current option is incorrect for 2+0 aka HRT strat.  See: https://raidplan.io/plan/zN-5X4rH7RzEE7gi or https://raidplan.io/plan/8gl9O5PGp0X467rO

The person who takes the first tower and doesn't get a debuff needs to go to the opposite instead of same tower.

wexxlee:
Closes #5630 .

Added three config options:  1+2 (current), 0+2 (newly added), and 'agnostic' (which will not call stays or swaps associated with any particular strat, and will only call a swap if the last tower color is now the same as the player's debuff color (meaning they will die and wipe if they don't swap).  Set agnostic as the config default since it will correctly function regardless of strat.

(Also, small note re: the comments on #5630 - in the 0+2 strat, the players who take the first towers _always_ both swap to take the second (opposite color) towers.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -2834,7 +2834,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'P12S Pangenesis Collect',
       type: 'GainsEffect',
       netRegex: { effectId: pangenesisEffectIds },
-      condition: (data) => !data.pangenesisDebuffsCalled && !data.isDoorBoss,
+      condition: (data) => !data.pangenesisDebuffsCalled && data.phase === 'pangenesis',
       run: (data, matches) => {
         const id = matches.effectId;
         if (id === pangenesisEffects.darkTilt) {
@@ -2965,7 +2965,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'P12S Pangenesis Tilt Gain',
       type: 'GainsEffect',
       netRegex: { effectId: [pangenesisEffects.lightTilt, pangenesisEffects.darkTilt] },
-      condition: (data, matches) => matches.target === data.me && !data.isDoorBoss,
+      condition: (data, matches) => matches.target === data.me && data.phase === 'pangenesis',
       run: (data, matches) => {
         const color = matches.effectId === pangenesisEffects.lightTilt ? 'light' : 'dark';
         data.pangenesisCurrentColor = color;
@@ -2975,7 +2975,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'P12S Pangenesis Tilt Lose',
       type: 'LosesEffect',
       netRegex: { effectId: [pangenesisEffects.lightTilt, pangenesisEffects.darkTilt] },
-      condition: (data, matches) => matches.target === data.me && !data.isDoorBoss,
+      condition: (data, matches) => matches.target === data.me && data.phase === 'pangenesis',
       run: (data) => data.pangenesisCurrentColor = undefined,
     },
     {
@@ -2983,7 +2983,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'Ability',
       // 8343 = Umbral Advent (light tower), 8344 = Astral Advent (dark tower)
       netRegex: { id: ['8343', '8344'] },
-      condition: (data, matches) => matches.target === data.me && !data.isDoorBoss,
+      condition: (data, matches) => matches.target === data.me && data.phase === 'pangenesis',
       run: (data, matches) => {
         const color = matches.id === '8343' ? 'light' : 'dark';
         data.lastPangenesisTowerColor = color;
@@ -2995,7 +2995,7 @@ const triggerSet: TriggerSet<Data> = {
       // 8343 = Umbral Advent (light tower), 8344 = Astral Advent (dark tower)
       // There's always 1-2 of each, so just watch one.
       netRegex: { id: '8343', capture: false },
-      condition: (data) => !data.isDoorBoss,
+      condition: (data) => data.phase === 'pangenesis',
       preRun: (data) => data.pangenesisTowerCount++,
       suppressSeconds: 3,
       alarmText: (data, _matches, output) => {
@@ -3019,7 +3019,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'GainsEffect',
       netRegex: { effectId: pangenesisEffects.lightTilt, capture: false },
       condition: (data) => {
-        if (data.isDoorBoss)
+        if (data.phase !== 'pangenesis')
           return false;
         return data.lastPangenesisTowerColor !== undefined && data.pangenesisTowerCount !== 3;
       },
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -406,6 +408,21 @@ const triggerSet: TriggerSet<Data> = {
...
+    {
+      id: 'pangenesisFirstTower',
+      name: {
+        en: 'Pangenesis: First Towers',
+      },
+      type: 'select',
+      options: {
+        en: {
+          'Call Required Swaps Only': 'agnostic',
+          '0+2 (Conga Line/HRT)': 'not',
```
quisquous, line 420: Aren't they all conga line?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -2808,7 +2834,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'P12S Pangenesis Collect',
       type: 'GainsEffect',
       netRegex: { effectId: pangenesisEffectIds },
-      condition: (data) => !data.pangenesisDebuffsCalled && !data.isDoorBoss,
+      condition: (data) => !data.pangenesisDebuffsCalled && data.phase === 'pangenesis',
```
quisquous, line 2837: I guess I don't understand this change or all of the other ones below.  When do these effects happen in the final boss other than in this phase? Is this a logic change or just cleanup?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -406,6 +408,21 @@ const triggerSet: TriggerSet<Data> = {
...
+      id: 'pangenesisFirstTower',
+      name: {
+        en: 'Pangenesis: First Towers',
+      },
+      type: 'select',
+      options: {
+        en: {
+          'Call Required Swaps Only': 'agnostic',
+          '0+2 (Conga Line/HRT)': 'not',
+          '1+2 (Yuki)': 'one',
```
quisquous, line 421: Is this also "Rinon" strat? Or am I misremembering?

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -2999,25 +3055,46 @@ const triggerSet: TriggerSet<Data> = {
...
+        // 2nd towers
+        if (data.pangenesisTowerCount === 1) {
+          if (strat === 'not') {
+            // in the 0+2 strat, 2nd tower responsibilities are fixed based on 1st tower soaks.
+            // the shortLight, shortDark, and both 'not' players always take the northern (opposite color) towers.
+            // the longLight, longDark, and both 'one' players soak the southern (same color) towers - per their still-active Pangenesis Initial call
+            const swapRoles: PangenesisRole[] = ['not', 'shortLight', 'shortDark'];
+            if (swapRoles.includes(myRole))
+              return { infoText: output[switchOutput]!() }; // infoText because, although a switch, it's 100% anticipated and should be treated as a reminder
+            return;
```
quisquous, line 3077: Should we call out the "stay output" for the other people waiting to take the second tower? I guess you can only know what the color is for the longs, though.  Hmm.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -2902,14 +2952,20 @@ const triggerSet: TriggerSet<Data> = {
           cn: '黑2: 踩第2轮白塔',
           ko: '긴 어둠 (두번째 빛 대상)',
         },
+        firstTower: {
+          en: 'First Tower',
+        },
+        secondTower: {
+          en: 'Second Tower',
+        },
```
quisquous, line 2960: Do we need a merge version as well for the 0+2 strat?

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Aren't they all conga line?

Are they?  I don't know 1+2, and I've seen 0+2 referred to as conga line, but I don't know if that's a feature in common.  If it's ambiguous though, I can take it out.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: I just watched Rinon's guide, and yes it is.  I'll add that.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > When do these effects happen in the final boss other than in this phase? Is this a logic change or just cleanup?

Purely stylistic/cleanup - they don't otherwise happen in the final phase.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Should we call out the "stay output" for the other people waiting to take the second tower? I guess you can only know what the color is for the longs, though. Hmm.

I avoided doing a call here on the 2nd towers for the other people, because they still have the Pangenesis Initial call up (longer duration) that tells them which second tower to take, and doing a new infoalert saying the same thing feels noisy.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Do we need a merge version as well for the 0+2 strat?

Sorry, not sure I understand what you're referring to by 'merge version.'  Do you mean the DNA slime making?  I don't think we call that out anywhere else in pangenesis.  Adding a "stack with partner" call for this or for towers would add more length to the text output or would get noisy if it's a second alert, so I'm hesitant.  (Also, tbh, I'm not entirely sure which sets of partners the slimes will come from - I think it can vary depending on who swaps?)

quisquous: Ah sorry, mostly saying rather than just saying "you're a 1, get second tower" should it say "you're a 1, get second tower but also merge first" (obviously not in those words).

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: Ok, sounds good!

ui/raidboss/data/06-ew/raid/p12s.ts
quisquous: We do 1+2 with a conga line.  I think everybody does conga line or some version of flex to figure out who goes to which side when you both have the same debuff.  But, I think that's strat agnostic.

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Ah sorry, mostly saying rather than just saying "you're a 1, get second tower" should it say "you're a 1, get second tower but also merge first" (obviously not in those words).

Ohhhh, okay.  Yeah, sorry, I was being dense.  Done.

PR #5644, opened 2023-07-04
raidboss: p9s additional limit cut 1 info

### LINKED ISSUE #5642: Feature request: Enhance P9S LC1 callouts
In P9S, at the start of the limit cut 1 phase, the callouts should be expanded as follows:

"2" -> "2 - First dash, third tower"
"4" -> "4 - Second dash, last tower"
"6" -> "6 - First tower, third dash"
"8" -> "8 - Second tower, last dash"
(No callout) -> "No number - stand by for defamation"

Rationale: 
- This informs the numbered players of their role in the mechanic's sequence. This sequence is the same for all strats and enforced by the mechanic. 
- It also gives positive confirmation to the non-numbered players that they are the ones who need to wait for their turn to be randomly selected for the defamation (ice) marker.

PS. Cactbot is also missing an raidwide AOE damage callout just after the LC1 phase.

wexxlee:
Closes #5642 .  I ran this against a reclear log from this morning in emulator and it looks fine, but could probably benefit from additional testing.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p9s.ts
+++ b/ui/raidboss/data/06-ew/raid/p9s.ts
@@ -631,7 +631,7 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
-      id: 'P9S Limit Cut Defamation',
+      id: 'P9S Limit Cut 1 Defamation',
       type: 'HeadMarker',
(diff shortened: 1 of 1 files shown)


quisquous:
Looks good to me!

PR #5645, opened 2023-07-05
raidboss: minor Eureka Orthos additions/fixes

quisquous:


xiashtra:
Do you have a 6.4 log of floor 40?  I suspect `Twintania Clone` was also renamed, to `Twintania's Clone`, and needs a similar update.

quisquous:
> Do you have a 6.4 log of floor 40? I suspect `Twintania Clone` was also renamed, to `Twintania's Clone`, and needs a similar update.

Oh! Good call. I guess I was too tired / focusing on mechanics to notice there weren't callouts there either.  <_<

@Akurosia are there updated translations in other languages for these?

Akurosia:
Sry currently not in reach of my pc. Will take a day or 2 to verify

Akurosia:
Regarding Eureka Orthos i did not see any other changes (as far as i am aware of the enemy names) but i made a small repo with the files i use (from what i had left) so i compared 6.38 and 6.41 which you can find here: https://github.com/Akurosia/ffxiv_translate_json/compare/37d362b..c66ded8

I also noted some other changes which might affect other fights. And i dont really know if these are really used once or not as we do not have a list of bnpcid in the files itself to easy reference them.

Might be a thing we should considere in generall to use bnpc IDs instead of Names whereever posible (easier to maintain, less to translate)

quisquous:
> Might be a thing we should considere in generall to use bnpc IDs instead of Names whereever posible (easier to maintain, less to translate)

Unfortunately there's only bnpcid on the added combatant line but not on other lines.  Other options here could be that the translation section could have a bnpcid -> name section (like the syncs, but just referenced differently) or alternatively alternatively we could consider dropping names from triggers (and timelines?)

xiashtra:
> > Might be a thing we should considere in generall to use bnpc IDs instead of Names whereever posible (easier to maintain, less to translate)
> 
> Unfortunately there's only bnpcid on the added combatant line but not on other lines. Other options here could be that the translation section could have a bnpcid -> name section (like the syncs, but just referenced differently) or alternatively alternatively we could consider dropping names from triggers (and timelines?)

Or, we could get Ravahn to include the bnpcid on the other log lines like it should be 😄

PR #5649, opened 2023-07-06
raidboss: TOP add missing replaceSync

Souma-Sumire:


quisquous:
Are these all of the missing translations? It looks like `missingTranslations: true` is still set.

Souma-Sumire:
I'm not sure, but it appears that way.

Additional question: So should we remove missingTranslations when it's complete?
(I was unaware of the purpose of this property, as it only exists in the ja and fr branches in the main branch, and it doesn't exist in the cn and de branches. I don't understand the other languages, so I can't discern their differences.)

quisquous:
Yes, definitely, remove it if unneeded.

If you run `npm run util -- findTranslations -f . -l fr` (or `-l ja` etc etc) it will show you all the missing translations for that language, including files where `missingTranslations` is set to true when not needed.

Souma-Sumire:
Thank you for the tutorial. After running the command, I see a lot of unresolved issues, and I believe this is a massive project that lacks long-term maintenance.

quisquous:
I think there's a number of issues, for sure. I think visibility (as probably only a few people know that command at this point, and so don't understand which translations are missing) is definitely part of it. I'll look into adding the missing translations as part of the build process (maybe the coverage generator??) and so it will be much more obvious which fights are missing translations and where.

quisquous:
https://github.com/quisquous/cactbot/pull/5657 should address some of this. I'll add to my list to write some better documentation for translators as well.

PR #5662, opened 2023-07-09
raidboss: fix P10S bond and other triggers for non-English

### LINKED ISSUE #5643: P10S Daemonic Bond  callout is opposite for non en users
Hi,

I am using a jp xiv client and I noticed cactbot is calling out the opposite for P10S Daemonic Bond(i.e. `partnersThenSpread` is called as `spreadThenPartners`).

After looking into the source code, I realized it's because boss name is hard [coded](https://github.com/quisquous/cactbot/blob/main/ui/raidboss/data/06-ew/raid/p10s.ts#L22C1-L22C44) as `const bossNameUnicode = 'Pand\u00e6monium';` while in jp client boss name is `パンデモニウム`.

This results in 

```
    {
      // Daemoniac Bonds starts casting
      // Then all of the Daemoniac Bonds DDE (spread) effects go out
      // Then 4x DDF or 2x E70 effects go out.
      id: 'P10S Pandaemoniac Bonds Cleanup',
      type: 'StartsUsing',
      netRegex: { id: '82A1', source: bossNameUnicode, capture: false },
      run: (data) => {
        delete data.daemonicBondsTime;
        delete data.bondsSecondMechanic;
        data.daemonicBondsCounter++;
      },
    },
```

not getting triggered and is still remembering the previous `data.daemonicBondsTime`(hence the opposite callout)

I can see in replaceSync the boss name is replaced with `'Pand\\\\u00e6monium': 'パンデモニウム',` [here](https://github.com/quisquous/cactbot/blob/34708993fef5a84adeefdd95953ece9e73282a98/ui/raidboss/data/06-ew/raid/p10s.ts#L708C5-L708C42) but probably the espacing is not working properly?

After I change this line to 

`[bossNameUnicode]: 'パンデモニウム',` it now works perfectly.

quisquous:
The translation of the boss name did not work properly and so no trigger with a `source` property was firing.

Fixes #5643.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p10s.ts
+++ b/ui/raidboss/data/06-ew/raid/p10s.ts
@@ -634,8 +634,8 @@ const triggerSet: TriggerSet<Data> = {
       'missingTranslations': true,
       'replaceSync': {
         'Arcane Sphere': 'arkan(?:e|er|es|en) Körper',
-        'Pand\\\\u00e6moniac Pillar': 'pand\\u00e6monisch(?:e|er|es|en) Turm',
-        'Pand\\\\u00e6monium': 'Pand\\u00e6monium',
+        'Pand\u00e6moniac Pillar': 'pand\u00e6monisch(?:e|er|es|en) Turm',
+        'Pand\u00e6monium': 'Pand\u00e6monium',
       },
       'replaceText': {
         '\\(cast\\)': '(Wirken)',
@@ -671,8 +671,8 @@ const triggerSet: TriggerSet<Data> = {
       'missingTranslations': true,
       'replaceSync': {
         'Arcane Sphere': 'sphère arcanique',
-        'Pand\\\\u00e6moniac Pillar': 'pilier pand\\u00e6moniaque',
-        'Pand\\\\u00e6monium': 'Pand\\u00e6monium',
+        'Pand\u00e6moniac Pillar': 'pilier pand\u00e6moniaque',
+        'Pand\u00e6monium': 'Pand\u00e6monium',
       },
       'replaceText': {
         'Bury': 'Impact',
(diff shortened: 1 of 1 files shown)


wexxlee:
> The translation of the boss name did not work properly and so no trigger with a `source` property was firing.

I wasn't aware that `timelineReplace` also did substitution on `netRegex` (or triggers generally) - I just assumed it was timeline-only functionality.  Worth a small update to RaidbossGuide?

quisquous:
O:)

ui/raidboss/data/06-ew/raid/p10s.ts
```diff
@@ -671,8 +671,8 @@ const triggerSet: TriggerSet<Data> = {
       'missingTranslations': true,
       'replaceSync': {
         'Arcane Sphere': 'sphère arcanique',
-        'Pand\\\\u00e6moniac Pillar': 'pilier pand\\u00e6moniaque',
```
YetAnotherIC, line 674: Thanks for the quick fix!

qq: If we already have `const bossNameUnicode = 'Pand\u00e6monium';` defined in the `ts` file, it might be better to just `[bossNameUnicode]: 'Pand\u00e6monium',` here to to ensure boss name is absolutely the same? I can submit a PR for this if you want. (I am sure this pr works, but just saying as a better code practice, we can even make this part of the RaidbossGuide translation guide to avoid issues like this in the future, for example "we recommend extract `replaceSync` keys as a shared variable to avoid mistype issues")?

quisquous: I think it is much more complicated than this, unfortunately. This is not a typo or a mistype, and the original entry is correct to translate timeline syncs. However, due to the way timelines vs triggers are parsed this same entry does not match the trigger. On top of that, there is an additional bug that find missing translations did not notice this mistake (when it normally should).

The way this bug would normally be detected is `missingTranslations` would not be set to true, and then `npm run test` would complain that there is not a match. However, one bug here hid the other.

It's also the case that due to translation collisions, the entry in `replaceSync` is not always the exact name of the boss. See  TOP for example, where `Omega` needs negative lookahead and lookbehind to not collide with `Omega-M` and `Alpha-Omega` and is not the literal key `Omega`. So, I don't think shared keys are generally the right solution.

PR #5666, opened 2023-07-09
raidboss: take two on P10S translations

### LINKED ISSUE #5664: P10S Missing German Translations
Given that I ran this through the raidemulator, I think #5662 is a correct fix, however it appears that the find missing translations script thinks that the p10s entries are no longer translated: https://quisquous.github.io/cactbot/util/coverage/missing_translations_de.html

This is also kind of related to #5437.

quisquous:
Sorry, this is more complicated than it should be. Maybe we can avoid unnecessary escaping in the future.

Caused by #5662 (sorry).
Fixes #5664.


quisquous:
There's also still some followup here to understand why find missing translations didn't catch #5643 in the first place.

PR #5675, opened 2023-07-10
i18n: More FR translations

mooondark:


mooondark:
@quisquous 
Yeah, I saw that page, this is why I can translate all the missing lines! Thanks for this 👍

PR #5679, opened 2023-07-10
raidboss: round timeline time to avoid micro jumps

quisquous:


quisquous:
I noticed this while doing #5577 and it seemed inefficient, as any change like this often causes timerbars to be re-added entirely. It's invisible to the user, but a little silly to do. A better change in the future might be to re-use and update the times on bars rather than just add/remove them entirely, but that's out of scope for me right now.

PR #5682, opened 2023-07-10
raidemulator: Various fixes for CombatantMemory lines

valarnin:
Fixes various issues found while writing some custom triggers.

### DIFF
--- a/ui/raidboss/emulator/data/CombatantTracker.ts
+++ b/ui/raidboss/emulator/data/CombatantTracker.ts
@@ -98,9 +98,8 @@ export default class CombatantTracker {
   }
 
   addCombatantFromCombatantMemoryLine(line: LineEvent0x105): void {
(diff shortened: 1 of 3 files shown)


valarnin:
As an aside, one of the custom triggers I was writing was to identify where exactly in p12s p1 the LC line baits need to happen, see screenshot. Not sure if there's interest in me PRing this trigger or not.

image

quisquous:
> As an aside, one of the custom triggers I was writing was to identify where exactly in p12s p1 the LC line baits need to happen, see screenshot. Not sure if there's interest in me PRing this trigger or not.
> 
> image

Do you feel like this is useful to know in practice? Like, that's a lot of information to remember and try to turn into useful information.  How do you use it?

Or, is there some other way to use that information, by saying like "(1 + 3 bait quick)" or "(2 + 4 bait, wait a bit)" and also adjust the durations on the bait triggers to stay up until they happen?

PR #5689, opened 2023-07-13
raidboss: add durations for p12s white flame

quisquous:
Also adjust durations to reflect this.

From @valarnin's comment here: https://github.com/quisquous/cactbot/pull/5682#issuecomment-1631198471

A follow-up to this would be to adjust the calls about baiting to say "quick" or "delayed" or "very delayed" or something depending on how long the wait is.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -291,6 +291,10 @@ const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker']) => {
   return (parseInt(matches.id, 16) - data.decOffset).toString(16).toUpperCase().padStart(4, '0');
 };
 
+export type LimitCutCombatantState = PluginCombatantState & {
+  order?: number;
+};
+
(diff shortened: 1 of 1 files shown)


quisquous:
I'm pretty sure this worked, but I forgot to stream so want to give this another round of testing next week (or somebody else could too) before merging.

Legends0:
I find it a bit confusing when it calls out the baits on #. I get after looking at it it is calling the adds as they come down, but it is easily confused with your number if seeing it for the first time, especially when the prevailing pf strat seems to not need to know anything about the add's jump order.

valarnin:
The main reason I wrote the initial version of this in the referenced PR is because I find it useful to know if the bait is fast or slow for 1/3 or 2/4 group.

quisquous:
Yeah, I think my thought is that this could be included in the bait laser calls (both alert and info) whether it's fast or slow (or very slow). The infotext call about the order would probably be dropped.

quisquous:
I'm going to land this first version with just the durations and then will follow up with more text on each bait.

PR #5691, opened 2023-07-13
raidboss: move P12S2 ultima ray call earlier

quisquous:


quisquous:
This is a WIP. I wanted the "which way to run" call to come out earlier. This works fine for Ultima Ray 1, but I'm not sure what's going on with Ultima Ray 2. I gave it a little bit more delay, but it needs more testing/fixing.

quisquous:
Mmmmmmm I think this sort of mostly works, but doesn't work on the last one. I think I'm just gonna close this as the earlier call is not worth it.

cc @wexxlee just FYI

Issue #5486, opened 2023-06-02
util: have log splitter apply rsv lines to ability names


quisquous:
It'd be convenient if when people used the log splitter (in util/logtools) that it applied any rsv lines (262 lines) from the same file to ability and effect names.

xiashtra:
This could maybe be expanded into allowing the log tools to use a file of substitutions we keep in the project tree.  It would be useful for some of the other log tools (such as the timeline tools), which might get used against a partial log that is missing 262 lines.  For instance, any log split with @valarnin's [log splitter plugin](https://github.com/valarnin/ACT_FFXIV_Log_Splitter_Plugin) will not have the 262 lines, since his plugin splits on 01 zone change lines, which appear after all of the 262 lines are written currently.

quisquous:
For what it's worth, [this log splitter](https://quisquous.github.io/cactbot/util/logtools/splitter.html) always includes 262 lines.  Maybe @valarnin's plugin could use this same logic.  The netlog_defs file has a `globalInclude` boolean which is intended to be a "always include all instances of this line even in a split log".  Similarly there's also `lastInclude` which is "always include the last instance of this line even in a split log".

quisquous:
I think this is done now.

PR #5701, opened 2023-07-16
util: apply rsvs to abilities/effects in log splitter

wexxlee:
Addresses the original scope of #5486 .  Updating make_timeline/test_timeline to use RSV lines (or, for that matter, to use a substitution table) would take some additional work.

Because 262 lines are zone-specific, I unset `globalInclude` and added logic to the log splitter to collect and push only those zone-specific RSV lines for each split log.  (From what I can tell, `globalInclude` is only used by the splitter, but if that's incorrect and we need to make adjustments elsewhere, let me know.)

### DIFF
--- a/resources/netlog_defs.ts
+++ b/resources/netlog_defs.ts
@@ -22,6 +22,8 @@ export type LogDefinition = {
   firstUnknownField?: number;
   // A map of all of the fields, unique field name to field index.
   fields?: { [fieldName: string]: number };
+  // A map of the field(s) that *may* contain RSV keys (for decoding)
+  possibleRsvFields?: { [fieldName: string]: number };
   subFields?: {
     [fieldName: string]: {
       [fieldValue: string]: {
@@ -342,6 +344,9 @@ const latestLogDefinitions = {
       z: 11,
       heading: 12,
     },
+    possibleRsvFields: {
+      ability: 5,
+    },
     blankFields: [6],
     playerIds: {
       2: 3,
@@ -389,6 +394,9 @@ const latestLogDefinitions = {
       sequence: 44,
       targetIndex: 45,
       targetCount: 46,
+    },
+    possibleRsvFields: {
+      ability: 5,
     },
     playerIds: {
       2: 3,
@@ -438,6 +446,9 @@ const latestLogDefinitions = {
       targetIndex: 45,
       targetCount: 46,
     },
+    possibleRsvFields: {
+      ability: 5,
+    },
     playerIds: {
       2: 3,
       6: 7,
@@ -459,6 +470,9 @@ const latestLogDefinitions = {
       id: 4,
       name: 5,
       reason: 6,
+    },
+    possibleRsvFields: {
+      name: 5,
     },
     playerIds: {
       2: 3,
@@ -550,6 +564,9 @@ const latestLogDefinitions = {
       targetMaxHp: 10,
       sourceMaxHp: 11,
     },
+    possibleRsvFields: {
+      effect: 3,
+    },
     playerIds: {
       5: 6,
       7: 8,
@@ -630,6 +647,9 @@ const latestLogDefinitions = {
       targetId: 7,
       target: 8,
       count: 9,
+    },
+    possibleRsvFields: {
+      effect: 3,
     },
     playerIds: {
       5: 6,
@@ -1118,7 +1138,6 @@ const latestLogDefinitions = {
       key: 4,
       value: 5,
     },
-    globalInclude: true,
     canAnonymize: true,
     firstOptionalField: undefined,
   },
--- a/resources/netlog_defs.ts
+++ b/resources/netlog_defs.ts
@@ -22,8 +22,8 @@ export type LogDefinition = {
   firstUnknownField?: number;
   // A map of all of the fields, unique field name to field index.
   fields?: { [fieldName: string]: number };
-  // A map of the field(s) that *may* contain RSV keys (for decoding)
-  possibleRsvFields?: { [fieldName: string]: number };
+  // A list of field ids that *may* contain RSV keys (for decoding)
+  possibleRsvFields?: readonly number[];
   subFields?: {
     [fieldName: string]: {
       [fieldValue: string]: {
@@ -344,9 +344,7 @@ const latestLogDefinitions = {
       z: 11,
       heading: 12,
     },
-    possibleRsvFields: {
-      ability: 5,
-    },
+    possibleRsvFields: [5],
     blankFields: [6],
     playerIds: {
       2: 3,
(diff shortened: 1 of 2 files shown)


resources/netlog_defs.ts
```diff
@@ -1118,7 +1138,6 @@ const latestLogDefinitions = {
       key: 4,
       value: 5,
     },
-    globalInclude: true,
```
quisquous, line 1121: What do you think about leaving this as `globalInclude: true`. My reasoning for setting this was mostly is so that if folks split/anonymize logs and then include them in the contribute logs file, then we'll always have the rsv lines if needed.

util/logtools/splitter.ts
```diff
@@ -89,20 +143,23 @@ export default class Splitter {
...
     lines.push(line);
 
     lines = lines.sort((a, b) => {
-      // Sort by earliest time first, then by earliest log id.
+      // Sort by earliest time first, then by the lowest-numbered type.
       // This makes the log a little bit fake but maybe it's good enough.
-      const aStr = a.slice(3, 36) + a.slice(0, 3);
-      const bStr = b.slice(3, 36) + b.slice(0, 3);
+      const aStr = (a.split('|')[1] ?? '') + (a.split('|')[0] ?? '');
+      const bStr = (b.split('|')[1] ?? '') + (b.split('|')[0] ?? '');
```
quisquous, line 154: What are these changes for?

util/logtools/splitter.ts
```diff
@@ -20,9 +26,33 @@ export default class Splitter {
     private notifier: Notifier,
     private includeGlobals: boolean,
   ) {
-    // Remap logDefinitions from log type (instead of name) to definition.
-    for (const def of Object.values(logDefinitions))
+    for (const def of Object.values(logDefinitions) as LogDefinition[]) {
```
quisquous, line 29: Is there a way to avoid `as LogDefinition[]` here?

resources/netlog_defs.ts
```diff
@@ -342,6 +344,9 @@ const latestLogDefinitions = {
       z: 11,
       heading: 12,
     },
+    possibleRsvFields: {
+      ability: 5,
```
quisquous, line 348: Is the name here ever used? I think this should probably be either `'ability'` or `'5'` and not redundantly both?

util/logtools/splitter.ts
```diff
@@ -67,6 +105,22 @@ export default class Splitter {
...
+        this.rsvLines = {};
+        this.rsvSubstitutionMap = {};
+      }
+      const idIdx = 4;
+      const valueIdx = 5;
+      const rsvId = splitLine[idIdx];
+      const rsvValue = splitLine[valueIdx];
+      if (rsvId !== undefined && rsvValue !== undefined) {
+        this.rsvLines[rsvId] = line;
+        this.rsvSubstitutionMap[rsvId] = rsvValue;
```
quisquous, line 122: Do you feel like there's any need to split statuses and actions? Proooobably we're ok in that they are wildly apart from each other and likely won't ever collide, but they are separate namespaces of ids.

Could you at least leave a TODO here about this?

quisquous (commented):
Ooooh, this is exciting, thank you!

quisquous:
> Addresses the original scope of #5486 . Updating make_timeline/test_timeline to use RSV lines (or, for that matter, to use a substitution table) would take some additional work.

Yeah, this could be a separate PR if somebody wanted to do it, or people could just run logs through the splitter first to un-RSV them if needed for now.

> Because 262 lines are zone-specific, I unset `globalInclude` and added logic to the log splitter to collect and push only those zone-specific RSV lines for each split log. (From what I can tell, `globalInclude` is only used by the splitter, but if that's incorrect and we need to make adjustments elsewhere, let me know.)

Yeah, globalInclude, lastInclude, canAnonymize, isUnknown, firstUnknownFields, playerIds, firstOptionalField, are all log splitter / anonymizer fields.

resources/netlog_defs.ts
wexxlee: > What do you think about leaving this as `globalInclude: true`. My reasoning for setting this was mostly is so that if folks split/anonymize logs and then include them in the contribute logs file, then we'll always have the rsv lines if needed.

So my thinking on removing this was two-fold.  First, they aren't really global includes since they are zone-specific and sent on zone-in.  Second, you can have multiple zone-ins (for that matter, multiple fights) in the same log file, and globalIncludes will export all of them in the split log file.  So if you end up trying to split a log file with a lot of prog hours (particularly across multiple fights), you'll end up with a lot of duplicate rsv line,ms (and possibly some irrelevant ones).

The logic I added to splitter will still export all of the rsv lines for a particular zone in the split/anonymized log, just like we do for combatants (which I think is analogous).  But no dups and no stuff from other zones.

util/logtools/splitter.ts
wexxlee: These slices don't work (or at least don't work as intended, I think), given that the logtype field can be either 2 or 3 characters long (eg all of the OverlayPlugin-emitted lines).  So rather than arbitrarily splitting the line at a fixed length, it splits by field (which I think was the original intent?)

util/logtools/splitter.ts
wexxlee: I tried, but I think we need the forced typing because possibleRsvFields is an optional property, and TS throws a tantrum if the def object doesn't have an explicit type.  I thought the forced typing would be okay in this circumstance because the object values are imported and already of that type (from netdefs) so there's not really any room for whoopsies.   But let me know if you have another idea in mind.

resources/netlog_defs.ts
wexxlee: > Is the name here ever used? I think this should probably be either `'ability'` or `'5'` and not redundantly both?

Nah, it's not used.  I did that to maintain the same structure as `fields`, and so it didn't look like a magic-number situation (also easier to visually see what fields we'll be decoding without having to look at `fields` -- although I suppose a comment would also accomplish the same thing.

util/logtools/splitter.ts
wexxlee: > Do you feel like there's any need to split statuses and actions? Proooobably we're ok in that they are wildly apart from each other and likely won't ever collide, but they are separate namespaces of ids.
> 
> Could you at least leave a TODO here about this?

Yeah, proooobably okay, and I will leave a comment.  Although more to the point, I'm not sure we get the namespace from the 262 line anyway.  There's the (commented-out) `unknown0` (3) field on the line that I suspected might have some relation to namespace, but it doesn't match up with the exd table ids so unless we make some assumptions and build a mapping, I'm not sure how else to validate that.

util/logtools/splitter.ts
quisquous:
```typescript
    const defs: LogDefinitionMap = logDefinitions;
    for (const def of Object.values(defs)) {
```

This will do it. The issue as you say is that because it's an optional property and not every value has this property, the intersection of all of these types does not include it. One option is to add it a non-optional member (that could be undefined, but present). The other option is to more loosely type it as the above.

util/logtools/splitter.ts
quisquous: Oh, oops!

PR #5702, opened 2023-07-16
util: Auto-generate header data in make_timeline

JLGarber:
This aims to simplify the recent "soft standard" of listing out all ignored abilities in detail, as well as the full list of abilities used by enemies in the encounter, ignored or not. Sample output:

```
make_timeline.ts -lf 10 -it Nidhogg -ii 179D 179F

### THE MINSTREL'S BALLAD: NIDHOGG'S RAGE
# ZoneID: 236

# -ii 179D 179F

# Ignored Abilities
# 179D Hot Wing
# 179F Hot Tail

# All Encounter Abilities
# 1795 the Scarlet Whisper
# 1796 Deafening Bellow
# 1797 Cauterize
# 1798 Touchdown
# 1799 --sync--
# 179A Horrid Roar
# 179B Horrid Roar
# 179C Hot Wing
# 179D Hot Wing
# 179E Hot Tail
# 179F Hot Tail
###Loads more elided for space

hideall "--Reset--"
hideall "--sync--"
```

(Those random "--sync--" entries are because the entry processing step prior to this one automatically replaces Unknown abilities. I don't think it's too big a deal, but we can probably add a rule to convert those back to Unknown_XXXX.)

This same technique can probably be applied to auto-listing ignored combatants, but unfortunately that's not quite as simple. The input `-ic MonsterOne "Two-Word Monster" "Three Word Monster"` is rendered in the `args.ignored_combatants` array as `['MonsterOne', 'Two-Word Monster', 'Three Word Monster']`. A simple `.join()` is not enough to restore the quote marks that would be necessary.

Currently the zone info generation does not operate on either raw start/end times or on timelines generated from FFLogs. For raw times, there is simply no way to know the zone information, since invoking this way bypasses the collector-creation step. For FFLogs reports, the data is available in the responses from FFLogs, but we would have to make some out-of-scope adjustments to how fflogs.ts operates. It's completely doable, but I didn't have the motivation for that this weekend.

None of this is elegant, and I should go through and add some in-line comments to bring this in line with the rest of the file, but it's ready for criticism and improvement.

### DIFF
--- a/util/logtools/make_timeline.ts
+++ b/util/logtools/make_timeline.ts
@@ -37,6 +37,8 @@ type TimelineEntry = {
   lineComment?: string;
 };
 
+type EncounterAbilityList = { [string: string]: string };
+
 class ExtendedArgsRequired extends Namespace implements TimelineArgs {
   'file': string | null;
   'force': boolean | null;
@@ -56,6 +58,7 @@ class ExtendedArgsRequired extends Namespace implements TimelineArgs {
   'report_id': string | null;
   'report_fight': number | null;
   'key': string | null;
+  'generate_ability_table': string | null;
 }
 
 type ExtendedArgs = Partial<ExtendedArgsRequired>;
@@ -145,6 +148,11 @@ timelineParse.parser.addArgument(['--phase', '-p'], {
 timelineParse.parser.addArgument(['--include_targetable', '-it'], {
   nargs: '+',
   help: 'Set this flag to include "34" log lines when making the timeline',
+});
+
+timelineParse.parser.addArgument(['--generate_ability_table', '-gat'], {
+  nargs: '?',
+  help: 'Set this flag to include a complete/unfiltered ability list after the timeline',
 });
 
 timelineParse.parser.addArgument(
@@ -449,7 +457,7 @@ const assembleTimelineStrings = (
       phases[ability] = parseFloat(time);
   }
 
-  const encounterAbilityList: { [string: string]: string } = {};
+  const encounterAbilityList: EncounterAbilityList = {};
   for (const entry of entries) {
     if (entry.lineType === 'ability') {
       // In order to list out all abilities we see in the timeline header,
@@ -528,8 +536,12 @@ const assembleTimelineStrings = (
     }
     lastEntry = entry;
   }
-  const headerLines = assembleHeaderAbilityStrings(args, encounterAbilityList);
-  return headerLines.concat(assembled);
+  const ignoreLines = assembledIgnoreHeaderStrings(args, encounterAbilityList);
+  const definiteLines = ignoreLines.concat(assembled);
+  // Generate a complete table of abilities if specified
+  if (args.generate_ability_table !== undefined)
+    return definiteLines.concat(assembleAbilityTableStrings(encounterAbilityList));
+  return definiteLines;
 };
 
 const assembleHeaderZoneInfoStrings = (fight: FightEncInfo): string[] => {
@@ -541,7 +553,7 @@ const assembleHeaderZoneInfoStrings = (fight: FightEncInfo): string[] => {
     headerInfo.push(zoneNameLine);
   }
   if (zoneId !== undefined) {
-    const zoneIdLine = `# ZoneID: ${zoneId}`;
+    const zoneIdLine = `# ZoneId: ${zoneId}`;
     headerInfo.push(zoneIdLine);
   }
   if (headerInfo.length > 0)
@@ -549,17 +561,24 @@ const assembleHeaderZoneInfoStrings = (fight: FightEncInfo): string[] => {
   return headerInfo;
 };
 
-const assembleHeaderAbilityStrings = (
+const assembledIgnoreHeaderStrings = (
   args: ExtendedArgs,
-  encounterAbilityList: { [string: string]: string },
+  encounterAbilityList: EncounterAbilityList,
 ): string[] => {
   const assembled = [];
-  const sortedIgnore = args.ignore_id?.sort();
-  if (sortedIgnore !== undefined) {
-    const iiLine = `# -ii ${sortedIgnore.join(' ')}\n`;
+
+  const sortedCombatantIgnore = args.ignore_combatant?.sort();
+  if (sortedCombatantIgnore !== undefined) {
+    const joinedIgnore = sortedCombatantIgnore.map((x) => `"${x}"`).join(' ');
+    assembled.push(`# -ic ${joinedIgnore}`);
+  }
+
+  const sortedAbilityIgnore = args.ignore_id?.sort();
+  if (sortedAbilityIgnore !== undefined) {
+    const iiLine = `# -ii ${sortedAbilityIgnore.join(' ')}\n`;
     assembled.push(iiLine);
     assembled.push('# Ignored Abilities');
-    for (const id of sortedIgnore) {
+    for (const id of sortedAbilityIgnore) {
       const abilityName = encounterAbilityList[id];
       if (abilityName !== undefined) {
         const detailedIgnoreLine = `# ${id} ${abilityName}`;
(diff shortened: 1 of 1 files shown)


JLGarber:
Now that I've done all that, it occurs to me that maybe this should be behind Yet One More Flag such as `-gh -- generate-header`, since this might get obnoxious when regenerating the timeline for the fifteenth time.

util/logtools/make_timeline.ts
```diff
@@ -512,6 +528,54 @@ const assembleTimelineStrings = (
...
+const assembleHeaderZoneInfoStrings = (fight: FightEncInfo): string[] => {
+  const zoneName = fight.zoneName;
+  const zoneId = fight.zoneId;
+  const headerInfo = [];
+  if (zoneName !== undefined) {
+    const zoneNameLine = `### ${zoneName.toUpperCase()}`;
+    headerInfo.push(zoneNameLine);
+  }
+  if (zoneId !== undefined) {
+    const zoneIdLine = `# ZoneID: ${zoneId}`;
```
quisquous, line 544:
```suggestion
    const zoneIdLine = `# ZoneId: ${zoneId}`;
```

The tiniest of nitpicks, but maybe `ZoneId` to match `ZoneId.SomeZoneTitleHere` in code.

quisquous (commented):
My only real question here is whether this should go at the top or the bottom of the timeline.

The "soft standard" I'd be trying to apply (and asked foophoof to do in P9S) was to put the ability list at the bottom because it's pretty long. Since the timeline can also be viewed in the cactbot config ui, it probably also gets in the way of users trying to figure out how to add custom timeline entries there.

I think many things are ok here, so more than willing to hear different opinions. If I had my druthers, I'd put the zone name, id, and args at the top of the file, and then the ability list at the bottom of the file after the timeline.

> (Those random "--sync--" entries are because the entry processing step prior to this one automatically replaces Unknown abilities. I don't think it's too big a deal, but we can probably add a rule to convert those back to Unknown_XXXX.)

I think that's fine! Unknown_XXXX is not very useful, and the id is plenty to know what that should be.

> This same technique can probably be applied to auto-listing ignored combatants, but unfortunately that's not quite as simple. The input `-ic MonsterOne "Two-Word Monster" "Three Word Monster"` is rendered in the `args.ignored_combatants` array as `['MonsterOne', 'Two-Word Monster', 'Three Word Monster']`. A simple `.join()` is not enough to restore the quote marks that would be necessary.

I think you could always apply double quotes to these and ~mostly be fine? `args.ignored_combatants.map((x) => `"${x}"`).join(' ')` kinda thing?

> Currently the zone info generation does not operate on either raw start/end times or on timelines generated from FFLogs. For raw times, there is simply no way to know the zone information, since invoking this way bypasses the collector-creation step. For FFLogs reports, the data is available in the responses from FFLogs, but we would have to make some out-of-scope adjustments to how fflogs.ts operates. It's completely doable, but I didn't have the motivation for that this weekend.

I think this is all fine. I think the collector step is used 99% of the time and fflogs is used to touch up enrages or find loops but not for most of the lifting.

> BONUS THOUGHTS

Yeah, Yet Another Option :tm: could make sense here to at least drop the full ability table.

A couple of follow-ups that would be nice for me:
- could this list whether abilities were cast+ability, just ability, self-targeted somehow?
- this is probably asking too much, but really the thing I'd want is to know if a particular log included an ability id that was not listed in the table (like a failure case, or a branch, or something), but it's probably an impossible thing to parse this text (and if we were parsing it, we'd probably put it in a separate json file or something)
  - (I mean arguably maybe it *should* be in a separate file at some point [this point??], but I think we're getting further afield from this PR)

xiashtra:
This is something I had a mind to tinker with, so I'm glad to see it.  One thing I would prefer to have in the abilities list is the source field also.  In fact, as the ability list here starts to approach a full dump of everything that occurs in an encounter, it becomes useful as a reference for making raidboss triggers, oopsy triggers, updating sync_files.ts, etc.  With that in mind, it would be nice to dump not only 15/16 lines, but maybe the 14 lines also for those times when cast IDs don't match damage IDs.

JLGarber:
> The "soft standard" I'd be trying to apply (and asked foophoof to do in P9S) was to put the ability list at the bottom because it's pretty long. Since the timeline can also be viewed in the cactbot config ui, it probably also gets in the way of users trying to figure out how to add custom timeline entries there.

This is reasonable and I think it should be relatively simple to split out "list the ignored abilities up top" vs "here's everything dumped at the bottom". This would make the helper functions for each more compact individually anyway, so that's even better.

> I think you could always apply double quotes to these and ~mostly be fine? `args.ignored_combatants.map((x) => "${x}").join(' ')` kinda thing?

Oh, fair point, I didn't consider that the double quotes are idempotent. That should work.

> could this list whether abilities were cast+ability, just ability, self-targeted somehow?

Mm, maybe? State tracking across timeline entries is effectively non-existent right now. Because any `startsUsing` lines have cast times included, it should in theory be Not-Impossible to do something like this. ("It is 2.7 seconds after a `startsUsing` line. Look out for a matching `Ability` line." etc etc) But unfortunately I don't see a good path for adding that quickly here.

> Log-sieve functionality

I think you're right that that's kinda out of scope for `make_timeline`. Honestly, it doesn't seem too difficult to put together a script that takes a log and an ability table as input and then outputs missing entries from each. Or am I misunderstanding what the purpose here is? Maybe we need to move some more utilities out of `make_timeline` and into `encounter_tools`?

> One thing I would prefer to have in the abilities list is the source field also. In fact, as the ability list here starts to approach a full dump of everything that occurs in an encounter, it becomes useful as a reference for making raidboss triggers, oopsy triggers, updating sync_files.ts, etc. With that in mind, it would be nice to dump not only 15/16 lines, but maybe the 14 lines also for those times when cast IDs don't match damage IDs.

That's reasonable! I think it's not too terribly difficult to get that done, but that's maybe a bit much for this PR. I can see a clean path forward to make the basic changes quisquous suggested, but it's going to be slightly longer to do something on this scale. (We would need to discuss how the output table is formatted, how best to handle abilities that can have multiple sources, what kind of argument overrides we might need, etc etc.)

quisquous:
For this PR, I think moving the table to the bottom is sufficient and we can talk about where to go after that elsewhere.

For the future, mostly my thinking here is this. It's useful to have a table of abilities. It's useful to have extra metadata (is this a cast, is this damage, does this hit players/E000000). It's useful to write some notes about these abilities too. Having this information in textual format in a timeline (and arguably in oopsy comments) is better than nothing, but it might be better to have it in a more structured format that can be written and read by tools. If tools could read it as well as write it, then we could have what I suggested above, where you could apply multiple log files and do some sort of in place update to add any abilities that weren't listed. It also would mean that somebody could write like a hypothetical log viewer that added this information to the side in some sort of annotated log viewer. It would probably also make sense to have this in some other file and format that's not a timeline comment or oopsy comment.

JLGarber:
Alright, have some changes. It works. I don't really like the name for the table generation argument. Do we have a better option I can drop in that maybe has a nice 3-letter acronym?

wexxlee:
> Alright, have some changes. It works. I don't really like the name for the table generation argument. Do we have a better option I can drop in that maybe has a nice 3-letter acronym?

Hmm.  How about --list-abilities (--la)?

util/logtools/make_timeline.ts
```diff
@@ -541,41 +553,52 @@ const assembleHeaderZoneInfoStrings = (fight: FightEncInfo): string[] => {
...
+    assembled.push(`# -ic ${joinedIgnore}`);
+  }
+
+  const sortedAbilityIgnore = args.ignore_id?.sort();
+  if (sortedAbilityIgnore !== undefined) {
+    const iiLine = `# -ii ${sortedAbilityIgnore.join(' ')}\n`;
     assembled.push(iiLine);
     assembled.push('# Ignored Abilities');
-    for (const id of sortedIgnore) {
+    for (const id of sortedAbilityIgnore) {
```
quisquous, line 581: Sorry for this nitpicky opinion, but I kind of feel like the `-ii` comment should go at the top, but the `# Ignored Abilities` should go at the bottom with the rest of the table.

xiashtra: Same opinion for the `-ic` comment.  I would like to be able to recreate the entire command line string from one location, preferably in the header.  I think only the ability table(s) should be at the bottom.

JLGarber: Sure thing, we can do that.

PR #5710, opened 2023-07-18
raidboss: remove bad DE Peridialogos translation

xiashtra:
Reported by user on Discord that the German translation for P12S Peridialogos trigger was backwards.  This removes the bad translation for now.


quisquous:
Fixed by #5709

PR #5719, opened 2023-07-22
resources: remove unneeded type assertion

quisquous:
Just a drive-by cleanup.


JLGarber (approved):
Have a drive-by review. This is much clearer after the change, as someone who doesn't do much with this section of the internals.

PR #5721, opened 2023-07-23
raidboss: call caloric 1 beacon for everyone

g0shu:


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -3625,17 +3625,31 @@ const triggerSet: TriggerSet<Data> = {
         data.caloric1Mine = undefined;
       },
       delaySeconds: 1,
-      alertText: (data, _matches, output) => {
+      response: (data, _matches, output) => {
```
quisquous, line 3628: For `response`, you have to put the output strings inside the response, e.g. `output.responseOutputStrings = {` rather than a separate field.

Can you also add a `// cactbot-builtin-response` comment at the top of the response function too so that the config ui knows it's safe to run the function.

See `P12S Palladion White Flame Initial` for an example of both of these.

PR #5725, opened 2023-07-24
raidboss: add simple call for classical concepts

g0shu:


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -294,7 +294,7 @@ const getHeadmarkerId = (data: Data, matches: NetMatches['HeadMarker']) => {
 export interface Data extends RaidbossData {
   readonly triggerSetConfig: {
     engravement1DropTower: 'quadrant' | 'clockwise' | 'tower';
-    classicalConceptsPairOrder: 'xsct' | 'cxts' | 'ctsx' | 'ctxs';
+    classicalConceptsPairOrder: 'xsct' | 'cxts' | 'ctsx' | 'ctxs' | 'shapeAndDebuff';
```
quisquous, line 297: Sorry for one last comment, but can you make this the default? I think this is a better default than bpog.

wexxlee: > Sorry for one last comment, but can you make this the default? I think this is a better default than bpog.

😭

quisquous: Sorry! I think it's better to not make assumptions about strats if we can. I'll make a release note.

PR #5734, opened 2023-07-28
docs: Update timeline guide

JLGarber:
This is (probably) not a complete modernization, but it should be a good first pass. I tried to structure the commits to be sensibly distinct for review and rollback purposes.

For readers' convenience, the formatted revision can be viewed at https://github.com/JLGarber/cactbot/blob/update-timeline-guide-new-table-key/docs/TimelineGuide.md.

### DIFF
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -639,7 +639,6 @@ there's a `Shield Skewer` 5.4 seconds after it.
 
 We'll add any necessary jumps in later.
 
-
 ### Adding Phases
 
 Now on to the second phase.
@@ -1406,7 +1405,6 @@ Here are some options:
 * get more data, and make a timeline for the most common case
 * leave a comment in the timeline
 * if this is an important ability (e.g. tankbuster) put a question mark on it so players know it's not guaranteed
-
 
 ### Handling "Linear" Timelines
 
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -199,13 +199,13 @@ If this is not possible, still try to add wide syncs to the beginning of each ph
 (If an ability name is uncomfortably long,
 or if it otherwise makes sense to modify how it's displayed,
 handle that modification in the timeline replacement section of the trigger file.)
-* loops should use `jump` or `forceJump` to return to an earlier point in the timeline,
+* loops should use `jump` or `forcejump` to return to an earlier point in the timeline,
 rather than using a wide window sync at the beginning of the loop.
 * liberally use whitespace and comments to make the timeline readable
 * do not put any triggers, tts, or any other form of alerts in the timeline file itself
 * use [timeline triggers](#timeline-triggers) from within a trigger file for any alerts
 * add a lookahead window of at least 30 seconds *and* 6 abilities for multi-possibility `jump`s that are not simple loops.
-(If it is a simple loop, use `forceJump`.)
+(If it is a simple loop, use `forcejump`.)
 * comment out syncs from any displayed abilities that are within 7 seconds of each other,
 but do not remove them. (This preserves the ability ID for future maintainers.)
 
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -225,7 +225,7 @@ Articles like `The` can be dropped.
 Raids are numbered through the tier,
 e.g. `t1` through `t13` and `a1s` through `a12s`.
 Savage fights should have an `s` suffix
-while normal fights have an 'n' suffix.
+while normal fights have an `n` suffix.
 (However, this does not apply to coil raids.)
 
 Examples:
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -69,7 +69,7 @@ Everything after that on the current line will be ignored.
 
 ### Entries
 
-Here is some grammar examples of timeline entries.
+Here are some grammar examples of timeline entries.
 Every timeline entry begins with the ability time and the ability name.
 
 `Number "String" (duration Number)`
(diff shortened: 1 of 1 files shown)


docs/TimelineGuide.md
```diff
@@ -167,20 +182,32 @@ which is still excellent.
...
 * add syncs for everything possible
-* always add an Engage! entry, but add syncs in case there's no /countdown
 * if the first boss action is an auto-attack, consider adding a sync for
-  the first auto-attack or at least the first "starts using" line
-* include the command line used to generate the timeline in a comment at the top
-* prefer actions for syncs over rp text, but rp text syncs if that's the only option
-* if you do sync a phase with rp text, add a large window sync for an action
-* use original names for ability text as much as possible
-* loops should use `jump` instead of having previous abilities have large windows
+  the first auto-attack, or at least the first "starts using" line.
```
quisquous, line 186: I think we should change this to say start on combat if there's only one boss fight in the zone.

docs/TimelineGuide.md
```diff
@@ -167,20 +182,32 @@ which is still excellent.
...
+  This will help ensure the timeline is synced correctly after entering combat.
+* include any special command line flags used to generate the timeline in a comment at the top.
+(The ignore-combatant and ignore-ability flags
+are automatically added by the timeline utility for your convenience.)
```
quisquous, line 190: This is a thought for another PR, but maybe we should add the other flags, like `-p` or `-it` etc

docs/TimelineGuide.md
```diff
@@ -167,20 +182,32 @@ which is still excellent.
...
 * liberally use whitespace and comments to make the timeline readable
-* do not put any triggers or tts or alerts in the timeline file itself
-* use [timeline triggers](#timeline-triggers) for any alerts
-* add at least a 30 second lookahead window for loops
-* comment out syncs from any abilities that are within 7 seconds of each other
-(This preserves the ability ID for future maintainers.)
+* do not put any triggers, tts, or any other form of alerts in the timeline file itself
+* use [timeline triggers](#timeline-triggers) from within a trigger file for any alerts
+* add a lookahead window of at least 30 seconds *and* 6 abilities for multi-possibility `jump`s that are not simple loops.
+(If it is a simple loop, use `forceJump`.)
```
quisquous, line 208:
```suggestion
(If it is a simple loop, use `forcejump`.)
```

docs/TimelineGuide.md
```diff
@@ -3,12 +3,14 @@
 This is a guide for people who want to write timelines,
 primarily for cactbot.
 
-NOTE: this guide is a good bit out of date:
+NOTE: When this guide was originally written,
+Cape Westwind was a standard 8-player trial.
+This was removed in Endwalker, and that version of the encounter is no longer accessible.
+However, the timeline creation process for it is still valid,
```
quisquous, line 9: This doesn't have to be done in this PR, but I think it would be useful to pick a fight (it doesn't have to be Cape Westwind) that can have an example log from content that people can run solo (unsynced) such that they can use the example log to make/test.

I think the Cape Westwind example is a little outdated, personally. I think an example that shows off as many of these as possible would be good:
* starting from in combat
* at least one loop
* autos that should be ignored
* multiple abilities, e.g. left OR right cleave
* maybe abilities that should be commented out for noisiness (like exaflare hits or something)

JLGarber: Agreed. I would have liked to have something like that to go along with the rest of this PR, but it's a *lot* of work to build that all out and I wasn't really in a place to do that here. I'm hoping that with just "replace the timeline to something modern" outstanding it will be easier to get that change made.

docs/TimelineGuide.md
JLGarber: Agreed. I should have thought of that when I made the previous listing updates. I'll keep that in mind over the next few weeks, since those should be relatively simple to just dump in as we did with ignored combatants.

docs/TimelineGuide.md
JLGarber: Maybe something along the lines of:
```
If there is only one boss in a zone, the timeline should start on entering combat.
(The timeline utility should handle this automatically.) Unless there's a pressing reason not to, 
add a wide sync to the first usage of a skill that isn't an auto-attack. 
This ensures that even if combat entry is early or late, the timeline will still be synced correctly.
```

quisquous: I think combat entry should be pretty reliable? Certainly more reliable than auto attacks. I guess my thinking is that with auto attacks, you'd want a wide sync, but ideally we shouldn't need to start from autos any more. The other reason for a wide sync is if you are starting from a combat log line (zone seal?) and you're worried people might not have their combat log turned on.

docs/TimelineGuide.md
quisquous: Oh yeah, for sure it's a lot of work. That's why I haven't done it either. Just thinking about the future, but this is a really good change on its own to not mislead people.

docs/TimelineGuide.md
JLGarber: I guess I'm mostly thinking of situations where the boss is pulled but the first auto timing, and therefore the entire timeline, is different depending on whether the pulling tank runs around or not. I've made an update that should probably be close to what you're thinking?

quisquous: Yeah, this is what I was looking for! And, re: tank running around, if you don't sync on autos, is this still an issue? Like, if the boss casts a raidwide at t=11, won't it still do that regardless of how many autos occurred? In other words, even if no autos, my understanding is that "modern" fights are still on script and won't adjust. I'm thinking of the video where people were pulling p5s very carefully with provokes and running and rescue and shirk to avoid any autos until the raidwide so that fflogs didn't start the fight until then and so two minutes didn't line up in awkward places and they'd still get a fast kill time.

JLGarber: I don't know. I haven't specifically checked whether it was changed recently. I know there were some timelines since Endwalker started where I noticed some slippage in the time to the first raidwide, but presumably running multiple times and verifying should catch any such slip. Regardless, the updated wording should cover it.

JLGarber:
One other thing that occurs to me, do we want to modernize the entries appearing at lines 158-168? (I didn't modify those so I can't add a comment in-line from Github's interface.)

```bash
677.0 "Heavensfall Trio"
1044 "Enrage" # ???
35.2 "Flare Breath x3" duration 4
1608.1 "Petrifaction" sync /:Melusine:7B1:/ window 1610,5
1141.4 "Leg Shot" sync /:Mustadio:3738:/ duration 20
# I am just a comment
hideall "--sync--"

28.0 "Damning Edict?" sync /:Chaos:3150:/ window 30,10 jump 2028.0
524.9 "Allagan Field" sync /:The Avatar:7C4:/ duration 31 jump 444.9
1032.0 "Control Tower" duration 13.5 sync /:Hashmal, Bringer Of Order starts using Control Tower on Hashmal/ window 20,20 # start of cast -> tower fall
```

quisquous:
> One other thing that occurs to me, do we want to modernize the entries appearing at lines 158-168? (I didn't modify those so I can't add a comment in-line from Github's interface.)
> 
> ```shell
> 677.0 "Heavensfall Trio"
> 1044 "Enrage" # ???
> 35.2 "Flare Breath x3" duration 4
> 1608.1 "Petrifaction" sync /:Melusine:7B1:/ window 1610,5
> 1141.4 "Leg Shot" sync /:Mustadio:3738:/ duration 20
> # I am just a comment
> hideall "--sync--"
> 
> 28.0 "Damning Edict?" sync /:Chaos:3150:/ window 30,10 jump 2028.0
> 524.9 "Allagan Field" sync /:The Avatar:7C4:/ duration 31 jump 444.9
> 1032.0 "Control Tower" duration 13.5 sync /:Hashmal, Bringer Of Order starts using Control Tower on Hashmal/ window 20,20 # start of cast -> tower fall
> ```

Ah yeah, I think that would make sense to do!

PR #5736, opened 2023-07-29
i18n: fix cn translation for trigger set config output

MaikoTan:
This changes was proposed in #5733 but it is not correct.

@lsl1225 此選項意在指示 TriggerSet 即觸發器集合之整體的輸出形式，但似乎您理解爲使用者自定義觸發器的相關設定。事實上使用者自定義的觸發器與 cactbot 內建的觸發器在設定介面是沒有區別的。
Please review and give an approve if you agree the change in this fixup PR.


lsl1225:
Oh, you should be correct. I have got misunderstood, only looking into my own custom trigger sets and not notice the default ones sorry.
Thank you btw!!

lsl1225 (approved):

PR #5741, opened 2023-08-01
config: add trigger comments

Souma-Sumire:
I attempted to add optional multilingual labels to the trigger configuration. This enhancement aims to assist non-English users in better understanding the purpose of each trigger, rather than relying on non-native language IDs.
image


ui/raidboss/raidboss_config.ts
```diff
@@ -684,11 +688,24 @@ class RaidbossConfigurator {
...
+        triggerDivId.innerHTML = trig.isMissingId ? '(???)' : trigId;
+        triggerDivId.classList.add('trigger-id');
+        triggerDiv.appendChild(triggerDivId);
+
+        // Build the trigger explanatory label
+        const trigLabel = trig.label?.[this.displayLanguage] ?? trig.label?.en;
+        if (trigLabel && trigLabel !== triggerDiv.innerHTML) {
+          const triggerLabel = document.createElement('div');
+          triggerLabel.innerHTML = trigLabel;
+          triggerLabel.classList.add('trigger-label');
```
quisquous, line 705: I think you should use a different CSS class here, as `trigger-label` is also used in the `addTriggerDetail` function on line 478 for the headings for each of the trigger fields. You could also call this "trigger comment" instead of "label" if you needed a different word.

Once you add a different CSS class, my only other suggestions are:

(1) I think it would be nice to have a little bit of vertical whitespace between the trigger id and the trigger label
(2) It would be nice to handle very long text better; right now very long text just pushes the right part of the grid over and it's very hard to read

Something like this will help the label be more visually distinctive and not look like part of the id:

```css
/* there's already a .trigger-label, so need something else here */
.trigger-comment {
  display: block; /* this might not be needed */
  font-size: 12px;
  margin-top: 8px;
  margin-left: 8px;
}
```

Something like this will help very long text (maybe 300px or 400px or something like that?)

```diff
 .trigger-details {
   display: grid;
-  grid-template-columns: repeat(1, auto 1fr);
+  grid-template-columns: repeat(1, minmax(auto, 300px) 1fr);
   grid-gap: 3px 10px;
 }
```

quisquous (commented):
This is a great idea, thank you! I have a few tiny CSS suggestions. If you feel up to it, feel free to add documentation for this in `RaidbossGuide.md` but I can do it myself as well.

(Also, if you want an idea for another similar PR after this one, I think labels on trigger config ui  options would also be helpful; it's not great that people have to read release notes to understand what options are, and it'd be good to put that text near the options.)

Souma-Sumire:
Thanks for the tips! I'll give it a shot, but I'm not too great at it, so it might take a while.

At first, I was inspired because I heard someone mention why the IDs were not translated. So, I placed the text near the IDs as optional translation text, while keeping the original IDs for easy reference and feedback on issues. However, now I agree more with your suggestion of using 'comment,' as it seems more appropriate.

I should be able to update the documentation in Chinese, but for other languages, I'm not able to do it myself as I heavily rely on translation software~

Souma-Sumire:
I think I've completed the functionality preliminarily. 
Although the empty string solution might not be the best.
Might have submitted messy git commits, sorry.

resources/user_config.ts
```diff
@@ -78,6 +78,7 @@ export type ConfigEntry = {
     options: BaseOptions,
     isDefault: boolean,
   ) => ConfigValue | void | undefined;
+  comment?: LocaleText;
```
quisquous, line 81: I agree with you about the empty string being slightly odd. I preferred this the way it was before when it was `Partial<LocaleText>`. I think most comments are not going to have an English entry, and so the empty string seems unnecessary a little.

Souma-Sumire: Updated! 'comment' type has been reverted to Partial<LocaleText>.

docs/RaidbossGuide.md
```diff
@@ -362,6 +363,17 @@ response: (data, matches, output) => {
 },
 ```
 
+**comment**
+An object where keys represent optional strings in various languages. This property is an optional auxiliary attribute used to display text around the trigger item in the cactbot configuration panel. You can use it to explain your trigger, leave some descriptive text, or even include a hyperlink.
```
quisquous, line 367:
```suggestion
An object where keys represent optional strings in various languages.
This property is an optional auxiliary attribute used to display text around the trigger item in the cactbot configuration panel.
You can use it to explain your trigger, leave some descriptive text, or even include a hyperlink.
```

See: https://sembr.org/ about semantic line breaks.

quisquous (commented):
A few last suggestions, my apologies.

Personally I think it makes more sense to have the comment at the top where you had it before, as it's kind of like `id` for other languages and might help with readability for people who aren't playing in English. So, I think it would make sense to include after `id`, personally.

Can you also add `comment` to `cactbot-trigger-property-order.js` for both raidboss and oopsy. Can you also add this in `test_trigger.ts` in the `has sorted trigger fields` test?

Souma-Sumire:
Thank you for your guidance. I have made the necessary corrections. 
However, I'm not sure how to clean up these excess commits. Would it be possible for you to squash them during the merge? 
I apologize for any inconvenience caused.

quisquous:
> However, I'm not sure how to clean up these excess commits. Would it be possible for you to squash them during the merge? I apologize for any inconvenience caused.

I will squash it all when I merge. Extra commits are better on a PR for me, because it's easy to see what has changed as I can look at the individual commit rather than having to look at re-squashed PR and re-review the entire thing. This PR is small enough that it's not a big deal, but on very large PRs it is very helpful to have lots of small commits.

quisquous:
Thank you, this is all really great!

PR #5745, opened 2023-08-03
raidboss: Replace player names with job names in popup text

Souma-Sumire:
This is an early implementation, and welcome more ideas and suggestions.

A highly popular feature on the CN server, previously promoted through user-customized JS. Now, I'm attempting to implement it directly in the source code.

Purpose:
1. Cater to different user preferences.
2. Ensure the same mechanics (e.g., tank-buster) have consistent TTS text, reducing token consumption for the TTS plugin. (We prefer using [FoxTTS](https://github.com/Noisyfox/ACT.FoxTTS) plugin to call Microsoft Azure TTS online rather than the system's built-in TTS.)

Potential improvements: Allow strings to be user-modifiable, with default values based on the user's AlertsLanguage. However, I haven't found a way to expose the string configuration in the config panel, as I can't obtain the user's AlertsLanguage before reading the config, resulting in hardcoded text for now.

Preview:
image

with duplicate job
image


quisquous:
Yeah, I know this is a popular suggestion. (I also play in a static, and so names are more useful for me, which is why I never got around to this myself.)

This idea is along the same lines as #5551 which has stalled out a little bit. I think #5551 is a more general solution though and will allow for more user customization. The idea is that you'd be able to not only have a global default (which we could have for job, job+name, name) but for individual triggers it'd be possible to say `${player.job}` or `${player.role}` for that trigger specifically as needed.

PR #5747, opened 2023-08-03
raidboss: add P12S delay info to white flame baits

quisquous:
Followup to #5689.

This also fixes a bug where the delays were incorrect because ongoing add updates caused the trigger to re-fire.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -1917,14 +1917,7 @@ const triggerSet: TriggerSet<Data> = {
         if (o1 === undefined || o2 === undefined || o3 === undefined || o4 === undefined)
           return;
 
-        // 1 = immediate
-        // 2 = fast
-        // 3 = (normal)
-        // 4 = delayed
-        // 5 = very delayed
         data.lcWhiteFlameDelay = [o1 + 1, o2 - o1, o3 - o2, o4 - o3];
-        console.log(`BAIT: ${JSON.stringify([o1, o2, o3, o4])}`);
-        console.log(`DELAY: ${JSON.stringify(data.lcWhiteFlameDelay)}`);
       },
     },
     {
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -167,6 +167,24 @@ const limitCutIds: readonly string[] = Object.keys(limitCutMap);
 const wingIds: readonly string[] = Object.values(wings);
 const superchainNpcBaseIds: readonly string[] = Object.values(superchainNpcBaseIdMap);
 
+const whiteFlameDelayOutputStrings = {
+  delay1: {
+    en: 'now',
+  },
+  delay2: {
+    en: 'soon',
+  },
+  delay3: {
+    en: 'delayed',
+  },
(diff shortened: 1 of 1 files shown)


quisquous:
This needs more testing. I fixed some bugs and looked at the emulator but haven't had a chance to re-run with it.

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -167,6 +167,24 @@ const limitCutIds: readonly string[] = Object.keys(limitCutMap);
 const wingIds: readonly string[] = Object.values(wings);
 const superchainNpcBaseIds: readonly string[] = Object.values(superchainNpcBaseIdMap);
 
+const whiteFlameDelayOutputStrings = {
+  delay1: {
+    en: 'now',
```
YuzukiTsuru, line 172: In some languages such expressions are ambiguous. There are no sufficient and clear words to describe these 5 delayed states. Generally speaking, problems often occur in group [1, 3] and [2, 4] group, which can be written to `wait for [1, 3] bait` when [2, 4] need for the bait after [1, 3] bite.

quisquous: It's ambiguous in English as well, and I don't know that these really differentiate it.  It's more just a very rough indication of how much waiting is required without just saying the numbers 1-5, which I think would be less understandable.

quisquous: I think these are every ~3s apart, so one alternative is just to say 3 seconds, 6 seconds, 9 seconds, etc.

quisquous:
@valarnin @wexxlee do either of you have opinions here on what this can look like? In practice, I found this pretty reasonable. My worry is similar to the above comment about this being too ambiguous (but now/soon/delayed seems mostly fine? maybe there's some other wording; I'm not sure if it can ever be "very delayed" or "verrrry delayed" in practice).

My other worry is also that this is just a LOT on screen at once. It's your number, it's your bait alert it's also the "shotcalling" bait call info. Should I drop the info when the bait is on you? Any other thoughts about how to make this more manageable?

Legends0:
> I'm not sure if it can ever be "very delayed" or "verrrry delayed" in practice).

I have seen "verrry delayed". I don't know about very delayed though. Had to read the text and not just listen to tts since it's similar.

quisquous:
> I have seen "verrry delayed". I don't know about very delayed though. Had to read the text and not just listen to tts since it's similar.

Thank you for giving this a test. I should have cc'd you too!

Do you have any opinions about this being too noisy?

Legends0:
> Do you have any opinions about this being too noisy?
It didn't seem too noisy to me. I find it sometimes useful to know for healing/mit how long to delay a cast or if I need to dodge someone's laser or perhaps take a second laser, but it's minor.

wexxlee:
Sorry for delayed response.  It didn't seem to noisy to me either (and I can confirm (1) verrrry delayed is definitely a real thing, and (2) it was actually helpful info).  

There's probably some options to simplify if you wanted to, but they all have drawbacks.  E.g., we could just tell the user one dash before their bait that it's their turn (and eliminate the "shotcaller" alerts or gate them behind a checkbox), but I'd actually kinda lean against removing current info that some people may still be relying on.  So on balance, I think this is a good change. 👍

PR #5750, opened 2023-08-05
oopsy: add isImmune to OopsyData

g0shu:
I'll look at actually making this work in p12s tomorrow. maybe.

### DIFF
--- a/types/data.d.ts
+++ b/types/data.d.ts
@@ -50,7 +50,7 @@ export interface OopsyData {
   role: Role;
   party: PartyTracker;
   inCombat: boolean;
-  isImmune: (x?: string) => boolean;
+  IsImmune: (x?: string) => boolean;
   ShortName: (x?: string) => string;
   IsPlayerId: (x?: string) => boolean;
   DamageFromMatches: (matches: NetMatches['Ability']) => number;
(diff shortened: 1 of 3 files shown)


types/data.d.ts
```diff
@@ -50,6 +50,7 @@ export interface OopsyData {
   role: Role;
   party: PartyTracker;
   inCombat: boolean;
+  isImmune: (x?: string) => boolean;
```
quisquous, line 53:
```suggestion
  IsImmune: (x?: string) => boolean;
```

I hate that past me made functions have capital letters in this class, but probably this should be consistent with the other functions here.

ui/oopsyraidsy/damage_tracker.ts
```diff
@@ -159,6 +159,12 @@ export class DamageTracker {
       role: this.role,
       party: this.playerStateTracker.partyTracker,
       inCombat: this.inCombat,
+      isImmune: (targetId?: string) => {
```
quisquous, line 162:
```suggestion
      IsImmune: (targetId?: string) => {
```

ui/oopsyraidsy/damage_tracker.ts
```diff
@@ -159,6 +159,12 @@ export class DamageTracker {
...
+        const invulnIds = ['52', '199', '32A', '32B', '72C'];
```
quisquous, line 163: Can you leave a comment as to what these all are? Mostly so that if/when one changes in the future it's easier to come and edit this.

PR #5753, opened 2023-08-06
raidboss: fix JA translation for Baelsar's Wall

stayakira:
I fixed some Japanese translations which are written in Chinese, in both `baelsars_wall.ts` and `baelsars_wall62.ts`.

Translated words (EN -> JA):
* Magitek Bit -> 魔導ビット (for mob name)
* Assault Cannon -> アサルトカノン
* Diffractive Laser -> 拡散レーザー
* Distress Beacon -> 援軍要請
* Dynamic Sensory Jammer -> 動体感知ジャマー
* Magitek Bit -> ビット射出 (for boss's ability)
* Magitek Cannon -> 魔導カノン
* Magitek Claw -> 魔導クロー
* Magitek Missile -> 魔導ミサイル
* Magitek Ray -> 魔導レーザー
* Restraint Collar ->鉄鎖

I also fixed a mistranslation for area name.

Translated words (EN -> JA):
* The Airship Landing -> 飛空戦艦発着場 (previously: `ゴールドソーサー・ランディング`, which means the airship landing too, but only for the Manderville Gold Saucer)


Akurosia:
Magitek Bit in the ReplacementSync section is the enemy not the bosses ability.
The boss ability should be in the replacementText section

stayakira:
Thank you Akurosia for your fast reviewing 😊

I found `Magitek Bit` in 2 sections, `replaceSync` and `replaceText`.
As you mentioned, `Magitek Bit` in `replaceSync` section should be the enemy, that is `魔導ビット` in Japanese.
I believe `Magitek Bit` in `replaceText` section is the bosses' abilities that ejects drones.
Therefore its translation should be `ビット射出` meaning ejecting the magitek bits.
These seem not to need changes.

I would add some addtional commits to remove location names from the files if cactbot does not use location names.
Should the location names removal be a separated PR since this PR is for Japanese translations? (I think so)

Akurosia:
afaik cactbot uses the zone names for the zone seal as you can see in the baelsar_wall.txt
```
0 "Start" sync / 00:0839::Via Praetoria will be sealed off/ window 0,1
1000.0 "Start" sync / 00:0839::The Magitek Installation will be sealed off/ window 1000,5
2000.0 "Start" sync / 00:0839::The Airship Landing will be sealed off/ window 2000,5
```

As i do not speak japanese (only english and german) i do many translations based on the first working hit in both languages.
Zonenames are mostly only included if they are needed (rare exceptions maybe for older content).

stayakira:
Ah, it was my misunderstanding.

I did not know tranlations may be based on pastly translated data.
Also, in this case the location name `The Airship Landing` is needed as baelsar_wall.txt uses...

quisquous:
I'm a little unclear based on the above discussion; is this ready to merge at this point?

Akurosia:
from my side there is nothing open, so i would say i can be merged

quisquous:
@stayakira Thank you!

There are not many folks contributing Japanese translations, as you can see by the really long list of missing translations here (https://quisquous.github.io/cactbot/util/coverage/missing_translations_ja.html) if you're interested in adding any other translations too.

PR #5754, opened 2023-08-06
raidboss: p12s classic2 option to call actual w/no flip

### LINKED ISSUE #5752: P12S Classical Concept 2 without movement
Could an alternate strat option be added to just skip the "initial" call altogether, call only the "actual" positions at the start, and NOT reverse the symbol order from Classical 1?  The mechanic's easier when you just turn around and look at the real pattern.

P12SClassical2

wexxlee:
Closes #5752 -- adds a config checkbox so that the 'Actual' (final) correct position during classical2 will be called out on the classical2 cast, and no follow up alert during Panta Rhei.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -452,7 +452,7 @@ const triggerSet: TriggerSet<Data> = {
       id: 'classicalConcepts2ActualNoFlip',
       comment: {
         en:
-          'Suppresses initial output & calls final position immediately in chosen pair order (e.g. no flip).  <a href="https://user-images.githubusercontent.com/67395342/258580930-1659573f-a492-4839-a275-e51449bb85c0.gif" target="_blank">Visual</a>',
+          'Only calls final position immediately in chosen pair order with no flip. For example, for BPOG, the blue X (crosses) will be far west. <a href="https://quisquous.github.io/cactbot/resources/images/06ew_raid_p12s_classic2_noflip.gif" target="_blank">Visual</a>',
       },
       name: {
         en: 'Classical Concepts 2: Actual only & no inversion',
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -299,6 +299,7 @@ export interface Data extends RaidbossData {
   readonly triggerSetConfig: {
     engravement1DropTower: 'quadrant' | 'clockwise' | 'tower';
     classicalConceptsPairOrder: 'xsct' | 'cxts' | 'ctsx' | 'ctxs' | 'shapeAndDebuff';
+    classicalConcepts2InitialSuppress: true | false;
     pangenesisFirstTower: 'agnostic' | 'not' | 'one';
   };
   decOffset?: number;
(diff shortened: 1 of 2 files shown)


wexxlee:
Please hold off on merge for now - discussion continuing in #5752 about further changes.

wexxlee:
@quisquous updated per discussion in #5752 - this should be ready for review now

ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -447,6 +448,18 @@ const triggerSet: TriggerSet<Data> = {
       },
       default: 'shapeAndDebuff',
     },
+    {
+      id: 'classicalConcepts2ActualNoFlip',
+      comment: {
+        en:
+          'Suppresses initial output & calls final position immediately in chosen pair order (e.g. no flip).  <a href="https://user-images.githubusercontent.com/67395342/258580930-1659573f-a492-4839-a275-e51449bb85c0.gif" target="_blank">Visual</a>',
```
quisquous, line 455:
```suggestion
          'Only calls final position immediately in chosen pair order with no flip.  For example, for BPOG, the crosses will be far west. <a href="https://user-images.githubusercontent.com/67395342/258580930-1659573f-a492-4839-a275-e51449bb85c0.gif" target="_blank">Visual</a>',
```

Also, can you check this image in somewhere so that we're not depending on github hosting some image from a comment and instead we can link to it in github pages?

quisquous (commented):
Looks good in general

ui/raidboss/data/06-ew/raid/p12s.ts
wexxlee: > Also, can you check this image in somewhere so that we're not depending on github hosting some image from a comment and instead we can link to it in github pages?

Sure, makes sense.  When you say check it in somewhere, do you mean like external image hosting svc (e.g. imgur), or do you actually want it checked in to this repo? (e.g. new subdir under resources maybe?)

quisquous: Yeah, I mean in the repo itself so it doesn't get lost.

And yeah, I wasn't sure where a good place was. It's not really (but sort of?) `docs/`, but I think `resources/images/` or something along those lines would be good.

wexxlee: > Yeah, I mean in the repo itself so it doesn't get lost.
> 
> And yeah, I wasn't sure where a good place was. It's not really (but sort of?) `docs/`, but I think `resources/images/` or something along those lines would be good.

I'm having trouble getting the animated gif file added to the repo.  It looks fine on my local fs, but when I commit it to the feature branch and sync, the file gets added to github without any animation (and with the bottom third of the image blank).  I just created a test feature branch and had the same thing happen.  Tried committing with both VS2022 and git bash (in case it was a weird VS bug).  And when I redownload the gif from the github branch, the downloaded file has the same display issue (and, incidentally, is 237 bytes smaller than the original source file).

I'm at a bit of a loss.  Any ideas?  Maybe some sort of line-ending issue since it's being uploaded from a windows machine?

quisquous: Mmm is it being added as text instead of binary? If you add `*.gif binary` to `.gitattributes` and then add the file, does it work ok?

wexxlee: > Mmm is it being added as text instead of binary? If you add `*.gif binary` to `.gitattributes` and then add the file, does it work ok?

Yep, looks like that did the trick.  Thanks!

PR #5755, opened 2023-08-07
raidboss: Add JA translations for DSR Gigaflare

stayakira:
Japanese translations for `directions`, `backLeft`, `backRight`, and `front`.
Arrow symbol `→` at `directions` can be `=>` for consistency, though `→` is more common in Japanese context.

### DIFF
--- a/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
+++ b/ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
@@ -2834,7 +2834,7 @@ const triggerSet: TriggerSet<Data> = {
         directions: {
           en: '${start} => ${rotation}',
           de: '${start} => ${rotation}',
-          ja: '${start} → ${rotation}',
+          ja: '${start} => ${rotation}',
           cn: '${start} => ${rotation}',
           ko: '${start} => ${rotation}',
         },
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/ultimate/dragonsongs_reprise_ultimate.ts
```diff
@@ -2834,24 +2834,28 @@ const triggerSet: TriggerSet<Data> = {
         directions: {
           en: '${start} => ${rotation}',
           de: '${start} => ${rotation}',
+          ja: '${start} → ${rotation}',
```
quisquous, line 2837:
```suggestion
          ja: '${start} => ${rotation}',
```

Mmm, I'm not sure it's worth changing this one instance. `git grep "ja:.*=>" | grep -v "ja: (" | wc -l` suggests there's 180 occurrences of `=>` for Japanese, and I think it'd be better to be consistent than to change this one instance.  (If you think we should change all of them, then maybe that's worth a different PR.)

MaikoTan: In addition, the `=>` is matched in the line below to be replaced as `や` in Japanese. If you're going to change the symbol, not only should you change all the instances for consistency, but it is also required to make the changes to the regex mentioned to make sure it is still functioning.

https://github.com/quisquous/cactbot/blob/31ef788ad883bdfc3108ee4332e99f526a63c9ed/ui/raidboss/popup-text.ts#L1472-L1474

stayakira:
Thank you for commenting.

I did not know the arrow symbol is used by regex for tts.
Commiting @quisquous 's suggestion.

MaikoTan (approved):
LGTM!

Issue #5619, opened 2023-06-26
Timeline improvements discussion


quisquous:
I'm breaking this off from #5577.

Restating the discussion here:

@valarnin [comment link](https://github.com/quisquous/cactbot/pull/5577#issuecomment-1596422841)
> Is this necessary before we end up reworking the format and therefore the underlying classes? Seems like potential breakage mid-raid-tier that we'll end up throwing out anyways later isn't really great? I'd like to get transitioned to a new timeline format before 7.0...

@quisquous [comment link](https://github.com/quisquous/cactbot/pull/5577#issuecomment-1596435795)
> > Is this necessary before we end up reworking the format and therefore the underlying classes?
> Uh, when and who is reworking this format? I haven't seen any momentum or discussion on this front, and personally haven't heard enough pain points to understand the motivation here.
> 
> EDIT: Sorry for the surprise on my part, but I was caught off guard with this sentiment.
> 
> The only discussion on my radar is this discord discussion: https://discord.com/channels/551474815727304704/594899820976668673/1032391897056022581. It included some discussion that locrian working on a new timeline format for lemegeton (which I asked to see, but didn't get shared; I think I saw it it once elsewhere and if that's what it is it's very XML/id heavy and hard to read; locrian would probably say "you're not supposed to read the raw format" but I think we continue to differ on the value of that). I didn't come away from that discussion thinking that "somebody will make a new timeline format and convert all timelines and tooling before 7.0".
> 
> In that same discord conversation you say "The timeline as it exists now is very, very fixed and inflexible. It would probably benefit from a significant rewrite", but it's not clear to me what your pain points are such that there would be benefits from a rewrite. (My complaints about the timeline code are largely about the the timeline controller and not the timeline format or parser.) What's on your mind, specifically?
> 
> Here are various issues I know about with timelines (in order of my own priority?):
> 
> * making loops is tedious, and tedious to review (this issue)
> * jumps to numbers is fragile (we could probably add named labels)
> * rather than commenting out syncs, it'd be nice to add `testsync` that test_timeline verifies syncs at the right time but doesn't sync or adjust the time at runtime
> * (inserting some space here between the above things that I think are higher priority and the below things which I think are even lower priority / could sit forever on the backburner)
[...]

PR #5760, opened 2023-08-10
raidboss: add named timeline jump labels

quisquous:
Mentioned in #5619.

### DIFF
--- a/docs/TimelineGuide.md
+++ b/docs/TimelineGuide.md
@@ -76,13 +76,18 @@ Every timeline entry begins with the ability time and the ability name.
 
 `Number "String" (duration Number)`
 
-`Number "String" sync /Regex/ (window Number,Number) (jump Number) (duration Number)`
+`Number "String" sync /Regex/ (window Number,Number) (jump NumberOrLabel) (duration Number)`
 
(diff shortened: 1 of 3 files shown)


quisquous:
A proof of concept. I think it's kind of nice to have labels on their own lines to make them more visually obvious even if it means repeating a timestamp.

docs/TimelineGuide.md
```diff
@@ -76,14 +76,19 @@ Every timeline entry begins with the ability time and the ability name.
...
+`Number "String" sync /Regex/ (window Number,Number) (forcejump NumberOrLabel) (duration Number)`
+
+`Number label "String"`
 
 (The parentheses here indicate optionality and are not literal parentheses.)
 
 **Number** can be an integer, e.g. `34`, or a float, e.g. `84.381`.
 
+**NumberOrLabel** can be a **Number** (e.g. `42` or `12.8`)
+or a label name with quotes (e.g. `"loop"` or `"branch2"`).
```
JLGarber, line 90:
```suggestion
or a label name with double quotes (e.g. `"loop"` or `"branch2"`).
```

JLGarber:
I like this. A significant benefit I can see right away is that named jumps don't need to be updated alongside timeline adjustments. That will be *really* nice.

PR #5766, opened 2023-08-15
plugin: update DEPS.json5

Akurosia:
After updateing DEPS.json5, i got following compilation error: 'CactbotEventSource.cs(46,13): error CS0104: "GameRegion" ist ein mehrdeutiger Verweis zwischen "Cactbot.VersionChecker.GameRegion" und "RainbowMage.OverlayPlugin.GameRegion"' thats why i touched CactbotEventSource.cs

### DIFF
--- a/util/DEPS.json5
+++ b/util/DEPS.json5
@@ -9,7 +9,7 @@
         "hash": ["sha256", "8822e08147a44d0b82548068033dc18177c1fac993e8564e1ac32ed94c570e08"],
     },
     "FFXIV_ACT": {
-        "url": "https://github.com/ravahn/FFXIV_ACT_Plugin/raw/master/Releases/FFXIV_ACT_Plugin_SDK_2.6.9.3.zip",
+        "url": "https://github.com/ravahn/FFXIV_ACT_Plugin/raw/master/Releases/FFXIV_ACT_Plugin_SDK_2.4.2.1.zip",
         "dest": "plugin/ThirdParty/FFXIV_ACT",
         "strip": 0,
         "hash": ["sha256", "76fa369d0c1c6f9ab8afa01bf657a7dcf611324f15cf380fbe6a4dc0e150bcc5"],
--- a/util/DEPS.json5
+++ b/util/DEPS.json5
@@ -3,21 +3,21 @@
 //   I just borrowed the basic concept and naming scheme.
 {
     "ACT": {
-        "url": "https://github.com/EQAditu/AdvancedCombatTracker/releases/download/3.4.6.267/ACTv3.zip",
+        "url": "https://github.com/EQAditu/AdvancedCombatTracker/releases/download/3.6.2.279/ACTv3.zip",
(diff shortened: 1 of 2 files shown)


util/DEPS.json5
```diff
@@ -3,21 +3,21 @@
...
-        "url": "https://github.com/ravahn/FFXIV_ACT_Plugin/raw/master/Releases/FFXIV_ACT_Plugin_SDK_2.0.6.1.zip",
+        "url": "https://github.com/ravahn/FFXIV_ACT_Plugin/raw/master/Releases/FFXIV_ACT_Plugin_SDK_2.6.9.3.zip",
         "dest": "plugin/ThirdParty/FFXIV_ACT",
         "strip": 0,
-        "hash": ["sha256", "ed98fa01ec2c7ed2ac843fa8ea2eec1d66f86fad4c6864de4d73d5cfbf163dce"],
+        "hash": ["sha256", "76fa369d0c1c6f9ab8afa01bf657a7dcf611324f15cf380fbe6a4dc0e150bcc5"],
     },
     "OverlayPlugin": {
-        "url": "https://github.com/ngld/OverlayPlugin/releases/download/v0.18.2/OverlayPlugin-0.18.2.zip",
+        "url": "https://github.com/OverlayPlugin/OverlayPlugin/releases/download/v0.19.21/OverlayPlugin-0.19.21.zip",
```
quisquous, line 18: I think it makes sense to update OverlayPlugin here, as users will always be on the latest.

However, I think it would be better to leave the FFXIV_ACT version, I think(?) it's an issue if the cactbot is built against 2.6.9.3 but Korean users try to use it with version 2.4.2.1 or whatever it's on. If you want to update it to whatever the Korean version is on, that's fine.

util/DEPS.json5
```diff
@@ -9,7 +9,7 @@
         "hash": ["sha256", "8822e08147a44d0b82548068033dc18177c1fac993e8564e1ac32ed94c570e08"],
     },
     "FFXIV_ACT": {
-        "url": "https://github.com/ravahn/FFXIV_ACT_Plugin/raw/master/Releases/FFXIV_ACT_Plugin_SDK_2.6.9.3.zip",
+        "url": "https://github.com/ravahn/FFXIV_ACT_Plugin/raw/master/Releases/FFXIV_ACT_Plugin_SDK_2.4.2.1.zip",
         "dest": "plugin/ThirdParty/FFXIV_ACT",
         "strip": 0,
         "hash": ["sha256", "76fa369d0c1c6f9ab8afa01bf657a7dcf611324f15cf380fbe6a4dc0e150bcc5"],
```
quisquous, line 15: Do you need to change the hash here too?

Akurosia: ohh, yeah rigth x.x will commit it quickly

Akurosia: good thing you are here to double check, thought ko version has a corresponding sdk version, but sdk jumps from 2.2x to 2.6x so i used the latest working for both

Issue #5764, opened 2023-08-14
Cactbot cannot be built in current 18.x NodeJS Version


jpdied:
If you follow the current build instructions (the ones to build timelines, not the actual plugin), the build process faills and it looks like its due to recent newer versions of NodeJS.

Node.js v18.17.1
```
$ git clone git@github.com:quisquous/cactbot.git
Cloning into 'cactbot'...
remote: Enumerating objects: 83196, done.
remote: Counting objects: 100% (1014/1014), done.
remote: Compressing objects: 100% (407/407), done.
remote: Total 83196 (delta 677), reused 884 (delta 601), pack-reused 82182
Receiving objects: 100% (83196/83196), 104.10 MiB | 3.83 MiB/s, done.
Resolving deltas: 100% (62511/62511), done.

$ cd cactbot/

$ npm install

$ npm test

> cactbot@0.29.22 test
> mocha

(node:41700) ExperimentalWarning: Custom ESM Loaders is an experimental feature and might change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
N:\git\cactbot\node_modules\ttypescript\lib\loadTypescript.js:13
        function __() { this.constructor = d; }
                                         ^

TypeError: Cannot set property constructor of [object Object] which has only a getter
    at new __ (N:\git\cactbot\node_modules\ttypescript\lib\loadTypescript.js:13:42)
    at __extends (N:\git\cactbot\node_modules\ttypescript\lib\loadTypescript.js:14:84)
    at N:\git\cactbot\node_modules\ttypescript\lib\loadTypescript.js:65:5
    at Object.<anonymous> (N:\git\cactbot\node_modules\ttypescript\lib\loadTypescript.js:100:2)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10)
    at Module.load (node:internal/modules/cjs/loader:1119:32)
    at Module._load (node:internal/modules/cjs/loader:960:12)
    at Module.require (node:internal/modules/cjs/loader:1143:19)
    at require (node:internal/modules/cjs/helpers:121:18)

Node.js v18.17.1
```

When performing the actions on a clean git install using version 18.12.0, things run correctly.

[...]

PR #5767, opened 2023-08-16
build/docs: Updated ttypescript version

Symulation12:
Tested with a fresh environment on node 18.17.1. As was mentioned in #5764 , I also added version numbers to the recommended global packages in CONTRIBUTING (the same ones as in package.json), and version numbers to the README for nodejs and npm (current LTS).  
Apologies if I used the wrong label for this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,8 +79,8 @@ in order to install local versions of husky, lint-staged, and eslint.
 It is recommended that you also install ts-node and typescript globally, e.g.
 
 ```shell
-npm install -g typescript@4.6.4
-npm install -g ts-node@10.8.0
+npm install -g typescript
+npm install -g ts-node
 ```
 
 This will let you run commands as `ts-node` directly instead of `npx ts-node`.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -79,8 +79,8 @@ in order to install local versions of husky, lint-staged, and eslint.
 It is recommended that you also install ts-node and typescript globally, e.g.
 
 ```shell
-npm install -g typescript
-npm install -g ts-node
+npm install -g typescript@4.6.4
+npm install -g ts-node@10.8.0
 ```
 
 This will let you run commands as `ts-node` directly instead of `npx ts-node`.
--- a/README.md
+++ b/README.md
@@ -246,8 +246,9 @@ To install npm and start Webpack, follow these steps:
 1. Run `npm install` in the root of the cactbot directory.
 1. Run `npm run build` or `npm start`.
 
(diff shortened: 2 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -79,8 +79,8 @@ in order to install local versions of husky, lint-staged, and eslint.
 It is recommended that you also install ts-node and typescript globally, e.g.
 
 ```shell
-npm install -g typescript
-npm install -g ts-node
+npm install -g typescript@4.6.4
+npm install -g ts-node@10.8.0
```
quisquous, line 83: One worry I have here is these getting out of date. It's going to be extremely likely that somebody will forget to bump this documentation in the future. This seems like the same problem, but in reverse. People will install this old version and then it won't work with whatever the new version needs.

I think this is more likely of a problem than the one you ran into, so I'd prefer not to change the versions.

(Happy to discuss more too, or maybe there's some solution to keep this doc up to date or a command that installs the package.json version globally that I don't know about?)

README.md
```diff
@@ -242,7 +242,7 @@ instead of changing your local cactbot files.
 
 To install npm and start Webpack, follow these steps:
 
-1. Install [nodejs and npm](https://nodejs.org/en/download/)
+1. Install [nodejs 18.17.1 and npm 9.6.7](https://nodejs.org/en/download/)
```
quisquous, line 245:
```suggestion
1. Install [nodejs 18.17.1 and npm 9.6.7](https://nodejs.org/en/download/)
```

This feels like the same issue here as above. Re: nodejs, I feel like cactbot should always work against the LTS version of nodejs and if it doesn't then it's a bug to be fixed. As soon as 18.17.1 is not the most current LTS version, then it's going to be much harder to find as well and folks will have to dig around in versions to get to it.

I'd rather make a note that says "cactbot should always work with the latest LTS and file an issue if it doesn't".

quisquous:
Also also, thanks for testing this and the update! Glad this was an easy fix.

README.md
Symulation12: Fair point, that is definitely a more sensible solution.

CONTRIBUTING.md
Symulation12: Ah, true, that is something I should have considered. I looked around to see if I could find a better solution, but I couldn't find anything cleaner than simply listing the most recent version. Putting it that way, I definitely agree that leaving it out is the better option for now.

Symulation12:
Of course! 
I removed the version numbers from contributing, and changed the readme to not reference a specific version but rather LTS as a whole. Hopefully that's a bit more reasonable.

quisquous:
Thank you!!

PR #5768, opened 2023-08-17
i18n: p12s cn translations for white flame baits delay

lsl1225:


lsl1225:
It is very tricky to translate those 'time state' words, so that I think these translated words need to be reviewed. Also 3 secs, 6 secs and 9 secs can make me confused in game so thats not a very good option?

lsl1225:
@YuzukiTsuru any opinions? I have tried other translations like `等待上一组 => 引导激光` (wait for last bait => bait laser in English), but it makes tts too long in this fast-paced mechanic and may not distinguish those 5 states as well.

lsl1225:
Just updated the translations, I think it's better and more obvious to differentiate between those words now.

quisquous:
Sorry that this was tricky to translate. Even in English it's kind of ambiguous, but it's still helpful information too so I didn't want to not include it. Thanks!!

lsl1225:
> Sorry that this was tricky to translate. Even in English it's kind of ambiguous, but it's still helpful information too so I didn't want to not include it. Thanks!!

It's okay! This information also helps me a lot during my game play. Difficulty in translating this is how to let people know what those words are talking about. Thanks for the reply btw!

PR #5772, opened 2023-08-26
raidboss: fix classical1 bug with noflip checkbox

### LINKED ISSUE #5771: PR #5754 also flips Classical 1 column order
Thanks for the work implementing the new Classical Concepts option!  I'm glad to report it works perfectly for Classical 2, however, having the checkbox enabled _also_ reverses the column order for Classical 1.  When I tested it with BPOG order, I was given the instructions for G when I was B, or O when I was P, and so on.  This only occurs during Classical 1.  Classical 2 works as intended.

Demonstration: https://youtu.be/wNmS1kU8OLY

_Originally posted by @DreyfussFrost in https://github.com/quisquous/cactbot/issues/5752#issuecomment-1694020061_

wexxlee:
Closes #5771.

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -3085,11 +3085,11 @@ const triggerSet: TriggerSet<Data> = {
           if (columnOrder?.length !== 4)
             return;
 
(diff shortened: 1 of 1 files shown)


quisquous:
Sorry for not catching this earlier either.

PR #5777, opened 2023-09-03
raidboss: update cn translation

Souma-Sumire:
The dir of cn's outputs already includes the side field.

before:
image

now:
image


Souma-Sumire:
Oops, did not pass the format test. This is difficult to handle.

quisquous:
Ah, I see. The issue is that Left and West are the same thing, so it looks redundant?

I think if you want to do this, maybe we need a "Left/West" and "Right/East" output string, and then use that instead of trying to combine it, if that makes sense.

Souma-Sumire:
How about this now

PR #5778, opened 2023-09-03
oopsy: fix P12N palladion spread id

g0shu:
`82C3` is Trinity of Souls (line 14)


g0shu:
as a thought, should there be a test checking if the same ability is being warned about from 2 triggers? because something like is probably never an intended behavior
image

quisquous:
> as a thought, should there be a test checking if the same ability is being warned about from 2 triggers? because something like is probably never an intended behavior

Mmmmm yeah, that's a good thought.

PR #5780, opened 2023-09-06
raidboss: uwu updates

quisquous:
* config options for custom jail order (by name or job)
* garuda sisters locations
* nail spawn locations
* ifrit charge directions (closes #3715)
* predation safe cardinal and rotation
* improved timeline

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -59,31 +59,31 @@ export default class PartyTracker {
   }
 
   // returns an array of the names of players in your immediate party
-  get partyNames(): string[] {
+  get partyNames(): readonly string[] {
     return this.partyNames_;
   }
 
-  get partyIds(): string[] {
+  get partyIds(): readonly string[] {
     return this.partyIds_;
   }
 
   // returns an array of the names of players in your alliance
-  get allianceNames(): string[] {
+  get allianceNames(): readonly string[] {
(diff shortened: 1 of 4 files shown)


Souma-Sumire:
I'm delighted to see that cactbot has started supporting custom sorting for uwu Titan Gaols！ 

Have you considered implementing sorting by job (e.g., WAR > WHM > MNK)? 

In my humble opinion, managing it by ID is not only cumbersome (requires configuration adjustments every time a party member changes) but also prone to input errors (and small errors are hard to notice). 

(Translated via software, please forgive any impoliteness.)

quisquous:
> I'm delighted to see that cactbot has started supporting custom sorting for uwu Titan Gaols！
> 
> Have you considered implementing sorting by job (e.g., WAR > WHM > MNK)? [24H paste sharing](http://pastie.org/p/3XHGCYlPFXTZNoNnx0OPBE)

Mmm, ok I can do that too. I like names because some people switch jobs around but the order should still be the same. I've implemented it so that you can write either names or jobs.

The one downside with jobs is that https://github.com/OverlayPlugin/OverlayPlugin/issues/60 is still a bug, so in some cases the party info might not be there.

quisquous:
Let me know if that's not sufficient for what you need and I'll give it another update.

PR #5789, opened 2023-09-08
raidboss: text in alerts lang, not display lang

Souma-Sumire:
before:
image

now:
image

### DIFF
--- a/ui/raidboss/popup-text.ts
+++ b/ui/raidboss/popup-text.ts
@@ -921,7 +921,7 @@ export class PopupText {
 
     trigger.output = TriggerOutputProxy.makeOutput(
       trigger,
-      this.options.DisplayLanguage,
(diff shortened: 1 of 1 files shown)


Souma-Sumire:
I'm not quite sure if the modification here is non-standard; please advise.

Souma-Sumire:
This can also occur in actual battles.

============================
image
image


============================
image
image

ui/raidboss/popup-text.ts
```diff
@@ -921,7 +921,7 @@ export class PopupText {
 
     trigger.output = TriggerOutputProxy.makeOutput(
       trigger,
-      this.options.DisplayLanguage,
+      this.options.AlertsLanguage ?? this.options.DisplayLanguage,
```
quisquous, line 924:
```suggestion
      this.displayLang,
```

I think this is meant to be the above, which includes even more fallback options.

PR #5792, opened 2023-09-10
resources: file size reduction

Souma-Sumire:
Size Change: 8.75 MB => 2.17 MB

There is a slight, acceptable loss in quality, but the size has significantly reduced, and it displays correctly in the new 'open' window of OverlayPlugin.

I also tried higher compression codecs, but they didn't display correctly in OverlayPlugin.


quisquous:
Thanks!

PR #5794, opened 2023-09-14
raidboss: prevent duplicate button creation

Souma-Sumire:
created only once, rather than multiple times.

### DIFF
--- a/ui/raidboss/autoplay_helper.ts
+++ b/ui/raidboss/autoplay_helper.ts
@@ -3,9 +3,7 @@ export default class AutoplayHelper {
   private static isButtonCreated = false;
 
   private static getContext(): AudioContext {
-    if (!AutoplayHelper.context) {
-      AutoplayHelper.context = new AudioContext();
-    }
+    AutoplayHelper.context ??= new AudioContext(); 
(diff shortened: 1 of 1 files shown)


ui/raidboss/autoplay_helper.ts
```diff
@@ -1,16 +1,29 @@
 export default class AutoplayHelper {
+  private static context: AudioContext | null = null;
```
quisquous, line 2:
```suggestion
  private static context?: AudioContext;
```

Just preferring `undefined` over null here.

ui/raidboss/autoplay_helper.ts
```diff
@@ -1,16 +1,29 @@
...
+  private static isButtonCreated = false;
+
+  private static getContext(): AudioContext {
+    if (!AutoplayHelper.context) {
+      AutoplayHelper.context = new AudioContext();
+    }
+    return AutoplayHelper.context;
```
quisquous, line 9:
```suggestion
    AutoplayHelper.context ??= new AudioContext(); 
    return AutoplayHelper.context;
```

Just simplifying.

PR #5795, opened 2023-09-15
raidboss: add p11s dike

Souma-Sumire:


ui/raidboss/data/06-ew/raid/p11s.ts
```diff
@@ -91,6 +93,32 @@ const triggerSet: TriggerSet<Data> = {
...
     },
+    {
+      id: 'P11S Dike',
+      type: 'StartsUsing',
+      netRegex: { id: ['8230', '822F'], capture: true },
+      preRun: (data, matches) => data.busterTargets.push(matches.target),
+      response: (data, _matches, output) => {
+        // cactbot-builtin-response
+        output.responseOutputStrings = {
+          busterOnYou: Outputs.tankBusterOnYou,
```
quisquous, line 104: Hmm. Should this say something different, as it's 2x tank busters?

I think most of the time the first two get invuln'd and the third one gets skipped by dps, but it is still a tank swap too?

Mostly just not sure what this should say that conveys "2x tankbuster, maybe you're invulning but technically it's a swap" in a concise way.

quisquous: e12s says "Tank Buster + Swap" which maybe we could reuse here?

PR #5801, opened 2023-09-20
i18n: more FR translations

mooondark:


quisquous:
if you feel up for it, https://quisquous.github.io/cactbot/util/coverage/missing_translations_fr.html still has a lot missing too

PR #5807, opened 2023-09-23
raidboss: p12s Paradeigma 2 Tower Strategy add 'tetherbase'

YuzukiTsuru:
Added Paradeigma 2 Tower Strategy tetherbase, which is commonly used by Chinese players.

- [【【菓子】FF14万魔殿天狱篇零式4层 P12S超详解攻略】 ](https://www.bilibili.com/video/BV1vF411S7oL)

Solution idea:
1. First look at the type of Anthropos generated on the sidelines, and the corresponding position of the Tether player
2. Determine the debuff type that you have
3. drop the tower according to the priority TH: A->B->C DPS: D->C->B

This solution avoids seeing the wrong type because the Tether player is not in place in time

### DIFF
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -426,7 +426,7 @@ const triggerSet: TriggerSet<Data> = {
           '方針なし': 'tower',
         },
         cn: {
-          '看小怪位置(菓子)': 'tetherbase',
+          '看小怪位置(果子)': 'tetherbase',
           '垂直拉线 (Game8)': 'quadrant',
           '对角拉线': 'clockwise',
           '仅提示塔颜色': 'tower',
@@ -1272,7 +1272,11 @@ const triggerSet: TriggerSet<Data> = {
 
           // Do the sort by role
           if (data.role === 'dps') {
-            if (tower0.clone === 'SSW' || tower0.clone === 'SSE') {
+            if (tower0.clone === 'WSW' || tower0.clone === 'WNW') {
+              towerResult = tower0.location;
+            } else if (tower1.clone === 'WSW' || tower1.clone === 'WNW') {
+              towerResult = tower1.location;
+            } else if (tower0.clone === 'SSW' || tower0.clone === 'SSE') {
               towerResult = tower0.location;
             } else if (tower1.clone === 'SSW' || tower1.clone === 'SSE') {
               towerResult = tower1.location;
@@ -1282,7 +1286,11 @@ const triggerSet: TriggerSet<Data> = {
               towerResult = tower1.location;
             }
           } else {
-            if (tower0.clone === 'ENE' || tower0.clone === 'ESE') {
+            if (tower0.clone === 'NNW' || tower0.clone === 'NNE') {
+              towerResult = tower0.location;
+            } else if (tower1.clone === 'NNW' || tower1.clone === 'NNE') {
+              towerResult = tower1.location;
+            } else if (tower0.clone === 'ENE' || tower0.clone === 'ESE') {
               towerResult = tower0.location;
             } else if (tower1.clone === 'ENE' || tower1.clone === 'ESE') {
               towerResult = tower1.location;
@@ -1292,23 +1300,6 @@ const triggerSet: TriggerSet<Data> = {
               towerResult = tower1.location;
             }
           }
-
-          // Check role Only tower clone location
-          towerList.forEach((tower) => {
-            if (data.role === 'dps') {
-              // handle DPS only WSW and WNW clone
-              if (tower.clone === 'WSW' || tower.clone === 'WNW') {
-                towerResult = tower.location;
-                return;
-              }
-            } else {
-              // handle TH only NNW and NNE clone
-              if (tower.clone === 'NNW' || tower.clone === 'NNE') {
-                towerResult = tower.location;
-                return;
-              }
-            }
-          });
 
           if (towerResult === undefined)
             return;
--- a/ui/raidboss/data/06-ew/raid/p12s.ts
+++ b/ui/raidboss/data/06-ew/raid/p12s.ts
@@ -327,7 +327,7 @@ export type LimitCutCombatantState = PluginCombatantState & {
 
 export interface Data extends RaidbossData {
   readonly triggerSetConfig: {
-    engravement1DropTower: 'quadrant' | 'clockwise' | 'tower';
+    engravement1DropTower: 'quadrant' | 'clockwise' | 'tower' | 'tetherbase';
     classicalConceptsPairOrder: 'xsct' | 'cxts' | 'ctsx' | 'ctxs' | 'shapeAndDebuff';
     classicalConcepts2ActualNoFlip: true | false;
     pangenesisFirstTower: 'agnostic' | 'not' | 'one';
@@ -404,12 +404,13 @@ const triggerSet: TriggerSet<Data> = {
         en: 'Paradeigma 2 Tower Strategy',
         de: 'Paradigma 2 Türme Strategy',
         ja: 'パラデイグマ2の塔処理方法',
-        cn: '第一次拉线踩塔方法',
+        cn: '范式 2 踩塔方法',
         ko: 'Paradeigma 2 기둥 공략',
       },
       type: 'select',
       options: {
         en: {
+          'Based on Tether': 'tetherbase',
           'Tether direct across + nearest quadrant tower (Game8)': 'quadrant',
           'Clockwise tower from tether': 'clockwise',
           'No strategy: just call tower color': 'tower',
@@ -425,6 +426,7 @@ const triggerSet: TriggerSet<Data> = {
           '方針なし': 'tower',
         },
         cn: {
+          '看小怪位置(果子)': 'tetherbase',
           '垂直拉线 (Game8)': 'quadrant',
           '对角拉线': 'clockwise',
           '仅提示塔颜色': 'tower',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/raid/p12s.ts
```diff
@@ -425,6 +426,7 @@ const triggerSet: TriggerSet<Data> = {
           '方針なし': 'tower',
         },
         cn: {
+          '看小怪位置(果子)': 'tetherbase',
```
superjump22, line 429: 是菓子不是果子

quisquous:
I watched the video and my understanding is that this looks the same as the `quadrant` strategy, in terms of where towers go relative to tethers. Am I understanding this correctly?

Re: A B C D. In general, I try not to put marker names in cactbot because people put markers in different places and the calls should still work without markers. If folks want to have marker locations, they can edit the [cactbot config ui](https://github.com/quisquous/cactbot/blob/main/docs/CactbotCustomization.md#changing-trigger-text-with-the-cactbot-ui) to say A or 1 instead of northeast. For consistency, I think it's better to just say directions.

If you want to add single tower locations, I think it would be better to implement this as an additional config option so that it would work with both quadrant and clockwise strategies. Rather than embedding "A+NW", it would use the previous "NW" logic, but could find the set of positions at the end and pull out the correct one for the role. This option would be a choice between "default (call out two locations)" and "support NE+clockwise". Does that make sense?

YuzukiTsuru:
The `quadrant`  is bassed on the tether player, but the tetherbase is based on the `Anthropos`, I'm thinking of solving this problem in a way that's closer to the actual processing logic.

A B C D it is for internal use only and will not be displayed. Since this strategy is strongly based on waymark, I think it's easier to understand using ABCD here (in code).

The A+NW method is used here to distinguish two different NW. According to the original `quadrant` method, only the target can be dropped on NW, and the actual position of Anthropos is not known (and it is not necessary to know). However, this strategy is based on the location of Anthropos, so it is necessary to determine the actual location of Anthropos. And for this strategy only.

quisquous:
> The quadrant is bassed on the tether player, but the tetherbase is based on the Anthropos, I'm thinking of solving this problem in a way that's closer to the actual processing logic.

image

I am not sure I understand what you are saying. This appears to be the same logic as `quadrant`, in that if there is a NNE light add, then there needs to be a dark SW tower and if there is a NNW light add then there needs to be a dark SE tower instead. This seems to be the same strategy.

> A B C D it is for internal use only and will not be displayed. Since this strategy is strongly based on waymark, I think it's easier to understand using ABCD here (in code).

Oh, I see. A B C D is just a way to do priority. I think it could be done in a simpler manner if that's true. I was trying to propose a second separate config option to do the priority ordering, and do the sorting after the tower locations had been found. Something like this:

```typescript
          const engravement1Keys: string[] = [];
          data.engravement1BeamsPosMap.forEach((value: string, key: string) => {
            if (value === 'light' && matches.effectId === engravementIdMap.lightTower)
              engravement1Keys.push(key);
            else if (value === 'dark' && matches.effectId === engravementIdMap.darkTower)
              engravement1Keys.push(key);
          });

          const outputMap: { [dir: string]: string } = {
            NW: output.northwest!(),
            NE: output.northeast!(),
            SE: output.southeast!(),
            SW: output.southwest!(),
          } as const;

          const [key0, key1] = engravement1Keys;
          if (key0 === undefined || key1 === undefined)
            return;

          // This is a new trigger config that would be need to be added.
          if (data.triggerSetConfig.engravement1SortOrder === 'SupportNEClockwise') {
            // By doing it this way, we could support multiple sort orders easily in the future.
            // Also having a config option lets this apply to both quadrant and clockwise strategies,
            // and doesn't intermix "priority" with "tower location" logic.
            const supportOrder = ['NE', 'SE', 'SW', 'NW'];
            const idx0 = supportOrder.indexOf(key0);
            const idx1 = supportOrder.indexOf(key1);
            const supportPreferKey0 = idx0 < idx1;
            const isSupport = data.role !== 'dps';
            const useKey0 = isSupport && supportPreferKey0 || !isSupport && !supportPreferKey0;
            const myKey = useKey0 ? key0 : key1;
            if (matches.effectId === engravementIdMap.lightTower)
              return output.lightTowerOneSide!({ pos1: outputMap[myKey] });
            return output.darkTowerOneSide!({ pos1: outputMap[myKey] });

          } else if (data.triggerSetConfig.engravement1SortOrder === 'default') {
            if (matches.effectId === engravementIdMap.lightTower) {
              return output.lightTowerSide!({
                pos1: outputMap[key0],
                pos2: outputMap[key1],
              });
            }

            return output.darkTowerSide!({
              pos1: outputMap[key0],
              pos2: outputMap[key1],
            });
          }
```

Souma-Sumire:
I think we cannot sort by tower positions because a dark SW tower could be either an NNW light add or an EES light add.

I gave two examples to demonstrate the results after sorting by the source, and you can see that if we use tower sorting, the order of these two results is exactly opposite.

image
image

YuzukiTsuru:
It seems impossible to simply sort the position of the tower, previously thought that a new collector could be added here to collect functions and tower colors, but it is possible to modify the current matching conditions.

quisquous:
Ohhhhhhhhh I see, the sorting is based on where the add is and not where the location is. Wild. (It's still sad to me that clockwise isn't a more popular strategy because it has the same advantage as looking at the tether base but you also don't need to look at which side of the line the add is on, only the tether color. Anyway~~~.)

Ok, then I'd rather still do something like this, where you annotate the beam data with the location explicitly, rather than having one of the paths have different values in the map that have to be processed. Sorry to be picky, I just think this is getting rather complicated (especially if somebody ever wants to add a fourth one or different sorting), and so I want to try to keep it easier to follow along.

What about something like this code. It keeps the sorting at the end and doesn't duplicate the tower location finding. It also only finds the towers that are valid for the current player rather than finding them all, which simplifies a lot.

```typescript
      alertText: (data, matches, output) => {
        data.engravement1Towers.push(matches.target);

        if (data.me !== matches.target)
          return;

        // if Only notify tower color
        if (data.triggerSetConfig.engravement1DropTower === 'tower') {
          if (matches.effectId === engravementIdMap.lightTower)
            return output.lightTower!();
          return output.darkTower!();
        }

        const locations = ['NE', 'SW', 'SE', 'NW'] as const;
        type TowerLocation = typeof locations[number];
        type CloneLocation = 'NNW' | 'NNE' | 'ENE' | 'ESE' | 'SSE' | 'SSW' | 'WNW' | 'WSW';
        type TowerData = {
          location: TowerLocation;
          clone: CloneLocation;
        };
        const towerList: TowerData[] = [];
        const towerStrategy = data.triggerSetConfig.engravement1DropTower;

        for (const combatant of data.combatantData) {
          const x = combatant.PosX;
          const y = combatant.PosY;

          const combatantId = combatant.ID;
          if (combatantId === undefined)
            return;

          const tempColor = data.engravement1TetherPlayers[combatantId.toString(16).toUpperCase()];
          const color = tempColor === 'light' ? 'dark' : 'light';

          const isCorrectColor = color === 'light' && matches.effectId === engravementIdMap.lightTower ||
            color === 'dark' && matches.effectId === engravementIdMap.darkTower;
          if (!isCorrectColor)
            continue;

          if (towerStrategy === 'quadrant' || towerStrategy === 'tetherbase') {
            if (x < 80 && y < 100) { // WNW: x = 75 && y = 97
              towerList.push({ location: 'NE', clone: 'WNW' });
            } else if (x < 100 && y < 80) { // NNW: x = 97 && y = 75
              towerList.push({ location: 'SW', clone: 'NNW' });
            } else if (x > 100 && y < 80) { // NNE: x = 103 && y = 75
              towerList.push({ location: 'SE', clone: 'NNE' });
            } else if (x > 120 && y < 100) { // ENE: x = 125 && y = 97
              towerList.push({ location: 'NW', clone: 'ENE' });
            } else if (x > 120 && y > 100) { // ESE: x = 125 && y = 103
              towerList.push({ location: 'SW', clone: 'ESE' });
            } else if (x > 100 && y > 120) { // SSE: x = 103 && y = 125
              towerList.push({ location: 'NE', clone: 'SSE' });
            } else if (x < 100 && y > 120) { // SSW: x = 97 && y = 125
              towerList.push({ location: 'NW', clone: 'SSW' });
            } else if (x < 80 && y > 100) { // WSW: x = 75 && y = 103
              towerList.push({ location: 'SE', clone: 'WSW' });
            }
          } else if (data.triggerSetConfig.engravement1DropTower === 'clockwise') {
            // Tether stretches across and tower is clockwise; e.g. N add stretches S, and tower is SW.
            if (x < 80 && y < 100) { // WNW: x = 75 && y = 97
              towerList.push({ location: 'SE', clone: 'WNW' });
            } else if (x < 100 && y < 80) { // NNW: x = 97 && y = 75
              towerList.push({ location: 'SW', clone: 'NNW' });
            } else if (x > 100 && y < 80) { // NNE: x = 103 && y = 75
              towerList.push({ location: 'SW', clone: 'NNE' });
            } else if (x > 120 && y < 100) { // ENE: x = 125 && y = 97
              towerList.push({ location: 'NW', clone: 'ENE' });
            } else if (x > 120 && y > 100) { // ESE: x = 125 && y = 103
              towerList.push({ location: 'NW', clone: 'ESE' });
            } else if (x > 100 && y > 120) { // SSE: x = 103 && y = 125
              towerList.push({ location: 'NE', clone: 'SSE' });
            } else if (x < 100 && y > 120) { // SSW: x = 97 && y = 125
              towerList.push({ location: 'NE', clone: 'SSW' });
            } else if (x < 80 && y > 100) { // WSW: x = 75 && y = 103
              towerList.push({ location: 'SE', clone: 'WSW' });
            }
          }
        }

        // Now use strategy and towerList (which only contains the correct color for the player)
        // to call out two spots or sort down to one spot.

        const outputMap: { [dir in TowerLocation]: string } = {
          NW: output.northwest!(),
          NE: output.northeast!(),
          SE: output.southeast!(),
          SW: output.southwest!(),
        } as const;

        const [tower0, tower1] = towerList;
        if (tower0 === undefined || tower1 === undefined)
          return;

        if (towerStrategy === 'clockwise' || towerStrategy === 'quadrant') {
          if (matches.effectId === engravementIdMap.lightTower) {
            return output.lightTowerSide!({
              pos1: outputMap[tower0.location],
              pos2: outputMap[tower1.location],
            });
          }

          return output.darkTowerSide!({
            pos1: outputMap[tower0.location],
            pos2: outputMap[tower1.location],
          });
        } else if (towerStrategy === 'tetherbase') {
          // TODO: do some sorting based on tower01.clone
        }
```

YuzukiTsuru:
It seems to work. I'll test it later.

YuzukiTsuru:
I tested it and it seems to works for me.

quisquous:
Sorry for the extra work. I think the end result will make it much better to add another strategy in the future if we need to. Thank you!

PR #5814, opened 2023-09-26
oopsy: report overwritten mit

g0shu:


ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -113,6 +163,62 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
+        const lastTime = data.mitTracker[target]?.[matches.id]?.time;
+        const lastSource = data.mitTracker[target]?.[matches.id]?.source;
+        const newTime = new Date(matches.timestamp).getTime();
+        const newSource = data.ShortName(matches.source);
+        const duration = isTargetMit
+          ? targetMitAbilityIdToDuration[matches.id]
+          : partyMitAbilityIdToDuration[matches.id];
+        if (lastTime !== undefined && lastSource !== undefined && duration !== undefined) {
+          const diff = newTime - lastTime;
+          const leeway = (duration * 1000 - diff) > Options.MinimumTimeForOverwrittenMit * 1000;
```
quisquous, line 190: I think you mean `data.options` here, rather than importing this and checking the default set of option values.

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -113,6 +163,62 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
+              text: {
+                en: 'overwrote ' + lastSource + '\'s ' + matches.ability,
+              },
+            };
+          }
+        }
+        (data.mitTracker[target] ??= {})[matches.id] = {
+          time: newTime,
+          source: newSource,
+        };
```
quisquous, line 209: Simplify by not duplicating the logic to assign to `data.mitTracker`

```suggestion
          (data.mitTracker[target] ??= {})[matches.id] = {
            time: newTime,
            source: newSource,
          };
          if (diff < duration * 1000 && leeway) {
            return {
              type: 'heal',
              blame: matches.source,
              reportId: matches.sourceId,
              text: {
                en: `overwrote ${lastSource}'s ${matches.ability}`,
              },
            };
          }
        }
```

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -113,6 +163,62 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
+        if (isTargetMit && matches.targetId === 'E0000000') // missed
+          return;
+        if (isPartyMit && !data.party.partyIds_.includes(matches.sourceId))
+          return;
+        if (isPartyMit && matches.targetId !== matches.sourceId)
+          return;
+        data.mitTracker ??= {};
+        const target = isTargetMit ? matches.targetId : 'party';
+        const lastTime = data.mitTracker[target]?.[matches.id]?.time;
+        const lastSource = data.mitTracker[target]?.[matches.id]?.source;
```
quisquous, line 182: Sorry for changing my mind here, but in looking at this, I think I'd prefer not to have a special "party" value here. It seems easy to get wrong (or misspell). Can you split the party mit out into a separate object so it can be more properly type checked?

e.g. something like this

```typescript
type MitTracker = { [abilityId: string]: {
    time: number;
    source: string;
  };
};

export interface Data extends OopsyData {
  // etc
  targetedMitTracker?: {
    [targetId: string]: MitTracker;
  };
  partyMitTracker?: MitTracker;
```



and then you could do something more like this

```suggestion
        const mitTracker = isTargetMit ? (targetedMitTracker[target] ??= {}) : partyMitTracker;
        const lastTime = mitTracker[matches.id]?.time;
        const lastSource = mitTracker[matches.id]?.source;
```

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -113,6 +165,71 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
+            time: newTime,
+            source: newSource,
+          };
+        }
+        if (isPartyMit) {
+          (data.partyMitTracker ??= {})[matches.id] = {
+            time: newTime,
+            source: newSource,
+          };
+        }
```
quisquous, line 201:
```suggestion
        const mitTracker = isTargetMit
          ? ((data.targetMitTracker ??= {})[matches.targetId] ??= {})
          : (data.partyMitTracker ??= {});
        const newTime = new Date(matches.timestamp).getTime();
        const newSource = data.ShortName(matches.source);
        const lastTime = mitTracker[matches.id]?.time;
        const lastSource = mitTracker[matches.id]?.source;

        mitTracker[matches.id] = {
          time: newTime,
          source: newSource,
        };
```

one last simplification

PR #5815, opened 2023-09-29
oopsy: better double raise detection

g0shu:


ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -155,16 +158,30 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
+      netRegex: NetRegexes.gainsEffect({ effectId: '94' }),
+      run: (data, matches) => {
+        ((data.raiseGainTracker ??= {})[matches.targetId] ??= []).push(matches.timestamp);
+      },
+    },
+    {
+      id: 'General Raise Removal',
       type: 'LosesEffect',
       netRegex: NetRegexes.losesEffect({ effectId: '94' }),
+      delaySeconds: 0.5,
```
quisquous, line 176: Does this have issues if two people try to raise at the same time, since this clears everything from the gain tracker, so the second raise will not be found?

Wondering if it would make more sense to instead of "all of the last gain timestamps" to have a single "last removal timestamp", and throw an error if a gain comes in after that with the same timestamp.

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -141,8 +142,10 @@ const triggerSet: OopsyTriggerSet<Data> = {
       }),
       // 7D = Raise; AD = Resurrection; E13 = Ascend; 1D63 = Verraise; 5EDF = Egeiro; 478D = BLU; 7423, 7426 = Variant
       mistake: (data, matches) => {
-        data.raiseTracker ??= {};
-        const lastRaiser = data.raiseTracker[matches.targetId];
+        if (matches.targetId === 'E0000000') // target raised before cast finished
```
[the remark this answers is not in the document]
g0shu: hmm probably, i'll look into it

ui/oopsyraidsy/data/00-misc/general.ts
g0shu: i noticed that might be an issue too during my attempted rewrite, so it's working a bit differently there

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -133,17 +134,38 @@ const triggerSet: OopsyTriggerSet<Data> = {
         };
       },
     },
+    {
+      id: 'General Raise Lost Effect Tracker',
+      type: 'LosesEffect',
+      netRegex: NetRegexes.losesEffect({ effectId: '94' }),
+      run: (data, matches) => {
+        ((data.raiseLostTracker ??= {})[matches.targetId] ??= []).push(matches.timestamp);
```
quisquous, line 142: Can this just be the last timestamp, rather than an array? I suspect if two people raised at the precise same time and both events were delivered and processed in the same network packet, you'd still get lose/gain/lose/gain ordering.

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -133,17 +134,38 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
+      },
+    },
     {
       id: 'General Double Raise',
-      type: 'Ability',
-      netRegex: NetRegexes.ability({
-        id: ['7D', 'AD', 'E13', '1D63', '5EDF', '478D', '7423', '7426'],
-      }),
-      // 7D = Raise; AD = Resurrection; E13 = Ascend; 1D63 = Verraise; 5EDF = Egeiro; 478D = BLU; 7423, 7426 = Variant
+      // delaySeconds: 0.5,
```
quisquous, line 147:
```suggestion
```

ui/oopsyraidsy/data/00-misc/general.ts
g0shu: oh true, since i'm no longer using `delaySeconds` i don't need it to be an array anymore

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -199,18 +199,18 @@ const triggerSet: OopsyTriggerSet<Data> = {
       netRegex: NetRegexes.ability({ id: raiseAbiliyIds }),
```
quisquous, line 199: What about:  `netRegex: NetRegexes.ability({ id: raiseAbiliyIds, targetId: 'E0000000' }),` and then you don't need an extra check.

(Also, I know this was there before, but can you rename `raiseAbiliyIds` to `raiseAbilityIds`?)

ui/oopsyraidsy/data/00-misc/general.ts
```diff
@@ -151,35 +151,35 @@ const triggerSet: OopsyTriggerSet<Data> = {
...
             type: 'warn',
             blame: matches.source,
             reportId: matches.sourceId,
             text: {
-              en: 'overwrote ' + lastRaiser + '\'s raise',
-              de: 'überschrieb ' + lastRaiser + '\'s Wiederbeleben',
-              ko: lastRaiser + '의 부활과 겹침',
+              en: 'overwrote ' + originalRaiser + '\'s raise',
+              de: 'überschrieb ' + originalRaiser + '\'s Wiederbeleben',
+              ko: originalRaiser + '의 부활과 겹침',
```
quisquous, line 182:
```suggestion
              en: `overwrote ${originalRaiser}'s raise`,
              de: `überschrieb ${originalRaiser}'s Wiederbeleben`,
              ko: `${originalRaiser}의 부활과 겹침`,
```

and the same below too

quisquous (commented):
I think logic looks good here, with a few cleanups.

I was wondering if we ever needed to clean up `raiseTargetTracker` for logic reasons, and I think it's fine, since the only time you'd ever get an E0000000 is if you casted, and so it'd be updated for the correct target. Mostly saying, looks correct to my eyes.

I do wonder about cleaning up the tracker in some longer sense if it gets full, but it's probably fine for now.

PR #5819, opened 2023-09-30
raidboss/oopsy: Thordan EX

### LINKED ISSUE #5742: Add Thordan Extreme
If anybody has time, it'd be great to get Thordan Extreme in before the unreal is added.

(cc @JLGarber maybe?)

JLGarber:
(This will close #5742 )

This should largely be self-explanatory. We are fortunate that this content is old enough that positions and headings are accurate without having to do any work for them.

I've tested this live both solo and with others (thanks quisquous!) but it will definitely need some polish here and there. In particular, some complex mechanics feel very noisy. It would be nice to figure out what alerts can be downgraded to info.

Note that currently there's no support for keeping Defamation and Pierce players out of the stack during the meteor trio. This can be done without too much effort but I have no brain left for today.

### DIFF
--- a/ui/raidboss/data/03-hw/trial/thordan-ex.ts
+++ b/ui/raidboss/data/03-hw/trial/thordan-ex.ts
@@ -17,7 +17,6 @@ export interface Data extends RaidbossData {
   shieldKnight?: string;
   shieldTarget?: string;
   defCounter: number;
-  pierceTargets: string[];
 }
 
 const unsafeMap: Partial<Record<DirectionOutput8, DirectionOutput8>> = {
@@ -53,7 +52,6 @@ const triggerSet: TriggerSet<Data> = {
       thrustPositions: [],
       seenThrust: false,
       defCounter: 1,
-      pierceTargets: [],
     };
   },
   timelineTriggers: [
@@ -82,14 +80,6 @@ const triggerSet: TriggerSet<Data> = {
       },
     },
     {
-      // This is timed off the towers appearing,
-      // so having no beforeSeconds parameter is correct.
-      id: 'ThordanEX Conviction',
-      regex: /--towers spawn--/,
-      suppressSeconds: 5,
-      response: Responses.getTowers(),
-    },
-    {
       id: 'ThordanEX Heavenly Slash',
       regex: /Heavenly Slash/,
       beforeSeconds: 4,
@@ -99,7 +89,7 @@ const triggerSet: TriggerSet<Data> = {
     {
       id: 'ThordanEX Faith Unmoving',
       regex: /Faith Unmoving/,
-      beforeSeconds: 5,
+      beforeSeconds: 6,
       condition: (data) => data.phase === 2,
       suppressSeconds: 5,
       response: Responses.knockback(),
@@ -190,8 +180,6 @@ const triggerSet: TriggerSet<Data> = {
       netRegex: { id: '0009' },
       condition: (data, matches) => data.me === matches.source || data.me === matches.target,
       alertText: (data, matches, output) => {
-        if (matches.source === undefined || matches.target === undefined)
-          return output.unknown!();
         const partner = data.me === matches.source ? matches.target : matches.source;
         return output.breakChains!({ partner: data.ShortName(partner) });
       },
@@ -199,8 +187,14 @@ const triggerSet: TriggerSet<Data> = {
         breakChains: {
           en: 'Break chains with ${partner}',
         },
-        unknown: Outputs.breakChains,
       },
+    },
+    {
+      id: 'ThordanEX Conviction',
+      type: 'StartsUsing',
+      netRegex: { id: '149D', source: 'Ser Hermenost', capture: false },
+      suppressSeconds: 5,
+      response: Responses.getTowers(),
     },
     {
       id: 'ThordanEX Dragons Gaze',
@@ -219,13 +213,6 @@ const triggerSet: TriggerSet<Data> = {
           en: 'Look away from Thordan and Eye',
         },
       },
-    },
-    {
-      id: 'ThordanEX Conviction Towers',
-      type: 'StartsUsing',
-      netRegex: { id: '149D', source: 'Ser Hermenost', capture: false },
-      suppressSeconds: 5,
-      response: Responses.getTowers(),
     },
     {
       id: 'ThordanEX Triple Spiral Thrust Collect',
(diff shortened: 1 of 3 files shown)


ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -0,0 +1,513 @@
...
+          return;
+        return output.pierceOthers!();
+      },
+      run: (data) => data.pierceTargets = [],
+      outputStrings: {
+        pierceYou: {
+          en: 'Line AoE on YOU',
+        },
+        pierceOthers: {
+          en: 'Avoid tether line AoEs',
```
quisquous, line 395: Do you think this should be dropped for noise reasons? Do folks always have ice if they don't have a tether?

ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -0,0 +1,513 @@
...
+        },
+        ...Directions.outputStrings8Dir,
+      },
+    },
+    {
+      id: 'ThordanEX Faith Unmoving Off Center',
+      type: 'Ability',
+      netRegex: { id: '1018', source: 'Ser Grinnaux' }, // Shared ability from all knights when they teleport in.
+      condition: (data) => data.phase === 4,
+      delaySeconds: 6, // Grinnaux insta-casts Faith Unmoving 13s after appearing. Give ~7s of warning.
```
quisquous, line 487: Maybe give 6s of warning, so arms length etc can cover it as soon as the call happens? (if arms length works?)

ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -0,0 +1,513 @@
...
+      alarmText: (_data, matches, output) => {
```
quisquous, line 488: I don't know that this needs to be an alarm

ui/raidboss/data/03-hw/trial/thordan-ex.txt
```diff
@@ -0,0 +1,281 @@
...
+365.0 "Faith Unmoving" sync / 1[56]:[^:]*:Ser Grinnaux:149B:/
+
+# Intermission part 4
+# Two sets of four Heavy Impacts used during this intermission.
+# They overlap heavily. Because they use different IDs for each ring,
+# it's safe to leave them all synced.
+366.0 "Holy Meteor" sync / 1[56]:[^:]*:Ser Noudenet:14B0:/ window 220,10
+374.6 "--sync--" sync / 1[56]:[^:]*:Ser Guerrique:149F:/ # Heavy Impact (intial cast, no earth ring)
+376.0 "Comet x4" duration 4 #sync / 1[56]:[^:]*:Ser Noudenet:14B5:/
+379.4 "Heavy Impact (round 1 ring 1)" sync / 1[56]:[^:]*:Ser Guerrique:14A0:/
```
quisquous, line 99: this is a little long; I think it could just say ring 1-4 and then again ring 1-4 (even if interleaved)

(alternatively, it could be a duration)

ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -0,0 +1,513 @@
...
+      infoText: (_data, _matches, output) => output.knockbackAoe!(),
+      outputStrings: {
+        knockbackAoe: {
+          en: 'AOE + knockback x7',
+        },
+      },
+    },
+    {
+      id: 'ThordanEX Ultimate End',
+      regex: /Ultimate End/,
```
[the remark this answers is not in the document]
JLGarber: Good point, Conviction actually is, not sure how I missed that. Heavensflame and Dimensional Collapse both have cast times, but the cast times are for the puddle explosions, not the bait snapshots.

The only other real change possible would be to add an ability trigger for Ultimate End. Thordan self-casts 189D eight seconds before the party is hit with 189E:
```
21|2023-09-24T22:58:10.0280000-04:00|40015D6E|King Thordan|148D|Ultimate End|40015D6E|King Thordan|ETC
22|2023-09-24T22:58:18.5350000-04:00|400162DA|King Thordan|148E|Ultimate End|PLAYERID|PLAYER|ETC
```
All other timeline triggers have no startsUsing lines.

ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -0,0 +1,513 @@
...
+      response: Responses.tankBuster(),
+    },
+    {
+      id: 'ThordanEX Holy Chains',
+      type: 'Tether',
+      netRegex: { id: '0009' },
+      condition: (data, matches) => data.me === matches.source || data.me === matches.target,
+      alertText: (data, matches, output) => {
+        if (matches.source === undefined || matches.target === undefined)
+          return output.unknown!();
```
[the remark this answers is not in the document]
JLGarber: No, you're right, that's a late-night mistake.

ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -0,0 +1,513 @@
...
+    },
+    {
+      id: 'ThordanEX Holy Bladedance Collect',
+      type: 'StartsUsing',
+      netRegex: { id: '1496', source: ['Ser Adelphel', 'Ser Janlenoux'] },
+      run: (data, matches) => {
+        if (data.swordKnight === matches.source)
+          data.swordTarget = matches.target;
+        if (data.shieldKnight === matches.source)
+          data.shieldTarget = matches.target;
```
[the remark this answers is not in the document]
JLGarber: They become equal for the check if the target is the same, not the source. (One tank dies, the other tank picks up the loose knight, and then both use Bladedance on the single surviving tank.) Once either knight dies, the other loses their buff, so this trigger will no longer activate, and the call trigger falls back to "heavy busters".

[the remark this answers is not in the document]
quisquous: Oh! Ok, makes sense.

ui/raidboss/data/03-hw/trial/thordan-ex.ts
JLGarber: They do, yes, that's reasonable. The tethers should be visible enough to avoid.

ui/raidboss/data/03-hw/trial/thordan-ex.ts
JLGarber: Whoops, that was from testing. I was missing the knockback notification and wasn't sure whether the trigger was broken or I was just not seeing it in the mess, so I set it to alarm to guarantee I would see it.

ui/raidboss/data/03-hw/trial/thordan-ex.txt
JLGarber: Tried the duration, it seems to be okayish. People can complain at release if it isn't great.

ui/raidboss/data/03-hw/trial/thordan-ex.ts
```diff
@@ -484,8 +452,8 @@ const triggerSet: TriggerSet<Data> = {
       type: 'Ability',
       netRegex: { id: '1018', source: 'Ser Grinnaux' }, // Shared ability from all knights when they teleport in.
       condition: (data) => data.phase === 4,
-      delaySeconds: 6, // Grinnaux insta-casts Faith Unmoving 13s after appearing. Give ~7s of warning.
-      alarmText: (_data, matches, output) => {
+      delaySeconds: 7, // Grinnaux insta-casts Faith Unmoving 13s after appearing. Give ~7s of warning.
```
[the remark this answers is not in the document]
JLGarber: Whoops!

PR #5828, opened 2023-10-04
raidboss: ex7 initial timeline

valarnin:
Just timeline for now. Haven't tested in-game yet, only against pulls from last night and against FFLogs reports.

P2 enrage I'm still not sure how to indicate.

I'll poke at some initial triggers as well but might not have time to actually add any until tomorrow, so if this timeline looks good feel free to merge it now and I'll do the triggers as a separate PR.

Also left my notes in the timeline file instead of deleting them like I normally do, figured they might be useful in the future if the timeline needs cleaned up by someone else.

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
@@ -1,0 +1,13 @@
+import ZoneId from '../../../../../resources/zone_id';
+import { RaidbossData } from '../../../../../types/data';
+import { TriggerSet } from '../../../../../types/trigger';
+
+export type Data = RaidbossData;
+
+const triggerSet: TriggerSet<Data> = {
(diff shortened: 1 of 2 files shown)


quisquous:
I think that's reasonable for the P2 enrage. This happens in a lot of other places (e.g. Sophia) and so there's really nothing to do. Usually I just make the "indeterminate length phase" loop forever and jump to an enrage cast if it starts with a note.

Thanks for this!

valarnin:
> Usually I just make the "indeterminate length phase" loop forever and jump to an enrage cast if it starts with a note.

The weird/annoying part is that both phases are indeterminate length. It's a %-based push, so if the boss doesn't drop below 20% it'll never go to phase two and still enrage at 11m with the exact same cast, but next expansion unsynced you could push the boss down below 20% really early and then phase two loops at some point.

PR #5831, opened 2023-10-05
raidboss/oopsy: Add Singularity Reactor Unreal

JLGarber:
Updated sync_files with data sourced from FFLogs. The skill IDs line up, but some of the actors appear to have changed. Notably, the Comet and Heavy Impact abilities are now sourced from Thordan rather than various knights, and The Light Of Ascalon is also sourced by Thordan. I don't know whether this is an actual change to the encounter or an artifact of FFLogs/ACT needing updates.

This is fully untested in-game, and I won't have time until Friday, but here it is.


JLGarber:
I managed a single instance timer tonight that got to enrage. Timeline looks okay now that I fixed the broken sync. All the directional triggers are non-functional, so I'll look at that tomorrow.

xiashtra:
> I managed a single instance timer tonight that got to enrage. Timeline looks okay now that I fixed the broken sync. All the directional triggers are non-functional, so I'll look at that tomorrow.

There's a log with some more enrage pulls posted in the ACT_FFXIV Discord also.

JLGarber:
Thanks for contributing that, it was helpful. I've corrected the abilities that were not just a simple offset from the EX version and I think it should be good now.

I haven't yet tested the revision in-game, but `test_timeline` shows nothing obviously broken, and no triggers are actively *wrong* that I can see. I think this should be okay for release and we can fix any missing triggers later.

PR #5841, opened 2023-10-07
raidboss: fix ex7 forked lightning

### LINKED ISSUE #5840: Zeromus EX triggers will call for players with a spread debuff to stack
Problem: Existing triggers don't detect the forked lightning debuff applied during meteors.

Explanation: Two DPS players get this debuff as all meteor debuffs are applied, and after meteors, chains, X's and spread markers go off, the last set of debuffs will go off at the same time. This includes the stack marker, placed on a healer, and the two forked lightning debuffs, which will put spread aoes on the two DPS. These triggers call for players marked with forked lightning to "stack" when the stack marker appears, this will paralyze 5 players at best and wipe the raid at worst.

Solution: Detect forked lightning debuff, and for the two dps that have it call for them to "spread" instead of "stack".

wexxlee:
Closes #5840

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
@@ -152,7 +152,6 @@ const triggerSet: TriggerSet<Data> = {
       id: 'ZeromusEx Forked Lightning Collect',
       type: 'GainsEffect',
       netRegex: { effectId: 'ED7' },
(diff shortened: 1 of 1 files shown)


quisquous:
We probably need to do the same thing with tankbusters as well. It's pretty obvious you don't need to stack with a giant red tankbuster on you, but it does tell you.

quisquous:
I'll land this and look at following up with other changes. Thanks for the fix.

PR #5847, opened 2023-10-08
raidboss: update wording for ex7 miasmic

quisquous:
Most players seem to dodge north for this (even though it makes the spread tough). Also, the current version orders the outputs by their ordering in the array, which means that sometimes the "north" direction is first and sometimes second.

To add clarity, only consider safe spots on the north side of the room, and then add explicit wording for where to go.

This is maybe controversial as "east wall" could be ambiguous between ENE and ESE but it felt like to me that if most people dodge north, we can just say the "north" safe spot with details in parentheses. I found this to be more clear, at least.

This output strings change would also let people change to using marker names in the output strings more easily.

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
@@ -222,7 +222,15 @@ const triggerSet: TriggerSet<Data> = {
           intercard: [-2, 0, 2, 3, 8, 13],
         };
 
-        let possibleSafeSpots = Directions.output16Dir;
+        // Filter to north half.
+        let possibleSafeSpots = [
+          'dirNNE',
+          'dirNE',
+          'dirENE',
+          'dirWNW',
+          'dirNW',
+          'dirNNW',
+        ];
 
         for (const blast of data.miasmicBlasts) {
           // special case for center - don't need to find relative dirs, just remove all intercards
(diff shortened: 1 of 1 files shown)


quisquous:
(cc @wexxlee for thoughts; sorry if this is too opinionated of a suggestion)

wexxlee:
Been moving all weekend so haven't had a chance to dive back into ex7.  Conceptually makes sense to me to simplify the calls -- but I seem to remember one pattern where the only north-ish safe spot is actually WSW?  I might be misremembering that though.  Will try to pull my prog logs tonight and run them through emulator with this fix and make sure I'm not hallucinating.

valarnin:
During miasmic, there's always a safe spot in the north half and a safe spot in the south half, at an exact 45º angle from each other. e.g. ENE is safe means that SSW is safe.

ui/raidboss/data/06-ew/trial/zeromus-ex.ts
```diff
@@ -246,23 +254,37 @@ const triggerSet: TriggerSet<Data> = {
...
-        },
         avoidUnknown: {
           en: 'Avoid Line Cleaves',
         },
-        ...Directions.outputStrings16Dir,
+        dirNNE: {
+          en: 'North Wall (NNE/WSW)',
+        },
+        dirNNW: {
+          en: 'North Wall (NNW/WSW)',
```
wexxlee, line 274: I think this should be "North Wall (NNW/ESE)".

quisquous: Whoops!

PR #5855, opened 2023-10-11
i18n: add ex7 zeromus CN translate

YuzukiTsuru:
- add ex7 zeromus CN translate
- `ZeromusEx Abyssal Nox` only aleart Healers

### DIFF
--- a/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
+++ b/ui/raidboss/data/06-ew/trial/zeromus-ex.ts
@@ -89,10 +89,7 @@ const triggerSet: TriggerSet<Data> = {
       type: 'GainsEffect',
       netRegex: { effectId: '6E9', capture: false },
       suppressSeconds: 5,
-      alertText: (_data, _matches, output) => {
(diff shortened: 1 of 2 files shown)


ui/raidboss/data/06-ew/trial/zeromus-ex.ts
```diff
@@ -89,11 +89,15 @@ const triggerSet: TriggerSet<Data> = {
       type: 'GainsEffect',
       netRegex: { effectId: '6E9', capture: false },
       suppressSeconds: 5,
-      alertText: (_data, _matches, output) => output.text!(),
+      alertText: (_data, _matches, output) => {
+        if (_data.role === 'healer')
```
quisquous, line 93: Sorry, I disagree a little here. A lot of players have self-heals that they can use here to help. There's also not anything else going on, so it's not like this is particularly noisy.

YuzukiTsuru: You're right, I will revert it

PR #5861, opened 2023-10-21
raidboss: Support for objects in trigger outputStrings

quisquous:
This is #5551, with some additional changes on top to resolve a few conversations.

Closes #5551.


quisquous:
@valarnin Sorry to step on your patch, but I wanted to get this in while it was on my mind since there's a bunch of changes after this that would be good to make.

valarnin:
All good, I'm fine with you taking it over. This fell a bit to the wayside due to IRL obligations.

quisquous:
> All good, I'm fine with you taking it over. This fell a bit to the wayside due to IRL obligations.

Yeah, I figured! Did you want to take a look at the changes?

resources/party.ts
```diff
@@ -14,6 +15,27 @@ const emptyRoleToPartyNames = () => {
   };
 };
 
+export type BasePartyMemberParamObject = {
+  role?: Role;
+  job?: Job;
+  id?: string;
+  name: string;
+  nick: string;
```
valarnin, line 23: Most triggers currently use `shortName` which is why I included it here to begin with. It seems that `nick` is taking a dual role/default to `shortName` logic path here which may be counterintuitive since everything else here should be a concrete value.

It may make sense to include both here instead, and have the default be `nick` still, so that users have the option of still using `shortName` for specific triggers?

resources/party.ts
```diff
@@ -14,6 +15,27 @@ const emptyRoleToPartyNames = () => {
...
+export interface PartyMemberParamObject
+  extends OutputStringsParamObject, BasePartyMemberParamObject {
+  toString: () => string;
+}
+
+// This is a partial interface of both RaidbossOptions and OopsyOptions.
+export interface PartyTrackerOptions {
+  DefaultPlayerLabel: keyof PartyMemberParamObject;
+  PlayerNicks: { [gameName: string]: string };
+}
```
valarnin, line 37: If this is meant to be a partial interface then it may make sense to use `Pick` to enforce that contract? Something like this?

```ts
export type PartyTrackerOptions = Pick<RaidbossOptions, 'DefaultPlayerLabel'> & {
  DefaultPlayerLabel: keyof PartyMemberParamObject;
}
```

resources/util.ts
```diff
@@ -459,6 +459,26 @@ const Util = {
     clearCombatantsOverride = clearFunc;
   },
   gameLogCodes: gameLogCodes,
+  shortName: (
+    name: string | undefined,
+    playerNicks: { [name: string]: string },
+  ): string => {
+    // TODO: make this unique among the party in case of first name collisions.
+    // TODO: probably this should be a general cactbot utility.
```
valarnin, line 467: I think this second `TODO` comment can be removed? Or at least replaced with a "we should clean up other instances to use this instead"?

ui/raidboss/popup-text.ts
```diff
@@ -260,7 +260,9 @@ class OrderedTriggerList {
   }
 }
 
-const isObject = (x: unknown): x is { [key: string]: unknown } => x instanceof Object;
+const isObject = (x: unknown): x is { [key: string]: unknown } => {
+  return x !== null && x instanceof Object && !Array.isArray(x);
```
valarnin, line 264: This makes sense to me but it may be worth a comment as to why this is necessary, something like "javascript considers `[]` to be an object"?

(also is the `null` check here required?)

valarnin (commented):
Also as an in-general comment, it feels kind of bad to be adding even more code to `popup-text.ts` when that file is already so massive, but I'm not sure if it's feasible to offload the new code to something else.

resources/party.ts
quisquous: It is meant to be a partial interface, but it's more that those two interfaces extend (implicitly) this one.

I feel like the contract is already enforced by typescript as if either of those options change to be not exist or be incompatible, then this callsite will break.

resources/party.ts
quisquous: I'm not sure what you mean by this, or by concrete value. `data.ShortName` the function is both "first name" and "nickname substitution". Users don't really know anything about "short name" as a function and I felt like "nick" better summed up what it did.

What would both options be? Would `shortName` just be `firstName`? Would `nick` be `firstName` if it didn't have a substitution? I'm not sure when you would ever want to pick between these.

quisquous (commented):
> Also as an in-general comment, it feels kind of bad to be adding even more code to `popup-text.ts` when that file is already so massive, but I'm not sure if it's feasible to offload the new code to something else.

Agreed, but also it seems like a preexisting problem that can be solved in a different PR.

resources/util.ts
quisquous: I think all other instances are using it, so I'll remove this comment.

ui/raidboss/popup-text.ts
quisquous: Left a comment, and removed the null check. Good thought! I was thinking of `typeof` and not `instanceof` when I added that.

resources/party.ts
quisquous: Gotcha.

I think we *could* leave the nick field blank if there's no nick and then also have the default field be the short name and it would have the same end result for a user who picks "nick". However, I don't think end users will notice this or be able to do anything with that change either?

resources/party.ts
quisquous: Ok, I added an explicit type here. Because this is TypeScript and it's all duck typing this felt ok to me originally, but you're right that being explicit is better here.

resources/party.ts
valarnin: Yeah, I think this is fine as-is. As I said in my last comment, I think that keeping `nick` as `nickname ?? firstName` is more user-friendly here.

Maybe it also makes sense to add `firstName` as a separate field but I feel like the user should be able to just set someone's nickname to their first name if they wanted to anyways.

PR #5869, opened 2023-10-24
raidboss: Add Option to Order Timeline Bottom-to-Top

quisquous:
This is #4811 with conflicts resolved due to refactoring and 3ae3f6c1eca6dd6a96d515d7d54ed654f5e9642e on top. 3ae3f6c1eca6dd6a96d515d7d54ed654f5e9642e moves the keep alive logic from html_timeline_ui to timeline so that it knows about how many bars are still on screen and also knows when a bar has been expired and it should consider adding more to the list.

Closes #4811.


ui/raidboss/timeline.ts
```diff
@@ -342,16 +352,45 @@ export class Timeline {
...
+            () => {
+              // Find and remove the first keepalive event with this id.
+              let found = false;
+              this.keepAliveEvents = this.keepAliveEvents.filter((x) => {
+                if (found)
+                  return true;
+                if (x.event.id === event.id) {
+                  found = true;
+                  return false;
+                }
```
panicstevenson, line 380: Does this need to `return true;` in the fallthrough where it is not found and the ids are not equal?

ui/raidboss/timeline.ts
```diff
@@ -342,16 +352,45 @@ export class Timeline {
...
 
     this.activeEvents = durationEvents;
   }
 
   private _RemoveExpiredTimers(fightNow: number): void {
     let activeEvent = this.activeEvents[0];
     while (this.activeEvents.length && activeEvent && activeEvent.time <= fightNow) {
-      this.ui?.OnRemoveTimer(activeEvent, true);
+      const event = activeEvent;
+      if (this.options.KeepExpiredTimerBarsForSeconds > 0) {
```
panicstevenson, line 367: (This introduces regression in the user experience where a previously valid negative KeepExpiredTimerBarsForSeconds value was set to the HTML5 minimum of [4ms](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout#Specification). /s)

panicstevenson (approved):
Nice! Thank you for bringing this across the finish line.

ui/raidboss/timeline.ts
quisquous: Oh yeah, thanks!

PR #5878, opened 2023-10-27
test: remove non-existent FileChanged event

quisquous:
It looks like this was removed in:
https://github.com/OverlayPlugin/OverlayPlugin/pull/165

### DIFF
--- a/types/event.d.ts
+++ b/types/event.d.ts
@@ -196,11 +196,6 @@ export interface EventMap {
     charName: string;
   }) => void;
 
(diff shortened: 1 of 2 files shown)


valarnin (approved):
The OverlayPlugin `FileChanged` event was never actually fired by anything. It may have been originally added to allow for a move of cactbot's `onUserFileChanged` event up to OverlayPlugin, but the cactbot equivalent event is also never fired. 🤷

PR #5882, opened 2023-10-28
raidboss: Ktisis Hyperboreia--Add Ladon Lord's Pyric Breath

JLGarber:


quisquous:
Oooooh thanks for this!

PR #5883, opened 2023-10-30
i18n: update missing ko translations

Bing-su:
After updating the find-translations script, it became easier to find missing translations. I used grep before to find only the typescript files.


quisquous:
> After updating the find-translations script, it became easier to find missing translations. I used grep before to find only the typescript files.

Glad to hear that was helpful!

quisquous:
This should have the up to date list too: https://quisquous.github.io/cactbot/util/coverage/missing_translations_ko.html

PR #5889, opened 2023-11-01
raidboss/oopsy: add Another Aloalo Island

quisquous:
Sorry for the awkwardness again, but I'm doing this one blind too (so please don't let me know where this is wrong). I'll put my own notes here in this PR if that's helpful for anybody else.


quisquous:
Ok, this should be ready to go. I feel pretty confident in it and I think there's nothing strat-specific, so I might merge it now and then can adjust for any feedback after.

PR #5890, opened 2023-11-02
docs: Update defs to add currentShield

valarnin:
Per the .cs that Ravahn shared.

Probably there's other updates that can be made, but this was the main driver.

Also `37` lines should probably have actual examples and such instead of having hardcoded values, but that's not really in the scope of this PR.


quisquous:
Thank you!!

PR #5891, opened 2023-11-02
raidboss: update arr and responses to use player objects

quisquous:
Followup to #5861.

### DIFF
--- a/resources/responses.ts
+++ b/resources/responses.ts
@@ -200,7 +200,7 @@ export const Responses = {
       if (target === data.me)
         return;
 
-      return output.busterOnTarget?.({ player: data.ShortName(target) });
+      return output.busterOnTarget?.({ player: data.party.member(target) });
     };
 
     const combined = combineFuncs(
@@ -217,6 +217,7 @@ export const Responses = {
   },
   tankBusterSwap: (busterSev?: Severity, swapSev?: Severity) => {
     const outputStrings = {
+      noTarget: Outputs.tankBuster,
       tankSwap: Outputs.tankSwap,
       busterOnYou: Outputs.tankBusterOnYou,
       busterOnTarget: Outputs.tankBusterOnPlayer,
@@ -236,7 +237,9 @@ export const Responses = {
 
       if (target === data.me)
         return output.busterOnYou?.();
-      return output.busterOnTarget?.({ player: data.ShortName(target) });
+      if (target === undefined)
+        return output.noTarget?.();
+      return output.busterOnTarget?.({ player: data.party.member(target) });
     };
 
     const combined = combineFuncs(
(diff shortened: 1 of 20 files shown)


quisquous:
@valarnin A couple of things this brings up is:

* should `data.party.member(undefined)` return `???` or should the trigger handle it explicitly (I am leaning towards making the trigger do this work but I guess I can see how awkward this becomes later)
* should output support not only objects but arrays of objects and join them with commas into the final string (I'm not sure about this)

valarnin:
> * should `data.party.member(undefined)` return `???` or should the trigger handle it explicitly (I am leaning towards making the trigger do this work but I guess I can see how awkward this becomes later)

I see two options here:
1. `data.party.member(undefined)` returns `undefined` and the trigger handles it
2. `data.party.member(undefined)` returns a dummy party member object with all fields set to `???`

I'm fine with either one, honestly. The latter option would probably be better to reduce boilerplate in triggers though, and still be detectable if a trigger explicitly needed to check for it?

> * should output support not only objects but arrays of objects and join them with commas into the final string (I'm not sure about this)

I think there are a couple of other questions that would need answered first:
1. Is `,` the most appropriate separator? Does that make sense for all the languages that cactbot supports, or would that need to be localize-able?
2. Should the input string array be `join`'d directly and then substituted, or should each input string in the array be substituted and then the resultant array be joined, or possibly both with some sort of selector? e.g.:
```js
const inputStrings = ['A', 'B', 'C'];
const outputString = 'something ${value}';

const option1Result = 'something A,B,C';
const option2Result = 'something A,something B,something C';
```

quisquous:
> > * should output support not only objects but arrays of objects and join them with commas into the final string (I'm not sure about this)
> 
> I think there are a couple of other questions that would need answered first:
> 
> 1. Is `,` the most appropriate separator? Does that make sense for all the languages that cactbot supports, or would that need to be localize-able?

As far as I can tell, this is generally what gets used, e.g. `T6 Thorn Whip`'s `twoTethers` output string. There are some triggers (see: https://github.com/quisquous/cactbot/pull/5891/files#diff-7796360647af53a46da16de5fff30e73d36d01e87dfa0489475c60354678f905L45) that just insert commas directly (which imo they shouldn't do, but unless `output` starts returning non-string values, there's not much to be done. A quick search for `/map.*ShortName/` shows ~10 more examples of this (although other different uses of that as well).

I could see an argument that this is ~ok and if it's awkward in a particular language that trigger could handle it specially.

> 2. Should the input string array be `join`'d directly and then substituted, or should each input string in the array be substituted and then the resultant array be joined, or possibly both with some sort of selector? e.g.:
> 
> ```js
> const inputStrings = ['A', 'B', 'C'];
> const outputString = 'something ${value}';
> 
> const option1Result = 'something A,B,C';
> const option2Result = 'something A,something B,something C';
> ```

I think `option1Result` is the one that makes more sense. It's usually a "X on a, b, c" or "X with a, b, c".

PR #5895, opened 2023-11-03
raidboss: handle player object arrays in output params

quisquous:
Followup to #5861 and discussions on #5891.

As the comment in the code specifies, if a trigger passes in `players: [player1, player2, player3]`, and a user specifies `${players.job}`, then return: `${players[0].job}, ${players[1].job}, ${players[2].job}`.

In general, this means that all array elements must either be simple strings/numbers or all share the same prop, or there will be errors below about non-existent properties. In practice, this likely will never happen.

This will also handle simpler cases like: `output.safeSpots!({ dirs: [output.front!(), output.back!(), output.left!()] })`

This also allows for `data.party.member(undefined)` to return '???', just to reduce boilerplate.

### DIFF
--- a/resources/party.ts
+++ b/resources/party.ts
@@ -159,7 +159,17 @@ export default class PartyTracker {
     return this.idToName_[id];
   }
 
-  member(name: string): PartyMemberParamObject | undefined {
+  member(name?: string): PartyMemberParamObject {
+    // For boilerplate convenience in triggers, handle undefined names.
+    if (name === undefined) {
+      const unknown = '???';
+      return {
+        name: unknown,
+        nick: unknown,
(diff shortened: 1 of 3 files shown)


quisquous:
cc @valarnin

valarnin:
This looks good to me at a glance but I don't really have time right now to deep dive on the potential recursive issues to make sure there's no problems there.

quisquous:
> This looks good to me at a glance but I don't really have time right now to deep dive on the potential recursive issues to make sure there's no problems there.

Thanks for taking a look. Mostly I was just looking for a high level "does this seem like a reasonable approach to take", and it sounds like that sounds good.

PR #5896, opened 2023-11-03
raidboss: Add 263/264 line types

valarnin:
Add the new line types, and their corresponding documentation.

I don't think I missed anything here but please double check as the info included in LogGuide.md may be excessive or not 100% accurate for all cases.

### DIFF
--- a/docs/LogGuide.md
+++ b/docs/LogGuide.md
@@ -181,6 +181,14 @@ This guide was last updated for:
     - [Structure](#structure-33)
     - [Regexes](#regexes-33)
     - [Examples](#examples-33)
+  - [Line 263 (0x107): StartsUsingExtra](#line-263-0x107-startsusingextra)
+    - [Structure](#structure-34)
+    - [Regexes](#regexes-34)
+    - [Examples](#examples-34)
+  - [Line 264 (0x108): AbilityExtra](#line-264-0x108-abilityextra)
+    - [Structure](#structure-35)
+    - [Regexes](#regexes-35)
+    - [Examples](#examples-35)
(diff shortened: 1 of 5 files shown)


docs/LogGuide.md
```diff
@@ -2590,3 +2598,113 @@ Parsed Log Line Examples:
...
 
 <!-- AUTO-GENERATED-CONTENT:END (logLines:type=RSVData&lang=en-US) -->
+
+<a name="line263"></a>
+
+### Line 263 (0x107): StartsUsingExtra
+
+This line contains extra data for ActorCast/StartsUsing network data.
+
+This line is always output for a given StartsUsing cast. If the ability does not target the
```
quisquous, line 2608: This is a belated thought for the OP patch, but I wonder if it'd be better if this line always printed the source position and then also included the target data if it existed (maybe with flags here too, if they exist). Mostly just saying that there's a lot of "figure out where things are when they start casting" that would benefit from always having the source actor position. (cc @xpdota too)

xpdota: Wouldn't it have the same issue as the native 20/21/22 lines, where a position read from memory might be stale?

valarnin: I think the hope was that it would have actor position data in the network packet as well, but I don't think that's the case?

quisquous: Yeah, that was my hope. I guess if it doesn't have it, then it doesn't have it.

xpdota: I think the only true solution for cast-before-move race conditions - without modifying the base plugin - is to hook ActorMove (and anything else that could move an actor), and have that data take priority over memory in 261-lines.

PR #5901, opened 2023-11-04
docs: Clarify 263/264 in LogGuide.md

xpdota:
Clarify how 263/264-lines work with regards to abilities that have a cast direction/angle but do not target a specific location.


quisquous:
Thank you!

PR #5907, opened 2023-11-04
docs: update documentation to use player objects

quisquous:
Followup to #5861.

### DIFF
--- a/docs/CactbotCustomization.md
+++ b/docs/CactbotCustomization.md
@@ -303,16 +303,18 @@ Options.Triggers.push({
         // or I am not a tank.
         return true;
       },
-      infoText: function(data, matches) {
-        let name = data.ShortName(matches.source);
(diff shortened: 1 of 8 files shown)


Souma-Sumire:
Is it uncertain whether this is the right place to ask, but I wanted to inquire .
If there are any plans to remove the data.shortName method in the future, or will it be retained indefinitely?

quisquous:
> Is it uncertain whether this is the right place to ask, but I wanted to inquire . If there are any plans to remove the data.shortName method in the future, or will it be retained indefinitely?

In general if I remove something, I'll add some warning about it. Since people are writing their own triggers, I try to keep backwards compatibility as much as I can. In other words, I'm not planning to remove it, but I am planning to add tests so that people don't add more uses of this to cactbot built-in triggers.

PR #5919, opened 2023-11-10
plugin: delete fishing code and overlay

quisquous:
Dalamud's GatherBuddy plugin (https://github.com/Ottermandias/GatherBuddy) is a much better fishing tracker than cactbot's fisher overlay. If you were using this, you should switch to that.

I'm sorry about your fish data; you could possibly get it out of IndexedDB and into GatherBuddy import with a good bit of work but it's not something I can help with.

The fishing code is very out of date, was never converted to TypeScript, and hasn't gotten any attention other than me. I mentioned that I was planning to remove it on discord and in release notes, and nobody spoke up complaining that it was being removed let alone spoke up saying they wanted to maintain it. Sorry!

If this is something you want to maintain, please speak up.

You could probably, at least until the bait memory signature changes (probably in Dawntrail, but maybe never) or until cactbot makes some breaking change to code that it's using, keep using the fisher overlay you have locally already indefinitely. It just won't get updates and will probably break at some point in the future.


Tykku:
People do read these, and appreciate your work. <3

DariusDawnbreaker:
You told me to speak up so i speak up. Since im not using Dalamud i was using your fisher overlay for a long time. If it would be updated and maintained i would appreciate it and i will use it happily again:) If you dont have the time or energy to do it, i can understand it as well;)

quisquous:
Your best bet is to start using dalamud and use GatherBuddy. Your next best alternative is to manually install an older cactbot version and use that (cancelling updates) until it stops working. If you want to maintain this yourself, I can give you a few pointers on how you might go about that.

PR #5926, opened 2023-11-13
raidboss: Lunar Subterrane -- Add Forsaken Fount triggers

JLGarber:
The in-line comments should largely cover everything. It's maybe more complex than other possibilities for how we can handle the situation, but it does work.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
@@ -234,8 +234,8 @@ const triggerSet: TriggerSet<Data> = {
           return;
         // If there are five orbs on the field, three of them will, guaranteed,
         // have the same X or Y value. Those three are in a straight line.
-        const hCount = data.fountY.filter((n) => n === 0).length;
-        const vCount = data.fountX.filter((n) => n === 0).length;
+        const hCount = data.fountY.filter((n) => Math.abs(n) < 1).length;
+        const vCount = data.fountX.filter((n) => Math.abs(n) < 1).length;
 
         if (hCount === 3) {
           // Horizontal lines are always east/west safe hammer patterns.
--- a/ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
@@ -212,6 +212,49 @@ const triggerSet: TriggerSet<Data> = {
       // (12.8, -422)
       // (-12.8, -422)
       // (Other spawn locations exist for rounds 1/2, but we can ignore those.)
+
+      // The primary known configurations for round three onward are:
+      // West Safe
+      // (-12.80,-422.00), (0.00,-422.00), (12.80,-422.00)
+      // (9.15,-431.10)
+      // (3.65,-412.90)
+
+      // North Safe
+      // (0.00,-409.20), (0.00,-422.00), (0.00,-434.80)
+      // (9.05,-412.80)
+      // (-9.05,-418.35)
+
+      // East Safe
+      // (-12.80,-422.00),  (0.00,-422.00), (12.80,-422.00)
+      // (-9.25,-431.05)
+      // (-3.65,-412.95)
+
+      // Hourglass
+      // (0.00,-409.20), (0.00,-422.00), (0.00,-434.80)
+      // (-9.05,-418.35)
+      // (9.15,-425.55)
+      // So far nobody has seen a configuration with South being safe.
+
+      // When the Y axis is normalized to 0, the sets look like this:
+      // West Safe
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
```diff
@@ -162,6 +172,111 @@ const triggerSet: TriggerSet<Data> = {
...
+      netRegex: { name: 'Aetheric Charge', capture: false },
+      condition: (data) => data.fountsSeen === 2,
+      delaySeconds: 0.5,
+      infoText: (data, _matches, output) => {
+        if (data.fountX.length < 5 || data.fountY.length < 5)
+          return;
+        // If there are five orbs on the field, three of them will, guaranteed,
+        // have the same X or Y value. Those three are in a straight line.
+        const hCount = data.fountY.filter((n) => n === 0).length;
+        const vCount = data.fountX.filter((n) => n === 0).length;
```
quisquous, line 238: Can you add some tolerance here for "close to zero"? This is maybe me being conservative but I think it just feels wrong to do `===` on a floating point value. Even `Math.abs(n) < 1` seems plenty.

ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
```diff
@@ -162,6 +172,111 @@ const triggerSet: TriggerSet<Data> = {
...
+
+        if (hCount === 3) {
+          // Horizontal lines are always east/west safe hammer patterns.
+          const xSum = data.fountX.reduce((a, b) => a + b, 0);
+
+          // Don't rely on the rounded sum to be precise, but the sign will be reliable.
+          if (xSum > 5)
```
quisquous, line 245: Can you give an example of what these >5, <5, >10, <10 patterns look like in terms of values? Summing makes sense as a way to differentiate between cases when you know the full set of examples, but it's a bit confusing to a reader. (Also, minorly and mostly curiously rather than something to fix, it makes me wonder if there's some simpler way to differentiate?)

ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
JLGarber: It's not a float, because `Math.round` returns integers, but I suppose we can still do that to avoid potential issues with -0 if the `matches` value ends up being in the -0.5 - 0.0 range.

ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
JLGarber: Sure, here's a raw list of configurations:

```
West Safe
(-12.80,-422.00), (0.00,-422.00), (12.80,-422.00)
(9.15,-431.10)
(3.65,-412.90)

North Safe
(0.00,-409.20), (0.00,-422.00), (0.00,-434.80)
(9.05,-412.80)
(-9.05,-418.35)

East Safe
(-12.80,-422.00),  (0.00,-422.00), (12.80,-422.00)
(-9.25,-431.05)
(-3.65,-412.95)

Hourglass
(0.00,-409.20), (0.00,-422.00), (0.00,-434.80)
(-9.05,-418.35)
(9.15,-425.55)
```

And Y-normalized:
```
West Safe
(-12.80,0), (0,0), (12.8,0)
(9.15,-9.1)
(3.65,9.1)

North Safe
(0,12.8), (0,0), (0,-12.8)
(9.05,9.2)
(-9.05,3.65)

East Safe
(-12.8,0),  (0,0), (12.8,0)
(-9.25,-9.05)
(-3.65,9.05)

Hourglass
(0,12.8), (0,0), (0,-12.8)
(-9.05,3.65)
(9.15,-3.55)
```

I'm not completely sure this list is exhaustive, but I'm confident the list of orb locations in the inline comment is.

ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
quisquous: Oh quite right, this is all rounded. sounds good.

ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
quisquous: That makes sense! Can you add this as a comment somehow?

JLGarber: Let's try this. I don't like how long it is, but the information is there. I also added some more in-line comments at the points where the output is actually returned.

quisquous: Thanks, I think that's helpful for future debugging, especially the "this probably does not exist" kind of comment.

PR #5930, opened 2023-11-14
raidboss: localized job names

Souma-Sumire:
Please provide any comments.


resources/party.ts
```diff
@@ -15,6 +17,184 @@ const emptyRoleToPartyNames = () => {
...
+};
+
+// TODO: ja(abbr), de, fr, ko
+const jobLocalized: Record<Job, Record<'abbr' | 'full', Partial<Record<Lang, string>>>> = {
+  NONE: {
+    abbr: { ja: 'すっぴん士', cn: '冒险' },
+    full: { en: 'Adventurer', ja: 'すっぴん士', cn: '冒险者' },
+  },
+  GLA: {
+    abbr: { ja: '剣術士', cn: '剑术' },
```
MaikoTan, line 37: Are you lacking of English abbreviations here? And Japanese has their abbreviations for most jobs in just one or two characters rather than call out the full names, e.g. 剣 for 剣術士, ナ for ナイト or 暗 for 暗黒騎士. If you're not sure about that, maybe you could leave it blank and Japanese folks would fill them up once the PR got merged.

resources/party.ts
```diff
@@ -181,12 +361,17 @@ export default class PartyTracker {
...
+      const jobFull = jobLocalized[job]?.full?.[Options.AlertsLanguage ?? 'en'] ?? job;
+      const role = Util.jobToRole(job);
+      const roleName = roleLocalized[role]?.[Options.AlertsLanguage ?? 'en'] ?? role;
       ret = {
         id: partyMember.id,
-        job: jobName,
-        role: role,
+        job: job,
+        jobAbbr: jobAbbr,
+        jobFull: jobFull,
```
MaikoTan, line 373: Since the names and the variables are the same for object properties, you could use the simpler alternative representation for it:

```js
let ret = {
  job,
  jobAbbr,
  jobFull,
}
```

represent the same as following, but simpler:

```js
let ret = {
  job: job,
  jobAbbr: jobAbbr,
  jobFull: jobFull,
}
```

types/party.d.ts
```diff
@@ -1,9 +1,11 @@
-import { Job, Role } from './job';
+import { Job } from './job';
 import { OutputStringsParamObject } from './trigger';
 
 export type BasePartyMemberParamObject = {
-  role?: Role;
+  role?: string;
```
MaikoTan, line 5: Why change the type of the `Role`? I think you should keep it remained but add `none` into `Role` instead.

ui/raidboss/raidboss_config.ts
```diff
@@ -2174,13 +2174,17 @@ const templateOptions: OptionsTemplate = {
...
         cn: {
-          '昵称 (Tini)': 'nick',
+          '昵称 (弗雷)': 'nick',
           '职能 (坦克)': 'role',
-          '职业 (战士)': 'job',
-          '全名 (Tini Poutini)': 'name',
+          '职业缩写 (DRK)': 'job',
+          '职业简称 (暗骑)': 'jobAbbr',
+          '职业全称 (暗黑骑士)': 'jobFull',
+          '全名 (弗雷)': 'name',
```
MaikoTan, line 2187: This change makes `nick` and `name` looks like the same behaviour, also it looks like `黑骑` is the well-known abbreviation for Dark Knights much than `暗骑` that you wrote here in Chinese servers.

resources/party.ts
Souma-Sumire: I didn't use abbreviations because the original code didn't have any abbreviations, such as 'name' and 'nick,' so I followed this style.

types/party.d.ts
Souma-Sumire: Because 'role' is not a fixed string after localization.

ui/raidboss/raidboss_config.ts
Souma-Sumire: I didn't modify the logic in this part; 'nick' and 'name' are the same in the Chinese server by default.

I use '暗'骑 instead of '黑'骑 because in PvP, the CN localization for DRK is translated as '暗,' while BLM is '黑.

types/party.d.ts
MaikoTan: Hmmm, in this case, I would recommend to add an extra field to store the localized name rather than changing the original one. What do you think?

ui/raidboss/raidboss_config.ts
MaikoTan: Okay, that sounds fair.

resources/party.ts
MaikoTan: Well, in general we prefer the shorter style, since there isn't a lint rule forcing this style currently, it is also ok to not changing it just for now.

Souma-Sumire: Oh, I tried to abbreviate it, but it violates the project's ESLint rules "object-shorthand".

types/party.d.ts
Souma-Sumire: I agree with your idea. Initially, I also added a new attribute.

However, later I thought that existing users might have saved options with 'role' as the key. 

For instance, if I added a new one, let's say 'roleName,' how should we handle users who previously saved options with 'role'? 

Do we need to delete the 'role' option, and if not deleted, this option would only be effective for English language users, while 'roleName' would only be effective for non-English users. This would be quite awkward. 

I haven't come up with a good solution.

resources/party.ts
MaikoTan: Oh, it looks like we are forcing to not use the shorter style, I recall the wrong side, my apologies. Never mind.

resources/party.ts
```diff
@@ -15,6 +17,184 @@ const emptyRoleToPartyNames = () => {
...
+  tank: { ja: 'タンク', cn: '坦克' },
+  healer: { ja: 'ヒーラー', cn: '治疗' },
+  dps: { ja: 'DPS', cn: '输出' },
+  crafter: { ja: 'クラフター', cn: '能工巧匠' },
+  gatherer: { ja: 'ギャザラー', cn: '大地使者' },
+  none: { ja: 'すっぴん士', cn: '冒险者' },
+};
+
+// TODO: ja de fr ko
+const jobLocalized: Record<Job, Record<'abbr' | 'full', Partial<Record<Lang, string>>>> = {
```
quisquous, line 31: Similar to the above, I know it makes more sense from a code perspective to put these in the same object, but if you separate this into `jobLocalizedAbbr: Record<Job, LocaleText>` and `jobLocalizedFull: Record<Job, LocaleText>` separately then it will show up in the find translations script (and no need for a TODO).

quisquous (commented):
Thank you!!!

resources/party.ts
```diff
@@ -181,12 +280,16 @@ export default class PartyTracker {
         nick: nick,
       };
     } else {
-      const jobName = Util.jobEnumToJob(partyMember.job);
-      const role = Util.jobToRole(jobName);
+      const job = Util.jobEnumToJob(partyMember.job);
+      const jobAbbr = jobLocalizedAbbr[job]?.[Options.AlertsLanguage ?? 'en'] ?? job;
+      const jobFull = jobLocalizedFull[job]?.[Options.AlertsLanguage ?? 'en'] ?? job;
+      const role = Util.jobToRole(job);
+      const roleName = roleLocalized[role]?.[Options.AlertsLanguage ?? 'en'] ?? role;
```
quisquous, line 287:
```suggestion
      const lang = this.options.DisplayLanguage;
      const jobAbbr = jobLocalizedAbbr[job]?.[lang] ?? job;
      const jobFull = jobLocalizedFull[job]?.[lang] ?? job;
      const role = Util.jobToRole(job);
      const roleName = roleLocalized[role]?.[lang] ?? role;
```

The `Options` object is the default and not (necessarily?) the one with user config applied. It's better to use the one passed into the constructor here. I think DisplayLanguage is more correct here as `this.options` doesn't know about `AlertsLanguage` for oopsy. You'd need to add it to the `PartyTrackerOptions` type.

resources/party.ts
```diff
@@ -15,6 +17,103 @@ const emptyRoleToPartyNames = () => {
   };
 };
 
+const roleLocalized: Record<Role, LocaleText> = {
+  tank: { en: 'tank', ja: 'タンク', cn: '坦克' },
```
quisquous, line 21: Is it possible to put these fields on separate lines for the find translations script, e.g.

```
const roleLocalized: Record<Role, LocaleText> = {
  tank: {
    en: 'tank',
    ja: 'タンク',
    cn: '坦克',
  },
```

resources/party.ts
quisquous: (sorry for the complication, I forgot this part)

types/party.d.ts
```diff
@@ -17,7 +18,7 @@ export interface PartyMemberParamObject
 }
 
 // This is a partial interface of both RaidbossOptions and OopsyOptions.
-export interface PartyTrackerOptions {
+export interface PartyTrackerOptions extends BaseOptions {
```
quisquous, line 21: Ah this is a good idea, thanks!

PR #5934, opened 2023-11-15
jobs: user options to control Job display

Souma-Sumire:
A rough attempt, please advise.


MaikoTan (commented):
I saw there are many `optionJob` keys with the same structure, consider use an IIFE?

```ts
[
  // ...
  ...(() =>
    jobs.map((job) => ({
      id: `Option${job}`,
      name: localizeJob(job),
      type: 'select',
      options: {
        en: {
          'Disabled': 'disabled',
          'Only Bar': 'onlyBar',
          'Enabled': 'enabled',
        },
        cn: {
          禁用: 'disabled',
          仅基础: 'onlyBar',
          完整显示: 'enabled',
        },
      },
      default: 'enabled',
    })))(),
]

function localizeJob(job: string) {
  // the `jobLocalizedFull` can be exported from resources/party.ts
  return {
    en: 'Option ' + jobLocalizedFull[job].en,
    cn: '模块：' + jobLocalizedFull[job].cn,
  }
}
```

But IMO it still is absurdly redundant options that basically records a mapping table.

------

Here's my two cents:

未命名绘图 drawio

As the image above shows, we could put a searchable dropdown on the left, and then the right side can be used to adjust the setting for the job selected. Since we are not using any JS framework, I am not sure how to implement this on vanilla JS, but I think it's doable.

Echoring:
These options looks well. 

However, I think most player do not truely need a pure basic HP/MP bar without other functions. 
In fact the most frequently asked question I've seen is how to hide the upper part (Resource zone) or the lower part (Tracking zone) alone (and BLM ask for only MP ticker bar). 
But it seems more complex to control the element in job component instead of preventing the whole job component, so be it for now (or leave an interface here for mionr option?)

ui/jobs/jobs_options.ts
```diff
@@ -25,6 +25,26 @@ export interface JobsNonConfigOptions {
   NotifyExpiredProcsInCombat: number;
   NotifyExpiredProcsInCombatSound: 'disabled' | 'expired' | 'threshold';
   CompactView: boolean;
+  OptionAST: 'disabled' | 'onlyBar' | 'enabled';
```
quisquous, line 28: I worry that this might be limiting for the future, if we ever want to have more toggles that won't fit in a single type.

I think you could do something like this:

```typescript
// in this data structure
PerJobOption: { [name in Job]?: JobOption };

// elsewhere
export type JobOption = {
  enabled: boolean;
  hideBoxes: boolean;
};
```

In `jobs_config.ts`, you *should* be able to use `default` to turn `options.PerJobOption` into the correct value when reading and use `setterFunc` to turn the ui value back into the `options.PerJobOption` when writing. See `TimeToShowDeathReportSeconds` for a `setterFunc` example, and e8s's `uptimeKnockbackStrat` for a `default` function example.

Let me know if that doesn't make sense. But, I think having this would allow us to have per job "hide box 1 / box 2 / box 3" or hide bars and not boxes or show mp / hp differently etc much more easily.

MaikoTan: Agreed to this thought but I think it would be better to apply @Echoring 's idea here, dividing the monitoring area into `resouces` (hp, mp, job-specific meters like oath for PLD) and `proc` (including DoT), then we have a structure like:

```
interface JobOption {
  resources: boolean;
  procs: boolean;
}
```

Am I understand it correctly, @Echoring ?

quisquous: I feel like this PR is a step in that direction. My comments above are mostly to make sure that it's possible to follow up and add those sorts of additional config tweaks. Do you agree with that, or do you think this would prevent that?

Mostly just trying to suggest that we don't make Souma-Sumire do all the work in this PR, and that we land something that will enable those possibilities in the future.

MaikoTan: That's fair, I think what you wrote is good to me. Then me and @Echoring would talk about the improvement in the next PR.

quisquous: Sorry if I added too much friction to this PR with my comments, I just wanted to make sure that this wouldn't create issues for other things that people wanted to do. Hopefully what I asked for makes sense; let me know if it's not possible or if I can help.

quisquous:
Hi, sorry for the hassle. I've moved cactbot to OverlayPlugin/cactbot instead of here. If you'd like to get this PR landed, please send it to <https://github.com/OverlayPlugin/cactbot/> instead.

(Sorry again if I caused too much trouble with my suggestions. I still think this is a really great idea, I just wanted to make sure things would be flexible for the future.)

Souma-Sumire:
I agree with your idea, but I encountered some difficulties in its implementation. I believe others might be able to handle it more effectively, so I chose to put it on hold. I apologize for the delay.

I'd like to close it until someone is ready to submit a PR in the new repository.

MaikoTan:
I think I could take over it in tomorrow or other days, please keep your branch live until I came up with a PR in OverlayPlugin's fork.

PR #5938, opened 2023-11-19
i18n: update JP translations

kshman:


quisquous:
Oh wow, thank you for all this work!

PR #5939, opened 2023-11-20
raidboss: Add support for netregexes in timelines

### LINKED ISSUE #5145: raidboss: use network log lines in timeline files
Currently timeline files are the only things still using processed log lines and not network log lines.  They're also using regexes directly, which is a little bit awkward.  It'd be nice to switch timeline syncs to using NetRegexes in some form, however it's not clear to me what this format should look like:

```txt
# currently
1000.0 "stuff" sync / 1[56]:[^:]*:thing:32:/

# options
1000.0 "stuff" sync { type: "ability", id: 32, source: "thing" } # json
1000.0 "stuff" ability { id: 32, source: "thing" } # keyword per line type, shorter
1000.0 "stuff" sync:ability { id: 32, source: "thing" } # hybrid
1000.0 "stuff" sync "ability,id,32,source,thing" # csv???
```

These are certainly more readable than the original, but I'm not sure I'm really sold on any of them.

Other thoughts I'm having are that "ability" is so common that maybe it's the default and doesn't have to be specified, for brevity?

cc @JLGarber @valarnin @panicstevenson

valarnin:
Closes #5145

Obviously needs plenty of testing etc.

Example added to `test.txt` timeline, can be triggered with `/e testNetRegexTimeline`.

Took the opportunity to refactor the `TimelineParser.parse` function to extract out some of the line matching, so there wasn't a need for a 30+-space-indented set of lines, and to allow reusing some of the logic for the rest of the line matching after the regex.

### DIFF
--- a/ui/raidboss/timeline_parser.ts
+++ b/ui/raidboss/timeline_parser.ts
@@ -17,15 +17,9 @@ const isStringArray = (value: unknown[]): value is string[] => {
   return value.find((v) => typeof v !== 'string') === undefined;
 };
 
-const isNumberArray = (value: unknown[]): value is number[] => {
-  return value.find((v) => typeof v !== 'number') === undefined;
-};
-
-const isStringOrNumberOrArray = (value: unknown): value is number[] | string[] => {
+const isStringOrStringArray = (value: unknown): value is number[] | string[] => {
   if (Array.isArray(value)) {
     if (isStringArray(value))
-      return true;
-    else if (isNumberArray(value))
       return true;
     return false;
   } else if (!['string', 'number'].includes(typeof value))
@@ -42,7 +36,7 @@ const isValidNetParams = <T extends LogDefinitionTypes>(
     if (!(key in logDefinitions[type].fields))
       return false;
     // Make sure our value is either a string/int or an array of strings/ints
-    if (!isStringOrNumberOrArray(params[key]))
+    if (!isStringOrStringArray(params[key]))
       return false;
   }
   return true;
@@ -507,6 +501,8 @@ export class TimelineParser {
     return this.buildRegexSync(
       uniqueid,
       `${netRegexType} ${syncCommand.netRegex}`,
+      // TODO: Use `translateRegexBuildParam` instead, store off params for use elsewhere.
+      // See https://github.com/quisquous/cactbot/pull/5939#discussion_r1399453530
       Regexes.parse(this.GetReplacedSync(buildNetRegexForTrigger(netRegexType, params))),
       syncCommand.args,
       seconds,
--- a/ui/raidboss/timeline_parser.ts
+++ b/ui/raidboss/timeline_parser.ts
@@ -22,7 +22,7 @@ const isStringOrStringArray = (value: unknown): value is string | string[] => {
     if (isStringArray(value))
       return true;
     return false;
-  } 
+  }
   return typeof value === 'string';
 };
 
--- a/ui/raidboss/timeline_parser.ts
+++ b/ui/raidboss/timeline_parser.ts
@@ -81,6 +81,7 @@ export type Error = {
 export type Sync = {
   id: number;
   origRegexStr: string;
+  regexType: 'parsed' | 'net';
   regex: RegExp;
   start: number;
   end: number;
@@ -479,6 +480,8 @@ export class TimelineParser {
 
     let params: unknown;
     try {
+      // TODO: Use `exec` => `JSON.stringify` => `JSON.parse` instead?
+      // See https://github.com/quisquous/cactbot/pull/5939#issuecomment-1830484703
       params = JSON.parse(syncCommand.netRegex);
     } catch (e) {
       this.errors.push({
@@ -499,6 +502,7 @@ export class TimelineParser {
     }
     return this.buildRegexSync(
       uniqueid,
+      'net',
       `${netRegexType} ${syncCommand.netRegex}`,
       // TODO: Use `translateRegexBuildParam` instead, store off params for use elsewhere.
       // See https://github.com/quisquous/cactbot/pull/5939#discussion_r1399453530
@@ -527,6 +531,7 @@ export class TimelineParser {
     line = line.replace(syncCommand.text, '').trim();
     return this.buildRegexSync(
       uniqueid,
+      'parsed',
       syncCommand.regex,
       Regexes.parse(this.GetReplacedSync(syncCommand.regex)),
       syncCommand.args,
@@ -539,6 +544,7 @@ export class TimelineParser {
 
   private buildRegexSync(
     uniqueid: number,
+    regexType: 'parsed' | 'net',
     regex: string,
     parsedRegex: RegExp,
     args: string | undefined,
@@ -550,6 +556,7 @@ export class TimelineParser {
     const sync: Sync = {
       id: uniqueid,
       origRegexStr: regex,
+      regexType: regexType,
       regex: parsedRegex,
       start: seconds - 2.5,
       end: seconds + 2.5,
--- a/ui/raidboss/timeline_parser.ts
+++ b/ui/raidboss/timeline_parser.ts
@@ -1,10 +1,57 @@
 import { Lang } from '../../resources/languages';
+import logDefinitions, { LogDefinitionTypes } from '../../resources/netlog_defs';
+import { buildNetRegexForTrigger } from '../../resources/netregexes';
 import { UnreachableCode } from '../../resources/not_reached';
 import Regexes from '../../resources/regexes';
 import { translateRegex, translateText } from '../../resources/translations';
+import { NetParams } from '../../types/net_props';
 import { LooseTimelineTrigger, TriggerAutoConfig } from '../../types/trigger';
 
 import defaultOptions, { RaidbossOptions, TimelineConfig } from './raidboss_options';
+
+const isLogDefinitionTypes = (type: string): type is LogDefinitionTypes => {
+  return type in logDefinitions;
+};
+
+const isStringArray = (value: unknown[]): value is string[] => {
+  return value.find((v) => typeof v !== 'string') === undefined;
+};
+
+const isNumberArray = (value: unknown[]): value is number[] => {
+  return value.find((v) => typeof v !== 'number') === undefined;
+};
+
+const isStringOrNumberOrArray = (value: unknown): value is number[] | string[] => {
+  if (Array.isArray(value)) {
+    if (isStringArray(value))
+      return true;
+    else if (isNumberArray(value))
+      return true;
+    return false;
+  } else if (!['string', 'number'].includes(typeof value))
+    return false;
+  return true;
+};
+
+const isValidNetParams = <T extends LogDefinitionTypes>(
+  type: T,
+  params: Record<string, unknown>,
+): params is NetParams[T] => {
+  for (const key in params) {
+    // Make sure all keys are present on our definition type
+    if (!(key in logDefinitions[type].fields))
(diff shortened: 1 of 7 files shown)


ui/raidboss/timeline_parser.ts
```diff
@@ -461,6 +454,168 @@ export class TimelineParser {
...
+        lineNumber: lineNumber,
+        line: originalLine,
+        error: 'Invalid NetRegex arguments',
+      });
+      return line;
+    }
+    return this.buildRegexSync(
+      uniqueid,
+      `${netRegexType} ${syncCommand.netRegex}`,
+      Regexes.parse(this.GetReplacedSync(buildNetRegexForTrigger(netRegexType, params))),
```
quisquous, line 510: Now that we have parameters, can this use `translateRegexBuildParam` instead of `GetReplacedSync` here like `popup-text.ts` does? The main goal is that it knows which fields need to be translated rather than trying to translate the entire regex in one go.

I think you also might want to store the parameter object on the sync itself so that `find_missing_timeline_translations.ts` can just use that directly along with the return value from `translateRegexBuildParam` to know if a sync is fully translated. This would also let the `Translate` function in `timeline_parser.ts` use the same logic. (Sorry this is in a bunch of places.)

quisquous (commented):
Thank you for this! Overall, I think the implementation and refactoring looks great. I have a couple of comments about some edge cases around translations and tests, but that's it.

quisquous:
Also cc @JLGarber here too for timeline thoughts

ui/raidboss/timeline_parser.ts
```diff
@@ -1,11 +1,58 @@
...
+
+const isValidNetParams = <T extends LogDefinitionTypes>(
+  type: T,
+  params: Record<string, unknown>,
+): params is NetParams[T] => {
+  for (const key in params) {
+    // Make sure all keys are present on our definition type
+    if (!(key in logDefinitions[type].fields))
+      return false;
+    // Make sure our value is either a string/int or an array of strings/ints
```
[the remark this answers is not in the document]
valarnin: I included `number` because your original examples in the linked issue implied that `number` was desired, but maybe I misunderstood?

ui/raidboss/timeline_parser.ts
valarnin: Yeah, it being so spread out is why I didn't include it in this PR. I wanted to keep this as limited in scope as possible while not having the code be a mess.

ui/raidboss/timeline_parser.ts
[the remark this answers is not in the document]
quisquous: Looking at it, I think that was just sloppy example code. Those should be strings, for sure.

ui/raidboss/timeline_parser.ts
quisquous: Ok, sounds good! Can you leave a TODO at least in these places that it should translate parameters and not regexes ideally?

valarnin: I think that `TODO` comment works?

ui/raidboss/timeline_parser.ts
```diff
@@ -1,11 +1,52 @@
...
+
+const isStringOrStringArray = (value: unknown): value is number[] | string[] => {
+  if (Array.isArray(value)) {
+    if (isStringArray(value))
+      return true;
+    return false;
+  } else if (!['string', 'number'].includes(typeof value))
+    return false;
+  return true;
+};
```
quisquous, line 28:
```suggestion
const isStringOrStringArray = (value: unknown): value is string | string[] => {
  if (Array.isArray(value)) {
    if (isStringArray(value))
      return true;
    return false;
  } 
  return typeof value === 'string';
};
```

Is this what you meant here?

ui/raidboss/timeline_parser.ts
quisquous: Looks good to me, thanks!

quisquous:
Thanks!

I think I want to give this a little bit more testing and I want to give this a little bit of time for feedback from anybody else too before merging it, but overall this looks good to me and I think we should be able to auto-convert most things to this format.

JLGarber:
Nothing to add here, this looks great. I'll look at updating `make_timeline` at some point in the future.

quisquous:
A bunch of comments here.

Sorry about the conflict, it's due to the moved regex block and is just from #5940

Am I right in understanding the code that we're matching all net/parsed log lines against all net/parsed regexes? Do you think it'd be worth separating out "this is a netregex" and "this is a regex" and only matching particular lines against particular ones? It'd certainly prevent duplicating the amount of matching work going on.

One question I have is if you think it'd be worth supporting bareword keys (via an eval -> json stringify -> json parse), e.g. `Ability { id: "8A03", source: "Eulogia" }` which I find a little easier on the eyes but maybe that's just me having read too much javascript. Maybe that's a little too hacky. I think an eval is definitely a little bit sketchy too, but I think we're already evaling trigger and user code so it's not any worse.

Finally, I ran into some timeline test issues. I did a search and replace on `sync / 1\[56\]:\[\^:\]\*:([^:]*):([^:]*):/` for `Ability { "id": "$2", "source": "$1" }`. I didn't do any further testing, but timeline test does not work here with this with either an fflogs or local file. All replaced lines are missed by the timeline test. Repro steps are:

(1) do that replace
(2) add a large window sync to the 14 line, e.g. `4009.6 "--sync--" sync / 14:[^:]*:Eulogia:8A03:/ window 10000,10`
(3) run `ts-node util/logtools/test_timeline.ts -t thaleia -k YOURKEYHERE -r wRA1yPgGx8McNjnf -rf 7`

If I do add a `0 "--sync--" Ability { "id": "1D6D" } window 10000,10000` to the test timeline for provoke, that does start the timeline, so it's not a fundamental issue with non-gamelog lines in the timeline code and so I suspect that test_timeline doesn't know anything about netregex log lines.

valarnin:
> Sorry about the conflict, it's due to the moved regex block and is just from #5940

That's fine, I'll resolve it like I normally do (with a rebase/force-push) before I continue any other changes on this PR.

> Am I right in understanding the code that we're matching all net/parsed log lines against all net/parsed regexes? Do you think it'd be worth separating out "this is a netregex" and "this is a regex" and only matching particular lines against particular ones? It'd certainly prevent duplicating the amount of matching work going on.

I'm assuming by this comment you're referring to this code, which just redirects `NetLog` to `LogEvent`?

https://github.com/valarnin/cactbot/blob/c8ba9bf79d4266c91b34b3cecd52fb3fa07abdf5/ui/raidboss/timeline.ts#L726-L733

I went for simplicity on this PR, rather than also trying to de-duplicate logic in `OnLogEvent`, as that would also entail adding additional support for `netRegex` equivalents of the regexes being checked there, as well as storing netRegex checks differently, etc. That would likely have entailed adding another 200+ lines of code to this PR, as a quick rough estimate.

It probably makes sense to make that a high-priority followup PR maybe, unless you're fine with adding a bunch of extra code to this PR?

> One question I have is if you think it'd be worth supporting bareword keys (via an eval -> json stringify -> json parse), e.g. `Ability { id: "8A03", source: "Eulogia" }` which I find a little easier on the eyes but maybe that's just me having read too much javascript. Maybe that's a little too hacky. I think an eval is definitely a little bit sketchy too, but I think we're already evaling trigger and user code so it's not any worse.

I'm fine with handling this either way, it makes no difference to me, though a comment detailing the "why" for that sort of process would be required I think.

> Finally, I ran into some timeline test issues. I did a search and replace on `sync / 1\[56\]:\[\^:\]\*:([^:]*):([^:]*):/` for `Ability { "id": "$2", "source": "$1" }`. I didn't do any further testing, but timeline test does not work here with this with either an fflogs or local file. All replaced lines are missed by the timeline test. Repro steps are:
> 
> (1) do that replace (2) add a large window sync to the 14 line, e.g. `4009.6 "--sync--" sync / 14:[^:]*:Eulogia:8A03:/ window 10000,10` (3) run `ts-node util/logtools/test_timeline.ts -t thaleia -k YOURKEYHERE -r wRA1yPgGx8McNjnf -rf 7`
> 
> If I do add a `0 "--sync--" Ability { "id": "1D6D" } window 10000,10000` to the test timeline for provoke, that does start the timeline, so it's not a fundamental issue with non-gamelog lines in the timeline code and so I suspect that test_timeline doesn't know anything about netregex log lines.

I'll investigate this. I didn't think to touch `test_timeline` because the actual unit tests didn't fail and I didn't have any issues with in-game tests and `raidemulator` tests.

quisquous:
> > Am I right in understanding the code that we're matching all net/parsed log lines against all net/parsed regexes? Do you think it'd be worth separating out "this is a netregex" and "this is a regex" and only matching particular lines against particular ones? It'd certainly prevent duplicating the amount of matching work going on.
> 
> I'm assuming by this comment you're referring to this code, which just redirects `NetLog` to `LogEvent`?
> 
> https://github.com/valarnin/cactbot/blob/c8ba9bf79d4266c91b34b3cecd52fb3fa07abdf5/ui/raidboss/timeline.ts#L726-L733

Yeah, that's exactly it.

> I went for simplicity on this PR, rather than also trying to de-duplicate logic in `OnLogEvent`, as that would also entail adding additional support for `netRegex` equivalents of the regexes being checked there, as well as storing netRegex checks differently, etc. That would likely have entailed adding another 200+ lines of code to this PR, as a quick rough estimate.
> 
> It probably makes sense to make that a high-priority followup PR maybe, unless you're fine with adding a bunch of extra code to this PR?

I'm fine with extra code in this PR. I don't think you need to deduplicate `OnLogEvent`, though. My understanding here is that for backwards compatibility in some medium term sense, we'll need to send both types of log lines to the timeline. If that's true, then I think moving the code in `OnLogEvent` that's manually checking countdowns etc to `netRegex` could be done in a followup and a TODO could be left here.

What I'm really asking for is having a separate `OnParsedLog` / `OnNetLog` sort of path into `TimelineController` and into `Timeline` itself, and then a separate set of `activeNetSyncs` vs `activeSyncs` to check.

quisquous:
> > One question I have is if you think it'd be worth supporting bareword keys (via an eval -> json stringify -> json parse), e.g. `Ability { id: "8A03", source: "Eulogia" }` which I find a little easier on the eyes but maybe that's just me having read too much javascript. Maybe that's a little too hacky. I think an eval is definitely a little bit sketchy too, but I think we're already evaling trigger and user code so it's not any worse.
> 
> I'm fine with handling this either way, it makes no difference to me, though a comment detailing the "why" for that sort of process would be required I think.

Also as a clarification, this is more of a style question than an implementation one (maybe there's a better way). This definitely is more of a follow-up sort of fix either way.

I think really it's test_timeline and separating out the log handling that are the two pieces that I want to land with this PR.

valarnin:
Rebased and force-pushed.

I think this addresses all the outstanding issues, but please verify because I did this while juggling 4 other things.

There were two issues in test_timeline, the fixes should be pretty obvious.

Added more TODOs to address what was discussed here.

Opted for a more simple splitting of syncs rather than a more thorough one for simplicity. Maybe someday I'll rewrite the entire Timeline codebase so it's less... dated.

quisquous (commented):
Thanks for all these changes. Looks good to me, and `test_timeline.ts` seems to work with both a file and fflogs with these fixes.

quisquous:
> Opted for a more simple splitting of syncs rather than a more thorough one for simplicity. Maybe someday I'll rewrite the entire Timeline codebase so it's less... dated.

Yeahhhhh, there's still a good bit of 2017/2018 code in there for sure.

PR #5944, opened 2023-11-23
raidboss: Lunar Subterrane--Add Ruinous Hex calls

JLGarber:
This would probably be a lot simpler with matrix math, but I don't know matrices enough to implement them here, so this is what we get instead. It should be largely self-explanatory with the in-line comments, but please criticize away, since this isn't necessarily the nicest implementation.

### DIFF
--- a/ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
+++ b/ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
@@ -193,7 +193,7 @@ const triggerSet: TriggerSet<Data> = {
           en: 'Inner northeast safe',
         },
         northwest: {
-          en: 'Inner northwest safe'
+          en: 'Inner northwest safe',
         },
         southeast: {
           en: 'Inner southeast safe',
@@ -547,6 +547,7 @@ const triggerSet: TriggerSet<Data> = {
     },
     {
       'locale': 'ja',
+      'missingTranslations': true,
       'replaceSync': {
         'Aetheric Charge': '魔力球',
         'Damcyan Antlion': 'ダムシアン・アントリオン',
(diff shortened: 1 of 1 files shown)


ui/raidboss/data/06-ew/dungeon/the_lunar_subterrane.ts
```diff
@@ -74,6 +96,114 @@ const triggerSet: TriggerSet<Data> = {
...
+        northwest: {
+          en: 'Inner northwest safe',
+        },
+        southeast: {
+          en: 'Inner southeast safe',
+        },
+        southwest: {
+          en: 'Inner southwest safe',
+        },
+        unknown: '???',
```
quisquous, line 204:
```suggestion
        unknown: Outputs.unknown,
```

quisquous (commented):
> This would probably be a lot simpler with matrix math, but I don't know matrices enough to implement them here, so this is what we get instead. It should be largely self-explanatory with the in-line comments, but please criticize away, since this isn't necessarily the nicest implementation.

I don't know what you mean by matrix math. 

I think my only concern here is that this seems a little fragile, but if you feel convinced from your data, then I'm fine to merge it.

If you wanted an optional alternate implementation suggestion, then I'd suggest something like this approach:

```typescript
let safeX = [0, 1, 2, 3];
let safeY = [0, 1, 2, 3];

for (const staff of staffs) {
  // map staff positions into 0-3 row/col coordinates
  const col = Math.round((staff.x - centerX) / something + something);
  const row = Math.round((staff.y - centerY) / something + something);
  
  safeX = safeX.filter((x) => x !== col);
  safeY = safeY.filter((y) => y !== row);
}

const [finalX] = safeX;
const [finalY] = safeY;
if (finalX == undefined || finalY === undefined || safeX.length !== 1 || safeY.length !== 1)
  return;

// use safeX and safe Y to return a square
```

JLGarber:
Okay, thank you, that's a much better approach. I know the implementation I'm using here involves comparisons against numeric positions, but I think it's safe enough given that we know these positions are accurate, and that we're rounding everything. This worked perfectly in testing, and it should handle any possible staff setup regardless of whether we have seen it before.

quisquous:
Thanks! Looks good to me!

PR #5947, opened 2023-11-24
util: Add basic network log seal support in timeline tools

JLGarber:
This wasn't as simple as I had hoped, but it does seem to work as it is. There's still a bit of fragility in `make_timeline` with this implementation, since it still relies on the seal name being present in game log form. We could change this to make "name from game log" optional in the `assembleTimelineStrings()` function, but that would involve some complex conditions to allow for use of either the network seal or the game log seal.

Alternatively, we could simply not care about game log seals at all, except to check "if there's a game log seal here, use the name it contains". Maybe something like this for the end of `assembleTimelineStrings()`?

```javascript
  if (fight !== undefined && fight.sealId !== undefined) {
    if (fight.sealName !== undefined) {
      const sealMessage = SFuncs.toProperCase(fight.sealName);
      const sealComment = `# ${sealMessage} will be sealed off`;
      assembled.push(sealComment);
    }
    const netLogSeal = `0.0 "--sync--" sync / 29:[^:]*:7DC:[^:]*:${fight.sealId}/ window 0,1`;
    assembled.push(netLogSeal);
  } else {
    assembled.push('0.0 "--sync--" sync / 104:[^:]*:1($|:)/ window 0,1');
  }
  ```

Ultimately it would be nice to have the seal names dumped to a resource file as we do with `ZoneId` and `ZoneName`, and then just look them up here, but I'm not feeling up to CSV stuff this weekend.

### DIFF
--- a/util/logtools/encounter_tools.ts
+++ b/util/logtools/encounter_tools.ts
@@ -24,6 +24,7 @@ type FightEncInfo = ZoneEncInfo & {
   fightName?: string;
   endType?: string;
   sealName?: string;
+  sealId?: string;
   logLines?: string[];
 };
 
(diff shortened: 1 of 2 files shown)


valarnin:
Why not just pull the names from `PlaceName.csv` like my convert script does?

JLGarber:
~~I didn't read the script because I wasn't going to be using it~~ That would be sensible, yes. I think we might still wish to support game log seals just so that we aren't dependent on XIVAPI to update, but you're right, for any other content that is by far the better way to handle it. I'll see what I can do.

quisquous:
I think this is working as-is and so I'm going to merge it and start adjusting existing timelines over to this format.

Looking up `PlaceName` is a good idea, but can be done as a follow-up.

PR #5953, opened 2023-11-28
ci: do release based on version not tag

quisquous:
This allows releases to be done via PR and not via git push.

Notes:
* .npmrc combines tag+commit logic and so if you turn off tagging the commit has to be done manually; this is now in bump_version.ts
* this largely copies from OverlayPlugin's release.yml


quisquous:
I'll see how this goes and revert if I broke stuff.

PR #5976, opened 2023-12-02
i18n: another mount rokkon cn translation

lsl1225:


lsl1225:
That's weird, seems no issue on my local machine while syncing files.

lsl1225:
> > That's weird, seems no issue on my local machine while syncing files.
> 
> The bot complaint is that you haven't run `ts-node util/sync_files.ts` to move your changes into another mount rokkon _savage_ automatically. If you run that and then commit that file, hopefully it should work.

Oh, I see. Thanks!

PR #5990, opened 2023-12-07
build: add back loader parameter to node commands

### LINKED ISSUE #5910: Node 20 requires loader parameter again
See conversation thread here: https://github.com/quisquous/cactbot/issues/5859#issuecomment-1794648537

With Node 18 this works, but with Node 20 it seems to revert to previous behavior.

### LINKED ISSUE #5989: NodeJS LTS 20 Support
NodeJS 18 LTS ended active support over a month ago although it still is receiving security updates. Node is pushing 20 LTS now on their website and on Windows, the winget package manager is actively trying to upgrade 18 to 20.

I propose adding support for the current mainstream v20 LTS.

When trying to set things up with 20, you get the following:
```
$ npm install
npm ERR! code EBADENGINE
npm ERR! engine Unsupported engine
npm ERR! engine Not compatible with your version of node/npm: cactbot@0.30.9
npm ERR! notsup Not compatible with your version of node/npm: cactbot@0.30.9
npm ERR! notsup Required: {"node":">=18.10.0 <=18.18.2"}
npm ERR! notsup Actual:   {"npm":"10.2.3","node":"v20.10.0"}
```

quisquous:
Instead of:
  `ts-node util/sync_files.ts`
Now we must do:
  `node --loader=ts-node/esm util/sync_files.ts`

Additionally, as `--loader` is deprecated, an error message suggests to do in JavaScript, so add a loader script `--import ./loader.js` to do this work instead

Fixes #5910.
Fixes #5989.


quisquous:
@xiashtra why the confused emoji?

quisquous:
An alternative to this PR is to switch to using `--loader` instead of `--import` which will work on node 18 and node 20. However, every script run will spit out this blob of text about how we should be using `--import` instead.

```
$ npm run test

> cactbot@0.30.9 test
> mocha

(node:19160) ExperimentalWarning: `--experimental-loader` may be removed in the futur
e; instead use `register()`:
--import 'data:text/javascript,import { register } from "node:module"; import { pathT
oFileURL } from "node:url"; register("ts-node/esm", pathToFileURL("./"));'

  [▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬▬..............]
```

xiashtra:
> @xiashtra why the confused emoji?

Nothing you are responsible for.  I don't follow Node's development, so I'm just confused why the upstream change.  It seems like a lot of tedious boilerplate just to run the utility scripts, and on top of that, if it's considered deprecated, what is the intended long-term solution?

quisquous:
The `--loader` parameter is deprecated and prints out an error. The way to get around this is to add a file (e.g. `./loader.js`) with these contents:

```javascript
import { register } from 'node:module';
import url from 'node:url';

const __filename = url.fileURLToPath(import.meta.url);
register('ts-node/esm', url.pathToFileURL(__filename));
```

and then instead of saying `node --loader=ts-node/esm util/sync_files.ts` you write `node --import ./loader.js util/sync_files.ts`

However, this doesn't work on node 18, and so we should only switch to it in the future once people stop using node 18.

See: a0679c91f7efb795a583146b8e7e4a318cc248bd

PR #5993, opened 2023-12-07
raidboss: timeline netregex for arr ability

quisquous:


quisquous:
I double checked all of these, and also verified with Shiva Ex that timelines are working in game (as there's a fork in the timeline immediately).

PR #6060, opened 2023-12-21
raidboss/util: redirect all html to op/cactbot

quisquous:


quisquous:
This is not the fanciest, as presumably we could write a script to preserve all url params, but it should do the trick for anybody who has bookmarked the `quisquo.us.github.io` versions.

PR #6070, opened 2024-01-09
ci: remove dependabot, hopefully disabling it

quisquous:


MaikoTan:
Just FYI, delete the `.dependabot.yml` would only disable the dependabot version update, but the dependabot security alert should be manually disabled, you could refer to [this page](https://github.com/quisquous/cactbot/settings/security_analysis) then click `Disable` button in dependabot alert section:

image

------

Also you could just [archive the repo](https://docs.github.com/en/repositories/archiving-a-github-repository/archiving-repositories)

quisquous:
Alerts were already turned off, I think? (They're certainly turned off now.)

Re: archiving. I thought about that, but I wasn't sure if it would prevent transferring issues. There's some stuff that's probably still worth doing in the issues, and valarnin (at least) should be able to transfer issues from here to the new repo as needed so that it keeps all the conversation and context.

